PR #1, opened 2015-04-21
Update c++ style guidelines with TOC links working

tammoippen:
Further describe that we require clang-format v3.6 and how to install it


jougs:
All fine.

PR #4, opened 2015-05-12
removed kwarg alternative from kstest, since default option was used

hannahbos:


jougs:
See #5 for the corresponding issue.

Issue #8, opened 2015-05-18
configure.ac is missing from examples/MyModule


heplesser:
File `configure.ac` was not imported into GitHub with the rest of `examples/MyModule`. Thus, MyModule cannot be built.

heplesser:
This most likely happened because `.gitignore` excludes `configure.ac`.  This problem disappears once the `fix-956-libtool` patch is merged. I for now force `configure.ac` addition.

PR #7, opened 2015-05-15
Fix  no 956 (old trac): libtool 2.4.3 or later breaks build on OS X and related configure clean-up

heplesser:
My very first attempt at a pull request ... nervous ;).

This pull request cleans up several issues related to NEST builds:

- It fixes #956 from the old trac: NEST can now be built with current versions of libtool on OS X
- The libtool configuration in `configure.ac` is modernized
- `configure.ac.in` is removed, as is patching for libtool < v2 and automake < 1.14; `bootstrap.sh` now issues an error message if those programs are outdated
- `install_autotools.sh` is updated to the latest versions
- In `configure.ac`, the `AC_PREREQ` is now `2.63`.  This is pretty defensive, as that version was released Sept 2008

Tested successfully on most systems I have access to, including a BlueGene/Q, OS X 10.10.3, a recent Debian VM (jennie) and Ubuntu VM (14.04). On some systems with old Linuxes, I had to install current Autotools.

PR #9, opened 2015-05-18
Fix mymodule build fail

heplesser:
- Added missing `examples/MyModule/configure.ac`, fixing #8 .
- Added missing `nest` namespace qualifiers fixing a problem reported by Janroel Koppen (NEST User, May 2015)
- Fixed minor formatting glitches for cross-references in MyModule code.

### DIFF
--- a/examples/MyModule/configure.ac
+++ b/examples/MyModule/configure.ac
@@ -1,0 +1,190 @@
+AC_PREREQ(2.52)
+AC_INIT([mymodule], [1.0], [nest_user@nest-initiative.org])
+
+# Exporting source and build directories requires full path names.
+# Thus we have to expand.
+# Here, we are in top build dir, since source dir must exist, we can just
+# move there and call pwd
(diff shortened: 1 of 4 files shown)


examples/MyModule/configure.ac
```diff
@@ -0,0 +1,190 @@
+AC_PREREQ(2.52)
```
tammoippen: In configure.ac we require 2.63. I think we should stick to one version across NEST?

heplesser: Yes, we should absolutely. To keep things simple, though, I fixed in this PR only those things that absolutely needed fixing to make things work again and solve Janroel's problem. I thought to leave the remaining necessary fixes to this file here until PR #7 is accepted. We should then update this file to libtool initialisation etc as in the newest version of the main configure.ac.

jougs:
I'm happy with this as well and agree that we should stick with a single required version.

PR #11, opened 2015-05-21
First Structural Plasticity Pull request

sdiazpier:
These changes include the first step in order to enable structural plasticity in NEST. The implementation corresponds to the formalism described in: M. Butz and A. van Ooyen. A simple rule for dendritic spine and axonal bouton formation can account for cortical reorganization after focal retinal lesions. PLoS Comput. Biol., 9(10):e1003259, Oct 2013. 

This Model of Structural Plasticity (MSP) considers synapses to be not just a weighted link between two neurons but as consisting of two connected parts, a presynaptic axonal element and a postsynaptic dendritic element. Matching synaptic elements develop independently from each other, only depending on the level of activity of the hosting neuron. 
Once new synaptic elements are formed, they are first considered as vacant. From that moment on, they can be connected with a matching counterpart to form a new synapse in a random process. 
Synapses can break apart when a bound synaptic element is deleted. The remaining and now freed synaptic element of the broken synapse can then rebind with a different synaptic element. 
Breaking of synapses and rebinding of synaptic elements causes the rewiring of the network until a homoeostatic point is reached.

This first request includes only the implementation of classes and data structures to represent synaptic elements and their dynamics within the network. No changes in the connectivity are present at this point.

### DIFF
--- a/nestkernel/Makefile.am
+++ b/nestkernel/Makefile.am
@@ -78,8 +78,8 @@ libnest_la_SOURCES=\
 		conn_parameter.h conn_parameter.cpp\
 		conn_builder.h conn_builder.cpp\
 		conn_builder_factory.h\
-		music_event_handler.h music_event_handler.cpp
-
+		music_event_handler.h music_event_handler.cpp\
+                synaptic_element.h synaptic_element.cpp
 libnest_la_LIBADD= @LIBLTDL@ @LIBADD_DL@
 
 EXTRA_DIST= README.md
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -28,7 +28,9 @@
  */
 
 #include "archiving_node.h"
+#include "synaptic_element.h"
 #include "dictutils.h"
+#include "network.h"
 
 namespace nest
 {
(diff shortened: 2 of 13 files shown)


naveau:
Hi Sandra,
Travis does not complain but there are still some tests that fail. Here is what I found in the logs:
- something is wrong with the headers of the files (please have a look at the report of the phase 4 generated by the tests)
- The python testsuite didn't run because python can't import msp_test
Can you have a look at the tests report locally? I don't know if we can get the report folder directly from Travis.

sdiazpier:
Hi Mikael, 
Thank you for letting me know of this. I think the problem with the headers is not related to MSP, but the one running the python testsuite definitely is. I will solve it right away.

lekshmideepu:
Hi Mikael,

    The header issue is for a file which is not related to MSP. The file is related to pip upgrade.
get-pip.py. But it has been fixed in my pull request which will be merged to the main.

apeyser:
@lekshmideepu fix #16 is now at the head of master. If that's needed, rebase your work on that.

sdiazpier:
I have rebased on top of Lekshmi's fix and made the corrections for the python tests. I will close this pull request and create a new one.

PR #15, opened 2015-05-22
Lekshmideep master

apeyser:


apeyser:
While we're still investigating the S3 deployment, I'll split this and report under #14

apeyser:
This will be closed for now, and a pull request will be put in when the s3 deployment has been figured out.

PR #12, opened 2015-05-21
Fix for the copyright check under regression tests and a minor change in the email notification section of travis

lekshmideepu:
Error:  'regressiontests/ticket-659-copyright.py'... Failed: inconsistent copyright header(s) in get-pip.py.
Fix: Added get-pip.py in the exclude section of ticket-659-copyright.py

By using default notification in .travis.yml file, email notification will be sent to all the people who has admin rights to the github repository.

Issue #13, opened 2015-05-22
Turn off notification noise


apeyser:
The current travis settings are noisy. By deleting the notification section, the default scheme is used, which is that messages only go to the submitter. Given the pull system, there's no need for everyone to get every build error, as in the SVN/Jenkins setup.

apeyser:
Since this was also part of the pull request @lekshmideepu #12, I've merged this, and then rebased & split #12 into #15 and #16.

PR #16, opened 2015-05-22
Added get-pip.py in the exclude section of regressiontests/ticket-659…

apeyser:
This is part 2 of #12. Part 1 is #14, and part 3 is #15


apeyser:
And this is also associated with issue #13

PR #26, opened 2015-06-09
Xcode Workflow article

tammoippen:
In this article I describe how to set-up NEST on a Mac in such a way, that you can use Xcode to do the actual coding, code completion and debugging.


tammoippen:
Homebrew tries to use precompiled packages, if possible (so-called bottles). You can disable this by setting the environment variable HOMEBREW_BUILD_FROM_SOURCE to some value. This line works for me:

    HOMEBREW_BUILD_FROM_SOURCE=1 brew install open-mpi

jougs:
This looks like a useful addition to the documentation to me. I did not test it, though. +1 for merging.

flinz:
@tammoippen I could follow this without problems!

I am not very familiar with xcode - would you have hints on how to debug (e.g. breakpoints etc) the C++ codebase in Xcode when running it via the python API? 

Or more generally: What is your typical workflow for debugging Nest? I am, sadly, mostly still resorting to ``cout``s up until now (blush).

tammoippen:
@flinz I ran into problems with debugging PyNEST as well and was never able to do it properly. Usually I write short `SLI` scripts and debug them via XCode:

* In `Products->Scheme->Manage Schemes...->Edit...->Run` you can edit your scheme: change the Executable (should point to `nest` in `<prefix>/bin`) and edit the CL arguments, e.g. the SLI script.
* Compile with `-O0 -g`.
* I find MPI difficult to debug this way, but threading works.

PR #27, opened 2015-06-12
Modify .gitignore file so that project files created by kdevelop are …

wschenck:
…ignored as well

This is just a trivial addition to the .gitignore file to make life easier for people working with kdevelop.


tammoippen:
Looks reasonable. +1 for merging.

PR #17, opened 2015-05-22
First Structural Plasticity Pull request, corrected

sdiazpier:
These changes include the first step in order to enable structural plasticity in NEST. The implementation corresponds to the formalism described in: M. Butz and A. van Ooyen. A simple rule for dendritic spine and axonal bouton formation can account for cortical reorganization after focal retinal lesions. PLoS Comput. Biol., 9(10):e1003259, Oct 2013.

This Model of Structural Plasticity (MSP) considers synapses to be not just a weighted link between two neurons but as consisting of two connected parts, a presynaptic axonal element and a postsynaptic dendritic element. Matching synaptic elements develop independently from each other, only depending on the level of activity of the hosting neuron.
Once new synaptic elements are formed, they are first considered as vacant. From that moment on, they can be connected with a matching counterpart to form a new synapse in a random process.
Synapses can break apart when a bound synaptic element is deleted. The remaining and now freed synaptic element of the broken synapse can then rebind with a different synaptic element.
Breaking of synapses and rebinding of synaptic elements causes the rewiring of the network until a homoeostatic point is reached.

This first request includes only the implementation of classes and data structures to represent synaptic elements and their dynamics within the network. No changes in the connectivity are present at this point.

PR #32, opened 2015-06-16
First Structural Plasticity Pull request, corrections after first code review

sdiazpier:
These changes include the first step in order to enable structural plasticity in NEST. Follow up to pull request #17

### DIFF
--- a/nestkernel/Makefile.am
+++ b/nestkernel/Makefile.am
@@ -79,7 +79,9 @@ libnest_la_SOURCES=\
 		conn_builder.h conn_builder.cpp\
 		conn_builder_factory.h\
 		music_event_handler.h music_event_handler.cpp\
-		synaptic_element.h synaptic_element.cpp
+		synaptic_element.h synaptic_element.cpp\
+		growth_curve.h growth_curve.cpp\
+		growth_curve_factory.h
 libnest_la_LIBADD= @LIBLTDL@ @LIBADD_DL@
 
 EXTRA_DIST= README.md
--- a/nestkernel/Makefile.am
+++ b/nestkernel/Makefile.am
@@ -79,7 +79,7 @@ libnest_la_SOURCES=\
 		conn_builder.h conn_builder.cpp\
 		conn_builder_factory.h\
 		music_event_handler.h music_event_handler.cpp\
-                synaptic_element.h synaptic_element.cpp
+		synaptic_element.h synaptic_element.cpp
 libnest_la_LIBADD= @LIBLTDL@ @LIBADD_DL@
 
 EXTRA_DIST= README.md
--- a/nestkernel/Makefile.am
+++ b/nestkernel/Makefile.am
@@ -78,8 +78,8 @@ libnest_la_SOURCES=\
 		conn_parameter.h conn_parameter.cpp\
 		conn_builder.h conn_builder.cpp\
 		conn_builder_factory.h\
-		music_event_handler.h music_event_handler.cpp
-
+		music_event_handler.h music_event_handler.cpp\
+                synaptic_element.h synaptic_element.cpp
 libnest_la_LIBADD= @LIBLTDL@ @LIBADD_DL@
 
 EXTRA_DIST= README.md
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -212,7 +212,7 @@ nest::Archiving_Node::get_status( DictionaryDatum& d ) const
 #endif
 
   synaptic_elements_d = DictionaryDatum( new Dictionary );
-  def< DictionaryDatum >( d, "synaptic_elements", synaptic_elements_d );
+  def< DictionaryDatum >( d, names::synaptic_elements, synaptic_elements_d );
   for ( std::map< Name, SynapticElement >::const_iterator it = synaptic_elements_map_.begin();
         it != synaptic_elements_map_.end();
         ++it )
@@ -266,7 +266,7 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
     clear_history();
   }
 
-  if ( not d->known( "synaptic_elements" ) )
+  if ( not d->known( names::synaptic_elements ) )
   {
     return;
   }
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -82,7 +82,7 @@ Archiving_Node::register_stdp_connection( double_t t_first_read )
   {
     ( runner->access_counter_ )++;
   }
-  
+
   n_incoming_++;
 }
 
@@ -240,7 +240,7 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
   {
     throw BadProperty( "All time constants must be strictly positive." );
   }
-  
+
   tau_minus_ = new_tau_minus;
   tau_minus_triplet_ = new_tau_minus_triplet;
 
@@ -265,7 +265,7 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
   {
     clear_history();
   }
-  
+
   if ( not d->known( "synaptic_elements" ) )
   {
     return;
@@ -283,7 +283,8 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
   {
     insert_result = synaptic_elements_map_.insert(
       std::pair< Name, SynapticElement >( i->first, SynapticElement() ) );
-    ( insert_result.first->second ).set( getValue< DictionaryDatum >( synaptic_elements_d, i->first ) );
+    ( insert_result.first->second )
+      .set( getValue< DictionaryDatum >( synaptic_elements_d, i->first ) );
   }
 }
 
@@ -376,7 +377,7 @@ nest::Archiving_Node::get_synaptic_elements() const
 void
 nest::Archiving_Node::update_synaptic_elements( double_t t )
 {
-  assert ( t >= Ca_t_ );
+  assert( t >= Ca_t_ );
 
   for ( std::map< Name, SynapticElement >::iterator it = synaptic_elements_map_.begin();
         it != synaptic_elements_map_.end();
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -166,7 +166,7 @@ nest::Archiving_Node::get_history( double_t t1,
 void
 nest::Archiving_Node::set_spiketime( Time const& t_sp )
 {
-  update_synaptic_element( t_sp.get_ms() );
+  update_synaptic_elements( t_sp.get_ms() );
   Ca_minus_ += beta_Ca_;
 
   if ( n_incoming_ )
@@ -272,7 +272,6 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
   }
   // we replace the existing synaptic_elements_map_ by the new one
   DictionaryDatum synaptic_elements_d;
-  //DictionaryDatum synaptic_element_d;
   std::pair< std::map< Name, SynapticElement >::iterator, bool > insert_result;
 
   synaptic_elements_map_ = std::map< Name, SynapticElement >();
@@ -282,10 +281,9 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
         i != synaptic_elements_d->end();
         ++i )
   {
-    //synaptic_element_d = getValue< DictionaryDatum >( synaptic_elements_d, i->first );
     insert_result = synaptic_elements_map_.insert(
       std::pair< Name, SynapticElement >( i->first, SynapticElement() ) );
-    ( insert_result.first->second ).set( getValue< DictionaryDatum >( synaptic_elements_d, i->first ) );/*synaptic_element_d );*/
+    ( insert_result.first->second ).set( getValue< DictionaryDatum >( synaptic_elements_d, i->first ) );
   }
 }
 
@@ -305,7 +303,7 @@ nest::Archiving_Node::clear_history()
 * Get the number of synaptic_elements
 * ---------------------------------------------------------------- */
 double_t
-nest::Archiving_Node::get_synaptic_element( Name n ) const
+nest::Archiving_Node::get_synaptic_elements( Name n ) const
 {
   std::map< Name, SynapticElement >::const_iterator se_it;
   se_it = synaptic_elements_map_.find( n );
@@ -319,7 +317,9 @@ nest::Archiving_Node::get_synaptic_element( Name n ) const
       return z_value;
     }
     else
+    {
       return std::floor( z_value );
+    }
   }
   else
   {
@@ -328,7 +328,7 @@ nest::Archiving_Node::get_synaptic_element( Name n ) const
 }
 
 int_t
-nest::Archiving_Node::get_synaptic_element_vacant( Name n ) const
+nest::Archiving_Node::get_synaptic_elements_vacant( Name n ) const
 {
   std::map< Name, SynapticElement >::const_iterator se_it;
   se_it = synaptic_elements_map_.find( n );
@@ -344,7 +344,7 @@ nest::Archiving_Node::get_synaptic_element_vacant( Name n ) const
 }
 
 int_t
-nest::Archiving_Node::get_synaptic_element_connected( Name n ) const
+nest::Archiving_Node::get_synaptic_elements_connected( Name n ) const
 {
   std::map< Name, SynapticElement >::const_iterator se_it;
   se_it = synaptic_elements_map_.find( n );
@@ -368,18 +368,15 @@ nest::Archiving_Node::get_synaptic_elements() const
         it != synaptic_elements_map_.end();
         ++it )
   {
-    n_map.insert( std::pair< Name, double_t >( it->first, get_synaptic_element( it->first ) ) );
+    n_map.insert( std::pair< Name, double_t >( it->first, get_synaptic_elements( it->first ) ) );
   }
   return n_map;
 }
 
 void
-nest::Archiving_Node::update_synaptic_element( double_t t )
+nest::Archiving_Node::update_synaptic_elements( double_t t )
 {
-  if ( t < Ca_t_ )
-  {
-    throw KernelException( "Synaptic elements already updated" );
-  }
+  assert ( t >= Ca_t_ );
 
   for ( std::map< Name, SynapticElement >::iterator it = synaptic_elements_map_.begin();
         it != synaptic_elements_map_.end();
@@ -393,7 +390,7 @@ nest::Archiving_Node::update_synaptic_element( double_t t )
 }
 
 void
-nest::Archiving_Node::decay_synaptic_element_vacant( double_t p )
+nest::Archiving_Node::decay_synaptic_elements_vacant( double_t p )
 {
   double_t z, z_vacant;
   for ( std::map< Name, SynapticElement >::iterator it = synaptic_elements_map_.begin();
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -28,7 +28,9 @@
  */
 
 #include "archiving_node.h"
+#include "synaptic_element.h"
 #include "dictutils.h"
+#include "network.h"
 
 namespace nest
 {
@@ -42,6 +44,11 @@ nest::Archiving_Node::Archiving_Node()
   , tau_minus_( 20.0 )
   , tau_minus_triplet_( 110.0 )
   , last_spike_( -1.0 )
+  , Ca_t_( 0.0 )
+  , Ca_minus_( 0.0 )
+  , tau_Ca_( 10000.0 )
+  , beta_Ca_( 0.001 )
+  , synaptic_elements_map_()
 {
 }
 
@@ -53,6 +60,11 @@ nest::Archiving_Node::Archiving_Node( const Archiving_Node& n )
   , tau_minus_( n.tau_minus_ )
   , tau_minus_triplet_( n.tau_minus_triplet_ )
   , last_spike_( n.last_spike_ )
+  , Ca_t_( n.Ca_t_ )
+  , Ca_minus_( n.Ca_minus_ )
+  , tau_Ca_( n.tau_Ca_ )
+  , beta_Ca_( n.beta_Ca_ )
+  , synaptic_elements_map_( n.synaptic_elements_map_ )
 {
 }
 
@@ -148,6 +160,9 @@ nest::Archiving_Node::get_history( double_t t1,
 void
 nest::Archiving_Node::set_spiketime( Time const& t_sp )
 {
+  update_synaptic_element( t_sp.get_ms() );
+  Ca_minus_ += beta_Ca_;
+
   if ( n_incoming_ )
   {
     // prune all spikes from history which are no longer needed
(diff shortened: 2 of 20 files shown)


sdiazpier:
Given the type of changes included in this branch (mainly in the nest kernel), I think @jougs, @suku248, @tammoippen and/or @wschenck would be good to give a second review on this code.
Thank you.

nestkernel/Makefile.am
```diff
@@ -78,8 +78,8 @@ libnest_la_SOURCES=\
 		conn_parameter.h conn_parameter.cpp\
 		conn_builder.h conn_builder.cpp\
 		conn_builder_factory.h\
-		music_event_handler.h music_event_handler.cpp
-
+		music_event_handler.h music_event_handler.cpp\
+                synaptic_element.h synaptic_element.cpp
```
heplesser: This line seems indented one too deep, should be aligned with lines above.

nestkernel/archiving_node.cpp
```diff
@@ -205,20 +245,64 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
...
     clear_history();
+  }
+  
+  if ( not d->known( "synaptic_elements" ) )
+  {
+    return;
+  }
+  // we replace the existing synaptic_elements_map_ by the new one
+  DictionaryDatum synaptic_elements_d;
+  //DictionaryDatum synaptic_element_d;
```
heplesser: You should remove commented-out code

nestkernel/archiving_node.cpp
```diff
@@ -205,20 +245,64 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
...
+  std::pair< std::map< Name, SynapticElement >::iterator, bool > insert_result;
+
+  synaptic_elements_map_ = std::map< Name, SynapticElement >();
+  synaptic_elements_d = getValue< DictionaryDatum >( d, "synaptic_elements" );
+
+  for ( Dictionary::const_iterator i = synaptic_elements_d->begin();
+        i != synaptic_elements_d->end();
+        ++i )
+  {
+    //synaptic_element_d = getValue< DictionaryDatum >( synaptic_elements_d, i->first );
```
heplesser: More commented-out code.

nestkernel/archiving_node.cpp
```diff
@@ -228,6 +312,126 @@ nest::Archiving_Node::clear_history()
...
+
+  if ( se_it != synaptic_elements_map_.end() )
+  {
+    z_value = ( se_it->second ).get_z();
+    if ( ( se_it->second ).continuous() )
+    {
+      return z_value;
+    }
+    else
+      return std::floor( z_value );
```
heplesser: Also this line should be surrounded by braces.

nestkernel/archiving_node.h
```diff
@@ -64,6 +65,36 @@ class Archiving_Node : public Node
    */
   Archiving_Node( const Archiving_Node& );
 
+  /**
+   * \fn double_t get_Ca_value()
+   * return the Ca_minus value
+   */
+  double_t
+  get_Ca_value() const
```
heplesser: Is this really the best method name? Generally, a getter for an attribute should be `get_<attr name>`, i.e., in this case `get_Ca_minus`. Is there any reason to do it differently?

nestkernel/archiving_node.h
```diff
@@ -154,6 +190,25 @@ class Archiving_Node : public Node
 
   // spiking history needed by stdp synapses
   std::deque< histentry > history_;
+
+  /*
+   * Structural plasticity
+   */
+
+  // Time of the last update of the Calcium concentration
+  double_t Ca_t_;
```
heplesser: Comments should state the units; for times, it is probably ms, but what is it for `Ca_minus_` and `beta_Ca_`?

nestkernel/nest_names.cpp
```diff
@@ -25,6 +25,9 @@ namespace nest
 {
 namespace names
 {
+const Name Ca( "Ca" );
```
heplesser: The names list should be kept in alphabetical order (applies also to the .h file).

nestkernel/node.h
```diff
@@ -477,6 +477,59 @@ class Node
   virtual void handle( DoubleDataEvent& e );
 
   /**
+   * return the Ca_minus value at t (in ms).
```
heplesser: I don't know the theory here, but it `Ca_minus` a time, i.e., in ms?

nestkernel/node.h
```diff
@@ -477,6 +477,59 @@ class Node
...
+   * return 0.0 if not overridden
+   */
+  virtual double_t
```
heplesser: Right now, I cannot see where all these getter functions are used. Inspection of the node should be via `get_status()`. Unless there is a very good reason to have these getters, I strongly suggest to remove them all.

nestkernel/node.h
```diff
@@ -477,6 +477,59 @@ class Node
...
+  get_Ca_value() const
+  {
+    return 0.0;
```
heplesser: Is it really sensible to return 0 for nodes not providing these methods?

nestkernel/node.h
```diff
@@ -477,6 +477,59 @@ class Node
...
+  }
+
+  /**
+   * get the number of synaptic element for the current Node at t (in ms)
+   * return 0.0 if not overridden
+   */
+  virtual double_t get_synaptic_element( Name ) const
```
heplesser: It appears that this method would return an integer. Why is the return type double?

nestkernel/node.h
```diff
@@ -477,6 +477,59 @@ class Node
...
+    return 0.0;
+  }
+
+  /**
+   * get the number of vacant synaptic element for the current Node
+   * return 0.0 if not overridden
+   */
+  virtual int_t get_synaptic_element_vacant( Name ) const
+  {
+    return 0.0;
```
heplesser: The return type is `int_t`,  so it should be `return 0;`

nestkernel/node.h
```diff
@@ -477,6 +477,59 @@ class Node
...
+  }
+
+  /**
+   * get the number of connected synaptic element for the current Node
+   * return 0.0 if not overridden
+   */
+  virtual int_t get_synaptic_element_connected( Name ) const
+  {
+    return 0.0;
```
heplesser: See above.

nestkernel/archiving_node.cpp
```diff
@@ -228,6 +312,126 @@ nest::Archiving_Node::clear_history()
...
+  {
+    n_map.insert( std::pair< Name, double_t >( it->first, get_synaptic_element( it->first ) ) );
+  }
+  return n_map;
+}
+
+void
+nest::Archiving_Node::update_synaptic_element( double_t t )
+{
+  if ( t < Ca_t_ )
```
heplesser: Could this happen in reality, or is this only a safeguard against errors elsewhere in the code? In the latter case, turn this into `assert( t >= Ca_t);`

nestkernel/archiving_node.cpp
```diff
@@ -228,6 +312,126 @@ nest::Archiving_Node::clear_history()
...
+        it != synaptic_elements_map_.end();
+        ++it )
+  {
+    n_map.insert( std::pair< Name, double_t >( it->first, get_synaptic_element( it->first ) ) );
+  }
+  return n_map;
+}
+
+void
+nest::Archiving_Node::update_synaptic_element( double_t t )
```
heplesser: Should this have `_elements` in the plural; applies also to some other methods below.

heplesser:
Hi! I added quite a number of comments throughout, but have not looked at the `synaptic_element.{h.cpp}` files yet. One thing I am concerned about is whether the changes would affect simulation speed in cases where one does not use structural plasticity. I am particularly worried about the `update_synaptic_element( t_sp.get_ms() );` call in `Archiving_Node::set_spiketime()`.

Have you benchmarked, e.g. a standard brunel model, in NEST with and without structural plasticity to see the impact on performance?

nestkernel/node.h
sdiazpier: The number of synaptic elements is stored in a double variable. An standard value for the growth of a synaptic element is around 0.0001 elements/ms. The actual usable number of synaptic elements is an integer truncated from this double value.

sdiazpier:
Hi Hans, thank you so much for reviewing our code. I will work on your comments now. We have performed tests  on the Brunel model and verified that, when MSP is not enabled, there is no impact in the performance of the simulations. @naveau has also made tests related to changes in connectivity during run time (this feature will come in a following pull request) and verified that this also does not affect the normal behaviour of NEST. Our first priority when introducing these features was to make sure that the normal performance in NEST is preserved.

yoginormandie:
@heplesser I understand your point. As @sdiazpier mentionned, we performed some simple benchmark using the standard brunel model. Here are some results we got (best of 10 simulation time):
simulation_time_vs_threads
vmpeak_vs_threads

These tests come from an earlier version of the MSP branch, but nothing changed at the level of the Archiving node in the meanwhile. You're right, the only thing that is executed for non-structural plasticity user is this `update_synaptic_element( t_sp.get_ms() );`. This method is basically a for loop for all declared synaptic elements. If there is no synaptic element specified by the user, then the method returns immediately. That should not impact too much the performance as shown by our benchmark.

naveau:
(sry for the last comment, I was log in with a different account)

@heplesser I understand your point. As @sdiazpier mentionned, we performed some simple benchmark using the standard brunel model. Here are some results we got (best of 10 simulation time):
simulation_time_vs_threads
vmpeak_vs_threads

These tests come from an earlier version of the MSP branch, but nothing changed at the level of the Archiving node in the meanwhile. You're right, the only thing that is executed for non-structural plasticity user is this `update_synaptic_element( t_sp.get_ms() );`. This method is basically a for loop for all declared synaptic elements. If there is no synaptic element specified by the user, then the method returns immediately. That should not impact too much the performance as shown by our benchmark.

naveau:
@gewaltig Thanks for your review.

Just as a side note, would it be possible to get all the comments at the same place? @heplesser wrote his comment directly in the pull request level while @gewaltig did it in the fork of @sdiazpier.
Is there any best practice defined for the code review comments?

gewaltig:
I thought I also commented the Pull request directly….
Sorry if I did not do so.

Best
Marc-Oliver

On 18 Jun 2015, at 13:48, Mikael Naveau <notifications@github.com<mailto:notifications@github.com>> wrote:


@gewaltig<https://github.com/gewaltig> Thanks for your review.

Just as a side note, would it be possible to get all the comments at the same place? @heplesser<https://github.com/heplesser> wrote his comment directly in the pull request level while @gewaltig<https://github.com/gewaltig> did it in the fork of @sdiazpier<https://github.com/sdiazpier>.
Is there any best practice defined for the code review comments?

apeyser:
Comments should definitely be attached to the pull request.

tammoippen:
Can you please run clang-format on your files (or run <NEST-src>/extras/format_all_c_c++_files.sh in the source dir of your NEST directory). The following files show differences to the formatted ones:
nestkernel/synaptic_element.h
nestkernel/node.h
nestkernel/nest_names.h
nestkernel/archiving_node.h
nestkernel/archiving_node.cpp

nestkernel/nest_names.h
```diff
@@ -130,6 +131,9 @@ extern const Name events;          //!< Recorder parameter
 extern const Name ex_spikes;       //!< Number of arriving excitatory spikes
 extern const Name exc_conductance; //!< Recorder parameter
 
+extern const Name Ca;
+extern const Name tau_Ca;
+extern const Name beta_Ca;
```
tammoippen: The names are ordered alphabetically. Please sort them as you did in the nest_names.cpp file.

sdiazpier:
Thank you @tammoippen for your comments. I ran the clang tool again on my files and I got some modifications in format for files we have not touched in our branch. I committed the format changes suggested only on files modified in our branch. I am wondering why the clang format tools suggests something different for those other files if I have not done any modifications to the .clang-format file at all. Any idea? The suggested changes are only related to the maximum number of characters per line and the indentation of new lines.

sdiazpier:
Travis tests are also randomly failing with the following error:
bash: /home/travis/wrapper.sh: No such file or directory
Anyone has an idea of why this could be happening?
Thank you all for your feedback and support.

apeyser:
Did you merge from master? I believe I merged @tammoippen's fix #43

abigailm:
Please take a look at this revised PR, @heplesser , @gewaltig , @tammoippen to see if your comments have been sufficiently addressed by the revisions

nestkernel/archiving_node.cpp
```diff
@@ -191,12 +214,29 @@ nest::Archiving_Node::set_spiketime( Time const& t_sp )
...
+  def< double >( d, names::Ca, Ca_minus_ );
+  def< double >( d, names::tau_Ca, tau_Ca_ );
+  def< double >( d, names::beta_Ca, beta_Ca_ );
   def< double >( d, names::tau_minus_triplet, tau_minus_triplet_ );
 #ifdef DEBUG_ARCHIVER
   def< int >( d, names::archiver_length, history_.size() );
 #endif
+
+  synaptic_elements_d = DictionaryDatum( new Dictionary );
+  def< DictionaryDatum >( d, "synaptic_elements", synaptic_elements_d );
```
gewaltig: I think the string "synaptic_elements" should be a name as well.

nestkernel/archiving_node.cpp
```diff
@@ -205,20 +245,63 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
...
+
   // check, if to clear spike history and K_minus
   bool clear = false;
   updateValue< bool >( d, names::clear, clear );
   if ( clear )
+  {
     clear_history();
+  }
+
+  if ( not d->known( "synaptic_elements" ) )
```
gewaltig: See my comment above. Make "synaptic_elements" a name. If a string is used more than once, make it a name to avoid misspellings.

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+  , eps( 0.7 )
+{
+  name = "gaussian";
+}
+
+void
+nest::GrowthCurveGaussian::get( DictionaryDatum& d ) const
+{
+  def< std::string >( d, "growth_curve", name );
+  def< double_t >( d, "eps", eps );
```
gewaltig: Please use the name:: versions for all strings throughout.

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+  def< double_t >( d, "eta", eta );
```
gewaltig: Please use the name:: rather than the string.

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+void
+nest::GrowthCurveGaussian::set( const DictionaryDatum& d )
+{
+  updateValue< double_t >( d, "eps", eps );
```
gewaltig: Please use name::

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+  updateValue< double_t >( d, "eta", eta );
```
gewaltig: Please use name::

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+    growth_rate_ = other.growth_rate_;
+    tau_vacant_ = other.tau_vacant_;
+  }
+  return *this;
+}
+
+nest::GrowthCurve*
+nest::SynapticElement::new_growth_curve( std::string name )
+{
+  if ( not name.compare( "gaussian" ) )
```
gewaltig: Is it possible to use name objects also here? I presume that these values are passed through a dictionary?

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+  throw BadProperty( "Unknown growth curve. Available growth curve are: linear, gaussian." );
+}
+
+/* ----------------------------------------------------------------
+* get function to store current values in dictionary
+* ---------------------------------------------------------------- */
+void
+nest::SynapticElement::get( DictionaryDatum& d ) const
+{
+  // Store current values in the dictionary
```
gewaltig: Please use names instead of verbatim strings

nestkernel/synaptic_element.h
```diff
@@ -0,0 +1,293 @@
...
+    z_connected_ += n;
+  }
+
+  /*
+   * Used to define the dynamics of the synaptic elements using a Growth Curve
+   */
+  void
+  set_growth_curve( GrowthCurve* g )
+  {
+    delete growth_curve_;
```
gewaltig: Is it possible that growth_curve_==g ? If so, you'll need to check for that.

gewaltig:
I have two comments still.
First, there are many verbatim strings throughout that should be names. 
Second, there is one pointer operation which may need a check. Please look at the line comment.

nestkernel/node.h
```diff
@@ -477,6 +477,77 @@ class Node
   virtual void handle( DoubleDataEvent& e );
 
   /**
+   * Return the Ca_minus value at time t
+   * Return 0.0 if not overridden
+   */
+  virtual double_t
```
heplesser: I am confused about the `at time t` in the comment, given that the method does not take any time parameter. So I guess it means "at the present time". That makes me wonder who calls this method when. If nodes called it on each other during `update()`, this could lead to problems with causality given that the order of node updates is undefined, so that it is not clear what time it actually is. I also see potential problems in parallel simulations.

Could you comment on this and maybe clarify it in a comment? This also applies to other getter-methods below.

nestkernel/node.h
```diff
@@ -477,6 +477,77 @@ class Node
   virtual void handle( DoubleDataEvent& e );
 
   /**
```
heplesser: I think it would be very useful if you grouped all the getter-functions related to structural plasticity in a doxygen group and added a group-level explanation of their purpose and usage. You can do this with `@defgroup` and `@ingroup` Doxygen tags, see, e.g., `librandom/randomgen.h`. It would make it much easier for other developers to understand and use this new NEST component.

nestkernel/node.h
heplesser: So what does this method return: the actual usable number or the internal value?

heplesser:
@sdiazpier @naveau @abigailm 
Thanks for fixing most things I commented on! There are still a few things I do not understand and where I think more comments would be useful, see my detailed comments on the code, especially my concerns about "what is t" and causality. I assume all is fine, but it would be helpful to have documentation explaining matters.

A nitpicky point: According to our [code style guidelines](http://nest.github.io/nest-simulator/coding_guidelines_c++#variables-and-class-members), all private member variables should have names ending in an underscore. Such things are rather easy to fix using refactoring functions in IDEs like Eclipse and XCode.

sdiazpier:
@heplesser and @gewaltig  thank you for all your comments and sorry for my delayed response. I will work on them and get back to you with the corresponding changes and added documentation for the unclear sections of code.

sdiazpier:
Dear @heplesser and @gewaltig, I have re-submitted the code following your suggestions and corrections. Please let me know if anything else is needed for the pull request to be accepted. Kind regards.

heplesser:
@sdiazpier Nice work! I commented a few more points in the code. 

I am also concerned about the way you create new `GrowthCurve` object by if-else chain in `SynapticElement::new_growth_curve()`. This does not implement a clear separation of concerns, nor does it scale well, or would allow others to add their own growth curve classes without modifying your code.

What we need here is a factory for `GrowthCurve` objects, based on the factory design pattern. Take a look at `ConnBuilderFactory` and `ConnBuilder` in nestkernel. The factory set up by those classes is then filled with `register_conn_builder()` calls in nestkernel.cpp, around line 2070. In exactly the same way, user modules could add more `ConnBuilder`s. By using a factory design, the same would become possible for growth curves.

sdiazpier:
@heplesser thank you very much for your feedback and comments. I will work on this and re-submit as soon as possible.

abigailm:
@heplesser - at this point I would suggest Sandra takes care of the minor points you mentioned, but that we open a ticket for the factory classes extension rather than require it to merge the current PR. Would you be ok with that?

nestkernel/archiving_node.h
```diff
@@ -147,6 +207,28 @@ class Archiving_Node : public Node
 
   // spiking history needed by stdp synapses
   std::deque< histentry > history_;
+
+  /*
+   * Structural plasticity
+   */
+
+  // Time of the last update of the Calcium concentration in ms
```
gewaltig: This comment is minor: I suggest that you make all comments describing the private state variables  (Ca_t_, Ca_minus_, etc.) to conform to the doxygen format so that they show up in the documentstion.

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+  double_t z_minus,
+  double_t tau_Ca,
+  double_t growth_rate ) const
+{
+  double_t Ca, z_value;
+  Ca = Ca_minus * std::exp( ( t_minus - t ) / tau_Ca );
+  z_value =
+    growth_rate * tau_Ca * ( Ca - Ca_minus ) / eps + growth_rate * ( t - t_minus ) + z_minus;
+
+  if ( z_value > 0 )
```
gewaltig: z_value is a double_t, so use 0.0 to avoid implicit conversions

nestkernel/synaptic_element.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+  for ( lag = t_minus; lag < ( t - h / 2.0 ); lag += h )
+  {
+    Ca = Ca - ( ( Ca / tau_Ca ) * h );
+    dz = h * growth_rate * ( 2.0 * exp( -pow( ( Ca - xi ) / zeta, 2 ) ) - 1.0 );
+    z_value = z_value + dz;
+  }
+
+  //  return z_value;
+  if ( z_value > 0 )
```
gewaltig: Same as before, use 0.0 in the comparison

nestkernel/synaptic_element.h
```diff
@@ -0,0 +1,297 @@
...
+
+class SynapticElement;
+class Archiving_Node;
+
+/**
+ * \class GrowthCurve
+ * Defines the way the number of synaptic elements changes through time
+ * according to the calcium concentration of the neuron.
+ */
+class GrowthCurve
```
gewaltig: This class and some derived ones have private/protected member variables whose names should end in an underscore to be consistent with the conventions.

nestkernel/synaptic_element.h
```diff
@@ -0,0 +1,297 @@
...
+
+  bool
+  continuous() const
+  {
+    return continuous_;
+  }
+
+private:
+  // The current number of synaptic elements at t = z_t_
+  double_t z;
```
gewaltig: This variable should end in an underscore as well

heplesser:
@abigailm I would be okay with moving the refactoring to factory classes to a separate issue and merging this PR after the minor aspects are fixes, provided that that new ticket would be be addressed in the immediate future ("zeitnah" as the Germans say), preferably before the structural plasticity code gets wrapped into a release.

sdiazpier:
Dear @heplesser and @gewaltig, I have submitted new changes to the code. I hope I have addressed most or all your comments. Thank you for spending so much time reviewing my code, all comments and suggestions are always very valuable to me. Please let me know if anything else is missing.

heplesser:
@sdiazpier Excellent work, it is really nice that you implemented the factory concept! I have just a few small formatting issues left, otherwise the code is fine for merging.

sdiazpier:
Dear @heplesser thanks for your comments. I have made all the suggested corrections. Hope everything is ok and ready for merging.

heplesser:
@sdiazpier Great work!
@tammoippen @jougs :+1: from me after this great piece of work and very productive code review!

lekshmideepu:
Since all the code reviewers gave :+1: I'm going to merge it

PR #34, opened 2015-06-17
Added the Travis information to the CI page [ci skip]

lekshmideepu:
Updated the docs page of CI

### DIFF
--- a/continuous_integration.md
+++ b/continuous_integration.md
@@ -37,6 +37,7 @@ The current CI implementation is now available at the following URL:
     https://travis-ci.org/nest/nest-simulator
 
 Travis CI service has been integrated with the Github repository to automatically run the tests when code is pushed. Github integration is done by adding a simple YAML file to the project root.Travis results will appear in the Github pull requests and the primary log is visible in the Travis interface. 
+
 Build jobs
 
 The CI system is closely integrated to Github repository. Whenever some changes in the code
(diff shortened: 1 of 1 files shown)


continuous_integration.md
```diff
@@ -25,81 +25,37 @@ to re-run the test suite for all target platforms on every single code
...
 
-If any of these steps fails (returns a non-zero exit code), the build
-is marked as failed and a notification is sent to the developers
-mailing list.
+If any of the first three steps fails (returns a non-zero exit code), the build
+is marked as failed and a notification is sent to the person who initiated the build.
 
-Additionally, a number of metrics are collected from the build log to
-obtain a cumulative project "health" indicators:
+Due to security reasons, Travis doesn't allow uploading of artifacts to S3 for the pull requests.
```
heplesser: @lekshmideepu I don't quite understand this line; on line 48 above you say that artefacts are uploaded to S3---what is the difference?

lekshmideepu: From a pull request, the secure values (for e.g.: the encrypted access keys for S3) couldn't be extracted properly and thus the artifacts can't be uploaded to S3. Secure access keys are set for the particular repository (in our case, nest/nest-simulator). Because of this reason, artifacts will be uploaded when we merge the commit to the upstream. 
May be I could add this information along with line 53.

PR #29, opened 2015-06-15
Refactoring .travis.yml and build.sh

tammoippen:
Reduce the number of apt-get packages to only necessary ones and order by use.
Order env.matrix in 'binary' meaningful way.
Update whitespace and indentation.

PR #35, opened 2015-06-17
Switched default AdEx model to aeif_cond_alpha_RK5 (ed)

gewaltig:
This changeset switches the default adex model with alpha synapses to the aeif_cond_alpha_RK5 implementation which is gsl independent, faster and more accurate than the legacy gsl version.


gewaltig:
according to the log, all tests - except for MPI - pass. 
I find it quite strange that many of the MPI selftests are failing, even though they do not depend on anything that I have done here.

gewaltig:
@apeyser @jougs I am a bit at loss with travis here. According to the logs, nest compiles and passes all tests, but still build.sh fails with code 2.
Any ideas?

tammoippen:
Hi, that was me. Apparently, TravisCI accept my pull request #29, but it was erroneous. I am working on this.

tammoippen:
#39 Fixed the TravisCI .yaml. To rerun the TravisCI build on the most current source state, you have to do another commit in the gewaltig:adex-rename branch, e.g.

    git commit --amend   # this just touchs the last commit and does not modify anything
    git push origin adex-rename

testsuite/unittests/test_aeif_cond_alpha_multisynapse.sli
```diff
@@ -63,14 +63,14 @@
 /spike_generator Create /sg Set
 sg << /spike_times [spike_time] >> SetStatus
 
-% Create the several aeif_conf_alpha_RK5 (one for each synapses of the previous neuron)
+% Create the several aeif_conf_alpha (one for each synapses of the previous neuron)
```
abigailm: typo: aeif_cond_alpha

pynest/examples/evaluate_aeif_cond_alpha.py
```diff
@@ -131,13 +135,13 @@ def run_model(model='aeif_cond_alpha', dt=0.1,reps=1):
 
 '''
 Now we run the different simulations.
-Three traces a produced. First, a reference trace of the original `aeif_cond_alpha model with 100 times higher resolution than the default.
-The second trace is from the `aeif_cond_alpha model with default resolution and the final trace is from the `aeif_cond_alphaRK5 model also at default resolution.
+Three traces a produced. First, a reference trace of the original `aeif_cond_alpha_gsl model with 100 times higher resolution than the default.
```
abigailm: typo: Three traces are produced

abigailm:
The PR looks fine to me, but what is the purpose of maintaining the legacy gsl version if tests show that the RK5 version is faster and more accurate? I would recommend simply removing the obsolete version and calling the RK5 version aeif_cond_alpha

gewaltig:
Good point. I simply was not daring enough ;-) Let's remove the old version entirely

gewaltig:
@abigailm I've removed the old _gsl implementation and fixed the test typos. The eval_*.py file, however is now gone, since it compared the GSL and non-GSL versions.

testsuite/unittests/test_aeif_cond_alpha_multisynapse.sli
gewaltig: fixed

gewaltig:
@tammoippen Travis still fails with the message:
"
+f_base=reports/aeif_cond_alpha.cpp
./build.sh: 121: ./build.sh: cannot create reports/aeif_cond_alpha.cpp_vera.txt: Directory nonexistent
+vera++ --root ./vera_home --profile nest models/aeif_cond_alpha.cpp
The command "./build.sh" exited with 2.
"

Any idea?

tammoippen:
This is exactly what I have fixed. Maybe you have to merge in the changes from master first:

    git checkout master
    git pull upstream master
    git checkout <your-branch>
    git merge master
    git push origin master

models/aeif_cond_alpha.h
```diff
@@ -378,7 +365,57 @@ aeif_cond_alpha::set_status( const DictionaryDatum& d )
   S_ = stmp;
 }
 
+/**
+ * Function computing right-hand side of ODE for the ODE solver.
+ * @param y State vector (input).
+ * @param f Derivatives (output).
+ */
+inline void
+aeif_cond_alpha::aeif_cond_alpha_RK5_dynamics( const double y[], double f[] )
```
heplesser: I'd suggest to move this method (back) to the *.cpp file. Since it is only ever called from code within the cpp file, it can be inlined entirely within that file, no need to have it in the h-file.

Another question is whether inlining helps performance. The method is quite large, so function call overhead is probably limited compared to the total runtime per call, while code-bloat due to inlining is probably considerable, given that the method is called seven times in the inner update loop, so inlining will paste close to 300 lines of code into the loop.  Would be interesting to test this empirically.

models/aeif_cond_alpha.h
```diff
@@ -74,7 +64,7 @@ Membrane Parameters:
   C_m        double - Capacity of the membrane in pF
   t_ref      double - Duration of refractory period in ms.
   V_reset    double - Reset value for V_m after a spike. In mV.
-  E_L        double - Leak reversal potential in mV.
+ E_L        double - Leak reversal potential in mV.
```
abigailm: misalignment intentional?

tammoippen:
PR #43 is merged now. Please add another commit to your branch (even if it is an empty one [git commit --amend]), so that TravisCI runs the build on the most current sources.

jougs:
@gewaltig: please note that we decided to close all issues and pull requests labelled "action required" after 3 months of inactivity after the label was added or after the last action in such an issue or pull request. The decision was taken in the NEST Core meeting on September 7, 2015.

heplesser:
See also #114: The model (and all other `aeif_cond_*` models do not behave correctly for `Delta_T==0`.

heplesser:
While working on #114, I realized that we should look at numeric methods for aeif-neurons once more, so that they (a) can handle the limiting case `Delta_T==0` (neuron spikes at once when `V_m >= V_th`, cf original paper) and (b) avoid numerical instabilities when then exponential current grows very fast within a time step, driving `V_m` to extremely large values, while the model definition states that `V_m` never can exceed `V_peak`. Integrating past `V_peak` will lead to incorrect values for `w`. I am not entirely sure how to handle this best. I tried simply capping `V_m` by
```
if ( not ( S_.yin[ State_::V_M ] <= P_.V_peak_ ) )
  S_.yin[ State_::V_M ] = P_.V_peak_;
```
after each substep, but that did not help (the somewhat strange comparison is written so that it would handle NaN values correctly).

I am also wondering if Dormand-Prince necessarily is the best approach, since the model becomes extremely stiff once `V_m > V_th`.

Finally, on an implementation side, we will need the integrator in all aeif-models, so I would suggest to factor it out into an `AeifIntegrator` class that is then used by all aeif models.

models/aeif_cond_alpha.cpp
```diff
@@ -395,46 +322,147 @@ nest::aeif_cond_alpha::calibrate()
...
+        {
+          S_.yref[ i ] = S_.y_[ i ]
+            + h * ( 5179.0 / 57600.0 * S_.k1[ i ] + 7571.0 / 16695.0 * S_.k3[ i ]
+                    + 393.0 / 640.0 * S_.k4[ i ]
+                    - 92097.0 / 339200.0 * S_.k5[ i ]
+                    + 187.0 / 2100.0 * S_.k6[ i ]
+                    + 1.0 / 40.0 * S_.k7[ i ] );
+        }
+
+        err = std::fabs( S_.ynew[ 0 ] - S_.yref[ 0 ] ) / MAXERR + 1.0e-200; // error estimate,
```
heplesser: Instead of `0`, `State_::V_M` would be a more informative index.

Silmathoron:
@gewaltig I just stumbled on this pull-request and since I am currently working with the aeif models, this sure startled me: on what kind of tests did you see that the RK5 performed faster than the GSL implementation? It sure is more precise, but on my computer, the following script leads to 7s for "aeif_cond_alpha" vs 18s for "aeif_cond_alpha_RK5"...
[gsl_vs_rk5.txt](https://github.com/nest/nest-simulator/files/135179/gsl_vs_rk5.txt) (convert it to .py)
Best

jougs:
@Silmathoron: i just ran your script and can confirm your observations: 10.33 s for `aeif_cond_alpha`, 27.43 s for `aeif_cond_alpha_RK45`.
@gewaltig: can you please provide measurements showing the improvements in accuracy? Thanks!

Silmathoron:
@jougs though it's not really proof since my models have not been tested by others yet, the implementation of aeif I proposed in the PR #229 (using interpolation to get rid of the imprecision of the reset) match very closely the dynamics of `aeif_cond_alpha_RK5`. This implies (if the models are good) that RK5 is indeed more precise than `aeif_cond_alpha`.

jougs:
@Silmathoron: how do your models compare speed-wise? Could you maybe also have a look at #114 and add a reference to your PR if that makes sense?

Silmathoron:
@jougs: interpolation takes virtually no time and it occurs only when a spike is emitted so I did not see any difference with the older `aeif_cond*` models in terms of speed, i.e. on my machine it runs more than twice faster than `aeif_cond_alpha_RK5`.

Regarding #114, I must say I'm a bit skeptical about setting `Delta_T=0`... the current implementation seems fine to me as the model is designed to be closer to the biological behaviour than the iaf.
My guess is that people setting `Delta_T=0` in fact want an adaptive integrate and fire ("aiaf") so I would advise implementing this special model instead of trying to setup a limit-case of the AdExp model which (to me) has no biophysical relevance.

jougs:
NEST Core decided to close this due to inactivity. We suggest opening a new issue regarding the addition of a `aiaf` model as suggested by https://github.com/nest/nest-simulator/pull/35#issuecomment-186309378

gewaltig:
Sorry, that I have missed this thread entirely.
I will look at the problem and dig out the scripts for the comparison

gewaltig:
Dear all,

the comparison between the two implementations is done by the script
evaluate_aeif_cond_alpha_RK5.py in the pynest examples.

On my mac, I get the following output:
Runtime GSL: 0.0493731832504 Test: 0.0436501979828 Ratio: 0.884087172613 Spike Times GSL - Reference:[ 0.082 0.02 0.053 0.153 0.163 0.165 0.167 0.169 0.171 0.173 0.275 0.277 0.279 0.281 0.283 0.385 0.387] Spike Times Test - Reference: [ 0.082 0.02 0.053 0.053 0.063 0.065 0.067 0.069 0.071 0.073 0.075 0.077 0.079 0.081 0.083 0.085 0.087] L2-Error (per s) GSL - Reference: 34.4949607855 L2-Error (per s) Test - Reference: 3.97167314363 L2-Error ratio: 0.115137778191

The first part compares the runtime. a ratio of 0.88 means that the RK5 implementation is about 12 percent faster that the GSL implementation.
At the same time, the error of the GSL implementation is almost 10 times larger!

What remains to be tested is why the RK5 implementation in a network is so much slower.

heplesser:
I suggest we discuss in #494, to keep it in one place.

PR #39, opened 2015-06-18
Fix TravisCI

tammoippen:
Apparently, TravisCI accept my pull request #29, but it was erroneous.
The important line is this in the builds was something like this:

    ./build.sh: 121: ./build.sh: cannot create reports/uniformint_randomdev.cpp_vera.txt: Directory nonexistent

I got confused about in what directory we perform the lint-checks – in the source directory.
But the ‘reports’ directory is in the build-directory.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -115,7 +115,7 @@ for f in $file_names; do
   case $f in
     *.h | *.c | *.cc | *.hpp | *.cpp )
       echo "Static analysis on file $f:"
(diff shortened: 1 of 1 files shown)


apeyser:
Maybe we should use a pattern of tags to mark mainline builds that are found to be broken, just to make chasing them down in the future easier? Such as a tag evil/vera-39 maybe?

PR #40, opened 2015-06-19
Travis now should build on pushes to any branch except gh-pages.

heplesser:
@lekshmideepu I patched the `.travis.yml` file as you suggested. After pushing it to my github repo, Travis did not build this branch, but maybe that is because the `.travis.yml` file in the main branch still is the old one?

Additional proposed reviewer: @apeyser

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ matrix:
   # do notify immediately about it when a job of a build fails.
   fast_finish: true
 branches:
- only:
-  - master
+  except:
+  - gh-pages
 cache:
   apt: true 
   ccache: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,6 @@ env:
 matrix:
   # do notify immediately about it when a job of a build fails.
   fast_finish: true
-branches:
-  except:
-  - gh-pages
 cache:
   apt: true 
   ccache: true


apeyser:
Actually, the two lines should be simply removed. gh-pages are blacklisted by default, so the lines are unnecessary.

gewaltig:
@lekshmideepu When I run travis on my (unmodified) fork of nest, I get the following error:
"
The following packages have unmet dependencies:
 libllvm3.6 : Depends: libstdc++6 (>= 4.8) but 4.6.3-1ubuntu5 is to be installed
E: Unable to correct problems, you have held broken packages.
The command "sudo apt-get install -y libllvm3.6 clang-format-3.6 vera++ cppcheck" failed and exited with 100 during .
"

See full logs at:
https://travis-ci.org/gewaltig/nest-simulator/jobs/67365553

Any ideas what goes wrong?

apeyser:
Check that your repository is activated on travis & that build pushes are activated for the repository. If you check under the settings drop down, there's a menu element "requests" which will show you push requests that are on the queue and pending --- it can take a while to get a build.

lekshmideepu:
@gewaltig: I checked your build system info. Your builds are not activated for Trusty. It's running under precise (Ubuntu 12.04 LTS) only. You can check that under the "Build system details" on the primary log of Travis. That's the reason why you are getting the above error.If you would like to test the builds on Trusty, i could ask the travis team to add your builds to it.

gewaltig:
I suspected something like this, but I could not figure out where to define the details of the build environment.
If you can somehow have trusty activated, that would be great.
It’s much better to be able to test pull requests before submitting them for review.

Best
Marc-Oliver
On 19 Jun 2015, at 11:04, lekshmideepu <notifications@github.com<mailto:notifications@github.com>> wrote:


@gewaltig<https://github.com/gewaltig>: I checked your build system info. Your builds are not activated for Trusty. It's running under precise (Ubuntu 12.04 LTS) only. You can check that under the "Build system details" on the primary log of Travis. That's the reason why you are getting the above error.If you would like to test the builds on Trusty, i could ask the travis team to add your builds to it.

lekshmideepu:
os details
@gewaltig I will let you know once your builds are activated for Trusty.

apeyser:
@gewaltig, @lekshmideepu 
Please keep one issue to each issue. Open up a new issue and move the discussion on gewaltig builds there.

heplesser:
@apeyser I may have fixed the wrong problem here, given that gh-pages is blacklisted by default (does this apply also to builds from personal repos?). What I really want to achieve is the following:

- build all branches derived from master (i.e., branches containing code)
- ignore all branches derived from gh-pages (i.e., branches containing web pages)

Is there any way to achieve this, short of carefully remembering to add a `[ci skip]` to every single commit message in a branch from `gh-pages`?

lekshmideepu:
@heplesser : Could you please check whether you have checked the "build only if .travis.yml" is present option. I'm attaching the screenshot.
travis settings

apeyser:
@heplesser It applies to all repos. The default for all repos is that every branch gets built except for gh-pages. If you want to ignore all branches derived from gh-pages, one solution would be to use a regular expression as your black-list. Name every gh-pages derived branch as gh-pages-derived/$something, and match /^gh-pages-derived/ as your black-list. (Just gh-pages/ won't work)

lekshmideepu:
As I understand, gh-pages are blacklisted by default. You always need .travis.yml file in all branches including master to start the build. If you check the above option, you don't have to explicitly mention the whitelist in your yaml file.

heplesser:
@apeyser @lekshmideepu 
Ok, I will then remove the `branches:` section completely from `.travis.yml`.

apeyser:
Looking in my travis account, it appears that the default rule isn't as I believe was document, that gh-pages is excluded --- but that all branches that match the regular expression gh-pages is excluded, which explains why it was ignored on heplesser's repo. On the other hand, we're merging here to master, so it's not ignored. Lesson: never name a branch x-gh-pages-y unless it's derived from gh-pages, and name all gh-pages branches as gh-pages-* or gh-pages@/* or such.

lekshmideepu:
gh-pages doesn't include a .travis.yml file. Thus, the branches derived from gh-pages will also don't include the yaml file. @heplesser is right. You can use [ci skip] or [skip ci] anywhere in the commit message for not initiating a travis build for that commit.

apeyser:
@lekshmideepu But  heplesser:travis_blacklist_gh_pages does have a .travis.yml file and didn't get built by travis in my repository even though travis saw it. In the Requests list it says "github pages branch" even though this branch is not a "github pages branch". Thus, it must be determined by name pattern.

PR #44, opened 2015-06-22
Fix #43

### LINKED ISSUE #43: TravisCI fails on removed files
When a file is removed in a commit, the build.sh still tries to run cppcheck, vera++ and clang-format on it, hence all builds will fail. The solution is to first check, whether the file in the changeset, which is under testing, is still present – so before build.sh:115 have something like:

    if test -e "$f" ...

tammoippen:
Add a test to validate that only existing files are analyzed with
vera++, cppcheck and clang-format during TravisCI builds.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -111,7 +111,11 @@ EOF
 file_names=`git diff --name-only $TRAVIS_COMMIT_RANGE`
 
 for f in $file_names; do
-   # filter files
+  if [ ! -f "$f" ]; then
+    echo "$f : Is not a file or does not exist anymore."
+    continue
+  fi
+  # filter files
   case $f in
     *.h | *.c | *.cc | *.hpp | *.cpp )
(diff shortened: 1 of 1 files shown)


heplesser:
Looks good to me. +1

tammoippen:
@lekshmideepu or @apeyser , can you please also look into this. Otherwise deleted files in commits will make TravisCI fail.

build.sh
```diff
@@ -111,7 +111,11 @@ EOF
 file_names=`git diff --name-only $TRAVIS_COMMIT_RANGE`
 
 for f in $file_names; do
-   # filter files
+  if [ ! -f "$f" ]; then
```
apeyser: Should be -e "$f" or else will fail if $f is a symbolic link --- unless we have a policy of no symbolic links, ever.

PR #47, opened 2015-06-24
Moving topology functions into main namespace on the SLI level (trac.290)

heplesser:
This PR should solve ticket 290 from the trac.

Suggested reviewers: @jougs @espenhgn

### DIFF
--- a/testsuite/mpitests/ticket-955.sli
+++ b/testsuite/mpitests/ticket-955.sli
@@ -37,8 +37,6 @@ Author: Hans Ekkehard Plesser, 2015-02-03 based on a reproducer by Janne Morén
 
 [1 2 4]
 {
-  topology using
-  
   << 
     /positions [[0.0 0.0]]
     /elements /iaf_neuron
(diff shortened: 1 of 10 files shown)


espenhgn:
It can probably be merged as is, but @jhnnsnk commented that 
- There may be an "endusing" missing in file ticket-955.sli
- There is inconsistent usage of importing namespaces, i.e.: "/unittest using" vs. "unittest using". 
General note:
- cvdict is implemented for masks but not for parameters. We wish to have this function also for parameters.

heplesser:
@espenhgn @jhnnsnk 
- You are right that there is no `endusing` in `ticket-955.sli`, but this is the same for all `mpitests`. The reason is that the code in MPI tests is executed in a peculiar and way, where tests files should only contain a single `distributed_*_or_die` call at the very end of the file.
- I found three cases of `unittest using` and updated them to `/unittest using`.
- Could you open a separate issue for the extension of `cvdict` to parameters? That makes it easier to keep track of things.

Thanks for reviewing!

jougs:
Many thanks for the thorough review, @espenhgn, @jhnnsnk. I am happy now and thus merging.

PR #51, opened 2015-06-26
Fix trac.730 synapse documentation

heplesser:
Note: I mixed up two tickets, so this ended up with an incorrect branch name. This is about **trac.730**.

This pull request should fix trac.730. I mainly updated and clarified synapse model descriptions, especially wrt common synapse properties. The problems with singular and plural parameter names were a non-issue, plural was only used in the documentation of `tsodyks_synapse`, while actual parameter names were singular as everywhere else.

I also added tests for `cont_delay_synapse` and added `DS*` and `DataLoggingRequest` events, so it is now a fully-fledged static synapse and can be used to connect generators to neurons.

I also replace `Note:` with `Remarks:` in online documentation comments where necessary, so that our documentation generator sees the correct keyword.

Suggested reviewers: @abigailm

### DIFF
--- a/models/tsodyks_connection.h
+++ b/models/tsodyks_connection.h
@@ -68,12 +68,12 @@
 
    Parameters:
      The following parameters can be set in the status dictionary:
-     Us         double - maximum probability of realease [0,1]
-     tau_pscs   double - time constants of synaptic current in ms
-     tau_facs   double - time constant for facilitation in ms
-     tau_recs   double - time constant for depression in ms
-     xs         double - initial fraction of synaptic vesicles in the readily releasable pool [0,1]
-     ys         double - initial fraction of synaptic vesicles in the synaptic cleft [0,1]
+     Us        double - maximum probability of realease [0,1]
+     tau_psc   double - time constants of synaptic current in ms
+     tau_fac   double - time constant for facilitation in ms
+     tau_rec   double - time constant for depression in ms
+     x         double - initial fraction of synaptic vesicles in the readily releasable pool [0,1]
+     y         double - initial fraction of synaptic vesicles in the synaptic cleft [0,1]
 
   References:
    [1] Tsodyks, Uziel, Markram (2000) Synchrony Generation in Recurrent Networks
--- a/models/tsodyks_connection.h
+++ b/models/tsodyks_connection.h
@@ -68,8 +68,8 @@
 
    Parameters:
      The following parameters can be set in the status dictionary:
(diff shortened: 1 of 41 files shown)


abigailm:
Everything else looks fine to me

abigailm:
+1
Maybe @jakobj or @sdiazpier could take a look as well

models/tsodyks_connection.h
```diff
@@ -68,12 +68,12 @@
...
    Parameters:
      The following parameters can be set in the status dictionary:
-     Us         double - maximum probability of realease [0,1]
-     tau_pscs   double - time constants of synaptic current in ms
-     tau_facs   double - time constant for facilitation in ms
-     tau_recs   double - time constant for depression in ms
-     xs         double - initial fraction of synaptic vesicles in the readily releasable pool [0,1]
-     ys         double - initial fraction of synaptic vesicles in the synaptic cleft [0,1]
+     U         double - maximum probability of realease [0,1]
+     tau_psc   double - time constants of synaptic current in ms
```
jakobj: constants -> constant

jakobj:
When calling GetConnections, the delay is returned in integers as well. I was wondering whether this is a problem with the synapse or with GetConnections. In any case, does it not make sense to mention this somewhere in the documentation of the cont_delay_connection?

jakobj:
otherwise I am happy :+1:

heplesser:
@jakobj I do not quite understand you comment about `GetConnections`. That function returns

    [source-gid target-gid target-thread synapse-model-id port]

but not the delay. Calling `GetStatus` on the connections object returns a dictionary with the correct delay:

```
SLI ] /cont_delay_synapse << /delay 1.2345 >> SetDefaults
SLI ] /iaf_psc_alpha_canon Create dup
SLI [2] /cont_delay_synapse Connect
SLI ] << >> GetConnections dup
SLI [2] ==
[<1,1,0,14,0>]
SLI [1] 0 get GetStatus info
--------------------------------------------------
Name                     Type                Value
--------------------------------------------------
delay                    doubletype          1.2345
receptor                 integertype         0
sizeof                   integertype         40
source                   integertype         1
synapse_model            literaltype         cont_delay_synapse
target                   integertype         1
weight                   doubletype          1
--------------------------------------------------
Total number of entries: 7
```

jakobj:
uhh, my bad. it seems that i was thinking about PyNN. everything works as expected. :)

jougs:
Looks solid to me. Thanks to all for fixing and reviewing. Merging.

PR #52, opened 2015-06-29
Fix trac.735 : Standardise generator parameter names.

heplesser:
This PR implements the fixes to trac.735 discussed in the NEST Core meeting 29 June.

- `sinusoidal_poisson_generator` and `sinusoidal_gamma_generator` parameters are changed as follows:
  - `dc` -> `rate`
  - `ac` -> `amplitude`
  - `freq` -> `frequency`
  - `phi` -> `phase`; units changed from radian to degrees
- `ppd_sup_generator` parameters are changed as follows:
  - `amplitude` -> `relative_amplitude` (name change only)

This aligns parameters names of the sinusoidal generators with the other generators. The change of parameter name for `ppd_sup_generator` makes explicit that the modulation amplitude is relative to the mean rate for this generator.

All these changes will require adjustment pertaining scripts, they will cause existing code to fail noisily.

Proposed reviewers: @jougs @wschenck


jougs:
Looks good to me and is corresponding to what has been discussed. One thing, though: Should the internal variable `phi` also be renamed to `phase?

heplesser:
@jougs I have been wondering a bit about whether to rename the member variables. 
- I think `om_` and `phi_` should remain what they are, since they are in radians, not degrees, so keeping names distinct will (hopefully) remind readers of the difference.
- I will rename `dc_` and `ac_` to `rate_` and `amplitude_`, respectively to match the parameter names, and fix their documentation a little. 

Just running the test now and will push this update shortly.

heplesser:
Moritz Deger agreed with the change of parameter name in `ppd_sup_generator` (email to @heplesser).

tammoippen:
The code looks good to me. I am going to merge it.

PR #54, opened 2015-07-01
Fix formatting

tammoippen:
1) There was a problem with the previous (pre-git) formatting of constructor initialization lists with comments: when clang-format processed them, comments and member variables were mixed. This fixes all those problems I found.

2) Apparently there were some further format errors introduced via pull requests. They are corrected with this PR. 

3) Updated `build.sh` so that future PR will make TravisCI fail, if the committed files do not conform to the clang-format formatting.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -109,6 +109,7 @@ EOF
 
 # Extracting changed files between two commits
 file_names=`git diff --name-only $TRAVIS_COMMIT_RANGE`
+format_error_files=""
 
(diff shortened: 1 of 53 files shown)


heplesser:
7/8 travis builds are successful, so I give this a +1 already now. Thanks @tammoippen , this was a very useful clean-up!

jougs:
Looks much better now. +1 and merging.

PR #55, opened 2015-07-01
Fix 28 mymodule bluegene

### LINKED ISSUE #28: MyModule does not work on BlueGene
User modules (`MyModule`) do not work on BlueGene at present. Loading them as dynamically linked modules fails as well as building NEST with `MyModule` linked in. This is most likely due to the fact that the `MyModule` build setup does not take BlueGene peculiarities into account.

heplesser:
This PR should fix #28 and also partially address trac.526 and trac.617. Specifically

- `DynModule` is completely removed, as they no longer differed in a meaningful way from `SLIModule`
- The last code bits left of the never fully implemented module unloading mechanism removed 
- When compiling for BlueGene, user modules (based on `MyModule`) are only built as static libraries and are linked statically into NEST. Other platforms are not affected by this change

Potential reviews: @jougs @tammoippen @apeyser


tammoippen:
The file sli/interpreter.cc does not conform to clang-format. Can you call clang-format on this file again.

tammoippen:
The code looks good to me. +1 for merge.

heplesser:
@tammoippen I don't understand the formatting problems. On my machine, I get

```
[plesser@Hanss-MBP] ~/NEST/code/trunk/src/sli [fix_28_mymodule_bluegene*](0)$ clang-format interpret.cc > interpret.cc.clang
[plesser@Hanss-MBP] ~/NEST/code/trunk/src/sli [fix_28_mymodule_bluegene*](0)$ diff interpret.cc interpret.cc.clang 
[plesser@Hanss-MBP] ~/NEST/code/trunk/src/sli [fix_28_mymodule_bluegene*](0)$ clang-format --version
clang-format version 3.7.0 (tags/google/testing/2015-06-18)
```

and this line about which clang-format complains on Travis

```
 out << std::endl << msg << errorname;
```

seems perfectly fine to me. 

I am also surprised that Travis only reports build 89.8 as failing, even though all other builds have the same code.

tammoippen:
Unfortunately, this is due to the version difference: TravisCI uses clang-format 3.6 and on your Mac you use 3.7 (same with me). Also there is no flag in 3.7, which mimics the 3.6 behaviour.

> I am also surprised that Travis only reports build 89.8 as failing, even though all other builds have the same code.

I am also surprised by this: all builds should have failed...

jougs:
I'm also happy with the code, so +1 for merging. @tammoippen Can you please click the button if you think that the failing Travis is just a glitch in the format checker? Thanks!

apeyser:
If this is a glitch, shouldn't this branch get pulled and run against clang-3.6 so that we don't introduce failing files into the stream?

jougs:
@apeyser Sounds reasonable.

tammoippen:
It seems TravisCI is using different versions of the `build.sh` script. In the failing build, the changes from #54 are included, which makes wrong formatted code fail. The other builds do not have these changes, but also report the wrong formatted code as before (but do not fail).

I agree with @apeyser, but do not have write permissions to @heplesser's branch.

Issue #59, opened 2015-07-01
Invalid URL in deprecation warning for new Connection Management


frankmichler:
### Problem: old URL
The deprecation warning about changes in the connection management contains a link to the old site:
http://nest-initiative.org/Connection_Management
When clicking the link the following message shows up:

*"This is somewhat embarrassing, isn’t it? It seems we can’t find what you’re looking for. Perhaps searching can help."*

### Error Reproduction
This code will reproduce the warning:
```python
import nest
nodes = nest.Create('iaf_psc_delta', 2)
nest.Connect(nodes[0:1], nodes[1:], 'all_to_all', model='static_synapse')
```

Output:
```
lib/python2.7/site-packages/nest/hl_api.py:84: UserWarning: 
The argument 'model' is there for backward compatibility with the old Connect function 
and will be removed in NEST 2.6. Please change the name of the keyword argument 
from 'model' to 'syn_spec'. For details, see the documentation at:
http://nest-initiative.org/Connection_Management
```

### Origin and probable fix of the problem
The wrong url is in pynest/nest/hl_api.py (SHA1: ae79942), lines 80 and 1098.
I assume the URL should point to this page instead:
http://www.nest-simulator.org/connection_management/

Additionally the wording of the message in line 1096 should be changed to 
"will be removed in a future version of NEST" like in line 78.

frankmichler:
The fix was merged in commit db80b18e53acac751c6c95b54711e37a75bd9f75
So I guess the issue can be closed now.

PR #60, opened 2015-07-02
Fix issue #59 in hl_api.py: Invalid URL in deprecation warning

frankmichler:
The deprecation warning in hl_api.py points to an outdated URL:
http://nest-initiative.org/Connection_Management
Changed to http://www.nest-simulator.org/connection_management/

And fixed wording from "will be removed in NEST 2.6"
to
"will be removed in a future version of NEST"


jougs:
Thanks for spotting this. +1 for merging.

PR #61, opened 2015-07-02
Simplify range check fix in uniformint_random_dev

apeyser:
Redundant cases were being tested. So, the code is decreased and the commentary is extended: we don't need to check that a - b < min when we know that a > b.

@heplesser 
@abigailm 
@wschenck

### DIFF
--- a/librandom/uniformint_randomdev.cpp
+++ b/librandom/uniformint_randomdev.cpp
@@ -61,14 +61,38 @@ librandom::UniformIntRandomDev::set_status( const DictionaryDatum& d )
 
   // The following test is based on
   // https://www.securecoding.cert.org/confluence/display/c/INT32-C.+Ensure+that+operations+on+signed+integers+do+not+result+in+overflow
-  // The third condition tests that we can add 1 to the difference; by the time we reach
-  // that clause, we know that the first two are false, so that the difference does not overflow.
-  if ( ( new_nmin > 0 && new_nmax < std::numeric_limits< long >::min() + new_nmin )
-    || ( new_nmin < 0 && new_nmax > std::numeric_limits< long >::max() + new_nmin )
-    || ( new_nmax - new_nmin > std::numeric_limits< long >::max() - 1 ) )
+
+  // Rollover cases (integer bounds min, max):
+  // new_nmax = n+, new_nmin = n-
+  //
+  //   a) n+ >= n-:
+  // 	   1) n- >= 0: 0 <= n+ - n- <= max
+  //       2) n- <  0: 0 <= n+ - n-
+  // 	     must confirm that n+ - n- <= max
+  //
+  //   b) n+ < n-:
+  // 	   1) n- <= 0: 0 > n+ - n- >= min
+  //       2) n- >  0: 0 > n+ - n-
+  // 	     must confirm that n+ - n- >= min
+  //
+  //  Case b) is eliminated by the first test above.
+  //  Case a) is checked by confirming that
+  //    n+ <= max + n-
+  //    which can not rollover when n- < 0
+  //
+  //  Additionally, we set range = 1 + (n+ - n-)
+  //  a1: n+ - n- < max
+  //    n+ < max + n-
+  //  a2: n+ - n- < max
+  //
+  //  so we need to confirm that n+ - n- != max.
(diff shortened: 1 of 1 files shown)


lib/sli/rcsinfo.sli
```diff
@@ -0,0 +1 @@
+statusdict /rcsinfo (no_rcsinfo_available) put
```
heplesser: I think this file should not be committed, should it? Isn't it automatically generated?

librandom/uniformint_randomdev.cpp
```diff
@@ -61,10 +61,17 @@ librandom::UniformIntRandomDev::set_status( const DictionaryDatum& d )
...
-    || ( new_nmin < 0 && new_nmax > std::numeric_limits< long >::max() + new_nmin )
+  //
+  // However, this one does not requre the range check for the case where
+  // the first operand is smaller than the second:
+  //    new_nmin > 0 && new_nmax < min + new_nmin
+  // because by the preceding, new_nmax >= new_nmin and thus new_nmix can not be
+  //    less than any number smaller than new_nmin
+  //
+  // The second condition tests that we can add 1 to the difference; by the time we reach
+  // that clause, we know that the first is false, so that the difference does not overflow.
```
heplesser: @apeyser I think your test is correct, but I had a hard time wrapping my mind around the comment. Here an alternative proposal for the entire comment:

```
/*
  Ensure that range = nmax - nmin + 1 does not overflow.

  The following test is based on https://...

  We know from the previous test that new_nmin <= new_nmax. Thus, if new_nmin >= 0, 
  then new_nmax-new_nmin definitely does not overflow. The first condition below tests 
  that the difference does not overflow if new_nmin < 0. The second condition then tests
  that we can add 1 to the difference.
*/
´´´

+1 for the code.

File: lib/sli/rcsinfo.sli (hunk 1 above)
Comment (apeyser) [inline lib/sli/rcsinfo.sli]: This is an empty file needed by the version tracking system which gets updated somewhere through the make files. On other people's system, it appears to end up empty --- and on mine, it ends up filled. I'd make a separate issue of it (@jougs).

So in short, it is in git, so it does get committed --- I can splice the update out, but I'm not sure what the point of this file is vis-a-vis github now.

File: librandom/uniformint_randomdev.cpp (hunk 2 above)
Comment (apeyser) [inline librandom/uniformint_randomdev.cpp]: Better than my original, but I just went out and explicitly described all the bounds and tests.

**Comment by apeyser** [comment]
For posterity's sake:

The original test:
```
!  ( 
        (new_nmin < 0 && new_nmax < max + new_nmin && new_nmax - new_min != max)
     || (new_nmin >= 0 && new_nmax - new_nmin != max)
)
```
We simplify the not:
```
   (new_nmin >= 0 || new_nmax >= max + new_nmin || new_nmax - new_nmin == max)
&& (new_nmin < 0 || new_nmax - new_nmin == max)
```
We distribute or's:
```
   new_nmin >= 0 && new_nmin < 0
|| new_nmin >= 0 && new_nmax - new_nmin == max
|| new_nmax >= max + new_nmin && new_nmin < 0
|| new_nmax >= max + new_nmin && new_namx - new_nmin == max
|| new_nmax - new_nmin == max && new_nmin < 0
|| new_nmax - new_nmin == max && new_nmax - new_nmin == max
```
We simplify each term and reorder internally:
```
   false
|| new_nmin >= 0 && new_nmax - new_nmin == max
|| new_nmin <  0 && new_nmax >= max + new_nmin
|| new_nmax - new_nmin == max
|| new_nmin < 0 && new_nmax - new_nmin == max
|| new_nmax - new_nmin == max
```
We eliminate empty terms and reduncancy:
```
   new_nmin >= 0 && new_nmax - new_nmin == max
|| new_nmin <  0 && new_nmax >= max + new_nmin
|| new_nmin <  0 && new_nmax - new_nmin == max
|| new_nmax - new_nmin == max
```
And eliminate redunancy between terms:
```
   new_nmin <  0 && new_nmax >= max + new_nmin
|| new_nmax - new_nmin == max
```

tammoippen:
Nice piece of logic arithmetics. :+1: from my side. Merging.

PR #66, opened 2015-07-10
Array connect

hannahbos:
Extended the Connect()-function such that in can handle parameter arrays for rules 'one_to_one' and 'all_to_all'. The testsuite is extended accordingly. In addition issued with the python testsuite and MPI are fixed.


nestkernel/conn_builder.h
```diff
@@ -87,6 +87,9 @@ class ConnBuilder
   //! Create connection between given nodes, fill parameter values
   void single_connect_( index, Node&, thread, librandom::RngPtr& );
 
+  //! Create connection between given nodes, fill parameter values
+  void skip_conn_parameter_( thread, librandom::RngPtr& );
```
tammoippen: Please write more explicit documentation. I believe, the `skip_conn_parameter_` does not create connections, but propagates `weight` and `delay` parameters in ParameterArrays.

nestkernel/conn_parameter.h
```diff
@@ -69,8 +69,8 @@ class ConnParameter
    * @param rng   random number generator pointer
    * will be ignored except for random parameters.
    */
-  virtual double value_double( index, index, librandom::RngPtr& ) const = 0;
-  virtual long_t value_int( index, index, librandom::RngPtr& ) const = 0;
+  virtual double value_double( thread, librandom::RngPtr& ) const = 0;
+  virtual long_t value_int( thread, librandom::RngPtr& ) const = 0;
```
tammoippen: Please update the documentation.

nestkernel/conn_parameter.h
```diff
@@ -83,7 +83,7 @@ class ConnParameter
     return 0;
   }
 
-  static ConnParameter* create( const Token& );
+  static ConnParameter* create( const Token&, const size_t );
```
tammoippen: Can you please write documentation for this method. It should also explain the new parameter.

nestkernel/conn_parameter.h
```diff
@@ -145,50 +145,82 @@ class ScalarIntegerParameter : public ConnParameter
...
+
+class ArrayDoubleParameter : public ConnParameter
 {
 public:
-  ArrayParameter( const std::vector< double >& values )
-    : values_( values )
-    , next_( values_.begin() )
+  ArrayDoubleParameter( const std::vector< double >& values, const size_t nthreads )
+    : values_( &values )
+    , next_( nthreads, values_->begin() )
```
tammoippen: Can you write some documentation explaining the architecture of `ArrayDoubleParameter` and `ArrayIntegerParameter`.

tammoippen:
The files ` pynest/nest/tests/test_connect_all_patterns.pyc` and `pynest/nest/tests/test_connect_one_to_one.pyc` should not be included, as they are compiled Python files.

tammoippen:
I cannot comment on the Python tests. Maybe somebody more familiar with the python tests can review the code as well? (@heplesser, @jougs or @mschmidt87)

tammoippen:
The code looks good for me. +1 for merge.

TravisCI is also satisfied. There is only one test case failing, because TravisCI cannot connect to the internet for `apt-get update` and blocks.

Still I have not reviewed the python tests.

nestkernel/conn_parameter.h
jougs: Cyn you please explain why you removed the possibility to use the `sgid` and `tgid` arguments in favor of the `thread` instead of just adding the `thread`? It might be useful to have the former to achieve a better mapping from PyNN and PyNEST.

pynest/examples/plot_weight_matrices.py
```diff
@@ -169,6 +169,7 @@ def plot_weight_matrices(E_neurons, I_neurons):
     pylab.tight_layout()
 
     plt3 = ax3.imshow(W_EI)
+    print W_EI
```
jougs: This looks like left-over debug output. Please remove it if that is the case.

pynest/nest/hl_api.py
```diff
@@ -1058,11 +1059,14 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
...
 
-    All other parameters can be scalars or distributions. In the case of scalar 
-    parameters, all keys must be doubles except for 'receptor_type' which must be
-    initialised with an integer. Any distributed parameter must be initialised with 
-    a further dictionary specifying the distribution type ('distribution', e.g. 'normal') 
-    and any distribution-specific parameters (e.g. 'mu' and 'sigma').
+    All other parameters can be scalars, arrays or distributions. 
+    In the case of scalar parameters, all keys must be doubles except for 'receptor_type' which must be
+    initialised with an integer. 
+    Parameter arrays are only avaibale for the rules 'one_to_one' and 'all_to_all'. For 'one_to_one' the
```
jougs: `avaibale` is misspelled

pynest/nest/hl_api.py
```diff
@@ -1058,11 +1059,14 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
...
+    array has to be a one-dimensional NumPy array with length len(pre) For 'all_to_all' the array has
```
jougs: Add a full stop before `For`.

pynest/nest/tests/test_connect_helpers.py
```diff
@@ -62,24 +62,25 @@ def mpi_barrier():
     if haveMPI4Py:
         MPI.COMM_WORLD.Barrier()
 
-def mpi_assert(data_original, data_test, label=None):
+def mpi_assert(data_original, data_test, TestCase):
     '''
-    Compares data_original and data_test in a manner depending on the form of the two input parameters and the label.
-    Returns True is the two data sets are considered equal and False otherwise.
```
jougs: is -> if

sli/fdstreamtest.cc
```diff
@@ -67,7 +67,8 @@ main()
   }
 
   o << "Line 1" << std::endl;
-  o << "Line 2" << std::endl << "Line 3" << std::endl;
+  o << "Line 2" << std::endl
+    << "Line 3" << std::endl;
```
jougs: This file has nothing to do with the pull request and just got in due to clang reformatting. Please remove from PR.

sli/interpret.cc
```diff
@@ -943,7 +946,8 @@ SLIInterpreter::message( std::ostream& out,
       if ( i != 0 && text_str.at( i - 1 ) == ' '
         && static_cast< int >( space - i ) > static_cast< int >( width - pos ) )
       {
-        out << std::endl << std::string( indent, ' ' );
+        out << std::endl
+            << std::string( indent, ' ' );
         pos = 0;
       }
```
jougs: This file has nothing to do with the pull request and just got in due to clang reformatting. Please remove from PR.

sli/tarrayobj.cc
```diff
@@ -582,7 +582,7 @@ bool TokenArrayObj::operator==( const TokenArrayObj& a ) const
   if ( size() != a.size() )
     return false;
 
-  Token *i = begin(), *j = a.begin();
+  Token* i = begin(), * j = a.begin();
   while ( i < end() )
     if ( !( *i++ == *j++ ) )
       return false;
```
jougs: This file has nothing to do with the pull request and just got in due to clang reformatting. Please remove from PR.

topology/mask.h
```diff
@@ -444,7 +444,7 @@ class DifferenceMask : public Mask< D >
   Mask< D >* clone() const;
 
 protected:
-  Mask< D > *mask1_, *mask2_;
+  Mask< D >* mask1_, *mask2_;
 };
```
jougs: This file has nothing to do with the pull request and just got in due to clang reformatting. Please remove from PR.

topology/parameter.h
```diff
@@ -897,7 +897,7 @@ class DifferenceParameter : public Parameter
   }
 
 protected:
-  Parameter *parameter1_, *parameter2_;
+  Parameter* parameter1_, *parameter2_;
 };
 
 /**
```
jougs: This file has nothing to do with the pull request and just got in due to clang reformatting. Please remove from PR.

jougs:
Sorry for the long delay.

Here's some comments in addition to the line comments above.

* The `.pyc`-files are still in the PR.
* I don't particularly like the passing around and storing of the number of threads in the `ConnParameters` class. However, after offline discussions I accept that there's no better solution currently, as the `Network` instance is not available where it would be needed. Can you please create a ticket so that this won't be forgotten when the `Network` class is restructured?

Many thanks for the nice contribution!

nestkernel/conn_builder.h
```diff
@@ -87,6 +87,11 @@ class ConnBuilder
   //! Create connection between given nodes, fill parameter values
   void single_connect_( index, Node&, thread, librandom::RngPtr& );
 
+  //! Calls value-function of all parameters being instantiations of ArrayDoubleParameter or
```
heplesser: This comment is not in proper doxygen format (http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html). The common format for multiline Doxygen comments in NEST is

```
/**
 *
 *
 */
```

The first line should be a short summary of the function, followed by an empty line and a more detailed explanation.

nestkernel/conn_parameter.h
```diff
@@ -145,50 +148,103 @@ class ScalarIntegerParameter : public ConnParameter
...
- * Array parameter, returning values in order.
+ * Array parameter classes, returning double values in order.
  *
  * - The array of values must not be empty
  *   (so return 0 for number_of_values can signal non-array parameter)
  * - Throws exception if more values requested than available.
- *
+ * - The class contains nthread number of pointers (one for each thread)
+ *   to an iterator, which runs over the parameters initialised in an array.
+ *   Each pointer is moved along the paramtere array by the function
```
heplesser: Typo in `paramtere`

pynest/nest/hl_api.py
```diff
@@ -1058,11 +1059,14 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
...
+    to be a two-dimensional NumPy array with shape len(post) x len(pre).
```
heplesser: For extra clarity, I would write "with shape `(len(post), len(pre))`.", i.e., give a tuple as returned by `.shape`. The documentation should also state explicitly that rows represent target, columns source neurons.

pynest/nest/hl_api.py
```diff
@@ -1116,9 +1128,30 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
...
+                if isinstance(value, (np.ndarray, np.generic)):
+                    if len(value.shape) == 1:
+                        if rule == 'one_to_one':
+                            if value.shape[0] is not len(pre):
+                                raise NESTError("'" + key + "' has to be an array of dimension " + str(len(pre)) + ", a scalar or a dictionary.")
+                        else:
+                            raise NESTError("'" + key + "' has the wrong type. One-dimensional parameter arrays can only be used in conjunction with rule 'one_to_one'.")
+                    elif len(value.shape) == 2:
+                        if rule == 'all_to_all':
+                            # TODO Jochen does len(pre), len(post) work for all possible types of node containers?
```
heplesser: Is this resolved? If yes, the comment should be removed.

nestkernel/conn_parameter.h
```diff
@@ -145,50 +148,103 @@ class ScalarIntegerParameter : public ConnParameter
...
-  value_int( index, index, librandom::RngPtr& ) const
+  value_int( thread, librandom::RngPtr& ) const
   {
     throw KernelException( "ConnParameter calls value function with false return type." );
   }
 
 private:
-  std::vector< double > values_;
-  mutable std::vector< double >::iterator next_;
+  const std::vector< double >* values_;
```
heplesser: I am concerned about the pointer to vector here, fearing a memory leak or worse. In the pertaining `create()` method we dereference the datum twice (`**dvd`) to access the data in the vector. Is there any risk that the actual data would be destroyed while we here still hold a pointer to it? Or that we mess up the reference count of the datum, so it would not be freed? Since these arrays may be large, we might see noticeable memory leaks. We need to look at this carefully. The same applies to the `ArrayIntegerParameter`.

nestkernel/conn_parameter.h
```diff
@@ -145,50 +148,103 @@ class ScalarIntegerParameter : public ConnParameter
...
+      throw KernelException( "Parameter values exhausted." );
+  }
+  double
+  value_double( thread, librandom::RngPtr& ) const
+  {
+    throw KernelException( "ConnParameter calls value function with false return type." );
+  }
+
+private:
+  const std::vector< long_t >* values_;
```
heplesser: See comment on `values_` in `ArrayDoubleParameter`.

nestkernel/conn_parameter.h
heplesser: I agree with Jochen. We discussed at an earlier point that we may want to introduce a general interface for parameter-generating functions which should receive `sgid` and `tgid` as input. If we want to retain this possibility, we need to pass them in, therefore we should pass them and the target thread in addition.

nestkernel/conn_builder.cpp
```diff
@@ -337,6 +337,61 @@ nest::ConnBuilder::single_connect_( index sgid,
   }
 }
 
+inline void
+nest::ConnBuilder::skip_conn_parameter_( thread target_thread, librandom::RngPtr& rng )
+{
+  if ( param_dicts_.empty() ) // indicates we have no synapse params
+  {
+    if ( default_weight_ )
```
heplesser: This is exactly the same code as at the end of the `else` clause. Thus, we can run it unconditionally and turn the test around to `not param_dicts_.empty()` as condition for what currently is the `else` clause.

nestkernel/conn_builder.cpp
```diff
@@ -337,6 +337,61 @@ nest::ConnBuilder::single_connect_( index sgid,
   }
 }
 
+inline void
+nest::ConnBuilder::skip_conn_parameter_( thread target_thread, librandom::RngPtr& rng )
```
heplesser: I am wondering about the performance consequences of the parameter skipping as currently implemented. Just the looping and branching in this method will cost a bit, but I am particularly worried about randomized parameters. If I understand the code right, then every time we skip a target because it is not MPI- or thread-local, we call this method, and then obtain (and ignore) a value for each parameter defined. In some cases, this would mean drawing several random numbers, which may be quite demanding depending on the distribution used. Unless we can clearly demonstrate that this has no bad side effects even under "extreme" conditions (costly distributions, very many threads), we need to find a better solution.

One would be a flag "skipping required", which is set if there is at least one array parameter. One could then here just have `if ( not skipping_required_ ) return;`. The problem with this solution would be that different sequences of random numbers would be used for randomized parameters depending on the presence/absence of array parameters.

Therefore, I believe that a better approach would be to collect pointers to parameters requiring skipping and then skip only those. That would also avoid all the branching logic here, we could just do

```
for ( p = double_params_needing_skipping.begin() ; ... )
  p->double_value(...);
for ( p =int_params_needing_skipping.begin() ; ... )
  p->int_value(...);
```

One more issue: Can we be certain that the compiler will not optimize the value-calls away, since the value generated is never used? In that case, skipping would not work. To be safe, we need an explicit `skip()` method.

nestkernel/conn_builder.cpp
```diff
@@ -337,6 +337,61 @@ nest::ConnBuilder::single_connect_( index sgid,
...
+{
+  if ( param_dicts_.empty() ) // indicates we have no synapse params
+  {
+    if ( default_weight_ )
+      delay_->value_double( target_thread, rng );
+    else if ( ( !default_weight_and_delay_ ) && ( !default_weight_ ) )
```
heplesser: I spent several minutes wrapping my poor brain around the logic of the Boolean and the variable names (yes, I invented them ...). Can we find a better set of names?

heplesser:
@hannahbos Very nice work, although I think we need to iterate some more as detailed above. I am in particular concerned about (i) the performance penalty of skipping when it is not needed, (ii) the removal of the `sgid` and `tgid` arguments and (iii) a possible memory leak or premature deallocation of parameter arrays.

nestkernel/conn_parameter.h
hannahbos: I agree with both of you, we planned to implement the connection parameter in an extendible fashion. I removed the sgid and tgid arguments for two reasons. First, to remove the false impression, that the required functionality using these arguments is already in place. Second, as far as I recall we introduced the arguments to allow for parameter matrices, which are now inserted as one-dimensional arrays in order to be iterable. If we decide to extend the interface to allow for parameter-generating functions which are also iterable, the arguments sgid and tgid might not be necessary. Therefore I suggest to leave the arguments out for now and re-insert them from history if required in future development. 
The argument thread was introduced to secure the parameter array of iterators when using multi-threading.

nestkernel/conn_builder.h
hannahbos: Thanks Jochen. The problem is now documented in issue #76.

nestkernel/conn_parameter.h
hannahbos: After offline discussions we came to the conclusion that we shouldn't be concerned about that. The datum is initialised in the conn_builder once and subsequently accessible for all threads via pointers. The value_()-function, which moves the pointer along the parameter array, is called within an omp parallel-block, thus securing that all threads are finished before executing further code.

nestkernel/conn_builder.cpp
hannahbos: The skip()-function is now simplified such that only parameters specified as arrays require skipping. This is accomplished by collecting all pointers to array parameter in a vector after their initialisation.

hannahbos:
@heplesser Thank you HEP for the thorough review. I hope all issues are sufficiently addressed.

heplesser:
@hannahbos Excellent work! I will add a few inline comments with nit-picky stuff, otherwise it looks very fine. 

But I am still not persuaded that we should leave out the `sgid` and `tgid` parameters from the `value_{double,int}()` interface only to reintroduce them later. I think adding functions determining parameter values will be very useful, and I do not quite see how they should work without information about sender and target. @jougs what do you think?

nestkernel/conn_parameter.h
```diff
@@ -59,18 +61,20 @@ class ConnParameter
...
    * will be ignored except for random parameters.
    */
-  virtual double value_double( index, index, librandom::RngPtr& ) const = 0;
-  virtual long_t value_int( index, index, librandom::RngPtr& ) const = 0;
+  virtual double value_double( thread, librandom::RngPtr& ) const = 0;
+  virtual long_t value_int( thread, librandom::RngPtr& ) const = 0;
+  virtual void skip( thread ) const
+  {
+  }
+  virtual bool is_array() = 0;
```
heplesser: Couldn't this method be `const`?

nestkernel/conn_builder.h
```diff
@@ -119,6 +130,9 @@ class ConnBuilder
   //! dictionaries to pass to connect function, one per thread
   std::vector< DictionaryDatum > param_dicts_;
 
+  //! pointers to connection parameters specified as arrays
+  std::vector< ConnParameter* > skip_array_parameters;
```
heplesser: Since this is a private member, the name should end in an underscore.

nestkernel/conn_builder.cpp
```diff
@@ -99,17 +100,29 @@ nest::ConnBuilder::ConnBuilder( Network& net,
   if ( !default_weight_and_delay_ )
   {
     weight_ = syn_spec->known( names::weight )
-      ? ConnParameter::create( ( *syn_spec )[ names::weight ] )
-      : ConnParameter::create( ( *syn_defaults )[ names::weight ] );
+      ? ConnParameter::create( ( *syn_spec )[ names::weight ], net_.get_num_threads() )
+      : ConnParameter::create( ( *syn_defaults )[ names::weight ], net_.get_num_threads() );
+    // If weight is an array it will be added to a vector of ConnParameters,
+    // ensuring thread-safety of the connection routine.
```
heplesser: This comment does not do much more than restate what the code already tells, particularly the first line. It is also, just as the code, repeated several times. In true Robert C Martin "Clean Code" spirit, we could delegate the task here:
```
register_parameters_requiring_skipping_( *weight_ );
```
with
```
void register_parameters_requiring_skipping_( ConnParam& param )
{
  if ( param.is_array() )
  {
    skip_array_parameters_.push_back( &param );
  }
}
```

And (again in total RC Martin mode), one could change `skip_array_parameters` to `parameters_requiring_skipping_`.

jougs:
Once the .pyc files are gone, I'm also happy and :+1: for merging! Many thanks for the nice code and the constructive work :-)

@heplesser: regarding the `sgid` and `tgid` arguments in the `value_{double,int}()` interface: I also think they'll be important in the future. But I also strongly agree with @hannahbos that they are currently not supported and the interface should not be cluttered with arguments that might be needed at some point in the future. We can re-add them once we have a concrete use case.

Maybe the removal of those arguments should go into a separate PR, though, as it is technically unrelated to this one? @tammoippen, what do you think?

heplesser:
I agree with @jougs and @hannahbos, +1 for merging (how do you add thumbs-up icons?). If removing the `sgid` and `tgid` arguments requires any more work than moving a small number of  commits as is from this PR to a new PR, I would not prioritise PR tidying and merge this PR as is.

hannahbos:
@jougs, @heplesser, @tammoippen  thanks for the review. I removed the arguments sgid and tgid separately in PR #83 . However, due to non trivial dependencies, I left the associated commits in this PR.

jougs:
@heplesser: You can get thumbs up with `:+1:` and thumbs down with `:-1:`. Just colon and waiting gives you a completion list with some weird things in it.

jougs:
@hannahbos: Thanks for your new revision. This is really excellent work!
+1 and merging.

jougs:
The revert from #84 mentioned above was undone using `git reset --hard`, as it turned out that we actually want #66 merged and not #83. Sorry for any inconveniences caused by this.

PR #68, opened 2015-07-20
fix invalid string formatting for python 2.6

luis-rr:
This creates a crash when trying to use deprecated methods instead of just printing a warning


tammoippen:
The changes look good for me. +1 for merge.

jougs:
Also +1. Merging immediately without copyright agreement due to the minor size of the changeset.

PR #69, opened 2015-07-21
Warn about too few local neurons for recording in hpc_benchmark.sli [ci skip] 

tammoippen:
When running the hpc_benchmark.sli script with MPI and the number of local neurons is smaller than the number of recorded neurons, this results in an undescriptive out of range exception. With this PR a descriptive warning message is printed and the number of recorded neurons is reduced.

### DIFF
--- a/examples/nest/hpc_benchmark.sli
+++ b/examples/nest/hpc_benchmark.sli
@@ -244,11 +244,11 @@ myrng rdevdict /normal get CreateRDV /normal_dv Set
 
     local_neurons Nrec lt
     {
-      M_WARNING (BuildNetwork)
+      M_ERROR (BuildNetwork)
       (Spikes can only be recorded from local neurons, but the number of local 
 neurons is smaller than the number of neurons spikes should be recorded from. 
-Hence, the number of recorded neurons is reduced.) message
-      /Nrec local_neurons def
+Aborting the simulation!) message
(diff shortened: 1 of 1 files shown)


heplesser:
Thanks for adding this check. 

I am uncertain about the risk of changing the number of neurons recorded from, particularly when running via a queuing system. Unless the user explicitly inspects the output file from the run (and who does that when the exit code is 0?), the user will not see the warning and will analyse the recorded spike trains assuming that the coded number of neurons was recorded. That could lead to quite some confusion.

Therefore, I think it would be better to either

 1. abort the simulation
 1. turn off recording entirely

Alternative 2 has the advantage that the user notices that the recording program was adjusted, but this might waste valuable core hours on simulations not useful to the user. Thus, following the "explicit is better than implicit" rule, I would suggest to abort. In that case the test should be performed as soon as all neurons have been created, about line 198.

jougs:
I think this is a very useful addition to the example, but I also agree with @heplesser that the script should rather abort with a meaningful error message than just printing a warning and run on happily but recording from fewer neurons than specified by the user.

You can exit NEST with a non-zero exit code (including abortion of all connected MPI processes) using
```
1 quit_i
```

jougs:
Travis is happy and all issues of the reviewers have been addressed. Thus merging.

Issue #56, opened 2015-07-01
Revise the build setup for user-defined modules (MyModule)


heplesser:
Once #28 is fixed, user-defined modules based on `MyModule` should work on all architectures, including BlueGene. At the same time, it is clear that our way of building such modules is not the most convenient one and breaks with some good practices for such modules. We should therefore revise the build mechanism. This is a follow-up to trac.526.

Note that the philosophy of trac.526 was to move entirely to dynamically loaded modules to be loaded by `Install`. At least for BlueGene, this not feasible/advisable, there we need statically linked libraries. But the build process in that case might still be improved, e.g., by making the main NEST build process build also user-defined modules when they are linked as static libraries.

For script compatibility, we should make `Install` a no-op in case the pertaining module has been linked in.

See also comments on #28 and code changes in #55.

jougs:
See also #213.

tammoippen:
With #213, modules can be build statically, dynamically or as plug-in ( `Install` mechanism) - also on BlueGene. But you still get an error, when the module is linked dynamically and you load it again with an `Install`.

I have not tried it yet, but there is a possibility, to have `cmake` build sub-cmake-projects just by adding it as a subdirectory ([here](https://stackoverflow.com/questions/35732776/using-cmake-to-create-main-and-sub-projects-and-be-able-to-compile-them-individ/35865768)).

tammoippen:
@jougs @heplesser @apeyser 
Dynamic loading of external user modules is the only reason NEST depends on `libltdl`. This leads to inconveniences, if the library is not present by default, e.g. on JUQUEEN, see [this comment](https://github.com/nest/nest-simulator/issues/390#issuecomment-224540108) in #390. Further it increases the code / project complexity:
* The `DynamicLoaderModule` is responsible for locating, loading and maintaining the modules.
* The script `extras/nest-config.in` must be filled correctly with compiler information, flags, include directories, linked libraries, s.t. the user module builds correctly.
* The user module has to initialize correctly. It can be build dynamically or statically. If build dynamically, it can be loaded dynamically by `Install()` or linked directly into NEST. If build statically, it must be linked directly into nest. The initialisation is different in every case:
  * **Dynamic / Install()** `MyModule` has to create a global instance named `mymodule_LTX_mod`. The `libltdl` will look for this instance in a dynamic loaded module when calling `Install('mymodule')`. Then the `DynamicLoaderModule` will call `init()` in this instance and the module is loaded.
[...]

PR #70, opened 2015-07-21
Some improvements to module infrastructure

tammoippen:
Regards #56 and trac.526.

With this PR the user of modules does not have to set `LD_LIBRARY_PATH` anymore. When modules are opened dynamically, NEST first looks in the user-defined search path, which is set to the standard install directory for modules 

    SLI_PREFIX "/lib/nest"
Afterwards it still looks for modules in the libltdl's search path (`LTDL_LIBRARY_PATH `) and the system library search path (`LD_LIBRARY_PATH`).

Further, this PR fixes some formatting and variable assignment in `configure.ac` .

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -730,7 +730,7 @@ SLI_USER_MODULES=""
 # SLI_USER_MODULES_LDADD collects what needs to be added to LDADD
 SLI_USER_MODULES_LDADD=""
 
-AC_ARG_WITH(modules,[  --with-modules=MODULES  external modules to be linked in],
+AC_ARG_WITH(modules,[  --with-modules=MODULES external modules to be linked in],
 [
 if test "$withval" != no; then
   for mod in $withval; do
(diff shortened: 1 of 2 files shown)


heplesser:
This looks good to me. Given that module loading is not tested by the testsuite at present, could you briefly document how you have tested that this works?

Otherwise +1.

tammoippen:
I manually build and installed the module `MyModule` and then started a fresh installation of `nest`:

    SLI ] (mymodule) Install

The module installed without setting `LD_LIBRARY_PATH` beforehand. I tested the behaviour on OS X 10.10.4 and Ubuntu 14.04.3 LTS.

configure.ac
```diff
@@ -730,11 +730,11 @@ SLI_USER_MODULES=""
 # SLI_USER_MODULES_LDADD collects what needs to be added to LDADD
 SLI_USER_MODULES_LDADD=""
 
-AC_ARG_WITH(modules,[  --with-modules = external modules to be linked in],
+AC_ARG_WITH(modules,[  --with-modules=MODULES  external modules to be linked in],
```
jougs: There's a space too much

jougs:
Other than the additional space mentioned above, I'm happy with this. +1 for merging.

PR #50, opened 2015-06-25
Fix 427 document noise generator [ci skip]

heplesser:
I have added an Jupyter [IPython] notebook under `docs/models` discussing the properties of the `noise_generator` in detail. This should solve trac.427. I think that `doc/models` over time could become a location for similar in-depth descriptions of other models.

Suggested reviewers: @mhelias and @jschuecker

PR #71, opened 2015-07-21
Add propagator fot tau_m==tau_syn in iaf dynamics.

jschuecker:
This fixes the ticket for the 'iaf_psc_alpha' neuron model. In the singular case the degenerate solution is used and in case of numeric instabilities a switching criterion is used. The latter is described in the documentation.
See https://trac.nest-initiative.org/trac/ticket/618 for the detailed discussion. 

I suggest @heplesser @mhelias as code reviewers.

### DIFF
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -1,5 +1,5 @@
 /*
- *  propagators.cpp
+ *  propagator_stability.cpp
  *
  *  This file is part of NEST.
  *
@@ -20,20 +20,22 @@
  *
  */
 
-#include "propagators.h"
+#include "propagator_stability.h"
 #include <cmath>
 #include "numerics.h"
 
 double
 propagator_31( double tau_syn, double tau, double C, double h )
 {
+  const double P31_linear =
+    1 / ( 3. * C * tau * tau ) * h * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
   const double P31 = 1 / C * ( std::exp( -h / tau_syn ) * numerics::expm1( -h / tau + h / tau_syn )
-                                 / tau * ( tau / tau_syn - 1 )
-                               - h * std::exp( -h / tau_syn ) ) / tau * ( -1 - -tau / tau_syn );
-
+                                 / ( tau / tau_syn - 1 ) * tau
+                               - h * std::exp( -h / tau_syn ) ) / ( -1 - -tau / tau_syn ) * tau;
   const double P31_singular = h * h / 2 / C * std::exp( -h / tau );
+  const double dev_P31 = std::abs( P31 - P31_singular );
 
-  if ( std::abs( tau - tau_syn ) < std::pow( 1., -15 ) )
+  if ( tau == tau_syn or ( std::abs( tau - tau_syn ) < 0.1 and dev_P31 > 2 * P31_linear ) )
   {
     return P31_singular;
   }
@@ -46,11 +48,15 @@ propagator_31( double tau_syn, double tau, double C, double h )
 double
 propagator_32( double tau_syn, double tau, double C, double h )
 {
+  const double P32_linear =
+    1 / ( 2. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
   const double P32_singular = h / C * std::exp( -h / tau );
   const double P32 = -tau / ( C * ( 1 - tau / tau_syn ) ) * std::exp( -h / tau_syn )
     * numerics::expm1( h * ( 1 / tau_syn - 1 / tau ) );
 
-  if ( std::abs( tau - tau_syn ) < std::pow( 1., -15 ) )
+  const double dev_P32 = std::abs( P32 - P32_singular );
+
+  if ( tau == tau_syn or ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P32_linear ) )
   {
     return P32_singular;
   }
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -51,7 +51,7 @@ double
 propagator_31( double tau_syn, double tau, double C, double h )
 {
   const double P31_linear =
-    h / ( 3. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
+    1 / ( 3. * C * tau * tau ) * h * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
   const double P31 =
     1 / C * ( ( std::exp( -h / tau_syn ) - std::exp( -h / tau ) ) / ( -1 / tau_syn - -1 / tau )
               - h * std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -54,7 +54,7 @@ propagator_32( double tau_syn, double tau, double C, double h )
     1 / C * ( std::exp( -h / tau ) - std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
   const double dev_P32 = std::abs( P32 - P32_singular );
 
-  if ( tau == tau_syn or ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P31_linear ) )
+  if ( tau == tau_syn or ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P32_linear ) )
   {
     return P32_singular;
   }
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -1,0 +1,73 @@
+/*
+ *  propagator_stability.cpp
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "propagator_stability.h"
+#include <cmath>
+
+double
+propagator_32( double tau_syn, double tau, double C, double h )
+{
+  const double P32_linear =
+    1 / ( 2. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
+  const double P32_singular = h / C * std::exp( -h / tau );
+  const double P32 =
+    1 / C * ( std::exp( -h / tau ) - std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
+  const double dev_P32 = std::abs( P32 - P32_singular );
+
+  if ( tau == tau_syn )
+  {
+    return P32_singular;
+  }
+  else if ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P32_linear )
+  {
+    return P32_singular;
+  }
+  else
+  {
+    return P32;
+  }
+}
+
+double
+propagator_31( double tau_syn, double tau, double C, double h )
+{
+  const double P31_linear =
+    h / ( 3. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
+  const double P31 =
+    1 / C * ( ( std::exp( -h / tau_syn ) - std::exp( -h / tau ) ) / ( -1 / tau_syn - -1 / tau )
+              - h * std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
+  const double P31_singular = h * h / 2 / C * std::exp( -h / tau );
+  const double dev_P31 = std::abs( P31 - P31_singular );
+
+  if ( tau == tau_syn )
+  {
+    return P31_singular;
+  }
+  else if ( std::abs( tau - tau_syn ) < 0.1 and dev_P31 > 2 * P31_linear )
+  {
+    return P31_singular;
+  }
+  else
+  {
+    return P31;
+  }
+}
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -24,30 +24,6 @@
 #include <cmath>
 
 double
-propagator_32( double tau_syn, double tau, double C, double h )
-{
-  const double P32_linear =
-    1 / ( 2. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
-  const double P32_singular = h / C * std::exp( -h / tau );
-  const double P32 =
-    1 / C * ( std::exp( -h / tau ) - std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
-  const double dev_P32 = std::abs( P32 - P32_singular );
-
-  if ( tau == tau_syn )
-  {
-    return P32_singular;
-  }
-  else if ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P32_linear )
-  {
-    return P32_singular;
-  }
-  else
-  {
-    return P32;
-  }
-}
-
-double
 propagator_31( double tau_syn, double tau, double C, double h )
 {
   const double P31_linear =
@@ -69,5 +45,29 @@ propagator_31( double tau_syn, double tau, double C, double h )
   else
   {
     return P31;
+  }
+}
+
+double
+propagator_32( double tau_syn, double tau, double C, double h )
+{
+  const double P32_linear =
+    1 / ( 2. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
+  const double P32_singular = h / C * std::exp( -h / tau );
+  const double P32 =
+    1 / C * ( std::exp( -h / tau ) - std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
+  const double dev_P32 = std::abs( P32 - P32_singular );
+
+  if ( tau == tau_syn )
+  {
+    return P32_singular;
+  }
+  else if ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P32_linear )
+  {
+    return P32_singular;
+  }
+  else
+  {
+    return P32;
   }
 }
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -35,7 +35,8 @@ propagator_31( double tau_syn, double tau, double C, double h )
   const double P31_singular = h * h / 2 / C * std::exp( -h / tau );
   const double dev_P31 = std::abs( P31 - P31_singular );
 
-  if ( tau == tau_syn or ( std::abs( tau - tau_syn ) < 0.1 and dev_P31 > 2 * P31_linear ) )
+  if ( tau == tau_syn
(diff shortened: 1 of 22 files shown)


ingablundell:
Ich hab noch ein paar Fragen:

1. Warum verwenden wir fuer den instabilen Fall den singulaeren Term (P32_ex_linear), wo Tau verwendet wird, und nicht zum Beispiel den Mittelwert aus den beiden (oder irgendetwas in der Art)?

2. Bei der Taylorentwicklung kommt fuer P32_lin meiner Meinung nach der Faktor 1/2 zu viel vor. Kann aber auch sehr gut sein, dass ich mich verrechnet habe.

3. Wie ist das denn mit der Dokumentation gedacht? Wo soll was stehen? Bei Bedarf kann ich das auch schreiben.

Fuer NESTML ist das auf jeden Fall so alles ohne Probleme machbar!

jschuecker:
1. This would indeed be a refinement of this case. Another possibility would be to take also the linear term into account.
2. I performed the calculations for the propagators using sympy, but of course there might be something wrong. 
3. The switching criterion is described in the documentation, but I agree, this might be to short to fully explain the strategy. Do you have an idea, where we can put this in more detail?

We can discuss these issues in our upcoming meeting.

ingablundell:
And in english and more detail:

Looks very good!

I have a few questions though:
 
1. Why do you use the singular matrix entry (P32_ex_linear) in which Tau is used instead of using some sort of mean value?

2. For the limit of P32 in Tau==Tau syn, h/C*e^(-h/Tau) makes sense and is therefor the first part of the Taylor series. I calculated the second part of the series (the linear part) and don't get the factor 1/2. Maybe I made a mistake though.

3. Please add documentation in the way it was done in #50. I think it should include why the Taylor series can be written/ is T(P32)(Tau_syn, Tau)= P32_0+P32_lin +O(2) (including the limit of P32 in Tau==Tau syn, so why P32_0 it is a continuous continuation of P32 in Tau=Tau_syn). And an explanation of wyh the error is of order norm(P_31_1,P_32_1). I'm happy to write something if you want to.

It is not a problem to do it this way for NESTML.

models/iaf_psc_alpha.cpp
```diff
@@ -240,16 +236,85 @@ iaf_psc_alpha::calibrate()
...
-  V_.P32_ex_ = 1 / P_.C_ * ( V_.P33_ - V_.P11_ex_ ) / ( -1 / P_.Tau_ - -1 / P_.tau_ex_ );
-
   V_.P21_in_ = h * V_.P11_in_;
-  V_.P31_in_ = 1 / P_.C_ * ( ( V_.P11_in_ - V_.P33_ ) / ( -1 / P_.tau_in_ - -1 / P_.Tau_ )
-                             - h * V_.P11_in_ ) / ( -1 / P_.Tau_ - -1 / P_.tau_in_ );
-  V_.P32_in_ = 1 / P_.C_ * ( V_.P33_ - V_.P11_in_ ) / ( -1 / P_.Tau_ - -1 / P_.tau_in_ );
+
+  // exact propagator, linear approximation and singular case for excitatory synaptic event
+
+  const double P32_ex_linear = 1 / 2. / P_.C_ / P_.Tau_ / P_.Tau_ * h * h * ( P_.tau_ex_ - P_.Tau_ )
```
heplesser: I find this long sequence of divisions difficult to read, I think `1. / ( 2 * P_.C_ * P_.Tau_ * P_.Tau_ ) * h * h ...` would be clearer.

models/iaf_psc_alpha.cpp
```diff
@@ -240,16 +236,85 @@ iaf_psc_alpha::calibrate()
...
+  const double P32_ex = 1 / P_.C_ * ( V_.P33_ - V_.P11_ex_ ) / ( -1 / P_.Tau_ - -1 / P_.tau_ex_ );
+  const double P31_ex_linear = h * 2 / 3. * P32_ex_linear;
+  const double P31_ex = 1 / P_.C_ * ( ( V_.P11_ex_ - V_.P33_ ) / ( -1 / P_.tau_ex_ - -1 / P_.Tau_ )
+                                      - h * V_.P11_ex_ ) / ( -1 / P_.Tau_ - -1 / P_.tau_ex_ );
+  const double P31_ex_singular = h * h / 2 / P_.C_ * V_.P33_;
+  const double dev_P31_ex = std::fabs( P31_ex - P31_ex_singular );
+  const double dev_P32_ex = std::fabs( P32_ex - P32_ex_singular );
+  const double dev_norm_ex = std::sqrt( std::pow( dev_P31_ex, 2 ) + std::pow( dev_P32_ex, 2 ) );
+  const double error_linear_ex = std::sqrt(
+    std::pow( std::fabs( P32_ex_linear ), 2 ) + std::pow( std::fabs( P31_ex_linear ), 2 ) );
```
heplesser: Do we really need to compute the roots here? If we didn't, we would just have to replace the factor 2 by 4 in the test below and have exactly the same criterion. 

I also think we could drop the `fabs()`,  since we are squaring anyways.

models/iaf_psc_alpha.cpp
```diff
@@ -240,16 +236,85 @@ iaf_psc_alpha::calibrate()
...
+    V_.P31_ex_ = P31_ex_singular;
+    V_.P32_ex_ = P32_ex_singular;
+  }
+  else
+  {
+    V_.P32_ex_ = P32_ex;
+    V_.P31_ex_ = P31_ex;
+  }
+
+  // exact propagator, linear approximation and singular case for inhibitory synaptic event
```
heplesser: We are duplicating a lot of code here. Could we delegate this to a function that computes the P31 and P32 matrix elements?

models/iaf_psc_alpha.h
```diff
@@ -96,10 +96,20 @@ Name: iaf_psc_alpha - Leaky integrate-and-fire neuron model.
...
+P_31 = P_31_0 + P_31_lin + O(2) ; P_32 = P_32_0 + P_32_lin + O(2).
+If tau_m == tau_syn_{ex,in} the singular solutions P_31_0 and P_32_0 are used.
+If tau_m and tau_syn_{ex,in} are close to each other we calculate the
+deviation of the propagator to the singular case
+P_31_dev = P_31 - P_31_0 ; P_32_dev = P_32 - P_32_0
+and if
+||(P_31_dev,P_32_dev)|| > 2 * ||(P_31_1,P_32_1)||
+we assume that the deviations in the propagators are due to numeric
+instabilities and we use the singular solutions. Here || || is the euclidean norm.
+With this criterion the maximal error is of order ||(P_31_1,P_32_1)||.
```
heplesser: I agree with @ingablundell's comment in the details should be documented in an IPython notebook as in #50. The online documentation would then just need a single sentence saying that the correct method is chosen automatically and point to the notebook. Maths is not very readable in our documentation format.

heplesser:
@jschuecker @ingablundell Very nice work so far! With proper #50-style documentation and delegation of computation of the matrix elements to a helper function to avoid code repetition, this should be fine for integration. I did not check the mathematics, though. 

NB: In moving the code computing the P31 and P32 matrix elements to a helper function, we might consider putting it into a separate compilable (part of `libnestutil` maybe?) so that we can use it in all models that have alpha synapses. Otherwise, we'll be copying a lot of code around.

@jschuecker I think you had a test/example demonstrating continuity as one crosses the border to the singular case. Could you include this?

jschuecker:
@heplesser: @ingablundell  and I like the idea of moving the calculation of the matrix elements to libnestutil, since the singularity appearing in the exponential current neuron (P32) is actually the same as P32 of the alpha-shaped current case, so they can share the same function. In consequence we think that we should apply the stability criterion on each single propagator element, rather than on the norm of the all relevant elements, as we have it now.

jschuecker:
I moved the calculations of the unstable propagators to libnestutil/propagator_stability, maybe the name can be improved. The stability criterion now checks for each element individually, so that it can be used by neurons with alpha- and exponentially-shaped currents. 
A remaining issue is how to inform the user about the chosen solution. In libnestutil we do not have access to the net_ class, so we cannot simply send a message. We could return this information to the model using a map, but this would increase the number of lines considerably. @heplesser and @ingablundell, do you have ideas on this issue?
In addition, with the help of @ingablundell I created an ipython notebook explaining the stability criterion and showing the limit of the non-singular case into the singular one. Please have a look.

libnestutil/propagator_stability.cpp
```diff
@@ -0,0 +1,73 @@
...
+  {
+    return P32;
+  }
+}
+
+double
+propagator_31( double tau_syn, double tau, double C, double h )
+{
+  const double P31_linear =
+    h / ( 3. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
```
heplesser: It seems that this line contains h^3. Why is the first `h` separate from the next two?

libnestutil/propagator_stability.h
```diff
@@ -0,0 +1,29 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef PROPAGATOR_STABILITY_H
+#define PROPAGATOR_STABILITY_H
+
+double propagator_32( double tau_syn, double tau, double C, double h );
+double propagator_31( double tau_syn, double tau, double C, double h );
```
heplesser: The functions should be ordered alphabetically, i.e., `propagator_31` first. The same applies to the definitions in the cpp-file.

models/iaf_psc_alpha.h
```diff
@@ -96,10 +96,11 @@ Name: iaf_psc_alpha - Leaky integrate-and-fire neuron model.
   I_e        double - Constant external input current in pA.
   V_min      double - Absolute lower value for the membrane potential.
 
-Remarks:
-  tau_m != tau_syn_{ex,in} is required by the current implementation to avoid a
-  degenerate case of the ODE describing the model [1]. For very similar values,
-  numerics will be unstable.
+Note:
```
heplesser: This should be `Remarks:` not `Note:`.  Only `Remarks` is a keyword in the NEST documentation system.

models/iaf_psc_alpha.h
```diff
@@ -96,10 +96,11 @@ Name: iaf_psc_alpha - Leaky integrate-and-fire neuron model.
...
+For very similar values of tau_m and tau_syn_{ex,in} the
+evaluation of the propagator entries P_31 and P_32 is numerically
+unstable. For details, please see IAF_Neruons_Singularity.ipynb in the
+NEST source code (docs/model_details)
```
heplesser: I would write this comment more positive and informative. As it is now, the user may become uncertain about wht precisely happens if the taus are very similar. Here a suggestion:

If tau_m is very close to tau_syn_ex or tau_syn_in, the model will numerically treat the model as if tau_m is equal to tau_syn_ex or tau_syn_in, respectively, to avoid numerical instabilities. The status dictionary provides information about which propagators are used through the /propagator_info entry. For details, please see IAF_Neruons_Singularity.ipynb in the NEST source code (docs/model_details).

heplesser:
@jschuecker Nice work! I added a few comments directly above, and I have a few more here.

I have some trouble with the IPython notebook. I get an error message when computing `PB`; Python tells me it cannot compute the eigenvector for one of the eigenvalues. I also cannot reproduce the figures without cloning @jschuecker's repo and building your version. Could you add the figures you created in the notebook to the discussion thread here? Would just make life a little easier for me.

I am also not sure whether we should have `propagator_{31,32}` as plain functions. I think it might be better to create a class and have them as static methods in the class. That would make it clearer where this code belongs. The class could be called `IAFPropagatorElements` or something like that. @jougs, what to you think about this from a software design perspective?

In such a class, we could then also define an enumeration, e.g. `enum ElementType = {Standard, Singular};` (not 100% sure about the C++ syntax) and then have the `propagator_*` methods return the `ElementType` via call-by-reference. Information can then be stored in the neuron and returned via the status dictionary (see also comment on online doc above). That should probably be returned as a dict, so that we can return info for the ex and inh synapses separately.

I don't think issuing messages via the network would be a good idea, since the message would be repeated for every single neuron.

libnestutil/propagator_stability.cpp
```diff
@@ -0,0 +1,73 @@
...
+  const double P32_singular = h / C * std::exp( -h / tau );
+  const double P32 =
+    1 / C * ( std::exp( -h / tau ) - std::exp( -h / tau_syn ) ) / ( -1 / tau - -1 / tau_syn );
+  const double dev_P32 = std::abs( P32 - P32_singular );
+
+  if ( tau == tau_syn )
+  {
+    return P32_singular;
+  }
+  else if ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * P32_linear )
```
heplesser: The if-else-if here is unnecessary and, in my opinion, confusing rather than clarifying. You should combine the condition for using the singular value in a Boolean or.

jschuecker:
@heplesser, thank you for fast review!

Actually I get the same error as @heplesser with the ipython notebook. I anyway pushed this, since it was working with @ingablundell 's version of sympy. I will track down the problem with @ingablundell. 

iaf_neurons_singularity_fig1
iaf_neurons_singularity_fig2

jschuecker:
I fixed the minor issues pointed out in the review by @heplesser.
A remaining issue is how to inform the user about the propagator which is actually used. This is connected to the question @heplesser raised, namely whether we want to have the propagator function as plain or defined in a class (see above). @jougs, what do you think about this from a software design perspective?

jschuecker:
As @heplesser pointed out, there is also one issue left with the ipython notebook. Python cannot compute the eigenvector for one of the eigenvalue in case of the exponential matrix. This seems to be a bug in sympy, actually when one declares tau_m to be positive and tau_s to be negative, i.e.
tau_m = sp.symbols('tau_m', positive=True)
tau_s = sp.symbols('tau_s', positive=False)
it works. Of course this is not a nice solution, the best way would be to declare tau_s != tau_m, but I did not figure out how to do this. @heplesser or @ingablundell , do you have an idea?

libnestutil/propagator_stability.h
```diff
@@ -0,0 +1,29 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef PROPAGATOR_STABILITY_H
+#define PROPAGATOR_STABILITY_H
+
+double propagator_31( double tau_syn, double tau, double C, double h );
+double propagator_32( double tau_syn, double tau, double C, double h );
```
jougs: @ingablundell: Do we have to generate this as well if the dimensionality of the matrix changes? Are `_31` and `_32` good names or is there a more generic name? At least we should be aware that this might be something to consider in the context of NESTML.

models/iaf_psc_alpha.cpp
```diff
@@ -230,6 +228,7 @@ iaf_psc_alpha::calibrate()
 
   const double h = Time::get_resolution().get_ms();
 
+
```
jougs: Please remove additional empty line.

jougs:
@heplesser, @jschuecker: I like the current code for its simplicity and think it does not have to go into a class. Most likely, we'll anyway have to restructure this again once NESTML is fully in place and I think we don't have to go the extra mile at the moment.

I am not completely sure about the reporting issue. Did the behavior of the model change compared to previous versions if the same parameters are used? Or are we just supporting more cases (i.e. tau_m == tau_s) now? In the latter case, I think we don't have to report anything. If we are doing something different, we should report. I'll think more about this once I know the answer to my questions!

jschuecker:
@jougs 
For very similar values of tau_m and tau_s, the numeric stability of the propagator is tested and if it is numerically unstable, the singular propagator (tau_m==tau_s) is used instead. This is a new behaviour, and thus we need to report the used propagator.

jougs:
@jschuecker: Thanks for explaining. My question was more if the *results* will differ in cases, where the new propagator is used? If that's the case, we definitely have to report (probably using a static boolean flag in the neuron class indicating if the report has already been printed).

If the results are the same even with the new propagator, I think that the note in the documentation and the Ipython Notebook suffices.

jschuecker:
@jougs: Sorry for being imprecise, that is exactly what I meant. The results will differ, since the two propagators are not the same. The new one is numerically stable but an approximation, while the old one is numerically unstable and therefore led to wrong results.  Note that this happens only for very very similar values of the time constants. I think we only have to report in this rare situations, where results will differ. I will implement a suggestion.

libnestutil/propagator_stability.h
heplesser: @jougs @ingablundell The names are historic and little informative. `amat2_psc_exp` shows the limitations, with elements from `P00_` to `P77_`. But I suggest we solve this in the context of NESTML, and for now leave working code at peace.

heplesser:
@jschuecker @jougs I think we can drop the information. In the past, `tau_m == tau_syn` was prohibited, so that it works is new. I assume that most people who hit this problem made the two constants differ by at least 0.1 ms, and in that case they won't be affected by the new code. If people used smaller differences, they might observe small changes, which in a chaotic system can lead to very different firing patterns. But I think these cases are best addressed by a clear and prominent warning in the release notes, and maybe in the online documentation for the model(s).

jougs:
@heplesser: Thanks for chiming in. I fully agree with you on your statements about issuing a warning and the naming of the variables, which means the PR is in principle good to go as it is.
@jschuecker: Can you please adapt the other neuron models and rename the pull request accordingly? I'm all :+1: for merging this before the release end of the week.

jschuecker:
@jougs: Ok!

jschuecker:
@jougs, @heplesser : Going through the several neuron models, I figured out the following: The same propagator were implemented in a different way 'iaf_psc_exp' and in 'iaf_psc_alpha'. The one in 'iaf_psc_exp' in numerically much more stable, probably due to the fact that an exponential and a time constants were factored out. Adapting the critical propagators accordingly also in the 'alpha' case, stability problems only occur if time constant differ by around 10^(-10). I am wondering, If we need to implement a stability criterion at all, since these cases probably never appear and it could be confusing. We could rather just add the singular case, and switch if tau_m == tau_s. Of course, it would still be nice to have the propagators in libnestutil, so that the neuron models can share them, or should we leave this for NESTML. What do you think?

heplesser:
@jschuecker I think it is mainly the exponential being factored out that makes the difference, see also the graph below. But we can still do one better on the `iaf_psc_exp` implementation, by replacing `1-exp(...)` with `-expm1(...)`. I'd suggest to use that in the implementation in libnestutil.

Testing floating point numbers for equality is never a particularly good idea, and if users should randomize time constants, one could end up with very close but not equal values. So we should have some $|\tau_m - \tau_s| < \eps$ criterium. Given that your analysis shows stability problems for differences below $10^{-10}$, I would use that or maybe, to be on the safe side $\eps=10^{-9}$ as criterium. We could then just state in the docs that we use the singular propagator if differences in time constants are below 1ps.

@jougs I we make this change, we should also note in the release notes that we improved the numerics of `iaf_psc_alpha` models and that that might lead to slightly different results if $\tau_m\approx\tau_s$.

screen shot 2015-09-25 at 05 49 59

jougs:
I happily put such a sentence in the release notes. They will be prepared from later today on in https://trac.nest-initiative.org/trac/wiki/ReleaseNotes28. If you'd like something to appear there, feel free to add it.

jschuecker:
@heplesser: My previous statement was based on a bug in my implementation. The stability problems are only slightly reduced by factoring off the exponential, so we still need the stability criterion. I now implemented the propagators following your suggestion using the expm1. So far the 'iaf_psc_exp', the 'iaf_psc_alpha' and the 'iaf_psc_alpha_multisynapse' use the common propagators. I will now go through the remaining models.

jschuecker:
I further edited the notebook. I took out the exponential case due to the bug in sympy. Anyway it was not needed, since the singularity is explained in the alpha case.

heplesser:
@jschuecker Great work! Thank you for working through all the models!

But you added a bug to the new stability criterium: we must test agains the absolute value of the linear term, see also comments in code.

Another thing is that in the `propagator_{31,32}()` functions, the computation of the `Pxy` and `Pxy_linear` terms contains `tau_syn - tau` in the denominator, which becomes zero for equality. As long as we have IEEE758, this will result in an `inf` result and keep us safe, otherwise the user will get a division by zero error. But since IEEE758 numerics is a prerequisite for NEST support, we do not need to worry much about this.

jschuecker:
@heplesser: Thank you for spotting the bug, I fixed it.

jougs:
Very nice work indeed. I'm :+1: from my side. @heplesser?

heplesser:
@jschuecker @jougs :+1: from the Thalys between Liége and Bruxelles on beautiful Sunday morning, travelling climate friendly by train to present NEST 2.8.0 at the HBP Summit.

PR #73, opened 2015-08-04
Replace valarray with vectors

tammoippen:
NEST uses valarray only in RingBuffers to store spike data.
It does not perform any arithmetic operations on the Buffers,
hence neglecting the purpose of valarray completely. For convenience
and standardization of the sources I removed the usage of valarray
and replaced it with vector, if appropriate.

* Replace valarray with vector in RingBuffers.
* Remove sli/array.h as it is not used anywhere.
* Remove all other occurrences of valarray in sli/ .


heplesser:
@tammoippen @lekshmideepu @apeyser In this case, all Travis builds seem to have failed with a timeuout in the syntax checking part. Could anyone with Travis competence investigate what is going on?

nestkernel/ring_buffer.cpp
```diff
@@ -33,21 +33,24 @@ nest::RingBuffer::resize()
...
 }
 
 void
 nest::RingBuffer::clear()
 {
-  resize();      // does nothing if size is fine
-  buffer_ = 0.0; // clear all elements
+  resize(); // does nothing if size is fine
+  // clear all elements
+  for ( std::vector< double_t >::iterator it = buffer_.begin(); it != buffer_.end(); ++it )
```
heplesser: Isn't there a copy iterator/operator in the STL that allows one to turn this explicit for-loop into a single line?

nestkernel/ring_buffer.cpp
```diff
@@ -57,15 +60,18 @@ nest::MultRBuffer::resize()
...
+    std::vector< double_t >( size, 0.0 ).swap( buffer_ );
   }
 }
 
 void
 nest::MultRBuffer::clear()
 {
-  buffer_ = 0.0;
+  // clear all elements
+  for ( std::vector< double_t >::iterator it = buffer_.begin(); it != buffer_.end(); ++it )
```
heplesser: See above.

sli/tokenarray.h
```diff
@@ -29,7 +29,7 @@
 #include <typeinfo>
 #include <cstddef>
 #include <vector>
-#include <valarray>
+#include <algorithm>
```
heplesser: Why do you need to add `algorithm` here?

sli/Makefile.am
```diff
@@ -18,7 +18,6 @@ noinst_LTLIBRARIES= libsli.la
 
 libsli_la_SOURCES=\
 		aggregatedatum.h\
-		array.h\
```
heplesser: I am surprised the file is not needed, but if it compiles .... It makes me wonder if there are other examples of dead code around.

heplesser:
Conditional on Travis problems really being Travis problems, not test failures, I am happy with this code, a nice example of cleaning up ghosts of the past. I added a little suggestion for making code even more compact, see above. Conditionally (+1).

sli/tokenarray.h
tammoippen: @heplesser : `TokenArray::rotate` uses the `std::rotate` function from `algorithm`

tammoippen:
TravisCI halts on the static analysis of `cppcheck --enable=all --inconclusive --std=c++03 sli/tokenutils.cc`
on my local machine this takes a fraction of a second... I am not sure, why Travis has problems here. All other tests (including `installcheck`) work out nicely.

tammoippen:
The TravisCI problems are solved by #79.

jougs:
The code looks nice and clean to me. Thus +1 and merging.

Issue #64, opened 2015-07-03
git diff --name-only $TRAVIS_COMMIT_RANGE


apeyser:
@tammoippen: in build.sh, we do
```
file_names=`git diff --name-only $TRAVIS_COMMIT_RANGE`
```
Inside my repo on some pushes, this seems to return incorrect/unreachable references. I haven't yet identified the cause, or a travis bug report on this, but it may be a problem with history rewrites on pushes that it may end up referencing old, non-existent commits.

tammoippen:
Hi @apeyser ,
TRAVIS_COMMIT_RANGE has some problems with history rewrites. I am working on correcting this with  the github api: https://developer.github.com/v3/pulls/#list-pull-requests-files

tammoippen:
I am closing this issue as PR #75 is closed. @apeyser please reopen if the issue persists.

PR #75, opened 2015-08-06
Fix travis commit range issue 64

tammoippen:
Using the `$TRAVIS_COMMIT_RANGE` variable in `build.sh` on the TravisCI has some issues, when some history rewrites happened on your branch (see [here](https://github.com/travis-ci/travis-ci/issues/2668) and issue #64  ). With this PR the issue should be fixed with respect to pull requests (by using the [github api](https://developer.github.com/v3/pulls/#list-pull-requests-files) ). 

Using TravisCI on your own repository still uses `$TRAVIS_COMMIT_RANGE` for commits pushed to it: if history rewrites happened, the script will not be able to get all changed files and thus cannot perform the static code analyses – it will just omit it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,9 @@ before_install:
   # used for building cppcheck-1.69
   - sudo apt-get install -y libpcre3 libpcre3-dev
 
+  # process json output from github api
+  - sudo apt-get install -y jq
+
 install: 
   - cython --version
   - python --version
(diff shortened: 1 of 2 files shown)


sli/fdstreamtest.cc
```diff
@@ -67,7 +67,8 @@ main()
   }
```
heplesser: As @jougs pointed out in another PR, this and the other files with formatting fixes should not be part of a pull request fixing a non-formatting issue.

tammoippen: I will put them into separate PR later. I am still testing with the GitHub API

jougs:
Looks good to me. +1 for merging.

tammoippen:
@heplesser can you please have another look at this PR?

heplesser:
:+1: for merge.

PR #78, opened 2015-08-14
Removed dead link.

jougs:
The information is now at http://www.nest-simulator.org/scheduling-and-simulation-flow/ .


heplesser:
@jougs  Very nice work and +1 for the merge.

Just one suggestion: `max_delay` appears suddenly towards the end of the text and is never properly introduced. E.g., after the sentence "Thus we freeze ...", you could add

"We also freeze the maximum delay `d_max` at this time."

Could you add a sentence about it. Also, in the last sentence you could add a little detail in parenthesis like this: "These settings should be used with care, though: setting the delay extrema too wide without need might lead to decreased performance due to a higher number of update calls and communication cycles (small `min_delay`), or increasing the memory consumption of NEST (large `max_delay`)". 

Buffer size is always `min_delay + max_delay`.

tammoippen:
Nice work! +1 for merge. Merging now.

jougs:
@heplesser: Thanks for the additional comments on how to improve the text. I've incorporated the definition of `max_delay` early on into the text and added references to it wherever possible.

heplesser:
:+1: for merge.

jougs:
JFYI: The merge already happened 3 days ago after you and @tammoippen said +1.

PR #79, opened 2015-08-14
Update cppcheck for static analysis

tammoippen:
The cppcheck version that comes along with Ubuntu 14.04 is 1.61.  Apparently, it is a bit buggy, as it halts when executing the line:

    cppcheck --enable=all --inconclusive --std=c++03 sli/tokenutils.cc

The version 1.69 has no such problems, but is not available from the package system. Hence, we fetch it from the git, set it to the correct version and build it ourselves. 

With this PR, #73 will go green on TravisCI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,10 +61,12 @@ before_install:
   - sudo apt-get install -y python-nose
 
   # static code analysis
-  - sudo apt-get install -y libllvm3.6 clang-format-3.6 vera++ cppcheck
+  - sudo apt-get install -y libllvm3.6 clang-format-3.6 vera++
+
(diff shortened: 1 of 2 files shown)


heplesser:
Good work. +1 for merging. Should we add a reminder-issue that we from time to time should see if 1.69 or later becomes available packaged?

jougs:
This also looks good to me, so +1 and merging. @tammoippen: Could you please add a new issue as @heplesser suggested and reference this PR? Thanks!

tammoippen:
@heplesser I do not think, that ubuntu 14.04 will ever get this as a package, as this is planned for Wily Werewolf (15.10).

PR #83, opened 2015-08-14
Remove arguments in value_()-function of ConnParameter class

hannahbos:
The arguments sgid and tgid describing the indices of potentially inserted parameter containers have been removed since the functionality is not yet implemented as discussed in PR #66 .

PR #84, opened 2015-08-17
Revert "Array connect": #83 needs to be merged first

jougs:
Reverts nest/nest-simulator#66


jougs:
This revert was undone using `git reset --hard`, as it turned out that we actually want #66 merged and not #83. Sorry for any inconveniences caused by this.

Issue #88, opened 2015-08-19
Rename function `test_connect_helpers.test_synapse`


heplesser:
Function `test_connect_helpers.test_synapse` is a helper function, it does not implement a test. We therefore need to change the name to something not containing `test`, so that nosetests won't run it as a test.

This has lead to failing tests for some time (see eg Travis build 141.8), which were not reported due to #89.

PR #90, opened 2015-08-19
Fix #89 by making `do_tests.sh` return non-zero exit code in case of failing tests

### LINKED ISSUE #89: Travis does not report test suite failures
Travis will report success on builds even if the testsuite (`make installcheck`) fails. An example is Travis build 141.8 with one failing PyNEST test; see also #88.

The underlying problem is that `make installcheck` returns exit code 0 even when tests fail.

heplesser:
Until the PR #91 fixing #88 is merged, the Travis tests including Python should fail once this PR is merged. This failure with and success without Python is expected behavior, because #88 makes one PyNEST test fail.

Potential reviewers: @tammoippen @jougs @lekshmideepu

### DIFF
--- a/testsuite/do_tests.sh.in
+++ b/testsuite/do_tests.sh.in
@@ -627,4 +627,8 @@ if test "x${INFO_OS}" = xDarwin ; then
     defaults write com.apple.CrashReporter DialogType "${TEST_CRSTATE}"
 fi
 
-exit 0
+if test ${TEST_FAILED} -gt 0 ; then
+    exit 1
+else
+    exit 0
+fi


tammoippen:
The code looks code for me. :+1:  for merging.

jougs:
I like the basic idea. However, the code could be simplified to just
```
exit $((${TEST_FAILED} > 0))
```
which will exit with the truth value of the comparison.

I'm not sure about the loss of explicitness, though.

heplesser:
@jougs If this had been Python, I'd be 100% in agreement with your suggestion. But if I had read that line of shell code in an unfamiliar context, I would have had a hard time understanding it. Assuming that I still know more shell programming than 80% of NEST developers, I'd argue for "explicit is better than implicit". I suggest we keep elegance for C++ and Python.

jougs:
@heplesser: agreed!

:+1: and merging as is.

PR #91, opened 2015-08-19
Renamed helper function `test_synapse` to `check_synapse` to avoid it…

### LINKED ISSUE #88: Rename function `test_connect_helpers.test_synapse`

heplesser:
… being picked up as a test by nosetests. This should fix #88. This PR should be merged after #90, so that we have a failing test around that test the Travis setup.

Suggested reviewers: @hannah @tammoippen


jougs:
Well spotted. :+1: and immediately merging due to it being only a minor change.

PR #93, opened 2015-08-21
Fix PyNEST test failures

### LINKED ISSUE #6: PyNEST tests using AssertGreater fail
All PyNEST tests using the compatibility versions of `AssertGreater` fail with older versions of Python and SciPy (here Python 2.6.6, SciPy 0.7.2). The error messages look like this:

```
======================================================================
ERROR: testRPortDistribution (nest.tests.test_connect_all_patterns.TestAllToAll)
----------------------------------------------------------------------
  File "/users/eppler/10kcollaps.build/install/lib64/python2.6/site-packages/nest/tests/test_connect_all_to_all.py", line 88, in testRPortDistribution
    self.assertGreater(p, self.pval, 'Chi2 test failed.')
TypeError: <lambda>() takes exactly 3 arguments (4 given)

======================================================================
ERROR: testRPortDistribution (nest.tests.test_connect_all_to_all.TestAllToAll)
----------------------------------------------------------------------
  File "/users/eppler/10kcollaps.build/install/lib64/python2.6/site-packages/nest/tests/test_connect_all_to_all.py", line 88, in testRPortDistribution
    self.assertGreater(p, self.pval, 'Chi2 test failed.')
TypeError: <lambda>() takes exactly 3 arguments (4 given)
```

jougs:
This PR does three things compared to test failures with older Python and NumPy versions:
* Check if the nosetests executable supports the `--with-xunit` switch. If it does not, we use our own test harness instead.
* Remove the error message given to `assertGreater`, as this leads to problems with Python <= 2.6. This fixes #6 .
* Add a simple implementation for `numpy.fill_diagonal` if it is not there (NumPy <= 1.4.0)

### DIFF
--- a/pynest/nest/tests/compatibility.py
+++ b/pynest/nest/tests/compatibility.py
@@ -20,6 +20,7 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 import unittest
+import numpy
 import inspect
 
 
@@ -65,3 +66,10 @@ if not hasattr(unittest.TestCase, 'assertIsInstance'):
 if not hasattr(unittest.TestCase, 'assertGreater'):
     unittest.TestCase.assertGreater = lambda self, a, b, **kwargs: \
         unittest.TestCase.assertTrue(self, a > b, **kwargs)
+
+# NumPy 1.4.0-
+if not hasattr(numpy, 'fill_diagonal'):
+    def fill_diagonal(a, b):
+        diagonal = numpy.arange(a.shape[0])
+        a[diagonal, diagonal] = b
+    numpy.fill_diagonal = fill_diagonal
--- a/pynest/nest/tests/compatibility.py
+++ b/pynest/nest/tests/compatibility.py
@@ -70,6 +70,6 @@ if not hasattr(unittest.TestCase, 'assertGreater'):
 # NumPy 1.4.0-
 if not hasattr(numpy, 'fill_diagonal'):
     def fill_diagonal(a, b):
-        diagonal = numpy.arange(a.shape[0])
-        a[diagonal, diagonal] = b
+        diagonal = [ numpy.arange( min( a.shape ) ) ] * a.ndim
+        a[diagonal] = b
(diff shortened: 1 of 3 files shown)


pynest/nest/tests/compatibility.py
```diff
@@ -65,3 +66,10 @@ def decorator(test_item):
 if not hasattr(unittest.TestCase, 'assertGreater'):
     unittest.TestCase.assertGreater = lambda self, a, b, **kwargs: \
         unittest.TestCase.assertTrue(self, a > b, **kwargs)
+
+# NumPy 1.4.0-
+if not hasattr(numpy, 'fill_diagonal'):
+    def fill_diagonal(a, b):
```
heplesser: Do we need to handle non-square matrices explicitly?

heplesser:
@jougs Fine, but I wonder if we need to protect against special cases, see above. Otherwise, :+1:.

pynest/nest/tests/compatibility.py
```diff
@@ -65,3 +66,10 @@ def decorator(test_item):
...
+        diagonal = numpy.arange(a.shape[0])
+        a[diagonal, diagonal] = b
```
tammoippen: You could write those two lines as follows, to handle matrices of all shapes:

    diagonal = [ numpy.arange( min( numpy.shape( a ) ) ) ] * len( numpy.shape( a ) )
    a[ diagonal ] = b

tammoippen:
The PR looks fine to me. After @jougs commented on the special cases mentioned, I am good to give my :+1: .

jougs:
@heplesser, @tammoippen: I've added a generalized version for arbitrarily shaped matrices.

pynest/nest/tests/compatibility.py
jougs: See new commit cde542f below.

heplesser:
:+1:

tammoippen:
Nice. Merging.

Issue #38, opened 2015-06-18
Feature request: Script for checking coding style compliance


heplesser:
The C++ code style guidelines suggest running three tools (clang-format, vera++, cppcheck) to verify code formatting. The build.sh for Travis script can run these automatically (although there seems to be some trouble).

I would very much appreciate a small script that I could use myself locally to run all code style checks in one go, something like `check_code_style.sh`.

@tammoippen you are maybe the ideal candidate to implement this?

tammoippen:
The script is merged: #95 #96 
Hence closing this issue.

PR #95, opened 2015-08-24
Add a script that checks static analysis from TravisCI locally.

tammoippen:
This solves #38.

It checks, if the selected executables are working and have the correct version. Then it performs the static code analysis from the TravisCI and gives further information to the user.
It has an incremental mode, which allows to check one file at a time or the user can specify the file he wants to be checked. The CLI arguments follow the implementation from `testsuite/do_tests.sh`.

The documentation can be found in #96 .

### DIFF
--- a/extras/check_code_style.sh
+++ b/extras/check_code_style.sh
@@ -1,0 +1,215 @@
+#!/bin/bash
+
+# static code analysis
+START_SHA=master
+END_SHA=HEAD
+
+NEST_SRC=.
+CPPCHECK=cppcheck
+VERA=vera++
+VERA_HOME=/usr/local/lib/vera++
+
+CLANG_FORMAT=clang-format
+INCREMENTAL=false
+
+ONLY_FILE=
+
+#
+# usage [exit_code bad_option]
+#
+usage ()
+{
+    if test $1 -ne 0 ; then
+        echo "Unknown option: $2"
+    fi
+
+    cat <<EOF
+Usage: check_code_style.sh [options ...]"
+
+Setup of Tooling is explained here: 
+    https://nest.github.io/nest-simulator/coding_guidelines_c++
+
+Options:
+
+    --help               Print program options and exit
+    --incremental        Do analysis one file after another.
+    --file=/path/to/file Perform the static analysis on this file only.
+    --git-start=SHA      Enter the default SHA for git to start the diff
+                         (default=master)
+    --git-end=SHA        Enter the default SHA for git to end the diff
+                         (default=HEAD)
+    --nest-src=/path     The base directory for the NEST sources
+                         (default=. assuming you execute check_code_style.sh 
+                         from the base directory.)
+    --cppcheck=exe       Enter the executable that is used for cppcheck.
+                         (default=cppcheck)
+    --clang-format=exe   Enter the executable that is used for clang-format.
+                         (default=clang-format)
+    --vera++=exe         Enter the executable that is used for vera++.
+                         (default=vera++)
+EOF
+
+    exit $1
+}
+
+#
+# bail_out message
+#
+bail_out ()
+{
+    echo "$1"
+    exit 1
+}
+
+while test $# -gt 0 ; do
+    case "$1" in
+        --help)
+            usage 0
+            ;;
+        --incremental)
+            INCREMENTAL=true
+            ;;
+        --file=*)
+            ONLY_FILE="$( echo "$1" | sed 's/^--file=//' )"
+            if [ ! -f $ONLY_FILE ]; then
+              bail_out "This is not a valid file: $ONLY_FILE"
+            fi
+            ;;
+        --git-start=*)
+            START_SHA="$( echo "$1" | sed 's/^--git-start=//' )"
+            ;;
+        --git-end=*)
+            END_SHA="$( echo "$1" | sed 's/^--git-end=//' )"
+            ;;
+        --nest-src=*)
+            NEST_SRC="$( echo "$1" | sed 's/^--nest-src=//' )"
+            ;;
+        --cppcheck=*)
+            CPPCHECK="$( echo "$1" | sed 's/^--cppcheck=//' )"
+            ;;
+        --vera++=*)
+            VERA="$( echo "$1" | sed 's/^--vera++=//' )"
+            ;;
+        --clang-format=*)
+            CLANG_FORMAT="$( echo "$1" | sed 's/^--clang-format=//' )"
+            ;;
+        *)
+            usage 1 "$1"
+            ;;
+    esac
+    shift
+done
+
+#
+# sed has different syntax for extended regular expressions
+# on different operating systems:
+# BSD: -E
+# other: -r
+#
+EXTENDED_REGEX_PARAM=r
+/bin/sh -c "echo 'hello' | sed -${EXTENDED_REGEX_PARAM} 's/[aeou]/_/g' "  >/dev/null 2>&1 || EXTENDED_REGEX_PARAM=E
+
+echo "Executing check_code_style.sh"
+echo "Tooling:"
+
+# check vera++ is set up appropriately
+$VERA $NEST_SRC/nest/main.cpp >/dev/null 2>&1 || usage 1 "Executable $VERA for vera++ is not working!"
+$VERA --profile nest $NEST_SRC/nest/main.cpp >/dev/null 2>&1 ||  bail_out "No profile called nest for vera++ installed. See https://nest.github.io/nest-simulator/coding_guidelines_c++#vera-profile-nest"
+echo "vera++ version: `vera++ --version`"
+
+# check cppcheck 1.69 is installed correctly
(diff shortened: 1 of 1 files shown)


extras/check_code_style.sh
```diff
@@ -0,0 +1,215 @@
+#!/bin/bash
+
+# static code analysis
+START_SHA=master
+END_SHA=HEAD
+
+NEST_SRC=.
+CPPCHECK=cppcheck
+VERA=vera++
+VERA_HOME=/usr/local/lib/vera++
```
heplesser: `VERA_HOME` is not needed, and it would be wrong if `vera++` were installed from MacPorts. One could try to infer it from `which vera++`.

extras/check_code_style.sh
```diff
@@ -0,0 +1,215 @@
...
+# other: -r
+#
+EXTENDED_REGEX_PARAM=r
+/bin/sh -c "echo 'hello' | sed -${EXTENDED_REGEX_PARAM} 's/[aeou]/_/g' "  >/dev/null 2>&1 || EXTENDED_REGEX_PARAM=E
+
+echo "Executing check_code_style.sh"
+echo "Tooling:"
+
+# check vera++ is set up appropriately
+$VERA $NEST_SRC/nest/main.cpp >/dev/null 2>&1 || usage 1 "Executable $VERA for vera++ is not working!"
```
heplesser: On my machine, this command fails because Vera doesn't find the default profile. Although a default profile is in place in `/usr/local/lib/vera++/profiles`. Running with `--profile nest` works.

This test will also fail if one is running outside the NEST source root directory with the `--file` option.

extras/check_code_style.sh
```diff
@@ -0,0 +1,215 @@
...
+echo "cppcheck version: $cppcheck_version"
+if [[ "x$cppcheck_version" != "x1.69" ]]; then
+  bail_out "Require cppcheck version 1.69. Not $cppcheck_version ."
+fi
+
+# clang-format is installed correctly
+$CLANG_FORMAT -style=$NEST_SRC/.clang-format $NEST_SRC/nest/main.cpp >/dev/null 2>&1 || usage 1 "Executable $CLANG_FORMAT for clang-format is not working!"
+clang_format_version=`$CLANG_FORMAT --version | sed -${EXTENDED_REGEX_PARAM} 's/^.*([0-9]\.[0-9])\..*/\1/'`
+echo "clang-format version: $clang_format_version"
+if [[ "x$clang_format_version" != "x3.6" ]]; then
```
heplesser: According to developer space, we require 3.6 *or later*. Why do you require precisely 3.6 here?

extras/check_code_style.sh
```diff
@@ -0,0 +1,215 @@
...
+done
+if $INCREMENTAL; then
+  echo "Press enter to continue."
+  read
+else
+  echo
+fi
+
+for f in $file_names; do
+  if [ ! -f "$NEST_SRC/$f" ]; then
```
heplesser: This won't work if one is outside the NEST root directory running with the `--file` option.

extras/check_code_style.sh
```diff
@@ -0,0 +1,215 @@
...
+Usage: check_code_style.sh [options ...]"
+
+Setup of Tooling is explained here: 
+    https://nest.github.io/nest-simulator/coding_guidelines_c++
+
+Options:
+
+    --help               Print program options and exit
+    --incremental        Do analysis one file after another.
+    --file=/path/to/file Perform the static analysis on this file only.
```
heplesser: Doc should specify that this must be a relative path and that the script must either be run from the nest root directory, or `--nest-src` must also be given. See details on problems that may arise otherwise below.

extras/check_code_style.sh
tammoippen: There a formatting differences between clang-format 3.7 and 3.6. In TravisCI we use 3.6, so the local test should use the same, to reduce surprises.

tammoippen:
@heplesser @jougs 

I updated the PR:
• Remove option to set nest src. The script should always be started at nest root.
• Remove VERA_HOME. It was a relict from a former implementation.
• Added documentation in usage ().

heplesser:
:+1:

extras/check_code_style.sh
```diff
@@ -0,0 +1,215 @@
...
+      echo " - vera++ for $f: (please consider fixing all warnings emitted here.)"
+      $VERA --profile nest $f
+
+      # perform other static analysis
+      echo " - cppcheck for $f: (suggestions for improvements)"
+      $CPPCHECK --enable=all --inconclusive --std=c++03 $f
+
+      ;;
+    *)
+      echo "$f : not a C/CPP file. Do not do static analysis / formatting checking."
```
jougs: "Not executing static analysis / formatting checking." instead of "Do not do ..."

jougs:
@abigailm: Could you please check for proper English in the strings in the script and comment?

@tammoippen: After running the tool, I still have several comments before liking this:
* `--help` should also give a one-line explanation of what the tool is good for.
* The tool expects version 3.6 of clang-format. Yet, after installing it from my Ubuntu 15.04 repositories as explained in [NEST Coding Style Guidelines#Tooling](https://nest.github.io/nest-simulator/coding_guidelines_c++#tooling), I get an error that it is not there. Supplying `--clang-format=clang-format-3.6` works, but I wonder if the default should then not be rather `clang-format-3.6` instead of `clang-format`. Or both should be checked if that's depending on the source of the installed package
* Do we really need version 1.69 of `cppcheck` exactly? That's not available in the Ubuntu 15.04 repositories (1.67 is). Either relax the requirement, or remove the line saying that it can be `apt-get` installed in #96, as that apparently does not work.
* Running on an unmodified branch yields an output of "Performing static analysis on:" but no file names (I assume that's because the delta is empty). That output could be "Nothing to check" or something similar instead.

tammoippen:
@jougs 
> • --help should also give a one-line explanation of what the tool is good for.

Added with 456b15f.
> • The tool expects version 3.6 of clang-format. Yet, after installing it from my Ubuntu 15.04 repositories ...

Changed with 2b06610. Apparently in both setups (OS X and Ubuntu) there is a clang-format-3.6, so we can very well use that one.
> • Do we really need version 1.69 of cppcheck exactly? That's not available in the ...

I have not done a version by version comparison. The version 1.61, the default version you get with our TravisCI machine,  halts with one source file (see #79). Version 1.69 does not have this error anymore. I changed the documentation with 3b73aa4.
> • Running on an unmodified branch yields an output of "Performing static analysis on:" but no ...

I changed the output message with b490fea.

jougs:
@tammoippen: Many thanks for the changes. :+1: from me if @abigailm is OK with the English.

extras/check_code_style.sh
```diff
@@ -0,0 +1,228 @@
...
+    fi
+
+    cat <<EOF
+Usage: ./extras/check_code_style.sh [options ...]"
+
+This script checks our coding style guidelines. The checks are
+performed in the same way as in our TravisCI setup. First it looks
+for changed files in the repository in the commit range <git-start>..<git-end>
+(by default this is master..HEAD). Only the changed files are checked.
+You can specify a different commit range or define the file, which
```
abigailm: You can specify a different commit range, or define the file to be checked, as an argument

extras/check_code_style.sh
```diff
@@ -0,0 +1,228 @@
...
+NEST sources, i.e. all executions should start like:
+    ./extras/check_code_style.sh ...
+
+Setup of the tooling is explained here: 
+    https://nest.github.io/nest-simulator/coding_guidelines_c++
+
+Options:
+
+    --help               Print program options and exit.
+    --incremental        Do analysis one file after another.
```
abigailm: Do analysis on ...

abigailm:
:+1: once the English has been adjusted as suggested by Jochen and me.

extras/check_code_style.sh
heplesser: Could you add a comment to explain this in the script?

extras/check_code_style.sh
heplesser: Or even better "Skipping ..."

heplesser:
I added a few more language and comment suggestions; with that, :+1:

extras/check_code_style.sh
```diff
@@ -0,0 +1,228 @@
...
+
+echo "Executing check_code_style.sh"
+echo "Tooling:"
+
+# check vera++ is set up appropriately
+$VERA ./nest/main.cpp >/dev/null 2>&1 || usage 1 "Executable $VERA for vera++ is not working!"
+$VERA --profile nest ./nest/main.cpp >/dev/null 2>&1 ||  bail_out "No profile called nest for vera++ installed. See https://nest.github.io/nest-simulator/coding_guidelines_c++#vera-profile-nest"
+echo "vera++ version: `vera++ --version`"
+
+# check cppcheck 1.69 is installed correctly
```
heplesser: I think it would be useful to comment in the code why we require 1.69 (and clang 3.6) exactly.

tammoippen:
@jougs @heplesser @abigailm 
Ping. I have updated the documentation and help text accordingly.

jougs:
I think all concerns were addressed adequately and am thus merging. Thanks for the code and the updates.

PR #96, opened 2015-08-24
Documentation for `check_code_style.sh`

tammoippen:
This PR depends on #95. Further see #38.

It adds the documentation for the script `extras/check_code_style.sh`.


coding_guidelines_c++.md
```diff
@@ -31,22 +32,17 @@ _Note that older code might not conform to this rules and should be fixed when c
...
 The [clang-format](http://clang.llvm.org/docs/ClangFormat.html) tool is build on the clang compile fronted. It allows to pretty print input files in a configurable manner, and has vim and emacs integration. We supply a [.clang-format](#clang-format-file) configuration file to enforce some parts of the coding style. During the code review process we check that there is no difference between the committed files and the formatted version of the committed files:
 
-    clang-format <committed file> > tmp_out.txt
-    diff <committed file> tmp_out.txt > /dev/null
-    if [ $? -ne 0 ]; then
-      # error
-    fi
 
-Developer can benefit from the tool by formatting their changes before issuing a pull request: `clang-format -i <committed file>`. We require clang-format version 3.6 or later.
+Developer can benefit from the tool by formatting their changes before issuing a pull request: `clang-format -i <committed file>`. We require clang-format version 3.6.
```
heplesser: Why precisely 3.6? That makes it a bit tricky to keep vera++ locally, as one needs to suppress updates arriving via brew or ports.

coding_guidelines_c++.md
```diff
@@ -78,7 +72,7 @@ OS X:
 
 Or consider installing `vera++` from the [binary packages](https://bitbucket.org/verateam/vera/wiki/Installation).
 
-Finally, we let [cppcheck](http://cppcheck.sourceforge.net/) statically analyse the committed files and check for severe errors:
+Finally, we let [cppcheck](http://cppcheck.sourceforge.net/) statically analyse the committed files and check for severe errors. We require cppcheck version 1.69.
```
heplesser: Same here, why not 1.69 or later?

coding_guidelines_c++.md
tammoippen: There a formatting differences between clang-format 3.7 and 3.6. In TravisCI we use 3.6, so the local test should use the same, to reduce surprises.

coding_guidelines_c++.md
tammoippen: I guess, we can add 'or later' to cppcheck, as the results here are not critical for success of TravisCI tasks.

coding_guidelines_c++.md
heplesser: @tammoippen I understand, but I think this is a rather unfortunate choice by the clang-format developers. The format that you get should be definable independent of the version of the formatter program. Won't we be locked to 3.6 for all eternity now?

- Could you add a short note to the documentation here why we require 3.6?
- Also, could you add information on https://nest.github.io/nest-simulator/coding_guidelines_c++ on how to install a particular version of `clang-format` with brew, and maybe also add the info on why we need 3.6?

tammoippen: Homebrew provides clang-format 3.7 only right now and no recipes for older versions. I supplied a download link of pre-build binaries for clang-format-3.6 for OS X.

heplesser:
:+1: for merging. 

@tammoippen How do you install the pre-built clang-format? I just extracted the archive and added it to my `PATH`.  Any better idea?

tammoippen:
@heplesser this is exactly how i did it.

coding_guidelines_c++.md
```diff
@@ -93,6 +87,49 @@ OS X:
...
 
+Build from scratch:
+
+    git clone https://github.com/danmar/cppcheck.git
+    cd cppcheck
+    git checkout tags/1.69
+    make PREFIX=$PWD/install/ CFGDIR=$PWD/install/cfg HAVE_RULES=yes install
+
+    # In `.bashrc` at install/bin to `PATH`:
+    export PATH=$PATH:/path/to/cppcheck/install/bin
```
jougs: I'd rather use
```
export PATH=/path/to/cppcheck/install/bin:$PATH
```
That way, the local version would even be picked up if a global version was lying around somewhere.

tammoippen:
@jougs @heplesser Ping. :smile:

jougs:
The new version looks good to me. Many thanks.

PR #97, opened 2015-08-27
Adding reset functionality to modules

jougs:
This adds reset functionality to modules as requested in trac.558. Further and more general refactoring of the dynamic loader code seems due, but is not the scope of the referenced ticket.

### DIFF
--- a/nestkernel/dynamicloader.cpp
+++ b/nestkernel/dynamicloader.cpp
@@ -300,7 +300,7 @@ DynamicLoaderModule::reset( SLIInterpreter& interpreter )
   vecDynModules& dm = loadmodule_function.dyn_modules_;
   for ( it = dm.begin(); it != dm.end(); ++it )
   {
-    SLIModule* mod = (*it).pModule;
+    SLIModule* mod = ( *it ).pModule;
     std::string msg = String::compose( "Resetting dynamic module '%1'", mod->name() );
     interpreter.message( SLIInterpreter::M_INFO, "DynamicLoaderModule::reset()", msg.c_str() );
     mod->reset( interpreter );
--- a/nestkernel/dynamicloader.cpp
+++ b/nestkernel/dynamicloader.cpp
@@ -114,7 +114,7 @@ DynamicLoaderModule::~DynamicLoaderModule()
 const std::string
 DynamicLoaderModule::name( void ) const
 {
-  return std::string( "NEST-Dynamic Loader" ); // Return name of the module
+  return std::string( "NEST Dynamic Loader" ); // Return name of the module
 }
 
 const std::string
--- a/nestkernel/dynamicloader.cpp
+++ b/nestkernel/dynamicloader.cpp
@@ -293,6 +293,19 @@ DynamicLoaderModule::init( SLIInterpreter* i )
   }
 }
 
+void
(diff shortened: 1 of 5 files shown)


jougs:
@heplesser and @mhelias: I nominate you as reviewers, as you both have worked on the dynamic loader before. Thanks!

sli/interpret.h
```diff
@@ -898,6 +905,18 @@ SLIInterpreter::addmodule( void )
 }
 
 inline void
+SLIInterpreter::reset_modules()
```
heplesser: Why should this function be inlined? It is in no way performance critical and not a template, so it could very well go into the `cc` file.

sli/interpret.h
```diff
@@ -898,6 +905,18 @@ SLIInterpreter::addmodule( void )
...
+{
+  std::list< SLIModule* >::iterator i;
```
heplesser: I think `i` is a very confusing variable name for a variable representing modules, in particular since we (ab)use `i` as variable name for the interpreter in many places.

sli/slimodule.h
```diff
@@ -47,6 +47,15 @@ class SLIModule
   virtual void init( SLIInterpreter* ) = 0;
 
   /**
+   * Reset the module. This function is currently only called when the
+   * NEST function ResetKernel is invoked.
```
heplesser: And one may begin to wonder about some of our design decisions when `slimodule` needs a `reset()` method solely to permit for `ResetKernel` in NEST ...

nestkernel/dynamicloader.cpp
```diff
@@ -293,6 +293,19 @@ DynamicLoaderModule::init( SLIInterpreter* i )
...
 
+void
+DynamicLoaderModule::reset( SLIInterpreter& interpreter )
+{
+  vecDynModules::iterator it;
+  vecDynModules& dm = loadmodule_function.dyn_modules_;
+  for ( it = dm.begin(); it != dm.end(); ++it )
+  {
+    SLIModule* mod = ( *it ).pModule;
+    std::string msg = String::compose( "Resetting dynamic module '%1'", mod->name() );
```
heplesser: I would rather write `dynamically loaded module`.

heplesser:
@jougs I am afraid the old trac.558 ticket was anticipating a future that never arrived. Specifically, we expected at some point that we might get to a situation in which modules (classes derived from `SLIModule`) would have data members that would need to be reset. This arose in conjunction with caching in Topology, but we chose a different implementation in the end. Therefore, there is no use case for this type of module-resetting. And indeed, your fine code in the end does nothing, since `SLIModule::reset()` does nothing. 

Therefore, I would propose not to merge this PR, as it unnecessarily increases code complexity, and close the ticket as somewhere between wontfix and invalid. Sorry that we didn't discover this before you wrote the code.

jougs:
@heplesser: It's not bad at all! I came to the exact same conclusion as you did, namely that a fix for the referenced ticket is not needed.

However, to come to this insight, I had to look carefully at the code and the time to implement this was not significant compared to the time it took to understand that we don't need it. I'm thus OK with closing this PR and commenting accordingly in the original ticket.

Thanks for the review!

PR #98, opened 2015-08-28
Added S3 deployment

lekshmideepu:
Added the code for deploying the zipped config.log file and the reports folder to amazon S3. Since the secure environment variables are not extracted correctly for a PR, I had created a test branch on the main repo and did the testing. Please have a look at the below build:
https://travis-ci.org/nest/nest-simulator/builds/77636786
@heplesser could verify the uploaded files as he knows the access credentials for S3. For others, I could sent you by other means.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ cache:
   pip: true
 #virtualenv:
  # system_site_packages: true
-#addons:
-#  artifacts:
- #  debug: true
 
 before_install:
   - echo $LANG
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -69,18 +66,33 @@ before_install:
...
-    cat config.log
-    for i in $(find reports -type f); do
-      echo "###########################"
-      echo "File: $i"
-      cat $i
-    done
+  - tar -zcvf configlog.tar.gz ./*.log
+  - tar -zcvf reports.tar.gz ./reports 
+  - mkdir -p $TRAVIS_BUILD_DIR/build/artefacts_upload
+  - cp configlog.tar.gz reports.tar.gz $TRAVIS_BUILD_DIR/build/artefacts_upload
```
heplesser: I almost overlooked the `reports.tar.gz` on this line. For better readability, I would suggest one `cp` command per file. And I guess we could use `mv`, since the archives are not used anywhere else.

build.sh
```diff
@@ -174,3 +174,7 @@ if [ "$format_error_files" != "" ]; then
   exit 42
 fi
 
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  echo "WARNING: Not uploading results as this is a pull request" >&2
+  exit 0
+fi
```
heplesser: Newline missing at end of file.

build.sh
```diff
@@ -174,3 +174,7 @@ if [ "$format_error_files" != "" ]; then
   exit 42
 fi
 
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
heplesser: The logic is a bit backward here (why not `== "true"`)? I am not an expert in shell programming, but from configure scripts I seem to remember that test are often written as `x$foo != xabc`. I believe this is to avoid problems in case `$foo` is empty.

heplesser: Sorry for the previous comment, my ignorance! After reading [the docs](http://docs.travis-ci.com/user/environment-variables/#Convenience-Variables) I realize that we must test against `"false"` here.

heplesser:
@lekshmideepu Very nice! I checked on S3, and the correct files are uploaded. I would maybe change the name `configlog.tar.gz` to `logfiles.tar.gz`, since that archive also includes `install-help.log`.

Otherwise :+1:

lekshmideepu:
Thanks @heplesser for the comments and suggestions. I have incorporated those in my code. Along with that, I used the travis environment variables instead of the encrypted AWS access credentials.

lekshmideepu:
I'm afraid, I might have to add the checks for the travis environment variables $ARTIFACTS_KEY and $ARTIFACTS_SECRET as they will not be exported for a PR.

lekshmideepu:
I'm still waiting for the second reviewer:wink: Could anyone of you @abigailm @apeyser @jougs or @tammoippen review it and give a green signal to it?

jougs:
Thanks for addressing this issue and the comments by @heplesser! :+1: and merging!

lekshmideepu:
Thanks @jougs :smile:

PR #104, opened 2015-09-09
Wrong comparison operator for connections using parameter arrays

hannahbos:
Changed the comparison operator checking the dimensionality of parameter arrays from 'is not' to !=. This caused problems for larger arrays since 'is not' compares the identity of objects (which appears to be fixed in Python for integers ranging from -5 to 256) rather than their value. @heplesser @jougs

### DIFF
--- a/pynest/nest/hl_api.py
+++ b/pynest/nest/hl_api.py
@@ -1137,13 +1137,13 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
                 if isinstance(value, (np.ndarray, np.generic)):
                     if len(value.shape) == 1:
                         if rule == 'one_to_one':
-                            if value.shape[0] is not len(pre):
+                            if value.shape[0] != len(pre):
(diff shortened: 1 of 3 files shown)


jougs:
Well spotted. :+1: for merging.

jougs:
Do we need to extend the test to also check larger matrices to prevent a re-occurence of this problem? Moreover, I assume that this might also cause trouble in other locations.

heplesser:
:+1: I had a quick look for `is` and `is not` in `hl_api.py` and did not spot any other problematic uses.
I think @jougs suggestion to add a test with larger matrices is a good one, but I could not see right away which test covers parameter arrays. But this PR can be merged in any case.

hannahbos:
Thanks @jougs @heplesser  for the review. I updated the tests to check larger input arrays.

jougs:
As we don't know a better name, I'm happy to merge this as is.

heplesser:
:+1:

PR #107, opened 2015-09-24
Split PyNEST high-level API into multiple files

jougs:
This PR splits the PyNEST high-level API into multiple files according to the responsibility of the different functions. It also makes loading of the submodules dynamic, allowing modules to just drop their Python files into the `lib` folder instead of doing fancy installations themselves.

I nominate @tammoippen, @apeyser, and @heplesser as reviewers.


tammoippen:
Test.. ignore me.

tammoippen:
Test.. ignore me.

pynest/nest/lib/hl_api_connections.py
```diff
@@ -0,0 +1,641 @@
...
+
+    try:
+        GetChildren(gids)
+        return True
+    except kernel.NESTError:
+        return False
+
+@check_stack
+def CGConnect(pre, post, cg, parameter_map=None, model="static_synapse"):
+    """
```
tammoippen: @jougs  I think, as the CG* functions provide access to external connection libraries, it would be good to put them into a separate file.

jougs: @tammoippen: technically I agree. I just thought that this PR should just split the HL API as it is now and further splitting can be done in our refactoring branch. The automatic module discovery implemented in `__init__.py` lines 179-181 should make it fairly simple to have modules install their own files if they want to.

pynest/nest/lib/hl_api_nodes.py
```diff
@@ -0,0 +1,150 @@
...
+    else:
+        sps(nodes)
+
+    sps(params)
+    sr('2 arraystore')
+    sr('Transpose { arrayload pop SetStatus } forall')
+
+
+@check_stack
+def GetStatus(nodes, keys=None):
```
heplesser: `GetStatus()` can also be used on connections (same for `SetStatus()`). So isn't it a bit unsystematic to put this in a `nodes` file? Should the two functions rather go into the `info` component?

pynest/nest/__init__.py
```diff
@@ -178,8 +173,12 @@ def test():
 
     hl_api.set_debug(debug)
 
+from .pynestkernel import *
+from lib.hl_api_helper import *
+
+for name in os.listdir(os.path.join(os.path.dirname(__file__), "lib")):
+    if name.endswith(".py") and not name.startswith('__'):
+        exec("from lib.{0} import *".format(name[:-3]))
```
heplesser: I am not quite sure what is going on here (maybe I am just too tired) but what kind of files of name `lib.xyz.py` are we expecting here?

heplesser:
@jougs In principle, I am positive, but see my comments above.

pynest/nest/lib/hl_api_nodes.py
jougs: @heplesser: agreed. This code probably was still in the nodes section of the old `hl_api.py` from the times in which this did not work for connections. I moved it in a3ec4d4.

pynest/nest/__init__.py
jougs: @heplesser, you're right. This code is indeed a bit cryptic (and the result of endless hours of failure) and deserves a comment that explains it. I've added one in 5afad17

heplesser:
:+1: for merge.

tammoippen:
Very nice. :+1:  from my side. Merging.

PR #108, opened 2015-09-25
Added example comparing precise and grid-constrained neuron. 

jougs:
This example is based on work by Jeyashree Krishnan and fixes trac.796. I suggest @suku248 and @abigailm as reviewers.


jougs:
Thanks for the quick and thorough review. I hope the new commits address your concerns.

suku248:
a definite :+1:

jougs:
As time is up for this HBP related PR and because it is "only" an example and got one positive review, I'm merging without waiting for a second review.

PR #109, opened 2015-09-25
Get RCS information from git

### LINKED ISSUE #63: lib/sli/rcsinfo.sli
Ok, I have this file in the tree, and it gets updated to:
```
statusdict /rcsinfo (no_rcsinfo_available) put
```

without which, tests fail. Is this thing still necessary?

@jougs

jougs:
This updates the script for retrieving RCS information to use git instead of svn. It also fixes a bug in the `doc` and `fulldoc` targets. This fixes #63.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -12,7 +12,7 @@ doc:
 if DOXYGEN_AVAILABLE
 
 	$(eval RCSINFO:=`@PKGSRCDIR@/extras/create_rcsinfo.sh @PKGSRCDIR@ /tmp`) \
-	  $(SED) -i '' "/^PROJECT_NUMBER/ s/=.*/= @SLI_VERSION@, $(RCSINFO)/" @PKGBUILDDIR@/doc/normaldoc.conf
+	  $(SED) -i "/^PROJECT_NUMBER/ s/=.*/= @SLI_VERSION@, $(RCSINFO)/" @PKGBUILDDIR@/doc/normaldoc.conf
 	$(DOXYGEN) @PKGBUILDDIR@/doc/normaldoc.conf
 
 endif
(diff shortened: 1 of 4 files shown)


jougs:
I recommend @apeyser and @tammoippen as reviewers.

@tammoippen: can you please also check if `make doc` and `make fulldoc` work on MacOS? The second commit reverts a change by @heplesser (78a2ccc) meant to fix something for that system.

apeyser:
We should also consider removing the rcsinfo.sli file that goes into the repo. It must always be wrong (it can't have the future git hash until after it's been checked in...), and in those cases it's better to make it a buildtime file. For major releases that have no associated repo, if we're supporting that, it's better to have a fixed file value rather than values like "master". We would then want the release tag in rcsinfo.sli.

jougs:
@heplesser: can you please check the fix for sed in 7baf37b on OSX?

jougs:
@apeyser: Do you think the combination of dc5d9a6 and 2ed253e is sufficient to address your question?

tammoippen:
Hi @jougs !
I am very happy with this PR. `doc` and `fulldoc` work like a charm. Branch and SHA1 show up in the docs and in nests `statusdict /rcsinfo get =`. Hence, the `sed` commands work on OS X. :+1:  from my side.

Side note: we should really document somewhere that one needs doxygen and graphviz in order to build the documentation (could not find anything on [nest-simulator.org](nest-simulator.org) or [nest.github.io/nest-simulator/](nest.github.io/nest-simulator/) ). But this is a matter of another Issue.

heplesser:
@tammoippen Would you create the Issue about documenting doxygen and graphviz dependencies?

heplesser:
@jougs Works like a charm for me as well, so :+1:

If you have time, one little improvement suggestion: The docxygen documentation includes `SLI_VERSION`.  Maybe it should include the rcsinfo as well. To do this, go to files

- `doc/normaldoc.conf.in`
- `doc/fulldoc.conf.in`

and modify the following line (approx line 41)

```sh
PROJECT_NUMBER         = @SLI_VERSION@
```

jougs:
@heplesser: Maybe I don't understand your suggestion properly, but this is exactly what the `sed` commands in the `doc` and `fulldoc` targets are doing. They replace `= @SLI_VERSION@` in the unmodified doxygen config files by `= @SLI_VERSION@, $(RCSINFO)`, before running doxgen. This means that the information in the generated docs always contains current information, even if the user calls `git pull` between calls to `make`.

jougs:
@tammoippen, @heplesser: thanks for the review. I regard the original issue fixed for now and will merge this PR.

@apeyser: please feel free to open a follow-up ticket for adding the current git tag to rcsinfo.sli.

heplesser:
@jougs: Sorry, my mistake. To test, I had just downloaded your branch as zip file from Github, and then, obviously, not Git version information appeared. I did not look carefully enough at the code then. 

I now cloned your repo and it worked! I just have a tiny comment on the format: `2.6.1 (git f33d3b9)` would be even nicer than `2.6.1,f33d3b9`. And would there be any way to avoid the comma or (in my suggestion the `(git )`) in case git is not available?

PR #110, opened 2015-09-26
Update releasetools

jougs:
This PR contains the following changes:
* Remove last existing references to SVN
* Update the script to roll tarballs to use Git instead of SVN
* Move the script to `extras/create_release.sh`
* Removed outdated scripts for module handling.

I suggest @heplesser and @tammoippen as reviewers. As this is a behind-the-scenes change, I assume that one reviewer is enough. Please comment on this.

PR #112, opened 2015-09-27
Add merging of master to branch checkout if tagging is enabled

jougs:
Just a minor follow-up to #110.


jougs:
Immediately merging due to minority.

PR #118, opened 2015-10-03
Extended parrot_neuron to allow connections onto port 0 and 1, not repeating spikes on port 1

flinz:
This allows setting pre- and postsynaptic spike times for STDP protocols by connecting two parrot neurons onto port 1 postsynaptically by, e.g., a stdp_synapse. Thanks to @heplesser for your advice.

Here is a short test/demonstration
```python
import nest
import pylab as pl
import numpy as np
nest.ResetKernel()

# set pre and postsynaptic spike times
dt = 50.
pre_times = np.arange(100, 1101, 200., float)
post_times = pre_times + dt

# create spike_generators with these times
pre_spikes = nest.Create("spike_generator", 1, {"spike_times": pre_times})
post_spikes = nest.Create("spike_generator", 1, {"spike_times": post_times})

# create parrot neurons and connect spike_generators
pre_parrot = nest.Create("parrot_neuron", 1)
post_parrot = nest.Create("parrot_neuron", 1)
nest.Connect(pre_spikes, pre_parrot)
nest.Connect(post_spikes, post_parrot)

# connect both parrot neurons with a stdp synapse onto port 1
# thereby spikes transmitted through the stdp connection are
# not repeated postsynaptically.
syn_spec = {
   "model": "stdp_synapse",
   "receptor_type": 1  # set receptor 1 postsynaptically, to not generate extra spikes
}
conn_spec = {
   "rule": "one_to_one",
}
nest.Connect(pre_parrot, post_parrot, syn_spec=syn_spec, conn_spec=conn_spec)

# simulate for ~1s and print connection weights every 100ms
nest.Simulate(75)
for _ in range(5):
   nest.Simulate(200)
   print nest.GetStatus(nest.GetConnections(source=pre_parrot))[0]['weight']
```


models/parrot_neuron.h
```diff
@@ -48,6 +48,10 @@ example a poisson generator with a high firing rate is connected
 to a parrot neuron, the communication cost associated with outgoing
 spikes is much bigger for the latter.
 
+Only spikes incoming on connections to port 0 will be repeated.
```
heplesser: incoming -> arriving

models/parrot_neuron.h
```diff
@@ -48,6 +48,10 @@ example a poisson generator with a high firing rate is connected
...
+Connections onto port 1 will be accepted, but spikes incoming
+through port 1 will be inignored.
```
heplesser: typo: inignored

I think it would be useful to briefly explain the purpose of allowing spike arrival via port 1 while ignoring those spikes.

models/parrot_neuron.h
```diff
@@ -145,9 +149,20 @@ parrot_neuron::send_test_event( Node& target, rport receptor_type, synindex, boo
 inline port
 parrot_neuron::handles_test_event( SpikeEvent&, rport receptor_type )
 {
-  if ( receptor_type != 0 )
+  // Allow connections to port 0 (spikes to be repeated)
+  // and port 1 (spikes to be ignored).
+  if ( receptor_type == 0 )
```
heplesser: I think
```C++
if ( receptor_type == 0 or receptor_type == 1 )
{
  return receptor_type;
}
else
{
  throw UnknownReceptorType( receptor_type, get_name() );
}
```
would be cleaner.

heplesser:
@flinz : Nice work! 

Could you turn the example script you included above into a proper test? The main point would be to test that the synaptic weight changes, and that only the spikes from the `post_spikes` generator are parroted by the `post` neuron, not those coming from the `pre` neuron. If you feel like writing the test in SLI, it should go into `testsuite/unittests`, otherwise into `pynest/nest/testsuite`.  In that case you also have to add your test module to `test_all.py` in that directory.

models/parrot_neuron.h
flinz: I agree, but tried to stick to the existing coding style (see https://github.com/nest/nest-simulator/blob/master/models/iaf_psc_exp.h#L354). Will change this here.

flinz:
Ok, good idea. Will add the test shortly.

flinz:
I wrote a python test for parrot_neuron, including also testing for the basic functionality of repeating spikes at correct times (I could not find any tests for parrot_neuron, both in the SLI and python tests).

flinz:
Just for your info, travis CI is having weird problems - to restart the build process I have/had to open/close the pull request.

jougs:
@flinz: re-opening and closing unfortunately restarts *all* jobs, which brings a certain likelihood that one of them fails again. I can instead monitor the jobs and restart them individually if they fail.

flinz:
@jougs: ah, sorry for the travis load then. I'll just not touch it anymore.

tammoippen:
Very nice work! :+1: from my side!

pynest/nest/tests/test_parrot_neuron.py
```diff
@@ -0,0 +1,162 @@
...
+        }
+        nest.Connect(pre_parrot, post_parrot, syn_spec=syn_spec, conn_spec=conn_spec)
+
+        # get STDP synapse and initial weight
+        syn = nest.GetConnections(source=pre_parrot, synapse_model="stdp_synapse")
+        initial_weight = nest.GetStatus(syn)[0]['weight']
+
+        nest.Simulate(1000)
+
+        # check that the STDP protocol has elicited a weight change.
```
abigailm: I would prefer that it checks that it is the correct weight change, to make sure the delays are properly accounted for etc.

abigailm:
happy to merge once my comment has been addressed

pynest/nest/tests/test_parrot_neuron.py
flinz: Sorry, this is not totally clear to me.. could you be more precise which delays you are referring to?

As i see it, all I should need to check here is that the parrot_neuron spikes at times that are as intended (which is generator_spiketimes+synaptic delays between the generators and the parrots). 

The actual weight change given some intended pre- and postsynaptic spiketimes of the parrots should be part of the STDP model, and its tests.

flinz:
@jougs this commit gave me some interesting travis fails. pynestkernel.so throws some errors - seems you might be [building/checking against python built without some unicode flag](http://stackoverflow.com/questions/2584198/undefined-symbol-pyunicodeucs2-decode-whilst-trying-to-install-psycopg2/2584554#2584554).

lekshmideepu:
There is an issue with Python 2.6.9 on Travis worker systems after releasing the new Trusty beta image. I'm working on this problem.

lekshmideepu:
https://github.com/travis-ci/travis-ci/issues/4948

pynest/nest/tests/test_parrot_neuron.py
```diff
@@ -0,0 +1,162 @@
...
+import nest
+import unittest
+
+
+@nest.check_stack
+class ParrotNeuronTestCase(unittest.TestCase):
+    """Check parrot_neuron spike repetition properties"""
+
+    def setUp(self):
+        nest.set_verbosity(100)
```
heplesser: I would not recommend shutting the kernel up so brutally. A bit of output makes it easier to find out what went wrong when a test fails. `nest.set_verbosity('M_WARNING')` should do fine.

pynest/nest/tests/test_parrot_neuron.py
```diff
@@ -0,0 +1,162 @@
...
+        self.source = nest.Create("spike_generator", 1, {"spike_times": [self.spike_time]})
+        self.parrot = nest.Create('parrot_neuron')
+        self.spikes = nest.Create("spike_detector")
+
+        # record source and parrot spikes
+        nest.Connect(self.source, self.spikes)
+        nest.Connect(self.parrot, self.spikes)
+
+        self.gid_source = nest.GetStatus(self.source)[0]["global_id"]
+        self.gid_parrot = nest.GetStatus(self.parrot)[0]["global_id"]
```
heplesser: Why so complicated? `nest.Create()` returns a tuple of GIDs, so `self.gid_source = self.source[0]` would do, if you need it at all.

pynest/nest/tests/test_parrot_neuron.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # assert spike was repeated at correct time
+        assert post_time, "Parrot neuron failed to repeat spike."
+        assert self.spike_time + delay == post_time, "Parrot neuron repeated spike at wrong delay"
+
+    def test_ParrotNeuronIgnoreSpike(self):
+        """Check parrot_neuron ignores spikes on port 1"""
+
+        # connect with arbitrary delay to port 1
+        nest.Connect(self.source, self.parrot, syn_spec={"receptor_type": 1})
```
heplesser: I would explicitly specify a delay here as well, just to make it definite that the spike would arrive within the simulated time. Now, you rely on the default delay being 1 ms, but that could in principle change.

pynest/nest/tests/test_parrot_neuron.py
heplesser: @flinz I think this is exactly what @abigailm was asking for: an explicit check that the correct weight change occurs. This is the same thing done as in `testsuite/unittest/test_stdp_synapse.sli`.

heplesser:
@flinz I added some minor comments to the code. We had some problems with Travis lately which are now fixed. To apply that fix to your PR, please pull the recent changes from master into your PR. This will automatically trigger Travis. Once you have updated your test as you suggest above, your PR should be ready for merging.

pynest/nest/tests/test_parrot_neuron.py
abigailm: @flinz makes a good point about the proper place for testing the stdp_synapse being in the test for that synapse (which already exists). Adding this test would indeed be redundant. I think it would be sufficiently reassuring to check that the change of weight is in the expected direction.

flinz:
@heplesser thanks for the helpful feedback and comments on the code. 

As @abigailm [pointed out here](https://github.com/nest/nest-simulator/pull/118#discussion_r43736299) I chose to not include a reproduction of the full stdp_synapse model here, and instead test for causality in the weight change, i.e. ``pre-post -> potentiation``, ``post-pre -> depression``.

PR is updated with your suggested improvements and this newer version of the test.

pynest/nest/tests/test_parrot_neuron.py
```diff
@@ -0,0 +1,175 @@
...
+        self.delay = .2
+        self.source = nest.Create("spike_generator", 1, {"spike_times": [self.spike_time]})
+        self.parrot = nest.Create('parrot_neuron')
+        self.spikes = nest.Create("spike_detector")
+
+        # record source and parrot spikes
+        nest.Connect(self.source, self.spikes)
+        nest.Connect(self.parrot, self.spikes)
+
+        self.gid_source = self.source[0]
```
abigailm: is this needed? I don't see where.

abigailm:
I like the new version. I think the assignation of gid_source is unnecessary but I might have overlooked something. :+1: when it is removed or I see the line where it is used.

flinz:
@abigailm you are right. also, i removed the explicit ``self.gid_parrot`` variable and just call ``self.parrot[0]``.

abigailm:
merging.

jougs:
Many thanks for your contribution!

Issue #99, opened 2015-09-07
Add summary for build process in TravisCI


jougs:
Currently, the log files of Travis contain different types of information in one large block of text:
* configuration log and summary
* build output
* output from the coding guideline checker
* static code analysis results
* logs from the testsuite
In case of failures, it is hard to find the exact cause of failure.

Once the log files are available on S3 (#98), the logs shown by TravisCI could be shortened to only contain summaries of the information detailed above and a link to the full logs. They could also point out more prominently, what the exact cause for the failure was.

@lekshmideepu @tammoippen: Could you please look into this?

lekshmideepu:
As we have merged #98, we could decide about the files which are to be uploaded to S3 to avoid the confusion in the primary log output. I think, the output from the coding guideline checker and the static code analysis results can be uploaded to S3.

heplesser:
Ideally, the build log on Travis should be compact and point to the source of problems. All "volume output" should be as artifacts on S3. Essentially, I would like to see on Travis only 

```
Setup ... passed
NEST coding guidelines check ... passed
NEST configure ... passed
NEST static code analysis ... passed
NEST build ... passed
NEST install ... passed
NEST testsuite ... passed
```

If any of those tests fails, instead of `passed`, a short message should be provided. For setup, configure, build and install, probably a tail of the console output is most suitable, for the coding guidelines/static analysis the error message and pertaining file, and for the testsuite a list of failing tests (if that is tricky, the count of failing tests as displayed on the console now). All else should then go onto S3.

Note that I put the coding guideline and static analysis early in the suggestion above, so that fast-running tests will fail early, reducing the time required for to flag failure. I suppose the static analysis must come after configure so that all  `HAVE_*` macros are defined.

tammoippen:
Please see PR #119.

Regarding the short TravisCI log: all Travis does comes into the log, i.e. all the setup stuff will be in there anyway. If I am not mistaken, this behaviour is hard to change.

I have also put the static analysis to the front, but Travis will not fail immediately when the formatting is wrong, but will first try to build NEST and perform `make installcheck`. This way we will have more information instead of only formatting is off.

tammoippen:
Fixed by #119

PR #119, opened 2015-10-07
Add a script to parse TravisCI log and give a nice summary at the end.

tammoippen:
This is a PR in response to #99. 

The file `extras/parse_travis_log.py` parses the output generated by the `./build.sh`, when it is called from TravisCI. The summary is appended to the TravisCI log. For an example of a failing log see [here](https://travis-ci.org/tammoippen/nest-simulator/jobs/84117280). Scroll down to the end of the log:
scroll
and you will see this:

    $ ./extras/parse_travis_log.py build.sh.log
    Bootstrapping:     Ok
    Vera init:         Ok
    Cppcheck init:     Ok
    Changed files:     ['extras/scan_travis_log.py', 'nestkernel/connector_model.h', 'nestkernel/    nestmodule.cpp', 'nestkernel/network.cpp']
    Formatting:        Error
    Configure:         Ok
    Make:              Ok
    Make install:      Ok
    Make installcheck: Error (101 / 735)
    
    Static analysis:
       ------------------------
       | nestkernel/network.cpp:
       |    * vera++:
       |      - source file is too long : 1
       |      - negation operator used in its short form : 54
       |      - full block {} expected in the control structure : 87
       |    * cppcheck:
       |      - (performance) Function parameter 'weights' should be passed by reference. : 4
       |      - (style, inconclusive) Technically the member function 'nest::Network::model_in_use'     can be const. : 1
       |      - (performance) Function parameter 'target_ids' should be passed by reference. : 2
       |      - (performance) Function parameter 'source_ids' should be passed by reference. : 2
       |      - (performance) Function parameter 'delays' should be passed by reference. : 4
       |    * clang-format:
       |      - diff : 
    ##############################
    1150c1150
    <        continue;
    ---
    >       continue;
    ##############################
       |      - Ok? : False
       ------------------------------
       | nestkernel/connector_model.h:
       |    * vera++:
       |    * cppcheck:
       |      - (performance) Function parameter 'name' should be passed by reference. : 2
       |      - (style, inconclusive) The variable 'delay' hides a typedef with the same name. : 2
       |      - (style, inconclusive) The variable 'weight' hides a typedef with the same name. : 2
       |    * clang-format:
       |      - diff : 
    ##############################
    215c215
    <   return num_connections_ +1;
    ---
    >   return num_connections_ + 1;
    ##############################
       |      - Ok? : False
       ---------------------------
       | nestkernel/nestmodule.cpp:
       |    * vera++:
       |      - horizontal tab used : 2
       |      - negation operator used in its short form : 15
       |      - full block {} expected in the control structure : 31
       |      - source file is too long : 1
       |      - too many consecutive empty lines : 1
       |    * cppcheck:
       |      - (performance) Variable 'time' is reassigned a value before the old one has been     used. : 3
       |      - (style, inconclusive) The variable 'port' hides a typedef with the same name. : 1
       |      - (style, inconclusive) The variable 'delay' hides a typedef with the same name. : 1
       |      - (style, inconclusive) The variable 'weight' hides a typedef with the same name. : 1
       |    * clang-format:
       |      - diff : 
    ##############################
    246c246
    <   ad.push_back( conn.get_synapse_model_id());
    ---
    >   ad.push_back( conn.get_synapse_model_id() );
    ##############################
       |      - Ok? : False

Each file has a separate section in the `Static analysis:` part. In it it describes the most dominant flaws of the individual tool for this file. The number at the end of each entry is the number of occurrences of this flaw. `* clang-format:` reports of the format is ok `Ok?`; if not, it shows the diff.


extras/parse_travis_log.py
```diff
@@ -0,0 +1,264 @@
...
+#
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+# This script parses the TravisCI output as it is generated
```
heplesser: Shouldn't this be a file docstring instead of a comment?

heplesser:
@tammoippen: This is very nice (except some incorrect copyright header)! I suppose the way this is appended to the end of the log is that Travis simply puts all output into the log, isn't it? Would there be any way to replace the travis log output and put this summary at the top? This would help in cases where Travis truncates the log (10000 lines). 

Or could we do something like this (maybe also for other stages):
```yaml
script:
  - set -o pipefail
  - ./build.sh 2>&1 > build.sh.log
  - python extras/parse_travis_log.py build.sh.log
  - cat build.sh.log
```
My idea here is to first write all output to the log file and not let it go into the log travis collects from stdout and stderr, then produce the summary, and finally append the full log. If one did this exactly as shown above, it would put the summary between all setup logs and the build log, which is not optimal, so we would need to redirect output at all stages to a file and then `cat` all the files into the log after the summary.


In any case, it would be nice to put a few lines of stars or so at the top of the output from `parse_travis_log.py` to make it 100% clear to the reader where the summary starts.

tammoippen:
I surrounded the `parse_travis_log.py` output by:

    --------<<<<<<<< Summary of TarvisCI >>>>>>>>---------

@heplesser When something goes wrong in `./build.sh 2>&1 > build.sh.log` and halts TravisCI we will not see, what caused the problem. Putting the output at the top requires a lot of these redirections. I think, having the summary in the end is a good place: you can reach it quite easily (one click), our log is well below 4000 lines and even if it surpasses the 10,000 margin, you can always download the raw log, which is complete.

heplesser:
@tammoippen OK, :+1:

tammoippen:
@apeyser or @lekshmideepu , can you please also have a look at this?

lekshmideepu:
@tammoippen : I would like to add one more information in the summary part like:
 "Logs uploaded to S3: Yes/No"

tammoippen:
@lekshmideepu Good idea! I added that line in the last commit. 

BTW: Is it a good idea to only upload the results on merges to nest/nest-simulator master? If everything goes well and we only merge well behaving PR, there will never be a need to look into the reports. More interestingly would be the reports on PR that fail.

lekshmideepu:
@tammoippen : Yes, I also had the same discussion with the travis team. But Travis doesn't allow uploading of build artifacts in case of a PR.

PR #120, opened 2015-10-08
Second Pull Request on Structural Plasticity. Connector base modifications in order to erase connections.

sdiazpier:
This second pull request includes the modifications in the connector base classes required to erase connections. This functionality is implemented originally to enable the structural plasticity functionality but it can also be used without enabling structural plasticity in the simulation.


jougs:
Thanks for the pull request. I assume that @tammoippen, @suku248, and @mhelias are good reviewers for this.

nestkernel/connector_base.h
```diff
@@ -75,6 +75,37 @@ suicide_and_resurrect( Told* connector, C connection )
...
+inline Tnew*
+suicide( Told* connector )
+{
+  Tnew* p = 0;
+#ifdef USE_PMA
+  connector->~Told();
+#else
+  delete connector; // suicide
+#endif
+  return p;
```
tammoippen: Why does `suicide` have to return `Tnew*` if it will return 0 any way? It should be a function returning `void`.

nestkernel/connector_base.h
```diff
@@ -98,6 +129,7 @@ class ConnectorBase
 
   virtual size_t get_num_connections() = 0;
   virtual size_t get_num_connections( synindex syn_id ) = 0;
+  virtual size_t get_num_connections( size_t target_gid, size_t thrd, synindex syn_id ) = 0;
```
tammoippen: What is the reason for introducing the third `get_num_connections` method? As far as I can see, it is called only in the `HetConnector::get_num_connections` version of this method and nowhere else... I am confused :confused: 

The same goes for the `get_target_gids` method.

nestkernel/connector_model_impl.h
```diff
@@ -442,6 +442,89 @@ GenericConnectorModel< ConnectionT >::add_connection( Node& src,
...
+  vector_like< ConnectionT >* vc;
+
+  if ( conn->homogeneous_model() )
+  {
+    assert( conn->get_syn_id() == syn_id );
+    vc = static_cast< vector_like< ConnectionT >* >( conn );
+    // delete the first Connection corresponding to the target
+    for ( size_t i = 0; i < vc->size(); i++ )
+    {
+      ConnectionT* connection = &vc->operator[]( i );
```
tammoippen: This looks funny... :smile: 
I only see the `operator[]` function used in to places: here and line 499. I guess, the proper way of using it would be `&( *vc )[ i ]`, which looks even more funny. Maybe we should go for the [`at`](http://www.cplusplus.com/reference/vector/vector/at/) function of vectors here: `&vc->at( i )`. If so, we should not use `assert` but throw an [`out_of_range`](http://www.cplusplus.com/reference/stdexcept/out_of_range/) exception.

nestkernel/connector_base.h
sdiazpier: Hi dear Tammo, thanks for taking the time to check the code and for your very useful comments. This function is used by the structural plasticity manager. I know it is not visible right now, because the structural plasticity manager class will be part of the third pull request. The whole implementation of structural plasticity was split into three pieces in order to make easier the code review process. I tried to make each piece self explanatory and independent from each other but in this case the usage is not clear. Sorry for the confusion.

sdiazpier:
Hi dear @tammoippen, I have made modifications according to your comments and committed the changes. Thank you for your review. Please let me know if anything else would be required.

nestkernel/connector_base.h
```diff
@@ -352,12 +506,51 @@ class Connector< 1, ConnectionT > : public vector_like< ConnectionT >
...
+
   Connector< 2, ConnectionT >&
   push_back( const ConnectionT& c )
   {
     return *suicide_and_resurrect< Connector< 2, ConnectionT > >( this, c );
   }
 
+  ConnectorBase& erase( size_t )
+  {
+    // Should destroy the Connector
```
jougs: Should or does? If it does, please remove the "should"

nestkernel/connector_base.h
```diff
@@ -352,12 +506,51 @@ class Connector< 1, ConnectionT > : public vector_like< ConnectionT >
...
+    return 1;
+  }
+
+  ConnectionT&
+  at( size_t i )
+  {
+
+    if ( i != 0 )
+    {
+      throw std::out_of_range( "Invalid attempt to access a connection at an out of range index" );
```
jougs: Can you please rephrase? Something like "Invalid attempt to access connection: index yyy out of range." For adding the number to the string you can use the String::compose function used in many places of NEST.

nestkernel/connector_base.h
```diff
@@ -154,7 +187,9 @@ class vector_like : public ConnectorBase
 
 public:
   virtual ConnectorBase& push_back( const ConnectionT& c ) = 0;
-
+  virtual ConnectorBase& erase( size_t i ) = 0;
+  virtual size_t size() = 0;
+  virtual ConnectionT& at( size_t i ) = 0;
   void
```
jougs: Please add an empty line before `void`

nestkernel/connector_base.h
```diff
@@ -177,6 +212,24 @@ class Connector : public vector_like< ConnectionT > // unfortunately, we need th
     C_[ K - 1 ] = c;
   }
 
+  /*
+   * Create a new connector and remove the ith connection. Returns a connector
+   * with size K from a connector of size K+1.
+   */
```
jougs: Please add some more explanation about the copying process implemented below. It took me quite a bit of time to figure it out. Always expect your readers to be as tired as I am currently ;-)

nestkernel/connector_base.h
```diff
@@ -216,12 +269,75 @@ class Connector : public vector_like< ConnectionT > // unfortunately, we need th
...
+   * in the same manner as it would work with an array.
+   * @param i the index of the connection to be retrieved.
+   * @return The connection stored at position i.
+   */
+  ConnectionT&
+  at( size_t i )
+  {
+    if ( i >= K || i < 0 )
+    {
+      throw std::out_of_range( "Invalid attempt to access a connection at an out of range index" );
```
jougs: Please rephrase. See [below](https://github.com/nest/nest-simulator/pull/120/files#r44633433).

nestkernel/connector_base.h
```diff
@@ -450,6 +652,25 @@ class Connector< K_cutoff, ConnectionT > : public vector_like< ConnectionT >
     C_[ K_cutoff - 1 ] = c;
   };
 
+  /**
+   * Create a new connector and remove the ith connection
+   * @param Cm1 Original connector of size K_cutoff.
+   * @param i The index of the connection to be deleted.
+   */
+  Connector( const Connector< K_cutoff, ConnectionT >& Cm1, size_t i ) //: syn_id_(Cm1.get_syn_id())
```
jougs: Please also add some information about the algorithm. See [above](https://github.com/nest/nest-simulator/pull/120/files#r44636702).

nestkernel/connector_base.h
```diff
@@ -496,6 +735,29 @@ class Connector< K_cutoff, ConnectionT > : public vector_like< ConnectionT >
...
+  size()
+  {
+    return C_.size();
+  }
+
+  ConnectionT&
+  at( size_t i )
+  {
+    if ( i >= C_.size() || i < 0 )
+      throw std::out_of_range( "Invalid attempt to access a connection at an out of range index" );
```
jougs: Same as above.

jougs:
Nice work, @sdiazpier! I'm :+1: for merging once my comments are addressed.

sdiazpier:
Hi dear @jougs, thank you very much for your review. I have committed the changes related to your comments. Please let me know if anything else is missing.

jougs:
@tammoippen: could you please merge if your comments have been addressed satisfactorily? I'm fine with the current version.

tammoippen:
Sorry for the delay. I am happy with the changes! Nice work :+1: . Merging.

PR #136, opened 2015-10-29
Added information on next steps after creating a release tarball [ci skip]

jougs:
The title says it all.

@heplesser: I guess one reviewer is enough, as this is only a minor extension.


tammoippen:
Very nice script. :+1: from my side.

heplesser:
:+1:

PR #143, opened 2015-11-02
Fix process_static_analysis.sh

tammoippen:
If all formats are ok, the static analysis section does not end with `+format_error_files=`. Also with #119 the static code analysis is moved before building NEST. This means, if static analysis parsing reads in all of the rest, the script will error. This PR sets a proper stop position for static code analysis.


heplesser:
:+1:

jougs:
Seems to make sense. Merging!

Issue #141, opened 2015-11-01
Error with non default tics_per_ms


frankmichler:
# Error with non default tics_per_ms
## Error description
When compiled with --with-tics-per-ms=4, on NEST import a NEST error
occurs in constructor of node class "correlospinmatrix_detector".

```
Nov 01 19:48:14 SLIInterpreter [Error]: 
    An error occured while loading module NEST Standard Models Module

Nov 01 19:48:14 SLIInterpreter [Error]: 
    InvalidDefaultResolution

Nov 01 19:48:14 SLIInterpreter [Error]: 
    The default resolution of 0.25 ms (= 1 tics = 1 step) is not consistent with the value 0 ms (= 0 tics = 0 steps) of 
    property 'delta_tau' in model UnknownNode.
    This is an internal NEST error, please report it at https://github.com/nest/nest-simulator/issues
```
## Root cause
in correlospinmatrix_detector.cpp, default constructor uses 0.1 for delta_tau, obviously assuming tics-per-ms=10.

```cpp
 nest::correlospinmatrix_detector::Parameters_::Parameters_()
  : delta_tau_( Time::ms( 0.1 ) )
  , tau_max_( 10 * delta_tau_ )
  , Tstart_( Time::ms( 0.0 ) )
  , Tstop_( Time::pos_inf() )
  , N_channels_( 1 )
{
}
```

## Proposed solution
Use 'Time::min()' instead of '0.1'

frankmichler:
Time::min() was wrong. I misunderstood it to be the length of the minimal time step, but it seems to be the lowest possible time value which is a huge negative number.

This should be the solution which gives the lenght of a step in ms.
```cpp
delta_tau_( Time::ms( Time::get_tics_per_step() / Time::get_tics_per_ms() ) )
```
For the default values tics_per_step==100 and tics_per_ms==1000. this should result in 0.1
Implemented in commit 4dcad7abf4b7aac4b944ac426fa43c1064a2113b pull request #144

frankmichler:
Pull request #144 with the proposed solution was merged in commit d971005728061322214dd706adebc8e0d1653991
Therefore I close the issue.

PR #144, opened 2015-11-02
Fix issue #141: Error with non default tics_per_ms

frankmichler:
When compiled with --with-tics-per-ms=4 --with-tics-per-step=1, on NEST import an error occurs in constructor of node class "correlospinmatrix_detector".

"The default resolution of 0.25 ms (= 1 tics = 1 step) is not consistent with the value
0 ms (= 0 tics = 0 steps) of property 'delta_tau' in model UnknownNode."

### DIFF
--- a/models/correlomatrix_detector.cpp
+++ b/models/correlomatrix_detector.cpp
@@ -35,7 +35,7 @@
  * ---------------------------------------------------------------- */
 
 nest::correlomatrix_detector::Parameters_::Parameters_()
-  : delta_tau_( Time::ms( 0.5 ) )
+  : delta_tau_( 5 * Time::get_resolution() )
   , tau_max_( 10 * delta_tau_ )
   , Tstart_( Time::ms( 0.0 ) )
   , Tstop_( Time::pos_inf() )
--- a/models/correlospinmatrix_detector.cpp
+++ b/models/correlospinmatrix_detector.cpp
@@ -35,7 +35,7 @@
  * ---------------------------------------------------------------- */
 
 nest::correlospinmatrix_detector::Parameters_::Parameters_()
-  : delta_tau_( Time::ms( Time::get_tics_per_step() / Time::get_tics_per_ms() ) )
+  : delta_tau_( Time::get_resolution() )
   , tau_max_( 10 * delta_tau_ )
   , Tstart_( Time::ms( 0.0 ) )
   , Tstop_( Time::pos_inf() )
--- a/models/correlospinmatrix_detector.cpp
+++ b/models/correlospinmatrix_detector.cpp
@@ -35,7 +35,7 @@
  * ---------------------------------------------------------------- */
 
 nest::correlospinmatrix_detector::Parameters_::Parameters_()
-  : delta_tau_( Time::ms( 0.1 ) )
+  : delta_tau_( Time::ms( Time::get_tics_per_step() / Time::get_tics_per_ms() ) )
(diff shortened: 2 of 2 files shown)


jougs:
@frankmichler: Thanks for your contribution!

Here's just a short word on the procedure: one of the NEST core developers will have a look at this pull request soon and we'll come back to you after a review. It is not a problem at this time if checks fail. You don't have to close the request, but instead you can add new commits to your branch in order to fix problems or reply to comments by the reviewers.

frankmichler:
I closed pull request #142 because I realized that the solution was wrong. 
```
NEST Testsuite Summary
----------------------
  NEST Executable: /home/travis/build/nest/nest-simulator/result/bin/nest
  SLI Executable : /home/travis/build/nest/nest-simulator/result/bin/sli
  Total number of tests: 199
     Passed: 76
     Failed: 123 
```
So I deleted the branch, and created it new with the solution I think is correct.
Now pull request #144  tests are passing:

```
NEST Testsuite Summary
----------------------
  NEST Executable: /home/travis/build/nest/nest-simulator/result/bin/nest
  SLI Executable : /home/travis/build/nest/nest-simulator/result/bin/sli
  Total number of tests: 199
     Passed: 199
     Failed: 0 
```

But the Travis build failes because of a log parsing error:

```
The command "python extras/parse_travis_log.py build.sh.log" exited with 1.
```

But good to know that next time I should make my corrections by just adding a new commit instead of deleting and recreating branch and pull request.

jougs:
@mhelias: can you please have a look at this? Especially, I am afraid that the division in the proposed solution can lead to nasty side effects.

frankmichler:
@jougs: what kind of side effects do you mean? There are only two things that come to my mind:
1. division by zero
2. accidental integer division with result zero

Both things are safe in this case:
1. Time::get_tics_per_ms() should never return zero. If someone configured "--with-tics-per-ms=0" I can't imagine that anything works at all.
2.  Time::get_tics_per_ms() returns an integer type. So there is cause to worry. But Time::get_tics_per_ms() has return type double_t. So the division should always be performed as an floating point division.

And in the default case, the result should be the same as before, because nest_time.cpp defines:
```cpp
#ifndef HAVE_TICS_PER_MS
#define CONFIG_TICS_PER_MS 1000.0
#endif

#ifndef HAVE_TICS_PER_STEP
#define CONFIG_TICS_PER_STEP 100
#endif
```

The result should be 0.1, which is the same as the literal "0.1" which is in the current code.
Am I missing something important here?

heplesser:
@frankmichler @jougs @mhelias The documentation states that `delta_tau` must be a multiple of the resolution, and the old default is just the resolution, but hard-coded as `0.1`, which is the resolution in the default case. The correct way to introduce it into the constructor is
```C++
delta_tau_( Time::get_resolution() )
```
This will always work and we need not worry about any side effects.

Similar issues occur in `correlomatrix_detector`, where the doc states that `delta_tau` has to be an odd multiple of the resolution, and `0.5` is hardcoded. I'd suggest
```C++
delta_tau_( 5 * Time::get_resolution() )
```

Finally, `correlation_detector` has
```C++
delta_tau_( Time::ms(1.0) )
```
I would argue for leaving that as it is---anyone choosing a time base that cannot represent 1 ms is in a state of sin.

@mhelias Any additional thoughts?

jougs:
@mhelias likes the change. Thus merging.

Issue #82, opened 2015-08-14
`parrot_neuron` should exploit spike multiplicity


heplesser:
`parrot_neuron` currently emits spikes using a loop

```
for ( ulong_t i_spike = 0; i_spike < current_spikes_n; i_spike++ )
  network()->send( *this, se, lag );
```

This is inefficient---why don't we use multiplicity and make just a single `network->send()` call?

flinz:
Since i am already working on the parrot_neuron, I changed this to use multiplicity (see the branch https://github.com/flinz/nest-simulator/tree/issue-parrot_neuron-multiplicity) and added a test to confirm. 

This branch is based off my branch already containing new tests for the parrot neuron (https://github.com/flinz/nest-simulator/tree/parrot-multiport), so I will open a pull request once that one is merged.

flinz:
@heplesser I think this can be closed, since a fix was merged.

jougs:
Definitely.

Issue #77, opened 2015-08-14
stdp synapses on pp_psc_delta neuron are not plastic


mdeger:
stdp_synapses onto pp_psc_delta neurons are apparently not plastic at the moment. I have made a simple example simulation script which is available here: 
https://github.com/mdeger/nest-simulator/blob/master/testsuite/manualtests/test_pp_psc_delta_stdp.py

The result of the script is shown in the attached pictures. Both neurons spike shortly after each spike of a sequence of spikes from the same presynaptic neuron. However, only the iaf_psc_delta neuron changes its synaptic weight, pp_psc_delta's weight is unchanged.

The only difference of the two models that I found, with respect to STDP behavior, is that archiver_length of the neurons is different. Both are instances of ArchivingNode, and this is important for STDP. However, I do not know how this different archiver_length may occur. I speculate that the problem occurs when the STDP connection is instantiated, and somehow the pp_psc_delta neuron does not increment its connection counter or so, but the iaf_psc_delta does it correctly.

Any help on the issue is appreciated.

![test_pp_psc_delta_stdp_fig1](https://cloud.githubusercontent.com/assets/11544025/9271562/338922ae-427b-11e5-91ad-5b3c50f8f1b3.png)
![test_pp_psc_delta_stdp_fig2](https://cloud.githubusercontent.com/assets/11544025/9271764/d48945de-427c-11e5-9717-79d8d0fc33b8.png)

mdeger:
the second picture was missing:

![test_pp_psc_delta_stdp_fig2](https://cloud.githubusercontent.com/assets/11544025/9271722/6516f502-427c-11e5-8e55-cf8ec31f43eb.png)

heplesser:
@mdeger Thank you for reporting this. I fear you have discovered a proper bug. The `update()` method of `pp_psc_delta` does not call `set_spiketime()` when emitting a spike, so that the synapse gets the impression that the neuron has never fired.

The same seems to be true for the following models:

- `ginzburg_neuron` and `mcculloch_pitts` (based on `binary_neuron`)
- `parrot_ps`
- `pp_psc_delta`
- `pp_pop_psc_delta`
- possibly `volume_transmitter`, if the synapse onto the volume transmitter should support plasticity

None of our tests seem to detect this problem. I will try to find time later today to write a SLI-regressiontest for this later, the main point would be to test that the `lastspike` entry is updated.

[...]

PR #146, opened 2015-11-03
Fix issue #82: parrot neuron now emits multiple spikes via spike multiplicity instead of a loop

flinz:
Addresses issue https://github.com/nest/nest-simulator/issues/82.
Added test for parrot neuron that checks for correct spike repetition with multiplicity.

### DIFF
--- a/models/parrot_neuron.cpp
+++ b/models/parrot_neuron.cpp
@@ -56,12 +56,16 @@ parrot_neuron::update( Time const& origin, const long_t from, const long_t to )
     const ulong_t current_spikes_n = static_cast< ulong_t >( B_.n_spikes_.get_value( lag ) );
     if ( current_spikes_n > 0 )
     {
-      // create a new SpikeEvent and set its multiplicity
+      // create a new SpikeEvent, set its multiplicity and send it
       SpikeEvent se;
       se.set_multiplicity( current_spikes_n );
-
       network()->send( *this, se, lag );
-      set_spiketime( Time::step( origin.get_steps() + lag + 1 ) );
+
+      // set the spike times, respecting the multiplicity
+      for ( ulong_t i = 0; i < current_spikes_n; i++ )
+      {
+        set_spiketime( Time::step( origin.get_steps() + lag + 1 ) );
+      }
     }
   }
 }
--- a/models/parrot_neuron.cpp
+++ b/models/parrot_neuron.cpp
@@ -51,18 +51,16 @@ parrot_neuron::update( Time const& origin, const long_t from, const long_t to )
   assert( to >= 0 && ( delay ) from < Scheduler::get_min_delay() );
   assert( from < to );
 
-  SpikeEvent se;
-
   for ( long_t lag = from; lag < to; ++lag )
   {
     const ulong_t current_spikes_n = static_cast< ulong_t >( B_.n_spikes_.get_value( lag ) );
-
     if ( current_spikes_n > 0 )
     {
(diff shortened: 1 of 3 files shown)


heplesser:
Looks good, I created a PR to @flinz repo with improved documentation and one more test.

models/parrot_neuron.cpp
```diff
@@ -51,18 +51,16 @@ parrot_neuron::update( Time const& origin, const long_t from, const long_t to )
...
-      for ( ulong_t i_spike = 0; i_spike < current_spikes_n; i_spike++ )
-      {
-        network()->send( *this, se, lag );
-      }
+      // create a new SpikeEvent and set its multiplicity
+      SpikeEvent se;
+      se.set_multiplicity( current_spikes_n );
+
+      network()->send( *this, se, lag );
       set_spiketime( Time::step( origin.get_steps() + lag + 1 ) );
```
heplesser: I need to double-check whether `set_spiketime()` needs to be called once per outgoing spike, i.e., as many times as multiplicity indicates.

flinz: https://github.com/nest/nest-simulator/issues/77#issuecomment-148698719 seems to suggest that this is indeed the case.

this would result in 
```c++
for (ulong_t i=0;  i<current_spikes_n; i++ )
    set_spiketime( Time::step( origin.get_steps() + lag + 1 ) );
```
with which we're almost back to the old form -- although it's a little cleaner with the handling of multiplicity.

It should be noted that the [original model](https://github.com/flinz/nest-simulator/blob/c229212d987023246b2fd7b1f072680b33e79e40/models/parrot_neuron.cpp) did not set the spike_times multiply.

jougs: I agree. Can you please add this change?

models/parrot_neuron.h
```diff
@@ -60,22 +52,19 @@ Receives: SpikeEvent
 Sends: SpikeEvent
 
 Parameters:
-
 No parameters to be set in the status dictionary.
 
-References:
-No references
-
 Author:  May 2006, Reichert, Morrison
```
jougs: This should now be
```
Author: David Reichert, Abigail Morrison, Alexander Seeholzer, Hans Ekkehard Plesser
FirstVersion: May 2006
```
```

File: models/parrot_neuron.h [hunk 3]
```diff
@@ -60,22 +52,19 @@ Receives: SpikeEvent
...
- * stripped of the dynamics and unneeded features.
- * Instead of the accumulated weigths of the incoming spikes the
+ * The parrot neuron emits one spike for every incoming spike,
+ * but may use multiplicity to indicate number of spikes in a single
+ * time step.
+ * Instead of the accumulated weigths of the incoming spikes, the
  * number of the spikes is stored within a ring buffer.
+ *
  * \author David Reichert
  * \date may 2006
```
jougs: Please remove the date and author lines here.

jougs:
Very nice work. I'm all :+1: once my comments above have been addressed.

models/parrot_neuron.cpp
heplesser: @flinz I agree with @jougs and after more meditation over the spike archiving and plasticity (see #77) think that we do have to call `set_spiketime()` once for every spike. One this PR is merged, I will extend `set_spiketime()` to accept multiplicity as argument in a separate PR, but I don't want to delay this PR further.

heplesser:
:+1: once the loop around `set_multiplicity()` is in place and Jochen's comments are addressed.

PR #148, opened 2015-11-04
Improve hpc_benchmark.sli

mschmidt87:
@suku248 discovered that the hpc_benchmark.sli script, which is supposed to represent the benchmark script used in the recent NEST Technology papers, lacked the functionality of randomizing initial membrane potential, which we used and described in Kunkel et al. (2014).

Therefore, I implemented this option.

Furthermore, @suku248 and I discovered two inconsistencies:
- The simulation time in hpc_benchmark.sli was set to 100. ms but for the benchmarks in the publications, we always used 1000 ms. Perhaps this was done to make simulating the script on a local laptop more convenient, but it should be set to the original simulation time to avoid confusion, in our opinion.

- The pre-simulation time is fixed to 100. ms in hpc_benchmark.sli. However, for the publication we simulated 1 timestep + 10 min_delay periods = 15.1 ms (for the default settings). I did not apply any changes, should I change this or should we leave it at a fixed value of 100. ms. For simplicity, I would leave things as they are, because the pre-simulation time is just meant to skip the initial transient in the network.

@suku248 , @diesmann , please take a look at this pull-request.

### DIFF
--- a/examples/nest/hpc_benchmark.sli
+++ b/examples/nest/hpc_benchmark.sli
@@ -50,7 +50,7 @@
 
 % -------------------------------------------------------------------------------------
 
-%%% DEFINE FUNCTION TO CONVERT SYNAPSE WEIGHT FROM mV to pA %%%%%%%%%%
+%%% Define function to convert synapse weight from  mV to pA %%%%%%%%%%
 
 /ConvertSynapseWeight
 {
@@ -86,7 +86,7 @@ def
 
   /model_params
    <<
-   % Set variables for iaf_psc_exp
+   % Set variables for iaf_neuron
     /E_L     0.0  mV  % Resting membrane potential [mV]
     /C_m   250.0  pF  % Capacity of the membrane [pF]
     /tau_m  10.0  ms  % Membrane time constant [ms]
--- a/examples/nest/hpc_benchmark.sli
+++ b/examples/nest/hpc_benchmark.sli
@@ -81,7 +81,7 @@
 
   /mean_potential 5.7 mV
   /sigma_potential 7.2 mV
- 
+  /randomize_Vm true
 
   /delay  1.5 ms         % synaptic delay, all connections [ms] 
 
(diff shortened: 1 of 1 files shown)


jougs:
Thanks for this contribution. The changes look good to me and I'm :+1: for merging the current code.

However, I have one additional request: as you probably know most about this script, could you please also look into trac.950 and add a corresponding commit to this PR? That'd be great!

mschmidt87:
Thanks for your comments, 
@suku248 : I forgot to mention that the values of mean and sigma potential given in Kunkel et al. (2014) seem to be wrong. I checked the values in several of the benchmark scripts that we actually used, and there they were as they are now in hpc_benchmark.sli.
So, it seems to an error in the paper. What to do?

suku248:
In this case, we keep 5.7 and 7.2 as these are also the original values that were used in Morrison et al (2007).

mschmidt87:
* implemented the changes suggested by @suku248 .
* I also addressed trac.950: Actually in the ticket, @diesmann thought that the script uses exponential PSC, but it uses `iaf_neuron`, which has alpha-shaped PSC. Thus, I could take basically take over the functionality from brunel-2000.sli to convert synaptic weights from mV to pA. Moreover, I added an explanatory section to clarify why the value of tau_syn was chosen.

suku248:
@mschmidt87 : Thanks for taking care of this. :+1: for the merge.

examples/nest/hpc_benchmark.sli
```diff
@@ -41,23 +41,64 @@
...
+  /tausyn set
+  
+  (
+  a = taumem / tausyn;
+  b = 1.0 / tausyn - 1.0 / taumem;
+  ) execmath
+  
+} def
+
+% Inverting this equation leads to tau_syn = 0.32582722403722841 ms, as specified in model_params below.
```
jougs: This comment does not make sense to me (anymore), as you remove the magic number in line 88 to 90 below. Can you please change it to point to the use of the `ConvertSynapseWeight` and `PSP_rise_time` functions or rather remove it altogether? I am happy with removing it.

examples/nest/hpc_benchmark.sli
```diff
@@ -79,15 +120,18 @@
...
+                          % here were used for the benchmark on JUQUEEN.
 
   /delay  1.5 ms         % synaptic delay, all connections [ms] 
 
-   % synaptic strengths
-  /JE 325.78288268410768 0.14 mul pA    %peak of EPSC
-
+   % synaptic weight
+  /JE 0.14 mV %peak of EPSP
+
```
jougs: Removing trailing spaces

jougs:
:+1: from me once my minor comments are addressed. Many thanks for the work!

examples/nest/hpc_benchmark.sli
mschmidt87: I added this comment to explain the choice of tau_syn which I haven't removed. The reason is that the function I defined here (`PSP_rise_time`) computes the rise-time of the PSP in dependence of tau_syn, i.e.,
rise_time PSP = f(tau_syn).
Below we define tau_syn such that rise_time PSP = 1.700759 ms. Thus, I could also define the inverse function here, which would involve some more math. Therefore, I decided to define this function to inform the user how we, in principle, arrived at tau_syn=0.32... ms.

jougs: Ah, OK. Thanks for the clarification. The comment lead me to think that I could find that exact value below. What about just adding a clarification like this verbally without the inverted function and the math? If it left me confused, it might do the same for others...

PR #150, opened 2015-11-05
Add stdp_triplet_synapse from Pfister & Gerstner (2006)

zifeo:
This PR add the connection described in Pfister & Gerstner (2006) with some tests.
The work is based on the file @abigailm gave us and was supervised by @flinz.

**Points to be discussed**

*Nearest-spike pairing*
This would require a capping of the postsynaptic trace to a maximum of 1. This is not possible without changing the way that archiving_node updates `Kminus` and `triplet_Kminus`.

*Tests involving the postsynaptic decay*
This would require access to the postsynaptic traces (states `Kminus` and `triplet_Kminus` of the archiving_node). This could be added. However, the correct decay of these states should be tested in the archiving_node itself.

*Naming conventions*
We chose to stick to similar names across all variables, i.e. Kplus, Kminus, etc..
In addition `Kminus` might be named `K_minus` according the [naming conventions](https://nest.github.io/nest-simulator/variables_parameters_naming) but the name in archiving node is `Kminus` without the underscore. To keep coherence we used the same rule in the connection. However `triplet_var` is used as a prefix in archiving node whereas it might make more sense to use it as a suffix, i.e. `var_triplet`.

*Inlining facilitate and depress*
Should the `facilitate_` and `depress_` be inlined or not? Pros/cons?

*Max weight parameters*
One could add additionally options to update the weights with either a hard maximal weight bound or multiplicative updating of weights. 

*Negative weight*
Currently the weights are set to a minimum of 0, so no negative weights are technically possible.

*Parameters names*
Should they be defined in `nest_names.h` or only the ones that are regularly re-used?

### DIFF
--- a/models/stdp_triplet_connection.h
+++ b/models/stdp_triplet_connection.h
@@ -1,0 +1,333 @@
+/*
+ *  stdp_triplet_connection.h
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+#ifndef STDP_TRIPLET_CONNECTION_H
+#define STDP_TRIPLET_CONNECTION_H
+
+/* BeginDocumentation
+  Name: stdp_triplet_synapse - Synapse type for spike-timing dependent 
+  plasticity accounting for spike triplets as described in [1].
+
+  Description:
+   stdp_triplet_synapse is a connector to create synapses with spike time 
+   dependent plasticity accounting for spike triplets (as defined in [1]). 
+   
+   Here, a multiplicative weight dependence is added (in contrast to [1]) 
+   to depression resulting in a stable weight distribution.
+  
+  Parameters:
+
+   tau_plus     time constant of STDP window, potentiation 
+                (tau_minus defined in post-synaptic neuron)
+   tau_x        time constant of triplet potentiation
+   tau_y        time constant of triplet depression
+   A_2p         weight of pair potentiation rule
+   A_2m         weight of pair depression rule
+   A_3p         weight of triplet potentiation rule
+   A_3m         weight of triplet depression rule
+
+
+  References:
+
+   [1] J.-P. Pfister & W. Gerstner (2006) Triplets of Spikes in a Model 
+   of Spike Timing-Dependent Plasticity.  The Journal of Neuroscience 
+   26(38):9673-9682; doi:10.1523/JNEUROSCI.1425-06.2006
+
+   Transmits: SpikeEvent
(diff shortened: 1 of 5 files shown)


jougs:
@mhelias, can you please review this code?

models/stdp_triplet_connection.h
```diff
@@ -0,0 +1,355 @@
...
+      neuron states Kminus_ and triplet_Kminus_ which decay on time-constants
+      tau_minus and tau_minus_triplet, respectively. These two time-constants can
+      can be set as properties of the postsynaptic neuron.
+    - This version implements the 'all-to-all' spike interaction of [1]. The 'nearest-spike'
+      interaction of [1] can currently not be implemented without changing the postsynaptic
+      archiving-node (clip the traces to a maximum of 1).
+
+  FirstVersion: Nov 2007
+  Author: Moritz Helias, Abigail Morrison, Eilif Muller, Alexander Seeholzer, Teo Stocco
+  SeeAlso: synapsedict, stdp_synapse, static_synapse
```
jougs: Please add `stdp_triplet_synapse_hpc` to the `SeeAlso` field.

flinz: @jougs ``stdp_triplet_synapse_hpc`` is not yet in the main branch.. Wouldn't it make more sense to add this to the SeeAlso field when the model is added?

jougs: I don't understand the question. It is also added in this PR (see [here](https://github.com/zifeo/nest-simulator/blob/triplet_stdp/models/modelsmodule.cpp#L357)). Or isn't this the point you're making. *confused*

jougs:
Regarding the naming conventions: I think that you should leave `Kminus`, but change `triplet_var` to `var_triplet`.

models/stdp_triplet_connection.h
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  void
+  set_weight( double_t w )
+  {
+    weight_ = w;
+  }
+
+private:
+  inline double_t // TBD
```
jougs: Was the `// TBD` leftover from earlier or is it actually still hinting at something to be done?

jougs:
Regarding inlining of `facilitate_` and `depress_`: I'd inline them, as they are reasonably short.

models/stdp_triplet_connection.h
flinz: Ah, sorry, my bad. I wasn't aware of the double registering in ``modelsmodule``.
Will add the reference.

models/stdp_triplet_connection.h
flinz: The ``TBD`` hints at points that were to be discussed before merge.

jougs: Ah! stupid me! See the answer below, then!

flinz: Since the points were raised already in the PR comment above, I'll remove the ``TBD``'s from the code.

flinz:
@jougs One of the travis jobs failed during startup.. could you rerun it please?

zifeo:
@mhelias Any progress on the review?

flinz:
@jougs @abigailm @mhelias: would be nice if this still made it into nest 2.10

heplesser:
The code looks clean and the tests are comprehensive, from that perspective :+1:. But I know too little about plasticity to evaluate whether the code is actually doing the right thing.

abigailm:
Sorry for the very long delay. I am happy with the plasticity code, but please add a maximum weight parameter to prevent weight explosion, similar to stdp_connection.h, and clip the weights to it in the facilitate method if the weight exceeds it.

mhelias:
Looks ok to me, but would like my name to be deleted from the list of authors, because I never saw the code before this review.

zifeo:
Updates are done!

flinz:
@mhelias @abigailm I guess this addresses all open issues?

abigailm:
:+1: and merging

PR #155, opened 2015-11-16
add check for openmp, protect openmp call on K when building without

JanneM:
When building nest without OpenMP on K, omp_get_thread_num() is undefined and the build will fail. Added checks to make sure we're using the right code path for this case.


tammoippen:
:+1:

jougs:
@JanneM: Thanks for the contribution. I'm also :+1: and thus merging.

PR #157, opened 2015-11-17
Tools and content for the NEST documentation

steffengraber:
Added some static website content in markdown and some tools for creating a user-friendly documentation from source files.
This is the first initial release. 
The plan is:
- using the .md files for static documentation and the github api to connect the content to the website.
- using the tools to generate an up to date and dynamic documenation from information found in the source files.


tammoippen:
Hi @steffengraber ,

The python files in `extras/userdoc/generator/` need the NEST copyright header, then travis will be satisfied: the first 21 lines of each python file will be matched against `doc/copyright_header.py`.

steffengraber:
Hi @tammoippen 

done.

tammoippen:
Hi @steffengraber ,

I tried the generator scripts in `userdoc/generator` and they they are really nice :smile: ! I have just a few comments (I would write them as line comments, but Github does not display the changes for me ... :cry:) :
* `generate_example2png.sh`
  * You ask for the NEST install path, to set the `PYTHONPATH` in the script. Depending on the system and the Python installation, this path can vary.  It would be better, if you expect a working NEST installation: the user should `source <install-path>/bin/nest_vars.sh` before.
  * `sed` works differntly on GNU and OSX/FreeBSD (see [here](http://www.grymoire.com/Unix/Sed.html#uh-62h))
  * in line 32: instead of `sed '/savefig/s/\.py//' -i $value` you should use `sed -i '' '/savefig/s/\.py//' $value`
  * in line 36 & 38: use `sed -i '' '$d' $value`
  * Not every system comes with `recode`. You should state the dependancy somewhere.

* `generate_help2jsonhtml.py`
  * In line 97 you hardcode the seperator '/nest-2.8.0/' to get the relative path of a file. This is problematic, if sources are not in that folder. Since you know the absolute path of the sources (variable `path`), I think it would be better to give `coll_data` the correct `relfile` (line147):

          all_data = coll_data(keywords, documentation, num, file.replace(path, ""), f_file)
          # instead of
          all_data = coll_data(keywords, documentation, num, file, f_file)
And remove the two lines 97 and 98.

* `generate_example2ipynb.py`
  * Before the imports, you could include `%matplotlib inline`. This makes plots appear inside the notebook.
  * The docstring of functions split the function signature and the function body into two cells. This is an error for ipython notebook.
  * The same goes for class definitions: they cannot be put into different cells.

mschmidt87:
@jougs , please take a look at this pull-request.

jougs:
I'll have a look at this asap and will comment here and probably send a PR against the branch of @steffengraber with larger changes.

Issue #121, opened 2015-10-12
Use of synapse types in NEST + PyNN Issue #377


mschmidt87:
This issue is related to the pyNN issue 377 "PyNN exhausts NEST 2.6/2.8 synapse model storage" (https://github.com/NeuralEnsemble/PyNN/issues/377).

1. In pyNN, whenever a pyNN.projection is created to establish connections, a new synapse type in NEST is created, which is a misuse of NEST's synapse type. A synapse type in NEST specifies the dynamical model of synapse (e.g. static, STDP etc.) and not the projection/connection it belongs to. It also does not specify whether the synapse is excitatory or inhibitory, because this is independently decided by the synaptic weights. 
This causes pyNN to exhaust NEST's maximal amount of synapse types (256).

2. Some of the NEST examples create new synapse types to distinguish between excitatory and inhibitory synapses, which also a misuse of the synapse type concept. We should discourage NEST users to use NEST synapse types for such things like the differentiation between exc. and inh. synapses, if the only difference between them are the positive/negative synaptic weights.

Examples are the brunel network scripts.

jschuecker:
Solution and implementation plan:
We will use a decorator around a synapse type, that optionally gives every synapse a projection id as an integer. This projection id can be optionally passed to the Connect function, so that this feature will be optional in pyNEST while in PyNN it will be set automatically.
This is independent of pyNN and would cost one integer for each synapse but adds the idea of projections in NEST, if requested by the user (of course the user has to pay for this feature with memory costs).
This could be implemented using a template and will be done by @tammoippen.
PyNN then can identify a projection by filtering synapses for this projection id, which will be implemented by @mschmidt87 and by me.

heplesser:
Closing #121 by #158

PR #158, opened 2015-11-17
Labeled Synapses

tammoippen:
In response to #121.

This PR introduces a decorator for connections that adds an integer label to the decorated synapse. 
* See the template class `ConnectionLabel`: the template argument can be any synapse model. The `ConnectionLabel` then subclasses this model and overrides the `get_status`, `set_status` and `get_label` methods. The `get_label` method is added to `Connection` as well, returns `UNLABELED_CONNECTION` (-1) and ensures valid return values for all synapse models.
* See `connector_model_impl.h`: all synapse models that are registered via `register_connection_model` (i.e. no `gap_junction` model) and which are not postfixed with `_hpc`, will also register a decorated version of this synapse type. The model name is postfixed with `_lbl`.

        SLI ] /iaf_neuron 2 Create ;
        SLI ] [1 2] [2 1] << /rule /one_to_one >> << /model /static_synapse_lbl /synapse_label 5 >> Connect
* The function `GetConnections` allows to search for synapses with a certain label:

        SLI ] << /synapse_label 5 >> GetConnections /proj_syn Set
        SLI ] proj_syn ==
        [<1,2,0,1,0> <2,1,0,1,0>]
        SLI ] proj_syn 0 get GetStatus info
        --------------------------------------------------
        Name                     Type                Value
        --------------------------------------------------
        delay                    doubletype          1
        receptor                 integertype         0
        sizeof                   integertype         40
        source                   integertype         1
        synapse_label            integertype         5
        synapse_model            literaltype         static_synapse_lbl
        target                   integertype         2
        weight                   doubletype          1
        --------------------------------------------------
        Total number of entries: 8
This allows to assign a label / projection id to a projection between neuron populations, retrieve all synapses of this projection and, get and set properties of these synapses.
* As a consequence, all synapse models with label use more memory (one `long_t` more).

### DIFF
--- a/lib/sli/nest-init.sli
+++ b/lib/sli/nest-init.sli
@@ -482,7 +482,8 @@ def
    Synopsis:
    << /source [sgid1 sgid2 ...] 
       /target [tgid1 tgid2 ...]
-      /synapse_model /smodel    >> GetConnections -> [ conn1 conn2 ... ]
+      /synapse_model /smodel    
+      /synapse_label label      >> GetConnections -> [ conn1 conn2 ... ]
 
    Parameters:
    A dictionary that may contain the following fields (all are optional):
@@ -491,7 +492,9 @@ def
    /target  - array with GIDs of post-synaptic nodes whose connections are sought.
               If not given, all neurons are searched as targets.
    /synapse_model - literal specifying synapse model
-                    If not given, connections of all synapse models are returned. 
+                    If not given, connections of all synapse models are returned.
+   /synapse_label - integer specifying synapse label
+                    If not given, connections of all synapse labels are returned.
 
    Description:
    1. If called with an empty dictionary, GetConnections returns all connections of the 
@@ -501,16 +504,19 @@ def
    3. The optional dictionary elements /source and /target can be used to filter 
       for specific pre- and post-synaptic neurons, respectively.
    4. The optional parameter /synapse_model can be used to filter for a specific synapse model.
-   5. In a parallel simulation, GetConnections only returns connections with *targets*
+   5. The optional parameter /synapse_label can be used to filter for a specific synapse label.
+   6. In a parallel simulation, GetConnections only returns connections with *targets*
       on the MPI process executing the function. 
 
    Remarks:
    1. See synapsedict for the synapse-model-id's for all synapse models.
-   2. The "port" enumerates connections per source, thread and synapse model. It is
+   2. The /synapse_label property can only be assigned to synapse models with its name ending
+      with '_lbl'. All other synapses have the default synapse_label UNLABELED_CONNECTION (-1).
+   3. The "port" enumerates connections per source, thread and synapse model. It is
       mainly important for NEST internally.
-   3. In OpenMP mode, GetConnections works thread-parallel for better performance.
-   4. Connection objects can be converted to SLI lists with cva.
-   5. Connection objects can be passed to SetSynapseStatus, GetSynapseStatus, and DataConnect
+   4. In OpenMP mode, GetConnections works thread-parallel for better performance.
+   5. Connection objects can be converted to SLI lists with cva.
+   6. Connection objects can be passed to SetSynapseStatus, GetSynapseStatus, and DataConnect
 
    SeeAlso: DataConnect, SetSynapseStatus, GetSynapseStatus, synapsedict
 */
--- a/lib/sli/nest-init.sli
+++ b/lib/sli/nest-init.sli
@@ -510,7 +510,7 @@ def
 
    Remarks:
    1. See synapsedict for the synapse-model-id's for all synapse models.
-   2. The /synapse_label property can only be assigned to synapse models with its name ending
+   2. The /synapse_label property can only be assigned to synapse models with names ending
       with '_lbl'. All other synapses have the default synapse_label UNLABELED_CONNECTION (-1).
    3. The "port" enumerates connections per source, thread and synapse model. It is
       mainly important for NEST internally.
--- a/nestkernel/Makefile.am
+++ b/nestkernel/Makefile.am
@@ -37,7 +37,8 @@ libnest_la_SOURCES=\
 		communicator.h communicator_impl.h communicator.cpp\
 		sibling_container.h sibling_container.cpp\
 		subnet.h subnet.cpp\
-		connection.h connection_label.h\
+		connection.h\
+		connection_label.h\
 		common_properties_hom_w.h\
 		syn_id_delay.h\
 		connector_base.h connector_base.cpp\
--- a/nestkernel/Makefile.am
+++ b/nestkernel/Makefile.am
@@ -37,7 +37,7 @@ libnest_la_SOURCES=\
 		communicator.h communicator_impl.h communicator.cpp\
 		sibling_container.h sibling_container.cpp\
 		subnet.h subnet.cpp\
-		connection.h\
+		connection.h connection_label.h\
 		common_properties_hom_w.h\
 		syn_id_delay.h\
 		connector_base.h connector_base.cpp\
--- a/nestkernel/connection_label.h
+++ b/nestkernel/connection_label.h
@@ -48,10 +48,10 @@ public:
    */
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
-  long get_label() const;
+  long_t get_label() const;
 
 private:
-  long label_;
+  long_t label_;
 };
 
 template < typename Connection_t >
@@ -68,7 +68,7 @@ template < typename Connection_t >
 void
 ConnectionLabel< Connection_t >::set_status( const DictionaryDatum& d, ConnectorModel& cm )
 {
-  long lbl;
+  long_t lbl;
   if ( updateValue< long >( d, names::synapse_label, lbl ) )
   {
     if ( lbl >= 0 )
@@ -84,7 +84,7 @@ ConnectionLabel< Connection_t >::set_status( const DictionaryDatum& d, Connector
 }
 
 template < typename Connection_t >
-inline long
+inline long_t
 ConnectionLabel< Connection_t >::get_label() const
 {
   return label_;
--- a/nestkernel/connection_label.h
+++ b/nestkernel/connection_label.h
@@ -32,25 +32,24 @@ namespace nest
 {
 class ConnectorModel;
 
-template <class Connection_t>
-class ConnectionLabel: public Connection_t
+template < class Connection_t >
+class ConnectionLabel : public Connection_t
 {
 public:
-  
   /**
    * Get all properties of this connection and put them into a dictionary.
    */
   void get_status( DictionaryDatum& d ) const;
-  
+
   /**
    * Set properties of this connection from the values given in dictionary.
    *
    * @note Target and Rport cannot be changed after a connection has been created.
    */
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
-  
+
   long get_label() const;
-  
+
 private:
   long label_;
 };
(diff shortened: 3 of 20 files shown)


tammoippen:
@mschmidt87 and @jschuecker : please have a look at this code and usage pattern and tell me if this will work for the PyNN integration from #121 .

lib/sli/nest-init.sli
```diff
@@ -501,16 +504,19 @@ def
...
    4. The optional parameter /synapse_model can be used to filter for a specific synapse model.
-   5. In a parallel simulation, GetConnections only returns connections with *targets*
+   5. The optional parameter /synapse_label can be used to filter for a specific synapse label.
+   6. In a parallel simulation, GetConnections only returns connections with *targets*
       on the MPI process executing the function. 
 
    Remarks:
    1. See synapsedict for the synapse-model-id's for all synapse models.
-   2. The "port" enumerates connections per source, thread and synapse model. It is
+   2. The /synapse_label property can only be assigned to synapse models with its name ending
```
heplesser: "models with names ending"

nestkernel/Makefile.am
```diff
@@ -37,7 +37,7 @@ libnest_la_SOURCES=\
 		communicator.h communicator_impl.h communicator.cpp\
 		sibling_container.h sibling_container.cpp\
 		subnet.h subnet.cpp\
-		connection.h\
+		connection.h connection_label.h\
```
heplesser: Put `connection_label.h` on a separate line. Only h, cpp and impl files belonging together should be combined on a single line.

nestkernel/connection_manager.cpp
```diff
@@ -330,14 +330,21 @@ ConnectionManager::get_connections( DictionaryDatum params ) const
...
   if ( not source_t.empty() )
     source_a = dynamic_cast< TokenArray const* >( source_t.datum() );
   if ( not target_t.empty() )
     target_a = dynamic_cast< TokenArray const* >( target_t.datum() );
 
+  if ( not label_token.empty() )
+  {
+    label = getValue< long_t >( label_token );
+  }
+
```
heplesser: I think you can do this much more easily, along the lines of
```C++
long_t label = UNLABELED_CONNECTION;
updateValue< long_t >( names::synapse_label, label );
```

nestkernel/connection_manager.h
```diff
@@ -94,6 +94,7 @@ class ConnectionManager
    * 'target' a token array with GIDs of target neuron.
    * If either of these does not exist, all neuron are used for the respective entry.
    * 'synapse_model' name of the synapse model, or all synapse models are searched.
+   * 'synapse_label' label (int) of the synapse, or all synapses are searched.
```
heplesser: You write `int` here, but it is a `long`, at least on the C++ level. Or is this SLI documentation?

nestkernel/nest.h
```diff
@@ -121,6 +121,8 @@ const index invalid_index = std::numeric_limits< index >::max();
 typedef unsigned char synindex;
 const synindex invalid_synindex = std::numeric_limits< synindex >::max();
 
+#define UNLABELED_CONNECTION -1
```
heplesser: @tammoippen Is it really necessary to use a C makro here instead of a C++ constant? And is `nest.h` really the best place to define the constant? Wouldn't `connection_label.h` be a better place, or would that require too many includes all across the code?

heplesser:
@tammoippen Nice work! I added a few comments above, the most critical one is the one on the definition of `UNLABELED_CONNECTION`.  I think it would also be useful to add a comment on the `_lbl` synapses in the documentation of `modeldict` and `nest.Models()`, so users understand what they see.

tammoippen:
@heplesser 
Thank you for your comments. I addressed them in the commit 4a78c28.

heplesser:
:+1: provided the tiny comments I just added are addressed as well.

nestkernel/connection_label.h
```diff
@@ -0,0 +1,99 @@
...
+  long_t lbl;
+  if ( updateValue< long >( d, names::synapse_label, lbl ) )
+  {
+    if ( lbl >= 0 )
+    {
+      label_ = lbl;
+    }
+    else
+    {
+      throw BadProperty( "Connection label must be positive." );
```
jschuecker: this should be "Connection label must not be negative", since zero is allowed, right?

jschuecker:
I also had a look at the code and tested it. Very nice work! Here some comments:

When I provide a float for the synapse type

`syn_dict = {'model': 'static_synapse_lbl', 'synapse_label': 1.1}`

I get the error message:

`pynestkernel.NESTError: BadProperty in Connect_g_g_D_D: Receptor type must be of type integer. ` Could you please adapt the error message to `Synapse label`.

Is there a particular reason that 'gap_junctions' cannot be labelled?  

The pyNEST function GetConnections does not accept the synapse label as an argument so far. 

The error message for negative labels should be "Connection label must not be negative", since zero is allowed. I added an inline-comment.

tammoippen:
@jschuecker 
I adjusted the error messages and implemented the Python side of `GetConnections`.

> Is there a particular reason that 'gap_junctions' cannot be labelled?

I am not sure. Maybe @mhelias can help here:  In the test `test_hh_psc_alpha_gap.sli` only the synapse model `gap_junction_lbl` succeeds. I think the problem is her:, when registering the `gap_junction` with `register_secondary_connection_model` (`connector_model_impl.h:508`), a static field is set:
 
    ConnectionT::EventType::set_syn_id( synid );

Then I would create the second `gap_junction_lbl` and set the syn_id of that EventType:

    ConnectionLabel< ConnectionT >::EventType::set_syn_id( synid );

but this overwrites the previous synid. I am not so familiar with the gap junction framework and left it untouched for this PR.

janhahne:
@tammoippen Regarding the gap junctions: I think you identified the reason correctly. The `synid` in the `GapJEvent` is used by the scheduler to distinguish which kind of `SecondaryEvent` is send (see `Scheduler::deliver_events_´` in `scheduler.cpp:1666-1694`). So far there was exactly one synapse model which used the `GapJEvent`, so the `synid` of this synapse model was "connected" to the `GapJEvent` class on registration.

Not sure if there is an easy fix to make this compatible.

janhahne:
Once #175 is merged, it should be possible to extend this on the gap junctions.

tammoippen:
I did a rebase on master to get #175. Now it is possible to also label secondary event connections. @jschuecker , @heplesser  and @janhahne , can you have a look again?

nestkernel/connection_label.h
```diff
@@ -0,0 +1,113 @@
...
+namespace nest
+{
+class ConnectorModel;
+
+/**
+ * Connections are unlabeled by default. Unlabeled connections cannot be specified
+ * as a search criterion in the `GetConnections` function.
+ * @see ConnectionLabel
+ */
+const long_t UNLABELED_CONNECTION = -1;
```
heplesser: Wouldn't it be even tidier to make this a `static const` in the `ConnectionLabel` class?

nestkernel/connection_label.h
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * The class ConnectionLabel enables synapse model to be labeled by a positive integer. The
+ * label can be set / retrieved with the `names::synapse_label` property in the parameter
+ * dictionary of `Set/GetStatus` or `Connect`.
+ * Using the `GetConnections` function, synapses with the same label can be specified.
+ *
+ * The name of synapse models, which can be labeled, end with '_lbl'.
+ * @see nest::ConnectionManager::get_connections
+ */
+template < class Connection_t >
```
heplesser: I think just as everywhere else, it should be `typename`, no `class`. An then the parameter should maybe be `ConnectionT`?

nestkernel/connection_label.h
```diff
@@ -0,0 +1,113 @@
...
+  long_t label_;
+};
+
+template < typename Connection_t >
+void
+ConnectionLabel< Connection_t >::get_status( DictionaryDatum& d ) const
+{
+  Connection_t::get_status( d );
+  def< long_t >( d, names::synapse_label, label_ );
+  // override, as the size changes here
```
heplesser: I do not quite understand this comment. Do you want to say that `ConnectionLabel` has a different size than `ConnectionT`?

nestkernel/connector_model_impl.h
```diff
@@ -558,32 +559,50 @@ GenericConnectorModel< ConnectionT >::delete_connection( Node& tgt,
 // by modules                                                                  //
 /////////////////////////////////////////////////////////////////////////////////
 
+inline bool
+ends_with( std::string const& value, std::string const& ending )
+{
+  if ( ending.size() > value.size() )
+    return false;
+  return std::equal( ending.rbegin(), ending.rend(), value.rbegin() );
+}
```
heplesser: This function could be useful elsewhere. Should it move to a `string_utils.h` or similar in `libnestutil`?

nestkernel/connector_base.h
```diff
@@ -132,12 +133,14 @@ class ConnectorBase
   virtual void get_connections( size_t source_gid,
     size_t thrd,
     synindex synapse_id,
+    long_t label,
```
heplesser: For symmetry with `synapse_id` and usage elsewhere, shouldn't this be `synapse_label` instead of just `label`? Applies to all overriding declarations, too.

nestkernel/connector_base.h
```diff
@@ -132,12 +133,14 @@ class ConnectorBase
...
     ArrayDatum& conns ) const = 0;
 
   virtual void get_connections( size_t source_gid,
     size_t target_gid,
     size_t thrd,
     size_t synapse_id,
+    long_t label,
```
heplesser: See above.

heplesser:
@tammoippen I added some comments concerning code tidiness. My main concern now is that there seems to be no test for the synapse labels. Could you create one that tests the various use cases?

jschuecker:
@tammoippen, @heplesser :+1: from my side, once we have a test.

janhahne:
@tammoippen The labeling of the secondary event connections looks fine to me. Once the other comments here are implemented :+1:

nestkernel/connection_label.h
tammoippen: This would require a template argument, whenever you want to access this field, like:

    ConnectionLabel< StaticConnection< TargetIdentifierPtrRport > >::UNLABELED_CONNECTION

I guess we should not go for this...

tammoippen:
@heplesser I pushed my work on the comments.

nestkernel/connection_label.h
heplesser: I agree!

> On 08 Dec 2015, at 13:50, Tammo Ippen <notifications@github.com> wrote:
> 
> In nestkernel/connection_label.h:
> 
> > +
> > +#include "nest.h"
> > +#include "nest_names.h"
> > +#include "dictdatum.h"
> > +#include "dictutils.h"
> > +
> > +namespace nest
> > +{
> > +class ConnectorModel;
> > +
> > +/**
> > + * Connections are unlabeled by default. Unlabeled connections cannot be specified
> > + * as a search criterion in the `GetConnections` function.
> > + * @see ConnectionLabel
> > + */
> > +const long_t UNLABELED_CONNECTION = -1;
> 
> This would require a template argument, whenever you want to access this field, like:
> 
> ConnectionLabel< StaticConnection< TargetIdentifierPtrRport > >::UNLABELED_CONNECTION
> 
> I guess we should not go for this...
> 
> —
> Reply to this email directly or view it on GitHub.
> 


--

Dr. Hans Ekkehard Plesser, Associate Professor
Section Head

Dept. of Mathematical Sciences and Technology
Norwegian University of Life Sciences
PO Box 5003, 1432 Aas, Norway

Phone +47 6723 1560
Email hans.ekkehard.plesser@nmbu.no
Home  http://arken.umb.no/~plesser

heplesser:
@tammoippen Nice work!

heplesser:
With this PR merged, #121 is fixed on the NEST side.

PR #160, opened 2015-11-18
Fix of issue #77 for pp_psc_delta

mdeger:
pp_psc_delta now calls set_spiketime on every spike that it emits. This is necessary for STDP synapses that connect to pp_psc_delta neurons to be plastic. 
A python script to compare the change of a STDP synapse on pp_psc_delta in comparison to iaf_psc_delta is also attached to this pull request (testsuite/manualtests/test_pp_psc_delta_stdp.py).
Many thanks to @heplesser for solving the problem (see issue #77).


heplesser:
@mdeger Thank you for your pull request and excuse me for not fixing this earlier. I have some more changes to add for a complete solution, which I already have in a branch of mine. I will now integrate your and my changes in my repository and then send you a pull request towards your branch. Once you merge that, this pull request will be updated automatically.

heplesser:
Suggested reviewers: @suku248 @abigailm

precise/iaf_psc_alpha_canon.cpp
```diff
@@ -169,8 +169,6 @@ nest::iaf_psc_alpha_canon::State_::get( DictionaryDatum& d, const Parameters_& p
   def< double >( d, names::V_m, y3_ + p.E_L_ ); // Membrane potential
   def< double >( d, "y1", y1_ );                // y1 state
   def< double >( d, "y2", y2_ );                // y2 state
```
mhelias: should define entries in nest_names.h for y1 and y2

mhelias:
apart from comment above looks good to me and can be merged, once fixed

precise/iaf_psc_alpha_canon.cpp
heplesser: @mhelias Thanks for catching this. But I am wondering if we should not rather drop these two undocumented recordables. They are not recorded by any test or example, and the other models do not offer them or their equivalent (except for `iaf_psc_alpha`, under a proper and long name).
I would suggest to remove them entirely. 

@jougs Do you think that could lead to any trouble with unhappy users?

jougs: I agree that those lines should be removed or get a better name. I guess they are just left over from a debugging session to make them easily readable. No need to worry about unhappyness from my side.

models/pp_pop_psc_delta.h
```diff
@@ -387,12 +386,6 @@ pp_pop_psc_delta::set_status( const DictionaryDatum& d )
   State_ stmp = S_;      // temporary copy in case of errors
   stmp.set( d, ptmp );   // throws if BadProperty
 
-  // We now know that (ptmp, stmp) are consistent. We do not
-  // write them back to (P_, S_) before we are also sure that
-  // the properties to be set in the parent class are internally
-  // consistent.
-  Archiving_Node::set_status( d );
```
krishnanj: archiving node- get and set_status calls seems to be removed only from this model

precise/iaf_psc_exp_ps.cpp
```diff
@@ -458,6 +452,7 @@ nest::iaf_psc_exp_ps::emit_spike_( const Time& origin,
   S_.is_refractory_ = true;
 
   // send spike
+  set_spiketime( Time::step( S_.last_spike_step_ ), S_.last_spike_offset_ );
   SpikeEvent se;
 
   se.set_offset( spike_offset );
```
krishnanj: se.set_offset(S_.last_spike_offset_);

precise/iaf_psc_alpha_canon.cpp
```diff
@@ -484,6 +478,7 @@ nest::iaf_psc_alpha_canon::emit_spike_( Time const& origin,
   S_.is_refractory_ = true;
 
   // send spike
+  set_spiketime( Time::step( S_.last_spike_step_ ), S_.last_spike_offset_ );
   SpikeEvent se;
   se.set_offset( spike_offset );
```
krishnanj: se.set_offset(S_.last_spike_offset_);

precise/iaf_psc_exp_ps.cpp
heplesser: I have fixed this via #167.

precise/iaf_psc_alpha_canon.cpp
heplesser: I have fixed this via #167.

models/pp_pop_psc_delta.h
heplesser: @krishnanj This is actually intentional. The `pp_pop_psc_delta` model represents an entire population, not a single neuron, so STDP would not be well defined for it. Since spike archiving is only required for STDP, this model does not archive. See also the [comment by @mdeger](https://github.com/nest/nest-simulator/issues/77#issuecomment-159389434).

heplesser:
This PR should be merged before #167.

krishnanj:
:+1:

precise/iaf_psc_alpha_canon.cpp
abigailm: Agree to throw those out.

abigailm:
:1 once the y_ and y_2 are gone.

heplesser:
I will merge now and then immediately create and merge a new PR that removes `y_1` and `y_2`.

heplesser:
I have just committed PR #194 that will remove `y_1` and `y_2`.

Issue #156, opened 2015-11-17
PyNEST+MUSIC+OpenMP fails to work


JanneM:
PyNEST and MUSIC (1.1.15) fails on the receiving side when OpenMP is enabled and two models are connected using MUSIC. The sending side works. SLI models are unaffected. It happens when explicitly setting either local_num_threads or _total_num_virtual_procs on the receiving side. This is reproducible on both architectures I have access to (Linux/AMD64 and K). 

The error message complains that the port "event_in" does not exist, but "event_in" is not used as a port name anywhere:

Traceback (most recent call last):
  File "./receivenest.py", line 13, in <module>
    params = {'port_name': 'music_in'})
  File "/home/janne/bin/n28/lib/python2.7/site-packages/nest/lib/hl_api_helper.py", line 138, in stack_checker_func
    return f(*args, **kwargs)
  File "/home/janne/bin/n28/lib/python2.7/site-packages/nest/lib/hl_api_nodes.py", line 44, in Create
    sr(cmd)
  File "/home/janne/bin/n28/lib/python2.7/site-packages/nest/__init__.py", line 79, in catching_sli_run
    raise _kernel.NESTError("{0} in {1}{2}".format(errorname, commandname, message))
pynestkernel.NESTError: MUSICPortUnknown in SetDefaults_l_D: The port event_in does not exist.

I'm attaching a simple test case (remove the ".txt" endings - is there a better way to add things like this?):

[receivenest.py.txt](https://github.com/nest/nest-simulator/files/36323/receivenest.py.txt)
[sendnest.py.txt](https://github.com/nest/nest-simulator/files/36325/sendnest.py.txt)
[ompfault.music.txt](https://github.com/nest/nest-simulator/files/36324/ompfault.music.txt)

Run it with:

$ mpirun -np 2 music ompfault.music

If you comment out the local_num_threads statement in receivenest, the test works. Keep it in, or uncomment total_num_virtual_procs, and it fails.

mdjurfeldt:
The reason for this is that setting total_num_virtual_threads causes Network::reset () to be called which causes clearing of the Network::music_in_portlist_. This also clears the port name which the music_event_in_proxy had registered (event_in) and music_event_in_proxy is never given a new chance to register it.

I guess there are two possible solutions:

1. Introduce a new virtual method in the API for Node which tells a model to setup its pristine state. Network::init_ should then call it when cloning pristine models.

[...]

PR #165, opened 2015-11-20
Fix issue #156: At Network reset, initialize music_in_portlist_ from …

mdjurfeldt:
…a stored pristine copy instead of clearing it

### DIFF
--- a/models/music_event_in_proxy.cpp
+++ b/models/music_event_in_proxy.cpp
@@ -104,7 +104,7 @@ nest::music_event_in_proxy::music_event_in_proxy( const music_event_in_proxy& n 
   , P_( n.P_ )
   , S_( n.S_ )
 {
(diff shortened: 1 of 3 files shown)


jougs:
This change looks good to me, so :+1:. Maybe @tammoippen can have a second look? Do we need a test to prevent this problem from happening again?

apeyser:
Maybe the comment for register_music_in_port could be a bit more expansive? If I were looking at this for the first time, I'd be confused about what 'pristine' is, and why you would want to set it this way.  So something about this being the default reset under certain conditions.

mdjurfeldt:
Better now? :)

jougs:
:+1: once my comments have been addressed.

mdjurfeldt:
Done!

apeyser:
Solves my request re comments

PR #168, opened 2015-11-26
Added check for OpenMP to protect another OMP call

JanneM:
One more unprotected OMP call snuck in with a recent commit. Added a check to see that we are building with OpenMP.


jougs:
Thanks for spotting this. I'm immediately merging, as the code change is minor, but important. This fixes #169.

PR #170, opened 2015-11-26
`NumericalInstability` exception incorrectly refered to GSL.

heplesser:
Just a small fix. We have a dedicated `GSLSolverFailure` exception for cases in which GSL ODE solvers run into trouble. `NumericalInstability` is for cases which do not use GSL.


jougs:
Looks good to me. :+1:

tammoippen:
:+1:  and merging.

PR #57, opened 2015-07-01
Remove last bits of model and connection unloading, fixing trac.617.

heplesser:
After #55 handled module unloading, this PR removes the last bits of the never functional model and connection unregistration code. It thus fixes trac.617, which should be closed once this PR is merged.

Potential reviewers: @abigailm @jougs

PR #24, opened 2015-06-03
Removed CPEX from Communicator.

heplesser:
This PR removes the CPEX algorithm from communicator, partially solving old ticket no 364.

PR #171, opened 2015-11-27
Third pull request for structural plasticity: connectivity manager and pynest functions

sdiazpier:
This third pull request on structural plasticity integrates the final functionality for the dynamic creation and deletion of synapses during simulation. It adds a new structural plasticity connectivity manager which takes care of updating the network by creating or deleting synapses until a certain level of activity is reached. PyNEST functions and tests are also included.
With this changes, the user now can also delete synapses manually, in a similar manner as with the connect functions.


heplesser:
Suggested reviewers: @heplesser @jschuecker @ddahmen

nestkernel/conn_builder.cpp
```diff
@@ -85,11 +85,13 @@ nest::ConnBuilder::ConnBuilder( Network& net,
   // be possible to set the delay without the weight.
   default_weight_ = !syn_spec->known( names::weight );
 
+  default_delay_ = !syn_spec->known( names::delay );
```
heplesser: Could this (and `default_weight_`) be `const`?

nestkernel/conn_builder.cpp
```diff
@@ -114,6 +116,19 @@ nest::ConnBuilder::ConnBuilder( Network& net,
       : ConnParameter::create( ( *syn_defaults )[ names::delay ], net_.get_num_threads() );
   }
   register_parameters_requiring_skipping_( *delay_ );
+  // Structural plasticity parameters
+  // Check if both pre and post synaptic element are provided
+  if ( syn_spec->known( names::pre_synaptic_element )
+    && syn_spec->known( names::post_synaptic_element ) )
```
heplesser: Should we raise an error if only one if pre and post is given? The user probably intended to do something and just forgot to provide complete information. Silently doing nothing in that case could lead to hard-to-detect errors or even incorrect results.

heplesser:
@sdiazpier I have started to review your code, and will complete my review soon. Right now, I am mainly wondering whether there was an important reason to re-introduce CPEX-communication. We had removed it from NEST some time ago because it is slower than Allgather. Unless there structural plasticity comes with some special communications requirements where Allgather is not suitable, we should avoid to reintroduce CPEX to keep code complexity down.

nestkernel/communicator.cpp
```diff
@@ -146,7 +149,9 @@ nest::Communicator::init( int* argc, char** argv[] )
   // generate and commit struct
   MPI_Type_struct( 2, blockcounts, offsets, source_types, &MPI_OFFGRID_SPIKE );
   MPI_Type_commit( &MPI_OFFGRID_SPIKE );
-
+  // set up order of communication if using CPEX
+  if ( !use_Allgather_ )
+    init_communication();
   initialized_ = true;
```
tammoippen: The CPEX communication was removed in #24. This seems like a relict of old times coming back. There are many parts in the Communicator class that are affected by this. One should make a diff of the Communicator before CPEX removal and this version and see, what actually needs changing. As far as I can see, only these functions need to be added:

    static void communicate( std::vector< ulong_t >& send_buffer,
        std::vector< ulong_t >& recv_buffer,
        std::vector< int >& displacements );
    static void communicate( std::vector< int_t >& send_buffer,
        std::vector< int_t >& recv_buffer,
        std::vector< int >& displacements );
and

    /*
     * Sum across all rank
     */
    static void communicate_Allreduce_sum_in_place( double_t buffer );
    static void communicate_Allreduce_sum_in_place( std::vector< double_t >& buffer );
    static void communicate_Allreduce_sum_in_place( std::vector< int_t >& buffer );
    static void communicate_Allreduce_sum( std::vector< double_t >& send_buffer,
      std::vector< double_t >& recv_buffer );

nestkernel/conn_builder.cpp
```diff
@@ -423,6 +545,186 @@ nest::OneToOneBuilder::connect_()
...
+            ++tgid, ++sgid )
+      {
+        assert( sgid != sources_.end() );
+
+        if ( *sgid == *tgid and not autapses_ )
+          continue;
+
+        Node* const target = net_.get_node( *tgid );
+        const thread target_thread = target->get_thread();
+        if ( !change_connected_synaptic_elements( *sgid, *tgid, tid, 1 ) )
```
tammoippen: I would suggest, that you get the `Node*` and `target_thread` after this `if`, because only then you know for sure, that the target is local. Further, I would suggest to rename the function (or split the functionality in half), so that it becomes clear, that we are also checking, if `*tgid` is a local node.

The same goes for `AllToAllBuilder::sp_connect_()`.

nestkernel/conn_builder.cpp
```diff
@@ -850,3 +1318,166 @@ nest::BernoulliBuilder::connect_()
...
+        // update the number of connected synaptic elements
+        source->connect_synaptic_element( pre_synaptic_element_name, 1 );
+      }
+
+      // check whether the target is on this mpi machine
+      if ( !net_.is_local_gid( *tgid ) )
+        continue;
+
+      Node* const target = net_.get_node( *tgid );
+      const thread target_thread = target->get_thread();
```
tammoippen: Are you missing a check, wether the `target` is on this thread?

nestkernel/conn_builder.cpp
```diff
@@ -850,3 +1318,166 @@ nest::BernoulliBuilder::connect_()
...
+  for ( thread thr = 0; thr < net_.get_num_threads(); ++thr )
+    if ( exceptions_raised_.at( thr ).valid() )
+      throw WrappedThreadException( *( exceptions_raised_.at( thr ) ) );
+}
+
+void
+nest::SPBuilder::connect_()
+{
+  connect_( sources_, targets_ );
+}
```
tammoippen: I cannot see any place, where this `protected` method will be called. Best would be to to indicate this by throwing a `NotImplemented` exception, or such (maybe even in the header).

nestkernel/conn_builder.cpp
```diff
@@ -850,3 +1318,166 @@ nest::BernoulliBuilder::connect_()
...
+  // Code copied and adapted from OneToOneBuilder::connect_()
+  // make sure that target and source population have the same size
+  if ( sources.size() != targets.size() )
+  {
+    net_.message( SLIInterpreter::M_ERROR,
+      "Connect",
+      "Source and Target population must be of the same size." );
+    throw DimensionMismatch();
+  }
+
```
tammoippen: The `#pragma omp parallel` is missing. This will be executed serially.

nestkernel/connection_manager.cpp
```diff
@@ -542,8 +543,11 @@ ConnectionManager::validate_source_entry( thread tid, index s_gid, synindex syn_
   // check, if entry exists
   // if not put in zero pointer
   if ( connections_[ tid ].test( s_gid ) )
-    return connections_[ tid ].get(
-      s_gid ); // returns non-const reference to stored type, here ConnectorBase*
+  {
+    return connections_[ tid ].get( s_gid ); // validate_pointer(connections_[ tid ].get(s_gid ));
+                                             // // returns non-const reference to stored type, here
+                                             // ConnectorBase*
+  }
```
tammoippen: Can you cleanup the comment, please?

nestkernel/network.cpp
```diff
@@ -1954,6 +2088,43 @@ Network::register_model( Model& m, bool private_model )
   return id;
 }
 
+void
+Network::unregister_model( index m_id )
```
tammoippen: Also relict of old times. `unregister_model` and such were removed with #57.

nestkernel/scheduler.cpp
```diff
@@ -567,6 +569,23 @@ nest::Scheduler::update()
...
+      if ( net_->structural_plasticity_enabled_
+        && ( clock_.get_steps() + from_step_ ) % SPManager::structural_plasticity_update_interval
+          == 0 )
+      {
+        for ( i = nodes_vec_[ t ].begin(); i != nodes_vec_[ t ].end(); ++i )
+          ( *i )->update_synaptic_elements(
+            Time( Time::step( clock_.get_steps() + from_step_ ) ).get_ms() );
+#pragma omp barrier
+#pragma omp single
+        {
```
tammoippen: Why do we need the `omp barrier` and `single`? The method `Scheduler::init_()` is private and only called in the constructor and in `reset()`, which are also not called in an `omp parallel` region.

nestkernel/conn_builder.cpp
naveau: I agree, I was more focused on not disturbing people who doesn't use structural plasticity but we can assume that someone specifing only one synatic_element is doing something wrong.

nestkernel/conn_builder.cpp
naveau: As far as I remember these values are defined and needed outside of the structural plasticity framework. We might discuss this point on another thread.

nestkernel/communicator.cpp
naveau: @tammoippen, @heplesser : it was not intended to reintroduce CPEX communication. I'm sorry that I didn't check this. In fact, these changes where made when CPEX communication was still implemented in the communicator. Sorry for the noise.

nestkernel/conn_builder.cpp
heplesser: @naveau My mistake, the `default_*` are members because they are also needed in `single_connect_()`; so they cannot be const.

nestkernel/communicator.cpp
sdiazpier: Hi @heplesser @tammoippen and @naveau, sorry it was my mistake these outdated methods got reinserted after rebasing.

nestkernel/conn_builder.cpp
sdiazpier: Hi @tammoippen, I agree a better name is needed, Working on that.

nestkernel/conn_builder.cpp
sdiazpier: Yes, you are right @tammoippen

heplesser:
@sdiazpier It appears that the tests in `test_sp` are not run when testing with nosetest, which is the standard test setup for NEST. Running `nosetests --verbose` inside `test_sp` in the install directory reports 0 tests. I suspect this is because the python file names do not begin with `test_`.

sdiazpier:
Hi dear @heplesser, @tammoippen and @naveau   thanks for all your comments and review. I have made most of the corrections you suggested. I also fixed the test names and removed all outdated methods. Please let me know what would be missing.

heplesser:
@sdiazpier Great work! The code looks very fine to me now. Just some polishing of the example, so that it will help users to get started with structural plasticity in NEST, and we are ready to merge from my side.

heplesser:
I am sorry I commented in the commits instead of in the "Files changed" tab above, so my comments do not show inline with the code until you click the links in the comments above. Still learning Github ;).

examples/nest/structural_plasticity.py
```diff
@@ -0,0 +1,234 @@
...
+            'tau_syn_in': 0.5,          # inhibitory synaptic time constant (ms)
+            't_ref': 2.0,               # absolute refractory period (ms)
+            'E_L': -65.0,               # resting membrane potential (mV)
+            'V_th': -50.0,              # spike threshold (mV)
+            'C_m': 250.0,               # membrane capacitance (pF)
+            'V_reset': -65.0,           # reset potential (mV)
+        } 
+        
+        self.nodes = ()
+        self.nodes_i = ()
```
heplesser: For symmetry, I would call the excitatory nodes `self.nodes_e`. And instead of initializing with an empty tuple, I would use `None`.

sdiazpier:
@heplesser thank you very much for your new comments. I will work right away on the modifications for the example.

sdiazpier:
Dear @heplesser, I have made the corrections you suggested. Please let me know if anything else is still needed.

heplesser:
@sdiazpier The Travis tests failed because you forgot to update the `Makefile.am` in `examples` and `pynest/examples`. With that in place, all works nicely. But both here on Github and in my IDE, the example code looks as if the indentation is all wrong in many places. At the same time, the Python code runs. Have you maybe mixed spaces and tabs for indentation? Indentation should be by spaces only.

sdiazpier:
Dear @heplesser, @naveau and @tammoippen 
In the synaptic plasticity tests, all synapse models except for those with dopamine and the gap junction one are tested. I explicitly skip their name from the list of models. For some strange reason, Travis is generating clone synapse models with the name: "model_lbl", for example "gap_junction_lbl". 
All tests were running fine until commit 0d907ea. After this commit I made some formatting changes to the example and moved it from /examples/ to /pynest/examples/ and the tests started failing because these cloned synapse models were being tested with structural plasticity. 
On my computer this does not happen.
Anyone has any idea of why this could be happening?

heplesser:
@sdiazpier The labeled synapses entered NEST with #158. They are automatically added for all existing synapse types, so that PyNN can keep track of projections. It is fine to exclude them in the way you have done.

:+1: from me!

sdiazpier:
@heplesser oh sorry I missed that. Thank you for clarifying it to me, I thought something very strange was going on.

nestkernel/conn_builder.cpp
```diff
@@ -423,6 +551,185 @@ nest::OneToOneBuilder::connect_()
...
+#pragma omp parallel
+  {
+    // get thread id
+    const int tid = net_.get_thread_id();
+
+    try
+    {
+      // allocate pointer to thread specific random generator
+      librandom::RngPtr rng = net_.get_rng( tid );
```
tammoippen: I do not think that we need a RNG during disconnect.

nestkernel/conn_builder.cpp
```diff
@@ -423,6 +551,185 @@ nest::OneToOneBuilder::connect_()
...
+
+      for ( GIDCollection::const_iterator tgid = targets_.begin(), sgid = sources_.begin();
+            tgid != targets_.end();
+            ++tgid, ++sgid )
+      {
+
+        assert( sgid != sources_.end() );
+
+        if ( *sgid == *tgid and not autapses_ )
+          continue;
```
tammoippen: Does `autapses_` (and `multapses_`) make sense during disconnect? I think not and would remove the skipping.

nestkernel/conn_builder.cpp
```diff
@@ -423,6 +551,185 @@ nest::OneToOneBuilder::connect_()
...
+
+#pragma omp parallel
+  {
+    // get thread id
+    const int tid = net_.get_thread_id();
+
+    try
+    {
+      // allocate pointer to thread specific random generator
+      librandom::RngPtr rng = net_.get_rng( tid );
```
tammoippen: Same as above.

nestkernel/conn_builder.cpp
```diff
@@ -423,6 +551,185 @@ nest::OneToOneBuilder::connect_()
...
+
+      for ( GIDCollection::const_iterator tgid = targets_.begin(), sgid = sources_.begin();
+            tgid != targets_.end();
+            ++tgid, ++sgid )
+      {
+        assert( sgid != sources_.end() );
+
+        if ( *sgid == *tgid and not autapses_ )
+          continue;
```
tammoippen: Same as above.

nestkernel/conn_builder.cpp
```diff
@@ -483,6 +790,172 @@ nest::AllToAllBuilder::connect_()
...
+{
+#pragma omp parallel
+  {
+    // get thread id
+    const int tid = net_.get_thread_id();
+
+    try
+    {
+      // allocate pointer to thread specific random generator
+      librandom::RngPtr rng = net_.get_rng( tid );
```
tammoippen: Same as above.

nestkernel/conn_builder.cpp
```diff
@@ -483,6 +790,172 @@ nest::AllToAllBuilder::connect_()
...
+          for ( GIDCollection::const_iterator sgid = sources_.begin(); sgid != sources_.end();
+                ++sgid )
+            skip_conn_parameter_( tid );
+          continue;
+        }
+
+        for ( GIDCollection::const_iterator sgid = sources_.begin(); sgid != sources_.end();
+              ++sgid )
+        {
+          if ( not autapses_ and *sgid == *tgid )
```
tammoippen: Same as above.

sdiazpier:
Hi dear @tammoippen, thanks for your comments. I have made the changes you suggested. Please let me know if anything else would be missing.

tammoippen:
Very nice @sdiazpier and @naveau ! :+1: from my side.

jougs:
Please note that the contained example is currently not running in our standard VM image. We see this as a blocker for the merge and the 2.10 release, but @sdiazpier is working on it.

sdiazpier:
Dear all, the corrections required for the example to run correctly also on the standard VM image have been verified and committed. Please let me know if anything else is required before merging.

heplesser:
@sdiazpier Very fine! I just have one little cosmetic point. In `communicator.cpp`, we now have some lines reading
```
displacements.resize( num_processes_, 0 );
```
and some reading
```
displacements.resize( num_processes_ );
```
Both do exactly the same, because if no value is given, new elements are initialized with the default constructor for the element type, which for built-in datatypes gives the value 0 for that type (Stroustrup §6.2.8). But this is confusing to the reader, who will wonder what the difference between the two lines is and why it is there. So we should choose one, and I would suggest the latter.

Also, to reduce confusion, I think I would use the same line also in the no-mpi-case, instead of
```
displacements.resize( 1 );
```
`num_processes_` is 1 if NEST was compiled without MPI, so the result will be the same, but the code reader won't have to wonder "why is it different here"?

sdiazpier:
Hi dear @heplesser. Thank you for your comments, I have committed the changes you suggested. Please let me know if anything else is missing.

heplesser:
@sdiazpier Thanks! I think the only thing left now is a little check in the structural plasticity example telling the user that the example only works for a single process; alternatively, adapt the script to work on multiple processes. But then one collects connection numbers and Ca levels only for neurons local to each process, so one would have to write them to file for each process and then combine them. That is too complicated for an introductory example, so I think I would leave it as it is and just make sure it is not run if `nest.num_processes() > 1`.

sdiazpier:
Dear @heplesser, I added the check and committed the changes.

sdiazpier:
Right @heplesser, sorry I missed that. Hope it is all good now.

heplesser:
:+1: !!!

jougs:
Many thanks for the nice work, the review and quick responses. It was a pleasure to watch.

Issue #172, opened 2015-12-01
`CopyModel` on `gap_junction` does not work


tammoippen:
See https://github.com/nest/nest-simulator/pull/158#issuecomment-160584748

When registering the `gap_junction` model, the `static` variable `GapJEvent::synid_` is set to the synapse id of the `gap_junction` model.  Late, the `Scheduler` uses this to distinguish, which kind of SecondaryEvent is send.
The problem is, that one cannot copy the model: if `GapJEvent::synid_` is set to the new model, the old model will not work anymore, otherwise the new model does not work.

PR #XXX provides a hot-fix solution by prohibiting the `CopyModel` on `gap_junction`, but in the long-run we should provide a working solution of `CopyModel` on `gap_junction`.

tammoippen:
fixed by #175

PR #173, opened 2015-12-01
Hot-fix for gap junction framework.

tammoippen:
In #158 we found, that copying `gap_junction` connections is not functional yet. This PR adds a warning, when trying to `CopyModel` the `gap_junction` model, and prohibits the copy.
See #172 for ongoing work for this issue.

### DIFF
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -597,7 +597,7 @@ NestModule::ResumeSimulationFunction::execute( SLIInterpreter* i ) const
    A copy of model is created and registered in modeldict or synapsedict
    under the name new_model. If a parameter dictionary is given, the parameters
    are set in new_model.
-   
+
    Note: Copying the `gap_junction` model (or any other model using SecondaryEvents)
          is not possible at this time. Please see issue #172.
  */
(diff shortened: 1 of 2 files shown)


nestkernel/nestmodule.cpp
```diff
@@ -609,6 +609,17 @@ NestModule::CopyModel_l_l_DFunction::execute( SLIInterpreter* i ) const
   const std::string newmodname = getValue< std::string >( i->OStack.pick( 1 ) );
   DictionaryDatum dict = getValue< DictionaryDatum >( i->OStack.pick( 0 ) );
 
+  if ( oldmodname.toString() == "gap_junction" )
+  {
+    get_network().message( SLIInterpreter::M_WARNING,
```
heplesser: I think this should be an error, i.e., not a warning, and an exception should be raised. The user is asking NEST to do something that is not possible, therefore it is an error.

nestkernel/nestmodule.cpp
```diff
@@ -609,6 +609,17 @@ NestModule::CopyModel_l_l_DFunction::execute( SLIInterpreter* i ) const
   const std::string newmodname = getValue< std::string >( i->OStack.pick( 1 ) );
   DictionaryDatum dict = getValue< DictionaryDatum >( i->OStack.pick( 0 ) );
 
+  if ( oldmodname.toString() == "gap_junction" )
```
heplesser: This concerns the documentation above: The warning about module unloading should go, since we have removed module unloading (it never worked). Also, we should add a note saying that gap junctions cannot be copied.

heplesser: Could you also add a comment pointing out that this is a hotfix, with reference to the relevant issues?

janhahne:
The hotfix might not be needed anymore. See PR #175.

tammoippen:
closing due to #175 fixing #172 properly.

PR #174, opened 2015-12-01
Allow newer versions of cppcheck.

tammoippen:
With TravisCI we perform the `cppcheck` with v1.69, since a previous version had a bug (see #79). With the script from #95, we can check the code style locally and there is no reason, why we should not use a newer `cppcheck`. With this PR we allow newer version of `cppcheck` in that script.

### DIFF
--- a/extras/check_code_style.sh
+++ b/extras/check_code_style.sh
@@ -131,10 +131,9 @@ cppcheck_version=`$CPPCHECK --version | sed 's/^Cppcheck //'`
 echo "cppcheck version: $cppcheck_version"
 IFS=\. read -a cppcheck_version_a <<< "$cppcheck_version"
 if [[ ${cppcheck_version_a[0]} -lt 1 ]]; then
-  bail_out "Require cppcheck version 1.69. Not $cppcheck_version ."
-fi
-if [[ ${cppcheck_version_a[1]} -lt 69 ]]; then
-  bail_out "Require cppcheck version 1.69. Not $cppcheck_version ."
+  bail_out "Require cppcheck version 1.69 or later. Not $cppcheck_version ."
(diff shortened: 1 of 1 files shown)


extras/check_code_style.sh
```diff
@@ -129,7 +129,11 @@ echo "vera++ version: `vera++ --version`"
 $CPPCHECK --enable=all --inconclusive --std=c++03 ./nest/main.cpp >/dev/null 2>&1 || usage 1 "Executable $CPPCHECK for cppcheck is not working!"
 cppcheck_version=`$CPPCHECK --version | sed 's/^Cppcheck //'`
 echo "cppcheck version: $cppcheck_version"
-if [[ "x$cppcheck_version" != "x1.69" ]]; then
+IFS=\. read -a cppcheck_version_a <<< "$cppcheck_version"
+if [[ ${cppcheck_version_a[0]} -lt 1 ]]; then
+  bail_out "Require cppcheck version 1.69. Not $cppcheck_version ."
+fi
+if [[ ${cppcheck_version_a[1]} -lt 69 ]]; then
```
heplesser: This works for me with 1.70 installed. But I think in both cases the bail-out message should be "... 1.69 or later. Not ...".

apeyser: You intend this to test for 2 > $cppcheck_version >= 1.69 && || $cppcheck_version >= 2.69?

tammoippen:
Thank you for the comments. In b2785a2 i adjust the version checking and bail out message.

apeyser:
:thumbsup:

apeyser:
@heplesser: Any more comments so we can clean up here?

heplesser:
@apeyser  :+1:

PR #175, opened 2015-12-01
Fix for issue #172: Added CopyModel compatibility for gap_junction model

janhahne:
The PR fixes the problem reported in issue #172.

With this PR it is possible to
- register more than one SecondaryConnectorModel with GapJEvents (e.g. to add gap_junction_lbl, as needed for PR #158)
- use CopyModel on the gap_junction connection model

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -234,7 +234,14 @@ ConnectionManager::copy_synapse_prototype( synindex old_id, std::string new_name
     throw KernelException( "Synapse model count exceeded" );
   }
   assert( new_id != invalid_synindex );
-
+  
+  // if the copied synapse is a secondary connector model the synid of the copy has to
+  // be mapped to the corresponding secondary event type
+  if ( get_synapse_prototype( old_id ).is_primary() == false )
+  {
+    ( get_synapse_prototype( old_id ).get_event() )->add_syn_id( new_id );
+  }
+  
   for ( thread t = 0; t < net_.get_num_threads(); ++t )
   {
     prototypes_[ t ].push_back( get_synapse_prototype( old_id ).clone( new_name ) );
@@ -721,7 +728,7 @@ ConnectionManager::send_secondary( thread t, SecondaryEvent& e )
 
         if ( p->homogeneous_model() )
         {
-          if ( p->get_syn_id() == e.get_syn_id() )
+          if ( e.has_syn_id ( p->get_syn_id() ) )
             p->send( e, t, prototypes_[ t ] );
         }
         else
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -237,7 +237,7 @@ ConnectionManager::copy_synapse_prototype( synindex old_id, std::string new_name
 
   // if the copied synapse is a secondary connector model the synid of the copy has to
   // be mapped to the corresponding secondary event type
-  if ( get_synapse_prototype( old_id ).is_primary() == false )
+  if ( not get_synapse_prototype( old_id ).is_primary() )
   {
     ( get_synapse_prototype( old_id ).get_event() )->add_syn_id( new_id );
   }
@@ -728,7 +728,7 @@ ConnectionManager::send_secondary( thread t, SecondaryEvent& e )
 
         if ( p->homogeneous_model() )
         {
-          if ( e.has_syn_id( p->get_syn_id() ) )
+          if ( e.supports_syn_id( p->get_syn_id() ) )
             p->send( e, t, prototypes_[ t ] );
         }
         else
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -234,14 +234,14 @@ ConnectionManager::copy_synapse_prototype( synindex old_id, std::string new_name
     throw KernelException( "Synapse model count exceeded" );
   }
   assert( new_id != invalid_synindex );
-  
+
   // if the copied synapse is a secondary connector model the synid of the copy has to
   // be mapped to the corresponding secondary event type
   if ( get_synapse_prototype( old_id ).is_primary() == false )
   {
     ( get_synapse_prototype( old_id ).get_event() )->add_syn_id( new_id );
   }
-  
+
   for ( thread t = 0; t < net_.get_num_threads(); ++t )
   {
     prototypes_[ t ].push_back( get_synapse_prototype( old_id ).clone( new_name ) );
@@ -728,7 +728,7 @@ ConnectionManager::send_secondary( thread t, SecondaryEvent& e )
 
         if ( p->homogeneous_model() )
         {
-          if ( e.has_syn_id ( p->get_syn_id() ) )
+          if ( e.has_syn_id( p->get_syn_id() ) )
             p->send( e, t, prototypes_[ t ] );
         }
         else
--- a/nestkernel/event.cpp
+++ b/nestkernel/event.cpp
@@ -121,7 +121,6 @@ void GapJEvent::operator()()
   receiver_->handle( *this );
 }
 
-synindex GapJEvent::synid_ = invalid_synindex;
 std::vector< synindex > GapJEvent::supported_syn_ids_;
 size_t GapJEvent::coeff_length_ = 0;
 }
--- a/nestkernel/event.cpp
+++ b/nestkernel/event.cpp
@@ -122,5 +122,6 @@ void GapJEvent::operator()()
 }
 
 synindex GapJEvent::synid_ = invalid_synindex;
+std::vector< synindex > GapJEvent::synid_vector_;
 size_t GapJEvent::coeff_length_ = 0;
 }
--- a/nestkernel/event.cpp
+++ b/nestkernel/event.cpp
@@ -122,6 +122,6 @@ void GapJEvent::operator()()
 }
 
 synindex GapJEvent::synid_ = invalid_synindex;
-std::vector< synindex > GapJEvent::synid_vector_;
+std::vector< synindex > GapJEvent::supported_syn_ids_;
 size_t GapJEvent::coeff_length_ = 0;
 }
--- a/nestkernel/event.h
+++ b/nestkernel/event.h
@@ -736,6 +736,10 @@ public:
 
   virtual synindex get_syn_id() const = 0;
 
+  virtual void add_syn_id( const synindex synid ) = 0;
+
+  virtual bool has_syn_id( const synindex synid ) const = 0;
+
   //! size of event in units of uint_t
   virtual size_t size() = 0;
   virtual fwit& operator<<( fwit& pos ) = 0;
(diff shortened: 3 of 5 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -235,6 +235,13 @@ ConnectionManager::copy_synapse_prototype( synindex old_id, std::string new_name
   }
   assert( new_id != invalid_synindex );
 
+  // if the copied synapse is a secondary connector model the synid of the copy has to
+  // be mapped to the corresponding secondary event type
+  if ( get_synapse_prototype( old_id ).is_primary() == false )
```
heplesser: This should be 
```
if ( not get_synapse_prototype( old_id ).is_primary() )
```
In general, one should not compare with `true` and `false`.

nestkernel/event.h
```diff
@@ -810,7 +814,19 @@ class CoeffArrayIterator
 class GapJEvent : public SecondaryEvent
 {
 private:
+  /*
+  Conceptual there is an one-to-one mapping between a SecondaryEvent
```
heplesser: "Conceptually, there is a ..."

nestkernel/event.h
```diff
@@ -810,7 +814,19 @@ class CoeffArrayIterator
...
+  and a SecondaryConnectorModel. The synindex of this particular
+  SecondaryConnectorModel is stored in the static synid_ member variable
+  on model registration. There are however reasons (e.g. the usage of
+  CopyModel or the creation of the labeled synapse model duplicates for pyNN)
+  which make it necessary to register several SecondaryConnectorModels with
+  one SecondaryEvent. Therefore the synindices of all these models are
+  stored  in the synid_vector_. The has_syn_id()-function allows testing
+  if a particular synid is mapped with the SecondaryEvent in question.
+  */
   static synindex synid_;
```
heplesser: Do we still need this member, or is it entirely replaced by `synid_vector_`?

nestkernel/event.h
```diff
@@ -810,7 +814,19 @@ class CoeffArrayIterator
...
+  static std::vector< synindex > synid_vector_;
```
heplesser: The only purpose of `synid_vector_` is to check whether the `syn_id` of a connection is in this vector. In general, `std::set< synindex >` would implement this more efficiently. But as long as this data structure is small, linear search through a vector might still be faster, so here we would need to think carefully about the relevant use cases.

In any case, the variable name should not contain the data structure. How about `supported_syn_ids_`?

nestkernel/event.h
```diff
@@ -827,7 +843,31 @@ class GapJEvent : public SecondaryEvent
...
+    bool has_syn_id = false;
+    for ( unsigned int i = 0; i < synid_vector_.size(); i++ )
+    {
+      if ( synid_vector_[ i ] == synid )
+      {
+        has_syn_id = true;
+        break;
+      }
+    }
+    return has_syn_id;
```
heplesser: You can reduce this method to (assuming here the renaming suggested above):
```
{
  return supported_syn_ids_.find( synid ) != supported_syn_ids_.end();
}
```
This gives you the full speed of the STL implementation and will work for any container type.

nestkernel/event.h
```diff
@@ -736,6 +736,10 @@ class SecondaryEvent : public Event
 
   virtual synindex get_syn_id() const = 0;
 
+  virtual void add_syn_id( const synindex synid ) = 0;
+
+  virtual bool has_syn_id( const synindex synid ) const = 0;
```
heplesser: Maybe `supports_syn_id()` would be an even better name. "Has" suggests that the synapse ID would be unique, but the main point here is that the `GapJEvent` now supports multiple synapse IDs.

I also wonder if we now still need the `get_syn_id()` method. Since the `GapJEvent` now supports multiple IDs, that method cannot return a uniquely defined value (unless one somewhat arbitrarily decides that the first gap junction synapse model defined is special).

heplesser:
@janhahne Fine work! I added som comments above and have another question related to this: `GapJEvent` has input and output operators. What are they used for? Are they needed for communication, or are they for providing debugging output? And how do they need to be adapted to the fact that the event now can support a number of synapse IDs?

nestkernel/event.h
heplesser: I suppose one could always use `*(synid_vector_.begin())`, which will store the same value as `synid_`.

nestkernel/event.cpp
```diff
@@ -122,5 +122,6 @@ void GapJEvent::operator()()
 }
 
 synindex GapJEvent::synid_ = invalid_synindex;
```
heplesser: I suppose we won't need the `invalid_synindex` any more, we could just check if `synid_vector_` is empty.

janhahne:
@heplesser Thanks for the comments. I addressed them with the new commits. I also added comments on the input and output operators of `GapJEvent`. They are used to build and read from the buffers of the MPI-communication. On the receiving side it is used with the following code

```cpp
        synindex synid;
        output_stream( synid, readpos );

        if ( synid == invalid_synindex )
          break;
        --readpos;

        net_->connection_manager_.assert_valid_syn_id( synid );

        *( secondary_events_prototypes_[ t ][ synid ] ) << readpos;

        net_->connection_manager_.send_secondary(
          t, *( secondary_events_prototypes_[ t ][ synid ] ) );
```

where `secondary_events_prototypes` is of type `std::vector< std::vector< SecondaryEvent* > >`. As long as the sended `synid` is any of the supported synids this works fine. So we choose to send the first element of the static vector, which is always existent, in the input operator.

nestkernel/event.h
```diff
@@ -827,13 +843,20 @@ class GapJEvent : public SecondaryEvent
...
   }
 
-  synindex
-  get_syn_id() const
+  bool
+  supports_syn_id( const synindex synid ) const
   {
-    return synid_;
+    return std::find( supported_syn_ids_.begin(), supported_syn_ids_.end(), synid )
+      != supported_syn_ids_.end();
```
heplesser: I'd put the entire expression in parentheses to make clear that the two line form one expression.

nestkernel/event.h
```diff
@@ -827,13 +843,20 @@ class GapJEvent : public SecondaryEvent
...
   set_syn_id( const synindex synid )
   {
-    synid_ = synid;
+    supported_syn_ids_.push_back( synid );
+  }
+
+  void
+  add_syn_id( const synindex synid )
+  {
+    supported_syn_ids_.push_back( synid );
```
heplesser: Should we protect this against multiple insertion, e.g. by
```
if ( not supported_syn_id( synid ) )
   ...
```
or maybe, since this really should not happen, even `assert( not ... )`.

nestkernel/event.h
janhahne: Usually there is a one-to-one mapping between the secondary connector model and the SecondaryEvent type. The only reasonable exceptions are the duplication for the labeled version of the synapse and each call to `CopyModel`. So I would guess that the size of this vector should never exceed 5 or 6 elements - especially as `CopyModel` might be ineffective for large simulations (keyword: HomConnector/HetConnector).

Would do we conclude on this issue?

nestkernel/event.h
```diff
@@ -827,13 +843,20 @@ class GapJEvent : public SecondaryEvent
   static void
   set_syn_id( const synindex synid )
```
heplesser: Is this method still called from anywhere? I not, we should remove it. If yes, we should replace the call by a call to `add`syn_id()`.

nestkernel/event.h
```diff
@@ -890,9 +913,12 @@ output_stream( T& d, fwit& pos )
   pos += size_uint_t( d );
 }
 
+// The following operator is used to read the information
+// of the GapJEvent from the buffer in Scheduler::deliver_events_
+// The synid can be skipped here as it is stored in a static vector.
```
heplesser: Could you turn this into a Doxygen comment in the place where the operator is declared?

nestkernel/event.h
```diff
@@ -907,9 +933,13 @@ inline fwit& GapJEvent::operator<<( fwit& pos )
   return pos;
 }
 
+// The following operator is used to write the information
+// of the GapJEvent into the secondary_events_buffer_
+// All GapJEvents are identified by the synid of the
+// first element in supported_syn_ids_
```
heplesser: See `operator<<`

nestkernel/event.h
janhahne: This is used in `connector_model_impl.h`

```cpp
/**
 * Register a synape with default Connector and without any common properties.
 */
template < class ConnectionT >
synindex
register_secondary_connection_model( Network& net, const std::string& name, bool has_delay = true )
{
  ConnectorModel& cm =
    *( new GenericSecondaryConnectorModel< ConnectionT >( net, name, has_delay ) );

  synindex synid = net.register_secondary_synapse_prototype( &cm );

  ConnectionT::EventType::set_syn_id( synid );

  return synid;
}
```

At this point it is important, that this is a `static` method. The other method `add_syn_id()` on the other side is called from an object.

nestkernel/event.h
janhahne: The way it is constructed this should indeed never happen. I will add `assert( ... )`

nestkernel/event.h
heplesser: Static methods can also be called on objects, and since both `set_syn_id()` and `add_syn_id()` only manipulate the static `supported_syn_ids_` member, we can remove `set_syn_id` and make `add_syn_id()` static.

heplesser:
@janhahne I can't find you comment on the vector vs set issue any more, but I think vector will probably do it. You may want to add a comment to the declaration of `supported_syn_ids_` that we chose vector over set because we expect this always to be short.

nestkernel/event.h
janhahne: But here is the problem:

The `add_syn_id`-function is called from `*SecondaryEvent`:
`( get_synapse_prototype( old_id ).get_event() )->add_syn_id( new_id );`
so it has to be defined as `virtual` in the `GapJEvent` base class. And it is not possible for functions to be both virtual and static.

So I do not see any way to get this in one joint function.

heplesser: You are right. But this also means that `set...` is only called once, when the connection type is registered, isn't it? So I would add an `assert( supported_syn_ids_.empty() );` to the `set_...` function, plus add doxygen comments explaining the roles of set and add, respectively.

janhahne: Yes, `set_...` is only called when the connection type is registered. But I guess it will also be called for the labeled version of the synapse, at least if it is implemented in the same way as for primary connection models. So maybe we could add an `assert( not supported_syn_id( synid ) )` to the `add_...`-function and leave `set_...` without assert. Or we could add the same `assert` for both, what do you think?

heplesser: I think we should have the `not supported` assertion in both places to detect attempts to register the same synapse model multiple times.

janhahne: The compiler just "reminded" me that we cannot add `assert( not supports_syn_id( synid ) );` to `set_...` because the call of `supports_syn_id` requires an object of the class. And we also cannot make `supports_syn_id` static as it is also a virtual function of the base class.

So I guess we will have to live without an `assert` here

janhahne:
@heplesser Hopefully I addressed all of your suggestions with the new commit. If I missed something please let me know.

heplesser:
@janhahne Very nice, this has been a fast and fun review round (I hope you enjoyed it too!), a good experience with code discussion over the net. I still think that one should add the assert in the set function, see my inline comment. Conditional on that and on passing tests, :+1:.

janhahne:
@heplesser Thanks for the fast feedback. Its always nice when things are finished quickly. Regarding the additional assert I posted a comment, you might have skipped:

The compiler just "reminded" me that we cannot add `assert( not supports_syn_id( synid ) );` to `set_...` because the call of `supports_syn_id` requires an object of the class. And we also cannot make `supports_syn_id` static as it is also a virtual function of the base class.

So I guess we will have to live without an `assert` here

nestkernel/event.h
heplesser: @janhahne We could implement the check explicitly, but that would probably be overdoing it for a problem that for all we know cannot happen. So I am fine without the check.

janhahne: @heplesser Agreed!

tammoippen:
Nice work you have done! I would use a `std::set` for `supported_syn_ids_`, but I am also happy with the `std::vector`. :+1: from my side. When @heplesser is satisfied, I will merge this PR.

heplesser:
:+1:

PR #177, opened 2015-12-03
Fix105 bad min delay

### LINKED ISSUE #105: NEST 2.6.0 crashes when using different synapse parameters with a large number of connections
When using NEST v2.6.0 - together with Music- to run one of our simulations, on some occasion we get this error, and the simulation crashes:

python: ../nestkernel/scheduler.h:789: static nest::delay nest::Scheduler::get_modulo(nest::delay): Assertion `static_cast::size_type>(d) < moduli_.size()' failed. 

With version 2.4.2 there are no problems.
After digging into the problem, and testing different cases, it seems that it only happens when running simulations with a large number of connections and different parameters for each of them (e.g. delay and weight change from one to another). And even in this case, it's not consistent.
It could be worth to check if it's related to the new rounding strategy introduced in 2.6, because in one of my tests, filtering the synapses with a delay below a certain threshold (i.e. 0.1) seemed to fix the issue.

heplesser:
This pull request should fix #105 and fix some other inaccuracies in setting min and max delays as well, in particular in cases where delays are not multiples of the resolution. See also trac 903 for the most recent change on how such delays are converted; those changes were not carried through to min/max delays.


heplesser:
I suggest @tammoippen @mhelias as reviewers.

tammoippen:
I like what you have done. This makes min/max delay much easier. :+1:

nestkernel/connector_model.cpp
```diff
@@ -59,71 +59,60 @@ ConnectorModel::ConnectorModel( const ConnectorModel& cm, const std::string name
...
-  const bool new_max_delay = new_delay > max_delay_;
+  const bool new_min_delay = new_delay < min_delay_.get_steps();
+  const bool new_max_delay = new_delay > max_delay_.get_steps();
 
   if ( new_min_delay )
   {
     if ( user_set_delay_extrema_ )
-      throw BadDelay( new_delay.get_ms(), "Delay must be greater than or equal to min_delay." );
+    {
+      throw BadDelay( new_delay_ms, "Delay must be greater than or equal to min_delay." );
```
jougs: Maybe we can remind the user here that she set the min_ and may_delay herself? She might not remember that and be confused.

nestkernel/connector_model.cpp
```diff
@@ -59,71 +59,60 @@ ConnectorModel::ConnectorModel( const ConnectorModel& cm, const std::string name
...
+      min_delay_ = Time( Time::step( new_delay ) );
+    }
   }
 
   if ( new_max_delay )
   {
     if ( user_set_delay_extrema_ )
-      throw BadDelay( new_delay.get_ms(), "Delay must be smaller than or equal to max_delay." );
+    {
+      throw BadDelay( new_delay_ms, "Delay must be smaller than or equal to max_delay." );
```
jougs: See comment for l.90

nestkernel/connector_model.cpp
```diff
@@ -149,19 +138,27 @@ ConnectorModel::assert_two_valid_delays_steps( long_t new_delay1, long_t new_del
   if ( new_min_delay )
   {
     if ( user_set_delay_extrema_ )
+    {
       throw BadDelay(
         Time::delay_steps_to_ms( ldelay ), "Delay must be greater than or equal to min_delay." );
```
jougs: See comment for l.90

nestkernel/connector_model.cpp
```diff
@@ -149,19 +138,27 @@ ConnectorModel::assert_two_valid_delays_steps( long_t new_delay1, long_t new_del
...
+      min_delay_ = Time( Time::step( ldelay ) );
+    }
   }
 
   if ( new_max_delay )
   {
     if ( user_set_delay_extrema_ )
+    {
       throw BadDelay(
         Time::delay_steps_to_ms( hdelay ), "Delay must be smaller than or equal to max_delay." );
```
jougs: See comment for l.90

jougs:
I agree with @tammoippen that this is very solid work and improves the handling of min/max_delay a lot. With my minor comments addressed I'm :+1: for merging. Many thanks for the contribution.

heplesser:
@jougs I improved the error messages, I hope you are happy with them.

jougs:
This looks very nice. Many thanks again.

PR #183, opened 2015-12-06
Fix178 juqueen compile failure

heplesser:
This PR fixes the circular template specialization problem on which xlC choked. Briefly, `Connector< 1, ConnectionT >::erase()` must never be called and now will trigger an assert/abort if called. `GenericConnectorModel< ConnectionT >::delete_connection()` ensures that `erase()` is not called on `vector_like< ConnectionT >` containers of size 1.

This will fix #178.

PR #184, opened 2015-12-06
Fix for #181: Made gap junction implementation C++-standard compliant

### LINKED ISSUE #181: Gap junction implementation not strictly C++-standard compliant
Compiling NEST with Clang reveals a conflict with C++ standard compliance in the implementation of gap junctions:
```
In file included from ../../src/nestkernel/sibling_container.cpp:23:
../../src/nestkernel/event.h:770:10: error: calling a private constructor of class 'std::__1::__wrap_iter<unsigned int *>'
   fwit iit( reinterpret_cast< uint_t* >( &( *rhs ) ) );
```
The problem here is that we attempt to initialize an iterator with a pointer. This is not standard-compliant, since STL implementations may choose to implement vector iterators with data types other than pointers.

The reason for the construction above is that we need to exchange numerical information (doubles) as unsigned long, and want to do so without copying data (for efficiency). Currently, we do this by having a vector of doubles and coercing vector<ulong> iterators onto that vector.

I have [written a little example program](https://gist.github.com/heplesser/d5a1b2e0067a817168a6#file-union_vector-cpp) that uses a vector of a double-ulong union to achieve the same effect. Maybe this can be used to implement the double/ulong duality in a standard-compliant way.

@janhahne @apeyser @tammoippen @jougs I hope you also consider this a neat little technical problem in C++.

janhahne:
This PR fixes #181 by removing the `CoeffArrayIterator` and replacing it by suitable `std::vector` iterators in `GapJEvent`.

There is however still an ongoing discussion in the issue #181, further actions might be required.

### DIFF
--- a/models/hh_psc_alpha_gap.cpp
+++ b/models/hh_psc_alpha_gap.cpp
@@ -625,9 +625,10 @@ nest::hh_psc_alpha_gap::handle( GapJunctionEvent& e )
   B_.sumj_g_ij_ += e.get_weight();
 
   size_t i = 0;
-  for ( std::vector< uint_t >::iterator it = e.begin(); it != e.end(); e.next( it ) )
+  std::vector< uint_t >::iterator it = e.begin();
+  while ( it != e.end() )
   {
-    B_.interpolation_coefficients[ i ] += e.get_weight() * e.get_value( it );
+    B_.interpolation_coefficients[ i ] += e.get_weight() * e.get_coeffvalue( it );
     i++;
   }
 }
--- a/models/hh_psc_alpha_gap.cpp
+++ b/models/hh_psc_alpha_gap.cpp
@@ -626,6 +626,7 @@ nest::hh_psc_alpha_gap::handle( GapJunctionEvent& e )
 
   size_t i = 0;
   std::vector< uint_t >::iterator it = e.begin();
+  // The call to get_coeffvalue( it ) in this loop also advances the iterator it
   while ( it != e.end() )
   {
     B_.interpolation_coefficients[ i ] += e.get_weight() * e.get_coeffvalue( it );
--- a/models/hh_psc_alpha_gap.cpp
+++ b/models/hh_psc_alpha_gap.cpp
@@ -575,7 +575,7 @@ nest::hh_psc_alpha_gap::update_( Time const& origin,
   }
 
   // Send gap-event
-  GapJEvent ge;
+  GapJunctionEvent ge;
   ge.set_coeffarray( new_coefficients );
   network()->send_secondary( *this, ge );
 
@@ -619,13 +619,13 @@ nest::hh_psc_alpha_gap::handle( DataLoggingRequest& e )
 }
 
 void
-nest::hh_psc_alpha_gap::handle( GapJEvent& e )
+nest::hh_psc_alpha_gap::handle( GapJunctionEvent& e )
 {
 
   B_.sumj_g_ij_ += e.get_weight();
 
   size_t i = 0;
-  for ( fwit it = e.begin(); it != e.end(); e.next( it ) )
+  for ( std::vector< uint_t >::iterator it = e.begin(); it != e.end(); e.next( it ) )
   {
     B_.interpolation_coefficients[ i ] += e.get_weight() * e.get_value( it );
     i++;
--- a/models/hh_psc_alpha_gap.cpp
+++ b/models/hh_psc_alpha_gap.cpp
@@ -625,9 +625,9 @@ nest::hh_psc_alpha_gap::handle( GapJEvent& e )
   B_.sumj_g_ij_ += e.get_weight();
 
   size_t i = 0;
-  for ( CoeffArrayIterator it = e.begin(); it != e.end(); ++it )
+  for ( fwit it = e.begin(); it != e.end(); e.next( it ) )
   {
-    B_.interpolation_coefficients[ i ] += e.get_weight() * ( *it );
+    B_.interpolation_coefficients[ i ] += e.get_weight() * e.get_value( it );
     i++;
   }
 }
--- a/nestkernel/event.h
+++ b/nestkernel/event.h
@@ -712,10 +712,6 @@ DoubleDataEvent::clone() const
   return new DoubleDataEvent( *this );
 }
 
-// needed to serialize and deserialize SecondaryEvents
-typedef std::vector< uint_t >::iterator fwit;
-typedef std::vector< double_t >::iterator dfwit;
-
 /**
  * Base class of secondary events. Provides interface for
  * serialization and deserialization. This event type may be
@@ -742,8 +738,8 @@ public:
 
   //! size of event in units of uint_t
   virtual size_t size() = 0;
-  virtual fwit& operator<<( fwit& pos ) = 0;
-  virtual fwit& operator>>( fwit& pos ) = 0;
+  virtual std::vector< uint_t >::iterator& operator<<( std::vector< uint_t >::iterator& pos ) = 0;
+  virtual std::vector< uint_t >::iterator& operator>>( std::vector< uint_t >::iterator& pos ) = 0;
 };
 
 template < typename T >
@@ -759,7 +755,7 @@ size_t size_uint_t( T )
 
 template < typename T >
 void
-input_stream( T d, fwit& pos )
+input_stream( T d, std::vector< uint_t >::iterator& pos )
 {
   memcpy( &( *pos ), &d, sizeof( d ) );
   pos += size_uint_t( d );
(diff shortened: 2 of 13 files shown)


nestkernel/event.h
```diff
@@ -746,63 +746,32 @@ class SecondaryEvent : public Event
...
 };
 
-/**
- * Iterator class for the GapJEvent data array.
- * Generating a copy of the array at several times is too time consuming,
- * therefore we use iterators pointing to the "old" memory location of the array
- */
-class CoeffArrayIterator
+template < typename T >
+size_t size_uint_t( T )
```
heplesser: @janhahne Could you add a Doxygen comment explaining the purpose of this function?
Also,  I am not entirely happy with the function name. "size" evokes associations to `sizeof()`, which returns the size of an object in bytes. This method, though, returns the number of uints required to hold an object of type `T`. Maybe `number_of_uints_covered()` would be better? Then the places where the function is used would become
```
pos += number_of_units_covered( d );
```

nestkernel/event.h
```diff
@@ -746,63 +746,32 @@ class SecondaryEvent : public Event
...
 {
+  if ( sizeof( T ) < sizeof( uint_t ) )
+    return 1;
+  else if ( sizeof( T ) % sizeof( uint_t ) != 0 )
+    return sizeof( T ) / sizeof( uint_t ) + 1;
+  else
+    return sizeof( T ) / sizeof( uint_t );
+}
```
heplesser: Wouldn't
```
size_t num_uints = sizeof( T ) / sizeof( uint_t );
if ( num_uints * sizeof( uint_t ) < sizeof( T ) )
{
  num_uints += 1;
}
return num_uints;
```
be a more compact implementation?

heplesser:
@janhahne Very nice!

nestkernel/event.h
apeyser: How about:
```c++
template<typename T>
size_t sizeof_uint<T>() {...}
```

which can then just be called with:
```c++
size_uint_t<T>()
```

so it's clear to users & compilers that it's a type constant?
In fact, I bet if it's done a bit more obscurely with a ?: operator, it turns into a compile time constant (not a practical suggestion).

nestkernel/event.h
```diff
@@ -746,63 +746,32 @@ class SecondaryEvent : public Event
...
-
-  double& operator*()
-  {
-    return *reinterpret_cast< double* >( &( *pos_ ) );
-  }
+template < typename T >
+void
+input_stream( T d, fwit& pos )
+{
+  memcpy( &( *pos ), &d, sizeof( d ) );
```
apeyser: I still think this isn't safe for the case where sizeof(d) > sizeof(*pos). I'd decompose into uint_t arrays, and do it one by one --- a good compiler should give you the same thing as a memcpy for the common case

nestkernel/event.h
```diff
@@ -746,63 +746,32 @@ class SecondaryEvent : public Event
...
-  bool operator!=( const CoeffArrayIterator& rhs )
-  {
-    return pos_ != rhs.pos_;
-  }
-};
+template < typename T >
+void
+output_stream( T& d, fwit& pos )
+{
+  memcpy( &d, &( *pos ), sizeof( d ) );
```
apeyser: Same as for the other direction.

janhahne:
@heplesser @apeyser I did a couple of changes, which hopefully address all comments here and in the corresponding issue. Would you please review the changes?

models/hh_psc_alpha_gap.cpp
```diff
@@ -619,15 +619,16 @@ nest::hh_psc_alpha_gap::handle( DataLoggingRequest& e )
...
 
   B_.sumj_g_ij_ += e.get_weight();
 
   size_t i = 0;
-  for ( CoeffArrayIterator it = e.begin(); it != e.end(); ++it )
+  std::vector< uint_t >::iterator it = e.begin();
+  while ( it != e.end() )
   {
-    B_.interpolation_coefficients[ i ] += e.get_weight() * ( *it );
+    B_.interpolation_coefficients[ i ] += e.get_weight() * e.get_coeffvalue( it );
```
heplesser: In this loop, it is not obvious at all how `it` is advanced. This happens as a side-effect of `e.get_coeffvalue( it )`, but to understand that one needs to read that function. Maybe this needs to be like this, but then you ought to add a comment explaining how `it` is updated.

nestkernel/event.h
```diff
@@ -742,73 +738,82 @@ class SecondaryEvent : public Event
...
-  CoeffArrayIterator( const fwit& rhs )
-    : pos_( rhs )
-  {
-  }
+/**
+ * This template function writes data of type T to a given position of a
+ * std::vector< uint_t >. Please note that this function does not increase the
+ * size of the vector, it just writes the data to the position given by the iterator.
+ * The function is used to write data from SecondaryEvents to the NEST communcation buffer.
+ */
```
heplesser: The documentation should point out that `pos` is advanced.

Also, doxygen docstrings should start with a single line briefly summarizing the method (used in tables of methods in the documentation).

nestkernel/event.h
```diff
@@ -742,73 +738,82 @@ class SecondaryEvent : public Event
...
+  size_t num_uints = number_of_uints_covered< T >();
+  size_t left_to_copy = sizeof( T );
 
-  CoeffArrayIterator( const CoeffArrayIterator& rhs )
-    : pos_( rhs.pos_ )
+  for ( int i = 0; i < num_uints; i++ )
   {
+    memcpy( &( *( pos + i ) ),
+      c + i * sizeof( uint_t ),
+      left_to_copy < sizeof( uint_t ) ? left_to_copy : sizeof( uint_t ) );
```
heplesser: `std::min(left_to_copy, sizeof( uint_t ))` would be shorter and more expressive.

nestkernel/event.h
```diff
@@ -742,73 +738,82 @@ class SecondaryEvent : public Event
...
-    else
-      inc = sizeof( double ) / sizeof( uint_t );
+  pos += num_uints;
+}
 
-    pos_ += inc;
+/**
+ * This template function reads data of type T from a given position of a std::vector< uint_t >.
+ * The function is used to read SecondaryEvents data from the NEST communcation buffer.
+ */
```
heplesser: See above, mention that `pos` is advanced.

nestkernel/event.h
```diff
@@ -742,73 +738,82 @@ class SecondaryEvent : public Event
...
+  size_t num_uints = number_of_uints_covered< T >();
+  size_t left_to_copy = sizeof( T );
 
-  double& operator*()
+  for ( int i = 0; i < num_uints; i++ )
   {
-    return *reinterpret_cast< double* >( &( *pos_ ) );
+    memcpy( c + i * sizeof( uint_t ),
+      &( *( pos + i ) ),
+      left_to_copy < sizeof( uint_t ) ? left_to_copy : sizeof( uint_t ) );
```
heplesser: `std::min()`, see above.

heplesser:
@janhahne Nice work! I added a little more nit-picking ;). The last check seems to be stuck due to technical problems at Travis, I hope it will have finished by the morning.

jougs:
I've restarted the TravisCI job as it was still in the "booting" state.

janhahne:
@heplesser @apeyser I made the changes according to your comments. Furthermore I compiled and run the code (together with #183) on Juqueen and K. Works fine on both supercomputers. Regarding speed it is a little bit slower on Juqueen (but still acceptable), but must faster then before on K.

apeyser:
Question: does NEST support heterogeneous MPI? Casting uints by MPI across different machines is safe, if the allgather is told that it is casting uints. Is that all that we're sending, so that currently our code is heterogeneous safe? 

If so, this break heterogeneity, both endian-wise and by double format (unless we claim to only use ieee doubles), which would require some documentation line and run time check. 

If not --- then nothing need be done, we just need it documented, and maybe a run time check that we're actually running in the correct environment (which would be a different project than this one).

apeyser:
So, all I have as a current suggestion is to add a comment on strict aliasing or a reference to this pull so that a year from now we can track down why we did it this way and why it should be safe. Then :+1:

heplesser:
@apeyser For all I know, we have never thought about heterogeneous MPI so far (@tammoippen mentioned it in private conversation a few days ago), so at present, I would consider NEST as unsafe on heterogeneous systems. Thanks for opening #186 to follow up on this.

heplesser:
@janhahne If you add the comment/reference requested by @apeyser (maybe with a TODO in front, so it is picked up by IDEs), I will merge this PR.

janhahne:
@heplesser Done.

apeyser:
Good job all --- this sets the standard for adding in new mpi communicated info over the spike buffers.

Issue #149, opened 2015-11-04
MyModule does not work with gcc 4.7.2 / missing symbols


heplesser:
As Oliver Breitwieser reported on NEST User 30 Oct 2015, loading mymodule fails when NEST 2.8.0 is built with gcc 4.7.2 under Linux. I have reproduced this failure, using 4.7.2, but not using gcc 5 under OSX.

To reproduce:
1. Build mymodule as instructed (http://nest.github.io/nest-simulator/extension_modules)
1. Run `nest -c "(mymodule) Install"
1. NEST fails, reporting it cannot open mymodule due to "file not found". 

As Oliver pointed out, the actual error is not that the file is not found, but that it does not load due to missing symbols:
```
_ZN4nest4Time11LimitNegInf4ticsE
_ZN4nest4Time11LimitPosInf4ticsE
```

I have a suspicion that the compiler gets confused by the rather compact 
```C++
  static struct LimitPosInf
  {
    static const tic_t tics = tic_t_max / Range::INF_MARGIN + 1;
    static const delay steps = delay_max;
#define LIM_POS_INF_ms double_t_max // because C++ bites
  } LIM_POS_INF;
```
code.

@apeyser, could you take a look at this as you wrote the code? I'd also appreciate a little longer comment explaining how "C++ bites", necessitating the macro.

apeyser:
:laughing: 
in C++ < 11, double literal constants aren't counted as compile time constants, so they can't be static const's in headers, but literal integers and a set of operations on them are. It's one of those poor compromises made at the C++ standardization committee. Thus, you can't have LIM_POS_INF.ms, but need to make it a macro.

For some reason, it appears that gcc 4.7.2 isn't inlining LIM_POS_INF.{tics,steps}. I'm not sure whether this is an obscure C++ rule case, a bug in gcc 4.7.2 or a move towards C++ 11 that is creating a corner case.

apeyser:
I looked into it, and it appears that in C++ 03 this is not fully defined, and in C++ 11 involves great legal gymnastics regarding "odr-used". So either there's a case somewhere downstream where tics is referenced in such a way that requires it's address (such as a reference) that under some compiles isn't optimized away, or gcc 4.7.2 is acting funny and failing to recognize it.

The safe thing is to define inside of nest_time.cpp:
    const tic_t Time::LimitPosInf::tics;
    const delay Time::LimitPosInf::delay_max;

jougs:
Closing via #189.

PR #189, opened 2015-12-10
Define linkage for time constants in c file

apeyser:
Fixes (I think) issue #149 but that one was specifically for modules under gcc 4.7.2. It needs testing by someone with that compiler -- so we need to contact the original reporter and have this tested.

So: @heplesser as original issue maker, and @jougs since he also knows Oli.


heplesser:
@apeyser I tested this on Stallo with 4.7.2, where I had also reproduced the error. With your fix, the module loads nicely. So :+1:.

jougs:
I tested the code on my machine and it looks as if it does not cause any harm when the bug did not occur, so :+1: and merging.

PR #195, opened 2015-12-17
Improved documentation on pack_pointer/validate_pointer

janhahne:
Added some documentation on this functions introduced by the gap-junction framework. No changes to the actual code, so probably no need for a review.


jougs:
Many thanks for the added documentation. As I'm :+1: and it does not touch any code, I'm directly merging.

PR #196, opened 2015-12-17
Fix pylab and matplotlib imports

jougs:
This removes imports of matplotlib where it is actually not used, and import aliases for pylab, as they are lowering comprehensibility.

@hannahbos: can you please have a look at this?


hannahbos:
:+1

abigailm:
:+1:  and merging

PR #204, opened 2016-01-10
Solve path problems with imports on Python 3.5

prcastro:
All tests are passing when I run from Python 3.5:

```
import nest
nest.test()
```

Would be nice to have travis testing nest against Python 3 also, so this kind of regression doesn't happen.

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -50,7 +50,7 @@ except:
 
 
 from . import pynestkernel as _kernel
-import lib.hl_api_helper as hl_api
+import nest.lib.hl_api_helper as hl_api
 
 engine = _kernel.NESTEngine()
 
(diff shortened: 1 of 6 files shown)


apeyser:
@prcastro 
Is there an issue that describes what's the history of this?

The paths in init should probably be an explicit relative import -- import .lib.yadda-yadda
What happened in test_all.py that the relative imports no longer work?

jougs:
@prcastro: I've had a look at your PR and was able to reproduce the problems using Python3.4. However, I came up with a smaller changeset that solved all problems with imports. Can you please check out if https://github.com/jougs/nest-simulator/tree/split_pynest_hlapi solves the problem for you?

If there are still problems for you, could you please elaborate on them? We can then iterate on this.

In any case, many thanks for your contribution and your interest in NEST.

jougs:
Closing due to the merge of #233.

Issue #152, opened 2015-11-10
Skip documentation when installing cross-compiled NEST


heplesser:
This is issue is based on a report by Babak Keshavarz on NEST User 10 Nov 2015.

Running `make install` on a cross-compiled NEST may lead to a crash during the installation of NEST Online documentation because the cross-compiled SLI executable may not work on the machine executing make. 

Installing the documentation is the last step in the NEST install process. We should check whether SLI can run, and if not, issue an message pointing out that online help is not installed.

See also #153.

steffengraber:
I added a set of helping tools for the NEST documentation. See pull request #157.
The python tool generate_help2jsonhtml.py in extras/userdoc/generator extract the whole documentation from the code files (".sli", ".cpp", ".cc", ".h") and write it to HTML and JSON files.
See the documentation: extras/userdoc/README.md.
It is very fast and maybe it is an solution.

apeyser:
A --without-doc configure flag should be added, since it appears to me that no such flag exists.

Every solution will require some set of tools, which will lead to corner cases with failures on top of wasted compilation time. We already have the "IS_BLUEGENE" flag to suppress slidoc because of this problem (among others). We should have a configuration that skips all of it from the get go.

jougs:
Also see trac.791 for a related issue.

apeyser:
Yes -- a number of options should be explicitly disable-able such as docs and readline, instead of being the result of special-case flags (such as enable-bluegene). Whether they are done by implicit checks and disabling, or by failing on tests and requiring explicit disabling is a further question of complexity. But a first cut is making sure that one can disable things like docs piecewise.

tammoippen:
With #213 being merged, installing NEST skips generating the help documentation when cross compiling; see [here](https://github.com/nest/nest-simulator/blob/master/sli/CMakeLists.txt#L145). Hence closing as fixed.

PR #213, opened 2016-01-29
Clean build system and installation

tammoippen:
In trac we have a milestone `Clean build system and installation`:

> The build system of NEST has grown over the years and is currently containing custom checks for many systems not in existence anymore. This milestone is about fixing and cleaning the build system.

In order to solve the contained tickets, I rewrote the build system from scratch
using [`cmake`](https://cmake.org/) for different reasons:

* Very good cross platform support, where autotools is very much rooted in GNU Li/Unix based operating systems.
* It has very good support for cross compilation using toolchain files for the different targets/compiler combinations.
* As a "meta"-build system you can generate various "target"-build systems: Unix Makefiles, Ninja and project files for many different IDE's. Every developer can work with his preferred tool.

Every build configuration, that was possible before, should be possible with the cmake version, too. I tried to keep most of the work-flows intact. For NEST installation details, please see INSTALL.

The trac tickets that should be solved by this PR:

* trac.523 "make dist target includes compilation byproducts with libltdl" as libltdl will not be generated during configuring / building and the system libltdl will be used.
* trac.768 "Use pkgconfig for library detection wherever possible" as cmake has a [find_package](https://cmake.org/cmake/help/v3.0/command/find_package.html) infrastructure which is "simple" to extend. Internally, they often use pkgconfig.
* trac.775 "Change OpenMP configuration option --with-openmp to take the flag as argument" is enabled.
* trac.791 "Remove `--enable-bluegene` ./configure flag" as the corresponding toolchain sets all necessary options.
* trac.865 "check cython version in configure" as we check for Cythen version now.
* trac.923 "Do have gcc/g++ default guess when setting compiler options" as cmake has support for many compilers [included](https://github.com/Kitware/CMake/tree/master/Modules/Compiler). I tried to use as few as possible default flags in the CMakeLists.txt files

Partly or not solved trac issues from this milestone:

* trac.654 "Move all examples from all over the source tree to the examples directory" _partially_ solved, as all examples are installed, but in the same directory structure as before.
* trac.781 "NEST build system doesn't respect --docdir" __NOT__ solved, but can be adjusted by modifying `INSTALL_DOC_DIR` to be a user option.
* trac.904 "Review examples" did __NOT__ happen (although all examples are installed now).

The GitHub Issues affected by this:
* #152 Documentation will be skipped when cross compiling.
* #56 Both, dynamic and static building of user modules is enabled with cmake. It might also be possible to build a user module as a cmake _add_subdirectory_, but this requires additional work.

I propose as reviewers @jougs, @apeyser, @heplesser, but again I also welcome @nest/nest-developers to review this PR (especially those with cmake experiences).


heplesser:
@tammoippen Very nice work! 

I have tested so far on OSX and a large Linux system and all works well. The cmake and make output produced is tidier than previously, and building seems faster as make now appears to compile files from multiple directories simultaneously instead of completing everything in one directory before moving to the next.

Some minor issues:
- I misspelled `-Dwith-mpi` as `-Dwith_mpi`, which cmake silently ignored; it just didn't turn on MPI then. Is there any way to make cmake complain about unknown options?
- After running cmake, the build directory contained to executables, `assert_value` and `segfault_value`. Should cmake delete such test programs?

I have not tested building `mymodule` yet.

tammoippen:
> * I misspelled -Dwith-mpi as -Dwith_mpi, which cmake silently ignored; it just didn't turn on MPI then. Is there any way to make cmake complain about unknown options?

Usually you get a warning at the end of the cmake config; something like:

    -- Configuring done
    -- Generating done
    CMake Warning:
      Manually-specified variables were not used by the project:

        with_mpi


    -- Build files have been written to: ...

> * After running cmake, the build directory contained to executables, assert_value and segfault_value. Should cmake delete such test programs?

Test programs will be removed with c7ca61c.

apeyser:
@tammoippen I'll try to test with JQ, and we probably need a K computer tester...

tammoippen:
@apeyser I added instructions for compilation on JQ and K in the `INSTALL` file. Thank you for testing.

jakobj:
just finished compiling and installing it on K. incredibly easy! great work! :) but it would be nice to include the optimize flag for K in the toolchain file (`-Kfast`). is that easily possible? in the best case this could easily enabled/disabled by the user.

JanneM:
Just to note that if you build NEST on K with Python and MUSIC enabled (I haven't tried other configurations), an internal compiler error - probably an out of memory situation - forces you to either omit -Kfast or disable synapse models that you don't use. For that reason it may not be the best idea to hardcode the -Kfast parameter in the build file.

tammoippen:
@JanneM The compilation also fails when we compile without Python and without MUSIK.

I have looked at what `-Kfast` enables (from `man fccpx`):

    -O3 -Kdalign,eval,fast_matmul,fp_contract,fp_relaxed,ilfunc,lib,mfunc,ns,omitfp,prefetch_conditional,rdconv -x-

I looked at all options and figured that `-x-` might be the most demanding option. Using `-Kfast -x0` disables this option: almost everything compiles, except for the final target `nest`. Apparently, the `ModelModule` is compiled wrong then. Using `-Kfast -x1` succeeds. I have not tested any impact on performance. In the last commit (ffce7e4) I adjusted the documentation for installation on K.

    -xn    Inline  expansion is performed for functions in which the number of initialized dec-
           larations and executed statements is equal to or less than the specified number n.
           n can be a number from 0 to 2147483647.

    -x- [...] This  option  determines  whether  to perform inline expansion by using the value of
              best n calculated by the compiler.

jakobj:
@tammoippen it is not in general true that the compilation fails whenever the `-Kfast` switch is enabled. in our experience it mainly depends on the login node and on how many times you're willing to restart compilation. usually you can get NEST to compile on K in two or three tries, _with_ the `-Kfast` option.
nevertheless, i think the way you now solved the problem by updating the documentation should be fine.

JanneM:
@jakobj "Randomly works one time in three" isn't really working, as such. And when I build on compute nodes (you can't cross-compile with python) I have never once managed to finish the compile with -Kfast. 

I'm trying a new build on K right now; let's see if @tammoippen solution above works.

jakobj:
just as a clarification: i was referring to building _without_ Python and _without_ MUSIC. sorry i should've been more precise. in addition you should try logging out and logging back in to get to a different login node once a build fails.

jougs:
Please also consider closing #56 when this is done.

apeyser:
Preliminary problem with JQ: the cython check confuses the cython/python used to cythonize with the run-time python executable that has nothing to do with cython. These are distinct -- in fact, this is kind of a general problem, in that I should be able to cythonize the tree on one machine and configure on another, since the C produced is generic and isn't a configure time event.

apeyser:
-- Configuring done
CMake Error at CMakeLists.txt:792 (add_custom_target):
  Error evaluating generator expression:

    $<CONFIG>

  $<CONFIG> expression requires exactly one parameter.

 Suggestion?

apeyser:
I need cmake to run make targets??? Really?? <<Long rant on cmake>>

apeyser:
The config.h.in files appear wrong:
#cmakedefine HAVE_DIRENT_H
should be
#cmakedefine HAVE_DIRENT_H 1
for example.  And HAVE_MUSIC doesn't get defined at all, it was missed.

This breaks builds of pynest and (probably) MUSIC, so I'm not sure how this would have gotten to this point. Did I somehow miss a switch in branch from nest3_cmake? Or have some of the tests been turned off?

apeyser:
And the bluegene q platform files seem to be missing dynamic library flags. This is becoming incredibly painful. 

I'm starting to think looking through the CMakeLists tree that this pull needs a bit more discussion and code review. A few things noted:
* Hard-coded flags (-Wall for example) that are gcc specific (and clang inherited) but should be at least organized into a platform specific module, if not simply left to the person compiling to set in their CXX_FLAGS.
* If-else(expr) syntax is evil, and the expr in else(expr) should be dropped as it is optional. 
* Bluegene platform files are incomplete
* The dependency of python on cython is a problem
* config.h needs to be checked
* Doxygen is a found flag, rather than an enabled element. Is that only used for make doc? Are any other targets dependent on doc target?
* We still have this "Don't install into /usr/local". Why? To avoid someone messing this up from time to time? I've never seen another project do this, because this is pure policy check.
* Are all options documented in INSTALL? Can we group the options and flags at the top of the CMakeLists so we can do an easy check against INSTALL instructions?

This a large chunk of crucial code. We need reviewers (and hopefully more reviewers) to go through this code more carefully, otherwise even more time will be spent on issues and user mail dealing with debugging after the fact.

apeyser:
* Cython version check requires format d+.d+.d+ --- but 0.22 can be (is) reported for v22

apeyser:
* libltdl removed -- and on bgq, you need to bring your own. Possibly a problem with FindLTDL not matching current platform files.

tammoippen:
Hello @apeyser,

Thank you for the valuable feedback! In no specific order my feedback:

> * libltdl removed -- and on bgq, you need to bring your own. Possibly a problem with FindLTDL not matching current platform files.

> And the bluegene q platform files seem to be missing dynamic library flags. This is becoming incredibly painful.

On BG/Q it is recommended to build static binaries, hence the `libltdl` is not necessary. Newer `cmake` versions contain BG/Q platform files for static and dynamic building with `XLC` and `GCC`, but since the standard `cmake` on JUQUEEN does not have them, I only created a platform file, that fulfills the recommended configuration and eases the installation. If these original cmake files are used the flags from `cmake/Platform/BlueGeneQ.cmake:26-32` also have to be added on the command line.

> * Cython version check requires format d+.d+.d+ --- but 0.22 can be (is) reported for v22

I hope it is fixed with fac0b4ff: the format is now `d+.d+(.d+)?`.

> The config.h.in files appear wrong:
> #cmakedefine HAVE_DIRENT_H
> should be
> #cmakedefine HAVE_DIRENT_H 1
> for example. And HAVE_MUSIC doesn't get defined at all, it was missed.

I defined the `HAVE_MUSIC` flag with 64ee3b3a -- must have missed it. For `HAVE_DIRENT_H`, this seems to be a relict from autoconf. It is not used anywhere. Along with it I removed some more unused defines and checks (c95feb38) to represent, what is actually used in NEST.

> I need cmake to run make targets??? Really?? <<< Long rant on cmake >>>

Apparently cmake uses itself... I was not aware of this.

I also added custom targets or extended previous targets to use cmake. We can also decide to use `bash`/`zsh`/`ksh`/`sh`... for running these targets, but I thought, since we anyway need cmake to build NEST after this PR, it would be more platform independent to use cmake here. The targets in question are:

* installing the online help (also for MyModule).
* setting the rcsinfo during `doc` and `fulldoc`.

> Added: I see, the python flag is just compile time. I'm not a fan of burying this inside of cmake.

This is similar to the autoconf setup (if you are referring to the `_IS_PYNEST` flag): if we are to compile the sli-nest executable, this flag should not be present. Only when we build the PyNEST `pynestkernel.so` library, a.k.a. the `nest_lib` target which is only used by the `pynestkernel`.

> Preliminary problem with JQ: the cython check confuses the cython/python used to cythonize with the run-time python executable that has nothing to do with cython. These are distinct -- in fact, this is kind of a general problem, in that I should be able to cythonize the tree on one machine and configure on another, since the C produced is generic and isn't a configure time event.
>
> Added: The python flag isn't just for cythonization -- it needs to be on in order to compile pynest as well. This definitely has to be separated --- the current setup makes cross-compiling with python very difficult.

I was not aware, that you can install cython without python, hence the configure time encapsulation. I cross-compile a cythonized pynestkernel.pyx (on my laptop) on BG/Q with success. We should talk about this, as I think I am not getting the issue. With f19c10b3 I made cythonizing optional, allowing compilation with a cythonized pynestkernel on a system with an old version of cython.

> CMake Error at CMakeLists.txt:792 (add_custom_target):
> Error evaluating generator expression:
> 
>     $<CONFIG>
> $ expression requires exactly one parameter.

I encountered this warning as well, and I think it is due to the dated `cmake` on JuQueen. I replaced it by giving the current `${CMAKE_BUILD_TYPE}` only (ea24f93f). This is only for the target `make check`, which uses `ctest` to run our tests (experimental as of now).

> * Doxygen is a found flag, rather than an enabled element. Is that only used for make doc? Are any other targets dependent on doc target?

Yes, doxygen is only used for `make doc` and `make fulldoc` and we only need the 
`DOXYGEN_EXECUTABLE` from the package. No other target is dependent on `doc`/`fulldoc`.

> * We still have this "Don't install into /usr/local". Why? To avoid someone messing this up from time to time? I've never seen another project do this, because this is pure policy check.

I did not want to break expected behavior from before (to much) and was explicitly asked to include it...

> * Are all options documented in INSTALL? Can we group the options and flags at the top of the CMakeLists so we can do an easy check against INSTALL instructions?

Yes. I grouped all options in CMakeLists.txt:168-196 and moved them up to 42-70 in f44ad182.

> * If-else(expr) syntax is evil, and the expr in else(expr) should be dropped as it is optional.

Done in 8de09d21.

> * Hard-coded flags (-Wall for example) that are gcc specific (and clang inherited) but should be at least organized into a platform specific module, if not simply left to the person compiling to set in their CXX_FLAGS.

There are three hard-coded flags: `-Wall` if `-Dwith-warning=ON` is defined (default), `-g` if `-Dwith-debug=ON`, and `-O3` if `-Dwith-optimize=ON`. All of these can easily be shut off, e.g. `-Dwith-warning=OFF`, or set to a different value, e.g. `-Dwith-optimize="-Kfast -x1"`. All other options/flags are set by cmake, which indeed has platform/compiler specific modules.

apeyser:
@tammoippen: 

>On BG/Q it is recommended to build static binaries, hence the `libltdl` is not necessary. Newer `cmake` versions contain BG/Q platform files for static and dynamic building with `XLC` and `GCC`, but since the standard `cmake` on JUQUEEN does not have them, I only created a platform file, that fulfills the recommended configuration and eases the installation. If these original cmake files are used the flags from `cmake/Platform/BlueGeneQ.cmake:26-32` also have to be added on the command line.

I'll check whether a statically built nest can load python shared library modules.

>> The config.h.in files appear wrong:
>> #cmakedefine HAVE_DIRENT_H
>> should be
>> #cmakedefine HAVE_DIRENT_H 1
>> for example. And HAVE_MUSIC doesn't get defined at all, it was missed.

>I defined the `HAVE_MUSIC` flag with 64ee3b3a -- must have missed it. For `HAVE_DIRENT_H`, >this seems to be a relict from autoconf. It is not used anywhere. Along with it I removed some more >unused defines and checks (c95feb38) to represent, what is actually used in NEST.

My point was that almost all the flags are incorrect in that way. The HAVE_XXX defs should be to 1 and not to nothing, since that's what the autoconf headers do and thus we avoid having a macro conflict (and neither us nor python have namespaced the config.h macros, so we need to avoid conflicts).

>> I need cmake to run make targets??? Really?? <<< Long rant on cmake >>>

> Apparently cmake uses itself... I was not aware of this.

Yes --- it's an unavoidable part of cmake, a design error at the heart of the machine.

>> Added: I see, the python flag is just compile time. I'm not a fan of burying this inside of cmake.

>This is similar to the autoconf setup (if you are referring to the `_IS_PYNEST` flag): if we are to compile the sli-nest executable, this flag should not be present. Only when we build the PyNEST `pynestkernel.so` library, a.k.a. the `nest_lib` target which is only used by the `pynestkernel`.

We should track some of these issues for discussion --- some of this may just be tradition.

>> Added: The python flag isn't just for cythonization -- it needs to be on in order to compile pynest as well. This definitely has to be separated --- the current setup makes cross-compiling with python very difficult.

>I was not aware, that you can install cython without python, hence the configure time encapsulation. I cross-compile a cythonized pynestkernel.pyx (on my laptop) on BG/Q with success. We should talk about this, as I think I am not getting the issue. With f19c10b3 I made cythonizing optional, allowing compilation with a cythonized pynestkernel on a system with an old version of cython.

Other way around --- but you solved the problem I was intending to communicate with f19c10b3 I believe.


>> * Doxygen is a found flag, rather than an enabled element. Is that only used for make doc? Are any other targets dependent on doc target?

>Yes, doxygen is only used for `make doc` and `make fulldoc` and we only need the 
`DOXYGEN_EXECUTABLE` from the package. No other target is dependent on `doc`/`fulldoc`.

So shouldn't the search be disable-able? I guess we can leave it until someone has a problem with a broken doxygen that causes a failure in the configure script.

>> * We still have this "Don't install into /usr/local". Why? To avoid someone messing this up from time to time? I've never seen another project do this, because this is pure policy check.

>I did not want to break expected behavior from before (to much) and was explicitly asked to include it...

One more thing to track and ask.

>> * Hard-coded flags (-Wall for example) that are gcc specific (and clang inherited) but should be at least organized into a platform specific module, if not simply left to the person compiling to set in their CXX_FLAGS.

>There are three hard-coded flags: `-Wall` if `-Dwith-warning=ON` is defined (default), `-g` if `-Dwith-debug=ON`, and `-O3` if `-Dwith-optimize=ON`. All of these can easily be shut off, e.g. `-Dwith-warning=OFF`, or set to a different value, e.g. `-Dwith-optimize="-Kfast -x1"`. All other options/flags are set by cmake, which indeed has platform/compiler specific modules. 

Is there a standard way to handle these kinds of compiler specific options?

I suggest @w-klijn take a look at the pull request.

tammoippen:
>>> The config.h.in files appear wrong:
>>> #cmakedefine HAVE_DIRENT_H
>>> should be
>>> #cmakedefine HAVE_DIRENT_H 1
>>> for example. And HAVE_MUSIC doesn't get defined at all, it was missed.

>> I defined the HAVE_MUSIC flag with 64ee3b3 -- must have missed it. For HAVE_DIRENT_H, this seems to be a relict from autoconf. It is not used anywhere. Along with it I removed some more unused defines and checks (c95feb3) to represent, what is actually used in NEST.

> My point was that almost all the flags are incorrect in that way. The HAVE_XXX defs should be to 1 and not to nothing, since that's what the autoconf headers do and thus we avoid having a macro conflict (and neither us nor python have namespaced the config.h macros, so we need to avoid conflicts).

Fixed with d724382.

tammoippen:
> [-Wall et al.] Is there a standard way to handle these kinds of compiler specific options?

Usually you define the [build type](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html) with cmake and it will add some compiler specific options. For GNU compilers you can see the flags at the end of [GNU.cmake](https://github.com/Kitware/CMake/blob/master/Modules/Compiler/GNU.cmake). In general you can see the effective build type flags in the generated `CMakeCache.txt` file. Most of these compiler files do not add warning flags, this is why I put the option `with-warning` in. As all of the build type options above `CMAKE_${lang}_FLAGS_DEBUG` add `-DNDEBUG` (which breaks our `make installcheck`) I added `with-optimize` and `with-debug`. BTW: these three options are present at NEST+autotools as well.

> [doxygen] So shouldn't the search be disable-able? I guess we can leave it until someone has a problem with a broken doxygen that causes a failure in the configure script.

Should NEST deal with broken doxygen? Here we use a standard [`find_package`](https://cmake.org/cmake/help/v3.0/module/FindDoxygen.html) command to find doxygen/dot. If it is not found the targets (`doc`/`fulldoc`) will emit a message (since ec90627).

apeyser:
>> [-Wall et al.] Is there a standard way to handle these kinds of compiler specific options?

>Usually you define the [build type](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html) with cmake and it will add some compiler specific options. For GNU compilers you can see the flags at the end of [GNU.cmake](https://github.com/Kitware/CMake/blob/master/Modules/Compiler/GNU.cmake). In general you can see the effective build type flags in the generated `CMakeCache.txt` file. Most of these compiler files do not add warning flags, this is why I put the option `with-warning` in. As all of the build type options above `CMAKE_${lang}_FLAGS_DEBUG` add `-DNDEBUG` (which breaks our `make installcheck`) I added `with-optimize` and `with-debug`. BTW: these three options are present at NEST+autotools as well.

That seems pretty broken, to put debug & optimization flags into a global namespace when flags are compiler specific (in principle). This is why autoconf lacks build_type machinery --- that's really either fully on the developers side of the problem, or you have to have a full by-compiler-version namespace.

>Should NEST deal with broken doxygen? Here we use a standard [`find_package`](https://cmake.org/cmake/help/v3.0/module/FindDoxygen.html) command to find doxygen/dot. If it is not found the targets (`doc`/`fulldoc`) will emit a message (since ec90627).

Not "deal" with it in any intelligent way. Just shouldn't break because something you're not interested in is broken. I tend to err on the side of making everything disable-able.

tammoippen:
> This is why autoconf lacks build_type machinery

In cmake it is optional: when you do not specify a build type, these optimization flags will not be used. This is why I advertised them nowhere and introduced the three options.

apeyser:
> Usually you define the build type with cmake and it will add some compiler specific options. For GNU compilers you can see the flags at the end of GNU.cmake.

Ah, I overlooked that. They are then isolated by compiler when you use GNU.cmake or other similiar extensions. But if I'm understanding correctly, our cmake infrastructure isn't currently including a GNU.cmake or equivalent, since the autoconf builds weren't isolated in such a manner, right?

tammoippen:
In the beginning of the configure stage, cmake decides which platform and compiler should be used (or you decide by setting `-DCMAKE_C_COMPILER=...` or `-DCMAKE_TOOLCHAIN_FILE=...`). Here it loads the corresponding compiler and platform file. Along with the  `CMAKE_${lang}_FLAGS_{DEBUG|RELEASE|...}` it also sets how to compile source files with this compiler, how to link static/dynamic libraries/modules and executables. The `CMAKE_${lang}_FLAGS_{DEBUG|RELEASE|...}` variables 'just' contain common optimization flags for that compiler; these I planned to ignore for the NEST configuration to be more similar to `autotools`.

apeyser:
It makes sense for the first draft to reproduce autotools... except will we follow up and correct it later? A lot of the autotool behaviors are historical artifacts without a good rationale, or without a discussed rationale. So, is the intent to do these "incorrect" things only as a first pass with a commitment to fix the issue later, to leave them as so, or should we build a fairly clean CMAKE build from the beginning as long as we're going through the pain of switching?

tammoippen:
I think this is a good platform to gather interesting stuff about our build system; then create issues to discuss them in detail. From what I see in this thread, there are three points:
* Why do we error, when somebody does not give an install prefix?
* The `_IS_PYNEST` flag merges the SLI-NEST and PyNEST startup behaviour? Is this necessary or can we refactor this to be more modular.
* Enabling the flag `NDEBUG` breaks `make installcheck`, as we relay on assertions to fail in certain tests. Is it ok to define it for release versions. If yes, we could use the `BUILD_TYPE` system from cmake as well.

Do you have more points?

heplesser:
I would vote for a clean CMAKE build. There is really a lot of old cruft in the autotools code that has neither been used nor tested for a long time, such as system support (Hitachi, Scali MPI) or tools support (Vampir). I would suggest a build system that works for the systems we are using today, e.g. Linux, OSX, BG, K, gcc, xlc, K, clang and Intel compilers.

jougs:
I fully agree with @heplesser. Please don't try to reproduce the behavior of the old setup if that's insane or against CMAKE standards.

mdjurfeldt:
But doesn't switching to CMAKE mean that you loose a lot of freedom to
adapt the NEST build to the environment? I mean, cmake is fine as long as
it is fine, but autotools provides considerably better tools for testing
things. Also, this is the standard build system for GNU software (which
might be an argument to some of us. :) Yes, autotools are horrible, but it
is the best thing and a sign of good taste. :-))
Den 9 mar 2016 11:04 skrev "Hans Ekkehard Plesser" <notifications@github.com

mdjurfeldt:
(Sorry, I didn't follow this issue from the beginning. Now I realize that
the whole point from the start of this issue is a build system written in
cmake. I don't think it is a good idea, but I'll be quiet now.)

apeyser:
Next stage to accept this into NEST-3 and place other questions into sub issues:
1. I need to confirm that statically linked nest under python can pull in python modules
2. One more person with cmake experience should give it a look over and a thumbs up

I believe that all the other issues should go into follow up issues.

apeyser:
So, I can build on JQ with under the latest pull:
````
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Platform/BlueGeneQ_XLC.cmake       -DCMAKE_INSTALL_PREFIX:PATH=../install       -Dwith-python=/bgsys/local/python3/3.4.2/bin/python3 -Dwith-readline=OFF -Dstatic-libraries=ON -DPYTHON_LIBRARY=/bgsys/local/python3/3.4.2/lib/libpython3.4m.a -DPYTHON_INCLUDE_DIR=/bgsys/local/python3/3.4.2/include/python3.4m ..
````

but on import load, I get an empty module. I think we do have to build with dynamic libraries to get cmake to properly build the nest.so for python to import. It's not that nest must be dynamically built, it's that the pynest part must be dynamically built, and cmake like autotools doesn't know how to distinguish them. I still need to investigate what cmake is putting out, which allows the import to happen "successfully" without actually loading any symbols.

File: CMakeLists.txt [hunk 1]
```diff
@@ -0,0 +1,969 @@
...
+##################         All User Defined options           ##################
+################################################################################
+
+# change NEST defaults
+set( tics_per_ms "1000.0" CACHE STRING "Specify elementary unit of time. [default 1000.0]" )
+set( tics_per_step "100" CACHE STRING "Specify resolution. [default 100]" )
+option( with-ps-arrays "Use PS array construction semantics. [default=ON]" ON )
+
+# add user modules
+option( external-modules "External modules to be linked in. Divided by ';'. [default=OFF]" OFF )
```
heplesser: Suggestion: "External NEST modules to be linked in, separated by ';'."

CMakeLists.txt
```diff
@@ -0,0 +1,969 @@
...
+  endforeach ()
+endif ()
+
+if ( with-libraries )
+  if ( with-libraries STREQUAL "ON" )
+    message(FATAL_ERROR "-Dwith-libraries requires full library paths.")
+  endif ()
+  foreach ( lib ${with-libraries} )
+    if ( EXISTS "${lib}" )
+      link_libraries( "${lib}" )
```
heplesser: Eclipse tells me that `link_libraries` is deprecated and that one should use `target_link_libraries` instead, not sure how important this is.

CMakeLists.txt
```diff
@@ -0,0 +1,969 @@
...
+    set( HAVE_READLINE ON )
+    include_directories( ${READLINE_INCLUDE_DIRS} )
+    link_libraries( ${READLINE_LIBRARIES} )
+  endif ()
+endif ()
+
+# Find GSL
+set( HAVE_GSL OFF )
+set( HAVE_GSL_1_2 OFF )
+set( HAVE_GSL_1_11 OFF )
```
heplesser: I wonder if we should still differentiate between these GSL versions. V 1.11 was released in 2008.

jougs: I agree with @heplesser and think we should use the opportunity to get rid of this distinction. This also brings some simplifications of `#ifdef`s in the code.

tammoippen:
I improved compilation on BG/Q with the last commits. Please see `INSTALL` for more detailed instructions. I can confirm building a working dynamic NEST and PyNEST with this on JUQUEEN.

@heplesser I saw the deprecation flag, but ignored it as it keeps the code cleaner. This [email](https://cmake.org/pipermail/cmake/2009-April/028439.html) clarifies the deprecation flag. Maybe I should go through the code and see, where which dependancy is actually used.

tammoippen:
@heplesser All `link_libraries` are replaced by corresponding `target_link_libraries` calls.

Further, I reorganized and formatted the CMake files according to the last meeting.

heplesser:
@tammoippen I tried to build `MyModule` and ran into some problems.

On first try, I did not specify which compiler to use when running `cmake` for mymodule, and got the following message
```
CMake Error at CMakeLists.txt:138 (message):
  NEST compiler and CMAKE compiler differ.  Please specify with
  `-DCMAKE_CXX_COMPILER=/usr/local/bin/g++-5` .
```
Couldn't CMake pick up the compiler information from `nest-config`? Might need additions to `nest-config`.

With the compiler flag in place, compilation went well, but CMake chose `/usr/local/llib/nest` as install target, not the NEST installation location, even though `$NEST_INSTALL_DIR/bin/nest-config --prefix` returns the correct prefix. I needed to specify the prefix explicitly with `-DCMAKE_INSTALL_PREFIX:PATH=...`

With that in place, everything works (OSX 10.10, g++-5 from Homebrew).

w-klijn:
So I did my review and the comments have been processed. I guess I should now give a :+1:

tammoippen:
@heplesser Commit 3ca304dd should solve the compiler selection: `CMAKE_CXX_COMPILER` must be set before the call to `project(...)`. I rearranged the call to `${NEST_CONFIG} --compiler` before the  `project(...)` and this solves it on my machine.

> With the compiler flag in place, compilation went well, but CMake chose `/usr/local/llib/nest` as install target, not the NEST installation location, even though `$NEST_INSTALL_DIR/bin/nest-config --prefix` returns the correct prefix. I needed to specify the prefix explicitly with `-DCMAKE_INSTALL_PREFIX:PATH=...`

I cannot reproduce this. Sometimes the build directory is confused from previous failed runs. Try to 

1. run `make rebuild_cache` or
1. remove the content of the build directory and run the cmake command again.

heplesser:
@tammoippen The compiler is now detected properly. I have also understood the problem with the prefix. The relevant code is
```
# If no `CMAKE_INSTALL_PREFIX` is given, set the NEST_PREFIX.
if ( "${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr/local" )
  # no prefix given?
  set( CMAKE_INSTALL_PREFIX "${NEST_PREFIX}" CACHE STRING "Install path prefix, prepended onto install directories." FORCE )
endif ()
```
Note that unless `CMAKE_INSTALL_PREFIX` is equal to `/usr/local`, the prefix will not be set to the value obtained from `nest_config`. This is the case on my system, and that is why I ran into problems. 

I would suggest that we just drop the check and unconditionally use the value in `NEST_PREFIX`, obtained from `nest_config`: The module shall be installed together with the NEST for which it is compiled.

tammoippen:
@heplesser Commit 87891f2 sets the install path unconditionally. In the developer space for [writing modules](https://nest.github.io/nest-simulator/extension_modules#building-mymodule), we should then also rewrite point 4. of **Building MyModule** in #234.

heplesser:
@tammoippen Excellent, now it works, so :+1:!

jougs:
I've just closed the related trac tickets 523, 768, 775, 791, 865, 923.

@tammoippen, can you please check the status of trac tickets 654, 781 and 904 and close them after creating follow-up tickets here with the remaining issues? Please also have a look at #152 and #56. Many thanks

Issue #214, opened 2016-01-29
hard coded decay of vacant elements in structural plasticity


jakobj:
in e.g., https://github.com/nest/nest-simulator/blob/ff2bdbb/nestkernel/simulation_manager.cpp#L489 the amount for the decay of vacant elements is hardcoded to be 10%. should this not be a parameter that the user can change? and if not, it would be better to define it as a global constant? @sdiazpier

sdiazpier:
Yes this is right @jakobj , I have already done the changes and will submit the PR asap.
Thanks for pointing it out.

PR #215, opened 2016-01-29
Changes to fix the hard coded decay of synaptic elements

sdiazpier:
This pull request addresses issue #214 , now allowing the user to set the amount using the tau_vacant variable in the synaptic elements class.

### DIFF
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -394,14 +394,11 @@ nest::Archiving_Node::update_synaptic_elements( double_t t )
 void
 nest::Archiving_Node::decay_synaptic_elements_vacant()
 {
-  double_t z, z_vacant;
   for ( std::map< Name, SynapticElement >::iterator it = synaptic_elements_map_.begin();
         it != synaptic_elements_map_.end();
         ++it )
   {
-    z = ( it->second ).get_z();
-    z_vacant = ( it->second ).get_z_vacant();
-    it->second.set_z( z - ( z_vacant * it->second.get_tau_vacant() ) );
+    it->second.decay_z_vacant();
   }
 }
 
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -392,7 +392,7 @@ nest::Archiving_Node::update_synaptic_elements( double_t t )
 }
 
 void
-nest::Archiving_Node::decay_synaptic_elements_vacant( double_t p )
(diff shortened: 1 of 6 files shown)


nestkernel/synaptic_element.cpp
```diff
@@ -47,7 +47,7 @@ nest::SynapticElement::SynapticElement()
   , z_connected_( 0 )
   , continuous_( true )
   , growth_rate_( 1.0 )
-  , tau_vacant_( 10.0 )
+  , tau_vacant_( 0.1 )
```
heplesser: @sdiazpier I am a bit surprised by the change of the default value of `tau_vacant_` from 10.0 to 0.1. This probably has no consequence, since the value was not used anywhere in the past, but could you just confirm that the change is intentional?

nestkernel/archiving_node.cpp
```diff
@@ -401,7 +401,7 @@ nest::Archiving_Node::decay_synaptic_elements_vacant( double_t p )
   {
     z = ( it->second ).get_z();
     z_vacant = ( it->second ).get_z_vacant();
-    it->second.set_z( z - ( z_vacant * p ) );
+    it->second.set_z( z - ( z_vacant * it->second.get_tau_vacant() ) );
   }
```
heplesser: @sdiazpier All quantities entering the operation in this loop are members of the `SynapticElement` object `it->second`. So I'd suggest to just call here `it->second.update_z()`, which should then look something like 
```C++
{ z_ -= get_z_vacant() * tau_vacant_; }
```
Or have I overlooked something?

heplesser:
I think @jakobj would be a good second reviewer.

nestkernel/synaptic_element.cpp
sdiazpier: Hi dear @heplesser, yes you are right, the change is intentional. This value was unused, but it was part of the original design. This change was done in order to adapt the value to the current implementation of the decay.

jakobj:
i agree with @heplesser's suggestion to introduce the `update_z()` function in order to avoid those getters, but otherwise i'm happy with the changes.

sdiazpier:
Dear @jakobj and @heplesser, thank you for your comments. I have done and submitted the changes you suggested. Please let me know if anything else is missing.

jakobj:
looks good to me :+1:

heplesser:
Thank you! :+1:

Issue #211, opened 2016-01-21
segmentation fault in connect with multiple threads


jakobj:
using multiple threads, when calling connect with a mixture of neuron and devices, a segmentation fault occurs as threads create connections in non-thread-local sparse tables.
the following lines will generate a segmentation fault:

``` postscript
0 << /total_num_virtual_procs 2 /print_time false >> SetStatus

/n1 /iaf_psc_delta Create def
/n2 /iaf_psc_delta Create def
/n3 /iaf_psc_delta Create def

/sd /spike_detector Create def

[n1 n1] [sd n3] Connect
```

in this case thread0 and thread1 both try to create a connection in the connector of neuron 1 on thread 1. this can be traced back to the use of  `get_node(gid, thread)` in the connection builders (`AllToAllBuilder` etc.) which defaults to thread=0 if no argument is given.
most likely this bug was not ecountered before since users rarely mix devices and neurons in a single list. the segmentation fault occurs with 2.10.0, 2.8.0, 2.6.0 and i did not go back further in time.

am i right in assuming this is not the desired behaviour?

ps: i am now working on a fix based on the nest3 branch.

jakobj:
there still seem to be old connection routines like DivergentConnect in the connection_builder_manager. i thought these were completely replaced by the new connect framework? does it make sense to fix them aswell as they exist for a good reason or should i ignore them? @heplesser @hannahbos

jougs:
We decided to do the restructuring first and then clean up for the release once everything is in master.

heplesser:
This has been fixed by #221, thus closing.

PR #221, opened 2016-02-02
fix211 connect segfault and remove deprecated connection routines

jakobj:
this PR fixes issue #211 by using the correct thread in calls to `get_node` in parallel sections of the `conn_builder.cpp` functions. due to these changes, connections between devices and neurons will be created by the thread on which the neuron is located, connections between devices will be created on the "suggested thread" of the device. a test is included. i would suggest @heplesser @hannahbos as reviewers.


heplesser:
@jakobj Thank's for addressing this quickly! I will review your code shortly. Right now, I have a few suggestions for the test: Since it is a regressiontest, it should go to the `regressiontest` directory and be called `issue-211.sli`. I am also not entirely sure that all cases are covered, e.g., globally receiving devices such as `volume_transmitter`. Could you add the missing cases? Finally, the code would probably be a bit more readable if you used the prefix `n` only for neurons proper, `g` for generators and `d` for detectors, while keeping the consecutive numbering.

jakobj:
thanks for the quick response. i am currently addressing the issues you mentioned.

jakobj:
the fixes introduced in this PR break deprecated connection routines (DivergentConnect, RandomConvergentConnect etc.). to avoid complications @tammoippen and me are working on removing these. these changes will also be part of this PR.

jakobj:
currently some of the examples still do not run/run extremely slow. we're currently looking into that.

jakobj:
examples are fixed now (i.e., they run), but results partly deviate from master. runtimes are also slightly different.

heplesser:
@jakobj I presume that the changes you made to the connection routines lead to different networks being generated with the new code. The probabilistic connectivity tests ensure that the connectivity is still statistically correct. Since the networks now have slightly different connectivity, slightly different results are plausible *for networks with randomized connectivity*.

I think that it still would make sense to check at least a few cases by hand to ensure that results are plausible.

jakobj:
here are some results from manually run examples:
`brunel-2000.sli`
_master_
> Brunel Network Simulation
> Number of Threads : 2
> Number of Neurons : 12500
>Number of Synapses: 15637600
>        Excitatory : 12512500
>        Inhibitory : 3125000
> Excitatory rate   : 32.36 Hz
> Inhibitory rate   : 32.72 Hz
> Building time     : 3.82 s
> Simulation time   : 49.48 s

_fix211_
> Brunel Network Simulation
> Number of Threads : 2
> Number of Neurons : 12500
> Number of Synapses: 15637600
>        Excitatory : 12512500
>        Inhibitory : 3125000
> Excitatory rate   : 32.36 Hz
> Inhibitory rate   : 32.72 Hz
> Building time     : 4 s
> Simulation time   : 50.67 s

`hpc_benchmark.sli`
_master_
> 0 869044 # virt_mem_after_sim
> 0 869044 # virt_mem_after_sim
> 0 461.46 # sim_time
> 0 461.46 # sim_time
> 0] Simulation time   : 461.46 s

_fix211_
> 0 870060 # virt_mem_after_sim
> 0 870060 # virt_mem_after_sim
> 0 512.49 # sim_time
> 0 512.49 # sim_time
> 0] Simulation time   : 512.49 s

for the hpc_benchmark, the rates are the same aswell.

jakobj:
`hpc_benchmark.sli` (4threads)
_master_
> 0 961004 # virt_mem_after_sim
> 0 961004 # virt_mem_after_sim
> 0 228.84 # sim_time
> 0 228.84 # sim_time
> 0] Simulation time   : 228.84 s

_fix211_
> 0 960696 # virt_mem_after_sim
> 0 960696 # virt_mem_after_sim
> 0 221.12 # sim_time
> 0 221.12 # sim_time
> 0] Simulation time   : 221.12 s

same rates in both cases.

heplesser:
@jakobj I am a bit surprise about the hpc_benchmark timings. Were the first times you posted for single thread? How many MPI processes? Do we have an explanation why the simulation time was significantly longer (+ 9%) in the first data you posted, but shorter in the second case?

jakobj:
the first times were for a single thread, single MPI process. the second times were 4 threads and also just a single process. i don't really have an explanation beyond whatever my operating system is causes since the activity is exactly the identical in master and fix211. however, the activity looks very pathological, with strongly correlated bursts of network activity intermingled with completely silent periods, so maybe not the best setting for a benchmark. i will run a scaled-up version on our cluster and report back once i have the data.

jakobj:
update: it seems like we have forgotten to update the manual tests, i will work on thes in parallel to the (still) missing benchmarks.

jakobj:
alright, here are some larger scale simulations of `hpc_benchmark.sli` (2 mpi procs x 24 threads = 48vps, scale=10, presimtime=100ms, simtime=200ms):

_master_
> 0] BuildNode time     : 0.31 s
> 0] BuildEdge time     : 703.14 s
> 0] Preparation time   : 92.93 s
> 0] Simulation time   : 206.25 s

_fix211_
> 0] BuildNode time     : 0.31 s
> 0] BuildEdge time     : 771.93 s
> 0] Preparation time   : 82.39 s
> 0] Simulation time   : 205.66 s

i would guess that this is within the expected fluctuations.
still waiting on the results of the `brunel-2000.sli` simulations.

jakobj:
`brunel-2000.sli` (adapted to new connect) results (2 mpi procs x 24 threads = 48vps, order=9000, simtime=500ms):

_master_
> Brunel Network Simulation
> Number of Threads : 24
> Number of Neurons : 45000
> Number of Synapses: 101272550
>        Excitatory : 81022500
>        Inhibitory : 20250000
> Excitatory rate   : 14.4 Hz
> Inhibitory rate   : 14.08 Hz
> Building time     : 67.12 s
> Simulation time   : 42.14 s

_fix211_
> Brunel Network Simulation
> Number of Threads : 24
> Number of Neurons : 45000
> Number of Synapses: 101272550
>        Excitatory : 81022500
>        Inhibitory : 20250000
> Excitatory rate   : 14.4 Hz
> Inhibitory rate   : 14.08 Hz
> Building time     : 62.88 s
> Simulation time   : 41.33 s

these results look like this PR does not do horrible things to NEST's performance.

heplesser:
@jakobj The benchmark data look good now, so from that perspective, the PR is fine.

@jakobj @jougs @hannahbos @tammoippen Could you check on the following, I hope I have not misunderstood anything here. 

As far as I can see, this PR removes the deprecated connection functions from `nest-init.sli`, but not from `hl_api_connections.py`. I suppose they should be removed from there as well, especially since the deprecated Python functions call the now-removed SLI functions.

I also noticed that the `Connect_i_i*_l` functions in NEST Module have grown considerably in complexity. These functions are only there to support `gid gid ... Connect` calls from the SLI level. I think it would make much more sense to remove those functions from the C++-level completely and just wrap them on the SLI level along the lines of
```
/Connect_i_i_l
{
  /synspec def
  dup cvgidcollection
  exch
  dup cvgidcollection
  exch
  /one_to_one
  synspec
  Connect
} def
```
That way, all connectivity would be created using `Connect()` on the C++ level. Performance would suffer a little, but since these functions will be used for individual connections only, that does not matter. I added `/one_to_one` above to avoid using the more complex `all_to_all` connection builder.

tammoippen:
> As far as I can see, this PR removes the deprecated connection functions from `nest-init.sli`, but not from `hl_api_connections.py`.

This is right. The PyNEST versions should be removed as well. I also agree on replacing the `/Connect_i_i*_l` functions as proposed.

jakobj:
i agree with @heplesser's suggestions. i will look into this as soon as possible. one thing i am wondering about: are the `Connect_i_i*_l` functions exclusively called when using SLI or do we also need to keep an eye on the Python interface?

heplesser:
@jakobj The `Connect_i_i*` functions are not used from the Python interface, nothing to worry about.
I think we should also remove `FindConnections`, i.e., remove everything  marked `@deprecated` from `hl_api_connections.py`.

jakobj:
As far as I can see, all deprecated functions are now removed and the relevant scripts updated. While working on this I found that the pynest example `plot_quantal_stp_synapse.py` and that quite a few manualtests are outdated and do not run any more. I did not edit these, since this is logically disconnected from this PR.

nestkernel/conn_builder.cpp
```diff
@@ -404,10 +404,23 @@ nest::ConnBuilder::single_connect_( index sgid,
...
       kernel().connection_builder_manager.connect(
         sgid, &target, target_thread, synapse_model_, delay_->value_double( target_thread, rng ) );
+    }
+    else if ( default_delay_ )
+    {
+      kernel().connection_builder_manager.connect( sgid,
+        &target,
+        target_thread,
+        synapse_model_,
+        NAN,
```
heplesser: This should be `numerics::nan`.

nestkernel/conn_builder.cpp
```diff
@@ -460,15 +473,29 @@ nest::ConnBuilder::single_connect_( index sgid,
...
         delay_->value_double( target_thread, rng ) );
+    }
+    else if ( default_delay_ )
+    {
+      kernel().connection_builder_manager.connect( sgid,
+        &target,
+        target_thread,
+        synapse_model_,
+        param_dicts_[ target_thread ],
+        NAN,
```
heplesser: This should be `numerics::nan`.

heplesser:
@jakobj This all looks very fine now, except for the explicit use of `NAN` from `cmath`, see inline comments. 

The root problem here is that `connection_builder_manager.{h,cpp}`, `connector_model.h` and `connector_model_impl.h` all contain comments pointing to `NAN` from `cmath` as value marking missing weight or delay, and `connection_builder_manager.h` also uses is in code. This should all be converted to `numerics::nan` to be consistent with the rest of NEST. Could you fix this?

With that in place, this PR should be ready for merging.

jakobj:
I needed to change the unittest for gap junctions, @janhahne should have a look whether I broke it.

testsuite/unittests/test_gap_junction.sli
```diff
@@ -82,7 +98,7 @@ neuron_no_gap neuron_no_gap Connect
 0 /max_delay get /old_max Set
 0 /min_delay get /old_min Set
 
-neuron_gap neuron_gap 4.0 4.0 /gap_junction Connect
+neuron_no_gap neuron_no_gap 2.0 2.0 Connect
```
janhahne: This has to be a gap-junction connection. 
Like for example `neuron_gap neuron_gap /gap_junction Connect`

Gap-junction connections contribute to the calculation of `min_delay` and `max_delay` with delay `wfr_comm`. The idea here is to check that this works properly.

heplesser:
@jakobj @janhahne The tests failed due to PEP-8 violations. Could one of you reformat the code?

testsuite/unittests/test_gap_junction.sli
heplesser: @janhahne Would `neuron_gap neuron_gap /gap_junction Connect` the proper code here? With the `4.0 4.0` in addition it must be incorrect, since one cannot set delay for gap junctions, or?

Note that you here have "naked" code in a test script (not under any `*_or_die` control). That is best avoided. Could you put it into a setup function and then call that from the individual tests? That ensures that tests are independent. If tests are naturally dependent, you can also test different things within a single `assert_or_die` and then just `and` the results of the different parts of the test.

jakobj: I agree with @heplesser on this part. we can not use gap junctions here as we explicitly check before that setting a delay is not possible for these. it seems to me like this code part is obsolete.

jakobj:
@heplesser since, as you already pointed out the test only fail due to PEP8, and not due to changes we made in this PR, would it make sense to merge it and then create a new PR in which we reformat _all_ Python files according to PEP8? this way we would keep reformating separate from "real" changes.

testsuite/unittests/test_gap_junction.sli
heplesser: @jakobj I think you understood me wrong, I did not consider the code obsolete. 

@janhahne I think things got messed up in the merging process. I don't fully understand what is to be tested here. My current interpretation of the code is that you first set `wfr_comm_interval` to 2.0 ms. Then you create one connection with default delay (1 ms), to provoke an update of min_delay and max_delay. This should result in min_delay = 1 ms and max_delay = 2 ms. Then you create another connection between normal neurons, with explicit delay, and check again min_delay and max_delay. For that second test to make sense, this last connect needs to use a delay smaller than 1 ms or larger than 2 ms, so the line should probably be
```
neuron_no_gap neuron_no_gap 1.0 4.0 Connect
```

heplesser:
@janhahne At the end of `testsuite/unittests/test_gap_junction.sli`, you test that `no_gap -> no_gap` and `no_gap -> gap` connections cannot be created when synapse type `/gap_junction` is requested. That is fine. But shouldn't there also be a similar test for `gap -> no_gap`?

janhahne:
@heplesser @jakobj This pull request adds a function `void set_delay( double_t )` to `gap_junction.h` that throws a `BadProperty( "gap_junction connection has no delay" );`. That was missing before and unfortunately I even used commands like `neuron_gap neuron_gap 2.0 2.0 /gap_junction Connect` in the original unittest to create gap-junction connection. With this new function this isn't possible anymore and that is the reason why the unittest needs to be adjusted. In general that is a good thing and important improvement, but to me it seems a bit off topic to this PR.

I created a pull request to Jakobs branch where I improved the unittest (removed the inter-dependencies, added the gap -> no_gap test, corrected the test that checks the setting of min- and max-delay). See https://github.com/janhahne/nest-simulator/blob/patch-1/testsuite/unittests/test_gap_junction.sli

testsuite/unittests/test_gap_junction.sli
janhahne: @heplesser Please see the improved version of the test. The original idea was to test, that gap-junction connections contribute to the calculation of the min- and max-delay with the value of `wfr_comm_interval`. Therefore I added two test cases in the new version:

In the first version the min- and max-delay is set to 3.0 ms and then a gap-junction connection is added. It is tested that the min-delay changes to 2.0 ms (the value of `wfr_comm_interval`) and that the max-delay stays unchanged.

In the second version the min- and max-delay is set to 1.0 ms and it is tested that the max-delay changes to 2.0 ms after the creation of the gap-junction connection and that the min-delay stays unchanged.

janhahne:
@heplesser @jakobj Just another thought: Wouldn't it be cleaner to just revert `gap_junction.h` and `test_gap_junction.sli` to the version of the master branch and exclude this changes from this PR. Then I could do another PR that deals with this issue. In this case just ignore my PR to your branch. What do you think?

jakobj:
@janhahne thanks for your PR. I agree that it's off topic, but since this whole PR turned out to be more involved than previously expected I don't think adapting the unittest is now the big problem. ;) if @heplesser agrees, I will merge your PR into my PR and then hopefully everything works fine.

heplesser:
@janhahne @jakobj I suggest we continue with this PR, rather than creating a new one for PR cleanliness, with @jakobj merging into his PR.  The PEP8 check currently fails, I think we need to merge #337 to master first, and then on into this PR.

heplesser:
@jakobj @janhahne Great work and 👍 from me.
@jougs Would you look it over as second reviewer?

jakobj:
@heplesser @jougs a grep just told me that i missed a few occurences of RandomConvergentConnect etc. Will update later this week, so please don't merge yet.

heplesser:
@jakobj I am through with almost all remaining stuff, grepping for `ergentConnect`, and will send you a PR towards your for soon.

@steffengraber What is the relation between `doc/quickref.html` and `extras/userdoc/md/quickref.md`?  Am I right to assume that the former will dissappear? I am asking here because outdated references to `RandomConvergentConnect` and friends appear in both.

heplesser:
@jakobj Just sent you a PR towards your fork.

jakobj:
@heplesser great! thanks a lot. I'll also have another look later today.

steffengraber:
Hi @heplesser
'extras/userdoc/md/quickref.md' is the template to create the website
and maybe a replacement for 'doc/quickref.html' in future.


Am 01.06.2016 um 13:11 schrieb Hans Ekkehard Plesser:
> @jakobj <https://github.com/jakobj> I am through with almost all
> remaining stuff, grepping for |ergentConnect|, and will send you a PR
> towards your for soon.
>
> @steffengraber <https://github.com/steffengraber> What is the relation
> between |doc/quickref.html| and |extras/userdoc/md/quickref.md|? Am I
> right to assume that the former will dissappear? I am asking here
> because outdated references to |RandomConvergentConnect| and friends
> appear in both.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/nest/nest-simulator/pull/221#issuecomment-222962283>, or
> mute the thread
> <https://github.com/notifications/unsubscribe/AB003ADIuTwbPgj39MtKsf7DdZGAhyGnks5qHWjcgaJpZM4HRe5x>.
>



--
Steffen Graber

Simulation Lab Neuroscience
Division HPC in Neuroscience         Phone:  +49 2461 61 8930
Jülich Supercomputing Centre         E-mail: s.graber@fz-juelich.de
Institute for Advanced Simulation
Forschungszentrum Jülich GmbH
http://www.fz-juelich.de/ias/jsc

Institute of Neuroscience and Medicine (INM-6)
Computational and Systems Neuroscience &
Institute for Advanced Simulation (IAS-6)
Theoretical Neuroscience
Jülich Research Centre and JARA
Jülich, Germany
http://www.csn.fz-juelich.de


------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------
Forschungszentrum Juelich GmbH
52425 Juelich
Sitz der Gesellschaft: Juelich
Eingetragen im Handelsregister des Amtsgerichts Dueren Nr. HR B 3498
Vorsitzender des Aufsichtsrats: MinDir Dr. Karl Eugen Huthmacher
Geschaeftsfuehrung: Prof. Dr.-Ing. Wolfgang Marquardt (Vorsitzender),
Karsten Beneke (stellv. Vorsitzender), Prof. Dr.-Ing. Harald Bolt,
Prof. Dr. Sebastian M. Schmidt
------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------

jakobj:
@heplesser it looks to me like the mymodule example is using a removed `divergent_connect`, could you have a look at that? apart from that, i would be happy if this PR could be merged. :)

heplesser:
@jakobj You are right, sorry that I overlook that one. Fixing this one will require a complete reimplementation of the StepConnect function in terms of the new ConnBuilder architecture. I will try to get that done within a few days.

heplesser:
@jakobj Just sent you a PR that should fix the `divergent_connect` in `MyModule`.

heplesser:
@jakobj Would you merge the latest changes from master so that Travis can have a proper look at the code?

jakobj:
Travis seems happy. so how about some thumbs up for this PR @jougs? per your request this also removes `FindConnections`, if that does not make you happy, what does? ;)

heplesser:
@jakobj Unfortunately, merging from master seems to have messed up MyModule again, so right now it is :-1:. I will try to fix this shortly.

heplesser:
@jakobj Sorry, my fault, I had gotten confused, with some assistance from Eclipse. All is fine from my side, :+1:.

pynest/nest/tests/test_onetooneconnect.py
```diff
@@ -30,79 +30,62 @@
...
-        self.assertEqual(weights, (2.0, 2.0))
-
-        # OneToOneConnect([pre], [post], [params])
+        # Connect([pre], [post], params)
         nest.ResetKernel()
         pre = nest.Create("iaf_neuron", 2)
         post = nest.Create("iaf_neuron", 2)
-        nest.OneToOneConnect(pre, post, [{"weight": 2.0}])
-        connections = nest.FindConnections(pre)
+        nest.Connect(pre, post, "one_to_one", syn_spec={"weight" : 2.0})
```
jougs: Please remove the space before the colon after "weight".

pynest/nest/tests/test_onetooneconnect.py
```diff
@@ -30,79 +30,62 @@
...
         self.assertEqual(weights, (2.0, 2.0))
 
-        # OneToOneConnect([pre], [post], [params, params])
+        # Connect([pre], [post], [params, params])
         nest.ResetKernel()
         pre = nest.Create("iaf_neuron", 2)
         post = nest.Create("iaf_neuron", 2)
-        nest.OneToOneConnect(pre, post, [{"weight": 2.0}, {"weight": 3.0}])
-        connections = nest.FindConnections(pre)
+        nest.Connect(pre, post, conn_spec={"rule": "one_to_one"}, syn_spec={"weight" : [2.0, 3.0]})
```
jougs: Please remove the space before the colon after "weight".

pynest/nest/tests/test_onetooneconnect.py
```diff
@@ -30,79 +30,62 @@
...
-        self.assertEqual(weights, (2.0, 2.0))
-
-        # OneToOneConnect([pre], [post], [w], [d])
+        # Connect([pre], [post], w, d)
         nest.ResetKernel()
         pre = nest.Create("iaf_neuron", 2)
         post = nest.Create("iaf_neuron", 2)
-        nest.OneToOneConnect(pre, post, (2.0, ), (2.0, ))
-        connections = nest.FindConnections(pre)
+        nest.Connect(pre, post, conn_spec={"rule": "one_to_one"}, syn_spec={"weight": 2.0, "delay": 2.0})
```
jougs: Please remove the space before the colon after "weight".

jougs:
Awesome work!

In addition to my comments above, I have another minor thing: please also remove `FindConnections` from `doc/quickref.html` and `extras/userdoc/md/documentation/quickref.md`. Or even better: grep for the names of all removed functions again to find remaining occurences, if there are any.

If my comments are addressed, I'm all :+1: and for merging! Thanks again for this big piece.

And for the future: stop me if I'm suggesting idiotic things like to add more code to an already large PR. This was really painful to review and it's only fair that I had to do it ;-)

jakobj:
@jougs thanks for the comments. let's see what Travis thinks of this.

jougs:
Apparently `unittests/test_inh_stdp_synapse.sli` is failing in all cases. The file came in through your merge of master into nest_3-fix211-connect_segfault. Can you please replace the use of `FindConnections` in that file by `GetConnections`? That should do the trick.

jakobj:
uuhh. yes, didn't check again after the last merge. should be fixed now.

heplesser:
Travis is happy, @jougs is happy (and all his requests are handled). Great job, @jakobj!

Merging!

Issue #203, opened 2016-01-08
Remark in iaf_neuron.h not applicable to model?


sanjayankur31:
https://github.com/nest/nest-simulator/blob/master/models/iaf_neuron.h#L99

refers to tau_syn_in/tau_syn_ex neither of which seem to be used in the model. Do these remarks apply to this model? Exactly the same remark is found in iaf_psc_exp and other models where these parameters are indeed used, and this is just making me wonder if there's something about the iaf_neuron model that I have missed.

heplesser:
@sanjayankur31 Thanks for pointing this out! `iaf_neuron` has just a single synaptic time constant, `tau_syn`, which applies to excitatory and inhibitory input spikes. The documentation in `iaf_neuron` is incorrect, probably due to a cut-and-paste mistake.

sanjayankur31:
@heplesser Ah! Thank you for confirming this. Should I remove the remark and open a pull request? (I know there's a code freeze in effect at the moment).

jougs:
Fixed by merge of #224.

PR #224, opened 2016-02-05
Fix remark. Fixes issue #203

sanjayankur31:


jougs:
In the NEST Core meeting on February 15, 2016, we decided to merge this without further review. @sanjayankur31 many thanks for fixing this.

Issue #114, opened 2015-09-28
Regression in NEST 2.8.0 for 'aeif_cond_exp' with Delta_T = 0


apdavison:
When Delta_T = 0, the membrane voltage should go to infinity as soon as "V_th" is reached ([Scholarpedia](http://www.scholarpedia.org/article/Adaptive_exponential_integrate-and-fire_model#Physiological_interpretation)). This is handled correctly in NEST 2.6.0, but in NEST 2.8.0 there is no spike until the voltage reaches "V_peak".

Example:

```
import matplotlib.pyplot as plt
import nest

neuron = nest.Create('aeif_cond_exp')
nest.SetStatus(neuron,
               dict(Delta_T=0.0, I_e=1000.0, V_th=-50.0, V_peak=-45.0))

recorder = nest.Create('multimeter')
nest.SetStatus(recorder, {'record_from': ['V_m'],
                          'interval': nest.GetKernelStatus('resolution')})
nest.Connect(recorder, neuron)

nest.Simulate(100.0)

data = nest.GetStatus(recorder, 'events')[0]
t = data['times']
vm = data['V_m']

plt.plot(t, vm)
plt.xlabel("Time (ms)")
plt.ylabel("V_m (mV)")
plt.title(nest.version())
plt.ylim(-70, -45)
plt.savefig("aeif_conf_exp_{}.png".format(nest.version().replace(" ", "_")))
```

![aeif_conf_exp_nest_2 6 0](https://cloud.githubusercontent.com/assets/260552/10142674/5f6d17ba-6614-11e5-8b13-d6d5df2da689.png)
![aeif_conf_exp_nest_2 8 0](https://cloud.githubusercontent.com/assets/260552/10142680/6508ed66-6614-11e5-9311-282225f3fa8e.png)

heplesser:
@apdavison Thanks for spotting and reporting this. I can confirm the observation.

I have traced it back to a change in how we handle excessively large exponents (svn r11833), where I employed a transformation that is mathematically, but not computationally, correct and did not take into account the possibility that `Delta_T == 0`.

This affects all `aeif_cond_*` models.

heplesser:
@gewaltig Could you take a look at this. I am not entirely certain what would be the best way to handle this case without adding `if ( Delta_T == 0 )` in several places, particularly since the adaptive stepsize algorithm needs to be able to handle whatever we do in the `dynamics()` function. Maybe we can solve this in conjunction with #35?

[...]

PR #229, opened 2016-02-17
More precise AEIF models

Silmathoron:
**Overview**
This is a proposition of additional models of the aeif type, adding post synaptic currents, interpolation and precise spiking. The precise spiking models used a modified version of the slice-ring buffer.

**Details**
Added models/devices:
1. In models:
    - aeif_cond/psc_alpha/exp_gridprecise
2. In precise:
    - aeif_cond/psc_alpha/exp_ps
    - iaf_psc_alpha_ps (this is to provide a comparison with an existing model and implement a separation between inhibitory and excitatory synapses)
    - slice_ring_buffer_new

Modified files:
  - iaf_psc_alpha_canon (added the synaptic current to the recordables)
  - nest_names (added dI_ex and dI_in)

**Objectives**
More than just adding these new models, I would like to propose a [new way of implementing neuronal models (or rather the update function) that would be systematic](https://github.com/Silmathoron/nest-models/blob/master/NewModelImplementation.pdf).

This is related to the  [NESTML](https://github.com/nest/nestml) project: this systematic approach would make the automatic generation of models much simpler because we would only have to select which functions are needed depending on a few options of the model (integration method, interpolation, precise spiking...).
The basic idea is to split the update function into a set of independent "subfunctions" that can be added or removed depending on what is necessary. Only interpolation and spiking have been turned into real functions here, but we can clearly separate the functions necessary for the interpolation from the rest (see [this file](https://github.com/Silmathoron/nest-models/blob/master/NewModelImplementation.pdf)).
Note that the models are also available as an extension working for NEST up to 2.10 on [another repository](https://github.com/Silmathoron/nest-models).

I would really like to have the opinion of the main developers on this matter to see whether we can use it in NESTML.


tammoippen:
@Silmathoron Thank you for the extensive contribution. To start the code review, we first have to satisfy TravisCI:

* Your test `unittests/test_grid_precise_iaf_aeif.py` fails.
* The copyright notice in some files is missing/wrong: `precise/aeif_cond_alpha_ps.h/cpp`, `precise/aeif_cond_exp_ps.h/cpp`, `precise/aeif_psc_alpha_ps.h/cpp`, `precise/aeif_psc_exp_ps.h/cpp`, in  `unittests/test_grid_precise_iaf_aeif.py` the first line `#!/usr/bin/env python` is not allowed. The copyright notice will always be compared to `doc/copyright_header.cpp/py`.
* The formatting in some files does not correspond to our [guidelines](https://nest.github.io/nest-simulator/coding_guidelines_c++#tooling). Please run the `clang-format-3.6` tool as described (best twice).

Silmathoron:
@tammoippen Thank you for the quick reply,
I think I corrected the tests:
- `regressiontests/ticket-686-positive-parameters.sli`
- `regressiontests/ticket-659-copyright.py`
However, if I read the travis log correctly, `regressiontests/ticket-77.sli` also fails, which I do not understand... do you know what could be wrong?
Moreover, I'm suprised that you said `unittests/test_grid_precise_iaf_aeif.py` fails since I saw `Running test 'unittests/test_grid_precise_iaf_aeif.py'... Success` in the [travis log file](https://s3.amazonaws.com/archive.travis-ci.org/jobs/109854692/log.txt).

tammoippen:
I looked into the TravisCI job [`xGSL=0 xMPI=0 xPYTHON=0`](https://travis-ci.org/nest/nest-simulator/jobs/109854685): It is clear, that the python test cannot succeed, when PyNEST is not build :smile:. The test is in the wrong location: can you put it in `<src>/pynest/nest/tests`.

Silmathoron:
@heplesser Sorry to bother, but I'm at a loss regarding your `ticket-77.sli` test: my models completely fail it but I cannot see where the difference with the older models is located... also for `psc` models the poisson noise is not sufficient to trigger spiking, so a modification of the weight might be needed.

Silmathoron:
@jougs @heplesser I saw a pull request about copyright transfer agreement pass, should I fill the form and send it to you before the PR can be merged?

testsuite/regressiontests/ticket-77.sli
```diff
@@ -107,6 +115,10 @@ M_ERROR setverbosity
                           % plastic synapses.
 
       /initial_weight 10.0 def
+      model higher_weight_list exch MemberQ
+      {
+        /initial_weight 80.0 def
+      } if
```
heplesser: @Silmathoron This is fine, but you could integrate it with the `/extra_params` dictionary by adding to that dictionary for each of the affected models 
```
   /aeif_psc_exp_ps << /initial_weight 80.0 >>
```
with the corresponding model names. Then you can put in the definition of `/syn_spec`
```
/weight 
   extra_params model known
   {
      extra_params model get /initial_weight known
      {
         extra_params model get /initial_weight get
      }
      {
         initial_weight 
      } ifelse
   }
   {
      initial_weight
   } ifelse
```
similar to  the setup for `/receptor_type`.

heplesser:
@Silmathoron This is a lot of code to review and I fear I won't have time to review it this week, but hope to get around to it during the coming week. 

And yes, before we can merge your contribution into NEST you need to send us a signed contributor license agreement. See the [NEST Developer Space](http://nest.github.io/nest-simulator/), towards the end of the page, for more information.

Silmathoron:
Ok, thanks for the comment, I'll modify the files accordingly and send you the contributor agreement next week.
Best

tammoippen:
@mdeger Can you please also look into this pull request? Thanks.

Silmathoron:
Hi! After the transfer to cmake, I'm trying to update the models... however, though ``cmake`` seems to go well, running ``make`` afterwards leads to the following error:

    /home/tfardet/Documents/NAS-PhD/NEST/nest/models/modelsmodule.cpp: Dans la fonction membre ‘virtual void nest::ModelsModule::init(SLIInterpreter*)’:
    /home/tfardet/Documents/NAS-PhD/NEST/nest/models/modelsmodule.cpp:300:47: erreur : ‘aeif_cond_alpha_gridprecise’ was not declared in this scope
       kernel().model_manager.register_node_model< aeif_cond_alpha_gridprecise >(
                                                   ^
    /home/tfardet/Documents/NAS-PhD/NEST/nest/models/modelsmodule.cpp:301:35: erreur : no matching function for call to ‘nest::ModelManager::register_node_model(const char [28])’
         "aeif_cond_alpha_gridprecise" );
                                       ^
    In file included from /home/tfardet/Documents/NAS-PhD/NEST/nest/models/modelsmodule.cpp:129:0:
    /home/tfardet/Documents/NAS-PhD/NEST/nest/nestkernel/model_manager_impl.h:42:1: note : candidate: template<class ModelT> nest::index nest::ModelManager::register_node_model(const Name&, bool)
     ModelManager::register_node_model( const Name& name, bool private_model )
     ^
    /home/tfardet/Documents/NAS-PhD/NEST/nest/nestkernel/model_manager_impl.h:42:1: note :   template argument deduction/substitution failed:
    /home/tfardet/Documents/NAS-PhD/NEST/nest/models/modelsmodule.cpp:301:35: erreur : patron de l'argument 1 est invalide
         "aeif_cond_alpha_gridprecise" );
                                       ^

I only added the models to the ``CMakeLists.txt`` inside the ``set( models_sources`` function (this is the only thing I modified, but it seems that it is not enough... I'm completely new to ``cmake``, though, so I don't see what's missing)
You can see the modified file on the [aeif-models branch](https://github.com/Silmathoron/nest-simulator/tree/aeif-models) of my fork.

tammoippen:
Hi @Silmathoron,

The GSL flags have changed: all `HAVE_GSL_X_XX` are replaced by `HAVE_GSL` and versions before GSL1.11 are rejected. Further, you have missed the `iaf_psc_alpha_ps.cpp/h` lines in `precise/CMAkeLists.txt`.

Silmathoron:
Thanks @tammoippen! This solved the first problem.

However, running ``make`` now fails with ``cython`` (I'm using version 0.24 with python 3.5)

    [ 99%] Built target nest
    warning: /home/tfardet/Documents/NAS-PhD/NEST/nest/pynest/pynestkernel.pyx:497:10: Unreachable code
    warning: /home/tfardet/Documents/NAS-PhD/NEST/nest/pynest/pynestkernel.pyx:497:10: Unreachable code
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    from cpython.ref cimport PyObject
    from cpython.object cimport Py_LT, Py_LE, Py_EQ, Py_NE, Py_GT, Py_GE
    cdef string SLI_TYPE_BOOL = "booltype"
                               ^
    ------------------------------------------------------------
    /home/tfardet/Documents/NAS-PhD/NEST/nest/pynest/pynestkernel.pyx:42:28: Unicode literals do not support coercion to C types other than Py_UNICODE/Py_UCS4 (for characters) or Py_UNICODE* (for strings).

This goes on for quite some time with the other ``cdef string``s.

This is only on my computer, though, Travis does not complain...

jougs:
I got the same Cython-errors on my machine. Can you please check if PR #301 also solves the problem for you?

Silmathoron:
@jougs : thanks for the fix, I'll test it on Monday (py35 is on my computer at the office only)

abigailm:
How does this PR handle the problem raised in #114 ?

Silmathoron:
It does not ;)
See [comment](https://github.com/nest/nest-simulator/pull/35#issuecomment-186309378).

heplesser:
Ok, my mistake. But the solution to #114 should also be applied to these models.

Silmathoron:
Er... sorry, which solution? I thought it was decided to leave it as it was... (I personally think that ``Delta_T = 0`` is not a biologically relevant parameter and that it should be provided by an ``aiaf`` neuron model and not included in ``aeif``)

heplesser:
I agree that `Delta_T == 0` is a very special case, but unfortunately, it is explicitly defined in the original paper and PyNN expects is to work. But the main issue is how to define the numerics of the AEIF-models correctly near the threshold. Strictly speaking, V_m never exceeds V_peak. But during integration from t to t+h, V_m may take on arbitrarily large values and this will affect w as well. Once a threshold crossing is detected, V_m is reset, but w remains "polluted" by the effect of the exponentially growing V_m.

But I think I have just realized how to solve this: In the dynamics function computing the derivative, we must use `min( y [ S::V_m ], node.P_.V_peak )` wherever `V_m` appears in expressions. I will fix this in the standard AEIF-model shortly.

Silmathoron:
Ok, I understand, but one the [advantages of the ``aeif`` model compared to the Izhikevich model](http://www.mitpressjournals.org/doi/full/10.1162/neco.2009.09-08-853) was that given the fast divergence, the effect on ``w`` is finite.

Regarding the new models, since I use interpolation to find the exact time of threshold crossing, this problem does not exist.

Silmathoron:
Ok... there's something wrong with Travis, but I don't know what it's trying to tell me...

heplesser:
@Silmathoron Looking at the Travis log, this seems to be due to a mix-up in filenames. In `models/CMakeLists.txt` you use the suffix `_gridprecise`, while actual file names use suffix `_gp`.

Silmathoron:
Ok, that was really dumb, sorry, I didn't go far enough up or down the log...

jougs:
@ingablundell and @DimitriPlotnikov will also have a look.

heplesser:
@ingablundell @DimitriPlotnikov @Silmathoron From a numerics point of view (and this relates to #114 even more than to this PR (and to the general problem of numerical instabilities in aeif models)), the essential problem is the `exp((V-Vth)/DeltaT)` term, which grows exponentially for `V > Vth`. This growth needs to be handled properly.

Current code bounds the argument of the exponential:
```
  const double_t exp_arg = ( V - node.P_.V_th ) / node.P_.Delta_T;
  const double_t MAX_EXP_ARG = 10.;
  const double_t I_spike = node.P_.Delta_T * std::exp( std::min( exp_arg, MAX_EXP_ARG ) );
```
This precise code was introduced with NEST 2.8, causing the regression reported in #114; I believe, though, that the old code in NEST 2.6 was rather less clean than the current code.

For `Delta_T == 0`, a spike should be fired as soon as `V >= V_th`, according to the original Brette-Gerstner paper. We need to avoid division by zero here.

Another problem is the following: According to the model definition proper, we always have `V < V_peak`, with default value `V_peak = 0 mV` (in NEST, 20 mV in the paper). But during numerical integration during a time step, `V` can take on far larger values, and these much larger values will then affect the integration of `w`, yielding incorrect values of `w`. And while `V` is reset to a fixed value and thus "forgets", `w` is not reset and will thus not forget.

One way to address this would be to bound the membrane potential during integration. In practice, I fear, this can only be done in the way of using `min(V, V_peak)` instead of `V` inside the right-hand side function. This does not limit the actual value of `V` computed for intermediate points. In experiments (see attached notebook) it seems to stabilize the integration, but also to lead to larger values of `V` reported at the end of intervals. I am not sure what the correct solution is here.

heplesser:
[aeif_experiments.ipynb.zip](https://github.com/nest/nest-simulator/files/277661/aeif_experiments.ipynb.zip)

Silmathoron:
@heplesser thanks for the summary and the test file!

**Regarding the bounds of V:** I agree that preventing divergence is indeed better regarding the minimal step size and the integration failure; however, when it comes to the behaviour of ``w``, the main source of error comes from the uncertainty on the spike time, not from ``V``'s divergence (see attached notebook), which is why I introduced these corrected models.
That being said, I have no problem with bounding ``V`` in addition to the first correction if it is deemed necessary...

**Regarding ``Delta_T = 0``:** this, however, I am still having a hard time understanding; the model is not designed to implement that (people wanting to use it are probably looking for an adaptive integrate and fire model) and since implementing it requires either numerical "tricks" or loosing efficiency, I really don't understand why we should do it...
It would seem much more reasonable to me to implement this adaptive integrate and fire and tell people to use it instead of the AdExp when they set ``Delta_T`` to zero...
Otherwise we could use something like a function pointer to switch between two different implementations.

[aeif_experiments_interpolate.ipynb.zip](https://github.com/nest/nest-simulator/files/277742/aeif_experiments_interpolate.ipynb.zip)

heplesser:
@Silmathoron Thanks for your notebook. I just realized I forgot the `-w` term in the rhs function for dw/dt. This becomes particularly important when running for a long period. I have posted a new version of my notebook above. 

Using your notebook with fixed dw/dt, I ran a longer simulation (10s). Spike times for bounded and non-bounded then differ from each other for a while. Another interesting effect is that for long times, maximum V values for the unbounded case are around -41 mV, while for the bounded case they sometimes go up to about -34 mV.

For further work on the notebook, I suggest that we create a notebook properly documenting AEIF numerics as `doc/model_details/aeif_models.ipynb`. We already have two notebooks on other issues there (we might even think about submitting to ReScience once we are done).

The interpolation seems very sensible. I wonder if linear interpolation is the best choice here, or whether one should use some form of exponential interpolation, given that the dynamics during the last time step is completely dominated by the exponential "explosion".

Concerning `Delta_T == 0`: Your function pointer idea is excellent! We could write a `aeif_cond_alpha_dynamics_DT0()` function in addition to the normal one. Then, in `init_buffers_()`, we set `B_.sys_.function` to the pertaining function. In `Parameters_::set()` need to make sure that `Delta_T` cannot be changed to/from 0 after the simulation has started, since that would require changing the RHS side function in `calibrate()`; that would also be possible, but ...

Silmathoron:
@heplesser thanks for the update and the correction! (I'll updated my notebook)

I'll propose and update with the function pointer and I'll make some comparison of the bounded and unbounded functions on large networks (but maybe not before this week-end).

However, I did not understand the following:

> Using your notebook **with fixed dw/dt**, I ran a longer simulation (10s). Spike times for bounded and non-bounded then differ from each other for a while. Another interesting effect is that for long times, **maximum V values** for the unbounded case are around -41 mV, while for the bounded case they sometimes go up to about -34 mV.

Could you elaborate on the emphasized parts?

heplesser:
@Silmathoron By "fixed dw/dt" I meant "using the correct code for dw/dt" (with -w in place), but I see that "fixed" is ambiguous here.

The figure below shows the last 500 ms of a 10s simulation, membrane potential only. Membrane potential for unbounded (blue) never exceeds -41 mV, for bounded (green) reaches -34 mV on every other spike. These are the values resulting from interpolation.

unknown

Silmathoron:
@heplesser regarding the last paragraph in your [previous](https://github.com/nest/nest-simulator/pull/229#issuecomment-221189343) to make sure ``Delta_T`` isn't switched between zero and nonzero, what am I supposed to check? From [this comment](https://github.com/nest/nest-simulator/blob/master/nestkernel/simulation_manager.h#L106) in ``simulation_manager``, I guess I cannot use ``has_been_simulated()``... when exactly is ``calibrate`` called for the first time, and which flag should I test?

Wouldn't it be better to set ``B_.sys_.function`` in calibrate? Your post seem to imply otherwise, but I don't understand what would be the problem...

heplesser:
@Silmathoron Thank you for exploring this! Concerning the bounding: Bounding the exponential argument was introduced just to keep the numerics stable. It is not based in the model equations. Since the numerics appear to remain stable even without bounding the exponential argument if we bound the membrane potential in a way that is consistent with the model equations, I think we should remove the explicit bound on the exponential argument. The bound on `V` implies a bound on the exponential argument (`(V_peak - V_th)/Delta_T`, makes 25.2 for default parameters, exp(25.2)=87.9e9). Unless Delta_T << 1, this should be stable; if not, users will need to live with it and consider the Delta_T=0 limit.

Considering `B_.sys_.function`: The basic logic behind model design in NEST is that anything in `B_` are buffers, i.e., internal variables that "survive" from one `Simulate` call to the next, since they contain implementation-dependent but lasting information, e.g., representation of spikes that have been technically delivered already but have not affected the neuron yet because their delay is not over yet. `V_` on the other hand contain helper variables that may change between `Simulate` calls, e.g., when a time constant is changed using `SetStatus()`. Method `init_buffers_()` is called upon the first `Simulate` call after a `ResetKernel` (or after simulation start), i.e., before a neuron is simulated for the first time. This method shall set up everything in `B_`. Method `calibrate_()` is called every tiem `Simulate` is called and should update everything in `V_`.

Maybe the best thing would actually be to move the `sys_.function` member from `B_` to `V_` and then update it in `calibrate_()`. Then, one would not have to introduce a protection against changing `Delta_T` to/from 0.

Silmathoron:
@heplesser regarding the bounding and the exponential, I think it would be a good thing to compare the behaviour with what is obtained through the other simulators since it induces a non-negligible change in the behaviour.
Are you familiar with PyNN? If so, do you know whether we could compare easily with BRIAN or Neuron?

Ok, thanks for the explanation; I guess moving ``sys.function`` to `V_`` seems reasonable...

heplesser:
@Silmathoron I have never really worked much with PyNN, but it should not be too difficult. We could also look at the source code of the other simulators to see how they implement the integration.

Silmathoron:
Ok, using PyNN to compare various simulators is far from easy... (their support for BRIAN does not work with python 3 and I there seem to be differences in the synaptic delays... plus I can't install NEURON on my computer, so if you have other simulators I could try, don't hesitate).
I'll ask on their mailing list for details.
By the way, just for fun: [AdExp and BRIAN](https://github.com/NeuralEnsemble/PyNN/issues/370), [and NEURON](https://github.com/NeuralEnsemble/PyNN/issues/266)

Unfortunately I'm afraid this means that we can't really compare the dynamics with anything... or could the scipy implementation be considered as a reference (since there is no clipping of the exponential)?

heplesser:
@Silmathoron Thanks for the research! In some way good to know that the other simulators are also struggling. I would dare say that few have thought very thoroughly about the numerics of the model. According to [ModelDB](https://senselab.med.yale.edu/ModelDB/showmodel.cshtml?model=153635), the data in the original 2005 paper was created with Brian, so in a way Brian should be considered the reference. But we are free to improve on that implementation if it should be suboptimal.

Btw, we have a [disk image with NEST 2.8.0, Neuron, Brian, PyNN and some more](http://www.nest-simulator.org/download/) available.

Silmathoron:
Ok, thanks, I will have a look!

In the meantime, here are the implementation of [AdExp in PyNN](https://github.com/NeuralEnsemble/PyNN/blob/master/pyNN/standardmodels/cells.py#L279) for [NEURON](https://github.com/NeuralEnsemble/PyNN/blob/master/pyNN/neuron/cells.py#L229) and [BRIAN](https://github.com/NeuralEnsemble/PyNN/blob/master/pyNN/brian/standardmodels/cells.py#L200) ([equation here](https://github.com/NeuralEnsemble/PyNN/blob/master/pyNN/brian/standardmodels/cells.py#L30)).

Silmathoron:
Okay, here is an update to continue the discussion regarding the behaviour of the AEIF model:
I implemented the new dynamics for ``Delta_T == 0`` by moving the ``gsl_odeiv_system`` struct to ``Variables``. I also created a temporary model bounding the membrane potential below ``V_peak`` instead of bounding the exponential as before.

Since the scipy solvers rely on thoroughly tested libraries, I would rather take them as reference. I implemented the interpolation solver to mimic what I did in the new NEST models with scipy.
Comparing the results, I see that:

- the usual bounding of the exponential's argument leads to behaviours that are very similar to the "exact" resolution of scipy
- bounding ``V`` instead of the exponential leads to a noticeably different behaviour

[compare_sp_nest.pdf](https://github.com/nest/nest-simulator/files/313370/compare_sp_nest.pdf)
[compare_sp_nest_dt0.pdf](https://github.com/nest/nest-simulator/files/313371/compare_sp_nest_dt0.pdf)

Here is the [updated notebook](https://github.com/nest/nest-simulator/files/313372/aeif_experiments_interpolate.ipynb.zip)

As a conclusion, though it is true that the current implementation of aEIF suffers from instabilities, it looks like it is the one that follows most closely what I would consider the reference behaviour...

heplesser:
@Silmathoron Thank you for investigating! I will try to look at it later this week.

Silmathoron:
Ok, I do not understand what happens inside the solver, but know I know where the difference between the bounded and unbounded methods come from:

- when unbounded, the value of ``V`` does not exceed ``V_peak`` by much (it went up to 17 in my tests) so the interpolation finds a threshold crossing around 70% of the timestep, meaning that ``w`` decreases for only 30% of a timestep
- when bounded, ``V`` reaches 1e8... so threshold crossing is at 0% of the timestep, which means that ``w`` decreases for a whole timestep

This is why the two models diverge, roughly one timestep at a time, after each spike. I must say the divergence of ``V`` does not look very natural to me... and I don't know why this occurs in the solver.

heplesser:
I found some posts concerning event detection (=threshold crossing), that may be relevant for us, although they don't have a perfect solution. Unfortunately, SciPy only wraps LSODA, not LSODAR from ODEPACK, which can handle threshold conditions:

- http://scipy-user.10969.n7.nabble.com/Zero-crossing-detection-td16233.html
- http://scipy-user.scipy.narkive.com/nwBCNM9a/integrate-odeint-and-event-handling
- https://github.com/scipy/scipy/issues/1696

The posts mention the [PyDSTool](http://www2.gsu.edu/~matrhc/PyDSTool.htm) and [PySUNDIALS](http://pysundials.sourceforge.net) as useful tools with more advanced integrators than SciPy. We might want to look at them to get to a definite reference solution.

abigailm:
@heplesser, what needs to happen now? Does the code have your approval or are you suggesting more revisions? Who would you suggest as a second reviewer?

heplesser:
We still need to understand the mathematics in detail and investigate solvers with threshold detection, see the links above.

Silmathoron:
Ok, [here is the updated notebook](https://github.com/nest/nest-simulator/files/422431/aeif_experiments_17082016.ipynb.zip) including the LSODAR implementation of the adexp model.
I also added tests with a new adexp model, ``"aeif_cond_alpha_ugp"``, which you can find on the [``test-aeif-dyn`` branch](https://github.com/Silmathoron/nest-simulator/tree/test-aeif-dyn). It bounds ``V`` to be smaller than ``V_peak`` instead of bounding the argument of the exponential.
As you can see, the LSODAR implementation is pretty neat, so I guess this can be our objective for the NEST implementation.
I think an important par is the last comparison in the notebook, where we can see that the V-bounded model converges towards the LSODAR solution as the resolution goes down, whereas the elder model does not!

heplesser:
@Silmathoron Thanks a lot, I think this is a very useful step forward! We now seem to have a reference solution using LSODAR and a NEST implementation that converges towards that solution. I will try to take a look at your code in detail within a few days.

heplesser:
@Silmathoron I think you forgot to add the source files for the `_ugp` model variant to the repo.

heplesser:
Publications on ODE solvers with threshold detection (not all may be equally relevant):

- http://www.radford.edu/~thompson/RP/eventlocation.pdf ([papers citing this one](https://scholar.google.de/scholar?client=safari&rls=en&oe=UTF-8&gfe_rd=cr&um=1&ie=UTF-8&lr&cites=4605889203153043867))
- http://www.sciencedirect.com/science/article/pii/S0377042712000684
- http://www.radford.edu/~thompson/RP/rootfinding.pdf
- https://computation.llnl.gov/casc/nsde/pubs/u88007.pdf
- http://www.cs.ucsb.edu/~cse/Files/SCE000136.pdf
- http://www.sciencedirect.com/science/article/pii/0377042789903348
- http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.455.2976&rep=rep1&type=pdf
- https://theses.lib.vt.edu/theses/available/etd-12092002-105032/unrestricted/etd.pdf

The original reference for LSODAR is 
```
Implicitly defined output points for solutions of ODEs
Document ID: SAND-80-0180
Research Org: Sandia Labs., Albuquerque, NM (USA);
Authors: Hiebert, K.L.; Shampine, L.F.;
Publication Date: 1980-02-01
```
which is [available here](https://ntrl.ntis.gov/NTRL/dashboard/searchResults.xhtml?searchQuery=sand-80-0180) (terms and conditions apply).

heplesser:
@Silmathoron I have now done some convergence tests for LSODAR by changing atol from 1e-1 to 1e-10. Results look very reassuring. I think we can safely use LSODAR with atol=1e-8 as reference solution. For details, see the end of the revised notebook.

One question: Why are the arrays you use in the LSODAR code to collect spike times, voltage and w at spikes class members and not instance members?

[aeif_experiments_25082016.ipynb.zip](https://github.com/nest/nest-simulator/files/436759/aeif_experiments_25082016.ipynb.zip)

Silmathoron:
@heplesser: sorry for the late reply, I took a week of holiday and I have little to no internet connection.

I indeed forgot the ugp model, I'll update next week.

As for the class attribute, it's dirty code, I was lazy and did not take time to make it cleaner since I expected the class would not be used much...

It's a good thing you confirmed the reliability of the LSODAR solution, thanks a lot!

Silmathoron:
Ok, I updated the [``test-aeif-dyn`` branch](https://github.com/Silmathoron/nest-simulator/tree/test-aeif-dyn) with the new implementation (I renamed it ``aeif_cond_alpha_vbound_gp`` temporarily to be more coherent with what we said in the notebooks; we can look for better names later).
I'm still not completely satisfied by the fact that the model converges only for very small values of the timestep (namely 0.001 ms)... I wonder whether we could not use other solvers inside NEST to make it better...
Because of this I had a look at the other solvers (steppers) in [GSL](https://www.gnu.org/software/gsl/manual/html_node/Stepping-Functions.html#Stepping-Functions) and I was quite surprised to see that the new ``odeiv2`` solvers do not seem to work with the implementation (I get overflow errors or the script never seem to terminate); I would be glad if someone could have a look at it to confirm the problem: all there is to do is replace ``odeiv`` by ``odeiv2`` everywhere.

heplesser:
@Silmathoron I agree that it would be good if we could reach convergence for larger simulation resolution in NEST. The Adams method might help, I believe it is more suitable for stiff problems. It might, though, be a problem that only can be solved by implementing an integrator with built-in threshold detection. I took a very brief look at the LSODAR Fortran code—learning anything from that code is non-trivial.

I am surprised that GSL ODEIV2 is causing problems. Which script did you work with that ran into these problems?

heplesser:
@Silmathoron In `slice_ring_buffer_new.cpp`, line 42 the `ulong` must be `unsigned long`.

Silmathoron:
@heplesser Thanks for ``ulong`` correction!
I tested ODEIV2 with the notebook's script, just changing everything in the new ``vbound_gp`` model...
If you can tell me whether it works on your machine with ODEIV2, I'll try to check with the other solvers.

As for a custom implementation, [one of the papers you found]() mentionned DASSL, RADAU IIA and DOPRI5 as alternatives to LSODAR so I'll try to see whether we can access and use those source codes.

Silmathoron:
regarding the convergence towards LSODAR, I realized that using the V-bound implementation without interpolation, the trajectory is closer to LSODAR than with the older method (bounding the exponential argument) and still converges. In fact, because we change the value of ``V``, the interpolation does not behave as it should, so this is the problem I need to address now.
**Otherwise I think we can safely replace all the currently implemented AEIF models with the new dynamics.**

heplesser:
@Silmathoron Thank you for your update! I am quite sure DoPri is available from GSL, but obviously without built-in threshold checks. I agree with your suggestion to use the V-bounded approach throughout. 
Interpolation is indeed problematic. By bounding V, we discard information: When we get to the end of a time step (i.e., when the GSL solver returns), we cannot know at what point during the time step V hit the boundary.

DimitriPlotnikov:
Dear @Silmathoron and @heplesser,
your discussion make really valuable contribution to understanding of underlying problems of adaptive stepsize integrators.

However, the code of this pull commit was built for the last time 5 months ago. Therefore, I'm not sure that the place where the discussion takes place is really appropriate. On the one hand it is hardly possible for any of reviewers to get through all 55 messages. On the other hand all the great findings are lost after the commit is closed.

Therefore:
a) Is the discussion is finished the the code review (state of 5 months ago) can start?
b) Can you somehow document it: blog, paper, or something else?

PS
Don't get me wrong. I really appreciate your work. It would be pity to lose it.

Silmathoron:
Hi @DimitriPlotnikov!
Don't worry, I understand. In fact I was thinking about closing this PR since I need to find a new way of providing interpolation for the AEIF model now that the implementation is about to change (cf. [PR474](https://github.com/nest/nest-simulator/pull/474)).
@heplesser [mentionned](https://github.com/nest/nest-simulator/pull/229#issuecomment-221189343) including a notebook in ``doc/model-details`` to keep the information in NEST so I will provide it.

In the meantime, though I would like to provide NEST with AEIF models using postsynaptic currents. So should I add them in PR474 or should I start yet another PR?
On the same idea, if I manage to find a way to implement interpolation for the AEIF models (which I think is still possible), should I add them separately like I did with ``_gridprecise`` in this PR or should I modify the existing model to provide an ``interpolation`` option in the neuron's dictionary?

Cheers

PR #232, opened 2016-02-17
bug fix for icpc compiler

lepmik:


tammoippen:
Good catch. :+1: from my side.

jougs:
Nice. :+1: and merging.

PR #236, opened 2016-02-19
Replacing internal C++ variable U0_ by E_L.

jougs:
Most neuron models use the internal variable `U0_` for the resting potential. The user interface uses the name `E_L` to conform to our policy of using Dyan&Abbot terminology wherever possible.

This is confusing and this pull request replaces `U0_` by `E_L_` and thus makes code, comments and UI wording consistent.

I suggest @heplesser and @tammoippen as reviewers.

### DIFF
--- a/examples/nest/brunel-2000.sli
+++ b/examples/nest/brunel-2000.sli
@@ -144,7 +144,7 @@ def
 /CMem    1.0 def    % membrane capacity [pF]
 /tauSyn  0.5 def    % synaptic time constant [ms]
 /tauRef  2.0 def    % refractory time [ms]
-/U0      0.0 def    % resting potential [mV]
+/E_L     0.0 def    % resting potential [mV]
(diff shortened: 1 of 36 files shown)


tammoippen:
:+1:  from my side! Nice. Do you think, we need to adjust the examples as well? There `U0` is used a lot as a local variable.

jougs:
Yes, probably we should also change the examples. I'll look into this.

heplesser:
@jougs This seems to make a lot of sense. I trust no U0 is left in the code now? Have you run all the examples to make sure that everything still works?

jougs:
@heplesser: I can confirm that U0 is gone from NEST completely. I've also run all examples using the `run_examples.sh` script provided in the `examples` directory and did not observe any problems.

heplesser:
:+1: then for a U0-free world! Any U2 in NEST ;)?

PR #251, opened 2016-03-02
Improves documentation for regexec function, solving trac.560.

heplesser:
This PR improves the documentation of the regexec function in the regex SLI module. This should fix #560. The regex modules is used in a few places internally (SLI file system functions, unittests) and works fine, so I see no need for more work on that module.

@jougs Could you take a look at this?


heplesser:
trac.560 should be closes as fixed once this PR is merged.

jougs:
This looks good to me. I'm immediately merging and closing the old trac issue. Thanks for taking care!

PR #253, opened 2016-03-02
Added vim syntax file for SLI

sanjayankur31:
Also reorganised directories as @heplesser suggested on the ML


tammoippen:
Hi @sanjayankur31,

You changed the paths of the `sli.el.in` file, but have not changed it in the build system. Please also change `<src>/Makefile.am:70,71` to:

    	    $(DESTDIR)@PKGDATADIR@/extras/EditorSupport/emacs/sli.el.in > $(DESTDIR)@PKGDATADIR@/extras/EditorSupport/emacs/sli.el
	rm $(DESTDIR)@PKGDATADIR@/extras/EditorSupport/emacs/sli.el.in

jougs:
@weidel-p: could you please test this and review the code?
@sanjayankur31: more out of interest: is this based on a vim postscript mode as our emacs mode is, or did you write it from scratch?

Thanks!

sanjayankur31:
> @sanjayankur31: more out of interest: is this based on a vim postscript mode as our emacs mode is, or did you write it from scratch?

@jougs : I wrote it from scratch. I'm really new to vim scripting - not even sure if there is a "postscript" mode. (The script is extremely basic at the moment)

weidel-p:
Hi @sanjayankur31,

SLI files look really nice now in vim, but I have some minor remarks:

1. The list of keywords need some cleanup. For example models and tests shouldn't be defined as keywords because they are actually not part of the syntax.
2. The list of constants (M_INFO, M_ERROR, ..) is not complete (see [http://www.nest-simulator.org/sli/setverbosity/](http://www.nest-simulator.org/sli/setverbosity/))
3. There are unresolved TODOs in the code, maybe it's worth to resolve them first.

sanjayankur31:
Hey @weidel-p  - I left the model names in there so that they're also highlighted - they're pretty much constants that Nest defines - what do you think? (I'll remove the tests, I thought I'd removed them already).

I'll work on the other bits and push new commits.

heplesser:
@sanjayankur31 I think that keeping the models in is problematic. The set of models provided by NEST will change over time. If adding a neuron or synapse model will require changes to the vim style file, this will diverge. Furthermore, users are increasingly writing their own NEST modules that supply models, and those models would not be marked as keywords either. To handle this properly, vim would have to run NEST, import all modules loaded into a NEST script and dynamically update the keyword list.

Concerning the units, I would add at least `Hz`, and there may be a few more.

sanjayankur31:
Ah, I don't think vim can do that in a syntax file - that's something that autocompletion plug-ins do, which will need to be developed separately. I'll look into it, though.

jougs:
Commonly used units are defined here: https://github.com/nest/nest-simulator/blob/master/lib/sli/nest-init.sli#L1401

sanjayankur31:
Need to rebase this. Working on it.

tammoippen:
@weidel-p Can you have another look into this PR, please? Thank you.

extras/EditorSupport/vim/syntax/sli.vim
```diff
@@ -0,0 +1,73 @@
...
...etprecision SetStatus_dict SetStatus SetStatus_v SetSynapseStatus setverbosity SFWdumpparameters ShowDefaults ShowOptions ShowStatus shpawn shrink signaldict Sign Simulate sin sinusoidal_gamma_generator sinusoidal_poisson_generator size sleep_i sleep SLIFunctionWrapper SliToMathematicaIndex Sort spawn spike_detector spike_dilutor spike_generator spikes spin_detector Split spoon sqr sqrt stack StandardDeviation start statusdict step_current_generator StimulatingDevice stopped stop StringQ str subnet SubsetQ sub switchdefault switch symbol symbol_s synapsedict SyncProcesses sysexec system systemtime Table Take taskset taskset_thisjob TensorRank tic TimeCommunicationAlltoall TimeCommunicationAlltoall TimeCommunicationAlltoallv TimeCommunicationAlltoallv TimeCommunicationOffgrid TimeCommunication TimeCommunication TimeCommunicationv time Times tmpnam toc token_is token token_s ToLowercase ToMathematicaExpression topinfo_d Total ToUppercase Transpose trie trieinfo trim trunc typebind typeinfo type typestack undef unit_conversion UnitStep usertime using validate values Variance variant verbosity voltmeter volume_transmitter waitPID wait welcome which who whos writePGM xifstream xor zeros
```
weidel-p: I think it doesn't change much, but do you want to keep the devices (like `spike_detector` in the list of keywords?

extras/EditorSupport/vim/syntax/sli.vim
```diff
@@ -0,0 +1,73 @@
...
+hi def link sliKeyword     Keyword
+
+syn keyword sliConstant M_INFO M_ERROR M_DEBUG M_WARNING M_STATUS M_FATAL
+hi def link sliConstant    Constant
+
+syn match sliUnit '\smV'
+syn match sliUnit '\spF'
+syn match sliUnit '\spA'
+syn match sliUnit '\sms'
+syn match sliUnit '\ss'
```
weidel-p: In some cases this line causes a wrong highlighting. For example, the in command `M_ERROR setverbosity` the `s` is highlighted as unit. You could add a `space` after the declaration of the units to prevent that to happen.

After you change that, :+1: from me. Thanks for this work, I already used it several times!

extras/EditorSupport/vim/syntax/sli.vim
sanjayankur31: I removed them after @heplesser 's remark that things from the API such as models may change in the future, and if we haven't a mechanism in place that will automatically update the list in sli.vim, we'd rather leave them out. I'm still looking at whether there's a way to autogenerate a list of keywords, but I think that requires a more complex plugin like YCM or something. I'm sure it can be done, I just haven't enough vim-fu to know how XD

extras/EditorSupport/vim/syntax/sli.vim
sanjayankur31: Looking into this - maybe I can add word boundaries or something to restrict the match.

sanjayankur31:
@weidel-p - this should work, please have a look when you have the time. Cheers.

extras/EditorSupport/vim/syntax/sli.vim
weidel-p: I totally agree with you. But there are still a lot of devices in the list of keywords. For example, there is the `ac_generator`, `dc_generator`, `poisson_generator`, `spike_detector`, `spike_generator`. `voltmeter`, `volume_transmitter`, `multimeter`, `music_cont_in_proxy`, ...

Please take a look again

sanjayankur31: I think I've removed all the devices and models now - hopefully.

weidel-p:
:+1:

PR #255, opened 2016-03-03
Fix80 check multiplicity handling

### LINKED ISSUE #80: Add test that neuron models heed event multiplicity
Events, in particular `SpikeEvent`s can include a multiplicity: one event object represents several spikes emitted by a single sender in a single time step. The `handle()` method of the receiving neuron must read out this information and apply it, typically by multiplying the weight with the multiplicity. I think this only makes sense for `SpikeEvent`s, for all other event types this should lead to an error.

We currently have no test checking that all models heed multiplicity information. Add a test for this.

heplesser:
This PR allows spike_generator to emit spikes with multiplicity, which is essential for testing multiplicity-handling in neurons. It also adds a test to check that (almost) all neurons handle multiplicity of incoming spikes properly. This should fix #80.


heplesser:
@jougs @tammoippen Would you review?

models/spike_generator.h
```diff
@@ -153,7 +153,9 @@ namespace nest
        stop           double - earliest time stamp of a potential spike event that is not emitted in
   ms
        spike_times    double array - spike-times in ms
-       spike_weights  double array - corrsponding spike-weights, the unit depends on the receiver
+       spike_weights  double array - corresponding spike-weights, the unit depends on the receiver
+       spike_multiplicities  int array - multiplicities of spikes, same length as spike_times;
+  mostly for debugging
```
tammoippen: The comment formatting seems off here.

tammoippen:
Very nice. :+1:  from my side.

models/spike_generator.h
```diff
@@ -148,15 +148,18 @@ namespace nest
...
-  ms
-       spike_times    double array - spike-times in ms
-       spike_weights  double array - corrsponding spike-weights, the unit depends on the receiver
-       precise_times        bool - see above
-       allow_offgrid_spikes bool - see above
-       shift_now_spikes     bool - see above
+       origin                double - Time origin for device timer in ms
+       start                 double - earliest possible time stamp of a spike to be emitted in ms
+       stop                  double - earliest time stamp of a potential spike event that is not
+  emitted in ms
```
abigailm: Are comments supposed to wrap like this, or should they be indented after the break?

heplesser:
@abigailm I merged the formatting changes from the master branch now and fixed the comment formatting. I hope you are now happy with the PR.

PR #258, opened 2016-03-04
BugFix: RandomFunction

tammoippen:
During refactoring I was not aware, that the `librandom::randomFunction` returns either a double or a long to the sli stack, depending on the RdvDatum argument. Now the API encapsulates the returned random value in a Token.

Additionally, the `RandomArray` was encapsulated twice in a `ArrayDatum`. Fixed it here.

### DIFF
--- a/librandom/randomdev.cpp
+++ b/librandom/randomdev.cpp
@@ -31,7 +31,8 @@ long librandom::RandomDev::ldev( RngPtr ) const
   return 0;
 }
 
-void librandom::RandomDev::get_status( DictionaryDatum& dict ) const
+void
+librandom::RandomDev::get_status( DictionaryDatum& dict ) const
 {
   def< bool >( dict, "is_discrete", has_ldev() );
 }
--- a/librandom/randomdev.cpp
+++ b/librandom/randomdev.cpp
@@ -22,8 +22,16 @@
 
 #include "randomdev.h"
 
+// Includes from sli:
+#include "dictutils.h"
+
 long librandom::RandomDev::ldev( RngPtr ) const
 {
   assert( false );
   return 0;
+}
+
+void librandom::RandomDev::get_status( DictionaryDatum& dict ) const
+{
+  def< bool >( dict, "is_discrete", has_ldev() );
 }
--- a/testsuite/unittests/test_rdv_discreteness.sli
+++ b/testsuite/unittests/test_rdv_discreteness.sli
@@ -38,46 +38,41 @@ FirstVersion: 160304
 (unittest) run
 /unittest using
 
-/skip_list [ /gsl_binomial ] def
-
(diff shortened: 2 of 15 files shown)


heplesser:
@tammoippen If I understand the history right, this was introduced by 036b8099994bb2ba7fd02f08cfd17899be7554f2 and does not affect nest-2.10 and previous releases. Can you confirm this?

Your fix seems sensible, but could you add a regression test, along the lines of
```
rngdict /MT19937 get 123456789 CreateRNG /rng Set
rng rdevdict /normal get CreateRDV /norm Set
rng rdevdict /binomial get CreateRDV << /p 0.2 /n 10 >> SetStatus /bino Set

norm Random DoubleQ exch ;  -> true
bino Random IntegerQ exch ;  -> true

norm 10 RandomArray 
true exch { DoubleQ exch ; and } Fold   -> true

bino 10 RandomArray 
true exch { IntegerQ exch ; and } Fold   -> true
```

tammoippen:
@heplesser confirm! I tried to get the spike trains from `hpc_benchmark.sli` equal to those in nest 2.10 and found this error. 2.10 and before is not affected.

Also `RandomArray` is not affected. I am working on such a test.

testsuite/unittests/test_rdv_discreteness.sli
```diff
@@ -0,0 +1,83 @@
...
+for Random and RandomArray.
+
+Author: Ippen
+FirstVersion: 160304
+*/
+
+(unittest) run
+/unittest using
+
+/skip_list [ /gsl_binomial ] def
```
heplesser: Why are you excluding the `gsl_binomial` generator? If it is necessary to exclude it, you should comment why.

I think the main problem is that the `gsl_binomial` generator requires a RNG from the GSL. But couldn't that be handled by something like

```
rngdict /gsl_knuthran known { /gsl_knuthran } { /knuthlfg } ifelse
rngdict exch get 123456789 CreateRNG /rng Set
```
Then, we would use a GSL RNG when available, and a built-in one otherwise.

librandom/randomdev.cpp
```diff
@@ -22,8 +22,17 @@
 
 #include "randomdev.h"
 
+// Includes from sli:
```
jougs: Can you please remove this comment? I don't see why this matters.

tammoippen: This is a convention I introduced with organizing the includes: all includes are grouped by 'module', aka the folder in which they live and alphabetically. Each 'module' gets a comment `// Includes from ...` .

jougs: OK. Then I agree that it is a good idea to have it.

jougs:
Very nice. :+1: once Travis passes.

heplesser:
:+1: from me as well conditional on Travis.

Issue #257, opened 2016-03-03
Autogenerate PyNEST reference documentation


flinz:
Currently the Python bindings for NEST are not really well documented. There is the [pynest introduction](http://www.nest-simulator.org/introduction-to-pynest/) and many example, but I personally often find myself lacking an API reference. Further, changes in the API are not automatically reflected in the pynest introduction.

Since most of the API is documented, an easy solution would be to autogenerate a reference documentation, to which we could also link from the introduction for further reference on function calls etc.

Setting this up would involve some of the below:
- choose a build system (sphinx-doc.org is a common choice)
- rewrite/format the python docstrings to adhere to a standard that the build system can format ([google style](http://google.github.io/styleguide/pyguide.html#Comments) for example)
- add hooks to the repository that include automatic rebuilding and deployment of the documentation 

I've tried around a little with Nest 2.10, and sphinx, the results of which you can find here: http://flinz.github.io/nest-simulator-apidoc

Is this something that is generally desired? I could look into reformatting the docstrings, which would probably also go hand in hand with https://github.com/nest/nest-simulator/issues/188

steffengraber:
Hi @flinz 
I am currently working on a new build system for the documentation. For the python part the decision is to use docstrings according to the NumPy/SciPy documentation standard.
From my side reformatting the docstrings would be helpful.

flinz:
@steffengraber nice - do you have a little more details on what you're doing?

Regarding the docstring reformatting: I will do that.

------------

**Here is a little sidenote on autogeneration of the python documentation, maybe its helpful if you're working on this**

PyNEST widely uses decorators and thus the autogenerated docs showed all decorated functions with the universal signature ``function(*args, **kwargs)``. 

Since this is not really informative about possible arguments, I introduced a little workaround for the decorators that disables them if a documentation build imports nest: [code is available here](https://github.com/flinz/nest-simulator/commit/56ad76686e6c7888cf81a5c891df4e7887bc4f6d). This was used to build the docs at http://flinz.github.io/nest-simulator-apidoc.

[...]

Issue #188, opened 2015-12-10
Python code should be checked against PEP8


heplesser:
We now have powerful and useful code-style checking for our C++ code, but we do not check the style of our Python code. We should add PEP8-conformity checks to our code-style checker.

apeyser:
Seems reasonable to me, much as I find PEP8 to be filled with nonsense, standardization has a lot of value (even when wrong). Shouldn't this become a pull?

tammoippen:
Thank you! :smile: On it.

flinz:
@apeyser Just in general, I would be interested why you're so opposed to PEP8?
I usually lint against PEP8 and find it a quite sensible set of style guidelines..

flinz:
@tammoippen Most of the PyNEST still needs to be formatted to comply to pep8.
Are already in the process of doing this? I couldn't see anything on your issue branch.

If not, I might as well clean it up when I am going over it (while [reformatting the doc strings](https://github.com/nest/nest-simulator/issues/257))

tammoippen:
@flinz I was not planning on reformatting them in that PR. It would be great when you do this. I start the PR as soon as #213 is merged, as I put the changes on top of that code.

flinz:
@tammoippen both pep8 and docs are WIP on https://github.com/flinz/nest-simulator/tree/pynest-docs

heplesser:
@flinz Great work, thanks a lot!

heplesser:
With #337 merged, I will close this issue.

PR #263, opened 2016-03-07
PyNEST: pep8 compliance and Numpy-formatted docstrings.

flinz:
This PR came about due to https://github.com/nest/nest-simulator/issues/188 and https://github.com/nest/nest-simulator/issues/257.

# Major changes:
- added numpy standardized docstring ([see here for specs](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)) to all functions in ``pynest/nest``, excluding ``pynest/nest/tests`` in preparation for autogeneration of documentation
- pep8 compliant formatting for almost all python files in ``pynest/nest``, so the nest directory now checks out as almost clean under pep8 (see issues below)
- ``raster_plot.py`` has been slightly rewritten to simplify redundant keyword arguments

# Remaining issues
- I tried as best I could to infer missing parameter types etc for the doc-strings, there might be some slight "mistypings".
- Problem with ``pynest/nest/__init__.py`` scipy bugfix: Right now pep8 checks out as almost clean, except the 4 errors below. This is due to the fact that the [imports of ``pynestkernel`` and ``hl_api_helper``](https://github.com/nest/nest-simulator/pull/263/files#diff-14ca38d825613569e5bb0ab8ea82e266R58) come after the scipy bugfix ([see here](https://github.com/nest/nest-simulator/pull/263/files#diff-14ca38d825613569e5bb0ab8ea82e266L37)). Does anybody know more about this scipy import issue? Can we move the 4 import lines to the top of the file? I do not see scipy being imported anywhere there, so in my opinion this should be safe to do.
```
% pep8 nest
nest/__init__.py:57:1: E402 module level import not at top of file
nest/__init__.py:58:1: E402 module level import not at top of file
nest/__init__.py:60:1: E402 module level import not at top of file
nest/__init__.py:61:1: E402 module level import not at top of file
```


heplesser:
@flinz Thanks for your work! At the moment, the checks fail due to some copyright header problem---could you check that?

Import order, especially wrt SciPy/NumPy has been a problem for a long time, and one sensitive to a number of details, if I remember correctly. I would tend not to fiddle with the order of imports just to make PEP8 happy, in particular in an `__init__.py` that is not part of the public Python interface of NEST. I think this is a place where we probably should suppress the PEP8 warning.

flinz:
@heplesser copyright header errors should be fixed now.
I also moved all imports in ``__init__.py`` back to the original order (I had already moved two) just to be sure.

pynest/nest/__init__.py
```diff
@@ -82,27 +87,42 @@ def catching_sli_run(cmd):
...
-       Examples:
-         r,q = sli_func('dup rollu add',2,3)
-         r   = sli_func('add',2,3)
-         r   = sli_func('add pop',2,3)
-         l   = sli_func('CreateLayer', {...}, namespace='topology')
-         opt = sli_func('GetOptions', '/RandomConvergentConnect', litconv=True)
+    """Convenience function for executing the sequence
+    sli_push(args); sli_run(s); y=sli_pop().
+
+    The function takes an
```
jougs: Something is missing here. Probably, the sentence can just be deleted, as the thing with the arbitrary number of arguments is written below.

pynest/nest/lib/hl_api_connections.py
```diff
@@ -27,44 +27,84 @@
...
+    source : list
+        Source GIDs, only connections from these
+        pre-synaptic neurons are returned
+    target : list, optional
+        Target GIDs, only connections to these
+        post-synaptic neurons are returned
+    synapse_model : str, optional
+        Only connections with this synapse model are returned
+    synapse_type : str, optional
+        Only connections with this synapse type are returned
```
jougs: I'd add the fact that this is the same as `synapse_model`. On the other side, the function is deprecated since long and can as well be deleted right after this PR is merged.

jougs: Forget this! I just saw that it is mentioned below. Nevermind.

heplesser:
:+1: from my side.

jougs:
I gave up reviewing after I've read half of it and did not find a single error. So, pragmatic :+1: from my side as this is definitely better than what we had before. @tammoippen or @abigailm: if you agree with my lax style of reviewing, could you please merge?

Issue #264, opened 2016-03-08
Raise exception when simulation times not commensurable with resolution requested


heplesser:
In certain places in NEST, times need to be multiples of the resolution. Currently, this is only checked in `spike_generator::Parameters_::set()`. Other places where this should be checked are
- setting origin/start/stop in `Device::Parameters_::set()`
- `SimulationManager::simulate()`, which currently only checks against overflow

Concerning the latter in particular at the following, look at the following result in a newly started NEST with default resolution 0.1 ms:

```
SLI ] 0.11 Simulate
Feb 29 15:28:38 Simulate [Info]:
    Simulating 0.11 ms.
SLI ] 0 /time get =
0.2
SLI ] 0.11 Simulate
Feb 29 15:28:55 Simulate [Info]:
    Simulating 0.11 ms.
SLI ] 0 /time get =
0.4
```

NEST always rounds the Simulation time up to 0.2 ms, but tells the user that it is simulating the requested 0.11 ms. Thus, after the second Simulate call, the user expects NEST to be at time 0.22ms, while NEST is at 0.4 ms --- without any warning.

heplesser:
Fixed by #265, therefore closing.

PR #265, opened 2016-03-08
Fixing #264: Raise exceptions on incommensurate times

heplesser:
Simulate now raises exception if simulation time is not multiple of the resolution. SetStatus on devices do the same if start, stop, origin are not multiples of the resolution. Some tests needed minor adaptation.

@jougs @apeyser Would you take a look?


jougs:
I think the code is very nice. :+1: once Travis is also OK with it.

apeyser:
:+1: I'll merge after ci.

Issue #238, opened 2016-02-26
No speedup from mpi parallelization (SLI and PyNEST)


antolikjan:
Nest 2.10.0, intel compiler suite, RedHat version 6.7

Compiled with following options:
configure CC=mpiicc CFLAGS="-openmp -mt_mpi -O3" CXX=mpiicpc CXXFLAGS="-openmp -mt_mpi -O3" LDFLAGS="-L/.../mkl/lib/intel64 -lmkl_intel_lp64 -lmkl_sequential -lmkl_core -lpthread -lm -L/.../compiler/lib/intel64 -liomp5"

To test parallelization we use the hpc_benchmark.sli and slightly modified version of the brunel_alpha_nest.py example script.

The parallelization with threads works perfectly, achieving nearly linear speedup with up to 32 threads (maximum cores on a single node).

However, if I schedule a job with multiple mpi processes (single thread per mpi process) I observe no speedup, if anything the simulation runs slightly slower than in single thread/process condition, both for the sli and pynest scripts.

The outputs of the hpc_benchmark.sli when run with 1 and 16 threads:
[hpc_benchmark_mpiproc=16_out.tar.gz](https://github.com/nest/nest-simulator/files/148454/hpc_benchmark_mpiproc.16_out.tar.gz)
[hpc_benchmark_mpiproc=1_out.tar.gz](https://github.com/nest/nest-simulator/files/148455/hpc_benchmark_mpiproc.1_out.tar.gz)

The LoadLeveler submit script with which I submit the jobs:
[sub.ll.tar.gz](https://github.com/nest/nest-simulator/files/148459/sub.ll.tar.gz)

antolikjan:
I have found another problem which might be related to above, so I am posting it here in the hope it will help somebody to diagnose what is going on.

If I increase the number of neurons to above ~50000 neurons (645 mil. synapses), regardless how many mpi processes I select (i.e. I am sure that sufficient amount of memory _should have been_ allocated), the simulation crashes with following alloc errors:

[mem_error.tar.gz](https://github.com/nest/nest-simulator/files/148532/mem_error.tar.gz)

This is using the brunel_alpha_nest.py script.

heplesser:
[...]

PR #269, opened 2016-03-09
Clean up logging upon calls to Simulate

jougs:
The output upon calls to `Simulate` has previously been spread out over multiple messages and cluttered by useless messages. This was hard to read by humans and hard to parse by machines. This PR cleans and compacts logging messages and adds the number of threads and processes to the output in order to prevent problems like seen in #238.

The old output looked like this:
```
Mar 09 18:22:58 Simulate [Info]: 
    Simulating 100 ms.

Mar 09 18:22:58 NodeManager::prepare_nodes_ [Info]: 
    Please wait. Preparing elements.

Mar 09 18:22:58 NodeManager::prepare_nodes_ [Info]: 
    Simulating 1000 local nodes.

Mar 09 18:22:58 SimulationManager::update [Info]: 
    Simulating using OpenMP.

Mar 09 18:22:58 SimulationManager::resume [Info]: 
    Simulation finished.
OpenMP
Mar 09 18:22:58 NodeManager::finalize_nodes() [Info]: 
     using OpenMP.
```

The new output for the same script looks like this:
```
Mar 09 18:26:55 NodeManager::prepare_nodes [Info]: 
    Preparing 1000 nodes for simulation.

Mar 09 18:26:55 SimulationManager::resume [Info]: 
    Number of local nodes: 1000
    Simulaton time (ms): 100
    Number of OpenMP threads: 2
    Number of MPI processes: 1

Mar 09 18:26:55 SimulationManager::resume [Info]: 
    Simulation finished.
```

If NEST was compiled without support for OpenMP, the third line of the table will read
```
Not using OpenMP
```

If NEST was compiled without support for MPI, the fourth line of the table will read
```
Not using MPI
```

While testing with 0 nodes, I found a segfault, which I then also fixed in this PR for convenience reasons.

### DIFF
--- a/nestkernel/nest.cpp
+++ b/nestkernel/nest.cpp
@@ -225,12 +225,7 @@ get_connections( const DictionaryDatum& dict )
 void
 simulate( const double_t& time )
 {
-  std::ostringstream os;
-  os << "Simulating " << time << " ms.";
-  LOG( M_INFO, "Simulate", os.str() );
-  Time t = Time::ms( time );
-
-  kernel().simulation_manager.simulate( t );
+  kernel().simulation_manager.simulate( Time::ms( time ) );
 }
 
 void
--- a/nestkernel/node_manager.cpp
+++ b/nestkernel/node_manager.cpp
@@ -115,6 +115,7 @@ NodeManager::initialize()
   /* END of code adding the root subnet. */
 
   // explicitly force construction of nodes_vec_ to ensure consistent state
+  nodes_vec_network_size_ = 0;
   ensure_valid_thread_local_ids();
 }
 
--- a/nestkernel/node_manager.cpp
+++ b/nestkernel/node_manager.cpp
@@ -646,12 +646,10 @@ NodeManager::set_status_single_node_( Node& target, const DictionaryDatum& d, bo
   }
 }
 
-void
+size_t
 NodeManager::prepare_nodes()
 {
   assert( kernel().is_initialized() );
-
-  LOG( M_INFO, "NodeManager::prepare_nodes_", "Please wait. Preparing elements." );
 
   /* We initialize the buffers of each node and calibrate it. */
 
@@ -701,23 +699,20 @@ NodeManager::prepare_nodes()
     if ( exceptions_raised.at( thr ).valid() )
       throw WrappedThreadException( *( exceptions_raised.at( thr ) ) );
 
-  if ( num_active_prelim_nodes == 0 )
+  std::ostringstream os;
+  std::string tmp_str = num_active_nodes == 1 ? " node" : " nodes";
+  os << "Preparing " << num_active_nodes << tmp_str << " for simulation.";
+
+  if ( num_active_prelim_nodes != 0 )
   {
-    LOG( M_INFO,
-      "NodeManager::prepare_nodes_",
-      String::compose(
-           "Simulating %1 local node%2.", num_active_nodes, num_active_nodes == 1 ? "" : "s" ) );
+    tmp_str = num_active_prelim_nodes == 1 ? " uses " : " use ";
+    os << " " << num_active_prelim_nodes << " of them" << tmp_str
+       << "iterative solution techniques.";
   }
-  else
-  {
-    LOG( M_INFO,
-      "Scheduler::prepare_nodes",
-      String::compose( "Simulating %1 local node%2 of which %3 need%4 prelim_update.",
-           num_active_nodes,
-           num_active_nodes == 1 ? "" : "s",
-           num_active_prelim_nodes,
-           num_active_prelim_nodes == 1 ? "s" : "" ) );
-  }
+
+  LOG( M_INFO, "NodeManager::prepare_nodes", os.str() );
+
+  return num_active_nodes;
 }
 
 //!< This function is called only if the thread data structures are properly set up.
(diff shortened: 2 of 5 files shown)


jougs:
@apeyser, @heplesser, would you please review? Thanks!

nestkernel/simulation_manager.cpp
```diff
@@ -387,12 +404,9 @@ nest::SimulationManager::resume_()
   LOG( M_INFO, "SimulationManager::resume", "Simulation finished." );
 }
 
-void
+size_t
 nest::SimulationManager::prepare_simulation_()
 {
-  if ( to_do_ == 0 )
-    return;
-
```
jougs: This could never be reached, as the calling function already has a check for `t != 0`

apeyser: Then it should have an assert and a comment on the invariant

nestkernel/simulation_manager.cpp
```diff
@@ -298,7 +298,7 @@ nest::SimulationManager::simulate( Time const& t )
   to_do_ += t.get_steps();
   to_do_total_ = to_do_;
 
-  prepare_simulation_();
+  size_t num_active_nodes = prepare_simulation_();
```
apeyser: This is a const, correct?

nestkernel/simulation_manager.cpp
```diff
@@ -325,16 +325,41 @@ nest::SimulationManager::simulate( Time const& t )
...
 
-  resume_();
+  resume_( num_active_nodes );
 
   finalize_simulation_();
 }
 
 void
-nest::SimulationManager::resume_()
+nest::SimulationManager::resume_( size_t num_active_nodes )
```
apeyser: Is this (by the way const) num_active_nodes commented as simply a logging value and not actually usable to alter the number of active nodes? Or should this be a SimulationManager member?

nestkernel/node_manager.cpp
```diff
@@ -724,14 +720,12 @@ void
...
-  LOG( M_INFO, "NodeManager::finalize_nodes()", " using OpenMP." );
-// parallel section begins
 #pragma omp parallel
   {
     index t = kernel().vp_manager.get_thread_id();
 #else
-    for ( index t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
-    {
+  for ( index t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
+  {
```
tammoippen: You managed to trick `clang-format`: it thinks these two lines should be indented one more level. This is why TravisCI complains. Either you indent them, or surround them with `// clang-format off` and `// clang-format on` or we need to think off something else (clang-format-3.8 still has the same problem).

jougs:
I've added an assert and constness as suggested by @apeyser and disabled clang-format for some lines as suggested by @tammoippen. Many thanks for the thorough review.

nestkernel/nest.cpp
```diff
@@ -225,12 +225,7 @@ get_connections( const DictionaryDatum& dict )
 void
 simulate( const double_t& time )
 {
-  std::ostringstream os;
-  os << "Simulating " << time << " ms.";
-  LOG( M_INFO, "Simulate", os.str() );
-  Time t = Time::ms( time );
-
-  kernel().simulation_manager.simulate( t );
+  kernel().simulation_manager.simulate( Time::ms( time ) );
```
heplesser: Merging may lead to a conflict with #265, but that should be easy to handle.

nestkernel/node_manager.h
```diff
@@ -194,8 +194,9 @@ class NodeManager : ManagerInterface
    * Prepare nodes for simulation and register nodes in node_list.
    * Calls prepare_node_() for each pertaining Node.
    * @see prepare_node_()
+   * @returns number of nodes that will be simulated.
    */
-  void prepare_nodes();
+  const size_t prepare_nodes();
```
heplesser: I think the `const` here makes little sense and would remove it.

heplesser:
@jougs Great and :+1: once you have moved the `const`, see my and @apeyser's comments in the code.

heplesser:
@jougs The conflict is due to my merging #265, where I added parameter checking to `simulate()` in `nest.cpp`. I think that should be rather easy to sort out.

jougs:
@heplesser: I've merged in your changes and removed the const-ness for `prepare_nodes()`

heplesser:
@jougs :+1: from me once CI is green, but I think @apeyser would appreciate a `const` on line 301 in `simulation_manager.cpp`, see his comment above.

nestkernel/node_manager.cpp
```diff
@@ -724,15 +720,13 @@ void
...
 #pragma omp parallel
   {
     index t = kernel().vp_manager.get_thread_id();
-#else
-    for ( index t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
-    {
-#endif
+#else // clang-format off
+  for ( index t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
+  {
```
tammoippen: This `for`-loop start should be indented one more level.

tammoippen:
Maybe we should also include the MPI rank in the status message or limit the status message to rank 0?

nestkernel/node_manager.cpp
jougs: No, it shouldn't. Depending on the definition of `_OPENMP`, the compiler either sees
```
#pragma omp parallel
  {
```
or
```
  for (...)
  {
```
The `for` statement in line in (new) line 727 is _not_ inside the `#pragma omp parallel` block, but replaces it conditionally. So the indentation level of the opening curly brace in line 724 and (new) line 728 have to match.

As `clang-format` does apparently not take conditional compilation into account, I had to disable it to get the correct indentation.

jougs:
Only printing on rank 0 seems strange to me, as the number of local nodes might differ and this message might be very useful for debugging.

But I agree that rank information would be useful. However, I think it would be better if that was added globally to the `LOG` function to have it included in all output. I can open a corresponding PR if you think that's a good idea.

tammoippen:
Missed the indentation. :+1: for this PR. (merging)

I think, both rank and thread would be useful information for `LOG` - in another PR.

PR #273, opened 2016-03-10
Improvement of iterative solution framework

janhahne:
This PR improves the iterative solution framework used for connections without delay (so far only gap junctions, but there is more to come).

- Renamed several internal and kernel variables from rather unspecific names to names of typ `wfr_*` in order to indicate that they belong to the iterative waveform relaxation (wfr) scheme.
```
max_num_prelim_iterations -> wfr_max_iterations
prelim_tol -> wfr_tol
prelim_interpolation_order -> wfr_interpolation_order
prelim_update -> wfr_update
needs_prelim_update -> node_uses_wfr
...
```

- Added kernel variable `wfr_comm_interval`. All connections without delay contribute with `wfr_comm_interval` to the `min_delay` and `max_delay`. The communication of the waveform relaxation method uses the regular communication points in intervals of the minimal network delay. Therefore it is desirable to be able to influence the `min_delay` for simulations with connections without delay

- Added kernel flag `use_wfr` to disable usage of iterative waveform relaxation method and use communication in every step instead ( single step method ). This is achieved by setting `wfr_comm_interval = resolution` and will be useful for rate models without delay where wfr is not always necessary and to reproduce data of the single step method from our gap-junction paper.

- Added a new unit test `test_wfr_settings.sli` that checks that `wfr_comm_interval` and `use_wfr` work properly and can only be set in a reasonable order. Adapted `test_gap_junction.sli` due to functionality of `wfr_comm_interval`.

- Improved and corrected Python-Code in both gap-junction examples

Most certainly there will be a small conflict with PR #269 in `node_manager.cpp`. I will adapt this PR once #269 is merged.

### DIFF
--- a/examples/nest/gap_junction/inhibitory_network.py
+++ b/examples/nest/gap_junction/inhibitory_network.py
@@ -43,6 +43,7 @@ network resulting in an average of 60 gap-junction connections per neuron.
 import pylab
 import nest
 import random
+import numpy
 
 n_neuron = 500
 gap_per_neuron = 60
@@ -64,23 +65,42 @@ random.seed(1)
 
 nest.ResetKernel()
 
-nest.SetKernelStatus({'resolution': 0.05, 'total_num_virtual_procs': threads})
-nest.SetKernelStatus({'max_num_prelim_iterations': 15, 'prelim_interpolation_order': 3, 'prelim_tol': 0.0001})
+nest.SetKernelStatus({'resolution': 0.05, 
+                      'total_num_virtual_procs': threads,
+                      'print_time': True,
+# Settings for waveform relaxation
+# 'use_wfr': False uses communication in every step 
+# instead of an iterative solution
+                      'use_wfr': True,
+                      'wfr_comm_interval': 1.0,
+                      'wfr_tol': 0.0001,
+                      'wfr_max_iterations': 15,
+                      'wfr_interpolation_order': 3})
 
-neuron = nest.Create('hh_psc_alpha_gap',n_neuron)
+neurons = nest.Create('hh_psc_alpha_gap',n_neuron)
 
 sd = nest.Create("spike_detector", params={'to_file': False, 'to_memory': True})
 pg = nest.Create("poisson_generator",params={'rate': 500.0})
 
-conn_dict = {'rule': 'fixed_indegree', 'indegree': inh_per_neuron, 'autapses': False, 'multapses': True}
-syn_dict = {'model': 'static_synapse', 'weight': j_inh, 'delay': delay}
-nest.Connect(neuron, neuron, conn_dict, syn_dict)
+conn_dict = {'rule': 'fixed_indegree', 
+             'indegree': inh_per_neuron, 
+             'autapses': False, 
+             'multapses': True}
 
-nest.Connect(pg,neuron,'all_to_all',syn_spec={'model': 'static_synapse', 'weight': j_exc, 'delay': delay})
-nest.Connect(neuron,sd,'all_to_all')
+syn_dict = {'model': 'static_synapse', 
+            'weight': j_inh, 
+            'delay': delay}
+
+nest.Connect(neurons, neurons, conn_dict, syn_dict)
+
+nest.Connect(pg,neurons,'all_to_all',syn_spec={'model': 'static_synapse', 
+                                              'weight': j_exc, 
+                                              'delay': delay})
+
+nest.Connect(neurons,sd)
 
 for i in range(n_neuron):
-  nest.SetStatus([neuron[i]], { 'V_m': (-40. - 40. * random.random()) })
+  nest.SetStatus([neurons[i]], { 'V_m': (-40. - 40. * random.random()) })
 
 """
 We must not use the 'fixed_indegree' oder 'fixed_outdegree' functionality of nest.Connect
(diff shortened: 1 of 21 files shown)


jougs:
I suggest @tammoippen and @jschuecker as reviewers.

examples/nest/gap_junction/inhibitory_network.py
```diff
@@ -64,23 +65,43 @@
...
+
+nest.Connect(neurons, neurons, conn_dict, syn_dict)
 
-nest.Connect(pg,neuron,'all_to_all',syn_spec={'model': 'static_synapse', 'weight': j_exc, 'delay': delay})
-nest.Connect(neuron,sd,'all_to_all')
+nest.Connect(pg,neurons,'all_to_all',syn_spec={'model': 'static_synapse', 
+                                              'weight': j_exc, 
+                                              'delay': delay})
+
+nest.Connect(neurons,sd,'all_to_all')
```
jschuecker: Since `all_to_all` is default you could leave it out.

nestkernel/connector_model_impl.h
```diff
@@ -216,9 +224,21 @@ GenericConnectorModel< ConnectionT >::add_connection( Node& src,
...
-  if ( not numerics::is_nan( delay ) && has_delay_ )
-    kernel().connection_builder_manager.get_delay_checker().assert_valid_delay_ms( delay );
-
+  if ( not numerics::is_nan( delay ) )
+  {
+    if ( has_delay_ )
+    {
+      kernel().connection_builder_manager.get_delay_checker().assert_valid_delay_ms( delay );
+    }
+    // Let connections without delay contribute to min_delay with wfr_comm_interval
```
jschuecker: to min_delay and max_delay, right? However the contribution to max_delay is not important, so I am not sure if we should mention both here!?

jschuecker:
@janhahne, is there a reason, why the pynest examples are not in pynest/examples?

I made only two minor comments in line. :+1: otherwise.

janhahne:
@jschuecker Thank you for the review! I addressed your two comments in the new commit.

There is no particular reason why the examples are in examples/nest instead of in pynest/examples. Should they be moved there?

nestkernel/connector_model_impl.h
```diff
@@ -180,6 +180,14 @@ GenericConnectorModel< ConnectionT >::used_default_delay()
         kernel().connection_builder_manager.get_delay_checker().assert_valid_delay_ms(
           default_connection_.get_delay() );
       }
+      // Let connections without delay contribute to the delay extrema with wfr_comm_interval
+      // For those connections the min_delay is important as it determines the length
+      // of the global communication interval.
+      else
+      {
+        kernel().connection_builder_manager.get_delay_checker().assert_valid_delay_ms(
+          kernel().simulation_manager.get_wfr_comm_interval() );
```
tammoippen: If I understand correctly, the `wfr_comm_interval` is set only during `SimulationManager::set_status` and cannot change otherwise. If yes, then we need to check for a valid value only once (not for every 'delay-less' connection) when it is changed in the SimulationManager or when when `min/max_delay` is change in the `ConnectionBuilderManager`. Also relevant for the subsequent changes in this file.

examples/nest/gap_junction/inhibitory_network.py
```diff
@@ -64,23 +65,43 @@
...
 
-neuron = nest.Create('hh_psc_alpha_gap',n_neuron)
+# Settings for waveform relaxation
+# 'use_wfr': False uses communication in every step 
+# instead of an iterative solution
+nest.SetKernelStatus({'use_wfr': True,
+                      'wfr_comm_interval': 1.0,
+                      'wfr_tol': 0.0001,
+                      'wfr_max_iterations': 15,
+                      'wfr_interpolation_order': 3})
```
tammoippen: Why do you have two subsequent calls to `SetKernelStatus`? Can't you call it once like this: 

    nest.SetKernelStatus({'resolution': 0.05, 
                          'total_num_virtual_procs': threads,
                          'print_time': True,
    # comments ...
                          'use_wfr': True,
                          'wfr_comm_interval': 1.0,
                          'wfr_tol': 0.0001,
                          'wfr_max_iterations': 15,
                          'wfr_interpolation_order': 3})

(Similar in the other example.)

janhahne:
@tammoippen Thank you for taking a look and your comments! 

I improved the calls to `SetKernelStatus` in the examples. Regarding your second comment I am not quite sure, if I understand your suggestion correctly. You are right, that `wfr_comm_interval` is set only during `SimulationManager::set_status` and you are also right, that it can be interpreted as a default delay for connections without delay (which is however used anyway, even if the user sets a different delay for the connection).

So I see your point, that once one 'delay-less' connection is created it is no longer necessary to call `assert_valid_delay_ms()` for any further 'delay-less' connection, because the `min/max delay` is set correctly. I am however not sure how to achieve this in a smart way. Could you give me another hint (or if easier for you create a PR to my branch)?

janhahne:
@tammoippen Please see commit 0505f81. I think I found a solution. I would appreciate a quick merge of this pull request in order to move forward with the rate model implementation.

nestkernel/connector_model_impl.h
```diff
@@ -463,6 +476,19 @@ GenericConnectorModel< ConnectionT >::add_connection( Node& src,
...
+  // The call to assert_valid_delay_ms needs to happen only once
+  // (either here or in used_default_delay()) when the first connection
+  // without delay is created.
+  if ( default_delay_needs_check_ && not has_delay_ )
+  {
+    kernel().connection_manager.get_delay_checker().assert_valid_delay_ms(
+      kernel().simulation_manager.get_wfr_comm_interval() );
+    default_delay_needs_check_ = false;
+  }
+
```
tammoippen: This solution looks very good. Maybe it should be placed at the start of the function (line 354ff), because if something is bad with the delay, the connection is not stored in the connection infrastructure.

tammoippen:
When my last note is addressed, I am all for merging.

nestkernel/connector_model_impl.h
janhahne: Good catch, thanks!

janhahne:
Great, thanks! Since @jschuecker also gave his :+1:, this should be ready for merging now :)

Issue #101, opened 2015-09-09
Review current input and its documentation in iaf_psc_exp


heplesser:
`iaf_psc_exp` handles current input supplied via `CurrentEvent`s differently than other neuron models. Input via rport 0 is handled the normal way, while input via rport 1 is filtered as excitatory synaptic input.
This behavior is documented in one sentence, but that sentence is too little visible for an additional feature such at this. It should at least be on a line of its own, probably be `Remark`.

I am also struggling to understand the physical basis of this filtering. An arriving spike releases transmitter vesicles which dock at receptors and thus evoke an input current. The exponential decay of the current captures the process in which "all" channels open instantly and then close over time as transmitter molecules detach again. Thus, there is a physical basis for a current that persists beyond the arrival time of the spike.

But when we inject currents via `CurrentEvent`s, we model injection of currents via an electrode. Consider now a case where the electrode injects current only during a single time step. Then, immediately after that time step, the physical input current to the neuron is zero. But in the "filtered current" model, this current would persist as an exponentially decaying current for several milliseconds. What is the model behind this? It should be explained in the documentation.  If one just wants to have a possiblity of injecting a low-pass filtered current into a neuron, one could use `step_current_generator`, setting the current amplitudes to filtered values.

@jschuecker, could you take a look at this?

jschuecker:
This filtering of the current is motivated by the diffusion approximation, i.e. replacing the incoming  spike-trains with a gaussian white noise. This noise has to be filtered by the synapse. I implemented this to prove our newly developed theory about neurons subjected to coloured noise, caused by synaptic filtering. Is this explanation satisfactory? 

I will improve the documentation.

sanjayankur31:
This one is fixed - needs closing @heplesser , @jschuecker

PR #276, opened 2016-03-13
Fix 101 Documentation of current filtering in iaf_psc_exp

jschuecker:
This PR improves the documentation of the option for synaptic filtering of external current, available in `iaf_psc_exp`. This addresses #101 .

@heplesser, could you take a look at this?

### DIFF
--- a/models/iaf_psc_exp.h
+++ b/models/iaf_psc_exp.h
@@ -96,12 +96,13 @@ Remarks:
    For details, please see IAF_Neruons_Singularity.ipynb in the
    NEST source code (docs/model_details).
 
-   If the current event corresponding to the piecewise constant
-   external current is received at port 1 the current is filtered by
-   the synapse. Hereby the time constant of the post-synaptic
-   excitatory currents is used. This filtering can be used for a
-   direct implementation of the diffusion approximation, i.e.,
-   replacing incoming spike trains with white noise.
+   iaf_psc_exp can handle current input in two ways: Current input
+   through receptor_type 0 are handled as stepwise constant current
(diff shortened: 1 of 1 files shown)


models/iaf_psc_exp.h
```diff
@@ -99,6 +96,13 @@ class Network;
    For details, please see IAF_Neruons_Singularity.ipynb in the
    NEST source code (docs/model_details).
 
+   If the current event corresponding to the piecewise constant
+   external current is received at port 1 the current is filtered by
+   the synapse. Hereby the time constant of the post-synaptic
+   excitatory currents is used. This filtering can be used for a
+   direct implementation of the diffusion approximation, i.e.,
+   replacing incoming spike trains with white noise.
```
heplesser: @jschuecker Thanks for the more detailed explanation. I would be yet more explicit, e.g.

"iaf_psc_exp can handle current input in two ways: Current input through receptor_type 0 are handled as stepwise constant current input as in other iaf models, i.e., this current directly enters the membrane potential equation. Current input through receptor_type 1, in contrast, is filtered through an exponential kernel with the time constant of the excitatory synapse, tau_syn_ex. For an example application, see [4; reference to paper inserted below]."

I am a bit concerned about the statement that this provides a "direct implementation ... with white noise". Since the noise generator generates stepwise constant input, the input is never really white noise, see doc/model_details/noise_generator.ipynb.

jschuecker:
@heplesser, I meant "direct" in the sense of a "straight forward" implementation. But you are right, that was not clear and could be misleading. I also included the reference. Thanks!

heplesser:
:+1: and merging.

PR #280, opened 2016-03-17
Music cont out proxy

uahic:
New neuron type to support forwarding continuous values via the MUSIC continuous output port. 

Notes:
- A description how it works can be found in the header file.
- An example is provided in the /pynest/examples/music_cont_out_proxy_example directory
- The code is not written to support the MusicManager class yet. If this is desired we need to discuss implementation details (how to encapsulate port buffers, etc.). 
- The max_buffered property for the port cannot be set by the user in the current state. 
- It might be desirable to clean up the code in the future


tammoippen:
Hello @uahic ,

TravisCI is complaining about the formatting of the code. Please follow the instructions on our [developer space](https://nest.github.io/nest-simulator/coding_guidelines_c++) on formatting the code with clang-format.

uahic:
@tammoippen I really did all the checks now and I used some vim plugins for clang-format but also did it via command line. Travis still complaints because of some line breaks but they are enforced by the screen width of 80 which is set in the .clang-format file. I am a bit lost ...

tammoippen:
Hello @uahic,
The formatting of some committed files does not confirm to our [coding guidelines](https://nest.github.io/nest-simulator/coding_guidelines_c++). Please, closely follow the instructions you find there, especially the [Tooling](https://nest.github.io/nest-simulator/coding_guidelines_c++#tooling) section gives information on automated formatting with [`clang-format`](http://llvm.org/releases/3.6.0/tools/clang/docs/ClangFormat.html) and [local testing](https://nest.github.io/nest-simulator/coding_guidelines_c++#check-static-analysis-local) of the coding guidelines. The automated formatting is obligatory for accepting PR; please ensure to use `clang-format` version 3.6, as the formatting varies with different versions. NEST provides a [`.clang-format`](https://github.com/nest/nest-simulator/blob/master/.clang-format) file in its sources.

tammoippen:
@weidel-p Can you please look into this PR? Thank you.

uahic:
I found a bug in my model in case when the ports are not correctly mapped, a patch will follow ASAP

uahic:
Update:  the error seems to originate from a message input port (segfault)

uahic:
The class will be refactored and the branch merged with the HEAD today, please keep the PR open

libnestutil/hashtable-common.h
```diff
@@ -40,12 +40,12 @@
...
 
-#include <sparseconfig.h>
 #include <assert.h>
-#include <stdio.h>
-#include <stddef.h> // for size_t
 #include <iosfwd>
+#include <sparseconfig.h>
+#include <stddef.h>  // for size_t
 #include <stdexcept> // For length_error
+#include <stdio.h>
```
jougs: Is there a specific reason for re-sorting all includes, which is mandated by the addition of the `music_cont_out_proxy`? If there isn't, please revert to the original order.

This is especially important for files not actually provided by the NEST Initiative, like the ones implementing the Google Sparsetable (i.e. `sparsetable.h`, `libc_allocator_with_realloc.h`, `hashtable-common.h`, `sparseconfig.h`, `template_util.h` in `libnestutil`). We want to have these files in the source directory exactly like they come from the upstream source.

For more information on the order of includes, please see our [C++ coding style guide, section _order of includes_](https://nest.github.io/nest-simulator/coding_guidelines_c++#order-of-includes). For files that are original parts of NEST, you can also run our include-checker `extras/includes_checker.py`.

libnestutil/libc_allocator_with_realloc.h
```diff
@@ -32,10 +32,10 @@
 #ifndef UTIL_GTL_LIBC_ALLOCATOR_WITH_REALLOC_H_
 #define UTIL_GTL_LIBC_ALLOCATOR_WITH_REALLOC_H_
 
+#include <new> // for placement new
 #include <sparseconfig.h>
-#include <stdlib.h> // for malloc/realloc/free
 #include <stddef.h> // for ptrdiff_t
-#include <new>      // for placement new
+#include <stdlib.h> // for malloc/realloc/free
```
jougs: See above.

libnestutil/sparsetable.h
```diff
@@ -237,14 +237,14 @@
...
-#include <memory>    // uninitialized_copy, uninitialized_fill
-#include <vector>    // a sparsetable is a vector of groups
-#include <type_traits.h>
+#include <assert.h>  // for bounds checking
 #include <hashtable-common.h>
+#include <iterator> // to define reverse_iterator for me
 #include <libc_allocator_with_realloc.h>
+#include <memory> // uninitialized_copy, uninitialized_fill
+#include <type_traits.h>
+#include <vector> // a sparsetable is a vector of groups
```
jougs: See above.

models/hh_psc_alpha_gap.h
```diff
@@ -39,8 +39,8 @@
 #include "event.h"
 #include "nest_types.h"
 #include "node.h"
-#include "ring_buffer.h"
 #include "recordables_map.h"
+#include "ring_buffer.h"
 #include "universal_data_logger.h"
```
jougs: See above. This file has nothing to do with the PR.

models/modelsmodule.cpp
```diff
@@ -109,9 +109,9 @@
 #include "stdp_connection_facetshw_hom.h"
 #include "stdp_connection_facetshw_hom_impl.h"
 #include "stdp_connection_hom.h"
-#include "stdp_triplet_connection.h"
 #include "stdp_dopa_connection.h"
 #include "stdp_pl_connection_hom.h"
+#include "stdp_triplet_connection.h"
```
jougs: See above. This file has nothing to do with the PR.

models/modelsmodule.cpp
```diff
@@ -41,8 +41,8 @@
 
 // Neuron models
 #include "aeif_cond_alpha.h"
-#include "aeif_cond_alpha_multisynapse.h"
 #include "aeif_cond_alpha_RK5.h"
+#include "aeif_cond_alpha_multisynapse.h"
 #include "aeif_cond_exp.h"
```
jougs: See above. This file has nothing to do with the PR.

models/music_event_in_proxy.cpp
```diff
@@ -39,8 +39,8 @@
 #include "logging.h"
 
 // Includes from nestkernel:
-#include "kernel_manager.h"
 #include "event_delivery_manager_impl.h"
+#include "kernel_manager.h"
```
jougs: See above. This file has nothing to do with the PR.

models/stdp_triplet_connection.h
```diff
@@ -79,8 +79,8 @@
   SeeAlso: stdp_triplet_synapse_hpc, synapsedict, stdp_synapse, static_synapse
 */
 
-#include <cmath>
 #include "connection.h"
+#include <cmath>
```
jougs: See above. This file has nothing to do with the PR.

nest/neststartup.cpp
```diff
@@ -42,9 +42,9 @@
 // Includes from nestkernel:
 #include "dynamicloader.h"
 #include "kernel_manager.h"
+#include "model_manager_impl.h"
 #include "nest.h"
 #include "nestmodule.h"
-#include "model_manager_impl.h"
```
jougs: See above. This file has nothing to do with the PR.

nestkernel/connection_label.h
```diff
@@ -23,10 +23,10 @@
 #ifndef CONNECTION_LABEL_H
 #define CONNECTION_LABEL_H
 
-#include "nest.h"
-#include "nest_names.h"
 #include "dictdatum.h"
 #include "dictutils.h"
+#include "nest.h"
+#include "nest_names.h"
```
jougs: See above. This file has nothing to do with the PR.

nestkernel/event.h
```diff
@@ -24,9 +24,9 @@
 #define EVENT_H
 
 // C++ includes:
+#include <algorithm>
 #include <cassert>
 #include <cstring>
-#include <algorithm>
```
jougs: See above. This file has nothing to do with the PR.

nestkernel/event_delivery_manager.h
```diff
@@ -31,8 +31,8 @@
 #include "manager_interface.h"
 
 // Includes from nestkernel:
-#include "mpi_manager.h" // OffGridSpike
 #include "event.h"
+#include "mpi_manager.h" // OffGridSpike
```
jougs: See above. This file has nothing to do with the PR.

nestkernel/nest.cpp
```diff
@@ -28,8 +28,8 @@
 // Includes from nestkernel:
 #include "exceptions.h"
 #include "kernel_manager.h"
-#include "nodelist.h"
 #include "mpi_manager_impl.h"
+#include "nodelist.h"
```
jougs: See above. This file has nothing to do with the PR.

pynest/examples/music_cont_out_proxy_example/README.md
```diff
@@ -0,0 +1,16 @@
+# MUSIC_CONT_OUT_PROXY EXAMPLE
```
jougs: Please remove all-upper formatting.

pynest/examples/music_cont_out_proxy_example/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Requirements
++ MUSIC 1.1.15 or higher
++ NEST 2.10.0 or higher compiled with MPI and MUSIC
++ Numpy
```
jougs: NumPy

sli/fdstream.h
```diff
@@ -33,8 +33,8 @@
 
 #ifndef HAVE_ISTREAM
 // C++ includes:
-#include <iostream>
 #include <fstream>
+#include <iostream>
```
jougs: See above. This file has nothing to do with the PR.

sli/slistartup.h
```diff
@@ -27,9 +27,9 @@
 #include <string>
 
 // Generated includes:
+#include "config.h"
 #include "dirent.h"
 #include "errno.h"
-#include "config.h"
```
jougs: See above. This file has nothing to do with the PR.

uahic:
Thanks for having a look at the code; I am about to move adapt the code such that it is in accordance with the music manager (outsourcing some internals)

jougs:
So, I'm waiting for a more thorough review of the proxy code until after that. Many thanks for the work on this.

uahic:
I had a look on the music manager and it appears that it is currently only used to manage ingoing connections (as multiple devices could listen to the same channel which is - as far as I know - not possible for outgoing music ports). So I think the code is ready for review;  

Note:
As a list of global IDs must be provided by the user, it would be necessary to allow passing GIDCollections (or similar constructs) via Python for very large lists of target GIDs but this goes a bit beyond this particular device and it already has been addressed in other issues.

uahic:
I found some bugs that need to be fixed first...

uahic:
@jougs The code is ready for review

jougs:
@uahic sorry for the delay. Can you please pull in the recent changes from master? Thanks!

jougs:
Before I properly review the code, can you please address my issues about the code formatting? Thanks!

uahic:
Done!, I hope that I did not forget any formatting again

jougs:
@uahic: the issue with the include files are still unresolved. I'll look at the code again later this week. Thanks!

jougs:
@uahic: thanks for taking care of the comments! I'll have a closer look at the code tomorrow.

uahic:
@jougs I fixed an issue with the examples for this neuron model, now everything looks fine to me. Who else would like to review it? (Provided that multiple reviewers are required).

gtrensch:
I will also review it.

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+ * ---------------------------------------------------------------- */
+
+void
+nest::music_cont_out_proxy::Parameters_::get( DictionaryDatum& d ) const
+{
+  ( *d )[ names::port_name ] = port_name_;
+  ( *d )[ names::interval ] = interval_.get_ms();
+
+  ArrayDatum ad_record_from;
+
```
jougs: remove empty line

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  for ( size_t j = 0; j < record_from_.size(); ++j )
+  {
+    ad_record_from.push_back( LiteralDatum( record_from_[ j ] ) );
+  }
+
```
jougs: remove empty line

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  ( *d )[ names::record_from ] = ad_record_from;
+  std::vector< long_t >* pInd_map_long =
+    new std::vector< long_t >( target_gids_.size() );
+
+  std::copy< std::vector< long_t >::const_iterator,
+    std::vector< long_t >::iterator >(
+    target_gids_.begin(), target_gids_.end(), pInd_map_long->begin() );
+
+  ( *d )[ names::targets ] = IntVectorDatum( pInd_map_long );
```
jougs: Can't lines 112 through 119 be simply written as
```C++   
  ( *d )[ names::targets ] = target_gids_;
```
?

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+}
+
+void
+nest::music_cont_out_proxy::Parameters_::set( const DictionaryDatum& d,
+  const Node& self,
+  const State_& states,
```
jougs: Why does state have a plural s here? As far as I see, it's just one

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  const Buffers_& buffs )
```
jougs: replace `buffs` by `buffers`

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+{
+
+  if ( states.published_ == false )
+  {
+    updateValue< string >( d, names::port_name, port_name_ );
+  }
+
+  if ( buffs.has_targets_
+    && ( d->known( names::interval ) || d->known( names::record_from ) ) )
```
jougs: Rather break the line after the `||`

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  {
+    if ( record_from_.empty() )
+    {
+      throw BadProperty(
+        "The property record_from must be set before passing target_gids." );
+    }
+
+    if ( states.published_ == false )
+    {
+      ArrayDatum mca = getValue< ArrayDatum >( d, names::targets );
```
jougs: I think this whole connection business needs to go to `calibrate()`. The way this is implemented currently, multiple calls to `SetStatus` are possible and would lead to multiple connections, which is not the desired behavior.

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+      for ( Token* t = mca.begin(); t != mca.end(); ++t )
+      {
+        target_gids_.push_back( getValue< long_t >( *t ) );
+      }
+      const Token synmodel =
+        kernel().model_manager.get_synapsedict()->lookup( "static_synapse" );
+      if ( synmodel.empty() )
```
jougs: This should be an assertion. Exceptions are for user errors only. In this case the user does not specify a synapse type and can't do anything if this fails. Something like
```C++
assert(!synmodel.empty() && "synapse mode 'static_synapse' not available")
```
should do the trick.

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+      {
+        throw UnknownSynapseType( "static_synapse" );
+      }
+      const index synmodel_id = static_cast< index >( synmodel );
+      std::vector< long_t >::const_iterator t;
+      //
```
jougs: Remove empty comment, leave empty line

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  , B_( n.B_ )
+{
+}
+
+
+void
+nest::music_cont_out_proxy::init_state_( const Node& /* np */ )
+{
+  // const Multimeter& asd = dynamic_cast< const Multimeter& >( np );
+  // device_.init_state( asd.device_ );
```
jougs: Remove commented out code

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,226 @@
...
+
+/* BeginDocumentation
+
+Name: music_cont_out_proxy - A device which sends continuous data from NEST to
+MUSIC.
+
+Description:
+A music_cont_out_proxy can be used to send continuous data from neurons over
+MUSIC to
+remote applications. It works in a similar like a multimeter model. The user
```
jougs: Reformat this block to avoid strange line breaks.

"similar like" should be "similar fashion like"

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,226 @@
...
+The target neurons are specified by a list of global neuron ids which must be
+passed via
+the "targets" parameter. The music_cont_out_proxy will be connected
+automatically to the
+specified target neurons. It is not possible to change the list of target
+neurons or observed
+quantities once they have been set or the simulation has been started for the
+first time.
+
+In case of multiple recordables the data can be read out of the receiving buffer
```
jougs: This section must clarify that the code below only works in PyNEST. This documentation is also available from SLI.

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,226 @@
...
+remote applications. It works in a similar like a multimeter model. The user
+has to specify the recordable values to observe (e.g. ["V_m"]) via the
+record_from parameter.
+The target neurons are specified by a list of global neuron ids which must be
+passed via
+the "targets" parameter. The music_cont_out_proxy will be connected
+automatically to the
+specified target neurons. It is not possible to change the list of target
+neurons or observed
+quantities once they have been set or the simulation has been started for the
```
jougs: Currently, these lists can be set as often as wished. But not after `Simulate` has been called (which is the correct behavior in my view). The documentation needs to reflect this behavior.

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,226 @@
...
+first time.
+
+In case of multiple recordables the data can be read out of the receiving buffer
+via the following access pattern (like a 2-dimensional array in the C language
+):
+buffer[ target_gid_index ][ recordable_index] = buffer[ target_gid_index *
+record_from.size() + recordable_index ]
+For example:  if target_gids = [ 2, 5, 4] and record_from = ["V_m"], then if we
```
jougs: "then if" should be "and"

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,226 @@
...
+want to get "V_m" for neuron with GID 5:
+buffer[ 1*1 + 0 ]
+
+
+
+Parameters:
+The following properties are available in the status dictionary:
+
+interval     double   - Recording interval in milliseconds
+targets 	 array    - Global id list of neurons to be observed
```
jougs: Fix indentation

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,226 @@
...
+FirstVersion: March 2016
+Availability: Only when compiled with MPI and MUSIC
+
+SeeAlso: music_cont_in_proxy, music_event_out_proxy, music_event_in_proxy,
+music_message_in_proxy
+*/
+
+namespace nest
+{
+// class Network;
```
jougs: Remove commented out forward-declaration

jougs:
I've reviewed the code. In general I quite like it, but the issues I raised need to be addressed. I'll look at the example later.

gtrensch:
When running the music_cont_out_proxy_example MUSIC causes Open MPI to terminate with:

>> Error in MUSIC library: rank #1: Setup constructor was called a second time.
>> MPI_ABORT was invoked on rank 1 in communicator MPI_COMM_WORLD with errorcode 1

I have Open MPI 1.6.5 installed. MUSIC seems to be untested with it. The change furthermore requires MUSIC version 1.1.15. It is probably a compatibility issue. I will look into that.

uahic:
Ah the issue here is that I've used pyMUSIC and pyNEST at the same time which is not supported by MUSIC as the Setup object is globally unique and only once allocable. I put a pull request to fix that on MUSIC but it has a minor bug (warning when shutting down MUSIC) but functionally it works. When I worked over the test example I totally forgot that pyMUSIC and PyNEST does not work along without this fix.

gtrensch:
Thanks for that hint. I applied your MUSIC-change from the branch below and re-run the example.
* multi_setups_common_structure
* https://github.com/INCF/MUSIC/pull/23

Technically it seems to work now. Unfortunately, MPI issues following messages reporting a violation of the MPI standard.

>> *** The MPI_Comm_free() function was called after MPI_FINALIZE was invoked.
>> *** This is disallowed by the MPI standard.
>> *** The MPI_Group_free() function was called after MPI_FINALIZE was invoked.
>> *** This is disallowed by the MPI standard.
>> *** Your MPI job will now abort.

Question: Which MPI version do you use ?

I furthermore saw that your MUSIC pull request is from April and did not successfully compile without MPI on Travis. Could this be the reason why it has not been merged yet ?

In general, this PR introduces some dependencies which need to be resolved, especially a new MUSIC version !

uahic:
@gtrensch 
I am using OpenMPI 1.6.5 and a pretty busy in the next weeks; The code was actually more to provide something such that Mikael and I can work something out based on that. I will provide another example, the proxy neuron itself does not depend on the changes to MUSIC and so this pull request should not - you're totally right there. 

In case of interest:
I've tried to made as little changes as possible to the MUSIC code to got the current PR working (for own purposes), the issue you've encountered is the way the MPI-Communicator is allocated in my proposal. In order to allow shared Setup instances, there must be either shared data-structures among the instances or using something like the singleton pattern (whereas the latter one is not so desireble). The shared datastructure is currently implemented as structure which gets assigned to a static pointer when creating the first Setup instance, so When NEST terminates it shuts down MPI the shared communicator probably still exists.

gtrensch:
I got  in contact with @mdjurfeldt. He will look after the MUSIC pull request ( https://github.com/INCF/MUSIC/pull/23 ).

apeyser:
@uahic @gtrensch 
That's what I thought here: because NEST and MUSIC attempt to abstract away the MPI Communicator, it becomes impossible to properly get them to play nicely in extended situations.

This will only be worse if a third library becomes involved which also tries to abstract away something so important. 

Instead, we should be exposing them, rather than hacking fixes as needed here. Ownership of the connection belongs to the program, and not to the libraries (NEST or MUSIC) --- and in fact, I can easily imagine situations where I'd want to have a more sophisticated MPMD setup than here, with MUSIC and NEST belong to a subcommunicator themselves.

We had an issue on exposing these structures earlier that was closed for lack of interest.

mdjurfeldt:
@apeyser I think we have two problems here:

1. MPI Communicator: The only abtraction in MUSIC, here, is its use of MPI_COMM_WORLD to manage communication between MUSIC-aware application. I think this abstraction is good, and it is actually exploited in recent MUSIC versions in that this is handled differently for different communication algorithms. I buy your argument regarding more sophisticated MPMD setups, though. This need could be served by adding an optional argument to the Setup constructor such that an arbitrary communicator can be passed to MUSIC for inter-application-communication.

2. Initialization and handling of the MUSIC Setup object: When discussing MUSIC use together with PyNEURON, Michael Hines suggested that multiple initializations of MUSIC should be possible (=> multiple Setup objects) but would all end up using the same data structures. This would make for a convenient, implicit, "synchronization" of simultaneous use of PyNEST and PyMUSIC. This would then prevent the simultaneous use of multiple instances of MUSIC in the same application, but the question is if that would make any sense. The alternative would be to explicitly handle the setup object between NEST and pyMUSIC, but this would probably introduce a lot of complications for example relating to order of initialization. I'm happy to discuss this now. However, the quick solution to @uahic s problem would be the Hines solution.

uahic:
@apeyser 
From the perspective of a application developer I agree

@mdjurfeldt 
I already implemented the solution proposed by Hines, however it probably can be done in a cleaner way (see Pull Requests in the MUSIC repository)

mdjurfeldt:
To prevent behind-the-scenes changes to the semantics of the API, one could introduce the Hines semantics through two new calls in the MUSIC API:

MUSIC::init (Setup*& setup, int& argc, ...)
MUSIC::finalize (Setup*& setup)

which would then replace "new Setup (...)" and "delete setup".

That would enable init to actually create only a single Setup object instance and pass it to everybody calling "init". An attempt to create multiple Setup instances using "new Setup (...)" would still throw an error.

mdjurfeldt:
@uahic Yes, sorry, I'm aware of that and should have referred to it more properly! For various reasons I have not had time to look at it but am doing it now. BTW, when you say that you agree with @apeyser regarding abstraction, does this mean that you differ with what I say in my response? I'm very interested in finding a good solution.

uahic:
Having more control over the communicators allows building up own communication channels that are parallel to those of MUSIC but not dependent on the simulator (and schedule), and other rather exotic use-cases like spawning MUSIC simulations as a service but involving the spawning process as a simulator itself - the master process only offers nest/PyNN devices for collecting data, transforming data and sending data - (we have this situation like that in the NRP of HBP-SP 10 right now, but chose another way to achieve this). I think for most of the users these features are not so important.

apeyser:
@mdjurfeldt 
1) Then we essentially agree. A default use of WORLD is fine, but being able to pass in your own communicator would be open to unexpected uses.
2) Using a singleton is a "quick fix" --- but I predict will lead to the same problem of closing off unexpected use cases. If there are no globals, and the COMM can be passed in, then there's no need for singletons. In the same way, NEST should by default create a MUSIC instance, but should be able to accept a MUSIC instance passed in.

In essence, my position is to use reasonable defaults with the option to set the internal by hand. Like a car, have a nice hood but make it possible to open it and change the oil.

mdjurfeldt:
@uahic OK, I see. But the thing is that inside MUSIC, such channels might be looking very different depending on the MUSIC configuration information. Therefore, I think such needs are best served by introducing API calls requesting from MUSIC a specific kind of communicator, e.g. "give me an intercommunicator for applications named A and B". That way, MUSIC can ensure that this exists, regardless if it has been created anyway or not. Such extensions need discussion and thinking, though.

mdjurfeldt:
@apeyser Regarding your 2): Being able to pass in a COMM unfortunately does not solve the singleton problem---at least not in a straightforward way. This is because there is a lot more data structures than the COMM which are being used for handling MUSIC ports.

apeyser:
@mdjurfeldt 
Re 2), you're right. What I was (poorly) trying to say was that just as MUSIC encapsulates MPI_COMM, NEST encapsulates both MPI_COMM and MUSIC. Those need to be exposed.

It would also simplify some of the NEST internals, since MUSIC is a special case through out the code.

uahic:
I think a good way to go would be a mix of the proposed solutions which is to adjust NEST such that it returns its MUSIC Setup instance and MUSIC to have an alternative constructor to accept this. Internally they would share the same data structure but this structure must not be a global then. If you agree on that I would do the MUSIC related part

uahic:
It seems that an error occurs which does not seem to be related to the code changes:

could somebody with sufficient privileges please have a look at this?

jougs:
Caused by a bug in Travis: https://github.com/travis-ci/packer-templates/issues/221

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,416 @@
...
+    && ( d->known( names::interval ) || d->known( names::record_from ) ) )
+  {
+    throw BadProperty(
+      "The recording interval and the list of properties to record "
+      "cannot be changed after the index_map has been set." );
+  }
+
+
+  double_t v;
+  if ( updateValue< double_t >( d, names::interval, v ) )
```
jougs: Remove `double_t` by `double`

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,416 @@
...
+      throw BadProperty(
+        "The property record_from must be set before passing target_gids." );
+    }
+
+    if ( state.published_ == false )
+    {
+      ArrayDatum mca = getValue< ArrayDatum >( d, names::targets );
+      for ( Token* t = mca.begin(); t != mca.end(); ++t )
+      {
+        target_gids_.push_back( getValue< long_t >( *t ) );
```
jougs: `long_t` → `long`

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,416 @@
...
+  // only publish the output port once,
+  if ( S_.published_ == false )
+  {
+    const Token synmodel =
+      kernel().model_manager.get_synapsedict()->lookup( "static_synapse" );
+    assert(
+      synmodel.empty() == false && "synapse 'static_synapse' not available" );
+
+    const index synmodel_id = static_cast< index >( synmodel );
+    std::vector< long_t >::const_iterator t;
```
jougs: `long_t` → `long`

models/music_cont_out_proxy.cpp
```diff
@@ -0,0 +1,416 @@
...
+void
+nest::music_cont_out_proxy::set_status( const DictionaryDatum& d )
+{
+  P_.set( d, *this, S_, B_ ); // throws if BadProperty
+}
+
+void
+nest::music_cont_out_proxy::update( Time const& origin,
+  const long_t from,
+  const long_t )
```
jougs: `long_t` → `long`

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,215 @@
...
+  void finalize();
+
+  /**
+   * Collect and output membrane potential information.
+   * This function pages all its targets at all pertinent sample
+   * points for membrane potential information and then outputs
+   * that information. The sampled nodes must provide data from
+   * the previous time slice.
+   */
+  void update( Time const&, const long_t, const long_t );
```
jougs: `long_t` → `long`

models/music_cont_out_proxy.h
```diff
@@ -0,0 +1,215 @@
...
+  struct Parameters_
+  {
+    Parameters_(); //!< Sets default parameter values
+    Parameters_(
+      const Parameters_& ); //!< Copy constructor for parameter values
+
+    Time interval_;                   //!< sampling interval, in ms
+    std::string port_name_;           //!< the name of MUSIC port to connect to
+    std::vector< Name > record_from_; //!< recordables to record from
+    std::vector< long_t > target_gids_; //!< Neuron GIDs to be observed
```
jougs: `long_t` → `long`

uahic:
@jougs Changed long_t and double_t to the standard type names

gtrensch:
:+1: music_cont_out_proxy_example successfully tested with music v1.1.15 ! There is no dependency on the custom music branch any more.

gtrensch:
@uahic 
In the course of setting up automated MUSIC tests for NEST (PR #471) a SLI version of your music_cont_out_proxy_example is required. It would be fantastic if you could provide a SLI example.

uahic:
@gtrensch 
I am willing to provide the SLI example/test case, however before digging deeper into SLI (which I have never worked with), I need a little hint from you: How do I add a custom neuron to the SLI models dict? Actually the music_cont_out_proxy is already in the modelsmodule ( within models folder) and I dont see another place where models need to be registered. Does the model miss any declaration/method or is SLI unhappy with the status dictionary arguments? Because SLI tells me that music_cont_out_proxy is unknown

jougs:
@uahic: Can you please provide the exact error messages? Just from your verbal description it does not become clear what the problem is. I.e. what are the SLI commands your running and what's the error message you receive? From the code it looks as if everything should work. And if it does fro Python, this hints at a problem with your SLI code.

uahic:
`SLI ] /music_cont_out_proxy Create /bla Set

Sep 16 13:57:41 Create_l_i [Error]: UnknownModelName
    /music_cont_out_proxy is not a known model 
    name. Please check the modeldict for a list
    of available models.
`

jougs:
I just compiled your branch and calling `/music_cont_out_proxy Create` works just fine. So your analysis is probably correct.

jougs:
Please note that `Create` in SLI does not return the range of IDs as it does in PyNEST, but just the ID of the last created node. This might lead to unexpected results if you create multiple nodes in a single call.

uahic:
@jougs Thanks for the hint! And yep, after re-compiling it works. Next week there should be some time to spare for creating the SLI test's/examples.

uahic:
I've wrote half of the unit-test and hopefully will be able to deliver until the end of the week, currently I am pretty busy with project work and want to excuse that I do not have enough time right now to do code reviews

gtrensch:
This pull request has been replaced by PR #798 and can be closed.

PR #283, opened 2016-03-22
Fix: Uninitialized variable in three models

obreitwi:
`<model>::Parameters_::num_of_receptors_` is uninitialized for the three
models:

 * `aeif_cond_alpha_multisynapse`
 * `iaf_psc_alpha_multisynapse`
 * `iaf_psc_exp_multisynapse`

By chance it seems to always be 0 in the current master branch. However,
when compiling modified source code, it can assume aribtrary values.
This leads to potential SEGFAULTs when the `calibrate()` method is
called on an unchanged node.


heplesser:
@obreitwi Thanks for catching this! :+1: and merging.

Issue #240, opened 2016-02-29
Weights in stdp_triplet_connection cannot be negative - these cannot be used for inhibitory synapses?


sanjayankur31:
I may be missing something here, but from the examples that I've come across, if one wants to use synapse models as inhibitory, one just provides a negative weight. Now, in the implementation of the stdp_triplet_connection, the weights are bounded to (0, Wmax) - they cannot be negative. Does this mean they cannot be used for inhibitory synapses at all?

https://github.com/nest/nest-simulator/blob/master/models/stdp_triplet_connection.h#L195

``` C++
  inline double_t
  depress_( double_t w, double_t kminus, double_t Kplus_triplet_ )
  {
    double_t new_w = w - kminus * ( Aminus_ + Aminus_triplet_ * Kplus_triplet_ );
    return new_w > 0.0 ? new_w : 0.0;
  }
```

If a synapse is to be inhibitory, the Wmax value will also be negative. So, changing the check to something like this is enough.

``` C++
  inline double_t
  depress_( double_t w, double_t kminus, double_t Kplus_triplet_ )
  {
    double_t new_w = w - kminus * ( Aminus_ + Aminus_triplet_ * Kplus_triplet_ );
    return (new_w / Wmax_ )> 0.0 ? new_w : 0.0;
  }
```

This ensures that w and Wmax are of the same sign.

sanjayankur31:
@flinz  - I ran into this while working on the symmetric stdp pull request - what do you think?

flinz:
@sanjayankur31 
The triplet learning rule is only intended for positive weights (reflecting the amount of neurotransmitter released). In general, this also seems to be the case for example for https://github.com/nest/nest-simulator/blob/master/models/stdp_connection.h, in which no negative weights are allowed

So far I was of the opinion that negative weights are a workaround in NEST when multiple spike receptors are not being used, however it seems that most models rely indeed in positive and negative weights. So, regardless of a debate whether the STDP models in question are at all viable for models of inhibitory plasticity, it would probably be worth doing some minor changes to enable also negative weights for STDP connections.

There are three options as far as I see it:

1. Allow negative weights for also the triplet model (amongst others): then all constants for depression and facilitation (``A_minus`` or ``A_plus``) would have to switch sign (since, facilitation on a negative weight should also decrease it). In addition one would have to check to prevent sign switching of weights. I do not favor this method.
[...]

PR #284, opened 2016-03-23
Enables inhibitory stdp

weidel-p:
With this pull request, STDP is possible for inhibitory connections by assigning negative values for `weight` and `Wmax` in STDPConnection and STDPTripletConnection.
The functions `facilitate` and `depress` assumes positive weight changes, but the actual sign of the weight change is applied in the `send` function.
The sign of `weight` determines the sign of `Wmax` during connection and SetStatus. During simulation, the sign of `Wmax` determines the sign of `weight`.

This addresses #240.


weidel-p:
I suggest @abigailm, @mhelias and/or @suku248 as reviewers.

models/stdp_connection.h
```diff
@@ -185,7 +187,7 @@ class STDPConnection : public Connection< targetidentifierT >
   {
     double_t norm_w =
       ( w / Wmax_ ) - ( alpha_ * lambda_ * std::pow( w / Wmax_, mu_minus_ ) * kminus );
-    return norm_w > 0.0 ? norm_w * Wmax_ : 0.0;
+    return norm_w > 0.0 ? norm_w * Wmax_ : copysign( 0.0, Wmax_ );
```
heplesser: `copysign( 0.0, Wmax_ )` returns a value with the magnitude of its first and the sign of its second argument. So here we would get +0 og -0. Does that make sense?

heplesser:
@weidel-p The `copysign()` function is part of C++11. Since we cannot rely on C++11 support on all relevant compilers, we need a configure test for whether `copysign()` is available and if not, we need to define it ourselves, e.g., in `libnestutil.numerics.h`.

models/stdp_connection.h
weidel-p: With this line, I wanted to make sure that the sign of `weight` does not flip from negative to positive if it becomes 0. After having another look into this and after testing, I agree that it was unnecessary.
I fixed that in 0ebe4ba3d9d112bf1af7d999f3bd3c0f1789b0ad.

weidel-p:
@heplesser  `copysign()` is also available in C99. In 0ebe4ba3d9d112bf1af7d999f3bd3c0f1789b0ad I changed the include to use the C header. Is this already sufficient to make sure all relevant compilers support this or shall I include the function to `libnestutil.numerics.h` as you suggested?

heplesser:
@weidel-p I tried
```
#include <math.h>
#include <iostream>

int main()
{
  double x = 10;
  double y = -2;

  double z = copysign(x, y);

  std::cout << x << '\t' << y << '\t'  << z << std::endl;
  
  return 0;
}
```
using xlc++, Intel, Portland and Clang compilers. It works everywhere. The only system I cannot test is K. Could you ask our K-colleagues to test it?

jakobj:
with FCCpx and mpiFCCpx @heplesser's script works also on K. no compiler warnings/errors and the output reads `10      -2      -10` which is correct as far as I can see.

heplesser:
@jakobj Thanks for checking on K!
@weidel-p Could you make a tiny, inconsequential change to trigger a new build? The last build failed partially due to time-outs. Conditional on all test passing :+1: from my side, but one of the plasticity experts should approve before merging.

sanjayankur31:
Just a thought - Is using copysign to copy the sign of weight to Wmax under the hood a good idea? Wouldn't it be better to have a bad parameter check (or something) that checks if the two signs are same instead - let users manually set both the signs so that they're aware of exactly what they're doing?

models/stdp_triplet_connection.h
```diff
@@ -185,7 +187,7 @@ class STDPTripletConnection : public Connection< targetidentifierT >
   facilitate_( double_t w, double_t kplus, double_t ky )
   {
     double_t new_w = w + kplus * ( Aplus_ + Aplus_triplet_ * ky );
-    return new_w < Wmax_ ? new_w : Wmax_;
+    return new_w < std::abs( Wmax_ ) ? new_w : std::abs( Wmax_ );
```
sanjayankur31: Won't implementation along these lines permit sign switching? As in an excitatory connection reducing, going negative and turning into an inhibitory one and vice versa? 

Example - start with 5,500 - depress and turn into -5, 500 -> check still passes?
Example - start with -5, -500 - facilitate and turn into 5,-500 -> check still passes?

heplesser: Looking at this code again after a comment from @sanjayankur31 which is strangely missing from Github, I find the code very difficult to read. The problem is that on l. 254f you call call `facilitate_()` with the absolute value of `w`, something that is not clear from the code here. Basically, you mix handling of signs in `send()` and `facilitate_()`. I would suggest to do all signe handling inside `facilitate_()`, so that the code in `send()` would again become
```
weight_ = facilitate_( weight_, Kplus_ * std::exp( minus_dt / tau_plus_ ), ky );
```
That way, all sign handling will be inside `facilitate_()`, which we then could code like this
```
const double_t new_w_abs = std::abs( w ) + kplus * ( Aplus_ + Aplus_triplet_ * ky );
return copysign( std::min( new_w_abs, std::abs( Wmax_ ) ), Wmax_ );
```
I suppose the same applies to `depress_()`.

sanjayankur31: Oh, sorry - I removed the comment - I wasn't quite sure if what I was asking was correct - hadn't worked it out on paper yet.

heplesser:
@sanjayankur31 I suppose in your comment on changing the sign under the hood you refer to `STDPConnection::set_weight()` and `STDPConnection::set_status()`. I agree that it would make sense to introduce a test in both places, raising an exception if one tries to set a weight (or weight limit) that is not consistent with the sign. Zero should in this case work for both signs. There should also be a test ensuring that this check is in place.

This would stop users from subtle and hard-to-find errors.

tammoippen:
@flinz Can you please look into this PR? Thank you.

models/stdp_triplet_connection.h
flinz: @weidel-p I agree with @heplesser and @sanjayankur31 that the handling of the sign issues for weight updates should be put into ``facilitate`` and ``depress`` respectively.

pynest/nest/tests/test_stdp_triplet_synapse.py
```diff
@@ -84,11 +85,11 @@ def decay(self, time, Kplus, Kplus_triplet, Kminus, Kminus_triplet):
 
     def facilitate(self, w, Kplus, Kminus_triplet):
         """Facilitate weight."""
-        return w + Kplus * (self.syn_spec["Aplus"] + self.syn_spec["Aplus_triplet"] * Kminus_triplet)
+        return abs(w + Kplus * (self.syn_spec["Aplus"] + self.syn_spec["Aplus_triplet"] * Kminus_triplet))
```
flinz: After incorporating the changes proposed by @heplesser in https://github.com/nest/nest-simulator/pull/284#discussion_r58529377 the test functions for ``facilitate`` and ``depress``, and calls to them, should be adapted accordingly.

flinz:
After the line comments are addressed, 👍 from me.

heplesser:
@weidel-p #391 is merged and Travis works again. Could you merge all changes from master into your PR so that Travis hopefully turns green again?

weidel-p:
Hopefully I addressed all comments with the latest commit. 

1. The signs are now exclusively handled in the `facilitate` and `depress` functions.
2. An exception is thrown if `Wmax` and `weight` are initialized with different signs. This ensures that users get an immediate feedback if `weight` and `Wmax` are set inconsistently.
3. If the issue solved by #333 is reintroduced again, the tests for #284 will fail.

models/stdp_connection.h
```diff
@@ -68,11 +68,12 @@
 
   FirstVersion: March 2006
   Author: Moritz Helias, Abigail Morrison
+  Adapted by: Philipp Weidel
   SeeAlso: synapsedict, tsodyks_synapse, static_synapse
 */
 
 // C++ includes:
-#include <cmath>
+#include <math.h>
```
heplesser: Comment why you include `math.h`.

models/stdp_triplet_connection.h
```diff
@@ -76,10 +76,11 @@
 
   FirstVersion: Nov 2007
   Author: Abigail Morrison, Eilif Muller, Alexander Seeholzer, Teo Stocco
+  Adapted by: Philipp Weidel
   SeeAlso: stdp_triplet_synapse_hpc, synapsedict, stdp_synapse, static_synapse
 */
 
-#include <cmath>
+#include <math.h>
```
heplesser: Comment why you include `math.h`.

testsuite/unittests/test_inh_stdp_synapse.sli
```diff
@@ -0,0 +1,245 @@
...
+      }
+      ifelse
+    }
+    ifelse
+  }
+  ifelse
+}
+loop
+
+%final_weight w w_max mul sub abs dup = 1e-13 leq =
```
heplesser: Remove this commented-out line.

testsuite/unittests/test_inh_stdp_synapse.sli
```diff
@@ -0,0 +1,245 @@
...
+final_weight w w_max mul sub abs 1e-13 leq assert_or_die
```
heplesser: It would be more in keeping the testsuite style to put the entire loop from l 185 on up to and including the `leq` into a block and the place the `assert_or_die` after the block. Also, there is a SLI function `ToUnittestPrecision` (or similar) that allows testing with a prescribed number of decimals.

heplesser:
@weidel-p I fear that in all the merging, the sign-handling has disappeared from `stdp_connection.h` again, at least there is no trace of `copysign()` left in that file. Could you double-check this?

models/stdp_connection.h
weidel-p: We were afraid that `copysign` is not available in the c++ header for some supercomputers. Therefore, I changed it to the C header and we tested this for several compilers. Please read the comments above.

models/stdp_triplet_connection.h
weidel-p: see above.

testsuite/unittests/test_inh_stdp_synapse.sli
weidel-p: This test is basically copied from `test_stdp_synapse` and adapted for negative signs. Therefore, I tried to modify both tests according to your comments. I am now using `ToUnittestPrecision`  and used a block for the loop. You didn't specify where to put the loop, so I left it where it is.

weidel-p:
@heplesser in `stdp_connection.h` no `copysign()` is needed. In `facilitate` and `depress`, the `weight` is normalized between 0 and 1 and afterwards scaled with `Wmax`. This works for both, negative and positive weights.
As stated in the initial post, this PR addresses #240. All work done on the `stdp_connection.h` is just ensuring consistent behavior between the `stdp_triplet_synapse` and `stdp_synapse` (throwing the same exception).

models/stdp_connection.h
heplesser: @weidel-p Yes, I know. But that information should be here in the C++ file, so that if someone reads this file five years from now will understand. E.g. `// C-header since copysign() is in C99 but not C++98`.

heplesser:
@weidel-p Okay, I overlooked that `stdp_connection` doesn't need `copysign()`. But why does it then include `math.h` instead of `cmath`?

heplesser:
@weidel-p The tests look fine now.

models/stdp_connection.h
weidel-p: Right, I added this comment to the cpp file in the `stdp_triplet_connection.h`.

weidel-p:
@heplesser Sorry, I used `copysign()` in `stdp_connection.h` in an earlier version and forgot to change the include back to `cmath`.

heplesser:
:+1: and merging. @weidel-p Thanks for you effort!

PR #288, opened 2016-03-28
fix compilation on K

jakobj:
this PR just adds one line that must have been accidentally deleted during refactoring. with this line added, compilation of current master works fine on K.


heplesser:
:+1: and merging.

PR #289, opened 2016-03-29
hpc_benchmark fix and cleanup

jakobj:
this PR fixes an issue that lead to poor scaling pf the `hpc_benchmark.sli` example. the issue was the use of `[E_from E_to] Range` in randomizing membrane potentials of neurons. this would allocate memory for an array containing the gids of the entire network and loop over this list on every process(!), which quickly becomes an issue for large scale benchmarks (in terms of time and memory). it is replaced by `E_net GetLocalNodes`.
in addition I have removed output via `=` which happens on every process. all output should happen via the logger, which restricts the number of processes that generate output.
I have added a rate calculation function, which was shamelessly copied from `brunel-2000.sli`.
minor changes in this PR:

- removed some whitespace at end of lines
- removed some blank lines
- fixed indentation
- replaced [unit] with (unit)
- minor addition to documentation
- changed default scale to 2 (0.2 was pointless since this amounts to a five-times fully connected network)

### DIFF
--- a/examples/nest/hpc_benchmark.sli
+++ b/examples/nest/hpc_benchmark.sli
@@ -107,13 +107,13 @@ def
   /model_params
   <<
     % Set variables for iaf_neuron
-    /E_L     0.0  mV  % Resting membrane potential [mV]
-    /C_m   250.0  pF  % Capacity of the membrane [pF]
-    /tau_m  10.0  ms  % Membrane time constant [ms]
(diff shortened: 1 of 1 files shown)


jakobj:
oh, right, forgot: I nominate @mschmidt87 and @suku248 as reviewers.

mschmidt87:
Nice work. :+1:

suku248:
Thanks @jakobj and :+1: for merging

PR #292, opened 2016-04-05
New user documentation in markdown

steffengraber:
These are the static pages of user documentation.  In the future, they will serve as a basis of the creation of the website and the documentation in the source code.

From now on changes in the documentation and adding new texts can be done here and editing of documents is as easy as creating code.


extras/userdoc/md/documentation/analog_recording_with_multimeter.md
```diff
@@ -0,0 +1,98 @@
...
+    Out[24]:
+    array([-70.        , -70.        , -70.        , -70.        ,
+           -70.        , -70.        , -70.        , -70.        ,
+
+An example
+----------
+
+As an example, here is the multimeter.py example from the PyNEST examples set:
+
+     prettyprint
```
heplesser: Why does `prettyprint` have an extra space in front of it?

heplesser:
@steffengraber I think this is very nice! I would propose to merge this rather quickly without detailed review of the content, since it is the same material that is on our website today. We can then rather open issues for fixing specific problems.

extras/userdoc/md/documentation/analog_recording_with_multimeter.md
steffengraber: Only an artefact. I have deleted 'prettyprint'.

jougs:
I agree with @heplesser that we should merge this immediately in order to get the whole infrastructure in place and allow people to contribute to the homepage via GitHub. I'll thus merge once the tests have completed.

jougs:
@steffengraber: very nice work, btw!

heplesser:
I will merge this now, we can then iterate in the repo.

@steffengraber Could you send a mail to our mailing lists explaining the purpose of the usedoc in the repo? Maybe we should also add some guidelines/info on the Developerspace web pages?

PR #294, opened 2016-04-05
Remove connection_manager source files

### LINKED ISSUE #293: Remove deprecated connection_manager sources
Hiya,

Just going through the source - I notice connection_manager.{h,cpp} are not to be used - can these please be removed from the tree completely. I assume they were kept around while the move to cmake was being made - now that the move is complete, there's no reason to keep them around. Folks can always checkout an older commit if they want to refer to these files.

sanjayankur31:
These are deprecated and not to be used any more.  Fixes #293

Checked to confirm that nothing references the header:

```
$ grep -nri "connection_manager.h" *
nestkernel/connection_manager.h:2: *  connection_manager.h
nestkernel/connection_manager.h:23:#ifndef CONNECTION_MANAGER_H
nestkernel/connection_manager.h:24:#define CONNECTION_MANAGER_H
nestkernel/connection_manager.h:42:#error Do not use connection_manager.h in any place in your nest code!
nestkernel/connection_manager.h:250:#endif /* #ifndef CONNECTION_MANAGER_H */
nestkernel/connection_manager.cpp:23:#include "connection_manager.h"
```


jougs:
@tammoippen, can you please check if these files are left-overs from the nest3 refactoring? :+1: from my side if this is the case.

heplesser:
:+1: from my side as well, but I will leave the last word to Tammo.

tammoippen:
@jougs @heplesser I prepared a similar PR and now joint forces. Can you please look again into the changes. Basically a renaming `ConnectionBuilderManager` -> `ConnectionManager` took place.

heplesser:
In view of previous :+1: and the confirmation from Tammo, I will merge this one.

PR #296, opened 2016-04-07
TravisCI: Count Errors and Warnings.

tammoippen:
This PR extends the parsing of TravisCI compiler output for number of warnings and errors.

### DIFF
--- a/extras/parse_travis_log.py
+++ b/extras/parse_travis_log.py
@@ -296,7 +296,7 @@ if __name__ == '__main__':
     print_static_analysis(static_analysis)
 
     print("\n\nWarnings:")
-    for k,v in actual_warnings.iteritems():
+    for k, v in actual_warnings.iteritems():
         print(" - {}: {}".format(k, v))
     print("--------<<<<<<<< Summary of TravisCI >>>>>>>>--------")
 
--- a/extras/parse_travis_log.py
+++ b/extras/parse_travis_log.py
@@ -72,7 +72,6 @@ def process_changed_files(f):
         if not line:
             return False, line
 
-        # +file_names=extras/scan_travis_log.py nestkernel/connector_model.h nestkernel/nestmodule.cpp nestkernel/network.cpp 
         if line.startswith('+file_names='):
             return filter(lambda x: x != '', line.strip().split('=')[1].split(' ')), line
 
@@ -92,7 +91,6 @@ def process_vera(f, filename):
             return res
 
         if line.startswith(filename):
-            # "nestkernel/network.cpp:107: full block {} expected in the control structure"
             key = line.split(":")[-1].strip()
             if not d.has_key(key):
                 d[key] = 0
@@ -153,7 +151,6 @@ def process_static_analysis(f, line):
     """
     Process static analysis output for a certain file.
     """
-    # Static analysis on file nestkernel/archiving_node.cpp:
     filename = line.split(' ')[-1].strip()[0:-1]
     d = {}
     res = {filename: d}
@@ -205,7 +202,6 @@ def count_warnings_errors(f):
         if line.strip() == '+make install':
             return warn, error, line
 
-        # /home/travis/build/tammoippen/nest-simulator/build/libnestutil/config.h:65:0: warning: "HAVE_DIRENT_H" redefined [enabled by default]
         if ': warning:' in line:
             file_name = line.split(':')[0]
             if file_name not in warn:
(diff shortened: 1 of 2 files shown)


extras/parse_travis_log.py
```diff
@@ -191,13 +191,41 @@ def print_static_analysis(d):
...
+    error = {}
+    while True:
+        line = f.readline()
+        if not line:
+            return warn, error, line
+
+        if line.strip() == '+make install':
+            return warn, error, line
+
+        # /home/travis/build/tammoippen/nest-simulator/build/libnestutil/config.h:65:0: warning: "HAVE_DIRENT_H" redefined [enabled by default]
```
heplesser: Should this comment be removed?

extras/parse_travis_log.py
```diff
@@ -191,13 +191,41 @@ def print_static_analysis(d):
...
+        if ': warning:' in line:
+            file = line.split(':')[0]
+            if file not in warn:
+                warn[file] = 0
+            warn[file] += 1
+
+        if ': error:' in line:
```
heplesser: I am not really fond of a string constant here, as compilers may change. But gcc 4 and 5, clang, Intel and Portland Group compilers all include this string in their error messages, so I guess it is a rather safe bet. `xlC` will not work, but I think that is not a problem for checking on Travis.

heplesser:
@tammoippen Looks mostly fine, but I have two issues. 

First, CMake hides warnings from us, except those that come from the pynest build process. I think we  need to use `make VERBOSE=1` to see the warnings.

Second, towards the end of the build log, I see
 ```
Built target installcheck
+[  !=  ]
+[ 296 != false ]
+echo WARNING: Not uploading results as this is a pull request
WARNING: Not uploading results as this is a pull request
+exit 0
```
What are the two lines with the square brackets and inequality doing there?

**Comment by tammoippen** [comment]
@heplesser  The two lines are generated by [`build.sh`](https://github.com/nest/nest-simulator/blob/master/build.sh#L175).

File: extras/parse_travis_log.py (hunk 1 above)
Comment (tammoippen) [inline extras/parse_travis_log.py]: Also in the other parse functions, I have the line I want match included as a comment. For consistency, I should remove them there as well.

File: extras/parse_travis_log.py (hunk 2 above)
Comment (tammoippen) [inline extras/parse_travis_log.py]: I know what you mean, but as the purpose of this script is to parse TravisCI output, i think it is ok to assume gcc 4 and 5.

**Comment by tammoippen** [comment]
I put this PR on a hold, as the warnings are not parsed... I have to investigate.

**Comment by tammoippen** [comment]
@heplesser Can you please have another look?

**Comment by heplesser** [comment]
:+1:

**Comment by tammoippen** [comment]
@jougs Can you please look into this PR? Thank you.

File: extras/parse_travis_log.py [hunk 3]
```diff
@@ -191,13 +191,41 @@ def print_static_analysis(d):
...
+        line = f.readline()
+        if not line:
+            return warn, error, line
+
+        if line.strip() == '+make install':
+            return warn, error, line
+
+        # /home/travis/build/tammoippen/nest-simulator/build/libnestutil/config.h:65:0: warning: "HAVE_DIRENT_H" redefined [enabled by default]
+        if ': warning:' in line:
+            file = line.split(':')[0]
```
jakobj: i'm fairly unhappy about the choice of using `file` as a variable name here, as this is also a Python object. I would use `file_name` instead.

extras/parse_travis_log.py
jougs: Please remove them or make it explicit that they are the expected output. Thanks!

extras/parse_travis_log.py
```diff
@@ -241,26 +267,43 @@ def print_static_analysis(d):
...
     print("Make install:        " + ("Ok" if make_install_ok else "Error"))
     print("Make installcheck:   " + ("Ok (" if make_installcheck_failed == 0 else "Error (") +
           str(make_installcheck_failed) + " / " + str(make_installcheck_all) + ")")
     print("Logs uploaded to S3: " + ("Yes" if uploading_results else "No"))
 
     print("\nStatic analysis:" )
     print_static_analysis(static_analysis)
+
+    print("\n\nWarnings:")
+    for k,v in actual_warnings.iteritems():
```
jougs: There's a space missing between "k," and "v" for PEP-8 compliance.

jougs:
Nice work. :+1: after my comment is addressed

Issue #62, opened 2015-07-03
clang-format line length


apeyser:
@tammoippen 

100 is way, way too long. 80 should be the absolute maximum length of a line, and I'd argue for 70.

I know that some editors will tend to drive you towards long lines (Xcode) with line wrapping, but on that point, objectively, the lines should be single concepts, single tests, and the wrapping should be fixed by the programmer. A line like:
```
  if ( ( new_nmin < 0 && new_nmax > max + new_nmin ) || ( new_nmax - new_nmin == max ) )
```
is way, way too long.

Look at the default line lengths for TeX --- thin columns, and that's for literature.

heplesser:
@apeyser 

I agree that in the example above, code probably will be more readable by splitting it over two lines.

At the same time, I also find that 80 can be a bit on the short side, in particular for header lines in function definitions when using expressive class and method names, forcing one to split parameter lists over many lines close to the right margin. Not sure I like that.

apeyser:
I'd say: typedefs. If the names are so expressive that they're filling up long lines, the expressiveness in the header will make the logic of the source code difficult to read.

So, you split the baby: you keep the long names, but you use them as comments with typedefs in your source code, and then use locally readable very short names.
```
typedef IamComputing_embedding_some_strangeAlgorithm< UsingAnAllocator< ThatWorksFor< AFamilyOfClassesThatIdeclaredEarlyer> > > Alg;

int func(Alg alg, int x);
````

The experience from typesetting is unequivocal: optimal line length is 40-50 characters per line, never more than 70. So the 80 is an outside rule to allow some exceptional cases. Linux kernel:
>		Chapter 2: Breaking long lines and strings
>
> Coding style is all about readability and maintainability using commonly
> available tools.
>
> The limit on the length of lines is 80 columns and this is a strongly
> preferred limit.
>
> Statements longer than 80 columns will be broken into sensible chunks, unless
> exceeding 80 columns significantly increases readability and does not hide
> information. Descendants are always substantially shorter than the parent and
> are placed substantially to the right. The same applies to function headers
> with a long argument list. However, never break user-visible strings such as
> printk messages, because that breaks the ability to grep for them.

XCode's pattern is just wrong.

tammoippen:
I put a [branch](https://github.com/tammoippen/nest-simulator/tree/columnLimit) online with column limit set to 80: 

    Showing 323 changed files with 6,627 additions and 3,513 deletions.

[...]

PR #297, opened 2016-04-07
Format: limit column width to 80

tammoippen:
This is purely a reformatting of code as requested in #62.


heplesser:
:+1: from  my side. @apeyser Would you take a look as well?

apeyser:
:+1:

Silmathoron:
Sorry to reopen a merged PR, but fixing the conflicts with mine, I realized that there are still some lines longer 80, for instance in [modelsmodule.cpp](https://github.com/nest/nest-simulator/blob/master/models/modelsmodule.cpp).

heplesser:
@Silmathoron Can you pull the current master into your branch and then use clang-format-3.6 to reformat your code where necessary?

Silmathoron:
Yes I can, but I'm not sure we are talking about the same thing: I was saying that files on the ``nest/master`` branch still have lines longer than 80 (see previous link).

heplesser:
@tammoippen Could you clarify how to proceed here?

tammoippen:
@heplesser @Silmathoron 
`clang-format` makes an estimate on what is best: sticking to 80 column lines and readability (I guess at least), and it also has some problems with breaking some lines, e.g. it tries to break comments by whitespace, hence it has some problems with comments like:

    //-------------------------------------------------------------------------------------

So, the code is at most 80 columns for most lines, with some exceptions. Also see [here](http://llvm.org/releases/3.6.0/tools/clang/docs/ClangFormatStyleOptions.html) for **PenaltyExcessCharacter**.

PR #298, opened 2016-04-07
TravisCI: Check for PEP8

tammoippen:
This PR lets TravisCI check changed Python files for PEP8 compliance, as requested by #188. The file `extras/check_code_style.sh` now also checks for PEP8.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -152,16 +152,13 @@ for f in $file_names; do
 
       ;;
     *.py )
-      set +e # pep8 returns 1 if the file does not comply with PEP8
       echo "Check PEP8 on file $f:"
-      pep8_result=`pep8 --first --ignore=$PEP8_IGNORES $f`
 
-      if [ "$?" -gt "0" ]; then
+      if ! pep8_result=`pep8 --first --ignore=$PEP8_IGNORES $f` ; then
         echo "$pep8_result"
 
         format_error_files="$format_error_files $f"
       fi
-      set -e
       ;;
     *)
       echo "$f : not a C/CPP/PY file. Do not do static analysis / formatting checking."
--- a/build.sh
+++ b/build.sh
@@ -111,6 +111,9 @@ else
 fi
 format_error_files=""
 
+# Ignore those PEP8 rules
+PEP8_IGNORES="E121,E123,E126,E226,E24,E704"
+
 for f in $file_names; do
   if [ ! -f "$f" ]; then
     echo "$f : Is not a file or does not exist anymore."
@@ -151,7 +154,7 @@ for f in $file_names; do
     *.py )
       set +e # pep8 returns 1 if the file does not comply with PEP8
       echo "Check PEP8 on file $f:"
-      pep8_result=`pep8 --first $f`
+      pep8_result=`pep8 --first --ignore=$PEP8_IGNORES $f`
 
       if [ "$?" -gt "0" ]; then
         echo "$pep8_result"
--- a/build.sh
+++ b/build.sh
@@ -148,8 +148,20 @@ for f in $file_names; do
       fi
 
       ;;
+    *.py )
+      set +e # pep8 returns 1 if the file does not comply with PEP8
+      echo "Check PEP8 on file $f:"
+      pep8_result=`pep8 --first $f`
+
+      if [ "$?" -gt "0" ]; then
+        echo "$pep8_result"
+
+        format_error_files="$format_error_files $f"
+      fi
+      set -e
+      ;;
     *)
-      echo "$f : not a C/CPP file. Do not do static analysis / formatting checking."
+      echo "$f : not a C/CPP/PY file. Do not do static analysis / formatting checking."
       continue
   esac
 done
--- a/extras/check_code_style.sh
+++ b/extras/check_code_style.sh
@@ -6,6 +6,7 @@ END_SHA=HEAD
 
 CPPCHECK=cppcheck
 VERA=vera++
+PEP8=pep8
 
 CLANG_FORMAT=clang-format-3.6
 INCREMENTAL=false
@@ -57,6 +58,8 @@ Options:
                          TravisCI setup. (default=clang-format-3.6)
     --vera++=exe         Enter the executable that is used for vera++.
                          (default=vera++)
+    --pep8=exe           Enter the executable that is used for pep8.
+                         (default=pep8)
 EOF
 
     exit $1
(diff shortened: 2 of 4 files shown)


heplesser:
:+1: from my side. @apeyser, would you take a look, too?

build.sh
```diff
@@ -111,6 +111,9 @@ else
 fi
 format_error_files=""
 
+# Ignore those PEP8 rules
+PEP8_IGNORES="E121,E123,E126,E226,E24,E704"
```
apeyser: Should the E24 be E24x? I don't see an E24 in the pep8 script's documentation.
Where did this particular choice of pep8 exceptions come from?

build.sh
```diff
@@ -148,8 +148,20 @@ for f in $file_names; do
       fi
 
       ;;
+    *.py )
+      set +e # pep8 returns 1 if the file does not comply with PEP8
+      echo "Check PEP8 on file $f:"
+      pep8_result=`pep8 --first $f`
```
apeyser: Why not just
````
if ! pep8_result=`pep8 --first $f`; then
````

File: extras/parse_travis_log.py [hunk 3]
```diff
@@ -177,6 +180,35 @@ def process_static_analysis(f, line):
...
 
+def process_pep8(f, line):
+    """
+    Process PEP8 output for a certain file.
+    """
+    # Check PEP8 on file pynest/nest/lib/hl_api_connections.py:
+    filename = line.split(' ')[-1].strip()[0:-1]
+    d = set()
+    res = {filename: d}
+    while True:
```
apeyser: Options here: 
````
fiter = iter(f)
for line in fiter:
```
and use the same iter at 206
or ... comment what's going on here with walking through the file.

build.sh
tammoippen: Hi @apeyser,

Without it TravisCI complains, and my local machine not: `extras/parse_travis_log.py:285:45: E126 continuation line over-indented for hanging indent`
The line looks much nicer with this over-indention. On my Mac when I execute `pep8 --help` I get:
```
pep8 --help
Usage: pep8 [options] input ...

Options:
...
  --ignore=errors      skip errors and warnings (e.g. E4,W) (default:
                       E121,E123,E126,E226,E24,E704)
  --show-source        show source code for each error
```
i.e. `E126` is ignored here by default. Maybe we should only ignore `E126`? Also see [here](https://github.com/PyCQA/pycodestyle/issues/167) for discussion.

apeyser: My python3?-pep8 doesn't give me the defaults. Maybe I should pull up the script and see what the defaults actually are.

extras/check_code_style.sh
```diff
@@ -216,8 +228,20 @@ for f in $file_names; do
       $CPPCHECK --enable=all --inconclusive --std=c++03 $f
 
       ;;
+    *.py )
+      set +e # pep8 returns 1 if the file does not comply with PEP8
+      echo "Check PEP8 on file $f:"
+      pep8_result=`pep8 --first $f`
+
+      if [ "$?" -gt "0" ]; then
```
apeyser: I thought I commented on this, but can't find it, so...
Why not:
````
if ! pep8_result=`pep8 --first $f`; then
```

extras/check_code_style.sh
```diff
@@ -216,8 +229,20 @@ for f in $file_names; do
       $CPPCHECK --enable=all --inconclusive --std=c++03 $f
 
       ;;
+    *.py )
+      set +e # pep8 returns 1 if the file does not comply with PEP8
+      echo "Check PEP8 on file $f:"
+      pep8_result=`pep8 --first --ignore=$PEP8_IGNORES $f`
+
+      if [ "$?" -gt "0" ]; then
```
apeyser: Why not: if ! pep8_result=`pep8 ...` ?

extras/parse_travis_log.py
tammoippen: I go through the file only once and `f` stores the "pointer" into where we are at reading the file. When i use `iter(f)`, each function has its own state there, correct? Then I have to rewrite the script, which might be not a bad idea, as it will make the individual parsing functions simpler. But I think this would be a topic for another PR.

tammoippen: I go through the file only once and `f` stores the "pointer" into where we are at reading the file. When i use `iter(f)`, each function has its own state there, correct? Then I have to rewrite the script, which might be not a bad idea, as it will make the individual parsing functions simpler. But I think this would be a topic for another PR.

apeyser: In fact, no -- the underlying file handles moves forward, so:
````
f = open(file, 'r')
f1 = iter(f)
f2 = iter(f)
for l in f1: print l
````

will move f2 forward as well as f1. But in general for iterators, you would be right. And you're probably right that everything should just move the iterator instead of the file handle around... but it is too wide for this pr.

I'd just do:
````
for line in f: ....
````

since that's anyway the same thing.

Comment (tammoippen) [inline extras/parse_travis_log.py]: There is another nested loop continuing the iteration throught the file:

```Python
if line.startswith('+echo ' + filename):
    while True:
        line = f.readline()
```

Does it work there, too?

apeyser: Should, since the iterator is simply an adaptor on the common underlying buffer and file descriptor at the very bottom.

extras/check_code_style.sh
```diff
@@ -229,7 +251,10 @@ if [ "$format_error_files" != "" ]; then
     echo "  $f"
   done
   echo
-  echo "Perform '$CLANG_FORMAT -i <filename>' on them, otherwise TravisCI will report failure!"
+  echo "TravisCI will report failure!"
+  echo "- On C/CPP files, perform '$CLANG_FORMAT -i <filename>' on them."
+  echo "- On Python files, perform 'pep8ify -w <filename>' on them and"
```
jougs: Please remove the suffix "on them". See [here](http://uncyclopedia.wikia.com/wiki/Redundancy).

extras/parse_travis_log.py
```diff
@@ -63,18 +64,20 @@ def process_installcheck(f):
 
 def process_changed_files(f):
     """
-    Run through the file f until a line starts with '+file_names='. After the
-    equal sign is a space separated list of changed files in the PR. Returns that
-    list.
+    Run through the file f until a line starts with '+file_names='.
+    After the equal sign is a space separated list of changed files in the PR.
```
jougs: The equal sign is followed by a space separated list of changed files in the PR.

extras/parse_travis_log.py
tammoippen: @apeyser I structured the script with the `readline()` method. When I replace it with the iterator, I get errors like:

    ValueError: Mixing iteration and read methods would lose data

As this is not directly related to PEP8, I would like to put this into another issue, and here just document, what the lines are intended to do.

apeyser: @tammoippen 

Sounds reasonable. I see in the documentation that the read-ahead buffer is broken, that it's only used in next calls but not readline calls.  It's even built on top of a FILE*, so that you have buffers of buffers on top of the POSIX buffers on top of the hardware buffers...
(This is also the answer to why I'm suspicious of all directives coming out of that design camp)

apeyser:
:+1:

PR #301, opened 2016-04-08
Add "b" classifier for string literals to avoid unicode coercion errors by Cython.

jougs:
```
pynestkernel.pyx:43:31: Unicode literals do not support coercion to C types other than Py_UNICODE/Py_UCS4 (for characters) or Py_UNICODE* (for strings).
```

See also #229 for an occurence of this problem.


jougs:
I suggest  @apeyser and @tammoippen as reviewers.

apeyser:
Looks good to me :+1:

heplesser:
:+1:

PR #305, opened 2016-04-11
Fix trivial doc error

sanjayankur31:
A one word fix.


heplesser:
@sanjayankur31 Thank's for spotting and fixing this!

PR #306, opened 2016-04-12
Update include check

sanjayankur31:
Leftover from the connection_builder_manager -> connection_manager
rename I think.

@tammoippen - could you please check if I have this right?


tammoippen:
@sanjayankur31 That is correct. Good catch! 👍

heplesser:
:+1: and merging.

Issue #307, opened 2016-04-14
cmake -Dwith-libneurosim=ON failed


steffengraber:
`cmake -DCMAKE_INSTALL_PREFIX=/home/steffen/work/opt/nest -Dwith-optimize=ON -Dwith-warning=ON -Dwith-mpi=OFF -Dwith-python=ON -Dwith-gsl=OFF -Dwith-libneurosim=ON

make
`
Failed with:

`/home/steffen/Downloads/nest-simulator-master/conngen/conngendatum.h:27:43: fatal error: 

neurosim/connection_generator.h: Datei oder Verzeichnis nicht gefunden

compilation terminated.
nest/CMakeFiles/nest.dir/build.make:77: recipe for target 'nest/CMakeFiles/nest.dir/neststartup.cpp.o' failed

make[2]: **\* [nest/CMakeFiles/nest.dir/neststartup.cpp.o] Error 1

CMakeFiles/Makefile2:595: recipe for target 'nest/CMakeFiles/nest.dir/all' failed

make[1]: **\* [nest/CMakeFiles/nest.dir/all] Error 2

Makefile:146: recipe for target 'all' failed

make: **\* [all] Error 2
`

Where can I find the libneurosim?

tammoippen:
Hi @steffengraber ,

You can find the sources [here](http://software.incf.org/software/libneurosim/code_repository). After installing them, you can give the install path to `with-libneurosim`, if the install dir is not found automatically:

```
... -Dwith-libneurosim=/path/to/installdir/libneurosim ...
```

But you should not get the error... When `cmake` did not find the libneurosim, it should build without it. I will investigate.

steffengraber:
Hi @tammoippen 

works.

Thank you.

steffengraber:
Have you tried `-Dwith-libneurosim=/home/ulianych/PycharmProjects/other/libneurosim/build/`?

dizcza:
> Have you tried `-Dwith-libneurosim=/home/ulianych/PycharmProjects/other/libneurosim/build/`?

Yes, I tried this. No luck.

steffengraber:
That's supposed to work. It does for me.
Can you tell me how you build `libneurosim` and `NEST`? Especially the `configure` line (`libneurosim`) and the `cmake` line (`NEST`).

dizcza:
libneurosim (github master branch):
```
$ ./configure --prefix=/home/ulianych/PycharmProjects/other/libneurosim/build
[...]

PR #308, opened 2016-04-14
Libneurosim build error.

tammoippen:
Only add `conngen`, when libneurosim is actually found. See #307.


heplesser:
Looks sensible to me :+1: .

PR #313, opened 2016-04-19
Bugfix in FindCython.cmake

weidel-p:
My cython version is `0.20.1post0` and `FindCython.cmake` was not able to detect it. 

This pull request is fixing this issue by changing the regex in `FindCython.cmake`.


tammoippen:
:+1:

PR #319, opened 2016-04-25
Improve include checker.

tammoippen:
The script `extras/include_checker.py` suggests an include order according to the [coding guidelines](https://nest.github.io/nest-simulator/coding_guidelines_c++#order-of-includes). This PR gives a warning for excluded files and recursively runs through directories.

### DIFF
--- a/extras/include_checker.py
+++ b/extras/include_checker.py
@@ -145,6 +145,7 @@ def print_includes(includes):
 
 def process_source(path, f, all_header, print_suggestion):
     if f in excludes_files:
+        print("Please do not change the includes in '" + f + "'.")
         return 0
     includes = get_includes_from(path + "/" + f, all_header)
     order_ok = is_include_order_ok(includes)
@@ -164,13 +165,15 @@ def process_source(path, f, all_header, print_suggestion):
 def process_all_sources(path, print_suggestion):
     all_header = all_includes(path)
     count = 0
-    dirs = [d for d in next(os.walk(path))[1] if d[0] != '.' and not d in excludes ]
-    for d in dirs:
-        for root, dirs, files in os.walk(path + "/" + d):
-            for f in files:
-                if re.search("\.h$|\.hpp$|\.c$|\.cc|\.cpp$",f):
-                    # valid source file
-                    count += process_source(root, f, all_header, print_suggestion)
+    for root, dirs, files in os.walk(path):
+        for f in files:
+            if re.search("\.h$|\.hpp$|\.c$|\.cc|\.cpp$",f):
+                # valid source file
+                count += process_source(root, f, all_header, print_suggestion)
+        for d in dirs:
+            if not d in excludes:
+                # valid directory
+                count += process_all_sources(root + "/" + d, print_suggestion)
     return count
 
 
--- a/extras/include_checker.py
+++ b/extras/include_checker.py
@@ -178,7 +178,7 @@ if __name__ == '__main__':
     print_suggestion = True
     if len(sys.argv) != 3:
         print("Use like:")
-        print("  " + sys.argv[0] + " (-f <filename> | -d <directory>)")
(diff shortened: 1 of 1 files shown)


extras/include_checker.py
```diff
@@ -145,6 +145,7 @@ def print_includes(includes):
 
 def process_source(path, f, all_header, print_suggestion):
     if f in excludes_files:
+        print("Please do not change the includes in '" + f + "'.")
```
jougs: Please re-phrase: 
```python
print("Not checking file " + f +" as it is in the exclude list. Please do not change the order of includes.")
```

Also, please add a comment to line 27 explaining why files are in the exclude list. For a possible sentence on this, see https://github.com/nest/nest-simulator/pull/280#discussion_r60908734

jougs:
See my comment above. Otherwise this is very nice. So :+1: once my comment is addressed. I think we can waive a second review as this is only a helper script.

extras/include_checker.py
```diff
@@ -145,6 +147,7 @@ def print_includes(includes):
 
 def process_source(path, f, all_header, print_suggestion):
```
heplesser: Should be `all_headers` (plural) throughout.

extras/include_checker.py
```diff
@@ -84,7 +86,7 @@ def to_string(self):
 
 def all_includes(path):
     result = {}
-    dirs = [d for d in next(os.walk(path))[1] if d[0] != '.' and not d in excludes ]
+    dirs = [d for d in next(os.walk(path))[1] if d[0] != '.' ]
     for d in dirs:
         for root, dirs, files in os.walk(path + "/" + d):
```
heplesser: Use `os.path.join()` to compose paths.

extras/include_checker.py
```diff
@@ -145,6 +147,7 @@ def print_includes(includes):
 
 def process_source(path, f, all_header, print_suggestion):
     if f in excludes_files:
+        print("Not checking file " + f + " as it is in the exclude list. Please do not change the order of includes.")
         return 0
     includes = get_includes_from(path + "/" + f, all_header)
```
heplesser: Use `os.path.join()` to compose paths.

extras/include_checker.py
```diff
@@ -164,21 +167,21 @@ def process_source(path, f, all_header, print_suggestion):
...
-    dirs = [d for d in next(os.walk(path))[1] if d[0] != '.' and not d in excludes ]
-    for d in dirs:
-        for root, dirs, files in os.walk(path + "/" + d):
-            for f in files:
-                if re.search("\.h$|\.hpp$|\.c$|\.cc|\.cpp$",f):
-                    # valid source file
-                    count += process_source(root, f, all_header, print_suggestion)
+    for root, dirs, files in os.walk(path):
+        for f in files:
+            if re.search("\.h$|\.hpp$|\.c$|\.cc|\.cpp$",f):
```
heplesser: Space missing before `f`.

extras/include_checker.py
```diff
@@ -164,21 +167,21 @@ def process_source(path, f, all_header, print_suggestion):
...
+                # valid source file
```
heplesser: Indentation of this comment?

extras/include_checker.py
```diff
@@ -164,21 +167,21 @@ def process_source(path, f, all_header, print_suggestion):
...
+                count += process_source(root, f, all_header, print_suggestion)
+        for d in dirs:
+            count += process_all_sources(root + "/" + d, print_suggestion)
```
heplesser: Use `os.path.join()` to compose paths.

heplesser:
I think it would be very helpful to have a docstring explaining what this file is actually doing. I also think some lines might be a little long---have you run a pep8 check?

heplesser:
Looks good now, merging.

PR #321, opened 2016-04-26
Fix issues around nosetest checking

heplesser:
This PR fixes problems with nosetest detection, see also #318. 

- Checking for nosetest xunit support now uses python-free directory to avoid confusion.
- "Phase 6" heading now in one place, some info given about which test scheme is used.
- Check for nosetest commented.

@tammoippen @jougs Could you take a look?


tammoippen:
👍

jougs:
Nice work. :+1: and merging.

PR #326, opened 2016-04-29
Have target/host triples "arch-vendor-os" as in autotools

tammoippen:
Many internal functions, e.g. `memory_thisjob`, require the target triples "arch-vendor-os" as they were supplied with autotools. The file `GetTriple.cmake` adds functions for retrieving these. The rest of this PR sets the appropriate defines in `config.h.in`.

I propose @jakobj and @jougs as reviewers.


cmake/GetTriple.cmake
```diff
@@ -0,0 +1,89 @@
...
+  # Get the vendor.
+  if ( "${CMAKE_HOST_SYSTEM_NAME}" MATCHES "^Darwin.*" )
+    set( vendor "apple" )
+  else ()
+    set( vendor "pc" )
+  endif ()
+
+  # Get os.
+  if ( "${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows" )
+    set( os "win32" )
```
jougs: Wouldn't just "win" be more correct on current versions of that OS?

cmake/GetTriple.cmake
```diff
@@ -0,0 +1,89 @@
...
+  # Get the vendor.
+  if ( "${CMAKE_SYSTEM_NAME}" MATCHES "^Darwin.*")
+    set( vendor "apple" )
+  else ()
+    set( vendor "pc" )
+  endif ()
+
+  # Get os.
+  if ( "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows" )
+    set( os "win32" )
```
jougs: See above

CMakeLists.txt
```diff
@@ -115,6 +115,10 @@ find_program( SED NAMES sed gsed )
 ##################   CPack, checks, ... in external modules   ##################
 ################################################################################
 
+include( GetTriple )
```
jougs: I don't like the name `GetTriple` for this file. Can you make it more specific, please?`Maybe `DetermineArchitecture` or something like this.

jougs:
This seems like a useful addition to me. See my comments above. :+1: once they are addressed.

jakobj:
I can confirm that this works nicely with my machine (Debian 8), and our cluster (CentOS 7.1). I can't really comment on the code, so :+1: from me. :)

cmake/GetTriple.cmake
```diff
@@ -0,0 +1,89 @@
...
+  # i686 is an enhanced version of x86 
+  if ( arch STREQUAL "x86" )
+    set( arch "i686" )
+  endif ()
+
+  # Get the vendor.
+  if ( "${CMAKE_SYSTEM_NAME}" MATCHES "^Darwin.*")
+    set( vendor "apple" )
+  else ()
+    set( vendor "pc" )
```
heplesser: Would this set vendor to pc also for BlueGene and K?

jakobj:
@heplesser on juqueen this creates the following triples:

> Host triple: ppc64-pc-linux
> Target triple: -pc-bluegeneq_base

target triple looks a bit broken to me?!

jakobj:
and on K:
> Host triple: x86_64-pc-linux
> Target triple: s64fx-pc-linux

this looks fine to me.

heplesser:
@jakobj That target triple for BG/Q looks a bit strange since the first component is empty and the last part quite long.

But all these triples label BG/Q and K as `pc`, which doesn't cause any harm, I believe, but is a bit bit misleading ...

cmake/GetTriple.cmake
tammoippen: I am not so sure how relevant this actually is, given that NEST is not officially supported on Windows ("we do not tests this regularly [building on Windows] and do not support NEST under Cygwin at present"), but the lines are only relevant, if the host system name is `Windows`, then this will default to `win32`. In other cases, the `CMAKE_HOST_SYSTEM_NAME` will just be cast to lower case. Maybe on a 64bit Windows this will default to `win64` - I have not tested this - or on MinGW this will default to `mingw32` (as the [autoconf counterpart](https://sourceforge.net/p/mingw-w64/wiki2/TypeTriplets/)).

CMakeLists.txt
tammoippen: In autotools they are called [`TypeTriplets`](https://sourceforge.net/p/mingw-w64/wiki2/TypeTriplets/). Other cmake implementations call this [`GetTriplet.cmake`](https://opensource.apple.com/source/clang/clang-421.0.57/src/projects/libcxx/cmake/Modules/GetTriple.cmake), or `Triplet.cmake`[[1]](https://github.com/roxlu/SDP/blob/master/build/Triplet.cmake) [[2]](https://github.com/roxlu/video_streamer/blob/master/tools/cmake/Triplet.cmake). As this is specifically for the autotools triplets, I think the name triplet should be contained.

tammoippen:
For NEST the relevant triple is the target triple, which is only different from the host triplet in the case of cross compiling. The target triplet for BG/Q is now:
> -- Target triple: ppc64-ibm-linux

For the K computer it is:
> -- Target triple: s64fx-fujitsu-linux

PR #337, opened 2016-05-11
Pep8ify all Python code in NEST

heplesser:
This PR makes all Python code in NEST PEP8 compliant. For code in some directories (all `examples` dirs and the scripts for the topology user manual), E402 (import not at beginning of file) is suppressed, because it is didactically useful to have imports in several places throughout the script.

@jougs @tammoippen @apeyser Could you review?


extras/check_code_style.sh
```diff
@@ -232,7 +239,13 @@ for f in $file_names; do
     *.py )
       echo "Check PEP8 on file $f:"
 
-      if ! pep8_result=`pep8 --first --ignore=$PEP8_IGNORES $f` ; then
+      if [[ $f =~ $EXAMPLE_DIRS ]]; then
```
apeyser: Isn't there a comment flag within the file to mark sections that should be ignored by pep8? I suggest it only to avoid a proliferation of cases here where ignores should be changed.

heplesser: @apeyser As far as I have been able to find out, the only possibility is to append `# noqa` to pertaining lines. I have used that very sparingly. I have not been able to find a way to disable specific PEP8 errors for individual files or sections of files by comments in the files. In example code that will go onto our website or scripts that are included in the Topology manual, the `# noqa` comment seems undesirable because it would distract readers.

apeyser:
:+1:

extras/check_code_style.sh
apeyser: @heplesser  I just checked it -- https://github.com/PyCQA/pycodestyle/pull/27#issuecomment-218397241 ; Not possible, and apparently will never be possible -- it's a "philosophical" problem until someone forks pep8. The kind of philosophical problems that tend to crop up particularly often in the python community.

jakobj:
running pep8 from the command line for me still complains about three things:

> ./extras/parse_travis_log.py:327:45: E126 continuation line over-indented for hanging indent
./testsuite/manualtests/test_facetshw_stdp.py:137:25: E126 continuation line over-indented for hanging indent
./pynest/nest/tests/compatibility.py:37:30: E701 multiple statements on one line (colon)

heplesser:
@jakobj Could you recheck the issues pep8 flagged on your machine?

jakobj:
@heplesser with the newest changes, pep8 does not complain about files anymore :+1:

extras/check_code_style.sh
apeyser: @heplesser I just found flake8 which is a wrapper around pep8 to make pep8 less religious, including turning off warnings on a file by file basis. Maybe worthwhile using?

heplesser: @apeyser Yes, that sounds interesting. Would you open a new issue for that?

PR #340, opened 2016-05-11
Fix copyright header test on backup files

### LINKED ISSUE #339: Copyright header test fails with backup files in source directory
Some editors (e.g., emacs) create automatic backup files, which are usually placed in the same directory as the actual file. Our copyright header test fails on some of these backup files:

```bash
  File "testsuite/regressiontests/ticket-659-copyright.py", line 104, in <module>
    with open(tested_file) as source_file:
IOError: [Errno 2] No such file or directory: '/home/jordan/opt/nest-simulator/nestkernel/.#common_properties_hom_w.h'
```

I'll create a PR fixing this issue soon.

jakobj:
As mentioned in #339 the test currently fails when backup files are present. This PR introduces a regular expression match on the filename to exclude such files from the check. Fixes #339.


testsuite/regressiontests/ticket-659-copyright.py
```diff
@@ -38,6 +39,15 @@
...
 ]
 
+# match all file names against these regular expressions. if a match
+# is found the file is excluded from the check
+exclude_file_patterns = [
+    '\.#.*',
+    '#.*',
+    '.*~',
+]
+exclude_file_regex = [re.compile(pattern) for pattern in exclude_file_patterns]
```
heplesser: I think it would be more elegant to have a single regular expression using the or operator, i.e., something like `\.#.*|#.*|.*~`. Maybe also add `.bak`.

jakobj: Sure, we could do it in a single one. My thinking was that this way it might be easier to add new patterns later? Like the one you suggested.

jougs: I agree with @jakobj that hsaving the patterns individually helps. But as they are short, maybe they can all be on a single line?

heplesser: To quite a degree, I think, it is a matter of taste: do you want to put the iteration on the Python side or the Regex side? Since most NEST developers are much more versed in Python than in regular expressions, doing it in Python makes sense. On the other hand, I would expect that placing the iteration in a compiled regular expression object will be more efficient. But run-time efficiency is no issue here, so we should choose a coding style most developers will be familiar with.

heplesser:
@jakobj Would you pull recent changes from master into this PR, so that the PEP8 check passes? Once Travis is happy, you have my :+1:.

jakobj:
I now added `.bak` to the excluded patterns, but left them separately. I hope that is fine. @jougs if you're also happy, please consider the thumbs up. :)

jougs:
The new code looks very nice to me. :+1: from me. Thanks for taking care!

PR #323, opened 2016-04-26
Remove GNUReadline link dependency from PyNEST 

tammoippen:
Only the executables `sli` and `nest` use the GNUReadline interaction. All other libraries and modules, e.g. PyNEST, do not need to be linked against the GNUReadline library. In this PR I create a separate target `sli_readline` that contains all the code necessary for the GNUReadline interaction and link it only to sli and nest executable. This way PyNEST is not linked against GNUReadline (and Python can use its readline library) and problems discussed in #210 and #209 will not occur anymore.

PR #341, opened 2016-05-18
Fix building nest without python

tammoippen:
The PR #323 removed GNUReadline from PyNEST, but when building nest without python the library `libnest.dylib` now expects GNUReadline, as no _IS_PYNEST is given, which does not compile (at least not on OS X). This PR allows to differentiate building the NEST CLI, PyNEST and a standalone
library.


tammoippen:
Closing due to wrong workflow.

PR #342, opened 2016-05-18
Fix building nest without python

tammoippen:
The PR #323 removed GNUReadline from PyNEST, but when building nest without python the library `libnest.dylib` now expects GNUReadline, as no _IS_PYNEST is given, which does not compile (at least not on OS X). This PR allows to differentiate building the NEST CLI, PyNEST and a standalone library.

### DIFF
--- a/nest/CMakeLists.txt
+++ b/nest/CMakeLists.txt
@@ -46,6 +46,7 @@ target_include_directories( nest PRIVATE
     ${PROJECT_SOURCE_DIR}/nestkernel
     ${SLI_MODULE_INCLUDE_DIRS}
     )
+
 target_compile_definitions( nest PRIVATE
     -D_BUILD_NEST_CLI
 )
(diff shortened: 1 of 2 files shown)


nest/CMakeLists.txt
```diff
@@ -46,6 +46,9 @@ target_include_directories( nest PRIVATE
     ${PROJECT_SOURCE_DIR}/nestkernel
     ${SLI_MODULE_INCLUDE_DIRS}
     )
+target_compile_definitions( nest PRIVATE
```
heplesser: There should be an empty line before this section.

heplesser:
@tammoippen Could one now remove the [`-D_IS_PYNEST`](https://github.com/nest/nest-simulator/pull/342/files#diff-b369dd2c1ddbc2b51e81a86e9c6d651bR64) macro? It should not play any role in this directory, should it?

tammoippen:
@heplesser No, we still need it, as it adds an argument to [`neststartup`](https://github.com/tammoippen/nest-simulator/blob/953420f93279402b69006970438383add872813c/nest/neststartup.cpp#L82) and adds the [`pynest-init.sli`](https://github.com/tammoippen/nest-simulator/blob/953420f93279402b69006970438383add872813c/nest/neststartup.cpp#L177) file.

heplesser:
@tammoippen Sorry I overlooked that! Then, :+1: from me.

jougs:
:+1: and merging.

PR #350, opened 2016-05-19
Refactor cmake includes into functions

tammoippen:
The includes of cmake modules from NEST (within `<src>/cmake/`) result in executing the contained code at the include position. Due to a not optimal distribution of the code into modules in commit 7488005 at PR #213, this resulted in, e.g., incorrect building of the `conngen` module, if libneurosim is given.

```bash
Built-in modules    : models;precise;topology           # << conngen is missing
• • •
Use libneurosim     : Yes (LibNeurosim 1.0.0)           # although libneurosim is given
    Includes        : /path/to/libneurosim/install/include
    Libraries       : /path/to/libneurosim/install/lib/libneurosim.dylib;/path/to/libneurosim/install/lib/libpyneurosim.dylib
```

In this PR the cmake functionality is put into fine grain functions, e.g. each command line parameter like `-Dwith-gsl=...` is put into a function `nest_process_with_gsl()` that performs all necessary checks and linking. These functions can then be executed in the correct order.

Most of the changes in this PR is putting code into [cmake functions](https://cmake.org/cmake/help/v3.0/command/function.html) and increasing the indent level of the function body. Thus the diff might be easier to view using Githubs [whitespace ignore feature](https://github.com/blog/967-github-secrets) - put `?w=1` behind the diff URL like [this](https://github.com/nest/nest-simulator/pull/350/files?w=1).
Further, `PARENT_SCOPE` is the 'return' feature from cmake functions, but it sets the variable *only* in the parent scope and not in the function scope, this results in some interesting constructs like `cmake/CheckExtraCompilerFeatures.cmake:98`, where a local variable is set again in parent scope. Also see [this](https://stackoverflow.com/questions/29892929/variables-set-with-parent-scope-are-empty-in-the-corresponding-child-scope-why) discussion on `PARENT_SCOPE`.

I propose @jougs and @apeyser as reviewers.


apeyser:
Has this been tested on all architectures? It looks like a simple refactor, but unexpected interactions can be painful.

tammoippen:
Have bin tested on OS X, Lubuntu 14.04 and BG/Q.

tammoippen:
@apeyser I also tested the recipes from #385 .

jougs:
I've just tested this on top of my conngen-refactoring branch. It now detects libneurosim flawlessly, which used to not work before. Nice work and :+1:

PR #366, opened 2016-05-24
Correct versions in FindXXX.cmake

tammoippen:
I missed in these two FindXXX.cmake files, that versions can have more than one digit and with
MUSIC already at version 1.1.15 (see [here](https://github.com/INCF/MUSIC/blob/master/configure.ac#L2)), this might become relevant.

### DIFF
--- a/cmake/FindLibNeurosim.cmake
+++ b/cmake/FindLibNeurosim.cmake
@@ -42,9 +42,9 @@ find_library( PYNEUROSIM_LIBRARY
     )
 
 if ( EXISTS "${LIBNEUROSIM_INCLUDE_DIRS}/neurosim/version.h" )
-  file( STRINGS "${LIBNEUROSIM_INCLUDE_DIRS}/neurosim/version.h" 
+  file( STRINGS "${LIBNEUROSIM_INCLUDE_DIRS}/neurosim/version.h"
                 version_h_contents REGEX "define LIBNEUROSIM_VERSION" )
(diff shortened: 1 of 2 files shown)


cmake/FindLibNeurosim.cmake
```diff
@@ -42,9 +42,9 @@ find_library( PYNEUROSIM_LIBRARY
     )
 
 if ( EXISTS "${LIBNEUROSIM_INCLUDE_DIRS}/neurosim/version.h" )
-  file( STRINGS "${LIBNEUROSIM_INCLUDE_DIRS}/neurosim/version.h" 
+  file( STRINGS "${LIBNEUROSIM_INCLUDE_DIRS}/neurosim/version.h"
                 version_h_contents REGEX "define LIBNEUROSIM_VERSION" )
-  string( REGEX REPLACE ".*([0-9].[0-9].[0-9]).*" "\\1" 
+  string( REGEX REPLACE ".*([0-9]+.[0-9]+.[0-9]+).*" "\\1"
```
heplesser: @tammoippen 
- Why is the dot between the version digits not escaped here (requiring the dot character), while it is escaped in `FindMusic.cmake`?
- Wouldn't it make sense to define all regular expressions as raw strings, so one avoids the double backslashes?

tammoippen: You are of course right. Escaped in e329a63c .

heplesser:
:+1: from me.

jougs:
:+1: and merging. Thanks for spotting and fixing.

PR #378, opened 2016-05-31
Add cmake option for colored gcc >= 4.9 output

jakobj:
This PR adds and option for cmake that enables colored output of errors and warnings. works fine with gcc >= 4.9. Is is maybe necessary to prevent this from being set when using an incompatible compiler? If so, whats the best way to do this? I propose @tammoippen and @apeyser as reviewers.


cmake/ProcessOptions.cmake
```diff
@@ -89,6 +89,13 @@ if ( with-defines )
   endforeach ()
 endif ()
 
+# add colored output from gcc
+if ( with-color )
+  if ( with-color STREQUAL "ON" )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fdiagnostics-color=auto" )
```
tammoippen: This is a GCC specific option, that is only fully supported for GCC >= 4.9. Older versions might choke:

```
g++-4.8 -fdiagnostics-color=auto -Wall -pedantic -c ...
g++-4.8: error: unrecognized command line option '-fdiagnostics-color=auto'
```

I would recommend to not have this as a user option, but always enable it, if possible (colors are cool and useful during coding), e.g. like this:

```cmake
if ( CMAKE_CXX_COMPILER_ID STREQUAL "GNU" )
  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fdiagnostics-color=auto" )
  endif ()
endif ()
```

And also do not forget to add the option for the C compiler (the checks are similar):

```cmake
set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fdiagnostics-color=auto" )
```

jakobj:
included @tammoippen's suggestions. thanks!

tammoippen:
👍  from my side.

heplesser:
@jakobj Would you merge the latest changes from master, so that Travis can test properly, now that #391 is merged?

heplesser:
:+1: and merging!

Issue #376, opened 2016-05-31
Provide verbose error information when MPI_Init_thread fails


sanjayankur31:
I've been having some weird issues with our cluster - my simulation just errors out with these:

```
[cli_71]: aborting job:
Fatal error in PMPI_Init_thread:
Other MPI error
```

The mpi_manager doesn't quite provide any more details and this makes the error quite hard to debug. Would it be possible to use the `MPI_Error_string` method in mpi_manager.cpp to provide more information to the user when the `MPI_Init_thread` method fails?

[Example error file from MPI simulation](https://github.com/nest/nest-simulator/files/290900/nest-v-s.e1491178.txt)

sanjayankur31:
Using `MV2_DEBUG_SHOW_BACKTRACE=1 /usr/local/bin/mpiexec -n $NUM_NODES ... ` gives some more info. Now I get:

```
[node039:mpi_rank_44][print_backtrace]   1: /usr/mpi/gcc/mvapich2-1.7/lib/libmpich.so.3(MPIDI_CH3_Abort+0x5f) [0x7f67310ec2df]
[node039:mpi_rank_44][print_backtrace]   2: /usr/mpi/gcc/mvapich2-1.7/lib/libmpich.so.3(MPID_Abort+0x3e) [0x7f673114afae]
[node039:mpi_rank_44][print_backtrace]   3: /usr/mpi/gcc/mvapich2-1.7/lib/libmpich.so.3(+0xd4a6d) [0x7f6731128a6d]
[node039:mpi_rank_44][print_backtrace]   4: /usr/mpi/gcc/mvapich2-1.7/lib/libmpich.so.3(MPIR_Err_return_comm+0xf0) [0x7f6731128b70]
[node039:mpi_rank_44][print_backtrace]   5: /usr/mpi/gcc/mvapich2-1.7/lib/libmpich.so.3(MPI_Init_thread+0xc0) [0x7f6731146840]
[node039:mpi_rank_44][print_backtrace]   6: /home/asinha/installed-software/nest/lib64/python2.7/site-packages/nest/../../../libnestkernel.so(_ZN4nest10MPIManager8init_mpiEPiPPPc+0x58) [0x7f6731944df8]
[node039:mpi_rank_44][print_backtrace]   7: /home/asinha/installed-software/nest/lib64/python2.7/site-packages/nest/../../../libnestkernel.so(_ZN4nest9init_nestEPiPPPc+0x2d) [0x7f67319343cd]
[...]

PR #379, opened 2016-05-31
Mpi error verbose

sanjayankur31:
First attempt at fixing #376. 

I'm not too well versed with MPI programming - some improvements will probably be needed.

### DIFF
--- a/nestkernel/mpi_manager.cpp
+++ b/nestkernel/mpi_manager.cpp
@@ -90,10 +90,6 @@ nest::MPIManager::init_mpi( int* argc, char** argv[] )
     int mpi_retcode, mpi_error_len;
     char* mpi_error_string;
 
-    // Set MPI error handlers
-    MPI_File_set_errhandler( MPI_ERRORS_RETURN );
-    MPI_Comm_set_errhandler( MPI_ERRORS_RETURN );
-
     mpi_retcode = MPI_Init_thread(
       argc, argv, MPI_THREAD_FUNNELED, &provided_thread_level );
     if ( mpi_retcode != 0 )
--- a/nestkernel/mpi_manager.cpp
+++ b/nestkernel/mpi_manager.cpp
@@ -87,19 +87,7 @@ nest::MPIManager::init_mpi( int* argc, char** argv[] )
     comm = kernel().music_manager.communicator();
 #else  /* #ifdef HAVE_MUSIC */
     int provided_thread_level;
-    int mpi_retcode, mpi_error_len;
-    char* mpi_error_string;
-
-    mpi_retcode = MPI_Init_thread(
-      argc, argv, MPI_THREAD_FUNNELED, &provided_thread_level );
-    if ( mpi_retcode != 0 )
-    {
-      MPI_Error_string( mpi_retcode, mpi_error_string, &mpi_error_len );
-      std::string msg = String.compose(
-        "Error initializing MPI. Error encountered: %1", mpi_error_string );
-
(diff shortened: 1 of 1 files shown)


sanjayankur31:
@heplesser, @jougs, @tammoippen - could you please take a look at this when time permits? Thanks.

nestkernel/mpi_manager.cpp
```diff
@@ -87,7 +87,19 @@ nest::MPIManager::init_mpi( int* argc, char** argv[] )
...
+
+    mpi_retcode = MPI_Init_thread(
+      argc, argv, MPI_THREAD_FUNNELED, &provided_thread_level );
+    if ( mpi_retcode != 0 )
+    {
+      MPI_Error_string( mpi_retcode, mpi_error_string, &mpi_error_len );
+      std::string msg = String::compose(
+        "Error initializing MPI. Error encountered: %1", mpi_error_string );
+
+      LOG( M_ERROR, "MPIManager::init_mpi", msg );
```
heplesser: I think this should be followed by throwing an exception. @tammoippen What is the relation now between LOGging messages and raising exceptions? At some point, we tried to remove error logging as much as possible and leave that to the exception handler.

tammoippen: `LOG( M_ERROR, ...` does not raise an exception. I agree that this should result in an exception like:

```C++
throw KernelException( msg );
```

My other question: does your code actually execute lines 95ff? In #376 it seams as if the code fails within the `MPI_Init_thread` function.

sanjayankur31: I'm not entirely sure. This seems to be the only way of getting information from a failed MPI_Init_thread call - or the only way I've found in the documentation. I'm running a simulation now that will hopefully fail and confirm if this works or not. Unfortunately, it's still stuck in the cluster queue.

The page [here](http://www.mpich.org/static/docs/v3.1/www3/MPI_Init_thread.html) says:

```
Errors

All MPI routines (except MPI_Wtime and MPI_Wtick) return an error value; C routines as the 
value of the function and Fortran routines in the last argument. Before the value is returned, 
the current MPI error handler is called. By default, this error handler aborts the MPI job. The 
error handler may be changed with MPI_Comm_set_errhandler (for communicators), 
MPI_File_set_errhandler (for files), and MPI_Win_set_errhandler (for RMA windows). The 
MPI-1 routine MPI_Errhandler_set may be used but its use is deprecated. The predefined 
error handler MPI_ERRORS_RETURN may be used to cause error values to be returned. 
Note that MPI does not guarentee that an MPI program can continue past an error; however, 
MPI implementations will attempt to continue whenever possible.

MPI_SUCCESS
    No error; MPI routine completed successfully. 

MPI_ERR_OTHER
    Other error; use MPI_Error_string to get more information about this error code. 
```

Note that it does mention setting an error handler - maybe that would be a better way of going about it - setting default error handlers right at the beginning? 

I also just noticed that `MPI_ERRORS_RETURN` needs to be set to get the method to return an error value. I'll look at how this is to be done and update the code.

sanjayankur31:
The last commit is wrong. Fixing.

sanjayankur31:
The error handlers only come in after MPI has been initialised, so these methods must return error codes as the docs say. Testing on a simulation now.

sanjayankur31:
This didn't seem to work. I still got no error information. :/

sanjayankur31:
I tried a try and catch block too but it didn't seem to help much. I'm at a loss now. Closing for the time being :/

PR #380, opened 2016-06-01
Fix TravisCI

weidel-p:
The [llvm.org/apt](http://llvm.org/apt) seams to be down due to 'excess load':

> APT mirror was temporary switched off due to excess load. We are working on bringing it back. Stay tuned!

Since we get `clang-format` from them, all current build jobs will fail. In this PR we get the pre-built binaries for `clang-format` and install them manually (similar to `cppcheck`). See 2daa77f for this.

We are not sure, if we are responsible for the 'excess load', hence we want to use the TravisCI caching feature (d3e2fa2), which does not work at the moment, as we are not using the 'container-based' infrastructure ([here](https://docs.travis-ci.com/user/migrating-from-legacy/)) and/or a payed plan, i.e. private repository([here](https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies) at the end).

We propose @heplesser and @tammoippen as reviewers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,21 +7,20 @@ dist: trusty
 
 env:
   matrix:
-    - xGSL=0 xMPI=0 xPYTHON=0
-    - xGSL=0 xMPI=0 xPYTHON=1
-    - xGSL=0 xMPI=1 xPYTHON=0
-    - xGSL=0 xMPI=1 xPYTHON=1
-    - xGSL=1 xMPI=0 xPYTHON=0
-    - xGSL=1 xMPI=0 xPYTHON=1
-    - xGSL=1 xMPI=1 xPYTHON=0
-    - xGSL=1 xMPI=1 xPYTHON=1
+    - xGSL=0 xMPI=0 xPYTHON=1 CACHE_NAME=JOB
+    - xGSL=0 xMPI=0 xPYTHON=0 CACHE_NAME=JOB
+    - xGSL=0 xMPI=1 xPYTHON=0 CACHE_NAME=JOB
+    - xGSL=0 xMPI=1 xPYTHON=1 CACHE_NAME=JOB
+    - xGSL=1 xMPI=0 xPYTHON=0 CACHE_NAME=JOB
+    - xGSL=1 xMPI=0 xPYTHON=1 CACHE_NAME=JOB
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,30 +7,24 @@ dist: trusty
...
   matrix:
-    - xGSL=0 xMPI=0 xPYTHON=0
-    - xGSL=0 xMPI=0 xPYTHON=1
-    - xGSL=0 xMPI=1 xPYTHON=0
-    - xGSL=0 xMPI=1 xPYTHON=1
-    - xGSL=1 xMPI=0 xPYTHON=0
-    - xGSL=1 xMPI=0 xPYTHON=1
-    - xGSL=1 xMPI=1 xPYTHON=0
-    - xGSL=1 xMPI=1 xPYTHON=1
+    - xGSL=0 xMPI=0 xPYTHON=1 CACHE_NAME=JOB
```
heplesser: What does the CACHE_NAME mean here? I cannot see it used anywhere in the script. And could there be a problem with using the same cache name on multiple processes running simultaneously?

heplesser:
I would assume that it takes bigger guns than ours to bring down the llvm mirror, but this would make us more robust. Is there any risk that concurrent jobs can lead to race conditions or similar parallel problems?

.travis.yml
tammoippen: You can have a different cache for each job. Giving a `CACHE_NAME` ensures, that we only have one cache. See this documentation about [TravisCI caching](https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices) (point 4. would make different caches in our case).

tammoippen:
@heplesser Each job is executed in a separate virtual machine, i.e. concurrent jobs do not have access to the installed packages of another job. Each job has to download the cache from S3 separately, see [how cache works in TravisCI](https://docs.travis-ci.com/user/caching/#How-does-the-caching-work%3F).

tammoippen:
👍  from my side.

lekshmideepu:
More details about the excess load problem are discussed in the following thread:
[[llvm-dev] IMPORTANT: APT repo temporary switched off](http://lists.llvm.org/pipermail/llvm-dev/2016-May/100303.html)

heplesser:
:+1: and merging.

PR #384, opened 2016-06-02
Make NEST compile statically on BlueGene again

tammoippen:
@jakobj Encountered that compiling NEST statically on JUQUEEN generated some `undefined symbols`.

CMake replaces library paths of libraries in implicit link directories by their `-l<name>` version. When building statically, it also adds linker flags, like `-Wl,-Bstatic -lfoo -Wl,-Bdynamic`. See [CMP0060](https://cmake.org/cmake/help/v3.3/policy/CMP0060.html) for more information.

This commit sets `LINK_SEARCH_END_STATIC` and `LINK_SEARCH_START_STATIC` to `TRUE` in the case of BlueGene and static linking. This should allow NEST to compile again. (I tried both static and dynamic construction of NEST.)

I suggest @jakobj, @jougs and/or @heplesser as reviewers.

### DIFF
--- a/cmake/Platform/BlueGeneQ_XLC.cmake
+++ b/cmake/Platform/BlueGeneQ_XLC.cmake
@@ -35,6 +35,8 @@ set( CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g -qarch=qp -qtune=qp -DNDEBUG" CACHE 
 set( OpenMP_C_FLAGS "-qsmp=omp" CACHE STRING "Compiler flag for OpenMP parallelization" FORCE )
 set( OpenMP_CXX_FLAGS "-qsmp=omp" CACHE STRING "Compiler flag for OpenMP parallelization" FORCE )
 
+set( with-warning "-qinfo=all" CACHE STRING "Enable user defined warnings. [default ON, when ON, defaults to '-Wall']" FORCE )
+
 if ( static-libraries )
   __bluegeneq_setup_static( XL CXX )
   __bluegeneq_setup_static( XL C )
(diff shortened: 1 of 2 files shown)


heplesser:
Seems sensible to me and will affect only BlueGene. Could this be relevant for K, too? :+1:

jakobj:
looks good. currently trying it out on JQ. in the meantime: is it possible to disable `-Wall` by default when compiling on JQ? all this output is quite annoying if you forget to disable warning manually.

jakobj:
works nicely. :+1:

tammoippen:
@jakobj Great idea. Is included in ed496ca.

tammoippen:
> Could this be relevant for K, too?

@heplesser I have not tried to build statically on K yet and I cannot access K anymore. @jakobj  maybe you can test this?

apeyser:
Someone should look at the latest cmake sources for the BlueGeneQ macros. We need it to compile dynamically as well, and I doubt that we are unique in the flags we use.

tammoippen:
@apeyser I can confirm that dynamic compiling works on BlueGeneQ. The last changes to the [BlueGeneQ files](https://github.com/Kitware/CMake/tree/master/Modules/Platform) were about 11 months ago, so the platform files in NEST are based on the most recent versions.

apeyser:
@tammoippen @jakobj I see the policy report now. So many CMake libraries will be broken through a long convoluted policy regarding library link lines that should have a fix, but no one will ever figure out where it's coming from or why, other than to simply kill the whole functionality aggressively.

Okay be me then. I'm sure this will break at sometime in the future, given the convoluted policy tree that is in CMake, but I don't see any other option.

heplesser:
Looks like everyone agrees that this is a working, if not great, solution, so merging.

PR #385, opened 2016-06-03
Add documentation for building NEST with GCC on BG/Q [skip ci]

tammoippen:
This PR adds documentation on how to build NEST with GCC on BG/Q.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -166,6 +166,11 @@ environment variables are set properly:
     # makes HOME and PYTHONPATH available for python
     runjob --exp-env HOME --exp-env PYTHONPATH ... : python script.py
 
+Compiling NEST with the GCC (`-DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_GCC`)
+might require you to use a GSL library compiled by the GCC, otherwise undefined
+symbols break your build. Changing the compiler should be done in the file
+`<src>/cmake/Platform/BlueGeneQ_GCC.cmake`.
+
 Compiling for Fujitsu Sparc64
 =============================
 
--- a/INSTALL
+++ b/INSTALL
@@ -133,16 +133,24 @@ Compiling for BlueGene/Q
 NEST provides a cmake tool-chain file for cross compilation for BlueGene/Q. When
 configuring NEST use the following `cmake` line:
 
-    cmake -DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_XLC 
-          -DCMAKE_INSTALL_PREFIX:PATH=</install/path>
-          -Dwith-python=OFF
-          -Dstatic-libraries=ON
+    cmake -DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_XLC \
+          -DCMAKE_INSTALL_PREFIX:PATH=</install/path> \
+          -Dwith-python=OFF \
+          -Dstatic-libraries=ON \
           </path/to/NEST/src>
 
-It is recommended to build statically on larger BlueGene/Q systems. If you need
-PyNEST on BlueGene/Q, you have to compile dynamically, i.e. `-Dstatic-libraries=OFF`
-and you have to cythonize the `pynest/pynestkernel.pyx/.pxd`
-on a machine with Cython installed:
+It is recommended to build statically on larger BlueGene/Q systems. If you compile
+dynamically, be aware that the BlueGene/Q system might not provide a `ltdl`
+library. If you want to dynamically load an external user module, you have to
+compile and install a `ltdl` yourself and add `-Dwith-ltdl=<ltdl-install-dir>`
+to the `cmake` line. Otherwise add `-Dwith-ltdl=OFF`.
+
+BlueGene/Q and PyNEST
+---------------------
+
+Building PyNEST on BlueGene/Q requires you to compile dynamically, i.e. 
+`-Dstatic-libraries=OFF`. Further, you have to cythonize the
+`pynest/pynestkernel.pyx/.pxd` on a machine with Cython installed:
 
     cythonize pynestkernel.pyx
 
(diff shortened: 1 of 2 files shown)


INSTALL
```diff
@@ -166,6 +166,11 @@ environment variables are set properly:
     # makes HOME and PYTHONPATH available for python
     runjob --exp-env HOME --exp-env PYTHONPATH ... : python script.py
 
+Compiling NEST with the GCC (`-DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_GCC`)
```
heplesser: "with GCC"

INSTALL
```diff
@@ -166,6 +166,11 @@ environment variables are set properly:
...
+might require you to use a GSL library compiled by the GCC, otherwise undefined
```
heplesser: "compiled using GCC"

INSTALL
```diff
@@ -166,6 +166,11 @@ environment variables are set properly:
...
+symbols break your build. Changing the compiler should be done in the file
+`<src>/cmake/Platform/BlueGeneQ_GCC.cmake`.
```
heplesser: I am a bit confused by the last sentence: Doesn't the `..._GCC.cmake` file define gcc as the compiler to use? Do you maybe mean that one needs to define the GSL version to use in that file?

tammoippen: The [`..._GCC.cmake`](https://github.com/nest/nest-simulator/blob/master/cmake/Platform/BlueGeneQ_GCC.cmake#L25) file indeed defines the `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER`, otherwise cmake will find the gcc for the front-end and one always has to state the compiler flags in the configure line. BTW: the same goes for the [`..._XLC.cmake`](https://github.com/nest/nest-simulator/blob/master/cmake/Platform/BlueGeneQ_XLC.cmake#L22) file.

heplesser: Ok. But what do you mean then by "Changing the compiler should be done ..."? Why would I want to change the compiler? After all, I specify the GCC Toolchain file to choose GCC?

tammoippen: You might be interested in compiling with GCC toolchain but use a different gcc version, e.g. see [here](http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JUQUEEN/UserInfo/CompilingTuning.html#CLANG)

heplesser: I think I just got totally confused because I thought that the second sentence explained how to do what the first sentence suggests. But first and second sentence have nothing to do with each other! Could you add after the first sentence another one  such as "Use `-D<whatever is needed to select the correct GSL>` to link NEST against the correct GSL version. Then start a new paragraph "If you want to use a different version of GCC, please edit ...". I think that will make things more transparent.

apeyser:
@heplesser @tammoippen  Have you tried this recipe from scratch? We have INSTALL recipes that don't work without "something extra".

tammoippen:
@heplesser @apeyser 
I adjusted the `INSTALL` according to the comments and the suggestions in #390. I tried all recipes (`XLC` and `GCC`) with and without Python.

In 0898c81 I removed the `FORCE` argument from the `XLC` `with-warning` option, so that one can silence the warnings.
In eae837b I added `\` for easier copy&paste of multiline commands.

apeyser:
@tammoippen @heplesser 
We still don't have a proper build, see my comment on #390, which may be a cmake problem, an installation problem, or a change on libraries on JQ. 

So, we're going to keep having problems with this pull being complete.

heplesser:
@tammoippen I tried to follow the instructions for building on Juqueen including PyNEST, using the following command line:

```
jinb3304@juqueen2:~/NEST/bld_master $ cmake -DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_XLC \
> -DCMAKE_INSTALL_PREFIX=`pwd`/install \
>           -Dstatic-libraries=OFF \
>           -Dcythonize-pynest=OFF \
>           -Dwith-python=/bgsys/tools/Python-2.7/bin/hostpython \
>           -DPYTHON_LIBRARY=/bgsys/tools/Python-2.7/lib64/libpython2.7.so.1.0 \
>           -DPYTHON_INCLUDE_DIR=/bgsys/tools/Python-2.7/include/python2.7 \
>           -Dwith-ltdl=OFF \
> ../src
```

CMake ran, but issued an error message in between:

```
-- Target triple: ppc64-ibm-linux
CMake Error at /usr/share/cmake/Modules/CPack.cmake:395 (message):
  CPack license resource file: "/homea/jinb33/jinb3304/NEST/src/LICENSE"
  could not be found.
  /usr/share/cmake/Modules/CPack.cmake:400 (cpack_check_file_exists)
  cmake/CPackVariables.cmake:55 (include)
  CMakeLists.txt:116 (include)

-- Looking for include file inttypes.h
```

It then produced then normal summary at the end, not mentioning any error. But it did **not** create a `Makefile`.

tammoippen:
@heplesser I cannot reproduce the reported issue. I tried with the default `cmake 2.8.12` and with a custom install of `cmake 3.5.0` and both succeeded with your commands copy-pasted. The error you report is a missing `LICENSE` file. Have you deleted it? Is your nest sources actually in `../src`?

heplesser:
@tammoippen My fault. I was a bit overeager when removing autotools-related files and got totally confused by the error message. After a clean clone NEST built and installed. I have not tried running.

apeyser:
@heplesser 
With #428 I've played a bit with the broken BGQ platform file. I'll check this as well.

apeyser:
Also, I seem to finally have figured out how to get Nest with Python to build on BGQ. There's a not-so-well documented --with-ltdl flag in the cmake configuration which allows a dynamic build without using the dynamic link libraries. Dynamic linking is needed if anything links dynamically (such as a python module) without failing with undefined symbol errors (pthread seems to be notorious for this). 

More for the documentation here.

apeyser:
So XLC & BGQ:
````
module load gsl
cmake .. -Dwith-mpi=ON -DCMAKE_INSTALL_PREFIX=<INSTALL> -DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_XLC -DCMAKE_C_COMPILER=/bgsys/drivers/ppcfloor/comm/xl/bin/mpixlc_r -DCMAKE_CXX_COMPILER=/bgsys/drivers/ppcfloor/comm/xl/bin/mpixlcxx_r -Dwith-python=/bgsys/local/python3/3.4.2/bin/python3 -DPYTHON_LIBRARY=/bgsys/local/python3/3.4.2/lib/libpython3.4m.a -DPYTHON_INCLUDE_DIR=/bgsys/local/python3/3.4.2/include/python3.4m -Dwith-ltdl=OFF -Dstatic-libraries=OFF
````

GCC & BGQ -GSL:
````
cmake .. -Dwith-mpi=ON -DCMAKE_INSTALL_PREFIX=<INSTALL> -DCMAKE_TOOLCHAIN_FILE=Platform/BlueGeneQ_GCC -DCMAKE_C_COMPILER=/bgsys/local/comm/mpigcc -DCMAKE_CXX_COMPILER=/bgsys/local/comm/mpig++ -Dwith-python=/bgsys/local/python3/3.4.2/bin/python3 -DPYTHON_LIBRARY=/bgsys/local/python3/3.4.2/lib/libpython3.4m.a -DPYTHON_INCLUDE_DIR=/bgsys/local/python3/3.4.2/include/python3.4m -Dwith-gsl=OFF -Dwith-ltdl=OFF
````

Does anyone happen to know the flags for BGQ, GSL & cmake?

apeyser:
Replaced with pull #447

PR #387, opened 2016-06-07
Documentation improvement to Hill-Tononi neuron and minor code prettification

heplesser:
- Improved documentation of mathematical details in HT neuron.
- Reorderd some statements, removed unnecessary namespace qualifiers, added const where possible.
- Renamed some variables.
- Switched orded of operands in prefactor and peak_value so all expressions are positive.


heplesser:
@ingablundell Could you review this one?

models/ht_neuron.cpp
```diff
@@ -618,21 +661,27 @@ ht_neuron::update( Time const& origin, const long_t from, const long_t to )
...
+    {
       S_.g_spike_ = false; // Deactivate potassium current.
+    }
 
     // Add new spikes to node state array
     for ( size_t i = 0; i < B_.spike_inputs_.size(); ++i )
+    {
       S_.y_[ 2 + 2 * i ] +=
         V_.cond_steps_[ i ] * B_.spike_inputs_[ i ].get_value( lag );
+    }
```
DimitriPlotnikov: It is really hard to understand the intention of this loop (an update of  DG_AMPA,     DG_NMDA,  DG_GABA_A,  DG_GABA_B state variables). Did you think about unrolling of the loop and the usage of explicit names of state variables would improve comprehensibility of the code?

DimitriPlotnikov: Another possible improvement: in the calibrate function there is a TODO:
In the original model there is a TODO (line 577, ht_neuron.cpp):    ```
// NOTE: code below initializes conductance step size for incoming pulses.
 // Variable and function names need to be changed!
```
Since you are already working on the ht_model, could your improve this?

ingablundell:
@heplesser The new names of the variabless in ht_neuron.cpp look very sensible to me! Also the ordering of operands seems to be correct. The documentation for the calculation of g_peak/t_peak looks nice. Why does $a$ have to be larger than $b$?

Can you also propose a reviewer for the continuous integration stuff who can look at the build.sh and the parse_travis_log.py files?

models/ht_neuron.cpp
heplesser: @DimitriPlotnikov I see your point, but I like the loop because it is more flexible in case more synapse models would be added. I have added a more detailed comment to clarify things. I also removed the second line of the comment on line 577, since, after looking again at that code, I didn't really think it would make much sense to change function/variable names there.

heplesser:
@ingablundell Thanks! Strictly speaking, one only needs a != b, and I have changed the comment accordingly. In biological reality, the rise time constant always seems to be shorter than the decay time constant.

I merged all recent changes from master now, so the changes in build.sh and parse_travis_log.py are no longer part of this PR.

Issue #377, opened 2016-05-31
Importing NEST into Python modifies sys.argv 


heplesser:
Importing NEST into Python modifies the content of `sys.argv`. It does not matter whether NEST is imported before or after `sys`.

To reproduce, save the following code into a file:
```
import sys
print "Before:", sys.argv
import nest
print "After :", sys.argv
```
Then run with varying number of command line arguments, e.g.:
```
python st.py A
Before: ['st.py', 'A']
After : ['pynest', 'A']   # (almost) OK

python st.py A B
Before: ['st.py', 'A', 'B']
After : ['pynest', 'A', 'B']  # (almost) OK

python st.py A B C
Before: ['st.py', 'A', 'B', 'C']
After : ['pynest', 'C', 'C', 'C']  # BAD

python st.py A B C D
Before: ['st.py', 'A', 'B', 'C', 'D']
After : ['pynest', 'C', 'D', 'C', 'D']  # BAD
```

Note that the first entry is changed (`pynest` instead of `st.py`) and, for more than two command line arguments, also subsequent arguments.

This bug is **not** present in 2.8.0.

heplesser:
This was probably introduced by db73559, see also issue #163 and PR #164.

@apeyser and @mdjurfeldt : You wrote or reviewed the PR, could you take a look? I am wondering how to write a good, automatic test for this.

apeyser:
@heplesser 
Does this happen with and without MUSIC, only with MUSIC or only without MUSIC? And how was this found --- does someone have a particular problem arising from this?

heplesser:
@apeyser I have only tested this without MUSIC. I noticed the problem when running a Python script that receives simulation parameters on the command line, see the toy example below. The problem is that this script will see the modified command line arguments unless I move `import nest` after the point where I read out the command line arguments. Scripts should not depend on such ordering

 ```
import nest
import sys

rate = float(sys.argv[1])
weight = float(sys.argv[2])
seed = int(sys.argv[3])

print rate, weight, seed

nest.ResetKernel()
nest.SetKernelStatus({'grng_seed': seed, 'rng_seeds': [seed+1]})
pg = nest.Create('poisson_generator', params={'rate': rate})
nrn = nest.Create('iaf_psc_alpha')
sd = nest.Create('spike_detector')
nest.Connect(pg, nrn, syn_spec={'weight': weight})
nest.Connect(nrn, sd)

nest.Simulate(1000)

[...]

PR #393, opened 2016-06-10
Issue #377: Python NEST import changes argv

apeyser:
In pynestkernel.pyx init, the encoded arguments were not being stored
as python objects, leading to premature deallocation of the c
strings. The symptom was a corrupted sys.argv.

### DIFF
--- a/pynest/pynestkernel.pyx
+++ b/pynest/pynestkernel.pyx
@@ -181,17 +181,18 @@ cdef class NESTEngine(object):
 
         cdef char* arg0 = "pynest\0"
         cdef char** argv_bytes = <char**> malloc((argc+1) * sizeof(char*))
-
         if argv_bytes is NULL:
             raise NESTError("couldn't allocate argv_bytes")
-
-        argv_bytes[0] = arg0        
-        argv_bytes[argc] = NULL
-
         try:
-            for i in range(1, argc):
-                arg_byte = argv[i].encode()
-                argv_bytes[i] = arg_byte
+            argv_bytes[0] = arg0
+            argv_bytes[argc] = NULL
+
+            # Need to keep a reference to encoded bytes #377
+            arg_bytes = []
+            argv_bytes_args = argv_bytes+1
+            for i, argvi in enumerate(argv[1:]):
+                arg_bytes.append(argvi.encode())
+                argv_bytes_args[i] = arg_bytes[-1]
 
             self.pEngine = new SLIInterpreter()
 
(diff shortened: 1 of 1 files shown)


apeyser:
@heplesser @mdjurfeldt 
Have a look and review

heplesser:
@apeyser Thank you for the fix! It works for my test case. But it took me a long time to understand the code, with the mix of C-style arrays and pointers with Python at the base of my confusion. Let me see if I got it half-way correct. The main problem is that we need to copy arguments from Python `argv` to a C-style array-of-char arrays `argv_bytes` to pass to `neststartup`, and that we in this process must convert any unicode to plain byte strings. Furthermore, `neststartup` may remove MPI-related arguments from `argv_bytes`. Therefore, we must afterwards construct a new Python `argv` that contains only those arguments that are still left. In this process, we must convert plain strings back to unicode.

The key problem in the old code was that in
```
for i in range(1, argc):
    arg_byte = argv[i].encode()
    argv_bytes[i] = arg_byte
```
`arg_byte` was a pythonic temporary variable and the we stored pointers to that.

In the new code
```
arg_bytes = []
argv_bytes_args = argv_bytes+1
for i, argvi in enumerate(argv[1:]):
    arg_bytes.append(argvi.encode())
    argv_bytes_args[i] = arg_bytes[-1]
```
the individual arguments are stored in a pythonic array `arg_bytes` and are thus "kept alive", so we can use pointers to them. `argv_bytes_args` is here a pointer into the `argv_bytes` array, isn't it?

If I got this right, then :+1: from my side, provided someone else (@mdjurfeldt ?) tests that removal of MPI/MUSIC-related options is handled properly and that you document in the code what is going on.

mdjurfeldt:
I can check the MUSIC side today.

Sorry for not thinking about storage lifetime when submitting the original
code!
Den 10 jun 2016 15:04 skrev "Hans Ekkehard Plesser" <
notifications@github.com>:

apeyser:
@heplesser 
Yup -- that's it. arg_byte is a temporary `byte` object, which contains a `char*` array that is extracted implicitly by assigning it to the `char* argv_bytes[i]`. When the `byte` object is cleaned up, the array is `free`d. 

The temporary in the original code was there because the Cython compiler flags this very case when a unnamed temporary is transiently created, but the temporary variable tricks the compiler into thinking that the byte object is stored for the proper duration somewhere.

@mdjurfeldt 
I don't think this was your code, from the subversion history. It seems to go back to they first version in the old repository for CyNest.

apeyser:
Cleaned up by commenting and simplifying operations.

Why does SLI require a non-empty argv[0]??? It seems totally arbitrary --- and conflicts with the additionally totally nonsensical rules that python uses. 

Arrgg... these conventions and definitions go back to the 70s...

apeyser:
@gtrensch 
Could you review so we can close?

mdjurfeldt:
I can confirm that this seems to work with MUSIC. (Sorry for the delay---other things in-between.)

Issue #392, opened 2016-06-08
Improve robustness, markup and parsing of build log on Travis


heplesser:
The work on #380 and #391 revealed that the build log generated on Travis should be improved:
- Instead of relying on text or commands, introduce explicit marker lines with special format for mark the beginning and end of certain sections, e.g. `======= BEGIN VERA =======` and  `======= END VERA =======`. This will allow for much more robust parsing, even as the content of sections evolves.
- If the statistical analysis has found problems, the final report may contain the entire diff result. Thus, the summary becomes very difficult to read. I suggest to keep it shorter, at least not include the full diff if it is longer than a few lines.
- On rare occasions, e.g. [Travis Build 841.4](https://travis-ci.org/nest/nest-simulator/jobs/136142790) the 

```
curl https://api.github.com/repos/nest/nest-simulator/pulls/389/files
```

  in [build.sh](https://github.com/nest/nest-simulator/blob/ef980a92fef66d9d5a43429eb987f851c3a607a6/build.sh#L117) may fail, presumably if Github cannot fullfill the request on time, leading to a 

```
jq: error (at <stdin>:4): Cannot index string with string "filename"
```

In that case, the `file_names` variable in `build.sh` remains empty, no files are tested for formatting errors, and the build may pass even if the PR contained files with formatting errors. This should not happen: if the command filling `file_names` fails, the build should stop.

tammoippen:
Please see #395 if it fixes most of the points.

heplesser:
@gtrensch Given your suggestions for further improvements, I leave this issue open even after #395 is merged and leave it in your care.

gtrensch:
This issue has been resolved by #546 and can be closed.

PR #395, opened 2016-06-13
TravisCI: More robust parsing of TravisCI log

tammoippen:
Solves #392.
This PR makes the parsing of the TravisCI more robust and allows to order the different steps of the `build.sh` in any order (for the parsing). It is also possible to remove different steps completely and the parser will not break. A summary of the changes:

In `build.sh`:
* Every block is surrounded by tags like 

        ======= Extract changed files start =======
        ...
        ======= Extract changed files end =======
* Output is reduced:
  - Comment [`set -x`](http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_02_03.html#sect_02_03_02) out, as we do not need to debug the travis log regularly. One can uncomment it, if needed.
  - Remove verbose flag of extracting `clang-format`.
* Make static analysis optional with env. variable `xSTATIC_ANALYSIS`.
* Fix [regex bug](https://travis-ci.org/nest/nest-simulator/jobs/136773603#L4188) (see line 4188) for `if [[ $f =~ $EXAMPLE_DIRS ]]`. I think, the TravisCI environment does not allow `if [[ ... ]]`.
* Add output that tells if results are not uploaded, if you are working on your fork.

In `extras/parse_travis_log.py`:
  * No state on reading/parsing the log. Every function opens the file itself and completely parses it for its issues. The actual parsing happens only within the starting/ending tags relevant for this function.
  * Separate `vera++`, `cppcheck` and `clang-format` parsing into standalone functions.
  * Add `print_...` functions for nicer output.
  * Cleanup "`main`" function.

I propose @heplesser and @gtrensch as reviewers.


gtrensch:
:+1: Everything works fine!  Just one remark from my side: There is a comment in parse_travis_log.py referring to the dependencies on build.sh. A similar comment in build.sh would be good.

build.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 set -e
-set -x
+#set -x
```
heplesser: Why commented out instead of deleted?

extras/parse_travis_log.py
```diff
@@ -49,81 +53,88 @@ def process_installcheck(f):
...
-                           line.strip().split('=')[1].split(' ')), line)
-
-
-def process_vera(f, filename):
+    in_changed_files_section = False
+    for line in open(f, "r"):
+        if (not in_changed_files_section and
+                line.strip() == "======= Extract changed files start ======="):
+            in_changed_files_section = True
+        if in_changed_files_section and line.startswith('file_names='):
```
heplesser: Shouldn't this be `line.strip().startswith('file_names=')` to be on the safe side? Applies also to all other uses of `line.startswith()`.

build.sh
tammoippen: @heplesser I documented the 'out-commenting'. When debugging the script, the flag is very [informative](http://julio.meroh.net/2010/01/set-e-and-set-x.html).

tammoippen:
@gtrensch I added the comment in 500f192. Thanks.

extras/parse_travis_log.py
tammoippen: @heplesser I added the `strip()`'s in 4033e3e.

tammoippen:
@heplesser in 749d929 I introduced some intentional failures that should trigger 'vera, clang-format, cppcheck, pep8, building, testsuite'. I will remove it afterwards.

tammoippen:
See this [build](https://travis-ci.org/nest/nest-simulator/jobs/138848521) for forced errors.

heplesser:
:+1: and merging.

PR #404, opened 2016-06-22
Removed duplicate include line, probably introduced by a merge.

heplesser:
@jougs Would you check this one? I think a single reviewer is enough here.

### DIFF
--- a/nestkernel/nest_types.h
+++ b/nestkernel/nest_types.h
@@ -60,9 +60,6 @@
  * (C) Copyright 1995-2006 The NEST Initiative.
  */
 
-// Generated includes:
-#include "config.h"
-
 /**
  * Namespace for the NEST simulation kernel.
  */


jougs:
What's this line a duplicate for? I don't see another include of `config.h` in that file. Don't we need it?

Also, do you know why there is no mention of Travis here?

lekshmideepu:
I don't see any changes in this file here:
https://github.com/heplesser/nest-simulator/blob/eafb4ed51277ac235ebf9a862277fdcd4cb572c8/nestkernel/nest_types.h
May be that is the reason Travis didn't trigger a build

heplesser:
@jougs `config.h` is included on [line 33](https://github.com/nest/nest-simulator/pull/404/files#diff-42615d7378f1abcdbe8ff2af37737101R33) and line 64. I removed the latter one. I am not sure why no travis run was triggered.

jougs:
Maybe I got confused in the same way as Travis ;-) Anyway: good catch! Merging.

Issue #300, opened 2016-04-08
Figure 3.6 in the toplogy user manual is empty


sanjayankur31:
I was going through the [topology user manual](https://github.com/nest/nest-simulator/blob/master/topology/doc/Topology_UserManual.pdf) and noticed that the graphs in figure 3.6 (Distance-dependent and randomized weights and delays. See text for details) are empty.

heplesser:
Fixed by #408.

Issue #352, opened 2016-05-20
Implement randomized initialization for neurons


heplesser:
With `NewConnect`,  we support randomised initialisation of synapses. Provide the same type of support for neuron initialisation, so that one could write

```
/iaf_neuron 20 << /V_m << /distribution /uniform /low -75. /high -55. >> >> Create
```

This was previously trac854.

jakobj:
If no one minds, I'd give this a try. @heplesser if that's OK please assign the ticket to me.

heplesser:
@jakobj Great that you want to work on this! All the "mechanics" should be in place from the Connect framework, but need to be applied to neuron initialization. The [`ConnParameter` class](https://github.com/nest/nest-simulator/blob/193594f9ce199a0db1322e7752da00237a671f4b/nestkernel/conn_parameter.h#L55) represents parameters and can be used for neuron as well as for synapse parameters. We should probably rename it; it really represents parameter initializers (fixed scalar values, random values, arrays).

Take a look at the [`ConnBuilder` constructor](https://github.com/nest/nest-simulator/blob/193594f9ce199a0db1322e7752da00237a671f4b/nestkernel/conn_builder.cpp#L50) for how parameters objects are created from the connection specification. It should be possible to do this in a quite similar way for parameters for neurons. One thing that might be slightly more complicated is how to know which parameters cannot be set in neuron models.

It may be a good idea to create a few test cases first, especially for initialization using random numbers or arrays in case of parallel building.

jakobj:
thanks @heplesser, i agree that we can reuse and should rename `ConnParameter`. from what i've gathered so far `SetDefaults` is called whenever `Create` is called. i am not sure yet, where i can see this in the code, it seems to be some sli dictionary magic. ;) the issue then is: if defaults are changed to initialize all nodes during one connect call with the specified parameters, at which point should we place the randomization? i can surely randomize the default value, but then all nodes created in the same call will have the same (random) value. this is not really what we want. one approach would be to check for the existence of a distribution in the dictionary, store it somewhere else, create all neurons, then randomize the corresponding values. i am not sure about the angle of attack here.

heplesser:
@jakobj The default parameter values are set and reset in [`Create_l_i_D` at the SLI level](https://github.com/nest/nest-simulator/blob/9faa71e51c9413657a909ad8ecbb8b5d5923693c/lib/sli/nest-init.sli#L160). 

We might want to move this down to the C++-level and proceed as follows:

[...]

PR #408, opened 2016-06-24
Rename Parameter class in Topology to TopologyParameter

heplesser:
The renaming is motivated by #352 and the need for a `Parameter` class in NEST. As an extra, it fixes a little detail in the scripts used to build the topology user manual. The main point of renaming in topology is that the `Parameter` class to be introduced in nestkernel will be the general parameter class for the future.

@jakobj @jougs Would you take a look---it is really just renaming.


topology/doc/user_manual_scripts/connections.py
```diff
@@ -24,6 +24,7 @@
 import nest
 import nest.topology as tp
 import matplotlib.pyplot as plt
+from mpl_toolkits.mplot3d.axes3d import Axes3D
```
jougs: Is this used somewhere in the file? If it is, why was this `import` not here before?

jougs:
Why did you re-add all the figures? For me they look the same, with only very minor shifts probably due to another version of some font library. Can you please remove them?

The code changes are trivial and look good to me. :+1: once the images are gone and my other comment is addressed.

topology/doc/user_manual_scripts/connections.py
heplesser: This is required to enable [3d projection here](https://github.com/heplesser/nest-simulator/blob/63fc287e0ba00d1c02c13bca4a4c8cbeea1c1352/topology/doc/user_manual_scripts/connections.py#L223). I am not sure why it was not needed before. That may be due to internal changes in matplotlib.

jougs: Thanks for the explanation.

heplesser:
@jougs I committed all the figure files because they were recreated with the modified script and topology code. Incidentally, this fixes #300: Fig 3.6. is now back in the PDF. But I agree that the figures should probably not be committed, just the final manuscript PDF. I will create a new issue for that change. I'd appreciate merging this PR soon, since #352 depends on it.

jougs:
Agreed. As this is just a simple rename otherwise, I'll merge without second review.

PR #409, opened 2016-06-27
Reduce the number of build jobs

jougs:
Many configuration options of NEST are actually independent. This means we don't have to run the full matrix, but just a minimal job, a maximal job, and some jobs for dependent options.

This PR simplifies the matrix from 8 runs to 4 runs while also adding a new option for builds with and without threading (04c668b).

Using this, I already found 3 tests that were failing when building without threading (5387bbc).

I suggest @heplesser and @lekshmideepu as reviewers.

PR #416, opened 2016-07-01
Actually use the configure flags for ltdl and readline

jougs:
This is a high-priority hot-fix for #409!


heplesser:
:+1: merging right away.

PR #418, opened 2016-07-06
Added: load modules from environment at startup

obreitwi:
This PR adds the capability for nest to automatically load modules at startup which are specified via the environment variable `NEST_MODULES`. Furthermore, additional search paths for SLI scripts may be specified via the `SLI_PATH` variable.

We use [gnumodules](http://modules.sourceforge.net/) to manage our software installation and keep separate installations of nest and modules. Therefore we need a way for nest to know which nest modules the user wishes to automatically load.

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -25,6 +25,7 @@ Initializer of PyNEST.
 
 import sys
 import os
+import re
 
 # This is a workaround for readline import errors encountered with Anaconda
 # Python running on Ubuntu, when invoked from the terminal
@@ -182,6 +183,9 @@ def init(argv):
         if not quiet:
             engine.run("pywelcome")
 
+        expand_sli_path()
+        load_modules()
+
         # Dirty hack to get tab-completion for models in IPython.
         try:
             __IPYTHON__
@@ -196,6 +200,40 @@ def init(argv):
 
     else:
         _kernel.NESTError("Initiatization of NEST failed.")
+
+
+def expand_sli_path():
+    """ Add paths defined in SLI_PATH to searchpath. """
+
+    sli_paths = [path for path in
+                 os.environ.get("SLI_PATH", "").split(os.path.pathsep)
+                 if path != ""]
+
+    assert_sli_valid(sli_paths)
+
+    for path in sli_paths:
+        sli_run("({}) addpath".format(path))
+
+
+def assert_sli_valid(names):
+    for n in names:
+        if re.search(r"[\(\)\\]", n):
+            raise _kernel.NESTError("Path or Module names must not contain "
+                "parentheses or backslashes.")
+
+
+
+def load_modules():
+    """ Load a list of modules that are specified via NEST_MODULES. """
+
+    modules = [module for module in
+               os.environ.get("NEST_MODULES", "").split(os.path.pathsep)
+               if module != ""]
+
+    assert_sli_valid(modules)
+
+    for module in modules:
+        sli_run("({}) Install".format(module))
 
 
 def test():


heplesser:
@obreitwi  This seems sensible to me in principle, but ideally, the auto-loading should happen at the SLI, not the Python level.  @jougs, could you take a look?

apeyser:
Do we need to escape filenames to SLI?
Other than that, looks good.

heplesser:
@apeyser What do you mean by "escape filenames to SLI"?

apeyser:
@heplesser: module)x.so

Is that a) something that should be escaped b) defined as an invalid filename that should cause an arbitrary failure within SLI, c) something that should return an error as an illegal filename, d) a problem in SLI that needs to be fixed.

apeyser:
@heplesser : any thoughts on whether we do anything with the escapes?

heplesser:
@apeyser @obreitwi Module names that confuse the SLI parser will confuse the user, especially when they are induced via PyNEST. I therefore suggest that we prohibit the following: `()\`. All else should be parsed properly by SLI and result in a proper error message (file not found) if the name is misspelled.

I suppose something like 
```python
def assert_valid(names):
    for n in names:
        if re.search(r'[\(\)\\]', n):
            raise nest.NESTError('Path or Module names must not contain parentheses or backslashes.')
```
should do. I don't really see a great likelihood for either of those being part of a valid file or path name.

apeyser:
ping @obreitwi

obreitwi:
@apeyser I added the verification step as suggested by @heplesser.

jougs:
Sorry to chip in so late.

I totally agree with @heplesser. This is a sensible addition to NEST, but it should definitely go to the SLI level. You're anyway mostly calling SLI commands from the Python level.

Something like this could go to the end of `lib/sli/sli-init.sli`:
```
(SLI_PATH) getenv { (:) breakup { addpath } forall } if
```

While something like this could go to the end of `lib/sli/nest-init.sli`:
```
(NEST_MODULES) getenv { (:) breakup { Install } forall } if
```

I don't see any reason why the environment variables shouldn't contain parentheses or backslashes in this implementations as there's no parsing going on. Probably colons are a bad idea, though.

If you're already at it, could you please remove the old comment about `SLIHOME` in `lib/sli/sli-init.sli` and add some documentation for these two new paths (and for the existing path `NEST_MODULE_PATH`) to a new file `extras/userdoc/md/documentation/environment-variables.md`?

Many thanks!

pynest/nest/__init__.py
```diff
@@ -198,6 +202,40 @@ def init(argv):
...
+    modules = [module for module in
+               os.environ.get("NEST_MODULES", "").split(os.path.pathsep)
+               if module != ""]
+
+    assert_sli_valid(modules)
+
+    for module in modules:
+        sli_run("({}) Install".format(module))
+
+
```
apeyser: Extra blank here ("pynest/nest/__init__.py:225:1: E303 too many blank lines (3)" -- I think pep8 is pointing at the wrong line

pynest/nest/__init__.py
```diff
@@ -198,6 +202,40 @@ def init(argv):
...
+        sli_run("({}) addpath".format(path))
+
+
+def assert_sli_valid(names):
+    for n in names:
+        if re.search(r"[\(\)\\]", n):
+            raise _kernel.NESTError("Path or Module names must not contain "
+                "parentheses or backslashes.")
+
+
```
apeyser: Extra blank here ("pynest/nest/__init__.py:225:1: E303 too many blank lines (3)"

pynest/nest/__init__.py
```diff
@@ -198,6 +202,40 @@ def init(argv):
...
+    for path in sli_paths:
+        sli_run("({}) addpath".format(path))
+
+
+def assert_sli_valid(names):
+    for n in names:
+        if re.search(r"[\(\)\\]", n):
+            raise _kernel.NESTError("Path or Module names must not contain "
+                "parentheses or backslashes.")
```
apeyser: pynest/nest/__init__.py:221:17: E128 continuation line under-indented for visual indent

pynest/nest/__init__.py
```diff
@@ -198,6 +202,40 @@ def init(argv):
...
+    sli_paths = [path for path in
+                 os.environ.get("SLI_PATH", "").split(os.path.pathsep)
+                 if path != ""]
+
+    assert_sli_valid(sli_paths)
+
+    for path in sli_paths:
+        sli_run("({}) addpath".format(path))
+
+
```
apeyser: This may make pep8 spit up

pynest/nest/__init__.py
```diff
@@ -198,6 +202,40 @@ def init(argv):
         _kernel.NESTError("Initiatization of NEST failed.")
```
apeyser: This may make pep8 spit up

apeyser:
Needs a bit of reformatting for pep8: some may be old extra spaces in the code.

jougs:
@obreitwi: I will add the functionality in SLI, so there's no need for you to fix up things here.

Many thanks for making us aware of the problem and implementing the PyNEST version. I'll add you as a possible reviewer to my PR later today.

jougs:
Please see #479 for a replacement PR that adds handling for the environment variables on the SLI level.

apeyser:
Since #479 should fix this in many fewer lines, let's close this, merge that one, and let @obreitwi get his fix without any more work from his part

PR #422, opened 2016-07-15
Use pandas if available

### LINKED ISSUE #400: Use pandas.read_csv instead of numpy.loadtxt
This explains it: http://akuederle.com/stop-using-numpy-loadtxt. (numpy.loadtxt is extremely slow compared to pandas.read_csv.)

One place where numpy.loadtxt is used is in the [raster_plot.py](https://github.com/nest/nest-simulator/blob/master/pynest/nest/raster_plot.py) file. 

Not the most urgent issue, but something that can be updated when there's time.

sanjayankur31:
It tries to load the pandas module and uses it if it can, otherwise, it
falls back to numpy.

Note that the nest.is_iterable test returns True for a string, since a
string is iterable and the function then tries to iterate over each
character in the string! I've, therefore, changed the test.

Tested this out with the gdf file generated from
`examples/nest/tsodyks_shortterm_bursts.sli` using the
`examples/nest/plot_tsodyks_shortterm_bursts.py` plotter script and it
works both with and without pandas.

Fixes #400 to begin with.

### DIFF
--- a/pynest/nest/raster_plot.py
+++ b/pynest/nest/raster_plot.py
@@ -19,9 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-"""
-Functions for raster plotting.
-"""
+""" Functions for raster plotting."""
 
 import nest
 import numpy
@@ -29,8 +27,7 @@ import pylab
 
 
 def extract_events(data, time=None, sel=None):
-    """Extracts all events within a given time interval or are from a
-    given set of neurons.
+    """Extract all events within a given time interval.
 
     Both time and sel may be used at the same time such that all
     events are extracted for which both conditions are true.
@@ -55,7 +52,6 @@ def extract_events(data, time=None, sel=None):
     numpy.array
         List of events as (gid, t) tuples
     """
-
     val = []
 
     if time:
@@ -92,7 +88,6 @@ def from_data(data, sel=None, **kwargs):
     kwargs:
         Parameters passed to _make_plot
     """
-
     ts = data[:, 1]
     d = extract_events(data, sel=sel)
     ts1 = d[:, 1]
@@ -106,8 +101,8 @@ def from_file(fname, **kwargs):
 
     Parameters
     ----------
-    fname : str
-        Name of file
+    fname : str or tuple(str) or list(str)
+        File name or list of file names
     kwargs:
         Parameters passed to _make_plot
     """
@@ -121,9 +116,21 @@ def from_file(fname, **kwargs):
 
 def from_file_pandas(fname, **kwargs):
     """Use pandas."""
-    if isinstance(fname, (list, tuple)):
+    if isinstance(fname, str):
+        fname = fname
+        dataFrame = pandas.read_csv(
+            fname, sep='\s+', lineterminator='\n',
+            header=None, index_col=None,
+            skipinitialspace=True)
+        data = dataFrame.values
+
+    elif isinstance(fname, (list, tuple)):
         data = None
         for f in fname:
+            if not isinstance(f, str):
+                print >> sys.stderr, 'List entry is not a python2 string'
+                return None
+
             if data is None:
                 dataFrame = pandas.read_csv(
                     f, sep='\s+', lineterminator='\n',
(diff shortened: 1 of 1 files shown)


apeyser:
Maybe the test for ["string"...] should be inverted? We should check for isinstance(str) for the single filename case rather than checking that we received a list or a tuple. 

In fact, the code can be simplified by just making 
````
if isinstance(fname, str):
    fname = (fname,)
````

but maybe there's a python2.7 issue with str vs byte vs unicode types for the test?

**Comment by sanjayankur31** [comment]
Working on this now!

File: pynest/nest/raster_plot.py [hunk 1]
```diff
@@ -121,9 +116,21 @@ def from_file(fname, **kwargs):
 
 def from_file_pandas(fname, **kwargs):
     """Use pandas."""
-    if isinstance(fname, (list, tuple)):
+    if isinstance(fname, str):
```
apeyser: Ahh -- I think my suggestion was unclear. I just meant that if you do
````
if isinstance(fname, str):
    fname = [fname]

if isinstance(fname, (list, tuple)):
    ....
````

then you could eliminate lines 121-125 here, which are a duplication of 134-146

File: pynest/nest/raster_plot.py [hunk 2]
```diff
@@ -121,9 +116,21 @@ def from_file(fname, **kwargs):
...
+
+    elif isinstance(fname, (list, tuple)):
         data = None
         for f in fname:
+            if not isinstance(f, str):
+                print >> sys.stderr, 'List entry is not a python2 string'
+                return None
+
             if data is None:
                 dataFrame = pandas.read_csv(
```
apeyser: And shouldn't this be:
````
for f in fname:
   dataFrame = panda.read_csv(...)
   newdata = dataFrame.values
   if data is None: data = newdata
   else: data = numpy.concatenate(...)
````

Currently it looks to me like this would only read the first fname.

**Comment by apeyser** [comment]
@sanjayankur31 
:+1: Much more concise!

@heplesser Comments?

File: pynest/nest/raster_plot.py [hunk 3]
```diff
@@ -106,21 +101,54 @@ def from_file(fname, **kwargs):
...
-            else:
-                data = numpy.concatenate((data, numpy.loadtxt(f)))
+    if isinstance(fname, str):
+        fname = [fname]
+
+    if isinstance(fname, (list, tuple)):
+        try:
+            global pandas
+            pandas = __import__('pandas')
+            from_file_pandas(fname, **kwargs)
```
heplesser: Don't we need a `return` here?

pynest/nest/raster_plot.py
```diff
@@ -106,21 +101,54 @@ def from_file(fname, **kwargs):
...
+        except ImportError:
+            from_file_numpy(fname, **kwargs)
```
heplesser: And here we need a `return` as well, don't we?

pynest/nest/raster_plot.py
```diff
@@ -106,21 +101,54 @@ def from_file(fname, **kwargs):
...
     else:
-        data = numpy.loadtxt(fname)
+        print >> sys.stderr, 'fname should be of str/list(str)/tuple(str).'
```
heplesser: Didn't now this redirection was possible, in Python. But I am rather sure this line won't work in Python 3, or?

pynest/nest/raster_plot.py
```diff
@@ -106,21 +101,54 @@ def from_file(fname, **kwargs):
...
+        return None
```
heplesser: This return as superfluous, I think.

pynest/nest/raster_plot.py
```diff
@@ -106,21 +101,54 @@ def from_file(fname, **kwargs):
 
     Parameters
     ----------
-    fname : str
-        Name of file
+    fname : str or tuple(str) or list(str)
+        File name or list of file names
```
heplesser: I think it would be good to document how lists of files are handled. If I understand it right, data from files are concatenated and plotted as if they had been stored in a single file?

heplesser:
@sanjayankur31 Looks quite good to me, I just added some small comments (and deleted some again, I had not realised at first that the loading functions call the plot function).

pynest/nest/raster_plot.py
```diff
@@ -117,8 +121,7 @@ def from_file(fname, **kwargs):
         except ImportError:
             from_file_numpy(fname, **kwargs)
     else:
-        print >> sys.stderr, 'fname should be of str/list(str)/tuple(str).'
-        return None
+        print('fname should be of str/list(str)/tuple(str).')
```
heplesser: "fname should be one of ..."

heplesser:
@sanjayankur31 Just one more little text fix (see in-text comment) and I will merge.

PR #428, opened 2016-07-22
Allow overriding path to compiler for BGQ

### LINKED ISSUE #427: CMake broken on BGQ (or CMake considered harmful)
CMakes design is broken for a) mpi and b) BGQ. For MPI, the cmake authors seem to believe that they can in general write a program that can extract all the necessary information from the mpi wrappers for all machines in all cases, and then use that information to directly construct the actions needed to reproduce that behavior. 

This would ultimately imply P==NP :astonished: 

For BGQ with sionlib, the combination of things that cmake can't handle because of its improper concept of abstraction leads to a failure (can't find MPI_send and such errors)

The way around this would be to set CMAKE_{C,CXX}_COMPILER to the proper wrapper, but that's unfortunately not possible, because the cmake community sets those unconditionally in the platform files without marking them as CACHE.

I'll go ahead and fix this, but this will lead in the future to updates of the cmake platform files that revert the changes made here by mistake.

apeyser:
Resolves #427, at least makes the compiler overrideable for mpi usages.

@wschenck @gtrensch ?

### DIFF
--- a/cmake/Platform/BlueGeneQ_GCC.cmake
+++ b/cmake/Platform/BlueGeneQ_GCC.cmake
@@ -22,8 +22,8 @@ include( Platform/BlueGeneQ_Base )
 set( ENV{PATH} "/bgsys/drivers/ppcfloor/gnu-linux/bin:$ENV{PATH}" )
 
 # set the compiler
-set( CMAKE_C_COMPILER powerpc64-bgq-linux-gcc CACHE FILEPATH "cmake is so awesomely misdesigned")
-set( CMAKE_CXX_COMPILER powerpc64-bgq-linux-g++ CACHE FILEPATH "cmake is so awesomely misdesigned")
+set( CMAKE_C_COMPILER powerpc64-bgq-linux-gcc CACHE FILEPATH "override C compiler" )
+set( CMAKE_CXX_COMPILER powerpc64-bgq-linux-g++ CACHE FILEPATH "override C++ compiler" )
 
 if ( static-libraries )
   __bluegeneq_setup_static( GNU CXX )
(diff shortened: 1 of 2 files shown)


apeyser:
@abigailm Any one else? This is really quite trivial.

cmake/Platform/BlueGeneQ_GCC.cmake
```diff
@@ -22,8 +22,8 @@ include( Platform/BlueGeneQ_Base )
 set( ENV{PATH} "/bgsys/drivers/ppcfloor/gnu-linux/bin:$ENV{PATH}" )
 
 # set the compiler
-set( CMAKE_C_COMPILER powerpc64-bgq-linux-gcc )
-set( CMAKE_CXX_COMPILER powerpc64-bgq-linux-g++ )
+set( CMAKE_C_COMPILER powerpc64-bgq-linux-gcc CACHE FILEPATH "cmake is so awesomely misdesigned")
```
jakobj: are these strings `"cmake is so awesomely misdesigned"` necessary to make it work? otherwise I would remove them.

jakobj: ahh, I see. just change it to something else then.

jakobj:
after my comments have been addressed :+1: from me.

wschenck:
I was not asked directly, but nevertheless I would like to approve this fix. :+1:

cmake/Platform/BlueGeneQ_GCC.cmake
apeyser: :disappointed:

PR #431, opened 2016-07-27
Changed default "precision" value of recording_device

SepehrMN:
The default value of "precision" of a recording_device is 3. If precise a model exists, a spike_detector prints a warning and sets the precision value to 15. But if the property "precise_times" is set on a spike_detector, this warning does not get printed and one would expect the value to be set to 15 too, but this would not be the case; it would be 3. This pull requests fixes this discrepancy and makes sure that if the value is not specified, it will always be 15.

### DIFF
--- a/nestkernel/recording_device.cpp
+++ b/nestkernel/recording_device.cpp
@@ -62,7 +62,7 @@ nest::RecordingDevice::Parameters_::Parameters_( const std::string& file_ext,
   , withgid_( withgid )
   , withtime_( withtime )
   , withweight_( withweight )
-  , precision_( 3 )
+  , precision_( 15 )
   , scientific_( false )
   , binary_( false )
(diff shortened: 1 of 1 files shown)


SepehrMN:
I suggest @abigailm as a reviewer.

heplesser:
@SepehrMN I do not think that setting the precision to 15 decimals by default is wise. Most people simulate with fixed time steps around 0.1 ms, so three decimals are more than enough. Using 15 decimals by default will only blow up file size unnecessarily. As long as you do not have precise spiking neurons, more decimals do not make sense, unless you timestep with less than microsecond resolution. That is why setting `precise_times` on the `spike_detector` does not change the number of decimals.

heplesser:
This is a subset of #432. I therefore suggest closing this one.

PR #432, opened 2016-07-27
Spike detector precision implicit setting and variable name

SepehrMN:
Two improvements:
1. "user_set_precise_times" is currently used to store whether the precise_times property been directly set and "precision" of the recording_device the actual value. By changing to "has_user_set_precise_times", one could readily understand the point of this variable without wasting time.
2. By setting "precision" one is implicitly setting the value of the variable above so in the set method it is necessary to set it to true if "precision" is directly set.

### DIFF
--- a/models/spike_detector.cpp
+++ b/models/spike_detector.cpp
@@ -42,21 +42,21 @@
 #include "integerdatum.h"
 
 nest::spike_detector::spike_detector()
-        : Node()
-        // record time and gid
-        , device_( *this, RecordingDevice::SPIKE_DETECTOR, "gdf", true, true )
-        , has_user_set_precise_times_( false )
-        , has_proxies_( false )
-        , local_receiver_( true )
+  : Node()
+  // record time and gid
+  , device_( *this, RecordingDevice::SPIKE_DETECTOR, "gdf", true, true )
+  , has_user_set_precise_times_( false )
+  , has_proxies_( false )
+  , local_receiver_( true )
 {
 }
 
 nest::spike_detector::spike_detector( const spike_detector& n )
-        : Node( n )
-        , device_( *this, n.device_ )
-        , has_user_set_precise_times_( n.has_user_set_precise_times_ )
-        , has_proxies_( false )
-        , local_receiver_( true )
+  : Node( n )
+  , device_( *this, n.device_ )
+  , has_user_set_precise_times_( n.has_user_set_precise_times_ )
+  , has_proxies_( false )
+  , local_receiver_( true )
 {
 }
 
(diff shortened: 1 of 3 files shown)


SepehrMN:
please do not review the second point yet.

SepehrMN:
I suggest @abigailm as a reviewer.

heplesser:
This is an extended version of #431; I suggest to close that one.

models/spike_detector.cpp
```diff
@@ -45,7 +45,7 @@ nest::spike_detector::spike_detector()
   : Node()
   // record time and gid
   , device_( *this, RecordingDevice::SPIKE_DETECTOR, "gdf", true, true )
-  , user_set_precise_times_( false )
+  , has_user_set_precise_times_( false )
```
heplesser: I don't agree with this name change, I think the original name is sufficiently clear (note that "set" is past tense in the original name).

nestkernel/recording_device.cpp
```diff
@@ -62,7 +62,7 @@ nest::RecordingDevice::Parameters_::Parameters_( const std::string& file_ext,
   , withgid_( withgid )
   , withtime_( withtime )
   , withweight_( withweight )
-  , precision_( 3 )
+  , precision_( 15 )
```
heplesser: This unnecessarily blows up file size, see my comment on #431.

models/spike_detector.cpp
```diff
@@ -139,7 +139,27 @@ void
...
+  if ( d->known( names::precision ) )
+  {
+
+    if ( d->known( names::precise_times ) )
+    {
+      if ( getValue< bool >( d, names::precise_times ) )
+      {
+        device_.set_precise( true, getValue< long >( d, names::precision ) );
+      }
+    }
```
heplesser: Using precise times makes only sense if we have precise spiking neurons. Activating it when setting the recording precision does not seem to make sense.

heplesser:
@SepehrMN I don't understand what you want to achieve with your changes. Could you describe the use case you are trying to address?

models/spike_detector.cpp
SepehrMN: Dear Hans-Ekkehard, the first time I saw it, I read "user_set" word as a compound adjective. I will revert this change.

models/spike_detector.cpp
SepehrMN: Yes, this was due to a confusion on my part. There is no reason to add this to the class.

SepehrMN:
There is two issue left, the discrepancy in setting the value of precision and the default value of  "precision". In #431 I offered a solution for the latter, but as @heplesser pointed out, it was not a good one. I try to offer a better one with a new pull request.

If there are precise models, and a user sets "precise_times" to true, everything goes fine. But if there are precise models and it is not set, there can be a few problems:

1. The precision value set by the user is ignored and it is set to 15. NEST does give the warning message below and it prints the value set, but it does not directly point to why the specified "precision" value was ignored and how one can set it to other than 15. I will push a fix for this.

    _Precise neuron models exist: the property precise_times of the 
    spike_detector with gid 2 has been set to true, precision has been set to 
    15._

2.   The default precision is 3 on recording_device, but if precise models exist and "precise times" is not set, spike_detector deviates from this and sets it to 15. A warning is printed, but would it not make sense for everything to be consistent and only 1 default value to exist and changing another parameter ("precise_times") not affect the default value? I suggest changing the 15 to 3 and push a new fix for this.

SepehrMN:
This has morphed into #446 so I am closing it. Thanks @heplesser !

PR #434, opened 2016-07-29
Additional timers and local spike counter for profiling and performance measurements

wschenck:
In this pull request two timers and one counter are added to the nest kernel.

The timers contain (1) the time spent in gather_events for spike communication via MPI and (2) the time spent in gather_events for collocating spike buffers. The counter contains the number of spike events generated on the specific rank (both on- and offgrid). All these measurements are for the last call to simulate. They are reset to zero at the beginning of a new call to simulate within prepare_simulation.

These measurements are useful for light-weight profiling and collecting of performance data.

### DIFF
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -51,6 +51,8 @@ EventDeliveryManager::EventDeliveryManager()
   , global_offgrid_spikes_()
   , displacements_()
   , comm_marker_( 0 )
+  , time_collocate_( 0.0 )
+  , time_communicate_( 0.0 )
 {
 }
 
@@ -67,6 +69,7 @@ void
 EventDeliveryManager::initialize()
 {
   init_moduli();
+  reset_timers();
 }
 
 void
@@ -89,6 +92,8 @@ void
 EventDeliveryManager::get_status( DictionaryDatum& dict )
 {
   def< bool >( dict, "off_grid_spiking", off_grid_spiking_ );
+  def< double >( dict, "time_collocate", time_collocate_ );
+  def< double >( dict, "time_communicate", time_communicate_ );
 }
 
 void
@@ -244,6 +249,13 @@ EventDeliveryManager::update_moduli()
     slice_moduli_[ d ] = ( ( kernel().simulation_manager.get_clock().get_steps()
                              + d ) / min_delay ) % nbuff;
   }
+}
+
+void
+EventDeliveryManager::reset_timers()
+{
+  time_collocate_ = 0.0;
+  time_communicate_ = 0.0;
 }
 
 void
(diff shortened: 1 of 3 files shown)


heplesser:
@wschenck Looks good to me. What is the reason for having the stopwatches as member variables instead of local variable where they are needed? And do we really need two of them?

wschenck:
@heplesser You are right, there is no real reason for the stopwatches being member variables. And one could live with only one of them. I will update the pull request in the beginning of next week. I tend towards defining the single stopwatch as a static local variable then.

nestkernel/event_delivery_manager.cpp
```diff
@@ -577,12 +596,26 @@ EventDeliveryManager::deliver_events( thread t )
 void
 EventDeliveryManager::gather_events( bool done )
 {
+  // IMPORTANT: Ensure that gather_events(..) is called from a single thread and
+  //            NOT from a parallel OpenMP region!!!
```
heplesser: Is there any way to enforce this by an assertion?

heplesser:
@wschenck Good. Since the stopwatch is only to be used in serial context, a static local should not cause any problems.

nestkernel/event_delivery_manager.cpp
apeyser: ! omp_in_parallel() might do the trick

apeyser:
:+1:

nestkernel/event_delivery_manager.cpp
wschenck: I looked into this: Unfortunately, gather_events(..) is called from within a parallel region, although only by one thread, which is enforced by "omp single" and "omp master". However, omp_in_parallel() and also omp_get_num_threads() refer to the superordinate parallel region and don't perform as expected for this reason.
Anyhow, the problem that gather_events(..) should be called only by a single thread exists regardless of the StopWatch introduced here. In this way, the problem is independant from this specific pull request. But of course, it would be good to find a better solution than to just indicate this requirement not at all or just by a comment.

wschenck:
My travis tests failed only for the configurations with MPI. The corresponding error message is shown below. It occurs in "make installcheck" and looks really weird. I doubt that this is the fault of my code changes. See in the following:

Phase 1: Testing if SLI can execute scripts and report errors
-------------------------------------------------------------
Running test 'selftests/test_pass.sli'... *** Error in `mpirun': munmap_chunk(): invalid pointer: 0x00002ad9a33e111d ***
Failed: unexpected exit code 134
Running  mpirun -np 1 /home/travis/build/wschenck/nest-simulator/result/bin/nest /home/travis/build/wschenck/nest-simulator/result/share/doc/nest/selftests/test_pass.sli
Error: Fail to load fglrx kernel module!
Error! Fail to load fglrx kernel module! Maybe you can switch to root user to load kernel module directly

jougs:
The failing is not due to your code, but is a known Travis issue at the moment. I've reported it to them, see here for details: https://github.com/travis-ci/packer-templates/issues/221

heplesser:
:+1: from me and merging.

PR #436, opened 2016-08-01
Lowpassfilter_spike_detector 

SepehrMN:
Optimized device for efficient generation of a continuous trace of spiking activity. Special thanks to Hans-Ekkehard Plesser for providing insight on how to create it.

Although this is the implementation for a specific computation, by editing the methods "calculate_decay_" and "add_impulse_" one can create a different one.


SepehrMN:
This pull request will be ready for review on Wednesday/Thursday. There will be minor code changes, compatibility with refactored kernel, improved documentation, tests, formatting with clang-format and rule checking with vera++.

SepehrMN:
This commit can be reviewed from tomorrow. The latest commit works. Will format and push my tests.

SepehrMN:
It is now safe to have a look at the code for the device. I will do the formatting, fix my pull request's commits and and upload tests later.

SepehrMN:
The last commit fixes all issues. Travis is failing because of https://github.com/travis-ci/packer-templates/issues/221 and this is independent of this pull request. I suggest @heplesser as a reviewer.

SepehrMN:
I'll add tests today.

SepehrMN:
The Test is fine and works on my machine. There seems to be an issue with how the floating numbers are represented on different machines and I welcome suggestions.

SepehrMN:
Can we please establish that the test is being run correctly by Travis? I created a pynest test in unittests folder, but the error Travis gives is "Failed: missed SLI assertion" for a pynest test. make installcheck gives no errors on my machines.

SepehrMN:
It is possible that we have some numerical instability as in #466 that is caused by different packages etc., perhaps someone experienced can comment after we establish that the test is actually being run correctly.

heplesser:
@SepehrMN I will look into this. I had a look at your test script. It currently is a long Python script with many plain `assert()`s. Could you refactor it into a proper set of Python `unittest`s using the higher-level assertions provided by `unittest`? See, e.g., [`test_vogels_sprekeler_synapse.py`](https://github.com/nest/nest-simulator/blob/b5ede29a3f18580f5ecebad9a1459cd7d9f51b9a/pynest/nest/tests/test_vogels_sprekeler_synapse.py) or [`test_stdp_multiplicity.py`](https://github.com/nest/nest-simulator/blob/b5ede29a3f18580f5ecebad9a1459cd7d9f51b9a/pynest/nest/tests/test_stdp_multiplicity.py) for examples.

SepehrMN:
@heplesser I've made the changes and also some of the code, but the problem still persists. The tests still work fine on my machine.

On an unrelated matter, I will have to do a new push when request 343 (alias changes) gets merged.

heplesser:
@SepehrMN @lekshmideepu Your test works well on my machine, too (OSX 10.11.6, gcc 6.1.0, Python 2.7.11). When building with Python 3.5, though, I get the following errors:
```
ERROR: test_SegmentedSimulation (__main__.LowpassFilterSpikeDetector)
5) Checking to see if the result is correct if simulation is stopped and is continued many times.
----------------------------------------------------------------------
  File "/Users/plesser/NEST/code/src/testsuite/unittests/test_lowpassfilter_spike_detector.py", line 246, in test_SegmentedSimulation
    self.assertAlmostEquals(device_trace, python_trace)
  File "/Users/plesser/anaconda/envs/py35/lib/python3.5/unittest/case.py", line 1305, in deprecated_func
    return original_func(*args, **kwargs)
  File "/Users/plesser/anaconda/envs/py35/lib/python3.5/unittest/case.py", line 858, in assertAlmostEqual
    if round(abs(second-first), places) == 0:
TypeError: type numpy.ndarray doesn't define __round__ method

```
But on Travis, we have Python 2.7.6.

heplesser:
@SepehrMN You should include some tests that use multiple threads to make sure everything works then as well, and preferably MPI (but that test must be written in SLI). If I see you test correctly, it seems that you only test with spike_generator as input, which delivers locally. You should also test with input from normal neuron models.

SepehrMN:
@heplesser Thanks! I'll add them once we figure out why tests only fail on Travis.

abigailm:
Sepehr is checking the PR on a virtual machine of Travis to try to identify the cause of the failed build

SepehrMN:
@heplesser , @abigailm As I suggested before, the error seems to be with Travis not running the test correctly. The error I found on the VM is a simple import error. For some reason the path is not added when Travis runs the test:

```
----------------------------------------
Running test 'unittests/test_lowpassfilter_spike_detector.py'...
   > Traceback (most recent call last):
   >   File "/home/travis/build/SepehrMN/nest-simulator/result/share/doc/nest/u$
   >     import nest
   > ImportError: No module named nest
-> 1 (Failed: missed SLI assertion)
----------------------------------------

```

I think this is the only pynest test in that folder.

abigailm:
@SepehrMN well done for resolving the issue on Travis. Could you summarise how this issue can be avoided in future, so we can add it to the documentation?

models/lowpassfilter_spike_detector.cpp
```diff
@@ -0,0 +1,863 @@
...
+
+    if ( spikes_device_.is_precision_user_set() )
+    {
+      // if user explicitly set the precision, there is no need to do anything.
+      msg += ".";
+    }
+
+    else
+    {
+      // it makes sense to increase the precision if precise models are used.
```
abigailm: Is this compatible with #446?

SepehrMN: Yes it is! I already took care of it in c7ec080db108120165f63c22e7f81079a03a418b and 6f9c841a119e7c3075b8a1bfcee5a6ad04a8a24b .

SepehrMN: The next push will remove "user_set_precise_times_" variable that is no longer needed after #446.

SepehrMN:
The readme file in `/testsuite` includes this : 
```
  The tests in `unittests` and `regressiontests` may be written in either
  Python or SLI. They are executed by the corresponding executable according
  to their filename extension. Tests that are written in Python are run after
  SLI tests.
```
But when I put my pynest test in `/unittests` , it was failing because the path to nest was not found.

models/lowpassfilter_spike_detector.cpp
abigailm: Great, thanks for clarifying

abigailm:
Thanks for the explanation. We will look into this so it doesn't catch anyone else out.

jougs:
@SepehrMN, @abigailm: the [readme](https://github.com/nest/nest-simulator/blob/master/testsuite/README.md) is indeed misleading. What the sentence wants to say is that *general* Python scripts can be put there and test things (e.g. the copyright header checker would be very hard to write in SLI). Tests for PyNEST should go to the directory where you put them.

The reasoning behind this is that these tests are only run when PyNEST is actually compiled and the corresponding environment is set up properly.

Could you two please review #503, which clarifies this?

jougs:
Before going into a more thorough review, I have a general question: Why is this implemented as a new device instead of an extension to the existing `spike_detector`? I have a strong feeling that it duplicates a lot of the functionality and code of the normal `spike_detector` making it hard to keep them in sync and porting bug fixes and improvements between them.

SepehrMN:
It has all of the functionality of a spike_detector (which can be turned on/off to save resources), but it has an overhead that may make spike_detector slower for everyone if combined. I did not want spike_detector to become complex like the multimeter which can seem overwhelming to the uninitiated.

SepehrMN:
I want to add that as previously explained, the computation performed can be changed by modifying two lines, and for that, a new class would not be necessary. However, as a consequence the code could grow larger which supports the argument that this should be a different class.

models/lowpassfilter_spike_detector.cpp
```diff
@@ -0,0 +1,856 @@
...
+   * Stage 3: Recording filtered data.
+   * Committing the processed data to memory or printing it to file/screen.
+   */
+  if ( B_.steps_to_filter_.size() != 0
+    && ( ( origin.get_steps() + from )
+           % kernel().connection_manager.get_min_delay()
+         == 0 ) )
+  {
+
+    if ( filtered_device_.to_file() || filtered_device_.to_screen() )
```
SepehrMN: The data to be recorded are on the new class so this has to happen here.

models/lowpassfilter_spike_detector.h
```diff
@@ -0,0 +1,332 @@
...
+  delay in the network (+ one resolution increment). For example, if the minimum
+  delay in the network is 1.5 ms,
+  but simulation resolution and filter_report_interval are 0.1, The results for
+  the first 1.5 ms will be recorded
+  once the simulation has exceeded 1.5 ms (i.e., nest.Simulate(x) where x is >=
+  1.6).
+  - Filtered traces and/or spikes are not necessarily written to file in
+  chronological order.
+  - To get the most performance when only needing the trace at certain times,
+  set start_times and start_times
```
SepehrMN: Thank you for spotting the typo.

SepehrMN:
Herr @weidel-p, many thanks for review and comments. I can look at modularizing the update method after r2. I think certain things should be left there. For example, the main loop over events. My goal is that it is kept similar to the spike_detector so that it would be easy to understand and transfer changes.

models/lowpassfilter_spike_detector.cpp
```diff
@@ -0,0 +1,856 @@
...
+      + std::ceil( ( ( double ) ( ( update_start + 1 )
+                       - ( filterblock_start_step + interval_step ) )
+          / interval_step ) ) * interval_step;
+
+    if ( ( filter_step - ( filterblock_start_step + interval_step ) )
+        % interval_step
+      != 0 )
+    {
+      filter_step = filter_step - ( filter_step % interval_step );
+    }
```
SepehrMN: Addressed. This is tested and works fast. In a separate pull request, attempts can be made to incorporate 2.2 into 2.1 or replace 2. with for loops (which have drawbacks compared to this and may or may not seem simpler).

weidel-p:
Dear @SepehrMN, sorry for the late reply and thank you for addressing the comments. You already started modularize the 'update' function by defining different phases in the comments of the code. But I still think the function is too long and complex. For the sake of comprehensibility, it should be restructured before merging.

SepehrMN:
Many thanks. Before working on that we must establish that r2 is fine with the algorithms.

jougs:
Dear @SepehrMN, yes, I am also fine with the algorithms. Can you please proceed as @weidel-p detailed? I am in personal contact with him and will review the code once it is modularized. Thanks!

SepehrMN:
I realized that installcheck prints this during testing:

1) Checking default values. ... ok
3) Setting invalid filter blocks. ... ok
2) Checking to see if setting Status works. ... ok

The tests are not in order and there are many more tests after those. It runs perfectly fun with manual python run and they pass. I do not have time to investigate this anytime soon.

SepehrMN:
If there are failures in function, please report them here. There are issues for formatting. I can try to find time to attend to other matters in the next couple of months. Modularized version looks more complicated to me.

heplesser:
Closing due to inactivity.

sepehrmn:
I want to reopen this PR since I can afford some resources to see to its completion now.

Much has changed since it was written. I guess i'd have to look at nestio. Where can I look for documentation?

I don't suppose the windowing formulation I wrote for this has been implemented elsewhere yet?

sepehrmn:
@heplesser  I want to reopen this PR since I can afford some resources to see to its completion now.

Much has changed since it was written. I guess i'd have to look at nestio. Where can I look for documentation?

I don't suppose the windowing formulation I wrote for this has been implemented elsewhere yet?

heplesser:
@sepehrmn I would suggest you create an issue proposing this detector type as a new feature first, to have a place for discussion, and then create a new PR instead of reopening this one. Once the new issue is in place, @jougs could maybe comment wrt NESTIO?

Issue #438, opened 2016-08-01
Multisynaptic neuron model with independent synaptic rise times and decay times


golosio:
The multisynaptic neuron models that are currently available do not allow one to define independent synaptic rise times and decay times. The alpha function depends on a single time constants, and it is not always suitable for modeling synaptic conductance. On the other end, the time course of most synaptic conductances can be well described by the sum of two exponentials (see A. Roth and M.C.W. van Rossum in Computational Modeling Methods for Neuroscientists, MIT Press 2013, Chapter 6). It would be important to have a multisynaptic neuron model with synaptic conductance modeled by a double exponential function, with independent synaptic rise times and decay times

DimitriPlotnikov:
@golosio You pull request still doesn't pass through CI. It would be very helpful, if you would get all tests pass.

@abigailm @heplesser aeif_cond_exp_multisynapse is an example of a neuron model, that currently doesn't integrate well with NESTML approach. Generally, it is not an issues to deal with in NESTML. Here is an example of model with 2 constants implemented: ( cf. https://github.com/nest/nestml/blob/master/models/iaf_cond_beta.nestml ).

golosio:
@DimitriPlotnikov Thank you. The build process and the install check work fine in my machine. I'm new with nest development and I still have to understand how the Travis check works.

DimitriPlotnikov:
@heplesser What is the general policy regarding including new models into NEST?

golosio:
@DimitriPlotnikov accordind to the guidelines in the developer space: http://nest.github.io/nest-simulator/code_review_guidelines, concerning new models, "each pull request needs an OK from the CI platform and at least two reviewers to be merged". Meanwhile, my pull request passed through the CI. It was blocked by very little formatting problems, the rules for formatting are very strict.

DimitriPlotnikov:
@golosio @heplesser 
Cf. my latest comment in the pull request.

DimitriPlotnikov:
Since there is already an active pull request solving this issues and this issue being quite  non-informative, from my point of view we can close it.

golosio:
ok

PR #261, opened 2016-03-05
Gif models

hesam-setareh:
Implementing generalized integrate-and-fire (GIF) neuron model for nest in four versions:
gif_psc_exp
gif_psc_exp_multisynapse
gif_cond_exp
gif_cond_exp_multisynapse

And also a unittest for each version.

PR #439, opened 2016-08-01
Added neuron model aeif_cond_beta_multisynapse using GSL ODE solver (see issue 438)

golosio:
Multisynaptic neuron model with synaptic conductance modeled by the sum of two exponential functions, with independent synaptic rise times and decay times, as described in A. Roth and M.C.W. van Rossum in Computational Modeling Methods for Neuroscientists, MIT Press 2013, Chapter 6. The model is an extension of the conductance based exponential integrate-and-fire neuron model according to Brette and Gerstner (2005), as implemented in aeif_cond_alpha_multisynapse
This pull request deal with the issue  https://github.com/nest/nest-simulator/issues/438


heplesser:
@golosio Thank you for your contribution! We are currently trying to properly understand and handle some numerical issues with aeif models in general (see #229). I would prefer to solve them in the simple case of the `aeif_cond_alpha` model first, before adding new aeif models. In that way, we can ensure that all new models will use the best numerics. I hope that this is okay for you.

Another pre-requisite for merging your model into the NEST master repository is a signed Contributor License Agreement, which you can [download from our website](https://nest.github.io/nest-simulator/NEST_Contributor_Agreement.pdf) and send me by mail.

golosio:
@heplesser Sure, that's fine. Meanwhile I added the (equivalent) model aeif_cond_beta_multisynapse. I compared the two and they give the same results, which is not surprising as the underlying differential equations have the same solutions. aeif_cond_beta_multisynapse is closer to the original model aeif_cond_alpha_multisynapse, and it is based on the same differential equations as those proposed by dnao in "Create iaf_cond_beta" #405 . so it should be a quicker review, when you feel you can do it. By the way, if you think it can be useful I can help with the revision of iaf_cond_beta. Let me know...

heplesser:
@golosio I am a bit confused: isn't the beta function precisely the difference-of-two-exponentials? So what precisely  is the difference between the `_beta_` and the `_2exp_` model?

Btw, our preferred integrator for AEIF neurons is at present the GSL ODE solver used in `aeif_cond_alpha`. Would you consider to convert your code to that solver? In #229, we are currently looking at even better solvers that handle threshold detection as part of the integration.

Concerning contributions to #405, I'd suggest that you contact @dnao directly, since it is his/her code.

golosio:
Right, the beta function is the difference of two exponential. The two models differ because they are based on a slightly different (but equivalent) system of differential equations, which have the same solution (i.e. the beta function) and it is probably superfluous to have both, but it was useful for me as a check of consistency. Sure, I'll have a look at the GSL ODE solver and see how difficult it would be to adapt my code for it.

heplesser:
@golosio Good. While cross-validation of models is very useful, we want to keep the number of models in the NEST distribution within sensible limits. So I would suggest that you remove the `_2exp_` variant.

golosio:
Sure, that's easy!

golosio:
@heplesser I removed the 2exp variant

golosio:
@heplesser ok, I'm working on the gsl ode version, it should not be too much work if I use aeif_cond_alpha as a base.

golosio:
@heplesser ok, I adapted my code for using the GSL ODE solver and tested it. You can find the new version in the updated pull request.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -0,0 +1,730 @@
...
+  , E_in( -85.0 )     // mV
+  , E_L( -70.6 )      // mV
+  , Delta_T( 2.0 )    // mV
+  , tau_w( 144.0 )    // ms
+  , a( 4.0 )          // nS
+  , b( 80.5 )         // pA
+  , V_th( -50.4 )     // mV
+  , I_e( 0.0 )        // pA
+  //  , MAXERR( 1.0e-10 ) // mV
+  //  , HMIN( 1.0e-3 )    // ms
```
heplesser: Please remove commented-out code.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -0,0 +1,162 @@
...
+multisynapse_neuron << /taus_decay taus_decay >> SetStatus
+multisynapse_neuron << /taus_rise taus_rise >> SetStatus
+multisynapse_neuron << /V_peak V_peak >> SetStatus
+multisynapse_neuron << /a a >> SetStatus
+multisynapse_neuron << /b b >> SetStatus
+multisynapse_neuron << /E_L Vrest >> SetStatus
+multisynapse_neuron << /g_L g_L >> SetStatus
+multisynapse_neuron << /E_in E_in >> SetStatus
+
+multisynapse_neuron ShowStatus
```
heplesser: Useful for development, but I suggest to remove it since it will not be of use in the actual test.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -0,0 +1,162 @@
...
+% creates an array of synaptic indexes 
+delays length 1 arraystore Range /synapses_idx exch def
+
+[delays weights synapses_idx] % loop on synaptic receptor ports
+{
+	/syn_id exch def
+	/W exch def
+	/delay exch def
+	% connect spike generator to each port
+	[sg] [multisynapse_neuron] /one_to_one << /model /static_synapse /delay delay /weight W /receptor_type syn_id >> Connect
```
heplesser: Break the line into several shorter lines.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -0,0 +1,162 @@
...
+
+   Description:
+   This test creates a multisynapse neuron with three receptor ports with
+   different synaptic rise times and decay times, and connect it to
+   two excitatory and one inhibitory signals. If the rise times and decay
+   times are much greater than the membrane time constant, the difference
+   between the membrane potential and the rest potential V(t) will be
+   approximately proportional to the synaptic conductance g(t), which can
+   be computed analitically. At the end, the script compares the simulated
+   values of V(t) with the approximate theoretical prediction.
```
heplesser: Could you add a derivation (or reference to a published derivation) of the approximation you use as reference?

heplesser:
@golosio This looks very fine now, just a few details left, see my inline comments. I will soon suggest a second reviewer.

golosio:
@heplesser I made all the changes that you indicated. As you suggested, I added a shortened version of the derivation as a comment to the sli test. Furthermore I included the possibility of the limiting case when the synaptic rise time is equal to the decay time, which corresponds to the alpha function. This can be useful because it gives the possibility to have, in the same neuron, both types of models on different synapses. Just to let you know, I made an additional check of consistency, setting the decay time very close to the rise time, and compared the V(t) curve with that obtained using the aeif_cond_alpha_multisynapse model, and they superimpose, as expected.

golosio:
There seem to be a problem with mpi
Running test 'selftests/test_pass.sli'... *** Error in `mpirun': munmap_chunk(): invalid pointer
I don't think this can depend on the changes that I made...

jougs:
@golosio this crash is most likely completely unrelated to your changes. `fglrx` is the graphics driver for AMD graphics cards and I have no clue why a reload of this was triggered. I've restarted the jobs.

golosio:
@jougs ok, thank you! I've merged the latest changes from the upstream repository and tried to start the check again, let's see...

jougs:
See also #454 for failing builds

golosio:
@jougs you're right, I should have paid attention to this as it was discussed in the VC. As I understand I have to replace double_t and long_t with double and long, but what about size_t?

jougs:
@golosio: to reduce confusion: the failing builds are most likely not related to the replacement of types.

But yes, you should replace `double_t` and `long_t` as you suggested to reduce work later. `size_t` is a standard C++ type and should stay.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -466,13 +464,20 @@ aeif_cond_beta_multisynapse::calibrate()
...
-      = ( 1. / P_.taus_rise[ i ] - 1. / P_.taus_decay[ i ] )
-      / ( std::exp( -t_p / P_.taus_decay[ i ] )
-          - std::exp( -t_p / P_.taus_rise[ i ] ) );
+    if ( P_.taus_decay[ i ] == P_.taus_rise[ i ] ) // alpha function limit
+    { // use normalization for alpha function in this case
+      V_.g0_ex_[ i ] = V_.g0_in_[ i ] = 1.0 * numerics::e / P_.taus_decay[ i ];
+    }
+    else
+    {
+      double t_p = P_.taus_decay[ i ] * P_.taus_rise[ i ]
```
heplesser: Could be const.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -466,13 +464,20 @@ aeif_cond_beta_multisynapse::calibrate()
...
-    double t_p = P_.taus_decay[ i ] * P_.taus_rise[ i ]
-      / ( P_.taus_decay[ i ] - P_.taus_rise[ i ] )
-      * std::log( P_.taus_decay[ i ] / P_.taus_rise[ i ] ); // peak time
-    V_.g0_ex_[ i ] = V_.g0_in_[ i ] // normalization factors for conductance
-      = ( 1. / P_.taus_rise[ i ] - 1. / P_.taus_decay[ i ] )
-      / ( std::exp( -t_p / P_.taus_decay[ i ] )
-          - std::exp( -t_p / P_.taus_rise[ i ] ) );
+    if ( P_.taus_decay[ i ] == P_.taus_rise[ i ] ) // alpha function limit
+    { // use normalization for alpha function in this case
+      V_.g0_ex_[ i ] = V_.g0_in_[ i ] = 1.0 * numerics::e / P_.taus_decay[ i ];
```
heplesser: I would suggest to avoid statements with serial assignments, there is always a risk that one overlooks the "sandwiched" assignment, especially if the expression to the right is not totally trivial (eg `a = b = c = 0`).

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -466,13 +464,20 @@ aeif_cond_beta_multisynapse::calibrate()
...
+    }
+    else
+    {
+      double t_p = P_.taus_decay[ i ] * P_.taus_rise[ i ]
+        / ( P_.taus_decay[ i ] - P_.taus_rise[ i ] )
+        * std::log( P_.taus_decay[ i ] / P_.taus_rise[ i ] ); // peak time
+      V_.g0_ex_[ i ] = V_.g0_in_[ i ] // normalization factors for conductance
```
heplesser: See above concerning serial assignment.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -217,10 +215,10 @@ aeif_cond_beta_multisynapse::Parameters_::set( const DictionaryDatum& d )
         throw BadProperty(
           "All synaptic time constants must be strictly positive" );
       }
-      if ( tau_tmp[ i ] >= taus_decay[ i ] )
+      if ( tau_tmp[ i ] > taus_decay[ i ] )
```
heplesser: Is there any strict mathematical reason why the rise time must be shorter than the decay time? This is the usual case in neuroscience, but I think the mathematics will work perfectly well otherwise, too, so I would remove this constraint.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -466,13 +464,20 @@ aeif_cond_beta_multisynapse::calibrate()
...
   for ( size_t i = 0; i < P_.num_of_receptors_; ++i )
   {
-    double t_p = P_.taus_decay[ i ] * P_.taus_rise[ i ]
-      / ( P_.taus_decay[ i ] - P_.taus_rise[ i ] )
-      * std::log( P_.taus_decay[ i ] / P_.taus_rise[ i ] ); // peak time
-    V_.g0_ex_[ i ] = V_.g0_in_[ i ] // normalization factors for conductance
-      = ( 1. / P_.taus_rise[ i ] - 1. / P_.taus_decay[ i ] )
-      / ( std::exp( -t_p / P_.taus_decay[ i ] )
-          - std::exp( -t_p / P_.taus_rise[ i ] ) );
+    if ( P_.taus_decay[ i ] == P_.taus_rise[ i ] ) // alpha function limit
```
heplesser: Testing doubles for equality is not really safe. What you in the end need is that
```
P_.taus_decay[ i ] - P_.taus_rise[ i ] != 0
```
and
```
std::exp( -t_p / P_.taus_decay[ i ] ) - std::exp( -t_p / P_.taus_rise[ i ] ) != 0
```
These expressions may be zero even if the rise and decay time differ slightly. `doc/model_details/IAF_neurons_singularity.ipynb` discusses how we handle singularities due to `tau_m ~ tau_syn` in linear models; you may want to take a look there.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -0,0 +1,735 @@
...
+
+  // The following code is verbose for the sake of clarity. We assume that a
+  // good compiler will optimize the verbosity away ...
+
+  // shorthand for state variables
+  const double_t& V = y[ S::V_M ];
+  // const double_t& dg_ex = y[ S::DG_EXC ];
+  // const double_t& g_ex = y[ S::G_EXC ];
+  // const double_t& dg_in = y[ S::DG_INH ];
+  // const double_t& g_in = y[ S::G_INH ];
```
heplesser: Please remove the commented-out code.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -0,0 +1,418 @@
...
+ being the index of the last element of the tau_ex and tau_in
+ arrays).
+ During connection, the ports are selected with the property "receptor_type".
+
+ Examples:
+ % PyNEST example, of how to assign synaptic rise time and decay time
+ % to a receptor type.
+
+ nest.SetDefaults('aeif_cond_beta_multisynapse', {'HMIN':0.001})
+ nest.SetDefaults('aeif_cond_beta_multisynapse', {'MAXERR':1e-10})
```
heplesser: I think these two parameters have been replaced by tolerance parameters for the GSL solver?

models/aeif_cond_beta_multisynapse.h
```diff
@@ -0,0 +1,418 @@
...
+ nest.SetStatus(neuron, {'taus_decay':[50.0,20.0,20.0,20.0],
+                         'taus_rise':[10.0,10.0,1.0,1.0]})
+ spike1 = nest.Create('spike_generator', params = {'spike_times':
+                                                  np.array([10.0])})
+ spike2 = nest.Create('spike_generator', params = {'spike_times':
+                                                  np.array([10.0])})
+ spike3 = nest.Create('spike_generator', params = {'spike_times':
+                                                  np.array([10.0])})
+ spike4 = nest.Create('spike_generator', params = {'spike_times':
+                                                  np.array([10.0])})
```
heplesser: Since all spike generators produce exactly the same spike train, you only really need one generator here, which then connects to the one neuron through four `Connect()` calls with a different synapse type each.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -0,0 +1,418 @@
...
+   * Internal variables of the model.
+   */
+  struct Variables_
+  {
+
+    /** initial value to normalise excitatory synaptic conductance */
+    std::vector< double_t > g0_ex_;
+
+    /** initial value to normalise inhibitory synaptic conductance */
+    std::vector< double_t > g0_in_;
```
heplesser: This will always have exactly the same values as `g0_ex_`, so you really need only one `g0_` vector.

heplesser:
@golosio Good work! I added a few more comments on details and noticed one more easy-to-fix but principal matter, but I will open a new issue for that, since it affects `aeif_cond_alpha_multisynapse` as well.

jougs:
Our failing builds actually seem to be a Travis bug. See https://github.com/travis-ci/packer-templates/issues/221 for the issue they opened after my mail.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -459,24 +459,31 @@ aeif_cond_beta_multisynapse::calibrate()
...
 
   for ( size_t i = 0; i < P_.num_of_receptors_; ++i )
   {
-    if ( P_.taus_decay[ i ] == P_.taus_rise[ i ] ) // alpha function limit
-    { // use normalization for alpha function in this case
-      V_.g0_ex_[ i ] = V_.g0_in_[ i ] = 1.0 * numerics::e / P_.taus_decay[ i ];
-    }
-    else
+    // denominator is computed here to check that it is != 0
+    double denom1 = P_.taus_decay[ i ] - P_.taus_rise[ i ];
```
heplesser: Your code is correct, but the logic a bit difficult to follow, especially the `if ( denom2 == 0 )` at the end. Since this code is in `calibrate()`, thus rarely executed, I would give legibility precedence over performance and use
```c++
const double denom1 = ...;
const double denom2 = ...;
if ( denom1 == 0 or denom2 == 0 )
{
  // alpha case
}
else
{ 
  // beta case
}

**Comment by heplesser** [comment]
@golosio If you include the prefactor (Roth & van Rossum, Eqs 6.4-6.6), the beta function is fully symmetric under exchange of rise and decay times. This means two things: (i) The function will always be positive, no matter which of the two times is larger. (ii) The shorter of the two times will always be the rise time, no matter what it is called. Thus, requiring tau_rise <= tau_decay makes sense, because otherwise one will just get confusing behavior.

I overlooked the problem with the restriction to a single synapse type (even though I implemented it myself, see trac ticket 737). The restriction is in place to ensure that a callback mechanism used by some stimulating devices works correctly.

But you don't need the different synapse models at all, you can just pass the parameters directly in the connect call:
```python
nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
                                      'weight': 1.0,
                                      'delay': 1.0,
                                      'receptor_type': 1})
nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
                                      'weight': 1.0,
                                      'delay': 100.0,
                                      'receptor_type': 2})
nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
                                      'weight': 1.0,
                                      'delay': 300.0,
                                      'receptor_type': 3})
nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
                                      'weight': 1.0,
                                      'delay': 500.0,
                                      'receptor_type': 4})
```
or even more compact (with slightly different delay values):
```python
for syn in range(4):
    nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
                                          'weight': 1.0,
                                          'delay': 1.0 + syn * 200.,
                                          'receptor_type': 1 + syn})
```

models/aeif_cond_beta_multisynapse.cpp
golosio: @heplesser I agree, indeed first I tried to write the code as you suggest, but the problem is that denom2 depends on t_peak, and before computing t_peak I must check that denom1 is !=0...
I could not find a more readable solution...

heplesser: @golosio I should think before I comment ... But how about the following:
```c++
const double denom1 = ...;
double denom2 = 0;
if ( denom1 != 0 )
{
  const double t_p = ...;
  denom2 = ...;
}
if ( denom2 == 0 )  // if denom1 == 0, denom2 == 0 is implied, no need to test denom1 here again
{
  // alpha
}
else
{
  // beta
}
```
This avoids have an `if ( denom2 !=0 )` in one place and an `if ( denom2 == 0 )` a few lines further down.

golosio: yes, this looks fine!

golosio:
@heplesser Done!

heplesser:
@golosio  :+1: from me, and thanks for the effort! Before merging, we should still wait until the test system works again for all configurations.

@jougs Any comments?

golosio:
@heplesser thank you for your constructive review!

heplesser:
@jougs Do you have any objections to merging this PR now that all tests have passed?

jougs:
@heplesser: I was just commenting here because of the failing builds, not as a reviewer.

Maybe @ingablundell or @DimitriPlotnikov could have a look?

DimitriPlotnikov:
@heplesser @jougs my first impression is that the model is implemented well. Before I proceed with more detailed review, I would have an acknowledgment that the model is important enough to be included into NEST repository.

golosio:
Thank you @DimitriPlotnikov.

heplesser:
@DimitriPlotnikov Yes, the model is relevant (otherwise I wouldn't have given it a thumb up).

DimitriPlotnikov:
@heplesser Thank you for your replay. I still a bit confused with this 'kitschy' emojies which can appear at arbitary place in the discussion. I'll tackle it next week.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -0,0 +1,401 @@
...
+  {
+    Buffers_( aeif_cond_beta_multisynapse& );
+    Buffers_( const Buffers_&, aeif_cond_beta_multisynapse& );
+
+    //! Logger for all analog data
+    UniversalDataLogger< aeif_cond_beta_multisynapse > logger_;
+
+    /** buffers and sums up incoming spikes/currents */
+    std::vector< RingBuffer > spike_exc_;
+    std::vector< RingBuffer > spike_inh_;
```
DimitriPlotnikov: Here I need qualified help of @heplesser:

We already discussed the issue with our interpretation of inhibitory and excitatory buffers in the current NEST state. In a model with one buffer (iaf_psc_exp) the logic to use the sigh of the weight to route spike events is at least plausible. In the case of the mutlisynapse model it leads to: a) you always have the same number of inh und exc buffers b) you have to buffers (inh/exc) at the same portnumber c) finally, this implementation is inconsistent with the implementation of the iaf_psc_alpha_multisynapse and iaf_psc_exp_multisynapse.

golosio:
@DimitriPlotnikov I took as reference aeif_cond_alpha_multisynapse, but I agree with your observations. If you already discussed this issue for the iaf_psc_*_multisynapse models, probably there is no reason to keep two separate buffers for inhibitory and excitatory connections. We have to be careful that in conductance based models the driving force is different for inhibitory vs excitatory signals, but this should not be a big deal. I will try to modify the code and let you  know.

golosio:
@DimitriPlotnikov I modified the code according to your comment. Now there is only one buffer for both types of events, inhibitory or excitatory.
I compared the results of the test, which uses both exc and inh connections, with the old and with the updated code, and they are exactly the same.
Probably this modification should be done to all conductance-based multisynapse and multicompartment models.

heplesser:
@golosio @DimitriPlotnikov The issue of the exc/inh buffer also occurs in #261 for the gif_cond_exp_multisynapse models. I put it on the agenda for the NEST developer VC on 19 Sept.

golosio:
@heplesser @DimitriPlotnikov maybe we should seek a more general solution (see the issue that i just opened  https://github.com/nest/nest-simulator/issues/477 ). My opinion is that I would not ask Hesam to review his code once more for this issue before merging it, after all it only affects performances, it can be fixed at a later stage.

golosio:
@DimitriPlotnikov Did you have a chance to check the code with the changes that I made to meet your observation? Do you have other amendments to propose?

DimitriPlotnikov:
I'll do it tomorrow.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -0,0 +1,701 @@
...
+  }
+
+  V_.g0_.resize( P_.num_of_receptors_ );
+
+  for ( size_t i = 0; i < P_.num_of_receptors_; ++i )
+  {
+    // denominator is computed here to check that it is != 0
+    double denom1 = P_.taus_decay[ i ] - P_.taus_rise[ i ];
+    double denom2 = 0;
+    if ( denom1 != 0 )
```
DimitriPlotnikov: Could you please add an informative comment why are you doing this computation and checks? Can it be already detected in the Parameters_::set?

models/aeif_cond_beta_multisynapse.h
```diff
@@ -0,0 +1,397 @@
...
+ nest.Connect(voltmeter, neuron)
+
+ Sends: SpikeEvent
+
+ Receives: SpikeEvent, CurrentEvent, DataLoggingRequest
+
+ Adapted by Bruno Golosio from aeif_cond_alpha_multisynapse
+ SeeAlso: aeif_cond_alpha_multisynapse
+ */
+
```
DimitriPlotnikov: I tried to run the example that you provide in the commet. I would add 
```
import nest 
import numpy as np

```
to make it an executable python program.

Another issue is that the resulting script fails at my machine with the following stacktrace:
```

  File "aeif_cond_alpha_multisynapse.py", line 24, in <module>
    nest.Connect(spike, neuron, syn_spec="synapse2")
  File "/home/nestml/repositories/nest_repo/bld_master_nompi/lib/python2.7/site-packages/nest/lib/hl_api_helper.py", line 230, in stack_checker_func
    return f(*args, **kwargs)
  File "/home/nestml/repositories/nest_repo/bld_master_nompi/lib/python2.7/site-packages/nest/lib/hl_api_connections.py", line 314, in Connect
    sr('Connect')
  File "/home/nestml/repositories/nest_repo/bld_master_nompi/lib/python2.7/site-packages/nest/__init__.py", line 91, in catching_sli_run
    errorname, commandname, message))
pynestkernel.NESTError: IllegalConnection in Connect_g_g_D_D: Creation of connection is not possible because:
All outgoing connections from a device must use the same synapse type.
```
Am I doing something wrong? Otherwise, could you fix the comment?

golosio:
@DimitriPlotnikov Thank you for your review. I made the changes that you requested. I improved the comments and completed the example. I already noticed that the example provided in aeif_cond_alpha_multisynapse don't work, see my comments at the end of https://github.com/nest/nest-simulator/pull/439#issuecomment-239041306 and the following reply from @heplesser . On the other hand, I fixed my  example and it should work.
It's interesting that you used the model in a Brunel simulation. I'm also working on a kind of balanced network, I'm trying to implement with a more realistic neuron model the network described in Lim S, Goldman M, Nature Neuroscience 16, 1306–1314 (2013) http://www.nature.com/neuro/journal/v16/n9/full/nn.3492.html

DimitriPlotnikov:
This PR can be merged.

golosio:
Thank you @DimitriPlotnikov

PR #441, opened 2016-08-01
Fix handling of params dictionary in Connect

### LINKED ISSUE #437: Connect ignores parameters for neuron->device and device->device connections
As discovered by @weidel-p #221 introduced a bug, which causes parameter dictionaries to be ignored when connecting a neuron to a device or a device to a device. This can be traced back to [connection_manager.cpp:450](https://github.com/nest/nest-simulator/blob/master/nestkernel/connection_manager.cpp#L450) and [connection_manager.cpp:461](https://github.com/nest/nest-simulator/blob/master/nestkernel/connection_manager.cpp#L461), where both function calls are missing the `params` parameter. I'll open a PR later.

jakobj:
If Connect was called with a parameter dictionary, this dictionary was ignored when connecting a normal node to a device or a device to a device. This is fixed here and a regression test is added. Thanks to @weidel-p for initial analysis of the bug. I suggest @weidel-p and @gtrensch as reviewers.
Fixes #437

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -446,7 +446,7 @@ nest::ConnectionManager::connect( index sgid,
 
     if ( source->has_proxies() ) // normal neuron->device connection
     {
-      connect_( *source, *target, sgid, target_thread, syn, d, w );
+      connect_( *source, *target, sgid, target_thread, syn, params, d, w );
     }
     else // create device->device connections on suggested thread of target
     {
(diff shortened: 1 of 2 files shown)


weidel-p:
The code and regressiontest looks nice and clean. Most important though, it solves the issue I had. 

:+1: from my side.

gtrensch:
I ran a couple tests with MUSIC version 1.1.15 using two channel connections (This is where @weidel-p went into that issue.). Everything works fine now. Also negative testing, particularly the regression test, behaves as expected.

:+1:  also from my side.

Issue #347, opened 2016-05-18
 Write tests for the MUSIC interface 


jougs:
Currently, MUSIC is not covered by any tests. This should be changed.

weidel-p:
We could start by including MUSIC in the Travis CI test matrix. As MUSIC is based on MPI, it only makes sense to compile and test NEST with MUSIC for cases including MPI. 
I know our time on Travis is limited and I'm not sure if we can afford compiling and testing MUSIC in the given time. Did someone think about (or test) that already?

To test the functionality of MUSIC we could convert the MUSIC examples to Python tests.

heplesser:
One aspect to test is whether MUSIC respects spike multiplicity, i.e., whether the MUSIC out proxy generates n spikes when receiving a SpikeEvent with multiplicity n.

jougs:
@weidel-p: with #409 merged, the build time is not a big problem anymore and adding MUSIC to the maximal job is high on my agenda. You're welcome to help!

Basically we need three things in place for this:
- Installation of MUSIC in Travis and addition of the corresponding flags to the configuration (easy)
- A test harness to run jobs via the `music` launcher (medium)
- Tests!

I'm also adding @mdjurfeldt, so he is aware of this.

jougs:
This was fixed by the merge of https://github.com/nest/nest-simulator/pull/471. I'm thus closing.

PR #443, opened 2016-08-01
Ignore MUSIC devices in multiplicity test

### LINKED ISSUE #442: Error in testsuite when compiling with MUSIC
When compiled with MUSIC the test `test_neurons_handle_multiplicity.sli` fails with `Key '/V_m' does not exist in dictionary.`, since NEST tries to get `V_m` from a MUSIC device. These devices need to be added to the ignore dictionary. PR will follow.

jakobj:
Since MUSIC devices do not have a membrane potential, test_neurons_handle_multicplicity.sli fails when trying to retrieve the value of the membrane potential of a MUSIC device. These should hence be ignored in this test. I suggest @heplesser and @weidel-p as reviewers.
Fixes #442

### DIFF
--- a/testsuite/unittests/test_neurons_handle_multiplicity.sli
+++ b/testsuite/unittests/test_neurons_handle_multiplicity.sli
@@ -47,6 +47,12 @@ M_ERROR setverbosity
              /parrot_neuron
              /parrot_neuron_ps
              /iaf_cond_alpha_mc    % cannot readout V_m directly
+             /music_event_in_proxy        % music device
+             /music_event_out_proxy       % music device
+             /music_cont_in_proxy         % music device
+             /music_cont_out_proxy        % music device
+             /music_message_in_proxy      % music device
+             /music_message_out_proxy     % music device
            ] def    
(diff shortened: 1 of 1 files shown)


heplesser:
@jakobj I think your changes are fine, but they raise an interesting question: How can we test that the MUSIC proxy for spikes from NEST handles spike multiplicity properly?

:+1: for this one.

weidel-p:
@jakobj thanks for this fix. :+1: from me.

@heplesser I'm not sure about that, does MUSIC handle spike multiplicity at all? Maybe we can involve @mdjurfeldt in this discussion.

Also I'd like to reopen the discussion in #347. We should think about including tests for MUSIC functionality in Travis CI.

heplesser:
With two thumbs up I will merge shortly.

@weidel-p The only way MUSIC would have to handle multiplicity would be as follows: If a spike event arrives with multiplicity n, then the MUSIC out proxy will have to transmit n spikes with the time stamp of the event. I will add this to #347.

PR #444, opened 2016-08-02
Small updates to Eclipse instructions. [skip ci]

heplesser:
Minor updates to instructions for using NEST with Eclipse.

### DIFF
--- a/eclipse_workflow.md
+++ b/eclipse_workflow.md
@@ -133,7 +133,7 @@ $NEST_ROOT/bld_master_nompi/install     # install directory
 ```
 
 You should configure, build and install NEST manually once (note that
-I want to build NEST with gcc 5.1 from Homebrew, therefore the
+I want to build NEST with gcc 6.x from Homebrew, therefore the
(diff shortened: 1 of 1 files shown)


heplesser:
Suggest @DimitriPlotnikov as reviewer.

eclipse_workflow.md
```diff
@@ -129,7 +134,7 @@ $NEST_ROOT/bld_master_nompi/install     # install directory
 
 You should configure, build and install NEST manually once (note that
 I want to build NEST with gcc 5.1 from Homebrew, therefore the
-`-DCMAKE_C_COMPILER=gcc-5 -DCMAKE_CXX_COMPILER=g++-5` arguments to `cmake`;
+`-DCMAKE_C_COMPILER=gcc-6 -DCMAKE_CXX_COMPILER=g++-6` arguments to `cmake`;
 NB: Make sure that you have checked out the master branch in the `src` directory):
```
DimitriPlotnikov: OK, now I'm a bit confused. You are using gcc 5.1, but also the compiler option DCMAKE_C_COMPILER=gcc-6 (which is probably not a part of gcc5.1). Is that correct?

Issue #411, opened 2016-06-28
NEST developer version specifies version to '2.10.0'


mschmidt87:
The NEST developer version from github specifies its version number to '2.10.0' and is therefore easily confused with the official release version. This is true for the header that is shown on the startup of NEST and for `nest.version()`.

Optimally, this could instead give the commit hash of the source code that was actually used for the code.

DimitriPlotnikov:
Hi, 
it is true an issue in the current version of NEST. However, I don't like the proposal to use the any hashcodes based on commits. It could lead to a problem, that NEST user would start to relay on a particular temporary version development version. 

Therefore, my proposal would be to call the current version like: 3.0.0-SNAPSHOT. It unambiguously states that it is a development version of the 3.0.0 release. This scheme is adopted from Maven, but it is already proved to work.

@mschmidt87 Would this proposal solve your problem? Could your otherwise elaborate on advantages of your proposal?

mschmidt87:
Hi, your proposal would at least avoid confusion of the developer version with the release version. 

However, I still think that it would be better to have the commit hash displayed, because this enables true provenance tracking of which version of NEST you actually compiled and are using, when e.g. discovering a bug. So, whenever a user reports a bug or a problem with NEST, we can easily ask him for the exact version, which will make it easier to track the problem down.

DimitriPlotnikov:
Hi,

from my point of view, all bugs must be reported on the latest version. If it is not there anymore, then the bug is solved:) We will discuss it one more time during the next NEST developer meeting and meet the final decision then.

mschmidt87:
Okay, you have a point here. Still, I think it would be useful to have the executable indicate the exact commit hash, I have stumbled across this quite a few times during development and simulation work.

DimitriPlotnikov:
Hello @mschmidt87,  
your points are correct and valid from the developer's point of view.  But for the enduser a commit hash is not really informative.  github enables you to checkout a particular pullrequest (https://help.github.com/articles/checking-out-pull-requests-locally/) and/or particular commit. In case that you really need to work with th specific commit, you can do it and keep track of the hash information locally. 

Beside that I made a pull request fixing the problem.

apeyser:
But bug reports on the developer branch indicate that we are talking about the developer's point of view.The hash can go into a built file, and for release versions there'd be no hash, since it's a fixed release. 

[...]

PR #448, opened 2016-08-04
Set current NEST version to '3.0.0-SNAPSHOT'. 

DimitriPlotnikov:
The version number for the NEST Head in the repo until the next release will be 2.11


extras/create_rcsinfo.sh
```diff
@@ -6,7 +6,7 @@
 
 # default value, if git is not installed or if the source directory is
 # not under version control
-version="v2.10.0"
+version="v3.1.0-SNAPSHOT"
```
heplesser: Shouldn't this be `v3.0.0-SNAPSHOT`? And wouldn't it be better to build this string automatically from `NEST_VERSION_VERSION` defined in `CMakeLists.txt`?

heplesser:
In principle ok, provided my inline comments are addressed.

@jougs @apeyser @mschmidt87 What to you think?

apeyser:
@heplesser 

As per issue #411 , I'd suggest we go to even/odd numbering. It's not a big deal, but if we go with -SNAPSHOT, immediately at the time of a release, we have to commit to whether the next release is a big release or a small release. We don't have an option without confusing ourselves about what's going on.

We've already had this happen with 3.0, with discussion after 2.8 whether the next release was a 2.x or a 3.0.

extras/create_rcsinfo.sh
apeyser: Hmm, making a file-creator script creator. Sounds complex. I guess it could grep out of config.h as well, or use a c preprocessor to extract it. A bit complex as well. Maybe source a very small file that does the version definition made by cmake?

Or maybe just stick this all in cmake and just add an operator that pulls it from config.h?

Tracking version control information inside of projects is always painful, and I tend to think not so useful without the build infrastructure, and if you have the build infrastructure, you already have that information in the tree.

Maybe we only want to track "Is this release version X" or "You built this yourself, you should know how to give us the hash"?

heplesser:
@apeyser I agree with you in principle. My thought right now was that 3.0 is going to be the next big one, so we could use this patch as a temporary fix and then define a long term policy in our meeting in two weeks. But we might as well wait two+ weeks and then fix it for real.

apeyser:
@heplesser Then we should accept, if the decision is fixed that we definitely will not be putting out any other major revision than 3.0. No practical consequence. Then afterwards, we can decide.

terhorstd:
I would agree to apeyser that we don't initially want to commit ourselves to the upcomming version number. What about first sorting the current issues to milestones and then actually planning the versions a bit more in advance? I don't see a reason why there shouldn't be a 2.12 and a 3.0 milestone in parallel. We could properly sort the issues with major changes to the new major version and the others to 2.12.

extras/create_rcsinfo.sh
apeyser: http://xit0.org/2013/04/cmake-use-git-branch-and-commit-details-in-project/

That could simplify (eliminate) all this code, if we really want to insert the hashes in sli; on the other hand, inserting it in sli instead of c defeats the goal of making a complete and clear c api.

heplesser:
We decided in the [Open NEST Developer VC today](https://github.com/nest/nest-simulator/wiki/2016-08-08-Open-NEST-Developer-Video-Conference) that we will use odd minor version numbers numbers for NEST in between releases, with no further detailing. Thus, the version number for the NEST Head in the repo until the next release will be `2.11`.

We may create `-RC` version numbers on release branches shortly before a release.

Issue #412, opened 2016-06-28
CMake confused by simultaneous installation of Python2 and Python3


tammoippen:
When using older versions of `cmake` (before version 3.4, e.g. [here](https://github.com/Kitware/CMake/blob/v3.3.2/Modules/FindPythonLibs.cmake)) the script to find the Python libraries is independent from the script to find the Python interpreter. Thus, it is possible that Python 2.7 interpreter is found, but NEST is linked against Python 3 libraries. In this case, loading the `pynestkernel` does not work.

With `cmake` 3.4 and following, [`FindPythonLibs`](https://github.com/Kitware/CMake/blob/v3.4.0/Modules/FindPythonLibs.cmake#L54) uses the Python interpreter as hint and looks for the corresponding libraries / include directory.

If you are using an earlier `cmake`, there is a workaround: specify the correct library and include directory on the `cmake` line:

```
cmake  ..... \
    -DPYTHON_LIBRARY=<path/to/python/lib.so> \
    -DPYTHON_INCLUDE_DIR=<path/to/where/Python.h/is/found> \
    <src>
```

To fix this issue, the NEST build system should either figure out the correct library / include or warn the user in case of different versions / prefixes / ... and hint to the workaround.

apeyser:
@jougs @gtrensch 
What cmake are we supporting? If it's >=3.4, we can close this as won't fix.

jougs:
Can we make this so, then? Ubuntu Xenial Xerus(16.04) has 3.5.1. I'd rather not invest time in supporting old packages that can either be gotten from the distribution or source.

If we agree that we drop support for older versions, we need to adapt [INSTALL](https://github.com/nest/nest-simulator/blob/master/INSTALL), which currently says 
> NEST is installed with cmake (at least v2.8.12)

tammoippen:
Keep in mind, that Juqueen and other compute clusters might have older versions installed. We would require users to install cmake in this case.

jougs:
@tammoippen: true.

There are pre-built statically linked versions on the cmake homepage, so installation is not so complicated. but I'm not sure if we should burden people with that if they just happen to have an older distribution.

On the other side, I seem to be the only one hitting this problem regularly and that certainly does not justify to change code.

apeyser:
@jougs 
I hit it regularly (as in today on BGQ, yesterday on my desktop...). 
"CMake considered harmful" is my catch phrase of the day.

[...]

PR #452, opened 2016-08-05
Prevent Python version mix and introduce NEST build option -Dwith-python=ON/OFF/2/3 

gtrensch:
Proposed solution / workaround for issue #412.

@apeyser please can you have a look at it ?

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -81,8 +81,8 @@ Set default libraries:
     -Dwith-ltdl=[OFF|ON|</path/to/ltdl>]         Find a ltdl library. To set a specific
                                                  ltdl, set install path.
                                                  [default=ON]
-    -Dwith-python=[OFF|ON|</path/to/python>]     Build PyNEST. To set a specific Python,
-                                                 set install path. [default=ON]
+    -Dwith-python=[OFF|ON|2|3]                   Build PyNEST. To set a specific Python
+                                                 version, set 2 or 3. [default=ON]
     -Dcythonize-pynest=[OFF|ON]                  Use Cython to cythonize pynestkernel.pyx.
                                                  If OFF, PyNEST has to be build from
                                                  a pre-cythonized pynestkernel.pyx.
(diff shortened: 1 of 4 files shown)


apeyser:
Maybe an error if python3 == ON but python == OFF ? Or maybe we should just assume that people know what they're doing, since we may want to switch the default from python2 to python3 at some point, and such an error check would then make code complex.

So: :+1:

gtrensch:
@apeyser: I removed the "path/to/python" option from "-Dwith-python". This is not sufficient anyway. It is much better to set all cmake built-in variables required for Python. I have described this in INSTALL. I also removed the "-Dwith-python3" cmake argument again.  Instead of "path/to/python" a specific version can now be set with "-Dwith-python=ON/2/3/OFF". ON is the default. In that case cmake will chose a Python version (usually 2, the lowest installed). The cmake script ensures that all Python components are of exact same version. There is also a parameter check now.

INSTALL
```diff
@@ -115,17 +121,42 @@ Configuring NEST for Distributed Simulation with MPI
...
+=========================
+
+To force a Python3-binding in a mixed Python2/3 environment pass
+
+    -Dwith-python=3
+
+as an argument to `cmake`.
+
+`cmake` usually autodetects your Python installation.
+In some cases `cmake` v2.8.12 might not be able to localize the Python interpreter
```
jougs: Do we need the version number here, i.e. are we sure that the problem only happens with this exact version? I think the sentence is just as clear without the version.

pynest/README.md
```diff
@@ -21,6 +21,14 @@ PyNEST to `$(pyexecdir)`, which is often expanded as follows:
 
     $(prefix)/lib{,64}/pythonX.Y/site-packages/nest
 
+
+To force to usage of a specific Python version pass
```
jougs: "force to" → "force the"

jougs:
:+1: after fixing my minor comments. Many thanks for this contribution!

INSTALL
apeyser: Isn't it cmake < 3.4?

apeyser:
Discussed w/ @gtrensch :+1: 
Just confirm that the install is correct in saying cmake == 2.8.12, or is a cmake < 3.4

gtrensch:
Version number removed to "be on the safe side".

jougs:
This fixes #412.

Issue #343, opened 2016-05-18
Replace alias datatypes by their standard variants


jougs:
We're often using our own typedef'd versions of standard datatypes throughout the code (e.g. `int_t` instead of just `int`; see line 104-109 in [nest_types.h](https://github.com/nest/nest-simulator/blob/dc5edb5808c2526229acbb3989a54acb4efd8d29/nestkernel/nest_types.h#L104))

The use is inconsistently and brings more confusion than readability. These custom types should be removed and their usage replaced by the standard types.

This issue was known as trac90. See also #344.

sanjayankur31:
@jougs - has a decision been made on this? Does someone just need to sit down and remove the aliases in favour of their standard variants now?

sanjayankur31:
The advantages I see of using aliases are:

- readability (`weight` makes more sense than `double_t`)
- aliases are usually shorter (`somealias` instead of `unsigned long long` and so on)

heplesser:
@sanjayankur31 This issue is only about the following aliases for standard C++ datatypes:
```
typedef double double_t;       ///< Double precision floating point numbers.
typedef float float_t;         ///< Single precision floating point numbers.
typedef int int_t;             ///< Integer number with at least 16 bit.
typedef long long_t;           ///< Integer number with at least 32 bit.
typedef unsigned int uint_t;   ///< Unsigned int_t.
typedef unsigned long ulong_t; ///< Unsigned long_t.
```
At some point very long ago we had an idea that you could change `double_t` to be an alias for `float` and thus reduce the memory required for network components. But that would have required fully consistent use of `double_t` throughout, which we do not have at all. We also learned since that memory issues are elsewhere (see eg Kunkel et al, 2014). Therefore, these aliases are mostly confusing.

I will put this on the agenda for next Monday's developer meeting, but unless there should be serious objections raised there, I think we should implement this soon.

heplesser:
In the  [Open NEST Developer VC today](https://github.com/nest/nest-simulator/wiki/2016-08-08-Open-NEST-Developer-Video-Conference) we decided to remove all the `_t` datatypes.

heplesser:
Closing as this has been fixed by#454.

Issue #453, opened 2016-08-08
C++ 11


apeyser:
Following #344, we should consider transitioning towards C++11.

All HPC machines currently support at least C++ 11, except possibly for K, which is in the process of porting their gcc-based compiler to gcc-4.8.2 at some point this year.

STLs are usable, curly brace constructors remove narrowing, constructors and converters are vastly improved, templates are vastly improved...

In fact given that most machines will have llvm compilers going forward, an even more aggressive position is tenable.

jougs:
At the NEST Conference 2017 we officially announced that C++11 is allowed in NEST from now on. I'm thus closing this issue.

PR #454, opened 2016-08-08
Removed _t aliases for built-in C++ data types, see #343.

heplesser:
This PR removes all `_t` "same name" aliases of C++ PODs (`double_t`, `float_t`, `int_t`, `long_t`, `uint_t`, `ulong_t`) from NEST, fixing #343.

I suggest @apeyser and @terhorstd as reviewers.


sanjayankur31:
Found these, mostly in comments from the looks of it:

```bash
[asinha@cs-as14aho-2-herts-ac-uk  nest-simulator(fix-343-remove-pod-alias-types=)]$ grep -Erni " (double_t|float_t|int_t|long_t|uint_t|ulong_t)" *

models/hh_psc_alpha_gap.h:190:   * @param double_t new value of the mebrane potential
models/pp_psc_delta.cpp:329:  // a double_t in ms. The grid based iaf_psp_delta can only handle refractory
models/binary_neuron.h:92:  // must have an double_t operator(double_t) defined
models/mat2_psc_exp.cpp:279:  // a double_t in ms. The grid based mat2_psc_exp can only handle refractory
models/iaf_psc_exp_multisynapse.h:206:    //    double_t PSCInitialValue_;
models/amat2_psc_exp.h:310:    //    double_t PSCInitialValue_;
models/music_message_in_proxy.cpp:161:    double_t acceptable_latency = P_.acceptable_latency_ / 1000.0;
models/iaf_tum_2000.h:278:    //    double_t PSCInitialValue_;
models/iaf_tum_2000.cpp:259:  // a double_t in ms. The grid based iaf_tum_2000 can only handle refractory
models/iaf_psc_alpha.cpp:265:  // a double_t in ms. The grid based iaf_psc_alpha can only handle refractory
models/iaf_neuron.cpp:234:  // a double_t in ms. The grid based iaf_neuron can only handle refractory
models/amat2_psc_exp.cpp:351:  // a double_t in ms. The grid based amat2_psc_exp can only handle refractory
models/mat2_psc_exp.h:282:    //    double_t PSCInitialValue_;
models/iaf_chs_2007.h:233:    //    double_t PSCInitialValue_;
models/music_event_out_proxy.cpp:222:  double_t time = e.get_stamp().get_ms() * 1e-3; // event time in seconds
models/hh_psc_alpha.h:167:   * @param double_t new value of the mebrane potential
models/iaf_psc_exp.cpp:257:  // a double_t in ms. The grid based iaf_psc_exp can only handle refractory
models/iaf_psc_delta.cpp:234:  // a double_t in ms. The grid based iaf_psp_delta can only handle refractory
models/iaf_psc_exp.h:280:    //    double_t PSCInitialValue_;
nestkernel/ring_buffer.h:89:   * @param  double_t Value to add.
nestkernel/ring_buffer.h:96:   * @param  double_t Value to set.
nestkernel/ring_buffer.h:206:   * @param  double_t Value to add.
nestkernel/ring_buffer.h:287:   * @param  double_t Value to append.
nestkernel/connector_model_impl.h:98:// double_t get_default_delay(const GenericConnectorModel<ConnectionT> &cm)
nestkernel/archiving_node.h:75:   * \fn double_t get_Ca_minus()
nestkernel/archiving_node.h:81:   * \fn double_t get_synaptic_elements(Name n)
nestkernel/archiving_node.h:104:   * \fn std::map<Name, double_t> get_synaptic_elements()
nestkernel/archiving_node.h:130:   * \fn double_t get_K_value(long t)
nestkernel/archiving_node.h:144:   * \fn double_t get_triplet_K_value(std::deque<histentry>::iterator &iter)
nestkernel/archiving_node.h:181:   * \fn void set_spiketime(Time const & t_sp, double_t offset)
nestkernel/archiving_node.h:187:   * \fn double_t get_spiketime()
nestkernel/mpi_manager.h:265:   * @note This class actually stores the GID as @c double_t internally.
nestkernel/node.h:590:   * @param t double_t time when the update is being performed
nestkernel/nest_time.h:148:  // double_t: milliseconds (double!)
nestkernel/nest_time.h:231:#define LIM_POS_INF_ms double_t_max // because C++ bites
nestkernel/music_manager.h:59:void set_music_in_port_acceptable_latency( std::string portname, double_t
nestkernel/synaptic_element.h:167:  * \fn double_t get_z_value(Archiving_Node const *a, double_t t) const
nestkernel/music_event_handler.cpp:98:    double_t acceptable_latency = acceptable_latency_ / 1000.0;
nestkernel/nest_types.h:101:const long long_t_max = LONG_MAX;
nestkernel/nest_types.h:102:const long long_t_min = LONG_MIN;
nestkernel/nest_types.h:110:#define double_t_max ( DBL_MAX )
nestkernel/nest_types.h:111:#define double_t_min ( DBL_MIN )
nestkernel/nest_types.h:190:const long delay_max = long_t_max;
nestkernel/nest_types.h:191:const long delay_min = long_t_min;
precise/iaf_psc_exp_ps.h:241:   * @param   double_t length of interval since previous event
precise/iaf_psc_alpha_canon.h:274:   * @param   double_t length of interval since previous event
sli/sli_io.cc:1320:    Token int_token( new IntegerDatum( c ) );
sli/sli_io.cc:1322:    i->OStack.push_move( int_token );
sli/fdstream.h:122:  int_type
sli/fdstream.h:135:  int_type
sli/fdstream.h:136:  overflow( int_type c )
topology/doc/Topology_UserManual.tex:1405:    double_t rx_, ry_;
```

heplesser:
@sanjayankur31 Thanks for noticing those leftovers. They should now be gone.

sanjayankur31:
Re-ran a tweaked grep. Are we not removing `longlong_t`?

```bash
sli/slistartup.cc:557:  typedef long long longlong_t;
sli/slistartup.cc:560:  typedef long longlong_t;
sli/slistartup.cc:566:    longlongsize_name, Token( new IntegerDatum( sizeof( longlong_t ) ) ) );
```

Some limits are still using the `_t` convention - are these to stay?

```
nestkernel/nest_time.h:231:#define LIM_POS_INF_ms double_t_max // because C++ bites
nestkernel/nest_time.h:238:#define LIM_NEG_INF_ms ( -double_t_max ) // c++ bites
nestkernel/simulation_manager.h:158:  bool print_time_; //!< Indicates whether time should be printed during
nestkernel/kernel_manager.h:93: print_time               booltype    - Whether to print progress information during the simulation
nestkernel/nest_types.h:101:const long long_t_max = LONG_MAX;
nestkernel/nest_types.h:102:const long long_t_min = LONG_MIN;
nestkernel/nest_types.h:110:#define double_t_max ( DBL_MAX )
nestkernel/nest_types.h:111:#define double_t_min ( DBL_MIN )
nestkernel/nest_types.h:190:const long delay_max = long_t_max;
nestkernel/nest_types.h:191:const long delay_min = long_t_min;
```

Otherwise LGTM - the test is failing because `fglrx` couldn't be loaded - not really sure what's happening there.

heplesser:
@sanjayankur31 LGTM??? Travis has some technical problems according to the status info on their site that probably causes the failures.

Concerning `longlong_t`, I had overlooked it because it is defined in the code for the SLI interpreter, not the NEST kernel. But it is used solely in creating the longlong have/size entries in the `statusdict/architecture` dictionary, so I will remove it.

@apeyser Concerning the `{double,long}_t_{min, max}` defined in `nest_types.h`: Do we really need those defines? Why don't we use the `{DOUBLE,LONG}_{MIN,MAX}` from the standard headers directly?

jougs:
See also #439 for failing builds due to fglrx not loading.

@heplesser: LGTM = looks good to me. See also http://lgtm.in/

jougs:
@heplesser: I'm not sure if the issue reported on the status page of Travis CI are causing our problems. I've emailed their support for further assistance.

jougs:
Our failing builds actually seem to be a Travis bug. See https://github.com/travis-ci/packer-templates/issues/221 for the issue they opened after my mail.

apeyser:
@heplesser Why do we have those variables? Because steps were long_t, and I had to search for what a long_t was and what it's max should be, so I labelled it. Since it was an alias that could be changed (I assumed, since why else alias), I wanted the label generalized in case that actually happened. (Further comment for historical record, steps were int_t with a complicated system of range checks that were simplified by going to long_t).

But if we get rid of long_t, there's no reason to label the max. The other option was to assume that long_t was a primitive type and use a numeric_limits type, but that wouldn't work for const types before constexpr in C++11. #453 

Thus, we should remove them now and use {DOUBLE,LONG}_{MIN,MAX}.

heplesser:
@apeyser Thanks for the clarification, I will update the PR shortly. Is there a particular reason why we are using `DBL_MAX` etc instead of `std::numeric_limits<double>::max()` and the like? I have a hunch that it has something to do with the "because C++ bites" comments?

apeyser:
@heplesser Yes --- exactly. In C++ < 11, a floating point constant can only be initialized with a floating point literal. Thus, #define for the double_t_max. For integer types, you can defined a constant in terms of another constant under very limited conditions (literal and const arithmetic expressions, basically). So std::numeric_limits<T>::max() can not be used to initialize a const, which means that it's difficult for the compiler to deduce the const-ness of the value throughout the program, which made many traits not-very-useful.

Thus, 15 years on after C++ being first widely used, const weren't const's. But C++11 adds constexpr that come much closer to allowing you to use as a const terms that are composed in ways that are easily deducible to be consts.

apeyser:
@heplesser looks like you've been stepped on by another merge...

heplesser:
@apeyser I have now merged all new commits from master and the conflicts are resolved.

Once we merge this PR, we need to inform all users via the mailing list that they need to pull these changes into their PRs and adapt their code.

apeyser:
:+1: Will merge if/when checks succeed

jougs:
Should we add a new regression test that `grep`s for the unwanted types? This way users are directly made aware of problems that may arise.

heplesser:
@jougs Doesn't the compiler provide that regression test with much better quality (no false positives) than any grep could? A script would only protect us against a developer re-inserting a typedef for double_t & Co. Or am I overlooking something?

Issue #449, opened 2016-08-05
Error when running code from connection management documentation


lepmik:
Running the following code from the [documentation](http://www.nest-simulator.org/connection_management/)

NEST 2.10.0
```
$git describe --tags
v2.10.0-298-g5e5aa48
$cat .git/refs/heads/master
5e5aa480c4c0fc2b54222b1d6af034b6ffd59614
```

```python
nest.ResetKernel()
n = 10
A = nest.Create("iaf_neuron", n)
B = nest.Create("iaf_neuron", n)
nest.CopyModel('static_synapse','excitatory',{'weight':2.5})
syn_dict = {'model': 'excitatory',
            'weight': 2.5,
            'delay': {'distribution': 'uniform', 'low': 0.8, 'high': 2.5},
            'alpha': {'distribution': 'normal_clipped', 'low': 0.5, 'mu': 5.0, 'sigma': 1.0}
           }
nest.Connect(A, B, syn_spec=syn_dict)
```

```
NESTError                                 Traceback (most recent call last)
<ipython-input-66-8534e8f15310> in <module>()
      9             'alpha': {'distribution': 'normal_clipped', 'low': 0.5, 'mu': 5.0, 'sigma': 1.0}
     10            }
---> 11 nest.Connect(A, B, syn_spec=syn_dict)

/home/mikkel/apps/nest/ins/lib/python2.7/site-packages/nest/lib/hl_api_helper.pyc in stack_checker_func(*args, **kwargs)
    136     def stack_checker_func(*args, **kwargs):
    137         if not get_debug():
--> 138             return f(*args, **kwargs)
    139         else:
    140             sr('count')

/home/mikkel/apps/nest/ins/lib/python2.7/site-packages/nest/lib/hl_api_connections.pyc in Connect(pre, post, conn_spec, syn_spec, model)
    476             raise kernel.NESTError("syn_spec needs to be a string or dictionary.")
    477 
--> 478     sr('Connect')
    479 
    480 

/home/mikkel/apps/nest/ins/lib/python2.7/site-packages/nest/__init__.pyc in catching_sli_run(cmd)
     77         commandname = sli_pop()
     78         engine.run('clear')
---> 79         raise _kernel.NESTError("{0} in {1}{2}".format(errorname, commandname, message))
     80 
     81 sli_run = hl_api.sr = catching_sli_run

NESTError: DictError in Connect_g_g_D_D: Unused dictionary items:  alpha
```

heplesser:
[...]

PR #462, opened 2016-08-16
Change Connection management

steffengraber:
Change in documentation about connection management, regarding issue  #449.


heplesser:
Minimal documentation fix, thus merging without further review.

PR #421, opened 2016-07-15
Remove autotools artefacts from git ignore file

jougs:

PR #464, opened 2016-08-18
NEST Help Generator

steffengraber:
Processing more in less time with less code.

The parser go through all sli and cc files to find documentation and 
convert it into .html and .hlp files.
This is a replacement for the old help generation mechanism in NEST.

CMAKE integration is also available.


CMakeLists.txt.orig
```diff
@@ -0,0 +1,378 @@
+# CMakeLists.txt
```
heplesser: @steffengraber Should this `.orig` file be in the repository?

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
...
+convert it into .html .hlp files.
+This is a replacement for the old help generation mechanism in NEST.
+
+CMAKE integration is also available.
+
+## Usage
+
+    python parse_help.py
+
+## In addition the parser solve issue \#363
```
heplesser: "solves"

extras/userdoc/generator/modules/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+def cut_it(trenner, text):
```
heplesser: `trenner` -> `separator` for the benefit of non-German code readers

cmake/GenerateHelp.cmake
```diff
@@ -0,0 +1,76 @@
...
+              -DDATA_DIR='${CMAKE_INSTALL_FULL_DATADIR}'
+              -DHELPDIRS='${HELPDIRS}'
+              -DINSTALL_DIR='${CMAKE_INSTALL_PREFIX}'
+              -P ${PROJECT_SOURCE_DIR}/cmake/generate_help.cmake
+            WORKING_DIRECTORY \"${PROJECT_BINARY_DIR}\"
+          )"
+        )
+        install( CODE
+            "execute_process(
+                COMMAND ${PYTHON} parse_help.py \"${CMAKE_INSTALL_FULL_DOCDIR}/help2\"
```
heplesser: What will happen here if we are on a system that has no Python? What if we build NEST without pynest, but on a system with Python in place?

extras/userdoc/generator/modules/writers.py
```diff
@@ -0,0 +1,420 @@
...
+            Command Index</a></td>
+        <td width="30%" align="center"><a href="../../quickref.html">
+            NEST Quick Reference</a></td>
+      </tr>
+    </tbody></table>
+    '''
+
+    copyright = '''
+    <p style="text-align:center">
+        &copy; 2016 <a href="http://www.nest-initiative.org">
```
heplesser: Since we copyright all code per 2004, I'd suggest to do so for the documentation, too.

heplesser:
@steffengraber Great work! I made some minor comments in the code and have a few more questions/comments:

1. `helpinit.sli` has a huge number of white-space changes, but I haven't noticed any real changes. Are there any? And do we still need the file at all if we move help generation to Python?
1. In general, moving all help generation to Python seems fine to me. I simply cannot see that there would be any system on which people would want to work interactively and that does not have Python. We just need to make sure that `make install` does not choke on python-free systems.
1. All test instances fail on Travis, but the MPI-free cases should pass even with the current problems on Travis. The failing test is `test_round_validate.sli` which is currently our *only* doctest-like test in the testsuite. It only tests a trivial case (the round function) and depends on help being installed correctly. I would suggest to remove the test and consider doctesting when we overhaul the testsuite.

CMakeLists.txt.orig
steffengraber: @heplesser deleted

extras/userdoc/generator/README.md
steffengraber: @heplesser  solved

extras/userdoc/generator/modules/helpers.py
steffengraber: @heplesser done

steffengraber:
@heplesser 

1. There are no real changes in `helpinit.sli`. Besides the help generation, there are some functions like `help` or `page` in the file. I think we should not delete it.
2. Ok.
3. Ok. Is there something I have to do?

heplesser:
@steffengraber Concerning 3: The tests not using mpi (subtests 1 and 3 on travis) fail because of PEP8 violations and because `test_round_validate.sli` fails. As a first step, you can remove the line
```
/- /--> def - 1. 1. eq exch pop assert_or_die
```
from the test, it is historic debris without any use. If the test then still fails on Travis, it is because the `validate` function does not find the help file for the `round` sli function. On systems without Python, the help file will not be available, because Python is required to build help now. The, that test would have to be skipped. But I think we can safely remove it---it only performs a highly trivial test.

.gitignore
```diff
@@ -1,21 +1,38 @@
...
 
 /juqueen/
 /extras/nest-config
 /extras/create_release.sh
+/extras/userdoc/cmds/
 /libnestutil/config.h
+/libnestutil/stamp-h1
 /librandom/randomtest
 /libtool
 /nest/nest
```
jougs: Your change brings back many lines that were removed in #421, probably due to a failed merge. Please remove them again.

CMakeLists.txt
```diff
@@ -108,6 +108,10 @@ find_program( SED NAMES sed gsed )
 ################################################################################
 ##################   CPack, checks, ... in external modules   ##################
 ################################################################################
+include(GenerateHelp)
+include( GetTriple )
+get_host_triple( NEST_HOST_TRIPLE NEST_HOST_ARCH NEST_HOST_VENDOR NEST_HOST_OS )
+get_target_triple( NEST_TARGET_TRIPLE NEST_TARGET_ARCH NEST_TARGET_VENDOR NEST_TARGET_OS )
```
jougs: Why is `GetTriple` included here and why are these function calls here? The include is already in line 125.

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
+# README for the new NEST Help Generator
```
jougs: Remove "new". No one will ever change this if it becomes old one day. Make "Help Generator" all lower case

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Processing more in less time with less code.
```
jougs: This slogan does not explain anything to a normal user, who does not know the horrors of the old SLI parser. Please remove it.

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+The parser go through all sli and cc files to find documentation and 
+convert it into .html and .hlp files.
```
jougs: "convert" → "converts"

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
+# README for the new NEST Help Generator
+
+Processing more in less time with less code.
+
+The parser go through all sli and cc files to find documentation and
```
jougs: .sli and .cc

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
+# README for the new NEST Help Generator
+
+Processing more in less time with less code.
+
+The parser go through all sli and cc files to find documentation and 
+convert it into .html and .hlp files.
+This is a replacement for the old help generation mechanism in NEST.
```
jougs: Remove

extras/userdoc/generator/README.md
```diff
@@ -0,0 +1,24 @@
...
+## In addition the parser solves issue \#363
+
+    'ISSUE 363
+    Currently, the help extractor reads all source files and extracts 
+    the documentation, no matter if the function is enclosed in a #ifdef
+    block or not. This means that the helpdesk may contain documentation
+    for functions or models that have not been compiled, which 
+    frequently confuses users.
+
+    This is a follow-up to trac 713.'
```
jougs: Remove this block. It does not belong into the README.

testsuite/unittests/test_round_validate.sli
```diff
@@ -44,7 +44,4 @@ SeeAlso: floor, cvi
 
 M_ERROR setverbosity
 
-
-/- /--> def - 1. 1. eq exch pop assert_or_die
-
```
jougs: Why?

jougs:
I kind of agree with @heplesser that `helpinit.sli` should go away. At least we should remove the functions for help generation in this PR. We can then move the remaining functions to more sensible places later on.

I'll test the new parser more thoroughly in a week, when I'm back from vacation.

testsuite/unittests/test_round_validate.sli
jougs: Ah! It was @heplesser's idea. Please ignore my comment!

steffengraber:
@jougs Thank you for having a deeper look into the branch. There was an failed merge end of June.

.gitignore
```diff
@@ -1,29 +1,19 @@
...
 *.lo
 *.la
 .*
 *.log
+<<<<<<< HEAD
 *.pyc
 config.status
+=======
+>>>>>>> master
 Makefile
```
jougs: These are more artefacts from an unresolved merge conflict. Please remove them and also the line with `config.status`.

extras/userdoc/generator/modules/writers.py
```diff
@@ -0,0 +1,420 @@
...
+    <h1><a name="top">NEST Command Index</a></h1>
+    <table class="headerfooter">
+      <tr>
+        <td width="50%" align=center>
+        <a href="../index.html">NEST HelpDesk</a></td>
+        <td width="50%" align=center>
+        <a href="../quickref.html">NEST Quick Reference</a></td>
+      </tr>
+    </table>
+    '''
```
jougs: I thing we should use a template engine here to increase code readability and maintainability. [string.Template](https://docs.python.org/2/library/string.html#template-strings) is in the Python standard library and should suffice for this.

I also vote for putting the different blocks in separate files called `nest.css`, `header.html`, `footer.html` and loading them here. This way they can be easily edited separately with an appropriate editor, syntax highlighting and all.

.gitignore
```diff
@@ -9,12 +9,18 @@ pynest/pynestkernel.cpp
 *.la
 .*
 *.log
+<<<<<<< HEAD
+*.pyc
+config.status
+=======
+>>>>>>> master
 Makefile
```
jougs: These are more artefacts from an unresolved merge conflict. Please remove them and also the line with `config.status`.

steffengraber:
The build still fails due to the test_round_validate.sli.
Do you have an idea, how to avoid this issue?

heplesser:
Unless @diesmann strongly objects, I would suggest to remove `test_round_validate.sli`. It performs a close to meaningless test, testing that
```
1.4 round  --> 1.0
1.5 round  --> 2.0
2   round  --> 2.0
```
works as advertised in the documentation. In a second step, we should consider how to implement doctesting properly (see #470 for a follow-up).

extras/userdoc/generator/modules/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+def cut_it(separator, text):
+    """
+    Cut it.
+
+    cut text by trenner
```
jougs: "trenner" → "separator"

extras/userdoc/generator/modules/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    import re
+    if separator:
+        return re.split(separator, text)
```
jougs: Please move the `import` into the `if` block.

jougs: On second thought, I think the `import re` should go to the top of the file as it is needed by all helpers.

steffengraber:
I removed the test in `test_round_validate.sli` only to see what’s happened. 
Travis CI now builds without errors.
But waiting for issue #470 .

heplesser:
To make sure `round` gets tested still, please add the following as `unittests/test_round.sli`, replacing `test_round_validate.sli`. It performs exactly the same test.

```
/*
 *  test_round.sli
 *
 *  This file is part of NEST.
 *
 *  Copyright (C) 2004 The NEST Initiative
 *
 *  NEST is free software: you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation, either version 2 of the License, or
 *  (at your option) any later version.
 *
 *  NEST is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
 *
 *  You should have received a copy of the GNU General Public License
 *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 *
 */


/* BeginDocumentation
Name: testsuite::test_round - check if round function works

Synopsis: (test_round) run -> 

Description:

Tests basic properties of round.
Author:  130531, Diesmann; 161006, Plesser
SeeAlso: floor, cvi
*/

(unittest) run
/unittest using

M_ERROR setverbosity

{ 1.4 round 1.0 eq } assert_or_die
{ 1.5 round 2.0 eq } assert_or_die
{ 2      round 2.0 eq } assert_or_die
```

cmake/GenerateHelp.cmake
jougs: `${PYTHON}` just points to the Python interpreter available on the standard `PATH`. I guess, we can safely assume that *some* Python is available on all modern machines. This is also what we hope for in the testsuite and it never failed on us. Extracting help with a different Python version than the one we compie PyNEST against is also not a problem.

apeyser: I'd add a way to disable -- it's just an extra if, and eventually you'll run into a case where you have to hack it and it's always when you're not in a state of mind to do it properly.

extras/userdoc/generator/modules/writers.py
```diff
@@ -0,0 +1,289 @@
...
+Collect all the data and write help files.
+"""
+
+import os
+import re
+import textwrap
+from helpers import cut_it
+from string import Template
+
+if os.path.isdir('../cmds'):
```
apeyser: Why not just ````if not```` rather than the ````if ...: else: pass```` ?

extras/userdoc/generator/modules/writers.py
```diff
@@ -0,0 +1,289 @@
...
+    footertempl = footerf.read()
+    footerf.close()
+
+    s = Template(templ)
+
+    htmllist = []
+    hlplist = []
+    name = ''
+
+    for key, value in doc_dic.iteritems():
```
apeyser: Why not make these two loops one, and cut out the old comments?

extras/userdoc/generator/modules/writers.py
```diff
@@ -0,0 +1,289 @@
...
+        if key == "File":
+            value = value.strip("###### ###### $$")
+            htmllist.append('<b>Source:</b><pre>%s</pre>' % value)
+            hlplist.append('Source:\n\n%s' % value)
+    htmllist.append('</div>')
+    htmlstring = ('\n'.join(htmllist))
+    cmdindexstring = s.substitute(indexbody=htmlstring, css=csstempl,
+                                  title=name, footer=footertempl)
+
+    if name:  # only, if there is a name
```
apeyser: Aren't these two following one piece of code, but just with a difference in the sli versus cc element of the filename?

jougs:
I've submitted two pull requests against @steffengraber's [branch](https://github.com/steffengraber/nest-simulator/tree/help-build). The first (https://github.com/steffengraber/nest-simulator/pull/2) removes the remaining parts of the old SLI help extraction code, the second (https://github.com/steffengraber/nest-simulator/pull/3) refactors and simplifies the new help generator.

@heplesser, @apeyser: could you please take a look at the two? Thanks!

jougs:
@steffengraber: can you please merge in master? Thanks!

heplesser:
@jougs @steffengraber I will wait with looking at this until @steffengraber has merged master to his branch to reduce the size of the changes.

apeyser:
@jougs, @steffengraber: Added comments on https://github.com/steffengraber/nest-simulator/issues/3

extras/help_generator/writers.py
```diff
@@ -123,7 +105,7 @@ def write_help_html(doc_dic, file, sli_command_list, keywords):
             hlplist.append('%s:\n' % key)
             htmllist.append('<ul>')
             for i in value:
-                see = i.strip("###### ###### $$")
+                # see = i.strip("###### ###### $$")
```
apeyser: Why commented and not just removed?

extras/help_generator/writers.py
```diff
@@ -209,54 +185,92 @@ def write_helpindex(index_dic_list):
...
-                elif (c < 16):
-                    hlp_list.append("%s\t\t\t%s" % (item['name'],
-                                                    item['fullname']))
-                elif(c < 24):
-                    hlp_list.append("%s\t\t%s" % (item['name'],
-                                                  item['fullname']))
-                elif(c < 32):
-                    hlp_list.append("%s\t%s" % (item['name'],
-                                                item['fullname']))
+                c = len(name)
```
apeyser: '\t' * (16- min(c,60)//4)

steffengraber: @apeyser I was looking for something like that. Perfect.

lekshmideepu:
Merge commit fails in Travis build as it is unable to find the log files in the build directory.

```
$ cd $TRAVIS_BUILD_DIR/build
$ tar -zcvf logfiles.tar.gz ./*.log
tar: ./*.log: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```
It is looking for the following files: 
```
$ tar -zcvf logfiles.tar.gz ./*.log
./install-help.log
$ tar -zcvf reports.tar.gz ./reports
./reports/
./reports/installcheck.log
./reports/TEST-core.phase_2.xml
./reports/TEST-core.phase_1.xml
./reports/TEST-core.phase_4.xml
./reports/TEST-core.phase_3.xml
```
solution: Need to edit the .travis.yml file to look for new .html and .hlp files

heplesser:
@lekshmideepu Thank you very much for your detective work! Could you turn this into a new issue with severity "Critical"?

lekshmideepu:
@heplesser  sure.done!

Issue #463, opened 2016-08-18
Parameter arrays cannot be used in connections with rules fixed_indegree and fixed_outdegree in Python


golosio:
Two-dimensional parameter arrays can only be used in conjunction with rule 'all_to_all'.
Apparently, there is no reason to apply this restriction for connection rules 'fixed_indegree' and 'fixed_outdegree', because with such rules there is a fixed number of connections, which can easily be matched to the elements of a two-dimensional array. Having the possibility to use parameter arrays with such rules would make the users able to use user-defined distributions with those two rules without using loops for connecting populations of neurons.

heplesser:
Closing as this was fixed by #465 .

ikitayama:
This test fails on JURON with the following messages:
```
Running test 'unittests/issue-463.sli'...
   > Running  /p/project/cpcp0/pcp0151/opt/nest-clang-normal/bin/nest /p/project/cpcp0/pcp0151/opt/nest-clang-normal/share/doc/nest/unittests/issue-463.sli
   > [INFO] [2019.1.17 8:6:7 /p/project/cpcp0/pcp0151/projects/nest-simulator/nestkernel/rng_manager.cpp:238 @ Network::create_rngs_] : Creating default RNGs
   > [INFO] [2019.1.17 8:6:7 /p/project/cpcp0/pcp0151/projects/nest-simulator/nestkernel/rng_manager.cpp:284 @ Network::create_grng_] : Creating new default global RNG
   > NEST gpus@35514c3 (C) 2004 The NEST Initiative
   >
   > Jan 17 08:06:07 MapThread_a [Error]: RangeCheck
-> 126 (Failed: error in test script)
```

I'd happy to provide more information if needed.

ikitayama:
@heplesser could you re-open this Issue? Thanks.

PR #465, opened 2016-08-18
Proposed fix for issue #463

golosio:
Proposed fix for issue https://github.com/nest/nest-simulator/issues/463
Added possibility to use two-dimensional parameter array in conjunction with connection rules fixed_indegree and fixed_outdegree in Python .
Added Python test for this possibility.


golosio:
I suggest @heplesser and @jougs as reviewers

jougs:
Adding @hannahbos as the original author of the array parameter variant. I'm on vacation currently and can only look properly on 29th of August.

golosio:
Thank you @jougs . @hannahbos and @heplesser actually the array parameter variant seems to work properly in nest, without modifications, in conjunction with the connection rules fixed_indegree and fixed_outdegree as well. It is only a pynest check that forbids this possibility from python. But apparently there is no reason to have this restriction, also because the array parameter variant can be very useful. What I did is simply to remove the restriction, put new checks on array dimensions for the fixed in/out degree cases, add a python test.

pynest/nest/tests/test_connect_array.py
```diff
@@ -0,0 +1,110 @@
...
+        nest.ResetKernel()
+
+        eps = 1.0e-8
+        N = 10
+        K = 3
+        neuron1 = nest.Create('iaf_neuron', N)
+        neuron2 = nest.Create('iaf_neuron', N)
+
+        Warr = [[y*10+x for x in range(K)] for y in range(N)]
+        Darr = [[1.0*y+0.1*x+1.0 for x in range(K)] for y in range(N)]
```
heplesser: You could use integers as well for delays, avoids the need for handling rounding errors.

pynest/nest/tests/test_connect_array.py
```diff
@@ -0,0 +1,110 @@
...
+
+        syn_dict = {'model': 'static_synapse', 'weight': Warr, 'delay': Darr}
+        conn_dict = {'rule': 'fixed_indegree', 'indegree': K}
+
+        nest.Connect(neuron1, neuron2, conn_spec=conn_dict, syn_spec=syn_dict)
+
+        w = [0.0, 0.0, 0.0]
+        d = [0.0, 0.0, 0.0]
```
heplesser: Here you imply that `K==3`. That should be more flexible.

pynest/nest/tests/test_connect_array.py
```diff
@@ -0,0 +1,110 @@
...
+            for j in range(len(conns)):
+                c = conns[j:j+1]
+                w[j] = nest.GetStatus(c, 'weight')[0]
+                d[j] = nest.GetStatus(c, 'delay')[0]
+                val = d[j]-1.0-w[j]/10.0
+                self.assertTrue(val < eps and val > -eps)
+                w1 = w[j]-10.0*i
+                self.assertTrue(w1 == 0.0 or w1 == 1.0 or w1 == 2.0)
+
+            self.assertTrue(w[0] != w[1] and w[0] != w[2] and w[1] != w[2])
```
heplesser: Also here you imply that `K == 3`.

pynest/nest/tests/test_connect_array.py
```diff
@@ -0,0 +1,110 @@
...
+        w = [0.0, 0.0, 0.0]
+        d = [0.0, 0.0, 0.0]
+        for i in range(N):
+            conns = nest.GetConnections(target=neuron2[i:i+1])
+            for j in range(len(conns)):
+                c = conns[j:j+1]
+                w[j] = nest.GetStatus(c, 'weight')[0]
+                d[j] = nest.GetStatus(c, 'delay')[0]
+                val = d[j]-1.0-w[j]/10.0
+                self.assertTrue(val < eps and val > -eps)
```
heplesser: Python permits `-eps < val < eps` or `abs(val) < eps`.

pynest/nest/tests/test_connect_array.py
```diff
@@ -0,0 +1,110 @@
...
+        Darr = [[1.0*y+0.1*x+1.0 for x in range(K)] for y in range(N)]
+
+        syn_dict = {'model': 'static_synapse', 'weight': Warr, 'delay': Darr}
+        conn_dict = {'rule': 'fixed_indegree', 'indegree': K}
+
+        nest.Connect(neuron1, neuron2, conn_spec=conn_dict, syn_spec=syn_dict)
+
+        w = [0.0, 0.0, 0.0]
+        d = [0.0, 0.0, 0.0]
+        for i in range(N):
```
heplesser: I wonder if you need the nested loops. Maybe it would be faster to make a single call to `GetConnections` and `GetStatus` and then sort things out on the Python level, maybe so that you could directly compare with the original arrays?

heplesser:
@golosio We will need to check this carefully. We originally permitted arrays only for one-to-one and all-to-all because we were afraid that the syntax might not be straightforward/intuitive enough and lead to subtle errors (NEST doing something else than the user expects). It appears to me that you logic is that (for fixed in-degree) the outer dimension indexes target neurons and the inner dimension the connections onto each target. That is certainly well-defined. But we should still discuss this in the Developer VC.

There are a number of things to do:
- Add the new feature to the documentation
- Make sure we have an mpitest for this feature. This is problematic, since our MPI testing mechanism currently requires SLI and you are implementing this on the Python level. We need to think about this.
- As a first step, you could extend the Python test to use multiple threads as a first stage of parallelization.

golosio:
@heplesser I'll try to fix the pull request according to all your  concerns. I'm also trying to write a sli version of the test, but I was not able to specify an  array for weights or delays in the dictionary syn_spec used with the Connect function, not even with the all_to_all rule. Also I can't find an example for this in the documentation. Can you give me some hints?

golosio:
@heplesser ok I think I solved. The weight/delay type in the dictionary is not supposed to be a sli array, but a DoubleVector (or IntVector) so I had to use the sli function cv_dv to convert the sli array to a vector. I had to look in the source code to understand this, as I could not find documentation. . .

jougs:
@golosio: Where in the documentation would you have expected this? I can add the corresponding text if you tell me, where! Thanks!

golosio:
@jougs I think that:
1) the commands cv_dv and cv_iv should be mentioned in http://www.nest-simulator.org/quickref/
in the section "Arrays, vectors, and matrices", subsection "conversions"
and in http://www.nest-simulator.org/objects_and_data_types/ in the section "object types", subsection "Conversion between types".
2) the types <doublevectortype> and  <intvectortype> should be mentioned in http://www.nest-simulator.org/objects_and_data_types/ in the section "arrays" (which could become "arrays and vectors") or next to it. It should be shortly explained why they are needed (maybe for compatibility with some python functions?). Maybe the list in the section "object types" could be expanded to include them...
3) There is no information about using synapse and connection dictionaries in http://www.nest-simulator.org/neural_simulations/ Just as a starting point, a very short paragraph analogous to that in pynest: http://www.nest-simulator.org/introduction-to-pynest/part-3-connecting-networks-with-synapses/ section "Connecting with synapse models" with a similar example using the corresponding SLI commands, could be very useful.
4) I would add a section "Connecting with array parameters" with an example similar (but simpler) to the one that I used in test_connect_array.sli
Let me know what you think and if I can help with the changes that you think are appropriate...

golosio:
@jougs  @heplesser if you want I can write a draft of what I would add to the documentation, and then you can review it...

pynest/nest/tests/test_connect_array.py
golosio: Done

golosio:
@heplesser @jougs As @heplesser suggested I made some test changing the number of virtual processes, and indeed there are problems. For instance if I run this sli script (in nest, also without my changes)
[t1.txt](https://github.com/nest/nest-simulator/files/443572/t1.txt)
I get wrong results if the number of virtual processes is greater than 1. Next days I'll try to see how difficult is to fix this problem.

golosio:
ok I got it. The relevant code is in nestkernel/conn_builder.cpp
For instance in the method AllToAllBuilder::connect_() lines 833-...
`
        // check whether the target is on this mpi machine
        if ( not kernel().node_manager.is_local_gid( *tgid ) )
        {
          for ( GIDCollection::const_iterator sgid = sources_->begin();
                sgid != sources_->end();
                ++sgid )
            skip_conn_parameter_( tid );
          continue;
        }
...
        // check whether the target is on our thread
        if ( tid != target_thread )
        {
          for ( GIDCollection::const_iterator sgid = sources_->begin();
                sgid != sources_->end();
                ++sgid )
            skip_conn_parameter_( tid );
          continue;
        }
`
the function skip_conn_parameter_() is used to skip array parameters handled in other virtual processes. Using the same function in a similar way in  FixedInDegreeBuilder::connect_() and in 
FixedOutDegreeBuilder::connect_() solves the issue, and make things work with multiple mpi processes / threads.

heplesser:
@golosio I will try to look at your comments and code later this week. I would suggest that you add any kind of failing test, such as your `t1.txt` to the testsuite.

golosio:
Sure, thank you @heplesser . Take into account that I want to modify test_connect_array.sli into an mpi test. I hope I will finish today or tomorrow.

jougs:
@golosio: Thanks for your ideas about the documentation. Please feel free to change the proposed things in another pull request. All the places you mention are part of the documentation that can be found in `extras/userdoc/md/documentation/`. The homepage is created from those files.

golosio:
@jougs ok, next week I'll try to do it.

golosio:
@heplesser @jougs I wrote two SLI tests for MPI/threads. The test for fixed_outdegree with more than 1 MPI process is more difficult, because connection info are accessible to a MPI process only if the target neuron is associated to that process, thus while in the case of the fixed_indegree rule all the elements of a row of the parameter array can be compared with connection parameters accessible in the same MPI process, for the fixed_outdegree rule this is not the case.
Therefore for this second rule I made a test that works in two levels:
- a first level checks that for each source neuron the connection parameters accessible to the actual MPI process are a subset of the corresponding row of the parameter array.
- a comprehensive test that is not run automatically, which could be run only in case of specific problems or major changes. It requires that the first test writes its results (by uncommenting a line in the SLI script) in a set of files, one for each MPI process. Then a bash script joins the files' content in a proper way to check it.

golosio:
@heplesser ok, I made all the changes that you indicated.
- I made the tests more general, for any values of K
- I prepared a simple failing test as t1.txt
- I wrote two SLI tests for MPI/threads.
- I wrote some documentation in the file hl_api_connections.py
Only I was not able to avoid a nested loop in test_connect_array.py
because (e.g. for fixed_indegree with K=3) I want to check that neuron 1 incoming connections have weights 1,2,3 (no matter in which order) neuron 2 have 4,5,6  and so on. Maybe I could do the same building couples (neuron index, connection weight) and sorting at two levels, but I am not so familiar with python and I am not sure it would be simpler.
I wait for your comments.

hannahbos:
@golosio. Thank you very much for all the nice work. We excluded these options, because we couldn't come up with a usecase. When it comes to biological networks one should be careful with this option since the resulting network could have a network structure that was not intended. For example, given neuron pairs could be connected with different weights or one neuron could be projecting with a positive weight on some and with a negative weight on other neurons. However, the restriction of the parameter array dimension you implemented as well as the documentation should make the usage straight forward. Below I propose a small change in the wording of the new documentation. Mentioning source neurons or target neurons when describing 'fixed_indegree' and 'fixed_outdegree', respectively, is a bit confusion. I tried to point out that the parameter specification of the indegree/outdegree does not depend on the identity of the source/target neuron. In addition to the docstring, the documentation on the homepage should be updated. Once the documentation issues are resolved I am all in for merging.

For 'fixed_indegree' the array has to be a two-dimensional NumPy array
with shape (len(post), indegree), where indegree is the number of
incoming connections per target neuron, therefore the rows describe the
target and the columns the connections converging to the target neuron, 
regardless of the identity of the source neurons.
For 'fixed_outdegree' the array has to be a two-dimensional NumPy array
with shape (len(pre), outdegree), where outdegree is the number of
outgoing connections per source neuron, therefore the rows describe the
source and the columns the connections starting from the source neuron
regardless of the identity of the target neuron.

golosio:
@hannahbos thank you! I will modify the docstring as you suggest and I will update the documentation as soon as possible.

golosio:
@hannahbos ok I changed the docstring as you suggested and I expanded the documentation (file extras/userdoc/md/documentation/connection-management.md) using the same description and including two simple examples.

nestkernel/conn_builder.cpp
```diff
@@ -1101,14 +1101,28 @@ nest::FixedInDegreeBuilder::connect_()
       {
         // check whether the target is on this mpi machine
         if ( not kernel().node_manager.is_local_gid( *tgid ) )
+        {
+          for ( long j = 0; j < indegree_; ++j )
+          {
+            // skip array parameters handled in other virtual processes
+            skip_conn_parameter_( tid );
```
heplesser: Looping over `skip_conn_parameter_()` can become very inefficient due to nested loops and function calls, including to `ConnParameter::skip()`, which requires a vtable lookup. I think it would be far more efficient to pass the number of elements to skip to `skip_conn_parameter()` as second argument (default value 1), pass that through to `ConnParamter::skip()` and then use the most efficient implementation "down there". For `Array*Parameter::skip()`, that probably would just mean replacing `*next_[ tid ]++` with `next_[ tid ] += j` (the derefercing in the current code makes no sense).

pynest/nest/tests/test_connect_array.py
```diff
@@ -0,0 +1,142 @@
...
+            # get row of original weight array, sort it
+            # and compare it with Warr1
+            Warr2 = sorted(Warr[i])
+            for k in range(K):
+                self.assertTrue(Warr1[k]-Warr2[k] == 0.0)
+
+        ############################################
+        # test with connection rule fixed_outdegree
+        ############################################
+        nest.ResetKernel()
```
heplesser: I would split the test into two here. In that way, test failure will be more specific and thus informative.

testsuite/mpitests/test_connect_array_fixed_indegree_mpi.sli
```diff
@@ -0,0 +1,117 @@
...
+
+/N 20 def          % number of neurons in each subnet
+/K 5 def           % number of connections per neuron
+/VP_per_MPI 2 def  % number of virtual processes per MPI process
+
+ResetKernel
+NumProcesses /num_mpi_procs Set  % gets number of MPI processes
+
+0 <<
+     /total_num_virtual_procs num_mpi_procs VP_per_MPI mul
```
heplesser: You can also set `/local_num_threads 2` directly, then NEST will do the math for you. That is the recommended approach.

testsuite/mpitests/test_connect_array_fixed_indegree_mpi.sli
```diff
@@ -0,0 +1,117 @@
...
+
+                /Warr1 0 array def % creates empty weight array
+
+                sdict_lst % loop on synapses that connect to target neuron
+                {
+                        /sdict exch def
+                        /w sdict /weight get def  % gets synaptic weight
+                        /d sdict /delay get def   % gets synaptic delay
+                        /diff d w sub def         % diff = delay - weight 
+                        diff 1.0 eq assert_or_die   % checks that d = w + 1
```
heplesser: You should avoid `assert_or_die` in MPI tests. Rather, set a boolean flag, e.g., `pass` here indicating a problem. This also applies to all following `assert_or_die`. Then, at the very end, have `pass n_loc 0 gt and`

testsuite/mpitests/test_connect_array_fixed_outdegree_mpi.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
heplesser: Since this cannot be run automatically, move it to the `manualtests` directory. With #471 @jougs added a way allowing more complex tests requiring output files. Once that goes "mainstream", we can use it for this test as well.

testsuite/mpitests/test_connect_array_fixed_outdegree_mpi.sli
```diff
@@ -0,0 +1,151 @@
...
+        } if
+
+        /i1 neuron1 1 sub K mul def  % first index = (n1 - 1) * K
+
+        % get row of original weight array
+        /Wrow Warr cva i1 K getinterval def
+        Warr1 {
+                /w exch def
+                % check that all elements of Warr are in Wrow
+                Wrow w MemberQ assert_or_die
```
heplesser: See comment on fixed-indegree test.

testsuite/unittests/issue-463.sli
```diff
@@ -0,0 +1,83 @@
...
+/iaf_neuron 6 Create
+/net1 [1 3] Range def
+/net2 [4 6] Range def
+
+% weight array is made of integers from 1 to 9
+/Warr [1 9] Range cv_dv def
+
+% Define connection and synapse dictionaries
+/syn_dict << /model /static_synapse /weight Warr >> def
+/conn_dict << /rule /fixed_indegree /indegree 3 >> def
```
heplesser: The indegree is equal to hte set of source neurons. Does that weaken the test? Could you use a larger `net1`?

nestkernel/conn_builder.cpp
golosio: Good idea. I've done it. However I also noticed that the same is true for the connection rule all_to_all
in conn_builder.cpp, line 849-853:

          for ( GIDCollection::const_iterator sgid = sources_->begin();
                sgid != sources_->end();
                ++sgid )
            skip_conn_parameter_( tid );

and same thing in other parts of the same file. Since the number of source neurons is fixed in the all_to_all rule, also in this case it should be possible to avoid those loops and to pass the number of source neurons as second argument of the call to skip_conn_parameter

pynest/nest/tests/test_connect_array.py
golosio: done

testsuite/mpitests/test_connect_array_fixed_indegree_mpi.sli
golosio: done

testsuite/mpitests/test_connect_array_fixed_outdegree_mpi.sli
golosio: done

testsuite/unittests/issue-463.sli
golosio: I've increased the number of source neurons(10) target neurons(10) and indegree(5).

golosio:
@heplesser I made all the changes that you requested, except for  setting local_num_threads in the sli tests, for which I wait for your help. I also wait that you tell me whether I should also improve the calls to skip_conn_parameters in the rule 'all_to_all'.

testsuite/mpitests/test_connect_array_fixed_indegree_mpi.sli
heplesser: @golosio Yes, you can :)!

nestkernel/conn_builder.h
```diff
@@ -144,7 +144,7 @@ class ConnBuilder
    * node is not located on the current thread or MPI-process and read of an
    * array.
    */
-  void skip_conn_parameter_( thread );
+  void skip_conn_parameter_( thread, int skip_num = 1 );
```
heplesser: Since the number of places to skip cannot be negative, I would suggest `size_t` as data type. I would also think `num_skip` or `n_skip` would be better names, closer to the full "number of positions to skip".

nestkernel/conn_parameter.h
```diff
@@ -297,10 +297,10 @@ class ArrayIntegerParameter : public ConnParameter
   }
 
   void
-  skip( thread tid ) const
+    skip( thread tid, int skip_num ) const
```
heplesser: Indentation seems off here?

nestkernel/conn_builder.cpp
golosio: ok, I created a new issue.

golosio:
@heplesser ok, I made the changes that you requested.

golosio:
Thank you @heplesser !

hannahbos:
:+1:  Thank you again for all the good work!

golosio:
Thank you @hannahbos !

PR #466, opened 2016-08-18
Improve numerical stability of some tests.

heplesser:
Revised tsodyks-test to use explicitly defined input spike trains instead of trains generated by a DC-driven neuron. This avoids numerical instabilities.


heplesser:
I suggest @jougs and @terhorstd as reviewers.

DimitriPlotnikov:
Improves readability of the test and makes it more numerically stable.

PR #467, opened 2016-08-22
Structural Plasticity steering

sdiazpier:
This pull request enables the dynamic steering of structural plasticity parameters. This can be used to progressively explore the connectivity parameter space in a network and allows the stable creation of connections. It is specially useful in networks with multiple populations, where each must reach different target activity or when a base fixed connectivity is provided and strong activity dependencies between populations are present.

### DIFF
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -303,10 +303,10 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
       }
     }
   }
-    if ( not d->known( names::synaptic_elements ) )
-    {
-      return;
-    }
+  if ( not d->known( names::synaptic_elements ) )
+  {
+    return;
+  }
   // we replace the existing synaptic_elements_map_ by the new one
   DictionaryDatum synaptic_elements_d;
   std::pair< std::map< Name, SynapticElement >::iterator, bool > insert_result;
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -285,10 +285,21 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
     clear_history();
   }
 
-  if ( not d->known( names::synaptic_elements ) )
+  
+  if ( d->known( names::update_synaptic_elements ) )
   {
-    return;
+    const DictionaryDatum synaptic_elements_a = getValue< DictionaryDatum >( d, names::update_synaptic_elements );
+    for ( std::map< Name, SynapticElement >::iterator it = synaptic_elements_map_.begin();
+        it != synaptic_elements_map_.end();
+        ++it )
+    {
+          it->second.set(synaptic_elements_a);
+    }
   }
+    if ( not d->known( names::synaptic_elements ) )
+    {
+      return;
+    }
   // we replace the existing synaptic_elements_map_ by the new one
   DictionaryDatum synaptic_elements_d;
   std::pair< std::map< Name, SynapticElement >::iterator, bool > insert_result;
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -285,7 +285,6 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
     clear_history();
   }
 
-
   if ( d->known( names::synaptic_elements_param ) )
   {
     const DictionaryDatum synaptic_elements_dict =
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -288,12 +288,20 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
   
   if ( d->known( names::update_synaptic_elements ) )
   {
-    const DictionaryDatum synaptic_elements_a = getValue< DictionaryDatum >( d, names::update_synaptic_elements );
-    for ( std::map< Name, SynapticElement >::iterator it = synaptic_elements_map_.begin();
-        it != synaptic_elements_map_.end();
-        ++it )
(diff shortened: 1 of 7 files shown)


sdiazpier:
I suggest @heplesser, @apeyser, @suku248  or @jougs as reviewers.

nestkernel/synaptic_element.cpp
```diff
@@ -120,7 +120,10 @@ nest::SynapticElement::set( const DictionaryDatum& d )
   double new_tau_vacant = tau_vacant_;
 
   // Store values
-  updateValue< double >( d, names::growth_rate, growth_rate_ );
+  if ( d->known( names::growth_rate ) )
```
apeyser: So, if the growth rate is in the dictionary, set the value, but if not, propagate the current value up. Is this right, that a set operation can instead change the dictionary passed? Maybe the set should do nothing if not known, and a get is needed to access the value?

nestkernel/archiving_node.cpp
```diff
@@ -275,6 +275,19 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
     clear_history();
   }
 
+
+  if ( d->known( names::update_synaptic_elements ) )
```
apeyser: Is update_synaptic_elements a good name? It represents the value to update the element with, and not a boolean. Hmm, I'm not inspired as to what would be a better short word for update which doesn't sound like a true/false but is the value to update with...

Silmathoron: Why not ``"param_synaptic_elements"``?

nestkernel/synaptic_element.cpp
Silmathoron: I agree with @apeyser than this looks weird.
Also, this might be a stupid question since I'm not completely familiar with how NEST deals with structural plasticity, yet, but is it clear that all synaptic elements share the same ``growth_rate`` value?
If not, wouldn't the current ``set`` make all of them equal in [``Archiving_Node::set_status``](https://github.com/sdiazpier/nest-simulator/blob/325ec678099cb251735793b28e14499b710ca216/nestkernel/archiving_node.cpp#L288)?

nestkernel/synaptic_element.cpp
```diff
@@ -120,7 +120,10 @@ nest::SynapticElement::set( const DictionaryDatum& d )
...
+    growth_rate_ = getValue< double >( d, names::growth_rate );
```
Silmathoron: NEST guidelines ask for braces even around single line blocks (though I personally find it ugly, apparently it is for safety reasons), so I guess this ``if/else`` should span 4 more lines ;)

nestkernel/archiving_node.cpp
```diff
@@ -275,6 +275,19 @@ nest::Archiving_Node::set_status( const DictionaryDatum& d )
...
+  {
+    const DictionaryDatum synaptic_elements_a =
+      getValue< DictionaryDatum >( d, names::update_synaptic_elements );
+    for ( std::map< Name, SynapticElement >::iterator it =
+            synaptic_elements_map_.begin();
+          it != synaptic_elements_map_.end();
+          ++it )
+    {
+      it->second.set( synaptic_elements_a );
```
sanjayankur31: Just a quick comment to inform the reviewers of an issue here (@sdiazpier is already aware of this):

The current implementation does not enable the user to set different parameter values for different synaptic elements. Usually, though, as in the Butz paper that introduces MSP, neurons will at least have two types of synaptic elements - the axonal and the dendtritic, and these will both have growth curves described by different sets of parameters. So, one must be able to set their parameters separately here too. The fix is quite simple - enable the user to mention which synaptic element is to be updated.

nestkernel/archiving_node.cpp
sanjayankur31: "synaptic_elements_growth_param" maybe? Since whatever parameters are updated here affect the growth dynamics of various synaptic elements? It's a bit wordy, though.

sdiazpier:
Dear @apeyser , @Silmathoron, @heplesser,
@sanjayankur31 and I have worked on the PR and solved the issues you have pointed out. Could you please let us know if there are any additional changes we should implement? Thanks a lot for reviewing our work.

Silmathoron:
I just took a look out of curiosity, I don't know enough about structural plasticity to judge this, I'm fine with whatever HEP and Alex decide ^^

nestkernel/synaptic_element.cpp
```diff
@@ -121,10 +121,10 @@ nest::SynapticElement::set( const DictionaryDatum& d )
 
   // Store values
   if ( d->known( names::growth_rate ) )
-    growth_rate_ = getValue< double_t >( d, names::growth_rate );
-  else
-    updateValue< double_t >( d, names::growth_rate, growth_rate_ );
-  updateValue< double_t >( d, names::tau_vacant, new_tau_vacant );
+  {
+    growth_rate_ = getValue< double >( d, names::growth_rate );
+  }
```
heplesser: This code seems too complicated, the original code was correct:
```c++
updateValue< double >( d, names::growth_rate, growth_rate_ );
```
This will change the value of `growth_rate_` if `d` contains a `growth_rate` entry and leave it unchanged otherwise. `updateValue()` does not change the dictionary `d`, but it returns `true` if `growth_rate_` has been changed, `false` otherwise. This is useful in some cases where checks or changes are required if a member has been modified.

sdiazpier:
Dear @heplesser I have reverted to the initial code. Please let me know if there is anything else to be fixed.

heplesser:
Release notes: Structural plasticity parameters can no be changed with SetStatus/SetKernelStatus.

Issue #81, opened 2015-08-14
Add a status flag indicating whether a model support precise events


heplesser:
For all I can see, it is currently not possible to detect whether a neuron model supports precise spike times, short of reading the documentation. This is unfortunate for the user, and makes automated testing problematic if that requires adaptation for models supporting precise times.

Add a flag to the status dictionary, returning the value provided by `Node::is_off_grid()`.

heplesser:
Closing as issue is fixed in current HEAD.

PR #469, opened 2016-08-25
'is_precise_neuron' dictionary entry

DimitriPlotnikov:
Adds a new dictionary entry that is supported by all ancestors of the ArchivingNode indicating whether the model is a precise neuron. By default, it is set to false. In precise neurons is redefined in the 'get_status' call. Proposal to solve the #81 issue.

### DIFF
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -216,7 +216,7 @@ nest::Archiving_Node::get_status( DictionaryDatum& d ) const
   def< double >( d, names::tau_minus_triplet, tau_minus_triplet_ );
   // By default, the neuron does't support the precise integration scheme.
   // This property can be only set by the neuron implementation and not from outside.
-  def< bool >(d, names::is_precise_neuron, false);
+  def< bool >(d, names::is_precise_neuron, this->is_off_grid());
 #ifdef DEBUG_ARCHIVER
   def< int >( d, names::archiver_length, history_.size() );
 #endif
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -214,9 +214,7 @@ nest::Archiving_Node::get_status( DictionaryDatum& d ) const
   def< double >( d, names::tau_Ca, tau_Ca_ );
   def< double >( d, names::beta_Ca, beta_Ca_ );
   def< double >( d, names::tau_minus_triplet, tau_minus_triplet_ );
-  // By default, the neuron does't support the precise integration scheme.
-  // This property can be only set by the neuron implementation and not from outside.
-  def< bool >(d, names::is_precise_neuron, this->is_off_grid());
+
 #ifdef DEBUG_ARCHIVER
   def< int >( d, names::archiver_length, history_.size() );
 #endif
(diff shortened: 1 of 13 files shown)


nestkernel/archiving_node.cpp
```diff
@@ -214,6 +214,9 @@ nest::Archiving_Node::get_status( DictionaryDatum& d ) const
   def< double >( d, names::tau_Ca, tau_Ca_ );
   def< double >( d, names::beta_Ca, beta_Ca_ );
   def< double >( d, names::tau_minus_triplet, tau_minus_triplet_ );
+  // By default, the neuron does't support the precise integration scheme.
+  // This property can be only set by the neuron implementation and not from outside.
+  def< bool >(d, names::is_precise_neuron, this->is_off_grid());
```
heplesser: I would move this all the way up to the `Node` base class. There is no principle reason why one would not derive a precise spiking subclass directly from `Node`.

Also, we don't need `this->` here.

nestkernel/nest_names.cpp
```diff
@@ -177,6 +177,7 @@ const Name input_currents_in( "input_currents_in" );
 const Name Interpol_Order( "Interpol_Order" );
 const Name interval( "interval" );
 const Name is_refractory( "is_refractory" );
+const Name is_precise_neuron( "is_precise_neuron" );
```
heplesser: I am not sure I like the name. I would prefer `supports_precise_spikes`.

precise/parrot_neuron_ps.cpp
```diff
@@ -101,6 +101,7 @@ void
 parrot_neuron_ps::get_status( DictionaryDatum& d ) const
 {
   Archiving_Node::get_status( d );
+  def< bool >(d, names::is_precise_neuron, true);
```
heplesser: There should be no need to insert this manually. `parrot_neuron_ps::is_off_grid()` returns `true`, so the default mechanism should work.

precise/poisson_generator_ps.h
```diff
@@ -224,6 +224,7 @@ poisson_generator_ps::get_status( DictionaryDatum& d ) const
 {
   P_.get( d );
   device_.get_status( d );
+  def< bool >(d, names::is_precise_neuron, false);
```
heplesser: Why do set this to `false`? The generator supports precise spikes.

testsuite/unittests/test_precise_neuron_flag.sli
```diff
@@ -0,0 +1,33 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+ 
+(unittest) run
+/unittest using
+
+/hh_psc_alpha Create /grid_neuron Set
+grid_neuron GetStatus /is_precise_neuron get false eq assert_or_die
```
heplesser: We generally write code SLI tests as blocks, since we then can see in the output which block failed. Since support for precise spikes is a model property, we can check it on the model defaults (also dropping comparison with true/false):
```
{ /hh_psc_alpha GetDefaults /supports_precise_spikes get not } assert_or_die
{ /parrot_neuron_ps GetDefaults /supports_precise_spikes get } assert_or_die
```

heplesser:
@DimitriPlotnikov Thanks! The test you added fails---does that also happen on your machine or only on Travis? See also the comments in the code.

DimitriPlotnikov:
@heplesser Thanks for the comments! Regarding the failing test: it seems to be a formating check (Formatting:          Failed.) I'll try to run clang-format with NEST properties after adopting your proposals.

DimitriPlotnikov:
The pull request accidentally changed unrelated files. I'll submit a new one.

heplesser:
@DimitriPlotnikov Did you ever create the new PR you posted in your last comment on this PR?

heplesser:
Closed #81 now as I see that this has been fixed.

PR #474, opened 2016-09-08
V-bounded implementation for AEIF + fixed Delta_T=0

Silmathoron:
This PR proposes a corrected implementation of the existing AEIF models that leads to a solution which is closer to the reference LSODAR implementation (see discussion in PR #229 and provided notebook).
It also includes a fix for issue [#114](https://github.com/nest/nest-simulator/issues/114) by setting a separate implementation for the case of ``Delta_T == 0``.

**Remaining question:** currently the argument of the exponential is not bounded anymore (except for the ``multisynapse`` implementation because the numerical instability test is different), which means that numerical overflow _will_ occur when ``Delta_T`` becomes small enough.
Should we consider this an unavoidable issue of the model and consider that the domain where it occurs are not biologically relevant anyway? Or should the exponential still be bounded?
If it should be bounded, I propose to use ``numeric_limits`` and ``MAX_DOUBLE`` instead of an ad hoc value, but it should be noted that in those cases, the model _will_ deviate from the real solution anyway...


golosio:
@Silmathoron  and @heplesser I finally find some time to read your messages and discussion in https://github.com/nest/nest-simulator/pull/229 It's a very interesting and valuable work!
Concerning this pull request, the first thing that I have to say is that there are several issues with the aeif_*_multisynapse models that are not related to your changes, see https://github.com/nest/nest-simulator/projects/1 , and discussion in https://github.com/nest/nest-simulator/pull/439 . I dealt with those issues, with the help of @heplesser and @DimitriPlotnikov, in the latter pull request for the aeif_cond_beta_multisynapse model. My suggestion is that you send a separate pull request for the single inh/exc synapse models, so that your pull request can be merged quickly, and meanwhile we can discuss the issues for multisynapse models.

Silmathoron:
Hi @golosio, thanks for your reply.
Just to be sure I understand: you suggest that I modify only the ``aeif_cond_alpha``, ``aeif_cond_exp``, ``aeif_cond_alpha_RK5``, keeping the ``aeif_cond_alpha_multisynapse`` unchanged, is that it?
I understand you point, but since this concerns the numerical implementation of the ODE and not the model design implementation, this does not really interfere with your PR, right? I mean we are modifying different parts of the code, so if I include the modification in ``aeif_cond_alpha_multisynapse``, it will save someone (probably you, in fact ^^) the trouble of modifying another part of the code in your PR, don't you think?

golosio:
@Silmathoron What I mean is that aeif_*_*_multisynapse need an extensive review, which have already been done in aeif_cond_beta_multisynapse. I am not sure that your PR can be merged without this review, even though your changes are related to other parts of the code. In my opinion the easiest way to proceed is that you present a first PR only for  aeif_cond_alpha, aeif_cond_exp, aeif_cond_alpha_RK5, which could be merged quickly. Then if you are available you could apply your changes to aeif_cond_beta_multisynapse (which was just accepted for merging) with a new PR (if you can't I can try to do it) and from this it would be straightforward to rewrite aeif_cond_alpha_multisynapse.

Silmathoron:
Ok, sounds good to me, the modification is really easy to do so it should not take long to modify your model afterwards.

golosio:
@Silmathoron In aeif_cond_alpha_RK5.cpp line 438

        ( this->*( V_.model_dynamics ) )( S_.yin, S_.k6 );
        // 5th order
        for ( int i = 0; i < S_.STATE_VEC_SIZE; ++i )
          S_.ynew[ i ] = S_.y_[ i ]
            + h * ( 35.0 / 384.0 * S_.k1[ i ] + 500.0 / 1113.0 * S_.k3[ i ]
                    + 125.0 / 192.0 * S_.k4[ i ]
                    - 2187.0 / 6784.0 * S_.k5[ i ]
                    + 11.0 / 84.0 * S_.k6[ i ] );
( this->*( V_.model_dynamics ) )( S_.yin, S_.k7 );

I've worked on RK5 in other projects and I think there is a mistake in the last line. The first argument should be  S_.ynew, not S_.yin Indeed, with this code k7 will always be the same as k6, which does not make much sense, because yIn is unchanged (see the code above). As k7 is only necessary for evaluating the error, it probably does not have dramatic effects on the solution. I've already noticed this problem in aeif_cond_alpha_multisynapse. I know it has nothing to do with your changes, however since we are here it is better to clarify this issue.

Silmathoron:
Ok, I'll have a look, thanks.
In the meantime, if you want to comment on code, please use the GitHub tool:
* browse the file
* find the line
* click on the ``+`` that appears on the right, next to the line number
* write your comment

golosio:
Concerning your "remaining question": if you bind V to V_peak, for  Delta_T!=0 the exponential argument maximum value should be (V_peak - V_th ) / Delta_T , am I right? Maybe this value could be checked in the Parameters_::set and give an error or warning, e.g. suggesting either to use Delta_T=0 or to use larger values?

golosio:
@Silmathoron thank you for the explanation on how to use the GitHub tool, I'll use it in the future.
However, in this case I've already done checking all your changes and apart from the previous comments everythig looks good to me.

golosio:
@helplesser I don't know if I'm allowed to act as a reviewer for this PR, however I've read all your previous discussion in https://github.com/nest/nest-simulator/pull/229 and I revised the code. After clarification of the few points that I raised above, I would suggest to merge this PR.

abigailm:
@golosio - you most certainly are welcome to act as a reviewer on this PR or, indeed, on any other!

golosio:
Thank you @abigailm

models/aeif_cond_alpha_RK5.cpp
```diff
@@ -427,7 +435,7 @@ void nest::aeif_cond_alpha_RK5::update( Time const& origin,
                     + 125.0 / 192.0 * S_.k4[ i ]
                     - 2187.0 / 6784.0 * S_.k5[ i ]
                     + 11.0 / 84.0 * S_.k6[ i ] );
-        aeif_cond_alpha_RK5_dynamics( S_.yin, S_.k7 );
+        ( this->*( V_.model_dynamics ) )( S_.yin, S_.k7 );
```
heplesser: I agree with @golosio that it should be `S_.ynew` instead of `S_.yin` here.

doc/model_details/aeif_models_implementation.ipynb
```diff
@@ -0,0 +1,711 @@
+{
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -219,45 +219,33 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
-    {
       throw BadProperty( "MAXERR must be positive." );
-    }
     MAXERR = tmp;
   }
 
   if ( updateValue< double >( d, names::HMIN, tmp ) )
   {
     if ( not( tmp > 0.0 ) )
-    {
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -219,45 +219,33 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
       throw BadProperty( "HMIN must be positive." );
-    }
     HMIN = tmp;
   }
 
-  if ( V_peak_ <= V_th )
-  {
+  if ( Delta_T != 0. && V_peak_ <= V_th )
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -219,45 +219,33 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
     throw BadProperty( "V_peak must be larger than threshold." );
-  }
+  else if ( Delta_T == 0. )
+    updateValue< double >( d, names::V_peak, V_th ); // expected behaviour
 
   if ( V_reset_ >= V_peak_ )
-  {
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -219,45 +219,33 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
     throw BadProperty( "Ensure that: V_reset < V_peak ." );
-  }
 
   if ( C_m <= 0 )
-  {
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -219,45 +219,33 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
     throw BadProperty( "Capacitance must be strictly positive." );
-  }
 
   if ( t_ref_ < 0 )
-  {
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -219,45 +219,33 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
     throw BadProperty( "Refractory time cannot be negative." );
-  }
 
   if ( tau_w <= 0 )
-  {
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -426,8 +414,17 @@ aeif_cond_alpha_multisynapse::calibrate()
     V_.g0_ex_[ i ] = 1.0 * numerics::e / P_.taus_syn[ i ];
     V_.g0_in_[ i ] = 1.0 * numerics::e / P_.taus_syn[ i ];
   }
-  V_.RefractoryCounts_ = Time( Time::ms( P_.t_ref_ ) ).get_steps();
-  assert( V_.RefractoryCounts_
+
+  // set the right function for the dynamics
+  if ( P_.Delta_T != 0. )
```
Silmathoron: Done

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -524,25 +521,25 @@ aeif_cond_alpha_multisynapse::update( Time const& origin,
         t_return = t + h; // update t
 
         // k1 = f(told, y)
-        aeif_cond_alpha_multisynapse_dynamics( S_.y_, S_.k1 );
+        ( this->*( V_.model_dynamics ) )( S_.y_, S_.k1 );
```
Silmathoron: I think so, see [SO post](http://stackoverflow.com/a/5499169/5962321)

models/aeif_cond_alpha_RK5.cpp
Silmathoron: Ok, I changed it.

Silmathoron:
Ok, I incorporated the changes and started to use the pep8 pip package, which finally allowed me to get the better of those awful pep8 errors!

Regarding the code, there are still remarks and questions:
* I chose to raise an error rather than switch the ``Delta_T`` behaviour because I think it's better to yell at the user that something is wrong
* I stored the LSODAR data in a .dat file; is there any easy way to access it when using ``nest.tests``?
* the RK5-based implementation do not pass the test

models/aeif_cond_alpha.cpp
```diff
@@ -280,6 +280,17 @@ nest::aeif_cond_alpha::Parameters_::set( const DictionaryDatum& d )
...
+    // spike time
+    double max_exp_arg = std::log( std::numeric_limits< double >::max() );
+    if ( ( V_peak_ - V_th ) / Delta_T >= max_exp_arg )
+      throw BadProperty(
+        "The current combination of V_peak, V_th and Delta_T \
+will lead to numerical overflow at spike time; try for instance to increase \
+Delta_T or to reduce V_peak to avoid this problem." );
+  }
   else if ( Delta_T == 0. )
     updateValue< double >( d, names::V_peak, V_th ); // expected behaviour
```
Silmathoron: Because the spike occurs at ``V_th`` for ``Delta_T = 0.``, just like in a regular iaf neuron. I'll add a longer comment to explain it in the code.

pynest/nest/tests/test_aeif_lsodar.py
```diff
@@ -0,0 +1,246 @@
...
+
+The new implementation binds V_m to be smaller than 0.
+
+Rationale of the test: all models should be close to the reference LSODAR when
+submitted to the same excitatory current.
+
+Details:
+  The models are compared and we assess that the difference is smaller than a
+  given tolerance.
+"""
```
Silmathoron: Yes, I'm currently doing it; however I don't know which path I need to provide to load the .dat file on Travis: does NEST provide a way to get the path to the source directory?

pynest/nest/tests/test_aeif_lsodar.py
```diff
@@ -0,0 +1,246 @@
...
+        overflow = (idx == len(array))
+        idx[overflow] -= 1
+        # for the others, find the nearest
+        tmp = idx[~overflow]
+        idx[~overflow] = tmp - (np.abs(values[~overflow] - array[tmp - 1]) <
+                                np.abs(values[~overflow] - array[tmp]))
+        return idx
+
+
+def _interpolate_lsodar(nest_times, lsodar):
```
Silmathoron: Because when the problem occurred I did not think about it and I did it on the fly, which is pretty stupid... I'll have a look at them.

models/aeif_cond_alpha_RK5.cpp
```diff
@@ -179,35 +185,49 @@ nest::aeif_cond_alpha_RK5::Parameters_::set( const DictionaryDatum& d )
...
 
   if ( t_ref_ < 0 )
+  {
     throw BadProperty( "Refractory time cannot be negative." );
+  }
 
   if ( tau_syn_ex <= 0 || tau_syn_in <= 0 || tau_w <= 0 )
+  {
     throw BadProperty( "All time constants must be strictly positive." );
+  }
```
Silmathoron: You're talking about the tests for ``Delta_T``, ``V_peak`` and ``V_th``, right?

models/aeif_cond_alpha.cpp
```diff
@@ -280,6 +280,17 @@ nest::aeif_cond_alpha::Parameters_::set( const DictionaryDatum& d )
 
   if ( Delta_T != 0. && V_peak_ <= V_th )
     throw BadProperty( "V_peak must be larger than threshold." );
+  else if ( Delta_T != 0. )
```
Silmathoron: This is much better, indeed, I'll change it!
No idea about ``V_peak_``; at first I thought it was because users are not supposed to access the variables directly but since only a few of them are like that, it doesn't make sense... do you want me to remove it? What about ``t_ref_`` 'n co. ?

pynest/nest/tests/test_aeif_lsodar.py
heplesser: @lekshmideepu Could you help @Silmathoron with this?

pynest/nest/tests/test_aeif_lsodar.py
heplesser: That would be great. Using standard tools usually makes code more maintainable.

models/aeif_cond_alpha_RK5.cpp
heplesser: @Silmathoron Now I managed to confuse myself. Was this about adding braces? Anyways, the code looks fine here.

models/aeif_cond_alpha.cpp
heplesser: Yes, but the coding here is unfortunate. On lines 258-259, we extract `V_peak` and `V_th` using `updateValue()`. If we now need to set `V_peak` to `V_th`, that should be done by direct assignment, not by calling `updateValue()` again. And strictly speaking, we should not change `V_th` (since that is a user-defined parameter). Rather, we should introduce an additional variable in `V_` that represents the `V_th` to use, and that is set in `calibrate()`:
```C++
if ( P_.Delta_T > 0 )
{
  V_.V_th = P_.V_th;
}
else
{
  V_.V_th = P_.V_peak;
}
```
(I hope I got the logic the right way around?)

Then the user could even switch back and forth between Delta_T=0 and > 0 and all would work as expected.

Silmathoron: This looks very handy, I'll implement it! (but it's ``V_peak`` that changes ;) )

pynest/nest/tests/test_aeif_lsodar.py
Silmathoron: Yes, it also takes only 10 s instead of 10 min and 2 lines of codes instead of 50...

Silmathoron:
Ok, all modifications have been done and I added a test to compare with ``iaf_cond_*`` when ``Delta_T = 0.``, ``a = 0.``, ``b = 0.`` and ``w = 0.``.
I managed to get the RK5 models to pass the test but I had to increase their specific tolerance and I need to keep the exponential bounded or it leads to either numeric instability or parasitic jumps in ``w`` at spike times because of the divergence of ``V``.

Now I just need to find out how to load the ``.dat`` from ``nest.tests`` before I push.

pynest/nest/tests/test_aeif_lsodar.py
Silmathoron: It's okay, I found the way by modifying the ``setup.py``

models/aeif_cond_alpha.cpp
```diff
@@ -280,6 +280,17 @@ nest::aeif_cond_alpha::Parameters_::set( const DictionaryDatum& d )
...
+  {
+    // check for possible numerical overflow with the exponential divergence at
+    // spike time
+    double max_exp_arg = std::log( std::numeric_limits< double >::max() );
+    if ( ( V_peak_ - V_th ) / Delta_T >= max_exp_arg )
+      throw BadProperty(
+        "The current combination of V_peak, V_th and Delta_T \
+will lead to numerical overflow at spike time; try for instance to increase \
+Delta_T or to reduce V_peak to avoid this problem." );
```
Silmathoron: corrected, thx

models/aeif_cond_alpha.cpp
```diff
@@ -280,6 +280,17 @@ nest::aeif_cond_alpha::Parameters_::set( const DictionaryDatum& d )
 
   if ( Delta_T != 0. && V_peak_ <= V_th )
     throw BadProperty( "V_peak must be larger than threshold." );
+  else if ( Delta_T != 0. )
+  {
+    // check for possible numerical overflow with the exponential divergence at
+    // spike time
+    double max_exp_arg = std::log( std::numeric_limits< double >::max() );
```
Silmathoron: done

models/aeif_cond_alpha.cpp
```diff
@@ -278,38 +278,50 @@ nest::aeif_cond_alpha::Parameters_::set( const DictionaryDatum& d )
...
+    throw BadProperty( "Delta_T must be positive." );
+  }
+  else if ( Delta_T > 0. )
   {
     // check for possible numerical overflow with the exponential divergence at
-    // spike time
-    double max_exp_arg = std::log( std::numeric_limits< double >::max() );
+    // spike time, keep a 1e20 margin for the subsequent calculations
+    const double max_exp_arg =
+      std::log( std::numeric_limits< double >::max() ) - 20.;
```
heplesser: @Silmathoron Here you mix natural and decadic log. I would suggest 
```C++
std::log( std::numeric_limits< double >::max() / 1e20 )
```
for clarity and correctness.

models/aeif_cond_alpha_RK5.cpp
```diff
@@ -200,18 +200,29 @@ nest::aeif_cond_alpha_RK5::Parameters_::set( const DictionaryDatum& d )
...
   }
-
-  if ( V_reset_ >= V_peak_ )
+  else if ( Delta_T > 0. )
   {
-    throw BadProperty( "Ensure that: V_reset < V_peak ." );
+    // check for possible numerical overflow with the exponential divergence at
+    // spike time, keep a 1e20 margin for the subsequent calculations
+    const double max_exp_arg =
+      std::log( std::numeric_limits< double >::max() ) - 20.;
```
heplesser: See comment on `aeif_cond_alpha`.

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -234,18 +234,29 @@ aeif_cond_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
   }
-
-  if ( V_reset_ >= V_peak_ )
+  else if ( Delta_T > 0. )
   {
-    throw BadProperty( "Ensure that: V_reset < V_peak ." );
+    // check for possible numerical overflow with the exponential divergence at
+    // spike time, keep a 1e20 margin for the subsequent calculations
+    const double max_exp_arg =
+      std::log( std::numeric_limits< double >::max() ) - 20.;
```
heplesser: See comment on `aeif_cond_alpha`.

models/aeif_cond_exp.cpp
```diff
@@ -269,42 +269,55 @@ nest::aeif_cond_exp::Parameters_::set( const DictionaryDatum& d )
...
+    throw BadProperty( "Delta_T must be positive." );
+  }
+  else if ( Delta_T > 0. )
   {
     // check for possible numerical overflow with the exponential divergence at
-    // spike time
-    double max_exp_arg = std::log( std::numeric_limits< double >::max() );
+    // spike time, keep a 1e20 margin for the subsequent calculations
+    const double max_exp_arg =
+      std::log( std::numeric_limits< double >::max() ) - 20.;
```
heplesser: See comment on `aeif_cond_alpha`.

Silmathoron: Fixed, sorry about that (propagate upwards)

golosio:
@Silmathoron Congratulations for all the good work that you've done!

Silmathoron:
Thanks at @heplesser and @golosio for the impressive review and constructive ideas ;)

PR #475, opened 2016-09-09
Introduce new flag 'supports_precise_spikes' 

DimitriPlotnikov:
Introduce new flag 'support_precise_spikes' for nodes which support off_grid events


nestkernel/nest_names.h
```diff
@@ -325,11 +325,13 @@ extern const Name structural_plasticity_update_interval; //!< Update interval
                                                          //!< plasticity
 extern const Name structure;                             //!< Node type
 extern const Name success;
-extern const Name symmetric;         //!< Connectivity-related
-extern const Name synapse;           //!< Node type
-extern const Name synapse_label;     //!< Label id of synapses with labels
-extern const Name synapse_model;     //!< Connection parameters
-extern const Name synapse_modelid;   //!< Connection parameters
+extern const Name supports_precise_spikes; //!< true iff. suports precise
```
heplesser: "if model supports"

testsuite/unittests/test_supports_precise_spikes_flag.sli
```diff
@@ -0,0 +1,35 @@
...
+supports_precise_flag is a) available b) is set correctly by the precise and
+normal neuronal models.
+
+Author: Dimitri Plotnikov, 2016-09-09
+*/
+
+(unittest) run
+/unittest using
+
+{ /hh_psc_alpha GetDefaults /supports_precise_spikes get not } assert_or_die
```
heplesser: I would add at least one test with a precise neuron as well to check that it returns `true`.

DimitriPlotnikov: Yes, of course. As you can see, I add stuff it inclemently. Last time, I had a problem that NEST build worked locally, but CI refused it. So, now I'm trying to find the first instruction that breaks the build. Of course, in the next step I' ll add: { /parrot_neuron_ps GetDefaults /supports_precise_spikes get } assert_or_die iff. the current build succeeds.

jougs:
@DimitriPlotnikov: you can enable Travis CI for your own repositories [here](https://travis-ci.org/profile/DimitriPlotnikov). This way you can test things in your branches *before* starting a pull request and flooding all of NEST Initiative with each change. This should minimize the risk that things work locally and fail on Travis.

DimitriPlotnikov:
@jougs It is really great!

Issue #455, opened 2016-08-09
Fully support GIDCollections in SLI and Python


heplesser:
We decided in the [Open NEST VC 8 Aug](https://github.com/nest/nest-simulator/wiki/2016-08-08-Open-NEST-Developer-Video-Conference) to removed subnets which represent structure in the simulation kernel and move all structure representation to the script layer in form of GIDCollections (see als #417). We further decided to
- create release 2.12 as a release that will be backward compatible (so old scripts still run) but issue deprecation warnings on functions and argument types that will be removed with NEST 3.0
- have subnets fully removed from the NEST 3.0 release.

To ensure a smooth transition, it is essential that release 2.12 provides the full user interface of NEST 3.0, so that users can convert their scripts using the 2.12 release, _before_ we release NEST 3.0. We should thus perform as much work as possible on master, before creating the 2.12 release branch (where we will add deprecation warnings) and the no-subnet-branch preparing 3.0.

In particular, we need to ensure that
- all connection functions support GIDCollections
- all get/set functions support GIDCollections
- `GetConnections` supports GIDCollections
- creation functions return GIDCollections
- topology supports GIDCollections (create separate issue)
- related issues are resolved (#266, #355, #369)
- all tests and examples use GIDCollections instead of subnets (create separate issue)
  before branching.

For `Create` and other related functions, we need to introduce `NewCreate` and similar which return GIDCollections, so that they can coexist with the old `Create` in 2.12. When converting scripts to 3.0 later, one will then just need to replace `NewCreate` with `Create`.

jougs:
Thanks for writing this up. However, I disagree about the creation of `NewCreate`, `NewGetStatus`, etc.

In my opinion, no one will use them anyway until things break. Adding them just creates more work for us now and again when the transition occurs. Let's rather issue a good deprecation warning in the functions for 2.12.

Moreover, this is only relevant for SLI users, which are the smaller part of the community nowadays and we can provide support for the transition on the mailing list if required.

In PyNEST, most code will probably work seamlessly even with the change, e.g. consider
```
inhibitory = nest.Create("iaf_neuron", 10000)
nest.SetStatus(inhibitory, "V_m", -45.0)
```

heplesser:
@jougs I think your argument is sound.

heplesser:
# Design of GIDCollections

[...]

PR #483, opened 2016-09-14
Refactor Connection Generator Interface

jougs:
In particular this PR
- removes one level of indirection by moving all functionality from the `cg_connect.{h,cpp}` files to the `conngen.{h,cpp}` files
- converts all `CGConnect` variants to support `intvectordatum` and `gidcollection` istead of `subnet`
- adds automatic testing of [libneurosim](http://github.com/INCF/libneurosim) to the Travis CI setup 
- fixes and extends the PyNEST test that uses the Python implementation of [CSA](http://software.incf.org/software/csa) throu the Connection Generator Interface
- adds a manual SLI test for the Connection Generator Interface
- adds more documentation for the Connection Generator Interface

I propose @mdjurfeldt and @uahic as reviewers.


extras/parse_travis_log.py
```diff
@@ -191,13 +191,41 @@ def print_static_analysis(d):
                 print(INDENT + ' | ' + 2 * INDENT +
                       ' - ' + k3 + ' : ' + str(v3))
 
+def count_warnings_errors(f):
+    """
+    Counts compiler warnings and errors. Stops when reading '+make install'.
```
mdjurfeldt: Explain the meaning of the "line" return value.

extras/parse_travis_log.py
```diff
@@ -229,8 +257,8 @@ def print_static_analysis(d):
                 configure_ok, line = process_until(
                     f, 'You can now build and install NEST with')
 
-            if not make_ok and line.strip() == '+make':
-                make_ok, line = process_until(f, '+make install')
+            if line.strip() == '+make':
+                warnings, errors, line = count_warnings_errors(f)
 
             if not make_install_ok and line.startswith('+make install'):
```
mdjurfeldt: Apparently, count_warnings_errors can return the empty string in "line". Then this fails.

extras/parse_travis_log.py
```diff
@@ -241,26 +269,40 @@ def print_static_analysis(d):
             if line.strip() == 'WARNING: Not uploading results as this is a pull request':
                 uploading_results = False
 
+            if 'Skipping a deployment with the s3 provider because' in line:
+                uploading_results = False
+
+    # post process values
+    actual_warnings = { k.split("nest-simulator/")[1] :v for k,v in warnings.iteritems() if k.startswith('/home/travis/build') }
```
mdjurfeldt: I don't get this. From the initialization of "warnings" and here, it seems like it is supposed to be a dict, but in the code above it is assigned a return value from count_warnings_errors, which is an int. Have I missed something?

conngen/conngen.cpp
```diff
@@ -165,13 +64,13 @@ cg_connect( ConnectionGeneratorDatum& cg,
     // connect source to target
     while ( cg->next( source, target, NULL ) )
     {
-      if ( kernel().node_manager.is_local_gid( target + target_offset ) )
+      if ( kernel().node_manager.is_local_gid( target_gids[ target ] ) )
```
mdjurfeldt: OK, so the function I commented on last time was replaced, but same thing applies here: Why this test?

conngen/sli/conngen-interface.sli
```diff
@@ -20,66 +20,64 @@
...
+  % We create the type trie for the list variants using a loop
+  [/arraytype /intvectortype /gidcollectiontype] {
+    /gidlisttype Set
 
-/CGConnect_cg_a_a_l {
-  << >> exch
-  CGConnect_cg_a_a_D_l
-} def
+    % using empty parameter map and default synapse model
+    [/connectiongeneratortype gidlisttype gidlisttype] {
```
mdjurfeldt: So, we are not supposed to be able to use different gidlisttypes for sources and targets?

uahic:
I'll have a look at this later today

extras/parse_travis_log.py
jougs: These changes are from a commit (81a7b8048e5744) in an earlier pull request., which was pulled into my pull request by a merge from master. I can't comment on this code. Feel free to open an issue to explain the problem.

extras/parse_travis_log.py
jougs: See above.

extras/parse_travis_log.py
jougs: See above.

conngen/sli/conngen-interface.sli
jougs: Yes. This is consistent with the ordinary [connection routines](http://www.nest-simulator.org/connection_management/) in NEST. I think this is acceptable, as `cvgidcollection` should be pretty cheap in most situations. See also https://github.com/nest/nest-simulator/issues/455.

conngen/conngen.cpp
```diff
@@ -147,71 +147,349 @@ nest::cg_connect( nest::ConnectionGeneratorDatum& cg,
...
-nest::cg_parse_file( const StringDatum& xml )
-{
-  return ConnectionGenerator::fromXMLFile( xml );
+  int source, target, num_parameters = cg->arity();
+  if ( num_parameters == 0 )
+  {
+    // connect source to target
+    while ( cg->next( source, target, NULL ) )
+    {
+      if ( kernel().node_manager.is_local_gid( target + target_offset ) )
```
jougs: I've removed it in the latest commits.

conngen/conngen.cpp
jougs: I've removed it in the latest commits.

heplesser:
@jougs Could you fix the formatting errors? I will try to review your changes in the near future.

heplesser:
@heplesser volunteers as second reviewer.

conngen/conngen.cpp
```diff
@@ -132,86 +50,278 @@ nest::cg_connect( nest::ConnectionGeneratorDatum& cg,
...
+  // (counting from 0). This is used for index translation.
+  size_t cg_idx_left = 0;
 
-  RangeSet target_ranges;
-  cg_get_ranges( target_ranges, ( *targets ) );
+  // For sources, we only need to translate from NEST to CG indices.
+  for ( RangeSet::iterator source = sources.begin(); source != sources.end();
+        ++source )
+  {
+    size_t num_elements = source->last - source->first;
```
mdjurfeldt: I think we should rename num_elements -> num_elements_1 and add a comment that it is the number of elements minus 1.

heplesser: @mdjurfeldt @jougs I have several comments here, some of them concerning libneurosim design.  (a) One could add a few `const`s to the code. (b) I think `num_elements_1` would be difficult to understand, since it is no clear what the `1` represents; `num_elements_minus_1` would be much clearer. (c) Why does `ConnectionGenerator::ClosedInterval` use a logic where `last` is included, instead of the following the normal C++ "excluded  `end`" logic? Wouldn't the latter make code much more legible? (d) I also wonder why we have `RangeSet` here, which seems to be the same as `ConnectionGenerator::IntervalSet`. (e) The `ConnectionGenerator` classes should, where sensible, have a `size()` method, so users do not need to compute then number of elements directly.

extras/userdoc/md/documentation/connection-generator-interface.md
```diff
@@ -0,0 +1,55 @@
+# The Connection Generator Interface
+
+Since 2014 ([Djurfeldt et al., 2014](http://dx.doi.org/10.3389/fninf.2014.00043)) NEST supports the Connection Generator Interface. This allows to couple connection generating libraries to NEST without having to modify NEST itself.
+
+In contrast to the [built-in connection functions](connection-management.md) and the Topology module, the Connection Generator Interface has a different way for specifying the connectivity:
+
+2. pre- and postsynaptic neurons are created
```
heplesser: The enumeration seems garbled, shouldn't all lines begin with 1. for automatic numbering?

extras/userdoc/md/documentation/connection-generator-interface.md
```diff
@@ -0,0 +1,55 @@
...
+1. a Connection Generator object is created, which specifies the connectivity pattern
```
heplesser: No comma after "created".

extras/userdoc/md/documentation/connection-generator-interface.md
```diff
@@ -0,0 +1,55 @@
...
+3. the Connection Generator is applied to the sets of source and target neurons using the `CGConnect` function.
```
heplesser: Are source and target really sets in the mathematical sense?

extras/userdoc/md/documentation/connection-generator-interface.md
```diff
@@ -0,0 +1,55 @@
...
+```python
+import nest
+import csa
+
+# Create the neurons
+sources = nest.Create("iaf_psc_alpha", 2)
+targets = nest.Create("iaf_psc_alpha", 2)
+
+# Create the Connection Generator for one-to-one connectivity
+cg = csa.cset(csa.oneToOne)
```
heplesser: `oneToOne` mixes upper and lower case strangely. Assuming that it is a class, shouldn't it be `csa.OneToOne`, otherwise `csa.one_to_one`? But I guess it is an external library dependency that we are stuck with.

extras/userdoc/md/documentation/connection-generator-interface.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Connect the neurons
+nest.CGConnect(sources, targets, cg)
+
+# Verify the connectivity
+conn = nest.GetConnections(sources)
+print map(lambda x: x["target"], nest.GetStatus(conn))
```
heplesser: Guido van R says recommends comprehensions instead of `map()`. The `print` is also not Py3-safe. How about
```Python
print((c['source'], c['target']) for c in nest.GetStatus(conn))
```

extras/userdoc/md/documentation/connection-generator-interface.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
```
heplesser: It would be very useful to have examples showing how to use CG with parameters for the rule (indegree) and initialization of synapse parameters (weight, delay, ...).

testsuite/manualtests/test_conngen_interface.sli
```diff
@@ -0,0 +1,111 @@
+/*
```
heplesser: Why is this test in manualtests?

pynest/nest/lib/hl_api_connections.py
```diff
@@ -431,9 +410,9 @@ def CGConnect(pre, post, cg, parameter_map=None, model="static_synapse"):
 
     Parameters
     ----------
-    pre : list
+    pre : list or numpy.array
```
heplesser: I guess this could be a GIDCollection as well?

pynest/nest/lib/hl_api_connections.py
```diff
@@ -431,9 +410,9 @@ def CGConnect(pre, post, cg, parameter_map=None, model="static_synapse"):
...
         must contain 1 subnet, or a list of GIDs
```
heplesser: Remove all mentions of `subnet`

pynest/nest/lib/hl_api_connections.py
```diff
@@ -431,9 +410,9 @@ def CGConnect(pre, post, cg, parameter_map=None, model="static_synapse"):
...
-    post : list
+    post : list or numpy.array
         must contain 1 subnet, or a list of GIDs
```
heplesser: See pre

pynest/nest/lib/hl_api_connections.py
```diff
@@ -431,9 +410,9 @@ def CGConnect(pre, post, cg, parameter_map=None, model="static_synapse"):
...
     cg : connection generator
         libneurosim connection generator to use
```
heplesser: This applies to the line below, where I cannot add a comment: has parameter_map the same syntax and semantics as `syn_spec` in `Connect()`? I think slightly more information would be useful.

pynest/nest/lib/hl_api_connections.py
```diff
@@ -457,23 +436,8 @@ def CGConnect(pre, post, cg, parameter_map=None, model="static_synapse"):
...
-                 parameter_map, '/'+model, litconv=True)
-
-    else:
-        sli_func('CGConnect', cg, pre, post,
-                 parameter_map, '/'+model, litconv=True)
+    sli_func('CGConnect', cg, pre, post,
+             parameter_map, '/'+model, litconv=True)
 
 
 @check_stack
```
heplesser: Applies to line 445 below: typo "correcponding" -> "corresponding".

Applies to lines 459 and 490 below: I am confused by the word "Description" on a single line in the docstrings.

conngen/conngenmodule.cpp
```diff
@@ -281,23 +264,32 @@ ConnectionGeneratorModule::CGSelectImplementation_s_sFunction::execute(
    purposes. Usually, connections are created from a
    ConnectionGenerator using CGConnect.
 
+   Availability: Only if compiled with libneurosim support
```
heplesser: Applies to lines 258f above: doc says `sources`, `targets` must be lists, but shouldn' that be GIDCollections?

conngen/conngenmodule.cpp
```diff
@@ -68,15 +67,14 @@ ConnectionGeneratorModule::init( SLIInterpreter* i )
...
-  i->createcommand( "CGConnect_cg_iV_iV_D_l", &cgconnect_cg_iV_iV_D_lfunction );
+  i->createcommand( "CGConnect_cg_g_g_D_l", &cgconnect_cg_g_g_D_lfunction );
   i->createcommand( "CGParse", &cgparse_sfunction );
   i->createcommand( "CGParseFile", &cgparsefile_sfunction );
   i->createcommand(
     "CGSelectImplementation", &cgselectimplementation_s_sfunction );
 
   // Register the low level functions of the connection generator interface
-  i->createcommand( "cgsetmask_cg_iV_iV", &cgsetmask_cg_iV_iVfunction );
+  i->createcommand( "cgsetmask", &cgsetmask_cg_g_gfunction );
```
heplesser: `cgsetmask` and some other functions are described in the documentation below as belonging to the low-level API. (a) Do they need to be part of the public API? (b) Should they get names beginning with `:` on the SLI level to mark them as internal?

conngen/conngen.cpp
```diff
@@ -37,92 +34,13 @@
...
-    synmodel_id );
-}
 
 void
-nest::cg_connect( nest::ConnectionGeneratorDatum& cg,
-  IntVectorDatum& sources,
-  IntVectorDatum& targets,
+cg_connect( ConnectionGeneratorDatum& cg,
+  GIDCollection& source_gids,
+  GIDCollection& target_gids,
```
heplesser: Couldn't `source_gids` and `target_gids` be passed as `const` references?

conngen/conngen.cpp
```diff
@@ -132,86 +50,278 @@ nest::cg_connect( nest::ConnectionGeneratorDatum& cg,
...
 
-  RangeSet source_ranges;
-  cg_get_ranges( source_ranges, ( *sources ) );
+  RangeSet source_ranges, target_ranges;
 
-  RangeSet target_ranges;
-  cg_get_ranges( target_ranges, ( *targets ) );
+  cg_get_ranges( source_ranges, source_gids );
+  cg_get_ranges( target_ranges, target_gids );
+  cg_set_masks( cg, source_ranges, target_ranges );
```
heplesser: Here and in a few other places, we always have two calls to `cg_get_ranges()` followed by `cg_set_masks()`. `cg_get_ranges()` does not seem to be used in any other context. Wouldn't it be easier (and make for more readable code) to pass `source_gids` and `target_gids` directly to `cg_set_masks()` and then let `cg_set_masks()` get the ranges. 

One thing I really dislike about C++ is that in the `cg_get_ranges()` calls, there is no clear, visual data flow.
```C++
source_ranges = cg_get_ranges( source_gids );
```
would look so much cleaner, but then one would need smart pointers ...

mdjurfeldt:
@heplesser: You asked: It seems to me from examples/docs that when using CG from SLI, you must parse some XML, while that is not necessary from Py. Why not? I am still at a bit of a loss what the parsing is about.

The python bindings of csa and libcsa are part of these libraries and maintained on the CG side of the CG interface. SLI bindings are maintained on the simulator side of the CG interface.

There is no way for the simulator side to provide bindings for the CG side since the CG interface (intentionally!) provides very little knowledge about the CG side, so the only possibility to provide SLI bindings would be if csa and libcsa did this natively. (For Python, the situation is that *both* sides of the CG interface provide bindings to the user.)

However, the CG interface provides a generalized means for restoring CG:s from XML.  This entirely decouples the simulator side from the CG side such that scripting on the simulator side (in SLI in this case) becomes entirely independent of the CG side, since it can bind commands for restoring CG objects from XML and thereby specifying, creating as well as using CG objects without touching the CG side.

extras/userdoc/md/documentation/connection-generator-interface.md
mdjurfeldt: @heplesser It's not a class. It's a singleton object---a set of connections.

extras/userdoc/md/documentation/connection-generator-interface.md
mdjurfeldt: I would say that they are: At the most abstract level, we are giving sets of neuron indices to the CG. However, due to practical issues such as performance, these sets have other forms at the level of data types.

jougs:
@mdjurfeldt: You just re-explained what the [documentation added in this PR](https://github.com/jougs/nest-simulator/blob/c78582354f51c90b9fcca07d16560d1183bec5c9/extras/userdoc/md/documentation/connection-generator-interface.md) was meant to explain...

@heplesser: Is that documentation completely incomprehensible? I can add the text from @mdjurfeldt's comment if that helps...

heplesser:
@jougs I guess you refer to the documentation on SLI in [connection_generator_interface.md](https://github.com/nest/nest-simulator/pull/483/files#diff-8f48c26866d1b03d8220f61c5336d239R36), but that leaves me still totally lost. I will try reading the paper again, but it may be that I am missing some very basic pieces here.

conngen/conngen.cpp
jougs: Good catch. Fixed in 2a774a2.

conngen/conngen.cpp
jougs: I dislike both of your suggestions for adding a suffix to the variable name. Instead I've added `+1`/`-1` to the assignment and uses of the variable in 292f690 to make things still work and keep the simple name.

Regarding the other points  by @heplesser I'm sorry, but I think that this is not the place to discuss the design of libneurosim or the Connection Generator Interface. Please feel free to open corresponding issues or pull requests here: https://github.com/INCF/libneurosim/.

conngen/conngenmodule.cpp
jougs: You are right. In 4ba0af1 I have marked the internal and removed them from the `SeeAlso`s of all other functions.

conngen/conngenmodule.cpp
jougs: Fixed in 77915a1.

extras/userdoc/md/documentation/connection-generator-interface.md
jougs: Fixed in 29d9d3a.

extras/userdoc/md/documentation/connection-generator-interface.md
jougs: I agree that we need more comprehensive documentation and examples, but would rather defer this request to another PR. Please feel free to add a corresponding ticket once this PR is merged.

pynest/nest/lib/hl_api_connections.py
jougs: Not in PyNEST. `GIDCollection`s are currently only available in SLI.

pynest/nest/lib/hl_api_connections.py
jougs: Fixed in 44e8881.

pynest/nest/lib/hl_api_connections.py
jougs: Fixed in 44e8881.

pynest/nest/lib/hl_api_connections.py
jougs: I've updated and extended the documentation in 44e8881 to point out more clearly what `parameter_map` is used for.

testsuite/manualtests/test_conngen_interface.sli
jougs: Because it requires a) the C++ version of libcsa, which is not yet publicly available and b) some extensions to our test harness (e.g. handling the installation and setting environment variables) that I don't dare implementing until a) is fulfilled.

jougs:
@heplesser: I've addressed most of your comments in the recent commits. Next, I will update the code to use v2.0 of the Connection Generator Interface. If required, I can also provide more extensive documentation and possibly more examples after that.

conngen/conngenmodule.cpp
```diff
@@ -378,14 +366,24 @@ ConnectionGeneratorModule::CGNext_cgFunction::execute( SLIInterpreter* i ) const
...
+  int source_id, target_id;
+  if ( cgd->next( source_id, target_id, tmp_values ) )
   {
-    i->OStack.push( j );
-    i->OStack.push( k );
+    for ( int m = 0; m < arity; ++m )
+    {
+      values[ m ] = tmp_values[ m ];
+    }
+    delete[] tmp_values;
```
w-klijn: This delete only occurs in this branch of the if statement

conngen/conngenmodule.h
```diff
@@ -53,15 +53,10 @@ class ConnectionGeneratorModule : public SLIModule
   const std::string name() const;
   const std::string commandstring() const;
 
-  class CGConnect_cg_i_i_D_lFunction : public SLIFunction
+  class CGConnect_cg_g_g_D_lFunction : public SLIFunction
```
w-klijn: I would try and give this function a more descriptive name

pynest/nest/tests/test_csa.py
```diff
@@ -48,129 +48,125 @@
...
-        sources = nest.Create("iaf_neuron", n)
-        targets = nest.Create("iaf_neuron", n)
+        sources = nest.Create("iaf_psc_alpha", n_neurons)
+        targets = nest.Create("iaf_psc_alpha", n_neurons)
 
-        cg = csa.cset(csa.oneToOne)
+        cs = csa.cset(csa.oneToOne, weight, delay)
 
-        nest.CGConnect(sources, targets, cg)
+        nest.CGConnect(sources, targets, cs, {"weight": 0, "delay": 1})
```
w-klijn: This could use some comments: You add a cs with a set weight and delay, and here you add these additionally as parameters again. 
In general the tests could use comments to explain why you expect certain things to be like this.

pynest/nest/tests/test_csa.py
```diff
@@ -48,129 +48,125 @@
...
-        self.assertEqual(nest.sli_pop(), False)
+        cs = csa.cset(csa.oneToOne)
+        nonnodes = [1, 2, 3]
+
+        self.assertRaisesRegex(nest.NESTError, "UnknownNode",
+                               nest.CGConnect, nonnodes, nonnodes, cs)
+
+        n_neurons = 4
+
+        sources = nest.Create("iaf_psc_alpha", n_neurons)
```
w-klijn: You have two tests in a single unit-test. Depending on the unittest framework this might cause the second tests to be skipped if the first one fails.

.travis.yml
```diff
@@ -1,22 +1,22 @@
...
+    # and some additional jobs for interdependent options.
     # See also https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices
-    - xTHREADING=0 xMPI=0 xGSL=0 xLTDL=0 xREADLINE=0 xPYTHON=0 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # minimal
-    - xTHREADING=0 xMPI=1 xGSL=0 xLTDL=1 xREADLINE=1 xPYTHON=0 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # only MPI
-    - xTHREADING=1 xMPI=0 xGSL=0 xLTDL=1 xREADLINE=1 xPYTHON=0 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # only threading
-    - xTHREADING=1 xMPI=0 xGSL=0 xLTDL=0 xREADLINE=0 xPYTHON=1 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # Python & Threading
-    - xTHREADING=0 xMPI=1 xGSL=0 xLTDL=0 xREADLINE=0 xPYTHON=1 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # Python & MPI
-    - xTHREADING=0 xMPI=0 xGSL=0 xLTDL=0 xREADLINE=0 xPYTHON=1 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # only Python
-    - xTHREADING=1 xMPI=1 xGSL=1 xLTDL=1 xREADLINE=1 xPYTHON=1 xMUSIC=1 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # full
+    - xTHREADING=0 xMPI=0 xGSL=0 xLIBNEUROSIM=0 xLTDL=0 xREADLINE=0 xPYTHON=0 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # minimal
```
w-klijn: It's not related to this PR, but it looks like you do want a matrix for Python, mpi and threading. In the current setup mpi, threading and python are not tested in isolation.

You would then only need a single manual job

.travis.yml
```diff
@@ -61,6 +62,22 @@ before_install:
   # process json output from github api
   - sudo apt-get install -y jq
 
+  # Install libneurosim
```
w-klijn: Does this mean that libneurosim is always installed? I don't see a conditional on xLIBNEUROSIM

conngen/conngenmodule.cpp
```diff
@@ -68,17 +67,16 @@ ConnectionGeneratorModule::init( SLIInterpreter* i )
   ConnectionGeneratorType.setdefaultaction( SLIInterpreter::datatypefunction );
 
   // Register the user functions of the connection generator interface
```
w-klijn: Comment does not match the name of the function

conngen/conngenmodule.cpp
```diff
@@ -106,48 +104,29 @@ ConnectionGeneratorModule::init( SLIInterpreter* i )
...
    Author: Jochen Martin Eppler
    FirstVersion: August 2012
    SeeAlso: Connect, synapsedict, GetOptions, CGParse, CGParseFile,
-   CGSelectImplementation, cgstart,
-   cgsetmask, cgnext
+   CGSelectImplementation
 */
 
-// Connect for conn_generator subnet subnet dict synapsetype
+// CGConnect for conngen gidcollection gidcollection dict literal
```
w-klijn: It might be a convention, but adding the parameter types of a function causes tight coupling: If you want to change the interface you have to change the name. The Api will get really unstable this way

build.sh
```diff
@@ -240,6 +247,7 @@ make install
 echo "======= Install NEST end ======="
 
 echo "======= Test NEST start ======="
+export PYTHONPATH=$HOME/.cache/csa.install/lib/python2.7/site-packages:$PYTHONPATH
```
w-klijn: Hard coded python version

gtrensch: w-klijn, indeed, but this has already been reworked and improved in the latest version.

.travis.yml
jougs: Part of what you suggest was done in #409 and #640. More is to come in #789.

.travis.yml
jougs: As this is in accordance with the handling of all other dependencies, I would not want to change this here and rather do this in a refactoring of the whole dependency handling. A good opportunity is probably when we move to a Docker-based setup.

conngen/conngenmodule.cpp
jougs: I guess we can live with this. SLI's function names are confusing at times and especially to those new to the code.

conngen/conngenmodule.cpp
```diff
@@ -68,17 +67,16 @@ ConnectionGeneratorModule::init( SLIInterpreter* i )
...
   i->createcommand( "CGParse", &cgparse_sfunction );
   i->createcommand( "CGParseFile", &cgparsefile_sfunction );
   i->createcommand(
     "CGSelectImplementation", &cgselectimplementation_s_sfunction );
 
   // Register the low level functions of the connection generator interface
-  i->createcommand( "cgsetmask_cg_iV_iV", &cgsetmask_cg_iV_iVfunction );
-  i->createcommand( "cgstart", &cgstart_cgfunction );
-  i->createcommand( "cgnext", &cgnext_cgfunction );
+  i->createcommand( ":cgsetmask", &cgsetmask_cg_g_gfunction );
```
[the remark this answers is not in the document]
jougs: Yes. The first argument is the name under which the function is available in the SLI interpreter, while the second is a pointer to the C++ implementation. All command creations throughout SLI/NEST look like this.

conngen/conngenmodule.cpp
jougs: You are right. However, this is indeed a convention that makes it much easier to get things right on the SLI level. The documentation for the type codes can be found [here](https://github.com/nest/nest-simulator/blob/master/nestkernel/nestmodule.h#L60).

conngen/conngenmodule.cpp
jougs: `cgd->next()` won't put any more values into `tmp_values` before it returns `false`. So we don't have to delete anything in the `else` branch.

conngen/conngenmodule.h
jougs: `CGConnect` will probably be merged with the normal `Connect` function at one point. For backward-compatibility, I'd rather leave it like this in this PR, which is only about refactoring.

pynest/nest/lib/hl_api_connections.py
jougs: Typo fixed in fbd843400, stray word "Description" removed in bd2fdfe62.

pynest/nest/tests/test_csa.py
jougs: Comments added in 84d73c1a3. Thanks for pointing this out. We all get a little blind to these things when working too much with some code, don't we?

pynest/nest/tests/test_csa.py
jougs: Right. Thanks. I've split them in 15411851c.

testsuite/manualtests/test_conngen_interface.sli
jougs: With the upcoming merge of #774, `manualtests` will be gone. I can change this file to an example instead until we have a) and b) fulfilled. What do you think?

heplesser: @jougs Turning it into an example for now would be fine for me, provided we create a follow-up ticket reminding us to create the test (with Milestone 3.0). I suppose this also depends on #817 being fixed?

heplesser:
@jougs I have finally managed to install CSA under macOS, after some fixes to `autogen.sh` and `Makefile.am` (and building inside source ...). I can then import `csa` to NEST, but when I try to execute the example, I get the following error (part of trace deleted):
```python
In [6]: nest.CGConnect(s, t, cg)
NESTError                                 Traceback (most recent call last)
<ipython-input-6-fe56a981a096> in <module>()
----> 1 nest.CGConnect(s, t, cg)

[snip]

--> 165     sli_push(args)       # push array of arguments on SLI stack
    166     sli_push(s)          # push command string
    167     sli_run(slifun)      # SLI support code to execute s on args

pynestkernel.pyx in pynestkernel.NESTEngine.push (/Users/plesser/NEST/code/bld_csa_27/pynest/pynestkernel.cxx:4498)()

pynestkernel.pyx in pynestkernel.python_object_to_datum (/Users/plesser/NEST/code/bld_csa_27/pynest/pynestkernel.cxx:5590)()

pynestkernel.pyx in pynestkernel.python_object_to_datum (/Users/plesser/NEST/code/bld_csa_27/pynest/pynestkernel.cxx:6889)()

NESTError: unknown Python type: <class 'csa._elementary.OneToOne'>
```
But I have
```Python
In [7]: csa._elementary.OneToOne
Out[7]: csa._elementary.OneToOne
```

Do you have any idea what may be going wrong here?

I do not find `libcsa.*` in my `install/lib` directory, only `libpycsa.*`, so I couldn't test the pure SLI approach.

conngen/conngenmodule.cpp
```diff
@@ -365,10 +353,10 @@ ConnectionGeneratorModule::CGStart_cgFunction::execute(
    purposes. Usually, connections are created from a
    ConnectionGenerator using CGConnect.
```
heplesser: The Description above claims that a "list containing the values" may be pushed to the stack. From the code, it seems that the values are pushed, not a list of them.

conngen/conngenmodule.cpp
```diff
@@ -281,34 +261,42 @@ ConnectionGeneratorModule::CGSelectImplementation_s_sFunction::execute(
...
+  GIDCollectionDatum sources =
+    getValue< GIDCollectionDatum >( i->OStack.pick( 1 ) );
+  GIDCollectionDatum targets =
+    getValue< GIDCollectionDatum >( i->OStack.pick( 0 ) );
+
+  RangeSet source_ranges;
+  cg_get_ranges( source_ranges, sources );
+
+  RangeSet target_ranges;
+  cg_get_ranges( target_ranges, targets );
```
heplesser: We never use `source_ranges` and `target_ranges`, except as argument to `cg_set_masks()`.  Wouldn't it be easier to just pass `sources` and `targets` to `cg_set_masks()`?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+
+  cg->start();
 
-  Subnet* targets =
-    dynamic_cast< Subnet* >( kernel().node_manager.get_node( target_id ) );
-  if ( targets == NULL )
+  int source, target, num_parameters = cg->arity();
```
heplesser: Please split into separate defintions. `source` and `target` should probably be `long` or  `index`, while `num_parameters` can be `const`.

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
-      "targets must be a homogeneous subnet." );
-    throw BadProperty();
+    if ( not params_map->known( names::weight )
+      or not params_map->known( names::delay ) )
+    {
+      throw BadProperty(
+        "The parameter map has to contain the indices of weight and delay." );
+    }
+
+    long w_idx = ( *params_map )[ names::weight ];
```
heplesser: Why not `size_t`? Could be `const`?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+    long d_idx = ( *params_map )[ names::delay ];
+    std::vector< double > params( 2 );
+
+    // connect source to target with weight and delay
+    while ( cg->next( source, target, &params[ 0 ] ) )
```
heplesser: Is it really safe to use `&params[0]` here and assume contiguous memory layout for data in a `vector`?  I think the standard does not absolutely demand it; a C-style array would be safe.

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+      // created by cg_set_masks() only contain local nodes.
+      Node* const target_node =
+        kernel().node_manager.get_node( target_gids[ target ] );
+      const thread target_thread = target_node->get_thread();
+      kernel().connection_manager.connect( source_gids[ source ],
+        target_node,
+        target_thread,
+        synmodel_id,
+        params[ d_idx ],
+        params[ w_idx ] );
```
heplesser: Here we assume without checking that `d_idx` and `w_idx` are either 0 or 1. Is that safe?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+ * \param cg The ConnectionGenerator to set the masks on
+ * \param sources The source ranges to create the source masks from
+ * \param targets The target ranges to create the target masks from
+ */
+void
+cg_set_masks( ConnectionGeneratorDatum& cg,
+  RangeSet& sources,
+  RangeSet& targets )
+{
+  long np = kernel().mpi_manager.get_num_processes();
```
heplesser: `const size_t`

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+ * interface MPI aware and communicating the masks during connection
+ * setup.
+*/
 void
-nest::cg_connect( nest::ConnectionGeneratorDatum& cg,
-  IntVectorDatum& sources,
-  IntVectorDatum& targets,
-  const DictionaryDatum& params_map,
-  const Name& synmodel_name )
+cg_create_masks( std::vector< ConnectionGenerator::Mask >* masks,
```
heplesser: Wouldn't it be better to pass a `std::vector< ... >&` instead of a pointer?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+  // The index of the left border of the currently looked at range
+  // (counting from 0). This is used for index translation.
+  size_t cg_idx_left = 0;
+
+  // For sources, we only need to translate from NEST to CG indices.
+  for ( RangeSet::iterator source = sources.begin(); source != sources.end();
+        ++source )
   {
-    throw UnknownSynapseType( synmodel_name.toString() );
+    size_t num_elements = source->last - source->first + 1;
```
heplesser: `const`

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+    size_t right = cg_idx_left + num_elements - 1;
```
heplesser: `const`

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
-    source_ranges,
-    ( *sources ),
-    target_ranges,
-    ( *targets ),
-    params_map,
-    synmodel_id );
-}
+        // right is set to the CG index of the right border of the
+        // range. This is the same for all ranks.
+        size_t right = cg_idx_left + num_elements - 1;
```
heplesser: `const`

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+      // Make sure that the range is only added on as many ranks as
+      // there are elements in the range, or exactly on every rank,
+      // if there are more elements in the range.
+      if ( proc <= num_elements - 1 )
+      {
+        // For the different ranks, left will take on the CG indices
+        // of all first local nodes that are contained in the range.
+        // The rank, where this mask is to be used is determined
+        // below when inserting the mask.
+        size_t left = cg_idx_left + proc;
```
heplesser: `const`

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+  {
+    size_t num_elements = target->last - target->first + 1;
+    for ( size_t proc = 0; proc
+            < static_cast< size_t >( kernel().mpi_manager.get_num_processes() );
+          ++proc )
+    {
+      // Make sure that the range is only added on as many ranks as
+      // there are elements in the range, or exactly on every rank,
+      // if there are more elements in the range.
+      if ( proc <= num_elements - 1 )
```
heplesser: Why `p <= n-1` instead of the more common `p < n`?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+*/
 void
-nest::cg_connect( nest::ConnectionGeneratorDatum& cg,
-  IntVectorDatum& sources,
-  IntVectorDatum& targets,
-  const DictionaryDatum& params_map,
-  const Name& synmodel_name )
+cg_create_masks( std::vector< ConnectionGenerator::Mask >* masks,
+  RangeSet& sources,
+  RangeSet& targets )
```
heplesser: As far as I can see, `sources` and `targets` are not modified below. Why not pass as `const &`?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+ * creates the masks from the given RangeSets sources and targets.
+ *
+ * \param cg The ConnectionGenerator to set the masks on
+ * \param sources The source ranges to create the source masks from
+ * \param targets The target ranges to create the target masks from
+ */
+void
+cg_set_masks( ConnectionGeneratorDatum& cg,
+  RangeSet& sources,
+  RangeSet& targets )
```
heplesser: As far as I can see, `sources` and `targets` are not modified below. Why not pass as `const&`?

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
   {
-    LOG( M_ERROR,
-      "CGConnect_cg_i_i_D_l",
-      "Only 1-dim subnets are supported as sources." );
-    throw BadProperty();
+    throw UnknownSynapseType( synmodel_name.toString() );
   }
+  const index synmodel_id = static_cast< index >( synmodel );
+
+  RangeSet source_ranges, target_ranges;
```
heplesser: Avoid multiple definitions on one line (at least for my taste).

conngen/conngen.cpp
```diff
@@ -37,185 +34,312 @@
...
+
+  cg_get_ranges( source_ranges, source_gids );
+  cg_get_ranges( target_ranges, target_gids );
+  cg_set_masks( cg, source_ranges, target_ranges );
```
heplesser: As far as I can see, `source_ranges` and `target_ranges` are only created to be passed to `cg_set_masks()`, but not used in subsequent code. Wouldn't it make code here more easily readable if the `cg_get_ranges()` calls were moved into `cg_set_masks()`, which would then just receive the `*_gids`?

conngen/conngenmodule.cpp
jougs: Good catch. Fixed in fe4408921.

conngen/conngenmodule.cpp
jougs: Fixed in 0066013c9.

conngen/conngenmodule.cpp
```diff
@@ -378,14 +366,24 @@ ConnectionGeneratorModule::CGNext_cgFunction::execute( SLIInterpreter* i ) const
   ConnectionGeneratorDatum cgd =
     getValue< ConnectionGeneratorDatum >( i->OStack.pick( 0 ) );
 
-  int j, k;
-  std::vector< double > values;
+  int arity = cgd->arity();
```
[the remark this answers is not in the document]
jougs: Fixed in 43e8d6abf.

conngen/conngenmodule.cpp
```diff
@@ -378,14 +366,24 @@ ConnectionGeneratorModule::CGNext_cgFunction::execute( SLIInterpreter* i ) const
...
+  double* tmp_values = new double[ arity ];
+  std::vector< double > values( arity );
+
   i->OStack.pop( 1 );
-  i->EStack.pop();
-  if ( cg_next( cgd, j, k, values ) )
+
+  int source_id, target_id;
```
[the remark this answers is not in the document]
jougs: * no, the `ConnectionGenerator` interface defines them as `int`
* Fixed in 43e8d6abf.

conngen/conngenmodule.cpp
```diff
@@ -378,14 +366,24 @@ ConnectionGeneratorModule::CGNext_cgFunction::execute( SLIInterpreter* i ) const
...
+  if ( cgd->next( source_id, target_id, tmp_values ) )
```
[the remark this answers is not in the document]
jougs: I don't know. I'm currently just implementing the bindings for NEST, not the interface they are using.

conngen/conngen.cpp
jougs: These lines were removed in fe4408921.

conngen/conngen.cpp
jougs: Fixed in fe4408921.

conngen/conngen.cpp
jougs: The split and addition of `const` was done in 43e8d6abf. I'll keep them as `int` as this is what the `ConnectionGenerator` interface defines them as.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: They are `const GIDCollection&` now. See fe4408921.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: They are `const GIDCollection&` now. See fe4408921.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

conngen/conngen.cpp
jougs: Fixed in 43e8d6abf.

jougs:
@heplesser: Thanks for the thorough review. I've addressed your minor comments in the latest commits and will look at the more complex ones either later today or tomorrow.

This is what's missing:
* Add doxygen documentation for all functions in `conngen.h`
* `conngenmodule.cpp:380`: Do we really need to copy here? It seems the only thing we do with `values` is to loop over it and push its elements individually to the stack.
* `conngenmodule.cpp:395`: Where is the `lock()` that we unlock here---maybe comment?
* `conngen.cpp:92`: Is it really safe to use `&params[0]` here and assume contiguous memory layout for data in a `vector`? I think the standard does not absolutely demand it; a C-style array would be safe.
* `conngen.cpp:104`: Here we assume without checking that `d_idx` and `w_idx` are either 0 or 1. Is that safe?

Regarding your [issues with libneurosim](https://github.com/nest/nest-simulator/pull/483#issuecomment-327026285): I've seen the same problem on JURECA last week and will investigate further as soon as the machine is back from maintenance. I'll report back when I know more.

jougs:
Just an additional note: this should probably not be merged before #817 was addressed. Currently, we are likely to miss proper testing of libneurosim/CSA related code.

jougs:
@heplesser: I have successfully reproduced your problem with libneurosim. Contrary to what I said before, I now think that this is a proper bug and not something that only occurs on some machines. I've opened #821 with some more explanations and a question for you.

jougs:
@heplesser I have addressed your remaining issues in the latest commits.

> Add doxygen documentation for all functions in `conngen.h`

960122341 adds a Doxygen comment for `cg_connect()`. As all other functions were already documented in the `.cpp` file. I thus also added this one there. I hope this is OK.

> `conngenmodule.cpp:380`: Do we really need to copy here? It seems the only thing we do with `values` is to loop over it and push its elements individually to the stack.

70d77c4e8 removes the copying.

> `conngenmodule.cpp:395`: Where is the `lock()` that we unlock here---maybe comment?

The `lock()` is implicit as the object is derived from a `lockptr`. The [doxygen comment](https://github.com/nest/nest-simulator/blob/master/libnestutil/lockptr.h#L77) there has an explanation, which I'd rather not repeat at the place of usage

> `conngen.cpp:92`: Is it really safe to use `&params[0]` here and assume contiguous memory layout for data in a `vector`? I think the standard does not absolutely demand it; a C-style array would be safe.

[This answer on Stack Overflow](https://stackoverflow.com/a/849190) says that this guarantee was missing in C++98 but added later. I assume it is safe enough as I can only think of rather complicated ways a compiler should implement a vector non-contiguously given all its other properties.

> `conngen.cpp:104`: Here we assume without checking that `d_idx` and `w_idx` are either 0 or 1. Is that safe?

156b88bff adds a check and a corresponding exception for this.

> Convert the manualtest to an example

a835bd47b200 converts the test to an example.

jougs:
Please note that my latest commits are rather untested due to https://github.com/INCF/csa/issues/6.

heplesser:
@jougs Thanks for your fixes! Travis was unstable, I have restarted the builds there (failed again).

Concerning the doxygen comment placement, we can leave it where it is now, but generally my understanding is that class and method documentation should be in the header file, since the header might be all that developers see (a bit of closed source thinking here, but I think it makes sense).

Concerning the testing: How much work would it be to create a very simple connection generation code, so that one at least would have basic testing in place?

heplesser:
@jougs Thanks to advice from @lekshmideepu I now realised that the problem causing Travis to fail is that the `autogen.sh` process for CSA fails because `README` is missing in  CSA (it has `README.md`). One way to fix this for now would be to create a symlink from `README` to `README.md` in our yaml file. Alternatively, should we temporarily replace CSA by some stub connectivity generator for test purposes?

heplesser:
@jougs With CSA issue 7 fixed it looks pretty fine now, just some clang-formatting issues to fix.

jougs:
Thanks for the review. I've addressed the remaining issues as follows:
* 892c7fe55b adds `const` for the bools and fixes formatting
* 1cf4325e40 fixes the testing issues by setting `LD_LIBRARY_PATH` in the build script
* cdec29539e adds improves the output of the CSA test and removes the libcsa test. It can be re-added once libcsa is available in the wild

heplesser:
@jougs Great work! But I am confused after a look at the test configuration including libneurosim: There is says (l 5868ff):
```
  One-to-one connectivity using CGConnect with id intvectors ... SKIP: Python CSA package is not available
  One-to-one connectivity using CGConnect with paramters ... SKIP: Python CSA package is not available
  One-to-one connectivity using CGConnect with synmodel ... SKIP: Python CSA package is not available
  One-to-one connectivity using CGConnect with id tuples ... SKIP: Python CSA package is not available
  Error handling of CGConnect in case of unknown nodes ... SKIP: Python CSA package is not available
  Error handling of CGConnect in case of unknown synapse model ... SKIP: Python CSA package is not available
```
So the tests in `test_csa.py` are apparently not run on Travis, maybe due to a problem with Python paths? Those tests should run, shouldn't they? Could you check this?

jougs:
[D'oh!](https://vignette1.wikia.nocookie.net/buddyfight/images/d/d8/Homer-simpson-quotes-doh-i11.jpg/revision/latest?cb=20141227165956)

The commands `make` and `make install` were missing from the CSA recipe. I've added them in 1e423f5. Let's see if this solves the issue with not running the test.

heplesser:
Well done! And merging ...

jougs:
Many thanks to all who made this long story come to a good end. What a relief :-)

PR #485, opened 2016-09-15
Improved performance of threaded connection generation for large networks

heplesser:
This PR integrates work by @tammoippen providing improved performance of threaded connection generation.

I suggest @jougs and @jakobj as reviewers.


jakobj:
short notice: I am currently running some simulations on JQ with this branch, just for double checking. I'll get around to reviewing the code this week.

nestkernel/conn_builder.cpp
```diff
@@ -602,34 +609,78 @@ nest::OneToOneBuilder::connect_()
...
+            skip_conn_parameter_( tid );
+            continue;
+          }
+
+          single_connect_( *sgid, *target, target_thread, rng );
         }
+      }
+      else
+      {
+        for ( SparseNodeArray::const_iterator it = get_local_nodes().begin();
```
jakobj: I don't really like that we need to iterate over a vector that is (for good reasons) hidden deep in the node manager. Would it be possible to ask the `NodeManager` for a `GidCollection` of all local nodes?

jakobj:
so here are some preliminary results from JQ weak scaling benchmarks (12500 neurons/node, 8 threads/node, 500ms simulation duration, `hpc_benchmark.sli`):
jq_weak_master_threadedperf2
dark blue: current master
pink: threaded_perf2

nestkernel/conn_builder.cpp
heplesser: @jakobj The problem with a `GidCollection` would be that it represents gobal sets of GIDs in a rather abstract fashion that first needs to be expanded. The `SparseNodeArray` gives us access to local node pointers right away. But we are risking encapsulation here, so I will look at how to implement this cleaner.

nestkernel/conn_builder.cpp
```diff
@@ -1337,18 +1475,18 @@ nest::FixedTotalNumberBuilder::connect_()
...
       double p_local = num_local_targets / ( size_targets - sum_dist );
       bino.set_p( p_local );
       bino.set_n( N_ - sum_partitions );
       num_conns_on_vp[ k ] = bino.ldev();
     }
 
-    sum_dist += static_cast< double >( targets_on_vp[ k ].size() );
-    sum_partitions += static_cast< unsigned int >( num_conns_on_vp[ k ] );
+    sum_dist += static_cast< double >( number_of_targets_on_vp[ k ] );
+    sum_partitions += static_cast< uint >( num_conns_on_vp[ k ] );
```
heplesser: Will be fixed in next commit.

nestkernel/connection_manager.cpp
```diff
@@ -78,7 +78,13 @@ nest::ConnectionManager::ConnectionManager()
...
 nest::ConnectionManager::~ConnectionManager()
 {
-  delete_connections_();
+  // Memory leak on purpose!
+  // The ConnectionManager is deleted, when the network is deleted, and
+  // this happens only, when main() is finished and we give the allocated memory
+  // back to the system anyway. Hence, why bother cleaning up our highly
+  // scattered connection infrastructure? They do not have any open files, which
+  // need to be closed or similar.
+  // delete_connections_();
```
heplesser: I removed the last line. The method is selves is needed elsewhere in the code in conjunction with `ResetKernel`.

nestkernel/connection_manager.cpp
```diff
@@ -195,38 +201,40 @@ nest::ConnectionManager::set_synapse_status( index gid,
...
+    for ( size_t t = 0; t < connections_.size(); ++t )
     {
+      for (
+        tSConnector::nonempty_iterator iit = connections_[ t ].nonempty_begin();
+        iit != connections_[ t ].nonempty_end();
+        ++iit )
+      {
 #ifdef USE_PMA
-      validate_pointer( *iit )->~ConnectorBase();
+        validate_pointer( *iit )->~ConnectorBase();
```
heplesser: As far as I can see, the indentation is correct wrt the bracket before the `#ifdef`.

nestkernel/connection_manager.cpp
```diff
@@ -195,38 +201,40 @@ nest::ConnectionManager::set_synapse_status( index gid,
...
-  for ( tVSConnector::iterator it = connections_.begin();
-        it != connections_.end();
-        ++it )
+#ifdef _OPENMP
+#pragma omp parallel
   {
-    for ( tSConnector::nonempty_iterator iit = it->nonempty_begin();
-          iit != it->nonempty_end();
-          ++iit )
+#pragma omp for schedule( static, 1 )
```
heplesser: I think the explicit `for schedule( static, 1 )` here makes sense, since it will ensure a fixed assignment of threads to loop iterations. I would rather think we should introduce it in other places as well (might improve performance?).

heplesser:
@jakobj Thank you for your review! I have tried to answer most points above and will fix the access to the `SparseNodeArray` shortly. We will soon convert NEST to `GidCollections` throughout, so that issue should be fixed.

heplesser:
@jakobj I just pushed a commit that should address the issues you raised. Access to the  `SparseNodeArray` is now encapsulated better, and `ConnBuilder` no longer a `friend` of `NetworkManager`. Since all access is via `const_iterator`s, it is also safe. Once we have removed subnets, we can simplify code in `ConnBuilder` further (see #493).

heplesser:
@suku248 could you look at this?

nestkernel/connection_manager.cpp
jakobj: you're right. my mistake.

nestkernel/connection_manager.cpp
jakobj: well, we are using the thread id in many places to directly access `connections_[ tid ]` in an `omp parallel` block, as you point out, so this assignment should also hold here. i agree with you that we should think about our threading model in nest, but this might be beyond the scope of this PR.

jakobj:
looks good so far. i don't have a strong opinion on the whole `omp for` vs `omp parallel` section, so i would also be fine if you left this unchanged.

nestkernel/connection_manager.cpp
heplesser: @tammoippen I think you added the `omp for schedule`. Do you have a comment on this?

tammoippen: My hope was, that this makes the code clearer: there are already a lot `#ifdef`'s and not using the `#pragma omp for schedule( static, 1 )` would increase this further. Another point is, that the indentation becomes 'more' wrong, if we are compiling with OpenMP, as the outer loop is gone and the inner loop is indented one level to much. Here is a version with the `pragma omp for` removed:

```C++
void
nest::ConnectionManager::delete_connections_()
{
#ifdef _OPENMP
#pragma omp parallel
  {
    size_t t = kernel().vp_manager.get_thread_id();
#else
    for ( size_t t = 0; t < connections_.size(); ++t )
    {
#endif
      for (
        tSConnector::nonempty_iterator iit = connections_[ t ].nonempty_begin();
        iit != connections_[ t ].nonempty_end();
        ++iit )
      {
#ifdef USE_PMA
        validate_pointer( *iit )->~ConnectorBase();
#else
      delete validate_pointer( *iit );
#endif
      }
      connections_[ t ].clear();
#ifdef _OPENMP
    }
#endif

#if defined _OPENMP && defined USE_PMA
#ifdef IS_K
    poormansallocpool[ t ].destruct();
    poormansallocpool[ t ].init();
#else
    poormansallocpool.destruct();
    poormansallocpool.init();
#endif
#endif

#ifdef _OPENMP
  }
#endif
}
```

heplesser: @tammoippen Thank you for the explanation! Makes sense to me.

nestkernel/node_manager.h
```diff
@@ -327,6 +349,24 @@ NodeManager::any_node_uses_wfr() const
...
+inline SparseNodeArray::const_iterator
+NodeManager::local_nodes_begin() const
+{
+  return local_nodes_.begin();
+}
+
+inline SparseNodeArray::const_iterator
+NodeManager::local_nodes_end() const
+{
+  return local_nodes_.begin();
```
heplesser: @suku248 Thank you for catching this! Nasty copy paste error, and rather unfortunate that it it not trigger any test. I will try to create tests that detect this.

heplesser:
@suku248 I have now pushed commits that add tests for the bug you discovered, fixes the issues you pointed out, cleans up code a little more, and merges recent changes from the kernel.

testsuite/mpitests/test_all_to_all.sli
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+
+/* BeginDocumentation
+    Name: testsuite::test_all_to_all - Test correct connection with many targets
+    Synopsis: (test_all_to_all) run -> - 
+
+    Description:
+    test_all_to_all.sli checks that all-to-all connections are created
+    correctly if then number of targets exceeds the number of local nodes.
```
suku248: Replace `then` with `the` in all 4 tests

testsuite/mpitests/test_fixed_indegree.sli
```diff
@@ -0,0 +1,57 @@
...
+/* BeginDocumentation
+    Name: testsuite::test_fixed_indegree - Test correct connection with many targets
+    Synopsis: (test_fixed_indegree) run -> - 
+
+    Description:
+    test_fixed_indegree.sli checks that all-to-all connections are created
+    correctly if then number of targets exceeds the number of local nodes.
+    
+    Author: Hans Ekkehard Plesser
+    SeeAlso: testsuite::test_one_to_one, testsuite::test_fixed_indegree,
```
suku248: Replace `testsuite::test_fixed_indegree` with `testsuite::test_all_to_all`

testsuite/mpitests/test_fixed_indegree.sli
```diff
@@ -0,0 +1,57 @@
...
+    testsuite::test_bernoulli
```
suku248: Replace with `testsuite::test_pairwise_bernoulli`

testsuite/mpitests/test_one_to_one.sli
```diff
@@ -0,0 +1,50 @@
...
+/* BeginDocumentation
+    Name: testsuite::test_one_to_one - Test correct connection with many targets
+    Synopsis: (test_one_to_one) run -> - 
+
+    Description:
+    test_one_to_one.sli checks that all-to-all connections are created
+    correctly if then number of targets exceeds the number of local nodes.
+    
+    Author: Hans Ekkehard Plesser
+    SeeAlso: testsuite::test_one_to_one, testsuite::test_fixed_indegree,
```
suku248: Replace `testsuite::test_one_to_one` with `testsuite::test_all_to_all`

testsuite/mpitests/test_one_to_one.sli
```diff
@@ -0,0 +1,50 @@
...
+    testsuite::test_bernoulli
```
suku248: Replace with `testsuite::test_pairwise_bernoulli`

testsuite/mpitests/test_pairwise_bernoulli.sli
```diff
@@ -0,0 +1,50 @@
...
+    Name: testsuite::test_pairwise_bernoulli - Test correct connection with many targets
+    Synopsis: (test_pairwise_bernoulli) run -> - 
+
+    Description:
+    test_pairwise_bernoulli.sli checks that all-to-all connections are created
+    correctly if then number of targets exceeds the number of local nodes.
+    
+    Author: Hans Ekkehard Plesser
+    SeeAlso: testsuite::test_one_to_one, testsuite::test_fixed_indegree,
+    testsuite::test_bernoulli
```
suku248: Replace with `testsuite::test_bernoulli` with `testsuite::test_all_to_all`

suku248:
:+1: for merge

heplesser:
@suku248 Thank you! I have just committed the requested changes to the documentation.

PR #490, opened 2016-09-22
Improve get_connections performance

alexeyshusharin:
Hi All,

I found that GetConnections is quite slow and consumes unreasonable amount of memory. Quick investigation has shown that ConnectionManager::get_connections pre-allocates memory for all connections in the model even if user requests connections of a single node. It reserves space only for Token (16 bytes), but that become megabytes in large models. In addition the memory consumption is doubled by similar per-thread reservation. Obviously it also decreases the performance.

This PR improves get_connections by:
 - Using stl deque for collecting connection data. Only final result is converted to ArrayDatum. This allows to avoid memory pre-allocation.
 - Removing per-thread connection number calculation since reservation is not required anymore.
 - Skipping iteration through thread connections if there are no connections with specified syn_id.

The simple test shows significant improvement of get_connection performance. It's about 118 times faster for requsting connections of a single source node. 55 times faster for a single target node. And 50% faster for requesting all connections (BTW, most time is spent outside c++ get_connection method).

The Python script measures average time of GetConnection call duration. It was run on my desktop with enabled OpenMP (16 threads).

```
Configuration:
Num of source nodes = 2500
Num of target nodes = 2500
Num of connections per node = 701
Num of all connections = 1757571
```
```
Master
------------------------------------------
Average time for getting connection is
per source =  0:00:00.463445
per target =  0:00:00.586650
all        =  0:00:01.772015
```
```
deque based
------------------------------------------
Average time for getting connection is
per source =  0:00:00.003906
per target =  0:00:00.010477
all        =  0:00:01.171967
```

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1018,7 +1018,7 @@ nest::ConnectionManager::get_num_connections( synindex syn_id ) const
 ArrayDatum
 nest::ConnectionManager::get_connections( DictionaryDatum params ) const
 {
-  std::deque<ConnectionID> connectome;
+  std::deque< ConnectionID > connectome;
 
   const Token& source_t = params->lookup( names::source );
   const Token& target_t = params->lookup( names::target );
@@ -1071,7 +1071,7 @@ nest::ConnectionManager::get_connections( DictionaryDatum params ) const
 
   while ( !connectome.empty() )
   {
-    result.push_back( ConnectionDatum ( connectome.front() ) );
+    result.push_back( ConnectionDatum( connectome.front() ) );
     connectome.pop_front();
   }
 
@@ -1079,14 +1079,17 @@ nest::ConnectionManager::get_connections( DictionaryDatum params ) const
 }
 
 void
-nest::ConnectionManager::get_connections( std::deque<ConnectionID>& connectome,
+nest::ConnectionManager::get_connections(
+  std::deque< ConnectionID >& connectome,
   TokenArray const* source,
   TokenArray const* target,
   size_t syn_id,
   long synapse_label ) const
 {
-  if ( get_num_connections( syn_id ) == 0)
+  if ( get_num_connections( syn_id ) == 0 )
+  {
     return;
+  }
 
   if ( source == 0 and target == 0 )
   {
@@ -1098,7 +1101,7 @@ nest::ConnectionManager::get_connections( std::deque<ConnectionID>& connectome,
     for ( thread t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
     {
 #endif
-      std::deque<ConnectionID> conns_in_thread;
+      std::deque< ConnectionID > conns_in_thread;
 
       for ( index source_id = 1; source_id < connections_[ t ].size();
             ++source_id )
@@ -1116,7 +1119,7 @@ nest::ConnectionManager::get_connections( std::deque<ConnectionID>& connectome,
         {
           while ( !conns_in_thread.empty() )
           {
-            connectome.push_back(conns_in_thread.front());
+            connectome.push_back( conns_in_thread.front() );
             conns_in_thread.pop_front();
           }
         }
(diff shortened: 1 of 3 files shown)


heplesser:
@jougs Could you take a look?

nestkernel/connection_manager.cpp
```diff
@@ -1122,7 +1116,13 @@ nest::ConnectionManager::get_connections( ArrayDatum& connectome,
 #ifdef _OPENMP
 #pragma omp critical( get_connections )
 #endif
-        connectome.append_move( conns_in_thread );
+        {
+          while ( !conns_in_thread.empty() )
```
heplesser: @alexeyshusharin This code repeats two more times below, so I would suggest to refactor it into a method.

Two more details: 
In newly written NEST code, logical operators (`!`, `&&`, ÌI`) should be written out as `not`, `and`, `or` for greater clarity.

I am also wondering if `std::copy()` wouldn't be faster and more expressive, although it would come at the price of temporarily having two copies of the connectivity data. But since the doubling is per-thread only, I think that would not be too bad.

alexeyshusharin: @heplesser, @jougs Thanks for review!

> In newly written NEST code, logical operators (!, &&, ÌI) should be written out asnot,and,or` for greater clarity.

Sure, will change it.

> I am also wondering if std::copy() wouldn't be faster and more expressive, although it would come at the price of temporarily having two copies of the connectivity data. But since the doubling is per-thread only, I think that would not be too bad.

Actually I did use std::copy() in the first version of the changes and I didn't notice any performance benefits of using it. Maybe it's faster but considering the whole method execution time it's not significant. So I decided that memory is more important than just nicely looking code. If you agree I will move this code to a separate method as you suggested.

heplesser: @alexeyshusharin Fine! Moving the code to a method with an informative name will make the code as readable as with `std::copy()` and save a bit of memory.

alexeyshusharin: @heplesser Done!

nestkernel/connection_manager.cpp
```diff
@@ -1163,13 +1157,7 @@ nest::ConnectionManager::get_connections(
...
 #pragma omp critical( get_connections )
 #endif
-        {
-          while ( !conns_in_thread.empty() )
-          {
-            connectome.push_back( conns_in_thread.front() );
-            conns_in_thread.pop_front();
-          }
-        }
+        append_move_connections_( conns_in_thread, connectome );
```
heplesser: This is fine in principle, but the data flow is not clear to the reader. If the method returned the reference to the connectome, one could write
```C++
connectome = append_from_thread( connectome, conns_in_thread );
```
In my eyes that would be much easier to read.

nestkernel/connection_manager.h
```diff
@@ -393,6 +400,17 @@ ConnectionManager::get_max_delay() const
   return max_delay_;
 }
 
+inline void
+ConnectionManager::append_move_connections_( std::deque< ConnectionID >& input,
```
heplesser: Inlining makes sense here, but the code should go into the `cpp`-file, since the method is used only there. In that way, changes to the code in the method won't require recompilation of all files including `connection_manager.h`.

nestkernel/connection_manager.cpp
alexeyshusharin: @heplesser Thanks for the comments!
Maybe you also like something like this. Looks good from my point of view. :)
```
static inline std::deque< nest::ConnectionID >&
operator<<( std::deque< nest::ConnectionID >& out,
  std::deque< nest::ConnectionID >& in )
{
  while ( not in.empty() )
  {
    out.push_back( in.front() );
    in.pop_front();
  }

  return out;
}
```

And then just write
`connectome << conns_in_thread;`

nestkernel/connection_manager.h
alexeyshusharin: I, personally, prefer to implement it as a static inline method in cpp file (local for cpp). But `grep` gave me just one place in the code where this approach is used. So I decided that this is not common practice for nest and made it as a class member.
Do you think it's ok to declare this method as static and place it just before `get_connections` in cpp file? Or do you think that moving implementation to cpp and leaving declaration in hpp is better approach?

nestkernel/connection_manager.cpp
heplesser: @alexeyshusharin Beautiful!

PR #491, opened 2016-09-22
Revise implementation of ht_neuron

heplesser:
Implement the full NMDA model from the Hill-Tononi paper with instantaneous blocking and fast and slow unblocking. This PR will change model behavior.

As reviewers, I suggest @ingablundell and @suku248.


models/ht_connection.h
```diff
@@ -200,8 +197,8 @@ template < typename targetidentifierT >
 HTConnection< targetidentifierT >::HTConnection()
   : ConnectionBase()
   , weight_( 1.0 )
-  , tau_P_( 50.0 )
-  , delta_P_( 0.2 )
+  , tau_P_( 500.0 )
+  , delta_P_( 0.5 )
```
Silmathoron: Since this could potentially have an influence on existing user scripts, could you justify the change in default parameters?

models/ht_neuron.cpp
```diff
@@ -49,8 +49,7 @@ RecordablesMap< ht_neuron >::create()
     Name( "Theta" ), &ht_neuron::get_y_elem_< ht_neuron::State_::THETA > );
   insert_(
     Name( "g_AMPA" ), &ht_neuron::get_y_elem_< ht_neuron::State_::G_AMPA > );
-  insert_(
-    Name( "g_NMDA" ), &ht_neuron::get_y_elem_< ht_neuron::State_::G_NMDA > );
+  insert_( Name( "g_NMDA" ), &ht_neuron::get_g_NMDA_ );
```
Silmathoron: Why is the recordable ``g_NMDA`` different from the state variable? (cf. ``get_g_NMDA``)

**EDIT:** ok, I understand, the state variable ``G_NMDA`` is the g(t) from the docstring and the recordable is the real ``g_NMDA = g(t)*m(V, t)``, which cannot be stored directly as a state variable since it is not what's updated.
But then the naming looks misleading to me, I guess we should change ``G_NMDA`` to something else and update the docstring... though I must say I'm at a loss regarding the name. Maybe ``G_T`` would be the most sensible choice...

models/ht_neuron.cpp
```diff
@@ -83,14 +82,29 @@ ht_neuron_dynamics( double, const double y[], double f[], void* pnode )
   // Synaptic channels
   double I_syn = 0;
 
+  /*
+   * NMDA conductance
+   *
+   * We need to take care to handle instantaneous blocking correctly.
+   * If the unblock-variables Mg_{fast,slow} > Mg_ss, the steady-state
```
Silmathoron: This looks unclear to me, you mean "if the unblock variables are greater than their respective steady-state values", right? I think the combination of equation and text sentence with commas is rather disturbing.

models/ht_neuron.cpp
```diff
@@ -208,11 +224,13 @@ nest::ht_neuron::Parameters_::Parameters_()
   , AMPA_Tau_2( 2.4 ) // ms
   , AMPA_E_rev( 0.0 ) // mV
   , NMDA_g_peak( 0.075 )
-  , NMDA_Tau_1( 4.0 )  // ms
-  , NMDA_Tau_2( 40.0 ) // ms
-  , NMDA_E_rev( 0.0 )  // mV
-  , NMDA_Vact( -58.0 ) // mV
-  , NMDA_Sact( 2.5 )   // mV
+  , NMDA_Tau_1( 4.0 )        // ms
+  , NMDA_Tau_2( 40.0 )       // ms
```
Silmathoron: If you think it's worth it (I'm not sure), having all ``tau``s in lowercase would be more coherent, though it would break existing codes... and ``Tau_m`` instead of ``tau_m`` is really too bad since it's supposed to be common to many models...

models/ht_neuron.h
```diff
@@ -280,7 +300,10 @@ class ht_neuron : public Archiving_Node
       DG_GABA_A,
```
Silmathoron: I would be in favour of changing ``G_NMDA`` to ``G_T`` but there should at least be a comment like
``G_NMDA, // state variable corresponding to g(t), with g_NMDA = m(V, t) * g(t)``
or something equivalent.

models/ht_neuron.h
```diff
@@ -409,6 +432,26 @@ class ht_neuron : public Archiving_Node
   {
     return S_.I_h_;
   }
+  double_t
```
Silmathoron: why are there ``double_t``?

models/ht_connection.h
heplesser: The actual values of these two parameters are not given in the Hill-Tononi (2005) paper. The new values are the default values in the original Synthesis simulator code by Sean Hill. I sent a mail to NEST User on 22 September warning of changes to the Hill-Tononi model and asking for comments from anyone who may be affected, but have not received a single answer, so I think we are safe here.

models/ht_neuron.cpp
heplesser: Naming is a bit of a mess here. In the paper, it is \tilde{g}_NMDA = m(V) g_NMDA(t), where g_NMDA(t) is the double exponential function.

I think that the end user will be most interested in the actual conductance of NMDA channels when Mg-blocking is taken into account (the m(V)). Therefore, I think the recordable should be `g_NMDA` and the getter `get_g_NMDA_()`, but I think I will change the name of the state variable, maybe to `G_NMDA_TIMECOURSE`.

models/ht_neuron.h
heplesser: I suppose because the branch is a bit old .... Fixed.

models/ht_neuron.cpp
Silmathoron: Yes, I agree that ``g_NMDA`` should be the variable recorded, and ``G_NMDA_TIMECOURSE`` looks good!

models/ht_connection.h
Silmathoron: Ok, sounds fine!

heplesser:
@Silmathoron I now noticed that the model needs a bit more work. I will change all names to use `names::` names, will add checks to parameter setting (currently, there are no checks at all) and change all `_Tau_` and `_Theta_` in names to lowercase equivalents.

Silmathoron:
Ok, thanks for your hard work!
May the code be with you!

models/ht_neuron.cpp
```diff
@@ -83,14 +82,29 @@ ht_neuron_dynamics( double, const double y[], double f[], void* pnode )
   // Synaptic channels
   double I_syn = 0;
 
+  /*
+   * NMDA conductance
+   *
+   * We need to take care to handle instantaneous blocking correctly.
+   * If the unblock-variables Mg_{fast,slow} > Mg_ss, the steady-state
+   * value for the present membrane potential, we cannot change those values
+   * in State_[], since the ODE Solver may call this function multiple times
```
ingablundell: I would prefer it if "those values" were specified

ingablundell:
The documentation of the NMDA current is very clear and helpful! Although I would add an explicit definition of g (what is g_NMDA in the paper). The calculation of the NMDA part of I_syn and propagation of Mg_fast and Mg_slow look correct to me! I approve this pull request.

models/ht_neuron.cpp
heplesser: I will fix this with the next push to the repo.

models/ht_neuron.h
```diff
@@ -59,29 +59,47 @@
...
 
-   In comparison to the model described in the paper, the following
-   modifications were mare:
+   NMDA conductances is modeled as follows
 
-   - NMDA conductance is given by g(t) = g_peak * m(V), where
+     g_NMDA(t, V) = m(V, t) g(t)
 
-       m(V) = 1 / ( 1 + exp( - ( V - NMDA_Vact ) / NMDA_Sact ) )
+   where g(t) is a beta function (difference of exponentials).
```
heplesser: I will fix this with the next update to the PR.

heplesser:
@Silmathoron I have now completed my revisions, so you can continue reviewing.

models/ht_neuron.h
```diff
@@ -59,48 +59,72 @@
...
+     g(t) = g_peak ( e^(-t/tau_1) - e^(-t/tau_2) )
+               / ( e^(-t_peak/tau_1) - e^(-t_peak/tau_2) )
 
-   - Several apparent typographical errors in the descriptions of
-     the intrinsic currents were fixed, hopefully in a meaningful
-     way.
+     t_peak = tau_2 tau_1 / ( tau_2 - tau_1 ) ln( tau_2 / tau_1 )
+
+     m(V, t) = a(V) m_fast*(V, t) + ( 1 - a(V) ) m_slow*(V, t)
+     a(V)    = 0.51 - 0.0028 V
```
Silmathoron: There is an apparent problem with the homogeneity, here... I suppose 0.0028 is in mV^{-1}, but could you specify it explicitely? (maybe introduce a named constant).
Shouldn't the user want to be able to set these two values? If not (which would be surprising to me), would you mind explaining their origin?

models/ht_neuron.h
```diff
@@ -224,25 +247,27 @@ class ht_neuron : public Archiving_Node
 
     // Parameters for synapse of type AMPA, GABA_A, GABA_B and NMDA
     double AMPA_g_peak;
```
Silmathoron: Why are there no units on all conductances?

models/ht_neuron.h
```diff
@@ -410,6 +438,27 @@ class ht_neuron : public Archiving_Node
     return S_.I_h_;
   }
 
+  double
+  get_g_NMDA_() const
+  {
+    const double A1 = 0.51 - 0.0028 * S_.y_[ State_::VM ];
```
Silmathoron: Why aren't these doubles parameters? (see comment in docstring)

nestkernel/nest_names.cpp
```diff
@@ -37,6 +37,10 @@ const Name Act_m( "Act_m" );
 const Name address( "address" );
 const Name alpha_1( "alpha_1" );
 const Name alpha_2( "alpha_2" );
+const Name AMPA_E_rev( "AMPA_E_rev" );
```
Silmathoron: Why start the names with ``AMPA``? ``E_rev_AMPA`` would look more in accordance with the naming standards to me... same applies for the others.

nestkernel/nest_names.cpp
```diff
@@ -137,15 +142,30 @@ const Name frozen( "frozen" );
...
 const Name g_Kv1( "g_Kv1" );
 const Name g_Kv3( "g_Kv3" );
 const Name g_L( "g_L" );
 const Name g_Na( "g_Na" );
+const Name g_NaL( "g_NaL" );
+const Name g_NMDA( "g_NMDA" );
 const Name g_rr( "g_rr" );
 const Name g_sfa( "g_sfa" );
+const Name g_spike( "g_spike" );
+const Name GABA_A_E_rev( "GABA_A_E_rev" );
```
Silmathoron: See related comment for ``AMPA``; also the conductances follow the "standard logic" with ``g_`` coming first, which comforts me in the idea that ``E_rev_GABA_A`` would be better.

nestkernel/nest_names.cpp
```diff
@@ -155,6 +175,8 @@ const Name growth_rate( "growth_rate" );
 const Name gsl_error_tol( "gsl_error_tol" );
 
 const Name h( "h" );
+const Name h_E_rev( "h_E_rev" );
```
Silmathoron: same comment as above

nestkernel/nest_names.cpp
```diff
@@ -179,6 +207,9 @@ const Name Interpol_Order( "Interpol_Order" );
 const Name interval( "interval" );
 const Name is_refractory( "is_refractory" );
 
+const Name KNa_E_rev( "KNa_E_rev" );
```
Silmathoron: see above

nestkernel/nest_names.cpp
```diff
@@ -202,8 +233,18 @@ const Name N_channels( "N_channels" );
 const Name n_events( "n_events" );
 const Name n_proc( "n_proc" );
 const Name n_receptors( "n_receptors" );
+const Name NaP_E_rev( "NaP_E_rev" );
```
Silmathoron: see above

nestkernel/nest_names.cpp
```diff
@@ -202,8 +233,18 @@ const Name N_channels( "N_channels" );
...
+const Name NaP_g_peak( "NaP_g_peak" );
 const Name needs_prelim_update( "needs_prelim_update" );
 const Name neuron( "neuron" );
+const Name NMDA_E_rev( "NMDA_E_rev" );
```
Silmathoron: see above

nestkernel/nest_names.cpp
```diff
@@ -288,6 +332,8 @@ const Name synapse_model( "synapse_model" );
 const Name synapse_modelid( "synapse_modelid" );
 const Name synaptic_elements( "synaptic_elements" );
 
+const Name T_E_rev( "T_E_rev" );
```
Silmathoron: see above

heplesser:
@Silmathoron I have just pushed a set of changes to the ht_neuron that, I hope, address most of your questions. I will now prepare a notebook detailing the model and its implementation, and also include some tests.

heplesser:
@Silmathoron  The Notebook on the model is still work in progess. I have pushed it nonetheless for your information during the review.

models/ht_connection.h
```diff
@@ -222,9 +219,9 @@ HTConnection< targetidentifierT >::get_status( DictionaryDatum& d ) const
 {
   ConnectionBase::get_status( d );
   def< double >( d, names::weight, weight_ );
-  def< double >( d, "tau_P", tau_P_ );
-  def< double >( d, "delta_P", delta_P_ );
-  def< double >( d, "P", p_ );
+  def< double >( d, names::tau_P, tau_P_ );
```
Silmathoron: Fair enough ;)

models/ht_connection.h
```diff
@@ -178,15 +180,12 @@ HTConnection< targetidentifierT >::send( Event& e,
   double t_lastspike,
   const CommonSynapseProperties& )
 {
-  double h = e.get_stamp().get_ms() - t_lastspike;
-  Node* target = get_target( t );
-  // t_lastspike_ = 0 initially
-
   // propagation t_lastspike -> t_spike, t_lastspike_ = 0 initially, p_ = 1
```
Silmathoron: change ``t_spike`` to ``t_ref``

models/ht_neuron.cpp
```diff
@@ -313,95 +326,224 @@ nest::ht_neuron::State_::~State_()
...
+  {
+    throw BadParameter( "g_peak_T >= 0 required." );
+  }
+  if ( g_peak_h < 0 )
+  {
+    throw BadParameter( "g_peak_h >= 0 required." );
+  }
+  if ( t_ref < 0 )
+  {
+    throw BadParameter( "t_spike >= 0 required." );
```
Silmathoron: change ``t_spike`` to ``t_ref``

models/ht_neuron.cpp
```diff
@@ -578,18 +722,18 @@ nest::ht_neuron::calibrate()
...
 
-  V_.cond_steps_[ GABA_B - 1 ] =
-    get_synapse_constant( P_.GABA_B_Tau_1, P_.GABA_B_Tau_2, P_.GABA_B_g_peak );
+  V_.cond_steps_[ GABA_B - 1 ] = get_synapse_constant(
+    P_.tau_rise_GABA_B, P_.tau_decay_GABA_B, P_.g_peak_GABA_B );
 
-  V_.PotassiumRefractoryCounts_ = Time( Time::ms( P_.t_spike ) ).get_steps();
+  V_.PotassiumRefractoryCounts_ = Time( Time::ms( P_.t_ref ) ).get_steps();
 
   // since t_spike_ >= 0, this can only fail in error
```
Silmathoron: change ``t_spike`` to ``t_ref``

models/ht_neuron.cpp
```diff
@@ -653,22 +797,28 @@ ht_neuron::update( Time const& origin, const long from, const long to )
...
 
-    // Deactivate potassium current after spike time have expired
-    if ( S_.r_potassium_ && --S_.r_potassium_ == 0 )
+    // Enforce instantaneous blocking of NMDA channels, see comment
+    // in ht_neuron_dynamics().
+    const double Mg_ss = Mg_steady_state_( S_.y_[ State_::VM ] );
+    S_.y_[ State_::Mg_slow ] = std::min( Mg_ss, S_.y_[ State_::Mg_slow ] );
+    S_.y_[ State_::Mg_fast ] = std::min( Mg_ss, S_.y_[ State_::Mg_fast ] );
+
+    // Deactivate potassium current after t_spike
```
Silmathoron: change ``t_spike`` to ``t_ref``

models/ht_neuron.h
```diff
@@ -59,49 +59,79 @@
...
 
-   Warning:
+   Note:
    THIS MODEL NEURON HAS NOT BEEN TESTED EXTENSIVELY!
 
    Parameters:
    V_m            -  membrane potential
-   spike_duration - duration of re-polarizing potassium current
-   Tau_m          - membrane time constant applying to all currents but
+   t_spike - duration of re-polarizing potassium current
```
Silmathoron: remove ``t_spike``

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+    "N_T & \\quad \\text{not given} \\\\\n",
+    "m_T^{\\infty}(V) &= 1/\\{1 +  \\exp[ -(V +  59.0)/6.2]\\} \\\\\n",
+    "\\tau_{m,T}(V) &= \\{0.22/\\exp[ -(V  + 132.0)/ 16.7]\\} +  \\exp[(V  + 16.8)/18.2] +  0.13\\\\\n",
+    "h_T^{\\infty}(V) &= 1/\\{1 +  \\exp[(V +  83.0)/4.0]\\} \\\\\n",
+    "\\tau_{h,T}(V) &= \\langle  8.2 +  \\{56.6 +  0.27 \\exp[(V +  115.2)/5.0]\\}\\rangle / \\{1.0 +  \\exp[(V +  86.0)/3.2]\\}\n",
+    "\\end{align}\n",
+    "\n",
+    "Note the following:\n",
+    "- $N_T=2$ from Synthesis code\n",
+    "- In the equation for $\\tau_{m,T}$, the second exponential term must be added to the first (in the denominator) to make dimensional sense\n",
```
Silmathoron: I guess this means that the constants 0.22 and 0.13 have the dimension of a time? could you precise this?
EDIT: ok, this appears just below... but still ^^

models/ht_connection.h
heplesser: No, here `t_spike` is actually the time of the spike passing through the synapse.

Silmathoron: Ok, right, I went a bit fast on that one ^^"

heplesser:
All should be fixed now. I am continuing work on the notebook, next step is a systematic test of the intrinsic currents (voltage-clamping is coming to NEST ...)

heplesser:
@Silmathoron @ingablundell I have just pushed a major revision of the model including a notebook exploring all aspects of the model (docs/model_details). Enjoy!

models/ht_neuron.cpp
```diff
@@ -74,26 +70,23 @@ ht_neuron_dynamics( double, const double y[], double f[], void* pnode )
...
    * NMDA conductance
    *
    * We need to take care to handle instantaneous blocking correctly.
-   * If the unblock-variables Mg_{fast,slow} are greater than the steady-state
-   * value M_ss for the present membrane potential, we cannot change
-   * Mg_{fast,slow} values in State_[], since the ODE Solver may call this
-   * function multiple times and in arbitrary temporal order. We thus need to
-   * use local variables for the values at the current time, and check the
-   * state variables once the ODE solver has completed the time step.
+   * If the unblock-variables m_NMDA_{fast,slow} are greater than the
```
Silmathoron: Naming is not in agreement with lines 87--88 (``m_{fast,slow}_NMDA``)

models/ht_neuron.cpp
```diff
@@ -191,17 +169,76 @@ ht_neuron_dynamics( double, const double y[], double f[], void* pnode )
...
+nest::ht_neuron::D_eq_KNa_( double V ) const
+{
+  const double D_influx_peak = 0.025;
+  const double D_thresh = -10.0;
+  const double D_slope = 5.0;
+  const double D_eq_0 = 0.001;
+
+  const double D_influx =
+    D_influx_peak / ( 1.0 + std::exp( -( V - D_thresh ) / D_slope ) );
+  return P_.tau_D_KNa * D_influx + D_eq_0;
```
Silmathoron: ``D_eq_0`` was not present before and the value returned by the function is apparently not homogeneous... could you explain the role of this new variable?

models/ht_neuron.cpp
```diff
@@ -535,15 +568,29 @@ nest::ht_neuron::Parameters_::set( const DictionaryDatum& d )
...
 {
-  def< double >( d, names::V_m, y_[ VM ] );      // Membrane potential
+  def< double >( d, names::V_m, y_[ V_M ] );     // Membrane potential
   def< double >( d, names::theta, y_[ THETA ] ); // Threshold
 }
 
 void
-nest::ht_neuron::State_::set( const DictionaryDatum& d, const Parameters_& )
+nest::ht_neuron::State_::set( const DictionaryDatum& d,
+  const ht_neuron& node,
```
Silmathoron: I don't understand this modification of the function

models/ht_neuron.cpp
```diff
@@ -577,7 +624,7 @@ nest::ht_neuron::Buffers_::Buffers_( const Buffers_&, ht_neuron& n )
 nest::ht_neuron::ht_neuron()
   : Archiving_Node()
   , P_()
-  , S_( P_ )
+  , S_( *this, P_ )
```
Silmathoron: See above

models/ht_neuron.h
```diff
@@ -128,14 +126,19 @@
...
                                   conductance, see above
-   instant_unblock_NMDA         - Instantaneous NMDA unblocking (default: false)
+   instant_unblock_NMDA         - instantaneous NMDA unblocking (default: false)
    {E_rev,g_peak}_{h,T,NaP,KNa} - reversal potential and peak conductance for
                                   intrinsic currents
    receptor_types               - dictionary mapping synapse names to ports on
                                   neuron model
-   recordables                  - list of recordable quantities.
+   recordables                  - list of recordable quantities
+   equilibrate                  - if given and true, time-dependent activation
```
Silmathoron: Maybe a precision that values are otherwise initialized at zero would be good

models/ht_neuron.h
```diff
@@ -285,12 +288,15 @@ class ht_neuron : public Archiving_Node
 
     double g_peak_KNa;
     double E_rev_KNa; // mV
+    double tau_D_KNa; // ms; in P_ for technical reasons, not meant to be public
```
Silmathoron: Why not?

doc/model_details/HillTononiModels.ipynb
```diff
@@ -0,0 +1,2069 @@
...
+    "\n",
+    "##### Corrected exponent\n",
+    "\n",
+    "According to Compte et al (2003), $N_{NaP}=3$, i.e.,\n",
+    "\\begin{equation}\n",
+    "I_{NaP} = g_{\\text{peak,NaP}}(m_{NaP}^{\\infty}(V))^3(V-E_{NaP})\n",
+    "\\end{equation}\n",
+    "This equation is also given in a comment in Synthesis, but is missing from the implementation.\n",
+    "\n",
+    "**Note: NEST implements the equation according to Compte et al (2003) with $N_{NaP}=3$, while Synthesis uses $N_{NaP}=1**\n",
```
heplesser: Done.

models/ht_neuron.cpp
heplesser: Done.

models/ht_neuron.cpp
heplesser: Renamed `D_eq` as in the paper/equations.

models/ht_neuron.cpp
heplesser: Several of the intrinsic channels have slow dynamics, O(1s) for I_h, e.g. If you change parameters, especially the membrane potential using `SetStatus/SetDefaults`, it will often be useful to set the state variables for these channels to their new steady-state values. Otherwise, one would have to simulate several seconds of biological time to get to equilibrium. Especially when setting model parameters at the beginning of a simulation, the new equilibration capability should be useful. When changing parameters during the simulation, on the other hand, you probably want to see how the system responds to the change (eg my voltage stepping tests and the wake-sleep transitions in HT05), so then one will not want to equilibrate.

models/ht_neuron.cpp
heplesser: ... and same answer, essentially: the state should be initialized to the correct steady-state values.

models/ht_neuron.h
heplesser: I added and "otherwise" part.

doc/model_details/HillTononiModels.ipynb
```diff
@@ -0,0 +1,2069 @@
...
+    "\n",
+    "where $X$ is \"slow\" or \"fast\". $a(V)$ expresses voltage-dependent weighting between slow and fast unblocking, $m^{\\infty}(V)$ the steady-state value of the proportion of unblocked NMDA-channels, the minimum condition in $m_X^*(V,t)$ the instantaneous blocking and the differential equation for $m_X(V,t)$ the unblocking dynamics.\n",
+    "\n",
+    "Synthesis uses tabluated values for $m^{\\infty}$. NEST uses the best fit of $V_{\\text{act}}$ and $S_{\\text{act}}$  to the tabulated data for conductance table `fNMDA`.\n",
+    "\n",
+    "**Note**: NEST also supports instantaneous NMDA dynamics using a boolean switch. In that case $m(V, t)=m^{\\infty}(V)$. \n",
+    "\n",
+    "#### Synaptic \"minis\"\n",
+    "\n",
+    "Synaptic \"minis\" due to spontaneous release of neurotransmitter quanta [HT05, p 1679] are not included in the NEST implementation of the Hill-Tononi model. This due to the fact that the total mini input rate for a cell was just 2 Hz and they cause PSP changes by $0.5 \\pm 0.25$mV only and thus should have minimal effect.\n",
```
Silmathoron: Ok, thanks

models/ht_neuron.cpp
Silmathoron: What about the homogeneity?
I noticed that, indeed, the equation for [D](https://github.com/nest/nest-simulator/pull/491/files#diff-a2b4ee8efed0b2df67fe89245f4fce87R227) does not seem to satisfy this requirement, unless ``D_influx`` has a dimension of $T^{-1}$, which does not seem to be the case...

models/ht_neuron.h
heplesser: I turned it into a normal, accessible parameter now.

heplesser:
I  have just pushed a new version addressing @Silmathoron's comments and also
- integrating insights from #547, especially moving threshold check into the ODE-integration loop and handling refractoriness in continuous time ("inside" approach)
- shortened the documentation in the code, since the details are now in the notebook; the details on NMDA-unblocking seemed out of balance, as no other, more basic equations were shown.

models/ht_neuron.cpp
heplesser: Here I need both parameters, since I compute some equilibrium values explicitly in the constructor.

models/ht_neuron.cpp
heplesser: I hadn't thought about this, but you are right. The dimension of D does not matter, as long as it is the same as of d_1/2, since the two are divided in the end, so dimensions drop out. I suggest to leave D and d_1/2 dimensionless, since they represent some "factor" which has no immediate physical counterpart (but models some sort of ion concentration). Then D_eq also is dimensionless, while D_influx and D_influx,peak must have units of inverse time, explicitly 1/ms. This makes sense, since these values essentially represent ion concentration *change*. I have updated the notebook.

heplesser:
@ingablundell Does your "thumbs up" still stand after the changes I made in response to @Silmathoron's comments?

models/ht_neuron.cpp
Silmathoron: Of course, I read too fast, sorry...

models/ht_neuron.cpp
Silmathoron: Ok, good!

doc/model_details/HillTononiModels.ipynb
```diff
@@ -221,22 +221,29 @@
...
     "| --: | --: | --: | --: | --: | --: |\n",
     "|$0.025\\text{ms}^{-1}$ |$0.001$|$1250\\text{ms}$|$-10\\text{mV}$|$5\\text{mV}$|$0.25$|\n",
     "\n",
+    "Note the following:\n",
+    "- $D_{eq}$ is the equilibrium value only for $D_{\\text{influx}}(V)=0$, i.e., in the limit $V\\to -\\infty$ and $t\\to\\infty$.\n",
+    "- The actual steady-state value is $D_{\\infty}$.\n",
+    "- $m_{DK}^{\\infty}$ is a steep sigmoid which is almost 0 or 1 except for a narrow window around $d_{1/2}$.\n",
+    "- To the left of this window, $I_{DK}\\approx 0$.\n",
+    "- To the right of this window, $I_{DK}\\sim -(V-E_{DK})$.\n",
+    "The equations implemented in NEST are thus\n",
```
Silmathoron: Either some other equation are missing or this is a wrong copy-paste

doc/model_details/HillTononiModels.ipynb
```diff
@@ -221,22 +221,29 @@
...
     "\n",
     "\\begin{align}\n",
-    " \\frac{dD}{dt} &= D_{\\text{influx}} - \\frac{D-D_{\\text{eq}}}{\\tau_D} \\\\\n",
-    " D_{\\text{influx}} &= \\frac{D_{\\text{influx,peak}}}{1+ \\exp\\left(-\\frac{V-D_{\\theta}}{\\sigma_D}\\right)} \\\\\n",
-    " m_{DK}^{\\infty} &= \\frac{1}{1 + \\left(\\frac{d_{1/2}}{D}\\right)^{3.5}}\n",
+    "I_{DK} &= - g_{\\text{peak},DK} (m_{DK}^{\\infty}(V,t))^3 (V - E_{DK})\\\\\n",
+    " m_{DK}^{\\infty} &= \\frac{1}{1 + \\left(\\frac{d_{1/2}}{D}\\right)^{3.5}}\\\\\n",
+    " \\frac{dD}{dt} &= D_{\\text{influx}}(V) - \\frac{D-D_{\\text{eq}}}{\\tau_D} = \\frac{D_{\\infty}(V)-D}{\\tau_D} \\\\\n",
+    " D_{\\infty}(V) &= \\tau_D D_{\\text{influx}}(V) + {D_{\\text{eq}}}\\\\\n",
+    " D_{\\text{influx}} &= \\frac{D_{\\text{influx,peak}}}{1+ \\exp\\left(-\\frac{V-D_{\\theta}}{\\sigma_D}\\right)} \n",
```
Silmathoron: Could you add in the "Note the following" list what you told me about the fact that, despite their ambiguous names, ``D_{influx}`` and ``D_{influx,peak}``
>  represent ion concentration change

If you could elaborate a little on this here, I think it would be a good thing.

heplesser: Done.

doc/model_details/HillTononiModels.ipynb
heplesser: Yes, that was a copy-paste leftover, fixed.

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The Neuron Model\n",
+    "\n",
+    "### Integration \n",
+    "\n",
+    "The original Synthesis implementation of the model uses Runge-Kutta integration with fixed 0.25 ms step size, and integrates channels dynamics first, followed by integration of membrane potential and threshold.\n",
+    "\n",
```
ingablundell: Maybe you could state which RK solver you are using?

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+    "\\begin{equation}\n",
+    "\\frac{\\text{d}V}{\\text{d}t} = \\frac{-g_{\\text{NaL}}(V-E_{\\text{Na}})\n",
+    "-g_{\\text{KL}}(V-E_{\\text{K}})+I_{\\text{syn}}+I_{\\text{int}}}{\\tau_{\\text{m}}}\n",
+    "-\\frac{g_{\\text{spike}}(V-E_{\\text{K}})}{\\tau_{\\text{spike}}}\n",
+    "\\end{equation}\n",
+    "\n",
+    "- The equation does not contain membrane capacitance. As a side-effect, all conductances are dimensionless.\n",
+    "- Na and K leak conductances $g_{\\text{NaL}}$ and $g_{\\text{KL}}$ are constant, although $g_{\\text{KL}}$ may be adjusted on slow time scales to mimic neuromodulatory effects.\n",
+    "- Reversal potentials are assumed constant.\n",
```
ingablundell: I would prefer: Reversal potentials $E_{NA}$ and $E_{K}$ are assumed constant.

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+    "\\end{align}\n",
+    "\n",
+    "Note that subscript $h$ in some cases above marks the $I_h$ channel, \n",
+    "\n",
+    "#### Low-threshold calcium current $I_T$\n",
+    "\n",
+    "Synthesis: `ItChannel`\n",
+    "\n",
+    "##### Equations given in paper \n",
+    "\n",
```
ingablundell: Why are you calling it "equations" and not "definitions"?

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+   "metadata": {},
+   "source": [
+    "### Intrinsic currents\n",
+    "\n",
+    "Note that not all intrinsic currents are active in all populations of the network model presented in [HT05, p1678f].\n",
+    "\n",
+    "Intrinsic currents are based on the Hodgkin-Huxley description, i.e.,\n",
+    "\n",
+    "\\begin{align}\n",
+    "I_X &= g_{\\text{peak}, X} m_X(V, t)^N_X h_X(V, t)(V-E_X) \\\\\n",
```
ingablundell: I would prefer it if the meaning of the variables were stated explicitly. But maybe this is not necessary.

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+    "\n",
+    "where $X$ is \"slow\" or \"fast\". $a(V)$ expresses voltage-dependent weighting between slow and fast unblocking, $m^{\\infty}(V)$ the steady-state value of the proportion of unblocked NMDA-channels, the minimum condition in $m_X^*(V,t)$ the instantaneous blocking and the differential equation for $m_X(V,t)$ the unblocking dynamics.\n",
+    "\n",
+    "Synthesis uses tabluated values for $m^{\\infty}$. NEST uses the best fit of $V_{\\text{act}}$ and $S_{\\text{act}}$  to the tabulated data for conductance table `fNMDA`.\n",
+    "\n",
+    "**Note**: NEST also supports instantaneous NMDA dynamics using a boolean switch. In that case $m(V, t)=m^{\\infty}(V)$. \n",
+    "\n",
+    "#### Synaptic \"minis\"\n",
+    "\n",
+    "Synaptic \"minis\" due to spontaneous release of neurotransmitter quanta [HT05, p 1679] are not included in the NEST implementation of the Hill-Tononi model. This due to the fact that the total mini input rate for a cell was just 2 Hz and they cause PSP changes by $0.5 \\pm 0.25$mV only and thus should have minimal effect.\n",
```
ingablundell: This *is* due...

doc/model_details/HillTononi.ipynb
```diff
@@ -0,0 +1,792 @@
...
+    "The conductance change due to a single input spike at time $t=0$ through a channel of type $X$ is given by (see below for exceptions)\n",
+    "\n",
+    "\\begin{align}\n",
+    "    \\bar{g}_X(t) &= g_X(t)\\\\\n",
+    "    g_X(t) &= g_{\\text{peak}, X}\\frac{\\exp(-t/\\tau_1) - \\exp(-t/\\tau_2)}{\n",
+    "                 \\exp(-t_{\\text{peak}}/\\tau_1) - \\exp(-t_{\\text{peak}}/\\tau_2)} \\Theta(t)\\\\\n",
+    "     t_{\\text{peak}} &= \\frac{\\tau_2 \\tau_1}{\\tau_2 - \\tau_1} \\ln\\frac{ \\tau_2}{\\tau_1}\n",
+    "\\end{align} \n",
+    "\n",
+    "where $t_{\\text{peak}}$ is the time of the conductance maximum and $\\tau_1$ and $\\tau_2$ are synaptic rise- and decay-time, respectively; $\\Theta(t)$ is the Heaviside step function. The equation is integrated using exact integration in Synthesis; in NEST, it is included in the ODE-system integrated using RK.\n",
```
ingablundell: again: which RK?

doc/model_details/HillTononi.ipynb
heplesser: Done.

doc/model_details/HillTononi.ipynb
heplesser: Done.

doc/model_details/HillTononi.ipynb
heplesser: Because physicist are a sloppy bunch and call most things with a "=" in the middle an equation. Fixing this here may introduce inconsistencies elsewhere in the notebook, so I prefer to leave it as it is.

doc/model_details/HillTononi.ipynb
heplesser: Done.

doc/model_details/HillTononi.ipynb
heplesser: Change that to a much simpler "..., because the total mini input rate".

heplesser:
@ingablundell Thank you! I have revised the notebook according to your suggestions, with one exception.

@apeyser Could you merge, as this is my PR?

PR #492, opened 2016-09-23
Erfc neuron model

tobikausk:
Add the neuron model "erfc-neuron" including the necessary tests. "erfc-neuron" is derived from the binary-neuron template and has got a complentary error function as activation function.
I propose @mhelias and @heplesser as reviewers.


heplesser:
@tobikausk Please fix the C++ and Python formatting errors.

jakobj:
seems like there are still formatting errors:
```
PEP8 analysis: (only first occurrence)
   ---------------------------------------
   | pynest/nest/tests/test_erfc_neuron.py:
   |   pynest/nest/tests/test_erfc_neuron.py:107:21: E128 continuation line under-indented for visual indent
   |   pynest/nest/tests/test_erfc_neuron.py:108:21: E122 continuation line missing indentation or outdented
```
it's possible that you have to fix them by hand if `auto-pep8` does not catch them.

DimitriPlotnikov:
Hello, 

could you, please, resolve conflicts?

tobikausk:
We fixed all conflicts. Please have another look on what we did.

nestkernel/nest_names.h
```diff
@@ -418,6 +418,7 @@ extern const Name sdev;             //!< Used in pulsepacket_generator
 extern const Name send_buffer_size; //!< mpi-related
 extern const Name senders;          //!< Recorder parameter
 extern const Name shift_now_spikes; //!< Used by spike_generator
+extern const Name sigma;       //!< Inverse steepness of erfc_neuron
```
heplesser: The indentation is wrong here. Please fix it to make Travis (clang-format) happy!

pynest/nest/tests/test_erfc_neuron.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        simulates erfc neuron for different parameter sets and compares
+        activity to theoretical value.
+        """
+        for sigma in self.sigma:
+            for theta in self.theta:
+                self.build_and_connect_nodes(sigma, theta)
+                nest.Simulate(self.T)
+                mean_activity = np.round(
+                    get_mean_activity(self.detector, self.T), 5)
```
heplesser: Why do you round here? Since you use  `AssertAlmostEqual` below, that should not be necessary.

pynest/nest/tests/test_erfc_neuron.py
```diff
@@ -0,0 +1,129 @@
...
+                mean_activity_theory = np.round(
+                    activation_function_theory(sigma, theta), 5)
```
heplesser: See above.

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef ERFC_NEURON_H
+#define ERFC_NEURON_H
+
+// Includes from models:
+#include "binary_neuron.h"
+#include "binary_neuron_impl.h"
```
heplesser: Header files should not include `*_impl.h` files. If you need methods defined in and `*_impl.h` file, you need to move that code to `erfc_neuron.cpp`.

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+
+namespace nest
+{
+/* BeginDocumentation
+   Name: erfc_neuron - Binary stochastic neuron with complementary error
+   function as activation function.
+
+   Description:
+   The erfc_neuron is an implementation of a binary neuron that
+   is irregularly updated as Poisson time points. At each update
```
heplesser: as -> at

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+   point the total synaptic input h into the neuron is summed up,
+   passed through a gain function g whose output is interpreted as
+   the probability of the neuron to be in the active (1) state.
+
+   The gain function g used here is
+   g(h) = 0.5 * erfc (( h - theta_ ) / ( sqrt( 2. ) * sigma)).
```
heplesser: Put empty lines around the equation for better visibility.

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+   This corresponds to a McCulloch-Pitts neuron receiving additional
+   Gaussian noise with mean 0 and standard deviation sigma
```
heplesser: Fullstop after sigma.

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+   The time constant tau_m is defined as the mean
+   inter-update-interval that is drawn from an exponential
```
heplesser: Is the mean interval drawn from the distribution? Aren't the intervals drawn from a distribution with this mean?

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+   References:
+   [1] Iris Ginzburg, Haim Sompolinsky. Theory of correlations in stochastic
+   neural networks (1994).
+   PRE 50(4) p. 3171
+   [2] W. McCulloch und W. Pitts (1943). A logical calculus of the ideas
+   immanent in nervous
+   activity. Bulletin of Mathematical Biophysics, 5:115-133.
+   [3] Abigail Morrison, Markus Diesmann. Maintaining Causality in Discrete Time
+   Neuronal
+   Simulations.
```
heplesser: Bad line breaks.

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+   tau_m      double - Membrane time constant (mean inter-update-interval) (ms)
+   theta      double - threshold for sigmoidal activation function (mV)
+   sigma      double - 1/sqrt(2pi) x inverse of maximal slope (mV)
+
+   References:
+   [1] Iris Ginzburg, Haim Sompolinsky. Theory of correlations in stochastic
+   neural networks (1994).
+   PRE 50(4) p. 3171
+   [2] W. McCulloch und W. Pitts (1943). A logical calculus of the ideas
+   immanent in nervous
```
heplesser: Bad line break.

models/erfc_neuron.h
```diff
@@ -0,0 +1,142 @@
...
+
+
+   Parameters:
+   tau_m      double - Membrane time constant (mean inter-update-interval) (ms)
+   theta      double - threshold for sigmoidal activation function (mV)
+   sigma      double - 1/sqrt(2pi) x inverse of maximal slope (mV)
+
+   References:
+   [1] Iris Ginzburg, Haim Sompolinsky. Theory of correlations in stochastic
+   neural networks (1994).
```
heplesser: Bad line break.

models/erfc_neuron.h
tobikausk: Right, that was unclear.

heplesser:
@tobikausk There are still a number of clang-format errors and the problem with the `_impl.h` file being included in a header file. If the latter should be necessary, please explain why.

tobikausk:
@heplesser @jakobj 
Thanks a lot for the very detailed commentary!
Concerning the question of the _impl.h-files, however, Jakob and I concluded that they have to be included in the erfc_neuron.h-file as it includes function definitions of the binary neuron template, which the linker otherwise misses. Therefore, we would keep this line, analogous to the case of the Ginzburg and the McCulloch-Pitts neuron.

heplesser:
@tobikausk "Umgekehrt wird ein Schuh daraus", as the German proverb says: Since `binary_neuron_impl.h` is always included together with `binary_neuron.h`, because the `impl` file provides essential definitions of templated methods, there should be no `binary_neuron_impl.h` file: all the code in that file should be included in `binary_neuron.h`. In this way we avoid the rather confusing `impl` file. 

Could you also merge the most recent changes from master and resolve the conflicts mentioned above?

jakobj:
@heplesser: we have reached exactly the same conclusion (that there should be no `_impl.h` file), however changing this would go well beyond the current PR, hence a reasonable strategy is to provide the new model in the same style as the old models and create a separate PR addressing the issue of the `_impl.h`.
The (questionable) reasoning behind the structure of separate `.h` and `_impl.h` files seems to be in this case the separation of `inline` functions and regular functions, tracing back before the move to git:
```bash
$git blame models/binary_neuron_impl.h | head -n1
^c229212d (Tiziano Zito    2015-05-11 17:11:02 +0200   1) /*
```
I agree that this should be fixed soon.

heplesser:
@jakobj @tobikausk  I agree with you that removing the `impl` file is a separate matter and should not bloat this PR. At the same time, I think it would be good to fix the impl-issue first, instead of adding new code with the impl.  

Eliminating the impl file is just a matter of copying the code in the impl file to the header file and removing existing `#include "binary_neuron_impl.h` files, plus updating `models/CMakeLists.txt`.  Could you create a PR for that? 

I would preferably merge that first, then update this PR correspondingly, so we avoid introducing new code which we know needs changing.

heplesser:
Thanks for #813! Merging this PR should wait until #813 is merged into this PR.

heplesser:
#813 is merged to master, so updating this PR with all the newest changes from Master should get this PR ready for merging, too.

tobikausk:
@jakobj @heplesser @DimitriPlotnikov 
Thank you very much for your help and for pushing me forward... I hope that the result is fine now.

heplesser:
@tobikausk Could you fix the code formatting problems that make Travis fail?

tobikausk:
@heplesser Fixed formatting issues, together with @jakobj .

models/erfc_neuron.h
```diff
@@ -0,0 +1,139 @@
...
+   Remarks:
+   This neuron has a special use for spike events to convey the
+   binary state of the neuron to the target. The neuron model
+   only sends a spike if a transition of its state occurs. If the
+   state makes an up-transition it sends a spike with multiplicity 2,
+   if a down transition occurs, it sends a spike with multiplicity 1.
+   The decoding scheme relies on the feature that spikes with multiplicity
+   larger 1 are delivered consecutively, also in a parallel setting.
+   The creation of double connections between binary neurons will
+   destroy the deconding scheme, as this effectively duplicates
```
hakonsbm: decoding, not "deconding"

models/erfc_neuron.h
```diff
@@ -0,0 +1,139 @@
...
+  /** sets default parameters */
+
+  gainfunction_erfc()
+    : theta_( 0.0 )
+    , sigma_( 1.0 )
+  {
+  }
+
+  void get( DictionaryDatum& ) const; //!< Store current values in dictionary
+  void set( const DictionaryDatum& ); //!< Set values from dicitonary
```
hakonsbm: dictionary, not "dicitonary"

jougs:
@tobikausk: could you please fix the typos pointed out by @hakonsbm, so we can merge this? Thanks!

jougs:
I really wonder if we shouldn't just reformat the whole `nest_names.h` file in the format
```C++
/** Description for variable */
extern const Name variable;
```

Using the current block format it seems to be impossible to arrive at a stable format of that file that
1. `clang-format` is not reformatting again and again on each call and
2. is still below (any reasonable) line-length limit

IMO this would best be done in another PR and merged to this one.

heplesser:
@jougs I very much like your suggestion! Could you implement it quickly? I guess it could be scripted ...

jougs:
@heplesser, please see #909 for a PR that fixes the problem with `Name`s and their comments. As most of the comments were actually wrong or outdated, I decided to remove them altogether instead of reformatting. I think that this is better, as many names are used in different places and their meaning should thus rather be defined and documented in the model.

jougs:
@tobikausk: can you please merge master? That should solve the failing checks and we could merge this.

Issue #332, opened 2016-05-03
current generating devices cannot be connected to binary neurons


tobikausk:
The ac-generator (and probably also dc- and step-current-generator) cannot be connected to binary neurons. This would be a useful feature (as you sometimes want to create "time varying thresholds") and would not create any inconsistencies.
The problem can be fixed rather easily and I will open a PR soon.

heplesser:
@mhelias What do you think about this proposal? Binary neurons are quite different from everything else, so interpreting "current" input as a signal that represents a time-varying threshold is probably something one could do. But documentation would have to be very clear.

mhelias:
Tobias and I discussed this. Actually, the connection of ac and dc denerators to binary neurons only fails since we implemented a stricter checking for compatibility of spiking and binary models. In previous revisions this has been possible and some of our published articles rely on this feature. The only thing to implement is the function send_signal() in all generators that should be compatible. The implementation of the binary neurons is already prepared to handle dc_input as an additive signal to the total input (thus equivalent to varying threshold). Documentation of the feature needs to be checked.

heplesser:
Closing due to inactivity.

PR #128, opened 2015-10-19
Issue #125: handshake binary / non-binary / documentation binary

mhelias:
Hi,

I implemented the check to prohibit connections between binary and non-binary models
and updated the documentation of binary neurons to better explain how and why the decoding
works.

Best wishes

Moritz

PR #496, opened 2016-09-27
ac_generator for all neuron models

tobikausk:
Change ac-generator such that it can send signals to neurons of arbitrary type. This is the pull request that I promised to open soon in May (issue #332)... The documentation does not have to be changes, I guess, because it does not specify the neuron type anyway. @mhelias @jschuecker


models/ac_generator.h
```diff
@@ -77,6 +77,13 @@ class ac_generator : public Node
 
   void get_status( DictionaryDatum& ) const;
   void set_status( const DictionaryDatum& );
+  using Node::sends_signal;
+
+  SignalType
+  sends_signal() const
+  {
+    return ALL;
+  }
```
heplesser: @tobikausk This works, but for the wrong reasons. For semantic consistency, `sends_signal()` has to return a value that represents the type of signal sent by a node. For the `ac_generator`, that would be `CONTINUOUS` or, to be even more precise, `STEPWISE_CONSTANT`. Thus, we should extend the `SignalType` enumeration in [`nest_types.h`](https://github.com/nest/nest-simulator/blob/48ca1d7090d8e0953c8e752425407a4c3b52d4b1/nestkernel/nest_types.h#L198) and adjust code correspondingly. `ALL` should only be used on the receiver side: A node may be able to handle any kind of input, but it will never send more than one type of signal.

heplesser:
@tobikausk There has been little activity on this pull request for a while. Are you still planning to fix the issues raised so that we can integrate it into NEST? Please let me know if you need any assistance!

tobikausk:
@heplesser Yes, I assume that I will have time to this in the next days. Being not totally sure what to do, I think that @mhelias can answer some questions. Thanks for the offer in any case!

models/ac_generator.h
jougs: @tobikausk: please change `ALL` to the new constant `STEPWISE_CONSTANT` as @heplesser pointed out.

models/binary_neuron.h
```diff
@@ -273,7 +273,7 @@ template < class TGainfunction >
 inline SignalType
 binary_neuron< TGainfunction >::receives_signal() const
 {
-  return BINARY;
+  return BINARY | STEPWISE_CONSTANT;
```
heplesser: @tobikausk This does not compile because the bitwise or produces an integer that cannot be converted into an enumeration value of type `SignalType`.  The corresponding operation works in the definition of the enumeration, because there an integer is expected on the right hand side of `=`.

@jougs What would be the best way to fix this? `bitset`?

jougs:
@tobikausk, I have a general request first: please always test (i.e. `make`, `make install`, `make installcheck`) your code locally before pushing. Any creation of/change to a pull request generates mails to the people subscribed, which might have other things to do than looking at failing builds.

----

Regarding [@heplesser's comment](https://github.com/nest/nest-simulator/pull/496#pullrequestreview-74594905): the cleanest solution for this is to add an overloaded `operator|` for the `SignalType` enum as I did in 7c2671c5fb809e7c962812121809ba37dafe0d54. However, even with the code compiling, Travis is still unhappy.

The reason for the continued failure has not so much to do with this PR, but rather with the fact that the `SignalType` mechanism added in #128 was merely meant to distinguish between `SPIKE` and `BINARY`, which are both transmitted through a `SpikeEvent` and not so much for distinguishing all kinds of events used in NEST (the [connection handshake](https://www.frontiersin.org/files/Articles/105658/fninf-08-00078-HTML/image_m/fninf-08-00078-g005.jpg) was actually meant for this).

I thus very much doubt that the current solution is a workable one as the obvious and simple fix (i.e. changing the default implementation of `sends_signal()` in the `Node` base class to also include `STEPWISE_CONSTANT`) is bound to fail sooner or later when people add more information to the enum. Adapting each and every model in NEST also seems wrong, as we already have the handshake mechanism.

@mhelias: Can you please (probably again) elaborate on why binary neurons also use `SpikeEvent` instead of a dedicated `BinaryEvent`? Wouldn't it be easier to have that and adapt the recorders accordingly? Many thanks!

mhelias:
Dear all,
the changes implemented by Tobias are those requested by HEP: He extended the signal enum type to include stepwise_constant and adapted the handshaking mechanism to check for compatibility of the sent and received signal.
The reason why binary neurons do not sent binary events is simple: Such an event type would have
to carry an additional information, namely 0 or 1. We could implement this.
The consequences are as follows:
- we need an additional send and receive buffer for binary events
- we need additional communication of these events
- we need to distinguish binary events in the entire delivery mechanism and synapse storage,
  becaue you should not deliver a binary event to a spiking neuron etc etc. 
-> in principle, the same suplication of work as we needed to introduce for rate neurons / gap junctions.

It is clear that the performance of all simulations will therefore suffer from the mere presence of binary neurons.
Since the main objective if NEST and it strength is in spiking models, I think this would be the
wrong way to go, if the only objective if semantic beauty.

The solution as it is right now conveys the binary information by sending two spike events for
an up-transition and one for down. This is a choice that has no side effects to the spiking
code besides the semantic problem that a connection between a spiking model and a binary
can be made, but does, of course, have no sense.
To prevent users from performing such potentally meaningless simulations, we implemented
the signal type, which is checked upon connection to make sure that one cannot connect
a spiking neuron to a binary neuron (even though my personal opinion is that performing
such a meaningless simulation would be an adequate punishment to such users).

Jochen's remark is of course well justified: we actually already have the handshake mechanism
(his attached jpg) to tell apart different event types. So we would not have to check for compatibility
of DC_current_events etc separately.

In my feeling the cleanest and most efficient solution would be like this:
- redefine the semantics of spike_event to point_event in time
- this redefinitions makes the use legitimate to indicate transitions of binary neurons
- since by this re-definition, the point_event may have several meanings
  (e.g. appearance of a spike or appearance of a transition of a binary variable)
  it would be cleanest if these different meanings would be a property of the event
  (the enum type which is now outside of the spike_event)
- in the standard handshake mechanism we anyway pass around these events
- only at this point we would need to check for this additional field and see if it is compatible
 with what the sender/receiver sets it to

One would need to check if the increase of the object size of event woud have any drawback on the
performance. If this is so, my proposed solution should be discarded and we should go with what
HEP proposed and Tobias implemented.

tobikausk:
@jougs Sorry for the spam, I did not intend to produce any extra work for anyone - I just wanted to document that I did something. Now I see that a commit is not the right form to do this.

jougs:
@mhelias: many thanks for your explanations and your implementation idea. I think this is sound and my gut feeling is that this should not be a bigger performance hit than the current solution might be.

@tobikausk: do you think you can implement the solution proposed by @mhelias? If yes, can you please do the work within your current development branch, so this PR updates accordingly?

tobikausk:
@jougs Together with @mhelias I could do it, I guess, but because my programming skills are very limited, as you know, I will not make it alone, I fear.

jougs:
I think we all agree that this pull request does not solve the problem adequately and I will thus close this pull request.

@tobikausk could you please start the work on a proper fix by transferring (and probably condensing) the information from this pull request to a new issue? Thanks!

PR #497, opened 2016-09-28
Device for recording weights of plastic synapses

weidel-p:
Hi,

with this PR I'd like to add a device (`weight_recorder`)  to NEST allowing the user to record the weight-changes of plastic synapses during the simulation.

Until now, it was necessary to stop the simulation, read the synaptic weights using `GetConnections` and `GetStatus` and resume the simulation  in order to record the changes of the synaptic strengths during a simulation. With the `weight_recorder` the changes of synaptic strengths are recorded on the fly and without the need to stop the simulation.

To achieve a consistent user experience and for minimizing the memory overhead, the usage of the `weight_recorder` is based on the `volume_transmitter`; the `weight_recorder` is registered to a synapse in the `CommonSynapsePorperties` using `SetDefaults` or `CopyModel`.

Here is an example pyNEST script using the `weight_recorder` with the `stdp_synapse`:

```
import nest

wr = nest.Create('weight_recorder', 1, {"to_file": True})
nest.CopyModel("stdp_synapse", "stdp_synapse_rec", {"weight_recorder": wr[0]})

sg = nest.Create("spike_generator", params={"spike_times": [10., 15., 55., 70.]})
pre = nest.Create("parrot_neuron", 1)
post = nest.Create("parrot_neuron", 1)

nest.Connect(pre, post, syn_spec="stdp_synapse_rec")
nest.Connect(sg, pre)

nest.Simulate(100)
```

Internally, the `weight_recorder` is based on the `spike_detector` but handles `WeightRecorderEvent` instead of `SpikeEvent` and can not be used as a global receiver. 
As a synapse has no `id` in NEST,  the `WeightRecorderEvent` has an additional field `receiver_gid` containing the gid of the postynaptic neuron, which allows the user to identify a synapse by the gid of the pre- and postsynaptic neuron. (This, of course, can not work for multapses)

Here is the output of the `weight_recorder` running the example script above:
The columns containing the `sender_gid`, `receiver_gid`, `time`, `weight` in this order.
```
3       4       11.000  1.000   
3       4       16.000  1.879   
3       4       56.000  3.450   
3       4       71.000  4.504
```

As mentioned above, the `weight_recorder` is registered to a synapse in the `CommonSynapseProperties`.  Also worth mentioning is that the `WeightRecorderEvent` is sent to the `weight_recorder` in `connector_base.h`. This implementation leads to an easy and general solution: every existing and future synapse can make use of the `weight_recorder` without changing the implementation or caring about it. The drawback of this solution is one additional `if` statement in the send function, which might slow down the simulation.

This PR also contains a pyNEST unittest, testing for consistent results of the weights recorded by the `weight_recorder` using a single and multiple threads.


pynest/nest/tests/test_weight_recorder.py
```diff
@@ -0,0 +1,105 @@
...
+
+@nest.check_stack
+class WeightRecorderTestCase(unittest.TestCase):
+    """Tests for the Weight Recorder"""
+
+    def testSingleThread(self):
+        """Weight Recorder Single Threaded"""
+
+        nest.ResetKernel()
+        nest.SetKernelStatus({"local_num_threads": 2})
```
heplesser: I thought this was `testSingleThread`?

pynest/nest/tests/test_weight_recorder.py
```diff
@@ -0,0 +1,105 @@
...
+
+        weights = [1.]
+        for i in range(100):
+            nest.Simulate(1)
+            weights.append(nest.GetStatus(connections, "weight")[0])
+
+        weights = np.round(weights, 6)
+        wr_weights = np.round(nest.GetStatus(wr, "events")[0]["weights"], 6)
+
+        self.assertTrue(all([w in weights for w in wr_weights]))
```
heplesser: I think unittest provides smarter asserts that compare collections elementwise within a given numerical precision. They provide more informative output on failure.

heplesser:
@weidel-p Great work! I suggest @jougs and @jakobj  as reviewers.

I guess plasticity could also change delays (at least in principle, I think we have no connections in NEST at the moment doing this), so I think it would be useful if the `weight_recorder` could also record delays.

jougs:
@abigailm: can you please comment on plasticity changing delays?

abigailm:
yes, I can. I think it's not necessary now, as we have no plasticity models that do this, and we would have to think carefully before adding them as the minimum synaptic delay is kind of important. I would recommend letting this PR through without this change, and adapted the device later, if such models are added.

abigailm:
The "advantage" of the GetConnections method is that you can select pre- and post-synaptic neurons in the query. So for example, I can connect a recurrent network randomly, and then record, say, all the outgoing connections from a selected pop1, or all the incoming connections to a selected pop2, or all the connections between pop1 and pop2. Is this functionality covered by this PR? It would be important for large scale simulations to be able to pick out specific groups to be recorded after the network has been wired.

weidel-p:
@abigailm: Recording the weight-changes of all synapses in a large scale simulation will lead to a large amount of data. To limit the amount of recorded data, the user can specify subpopulations using a `weight_recorder` before/during the wiring of the network using `CopyModel`. 

So in general this PR covers this aspect, but I agree that the workflow could be nicer. I would prefer to register a `weight_recorder` to a subset ob synapses after wiring the network using `SetStatus` on selected connections. This means that the `weight_recorder` would have to be registered in the properties of every synapse instead of in the `CommonSynapseProperties` resulting in an increased memory overhead. I don't know how bad this would be for HPC applications, but I didn't want to limit the HPC capacity of NEST with this device. 

I am open for suggestions to improve the implementation. Maybe I missed a smart way to implement a nicer workflow without increasing the memory overhead.

heplesser:
@jougs @abigailm You are right, plastic delays would raise a number of questions. So let's proceed with this PR recording weights only.

abigailm:
@weidel-p I see how that would work in many instances, but I think it is problematic. Often the user will not be able to determine pre-connection the synapses that he or she wants to record. For example, I connect a network with fixed indegree, and then I want to record all outgoing connections of a particular sub-population. There would be no way of labelling those particular ones ahead of time with a `CopyModel`, because they were selected randomly. Would it be possible to supplement the `weight_recorder` with properties identifying the pre- and post-synaptic populations, enabling it to do filtering of results on the fly? The default behaviour would be to record all synapses it is attached to, but the user could specify those variables if needed.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+void
+nest::weight_recorder::update( Time const&, const long from, const long to )
+{
+
+  for ( std::vector< WeightRecorderEvent >::iterator e = B_.events_.begin();
+        e != B_.events_.end();
+        ++e )
+  {
```
abigailm: So, for example, couldn't we test here whether the sender is one of the desired presynaptic population (if specified), ditto receiver, and only record it if it was?

models/weight_recorder.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+void
+nest::weight_recorder::handle( WeightRecorderEvent& e )
+{
+  // accept spikes only if detector was active when spike was
+  // emitted
+  if ( device_.is_active( e.get_stamp() ) )
+  {
+    WeightRecorderEvent* event = e.clone();
+    B_.events_.push_back( *event );
```
abigailm: It might be better to check for sender/receiver here, to avoid events_ getting swamped in a large simulation. I suggest testing for performance in a large network.

heplesser:
I like @abigailm 's suggestions about selecting by senders/targets. An interesting questions is the most efficient way to implement this. From @abigailm 's use-case description, I assume that the senders/targets one would want to select are not contiguous ranges of GIDs. Thus, STL `set` is probably the most useful data structure to keep the "wanted" nodes and check membership. But given that the set of senders/targets to select from does not change over time (or only at large intervals when the user reconfigures the recorder), we could also put them in a sorted `vector` and then do a binary search, maybe even with linear interpolation to determine the starting point (see `SparseNodeArray`). 

@abigailm How many senders and targets would you typically select, just a ballpark number?

abigailm:
Another advantage of the `GetConnections` method is that it allows the weights to be sampled with a lower rate, say every second, which massively reduces the file size on long simulations of large networks. I recommend that this not be a requirement for the current PR, but to open it as a ticket for a later enhancement of the `weight_recorder`. In the mean time, those who need the down-sampling can still use the `GetConnections` approach as before.

abigailm:
@heplesser For example, Kunkel et al. (2011) records weights from ~100s of neurons for ~100s of seconds

models/weight_recorder.cpp
```diff
@@ -155,16 +158,73 @@ nest::weight_recorder::set_status( const DictionaryDatum& d )
...
   // accept spikes only if detector was active when spike was
   // emitted
   if ( device_.is_active( e.get_stamp() ) )
   {
+
+    // P_pre_ is defined and sender is not in it
+    if ( !P_.pre_.empty()
+      && std::find( P_.pre_.begin(), P_.pre_.end(), e.get_sender_gid() )
+        == P_.pre_.end() )
+      return;
```
heplesser: @weidel-p  I would suggest that you combine the pre and post tests. I think that will make the code more readable. Finally, we decided to write out logical operators in new NEST C++ code as in Python:
```C++
if ( ( not P_.pre_.empty() and std::find(...) == P_.pre_.end() )
      or ( not P_.post_.empty() and std::find(...) == P_.post_.end() ) ) 
```

models/weight_recorder.cpp
```diff
@@ -155,16 +158,73 @@ nest::weight_recorder::set_status( const DictionaryDatum& d )
...
+  : pre_( p.pre_ )
+  , post_( p.post_ )
+{
+}
+
+void
+nest::weight_recorder::Parameters_::get( DictionaryDatum& d ) const
+{
+  //( *d )[ "pre" ] = pre_;
+  //( *d )[ "post" ] = post_;
```
heplesser: Please remove the commented lines.

models/weight_recorder.cpp
```diff
@@ -155,16 +158,73 @@ nest::weight_recorder::set_status( const DictionaryDatum& d )
...
+  ( *d )[ "pre" ] = std::vector< long >( pre_.begin(), pre_.end() );
+  ( *d )[ "post" ] = std::vector< long >( post_.begin(), post_.end() );
```
heplesser: See `Parameters_::set()` below for comments on pre/post names.

I have a design question here, and maybe @abigailm and @jougs could comment: How large would pre/post be, and how likely would they contain contiguous ranges of GIDs or isolated GIDs? That would influence the question of whether pre/post should be passed as GID vectors or as GIDCollections.

models/weight_recorder.cpp
```diff
@@ -155,16 +158,73 @@ nest::weight_recorder::set_status( const DictionaryDatum& d )
...
+    std::vector< long > tmp =
+      getValue< std::vector< long > >( d->lookup( "pre" ) );
+    pre_ = std::set< long >( tmp.begin(), tmp.end() );
+  }
+  if ( d->known( "post" ) )
+  {
+    std::vector< long > tmp =
+      getValue< std::vector< long > >( d->lookup( "post" ) );
+    post_ = std::set< long >( tmp.begin(), tmp.end() );
+  }
```
heplesser: Please add `"pre"` and `"post"` to NEST names. And are pre and post really good names here? In all connect routines we use "source" and "target", so that might be better to use here, too, to minimize user confusion.

See  `Parameters_::get()` for comments on vector vs GidCollection.

models/weight_recorder.h
```diff
@@ -131,7 +131,19 @@ class weight_recorder : public Node
   bool user_set_precise_times_;
   bool has_proxies_;
   bool local_receiver_;
-  omp_lock_t writelock;
+
+  struct Parameters_
+  {
+    std::set< long > pre_;
+    std::set< long > post_;
```
heplesser: I think `set` is a good first choice, we can later consider if we need something even more efficient. Note that we usually use the alias `index` for GIDs, I'd suggest using that here as well.

models/weight_recorder.cpp
```diff
@@ -24,6 +24,7 @@
 
 // C++ includes:
 #include <numeric>
+#include <typeinfo>
```
heplesser: Do you need the `typeinfo` header for anything?

models/weight_recorder.cpp
weidel-p: As suggested, I changed the names to `"source"` and `"target"`.

models/weight_recorder.h
weidel-p: I agree, but actually I had an issue with Travis and `std::set`. So I picked up your earlier idea and changed these parameters to sorted `vector` and use a binary search to access elements. Fortunately, these two fields will only be set once (or a few times) in the build phase, so the `vector` has not to be resorted during the simulation.

Overall, I think this solution is better than a `set` as the complexity to access elements is also O(log N) and I think a `vector` needs less memory.

weidel-p:
@heplesser and @abigailm: thanks for your comments and help. I tried to implement most suggestions in f6f7f23. I did not run any performance tests, yet.

The idea of having a sampling rate is probably very important for many use-cases and I will try to implement that asap. Another minor change which came up in a discussion with @jakobj is the possibility to record the `port` of a synapse to identify unique synapses even if there are `multapses`. 
I would prefer to add these two features to this PR before merging.

models/weight_recorder.cpp
```diff
@@ -155,16 +157,70 @@ nest::weight_recorder::set_status( const DictionaryDatum& d )
...
+  : sources_( p.sources_ )
+  , targets_( p.targets_ )
+{
+}
+
+void
+nest::weight_recorder::Parameters_::get( DictionaryDatum& d ) const
+{
+  ( *d )[ nest::names::source ] = sources_;
+  ( *d )[ nest::names::target ] = targets_;
```
heplesser: Since you are inside namespace `nest` here, you do not need the `nest::` qualifiers. Applies also in `set()`.

abigailm:
@heplesser the sampling rate is based on time rather than numbers of spikes. So the simulation is run for time t, we query the synapses as to what was their weight the last time they produced a spike, simulate for time t, query the synapses, etc. No maintenance of spike count required.

heplesser:
@abigailm That works indeed very nicely with the stop/GetStatus/resume approach, but could it be carried over to continuous recording?

--

Dr Hans Ekkehard Plesser


On 1 Oct 2016, at 15:38, Abigail Morrison <notifications@github.com<mailto:notifications@github.com>> wrote:


@heplesser<https://github.com/heplesser> the sampling rate is based on time rather than numbers of spikes. So the simulation is run for time t, we query the synapses as to what was their weight the last time they produced a spike, simulate for time t, query the synapses, etc. No maintenance of spike count required.

-

abigailm:
@heplesser Well I don't know for sure, which is why I would be ok with Philipp submitting this PR as it is and treating that as a separate issue (if he wants to handle it like that). I think it would be compatible, but the mechanism of sending the synaptic information to the `weight_recorder` would have to be changed, I guess by switching to a polling approach like `Multimeter`.

heplesser:
@abigailm @weidel-p I would suggest that we leave the sampling rate out of this PR and first collect some experience with this recorder in practice to see how it performs.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,226 @@
...
+  , user_set_precise_times_( n.user_set_precise_times_ )
+  , has_proxies_( false )
+  , local_receiver_( true )
+{
+}
+
+void
+nest::weight_recorder::init_state_( const Node& np )
+{
+  const weight_recorder& wr = dynamic_cast< const weight_recorder& >( np );
```
jakobj: this should be a `static_cast`.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+  ( *d )[ nest::names::source ] = sources_;
+  ( *d )[ nest::names::target ] = targets_;
+}
+
+
+void
+nest::weight_recorder::Parameters_::set( const DictionaryDatum& d )
+{
+
```
jakobj: you should allow the user to specify the index of the connection (the last entry in every element returned by `GetConnections`) or record the index as well. otherwise weights from multapses are not distinguishable.

nestkernel/recording_device.cpp
```diff
@@ -290,6 +295,14 @@ nest::RecordingDevice::State_::get( DictionaryDatum& d,
       dict, names::weights, std::vector< double >( event_weights_ ) );
   }
 
+  if ( p.withreceivergid_ )
```
jakobj: i was wondering, what's the use case of not recording the receiver gid? don't you always need at least the tuple (sender, receiver) to make sense out of the recorded data? or is this just signalling that this is a `weight_recorder`? in that case, maybe you could use a more explicit variable name?

jakobj:
very nice work @weidel-p. i'll let you know as soon as i have same results from benchmarks.

nestkernel/connector_base.h
```diff
@@ -434,14 +462,14 @@ class Connector : public vector_like< ConnectionT >
...
     for ( size_t i = 0; i < K; i++ )
     {
       e.set_port( i );
-      C_[ i ].send( e,
-        t,
-        ConnectorBase::get_t_lastspike(),
-        static_cast< GenericConnectorModel< ConnectionT >* >( cm[ syn_id ] )
-          ->get_common_properties() );
+      C_[ i ].send( e, t, ConnectorBase::get_t_lastspike(), cp );
+      ConnectorBase::send_weight_event( cp, e, t );
```
mschmidt87: But the `weight_events` are only communicated locally to the `weight_recorder`, right? So, at least this should not increase MPI communication.

nestkernel/recording_device.cpp
weidel-p: There was actually a bug in the code. It was meant that only the `weight_recorder` is using `withreceivergid`. But actually, the user can set this parameter to every recording device (i.e. the `spike_detector`), which could lead to undefined behavior as `get_receiver_gid()` was only implemented in the `WeightRecorderEvent`.  I fixed that in 72c7fae, thanks.

weidel-p:
@heplesser @abigailm: instead of going to a polling approach I added two parameter to the `weight_recorder`: `interval` and `duration`. By setting `interval` and `duration`, only events with a `timestamp` within this `interval` and `duration` will be recorded. These parameters are of course optional, and not the perfect solution, but can help to reduce the amount of data.

@jakobj: thanks for your comments. I tried to implement them all.

mschmidt87:
Hi, I tried the new parameters `interval` and `duration` and don't understand how they are supposed to work. From the code, I understand that if I set `interval`, only spikes that are a multiple of `interval`, are recorded. I confirmed this by using your example code from the first post and setting `interval=16.`. Then, only the spike at 15. ms (+1ms delay) being written out, because it is the only multiple integer of 16.
Is that the desired behavior?

If I set `duration` to a finite value, e.g. 10. ms, then the spikes at 15., 55. and 75. ms are written out. 

Am I missing here something? If I understood @abigailm  correctly, then she is suggesting to implement a fixed time interval at which the current weight of the recorded synapses are written out.

weidel-p:
@mschmidt87: sorry, I should have explained that a bit more in detail. My idea was to record events only if they have a `timestamp` which are within a certain `duration` in a repeating `interval`. I.e. to record "spikes" if they happen in the first 100ms of each second in the simulation.

To make this example, assume to set `interval: 1000ms` and `duration: 100ms` and there are events happening at `[50 ms, 700 ms, 1010 ms, 1400 ms]` then, only the first and third event are recorded.

If I understood correctly, @abigailm and @heplesser both actually suggested to postpone the topic of sampling. But I think it is very important to have some functionality of down-sampling now (also for my own research), even if it is not the most popular solution. We can still improve the `weight_recorder` and the sampling method in a new issue after this PR.

mschmidt87:
Okay, I see. This sounds like a relatively simple and smart solution without having to change the implementation altogether. Although I am not sure how practicable this might be in the end, because the risk is to miss a lot of events just by chance if `interval` and `duration` are not carefully tuned.

In any case, I would suggest to rename the parameters from `interval` and `duration` to `interval_step` and `interval_length`. In addition, this feature should be documented in the doc string of the `weight_recorder`.

heplesser:
@weidel-p Your interval idea is interesting in principle, but I would still prefer to keep it out of this PR for reasons I will explain below. Note that any form of windowing or sub-sampling will be an addition to thw recording itself, so it can easily be added in a later PR (and you can even use it for your own work as is till then).

My main reason for asking you to leave the windowing out is that we should take a little more time to think through approaches, syntax and semantics before committing to a solution. Windowing will probably be useful for many types of recordings, so we should aim for a more general solution. It also needs to be compatible (reasonably) with existing solutions to ensure a consistent user experience across recorders. I opened #505 for suggestions and further discussion on this.

Concerning weight recording, I am also wondering a little whether windowing really is the best approach, since weights are only observable when a spike passes through a synapse. Could it be an idea to simply use random sub-sampling of weights: spikes arriving at the weight recorder are stored only with a fixed, given probability (usually quite small). I think it would depend on the scientific questions investigated whether this would make more or less sense than windowing.

nestkernel/connector_base.h
jakobj: yes, you're right. they do not contribute to MPI communication, but adds an additional `if` for every spike event that is delivered. turns out it's not a performance issue (see below).

jakobj:
here are some results from a weak scaling benchmark on JQ, comparing master and this branch. turns out the additional `if` in `send` has basically no influence on performance -- apparently branch prediction works well. 
juqueen_weak_scaling_weight_recorder_master
this also means that I am happy with this PR now, so :+1:, conditional on @heplesser concerns being addressed.

mschmidt87:
Wouldn't it be interesting to check the performance of actually recording weights vs. no recording? Perhaps not on the supercomputer, but on the cluster for some configurations? Or did you already do that, @weidel-p ? Perhaps there is some performance issue that we don't see from the code itself?

heplesser:
@mschmidt87 I think the benchmark by @jakobj shows that there is no cost when not recording, and that is the most important thing. Recording will obviously cost, but I don't think we need to benchmark that explicitly at this point.

@weidel-p Could you merge the latest changes from master to solve the merge conflicts that have popped up.

weidel-p:
@heplesser: in bc04f15 I removed the sampling parameters and in 522e3fa I merged master. Are all issues addressed now, or did I forget about something?

nestkernel/common_synapse_properties.cpp
```diff
@@ -47,13 +49,31 @@ CommonSynapseProperties::~CommonSynapseProperties()
 }
 
 void
-CommonSynapseProperties::get_status( DictionaryDatum& ) const
+CommonSynapseProperties::get_status( DictionaryDatum& d ) const
 {
+  if ( weight_recorders_.size() != 0 )
```
heplesser: For vectors, better use `not w_r_.empty()`. Once this becomes an SC pointer, you can turn it into
```C++
def< long >( d, names::weight_recorder, weight_recorder_ ? weight_recorder->get_gid() : -1 );
```
Note: Please add weight_recorder to `nest::names`.

nestkernel/common_synapse_properties.cpp
```diff
@@ -47,13 +49,31 @@ CommonSynapseProperties::~CommonSynapseProperties()
...
+  long wrgid;
+  if ( updateValue< long >( d, "weight_recorder", wrgid ) )
+  {
+    weight_recorders_ =
+      std::vector< Node* >( kernel().vp_manager.get_num_threads() );
+    for ( thread tid = 0; tid < weight_recorders_.size(); ++tid )
+    {
+      weight_recorders_[ tid ] = kernel().node_manager.get_node( wrgid, tid );
+    }
+  }
```
heplesser: This will simplify to something like
```C++
weight_recorder_ = kernel().node_manager.get_thread_siblings( wrgid );
```

nestkernel/connector_base.h
```diff
@@ -43,6 +43,7 @@
 #include "nest_names.h"
 #include "node.h"
 #include "spikecounter.h"
+#include "common_synapse_properties.h"
```
heplesser: Sort includes alphabetically.

nestkernel/connector_base.h
```diff
@@ -196,10 +197,39 @@ class ConnectorBase
     t_lastspike_ = t_lastspike;
   }
 
+  void send_weight_event( const CommonSynapseProperties& cp,
+    const Event& e,
+    const thread t );
```
heplesser: Move up to behind `send()` for better readability.

nestkernel/connector_base.h
```diff
@@ -196,10 +197,39 @@ class ConnectorBase
...
+    WeightRecorderEvent wr_e;
+    wr_e.set_port( e.get_port() );
+    wr_e.set_rport( e.get_rport() );
+    wr_e.set_stamp( e.get_stamp() );
+    wr_e.set_sender( e.get_sender() );
+    wr_e.set_sender_gid( e.get_sender_gid() );
+    wr_e.set_weight( e.get_weight() );
+    wr_e.set_delay( e.get_delay() );
+    wr_e.set_receiver(
+      *cp.weight_recorders_[ t ] ); // receiver is the weight recorder
```
heplesser: With `SiblingContainer`, this will turn into `wr_e.set_receiver( *( cp.weight_recorder->get_thread_sibling( t ) ) )`.

nestkernel/connector_base.h
```diff
@@ -196,10 +197,39 @@ class ConnectorBase
...
+    wr_e.set_receiver_gid(
+      e.get_receiver()
+        .get_gid() ); // the gid of the postsynapic neuron is to be recorded
```
heplesser: I think this would be more legible with the comment on a separate line, followed by the code that would then fit on one line.

nestkernel/event.h
```diff
@@ -363,6 +369,55 @@ SpikeEvent::get_multiplicity() const
...
+
+  /**
+   * Return GID of receiving Node.
+   */
+  index get_receiver_gid() const;
+
+  /**
+   * Change GID of receiving Node.
+   */
+
```
heplesser: No empty line here.

nestkernel/nest_names.cpp
```diff
@@ -176,6 +176,7 @@ const Name inh_conductance( "inh_conductance" );
 const Name input_currents_ex( "input_currents_ex" );
 const Name input_currents_in( "input_currents_in" );
 const Name Interpol_Order( "Interpol_Order" );
+const Name duration( "duration" );
```
heplesser: I think `duration` will not be needed for now.

nestkernel/nest_names.cpp
```diff
@@ -257,6 +259,7 @@ const Name S( "S" );
 const Name scientific( "scientific" );
 const Name screen( "screen" );
 const Name senders( "senders" );
+const Name receivers( "receivers" );
```
heplesser: Do we really want to introduce "receivers"? Wouldn't "targets" be much more in line with terminology we have in NEST otherwise? These are terms the user will see.

nestkernel/recording_device.h
```diff
@@ -411,9 +417,24 @@ class RecordingDevice : public Device
   void print_weight_( std::ostream&, double );
 
   /**
+   * Print the receiver gid of an event.
+   */
+  void print_receiver_( std::ostream&, index );
+
+  /**
+   * Print the receiver gid of an event.
```
heplesser: Incorrect comment.

nestkernel/recording_device.h
```diff
@@ -411,9 +417,24 @@ class RecordingDevice : public Device
...
+   */
+  void print_port_( std::ostream&, long );
+
+  /**
+   * Print the receiver gid of an event.
```
heplesser: Incorrect comment.

nestkernel/recording_device.h
```diff
@@ -411,9 +417,24 @@ class RecordingDevice : public Device
...
+   */
+  void print_rport_( std::ostream&, long );
+
+  /**
    * Store data in internal structure.
    */
-  void store_data_( index, const Time&, double, double );
+  void store_data_( index, const Time&, double, double, index, long, long );
```
heplesser: Could you add parameter info to the docstring above?

nestkernel/nest_names.cpp
```diff
@@ -364,6 +367,9 @@ const Name withgid( "withgid" );
 const Name withpath( "withpath" );
 const Name withtime( "withtime" );
 const Name withweight( "withweight" );
+const Name withreceivergid( "withreceivergid" );
```
heplesser: Would not `withtarget` be a much better name (also for pertaining variable names in the code)?

nestkernel/recording_device.h
```diff
@@ -447,13 +468,16 @@ class RecordingDevice : public Device
...
-    bool withtime_;       //!< true if time of event is to be printed, default
-    bool withweight_;     //!< true if weight of event is to be printed
+    bool to_accumulator_;  //!< true if data is to be accumulated; exclusive to
+                           //!< all other to_*
+    bool time_in_steps_;   //!< true if time is printed in steps, not ms.
+    bool precise_times_;   //!< true if time is computed including offset
+    bool withgid_;         //!< true if element GID is to be printed, default
+    bool withtime_;        //!< true if time of event is to be printed, default
+    bool withweight_;      //!< true if weight of event is to be printed
+    bool withreceivergid_; //!< true if receiver GID is to be printed, default
```
heplesser: I'd suggest `withtarget_`.

nestkernel/recording_device.h
```diff
@@ -447,13 +468,16 @@ class RecordingDevice : public Device
...
+    bool withport_;        //!< true if receiver GID is to be printed, default
```
heplesser: Bad comment.

nestkernel/recording_device.h
```diff
@@ -447,13 +468,16 @@ class RecordingDevice : public Device
...
+    bool withrport_;       //!< true if receiver GID is to be printed, default
```
heplesser: Bad comment.

nestkernel/recording_device.h
```diff
@@ -494,6 +522,9 @@ class RecordingDevice : public Device
   {
     size_t events_;                         //!< Event counter
     std::vector< long > event_senders_;     //!< List of event sender ids
+    std::vector< long > event_receivers_;   //!< List of event receivers ids
```
heplesser: targets

nestkernel/recording_device.cpp
```diff
@@ -595,12 +642,30 @@ nest::RecordingDevice::record_event( const Event& event, bool endrecord )
...
+      dynamic_cast< const WeightRecorderEvent* >( &event );
+    if ( wr_e != 0 )
+    {
+      receiver = wr_e->get_receiver_gid();
+    }
+    else
+    {
+      receiver = event.get_receiver_gid();
+    }
+  }
```
heplesser: I think you are manually doing run-type type resolution here. `e.get_receiver_gid()` should be all that is needed. So you could do
```C++
const long receiver = P_.withtarget_ ? e.get_receiver_gid() : -1;
```
(if it is at all necessary to use the dummy -1, why not pass the target GID in any case?)

models/modelsmodule.cpp
```diff
@@ -91,6 +91,7 @@
 #include "correlospinmatrix_detector.h"
 #include "multimeter.h"
 #include "spike_detector.h"
+#include "weight_recorder.h"
```
heplesser: Please keep things alphabetical.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,223 @@
...
+      "of the %1 with gid %2 has been set to true",
+      get_name(),
+      get_gid() );
+
+    if ( device_.is_precision_user_set() )
+    {
+      // if user explicitly set the precision, there is no need to do anything.
+      msg += ".";
+    }
+
```
heplesser: Remove empty line.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,223 @@
...
+    else
+    {
+      // it makes sense to increase the precision if precise models are used.
+      device_.set_precision( 15 );
+      msg += ", precision has been set to 15.";
+    }
+
+    LOG( M_INFO, "spike_detector::calibrate", msg );
```
heplesser: NB: `weight_detector`

models/weight_recorder.cpp
```diff
@@ -0,0 +1,223 @@
...
+                 P_.targets_.end(),
+                 e.get_receiver_gid() ) ) )
+      return;
+
+    WeightRecorderEvent* event = e.clone();
+    B_.events_.push_back( *event );
+  }
+}
+
+nest::weight_recorder::Parameters_::Parameters_()
```
heplesser: Please move `Parameters_` code forward so it is in the same location within the file as for other models.

models/weight_recorder.h
```diff
@@ -0,0 +1,182 @@
...
+
+Description:
+The weight_recorder device is a recording device. It is used to record
+weights from synapses. Data is recorded in memory or to file as for all
+RecordingDevices.
+By default, source GID, target GID, time and weight of each spike is recorded.
+
+The weight detector can also record weights with full precision
+from neurons emitting precisely timed spikes. Set /precise_times to
+achieve this.
```
heplesser: Please document sources and targets parameters.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,223 @@
...
+// Includes from sli:
+#include "arraydatum.h"
+#include "dict.h"
+#include "dictutils.h"
+#include "doubledatum.h"
+#include "integerdatum.h"
+
+nest::weight_recorder::weight_recorder()
+  : Node()
+  // record time, gid, weight and receiver gid
```
heplesser: Remove comment here, rather add one per line below.

models/weight_recorder.cpp
```diff
@@ -0,0 +1,223 @@
...
+  , device_( *this,
+      RecordingDevice::SPIKE_DETECTOR,
```
heplesser: `SPIKE_DETECTOR`?

models/weight_recorder.cpp
```diff
@@ -0,0 +1,223 @@
...
+      "csv",
```
heplesser: "csv"  is misleading since we separate by tab, not comma. I would suggest either "tsv" or "dat".

weidel-p:
@heplesser thanks for looking over the code again; it was clearly necessary. I will wait for your `SiblingContainer` PR to address your comments.

nestkernel/nest_names.cpp
weidel-p: As the parameter for the sender is called `withgid` I chose to call this parameter `withtargetgid`. I think this is more consistent.

nestkernel/recording_device.cpp
weidel-p: What you are suggesting is unfortunately not working, I need a `dynamic_cast` here.

Consider that other recording devices like the `spike_detector` also have the parameter `withtargetgid` now, and this function receives also `SpikeEvents` which does not have the field `receiver_gid`. To get a reasonable output for this case, I implemented `get_receiver_gid()` in `Event` which returns the `gid` of the receiver (unlike the `WeightRecorderEvent` which returns the `gid` of the post-synaptic neuron).

I also can't use a `static_cast` from one derived class to another (i.e. `SpikeEvent` to `WeightRecorderEvent`), as this results in undefined behavior.

models/weight_recorder.cpp
weidel-p: I would like to keep `csv` and in my opinion it is not misleading. It is very common to have a tab-separated csv file. Every csv reader I've seen so far (i.e. in Python, Excel, LibreOffice, etc) has the option to choose the delimiter. Often this is even done automatically.

The advantage of using a known file extension like 'csv' is that user immediately knows how to open the file.

weidel-p:
@heplesser thanks again for your last comments, I hope I addressed them all.

alberto-antonietti:
Hi @weidel-p, I am currently using the current version of weight_recorder. I saw that you mentioned the possibility of having a down-sampling (e.g. saving weights each second, useful for long simulations). Are you (or someone else) working on this or could I try to add this new functionality to the current version?

Thanks!
Alberto

weidel-p:
Hi @alberto-antonietti, if I remember correctly, we decided to leave the down-sampling mechanism out of this PR as this is a mechanism which could be implemented for many devices. 
@heplesser created a special issue for this topic (see #505), but far as I know, no-one works directly on this at the moment. So, feel free to implement this functionality, your contribution would be highly appreciated.

alberto-antonietti:
Thank you for the reply @weidel-p, since no-one is working on this I can try to implement it for the weights_recorder, having in mind that it should be applied also to other devices as discussed in #505.

PR #503, opened 2016-10-04
Fix READMEs for testsuite and PyNEST [ci skip]

jougs:
- Add CyNEST reference
- Extend the testsuite README to provide a more comprehensive explanation about how Python tests are handled


SepehrMN:
@jougs Thank you for fixing this. :+1:

PR #507, opened 2016-10-06
Use 1.5x vector growth factor for connections

alexeyshusharin:
Hi,

I did some measurements of connection memory consumption and found that connections consumes 50% more memory than it's expected.

The reason is that the current implementation of `Connector` (with 3 and more connections) use `std::vector` to store the connections. It is well known that a capacity of `std::vector` grows exponentially, in particular, gcc libstdc++ uses a growth factor of 2. In other words, vector doubles capacity each time when it needs to extend. That explains 50% overhead since this is average value for large number of vectors.

The growth factor 2 is not a good choice for several reason even for generic use case (very good explanation can be found [here](https://github.com/facebook/folly/blob/master/folly/docs/FBVector.md)). And it's even worse for connecton storage case.

This PR provides very simple fix that forces `std::vector` to use 1.5x grow factor. This factor is most popular choice of modern c++ libraries (unfortunatelly except gcc). It allows to decrease average overhead of connection memory consumption from 50% to 25%. The cost is more reallocations but with 1.5x grow factor it's not very significant.

I used one of my models to test these changes. This model includes several types of synapses with `17338400` connections in total. The sum of all vector sizes in bytes is `595840000`.

With default (2x) `std::vector` growth factor:

```
Capacity: 26289600
CapacityBytes: 900556800
Memory overhead: 50%
Number of reallocations: 1901600
```

With growth factor 1.5x:

```
Capacity: 20875200
CapacityBytes: 713011200
Memory overhead: 19%
Number of reallocations: 2119200 (111% of default)
```

Overall memory consumed by linux process decreased from 1.1 GiB to 909.9 MiB.
I didn't test performance of connection creation but we can guess looking at number of reallocations.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -74,6 +74,9 @@ nest::ConnectionManager::ConnectionManager()
   , connbuilder_factories_()
   , min_delay_( 1 )
   , max_delay_( 1 )
+  , init_conn_capacity_(CONFIG_CONNECTOR_CUTOFF)
+  , large_conn_limit_(CONFIG_CONNECTOR_CUTOFF * 2)
+  , large_conn_growth_(1.5)
 {
 }
 
@@ -130,6 +133,45 @@ nest::ConnectionManager::finalize()
 void
 nest::ConnectionManager::set_status( const DictionaryDatum& d )
 {
+  long init_cap = init_conn_capacity_;
+  if ( updateValue< long >( d, "init_connector_capacity", init_cap ) )
+  {
+    if ( init_cap < CONFIG_CONNECTOR_CUTOFF )
+    {
+      throw KernelException(
+        "The initial connector capacity should be higher or equal to "
+        "connector_cutoff value specified via cmake flag [default 3]");
+    }
+
+    init_conn_capacity_ = init_cap;
+  }
+
+  long large_lim = large_conn_limit_;
+  if ( updateValue< long >( d, "large_connector_limit", large_lim ) )
+  {
+    if ( large_lim < CONFIG_CONNECTOR_CUTOFF )
+    {
+      throw KernelException(
+        "The large connector limit should be higher or equal to "
+        "connector_cutoff value specified via cmake flag [default 3]");
+    }
+
+    large_conn_limit_ = large_lim;
+  }
+
+  double large_growth = large_conn_growth_;
+  if ( updateValue< double >( d, "large_connector_growth", large_growth ) )
+  {
+    if ( large_growth <= 1.0 )
+    {
+      throw KernelException(
+        "The large connector capacity growth factor should be higher than "
+        "1.0");
+    }
+
+    large_conn_growth_ = large_growth;
+  }
+
   for ( size_t i = 0; i < delay_checkers_.size(); ++i )
   {
     delay_checkers_[ i ].set_status( d );
@@ -148,6 +190,10 @@ nest::ConnectionManager::get_status( DictionaryDatum& d )
   update_delay_extrema_();
   def< double >( d, "min_delay", Time( Time::step( min_delay_ ) ).get_ms() );
   def< double >( d, "max_delay", Time( Time::step( max_delay_ ) ).get_ms() );
+
+  def< long >( d, "init_connector_capacity", init_conn_capacity_ );
+  def< long >( d, "large_connector_limit", large_conn_limit_ );
+  def< double >( d, "large_connector_growth", large_conn_growth_ );
 
   size_t n = get_num_connections();
   def< long >( d, "num_connections", n );
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -74,9 +74,9 @@ nest::ConnectionManager::ConnectionManager()
   , connbuilder_factories_()
   , min_delay_( 1 )
   , max_delay_( 1 )
-  , init_conn_capacity_(CONFIG_CONNECTOR_CUTOFF)
-  , large_conn_limit_(CONFIG_CONNECTOR_CUTOFF * 2)
-  , large_conn_growth_(1.5)
+  , init_conn_capacity_( CONFIG_CONNECTOR_CUTOFF )
+  , large_conn_limit_( CONFIG_CONNECTOR_CUTOFF * 2 )
+  , large_conn_growth_( 1.5 )
 {
 }
 
@@ -140,7 +140,7 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
     {
       throw KernelException(
         "The initial connector capacity should be higher or equal to "
-        "connector_cutoff value specified via cmake flag [default 3]");
+        "connector_cutoff value specified via cmake flag [default 3]" );
     }
 
     init_conn_capacity_ = init_cap;
@@ -153,7 +153,7 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
     {
       throw KernelException(
         "The large connector limit should be higher or equal to "
-        "connector_cutoff value specified via cmake flag [default 3]");
+        "connector_cutoff value specified via cmake flag [default 3]" );
     }
 
     large_conn_limit_ = large_lim;
@@ -166,7 +166,7 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
     {
       throw KernelException(
         "The large connector capacity growth factor should be higher than "
-        "1.0");
+        "1.0" );
     }
 
     large_conn_growth_ = large_growth;
--- a/nestkernel/connector_base.h
+++ b/nestkernel/connector_base.h
@@ -869,7 +869,7 @@ public:
     if ( sz == C_.capacity()
       and sz >= kernel().connection_manager.get_large_conn_limit() )
     {
-      const size_t cap = static_cast<double>( sz )
+      const size_t cap = static_cast< double >( sz )
         * kernel().connection_manager.get_large_conn_growth();
 
       C_.reserve( cap > sz ? cap : sz + 1 );
--- a/nestkernel/connector_base.h
+++ b/nestkernel/connector_base.h
@@ -120,7 +120,8 @@ suicide( Told* connector )
 }
 
 // when to truncate the recursive instantiation
-#define K_CUTOFF 3
+// can be specified via cmake flag -Dconnector_cutoff=value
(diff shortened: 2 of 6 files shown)


nestkernel/connector_base.h
```diff
@@ -817,6 +820,16 @@ class Connector< K_CUTOFF, ConnectionT > : public vector_like< ConnectionT >
   ConnectorBase&
   push_back( const ConnectionT& c )
   {
+    // Replace default vector grow strategy that is quite bad for our case.
+    // Use vector grow strategy 1.5 then size >= K_SLOW_GROWING.
```
heplesser: then -> when

nestkernel/connector_base.h
```diff
@@ -817,6 +820,16 @@ class Connector< K_CUTOFF, ConnectionT > : public vector_like< ConnectionT >
...
+    // Call vector::reserve() manually then size() == capacity().
```
heplesser: then -> if

nestkernel/connector_base.h
```diff
@@ -817,6 +820,16 @@ class Connector< K_CUTOFF, ConnectionT > : public vector_like< ConnectionT >
...
+    const size_t sz( C_.size() );
```
heplesser: To keep with practice elsewhere in NEST, I'd suggest
```C++
const size_t sz = C_.size();
```

alexeyshusharin:
@heplesser Thanks! I will provide performance test results later today.

Honestly, I don't like this fix much despite the fact that I'm proposing it. :) It just optimizes things a little bit but doesn't solve the problem in general. The true solution should avoid memory reallocation at all and has much less memory overhead. I have some thoughts how it can be done, but this solution requires changes in architecture and I don't want to start working on it without preliminary discussion.

Could you please let me know if you think such improvements are useful for NEST and you are interested in this discussion? Is a new Issue with good description the right place to start it?

alexeyshusharin:
@heplesser I did the performance tesing on the same model as used before and I'm really surprised! Performance of connection creation with 1.5x growth factor is better though number of vector reallocations is higher!

I run two types of tests:
- **cold start test** when I run python script so start a new linux process each time
- **reset kernel test** when I reset NEST kernel and create the connections again

For default vector growth factor:
```
Average time for cold start: 7.343313 sec
Average time for reset kernel: 3.2542215 sec
```

For 1.5x vector growth factor:
```
Average time for cold start: 6.0488181 sec
Average time for reset kernel: 3.1935857 sec
```

The better performance of 1.5x factor for cold start test can be explained by less amount of memory that needs to be requested from the operating system. But I have no idea how to explain the timing for reset kernel test. Probably this is some implication of allocator internals.

heplesser:
@alexeyshusharin Our general approach has been to let science drive development: If someone needed  a new capability or improved performance to do a certain science project, then he/she developed the necessary code for NEST (in collaboration with others). Our work on reducing the memory footprint and handling memory in general is described in papers, see especially Kunkel S et al (2014) [Spiking network simulation code for petascale computers](http://dx.doi.org/10.3389/fninf.2014.00078). More publications on NEST technology are on [our homepage](http://www.nest-simulator.org/publications/?sort=search&type=nest_technology).

Avoiding reallocation would probably a great step forward, but it would require a lot of knowledge about the network to be built, which may be difficult when building random networks.

alexeyshusharin:
@heplesser Sorry, maybe my question was not very clear. I have a proposal how to reduce memory overhead and avoid reallocations for connection management. And I'm ready to implement it. What I'm asking for is the best way to discuss this proposal with all stakeholders.

heplesser:
@alexeyshusharin I think the best way would be to create an Issue here describing your ideas so that we can discuss them. An Issue is better than a Pull Request for the initial of discussion, so that we avoid the risk of having discussions on principal design decisions in a PR. We also have the Open NEST Developer Video Conference, which takes place every other Monday 11.30-12.30 (CEST), next on Oct 17. Maybe you could join us there?

alexeyshusharin:
@heplesser Thanks! I will prepare a detailed description of the proposal and post it as an Issue. It will take a couple of days. Joining a video conference is a good idea. I will try to join.

alexeyshusharin:
@heplesser I've submitted the issue #523 as we discussed. You feedback is welcome!

mschmidt87:
Hi,
first of all, my apologies for the long delay of my reply.

I have executed a benchmark test using the hpc_benchmark script of NEST and compared the performance of the current master (master, 8069e3b...) with the branch of @alexeyshusharin (conn-mem, 4d8d070...). I performed a weak scaling with 12500 neurons per MPI process, 8 threads per MPI process and 1 MPI process per node. I varied the number of MPI processes from 32 to 1024 and obtained the following results:
juqueen_git8069e3b_git4d8d070_bympiprocesses

- Performance in terms of build and simulation time (left panel) is very similar, deviations are probably due to statistical variation.
- Memory consumption (right panel) for connections (crosses) is reduced for smaller job sizes (MT=256, 512) and approx. equal for the larger jobs.

Do these results make sense for the implied changes? Is is possible that for large networks, the benefit of the suggested changes vanishes some reason?

heplesser:
@mschmidt87 Thank you for the benchmarks! I think your results make sense, in particular the vanishing difference in memory consumption for MT==8192. In that case, each source neuron has roughly 12500/8192 ~ 1.5 targets, so most connectors will be fixed-size homogenous connectors with 1 or 2 elements, and extremely few target lists will be longer than 3, where the growth factor would come into play. Given your results, I would leave the growth factor as it is for now.

mschmidt87:
Yes, true, that makes sense.

I should add that these benchmarks have been performed on JUQUEEN, using Bluegene's XLC compiler instead of gcc. I had to make this choice because of the problems with gcc and gsl on JUQUEEN documented in #657 .

heplesser:
@mschmidt87 I fear you misunderstood me: You benchmarks show a limited to now advantage of growth-factor 1.5 for large simulations where memory is most crucial. At the same time, the change implies more complicated code. Therefore, my suggestion was to **not** accept this pull request.

mschmidt87:
I see. I guess, one can argue for both sides, but I see your point, so I am also fine with not accepting this PR.

heplesser:
@mschmidt87 @alexeyshusharin We discussed this PR in light of @mschmidt87's benchmarks in the [Developer VC on 6 March](https://github.com/nest/nest-simulator/wiki/2017-03-06-Open-NEST-Developer-Video-Conference) and came there up with the idea of making this configurable by a CMake flag. Thinking about this now, the overhead for the cmake flag would be far more than the code in this PR, so that makes little sense.

I also looked at the effect of the effect of the `K_SLOW_GROWING` parameter: With `K_CUTOFF==3` and `K_SLOW_GROWING==6`, we get the following sizes 

- with the slow-growing-limit in place: 3 -> 6 -> 9 -> 14 -> ...
- without the limit in place: 3 -> 5 -> 8 -> 12 -> ...

so the limit makes very little difference and we could drop it.

But this made me wonder if we shouldn't go for a more flexible approach:

- `K_CUTOFF` becomes configurable via a CMake flag (it has to be a compile-time constant for template resolution; not sure if it needs to be `#define`d of if a `static const` would do)
- We introduce new kernel parameters:
    - `initial_connector_capacity`: When a connector is first created, it starts with this capacity reserved (if >= K_CUTOFF)
    - `large_connector_limit`: Capacity doubling is used up to this limit
   -  `large_connector_growth`: Factor (double) to use beyond limit

This will give users the flexibility to adjust memory allocation policy to their network.

@alexeyshusharin Would you be interested in implementing this? Otherwise, I would create a new issue for this and close this PR.

alexeyshusharin:
@heplesser Sure! I'm a little bit busy and not able to work on my proposal about connection infrastructure improvements. But this change doesn't look huge so I can do this.

nestkernel/connection_manager.cpp
```diff
@@ -130,6 +133,45 @@ nest::ConnectionManager::finalize()
 void
 nest::ConnectionManager::set_status( const DictionaryDatum& d )
 {
+  long init_cap = init_conn_capacity_;
+  if ( updateValue< long >( d, "init_connector_capacity", init_cap ) )
```
heplesser: `names::init_connector_capacity`

nestkernel/connection_manager.cpp
```diff
@@ -130,6 +133,45 @@ nest::ConnectionManager::finalize()
...
+      throw KernelException(
+        "The initial connector capacity should be higher or equal to "
+        "connector_cutoff value specified via cmake flag [default 3]" );
+    }
+
+    init_conn_capacity_ = init_cap;
+  }
+
+  long large_lim = large_conn_limit_;
+  if ( updateValue< long >( d, "large_connector_limit", large_lim ) )
```
heplesser: `names::large_connector_limit`

nestkernel/connection_manager.cpp
```diff
@@ -130,6 +133,45 @@ nest::ConnectionManager::finalize()
...
+      throw KernelException(
+        "The large connector limit should be higher or equal to "
+        "connector_cutoff value specified via cmake flag [default 3]" );
+    }
+
+    large_conn_limit_ = large_lim;
+  }
+
+  double large_growth = large_conn_growth_;
+  if ( updateValue< double >( d, "large_connector_growth", large_growth ) )
```
heplesser: `names::large_connector_growth`

nestkernel/connection_manager.cpp
```diff
@@ -149,6 +191,10 @@ nest::ConnectionManager::get_status( DictionaryDatum& d )
   def< double >( d, "min_delay", Time( Time::step( min_delay_ ) ).get_ms() );
   def< double >( d, "max_delay", Time( Time::step( max_delay_ ) ).get_ms() );
 
+  def< long >( d, "init_connector_capacity", init_conn_capacity_ );
+  def< long >( d, "large_connector_limit", large_conn_limit_ );
+  def< double >( d, "large_connector_growth", large_conn_growth_ );
```
heplesser: Should all be `names::...` objects, see above.

jougs:
@alexeyshusharin: can you please update your branch and answer the comments by @heplesser? Thanks!

alexeyshusharin:
@jougs, @heplesser Sorry, completely forget about this PR. Will do appropriate changes soon. Thanks!

jougs:
@alexeyshusharin: many thanks for your efforts. To finally get this ready and merged, I took your branch, fixed the remaining issues and submitted #850 as a replacement for this PR. I'm thus closing this one.

PR #513, opened 2016-10-06
aEIF psc models

Silmathoron:
This PR adds implementations of the aEIF model with post-synaptic currents:

* ``aeif_psc_alpha``
* ``aeif_psc_exp``

Test is provided via an update of ``test_aeif_lsodar.py`` and by some of the usual suites such as ``issue-77.sli``, which was modified.


golosio:
@Silmathoron @heplesser if you agree I can review this PR

Silmathoron:
Of course you can ;)
but I'm waiting for the VC on Monday to (most probably) remove the 2nd dynamics function and solve the conflicts

models/aeif_psc_alpha.cpp
```diff
@@ -94,19 +94,20 @@ nest::aeif_psc_alpha_dynamics( double,
   // good compiler will optimize the verbosity away ...
 
   // shorthand for state variables
-  const double& V = std::min( y[ S::V_M ], node.P_.V_peak_ ); // bound V
+  const double& V = std::min( y[ S::V_M ], node.P_.V_peak_ ); // bind V to the
```
heplesser: I would suggest `bound V by V_peak`

models/aeif_psc_alpha.cpp
```diff
@@ -94,19 +94,20 @@ nest::aeif_psc_alpha_dynamics( double,
...
   const double& dI_syn_ex = y[ S::DI_EXC ];
   const double& I_syn_ex = y[ S::I_EXC ];
   const double& dI_syn_in = y[ S::DI_INH ];
   const double& I_syn_in = y[ S::I_INH ];
   const double& w = y[ S::W ];
 
-  const double I_spike =
-    node.P_.Delta_T * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T );
+  const double I_spike = node.P_.Delta_T == 0. ? 0. : node.P_.g_L
+      * node.P_.Delta_T * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T );
```
heplesser: For clarity, I would put the entire expression after `:` in parentheses, even though they are not strictly needed in C++.

models/aeif_psc_exp.cpp
```diff
@@ -94,17 +94,18 @@ nest::aeif_psc_exp_dynamics( double, const double y[], double f[], void* pnode )
   // good compiler will optimize the verbosity away ...
 
   // shorthand for state variables
-  const double& V = std::min( y[ S::V_M ], node.P_.V_peak_ );
+  const double& V = std::min( y[ S::V_M ], node.P_.V_peak_ ); // bind V to the
+  // USER DEFINED V_peak_ value in Parameters.
```
heplesser: See suggestion or `alpha` above.

models/aeif_psc_exp.cpp
```diff
@@ -94,17 +94,18 @@ nest::aeif_psc_exp_dynamics( double, const double y[], double f[], void* pnode )
...
   const double& I_syn_ex = y[ S::I_EXC ];
   const double& I_syn_in = y[ S::I_INH ];
   const double& w = y[ S::W ];
 
-  const double I_spike =
-    node.P_.Delta_T * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T );
+  const double I_spike = node.P_.Delta_T == 0. ? 0. : node.P_.g_L
+      * node.P_.Delta_T * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T );
```
heplesser: Also here parentheses around all afer `:` would be helpful.

models/aeif_psc_alpha.cpp
Silmathoron: I think this comment relates to an outdated version, current comment is
``// bind V to the USER DEFINED V_peak_ value in Parameters.``

heplesser: @Silmathoron The English language is a challenging one: "bound" is both past tense of "bind" and a verb in its own right meaning "to confine" or "to limit". The original comment uses "bound" in the second meaning. "Bind" would be incorrect here, since it implied tying the membrane potential to V_peak, instead of just having V_peak as an upper limit. But this discussions shows the need for clearer language, so I would suggest: `// enforce upper limit on V_m`. 

I would not mention "USER DEFINED V_peak_ value in Parameters", since that is clear from the code.

Silmathoron: done!

heplesser:
@golosio I am happy with the code, how about you?

golosio:
I have just very little suggestions/doubts

models/aeif_psc_alpha.cpp
```diff
@@ -0,0 +1,538 @@
...
+  updateValue< double >( d, names::a, a );
+  updateValue< double >( d, names::b, b );
+  updateValue< double >( d, names::Delta_T, Delta_T );
+  updateValue< double >( d, names::tau_w, tau_w );
+
+  updateValue< double >( d, names::I_e, I_e );
+
+  updateValue< double >( d, names::gsl_error_tol, gsl_error_tol );
+
+  if ( V_reset_ >= V_peak_ )
```
golosio: It seems that also here the check that V_peak_ <= V_th  is missing

models/aeif_psc_alpha.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+  return GSL_SUCCESS;
+}
+
+/* ----------------------------------------------------------------
+ * Default constructors defining default parameters and state
+ * ---------------------------------------------------------------- */
+
+nest::aeif_psc_alpha::Parameters_::Parameters_()
+  : V_peak_( 0.0 )    // mV, should not be larger that V_th+10
```
golosio: I've seen this comment in all aeif models, but I do not understand it. Why shouldn't V_peak be larger that V_th+10? Even looking at the default values, the default value of V_th is -50.4, while V_peak is zero, so the difference is larger than 10. Am I missing something?

models/aeif_psc_exp.cpp
```diff
@@ -0,0 +1,520 @@
...
+  updateValue< double >( d, names::a, a );
+  updateValue< double >( d, names::b, b );
+  updateValue< double >( d, names::Delta_T, Delta_T );
+  updateValue< double >( d, names::tau_w, tau_w );
+
+  updateValue< double >( d, names::I_e, I_e );
+
+  updateValue< double >( d, names::gsl_error_tol, gsl_error_tol );
+
+  if ( V_reset_ >= V_peak_ )
```
Silmathoron: Right, I forgot about that, thanks for reminding me! ;)

golosio: Besides the very little issues above, everything looks fine to me.

Silmathoron:
Ok, I readded the missing test to all ``aeif_*_alpha`` models. I guess it went missing during some rearrangement in the ``dynamics`` modification.

golosio:
@Silmathoron @heplesser everythng is ok for me, I think this PR can be merged

models/aeif_psc_alpha.cpp
heplesser: I think the idea here is that the larger the V_th to V_peak is, the worse does the exploding-exponential problem become. But I think we handle this well elsewhere now, so the comment can be removed.

heplesser:
@Silmathoron Thanks for putting the test back in! But I think we should allow V_peak == V_th. Also, I would write the error message as `"V_peak >= V_th required."`. That is closer to what we use elsewhere and explicitly names the parameters involved.

Silmathoron:
Ok, done! @golosio I'll let you do the changes on the ``beta_multisynapse`` model to avoid conflicts

heplesser:
@Silmathoron Thanks for your efforts, all is well now! Together with @golosio's :+1:, I will merge.

Issue #514, opened 2016-10-07
Update aeif_cond_beta_multisynapse to new aeif numerics approach


heplesser:
`aeif_cond_beta_multisynapse` should be updated to the same voltage-limiting scheme introduced for the `aeif_cond_alpha_*` models (except `RK5` variant) with #474.

@golosio Could you take care of this, maybe together with #510 ?

golosio:
ok, if you agree I'll put the changes together in a single PR

2016-10-07 7:18 GMT+02:00 Hans Ekkehard Plesser <notifications@github.com>:

golosio:
@Silmathoron I was taking care of this issue and I was wondering if it is really necessary to have two aeif_cond_alpha_dynamics functions in aeif_cond models, one for Delta_t>0 and the other for Delta_t=0, but very similar to each other. I tried to modify your version of aeif_cond_alpha using only one  aeif_cond_alpha_dynamics function with a check on Delta_t:
  if ( node.P_.Delta_T > 0. )
  {
    I_spike =
      node.P_.Delta_T * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T );
  }
  else
  {
    I_spike = 0;
  }
See: https://github.com/golosio/nest-simulator/blob/aeif_DT0/models/aeif_cond_alpha.cpp
and the corresponding header file.
I run simulations with Delta_t = 0 and != 0, using the same parameters that you used for comparing this model with lsodar (see the attached files) and the results are identical. Also, I run a long simulation to evaluate the simulation time, but there are no significant differences.

[aeif.py.txt](https://github.com/nest/nest-simulator/files/517723/aeif.py.txt)
[aeif_DT0.py.txt](https://github.com/nest/nest-simulator/files/517724/aeif_DT0.py.txt)

Silmathoron:
Well, I'm not expert enough to know whether the memory/speed gain will have any influence... maybe @heplesser can help on that or knows someone who might tell us. In the meantime I'll try to run some tests on a cluster to compare the runtimes, but this might take time because this week schedule is rather tight...
I agree, though, that having redundant code might not be the best option; yet it might make the distinction between the two situations clearer.

golosio:
[...]

Issue #510, opened 2016-10-06
Correct input handling for aeif_cond_beta_multisynapse


heplesser:
Input handling for `aeif_cond_beta_multisynapse` needs to be modified according to the decision made in the [NEST Developer VC 19 Sep 2016](https://github.com/nest/nest-simulator/wiki/2016-09-19-Open-NEST-Developer-Video-Conference):

- Replace `E_ex`, `E_in` with vector of reversal potentials the same size as `taus_syn` (one per synapse)
- Replace  `I_syn_ex`, `I_syn_in` recordables with one recordable `I_syn` representing the total input current
- In `handle(SpikeEvent&)`, consider adding a check for non-negative weight.
- Update `*_dynamics` function to handle input current correctly, along the lines of
```C++
  double I_syn = 0.0;
  for ( size_t i = 0; i < node.P_.num_of_receptors_ ; ++i )
  {
      I_syn += y[ S::G + i ] * ( node.P_.E_rev[i] - V );
  }
```
- Provide examples (python based).

See also #509, #511.

heplesser:
@golosio Could you look at this? cond_beta is already furthest, since it has the proper `handle()` method.

golosio:
Sure, I'll have a look.

golosio:
We have to take care that the same quantity should have the same name in different models. How should I call the reversal potential? Maybe V_rev (I prefer this) or E_rev?

heplesser:
I think in the literature reversal potentials are usually given as E, so I would suggest `E_rev`. @Silmathoron Any thoughts on this?

Silmathoron:
For conductance-based flows, I've never seen something other than  E be used... so ``E_rev`` looks good!

Silmathoron:
By the way, just a thought since we're at it: should there be a possibility of giving only a double for the reversal potentials and taus if the user want them to be all equal?

golosio:
Ok. Also, we should take a decision on what to do when the number of receptor ports is changed, and when E_rev or taus_syn ( taus_rise or taus_decay for _cond_beta_ ) are defined or redefined with a number of elements that does not match the current number of receptor ports.
In aeif_cond_beta_multisynapse I tried to follow the same approach of aeif_cond_alpha_multisynapse and I used one of the vectors (taus_decay) to control the number of receptor ports. However I thought about this a lot and I think it was not the best approach.
My proposal is:
[...]

PR #516, opened 2016-10-07
Correct input handling for aeif_cond_beta_multisynapse #510 and Update aeif_cond_beta_multisynapse to new aeif numerics approach #514

golosio:
I modified input handling for aeif_cond_beta_multisynapse according to the decision made in the NEST Developer VC 19 Sep 2016, see issue https://github.com/nest/nest-simulator/issues/510


golosio:
I also adapted the example in aeif_cond_beta_multisynapse.h and the SLI test.
I still have to do the check that weights are positive and take care of issue https://github.com/nest/nest-simulator/issues/514

golosio:
@heplesser about the weight check, should I use a simple assert, as
  assert( e.get_delay() > 0 );
on the same line as the other checks that are done in handle( SpikeEvent& e ), or should I throw an exception with an error message, as
throw BadProperty("Synaptic weights for conductance based models must be positive." );

heplesser:
@golosio Raising an exception is certainly more user friendly, so I would suggest that.

golosio:
Ok, I should have finished. I made a test using the same parameters of the lsodar test, with one receptor port and synaptic rise time = decay time so that the models behaves as the alpha model, and the output is the same as that of aeif_cond_alpha, see the scripts:
[aeif_ms.py.txt](https://github.com/nest/nest-simulator/files/518172/aeif_ms.py.txt)
[aeif_DT0_ms.py.txt](https://github.com/nest/nest-simulator/files/518173/aeif_DT0_ms.py.txt)
@heplesser and @Silmathoron I wait for your revision.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -144,84 +153,75 @@ aeif_cond_beta_multisynapse::Parameters_::get( DictionaryDatum& d ) const
...
+    updateValue< std::vector< double > >( d, names::taus_decay, taud_tmp );
+  if ( Erev_flag || taur_flag || taud_flag )
+  { // receptor arrays have been modified
+    if ( !Erev_flag || !taur_flag || !taud_flag )
     {
       throw BadProperty(
-        "The neuron has connections, therefore the number of ports cannot be "
-        "reduced." );
+        "The reversal potential, synaptic rise time and synaptic decay time "
+        "arrays must be modified altogether." );
```
Silmathoron: I don't think "altogether" is the appropriate term here...
Beyond that, I think what we want to check is only change in length... users should be able to modify only one of the properties as long as they change the values inside and not the number of values, right?

golosio: It's a problem with my English, I will fix the error message.... About the implementation of the constraints, I followed @heplesser suggestions in comment https://github.com/nest/nest-simulator/issues/510#issuecomment-252402610 when he says:
"if a call to SetStatus contains taus_syn, it also must contain E_rev, and they must be of the same length". In case of aeif_cond_alpha_multisynapse there is no taus_syn, but taus_rise and taus_decay, so I guess that SetStatus should either contain none of them or all three...
If I understand well, what you are proposing is that you can also call SetStatus with one or two of them as far as they have the same length that they had previously. For me it would be fine. @heplesser what do you think about this?

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -76,23 +79,28 @@ aeif_cond_beta_multisynapse::Parameters_::Parameters_()
...
-  , E_ex( 0.0 )       // mV
-  , E_in( -85.0 )     // mV
   , E_L( -70.6 )      // mV
   , Delta_T( 2.0 )    // mV
   , tau_w( 144.0 )    // ms
   , a( 4.0 )          // nS
   , b( 80.5 )         // pA
   , V_th( -50.4 )     // mV
-  , I_e( 0.0 )        // pA
+  , num_of_receptors( 1 )
```
heplesser: I wonder if we should have this as a member. Since it will always be the same as `E_ref.size()`, I think we should should just use
```C++
inline size_t n_receptors() { return E_rev.size(); }
```

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -144,84 +153,75 @@ aeif_cond_beta_multisynapse::Parameters_::get( DictionaryDatum& d ) const
...
-  updateValue< double >( d, names::E_in, E_in );
 
   updateValue< double >( d, names::C_m, C_m );
   updateValue< double >( d, names::g_L, g_L );
 
-  std::vector< double > tau_tmp;
-  if ( updateValue< std::vector< double > >( d, names::taus_decay, tau_tmp ) )
-  {
-    if ( tau_tmp.size() < taus_decay.size() && has_connections_ == true )
+  std::vector< double > Erev_tmp, taur_tmp, taud_tmp;
```
heplesser: You should avoid defining multiple variables in one statement.

You also don't need temporary variables here, since `set()` is called on a temporary copy of `P_` that will only be copied back to the real `P_` if no exception is raised. This may also simplify some of the error checking below.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -144,84 +153,75 @@ aeif_cond_beta_multisynapse::Parameters_::get( DictionaryDatum& d ) const
...
         throw BadProperty(
           "Synaptic rise time must be smaller than or equal to decay time." );
       }
     }
-    taus_rise = tau_tmp;
+    E_rev = Erev_tmp;
+    taus_rise = taur_tmp;
+    taus_decay = taud_tmp;
+    num_of_receptors = taur_tmp.size();
+    receptor_types.resize( num_of_receptors );
```
heplesser: See above, I think we can drop `receptor_types`.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,44 +686,37 @@ aeif_cond_beta_multisynapse_dynamics( double,
   // good compiler will optimize the verbosity away ...
 
   // shorthand for state variables
-  const double& V = y[ S::V_M ];
+  const double& V = std::min( y[ S::V_M ], node.P_.V_peak_ ); // bound V
   const double& w = y[ S::W ];
 
   double I_syn = 0.0;
```
heplesser: I would remove this empty line to tie the initialization more directly to the loop.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,44 +686,37 @@ aeif_cond_beta_multisynapse_dynamics( double,
...
-    double g = y[ S::G + i ];
-    if ( g > 0 )
-    {
-      I_syn += g * ( node.P_.E_ex - V ); // g>0, E_ex>V => I_syn increases
-    }
-    else
-    {
-      I_syn += g * ( V - node.P_.E_in ); // g<0, V>E_in => I_syn decreases
-    }
+    size_t j = i * S::NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR;
```
heplesser: Add `const`.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,44 +686,37 @@ aeif_cond_beta_multisynapse_dynamics( double,
...
+    // I_syn = I_syn + g*(E_rev - V)
```
heplesser: Maybe move this comment above the `I_syn = 0` initialization and make the summation explicit: `I_syn = - sum_k g_k (V - E_rev_k)`.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,44 +686,37 @@ aeif_cond_beta_multisynapse_dynamics( double,
...
+  f[ S::V_M ] = ( -node.P_.g_L * ( V - node.P_.E_L - I_spike ) + I_syn - w
                   + node.P_.I_e + node.B_.I_stim_ ) / node.P_.C_m;
 
   // Adaptation current w.
   f[ S::W ] = ( node.P_.a * ( V - node.P_.E_L ) - w ) / node.P_.tau_w;
 
-  for ( size_t i = 0; i < node.P_.num_of_receptors_; ++i )
+  for ( size_t i = 0; i < node.P_.num_of_receptors; ++i )
   {
     size_t j = i * S::NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR;
```
heplesser: `const`

models/aeif_cond_beta_multisynapse.h
```diff
@@ -74,16 +74,18 @@
 
  neuron = nest.Create('aeif_cond_beta_multisynapse')
  nest.SetStatus(neuron, {"V_peak": 0.0, "a": 4.0, "b":80.5})
- nest.SetStatus(neuron, {'taus_decay':[50.0,20.0,20.0,20.0],
+ nest.SetStatus(neuron, {'num_of_receptors':4,
```
heplesser: `num_of_receptors` should never be set explicitly.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -74,16 +74,18 @@
...
+                                                 np.array([10.0])})
 
  voltmeter = nest.Create('voltmeter', 1, {'withgid': True})
 
  delays=[1.0, 300.0, 500.0, 700.0]
- w=[1.0, 1.0, 1.0, -1.0]
+ w=[1.0, 1.0, 1.0, 1.0]
  for syn in range(4):
      nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
                                            'weight': w[syn],
```
heplesser: Don't you need to specify `receptor_type` here?

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -75,12 +75,14 @@ if
 
 /tolerance   	5e-2 def %mV
 /simulation_t	2500 def %ms
-/V_peak			0. def
-/a 				4. def
-/b 				80.5 def
+/V_peak		0. def
```
heplesser: Fix whitespace here and in the following lines.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -75,12 +75,14 @@ if
...
+/a 		4. def
+/b 		80.5 def
+/num_of_receptors       3 def
```
heplesser: Should not be set explicitly.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -75,12 +75,14 @@ if
...
+/E_rev		        [0.0 0.0 -85.0] def % synaptic reversal potentials  
 /taus_decay		[400.0 200.0 200.0] def % synaptic decay times  
 /taus_rise		[200.0 100.0 100.0] def % synaptic rise times  
-/weights		[0.1 0.1 -0.1] def % synaptic weights
+/weights		[0.1 0.1 0.1] def % synaptic weights
```
heplesser: Wouldn't it be better to make all reversal potentials, time constants and weights different from each other to make sure nothing gets mixed up in the code?

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -75,12 +75,14 @@ if
...
 /delays     	[1. 1000. 1500.] def % ms - synaptic delays
```
heplesser: Do you need such long delays? Shorter simulation times would make the test run faster.

models/aeif_cond_beta_multisynapse.cpp
golosio: I changed the error message. Please check if it is ok now, otherwise please suggest a suitable one. I modified the behavior for changes in the arrays according to @Silmathoron indication.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -20,6 +20,8 @@
  *
  */
 
+#include <stdio.h>
```
golosio: Ops... sorry I use it for myself for debugging and I forgot it. Removed.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -41,6 +43,7 @@
 #include "doubledatum.h"
 #include "integerdatum.h"
 
+using namespace std;
```
golosio: removed

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -76,23 +79,28 @@ aeif_cond_beta_multisynapse::Parameters_::Parameters_()
...
-  , I_e( 0.0 )        // pA
+  , num_of_receptors( 1 )
+  , I_e( 0.0 ) // pA
   , gsl_error_tol( 1e-6 )
-  , num_of_receptors_( 1 )
   , has_connections_( false )
 {
+  static const double E_rev_def = 0.0;       // E_rev default value
+  static const double taus_rise_def = 2.0;   // taus_rise default value
+  static const double taus_decay_def = 20.0; // taus_decay default value
```
golosio: good, now the code is much more compact, there is only the initializer list and the body is empty. For now I'm still using the default with one receptor port, but it will be easy to remove it.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -76,23 +79,28 @@ aeif_cond_beta_multisynapse::Parameters_::Parameters_()
...
+  receptor_types.clear();
+  E_rev.clear();
   taus_rise.clear();
   taus_decay.clear();
```
golosio: removed.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -76,23 +79,28 @@ aeif_cond_beta_multisynapse::Parameters_::Parameters_()
...
-  taus_decay.push_back( 20.0 );
-  taus_rise.push_back( 2.0 );
+  receptor_types.push_back( 1 );
```
golosio: Right, it was not used anywhere, I removed it.

models/aeif_cond_beta_multisynapse.cpp
golosio: removed.

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.h
golosio: removed

models/aeif_cond_beta_multisynapse.h
golosio: You mean that the order should be important? I do not know, I did not see any difference in the simulations if I put 'receptor_type' at the end of the syn_spec list, in any case now I've put it in the second position after 'model'.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
golosio: removed

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
golosio: ok I used values different from each other.

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
golosio: ok I increased g_L so the approximation used for the analytical calculation is also valid for smaller synaptic times, and I could also decrease the tolerance

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
golosio: done

golosio:
@heplesser @Silmathoron ok I made all the changes that you requested.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -457,15 +439,9 @@ aeif_cond_beta_multisynapse::calibrate()
...
   B_.logger_.init();
 
-  P_.receptor_types.resize( P_.num_of_receptors );
-  for ( size_t i = 0; i < P_.num_of_receptors; i++ )
-  {
-    P_.receptor_types[ i ] = i + 1;
-  }
-
-  V_.g0_.resize( P_.num_of_receptors );
+  V_.g0_.resize( P_.n_receptors() );
```
heplesser: I think it would be good for readability to point out that `g0_` will be initialized in the loop below.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -501,10 +477,9 @@ aeif_cond_beta_multisynapse::calibrate()
   assert( V_.RefractoryCounts_
     >= 0 ); // since t_ref_ >= 0, this can only fail in error
 
-  B_.spikes_.resize( P_.num_of_receptors );
+  B_.spikes_.resize( P_.n_receptors() );
   S_.y_.resize( State_::NUMBER_OF_FIXED_STATES_ELEMENTS
-    + ( State_::NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR
-                  * P_.num_of_receptors ) );
+    + ( State_::NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR * P_.n_receptors() ) );
```
heplesser: I think it would be good to explicitly give the initial value `0` as second argument to `resize()`.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -168,60 +156,54 @@ aeif_cond_beta_multisynapse::Parameters_::set( const DictionaryDatum& d )
   updateValue< double >( d, names::C_m, C_m );
   updateValue< double >( d, names::g_L, g_L );
 
-  std::vector< double > Erev_tmp, taur_tmp, taud_tmp;
+  size_t old_n_receptors = n_receptors();
```
heplesser: Should be const.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -168,60 +156,54 @@ aeif_cond_beta_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
   bool Erev_flag =
-    updateValue< std::vector< double > >( d, names::E_rev, Erev_tmp );
+    updateValue< std::vector< double > >( d, names::E_rev, E_rev );
   bool taur_flag =
-    updateValue< std::vector< double > >( d, names::taus_rise, taur_tmp );
+    updateValue< std::vector< double > >( d, names::taus_rise, taus_rise );
   bool taud_flag =
-    updateValue< std::vector< double > >( d, names::taus_decay, taud_tmp );
+    updateValue< std::vector< double > >( d, names::taus_decay, taus_decay );
   if ( Erev_flag || taur_flag || taud_flag )
```
heplesser: I think it should be possible to simplify this some more:
```C++
// n_receptors() == E_rev.size()
if ( n_receptors() != taus_rise.size() || n_receptors() != taus_decay.size() )
{
  throw BadProperty("E_rev, taus_raus and taus_decays must have the same size.");
}
```
Now we know all is consistent I would then check `n_receptors() != old_n_receptors && has_connections` and `n_receptors > 0`. Then the check that the taus > 0 and tau_rise < tau_decay. That should maybe be protected by `if ( taur_flag || taud_flag )` since we have loops here and would want to avoid running through them if the taus have not been changed.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -168,60 +156,54 @@ aeif_cond_beta_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
         "The reversal potential, synaptic rise time and synaptic decay time "
         "arrays must have the same size." );
     }
-    if ( taur_tmp.size() == 0 )
+    if ( taus_rise.size() == 0 )
     {
       throw BadProperty( "The neuron must have at least one port." );
     }
-    if ( taur_tmp.size() < num_of_receptors && has_connections_ == true )
+    if ( taus_rise.size() < old_n_receptors && has_connections_ == true )
```
heplesser: Avoid comparisons with `true` and `false`.

golosio:
Ok I'll implement the changes that you suggest. Just one note and a question. When you say

>> It is a bit unusual that State_::set() allows setting of conductances and their derivatives. We don't support that for other models and I am a bit sceptical about adding it here, especially the derivative.

Actually this is  how it is done in aeif_cond_alpha_multisynapse, which was the base for this model; I do not know about other models. Where do you suggest to move this part of the code?

models/aeif_cond_beta_multisynapse.cpp
golosio: I agree that the code can be made more compact if we simply want to check consistency. However my idea is that we should not only check for consistency, but also give the user a feedback that suggests him how he should modify his command. For instance, if the current number of receptor ports is 4 (or if they have not yet been initialized so the number is 0 or 1) and the user types a command as:
nest.SetStatus(neuron, {'taus_decay':[50.0,20.0,20.0]})
I think a message as:
"If the number of receptor ports is changed, all three arrays E_rev, taus_rise and taus_decay must be provided."
is more helpful to the user than:
"E_rev, taus_rise and taus_decay must have the same size."
because with the latter message it may not be obvious to all users that the problem can be solved by providing all three arrays as arguments.

Silmathoron:
@golosio I think what @heplesser means is that maybe the user should not be able to modify these variables.
I must say that I do not really agree on that, but I guess we should discuss it on Monday (actually I'm waiting for the VC before I start my review, which is why I haven't been very active lately; I hope all these issues will be clearer afterwards).
In the meantime, I guess you should leave that as it is until then ;)

models/aeif_cond_beta_multisynapse.h
```diff
@@ -202,16 +201,18 @@ class aeif_cond_beta_multisynapse : public Archiving_Node
...
-    std::vector< long > receptor_types;
-
     // boolean flag which indicates whether the neuron has connections
     bool has_connections_;
 
     Parameters_(); //!< Sets default parameter values
 
     void get( DictionaryDatum& ) const; //!< Store current values in dictionary
     void set( const DictionaryDatum& ); //!< Set values from dictionary
+    inline size_t
```
golosio: done

models/aeif_cond_beta_multisynapse.cpp
golosio: done

golosio:
@heplesser I made the changes that you suggested, except for the two for which I wait for your feedback. I also wait for @Silmathoron review after the VC.

Silmathoron:
Ok, the VC concluded that default behaviour should be that of the single synapse model, so you should implement a default with 1 synapse and the same parameters as the aeif_cond_alpha, except eventually for the synaptic time constants... I'll let you judge on that unless @heplesser has specific suggestions.
As for the ``dynamics`` functions, the VC agreed on removing the ``_DT0`` so you can update the function according to your suggestion and @heplesser['s comment](https://github.com/nest/nest-simulator/issues/514#issuecomment-253451402).
Eventually, I think the people in the VC mostly thought giving access to all state variables in set was rather uninteresting but no one seemed stronlgy opposed to it either... so my guess is you can do as you feel like, but maybe @heplesser would like to comment on that...

Silmathoron:
Ok, there are a few problems that were missed previously on lines [303](https://github.com/nest/nest-simulator/pull/516/files#diff-230597c8da565f8ce73398a1bd42067fR303) and [310](https://github.com/nest/nest-simulator/pull/516/files#diff-230597c8da565f8ce73398a1bd42067fR310):
unless I'm mistaken, ``dg`` is the derivative, so error message on L303 should be ``"Conductances and derivatives arrays should have the same size"``, and on L310, ``dg`` should be allowed to be negative.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -425,6 +422,15 @@ aeif_cond_beta_multisynapse::init_buffers_()
   B_.sys_.params = reinterpret_cast< void* >( this );
   // B_.sys_.dimension is assigned in calibrate()
   B_.I_stim_ = 0.0;
+  // set the right threshold depending on Delta_T
+  if ( P_.Delta_T > 0. )
+  {
+    B_.V_peak = P_.V_peak_;
```
Silmathoron: I'm almost sure this is supposed to be in calibrate since we want this check to be done every time ``Simulate`` is called, so ``V_peak`` should be in ``V_``; @heplesser, could you confirm?

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,46 +662,30 @@ aeif_cond_beta_multisynapse_dynamics( double,
   // good compiler will optimize the verbosity away ...
 
   // shorthand for state variables
-  const double& V = y[ S::V_M ];
+  const double& V = std::min( y[ S::V_M ], node.P_.V_peak_ ); // bound V
```
Silmathoron: Maybe we should add a comment to assure people that we indeed want to use ``P_.V_peak_`` and not ``V_.V_peak`` here...

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -563,7 +563,7 @@ aeif_cond_beta_multisynapse::update( Time const& origin,
       {
         S_.y_[ State_::V_M ] = P_.V_reset_;
       }
-      else if ( S_.y_[ State_::V_M ] >= P_.V_peak_ )
+      else if ( S_.y_[ State_::V_M ] >= B_.V_peak )
```
Silmathoron: Switch ``V_peak`` to ``V_`` (see above)

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,46 +662,30 @@ aeif_cond_beta_multisynapse_dynamics( double,
...
-  // We pre-compute the argument of the exponential
-  const double exp_arg = ( V - node.P_.V_th ) / node.P_.Delta_T;
-
-  // Upper bound for exponential argument to avoid numerical instabilities
-  const double MAX_EXP_ARG = 10.;
-
-  // If the argument is too large, we clip it.
-  const double I_spike =
-    node.P_.Delta_T * std::exp( std::min( exp_arg, MAX_EXP_ARG ) );
+  const double I_spike = node.P_.Delta_T == 0. ? 0 : node.P_.Delta_T
```
Silmathoron: To be consistent (and homogeneous), change to ``node.P_.Delta_T * node.P_.g_L * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T )``

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -656,46 +662,30 @@ aeif_cond_beta_multisynapse_dynamics( double,
...
+      * std::exp( ( V - node.P_.V_th ) / node.P_.Delta_T );
 
   // dv/dt
-  f[ S::V_M ] = ( -node.P_.g_L * ( ( V - node.P_.E_L ) - I_spike ) + I_syn - w
+  f[ S::V_M ] = ( -node.P_.g_L * ( V - node.P_.E_L - I_spike ) + I_syn - w
```
Silmathoron: Switch to ``( -node.P_.g_L * ( V - node.P_.E_L ) + I_spike + I_syn + ...``

models/aeif_cond_beta_multisynapse.h
```diff
@@ -294,6 +298,7 @@ class aeif_cond_beta_multisynapse : public Archiving_Node
      * the first simulation, but not modified before later Simulate calls.
      */
     double I_stim_;
+    double V_peak;
```
Silmathoron: Should be in ``V_`` because it can be modified before ``Simulate``

golosio:
Thank you @Silmathoron , I implemented the default with one synapse and the same parameters as the aeif_cond_alpha, except for the synaptic rise time and decay time. I updated the dynamic function according to @heplesser suggestion. 
About State_::set() allowing to set conductances and their derivatives, I think @heplesser already said his opinion, so I would remove this possibility, unless you think there is really a reason to allow it

Silmathoron:
I would rather say that I don't really see a reason to remove it if you implemented it ^^
Otherwise I don't really care since I don't have a use case for the moment... it's up to you, as far as I am concerned.

golosio:
@Silmathoron 
ok since @heplesser was skeptic about it I will remove it.

>> unless I'm mistaken, dg is the derivative, so error message on L303 should be "Conductances and derivatives arrays should have the same size", and on L310, dg should be allowed to be negative.

As I will remove this part of the code, those changes are not necessary any more. However, note that what is called dg in the code IS NOT the derivative of g. If you look for instance Roth and van Rossum, "Modeling synapses"
http://homepages.inf.ed.ac.uk/mvanross/reprints/roth_mvr_chap.pdf
equations 6.7 - 6.9 and compare them with the equations in aeif_cond_beta_multisynapse_dynamic you can observe that what is called dg (or y[ S::DG + j ] ) in the code is actually the function that is called h in those equations, and from equation 6.8 you can see that the derivative dg/dt is not h, because there is another term. Sorry, but the code of aeif_cond_beta_multisynapse was originally based on aeif_cond_alpha_multisynapse, and I did not modify the name of the variable dg although I thought it was somehow inappropriate.

Silmathoron:
Ok, I did not see that ``S::DG`` and ``dg`` where different; this is indeed not optimal and speaks in favour of removing it.
By the way, I just noticed that the docstring of the model in ``aeif_cond_beta_multisynapse.h`` does not include a description of the parameters, so could you add one? (see example in [``aeif_cond_alpha``](https://github.com/golosio/nest-simulator/blob/05eef9ae2f5214a7ea3dda967578aba7c03c0e67/models/aeif_cond_alpha.h#L65)

golosio:
>> the docstring of the model in aeif_cond_beta_multisynapse.h does not include a description of the parameters, so could you add one? (see example in aeif_cond_alpha

sure, thank you for indicating the example!

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -144,84 +141,69 @@ aeif_cond_beta_multisynapse::Parameters_::get( DictionaryDatum& d ) const
...
     {
-      for ( size_t i = 0; i < taus_decay.size(); ++i )
-      {
-        taus_rise.push_back( taus_decay[ i ] / 10. );
-        // if taus_rise is not defined explicitly or if it has less elements
-        // than taus_decay, it will be set to taus_decay/10
-      }
+      throw BadProperty(
+        "The reversal potential, synaptic rise time and synaptic decay time "
+        "arrays must have the same size." );
```
golosio: About this point, please have a look at my answer in
https://github.com/nest/nest-simulator/pull/516#discussion_r83532267
I agree that if we simply want to check  that the changes are consistent, the code can be made more compact. However I think that we should also think about what error messages would be more helpful for the user.
I think we should distinguish those two erroneous situations:
1) The user types a command that provides only one or two arrays, and they have a different size from the old one. Then the error message should be something like:
"If the number of receptor ports is changed, all three arrays  E_rev, taus_rise and taus_decay must be provided."
2) The user types a command that provides three arrays, but they have inconsistent sizes
Then the error message should be something like:
"The reversal potential, synaptic rise time and synaptic decay time arrays must have the same size."
If we first agree on which error messages should be printed in different situations, then we can discuss on what is the best way to implement it in the code.

Silmathoron: My bad, I remember reading about something like what you said. Ok, I understand your point.

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.cpp
golosio: done. However when I will commit please check if this is sufficient or please suggest a more suitable comment.

models/aeif_cond_beta_multisynapse.cpp
golosio: done

models/aeif_cond_beta_multisynapse.h
golosio: done

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -233,14 +215,29 @@ aeif_cond_beta_multisynapse::Parameters_::set( const DictionaryDatum& d )
 
   updateValue< double >( d, names::gsl_error_tol, gsl_error_tol );
 
-  if ( V_peak_ <= V_th )
+  if ( V_reset_ >= V_peak_ )
```
Silmathoron: Ok, then this is probably my fault ^^"
No, just change it here, I'll correct the ``aeif_cond_alpha`` it in the PR suppressing the ``dynamics_DT0``

golosio:
@Silmathoron Thank you for your review. I think I'm done in making the changes that you requested.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -64,6 +64,48 @@
  reversal potentials are supplied by the array "E_rev". The port numbers
  are automatically assigned in the range from 1 to n_receptors.
  During connection, the ports are selected with the property "receptor_type".
+The membrane potential is given by the following differential equation:
+C dV/dt= -g_L(V-E_L)+g_L*Delta_T*exp((V-V_T)/Delta_T)-g_e(t)(V-E_e)
+                                                     -g_i(t)(V-E_i)-w +I_e
```
Silmathoron: Ok, sorry, one last comment: since it's multisynapse, here, you should probably detail:

    C dV/dt= -g_L(V-E_L) + g_L*Delta_T*exp((V-V_T)/Delta_T) + I_syn_tot(V, t) - w + I_e
    
    where:
    I_syn_tot(V,t) = \sum_i g_i(t) (V - E_{rev,i})

    and the synapse i is excitatory or inhibitory depending on the value of E_{rev,i}.

golosio:
@Silmathoron done!

nestkernel/nest_names.h
```diff
@@ -407,6 +407,8 @@ extern const Name taus_eta;   //!< Specific to population point process model
 extern const Name taus_syn;   //!< Synapse time constants (array)
 extern const Name taus_rise;  //!< Synapse rise constants (array)
 extern const Name taus_decay; //!< Synapse decay constants (array)
+extern const Name E_rev;      //!< Reversal potential (array)
```
heplesser: Entries in `nest_names.[h,cpp]` should be in strictly alphabetical order.

nestkernel/nest_names.h
```diff
@@ -407,6 +407,8 @@ extern const Name taus_eta;   //!< Specific to population point process model
...
+extern const Name num_of_receptors; //!< number of receptor ports
```
heplesser: In other names, we have used `n_`, e.g., `n_events`.  I suggest `n_receptors` here for consistency.

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -473,14 +470,24 @@ aeif_cond_beta_multisynapse::calibrate()
...
   V_.RefractoryCounts_ = Time( Time::ms( P_.t_ref_ ) ).get_steps();
   assert( V_.RefractoryCounts_
     >= 0 ); // since t_ref_ >= 0, this can only fail in error
 
-  B_.spikes_.resize( P_.num_of_receptors_ );
+  B_.spikes_.resize( P_.n_receptors() );
   S_.y_.resize( State_::NUMBER_OF_FIXED_STATES_ELEMENTS
-    + ( State_::NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR
-                  * P_.num_of_receptors_ ) );
+    + ( State_::NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR * P_.n_receptors() ) );
```
heplesser: I'd suggest explicitly initializing with `0.0` here, as second argument.

models/aeif_cond_beta_multisynapse.cpp
heplesser: I think for clarity, it would be good to place everything after the `:` in parentheses, even if not strictly necessary.

Silmathoron:
By the way, @golosio, I think you also forgot to readd the ``V_peak_ > V_th`` check in your branch... I was looking for it in your code to add it again in the psc models x)

golosio:
@Silmathoron maybe you are looking to an updated version of the file... it should be in line 218.

Silmathoron:
My bad, you're probably right, must have been an outdated file.
All good!

heplesser:
@golosio As in #513 , I would suggest we allow V_th == V_peak and rephrase the error message.

golosio:
@heplesser done!

models/aeif_cond_beta_multisynapse.h
```diff
@@ -51,20 +51,66 @@
...
 
- aeif_cond_beta_multisynapse is an extension of
- aeif_cond_alpha_multisynapse.  It allows an arbitrary number of synaptic
- rise time and decay time constant. Synaptic conductance is modeled by a
+ aeif_cond_beta_multisynapse is a conductance based adaptive exponential
+ integrate-and-fire neuron model. It allows an arbitrary number of synaptic
+ rise time and decay time constants. Synaptic conductance is modeled by a
  beta function, as described by A. Roth and M.C.W. van Rossum
  in Computational Modeling Methods for Neuroscientists, MIT Press 2013,
  Chapter 6
```
heplesser: Fullstop missing at end of sentence.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -51,20 +51,66 @@
...
 
- The time constants are supplied by two arrays taus_rise and taus_decay for
- the synaptic rise time and decay time, respectively.  Port numbers
- are then automatically assigned and there range is from 1 to n.  (n
- being the index of the last element of the tau_ex and tau_in
- arrays).
+ The number of receptor ports is supplied by the function "n_receptors".
```
heplesser: I would drop this sentence, and `n_receptors` is not a function. I think the meaning of n_receptors becomes clear from the last sentence in this paragraph.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -51,20 +51,66 @@
...
+ The membrane potential is given by the following differential equation:
+ C dV/dt = -g_L(V-E_L) + g_L*Delta_T*exp((V-V_T)/Delta_T) + I_syn_tot(V, t)
+           - w + I_e
+
+ where:
+ I_syn_tot(V,t) = \sum_i g_i(t) (V - E_{rev,i})
+
+ and the synapse i is excitatory or inhibitory depending on the value of
+ E_{rev,i}.
+
```
heplesser: The differential equation for w fell out in one of your previous commits. I think it should be here. Would you add it again?

models/aeif_cond_beta_multisynapse.h
```diff
@@ -51,20 +51,66 @@
...
+                      function).
+
+Integration parameters
+  gsl_error_tol  double - This parameter controls the admissible error of the
+                          GSL integrator. Reduce it if NEST complains about
+                          numerical instabilities.
+
  Examples:
  % PyNEST example, of how to assign synaptic rise time and decay time
  % to a receptor type.
```
heplesser: I think the comment here is not needed. If you want to keep it, mark the comment with Python's `#` instead of `%`, which introduces comments in SLI.

models/aeif_cond_beta_multisynapse.h
golosio: done.

models/aeif_cond_beta_multisynapse.h
golosio: done.

models/aeif_cond_beta_multisynapse.h
golosio: removed.

golosio:
@heplesser ok I made the changes that you requested!

heplesser:
@golosio Thank you for your efforts! Together with @Silmathoron's "all good" above, I will merge.

golosio:
Thank you @heplesser and @Silmathoron !

Issue #440, opened 2016-08-01
Add warning when two-way connections are created without symmetric=true


janhahne:
PR #315 added a symmetric flag to the `conn_dict` in order to be able to create symmetric connections ( a->b , b->a ) with a single call to `nest.Connect`. Two-way connections like `gap_junction` demand this flag to be set to `true` in order to create accurate connections. 

The default setting for `symmetric` is however `false`. Therefore it would be desirable to add a warning to NEST that is displayed whenever a two-way connection type is called with `symmetric = false`. This warning should remind the user, that the `Connect`-call does not create an accurate connection and that the other direction of the connection needs to be created with either a second call to `Connect` or by setting `symmetric = true` first.

heplesser:
@janhahne I wonder if we shouldn't throw an exception instead of just issuing a warning. My experience is that warnings tend to be overlooked.

Would you create a PR for this?

janhahne:
@heplesser I agree with you. An exception is the better choice. There is however one tricky case: All-to-all connections are symmetric by default. So we need to allow the creation of all-to-all connections with `symmetric=false`.

I created a PR (#524) which implements this and hopefully resolves this issue.

heplesser:
This was fixed by #524, therefore closing.

PR #524, opened 2016-10-19
Implemented requires_symmetric flag for synapse models

janhahne:
This PR is an addition to PR #315 and fixes issue #440. It adds an `requires_symmetric` flag to the ConnetorModel class. This way synapse models can be registered as two-way connections that require symmetric connections (1->2 and 2->1). If a connection of such a synapse model is created without setting the `symmetric` flag of the connection dicitonary to `true` an exception is thrown. All-to-all connections can be created without setting `symmetric` to `true` as they are symmetric by definition (and we anyway do not allow symmetric all-to-all connections for good reasons).

So far gap junctions are the only two-way connections that use this new flag.


janhahne:
`test_labeled_synapses.py` fails, as it creates non-symmetric one-to-one gap-junction connections. Not sure about the best fix for this. Gap junctions probably should not be excluded from this test. Can anyone have a look at this?

jougs:
I have a question and a possible solution.

Shouldn't the `requires_symmetric` flag be also included in the status dictionary of the synapse such that users can find out about its value using `GetDefaults` and `GetStatus`?

If that was the case, you could just query the model in the conflicting test and add `'symmetric': True|False` to the rule dictionary depending on the value of that property.

janhahne:
@jougs Thank you for the suggestion of this solution! Hopefully the PEP8 errors are gone with the latest commit.

nestkernel/conn_builder.cpp
```diff
@@ -382,9 +382,18 @@ nest::ConnBuilder::change_connected_synaptic_elements( index sgid,
 void
 nest::ConnBuilder::connect()
 {
+  if ( kernel().model_manager.connector_requires_symmetric( synapse_model_ )
+    && not symmetric_ && not is_symmetric() )
```
heplesser: @janhahne It seems to me we get a bit of an inflation here: `symmetric_`, `is_symmetric()` and `supports_symmetric()`. Could we at least integrate `symmetric_` and `is_symmetric()`?

nestkernel/conn_builder.h
```diff
@@ -240,6 +246,12 @@ class AllToAllBuilder : public ConnBuilder
   {
   }
 
+  bool
+  is_symmetric() const
+  {
+    return true;
+  }
```
heplesser: I think this is not always true. If all-to-all has a randomized synapse parameter, it is no longer automatically symmetric. We will still have a backward connection for each forward connection, but their properties will differ.

nestkernel/conn_builder.cpp
janhahne: @heplesser I agree that this is a bit unfortunate, but all three of them serves a purpose:

-  `symmetric_` is the property of the created connection (is `symmetric` set to `true` or `false`?)
- `supports_symmetric` determines if a specific ConnBuilder supports the creation of symmetric connection. This is only implemented for the OnetoOne builder.
- `is_symmetric` is needed to allow all-to-all connections without setting `symmetric` to `true`. Otherwise the creation of all-to-all connection would not be possible at all for gap-junctions (all-to-all does not support `symmetric=true`)

Unless we want to prohibit any kind of all-to-all connections for gap-junctions I do not see any way of reducing the number of symmetric parameters. Or is there any solution that I am not thinking of?

nestkernel/conn_builder.h
janhahne: @heplesser Good catch, I did not think of that. Is there any easy way to determine if all-to-all uses randomized synapse parameters? I am thinking of a solution where this functions first checks if there are any randomized parameters involved (if so throws an `BadProperty`) and (otherwise) then returns `true`.

nestkernel/conn_builder.cpp
heplesser: @janhahne I wonder if we should take a step back and think this through once more. All-to-all is inherently symmetric, provided source and target populations are identical and all synapses are created with the same parameters. Guaranteeing symmetric connections when randomized parameters are used would be very difficult, and for explicit parameter arrays it would require very cumbersome checks for matrix symmetry.

So I would think the best approach might be to prohibit both array and random parameters if the synapse model is derived from `gap_junction`, and to also require that source and target populations are identical. Then, you get symmetry for free. @jougs Is it possible to compare two GIDCollections for equality?

jougs:
@heplesser: currently you can't compare `GIDCollections`, but that should be easy to implement. I can probably look into this next week.

nestkernel/conn_builder.cpp
janhahne: @heplesser Good points! I think now we found all conditions that need to be checked in order to ensure that an all-to-all connection is symmetric. I would also vote for a solution where we prohibit both array and random parameters, rather than allowing those in general and check them for symmetry case by case.

So I would imagine our solution like this: Change the `is_symmetric()` function of the all-to-all ConnBuilder to
```
bool
is_symmetric() const
{
  if( sources != targets )
    return false;

  if( parameters are arrays or randomized )
    return false;
 
  return true;
}
```
and adjust the `BadProperty`-Message to `"This synapse model requires symmetric (or suitable uniform all-to-all) connections"`

The check in `is_symmetric()` will only be performed, if the first two conditions of the if-statement are met, ergo only if a synapse model requires symmetric connections and the `symmetric` flag is set to `false`

Would you agree on this general plan? In this case we would need to figure out how to implement the two `if` statements. @jougs could help with this first one, any idea how to check for the second one most efficiently?

heplesser: @janhahne @jougs This looks good, although I think that we could simplify the method to
```C++
inline bool
is_symmetric() const
{
  return sources == targets && all_parameters_scalar();
}
```
The other method would be something like
```C++
bool
ConnBuilder::all_parameters_scalar_() const
{
  bool all_scalar = weight_->is_scalar() && delay_->is_scalar();
  for ( ConnParameterMap::const_iterator it = synapse_params_.begin() ; ... ; ++it )
  {
     all_scalar = all_scalar && it->second->is_scalar();
  }
  return all_scalar;
}
```
We then need in `ConnParameter` and extra method `is_scalar()` that returns true only for `Scalar{Double,Integer}Parameter`.  We cannot use existing methods since `is_array()` returns false for random numbers and `number_of_values()` returns 0 for scalar and random.

janhahne:
@heplesser @jougs Implemented our solution. Added tests to the `symmetric_connection` unittest. For some reason the comparison of the ` GIDCollections` works fine already?

I would consider this PR as complete now, please have a look

testsuite/unittests/test_symmetric_connections.sli
```diff
@@ -149,4 +151,91 @@ fail_or_die
...
+{
+  ResetKernel
+  
+  /hh_psc_alpha_gap 4 Create
+  
+  /set1 1 2 cvgidcollection def
+  /set2 3 4 cvgidcollection def
+} def
+
+% Check that connections that require_symmetric cannot be created without symmetric=true
```
heplesser: Shouldn't there be one test checking that one-to-one Connect with gap_junction actually creates the expected connections? Or is that elsewhere? All further tests are just fail/pass, but do not check the result.

testsuite/unittests/test_symmetric_connections.sli
```diff
@@ -29,14 +29,16 @@ Synopsis: (test_symmetric_connections) run -> NEST exits if test fails
 
 Description: 
 This test ensures that the functionality to create symmetric connections
-works properly.
+works properly. It also ensures that the flag requires_symmetric works properly.
```
heplesser: This is confusing, since `requires_symmetric` is built-in property of synapse models, not a flag visible on the SLI level. On the other hand, we have the `/symmetric` flag very visible in this test.

nestkernel/conn_builder.cpp
```diff
@@ -386,7 +386,8 @@ nest::ConnBuilder::connect()
     && not symmetric_ && not is_symmetric() )
   {
     throw BadProperty(
-      "This synapse model requires symmetric (or all-to-all) connections" );
+      "This synapse model requires symmetric (or suitable uniform "
+      "all-to-all) connections" );
```
heplesser: "suitable" is unclear. How about "Connections with synapse model {0} can only be created as one-to-one connections with "make_symmetric" set to true or as all-to-all connections with equal source and target populations and default or scalar parameters."?

nestkernel/conn_builder.cpp
```diff
@@ -386,7 +386,8 @@ nest::ConnBuilder::connect()
     && not symmetric_ && not is_symmetric() )
```
heplesser: I find this still confusing, and I fear anyone reading this in five years time will not feel enlightened. How about renaming `symmetric_` to `make_symmetric_` and turning the order of checks a little:
```C++
if ( kernel().model_manager.connector_requires_symmetric( synapse_model_ )
     and not ( is_symmetric() or make_symmetric_ ) )
```
That seems more readable to me.

testsuite/unittests/test_symmetric_connections.sli
heplesser: @janhahne Sorry, my oversight!

janhahne:
@heplesser I agree, `make_symmetric` is less confusing. Addressed all your change requests. Please have another look

pynest/nest/tests/test_labeled_synapses.py
```diff
@@ -142,17 +156,21 @@ def test_SetLabelToNotLabeledSynapse(self):
...
+
             # try set a label during SetDefaults
             with self.assertRaises(nest.NESTError):
                 nest.SetDefaults(syn, {'synapse_label': 123})
 
             # try set on connect
             with self.assertRaises(nest.NESTError):
-                nest.Connect(a, a, {"rule": "one_to_one"}, {
-                             "model": syn, "synapse_label": 123})
+                nest.Connect(a, a, {"rule": "one_to_one", "make_symmetric": symm},
```
heplesser: This line is a little too long according to Travis/PEP8.

janhahne:
@heplesser Should be fixed now. Travis ended one test after 18s with error `Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/main/source/Sources`. That is the reason for the remaining failed test...

PR #535, opened 2016-11-01
Use NEST variable (in ms) for printing acc. latency

mdjurfeldt:


mdjurfeldt:
Requested changes committed.

Issue #406, opened 2016-06-23
PyNEST does not handle NEST error messages with non-ASCII characters correctly


heplesser:
PyNEST does not handle error messages containing non-ASCII characters correctly, at least under Python 2.7. 

This is an example of a properly handled error (shortened):

```
In [2]: nest.sli_run('foo')
NESTError                                 Traceback (most recent call last)
[snip]
NESTError: DictError in SLIInterpreter::execute: Key '/foo' does not exist in dictionary.
```

and here an example that breaks: 

```
In [3]: nest.sli_run('føø')
UnicodeDecodeError                        Traceback (most recent call last)

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128)
```

Note that SLI handles this fine:

```
SLI ] føø

Jun 23 15:45:59 SLIInterpreter::execute [Error]: DictError
    Key '/føø' does not exist in dictionary.
```

This issue was originally mentioned in #405.

stinebuu:
PyNEST manages to handle the error correctly with Python 3, we are looking at the problem with Python 2.7 now.

heplesser:
#536 will fix this.

hakonsbm:
@heplesser This issue has been fixed by #536.

PR #536, opened 2016-11-01
Fix for #406 (ascii issue)

hakonsbm:
PyNEST will now raise a `NESTError` when trying to use non-ascii characters under Python 2.7.

We did not manage to find a way to make PyNEST work with non-ascii characters under Python 2.7. Note that this works under Python 3.

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -88,9 +88,11 @@ def catching_sli_run(cmd):
         def decode(s):
             return s
     else:
-        def encode(s): return s.encode('utf-8')
+        def encode(s):
+            return s.encode('utf-8')
 
-        def decode(s): return s.decode('utf-8')
+        def decode(s):
+            return s.decode('utf-8')
     engine.run('{%s} runprotected' % decode(cmd))
     if not sli_pop():
         errorname = sli_pop()
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -93,6 +93,7 @@ def catching_sli_run(cmd):
 
         def decode(s):
             return s.decode('utf-8')
+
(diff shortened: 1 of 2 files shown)


jougs:
@hakonsbm: can you please remove the trailing whitespace in pynest/pynestkernel.pyx? Thanks!

heplesser:
@apeyser Could you have a look at this one and see if it is ready for a second thumb up?

apeyser:
I'd suggest wrapping up
````
if VERSION(3):
def encode(s): return s.encode('utf-8')
def decode(s): return s.decode('utf-8')
else:
... for python2.7
````

just to remove the repeated ifs scattered through the code.

**Comment by hakonsbm** [comment]
@apeyser Something like this?

File: pynest/nest/__init__.py [hunk 1]
```diff
@@ -88,9 +88,11 @@ def encode(s):
         def decode(s):
             return s
     else:
-        def encode(s): return s.encode('utf-8')
+        def encode(s):
+            return s.encode('utf-8')
 
-        def decode(s): return s.decode('utf-8')
+        def decode(s):
+            return s.decode('utf-8')
```
apeyser: Add a newline after this

pynest/nest/__init__.py
```diff
@@ -82,19 +82,21 @@ def catching_sli_run(cmd):
     """
 
     if sys.version_info >= (3,):
-        engine.run('{%s} runprotected' % cmd)  # Python 3
+        def encode(s): return s
```
apeyser: Precisely like this. The rest of the code is simplified and clear.

hakonsbm:
@apeyser Fixed.

PR #540, opened 2016-11-02
Refactoring ConnBuilder

hakonsbm:
- More systematic use of `names::`
- Remove parameter skipping on synapse deletion

I propose @sdiazpier and @heplesser as reviewers.


nestkernel/connector_model_impl.h
```diff
@@ -140,7 +140,7 @@ GenericConnectorModel< ConnectionT >::get_status( DictionaryDatum& d ) const
   default_connection_.get_status( d );
 
   ( *d )[ names::receptor_type ] = receptor_type_;
-  ( *d )[ "synapsemodel" ] = LiteralDatum( name_ );
+  ( *d )[ names::synapse_model ] = LiteralDatum( name_ );
   ( *d )[ "requires_symmetric" ] = requires_symmetric_;
   ( *d )[ "has_delay" ] = has_delay_;
```
heplesser: Why are these two still given as strings?

PR #541, opened 2016-11-02
Improve iaf_psc_alpha_multisynapse and iaf_psc_exp_multisynapse models

### LINKED ISSUE #472: Review and improve iaf_psc_*_multisynapse models
The `iaf_psc_*_multisynapse` models need review. Once #261 is merged, `gif_psc_exp_multisynapse` should be taken as reference for the review. Note in particular:
-  While all other neuron models require only that the refractory time is non-negative, `iaf_psc_*_multisynapse` require that the refractory time is at least one time step (see `calibrate()` method). This inconsistency should be removed and the multisynapse models should also accept zero refractory time.
- `handle(SpikeEvent&)` uses a loop to find the spike buffer to deliver to. This can be selected by direct indexing using `e.get_rport() - 1` as index.

hakonsbm:
This fixes #472 
- Improved input handling
- Consistent use of `names::`
- Better variable names

I propose @heplesser and @golosio as reviewers.


golosio:
@hakonsbm Sure, I can review your PR next week.

golosio:
in iaf_psc_alpha_multisynapse.h line 140:
`
    /** Time constants of synaptic currents in ms. */
     std::vector< double > tau_syn_;
 
     // type is long because other types are not put through in GetStatus
std::vector< long > receptor_types_;
`
it seems that the receptor_types member is not used anywhere, so we decided to drop it, see:
https://github.com/nest/nest-simulator/pull/516#discussion_r83170793
I think the same is true for this model.
In iaf_psc_alpha_multisynapse.h line 141:
`
size_t num_of_receptors_;
`
for being consistent with other names we decided it is better to call it n_receptors.
Also, since it is always the same as tau_syn_.size() , we don't need to have it as a member, it can be replaced by an inline function as:
`
inline size_t n_receptors() { return tau_syn_.size(); }
`
see:
https://github.com/nest/nest-simulator/pull/516#discussion_r83172171

models/iaf_psc_alpha_multisynapse.cpp
```diff
@@ -122,53 +122,73 @@ iaf_psc_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
+  {
+    throw BadProperty( "Capacitance must be strictly positive." );
+  }
   if ( Tau_ <= 0. )
-    throw BadProperty( "Membrane time constant must be > 0." );
-
+  {
+    throw BadProperty( "Membrane time constant must be strictly positive." );
+  }
   std::vector< double > tau_tmp;
```
golosio: You do not need to define a temp variable. This is already done by member function
`iaf_psc_alpha_multisynapse::set_status( const DictionaryDatum& d )`
(line 300 in the h file), which creates a temporary copy of parameters and state variables, and restore them if you throw an exception.

models/iaf_psc_alpha_multisynapse.cpp
```diff
@@ -122,53 +122,73 @@ iaf_psc_alpha_multisynapse::Parameters_::set( const DictionaryDatum& d )
...
   if ( updateValue< std::vector< double > >( d, "tau_syn", tau_tmp ) )
   {
     for ( size_t i = 0; i < tau_tmp.size(); ++i )
     {
       if ( tau_tmp.size() < tau_syn_.size() && has_connections_ == true )
```
golosio: This check does not depend on the loop index, so it does not need to be done for every step. It should be done before (outside of) the loop.

models/iaf_psc_alpha_multisynapse.cpp
```diff
@@ -177,18 +197,22 @@ void
...
 }
 
 void
 iaf_psc_alpha_multisynapse::State_::set( const DictionaryDatum& d,
   const Parameters_& p,
   const double delta_EL )
 {
-  if ( updateValue< double >( d, names::V_m, y3_ ) )
-    y3_ -= p.E_L_;
+  if ( updateValue< double >( d, names::V_m, V_m_ ) )
```
golosio: I know you already did it before, but I think the code would be more clear if you write again a short comment explaining why this `if` condition is used.

golosio:
@hakonsbm I requested just some minor changes for iaf_psc_alpha_multisynapse . The same hold for iaf_psc_exp_multisynapse as well.

hakonsbm:
@golosio I have performed the requested changes, if you want to take a look.

golosio:
All good, thanks @hakonsbm, for me this PR can be merged.

PR #542, opened 2016-11-02
Better handling of exceptions during update (fixes #311)

### LINKED ISSUE #311: No reset of simulating_ bool after error and ResetKernel
On NEST 2.10.0 I'm doing many consecutive runs to test the behaviour of my models with various parameters. Some of the combinations of parameters lead to numerical instability in `aeif` models so I wanted to catch this error and use `ResetKernel` to continue the loop over the remaining parameters.

The error is:
`NumericalInstability in Simulate_d: NEST detected a numerical instability while updating gp_aeif_psc_alpha.`

However, though it is properly catched, after the `ResetKernel`, I get the following error when trying to perform `nest.SetKernelStatus({"resolution": 0.1})`:

```
Apr 19 11:17:50 Network::clear_models [Info]: 
    Models will be cleared and parameters reset.
python: ../../.nest-2.10.0/nestkernel/scheduler.h:649: const nest::Time nest::Scheduler::get_time() const: Assertion `!simulating_' failed.
[my-pc:19489] *** Process received signal ***
[my-pc:19489] Signal: Aborted (6)
[my-pc:19489] Signal code:  (-6)
[my-pc:19489] [ 0] /usr/lib/libpthread.so.0(+0x10e80)[0x7fdc5c0d9e80]
[my-pc:19489] [ 1] /usr/lib/libc.so.6(gsignal+0x38)[0x7fdc5bd5b2a8]
[my-pc:19489] [ 2] /usr/lib/libc.so.6(abort+0x16a)[0x7fdc5bd5c72a]
[my-pc:19489] [ 3] /usr/lib/libc.so.6(+0x2c1b7)[0x7fdc5bd541b7]
[my-pc:19489] [ 4] /usr/lib/libc.so.6(+0x2c262)[0x7fdc5bd54262]
[my-pc:19489] [ 5] /home/me/.nest-2.10.0-install/lib/python3.5/site-packages/nest/pynestkernel.so(_ZNK4nest9Scheduler10get_statusER12lockPTRDatumI10DictionaryXadL_ZN14SLIInterpreter14DictionarytypeEEEE+0x799)[0x7fdc45308529]
[my-pc:19489] [ 6] /home/me/.nest-2.10.0-install/lib/python3.5/site-packages/nest/pynestkernel.so(_ZN4nest7Network10get_statusEm+0x71)[0x7fdc452eb651]
[my-pc:19489] [ 7] /home/me/.nest-2.10.0-install/lib/python3.5/site-packages/nest/pynestkernel.so(_ZNK4nest10NestModule19GetStatus_iFunction7executeEP14SLIInterpreter+0x4b)[0x7fdc452cb03b]
[my-pc:19489] [ 8] /home/me/.nest-2.10.0-install/lib/python3.5/site-packages/nest/pynestkernel.so(_ZN14SLIInterpreter8execute_Em+0x315)[0x7fdc45345da5]
[my-pc:19489] [ 9] /home/me/.nest-2.10.0-install/lib/python3.5/site-packages/nest/pynestkernel.so(_ZN14SLIInterpreter7executeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x138)[0x7fdc453462d8]
```

or later if I don't use it:

```
Apr 19 12:16:00 Scheduler::prepare_nodes [Info]: 
    Please wait. Preparing elements.
python: /home/me/.nest-2.10.0/nestkernel/scheduler.h:649: const nest::Time nest::Scheduler::get_time() const: Assertion `!simulating_' failed.
```

I don't really get what's going wrong since [`nest::SimulationManager::initialize()`](https://github.com/nest/nest-simulator/blob/7537a6b68a693eefe375176e2eac31276f08ee77/nestkernel/simulation_manager.cpp) should set `simulating_` to `False`...

Here is an example of the code I used, **the problem does not occur if the multimeter is not present** (even if `nest.SetKernelStatus({"resolution": 0.1})` is used):

``` python
import nest
nest.SetKernelStatus({'local_num_threads': 14})

di_param = {
    'V_reset': -58.,
    'V_peak': 0.0,
    'V_th': -50.,
    'I_e': 300.,
    'g_L': 9.,
    'tau_w': 300.,
    'E_L': -70.,
    'Delta_T': 2.5,
    'a': 2.,
    'b': 50.,
    'C_m': 200.,
    'V_m': -60.,
    'w': 0.,
    'E_ex': 0.,
    'tau_syn_ex':0.02
}

val_c = (20., 150., 200.)

for val in val_c:
    nest.ResetKernel()
    nest.SetKernelStatus({'local_num_threads': 14})
    di_param["C_m"] = val
    gids = nest.Create("aeif_cond_alpha", 1000, di_param)
    recordables = [ str(x) for x in nest.GetStatus((gids[0],), 'recordables')[0] ]
    nest.Connect(gids, gids, conn_spec={'rule': 'fixed_indegree', 'indegree': 100, 'autapses': False}, syn_spec={'weight':10.})
    mm_param = {
         'record_from': ["V_m"],
         'interval': 0.1,
         'to_accumulator': True
    }
    mm = nest.Create("multimeter", params=mm_param)
    nest.Connect(mm, gids)
    try:
        nest.Simulate(200.)
    except Exception as e:
        print(e)
    print("done")
```

heplesser:
SimulationManager now flagged as in inconsistent state after exceptions during update. Inspection is still possible, but simulation cannot be continued. This fixes #311. See issue-311.sli test for logic of what should and should not work.

Also removed `terminate_` flag, since this is handled properly via exceptions. `terminate_` did not play any actual role in handling exceptions, since it was bypassed by throws. It only had a role in terminating upon a Posix signal or an error in SLINeuron, but also there exceptions would have been generated in the end. Now, exceptions are generated right away.

I propose @jougs and @apeyser as reviewers.


heplesser:
The changes I made to `SLIsignalflag` handling may have been too brutal, I will review and post an  updated PR shortly if necessary.

nestkernel/simulation_manager.cpp
```diff
@@ -40,16 +40,16 @@
...
-  , clock_( Time::tic( 0L ) )
+  : clock_( Time::tic( 0L ) )
   , slice_( 0L )
   , to_do_( 0L )
   , to_do_total_( 0L )
   , from_step_( 0L )
   , to_step_( 0L ) // consistent with to_do_ == 0
   , t_real_( 0L )
-  , terminate_( false )
+  , simulating_( false )
```
apeyser: We should try to simplify this multivariable state machine, and we should document what the state graph is.

nestkernel/simulation_manager.h
```diff
@@ -151,13 +145,15 @@ class SimulationManager : public ManagerInterface
   delay to_step_;         //!< update clock_+from_step<=T<clock_+to_step_
   timeval t_slice_begin_; //!< Wall-clock time at the begin of a time slice
   timeval t_slice_end_;   //!< Wall-clock time at the end of time slice
-  long t_real_;    //!< Accumunated wall-clock time spent simulating (in us)
-  bool terminate_; //!< Terminate on signal or error
+  long t_real_;     //!< Accumunated wall-clock time spent simulating (in us)
+  bool simulating_; //!< true if simulation in progress
```
apeyser: Same state machine comment here: what are the possible states of the object? Maybe a single enum is needed here with the state nodes, and a description of the transitions between nodes?

testsuite/regressiontests/issue-311.sli
```diff
@@ -0,0 +1,75 @@
...
+Name: testsuite::issue-311 Ensure NEST behaves well after exception during update
+
+Synopsis: (issue-311) run -> NEST exits if test fails
+
+Description:
+This test ensures that NEST behaves properly when a model triggers and
+exception during update.
+
+Author: Hans Ekkehard Plesser, 2016-11-02
+ */
```
apeyser: We could even add: https://github.com/nest/nest-simulator/issues/311 to help people find the original issue...

heplesser:
@apeyser I will try to address the issues you mentioned, a proper state graph with well-defined (and well-described) transitions is certainly useful. I am a little surprised you approved the PR, your comments sound rather like changes are requested?

apeyser:
@heplesser: I was leaving it open on whether those changes should be in the same issue or another. The state transitions may be needed for splitting simulation loop into three pieces, so I'm still thinking about it -- and will probably do it for that issue if it's not done here.

heplesser:
I have now re-instated the handling of POSIX signals in case anyone should use them. They are only available if NEST was compiled without MPI. Signal handling is not automatically tested, since that is close to impossible with our current test setup (put nest into a long-running simulate loop, then send SIGUSR1 from the outside). I created #554, #555 and #556 as follow-ups for the points @apeyser raised.

@apeyser @jougs Would you take a look to see if you are content?

heplesser:
@jougs I have now fixed the merge conflicts.

jougs:
@heplesser: can you please provide a short summary for the PR that can be used in the release notes and as commit message for the squash&merge commit? Thanks!

apeyser:
@jougs @heplesser When I worked on the code for interrupting and continuing runs, the "implicit state machine" was a real problem. The fix of properly documenting the states should probably go into that when I manage to put together my PR, but I'd rather this go ahead so that I we don't have a future merge mess if both PRs are sitting in the queue simultaneously.

heplesser:
@jougs One liner for release notes: "NEST will no longer allow simulations to be resumed after a C++ exception has been raised during simulation. This ensures consistency."

Issue #338, opened 2016-05-11
Eliminate duplicity between build.sh and check_code_style.sh


heplesser:
Currently, developers use `check_code_style.sh` to check style compliance on their own computers, while Travis checks style compliance with different code in `build.sh`. This risks that checks differ. `build.sh` should use `check_code_style.sh`, or both should use the same helper script to do the work.

We should also consider to terminate building if the style check fails for faster turn-around and lower Travis load.

jougs:
This was fixed by #546 and #789.

PR #546, opened 2016-11-13
Enhance and improve robustness of the Travis CI log parsing.

gtrensch:
This pull request resolves issue #392. The major changes are:
- decouple 'parse_travis_log.py' from 'bash.sh' as good as it can be done for log parsers by introducing message numbers
- re-worked 'parse_travis_log.py' and improved 'bash.sh' as the parsing process had some difficulties with catching all messages and files correctly
- beautification of the build report and static code analysis print out

Some remarks:
- The existing implementation ignores VERA++ and cppcheck messages. Those messages will not cause the build to fail. I kept this behavior as is but added some code which can easily be activated.
- It is probably desirable to have a build report in HTML format. This has been already discussed and the code is structured in a way that this would be easily possible.


.travis.yml
```diff
@@ -15,50 +37,57 @@ env:
     - xTHREADING=1 xMPI=0 xGSL=0 xLTDL=1 xREADLINE=1 xPYTHON=0 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # only threading
     - xTHREADING=1 xMPI=1 xGSL=1 xLTDL=1 xREADLINE=1 xPYTHON=1 xMUSIC=1 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # full
 matrix:
-  # do notify immediately about it when a job of a build fails.
+  # Do notify immediately about it when a job of a build fails.
```
jougs: Change to `# Notify immediately when a job of a build fails`

.travis.yml
```diff
@@ -15,50 +37,57 @@ env:
...
   - echo $PATH
 
-  # update package repository status (-qq is more quiet)
+  # Update package repository status (-qq is more quiet).
   - sudo rm -rf /var/lib/apt/lists/*
   - ls /etc/apt/sources.list.d/
   - sudo apt-get update -qq
 
-  # remove any obsolete libraries
+  # Remove any obsolete libraries.
```
jougs: Change to `# Remove automatically all unused package`

(this is the text, which the the help for `apt-get autoremove` prints)

.travis.yml
```diff
@@ -15,50 +37,57 @@ env:
...
   - sudo apt-get autoremove
 
-  # from installation webpage http://www.nest-simulator.org/installation-2/#Standard_configuration
+  # Install packages required by NEST.
+  # (Please see also: http://www.nest-simulator.org/installation-2/#Standard_configuration )
```
jougs: The link is broken. This is the correct one:
`http://www.nest-simulator.org/installation/#standard-configuration`

.travis.yml
```diff
@@ -15,50 +37,57 @@ env:
...
 
-  # Install pip and cython
+  # Install pip and Cython.
   - wget https://bootstrap.pypa.io/get-pip.py
   - sudo python get-pip.py
   - sudo pip install -U setuptools
   - pip install --user cython==0.23.4
 
-  # MPI
+  # Install package 'terminaltables' required by 'parse_travis_log.py' to create the build summary.
```
jougs: If you use quotes around the packages here, quotes probably should also be used for command and packages everywhere else in the file. Alternatively remove them here, as the meaning is also clear without them.

.travis.yml
```diff
@@ -15,50 +37,57 @@ env:
...
+  - sudo pip install terminaltables
+
+  # Install OpenMPI.
   - sudo apt-get install -y openmpi-bin libopenmpi-dev
 
-  # python testsuite nosetests
+  # Install Python testsuite nosetests.
```
jougs: Change to `Install nose framework for the Python testsuite`

build.sh
```diff
@@ -1,25 +1,46 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+# This shell script is part of the NEST Travis CI build and test environment.
+# It is invoked by the top-level Travis script '.travis.yml'.
+#
+# NOTE: This shell script is pysically coupled to python script
```
jougs: `pysically` -> `physically`, however, I'd replace `physically` by `tightly`.
`python` -> `Python`

build.sh
```diff
@@ -62,128 +83,137 @@ else
...
-    echo "======= Static analysis on file $f ======="
+    echo "MSGBLD0120: Perform static code analysis for file $f."
     case $f in
       *.h | *.c | *.cc | *.hpp | *.cpp )
         f_base=$NEST_VPATH/reports/`basename $f`
-        # Vera++ checks the specified list of rules given in the profile
-        # nest which is placed in the <vera++ root>/lib/vera++/profile
 
-        echo "\n======= - vera++ for $f ======="
+        echo "\n"
```
jougs: Do we need the new line here? There's none for the other messages.

build.sh
```diff
@@ -62,128 +83,137 @@ else
...
           format_error_files="$format_error_files $f"
         fi
 
         ;;
       *.py )
-        echo "======= Check PEP8 for $f ======="
-
-        # Ignore those PEP8 rules
+        echo "MSGBLD0190: Run PEP8 check for file: $f"
+        # Ignore those PEP8 rules.
```
jougs: Change to `# PEP8 rules to ignore`

build.sh
```diff
@@ -193,30 +223,38 @@ if [ "$xSTATIC_ANALYSIS" = "1" ] ; then
...
+            echo "MSGBLD0195: PEP8 $line"
+          done
           format_error_files="$format_error_files $f"
         fi
-        echo "======= Check PEP8 end ======="
+        echo "MSGBLD0200: PEP8 check for file $f completed."
         ;;
       *)
-        echo "$f : not a C/CPP/PY file. Do not do static analysis / formatting checking."
+        echo "MSGBLD0210: File $f is not a C/CPP/PY file. Static code analysis and formatting check skipped."
```
jougs: Change `CPP` to `C++` and `PY` to `Python` so that the sentence refers to languages rather than file name extensions.

build.sh
```diff
@@ -193,30 +223,38 @@ if [ "$xSTATIC_ANALYSIS" = "1" ] ; then
...
 
   if [ "x$format_error_files" != "x" ]; then
-    echo "There are files with a formatting error: $format_error_files ."
+    printf '%s\n' "$format_error_files" | while IFS= read -r line
+    do
+      echo "MSGBLD0220: Formatting error in file: $line"
+    done
   fi
-fi # if [ "$xSTATIC_ANALYSIS" = "1" ] ; then
+fi   # static code analysis
```
jougs: `static` -> `Static`

extras/parse_travis_log.py
```diff
@@ -20,396 +20,894 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-This script parses the TravisCI output as it is generated
-by ./build.sh and outputs a shorter summary. It is hard-coded
-to the output of ./build.sh, make installcheck,
-and cmake -DCMAKE_INSTALL_PREFIX=... . Changing any of those, requires
-adapting this script.
+This python script is part of the NEST Travis CI build and test environment.
```
jougs: `python` -> `Python`

extras/parse_travis_log.py
```diff
@@ -20,396 +20,894 @@
...
+    Read the NEST Travis CI build log file and return 'True' in case both
+    messages are found in correct order. Return 'False' if only the first
+    message was found. Return 'None' in case the first or both messages
+    are not contained in the file.
+
+    :param log_filename:         NEST Travis CI build log file name.
+    :param msg_start_of_section: Message number string, e.g. "MSGBLD1234"
+    :param msg_end_of_section:   Message number string, e.g. "MSGBLD1234"
+    :return:                     True, False or None
+    """
```
jougs: Very nice!

We usually use the [NumPy documentation style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for our Python docstrings. Could you please reformat the ones in this script for consistency reasons? Thanks!

extras/parse_travis_log.py
```diff
@@ -20,396 +20,894 @@
...
+    pair_found = None
+    for line in open(log_filename, "r"):
```
jougs: Rather use Python's [`with`](http://preshing.com/20110920/the-python-with-statement-by-example/) statement here. That also closes the file again when all reading is done.

extras/parse_travis_log.py
```diff
@@ -20,396 +20,894 @@
...
-        print(INDENT + ' | ' + k1 + ':')
-        for v2 in v1:
-            print(INDENT + ' | ' + INDENT + v2)
+    Convert a boolean value, e.g. the value returned by
+    is_message_pair_in_logfile(), into a meaningful string representation.
 
+    :param value:  Boolean value: True, None or False
+    :return:       "Passed Successfully", "Skipped" or "Failed" (default)
+    """
+    if value is True:
```
jougs: From the [Python Style Guide](https://www.python.org/dev/peps/pep-0008/#id49):
>Don't compare boolean values to True or False using == .
>```
>Yes:   if greeting:
>No:    if greeting == True:
>Worse: if greeting is True:
>```

extras/parse_travis_log.py
```diff
@@ -20,396 +20,894 @@
...
+    header = '\n+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + ' \
+             '+ + + + + + + + + +' + \
+             '\n+                                                           ' \
+             '                  +' + \
+             '\n+           N E S T   T r a v i s   C I   B u i l d   S u m ' \
+             'm a r y           +' + \
+             '\n+                                                           ' \
+             '                  +' + \
+             '\n+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + ' \
+             '+ + + + + + + + + +'
```
jougs: Please make this nice by ignoring the line width restriction.

jougs:
Are there any problems causedl if you turn on VERA++ and cppcheck? If no, I think we should activate those tests. If there are problems in the current code, we should fix them in another PR and turn on the stricter checks only then.

gtrensch:
@jougs, many thanks for the Python hints! I addressed all your comments. I also activated the check of VERA++ and cppcheck messages.

extras/parse_travis_log.py
```diff
@@ -638,55 +716,51 @@ def printable_summary(list_of_changed_files,
...
-
-    build_summary = header + '\n\n'
+
+    header_l1 = '\n+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +'  # noqa
+    header_l2 = '\n+                                                                             +'  # noqa
+    header_l3 = '\n+           N E S T   T r a v i s   C I   B u i l d   S u m m a r y           +'  # noqa
+    header_l4 = '\n+                                                                             +'  # noqa
+    header_l5 = '\n+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +'  # noqa
+
+    build_summary = header_l1 + header_l2 + header_l3 + header_l4 + header_l5 + '\n\n'               # noqa
```
jougs: Actually, you could use a multi-line string here:
```
build_summary = """
+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +
+                                                                             +
+           N E S T   T r a v i s   C I   B u i l d   S u m m a r y           +
+                                                                             +
+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +
"""
```
I'm not sure what PEP8 says about line lengths in this case, though.

heplesser:
@gtrensch I really like the new look of the build logs. Great work! But I also see one problem, see #338: Travis will now consider cppcheck and Vera++ errors a failure and stop, while`check_code_style.sh` will only report formatting problems as proper errors. This will certainly lead to a lot of confusion and frustration for developers. I think it would be very helpful to adjust `check_code_style.sh` as well, so that local and Travis testing will be "in sync".

jougs:
@heplesser: good catch! I agree that these should be made consistent and that this PR is probably a good place to do so.

extras/parse_travis_log.py
gtrensch: @jougs: Thanks for the Python multi-line hint!  I finally made the layout of the header smaller. 4292033 The underlaying problem is the pep8 E501 rule limiting the line size to 79. Neither continuation lines nor multi-line comments allow '# noqa' to suppress pep8 checking. 
In my opinion this is a strong limitation which makes the code less readable. Readability is important for maintainability. As a result this limitation impacts code quality.

heplesser:
@gtrensch Could you take a look at sync'ing `check_code_style.sh` with Travis as part of this PR? See my and @jougs comment above.

gtrensch:
@heplesser, @jougs sorry that I did not respond. I already re-worked the scripts. To avoid maintaining two files I unified and outsourced the static code analysis part to a common script file.

heplesser:
@gtrensch Great, but are you sure you committed the common script file? I don't see it among the changed files.

gtrensch:
@heplesser  It's now committed! Testing, which meant forcing errors on Travis, was more time-consuming than I thought. Hence the delay in pushing the changes.

The commit contains the following:
- On Travis, the VERA++ NEST profile is no longer installed in the user home directory. It is now added to the VERA++ installation profiles folder. (suggested by @jougs)
- All static code analysis processing has been harmonised and moved to script '/extras/static_code_analysis.sh'. The script is executed on Travis and also called by 'check_code_style.sh'.
- The CPPCHECK part of the script did not work properly. This has been fixed.
- CPPCHECK is bad in finding system includes. Thus, option --suppress=missingIncludeSystem was added. 
- The script/tool 'check_code_style.sh' was re-worked. (added also a new option --i)

Remarks:
- With this PR we turn on all checks! CPPCHECK may produce a lot of messages which are informational, e.g. related to performance. These are filtered out by the Python script on Travis but not when running 'check_code_style.sh'. This might be confusing and perhaps a point for discussion and a topic for the next developer meeting.
- Checking for a license text is done by VERA++ and a regression test. The behavior is different.

@heplesser, @jougs Could you test 'check_code_style.sh' ? I never used it before and changed the behavior and printout a bit. Hope that it still meets the expectations.

extras/check_code_style.sh
```diff
@@ -1,279 +1,212 @@
...
-            CLANG_FORMAT="$( echo "$1" | sed 's/^--clang-format=//' )"
-            ;;
-        *)
-            usage 1 "$1"
-            ;;
-    esac
-    shift
+  case "$1" in
+      --help)
+          print_usage
```
heplesser: This should call `usage`, so that the script is exited after help has been printed.

extras/parse_travis_log.py
```diff
@@ -20,396 +20,971 @@
...
+                    if file_name not in warning_summary:
+                        warning_summary[file_name] = 0
+                    warning_summary[file_name] += 1
+                    number_of_warning_msgs += 1
+
+            if is_message(line, msg_make_section_end):
+                # The log file contains only one 'make' section, return.
+                if number_of_error_msgs == 0:
+                    return True, number_of_error_msgs, error_summary, \
+                           number_of_warning_msgs, warning_summary
```
heplesser: PEP8 discourages continuation lines. This should be 
```Python
 return (True, number_of_error_msgs, error_summary,
              number_of_warning_msgs, warning_summary)
```

extras/parse_travis_log.py
```diff
@@ -20,396 +20,971 @@
...
+                else:
+                    return False, number_of_error_msgs, error_summary, \
+                           number_of_warning_msgs, warning_summary
```
heplesser: This should be
```Python
return (False, number_of_error_msgs, error_summary,
   number_of_warning_msgs, warning_summary)

File: extras/static_code_analysis.sh [hunk 20]
```diff
@@ -0,0 +1,204 @@
...
+    echo "$2"
+  fi
+}
+
+# Print version information.
+VERA_VERS=`$VERA --version`
+CPPCHECK_VERS=`$CPPCHECK --version | sed 's/^Cppcheck //'`
+CLANG_FORMAT_VERS=`$CLANG_FORMAT --version`
+PEP8_VERS=`$PEP8 --version`
+print_msg "MSGBLD0105: " "Following tools are in use:"
```
heplesser: I think it would be helpful to offset this block with version information with at least an empty line above and below on output to make it easier to recognise. May even underline the header line.

extras/static_code_analysis.sh
```diff
@@ -0,0 +1,204 @@
...
+  if [ ! -f "$f" ]; then
+    print_msg "MSGBLD0110: " "$f is not a file or does not exist anymore."
+    continue
+  fi
+  if ! $RUNS_ON_TRAVIS && $INCREMENTAL; then
+    print_msg "" ""
+    print_msg "" "Press [Enter] to continue.  (Static code analysis for file $f.)"
+    read continue
+  fi
+  print_msg "MSGBLD0120: " "Perform static code analysis for file $f."
```
heplesser: I think the messages as printed are a bit difficult to read. I think it would be more informative to print just a single short line per file per test, e.g.,
```
Skipping: .travis.yml
Running PEP8: pynest/nest/lib/hl_api_parallel_computing.py
Running Vera++: nestkernel/event.h
Running Cppcheck: nestkernel/event.h
Running Clang-format: nestkernel.event.h
```
Maybe even with an indent so all colons are lined up.

extras/static_code_analysis.sh
```diff
@@ -0,0 +1,204 @@
...
+  if [ "x$format_error_files" != "x" ]; then
+    for f in $format_error_files; do
+      print_msg "MSGBLD0220: " "Formatting error in file: $f"
+    done
+  fi
+else
+  if [ "x$format_error_files" != "x" ]; then
+    print_msg "" ""
+    print_msg "" "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
+    print_msg "" "+                 TRAVIS CI MAY REPORT FORMAT ERRORS !                        +"
```
heplesser: I would rather write ```STATIC CODE ANALYSIS DETECTED PROBLEMS !```

extras/static_code_analysis.sh
```diff
@@ -0,0 +1,204 @@
...
+    print_msg "" "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
+    for f in $format_error_files; do
+      print_msg "" "Formatting error in file: $f"
+    done
```
heplesser: I would offset this output from the header above and the text below by empty lines for better visibility.

Maybe even better: Could you collect information on C++ and Python files separately? Then you could print something like
```
C++ files with formatting errors:
   node.h
   event.cpp
Run $CLANG_FORMAT -i <filename> for autocorrection.

Python files with formatting errors:
  a.py
  b.py
Run pep8ify -w <filename> for autocorrection.

For other problems, please correct your code according to the [VERA], [CPPC], [DIFF], and [PEP8] messages above.

Comment (heplesser) [inline extras/static_code_analysis.sh]: Or maybe one could differentiate even further: If we have detected errors that will make Travis fail, report them in a section for errors that will make Travis fail, and report other problems in a separate section. For best visibility, I think that the section with Travis-fail issues should come last.

File: extras/check_code_style.sh (hunk 17 above)
Comment (gtrensch) [inline extras/check_code_style.sh]: Yes indeed, this was the idea. Correction committed: 0683add

File: extras/parse_travis_log.py (hunk 18 above)
Comment (gtrensch) [inline extras/parse_travis_log.py]: I have corrected this. Additionally, I deactivated cppcheck again with this commit: bf46de3

File: extras/parse_travis_log.py (hunk 19 above)
Comment (gtrensch) [inline extras/parse_travis_log.py]: Commit: bf46de3, same as above.

File: extras/static_code_analysis.sh (hunk 20 above)
Comment (gtrensch) [inline extras/static_code_analysis.sh]: Commit:  	77d1197

File: extras/static_code_analysis.sh (hunk 21 above)
Comment (gtrensch) [inline extras/static_code_analysis.sh]: Commit:	9dcbac3
Please can you run some tests ?

File: extras/static_code_analysis.sh (hunk 22 above)
Comment (gtrensch) [inline extras/static_code_analysis.sh]: I re-worked that section. See the comment below.

File: extras/static_code_analysis.sh (hunk 23 above)
Comment (gtrensch) [inline extras/static_code_analysis.sh]: Commit:  	4588e37
I re-worked the whole section to be able to distinguish between C/C++ and Python files. All errors reported by the tool will also make Travis fail. I deactivated cppcheck and I will add additional command-line options to the tool to turn on and off checks. (Not done yet.) The default for cppcheck will be set to off. This will allow us to easily change the code analysis by toggle defaults, and still gives us the possibility to run additional checks not performed on Travis.

File: extras/parse_travis_log.py (hunk 18 above)
Comment (gtrensch) [inline extras/parse_travis_log.py]: For a uniform syntax I added brackest to all other return-statements too. Commit: 36136a4

File: extras/static_code_analysis.sh (hunk 23 above)
Comment (gtrensch) [inline extras/static_code_analysis.sh]: I added additional command-line options to turn on/off VERA++, CPPCHECK, CLANG-FORMAT and PEP8:
```
--perform-vera=on/off
--perform-cppcheck=on/off
--perform-clang-format=on/off
--perform-pep8=on/off
```
The behavior is now the following:
On Travis CI all checks are performed and the results are evaluated. The build will not fail if there are CPPCHECK messages. This is turned off in 'parse_travis_log.py'.
The command-line tool bypasses CPPCHECK. The default setting for '--perform-cppcheck' is 'off'. For all other checks the default is 'on'.

Commit: 79074a4

heplesser:
@gtrensch Looks great! Thank you for your efforts!

heplesser:
@jougs This is ready to merge, but I leave it to you whether to merge it before creating NEST 2.12 or after.

PR #551, opened 2016-11-21
Removed plural forms of certain parameter names (fixes #531)

### LINKED ISSUE #531: Use singular parameter names throughout
Several multisynapse models currently use plural versions of parameter names, such as `taus_syn`. The Developer VC on 31 Oct decided to standardize to singular names (`tau_syn`) throughout; there had been no protest after @heplesser's mail in this matter to NEST User on 26 October.

heplesser:
Removed plural forms of taus_* parameter names, removed duplicate and unused declarations and names from nest_names.

This fixes #531 .

I suggest @Silmathoron and @golosio as reviewers.


golosio:
@heplesser all the changes are fine for me.

heplesser:
@Silmathoron Thanks for your careful check. I checked all removed names by a project wide search, and indeed compilation would break if a name were missing. 
- I will change the Traub-model test to use `g_exc` and `g_inh` instead of the `..._conductance` names.
- I don't understand your comment about `I` and ticket-466.sli. Did you mean another ticket?
I will now try to fix the merge conflicts. I fear we may be getting into a mess here ...

Silmathoron:
Ok, good!
For ``ticket-466.sli``, unless I'm mistaken, ``mirollo_strogatz_ps`` does not exist anymore (which is why you could remove ``I``...).
Yes, this looks rather bad :s good luck with that!

heplesser:
@Silmathoron I merged, wasn't as bad as I thought. I decided to keep the Mirollo-Strogatz-related names in, I assume someone is tending the model somewhere and it might come back. Would you have another look and the approve (or not)?

Issue #520, opened 2016-10-13
Review and improve *_cond_*_multisynapse models


heplesser:
There are probably some ways to simplify the `*_cond_*_multisynapse` that should be exploited, e.g., replacing the `num_of_receptors` counter with a getter method and removing unused `receptor_type` array.

See also #472, #511, #509, #510.

Issue #509, opened 2016-10-06
Correct input handling for aeif_cond_alpha_multisynapse


heplesser:
Input handling for `aeif_cond_alpha_multisynapse` needs to be modified according to the decision made in the [NEST Developer VC 19 Sep 2016](https://github.com/nest/nest-simulator/wiki/2016-09-19-Open-NEST-Developer-Video-Conference):
- Replace `E_ex`, `E_in` with vector of reversal potentials the same size as `taus_syn` (one per synapse)
- Replace  `I_syn_ex`, `I_syn_in` recordables with one recordable `I_syn` representing the total input current
- Update class and `handle(SpikeEvent&)` to a single input buffer per rport, see [`aeif_cond_beta_multisynapse`](https://github.com/nest/nest-simulator/blob/446d340da0830edcc78ac85ae0a80e58695b3e37/models/aeif_cond_beta_multisynapse.cpp#L607) for an example; consider adding a check for non-negative weight.
- Update `*_dynamics` function to handle input current correctly, along the lines of

``` C++
  double I_syn = 0.0;
  for ( size_t i = 0; i < node.P_.num_of_receptors_ ; ++i )
  {
      I_syn += y[ S::G + i ] * ( node.P_.E_rev[i] - V );
  }
```
- Move `*_dynamics` function to cpp-file.
- Provide examples (python based).

See also #510, #511.

heplesser:
This is now implemented in PR #557.

heplesser:
Fixed by #557.

PR #557, opened 2016-11-22
Polished aeif_cond_{alpha,beta}_multisynapse

heplesser:
Polished aeif_cond_beta_multisynapse so it can serve as example for other _cond_ and _multisynapse_ models.

- Refractoriness now handled correctly (see #547), test extended
- Moved dynamics function into usual location
- Fixed some warnings
- Minor improvements to documentation

This issue partially addresses #520 and #547, and provides and example for #509 and #511.

I suggest @golosio and @Silmathoron as reviewers.


heplesser:
In this PR, I moved the spike-detection out of the `while()` loop encapsulating the ODE integration. I believe that this is the correct approach: All models except precise-timing models lock spikes to time steps. Thus, we should only check for a spike once we have integrated to the next step. This also makes handling refractoriness correctly much easier.

Silmathoron:
Ok, thanks a lot for the reference @heplesser, I'll have a look by the end of the week!

golosio:
@heplesser all the changes and also the extension of the SLI test are fine for me.

Silmathoron:
@heplesser: what is the status of this PR, now?

heplesser:
@Silmathoron I updated the model now to "inside" thresholding and also ensure that dV/dt == 0 during the refractory period.

@golosio I hope your thumbs-up is still valid with these changes?

golosio:
Sure it's valid!


2016-12-02 16:35 GMT+01:00 Hans Ekkehard Plesser <notifications@github.com>:

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -67,6 +67,67 @@ RecordablesMap< aeif_cond_beta_multisynapse >::create()
...
+  const nest::aeif_cond_beta_multisynapse& node =
+    *( reinterpret_cast< nest::aeif_cond_beta_multisynapse* >( pnode ) );
+
+  const bool is_refractory = node.S_.r_ > 0;
+
+  // Clamp membrane potential to V_reset while refractory, otherwise bound
+  // it to V_peak. Do not use V_.V_peak_ here, since that is set to V_th if
+  // Delta_T == 0.
+  const double& V =
+    is_refractory ? node.P_.V_reset_ : std::min( y[ S::V_M ], node.P_.V_peak_ );
```
Silmathoron: Since you set dV/dt to zero during the refractory period, is this test really necessary?

testsuite/unittests/test_aeif_cond_beta_multisynapse.sli
```diff
@@ -181,4 +181,145 @@ d /times get cva /ts Set % get times
...
+% test for t_ref == 0
+{ 
+  << >> begin
+  ResetKernel
+  /res 0.1 def
+  /V_reset -111. def
+  
+  0 << /resolution res >> SetStatus
+  /nrn /aeif_cond_beta_multisynapse 
+    << /w 0. /a 0. /b 0. /Delta_T 0. /t_ref 0. /I_e 1000.
```
Silmathoron: Why set all adaptation parameters to zero in the 1st 2 tests?

models/aeif_cond_beta_multisynapse.cpp
heplesser: Better safe than sorry.

heplesser:
I changed the section heading to "Adaptation parameters" and moved Delta_T, V_th and V_peak up.

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -65,6 +67,67 @@ RecordablesMap< aeif_cond_alpha_multisynapse >::create()
...
+  // not the state vector in the node, node.S_.y[].
+
+  typedef nest::aeif_cond_alpha_multisynapse::State_ S;
+
+  // get access to node so we can almost work as in a member function
+  assert( pnode );
+  const nest::aeif_cond_alpha_multisynapse& node =
+    *( reinterpret_cast< nest::aeif_cond_alpha_multisynapse* >( pnode ) );
+
+  const bool is_refractory = node.S_.r_ > 0;
```
Silmathoron: I fear we still have a misunderstanding: now, via your modification of ``S_.r_`` in ``update``, the ``t_ref = 0`` test passes, but the discrepancy between this model and the refractory LSODAR implementation still exists at spiketime because if ``t_ref > 0``, then we get the refractory condition during the "spiketime step", which we do not want since it modifies the spike trajectory.

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -516,167 +549,75 @@ aeif_cond_alpha_multisynapse::update( Time const& origin,
...
-                                       // refractory period
+        S_.y_[ State_::W ] += P_.b; // spike-driven adaptation
+
+        /* Initialize refractory step counter.
+         * - We need to add 1 to compensate for count-down immediately after
+         *   while loop.
+         * - If neuron has no refractory time, set to 0 to avoid refractory
+         *   artifact inside while loop.
+         */
+        S_.r_ = V_.refractory_counts_ > 0 ? V_.refractory_counts_ + 1 : 0;
```
Silmathoron: This is not what I meant (see above)

models/aeif_cond_beta_multisynapse.cpp
```diff
@@ -633,9 +633,13 @@ aeif_cond_beta_multisynapse::update( Time const& origin,
...
-        // initialize refractory steps, adding 1 to compensate for immediate
-        // subtraction after the while loop
-        S_.r_ = V_.refractory_counts_ + 1;
+        /* Initialize refractory step counter.
+         * - We need to add 1 to compensate for count-down immediately after
+         *   while loop.
+         * - If neuron has no refractory time, set to 0 to avoid refractory
+         *   artifact inside while loop.
+         */
+        S_.r_ = V_.refractory_counts_ > 0 ? V_.refractory_counts_ + 1 : 0;
```
Silmathoron: Corrects the ``t_ref =0`` case but not the discrepancy at spiketime when ``t_ref > 0``, see above and code in #575 (which does precisely what I intended it to do)

heplesser:
@Silmathoron Sorry I didn't fully understand your approach. After re-reading your code in #575 and posts, my understanding is now that if during step (t, t+h] the membrane potential crosses threshold at t+d with 0 < d < h, you
1. reset V_m to V_reset at t+d
1. do *not* clamp V_m during (t+d, t+h]
1. clamp V_m from t+h for refractory_count steps

This solves one problem with the "inside" approach that also the "inside_pr" approach (see #547) tries to address: With the inside approach, the neuron becomes fully refractory during (t+d, t+h], plus for refractory_count steps afterwards, so the average refractory time becomes h/2 longer than it should be. You approach addresses this by not having the neuron fully refractory during (t+d, t+h], while "inside_pr" solves it by letting the neuron return from refractoriness at t*+d instead of t*+h, where t* is the beginning of the last step during which the neuron is refractory. Note that without the reset at t+d in your approach, we would be back to the "outside" method, which is worst (due to the large V_m in (t+d, t+h]).

I tested your approach against mine with the parameter sets from the aeif_models_implementation notebook with refractory time 2 ms (both implementations give the same results for 0 ms refractory time).  For regular spiking and bursting, your and my method show very similar voltage traces. I only found significant differences between your an my approach for the "chaotic" parameters and 0.1 ms time step. In this case, the spike-time errors relative to the LSODAR reference solution are distributed equally for both approaches:

htc

Therefore, my conclusion at present is that none of the two approaches is numerically better. 

From a point of model logic, though, I find my solution more consistent, since it treats the neuron as fully refractory from the moment of threshold crossing t+d onward, while your approach puts it into a special transition state until t+h: already reset but not yet refractory.

Thus, and because my implementation is closer to implementations in other models using GSL integrators, I would suggest that we stick to my solution for the time being.

Silmathoron:
@heplesser thanks for the extensive checks, as usual!
I'm surprised the difference is so small, but that's fine with me, let's stick with your implementation.

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -318,40 +341,29 @@ aeif_cond_alpha_multisynapse::State_::set( const DictionaryDatum& d )
...
-    if ( ( g_exc.size() != dg_exc.size() ) || ( g_exc.size() != g_inh.size() )
-      || ( g_exc.size() != dg_inh.size() ) )
+    const std::vector< double > dg =
+      getValue< std::vector< double > >( d->lookup( names::dg ) );
+    const std::vector< double > g =
+      getValue< std::vector< double > >( d->lookup( names::g ) );
+
+    if ( ( dg.size() != g.size() ) )
     {
       throw BadProperty( "Conductances must have the same sizes." );
```
Silmathoron: I think this should be something like "There should be one time derivative entry for each conductance", or "The number of conductances and their first derivatives should be the same", but the arrays do not not both contain conductances.

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -318,40 +341,29 @@ aeif_cond_alpha_multisynapse::State_::set( const DictionaryDatum& d )
...
     }
 
-    for ( size_t i = 0; i < g_exc.size(); ++i )
+    for ( size_t i = 0; i < dg.size(); ++i )
     {
-      if ( ( g_exc[ i ] < 0 ) || ( dg_exc[ i ] < 0 ) || ( g_inh[ i ] < 0 )
-        || ( dg_inh[ i ] < 0 ) )
+      if ( ( dg[ i ] < 0 ) || ( g[ i ] < 0 ) )
```
Silmathoron: dg can be negative

models/aeif_cond_beta_multisynapse.h
```diff
@@ -74,7 +74,9 @@
  the synapse i is excitatory or inhibitory depending on the value of E_{rev,i}
  and the differential equation for the spike-adaptation current w is:
 
- tau_w * dw/dt = a(V - E_L) - W
+ tau_w * dw/dt = a(V - E_L) - w
+
+ When the neuron fires a spike, the adaptation current w <- w + b.
```
Silmathoron: Inserting "undergoes a step increment of value b" before the formula would make the sentence less abrupt (and possibly clearer (to me).

testsuite/unittests/test_aeif_cond_alpha_multisynapse.sli
```diff
@@ -37,12 +37,19 @@
...
+   Author: Daniel Peppicelli, revised Hans Ekkehard Plesser
  */
  
 (unittest) run
 /unittest using
 
+% This test should only run if we have GSL
+statusdict/have_gsl :: not
+{
+  exit_test_gracefully
```
Silmathoron: I love it! x)

models/aeif_cond_beta_multisynapse.h
heplesser: Fixed.

heplesser:
@golosio @Silmathoron In the `aeif_cond_*_multisynapse` models, it is currently possible to set `g` and `dg`. This complicates things a bit, since one needs to account for changes in receptor number. This can also complicates `init_state_()`, since the number of receptors in the prototype model may differ from that in then neuron being reset by `ResetNetwork`.

Therefore, I would strongly suggest to remove the possibility to change `g` and `dg`. We may re-introduce it later when generating code via NESTML. `dg` is a bit problematic anyways, since it is not dg/dt but a more complex expression.

Do you have any strong objections (i.e., would removing the possibility to change `g` and `dg` be a showstopper for your neuroscience work)?

Silmathoron:
Ok, I understand.
No, I never use this... In general I think it's better to give the user access to as many things as possible, but if you think this overcomplicates things, I don't think it's a vital feature, I would approve its removal until a satisfactory solution is found.

golosio:
Sorry, I had a strong flu and only today I'm back to work. I agree with the removal of the possibility to change g and dg, it seems that this possibility is only complicating things at the moment, and as you say we could re-introduce it later if we see it can be useful, and in such case we can look for a better way to handle such possibility.
@Silmathoron indeed long commands in SLI can become quite cryptic....
For the rest, I agree with all the changes.

heplesser:
@golosio @Silmathoron Thank you for your reviews. I have now removed the possibility to set `g` and `dg` from the `aeif_cond_{alpha,beta}_multisynapse` models. One can still read the values with `GetStatus`. Provided the checks pass and I don't here anything contrary from you, I will merge in the near future.

heplesser:
@Silmathoron Thanks for noticing my blunder! It should be fixed now.

PR #560, opened 2016-11-22
Added protection against probabilistic wiring to spike_detector (fixes #351)

### LINKED ISSUE #351: Fixed indegree connection rule creates wrong number of connections when connecting to devices on multiple MPI processes
Connecting a population of neurons to a device such as `spike_detector` using a random rule such as `fixed_indgree` or `fixed_total_number` using multiple MPI processes leads to an incorrect total number of connections. E.g., the following script

```
0 << /total_num_virtual_procs 2 >> SetStatus
/iaf_psc_alpha 1000 Create ;
/nodes 0 GetGlobalLeaves def
/sd /spike_detector Create def

nodes [ sd ] << /rule /fixed_indegree /indegree 100 >> Connect

0 /num_connections get =
```

will report 50 and 45 connections on two MPI processes, i.e., 95 in total, not the expected 100. When random seeds are changed, the connection numbers will also change.

The current work-around is not to use random connection rules when connecting to devices.

The cause of the problem is as follows:

Most (random) connection rules have a target-centric view that works fine for normal target nodes: Each target lives on exactly one VP and the VP-specific RNG is used to choose sources for the target. We must use the VP-specific RNG since the number of target neurons per VP can differ, so that different numbers of random numbers are needed.

`spike_detector` and similar devices, though, live on all VPs and `fixed_indegree` will try to create connections on each VP, but creates actual connections only for local _sources_. Thus, connections are de facto created in a source-centric fashion. Since every VP will select a different set of sources, different numbers of connections can result when running with different numbers of processes.

To avoid this, one would need to use the global random number generator when connecting to nodes without proxies. It is important (if slow) to choose the full set of sources on each VP (instead of just selecting among local sources); otherwise, different numbers of random numbers might be consumed on different VPs and the GRNG instances on the various processes would desynchronize (this is detected by NEST on return from Simulate). Essentially, this takes us back to the source-centric, non-parallel approach of `fixed_outdegree`.

This was formerly trac709.

heplesser:
This adresses #351 .

I suggest @hannahbos and @jougs as reviewers.


.gitignore
```diff
@@ -56,3 +56,4 @@ doc/normaldoc.conf
 lib/sli/rcsinfo.sli
 extras/emacs/sli.el
 extras/nest_vars.sh
+*.pyc
```
mschmidt87: Why is this part of the pull request

heplesser:
@mschmidt87 .pyc files should certainly be ignored by Git. I, as probably many other users, had `*.pyc` in my global gitignore file. After that file caused problems in a different context (not related to pyc at all), I decided not to have a global gitignore file any more. As a consequence, I needed to add `*.pyc` to project `.gitignore` files, and this branch was simply the first I committed to after I deleted by global gitignore.

mschmidt87:
Maybe I am missing something here, but isn't `*.pyc` already included in `.gitignore`, precisely in line 12? So, adding it another time shouldn't be necessary, should it?

heplesser:
@mschmidt87 I guess my 351-branch didn't contain all merges from master and thus didn't have the *.pyc on line 12. I have now removed the duplicate *.pyc.

mschmidt87:
I am not sure how picky we are in general with these things, but it would be cleaner if you would throw away the commits touching `.gitignore` e.g. by cherry-picking just the first commit of this PR. This way, `.gitignore` would remain completely untouched (currently, you're adding an empty line to the file) and we wouldn't have the unnecessary commits in the history.

hannahbos:
Greate. Thanks! :+1:

heplesser:
@mschmidt87 You are right that my changes to `.gitignore` will leave a bit of unnecessary back-and-forth in NEST history, but at present I would prioritize completion of milestone 2.12 and robustness before history neatness. I am also afraid that attempts at cherry-picking/rebasing might introduce even more of mess, so I hope you are willing to approve this PR as it is?

Issue #565, opened 2016-11-25
Structural plasticity disables valid connectors


jakobj:
Structural plasticity allows the creation and deletion of connections during runtime. In order to manage memory consumption, a heterogenenous connector containing multiple synapse types can be converted to a homogeneous connector when a deletion of a synapse leads to the situation that only a single synapse type is present in the connector (https://github.com/nest/nest-simulator/blob/master/nestkernel/connector_model_impl.h#L580 ff). With the introduction of gap junctions, packing/unpacking of pointers was introduced to hide information about the type of connection (primary, i.e., spike or secondary, e.g., gap events) in the pointer to the connector (https://github.com/nest/nest-simulator/blob/master/nestkernel/connector_model.h#L61 ff). This information is however, is only present in the pointer that is stored in the sparse table, but not in pointers that are part of heterogeneous connectors (https://github.com/nest/nest-simulator/blob/master/nestkernel/connector_model_impl.h#L443 ff).
When "downgrading" a connector from a heterogeneous connector to a homogeneous connector, the flags of the heterogeneous connector are incorrectly used to pack the pointer of the homogeneous connector (https://github.com/nest/nest-simulator/blob/master/nestkernel/connector_model_impl.h#L588) which can lead to lead to whole homogeneous connectors being ignored during delivery of spikes. In addition the homogeneous connector is cast to the wrong type, namely the type of the connector from which connections were removed (https://github.com/nest/nest-simulator/blob/master/nestkernel/connector_model_impl.h#L587).
In the example code below, a single neuron is connected to a spike detector and another neuron. The latter connection is deleted using functions from the structural plasticity framework. As a result, spike from the neuron are not transmitted to the spike detector any more, which incorrectly reports 0 received events. If the command for deleting the connection is commented out, the spike detector reports 3 recorded events.
```python
import nest

neurons = nest.Create('iaf_neuron', 2, {'I_e': 400.})
spike_detector = nest.Create('spike_detector', 1)

nest.Connect([2], [1], {'rule': 'all_to_all'}, {'model': 'static_synapse', 'weight': 0.})

nest.CopyModel('static_synapse', 'dev_syn')
nest.Connect([2], spike_detector, 'all_to_all', {'model': 'dev_syn'})

nest.Disconnect([2], [1], {'rule': 'all_to_all'}, {'model': 'static_synapse'})

nest.Simulate(100)

print nest.GetStatus(spike_detector, 'n_events')[0]
```
[...]

PR #526, opened 2016-10-21
Remove iaf_neuron

jougs:
This PR replaces all uses of `iaf_neuron` by `iaf_psc_alpha` and removes `iaf_neuron` completely from the NEST sources.

This is possible because the two were basically the same model. The only difference was the number of synaptic time constants. `iaf_neuron` only had a single _tau_syn_, while `iaf_psc_alpha` has _tau_syn_ex_ for excitatory synapses and _tau_syn_in_ for inhibitory synapses. The model was removed to avoid confusion and make simulation scripts more concise.

This PR resolves #21.

PR #566, opened 2016-11-25
Fix conversion from heterogeneous to homogeneous connector when deleting last synapse of a type

jakobj:
See #565 for a discussion of the issue. This PR fixes that issue and provides a test.
I suggest @sdiazpier and @suku248 as reviewers.

### DIFF
--- a/testsuite/unittests/issue-565.sli
+++ b/testsuite/unittests/issue-565.sli
@@ -1,0 +1,83 @@
+/*
+ *  issue-565.sli
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+ /* BeginDocumentation
+Name: testsuite::issue-565
+
+Synopsis: (issue-565) run -> NEST exits if test fails
+
+Description:
+This test provokes a switch from a heterogeneous connector to a
+homogeneous connector using the structural plasticity framework and
+asserts that the resulting homogeneous connector is correctly used.
+
+Author: Jakob Jordan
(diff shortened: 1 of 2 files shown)


testsuite/unittests/issue-565.sli
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+
+ /* BeginDocumentation
+Name: testsuite::issue-565
+
+Synopsis: (issue-565) run -> NEST exits if test fails
+
+Description:
+This test provokes a switch from a heterogeneous connector to a homogeneous connector using the structural plasticity framework and asserts that the resulting homogeneous connector is correctly used.
```
janhahne: This line could use a line break

testsuite/unittests/issue-565.sli
```diff
@@ -0,0 +1,64 @@
...
+[2] [1] << /rule /all_to_all >> << /model /static_synapse /weight 0. >> Connect
+
+% Create second connection with new synapse type, causing the
+% homogeneous connector to be turned into a heterogeneous connector
+/static_synapse /dev_syn CopyModel
+[2] [3] << /rule /all_to_all >> << /model /dev_syn >> Connect
+
+% Delete the first connection again, causing the heterogeneous
+% connector to be turned into a homogeneous connector
+[2] cvgidcollection [1] cvgidcollection << /rule /all_to_all >> << /model /static_synapse >> Disconnect_g_g_D_D
```
janhahne: Same here, maybe add a line break after `/all_to_all >>`

testsuite/unittests/issue-565.sli
```diff
@@ -0,0 +1,64 @@
...
+
+100 Simulate
+
+% The spike detector should have recorded three events
+{
+  3 GetStatus /n_events get 3 eq
+} assert_or_die
```
janhahne: I see from your example in the issue #565 how this ensures that everything works properly, but wouldn't it be much more fail-safe to add another simulation to this unit test where one neuron with the same parameters is just connected to the spike detector (without disconnecting a temporary connection) and then (after both simulations) to compare the actual spike times recorded by the spike detector?

testsuite/unittests/issue-565.sli
```diff
@@ -0,0 +1,64 @@
...
+SeeAlso:
+*/
+
+(unittest) run
+/unittest using
+
+M_ERROR setverbosity
+
+% Create neurons and spike detector
+/iaf_neuron 2 << /I_e 400. >> Create /neurons Set
```
janhahne: Use `iaf_psc_alpha` instead (see PR #526)

jakobj:
Thanks @janhahne for the quick review. I have pushed a new version addressing all your comments.

sdiazpier:
Dear @jakobj thanks for fixing this issue. I agree with the solution :+1:  from my side.

heplesser:
@janhahne Could you confirm that @jakobj has addressed your concerns successfully?

janhahne:
@heplesser @jakobj Yes, he did! Nice work :+1:

PR #569, opened 2016-11-29
Added Include Files for targetidentifier.h

maharjun:
Added the relevant include definitions in targetidentifier.h as including it would often cause undefined symbol issues


heplesser:
@jougs Could you take a look?

jougs:
LGTM. Thanks for spotting and fixing.

Issue #572, opened 2016-12-01
Travis unable to find the build log files in the Travis build directory


lekshmideepu:
Merge commit fails in Travis build as it is unable to find the log files in the build directory.

```
$ cd $TRAVIS_BUILD_DIR/build
$ tar -zcvf logfiles.tar.gz ./*.log
tar: ./*.log: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```
It is looking for the following files: 
```
$ tar -zcvf logfiles.tar.gz ./*.log
./install-help.log
$ tar -zcvf reports.tar.gz ./reports
./reports/
./reports/installcheck.log
./reports/TEST-core.phase_2.xml
./reports/TEST-core.phase_1.xml
./reports/TEST-core.phase_4.xml
./reports/TEST-core.phase_3.xml
```
 Need to edit the .travis.yml file to look for new .html and .hlp files

PR #573, opened 2016-12-01
Modified the before_deply section in the .travis.yml file

lekshmideepu:
Modified the before_deply section in the .travis.yml file by adding the compressed doc files and removed the install log file which doesn't exist.
This will solve the issue #572


lekshmideepu:
I would like to suggest @heplesser and @jougs as the reviewers.

jougs:
Great, thanks for finding and Fixing!

PR #575, opened 2016-12-03
Refractory implementation aeif models (fixes #547)

### LINKED ISSUE #547: Inconsistent implementation of refractory period between iaf and aeif models
The implementation of the refractory time is inconsistent between the ``iaf_*`` and ``aeif_*`` models:

- aeif models "seem to consider" that when a spike occurs, the spike was emitted at the beginning of the step, so the timestep containing the spike is considered in the refractory period.
- iaf and gif models do not see it that way, so there is an additional timestep compared to aeif...

Here is a MWE to show the difference between the ``iaf_psc_alpha`` model and the ``aeif_psc_alpha`` model for which adaptation was removed and ``Delta_T`` was set to zero.
[refractory_iaf_aeif.zip](https://github.com/nest/nest-simulator/files/594897/refractory_iaf_aeif.zip)

I think this problem needs to be solved before #473

Silmathoron:
Following #547, this corrects the refractory implementation for:
* aeif_cond_alpha
* aeif_cond_exp
* aeif _psc_exp
* aeif_psc_alpha

This PR also removes the ``DT0`` dynamics function in the ``aeif_cond_alpha/exp`` models.

### DIFF
--- a/doc/model_details/aeif_models_implementation.ipynb
+++ b/doc/model_details/aeif_models_implementation.ipynb
@@ -265,11 +265,24 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 1,
    "metadata": {
-    "collapsed": true
+    "collapsed": false
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "ImportError",
+     "evalue": "libsundials_nvecserial.so.2: cannot open shared object file: No such file or directory",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[0;31m---------------------------------------------------------------------------\u001b[0m",
+      "\u001b[0;31mImportError\u001b[0m                               Traceback (most recent call last)",
+      "\u001b[0;32m<ipython-input-1-6fb744ce4318>\u001b[0m in \u001b[0;36m<module>\u001b[0;34m()\u001b[0m\n\u001b[0;32m----> 1\u001b[0;31m \u001b[0;32mfrom\u001b[0m \u001b[0massimulo\u001b[0m\u001b[0;34m.\u001b[0m\u001b[0msolvers\u001b[0m \u001b[0;32mimport\u001b[0m \u001b[0mLSODAR\u001b[0m\u001b[0;34m\u001b[0m\u001b[0m\n\u001b[0m\u001b[1;32m      2\u001b[0m \u001b[0;32mfrom\u001b[0m \u001b[0massimulo\u001b[0m\u001b[0;34m.\u001b[0m\u001b[0mproblem\u001b[0m \u001b[0;32mimport\u001b[0m \u001b[0mExplicit_Problem\u001b[0m\u001b[0;34m\u001b[0m\u001b[0m\n\u001b[1;32m      3\u001b[0m \u001b[0;34m\u001b[0m\u001b[0m\n\u001b[1;32m      4\u001b[0m \u001b[0;32mclass\u001b[0m \u001b[0mExtended_Problem\u001b[0m\u001b[0;34m(\u001b[0m\u001b[0mExplicit_Problem\u001b[0m\u001b[0;34m)\u001b[0m\u001b[0;34m:\u001b[0m\u001b[0;34m\u001b[0m\u001b[0m\n\u001b[1;32m      5\u001b[0m \u001b[0;34m\u001b[0m\u001b[0m\n",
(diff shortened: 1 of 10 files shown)


Silmathoron:
@heplesser this new implementation seems a lot less precise than the previous one with regard to LSODAR... I'm investigating

Silmathoron:
@heplesser comparing the relative difference between nest and the LSODAR solution shows that the smallest divergence is obtained without applying ``V = V_reset`` and ``f[ S::V_M ] = 0.`` in the dynamics function.
I think this should be applied to #557 also.

heplesser:
@Silmathoron Thanks for adding the tests, but I think you may have gotten it the wrong way around: I believe the LSODAR reference solution is incorrect, since it does not know about refractoriness. 
I tested `aeif_cond_beta_multisynapse` against LSODAR with refractoriness in my heplesser/Issue-547-Refractoriness branch and there it worked perfectly. Here are the essential bits of my LSODAR code, you will find the rest in the doc/model_details notebook in my branch:
```Python
    def rhs(self, t, y, sw):
        """
        This function returns the dV/dt, dw/dt
        """
        
        is_refractory = t < self.t_end_refractory

        V = self.p['V_reset'] if is_refractory else y[0]
        w = y[1]

        Ispike = ( 0. if self.p['Delta_T'] == 0 else
                   self.p['g_L'] * self.p['Delta_T'] * 
                   np.exp( (V-self.p['V_th']) / self.p['Delta_T'] ) )

        dotV = ( 0. if is_refractory else
                 ( -self.p['g_L'] * (V-self.p['E_L']) + Ispike + self.p['I_e'] - w ) / 
                   self.p['C_m'] )
        dotW = ( self.p['a'] * (V-self.p['E_L']) - w ) / self.p['tau_w']

        return np.array([dotV, dotW])

    def handle_event(self, solver, event_info):
        """
        Event handling. This functions is called when Assimulo finds an event as
        specified by the event functions.
        """
        # only look at the state events information.
        if event_info[0][0] > 0:
            solver.sw[0] = True
            solver.y[0] = self.p['V_reset']
            solver.y[1] += self.p['b']
            self.t_end_refractory = solver.t + self.p['t_ref']
        else:
            solver.sw[0] = False
```
One could (should?) probably go yet a step further and code return from refractoriness as an event as well.

Silmathoron:
@heplesser Ok, thanks for the answer!
I'm may indeed have gotten a bit too far in removing all the changes in the ``dynamics`` function; however:
* regardless of the refractory implementation, the LSODAR data was exact when ``t_ref = 0``, therefore we cannot allow the implementation to diverge from it in this limit (especially since ``t_ref`` is seldom used in the aeif models)
* this discrepancy between NEST and LSODAR came from the divergence (as usual) during the "spike timestep" and was therefore not completely "refractory-related".

Therefore I propose the following correction:
* make sure that the spiking-step remains unchanged by testing ``if ( S_.r_ > 0 && S_.r_ <= V_.refractory_counts_ + 1)`` in the ``update`` function
* use the same condition for ``is_refractory`` in the ``dynamics`` function.

As I said before, we need to make sure that the models still pass the tests in the ``t_ref = 0`` limits, so I think this should do the trick, while correcting the refractory dynamics.

heplesser:
@Silmathoron You are right, we need to handle `t_ref==0` correctly and I also see the problem with the "inside" variant proposed in #547. Your code is much closer to "inside_pr". It took me a long time to understand how it worked, though, so I will shortly send you a PR with a slight reformulation which allows us to keep the simpler `S_.r_ > 0` tests without checking the upper limit.

heplesser:
@Silmathoron You are right, we need to handle `t_ref==0` correctly and I also see the problem with the "inside" variant proposed in #547. Your code is much closer to "inside_pr". It took me a long time to understand how it worked, though, so I will shortly send you a PR with a slight reformulation which allows us to keep the simpler `S_.r_ > 0` tests without checking the upper limit.

heplesser:
Assigning this to 2.12 milestone since it is part of the aeif_* cleanup process.

Silmathoron:
@heplesser I don't know whether the labels are really important or for posterity, but shouldn't this be reported as Bug?

I'll switch the implementation to the proposal in #557 shortly.

Silmathoron:
@heplesser I pushed the commit 2 days ago but just realized they are not appearing on GitHub...
``git`` tells me everything is up to date, though, and ``git ls-remote`` indeed shows the right commit number for ``t_ref_aeif``...
Did this ever occur to you?

heplesser:
@Silmathoron This is strange indeed. I can fetch the two commits you mention above from Github as part of your t_aeif_ref branch, so they are on their servers, but somehow do not get displayed. Could you check with Github support? One little thing that I noted was that your commit ec8d479 (one parent in the merge) is marked with a different email address than all your other commits, but that shouldn't derail things.

Silmathoron:
Yes, this is what Jochen advised me a little earlier so I sent them a message...

Silmathoron:
Ok, problem solved thanks to the support team!

heplesser:
@golosio Could you be the second reviewer for this PR?

golosio:
Sure. Indeed, even though I did not comment, I already followed this discussion, as it was closely related to (and referred by) that in PR https://github.com/nest/nest-simulator/pull/557/files , and I've also seen the plot in https://github.com/nest/nest-simulator/pull/557#issuecomment-265107011 and the following comments.
I approve this PR.

PR #577, opened 2016-12-04
Added MUSIC support to recording_device

uahic:
## **Proposal:**

- Added a 'music_connector' to unify the user-interface to specify music_channel ids
- Adding to_music (MUSIC support) to the recording_device class, if compiled with MUSIC. This enables **all** recorders such as multimeter and weight_recorder to stream their data via MPI without much effort in changing their code

## **Discussion:**
**Cont. Out. Port Streaming**
I would like to strap the MUSIC-related code out of the music_cont_out_proxy into a module that can be re-used by any recording neuron which handles continuous data. Two plausible options are to add the code to a new class or to stuff it into the recording_device (as it has been done in this pull request). The phrase 'to stuff it into' is not far from the truth; 

Even though it appears to me very natural to have a to_music flag in the recording device there are two things that were disturbing during the implementation:  

1. The multimeter is already kind of overloaded with functionality
2. The DataLoggingReply contains an array of arrays with the recorded values, but without going into misusing MUSIC-channels to also stream older values out, we are actually only interested in the most recent value vector, thus throwing away the other values contained within a single DataLoggingReply. This also implies that a sampling interval parameter (as given in the multimeter) is not very useful as it only describes a lower-bound on the refresh-rate of the MUSIC-buffer values but the upper bound is given by the MUSIC scheduling (when data gets streamed out). 
3. Quite a few pre-processor flags are required in the code

An alternative solution could be to strip the code out into an external class, keep the multimeter as it is and just refactor the music_cont_out_proxy and all other devices that should be able to use the continuous ports. 

**Music_synapse**
A second thing is the music_channel indexing; The issue here is that each NEST-process must know how to map each local neuron onto a global MUSIC id. This requires global knowledge which is not given out-of-the-box and should be done such that the user does not have to do much work. The best ways are to provide the channel-assignment via synapses ( mapping the music_channel property onto the receptor port number ) or to let the user pass a global index map via StatusDictionary. The receptor-port mapping only works for synapses that are inbound, e. g. the music_event_out_port. So this seems not to be a valid option for all kinds of recording devices. 

A global index map is a real option ( as it is implemented in the music_cont_out_proxy ), however, it is not very intuitive for a user to use synapse parameters for music_event_out_proxy and index maps for a music_cont_out_proxy/multimeter/...  Therefore, I propose to derive a synapse (from a static_connector) with an additional music_channel parameter. I assume that no learning synapse is directly connected to a device/neuron that interfaces MUSIC so a static_synapse should be sufficient. 

## **TODO:**

- Decide whether to include the Cont. Port functionality in recording_device or not
- Decide whether to use global index maps or music_synapses 
- Adapt music_event_out_proxy to read out channel ids from music_synapses or revert it to use receptor ports (I've removed them from the connector_base.h)
- Code cleaning
- Unittests (I don't really feel comfortable with SLI though :) )

If there is a need, the same thing could also be done for Cont. Input Ports to steer internals of generators etc. 

Please let me know your opinions!

### DIFF
--- a/models/modelsmodule.cpp
+++ b/models/modelsmodule.cpp
@@ -139,6 +139,7 @@
 #include "music_event_out_proxy.h"
 #include "music_cont_in_proxy.h"
 #include "music_message_in_proxy.h"
+#include "music_connection.h"
 #endif
 
 namespace nest
@@ -383,6 +384,13 @@ ModelsModule::init( SLIInterpreter* )
 
      SeeAlso: synapsedict, static_synapse
   */
+#ifdef HAVE_MUSIC
+  kernel()
+    .model_manager
+    .register_connection_model< MusicConnection< TargetIdentifierPtrRport > >(
+      "music_synapse" );
+#endif
+
   kernel()
     .model_manager
     .register_connection_model< StaticConnection< TargetIdentifierPtrRport > >(
--- a/models/multimeter.cpp
+++ b/models/multimeter.cpp
@@ -24,6 +24,9 @@
 
 // Includes from nestkernel:
 #include "event_delivery_manager_impl.h"
+#include <string>
+#include "compose.hpp"
+#include "logging.h"
 
 namespace nest
 {
@@ -123,6 +126,7 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
     for ( Token* t = ad.begin(); t != ad.end(); ++t )
       record_from_.push_back( Name( getValue< std::string >( *t ) ) );
   }
+
 }
 
 void
@@ -142,6 +146,11 @@ Multimeter::init_buffers_()
 void
 Multimeter::calibrate()
 {
+#
+  if ( !device_.is_published() )
+  {
+    device_.set_virtual_channel_width( P_.record_from_.size() );
+  }
   device_.calibrate();
   V_.new_request_ = false;
   V_.current_request_data_start_ = 0;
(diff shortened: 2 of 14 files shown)


models/static_connection.h
```diff
@@ -58,6 +58,7 @@ namespace nest
 template < typename targetidentifierT >
 class StaticConnection : public Connection< targetidentifierT >
 {
+protected:
```
heplesser: Why can `weight_ ` no longer be `private`?

heplesser:
@uahic Please make sure that your code is properly formatted, compiles even if Music is not available and passes all tests.

uahic:
The code is prototypic, before spending time to polish the code (that might be changed anyways) I would like to know others opinions about the design decisions (see first two items in the TODO section)

models/static_connection.h
uahic: True, this fragment will be undone

uahic:
Just a little update: 

It seems actually more complicated/impossible to get a 100% satisfying solution, currently I am checking if and how the recording device along with the devices can be refactored such that

- Coupling between recording_device and the devices is avoided as much as possible
- No HAVE_MUSIC preprocessor includes in the device codes
- the class recording device potentially could be splitted into subclasses or using another pattern (the class has grown to something pretty complicated and is tightly coupled to the devices)

apeyser:
We need to place a moratorium on recording device work outside of the nestio branch. The recording devices can have multiple specializations, and a MUSIC extension could be added there, but an ad-hoc approach will not merge with the other changes.

uahic:
@apeyser What do you mean by the nestio branch? Does another unmerged IO thing exist?

apeyser:
@uahic : @jougs @wschenck are involved with me in another merge io branch which does a full refactor. I think they should chime in and speak with you before we end up with two refactors of io.

uahic:
@apeyser Hmh, is the other refactoring already done or can I contribute/have alook at it?

apeyser:
@uahic : another refactor is in testing - @jougs @wschenck should comment on this.

PR #579, opened 2016-12-07
Correct input handling for gif_cond_exp_multisynapse (fixes #511)

### LINKED ISSUE #511: Correct input handling for gif_cond_exp_multisynapse
This issues becomes relevant once #261 is merged.

Input handling for `gif_cond_exp_multisynapse` needs to be modified according to the decision made in the [NEST Developer VC 19 Sep 2016](https://github.com/nest/nest-simulator/wiki/2016-09-19-Open-NEST-Developer-Video-Conference):
- Replace `E_ex`, `E_in` with vector of reversal potentials the same size as `taus_syn` (one per synapse)
- Replace  `I_syn_ex`, `I_syn_in` recordables with one recordable `I_syn` representing the total input current
- Update class and `handle(SpikeEvent&)` to a single input buffer per rport, see [`aeif_cond_beta_multisynapse`](https://github.com/nest/nest-simulator/blob/446d340da0830edcc78ac85ae0a80e58695b3e37/models/aeif_cond_beta_multisynapse.cpp#L607) for an example; consider adding a check for non-negative weight.
- Update `*_dynamics` function to handle input current correctly, along the lines of

``` C++
  double I_syn = 0.0;
  for ( size_t i = 0; i < node.P_.num_of_receptors_ ; ++i )
  {
      I_syn += y[ S::G + i ] * ( node.P_.E_rev[i] - V );
  }
```
- Move `*_dynamics` function to cpp-file.
- Provide examples (python based).

See also #509, #510.

hakonsbm:
This fixes #511 by, among other things
- Using a vector of reversal potentials
- Improving `handle` and `update` functions
- Adding more tests

I suggest @heplesser and @hesam-setareh as reviewers.

### DIFF
--- a/models/gif_cond_exp_multisynapse.h
+++ b/models/gif_cond_exp_multisynapse.h
@@ -130,13 +130,45 @@
   Synaptic parameters
     tau_syn   vector of double - Time constants of the synaptic conductance
                                  in ms.
-    E_ex       double - Excitatory reversal potential in mV.
-    E_in       double - Inhibitory reversal potential in mV.
+    E_rev      double vector - Reversal potential in mV.
 
   Integration parameters
     gsl_error_tol  double - This parameter controls the admissible error of the
                             GSL integrator. Reduce it if NEST complains about
                             numerical instabilities.
+
+  Example:
+
+  import nest
+  import numpy as np
+
+  neuron = nest.Create('gif_cond_exp_multisynapse')
+  nest.SetStatus(neuron, {'E_rev':[0.0, -85.0],
+                         'tau_syn':[4.0, 8.0]})
+
+  spike = nest.Create('spike_generator', params = {'spike_times':
+                                                 np.array([10.0])})
+
+  voltmeter = nest.Create('voltmeter', 1, {'withgid': True})
+
+  delays=[1.0, 300.0]
+  w=[1.0, 1.0, 1.0, 1.0]
+  for syn in range(2):
+     nest.Connect(spike, neuron, syn_spec={'model': 'static_synapse',
+                                           'receptor_type': 1 + syn,
+                                           'weight': w[syn],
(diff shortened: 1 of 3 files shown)


models/gif_cond_exp_multisynapse.h
```diff
@@ -130,14 +130,46 @@
...
+  import numpy as np
+
+  neuron = nest.Create('gif_cond_exp_multisynapse')
+  nest.SetStatus(neuron, {'E_rev':[0.0, -85.0],
+                         'tau_syn':[4.0, 8.0]})
+
+  spike = nest.Create('spike_generator', params = {'spike_times':
+                                                 np.array([10.0])})
+
+  voltmeter = nest.Create('voltmeter', 1, {'withgid': True})
```
heplesser: I think I would leave the voltmeter out as well, it does not really tell us much about the model. You can also drop the plotting section towards the end of the example. Only up to the `Connect()` call is really interesting.

heplesser:
:+1: from me.

hesam-setareh:
I've gone through the code and it is fine. The only thing I do not get is that why the neuron should always have at least one synapse (`tau_syn_.size() == 0` causes error). For example if we only have a neuron connected to a current injector, it does not necessarily need a synapse.

heplesser:
@hesam-setareh When we reviewed aeif*multisynapse models, we thought that it would make life easier for users to be able to have one synapse pre-configured. But there is no good reason to prohibit the user from configuring the neuron without any synapses.

@hakonsbm Could you remove the prohibition against empty `tau_syn`?  I will make the same change for the remaining multisynapse models.

hakonsbm:
@hesam-setareh @heplesser The model now accepts an empty `tau_syn`, i.e. a configuration without any synapses.

hesam-setareh:
All good for me.

Issue #558, opened 2016-11-22
Ensure all examples work with 2.12 and avoid deprecated functions


heplesser:
Once all other issues for this milestone are complete, including #552, ensure that all examples work and do not use deprecated functionality.

heplesser:
Any issues arising here are fixed in #580.

Silmathoron:
There are some subtle problems remaining concerning deprecation matters.
For instance ``iaf_neuron`` is still present in many docstrings' examples:

- pynest/nest/lib/hl_api_models.py
- models/volume_transmitter.h
- models/stdp_dopa_connection.h
- topology/topologymodule.cpp
- nestkernel/nestmodule.cpp
- models/music_event_in_proxy.h
- models/music_event_out_proxy.h
- nestkernel/nestmodule.h
- lib/sli/nest-init.sli
- models/iaf_psc_delta.h
- models/pp_psc_delta.h
- nestkernel/model_manager.h
- sli/slidict.cc
- nestkernel/node.h
- topology/pynest/hl_api.py

Also in some documentation/examples code:

- doc/nest_by_example/scripts/one_neuron_with_sine_wave.py
- doc/model_details/noise_generator.ipynb
- extras/userdoc/md/documentation/*
- extras/ConnPlotter/examples/*
- examples/nest/*.sli
- examples/nest/music/*
- examples/neuronview/neuronview.py
- topology/doc/user_manual_scripts/*
- doc/conngen/conngen.txt

Eventually, there was a mistake in #580 as [``iaf_neuron`` is said to be replaced by ``iaf_cond_alpha`` instead of ``iaf_psc_alpha``](https://github.com/nest/nest-simulator/blob/master/pynest/nest/lib/hl_api_helper.py#L417).

PR #580, opened 2016-12-07
Deprecation warnings for features that will disappear (fixes #552)

### LINKED ISSUE #552: Add deprecation warnings for features that will disappear with NEST 3.
NEST 2.12 should issue deprecation warnings for any functions or models that will not be part of NEST 3 and suggest alternatives. Warnings are limited to PyNEST, since few people write SLI these days except for tests.

The critical change in NEST 3 is that subnets will disappear and thus all functions based on subnets. For more on `GIDCollection`s, see #455. Below is a list of model and functions that I suggest to be deprecated.

## Models
- `subnet`: work with `GIDCollection`s instead
- `aeif_cond_alpha_RK5`: we need to converge on one implementation for this model

## PyNEST functions
- `BeginSubnet`: no longer needed, work with `GIDCollection`s
- `EndSubnet`: no longer needed, work with `GIDCollection`s
- `CurrentSubnet`: irrelevant without subnets
- `GetChildren`: irrelevant without subnets, work with `GIDCollection`s 
- `GetLID`: LIDs disappear with subnets; used index into `GIDCollection`
- `GetLeaves`: elements of `GIDCollection`s are leaves
- `GetNetwork`: irrelevant without subnets, script is responsible for retaining structure information if needed
- `GetNodes`: you have them in `GIDCollection`s
- `LayoutNetwork`: use `Create(<model>, n=<number>)` and get a `GIDCollection`
- `DataConnect`: use `Connect()` with `one_to_one` rule.

## Notes
- `PrintNetwork` in its current form depends on subnets. I suggest we let it print information from `ModelRangeManager` instead, i.e., list showing GID ranges and pertaining models.

stinebuu:
Added deprecation warnings for features that will disappear in NEST 3, which includes all subnet functions, the `subnet` model, `DataConnect()` and the `aeif_cond_alpha_RK5` model. This fixes #552.

The PR also includes partial work on #558, by making sure the examples do not display introduced deprecation warnings.

I suggest @heplesser  and @jougs  as reviewers.


topology/examples/ctx_2n.py
```diff
@@ -40,22 +40,26 @@
...
-                        'elements': ['pyr', 'in']})
+                        'elements': el})
 
-nest.PrintNetwork()
+nest.PrintNetwork(1, (0,))
 
-nest.PrintNetwork(2)
+nest.PrintNetwork(2, (0,))
 
 nest.PrintNetwork(2, ctx)
```
heplesser: I think we should use the opportunity here to improve the example. All these different `PrintNetwork` calls make very limited sense, and `PrintNetwork` will change anyways in NEST3. So I would reduce it to just one call to
```
nest.PrintNetwork(2)
```
I we then suppress the deprecation warning for `CurrentSubnet()` in `PrintNetwork()`, all should work nicely.

pynest/nest/lib/hl_api_subnets.py
```diff
@@ -277,6 +287,8 @@ def EndSubnet():
 
 
 @check_stack
+@deprecated("", "LayoutNetwork: use Create(<model>, n=<number>) and get a "
+            "GIDCollection")
```
heplesser: Better: "LayoutNetwork is deprecated and will be removed in NEST3. Use Create(<model>, n=<number>) instead."
```Python
@deprecated('LayoutNetwork', 'Create(<model>, n=<number>)')
```

pynest/nest/lib/hl_api_subnets.py
```diff
@@ -236,6 +244,7 @@ def GetNetwork(gid, depth):
 
 
 @check_stack
+@deprecated("", "BeginSubnet: no longer needed, work with GIDCollections")
```
heplesser: Better: "BeginSubnet is deprecated and will be removed in NEST3. See documentation for details."

pynest/nest/lib/hl_api_connections.py
```diff
@@ -350,6 +350,7 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
 
 
 @check_stack
+@deprecated("", "DataConnect: use Connect() with one_to_one rule")
```
heplesser: Better: `@deprecated('DataConnect', 'Connect() with one_to_one rule')`. In that way, the message will automatically  have standard format.

jougs:
I have a very general remark: shouldn't deprecation warnings be handled on the SLI level wherever possible? This way they'd automatically bubble up to PyNEST and would only have to be implemented once.

For models, we could for example add two flags in the `Node` base class (`deprecated`, `deprecation_warning_issued`) that are checked in all functions dealing with nodes. A similar mechanism could be thought of for `SLIFunction`s, where the interpreter could issue warnings whenever a function is used.

heplesser:
@jougs Since we already had the nice `@deprecated` decorator in PyNEST and almost all end users use PyNEST today, I figured that doing it at the Python level would be easier and cover most cases.

stinebuu:
@heplesser I have made changes in accordance to your comments if you want to take a look.

heplesser:
@jougs We have now added support for deprecation at the C++ level. I am wondering what to do about the SLI examples. We would have to write quite a bit of code programming around the lack of full support for GIDCollections, which will simplify significantly again in NEST 3. Therefore, I wonder if it isn't better to leave the SLI examples as they are, even though the user then will see some warnings. The problem does not (or to a much lesser degree) exist at the PyNEST level, since lists of GIDs there behave almost like GIDCollections will in NEST 3.

stinebuu:
There are now deprecation warnings on PyNEST and SLI level. Some PyNEST examples, like `ctx_2n.py` and `hill_tononi_Vp.py` still show deprecation warnings, but I have made comments saying it's a work-around. I have not done anything with the SLI-examples.

topology/examples/test_3d_exp.py
```diff
@@ -27,6 +27,10 @@
 3d layers are currently not supported, use at your own risk!
 
 Hans Ekkehard Plesser, UMB
+
+This example uses the function GetChildren, which is deprecated. A deprecation
+warning is therefore released. For details about deprecated functions, see
```
heplesser: released -> issued (applies also in other places)

pynest/nest/lib/hl_api_helper.py
```diff
@@ -413,11 +413,74 @@ def model_deprecation_warning(model):
...
 
     if model in deprecated_models:
         text = "The {0} model is deprecated and will be removed in a \
         future verstion of NEST, use {1} instead\
         ".format(model, deprecated_models[model])
         text = get_wrapped_text(text)
         warnings.warn('\n' + text)
+
+
+def turn_of_deprecation_warning(deprecated_func_name):
```
heplesser: `turn_of_...` -> `turn_off_...`

heplesser:
@apeyser Since @jougs is unavailable at the moment, could you be the second reviewer for this one?

lib/sli/nest-init.sli
```diff
@@ -732,9 +732,18 @@ def
...
+/LayoutNetwork
+<< /deprecation_warning true >>
+Options
+
 /LayoutNetwork_l_a_dict
 {
   << >> begin
+  /LayoutNetwork /deprecation_warning GetOption true eq
+  {
+    (Sli function LayoutNetwork is deprecated in NEST 3.0.) M_DEPRECATED message
```
jougs: "Sli" -> "SLI"

There are some more changes like this that I won't comment on. Please search and replace all of them.

libnestutil/logging_event.cpp
```diff
@@ -63,6 +63,9 @@ std::ostream& operator<<( std::ostream& out, const LoggingEvent& e )
   case M_INFO:
     out << "[INFO] ";
     break;
+  case M_DEPRECATED:
+    out << "[INFO] ";
```
jougs: Wouldn't it be even more expressive if "[INFO]" was "[DEPRECATED]"? This would also allow to search the output log for such functions more easily.

models/modelsmodule.cpp
```diff
@@ -353,7 +358,9 @@ ModelsModule::init( SLIInterpreter* )
 
   // This version of the AdEx model does not depend on GSL.
   kernel().model_manager.register_node_model< aeif_cond_alpha_RK5 >(
-    "aeif_cond_alpha_RK5" );
+    "aeif_cond_alpha_RK5",
+    /*private_model*/ false,
+    /*deprecation_model*/ "NEST 3.0" );
```
jougs: "deprecation_model" -> "deprecation_info"

pynest/nest/lib/hl_api_connections.py
```diff
@@ -350,6 +352,8 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
 
 
 @check_stack
+@deprecated('', 'DataConnect is deprecated and will be removed in NEST3. Use \
```
jougs: "NEST3" -> "NEST 3.0" to be consistent with the other messages

jougs:
@heplesser: I agree regarding the SLI examples. We can't possibly re-formulate all of them now and they will be naturally adapted when the deprecated models/functions are removed. I'm happy to accept this PR with the current current scope once my comments are addressed.

heplesser:
@apeyser @jougs has reviewed the PR now, so no need for action on your side.

Issue #576, opened 2016-12-04
[MPI][Structural plasticity] Running structural plasticity with MPI results in NEST crash


sanjayankur31:
Description of problem:
------------------------------
Running structural plasticity simulations using MPI results in a crash. I ran into this with my research simulation first and have tried to strip the simulation down in an attempt to pin point the issue. Not too successful yet, I'm afraid.

Version-Release or Git commit:
-----------------------------------------
```
[asinha@ankur  nestkernel(master=)]$ git lg | head -1
*   782d30a - (3 days ago) Merge pull request #573 from lekshmideepu/travis_build_log_fix — Jochen Martin Eppler
```

How reproducible: (Always/Sometimes/Unsure)
-----------------------------------------------------------
Always in this configuration

Also, if the simulation is run without using MPI, the simulation does NOT crash.

Steps to Reproduce:
--------------------------
1. run simulation: `mpirun -n 2 testcase.py`
2.
3.

Actual results:
------------------
Crash output (complete output log attached):

```
python: /home/asinha/Documents/02_Code/00_repos/01_others/nest-simulator/nestkernel/connector_model_impl.h:518: nest::ConnectorBase* nest::GenericConnectorModel<ConnectionT>::delete_connection(nest::Node&, size_t, nest::ConnectorBase*, nest::synindex) [with ConnectionT = nest::StaticConnection<nest::TargetIdentifierPtrRport>; size_t = long unsigned int; nest::synindex = unsigned char]: Assertion `conn != 0' failed.
[ankur:26916] *** Process received signal ***
[ankur:26916] Signal: Aborted (6)
[ankur:26916] Signal code:  (-6)
[ankur:26916] [ 0] /lib64/libpthread.so.0(+0x115c0)[0x7fcdfdc605c0]
[ankur:26916] [ 1] /lib64/libc.so.6(gsignal+0x9f)[0x7fcdfd1ae92f]
[ankur:26916] [ 2] /lib64/libc.so.6(abort+0x16a)[0x7fcdfd1b052a]
[ankur:26916] [ 3] /lib64/libc.so.6(+0x2de37)[0x7fcdfd1a6e37]
[ankur:26916] [ 4] /lib64/libc.so.6(+0x2dee2)[0x7fcdfd1a6ee2]
[ankur:26916] [ 5] /opt/lib64/python2.7/site-packages/nest/../../../libmodels.so(+0x36f573)[0x7fcdf5977573]
[...]

PR #582, opened 2016-12-08
Check if sparsetable has any targets before getting them

sanjayankur31:
The get() method does not first check to see if the sparsetable
contains any entries. Instead, it uses an assert() statement to check
if the ith entry is a valid one. This assert causes simulations to crash in
certain scenarios. One example scenario:

- two mpi ranks with one thread each, so two virtual processes vp0, vp1
- two sparsetables, one for each rank - sparsetable0, sparsetable1
- one neuron, so, on rank1. rank0 has no nodes
- structural plasticity enabled
- when the neuron forms a connection (autapse), both source and target
  are on rank1/vp1 - so sparsetable1 is valid and non empty.
- rank0/vp0 has no targets, sparsetable0 is empty
- when neuron needs to delete connections to reduce synaptic elements,
  on each rank, it gets the global ids of all neurons that need to
  delete an element. Then, it tries to get targets for these global ids
  on its rank. Since sparsetable0 is empty, this causes the assert in
  the get() in sparsetable.h to fail, causing the simulation to crash.

Example testcase simulation: https://github.com/nest/nest-simulator/files/634198/testcase-crash-1.txt

This fix is related to issue #576 but I'm not sure if it fixes both crashes that I've documented there.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1312,7 +1312,7 @@ nest::ConnectionManager::get_targets( std::vector< index > sources,
     target_it = targets.begin();
     for ( ; source_it != sources.end(); source_it++, target_it++ )
     {
-      if ( ( *it ).get( *source_it ) != 0 )
+      if ( not( *it ).empty() and ( *it ).get( *source_it ) != 0 )
       {
         validate_pointer( ( *it ).get( *source_it ) )
           ->get_target_gids( ( *target_it ), thread_id, synapse_model );


heplesser:
@sdiazpier @suku248 Could you review this one as well as #584?

sdiazpier:
Hi dear @sanjayankur31, I think that at the point you are doing this check it will be too late to make things right. The problem needs to be stopped a bit earlier. I am right now exploring also the solution suggested by @jakobj in issue #576.

heplesser:

PR #584, opened 2016-12-08
Check if conn exists before attempting to delete it

sanjayankur31:
Fixes the first crash reported in issue #576

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -643,22 +643,24 @@ nest::ConnectionManager::disconnect( Node& target,
     // get the ConnectorBase corresponding to the source
     ConnectorBase* conn =
       validate_pointer( validate_source_entry_( target_thread, sgid, syn_id ) );
-    if ( conn != 0 )
+    if ( conn == 0 )
     {
-      ConnectorBase* c =
-        kernel()
-          .model_manager.get_synapse_prototype( syn_id, target_thread )
-          .delete_connection( target, target_thread, conn, syn_id );
-      if ( c == 0 )
-      {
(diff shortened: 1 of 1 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -643,19 +643,22 @@ nest::ConnectionManager::disconnect( Node& target,
     // get the ConnectorBase corresponding to the source
     ConnectorBase* conn =
       validate_pointer( validate_source_entry_( target_thread, sgid, syn_id ) );
-    ConnectorBase* c =
-      kernel()
-        .model_manager.get_synapse_prototype( syn_id, target_thread )
-        .delete_connection( target, target_thread, conn, syn_id );
-    if ( c == 0 )
+    if ( conn != 0 )
```
heplesser: Could we simplify code a little by writing
```C++
if ( conn == 0 )
{
  return;
}
```
and then the rest outside any block?

heplesser:
@sdiazpier @suku248 Could you review this?

jakobj:
I am not 100% sure, but I think disconnect should in any case never be called with a non-existent connection, making this check unnecessary. If it is called with a non-existent connection, this is caused be some bug in the preceeding code which should be fixed rather than correcting for it at this point.

sdiazpier:
I agree with the comment by @jakobj. The source of the problem is surely higher up.

sanjayankur31:
Sure, extremely likely - I haven't gotten down to tracing the various program paths that can lead to this. Should we close this (and #582) and focus on logical fixes like the one in #595 instead then?

heplesser:

PR #586, opened 2016-12-11
Restore Data Connect

hannahbos:
The connection routine DataConnect was made thread safe and some parts were rewritten. The routine now takes at least double the time to create sparse networks, compared to its performance in NEST 2.10. DataConnect is sped up again by removing a clearing instruction of the data dictionary  after initialisation. As reviewers I suggest @heplesser and @jougs.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -805,7 +805,6 @@ nest::ConnectionManager::divergent_connect( index source_id,
 
       // here we fill a parameter dictionary with the values of the current loop
       // index.
-      par_i->clear();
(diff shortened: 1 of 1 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -805,7 +805,6 @@ nest::ConnectionManager::divergent_connect( index source_id,
 
       // here we fill a parameter dictionary with the values of the current loop
       // index.
-      par_i->clear();
```
heplesser: In the loop following on lines 808-814, we always fill all entries in the dictionary with new values (on first pass, the dict is empty). Therefore, it is not necessary to explicitly clear the dict.

heplesser:
@mschmidt87 Could you be the second reviewer for this one?

Issue #581, opened 2016-12-07
NEST master (92d53f9) does not compile on K


jakobj:
The current master branch does not compile on K due to the following error:

```bash
"/home/ra000024/a03303/opt/nest-simulator/nestkernel/connection_manager.cpp", line 1141: error: no operator "<<" matches these operands
            operand types are: std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>> << std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>>
          connectome << conns_in_thread;
                     ^

"/home/ra000024/a03303/opt/nest-simulator/nestkernel/connection_manager.cpp", line 1182: error: no operator "<<" matches these operands
            operand types are: std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>> << std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>>
          connectome << conns_in_thread;
                     ^

"/home/ra000024/a03303/opt/nest-simulator/nestkernel/connection_manager.cpp", line 1233: error: no operator "<<" matches these operands
            operand types are: std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>> << std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>>
          connectome << conns_in_thread;
```
However, a few lines below it complains, that exactly that function is not referenced from anywhere (modulo the `&`, changing the function definition to pass-by-value also doesn't work, and in any case would not be a desirable change ):

```bash
"/home/ra000024/a03303/opt/nest-simulator/nestkernel/connection_manager.cpp", line 1090: warning: function "operator<<(std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>> &, std::deque<nest::ConnectionID, std::allocator<nest::ConnectionID>> &)" was declared but never r
eferenced
  static inline std::deque< nest::ConnectionID >& operator<<(
```

Can anyone explain this imho rather strange behaviour of the compiler? It does not appear on JUQUEEN, our local cluster, or my laptop.

Here's my cmake line
```bash
[...]

PR #587, opened 2016-12-11
Replace operator<< with regular function to avoid issues with K compiler

jakobj:
This PR fixes an incompatibility with the K compiler as discussed in #581.
I suggest @janhahne and @JanneM as reviewers.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1087,8 +1087,8 @@ nest::ConnectionManager::get_connections( DictionaryDatum params ) const
 
 // Helper method, implemented as operator<<(), that removes ConnectionIDs from
 // input deque and appends them to output deque.
-static inline std::deque< nest::ConnectionID >& operator<<(
-  std::deque< nest::ConnectionID >& out,
+static inline std::deque< nest::ConnectionID >&
+extend_connectome( std::deque< nest::ConnectionID >& out,
   std::deque< nest::ConnectionID >& in )
 {
   while ( not in.empty() )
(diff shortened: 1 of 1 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -1087,8 +1087,8 @@ nest::ConnectionManager::get_connections( DictionaryDatum params ) const
 
 // Helper method, implemented as operator<<(), that removes ConnectionIDs from
 // input deque and appends them to output deque.
-static inline std::deque< nest::ConnectionID >& operator<<(
-  std::deque< nest::ConnectionID >& out,
+static inline std::deque< nest::ConnectionID >&
+join_deques( std::deque< nest::ConnectionID >& out,
```
heplesser: The `static` is confusing here, since this use of `static` is so rare hardly anyone knows what it does. It does make sense, as it makes sure there won't be any conflicts with functions of same name and signature defined in other cpp-files, but you should probably comment on this here.

I am also wondering a little about the name. Maybe `extend( connectome, conns_in_thread )` or `add_to( connectome, conns_in_thread )` would make for even more readable code.

JanneM: I don't have enough experience with the code base to review the patch, I think. I agree that the name should reflect what the code does to nest rather than what it does to the data structure.

The original code uses static, so it makes sense to keep it for this patch. If it's unneeded, then it's probably better to create a separate issue to change that.

jakobj: I agree with Janne on the second part. Regarding the name I don't have a strong opinion, I would be fine with `extend` or even something more specific `extend_connectome` that actually describes would be would like to achieve here.

jakobj:
Thanks @janhahne. Yes I compiled it on K and will do so again, before merging, after we have settled on a name.

nestkernel/connection_manager.cpp
heplesser: `extend_connectome()` should be fine.

jakobj:
Renamed and made sure that it compiles on K. A second thumbs up anyone?

heplesser:
:+1: and merging.

PR #590, opened 2016-12-13
Refractory tests and corrected hh_cond_exp_traub (fixes #473)

### LINKED ISSUE #473: hh_cond_exp_traub handles refractory time incorrectly
[`hh_cond_exp_traub::calibrate()`](https://github.com/nest/nest-simulator/blob/0d58accc82b6b44801c82e502976889e3c37d80b/models/hh_cond_exp_traub.cpp#L371) sets `V_.RefractoryCounts_` to a hard-wired value of 20 time steps, independent of the simulation resolution. Neuron responses will therefore depend on resolution.

Note also that the refractory period cannot be changed by the user.

Silmathoron:
This PR adresses #473 by correcting the refractory behaviour of ``hh_cond_exp_traub``
It also implements a random refractory test for many hybrid models.


pynest/nest/tests/test_refractory.py
```diff
@@ -0,0 +1,200 @@
...
+# -------------------------
+#
+
+ignore_model = [
+    "aeif_cond_alpha_RK5",
+    "aeif_cond_alpha_multisynapse",
+    "aeif_cond_beta_multisynapse",
+    "amat2_psc_exp",
+    "ginzburg_neuron",
+    "hh_cond_exp_traub",
```
heplesser: Why do you need to ignore this model? Isn't that the one you are fixing in this PR?

pynest/nest/tests/test_refractory.py
```diff
@@ -0,0 +1,200 @@
...
+
+# --------------------------------------------------------------------------- #
+#  Models, specific parameters
+# -------------------------
+#
+
+ignore_model = [
+    "aeif_cond_alpha_RK5",
+    "aeif_cond_alpha_multisynapse",
+    "aeif_cond_beta_multisynapse",
```
heplesser: Why do you need to exclude the multisynapse models?

pynest/nest/tests/test_refractory.py
Silmathoron: Yes, but the ``hh_*`` models do not clamp the potential, so I cannot easily check the refractory period... I'll try using a huge current

Silmathoron: No, I think there is just no way to check the HH models (I'm not even sure ``t_ref`` makes any sense for them)

Silmathoron:
Ok, I corrected the test (and recorrected it when I remembered why I had built the list of neurons to be tested that way... -_-')
I think this is the best I can do, but please tell me if you see how it might be enhanced.

heplesser:
@Silmathoron Very nice. But could you do one more update? The Hill-Tononi revision has just been merged this morning. Could you merge that into your branch and make sure that your test still works?

Silmathoron:
This is what I did this morning just after abigail merged it ;)

heplesser:
Sorry, my mistake. Github had listed all your recent commits under "... added some commits 2 days ago", so I thought it couldn't contain the revised HT.

heplesser:
@jschuecker Could you be the second reviewer for this one? It is just a small (but important fix) and a new test.

pynest/nest/tests/test_refractory.py
```diff
@@ -0,0 +1,271 @@
...
+
+# additional parameters for the connector
+add_connect_param = {
+    "iaf_cond_alpha_mc": {"receptor_type": 7},
+}
+
+
+# --------------------------------------------------------------------------- #
+#  Simulation and refractory time
+# -------------------------
```
jschuecker: you could change this to "refractory time limits" to avoid confusion as this are only the lower and upper bounds right?

pynest/nest/tests/test_refractory.py
```diff
@@ -0,0 +1,271 @@
...
+        spike_times = nest.GetStatus(sd, "events")[0]["times"]
+        if model in neurons_interspike:
+            # spike emitted at next timestep so substract resolution
+            return spike_times[1]-spike_times[0]-resolution
+        elif model in neurons_interspike_ps:
+            return spike_times[1]-spike_times[0]
+        else:
+            Vr = nest.GetStatus(neuron, "V_reset")[0]
+            times = nest.GetStatus(vm, "events")[0]["times"]
+            idx_max = (np.argwhere(np.isclose(times, spike_times[1]))[0][0]
```
jschuecker: is there a reason why np.isclose is used here? Otherwise you could make this consistent with line 209.

jschuecker:
@Silmathoron: Very nice new test! I only have two minor suggestions (see my inline comments).

pynest/nest/tests/test_refractory.py
Silmathoron: Yes, it's for precise models ;) I'll add a comment

pynest/nest/tests/test_refractory.py
Silmathoron: Done (and additional comments added)

pynest/nest/tests/test_refractory.py
```diff
@@ -134,14 +134,14 @@
 
 
 # --------------------------------------------------------------------------- #
-#  Simulation and refractory time
+#  Simulation and refractory time limits
```
jschuecker: @Silmathoron, Sorry, I meant to change this to "Simulation time and refractory time limits"

Silmathoron: easy ;)

jschuecker:
@Silmathoron thanks for the corrections. :+1: from my side

PR #595, opened 2016-12-15
Remove invalid partners before attempting to break synapses

sanjayankur31:
As discussed on the mailing list, the list of "partners" that should be dealt with should be limited to neuronal nodes that contain the corresponding synaptic element. This commit attempts to make this check and remove any partners that do not fit this criteria from the list.

### DIFF
--- a/nestkernel/sp_manager.cpp
+++ b/nestkernel/sp_manager.cpp
@@ -384,7 +384,7 @@ SPManager::update_structural_plasticity( SPBuilder* sp_builder )
   // Vector of displacements for communication
   std::vector< int > displacements;
 
-  // Get pre synaptic elements data from global nodes
+  // Get pre synaptic elements data from local nodes
   get_synaptic_elements( sp_builder->get_pre_synaptic_element_name(),
     pre_vacant_id,
     pre_vacant_n,
@@ -402,6 +402,9 @@ SPManager::update_structural_plasticity( SPBuilder* sp_builder )
   kernel().mpi_manager.communicate(
     pre_deleted_n, pre_deleted_n_global, displacements );
 
+  // This implies that there are no more vacant pre-synaptic elements.
+  // Instead, some synapses need to be deleted to reduce the number of synaptic
+  // elements.
   if ( pre_deleted_id_global.size() > 0 )
   {
     delete_synapses_from_pre( pre_deleted_id_global,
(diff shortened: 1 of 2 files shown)


jakobj:
Could you please elaborate a bit more what defines an "invalid partner" conceptually? From your code I can see that one requirement is that it needs to have proxies. Is this reasonable? I understand the intuitive appeal of excluding devices, but maybe someone else does find a reasonable use case for using structural plasiticity on devices. Is it not enough, as @sdiazpier pointed out to make sure that only synapses are deleted which have the correct type specified in the structural plasticity manager?

sanjayankur31:
Hiya,

I don't actually see how structural plasticity can be used with devices at all. 

Structural plasticity in this model that we're implementing in NEST, is homoeostatic, which implies that neurons attempt to either increase or reduce their connectivity with other neurons to maintain their electrical activity level. So, two things - first, devices are not part of the biological network, they are man made entities we introduce for experimentation - they are meant to be static entities; second - they don't contribute to this homoeostatic process in the same (any?) way. 

Devices do not have a calcium concentration to represent electrical activity, and do not gain or lose synaptic elements either since they don't have them at all (they shouldn't - I've got to check this). Even if we use "static synapses" for connections to and from devices, this is a simplification we've introduced in the nest implementation - it isn't really formed between a dendritic spine and an axonal bouton as far as devices are concerned. So, considering them for formation or deletion of synapses will result in an erroneous implementation of the model of structural plasticity since the network's topology will not be modified in the correct way.

@sdiazpier - does this sound correct, what do you reckon?

jakobj:
Yes, I agree with you argument. As @sdiazpier also pointed out, only nodes which actually can gain/loose synaptic elements should be considered by the structural plasticity framework. Since this is not the case for devices, as you already argued, they should not be part of the list of potentially deleted connections. However, I think rather than removing them at this point, they should never be added to the array in the first place. This is, imho, what the fix should address.

sanjayankur31:
+1

Not adding these entities to the target list at all would be more efficient. I did look at that a bit but from what I understood of the data structures that hold target information, e.g., `ConnectionManager`, they don't hold information about the type of target node. They are generic by design - to cater to all types of targets. They only seem to hold information of the synapse type, which, given that synapses that connect devices are `static_synapses` doesn't help much.

heplesser:

PR #601, opened 2016-12-21
Rename DataConnect related functions

jakobj:
This PR proposes to rename the functions used only by `DataConnect`, since the previous names (`divergent_connect`, `connect`) were rather confusing considering the NewConnect framework. The names are only suggestions, and I am happy to hear other (better) ideas.
I propose @hannahbos and @heplesser as reviewers.


hannahbos:
Thanks! :+1:

Issue #600, opened 2016-12-21
Nodes supporting waveform relexation can lead to desynchronization of MPI processes


jakobj:
when simulating with two MPI processes, the following code causes the simulation to hang:
```postscript
ResetKernel

0 << /total_num_virtual_procs 2 >> SetStatus

/hh_psc_alpha_gap 2 Create

1 << /node_uses_wfr true >> SetStatus
2 << /node_uses_wfr false >> SetStatus

10 Simulate
```

the cause for this can be found in https://github.com/nest/nest-simulator/blob/master/nestkernel/node_manager.cpp#L614 ff, where `any_node_uses_wfr` is only determined locally on every process. since in the simulation above, one neuron is located on rank 0, entering the wfr loop (https://github.com/nest/nest-simulator/blob/master/nestkernel/simulation_manager.cpp#L654), while the other rank directly proceeds with a regular update of the nodes.
if the user is allowed to set this flag, we need to communicate the value of `any_node_uses_wfr` among all processes somewhere in `prepare_simulation`. another option would be to not expose this setting to the user, but determine during all `Create` calls, whether nodes which support wfr have been created.
this we would like to discuss. please have a look at this @janhahne @suku248 @mhelias @ddahmen @jschuecker

janhahne:
@jakobj Good catch and thank you for the analysis! 

In general the node parameter `node_uses_wfr` should not be set by the user, but be determined as `false` if the neuron does not support waveform relaxation and equal to the kernel parameter `use_wfr` if the neuron supports wfr. So, a manual setting of `node_uses_wfr` can cause wrong results anyway. Therefore we should disable the possibility to set `node_uses_wfr` (this should have happend anyway, I just forgot about it...)

This issue however describes a even bigger problem: Code like

```
ResetKernel

0 << /total_num_virtual_procs 2 >> SetStatus

/hh_psc_alpha_gap 1 Create
/iaf_psc_alpha 1 Create

10 Simulate 
```

also fails when simulated with two MPI processes, although this is a perfectly fine simulation script. In general this problem will always occur if there are mpi processes with neurons that use wfr and others without. If one creates all `hh_psc_alpha_gap` neurons with a single `Create` call this does not happen very often, which is probably why this stayed undetected.

[...]

PR #603, opened 2016-12-22
Fixed desynchronization of MPI processes due to waveform relaxation (issue #600)

janhahne:
This PR fixes issue #600 and disables the possibility to set `node_uses_wfr`. 

Since there is no `MPI_BOOL` i used `MPI_CHAR` instead. This needs careful review and maybe some improvements. I suggest @heplesser and @jakobj as reviewers.


janhahne:
I just noticed that the error still occurs with this "fix". I will have another look at this tomorrow and create a new pull request once i found an actual fix. sorry!

PR #604, opened 2016-12-23
Fixed get targets for structural plasticity

sdiazpier:
This PR addresses issue #576 and issue #589 and substitutes PRs #582 and #584 and #595 by adding a new get targets function which checks if the target has defined synaptic elements before adding them to the possible targets list for deletion. After a deep analysis, I realized that the list of targets only cares for the synapse model and not for the existence of the synaptic elements in the target node. Usually a new synapse model is created for each synaptic elements pair. However this is not enforced in any way in the code and when one uses the same synapse model for all connections, this error will arise.
Checking if the node contains the desired synaptic elements declared solves this problem.
Thank you to @sanjayankur31 and @jakobj who have been working hard to solve and understand this issue. This PR is the result of our accumulated efforts.


sdiazpier:
@sanjayankur31 and @jakobj could you please review this PR?

nestkernel/connection_manager.cpp
```diff
@@ -908,17 +924,26 @@ nest::ConnectionManager::validate_source_entry_( thread tid,
   synindex syn_id )
 {
   kernel().model_manager.assert_valid_syn_id( syn_id );
+  return validate_source_entry_( tid, s_gid );
+}
 
+nest::ConnectorBase*
+nest::ConnectionManager::validate_source_entry_( thread tid, index s_gid )
```
jakobj: these parameters could be `const`, same for function above.

nestkernel/connection_manager.cpp
```diff
@@ -1311,11 +1336,48 @@ nest::ConnectionManager::get_targets( std::vector< index > sources,
...
-        validate_pointer( ( *it ).get( *source_it ) )
+        validate_pointer( validate_source_entry_( thread_id, *source_it ) )
           ->get_target_gids( ( *target_it ), thread_id, synapse_model );
       }
     }
   }
 }
+
+void
+nest::ConnectionManager::get_targets( std::vector< index > sources,
```
jakobj: this function replaces the original `get_targets` function, correct? if so, you should remove the outdated function.

nestkernel/connection_manager.cpp
```diff
@@ -1311,11 +1336,48 @@ nest::ConnectionManager::get_targets( std::vector< index > sources,
...
+  index synapse_model,
+  std::string post_synaptic_element )
+{
+  thread thread_id;
+  std::vector< index >::iterator source_it;
+  std::vector< std::vector< index > >::iterator target_it;
+  targets.resize( sources.size() );
+  for ( std::vector< std::vector< index > >::iterator i = targets.begin();
+        i != targets.end();
+        i++ )
```
jakobj: `++i`

nestkernel/connection_manager.cpp
```diff
@@ -1311,11 +1336,48 @@ nest::ConnectionManager::get_targets( std::vector< index > sources,
...
+  {
+    ( *i ).clear();
+  }
+
+  for ( tVSConnector::iterator it = connections_.begin();
```
jakobj: could you turn this into a loop of this sort
```c++
for ( thread tid = 0; tid < connections_.size(); ++tid )
{
  [...]
}
```
or are there reasons to stick to the iterators. I find it quite hard to read. this would also make it easier to introduce a threaded version later.

nestkernel/connector_base.h
```diff
@@ -461,6 +466,32 @@ class Connector : public vector_like< ConnectionT >
...
+ */
+  void
+  get_target_gids( std::vector< size_t >& target_gids,
+    size_t thrd,
+    synindex synapse_id,
+    std::string post_synaptic_element ) const
+  {
+    if ( get_syn_id() == synapse_id )
+    {
+      for ( size_t i = 0; i < K; i++ )
```
jakobj: `++i`

nestkernel/connector_base.h
```diff
@@ -461,6 +466,32 @@ class Connector : public vector_like< ConnectionT >
...
 
+  /**
+ * Return the GIDs of the target nodes in a given thread, for all connections
+ * on this Connector which match a defined synapse_id.
+ * @param target_gids Vector to store the GIDs of the target nodes
+ * @param thrd Thread where targets are being looked for
+ * @param synapse_id Synapse type
+ */
+  void
+  get_target_gids( std::vector< size_t >& target_gids,
```
jakobj: this replaces the original `get_target_gids` function, which should be removed.

nestkernel/connector_base.h
```diff
@@ -461,6 +466,32 @@ class Connector : public vector_like< ConnectionT >
...
+    size_t thrd,
+    synindex synapse_id,
+    std::string post_synaptic_element ) const
```
jakobj: these three parameters could be declared `const`

nestkernel/connector_base.h
```diff
@@ -691,6 +722,20 @@ class Connector< 1, ConnectionT > : public vector_like< ConnectionT >
   }
 
   void
+  get_target_gids( std::vector< size_t >& target_gids,
```
jakobj: (see above)

nestkernel/connector_base.h
```diff
@@ -927,6 +972,25 @@ class Connector< K_CUTOFF, ConnectionT > : public vector_like< ConnectionT >
...
```
jakobj: (see above)

nestkernel/connector_base.h
```diff
@@ -927,6 +972,25 @@ class Connector< K_CUTOFF, ConnectionT > : public vector_like< ConnectionT >
...
+    size_t thrd,
+    synindex synapse_id,
+    std::string post_synaptic_element ) const
+  {
+    typename std::vector< ConnectionT >::const_iterator C_it;
+    if ( get_syn_id() == synapse_id )
+    {
+      for ( C_it = C_.begin(); C_it != C_.end(); C_it++ )
```
jakobj: `++C_it`

nestkernel/connector_base.h
```diff
@@ -1094,6 +1158,22 @@ class HetConnector : public std::vector< ConnectorBase* >, public ConnectorBase
   }
 
   void
+  get_target_gids( std::vector< size_t >& target_gids,
```
jakobj: (see above)

nestkernel/connector_base.h
```diff
@@ -1094,6 +1158,22 @@ class HetConnector : public std::vector< ConnectorBase* >, public ConnectorBase
...
+    size_t thrd,
+    synindex synapse_id,
+    std::string post_synaptic_element ) const
+  {
+    for ( size_t i = 0; i < size(); i++ )
```
jakobj: `++i`

nestkernel/sp_manager.cpp
```diff
@@ -227,12 +227,21 @@ SPManager::disconnect_single( index sgid,
...
+      kernel().model_manager.get_synapsedict()->lookup( syn_name ) );
+  }
+  else
+  {
+    // Disconnect if Structural plasticity is not activated.
+    const std::string syn_name = ( *syn )[ names::model ];
+    disconnect( sgid,
+      target,
+      target_thread,
+      kernel().model_manager.get_synapsedict()->lookup( syn_name ) );
```
jakobj: isn't this the same code as in the `if` block? why did you duplicate it?

pynest/nest/tests/test_sp/test_disconnect.py
```diff
@@ -30,38 +31,61 @@ class TestDisconnectSingle(unittest.TestCase):
...
             'stdp_dopamine_synapse',
             'stdp_dopamine_synapse_lbl',
             'stdp_dopamine_synapse_hpc',
             'stdp_dopamine_synapse_hpc_lbl',
             'gap_junction',
-            'gap_junction_lbl'
+            'gap_junction_lbl',
         ]
-
+
```
jakobj: please remove whitespace

pynest/nest/tests/test_sp/test_issue-578-sp.py
```diff
@@ -0,0 +1,80 @@
...
+nest.SetStructuralPlasticityStatus({
+    'structural_plasticity_synapses': {
+        'synapseEE': synDictE,
+    }
+})
+
+
+try:
+    nest.Simulate(200*1000)
+except ExceptionType:
```
jakobj: i think this should just read `except:` if you want to catch all exceptions.

nestkernel/connection_manager.cpp
sdiazpier: Yes, I thought it was being used somewhere else but it is not the case.

nestkernel/sp_manager.cpp
sdiazpier: Yes this is right thanks for noticing, I was moving things around and at the end remained the same code. I am correcting it now.

sdiazpier:
@sanjayankur31  and @jakobj thanks for your comments. I have addressed them and fixed the formatting. Please let me know if you think all issues have been solved.

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,36 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
+        ++i )
   {
     ( *i ).clear();
   }
 
-  for ( tVSConnector::iterator it = connections_.begin();
+  /*for ( tVSConnector::iterator it = connections_.begin();
         it != connections_.end();
-        ++it )
+        ++it )*/
```
sanjayankur31: Cosmetic change - remove the commented code. It doesn't contribute to the readability much here. If you think a comment describing the logic here is needed, maybe you can add a descriptive comment instead?

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,36 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
+  for ( thread_id = 0; thread_id < connections_.size(); ++thread_id )
   {
-    thread_id = it - connections_.begin();
+    // thread_id = it - connections_.begin();
```
sanjayankur31: Same cosmetic change as above - the commented code isn't needed and a descriptive comment will fit better if you think it's needed here.

pynest/nest/tests/test_sp/test_issue_578_sp.py
```diff
@@ -0,0 +1,91 @@
...
+        nest.SetStructuralPlasticityStatus({
+            'structural_plasticity_synapses': {
+                'synapseEE': synDictE,
+            }
+        })
+
+        try:
+            nest.Simulate(200 * 1000)
+        except:
+            self.fail("Exception during structural plasticity deletion")
```
sanjayankur31: The test isn't only forcing deletion of synapses - all exceptions are being caught too, so the exception error message should be made more general. What do you think?

sdiazpier:
@sanjayankur31 thanks for the comments. I have removed the comments and I hope it is all good. Please let me know if anything is missing.

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
 void
 nest::ConnectionManager::get_targets( std::vector< index > sources,
   std::vector< std::vector< index > >& targets,
-  index synapse_model )
+  index synapse_model,
+  std::string post_synaptic_element )
 {
-  thread thread_id;
+  unsigned int thread_id;
```
jakobj: you can define this variable in line 1335, and type should be `thread`. I would prefer homogeneous usage of `tid` instead of `thread_id` but maybe this should be part of a maintanance PR, not this one.

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
 void
 nest::ConnectionManager::get_targets( std::vector< index > sources,
```
heplesser: Since `sources` is not modified inside the method, this should be `const std::vector< index>& sources` to avoid unnecessary copying.

nestkernel/connection_manager.cpp
heplesser: I agree with @jakobj

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
   std::vector< std::vector< index > >& targets,
-  index synapse_model )
+  index synapse_model,
+  std::string post_synaptic_element )
 {
-  thread thread_id;
+  unsigned int thread_id;
   std::vector< index >::iterator source_it;
   std::vector< std::vector< index > >::iterator target_it;
```
heplesser: I would also move these two declarations down to lines 1338, 1339.

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
   targets.resize( sources.size() );
   for ( std::vector< std::vector< index > >::iterator i = targets.begin();
         i != targets.end();
-        i++ )
+        ++i )
   {
     ( *i ).clear();
   }
```
heplesser: I think the entire code from `targets.resize()` can be simplified. If I understand the code right, in the end `targets` should be a vector of `sources.size()` empty vectors. I think the idiomatic way to achieve this is
```c++
std::vector< std::vector< index > >( sources.size() ).swap( targets );
```

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
+  // We go through the connections data structure to retrieve all
+  // targets which have an specific post synaptic element for each
+  // source.
+  for ( thread_id = 0; thread_id < connections_.size(); ++thread_id )
   {
-    thread_id = it - connections_.begin();
     // loop over the targets/sources
     source_it = sources.begin();
     target_it = targets.begin();
     for ( ; source_it != sources.end(); source_it++, target_it++ )
```
heplesser: `++source_it, ++target_it` is more efficient.

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
     {
-      if ( ( *it ).get( *source_it ) != 0 )
+      if ( validate_source_entry_( thread_id, *source_it ) != 0 )
       {
-        validate_pointer( ( *it ).get( *source_it ) )
-          ->get_target_gids( ( *target_it ), thread_id, synapse_model );
+        validate_pointer( validate_source_entry_( thread_id, *source_it ) )
+          ->get_target_gids(
+            ( *target_it ), thread_id, synapse_model, post_synaptic_element );
```
heplesser: I think the `()` around `*target_it` are superfluous.

nestkernel/connection_manager.cpp
```diff
@@ -1288,33 +1315,35 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
     source_it = sources.begin();
     target_it = targets.begin();
     for ( ; source_it != sources.end(); source_it++, target_it++ )
     {
-      if ( ( *it ).get( *source_it ) != 0 )
+      if ( validate_source_entry_( thread_id, *source_it ) != 0 )
       {
-        validate_pointer( ( *it ).get( *source_it ) )
-          ->get_target_gids( ( *target_it ), thread_id, synapse_model );
+        validate_pointer( validate_source_entry_( thread_id, *source_it ) )
```
heplesser: Here you call `validate_source_entry_()` for a second time with the same arguments. Perhaps better (guessing that `ConnectorBase` is the right type here
```c++
ConnectorBase const * const source = validate_source_entry_( thread_id, *source_it );
if ( source_ptr != 0 )
{
  source->get_target_gids( *target_it, thread_id, synapse_model );
}
```
Note that I have here assumed that the `validate_pointer()` call is integrated in `validate_source_entry_()`; that would make code here more readable, but if it has side-effects elsewhere, it may mean too much work for now.

Finally, I think the code would become more readable if you changed `target_it` to `targets_it`,  since this is an iterator pointing to a vector.

sdiazpier:
Dear @heplesser  and @jakobj thank you for your suggestions. I have now addressed your last comments. Please let me know if I missed anything.

nestkernel/connection_manager.cpp
```diff
@@ -1313,37 +1313,32 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
 }
 
 void
-nest::ConnectionManager::get_targets( std::vector< index > sources,
+nest::ConnectionManager::get_targets( const std::vector< index > sources,
```
heplesser: The `&` is missing to make it a reference.

nestkernel/connection_manager.cpp
```diff
@@ -1313,37 +1313,32 @@ nest::ConnectionManager::get_sources( std::vector< index > targets,
...
   std::vector< std::vector< index > >& targets,
-  index synapse_model,
-  std::string post_synaptic_element )
+  const index synapse_model,
+  const std::string post_synaptic_element )
```
heplesser: Should also be a reference, `const std::string&` to avoid copying.

nestkernel/connection_manager.h
```diff
@@ -231,10 +231,10 @@ class ConnectionManager : public ManagerInterface
     std::vector< std::vector< index > >& sources,
     index synapse_model );
 
-  void get_targets( std::vector< index > sources,
+  void get_targets( const std::vector< index > sources,
```
heplesser: See above.

nestkernel/connection_manager.h
```diff
@@ -231,10 +231,10 @@ class ConnectionManager : public ManagerInterface
...
     std::vector< std::vector< index > >& targets,
-    index synapse_model,
-    std::string post_synaptic_element );
+    const index synapse_model,
+    const std::string post_synaptic_element );
```
heplesser: See above.

sdiazpier:
Dear @heplesser I have solved the last errors in spelling and references. Please let me know if anything else is missing.

PR #607, opened 2016-12-27
Fixed desynchronization of MPI processes due to waveform relaxation (issue #600)

janhahne:
This PR fixes issue #600 by

- adding a function `wfr_synchrony` that communicates the local values of `any_node_uses_wfr` and returns a global one
- removing `static` from the `GapJunctionEvent` variable `coeff_length_` and communicating it instead (if any MPI process does not contain a neuron that uses wfr the static variable is not set correctly when it is used to read `GapJunctionEvent`s from the receive buffer in `deliver_events()`. This resulted in a segmentation fault.)

Furthermore this PR

- removes the possibility for the user to set `node_uses_wfr` (as this is not required and leads to undefined/wrong results)
- adds a unittests `issue-600.sli`

I suggest @heplesser and @jakobj as reviewers.


nestkernel/mpi_manager.cpp
```diff
@@ -713,6 +713,30 @@ nest::MPIManager::grng_synchrony( unsigned long process_rnd_number )
   return true;
 }
 
+// wfr_synchrony: called at the beginning of each simulate
+bool
+nest::MPIManager::wfr_synchrony( bool any_node_uses_wfr )
```
heplesser: I think we should be more abstract at the level of `MPIManage` and choose a method name that expresses the logical operation, not the semantic purpose in one special case. If I understand the code and context right, this method takes a single bool, exchanges with all other processes, and returns `true` if one or more processes provide `true`. In NumPy-parlance, this is `any()`.

I am wondering about two things:
1. What would be a good name? I fear `any()` is a bit too short.
1. Is there an MPI function that provides this type of aggregation built-in?

I realize that I wrote the similarly problematic `grng_synchrony()` myself ...

janhahne: @heplesser You understand it right and I agree. I made an attempt to improve the naming

1. `any` is indeed very short. I chose `any_true` instead. This indicates that this is an operation with boolean parameters.
2. I took a look at the MPI standard but did not find anything like this.

janhahne:
@heplesser Thank you for the review! I addressed your comments, please have another look.

nestkernel/node_manager.h
```diff
@@ -205,11 +205,16 @@ class NodeManager : public ManagerInterface
   void finalize_nodes();
 
   /**
-   *
+   * Returns if any node uses waveform relaxation
```
heplesser: "if" -> "whether"

nestkernel/simulation_manager.cpp
```diff
@@ -551,6 +551,15 @@ nest::SimulationManager::prepare_simulation_()
...
   kernel().model_manager.create_secondary_events_prototypes();
 
+  if ( kernel().mpi_manager.get_num_processes() > 1 )
+  {
+    // Check if any MPI process uses waveform relaxation and set
+    // kernel().node_manager.any_node_uses_wfr() correspondingly
+    // across all MPI processes
+    kernel().node_manager.set_any_node_uses_wfr( kernel().mpi_manager.any_true(
+      kernel().node_manager.any_node_uses_wfr() ) );
+  }
```
heplesser: This code is less legible than I had hoped. Especially, it is not really clear that we exchange local information to get a global picture. I wonder if it would not be better to replace the entire `if` block with a single call
```c++
kernel().node_manager.check_wfr_use();
```
and move the actual check to the `NodeManager` class, replacing `set_any_node_uses_wfr()` with `check_wfr_use()`, which would just be
```c++
{
  wfr_is_used_ = kernel().mpi_manager.any_true( wfr_is_used_ );
}
```
I would also change the `any_node_uses_wfr` variable and method names to `wfr_is_used`, which I think is a little clearer, and would move the check for multiple processes into the `any_true()` method.

nestkernel/mpi_manager.cpp
```diff
@@ -713,6 +713,32 @@ nest::MPIManager::grng_synchrony( unsigned long process_rnd_number )
...
 
+// any_true: takes a single bool, exchanges with all other processes,
+// and returns "true" if one or more processes provide "true"
+bool
+nest::MPIManager::any_true( bool my_bool )
+{
+  if ( get_num_processes() == 1 )
+    return my_bool;
+
+  int my_int = 0;
```
jakobj: this could be written as
```c++
int my_int = my_bool;
```
using implicit conversion bool -> int

nestkernel/mpi_manager.cpp
```diff
@@ -713,6 +713,32 @@ nest::MPIManager::grng_synchrony( unsigned long process_rnd_number )
   return true;
 }
 
+// any_true: takes a single bool, exchanges with all other processes,
+// and returns "true" if one or more processes provide "true"
+bool
+nest::MPIManager::any_true( bool my_bool )
```
jakobj: argument could be const

nestkernel/node_manager.cpp
```diff
@@ -843,6 +843,12 @@ NodeManager::finalize_nodes()
 }
 
 void
+NodeManager::check_wfr_use( void )
```
jakobj: please remove the `void`

janhahne:
@jakobj Thank you for the suggestions. I addressed all of them and also improved the handling of the `coeff_lenght_` parameter. Now it isn't communicated anymore. This is possible since the size of the coeff_array can also be computed in the `check_wfr_use` function. Would you please have another look?

jakobj:
very nice! i'm happy. 👍

nestkernel/mpi_manager.cpp
```diff
@@ -716,14 +716,12 @@ nest::MPIManager::grng_synchrony( unsigned long process_rnd_number )
...
-nest::MPIManager::any_true( bool my_bool )
+nest::MPIManager::any_true( const bool my_bool )
 {
   if ( get_num_processes() == 1 )
     return my_bool;
 
-  int my_int = 0;
-  if ( my_bool )
-    my_int = 1;
+  int my_int = my_bool;
```
heplesser: I think it would be useful to add a comment here, explaining why you need to convert to an int here. Two years from now a reader may not remember that MPI does not support bool. The declaration can also be `const`.

Then on line 734 below (I cannot add comments there in Github), I think it would be even safer to check `all_int[i] != 0`)  instead of checking for `== 1`.

nestkernel/event.h
```diff
@@ -952,6 +952,12 @@ class GapJunctionEvent : public SecondaryEvent
     supported_syn_ids_.push_back( synid );
   }
 
+  static void
+  set_coeff_length( const size_t coeff_length )
+  {
+    coeff_length_ = coeff_length;
```
heplesser: From the context it seems certain that this will never be called from a thread-parallel context, but if that should happen, we would be in trouble. Can we add an assertion here that we are running single-threaded, or explicitly mark this as critical section?

nestkernel/mpi_manager.cpp
janhahne: I will add a comment on this and change the check to `!= 0`. This indeed seems safer, although the C++ standard actually does guarantee that a boolean `true` is converted to `1`

nestkernel/event.h
janhahne: I gave it another thought:

Something like
```cpp
static void
set_coeff_length( const size_t coeff_length )
{
  #pragma omp critical
  coeff_length_ = coeff_length;
}
```
would not solve this problem: There is no parallel region within this function. Also checking the number of threads in this function is no option (this would always be 1).

One just needs to make sure that this function is never called from several threads at the same time. As the `check_wfr_use` function is in a omp single region this is true here. So from my point of view there is nothing we could do here, do you agree?

heplesser: @janhahne I think
```C++
#ifdef _OPENMP
  assert( omp_get_num_threads(void) == 1 );
#endif
```
should make sure this function breaks if it is ever called from a thread-parallel context (see
https://computing.llnl.gov/tutorials/openMP/#OMP_GET_NUM_THREADS).

janhahne: You are right! I tested it by calling the function from a thread-parallel region, it works. I also added this assert to the other two functions from `GapJunctionEvent` that modify static variables

janhahne:
@heplesser Applied your suggestions. Please have a (hopefully) final look.

heplesser:
@janhahne Very nice! I just sent you a PR towards you repo with a little tidy-up. It will automatically appear here when you merge it into your repo and then I will approve and merge this PR.

janhahne:
@heplesser Thank you!

PR #608, opened 2016-12-27
Added gap junction Markdown documentation

janhahne:
As discussed in the meeting on 19.12. this PR adds a markdown documentation file `simulations-with-gap-junctions.md` to the documentation explaining the use of gap junctions in NEST 2.12.

I suggest @steffengraber (is everything in the right place?) and @jschuecker (is the documentation understandable?) as reviewers.

### DIFF
--- a/extras/userdoc/md/documentation/simulations-with-gap-junctions.md
+++ b/extras/userdoc/md/documentation/simulations-with-gap-junctions.md
@@ -3,7 +3,7 @@
 **Note:** This documentation describes the usage of gap junctions in NEST 2.12.
 A documentation for NEST 2.10 can be found in
 [Hahne et al. 2016](http://link.springer.com/chapter/10.1007/978-3-319-50862-7_4).
-It is however recommended to use NEST 2.12 (or higher), 
+It is however recommended to use NEST 2.12 (or later), 
 due to several improvements in terms of usability. 
 
 ## Introduction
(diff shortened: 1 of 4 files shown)


steffengraber:
:+1:  from my side. Documentation and examples are on the right place.

extras/userdoc/md/documentation/simulations-with-gap-junctions.md
```diff
@@ -0,0 +1,121 @@
...
+```python
+import nest
+
+a = nest.Create('hh_psc_alpha_gap')
+b = nest.Create('hh_psc_alpha_gap')
+# Create gap junction between neurons a and b
+nest.Connect(a, b, {'rule': 'one_to_one'}, {'model': 'gap_junction', 'weight': 0.5})
+nest.Connect(b, a, {'rule': 'one_to_one'}, {'model': 'gap_junction', 'weight': 0.5})
+```
+
```
jschuecker: It might be dangerous to start with code lines which are actually not working, in particular if the user quickly wants to look up the call and does not read the whole documentation. I would suggest to skip the lines 16 to 28 and directly start with the correct function call. If you think these lines are necessary you could put them as a **note** after the correct call (around line 50) and say something like "these lines would in principle lead to the same gap junction connection but are forbidden in NEST 2.12 to prevent the user from creating incomplete gap junctions and thus result in an error."

jschuecker:
@janhahne: nice and clear documentation, I have only one issue (see my in line comment).

janhahne:
@jschuecker Thank you for the good suggestion. I removed the first script and it indeed improved readability. Could you have another look and approve the PR if you are satisfied?

jschuecker:
@janhahne  I had a look again, :+1:  from my side.

PR #614, opened 2017-01-04
Fix minor encoding issue

heplesser:
Some source code files contain Umlaute in author names, encoded in UTF-8. The copyright checking script opened files with the default encoding for the system on which it runs, which could lead to problems. This PR enforces opening files with UTF-8. The PR also replaces some non-ASCII punctuation with their ASCII equivalents.


testsuite/regressiontests/ticket-659-copyright.py
```diff
@@ -121,7 +125,7 @@
                 line_exp = template_line.replace('{{file_name}}', fname)
                 if line_src != line_exp:
                     print("Incorrect copyright header in '{0}':".format(tested_file))           # noqa
-                    print("    expected -> '{0}', actual -> '{1}'\n".format(line_exp.strip(),   # noqa
+                    print("    expected -> '{0}', actual -> '{1}'\n".format(line_exp.strip(),  # noqa
                                                                             line_src.strip()))  # noqa
                     total_errors += 1
```
jougs: This change looks wrong to me as it destroys the column of comments. Or is this just a glitch in the rendering by GitHub?

jougs:
Very good! These things drove me crazy so often. Thanks for fixing.

testsuite/regressiontests/ticket-659-copyright.py
heplesser: This change was done magically by pep8ify. I will have another look, not quite sure why the "noqa's" should be necessary.

PR #618, opened 2017-01-05
correct hl_api_helper import (fix #617)

Silmathoron:
Fix #617 (wrong import for python 3)

Issue #617, opened 2017-01-05
Wrong import in hl_api from python 3


Silmathoron:
After merging #580, an incorrect import for python 3 was included, this make ``import nest`` lead to

    /home/tfardet/.nest-2.10.0-install/lib64/python3.5/site-packages/nest/lib/hl_api_connections.py in <module>()
         29 from .hl_api_simulation import GetKernelStatus, SetKernelStatus
         30 from .hl_api_subnets import GetChildren
    ---> 31 import hl_api_helper as hlh
         32 import numpy
         33 
    
    ImportError: No module named 'hl_api_helper'

heplesser:
Fixed by merge of #620.

PR #620, opened 2017-01-05
Fix py3 issues

heplesser:
This PR fixes a number of Python 3 issues, including incorrect imports in pynest/nest/lib, thus addressing #617 among other issues. It is a superset of #618.

### DIFF
--- a/examples/nest/Potjans_2014/spike_analysis.py
+++ b/examples/nest/Potjans_2014/spike_analysis.py
@@ -57,16 +57,16 @@ gids = []
 for l in gidfile:
     a = l.split()
     gids.append([int(a[0]), int(a[1])])
-print 'Global IDs:'
-print gids
-print
+print('Global IDs:')
+print(gids)
+print()
 
 # number of populations
 
 num_pops = len(gids)
-print 'Number of populations:'
-print num_pops
-print
+print('Number of populations:')
+print(num_pops)
+print()
 
 # first GID in each population
 
@@ -98,9 +98,9 @@ last_gids = [int(np.sum(pop_sizes[:i + 1]))
 
 Pop_sizes = [pop_sizes[i:i + 2]
              for i in xrange(0, len(pop_sizes), 2)]
-print 'Population sizes:'
-print Pop_sizes
-print
+print('Population sizes:')
+print(Pop_sizes)
+print()
 
 Raw_first_gids = [raw_first_gids[i:i + 2] for i in
                   xrange(0, len(raw_first_gids), 2)]
@@ -112,9 +112,9 @@ Last_gids = [last_gids[i:i + 2] for i in xrange(0, len(last_gids), 2)]
 # total number of neurons in the simulation
 
 num_neurons = last_gids[len(last_gids) - 1]
-print 'Total number of neurons:'
-print num_neurons
-print
+print('Total number of neurons:')
+print(num_neurons)
+print()
 
 # load spikes from gdf files, correct GIDs and merge them in population files,
 # and store spike trains
@@ -135,8 +135,8 @@ for layer in ['0', '1', '2', '3']:
         file_pattern = os.path.join(datapath,
                                     'spikes_{}_{}*'.format(layer, population))
         files = glob.glob(file_pattern)
-        print 'Merge ' + str(
-            len(files)) + ' spike files from L' + layer + 'P' + population
+        print('Merge ' + str(
+            len(files)) + ' spike files from L' + layer + 'P' + population)
         if files:
             merged_file = open(output, 'w')
             for f in files:
@@ -190,9 +190,9 @@ for i in np.arange(num_pops):
     rates.append(temp / (rec_sizes[i] * (T - T_start)) * 1e3)
     temp = 0
 
-print
-print 'Firing rates:'
-print rates
+print()
+print('Firing rates:')
+print(rates)
 
 plt.figure(2)
 ticks = np.arange(num_pops)
--- a/extras/help_generator/writers.py
+++ b/extras/help_generator/writers.py
@@ -212,7 +212,7 @@ def write_helpindex(helpdir):
                 hlp_list.append(name + '\t' * (16 - min(c, 60) // 4)
                                 + fullname)
             elif not os.path.isfile(item):
-                print 'WARNING: Checkfile ' + item + ' not exist.'
+                print('WARNING: Checkfile ' + item + ' not exist.')
 
         html_list.append('</table></center>')
         html_list.append('</table></center>')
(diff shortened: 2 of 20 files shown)


pynest/nest/lib/hl_api_connections.py
```diff
@@ -453,15 +452,15 @@ def _is_subnet_instance(gids):
...
-        # Turn off _deprecation_warning as users shouldn't change
-        # implementation of this function, it is done by the developers
-        deprecation_bool = hlh._deprecation_warning['GetChildren']
-        hlh._deprecation_warning['GetChildren'] = False
+        # Turn off deprecation warning to avoid confusing users with
+        # internals.
+        # deprecation_status = suppress_deprecation_warning('GetChildren')
 
-        GetChildren(gids)
+        with SuppressedDeprecationWarning(['GetChildren']):
```
Silmathoron: Where is this defined? Couldn't find it in ``hl_api_helper``
**EDIT:** ok, just below ^^

pynest/nest/lib/hl_api_connections.py
```diff
@@ -453,15 +452,15 @@ def _is_subnet_instance(gids):
...
+            GetChildren(gids)
 
         # Need to reset the deprecation warning to its old value
-        hlh._deprecation_warning['GetChildren'] = deprecation_bool
+        # restore_deprecation_warning('GetChildren', deprecation_status)
```
Silmathoron: why is this here if it is commented (same question for L457)

pynest/nest/lib/hl_api_helper.py
```diff
@@ -424,63 +454,43 @@ def model_deprecation_warning(model):
         warnings.warn('\n' + text)
 
 
-def turn_off_deprecation_warning(deprecated_func_name):
-    """Turns off deprecation warnings on SLI and Python level for given
-    function.
+class SuppressedDeprecationWarning(object):
```
Silmathoron: I think this should not be exposed to users, so maybe add a ``_`` before and a ``__all__`` in the file?

pynest/nest/lib/hl_api_helper.py
```diff
@@ -402,6 +400,38 @@ def broadcast(item, length, allowed_types, name="item"):
     return item
 
 
+@check_stack
```
Silmathoron: Why did you switch it here? (I don't mind, though, just curious)

pynest/nest/lib/hl_api_subnets.py
```diff
@@ -49,16 +47,9 @@ def PrintNetwork(depth=1, subnet=None):
...
-        # developers.
-        deprecation_bool, verbosity_level = turn_off_deprecation_warning(
-            'CurrentSubnet')
-        subnet = CurrentSubnet()
-        # Need to reset the deprecation warning to its old value
-        turn_on_deprecation_warning('CurrentSubnet', deprecation_bool,
-                                    verbosity_level)
-
+        # Avoid confusing user by deprecation warning
+        with SuppressedDeprecationWarning('CurrentSubnet'):
```
Silmathoron: Note that if you indeed chose to make this not available to users (see comment above), you'll need to import it explicitly

topology/pynest/hl_api.py
```diff
@@ -1619,20 +1619,12 @@ def GetTargetNodes(sources, tgt_layer, tgt_model=None, syn_model=None):
...
-
-    # obtain local nodes in target layer, to pass to GetConnections
-    tgt_nodes = nest.GetLeaves(tgt_layer,
-                               properties={
-                                   'model': tgt_model} if tgt_model else None,
-                               local_only=True)[0]
-    # Need to reset the deprecation warning to its old value
-    nest.turn_on_deprecation_warning('GetLeaves', deprecation_bool,
-                                     verbosity_level)
+    with nest.SuppressDeprecationWarning('GetLeaves'):
```
Silmathoron: Same comment as above on accessibility of this function

examples/nest/Potjans_2014/spike_analysis.py
```diff
@@ -57,16 +57,16 @@
...
+print(gids)
+print()
 
 # number of populations
 
 num_pops = len(gids)
-print 'Number of populations:'
-print num_pops
-print
+print('Number of populations:')
```
heplesser: I agree with you in principle, but in this case I prioritised getting things done over doing them perfectly. Example beautification can come in a separate PR later.

extras/help_generator/writers.py
```diff
@@ -59,10 +59,10 @@ def write_help_html(doc_dic, helpdir, fname, sli_command_list, keywords):
     hlplist = []
     name = ''
 
-    for key, value in doc_dic.iteritems():
+    for key, value in doc_dic.items():
```
heplesser: I think wrapping with `iter()` will not provide any gains and rather make the code less readable. The only point of using `iteritems()` in Python 2 is to get a generator instead of a full list of items. `iter(d.items())` will still create the full list and then wrap an iterator around it in Python 2.  And `for k, v in d.items():` is the pythonic way as far as I know.

pynest/nest/lib/hl_api_connections.py
heplesser: Fixed.

pynest/nest/lib/hl_api_helper.py
heplesser: I ran into a circular inclusion problem otherwise. The `SuppressedDeprecationWarning` class definition fits best here among the helpers. It needs to manipulate the verbosity level. Having the verbosity functions defined in `hl_api_info.py` requires helper to include info an vice-versa. Moving the verbosity function definitions here means that `hl_api_helper.py` does not need to include any other `hl_api_*.py` files.

examples/nest/Potjans_2014/spike_analysis.py
Silmathoron: No problem

extras/help_generator/writers.py
Silmathoron: Ok, I thought they would find a clever way to do this in py2... but if the list is created, then there is indeed really no point

pynest/nest/lib/hl_api_helper.py
Silmathoron: ok

pynest/nest/lib/hl_api_helper.py
heplesser: Fixed. I will add explicit imports where needed.

heplesser: I thought about it a little more. `SuppressedDeprecationWarning` is also needed in the Topology hl_api, so I would have to access it there as a private member; possible, but slightly breaking with the logic of privacy. It is also not an implementation detail that may change at any time and does not require privacy from that perspective. Users should realise that they should use this context with care based on its name and documentation. Thus, I will leave the name as it is after all.

Silmathoron: Fair enough

PR #628, opened 2017-01-11
Add Allreduce_min and Allreduce_max functions, use them to determine delay extrema

jakobj:
This PR replaces MPI::Allgather and subsequent std::min/max_element with a single call to MPI::Allreduce with the min/max operations to reduce overhead for a large number of MPI processes. Some timing measurements will follow, but someone with experience in the min delay intricacies could already comment on whether this change is appropiate.
I suggest @heplesser @jougs @suku248 as reviewers.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -344,19 +344,19 @@ nest::ConnectionManager::update_delay_extrema_()
 
   if ( kernel().mpi_manager.get_num_processes() > 1 )
   {
-    std::vector< delay > min_delays( kernel().mpi_manager.get_num_processes() );
-    min_delays[ kernel().mpi_manager.get_rank() ] = min_delay_;
-    kernel().mpi_manager.communicate( min_delays );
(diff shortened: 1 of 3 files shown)


jakobj:
I am bringing this issue up, since every call to `GetStatus` causes an update of delay extrema (https://github.com/nest/nest-simulator/blob/master/nestkernel/connection_manager.cpp#L147), and hence communcation of min/max delays via MPI. This is quite painful for large scale simulations with many calls to `GetStatus` (like the hpc_benchmark). I will create a separate PR in which I have removed most `GetStatus` calls to avoid communication overhead.

jakobj:
Nevermind, here are the timing results for 8192 with one call to `GetStatus`
`MPIAllgather`: 2.415666666666667s
`MPIAllreduce`: 76.03766666666667s
Can anyone explain these strange results? Shouldn't Allreduce outperform Allgather?

heplesser:
@jakobj Given the benchmark results,  I close this PR for now. Feel free to re-open it or to create a new one once benchmarks look better.

Issue #574, opened 2016-12-03
`sudo make install` leaves a `doc/help` directory owned by root in the source tree


sanjayankur31:
Description of problem:
-----------------------------

Running `make install` to install nest to the expected directory, such as `/opt` leaves a `doc/help` directory owned by the `root` user in the source tree. 

Version-Release or Git commit:
----------------------------------------

```
[asinha@ankur  nest-simulator(master %=)]$ git lg | head -1
*   782d30a - (2 days ago) Merge pull request #573 from lekshmideepu/travis_build_log_fix — Jochen Martin Eppler 
```

How reproducible: (Always/Sometimes/Unsure)
------------------------------------------------------------

Always

Steps to Reproduce:
---------------------------

1. Run cmake:

```
INSTALL_PATH="/opt/"
cmake -DCMAKE_INSTALL_PREFIX:PATH=$INSTALL_PATH -Dwith-python:BOOL=ON -Dwith-mpi:BOOL=ON .
```

2. Run make and make install

```
make -j6; sudo make install; 
```

Actual results:
------------------

A `doc/help` directory that is owned by `root` is left behind after the `sudo make install`.

Expected results:
----------------------

Directory should not be left behind, and certainly shouldn't be owned by root? Only noticed it when `git clean -dfx` failed to completely clear the source tree.

Additional info:
-------------------
Directory is not created after `make`, it only turns up after `make install`:

```
[asinha@ankur  nest-simulator(master %=)]$ pwd
/home/asinha/Documents/02_Code/00_repos/01_others/nest-simulator

[asinha@ankur  nest-simulator(master %=)]$ lash /opt/share/doc/nest/help/ ./doc/help/
./doc/help/:
total 348K
 36K drwxr-xr-x. 2 root root  36K Dec  3 11:22 cc
 72K -rw-r--r--. 1 root root  71K Dec  3 11:22 helpindex.hlp
172K -rw-r--r--. 1 root root 172K Dec  3 11:22 helpindex.html
 68K drwxr-xr-x. 2 root root  64K Dec  3 11:22 sli

/opt/share/doc/nest/help/:
total 348K
 36K drwxr-xr-x. 2 root root  36K Dec  3 11:22 cc
 72K -rw-r--r--. 1 root root  71K Dec  3 11:22 helpindex.hlp
172K -rw-r--r--. 1 root root 172K Dec  3 11:22 helpindex.html
 68K drwxr-xr-x. 2 root root  68K Dec  3 11:22 sli
[...]

PR #630, opened 2017-01-13
Remove `doc/help` directory in the source tree after installation

steffengraber:
Fix for #574.

### DIFF
--- a/extras/help_generator/parse_help.py
+++ b/extras/help_generator/parse_help.py
@@ -135,3 +135,4 @@ write_helpindex(helpdir)
 
 shutil.rmtree(install_dir, ignore_errors=True)
 shutil.copytree(helpdir, install_dir)
+shutil.rmtree(helpdir, ignore_errors=True)
--- a/extras/help_generator/parse_help.py
+++ b/extras/help_generator/parse_help.py
@@ -134,5 +134,4 @@ for fname in allfiles:
 write_helpindex(helpdir)
 
 shutil.rmtree(install_dir, ignore_errors=True)
-shutil.copytree(helpdir, install_dir)
-shutil.rmtree(helpdir, ignore_errors=True)
+shutil.move(helpdir, install_dir)


extras/help_generator/parse_help.py
```diff
@@ -135,3 +135,4 @@
 
 shutil.rmtree(install_dir, ignore_errors=True)
 shutil.copytree(helpdir, install_dir)
+shutil.rmtree(helpdir, ignore_errors=True)
```
heplesser: Wouldn't it be more efficient to use `shutil.move()` instead of `copytree()` followed by `rmtree()`?

steffengraber: Yes, better. I changed it.

jougs:
I think the actual problem with #574 is that help is generated during the installation step (which may be run by root) rather than earlier on (where things are normally run by an unprivileged user). In this sense, this PR just addresses a symptom and not the cause of the problem.

If help was generated after compilation, the resulting files would belong to the user which ran `make all` and installation could then be just a plain copy as it is already for most other artifacts. That final step could be run by root or normal user without any unclear consequences for file ownerships in the build directory.

@steffengraber: do you think you can make this change? This would involve changing the [current hook for help extraction](https://github.com/steffengraber/nest-simulator/blob/eb34551fabdb8f5da2150652ff2ed53d12963d33/cmake/GenerateHelp.cmake#L51) to a [build event](https://cmake.org/cmake/help/v3.2/command/add_custom_command.html#build-events) added to the appropriate target and adding the generated help to an appropriate [install](https://cmake.org/cmake/help/v3.2/command/install.html?highlight=install#installing-directories) block in the `doc/CMakeLists.txt` file. Something similar would also be required for the build system of the `MyModule` example.

If you have troubles with this, I can implement the changes. It is probably anyway my fault that this problem exists at all ;-)

steffengraber:
@jougs Sorry, I know nothing about cmake. It would be nice if you could do the changes.

jougs:
Replaced by #647.

PR #634, opened 2017-01-19
Update setting of dlopenflags() to current Python standards (fixes #632)

### LINKED ISSUE #632: Py3.6: invalid mode for dlopen()
With Python 3.6, loading ``pynestkernel.so`` with dlopen cannot be done with the actual ``__init__.py``

```python
try:
    try:
        import dl
    except:
        import DLFCN as dl
    sys.setdlopenflags(dl.RTLD_NOW | dl.RTLD_GLOBAL)
except:
    # this is a hack for Python 2.6 on Mac, where RTDL_NOW is nowhere
    # to be found. See trac ticket #397
    import ctypes
    sys.setdlopenflags(ctypes.RTLD_GLOBAL)
```

With Py3.6, then correct call (or at least one that seems to work) is:
`` sys.setdlopenflags(os.RTLD_GLOBAL)`` (see [PyDoc](https://docs.python.org/3/library/sys.html#sys.setdlopenflags))

A correction could be

```python
try:
    try:
        import dl
    except:
        try:
            import DLFCN as dl
        except:
            import os as dl
    sys.setdlopenflags(dl.RTLD_NOW | dl.RTLD_GLOBAL)
except:
    # this is a hack for Python 2.6 on Mac, where RTDL_NOW is nowhere
    # to be found. See trac ticket #397
    import ctypes
    sys.setdlopenflags(ctypes.RTLD_GLOBAL)
```

If I understand correctly the [doc](https://docs.python.org/3/library/os.html#os.RTLD_GLOBAL), I think the dl flags were switch in ``os`` from 3.3, but I don't know if we can do better since I don't know what versions ``DLFCN`` is supposed to cover.

heplesser:
This solves Python 3.6 compatibility problems (#632) and removes code deprecated since Python 2.6.

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -45,14 +45,27 @@ except:
 try:
     # Python 3.3 and later has flags in os
     sys.setdlopenflags(os.RTLD_NOW | os.RTLD_GLOBAL)
-except:
-    # Python 2.6 and later have flags in ctypes, RTLD_NOW may be missing
-    import ctypes
+except AttributeError:
+    # Python 2.6 and 2.7 have flags in ctypes, but RTLD_NOW may only
+    # be available in dl or DLFCN and is required at least under 
+    # Ubuntu 14.04. The latter two are not available under OSX,
+    # but OSX does not have and does not need RTLD_NOW. We therefore
+    # first try dl and DLFCN, then ctypes just for OSX.
     try:
-        sys.setdlopenflags(ctypes.RTLD_GLOBAL | ctypes.RTLD_NOW)
-    except:
-        # RTLD_NOW missing from ctypes eg on OSX
-        sys.setdlopenflags(ctypes.RTLD_GLOBAL)
+        import dl
+        sys.setdlopenflags(dl.RTLD_GLOBAL | dl.RTLD_NOW)
+    except ImportError, AttributeError:
+        try:
+            import DLFCN
+            sys.setdlopenflags(DLFCN.RTLD_GLOBAL | DLFCN.RTLD_NOW)
+        except ImportError, AttributeError:
+            import ctypes
+            try:
+                sys.setdlopenflags(ctypes.RTLD_GLOBAL | ctypes.RTLD_NOW)
+            except AttributeError:
+                # We must test this last, since it is the only case without
+                # RTLD_NOW (OSX)
+                sys.setdlopenflags(ctypes.RTLD_GLOBAL)
 
 from . import pynestkernel as _kernel      # noqa
 from .lib import hl_api_helper as hl_api   # noqa
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -39,20 +39,20 @@ try:
 except:
     pass
 
-# This is a workaround to make MPI-enabled NEST import properly. The
-# underlying problem is that the shared object pynestkernel
-# dynamically opens other libraries that open other libraries...
+# Make MPI-enabled NEST import properly. The underlying problem is that the
+# shared object pynestkernel dynamically opens other libraries that open
+# yet other libraries.
 try:
-    try:
-        import dl
-    except:
-        import DLFCN as dl
-    sys.setdlopenflags(dl.RTLD_NOW | dl.RTLD_GLOBAL)
+    # Python 3.3 and later has flags in os
+    sys.setdlopenflags(os.RTLD_NOW | os.RTLD_GLOBAL)
 except:
-    # this is a hack for Python 2.6 on Mac, where RTDL_NOW is nowhere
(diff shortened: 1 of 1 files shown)


heplesser:
@Silmathoron Could you test this with Python 3.6? I have only tested 3.5 and 2.7 on OSX, while Travis will test 2.7 under Linux.

heplesser:
Hm, seems that at least Ubuntu 14.04 does not provide `RTLD_NOW` in `ctypes`, and that importing `pynestkernel` fails if it is not set. So we may need to import `DLFCN`. I will update the PR tomorrow.

heplesser:
@Silmathoron I just (about five minutes ago) pushed some fixes. Is your 3.6 approval with those fixes, or was that based on my push from about 15 hours ago?

Silmathoron:
Yes, don't worry ;)

DimitriPlotnikov:
Dear @Silmathoron and @heplesser,

I have been working quite a bit on the PR and corresponding issue. Now, I'm a bit confused what is  the actual problem been solved in the.

I'll explain what did I try. In a clean Ubuntu 16.4.1LTS with python 3.6 I tried:

Building nest **with** python3 flags (-Dwith-python=3)
nest-simulator-master: python2 doesn't work with nest. python3 interpreter works with nest. 
nest-simulator-PR: python2 doesn't work with nest. python3 interpreter works with nest.

Building nest **without** python 3 flags (-Dwith-python=3)
nest-simulator-master: python2 works with nest. python3 doesn't work with nest (problem mentioned in the issue).
nest-simulator-PR: python2 works with nest. python3 works with nest.

In my opinion It would be helpful to complement the issue with the information about:
a) Which software version were used
b) How nest-simulator were build
c) How to reproduce the error

Silmathoron:
Hi @DimitriPlotnikov!
Unfortunately I'm a bit unsure about what your problem is (and I think there are unfortunate copy-paste in your comment ^^")

Just to make sure everything is clear:

- using ``-Dwith-python=X`` will install PyNEST for ``python X.Y`` (with highest available ``Y``) only, e.g. ``-Dwith-python=2`` will most probably install it in the path of ``python 2.7``, meaning that any ``python 3.Y`` on your machine won't be able to use it (and the reverse is true if you use ``-Dwith-python=3``
- the actual problem with the previous ``__init__.py`` was that, for ``python 3.6`` (i.e. installing with ``-Dwith-python=3`` and ``python 3.6`` available on your machine), the ``import`` failed.

Thus, HEP's changes make sure that this problem is corrected:

1. ``sys.setdlopenflags(os.RTLD_NOW | os.RTLD_GLOBAL)`` works from 3.3 on
2. ``sys.setdlopenflags(dl.RTLD_GLOBAL | dl.RTLD_NOW)`` should work for 2.6--3.2 on most platforms
3. if not one of ``sys.setdlopenflags(DLFCN.RTLD_GLOBAL | DLFCN.RTLD_NOW)`` or ``sys.setdlopenflags(ctypes.RTLD_GLOBAL | ctypes.RTLD_NOW)`` should do the trick
4. unless we're on a mac, then we need ``sys.setdlopenflags(ctypes.RTLD_GLOBAL)``

DimitriPlotnikov:
@Silmathoron 

You provided me the information which I more or less wanted to know. I'll conduct a test with `python 3.6`

And no, there is no copy&paste errors in my previous message. I just summarized my observations and tests which I done.

Silmathoron:
> Building nest without python 3 flags (`-Dwith-python=3`)

Looks highly suspect to me, though... but ok, let us know ;)

DimitriPlotnikov:
From: https://github.com/nest/nest-simulator/blob/master/INSTALL

```
-Dwith-python=[OFF|ON|2|3]                   Build PyNEST. To set a specific Python version, set 2 or 3. [default=ON]
```
Since there is no information regarding such options in the PR and issue I tried to reconstruct it from the docu. If I did it wrong, please, let me known.

Silmathoron:
Sorry, I should have made it clearer: I just meant I think you used 3 when you wanted to use 2.
I would expect

> Building nest **without** python 3 flags (``-Dwith-python=``**2**)

But as I said, I think it's a copy-paste error since you seem to understand the option without any problem ;)

heplesser:
@DimitriPlotnikov Did your test with Python 3.6 go well?

DimitriPlotnikov:
@heplesser 
I'm still actively testing the PR in different configurations. Update the PR and issue today during the day.

DimitriPlotnikov:
Tested the PR in a Ubuntu 16.10 instance with following python version: ```2.7.12```, ```3.5```, ```3.6```. I was always able to ```import nest``` in every of the mentioned python interpreters.

heplesser:
Summary: NEST now works with Python 3.6.

PR #642, opened 2017-01-26
Added check on threading in two python tests which require it.

sdiazpier:
This PR solves a problem with two python tests which require threading but were not being skipped if threading was not enabled.

### DIFF
--- a/pynest/nest/tests/test_sp/test_enable_multithread.py
+++ b/pynest/nest/tests/test_sp/test_enable_multithread.py
@@ -29,7 +29,10 @@ __author__ = 'sdiaz'
 # and multiple threads are set, or if multiple threads are set and
 # the enable_structural_plasticity function is called.
 
+HAVE_OPENMP = nest.sli_func("statusdict/threading ::")
 
+
+@unittest.skipIf(HAVE_OPENMP == 'no', 'OpenMP is not available')
 class TestEnableMultithread(unittest.TestCase):
 
     def setUp(self):
--- a/pynest/nest/tests/test_sp/test_enable_multithread.py
+++ b/pynest/nest/tests/test_sp/test_enable_multithread.py
@@ -29,10 +29,10 @@ __author__ = 'sdiaz'
 # and multiple threads are set, or if multiple threads are set and
 # the enable_structural_plasticity function is called.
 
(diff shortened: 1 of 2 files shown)


pynest/nest/tests/test_sp/test_enable_multithread.py
```diff
@@ -29,7 +29,10 @@
 # and multiple threads are set, or if multiple threads are set and
 # the enable_structural_plasticity function is called.
 
+HAVE_OPENMP = nest.sli_func("statusdict/threading ::")
```
heplesser: It would be even better to do as follows:
```Python
HAVE_OPENMP = nest.sli_func("is_threaded")
```
This insulates the details of how threading is marked at the sli level. The result is a bool.

pynest/nest/tests/test_sp/test_enable_multithread.py
```diff
@@ -29,7 +29,10 @@
...
 
+
+@unittest.skipIf(HAVE_OPENMP == 'no', 'OpenMP is not available')
```
heplesser: With the change above, this would simplify to
```Python
@unittest.skipIf(HAVE_OPENMP, 'OpenMP is not available')
```

pynest/nest/tests/test_sp/test_growth_curves.py
```diff
@@ -28,6 +28,7 @@
 import nest
 from nest import raster_plot
 import time
+HAVE_OPENMP = nest.sli_func("statusdict/threading ::")
```
heplesser: See above

pynest/nest/tests/test_sp/test_growth_curves.py
```diff
@@ -216,6 +217,7 @@ def growth_curve(self, t):
         )
 
 
+@unittest.skipIf(HAVE_OPENMP == 'no', 'OpenMP is not available')
```
heplesser: See above.

sdiazpier:
Dear @heplesser thanks for the suggestion. I have made the changes. Please let me know if anything else is missing.

heplesser:
Summary: Suppresses thread-dependent threads if NEST was build without threads.

PR #647, opened 2017-01-27
Separate help extraction and installation

jougs:
This PR replaces #630 and constitutes a fix for #574. For details about the underlying problem, see the descriptions there.

In addition to separating help extraction from installation, it
* extends the configuration summary of the `MyModule` example
* adds natural order sorting for the helpindex
* re-generates the helpindex upon installation of additional help files


doc/CMakeLists.txt
```diff
@@ -17,6 +17,23 @@
...
+    WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/extras/help_generator"
+    )
+  install( DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/help"
+    DESTINATION "${CMAKE_INSTALL_DOCDIR}"
+    )
+  install( CODE
+    "execute_process(
+      COMMAND python -B generate_helpindex.py \"${CMAKE_INSTALL_DOCDIR}\"
+      WORKING_DIRECTORY \"${PROJECT_SOURCE_DIR}/extras/help_generator\"
+    )"
```
heplesser: I am a bit confused why you need to repeat the same code that is given in `add_custom_command()` here again; my confusion is probably just due to my very limited experience with CMake.

extras/help_generator/helpers.py
```diff
@@ -66,15 +66,13 @@ def makedirs(path):
     """
     try:
         os.makedirs(path)
-    except OSError as exc:
-        if exc.errno != errno.EEXIST or os.path.isdir(path):
-            raise
+    except:
+        pass
```
jougs: You are right. The test failed on me during testing so I've removed it and forgot to put it back in later on. A fixed version of the original check is back in cd2c7f7.

extras/help_generator/writers.py
```diff
@@ -184,13 +184,14 @@ def write_helpindex(helpdir):
         html_list.append('</tr></table></center>')
         html_list.append('<center><table class="commands" id="%s">'
                          % doubles[0])
-        for item in filelist:
+        keyfunc = lambda x: x.lower().rsplit('/', 1)[1]
+        for item in sorted(filelist, key=keyfunc):
```
jougs: The main reason to do it like this was to respect the line length limit of PEP8. But you are right with the lambda and I've factored `keyfunc` out into a proper function in ae6880b.

doc/CMakeLists.txt
jougs: Actually it is not the same code. The code in `add_custom_command()` extracts the help during the `make` step and puts the result in `doc/help` in the current build directory together with a `helpindex.html` containing all help files in that directors. This operation is performed by the script `generate_help.py`. The code in `execute_process()` re-generates the help index so that it contains *all* help files from the global install directory and not just the ones generated in the last call. The script used here is `generate_helpindex.py`. This two-step process is needed to allow modules to install their help to the global help directory and still have an up-to-date index. Otherwise the index would just contain the help files generated by the last call.

I've added comments to the corresponding `CMakeLists.txt` files to clarify this in d739c2a.

jougs:
@heplesser: thanks for the review and testing. I've addressed all of your concerns, see comments above.

The issue with missing help files for `MyModule` was very subtle. Actually I expected people to not directly build it out-of-tree, but rather make a copy somewhere. Thus, the check to exclude help extraction from `MyModule` when the whole NEST directory was built was too strict. I've fixed that in 74f3379.

extras/help_generator/writers.py
```diff
@@ -184,8 +187,7 @@ def write_helpindex(helpdir):
         html_list.append('</tr></table></center>')
         html_list.append('<center><table class="commands" id="%s">'
                          % doubles[0])
-        keyfunc = lambda x: x.lower().rsplit('/', 1)[1]
-        for item in sorted(filelist, key=keyfunc):
+        for item in sorted(filelist, key=key_func):
```
jougs: There's no technical reason. I find find these multi-line expressions ugly and try to avoid them whenever possible.

jougs:
I've answered to your suggestions in https://github.com/jougs/nest-simulator/pull/2.

heplesser:
@sanjayankur31 Thank you for your review! But I was confused by "git clean -dfx functions without sudo." What do you mean by that?

heplesser:
Summary: NEST Help files will now be generated during build, not during install. A full set of documentation is created in the build directory, and later installed in the install directory. The help index in the install directory will also include components installed by user modules.

sanjayankur31:
@heplesser - sorry, should've linked to the issue - was confirming that the PR also fixes #574.

PR #650, opened 2017-01-31
split simulate into prepare, run and finalize

apeyser:
Needs careful check on simulation_manager, given that it's a conflict bottleneck on master.

As requested by several projects including visualization, this PR makes it possible to split a run into multiple time chunks, without doing the calibration and teardown steps for every run.
Thus, simulate(n) = prepare(); run(n); cleanup();
as well as prepare(); run(n/2); run(n/2); cleanup();

A follow-up issue would be to make a run interruptible, returning the time actually run so that the simulation can be continued with another run call.


nestkernel/nest.h
```diff
@@ -82,6 +82,9 @@ ArrayDatum get_connections( const DictionaryDatum& dict );
 
 void simulate( const double& t );
 void resume_simulation();
+void run( const double& t );
+void prepare();
+void cleanup();
```
heplesser: Since this is the high-level API, the functions should be document with doxygen comments.
And to we still need `resume_simulation()` after this split?

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -39,6 +39,49 @@ def Simulate(t):
...
+        Time to simulate in ms
+    """
+
+    sr('Cleanup')
+
+@contextmanager
+def IterateRuns():
+    Prepare()
+    try: yield
+    finally: Cleanup()
```
heplesser: This Python is slightly beyond me ... It seems `yield` yields nothing and I cannot see when this generator would fail to yield anything and thus exit.

nestkernel/simulation_manager.cpp
```diff
@@ -443,13 +499,17 @@ nest::SimulationManager::simulate( Time const& t )
...
+nest::SimulationManager::cleanup()
+{
   finalize_simulation_();
 }
 
 void
-nest::SimulationManager::resume_( size_t num_active_nodes )
+nest::SimulationManager::resume_()
 {
   assert( kernel().is_initialized() and not inconsistent_state_ );
```
heplesser: The next section of code is lengthy and thus puts together a status message. Should that be delegated to a method, so that the flow of important logic becomes more transparent here?

nestkernel/simulation_manager.cpp
```diff
@@ -375,6 +376,61 @@ nest::SimulationManager::simulate( Time const& t )
...
+  prepare();
+  run( t );
+  cleanup();
+}
+
+void
+nest::SimulationManager::run( Time const& t )
+{
   if ( t == Time::ms( 0.0 ) )
     return;
```
heplesser: The next section of code is about error checking. Should that be delegated to a separate method to make the flow of logic clearer?

nestkernel/nest.h
apeyser: prepare -> run -> run -> ... -> run -> cleanup
resume_simulation has a different function, to run a simulation after cleanup. But it should be checked, as you point out. I would, however, make that another issue -- I wouldn't remove any functionality until we have this functionality in.

nestkernel/simulation_manager.cpp
apeyser: Maybe -- I tried to minimize changes, this was all in simulate before.

nestkernel/simulation_manager.cpp
```diff
@@ -443,13 +499,17 @@ nest::SimulationManager::simulate( Time const& t )
       "is called repeatedly with simulation times that are not multiples of "
       "the minimal delay." );
 
-  resume_( num_active_nodes );
+  resume_();
```
apeyser: ✅

nestkernel/simulation_manager.cpp
apeyser: Same as before -- maybe. How much should we take advantage here to clean up code, and how much should we avoid making extra changes in this same set of commits?

nestkernel/simulation_manager.h
```diff
@@ -163,6 +166,7 @@ class SimulationManager : public ManagerInterface
                             //!< relaxation
   size_t wfr_interpolation_order_; //!< interpolation order for waveform
                                    //!< relaxation method
+  size_t num_active_nodes;
```
apeyser: ✅

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -39,6 +39,49 @@ def Simulate(t):
     sps(float(t))
     sr('ms Simulate')
 
+@check_stack
+def Run(t):
+    """Simulate the network for t milliseconds.
```
apeyser: ✅

pynest/nest/lib/hl_api_simulation.py
apeyser: The name is wrong -- it's not iterate runs, it's manage runs. It should also be commented as to use:
```
with nest.IterateRuns():
  for i in range(0, 10):
    nest.run(5)
```

It's the messed up fix for a lack of proper lambda's in python. See `contextlib.contextmanager`.

pynest/nest/tests/test_split.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(TestSplit, self).__init__(*args, **kwargs)
+        self.spike = None
+
+    def setup(self):
+        nest.ResetKernel()
+        nest.SetDefaults('spike_detector', {'withtime': True})
+
+        n1 = nest.Create("iaf_neuron")
```
apeyser: ✅

apeyser:
Do we have a proper image to run clang in? The formatting is tightly fixed to clang-3.6 which is no longer available in many distributions.

nestkernel/nest.h
apeyser: Pointer to an example of the doxygen documentation? It hasn't been used in this header previously, so I'm not precisely sure without an example.

nestkernel/nestmodule.cpp
```diff
@@ -474,6 +474,39 @@ NestModule::SimulateFunction::execute( SLIInterpreter* i ) const
...
+  run( time );
+
+  // successful end of simulate
+  i->OStack.pop();
+  i->EStack.pop();
+}
+
+
+void
+NestModule::PrepareFunction::execute( SLIInterpreter* i ) const
```
apeyser: ✅

nestkernel/nestmodule.cpp
```diff
@@ -474,6 +474,39 @@ NestModule::SimulateFunction::execute( SLIInterpreter* i ) const
...
+{
+  prepare();
+
+  // successful end of simulate
```
apeyser: ✅

nestkernel/nestmodule.cpp
```diff
@@ -474,6 +474,39 @@ NestModule::SimulateFunction::execute( SLIInterpreter* i ) const
...
+  i->EStack.pop();
+}
+
+void
+NestModule::CleanupFunction::execute( SLIInterpreter* i ) const
```
apeyser: ✅

nestkernel/nestmodule.cpp
```diff
@@ -474,6 +474,39 @@ NestModule::SimulateFunction::execute( SLIInterpreter* i ) const
...
+{
+  cleanup();
+
+  // successful end of simulate
```
apeyser: ✅

nestkernel/simulation_manager.cpp
apeyser: ✅

nestkernel/simulation_manager.cpp
```diff
@@ -414,7 +470,7 @@ nest::SimulationManager::simulate( Time const& t )
   to_do_ += t.get_steps();
   to_do_total_ = to_do_;
 
-  const size_t num_active_nodes = prepare_simulation_();
+  prepare_simulation_();
```
apeyser: I just removed it as cruft.

nestkernel/simulation_manager.cpp
apeyser: resume_ -> start_updating, finalize_ removed

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -39,6 +39,49 @@ def Simulate(t):
...
+    t : float
+        Time to simulate in ms
+    """
+
+    sps(float(t))
+    sr('ms Run')
+
+@check_stack
+def Prepare():
+    """Simulate the network for t milliseconds.
```
apeyser: ✅

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -39,6 +39,49 @@ def Simulate(t):
...
+    t : float
+        Time to simulate in ms
+    """
+
+    sr('Prepare')
+
+
+@check_stack
+def Cleanup():
+    """Simulate the network for t milliseconds.
```
apeyser: ✅

pynest/nest/lib/hl_api_simulation.py
apeyser: Renamed RunManager + better doc string.

nestkernel/simulation_manager.cpp
```diff
@@ -410,11 +466,19 @@ nest::SimulationManager::simulate( Time const& t )
...
+{
+  check_run( t );
 
   to_do_ += t.get_steps();
   to_do_total_ = to_do_;
+  assert( to_do_ != 0 );
 
-  const size_t num_active_nodes = prepare_simulation_();
+  // Check whether waveform relaxation is used on any MPI process
+  kernel().node_manager.check_wfr_use();
```
wschenck: Why was check_wfr_use() moved to this point? In the old code it was part of prepare_simulation_().

nestkernel/simulation_manager.cpp
```diff
@@ -443,20 +507,39 @@ nest::SimulationManager::simulate( Time const& t )
...
       "the minimal delay." );
 
-  resume_( num_active_nodes );
+  start_updating_();
+}
 
-  finalize_simulation_();
+void
+nest::SimulationManager::cleanup()
+{
```
wschenck: What do you think about a state flag within SimulationManager, which enables the execution of run()? It could be set to true within prepare() and afterwards set to false within cleanup().

nestkernel/simulation_manager.cpp
```diff
@@ -443,20 +507,39 @@ nest::SimulationManager::simulate( Time const& t )
...
+  if ( not simulated_ )
+    return;
+
+  // Check for synchronicity of global rngs over processes
+  // TODO: This seems double up, there is such a test at end of simulate()
```
wschenck: Remove the comment line with the TODO.

nestkernel/simulation_manager.cpp
```diff
@@ -375,6 +376,61 @@ nest::SimulationManager::simulate( Time const& t )
   t_slice_begin_ = timeval();
   t_slice_end_ = timeval();
```
wschenck: I don't understand the reasoning behind t_slice_begin and t_slice_end. There are both used for print_progress_() and not changed anywhere else. But what does timeval() return? Does this refer to "struct timeval" from the C standard library?

nestkernel/simulation_manager.cpp
```diff
@@ -813,27 +842,6 @@ nest::SimulationManager::update_()
 }
```
wschenck: Additional comment reg. line 791 in the OLD code (in function update_()): I wonder if this part of the code outside of the update loop will work properly if running directly again without a cleanup/prepare cycle. It seems to be related to structural plasticity and updates synaptic elements.
@sdiazpier : May I ask you for help regarding this question?

nestkernel/simulation_manager.cpp
```diff
@@ -375,6 +376,61 @@ nest::SimulationManager::simulate( Time const& t )
   t_slice_begin_ = timeval();
   t_slice_end_ = timeval();
 
+  // Reset profiling timers and counters within event_delivery_manager
+  kernel().event_delivery_manager.reset_timers_counters();
```
wschenck: This line (resetting timers and counters) may be moved into run().

nestkernel/simulation_manager.cpp
apeyser: An earlier version had an enum that made a state machine checking that prepare, run and cleanup were called in the right order. I saw that PR #542 was playing with that, so I removed until merge. Apparently merged, so I should look again at exception throwing in terms of state.

nestkernel/simulation_manager.cpp
apeyser: Seems to. I think someone was trying to set them to {0, 0} (maybe?).

wschenck: I think this functionality is broken. Not by this PR, but already in the main branch. Shall we leave it here and repair it in an additional PR afterwards?

heplesser: `t_slice_end_` is actually updated in `simulation_manager.cpp`, approx line 780 (master), by
```C++
gettimeofday( &t_slice_end_, NULL );
```
It is used to calculate the real-time factor for the progress display.

nestkernel/simulation_manager.cpp
```diff
@@ -375,6 +376,61 @@ nest::SimulationManager::simulate( Time const& t )
...
+  // Check for synchronicity of global rngs over processes.
+  // We need to do this ahead of any simulation in case random numbers
+  // have been consumed on the SLI level.
+  if ( kernel().mpi_manager.get_num_processes() > 1 )
+  {
+    if ( !kernel().mpi_manager.grng_synchrony(
+           kernel().rng_manager.get_grng()->ulrand( 100000 ) ) )
+    {
+      LOG( M_ERROR,
+        "SimulationManager::simulate",
```
apeyser: ✅

nestkernel/simulation_manager.cpp
apeyser: And it looks like value needs to be nulled in order to check whether it was set at all -- the tv_sec is checked to see if it's zero (not precisely correct) to see whether it's been set somewhere else.

This code is messy, probably should be fixed, but outside of the purview of the current PR.

nestkernel/simulation_manager.cpp
apeyser: That's where `prepare_simulation_` was, the only part surviving is the check_wfr_use call. Maybe it could happen only once --- but I'm not sure, not tested.

nestkernel/simulation_manager.cpp
```diff
@@ -443,20 +507,39 @@ nest::SimulationManager::simulate( Time const& t )
       "is called repeatedly with simulation times that are not multiples of "
```
apeyser: ✅

nestkernel/simulation_manager.cpp
apeyser: ✅

nestkernel/simulation_manager.cpp
```diff
@@ -443,20 +507,39 @@ nest::SimulationManager::simulate( Time const& t )
...
+    return;
+
+  // Check for synchronicity of global rngs over processes
+  // TODO: This seems double up, there is such a test at end of simulate()
+  if ( kernel().mpi_manager.get_num_processes() > 1 )
+    if ( !kernel().mpi_manager.grng_synchrony(
+           kernel().rng_manager.get_grng()->ulrand( 100000 ) ) )
+    {
+      throw KernelException(
+        "In SimulationManager::simulate(): "
```
apeyser: ✅

nestkernel/simulation_manager.cpp
apeyser: It's still in the update loop, it's inside of run/start_updating, in the same spot as simulate/resume

nestkernel/simulation_manager.cpp
apeyser: ✅ 
Should be or maybe? Should they be reset every time you call run, or only once for a simulate function? Wouldn't this change the meaning, where before you called simulate (and the counters where reset) and then resume multiple times?

nestkernel/simulation_manager.cpp
wschenck: Sorry, I missed the part where t_slice_begin_ and t_slice_end_ are changed. t_slice_begin_ is also updated in line 593 (approx. in master). The initialization with timeval() is also fine according to C++ standard. Thus no reason to worry.

nestkernel/simulation_manager.cpp
wschenck: I don't think so. It is outside of the do-while-loop within `update_()`. Thus it is called only once at the end of `run()` in the changed code. (In addition, the update of synaptic elements also happens within the do-while-loop, but there is this extra call at the end outside of the loop, and this concerns me slightly - at least it is something to consider.)

nestkernel/simulation_manager.cpp
wschenck: I think it needs to happen only once after `update_delay_extrema_()`, but it should not hurt here either. The gap junction team must know.
@suku248 : ???

nestkernel/simulation_manager.cpp
wschenck: I would move it, because now - if you think of using sionlib as I/O backend - you have to call `run` instead of `simulate` to record all data in one container. And if you want to have separate measurements for each invocation of `run` (what would have been `simulate` before) you need the reset within `run`.

nestkernel/simulation_manager.cpp
apeyser: @wschenck I think I may be confused about the code we're discussing. Could you add the snippet or a link to it? Currently 791 on the master is to:
```
    } while ( to_do_ > 0 and not exit_on_user_signal_
      and not exceptions_raised.at( thrd ) );

>>>>>    // End of the slice, we update the number of synaptic elements
    for ( std::vector< Node* >::const_iterator i =
            kernel().node_manager.get_nodes_on_thread( thrd ).begin();
          i != kernel().node_manager.get_nodes_on_thread( thrd ).end();
          ++i )
    {
      ( *i )->update_synaptic_elements(
        Time( Time::step( clock_.get_steps() + to_step_ ) ).get_ms() );
    }

  } // end of #pragma parallel omp
```
which is in the same place in the  current *monster* update_ () function, right after the update loop. That needs to be refactored! 200 lines seperating the do {} while... and the function still has another 100 lines...

nestkernel/simulation_manager.cpp
apeyser: Is the nestio backend being re-initialized with each `run` in this set up, after we merge? And do we want to have different measurements for each run? It's a non-rhetorical question, what should the semantics here be. We want as little as possible being reinitialized in run, so that you can use run to handle interrupts or do external updates down from the python layer at regular intervals.

nestkernel/simulation_manager.cpp
apeyser: To confirm, do you mean: https://github.com/nest/nest-simulator/pull/650/commits/669dc4960566e289119eae6495cec1d365df6b76?diff=split#diff-06fecbcc67899362e4c211274b21fb71L791 ?
If so, that code is repeated, once after the loop, and once inside the loop:  https://github.com/nest/nest-simulator/pull/650/commits/669dc4960566e289119eae6495cec1d365df6b76?diff=split#diff-06fecbcc67899362e4c211274b21fb71L791

nestkernel/simulation_manager.cpp
apeyser: But I'll comment it, since it's clearly a mess. The code really needs to be refactored into several classes, splitting the current state of the machinery... but I don't want the job.

nestkernel/simulation_manager.cpp
wschenck: Now I am slightly confused about the links in your previous posting because they just open the complete diff.
But you already got the right code snippet in your posting before, concerning the location **outside** of the do-while-loop. And in line 600-607 (in master) the same code exists **within** the do-while-loop.
Concerning the outside position of this code snippet, this would be called in a kind of irregular way in subsequent calls to `run()` (e.g., one run could be for just a fraction of a min-delay interval, and the next for 100 min-delay intervals, and so on...). I just wonder if this may hurt the structural plasticity algorithm. Most likely not, but given the complexity of interacting algorithms in the kernel I am on the paranoid side.

nestkernel/simulation_manager.cpp
wschenck: Not sure what you mean about the nestio backend. In my view it is not re-initialized with each `run`, otherwise we would still have the sionlib problems.

My opinion about the position of `reset_timers_counters` is not very strong. Thus, if you like it as it is, I have no further objections.

nestkernel/simulation_manager.cpp
apeyser: I see now -- but this would also be a problem for multiple simulate() calls. But this may break the ideas that `simulate(n) = prepare(); run(n/2); run(n/2); cleanup()`
I think @sdiazpier is the only person who can answer this, or give a good suggestion on what we should do.

nestkernel/simulation_manager.cpp
apeyser: I was just trying to figure out the semantics -- what gets re-run on each cycle. Since you use the counters, I'll put them where you like them, given that they're light anyhow.

nestkernel/nest.h
apeyser: @heplesser ?

heplesser: See, e.g., `event.h`.

nestkernel/simulation_manager.cpp
apeyser: Done.

nestkernel/simulation_manager.cpp
wschenck: As soon as this is sorted out, I would say :+1: for pulling!

sdiazpier: Dear @apeyser and @wschenck, sorry for the late reply. The update of the synaptic elements can be done in any irregular fraction of simulation time. This is not relevant because the structural changes actually take place on a much larger time scale, so the accumulation of these small changes in the synaptic elements can be achieved at partial steps and the final results should be the same. 
To verify this I ran the simple structural plasticity examples and they provide the same output using the new Prepare() - Run() - Cleanup() functions as with Simulate()

nestkernel/nest.h
```diff
@@ -82,6 +82,47 @@ ArrayDatum get_connections( const DictionaryDatum& dict );
...
+ * @fn run(const double& time)
+ * @brief Run a partial simulation for `time` ms
+ *
+ * Runs a partial simulation for `time` ms
+ * after a call to prepare() and before a cleanup().
+ * Can be called multiple times between a prepare()/cleanup()
+ * pair to divide a simulation into multiple pieces with access
+ * to the API in between.
+ *
+ * Thus, simulate(t) = prepare(); run(n/2); run(n/2); cleanup()
```
Silmathoron: I think this should be ``t/2`` instead of ``n/2``

nestkernel/simulation_manager.cpp
apeyser: ✅

nestkernel/nest.h
apeyser: ✅

apeyser:
@heplesser Any more suggestions before merging?

apeyser:
Issue #659 has been added as a follow up to this PR, to add state checks for the prepare-run-cleanup loop.

heplesser:
@apeyser Thanks for your efforts. I will try to take a look within the next few days.

nestkernel/nestmodule.cpp
```diff
@@ -475,6 +475,69 @@ NestModule::SimulateFunction::execute( SLIInterpreter* i ) const
...
 
 /* BeginDocumentation
+   Name: Run - simulate n milliseconds
+
+   Synopsis:
+   n(int) Run -> -
+
+   Description: Simulate the network for n milliseconds.
+   Call prepare before, and cleanup after.
+   n simulate = prepare n/m run ... m times... cleanup
```
jougs: Make `prepare`, `run` and `simulate` uppercase, wherever they refer to the corrresponding SLI commands. Those are case sensitive. This comment applies throughout this PR.

pynest/nest/tests/test_split.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
```
jougs: Can you please rename this file to something like `test_split_simulation.py`? I think the current name is a bit too general.

nestkernel/nestmodule.cpp
apeyser: ✅

pynest/nest/tests/test_split.py
apeyser: ✅

apeyser:
@jougs I'll rebase on 2.12.0 (hopefully that's happening very soon). Then the whole thing can be merge into nestio.

jougs:
We're going to release current master as 2.12.0 probably already tomorrow.

nestkernel/simulation_manager.cpp
```diff
@@ -435,28 +499,46 @@ nest::SimulationManager::simulate( Time const& t )
...
+        "Global Random Number Generators are not "
+        "in sync at end of simulation." );
+    }
+
+  kernel().node_manager.finalize_nodes();
 }
 
 void
-nest::SimulationManager::resume_( size_t num_active_nodes )
+nest::SimulationManager::start_updating_()
```
heplesser: I am not really happy with the name `start_updating_()` because it makes me look for `stop_updating_()`. Since we already have `update_()`, maybe `update_network_()` or `perform_update_ ()` would be better.

nestkernel/simulation_manager.cpp
```diff
@@ -372,16 +373,68 @@ nest::SimulationManager::simulate( Time const& t )
...
+      throw KernelException();
+    }
+  }
+
+  // if at the beginning of a simulation, set up spike buffers
+  if ( !simulated_ )
+    kernel().event_delivery_manager.configure_spike_buffers();
+
+  kernel().node_manager.ensure_valid_thread_local_ids();
+  num_active_nodes_ = kernel().node_manager.prepare_nodes();
```
heplesser: Does it really make sense to have `num_active_nodes_` as a member variable? It is used only once to provide an informational message. Wouldn't it make more sense to add a `num_active_nodes()` method to `NodeManager`?

nestkernel/simulation_manager.cpp
```diff
@@ -372,16 +373,68 @@ nest::SimulationManager::simulate( Time const& t )
...
+void
+nest::SimulationManager::simulate( Time const& t )
+{
+  prepare();
+  run( t );
+  cleanup();
+}
 
+void
+nest::SimulationManager::check_run( Time const& t )
```
heplesser: `check_run()` does not express what this method does. How about `assert_valid_simtime()`?

nestkernel/nestmodule.cpp
apeyser: ✅

nestkernel/nestmodule.cpp
```diff
@@ -510,7 +515,13 @@ NestModule::RunFunction::execute( SLIInterpreter* i ) const
    Description: sets up network calibration before run is called
    any number of times
 
-   SeeAlso: run, cleanup, simulate
+   Note: Run must only be used after Prepare is called, and
+   before Cleanup to finalize state (close files, etc).
+   Any changes made between Prepare and Cleanup may cause
+   undefined failures (the famed nose-inhabiting monkeys of
+   internet lore).
```
heplesser: @apeyser Could you remove the text in parentheses and add "or yield incorrect results"? Open failures (crashes, stop with error message) are a lesser problem than simulations generating incorrect data.

nestkernel/nestmodule.cpp
```diff
@@ -528,7 +539,13 @@ NestModule::PrepareFunction::execute( SLIInterpreter* i ) const
    Description: tears down a network after run is called
    any number of times
 
-   SeeAlso: run, prepare, simulate
+   Note: Run must only be used after Prepare is called, and
+   before Cleanup to finalize state (close files, etc).
+   Any changes made between Prepare and Cleanup may cause
+   undefined failures (the famed nose-inhabiting monkeys of
+   internet lore).
```
heplesser: See above.

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -41,6 +42,61 @@ def Simulate(t):
...
+def RunManager():
+    """ContextManager for Run.
+
+    Calls Prepare() before a series of Run() calls,
+    and  adds a Cleanup() at end.
+
+    So:
+    with RunManager():
+    	for i in range(10):
+    	    Run()
```
heplesser: Doesn't `Run()` need an argument? I think it would be very useful to explain this context a bit more, so that users (neuroscientist with some Python knowledge) understand how they can use this context manager sensibly. Maybe put a little plot into the loop?

apeyser:
@heplesser @jougs I add some of the suggestion from hep that affected the node-manager, and now with the new vera++ analysis, node_manager fails with 65 errors -- apparently the file was very out of sink with vera++, and as soon as the file is changed at all, all fixes must be made.

That doesn't seem to be in the scope of this pull request, but should be a separate issue to correct this file. What should we do about this?

heplesser:
@apeyser I just created #672 to fix the Vera++ issues, including a problem with our Vera++ rules. Could you also merge from master (should work without a problem) to get the new static code checking in its final version in place here?

nestkernel/nestmodule.cpp
apeyser: That's what undefined means - up to crashes, but generally simply unexpected results (reordering of increments is the classic case).

apeyser:
Added apeyser/nest-simulator#1. I believe that it's only formatting changes -- please correct if I'm wrong @heplesser

heplesser:
@apeyser Almost all changes were formatting (adding curly braces around one-line blocks, `!` -> `not`, line length, spaces), a few tiny spelling corrections and in three places I changed "undefined behavior (the famed nose-inhabiting monkeys of internet lore)" to "undefined behavior and incorrect results" in comments (nestmodule.cpp, l 490, 520, 543).

PR #654, opened 2017-02-06
Add pynest binding for SetMaxBuffered

mhoff:
Function has been available via SLI, but not in pynest.

### DIFF
--- a/pynest/nest/lib/hl_api_parallel_computing.py
+++ b/pynest/nest/lib/hl_api_parallel_computing.py
@@ -73,32 +73,34 @@ def SetNumRecProcesses(nrp):
 
 
 @check_stack
-def SetAcceptableLatency(port, latency):
+def SetAcceptableLatency(port_name, latency):
     """Set the acceptable latency (in ms) for a MUSIC port.
 
     Parameters
     ----------
-    port : object
+    port_name : str
         MUSIC port to set latency for
     latency : float
         Latency in ms
     """
 
+    sps(kernel.SLILiteral(port_name))
     sps(latency)
-    sr("/%s exch SetAcceptableLatency" % port)
+    sr("SetAcceptableLatency")
 
 
 @check_stack
-def SetMaxBuffered(port, size):
+def SetMaxBuffered(port_name, size):
     """Set the maximum buffer size for a MUSIC port.
 
     Parameters
     ----------
-    port : object
+    port_name : str
         MUSIC port to set buffer size for
     size : int
         Buffer size
     """
 
+    sps(kernel.SLILiteral(port_name))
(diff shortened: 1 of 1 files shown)


pynest/nest/lib/hl_api_parallel_computing.py
```diff
@@ -86,3 +86,19 @@ def SetAcceptableLatency(port, latency):
...
     sr("/%s exch SetAcceptableLatency" % port)
+
+
+@check_stack
+def SetMaxBuffered(port, size):
+    """Set the maximum buffer size for a MUSIC port.
+
+    Parameters
+    ----------
+    port : object
```
apeyser: So port is a port name, in SLI /port_name? The comment should probably be more specific than "object"

pynest/nest/lib/hl_api_parallel_computing.py
```diff
@@ -86,3 +86,19 @@ def SetAcceptableLatency(port, latency):
...
+        MUSIC port to set buffer size for
+    size : int
+        Buffer size
+    """
+
+    sps(size)
+    sr("/%s exch SetMaxBuffered" % port)
```
apeyser: 1. Couldn't this be 
```
sps(SLILiteral(port))
sps(size)
sr("SetMaxBuffered")
```
That would be clearer, if that works correctly and isn't too much work
2. Otherwise, I'd use >= python2.7 calls:
```
sr("/{} exch SetMaxBuffered".format(port))
```
or even (to keep it simple)
```
sr("/{} {} SetMaxBuffered".format(port, size)
```
with no `sps(size)`

apeyser:
We probably need better descriptions of how the pynest bindings can be extended. The rest of the file which were the examples followed here could use some cleanup.

pynest/nest/lib/hl_api_parallel_computing.py
mhoff: Yes, I suppose it should be `str`. But as you mentioned beneath, I was sticking to the prior convention.

pynest/nest/lib/hl_api_parallel_computing.py
mhoff:
```python
sps(kernel.SLILiteral(port))
sps(size)
sr("SetMaxBuffered")
```
Does compile and execute just fine. I suppose that does the trick. (Even though I did no in-deep testing whether the call is really conducted).

mhoff:
@apeyser Is 9bdbba2 this what you had in mind?

apeyser:
@mhoff: Precisely. Looks good to me - and necessary, since some of the MUSIC defaults may be surprising to some.

Issue #655, opened 2017-02-08
NEST runs into assertion when gap junctions are used without wfr


janhahne:
NEST runs into an assertion (event.h, line 978) when gap junctions are used without waveform relaxation. The error was introduced with PR #607 - the `GapJunctionEvent` `coeff_length_` parameter is only set when wfr is used, but needs to be set anyway (also without wfr use). The fix is a one liner, i will submit a PR shortly.

heplesser:
Fixed by #656

PR #656, opened 2017-02-08
Fix for usage of gap junctions without wfr (issue #655)

janhahne:
This PR fixes issue #655 by setting the coeff_length parameter of the `GapJunctionEvent` regardless of the use of the waveform relaxation method.


heplesser:
Fixes minor bug.

PR #664, opened 2017-02-27
Added elliptical mask and a function to obtain GIDs inside masked area

stinebuu:
This work is done in preparation for a connectivity app that is being developed.

The PR contains
- Elliptical mask in 2D
- A function to obtain all the GIDs inside a masked area.


topology/pynest/hl_api.py
```diff
@@ -2115,3 +2115,34 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
...
+def SelectNodesByMask(layer, xpos, ypos, mask_obj):
+    """
+    Obtain GIDs inside a specified area.
+
+    Parameters
+    ----------
+    layer : tuple/list of int
+        List containing the single layer to get GIDs from.
+    xpos : double
+        Double value specifying x-coordinate of where we start to search. Can
```
heplesser: It makes more sense to pass this position as a tuple or numpy array; in that way, 2D and 3D are covered by the same code. `anchor` could be a good variable name, as it is the position where the mask is anchored.

topology/pynest/hl_api.py
```diff
@@ -2115,3 +2115,34 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
...
+    ypos : double
+        Double value specifying y-coordinate of where we start to search. Can
+        for instance be center of layer.
+    mask_obj: object
+        Mask object specifying chosen area.
+
+    Returns
+    -------
+    out : list of int(s)
+        GID(s) of neurons inside the specified area.
```
heplesser: "inside the mask"

topology/pynest/hl_api.py
```diff
@@ -2115,3 +2115,34 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
...
+    """
+
+    mask_datum = mask_obj._datum
+
```
heplesser: Raise an exception if `layer` does not have exactly one element.

topology/pynest/hl_api.py
```diff
@@ -2115,3 +2115,34 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
...
+    gid_list = topology_func('SelectNodesByMask', ypos, xpos, mask_datum,
```
heplesser: May require adjustment to changes in SLI interface, see below.

topology/sli/topology-interface.sli
```diff
@@ -158,6 +158,11 @@
     
   /get [/masktype /literaltype] {exch cvdict_M exch get} def
   /get [/masktype /arraytype] {exch cvdict_M exch get} def
+  
+  /SelectNodesByMask [/doubletype /doubletype /masktype /integertype]
+    /SelectNodesByMask_L_M_d_d load
```
heplesser: The anchor position should be supplied as an array or doublevector, so that we can support 2D and 3D with one function.

topology/topologymodule.cpp
```diff
@@ -369,6 +368,9 @@ TopologyModule::init( SLIInterpreter* i )
 
   i->createcommand( "cvdict_M", &cvdict_Mfunction );
 
+  i->createcommand(
+    "SelectNodesByMask_L_M_d_d", &selectnodesbymask_L_M_d_dfunction );
```
heplesser: Change arguments to anchor can be passed as array or vector.

topology/topologymodule.cpp
```diff
@@ -381,6 +383,8 @@ TopologyModule::init( SLIInterpreter* i )
   // Register mask types
   register_mask< BallMask< 2 > >();
   register_mask< BallMask< 3 > >();
+  register_mask< EllipseMask< 2 > >();
+  register_mask< EllipseMask< 3 > >();
```
heplesser: Until `EllipseMask` supports 3D, the 3D variant should not be registered.

topology/topologymodule.cpp
```diff
@@ -1239,6 +1243,39 @@ TopologyModule::Cvdict_MFunction::execute( SLIInterpreter* i ) const
...
+void
+TopologyModule::SelectNodesByMask_L_M_d_dFunction::execute(
+  SLIInterpreter* i ) const
+{
+  i->assert_stack_load( 4 );
+
+  const index& layer_gid = getValue< long >( i->OStack.pick( 0 ) );
+  MaskDatum mask = getValue< MaskDatum >( i->OStack.pick( 1 ) );
+  double xpos = getValue< double >( i->OStack.pick( 2 ) );
+  double ypos = getValue< double >( i->OStack.pick( 3 ) );
```
heplesser: Change to `anchor` as array instead of `xpos` and `ypos`. Also, I think the following order of arguments would be more logical: layer - anchor - mask, reading it as "from layer around anchor select all in mask".

topology/topologymodule.cpp
```diff
@@ -1239,6 +1243,39 @@ TopologyModule::Cvdict_MFunction::execute( SLIInterpreter* i ) const
...
+
+  Layer< 2 >* l =
```
heplesser: Code here must support 2D and 3D layers and masks.

topology/mask.h
```diff
@@ -312,6 +312,70 @@ class BallMask : public Mask< D >
...
+template < int D >
+class EllipseMask : public Mask< D >
+{
+public:
+  /**
+   * @param center Center of ellipse
+   * @param x_side Length between center and x vertex
+   * @param y_side Length between center and y vertex
+   */
+  EllipseMask( Position< D > center, double x_side, double y_side )
```
stinebuu: @heplesser Thanks for your comments! We called the parameter names x_side and y_side because when we check if a point is inside the ellipse (in `EllipseMask< D >::inside`) we use the ellipse equation (x-x_center)^2 / x_side^2 + (y-y_center)^2 / y_side^2 <= 1, and it is important that we use the x-axis radius when dividing with x and same with y. I am unsure how this should be done with major/minor axis, because we then no longer know if the major axis comes from the x-axis or the y-axis. Do you have any suggestions as to how I could deal with this?

topology/mask.h
```diff
@@ -312,43 +312,48 @@ class BallMask : public Mask< D >
...
-    , intermediate_axis_( intermediate_axis )
-    , angle_( angle )
+    , polar_axis_( polar_axis )
+    , azimuth_angle_( azimuth_angle )
+    , polar_angle_( polar_angle )
+    , x_scale_( 4.0 / ( major_axis_ * major_axis_ ) )
+    , y_scale_( 4.0 / ( minor_axis_ * minor_axis_ ) )
+    , z_scale_( 4.0 / ( polar_axis_ * polar_axis_ ) )
+    , azimuth_cos_value_( cos( azimuth_angle_ ) )
+    , azimuth_sin_value_( sin( azimuth_angle_ ) )
```
heplesser: You could drop `value_` here.

heplesser: All calls to mathematical functions should be `std::cos()`, `std::sin()`, ...

topology/mask.h
```diff
@@ -733,42 +745,68 @@ EllipseMask< D >::EllipseMask( const DictionaryDatum& d )
...
   }
   else
   {
-    angle_ = 0.0;
+    polar_angle_ = 0.0;
   }
-  cosine_value_ = cos( angle_ );
-  sine_value_ = sin( angle_ );
+  azimuth_cos_value_ = cos( azimuth_angle_ );
+  azimuth_sin_value_ = sin( azimuth_angle_ );
```
heplesser: `std::`

topology/mask_impl.h
```diff
@@ -395,30 +410,51 @@ EllipseMask< 3 >::inside( const Box< 3 >& b ) const
...
   }
   else
   {
-    // This is a simplification and can be more refined
-    radii.push_back( std::abs( major_axis_ ) );
-    radii.push_back( std::abs( major_axis_ ) );
+    // If the ellipse or ellipsoid is tilted, we make the boundary box
+    // quadratic, with the length of the sides equal to the axis with greatest
+    // length. This could be more refined.
+    double greatest_semi_axis = std::max( major_axis_, polar_axis_ ) / 2.0;
```
heplesser: Can be `const`

topology/mask.h
```diff
@@ -388,15 +393,22 @@ class EllipseMask : public Mask< D >
...
-  double z_dividend_;
+  double azimuth_cos_value_;
+  double azimuth_sin_value_;
 
-  double cosine_value_;
-  double sine_value_;
+  /*
+   * @returns boundary box of ellipse/ellipsoid.
+   */
+  void create_bbox();
```
heplesser: - This does not return the bbox, it sets it; the comment needs to be adapted.
- Since it is a private member function, its name should end with an `_`
- It is confusing to mix methods and data members; it would be better to move this declaration before the data member declarations.

topology/mask.h
```diff
@@ -312,43 +312,48 @@ class BallMask : public Mask< D >
...
+    , azimuth_cos_value_( cos( azimuth_angle_ ) )
+    , azimuth_sin_value_( sin( azimuth_angle_ ) )
   {
-    x_dividend_ = 1.0 / ( major_axis_ * major_axis_ );
-    y_dividend_ = 1.0 / ( minor_axis_ * minor_axis_ );
-    z_dividend_ = 1.0 / ( intermediate_axis_ * intermediate_axis_ );
-
-    cosine_value_ = cos( angle_ );
-    sine_value_ = sin( angle_ );
+    create_bbox();
```
heplesser: Parameter checks are missing here (positive axis, minor <= major).

topology/mask.h
```diff
@@ -733,42 +745,68 @@ EllipseMask< D >::EllipseMask( const DictionaryDatum& d )
...
         "All axis > 0 required." );
     }
 
-    z_dividend_ = 1.0 / ( intermediate_axis_ * intermediate_axis_ );
+    z_scale_ = 4.0 / ( polar_axis_ * polar_axis_ );
   }
   else
   {
-    z_dividend_ = 0.0;
+    z_scale_ = 0.0;
```
heplesser: Should't `polar_axis_` be set to some dummy value, to? I think 0 makes most sense.

topology/mask_impl.h
```diff
@@ -395,30 +410,51 @@ EllipseMask< 3 >::inside( const Box< 3 >& b ) const
...
+}
+
+template < int D >
+bool
+EllipseMask< D >::outside( const Box< D >& b ) const
+{
+  // Currently only checks if the box is outside the bounding box of
+  // the ellipse. This could be made more refined.
+
+  Box< D > bb = bbox_;
```
heplesser: Could be a `const&` to avoid copying.

topology/mask_impl.h
```diff
@@ -430,28 +466,7 @@ template < int D >
 Box< D >
```
heplesser: This should return a `const&` to avoid copying.

topology/pynest/hl_api.py
```diff
@@ -2100,10 +2100,10 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
...
                           zorder=-1000, fc='none', ec=mask_color, lw=3))
     elif 'elliptical' in mask:
-        width = 2*mask['elliptical']['major_axis']
-        height = 2*mask['elliptical']['minor_axis']
-        if 'angle' in mask['elliptical']:
-            angle = mask['elliptical']['angle']*180/pi
+        width = mask['elliptical']['major_axis']
+        height = mask['elliptical']['minor_axis']
+        if 'azimuth_angle' in mask['elliptical']:
+            angle = mask['elliptical']['azimuth_angle']*180/pi
```
heplesser: I just realized that in all stimulus generators with sinusoids, we specify the phase in degrees. So I think we should use angles in degrees as well in masks.

topology/mask.h
```diff
@@ -312,43 +312,48 @@ class BallMask : public Mask< D >
...
-   * @param major_axis Length between center and vertex on major axis
-   * @param minor_axis Length between center and vertex on minor axis
-   * @param intermediate_axis Length between center and vertex on z-axis
-   * @param angle Angle between x-axis and major axis of the ellipse
+   * @param major_axis Length of major axis of ellipse or ellipsoid
+   * @param minor_axis Length of minor axis of ellipse or ellipsoid
+   * @param polar_axis Length of polar axis of ellipsoid
+   * @param azimuth_angle Angle between x-axis and major axis of the ellipse or
+   *        ellipsoid
+   * @param polar_angle Angle between z-axis and polar axis of the ellipsoid
```
heplesser: After looking at sinusoidal stimulus generators, where we specify phase in degrees, I think we should specify angles in degree here as well, and then convert to radians below.

topology/mask.h
```diff
@@ -793,20 +820,21 @@ EllipseMask< D >::EllipseMask( const DictionaryDatum& d )
...
-  azimuth_cos_value_ = cos( azimuth_angle_ );
-  azimuth_sin_value_ = sin( azimuth_angle_ );
 
-  create_bbox();
+  pi_ = 3.141592653589793;
+
+  azimuth_cos_ = std::cos( azimuth_angle_ * pi_ / 180 );
+  azimuth_sin_ = std::sin( azimuth_angle_ * pi_ / 180 );
+  polar_cos_ = std::cos( polar_angle_ * pi_ / 180 );
+  polar_sin_ = std::sin( polar_angle_ * pi_ / 180 );
```
heplesser: I would use `180.` here just to be 100% sure we use floating point division.

topology/mask_impl.h
```diff
@@ -306,13 +306,20 @@ template <>
 bool
 EllipseMask< 3 >::inside( const Position< 3 >& p ) const
 {
-  const double new_x = ( p[ 0 ] - center_[ 0 ] ) * azimuth_cos_value_
-    + ( p[ 1 ] - center_[ 1 ] ) * azimuth_sin_value_;
+  // xx = (x-x_c)*cos(azimuth) + (y-y_c)*sin(azimuth) tilts x-y plane
```
heplesser: Maybe not a bad idea to add a reference?

topology/mask.h
```diff
@@ -339,13 +340,35 @@ class EllipseMask : public Mask< D >
     , polar_axis_( polar_axis )
     , azimuth_angle_( azimuth_angle )
     , polar_angle_( polar_angle )
+    , pi_( 3.141592653589793 )
```
heplesser: It is better to use `numerics::pi` from `numerics.h` for consistency across the code.

topology/pynest/hl_api.py
```diff
@@ -2115,3 +2130,33 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
                              'version of PyTopology')
 
     plt.draw()
+
+
+def SelectNodesByMask(layer, anchor, mask_obj):
+    """
+    Obtain GIDs inside a specified area.
```
jhnnsnk: Please make the description a bit more detailed, i.e., point out that GIDs are obtained inside one topology layer only and that the area is specified by a mask.

topology/pynest/hl_api.py
```diff
@@ -2115,3 +2130,33 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
...
+    anchor : tuple/list of double
+        List containing center position of the layer. This is the point from
+        where we start to search.
+    mask_obj: object
+        Mask object specifying chosen area.
+
+    Returns
+    -------
+    out : list of int(s)
+        GID(s) of neurons inside the mask.
```
jhnnsnk: Does it work only for neurons or also for other nodes/elements inside the mask?

topology/pynest/tests/test_selection_function_and_elliptical_mask.py
```diff
@@ -0,0 +1,363 @@
...
+        self.assertEqual(gid_list, (9, 14, 19,))
+
+        maskdict = {'major_axis': 3.0, 'minor_axis': 1.0, 'anchor': [1., 1.]}
+        mask = topo.CreateMask('elliptical', maskdict)
+
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+
+        self.assertEqual(gid_list, (13, 18, 23,))
+
+    def test_EllpsoidalMask3D(self):
```
jhnnsnk: Ellpsoidal -> Ellipsoidal

topology/pynest/tests/test_selection_function_and_elliptical_mask.py
```diff
@@ -0,0 +1,363 @@
...
+        self.assertEqual(gid_list, (2, 3, 13, 14, 123, 124, 134, 135,))
+
+    def test_CreateEllipticalMask2D(self):
+        """Creates simple elliptical mask"""
+        mask_dict = {'major_axis': 6.0, 'minor_axis': 3.0}
+        mask = topo.CreateMask('elliptical', mask_dict)
+
+        self.assertTrue(mask.Inside([0.0, 0.0]))
+
+    def test_EllpticalMask2D(self):
```
jhnnsnk: Ellptical -> Elliptical

topology/pynest/tests/test_selection_function_and_elliptical_mask.py
```diff
@@ -0,0 +1,363 @@
...
+        cntr = [0., 0., 0.]
+
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+
+        self.assertEqual(gid_list, (2, 3, 13, 14, 123, 124, 134, 135,))
+
+    def test_CreateEllipticalMask2D(self):
+        """Creates simple elliptical mask"""
+        mask_dict = {'major_axis': 6.0, 'minor_axis': 3.0}
+        mask = topo.CreateMask('elliptical', mask_dict)
```
stinebuu: Done! :)

Issue #665, opened 2017-02-27
ConnectLayers does not connect to thread siblings of devices


heplesser:
In multithreaded simulations, `ConnectLayers` does not connect to thread siblings of devices. The following code:
```
0 << /local_num_threads 4 >> SetStatus

/nrns << /elements /iaf_psc_alpha /rows 2 /columns 2 >> CreateLayer def
/det << /elements /spike_detector /rows 1 /columns 1 >> CreateLayer def
nrns det  << /connection_type /convergent >> ConnectLayers

<< >> GetConnections ==
```
will connect only the neuron managed by thread 0 (neuron with GID 4) to the spike detector (instance on thread 0):
```
[<4,7,0,0,0>]
```
The other neurons should be connected to the thread-siblings of the spike detector on their respective threads. This is a regression from NEST 2.10, which created all connections:
```
[<5,7,1,0,0> <3,7,3,0,0> <2,7,2,0,0> <4,7,0,0,0>]
```

Originally reported by Elena Pastorelli on NEST User 20 Feb 2017.

terhorstd:
closed by merge of #666

PR #666, opened 2017-02-28
Fix 665: Connecting to layers of recording devices

heplesser:
This PR addresses #665. When connecting to a layer of recording devices without proxies, i.e., with thread siblings, `ConnectLayers` only connected on layer 0. This is now fixed for convergent connections without prescribed number of connections, which was the approach recommended in the manual, Sec 3.10, previously. The fix uses the same technique as in `ConnBuilder`. For all other topology connection types, connecting to layers containing devices without proxies is now explicitly prohibited.


topology/connection_creator_impl.h
```diff
@@ -218,36 +218,39 @@ ConnectionCreator::target_driven_connect_( Layer< D >& source,
...
+      // check whether the target is on our thread
+      if ( thread_id != target_thread )
+      {
         continue;
+      }
 
       if ( target_filter_.select_model()
-        && ( ( *tgt_it )->get_model_id() != target_filter_.model ) )
+        && ( tgt->get_model_id() != target_filter_.model ) )
         continue;
```
terhorstd: this should also be in { }, as the one above

topology/connection_creator_impl.h
```diff
@@ -218,36 +218,39 @@ ConnectionCreator::target_driven_connect_( Layer< D >& source,
...
       if ( mask_.valid() )
         connect_to_target_( pool.masked_begin( target_pos ),
           pool.masked_end(),
-          *tgt_it,
+          tgt,
           target_pos,
-          target_thread,
+          thread_id,
           source );
       else
```
terhorstd: since the connect_to_target call is getting kind of multi-line, adding { } to the 'if' would ease the reading a lot

heplesser:
I will fix this later tonight. I think the figures changed because I installed Matplotlib 2.0 recently. I will re-run with 1.5.


On 28 Feb 2017, at 17:19, Jochen Martin Eppler <notifications@github.com<mailto:notifications@github.com>> wrote:


@jougs requested changes on this pull request.

Thanks! I have just a very minor comment on the code.

However, all the figures are changed and I think the new versions are problematic because of two reasons:

  *   the labels are harder to read because of the decreased size
  *   the contrast of gray boxes is gone in figures where the grid lines were changed from dotted gray to solid gray.

I could not see changes in the code that lead to theese changes. Did you change your matplotlib stylesheet?

________________________________

In testsuite/regressiontests/issue-665.sli<https://github.com/nest/nest-simulator/pull/666#discussion_r103484073>:

> +  << /connection_type /convergent
+     /number_of_connections 4
+     /allow_multapses false >> layer_to_det_test
+}
+fail_or_die
+clear
+
+{
+  % one outgoing connection from each neuron in layer
+  << /connection_type /divergent
+     /number_of_connections 1 >> layer_to_det_test
+}
+fail_or_die
+clear
+
+endusing


Please add a newline at the end of the file.

-

topology/connection_creator_impl.h
heplesser: Done.

topology/connection_creator_impl.h
heplesser: Done.

heplesser:
@jougs @terhorstd The points you raised should be fixed now.

jougs:
@heplesser: I'm afraid that the merge of #546 (which you approved) activated full VERA++ checks and now this pull request requires the addition of more curly braces in conditionals. Bad luck ;-)

heplesser:
@jougs I have hopefully fixed everything now after merging the new style checker into this PR. I even found and fixed a tiny, topology-related bug in the style checker. Travis is not running properly right now due to S3 problems, so we may need to wait a little to be really sure.

jougs:
Thanks for the fast reaction. Do you have an idea why the Travis display disappeared from here?

jougs:
Ah, OK. I got it. Travis is not running *at all* due to the problems. I assume we have to wait then. Nevermind.

heplesser:
@jougs @terhorstd The tests finally passed after some more fixes to the Travis checks. Sorry for not testing #546 on Travis thoroughly enough.

Issue #505, opened 2016-10-05
Improved windowing for devices


heplesser:
This is a follow-up to discussions in #497, which raised the issue of windowing recordings.

We should think through approaches, syntax and semantics before committing to a solution. Windowing will probably be useful for many types of recordings, so we should aim for a more general solution. It also needs to be compatible (reasonably) with existing solutions to ensure a consistent user experience across recorders. 

We currently have the possibility to define "one-time" windows using `start` and `stop`, which can then be shifted during simulation breaks by changing `origin`. I have no data on how much this is used, but I think this approach may be a bit more dated.

Then we have, for `multimeter`, the sampling `interval`: data is recorded only at time steps that are a multiple of `interval`.  For continuous signals that can be sampled at any time, this is a sensible approach. If combined with a `start`-`stop` window, only multiples of `interval` inside the window are recorded.

I would suggest to extend this approach in two directions. For all devices (include stimulation devices) we should support flexible, repeated windowing (extending what @weidel-p suggested in #497). Ideally, we should provide both regular windowing with a fixed window width repeated at fixed intervals, and user-defined windowing based on an explicit list of window start times and durations. This could be implemented in the `Device`/`RecordingDevice`/`StimulatingDevice` classes. I would suggest `window_width` and `window_repeat` (or `window_interval`) as names, and `windows` for the user-defined case. In that case, the argument should be a nested list or 2D array with `start time`-`width` pairs.

SepehrMN:
@heplesser @weidel-p @abigailm I have implemented a similar mechanism in my new device #436. There are multiple start and stop times, and one interval can be specified. Those should not ideally be in the device,  but in the recording device.

SepehrMN:
Hi,
Is there a timetable for merging the new windowing system?

heplesser:
@SepehrMN At present, the new windowing system is a proposal waiting for a volunteer to implement it. Many core developers are currently busy with major changes behind the scenes (subnet removal, improved output), so it will probably be a while until we get around to this one.

SepehrMN:
Many thanks. I think what I wrote can be very easily be turned into how your `windows` suggestion would work. Furthermore, it would work with a `window_stop` as well.

sepehrmn:
I personally do not need this feature anymore. I have an old pull request that has an implementation inside the class. I hope to reopen it for merging soon. 

This issue can be closed.

heplesser:
[...]

PR #669, opened 2017-03-03
Fix for multimeter origin bug. 

zbarni:
Modified multimeter to actually consider the origin variable when sampling. Modification of the DataLoggingRequest class was also required to keep track of the origin. Main issue was that the initial sampling timepoint (relative to which the next intervals are calculated) was computed without regard to the origin.


heplesser:
@zbarni As I explained in #670, `multimeter` works as intended, although some documentation is misleading. The main problem with your proposal here is that the purpose of `origin` is to allow the user to move the `start`-`stop` window along the time axis by changing `origin` during stimulation breaks. Your intended use of `origin` would conflict with this. Also, a logger will only take into account the value of `origin` at the first `Simulate` call after the multimeter is first connected. Any later changes would be difficult, because it would require resizing of logger buffers and thus risk losing or invalidating data that has been recorded already in the logger inside the neuron recorded from, but that has not yet been fetched by the multimeter.

I therefore think that you approach is not feasible.

zbarni:
I think the discussion has moved to #670 .

heplesser:
I think we should aim for a more general solution along the lines of what I sketched in #505. Therefore, I will close this PR for now.

heplesser:
@zbarni After thinking it through a bit further, here a suggestion for how to handle subsampling with arbitrary offset:

1. Sampling is parameterized in terms of `offset` and `interval` such that data is recorded at all times `t == n h == offset + k interval`, where `h` is the simulation resolution and `n` and `k` are non-negative integers (n>=1, k>=0).
1. Just as `interval`, `offset` can only be set before the multimeter is connected to any neurons; we cannot change it later because that would require modifying the recording buffers, risking loss of data. Since `interval` already has this restriction, this should not be problematic.
1. Windowing (anything that has to do with `start`, `stop`, `origin`) is orthogonal to this and handled entirely by base classes.

Could you try to implement this? It would also be excellent if you could add a test for the new `offset` feature, including tests that modifying `offset` after connecting is prohibited.

Sorry for the back-and-forth!

zbarni:
@heplesser no problem, probably the role of offset is clearer this way. Finished the changes you suggested (offset as a new variable instead of overlapping with origin), and added a simple unit test for correctness and changing the offset after connection. Let me know if anything else should be changed or extended.

models/multimeter.cpp
```diff
@@ -91,27 +98,55 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
...
-        "the simulation resolution" );
+        "The sampling interval must be a multiple of the simulation "
+        "resolution" );
+    }
+  }
+
+  if ( updateValue< double >( d, names::offset, v ) )
+  {
+    // only throw error if given offset is > 0., otherwise it's not an error
+    if ( Time( Time::ms( v ) ) < Time::get_resolution() && v > 0. )
```
heplesser: I think this code would be easier to read with the tests the other way around; also, we now prefer boolean operators written out, i.e.,
```c++
if ( v > 0 and Time( ... ) < ... )
```
It is also not necessary to write `0.` here, `0` will do.

models/multimeter.cpp
```diff
@@ -176,8 +215,8 @@ Multimeter::update( Time const& origin, const long from, const long )
   // following Reply data is then added.
   //
   // Note that not all nodes receiving the request will necessarily answer.
-  V_.new_request_ =
-    B_.has_targets_ && !P_.record_from_.empty(); // no targets, no request
+  V_.new_request_ = B_.has_targets_
+    && P_.record_from_.empty() == false; // no targets, no request
```
heplesser: One should not compare with `true` and `false`. The following would be better (comment moved for better line structure):
```c++
// no targets or no recordables means no request
V_.new_request_ = B_.has_targets_ and not P_.record_from_.empty();
```

models/multimeter.cpp
```diff
@@ -200,9 +241,11 @@ Multimeter::handle( DataLoggingReply& reply )
   for ( size_t j = 0; j < info.size(); ++j )
   {
     if ( not info[ j ].timestamp.is_finite() )
+    {
       break;
+    }
 
-    if ( !is_active( info[ j ].timestamp ) )
+    if ( is_active( info[ j ].timestamp ) == false )
```
heplesser:
```c++
if ( not is_active( ... ) )
```

models/multimeter.cpp
```diff
@@ -213,22 +256,28 @@ Multimeter::handle( DataLoggingReply& reply )
 
     // record sender and time information; in accumulator mode only for first
     // Reply in slice
-    if ( !device_.to_accumulator() || V_.new_request_ )
+    if ( device_.to_accumulator() == false || V_.new_request_ )
```
heplesser:
```c++
if ( not device_.to_accumulator() or V_.new_request_ )
```

models/multimeter.cpp
```diff
@@ -213,22 +256,28 @@ Multimeter::handle( DataLoggingReply& reply )
...
+    {
       device_.record_event( reply, false ); // false: more data to come
+    }
 
-    if ( !device_.to_accumulator() )
+    if ( device_.to_accumulator() == false )
```
heplesser:
```c++
if ( not device_.to_accumulator() )
```

models/multimeter.h
```diff
@@ -138,9 +139,9 @@ namespace nest
  * with a given time stamp.
  *
  * Start and stop are handled as follows: the first recorded
- * data is with time stamp origin+start+1, the last recorded one
- * that with time stamp origin+stop. Only such times are recorded
- * for which (T-(origin+start)) mod interval is zero.
+ * data is with time stamp offset+start+1, the last recorded one
+ * that with time stamp offset+stop. Only such times are recorded
+ * for which (T-(offset+start)) mod interval is zero.
```
heplesser: start, stop and origin are independent of offset and interval. Isn't the correct description now "Data is recorded at time steps T for which start < T <= stop and ( T - offset ) mod interval == 0."?

models/multimeter.h
```diff
@@ -237,7 +238,8 @@ class Multimeter : public Node
 
   struct Parameters_
   {
-    Time interval_;                   //!< recording interval, in ms
+    Time interval_; //!< recording interval, in ms
+    Time offset_;   //!< offset relative to which interval is calculated
```
heplesser: add "in ms"

nestkernel/event.h
```diff
@@ -582,18 +587,22 @@ class DataLoggingRequest : public Event
 inline DataLoggingRequest::DataLoggingRequest()
   : Event()
   , recording_interval_( Time::neg_inf() )
+  , recording_offset_( Time::ms( 0. ) )
```
heplesser: `0` would suffice

testsuite/unittests/test_multimeter_offset.sli
```diff
@@ -0,0 +1,115 @@
...
+      /result mm /events get def
+      result /times get cva 
+
+      % compare to reference data
+      refdata eq
+    }   
+    { ; true }
+    ifelse
+  }
+  Map
```
heplesser: It isn't really necessary to run this test for all neuron models and all recordables for each model. Running it for a single model with two recordables would suffice and make the test run faster.

It would be useful to add another test that uses start/stop and origin (changing origin during a simulation break) to see that offset still works correctly then.

nestkernel/event.h
```diff
@@ -555,7 +555,7 @@ class DataLoggingRequest : public Event
   DataLoggingRequest();
 
   /** Create event for given time stamp and vector of recordables. */
```
heplesser: Could you extend this comment to explain the different parameters? That becomes more important since we have two `Time` arguments now.

nestkernel/event.h
zbarni: This doesn't work.. initializing recording_offset_ with ( 0 ) leads to a compile error.

heplesser:
@zbarni Thanks for the fixes, nice work!

zbarni:
Glad I could contribute!

models/multimeter.cpp
```diff
@@ -91,27 +98,55 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
...
+      throw BadProperty(
+        "The sampling interval must be a multiple of the simulation "
+        "resolution" );
+    }
+  }
+
+  if ( updateValue< double >( d, names::offset, v ) )
+  {
+    // only throw error if given offset is > 0., otherwise it's not an error
+    if ( v > 0 && Time( Time::ms( v ) ) < Time::get_resolution() )
```
apeyser: `0 < v < t_resolution` is an error? But not `v > t_resolution`?
Please comment.

Also, v should just be converted to time once (say `v_time`), rather than assuming that the compiler can optimize it out -- it's also easier to read/

nestkernel/event.h
apeyser: @zbarni But Time::ms(0) rather than (0.) should be enough.

zbarni:
@apeyser as @rcfduarte put it in #670: 

> ...we're running very long simulations where we need to sample the values of population state variables at regular intervals. For example, we would like to sample the population activity at 100.1, 200.1, 300.1, etc (where the extra 0.1 ms is added to account for a delay of 0.1 ms from an input generator to the network). As @zbarni mentioned, as it is currently implemented, the multimeter doesn't allow this (unless we set interval=0.1, which is not a very efficient solution for long simulations)...

Let me know if this answers your question.

zbarni:
@apeyser regarding

> But Time::ms(0) rather than (0.) should be enough.

Doesn't compile, yielding the error
```
.....event.h:591:36: error: call of overloaded ‘ms(int)’ is ambiguous
   , recording_offset_( Time::ms( 0 ) )
```
Same case for ` recording_offset_( 0 )`

models/multimeter.cpp
zbarni: `v > 0` is wrong, should have been `v < 0`, but I removed it entirely since it's contained in the second condition. Thanks for pointing it out.

models/multimeter.cpp
```diff
@@ -91,27 +98,55 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
...
     if ( Time( Time::ms( v ) ) < Time::get_resolution() )
+    {
       throw BadProperty(
-        "The sampling interval must be at least as long "
-        "as the simulation resolution." );
+        "The sampling interval must be at least as long as the simulation "
+        "resolution." );
+    }
 
     // see if we can represent interval as multiple of step
```
zbarni: This is part of the original code, I just copied the whole if block.

models/multimeter.cpp
zbarni: `0 < v < t_resolution` is an error, because if `v` is other than the default value (0), it has to be multiple of the resolution (this means positive and at least as large as the resolution itself.)

`v > 0` was wrong, should have been `v != 0`. Not sure about introducing another variable `v_time `, or creating a new object by converting to Time(), as at that moment `v` is still double, and we're not comparing it against another object, just 0.

models/multimeter.cpp
apeyser: Understood -- but if possible, add some clarification to the block you copied, because otherwise the lack of commentary will just propagate through the code, leading to continuing confusion (for me).

nestkernel/event.h
apeyser: Good to know -- I should open an issue on ambiguities in time.

models/multimeter.cpp
zbarni: Can you point out which lines are you referring to exactly? 121-129? It seems to me that it's just one way to get around the numerical issues, but the main idea is present in the comment

> // see if we can represent interval as multiple of step

Even if the interval is a multiple of step, `std::abs( 1 - interval_.get_ms() / v )` might be very close but still != 0, hence the comparison with epsilon. Should I expand on this?

nestkernel/event.h
apeyser: There's an `explicit ms(long t)`. I wonder why there's an ambiguity on the constructor. The long/double constructors should probably anyway be reduced to just the double constructor.

mschmidt87:
Hi, I just stumbled upon the same issue in my simulations. I have read this PR and the related issue #670 . If I understood it correctly, this PR now aims to implement the suggestion by @heplesser , namely that the recording device should sample at times 

> `t == n h == offset + k interval`, where h is the simulation resolution and n and k are non-negative integers (n>=1, k>=0).

However, it seems to me that with the code implements `k>0`. Executing the following code:
```
import nest
nest.ResetKernel()
n = nest.Create('iaf_psc_exp', 1)
vm = nest.Create('voltmeter', 1)
nest.SetStatus(vm, {'interval': 5.,
                    'offset': 30.1})
nest.Connect(vm, n)
nest.Simulate(100.)
data = nest.GetStatus(vm, 'events')[0]['times']
```
gives recorded times `[ 35.1  40.1  45.1]`, whereas I would have expected `[30.1 35.1  40.1  45.1]`, which would be equivalent to setting `k=[0,1,2,3]` in the equation suggested by @heplesser (with `offset=30.1`, `interval=5.`).

Did I misunderstand something here?

models/multimeter.cpp
```diff
@@ -91,27 +98,56 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
...
+    && ( d->known( names::interval ) || d->known( names::offset )
+         || d->known( names::record_from ) ) )
+  {
     throw BadProperty(
-      "The recording interval and the list of properties to record "
-      "cannot be changed after the multimeter has been connected to "
-      "nodes." );
+      "The recording interval, the interval offset and the list of properties "
+      "to record "
+      "cannot be changed after the multimeter has been connected to nodes." );
```
heplesser: Could you fix the line breaks to that the last line is shortest?

zbarni: sure

heplesser:
@zbarni The conflict reported by Github seem to be just line break/empty line/whitespace changes and should be easy to fix. The easiest is probably to pull the current master into your branch.

models/multimeter.cpp
```diff
@@ -91,27 +98,56 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
...
+      throw BadProperty(
+        "The offset for the sampling interval must be at least as long as the "
+        "simulation resolution." );
+    }
+
+    // see if we can represent offset as multiple of step
+    offset_ = Time::step( Time( Time::ms( v ) ).get_steps() );
+    if ( std::abs( 1 - offset_.get_ms() / v ) > 10
+        * std::numeric_limits< double >::epsilon() )
+    {
```
zbarni: applied both changes along with the fix for @mschmidt87 's example and edited the unit test to account for this. Let me know if anything else needs to be done.

models/multimeter.cpp
```diff
@@ -103,8 +103,8 @@ nest::Multimeter::Parameters_::set( const DictionaryDatum& d,
   {
     throw BadProperty(
       "The recording interval, the interval offset and the list of properties "
-      "to record "
-      "cannot be changed after the multimeter has been connected to nodes." );
+      "to record cannot be changed after the multimeter has been connected "
```
heplesser: This needs a space at the end of the string, since the message output mechanism may change line breaks when the message is printed.

nestkernel/universal_data_logger_impl.h
```diff
@@ -116,10 +116,18 @@ nest::UniversalDataLogger< HostNode >::DataLogger_::init()
...
+    next_rec_step_ = kernel().simulation_manager.get_time().get_steps()
+      + recording_offset_.get_steps() - 1;
+  }
+  else
+  {
+    next_rec_step_ =
+      ( kernel().simulation_manager.get_time().get_steps() / rec_int_steps_
+        + 1 ) * rec_int_steps_
+      - 1;
+  }
```
heplesser: Could you add a comment on why you need to add 1 and move the `-1` up one line?

zbarni:
@heplesser I'm actually unsure whether the offset is supposed to be relative to 0. or relative to the elapsed simulation time since 0.? For instance,

```python
nest.Simulate(10.)
nest.Create("multimeter", 1, {"interval": 3., "offset": 5.})
(...)
nest.Simulate(10.)
```
should give `[15.  18.]` (relative to elapsed time, 10. ms) or `[11. 14. 17. 20.]` (relative to 0.)? Currently, the intervals are calculated relative to 0. when a new multimeter is created and not the elapsed time..

heplesser:
@zbarni We should stick to the current approach (relative to 0.), so that the recording times are given by `t_k = offset + k * interval`. This makes `offset`/`interval` semantics consistent with the `origin`/`start`/`stop` semantics which also use 0 as a reference point. Furthermore, it ensures that recording times remain fixed even if you change simulation intervals or move code lines around.

I just suggested to add the test to ensure that recording times are correctly based on 0 even if a multimeter is created after the simulation has run for some time.

nestkernel/universal_data_logger_impl.h
zbarni: clang requires the -1 to be on a separate line... 

```

MSGBLD0175: [DIFF] <     ( kernel().simulation_manager.get_time().get_steps() / rec_int_steps_
MSGBLD0175: [DIFF] <       + 1 ) * rec_int_steps_ - 1;
MSGBLD0175: [DIFF] ---
MSGBLD0175: [DIFF] >     ( kernel().simulation_manager.get_time().get_steps() / rec_int_steps_ + 1 )
MSGBLD0175: [DIFF] >       * rec_int_steps_
MSGBLD0175: [DIFF] >     - 1;
```

heplesser:
Add possiblity to set offset for multimeter recording. Thanks, @zbarni!

PR #675, opened 2017-03-09
Remove trailing tab in lines of output files of the weight detector

### LINKED ISSUE #674: Trailing tab in output files of weight_recorder
The `weight_recorder` introduced in #497 stores the recorded events in a tab-separated file. It introduces a trailing tab in each line. This is certainly a tiny detail but it causes some trouble when loading the data.
For instance, when I'm trying to load the data using the pandas library, e.g. using
    
    pd.read_csv('weights.csv', sep='\t')

this leads to errors in the create data object because the routine misinterprets the trailing tab to precede a data entry. To my understanding, trailing separators also violate the definition of the csv/tsv format. 
I therefore suggest to get rid of the trailing tab, which can be easily done by changing line 732 in recording_device.cpp . I will create a pull request for this.

mschmidt87:
This pull request fixes #674 by simply removing the `'\t'` from the output of the weight value which is the last value in each line of the output files of the `weight_recorder` .

Being such a simple pull request, this PR might need only one reviewer, which I suggest to be @weidel-p

### DIFF
--- a/nestkernel/recording_device.cpp
+++ b/nestkernel/recording_device.cpp
@@ -729,7 +729,7 @@ void
 nest::RecordingDevice::print_weight_( std::ostream& os, double weight )
 {
   if ( P_.withweight_ )
-    os << weight << '\t';
+    os << weight;
 }
 
 void
--- a/nestkernel/recording_device.cpp
+++ b/nestkernel/recording_device.cpp
@@ -56,43 +56,43 @@ nest::RecordingDevice::Parameters_::Parameters_( const std::string& file_ext,
   bool withtargetgid,
   bool withport,
   bool withrport )
-  : to_file_( false )
-  , to_screen_( false )
-  , to_memory_( true )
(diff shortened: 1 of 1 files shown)


nestkernel/recording_device.cpp
```diff
@@ -688,69 +764,91 @@ nest::RecordingDevice::record_event( const Event& event, bool endrecord )
...
+  }
 }
 
 void
 nest::RecordingDevice::print_weight_( std::ostream& os, double weight )
 {
   if ( P_.withweight_ )
-    os << weight << '\t';
+  {
+    os << weight;
```
heplesser: @mschmidt87 As you already remarked in a comment in #674, the tab is appended to all output so that arbitrary output can be added later. We need to find a better way of handling this to avoid trailing tabs. I suggest to have that discussion in #674.

heplesser:
Closing, see discussion in #674.

Issue #680, opened 2017-03-15
PyNEST testsuite: test_quantal_stdp_synapse fails with Numpy 1.12.0 due to float index


mschmidt87:
I recently installed the latest release of numpy, which is 1.12.0. Note that the release contained in the Ubuntu repositories is 1:1.11.1~rc1-1ubuntu1 for Ubuntu 16.10. To my understanding, numpy 1.12.0 will be contained in the repositories of Ubuntu 17.04 (see [http://packages.ubuntu.com/search?keywords=python-numpy](http://packages.ubuntu.com/search?keywords=python-numpy)).

In this version of numpy, using `float` as an index for numpy arrays throws an error (instead of a warning as in the previous releases). This causes the `quantal_stdp_synapse` test to file with the error:

`TypeError: 'float' object cannot be interpreted as an index`

To solve this, I suggest to replace the float variables used as indices by `int()` when they are used as indices to numpy arrays. Alternatively, one could also define them as `int` in the first place, but the first option is already being used in the test script, so I will implement this and create a pull request.

If I understand the configuration file for travis correctly, we are using 14.04 (Trusty) for Travis. is this intentional? Don't we want to use the newest version of Ubuntu? Or at least the newest LTS version?

To use the newest version of numpy, I suggest to add numpy to the packages installed by pip rather than apt-get and specify 1.12.0 directly, just like it is done for cython.

heplesser:
@mschmidt87 Thanks for reporting and fixing this!

Concerning the change of tools/versions, this is a more complicated issue and should be handled in a separate issue here. I much agree that we need to test with up-to-date tools, but we also need to test against old versions of tools. Otherwise, we risk that people running, e.g., LTS versions of Linux distributions or work on supercomputers (which tend to be rather late wrt versions), may no longer be able to run NEST because we added features that are not supported in older versions. We need to agree on a range of versions that we support and have a test for the oldest as well as cutting-edge. Could you open a separate issue for this?

mschmidt87:
Okay, sure, I certainly agree to this. I will open an issue with regards to this and take the travis modification out of #681 .

mschmidt87:
Okay, sure, I certainly agree. I will open an issue with regards to this and take the travis modification out of #681 .

heplesser:
Fixed by merge of #681.

PR #681, opened 2017-03-15
Fix quantal stdp test for numpy

mschmidt87:
This pull request addresses #680  and contains the following changes:

- Convert the float variables in the test to integers when they are used as indices to numpy arrays. I decided to do this rather then defining them as integers in the first place because a) defining times as floats seems plausible to me and b) the conversion of `t_tot` had been implemented in this test at a different location earlier.
- Remove two parameter dictionaries that were not used in the script.
- Make travis download the latest release of numpy via pip.

### DIFF
--- a/pynest/nest/tests/test_quantal_stp_synapse.py
+++ b/pynest/nest/tests/test_quantal_stp_synapse.py
@@ -101,15 +101,16 @@ class QuantalSTPSynapseTestCase(unittest.TestCase):
         vm_reference = numpy.array(nest.GetStatus(
             [voltmeter[0]], 'events')[0]['V_m'])
 
-        vm.shape = (n_trials, t_tot)
-        vm_reference.shape = (n_trials, t_tot)
+        assert(len(vm) % n_trials == 0)
+        n_steps = int(len(vm) / n_trials)
+        vm.shape = (n_trials, n_steps)
+        vm_reference.shape = (n_trials, n_steps)
 
-        vm_mean = numpy.array([numpy.mean(vm[:, i])
-                               for i in range(int(t_tot))])
-        vm_ref_mean = numpy.array(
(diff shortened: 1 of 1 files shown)


pynest/nest/tests/test_quantal_stp_synapse.py
```diff
@@ -101,15 +97,16 @@ def test_QuantalSTPSynapse(self):
         vm_reference = numpy.array(nest.GetStatus(
             [voltmeter[0]], 'events')[0]['V_m'])
 
-        vm.shape = (n_trials, t_tot)
-        vm_reference.shape = (n_trials, t_tot)
+        vm.shape = (n_trials, int(t_tot))
+        vm_reference.shape = (n_trials, int(t_tot))
```
heplesser: Using `int(t_tot)` here to dimension the result array is not a good idea, since you implicitly require that the voltmeter records at 1 ms intervals. A cleaner implementation would be (not 100% sure of spelling of assertions)
```python
self.assert_equal(len(vm) % n_trials, 0)
n_steps = int(len(vm) / n_trials)    # int to be absolutely Py3-safe
vm.shape = (n_trials, n_steps)
...
```

pynest/nest/tests/test_quantal_stp_synapse.py
```diff
@@ -101,15 +97,16 @@ def test_QuantalSTPSynapse(self):
...
 
         vm_mean = numpy.array([numpy.mean(vm[:, i])
                                for i in range(int(t_tot))])
```
heplesser: I think you can avoid the list comprehension here, NumPy supports row- and column-wise means.

mschmidt87:
Thank you for your feedback. I removed the commit touching .travis.yml from the PR and addressed your comments. I furthermore removed `t_plot` from the script because it was only used for computing the error and I didn't see a reason to not take the entire dataset into account here.

PR #685, opened 2017-03-20
Avoid divisions/calulcation of steps

jakobj:
This PR implements the changes suggested by Harald Servat to reduce run time by avoiding divisions and adding additional member in Event and Time objects that store frequently used values instead of calculating them from the other members on every invocation. Preliminary benchmarks show an increase of ~15% on JUQUEEN. Plots will follow.
I suggest @heplesser @apeyser @terhorstd as reviewers. If Harald Servat could also have a look that would be great, unfortunately I don't know his GitHub handle.


apeyser:
Please see my poster from two years ago with @wschenck  on why you shouldn't cache values like this.

On the other hand, inverting constants or statics and using them as multiplicative values is good -- as long as you aren't pushing the envelope of the floating point ranges and losing precision.

jakobj:
Thanks for the input @apeyser. If I understand you correctly, I should remove the steps member in the Time class, the rest can stay?

heplesser:
@apeyser Is you poster available online somewhere? It would be very useful for reference.

apeyser:
@heplesser https://juser.fz-juelich.de/record/256297/files/poster.pdf , just search alexander peyser sfn 2015 on google scholar

@jakobj Yes -- I think storing the inversions is a very good idea -- someone should check the boundary conditions though, whether you lose precision for extreme values (probably not, but needs checked). For the statics, particularly good --- I didn't do it originally only because they changes where already so big, I didn't want to confuse matters (if you look, some of it is that way already).

jakobj:
Here are some benchmarks of the proposed changes. Sorry for the labels: div3 = all changes as propsed, div4 = all changes as proposed without the extra member in the time objects. These seem to differ from your data @apeyser, as the additional member in Time objects leads to a decrease in runtime, not an increase. something we should discuss in today's meeting.
[JUQUEEN_master_jemalloc_git92d53f9_master_jemalloc_nodiv3_git0bd90b7_master_jemalloc_nodiv4_gitc2bd0b4_bympiprocesses.pdf](https://github.com/nest/nest-simulator/files/871956/JUQUEEN_master_jemalloc_git92d53f9_master_jemalloc_nodiv3_git0bd90b7_master_jemalloc_nodiv4_gitc2bd0b4_bympiprocesses.pdf)

apeyser:
@jakobj What's the difference between A & B?

jakobj:
Right, sorry. So (A) is run time, (B) is build time.

apeyser:
@jakobj 
So, to summarize: there's a small but significant (< 20%) improvement in run time by removing divisions and adding a caching member in runtime; but the difference between adding and not adding the caching of the step time is insignificant. In build time, there a small improvement by the current code (10%?) and the difference between div3 and div4 is insignificant. 

Is this correct? If so, I'd say it worth removing the division but not adding the complexity of the caching. I don't think this conflicts with my results -- in my results, we removed two extra members (both steps and ms),  and Wolfram had eliminated all together the time object construction for event dispatch.  (That was the single biggest problem, that the fat time object was being passed, constructed, extracted, reconstructed with an integer + floating point division, and then repeated).

These changes wouldn't affect the event dispatch, because now the code path is completely gone, and the conversion from int tics to int steps is a fraction of the computational cost from int tics to int steps and float ms. Thus the two biggest problems are unaffected here. 

I'm though curious about B and why the build time increases at all. That shouldn't happen from what I know of the functionality. Something's fishy.

nestkernel/archiving_node.cpp
```diff
@@ -66,6 +68,8 @@ nest::Archiving_Node::Archiving_Node( const Archiving_Node& n )
   , beta_Ca_( n.beta_Ca_ )
   , synaptic_elements_map_( n.synaptic_elements_map_ )
 {
+  tau_minus_inv_ = 1. / tau_minus_;
```
apeyser: This should be copied directly from `const Archiving_Node& n` rather than recomputed from the base value.

nestkernel/event.cpp
```diff
@@ -47,6 +47,7 @@ Event::Event()
   , offset_( 0.0 )
   , w_( 0.0 )
 {
+  steps_ = stamp_.get_steps();
```
apeyser: Shouldn't this just be in the initializer as steps(Time::step(0))?

apeyser: Actually, that should just be steps(0) I think.

nestkernel/nest_time.h
```diff
@@ -473,6 +485,7 @@ class Time
   {
```
apeyser: This is where you're getting extra complexity that will also cost performance: anyplace where you have a loop around `time += n secs` you're going to compute unneeded steps. It's better to cache as in event.h, where you know you need to keep the value, and not everywhere in the code where, if you're lucky, it all balances out to zero, or gives you sometimes a gain, and sometimes a loss.

nestkernel/nest_time.h
```diff
@@ -509,14 +522,26 @@ class Time
...
-    if ( tics == LIM_POS_INF.tics )
-      return LIM_POS_INF.steps;
-    if ( tics == LIM_NEG_INF.tics )
-      return LIM_NEG_INF.steps;
+    return steps_;
+  }
 
-    // round tics up to nearest step
-    // by adding TICS_PER_STEP-1 before division
-    return ( tics + Range::TICS_PER_STEP_RND ) / Range::TICS_PER_STEP;
```
apeyser: But here would be a good place to use an inverted value

nestkernel/nest_time.h
```diff
@@ -537,6 +562,9 @@ class Time
   {
     return ld_round( ms * Range::STEPS_PER_MS );
   }
+
+private:
+  unsigned long steps_;
```
apeyser: Otherwise, if I were going complicated, I wouldn't precompute steps, but cache the value computed in calculate_steps, which would only be computed for get_steps, with an invalid marker in steps_ to tell me when I needed to compute. If the caching is worth it, that's a much more efficient way to do so. 

But I think there's probably places in the code where time is being repeatedly converted to steps (like in event), and simply converting and storing the steps is what you really want.

apeyser: And C is memory usage for the extra members? I wouldn't expect to see anything here -- the issue wouldn't be memory usage, but the extra demand on bandwidth between RAM and caches, and multiple registers used unnecessarily, plus confusing the optimizer by operating on multiple operands that are glued together.

nestkernel/nest_time.h
```diff
@@ -509,14 +522,26 @@ class Time
...
+    return steps_;
+  }
 
-    // round tics up to nearest step
-    // by adding TICS_PER_STEP-1 before division
-    return ( tics + Range::TICS_PER_STEP_RND ) / Range::TICS_PER_STEP;
+  void
+  calculate_steps()
+  {
+    if ( tics == LIM_POS_INF.tics )
```
heplesser: For safety's sake `tics >= ...`

nestkernel/nest_time.h
```diff
@@ -509,14 +522,26 @@ class Time
...
+    {
+      steps_ = LIM_POS_INF.steps;
+    }
+    else if ( tics == LIM_NEG_INF.tics )
```
heplesser: `tics <= ...`

nestkernel/nest_time.h
heplesser: Time can be negative, so `steps_` must be `long`, not `unsigned long`. This, unfortunately, makes it difficult to mark `steps_` as invalid by assigning a special value.

nestkernel/event.h
```diff
@@ -309,6 +309,8 @@ class Event
    */
   Time stamp_;
 
+  unsigned long steps_;
```
heplesser: This is actually "doppeltgemoppelt": If we cache the steps value already in the `Time` object `stamp_`, we do not need to store it additionally in the `Event` object.

heplesser:
@apeyser @jakobj I think I understand now why we see the slight increase in build times: For each connection created, `sender->send_test_event()` creates an `Event` object, which creates a `Time` object `stamp_`. In the current version, this requires computing the `step_` value of the `Time` object, which is never used in connectivity testing. Since this is done once for every single synapse, it hurts.

Furthermore, the current solution now caches the step count twice in each `Event`: once inside the `Time stamp_` and once explicitly as `steps_`. This makes little sense.

The reason that we save time by caching during event delivery is that the spike time is [set once](https://github.com/nest/nest-simulator/blob/baad007903f423c3b5cf45756a4f8441b9239765/nestkernel/event_delivery_manager.cpp#L504) and the event is then sent to all spike targets on the local process. I believe this is the only case in which the stamps value will be re-used many times and where caching thus makes sense.

Therefore, I would propose the following approach:
- The `Time` class does not cache the steps value; it just replaces division by multiplication; this pays off, since the scaling factor needs to be recomputed only when the resolution changes.
- The `Event` class caches the step value of the time stamp using lazy evaluation.
- In the `Event` class, we can exploit the fact that events only occur for `t > 0`, so that we can use `steps_ == 0` as invalid marker (or -1 to be even more on the safe side).
This will avoid the unnecessary computation during the build phase and still ensure gains during the simulation phase.

heplesser:
@jakobj Which benchmark did you use?

nestkernel/nest_time.cpp
```diff
@@ -162,7 +165,7 @@ Time::fromstamp( Time::ms_stamp t )
   // intended ones.
   tic_t n = static_cast< tic_t >( t.t * Range::TICS_PER_MS );
   n -= ( n % Range::TICS_PER_STEP );
-  long s = n / Range::TICS_PER_STEP;
+  long s = n * Range::TICS_PER_STEP_INV;
   double ms = s * Range::MS_PER_STEP;
```
apeyser: What we need here is just MS_PER_TIC (MS_PER_STEP * STEP_PER_TIC).

This still seems like an overcomplex as per the comment on lines 160 + 161, a computation with ceil should be sufficient, but there's corner cases I haven't checked.

nestkernel/nest_time.cpp
```diff
@@ -54,6 +54,7 @@ const double Time::Range::TICS_PER_MS_DEFAULT = CONFIG_TICS_PER_MS;
 const tic_t Time::Range::TICS_PER_STEP_DEFAULT = CONFIG_TICS_PER_STEP;
 
 tic_t Time::Range::TICS_PER_STEP = Time::Range::TICS_PER_STEP_DEFAULT;
+double Time::Range::TICS_PER_STEP_INV = 1. / static_cast< double >( Time::Range::TICS_PER_STEP );
```
apeyser: The cast is unnecessary since the op is 1. / int, it's a float result. But it's a matter of taste what is clearer.

nestkernel/nest_time.cpp
```diff
@@ -77,7 +78,7 @@ Time::compute_max()
   const tic_t tmax = std::numeric_limits< tic_t >::max();
 
   tic_t tics;
-  if ( lmax < tmax / Range::TICS_PER_STEP ) // step size is limiting factor
+  if ( lmax < tmax * Range::TICS_PER_STEP_INV ) // step size is limiting factor
```
apeyser: And we probably don't need to bother with changing these, or inside limit, or inside set_resolution and so on... just in the code that gets deeply looped, which is Time constructors and operators on Time objects.

nestkernel/nest_time.h
apeyser: Is that safer? I think an assert for `tics <= ...` should be there -- it should never be greater than the limit, that's just as broken as missing it altogether.

jakobj:
@apeyser @heplesser thanks for all the helpful comments. I am now removing the caching in the `Time` objects and implement lazy evaluation in `Event`. Should I nevertheless implement your suggested changes in the `Time` class (like `if ( tics == LIM_POS_INF.tics )` -> `if ( tics >= LIM_POS_INF.tics )`)? 
For benchmarks I am using the `hpc_benchmark.sli` with 12500 neurons per node, 8 threads/process. I've added memory consumption for completeness (and because my benchmarking pipeline produces that panel anyway ;) ).

nestkernel/nest_time.cpp
heplesser: @apeyser This does not matter to performance here, but isn't it a good idea to do calculations consistent all over?

nestkernel/nest_time.h
heplesser: @apeyser I am a bit confused---did this comment end up in the wrong place?

nestkernel/nest_time.h
heplesser: I agree with @apeyser

nestkernel/nest_time.h
apeyser: In commit d05335a04bd81bbe56431641e8c49faa88c9d2fb it appears in the right place... but not in the PR discussion it seems. To reduce the confusion I created, I was referring to `calculate_steps()` in `Time(tic_t)` and `Time()`.  By caching there, under some usage patterns, you'll have a huge number of caching steps that are never used. For example, in `vector<Time>` or in temporaries created in computations.

But I believe downstream we've already agreed to just not do this, and do lazy evaluation in the Event class.

nestkernel/nest_time.cpp
gtrensch: I agree with Alex.  To convert milliseconds into tics to step boundary, something like this
`ceil( ms * STEPS_PER_MS ) * TICS_PER_STEP`
will do the same, I believe. Also preserving infinity is an issue here. But however, I think this goes beyond this PR and would be a refactoring task.

nestkernel/nest_time.cpp
gtrensch: @apeyser @heplesser I'd vote for readability. In `struct range{}` there are e.g. definitions for `STEPS_PER_MS` and `MS_PER_STEP`. Wouldn't it be consequent to have `STEPS_PER_TIC` ?

apeyser: @gtrensch Agreed -- in general the "INV"  of "X_PER_Y" would be clearer as "Y_PER_X".

apeyser: @heplesser This is also a question of taste. On the one hand, not changing anything that doesn't need to be changed is "clearer" since we have a minimal change; on the other hand, using all * instead of / is also "clearer".

apeyser:
@jakobj My tendency is to make a minimal change in a change set. Maybe open another pull request with just the fixes to the Time class class?

nestkernel/nest_time.cpp
apeyser: @heplesser @jakobj @gtrensch Agreed with guido -- I suggested later on that we make another PR with the changes in Time that aren't absolutely connected with the caching in Events or inverting the div operations.

heplesser: Agreed.

models/stdp_pl_connection_hom.cpp
```diff
@@ -65,6 +66,7 @@ STDPPLHomCommonProperties::set_status( const DictionaryDatum& d,
   CommonSynapseProperties::set_status( d, cm );
 
   updateValue< double >( d, "tau_plus", tau_plus_ );
+  tau_plus_inv_ = 1. / tau_plus_;
```
apeyser: Something to remember for further reviews and checks, since apparently that's slipped in, in many places earlier and only became apparent now.

jakobj:
I have now implemented the comments as far as I can see, and created a separate PR (#706) to implement all changes in the time class that are not directly related to this  PR. Please have another look.

jakobj:
@heplesser @apeyser I just fixed the merge conflicts, so this is ready to be checked again.

models/stdp_pl_connection_hom.cpp
```diff
@@ -66,7 +66,14 @@ STDPPLHomCommonProperties::set_status( const DictionaryDatum& d,
   CommonSynapseProperties::set_status( d, cm );
 
   updateValue< double >( d, names::tau_plus, tau_plus_ );
-  tau_plus_inv_ = 1. / tau_plus_;
+  if ( tau_plus_ > 0. )
```
apeyser: Is there a constraint on how small tau_plus is? If there wasn't before, then you don't need to add it, but it's probably worth documenting if there's a numeric instability here.

nestkernel/archiving_node.cpp
```diff
@@ -99,7 +103,7 @@ nest::Archiving_Node::get_K_value( double t )
     if ( t > history_[ i ].t_ )
     {
       return ( history_[ i ].Kminus_
-        * std::exp( ( history_[ i ].t_ - t ) / tau_minus_ ) );
+        * std::exp( ( history_[ i ].t_ - t ) * tau_minus_inv_ ) );
```
apeyser: I'd be curious if there's a way to state this in a more numerically stable way -- just to keep in mind for the future

heplesser: @apeyser What kind of numerical stability problems do you see here?

heplesser:
@jakobj It seems something may have gone wrong during the merge, since NEST does not build on Travis:
```
home/travis/build/nest/nest-simulator/nestkernel/nest_time.cpp: In function ‘std::ostream& operator<<(std::ostream&, const nest::Time&)’:
/home/travis/build/nest/nest-simulator/nestkernel/nest_time.cpp:220:1: error: expected ‘}’ at end of input
```

nestkernel/event.h
```diff
@@ -178,7 +178,7 @@ class Event
    *
    * @see NEST Time Memo, Rule 3
    */
-  long get_rel_delivery_steps( const Time& t ) const;
+  long get_rel_delivery_steps( const Time& t );
```
heplesser: Could you add a note in the doxygen comment for the method why it is not `const`? It will make it easier to understand the code, given that developers usually assume that getters are const.

Alternatively, one could keep the method `const` and declare `stamp_steps_` a `mutable`. @apeyser, do you have an opinion on which approach would make for more maintainable code in the long run?

nestkernel/nest_time.cpp
```diff
@@ -211,7 +215,6 @@ std::ostream& operator<<( std::ostream& strm, const Time& t )
     strm << t.get_ms() << " ms (= " << t.get_tics()
          << " tics = " << t.get_steps()
          << ( t.get_steps() != 1 ? " steps)" : " step)" );
-  }
```
heplesser: @jakobj This closing brace is required.

apeyser: I'd go with `mutable` -- it's precisely for this case of cached values.

nestkernel/archiving_node.cpp
apeyser: We'd have to run the numbers -- but we're combining root(tau, e^t1/e^t2), so it's worth looking at the relative scales of all three numbers (how big can t1, t2 and t3 get?) to see whether we can blow up in different ways. This formulation looks reasonable -- but I don't know the cases, so i can't say of the top of my head -- just asking whether it's been thought about.

jakobj:
Thanks for the input. Done implementing your proposed changes. Didn't know about `mutable` before - nice feature for this type of situation.

heplesser:
@jakobj It seems clang-format is unhappy.

heplesser:
Release notes: Performance improvements through more efficient numerics in Time

Issue #683, opened 2017-03-16
Custom connector models in NEST 2.12


kappeld:
This issue effects the modified API for registering connector models at the new ModelManager of NEST 2.12.

We have developed a dynamically loadable Module for NEST that implements a reward-based learning rule in a custom connector model. To implement this synapse model it was necessary for us to create a new type of synapses that is guaranteed to be updated after a certain maximum time lag instead of only waiting for a presynaptic spike event. The best way to achieve this, we found, was to write a new connector model that takes care of the regular updates, and which is registered using NEST's standard register_connection_model, at the NEST kernel. This solution has proven quite efficient and flexible, and worked very well with NEST 2.10. We are sharing this Module already with two other groups and were planning to make it available to the public soon.

However, the feature that we were using to register our connector model has been removed in NEST 2.12, more precisely the method `synindex register_connection_model_( ConnectorModel* );`  of the ModelManager is now private. This makes it impossible for a user of NEST to develop custom connector models as far as I can see.

The feature we have been implementing (synapses that are updated on a regular basis) is something that is of greater interest, I believe, since it is needed for many interesting synapse models such as the reinforce / Clopath- like update rules. It would be a pity if this class of models is lost for the NEST community. Is there a plan to introduce this feature in NEST in one or the other way in future releases? To get this feature functioning again I would need one of the following things:

1) Give the user again access to `ModelManager->register_connection_model_( ConnectorModel* );` I do understand why you decided to make this method private since it somehow provides a "backdoor" which an inexperienced user could use to break many things. However, with this step you made it impossible for a user of NEST to write a new ConnectorModel. On the long run, giving developers who write custom models for NEST the opportunity to register synapses that use new connector models might be an important feature to increase the flexibility of NEST. On could do this, e.g. through a protected member function of `ConnectorModel` such that only developers that really want to write a new connector model have access to this.

2) A second solution would be to put the option to have regular updates of synapse models into the NEST kernel. Since this feature might be interesting to many users of NEST, as described above, this option might be worthwhile considering. However, it probably comes with quite some overhead.

I'm quite agnostic to what solution is chosen to get this feature back (any of the two I proposed or another option). In any case I would be happy to provide code and help. For the first option I could easily implement a solution and make a pull request. Is there already a plan to realize something of this kind in a future version of NEST?

heplesser:
[...]

PR #686, opened 2017-03-21
Fix to issue #683.

kappeld:
- A second template argument to member function register_connection_model of the ModelManager has been added, which allows the user to specify the type of connector model.
- An overloaded function to register_connection_model has been added that uses the default connector model GenericConnectorModel.

### DIFF
--- a/nestkernel/model_manager.h
+++ b/nestkernel/model_manager.h
@@ -180,11 +180,21 @@ public:
   void set_model_defaults( Name name, DictionaryDatum params );
 
   /**
-   * Register a synape with default Connector and without any common properties.
+   * Register a synape model with default Connector and without any common
+   * properties. Convenience function that used the default Connector model
+   * GenericConnectorModel.
    * @param name The name under which the ConnectorModel will be registered.
-   * @return an ID for the synapse prototype.
    */
   template < class ConnectionT >
+  void register_connection_model( const std::string& name,
+    bool requires_symmetric = false );
+    
+  /**
+   * Register a synape model with a costom Connector model and without any
+   * common properties.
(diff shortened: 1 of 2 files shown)


nestkernel/model_manager.h
```diff
@@ -180,14 +180,24 @@ class ModelManager : public ManagerInterface
...
+   * GenericConnectorModel.
    * @param name The name under which the ConnectorModel will be registered.
-   * @return an ID for the synapse prototype.
    */
   template < class ConnectionT >
   void register_connection_model( const std::string& name,
     bool requires_symmetric = false );
 
+  /**
+   * Register a synape model with a costom Connector model and without any
```
heplesser: costom -> custom

nestkernel/model_manager.h
```diff
@@ -180,14 +180,24 @@ class ModelManager : public ManagerInterface
...
+   * common properties.
+   * @param name The name under which the ConnectorModel will be registered.
+   */
+  template < class ConnectionT, template < class > class ConnectorModelT >
+  void register_connection_model( const std::string& name,
+    bool requires_symmetric = false );
```
heplesser: For new code, please use `typename` instead of `class` for the template argument type; maybe fix that in the old version of the function as well.

kappeld:
I also changed the template argument type of register_secondary_connection_model to typename such that ConnectionT is consistently used with the same type in all methods. typename is not allowed in the context of ConnectorModelT. Here we have to use "class".

heplesser:
- Extend internal API to allow registration of ConnectorModels not based on GenericConnectorModel.
- Fixes #683.

Issue #246, opened 2016-03-01
Make help and helpdesk more robust


heplesser:
Currently, the NEST `help` and `helpdesk` commands can be rather difficult to use. In particular, they won't work if NEST has been compiled with MPI support, since both require spawning subprocesses.
- From Python, display of help information and opening of the helpdesk should bypass the NEST kernel completely, thus avoiding any problems with spawning
- It should not be necessary to specify the command to use for opening the helpdesk in `.nestrc`, the default browser should be used unless the user specifies a different browser explicitly.

This is a follow-up to trac.963.

heplesser:
There is an additional issue here: The help system does not work properly with Jupyter notebooks. Calling `nest.help('iaf_psc_alpha')` will use `more` (or whichever pager is defined in `.nestrc` to display the help in the terminal in which `jupyter` was started and the notebook will hang until the user closes the pager. This is confusing at best and completely useless on web-based systems such as the HBP Collab.

I would suggest a solution along the following lines:
- From the SLI level
    - without MPI: current solution
    - with MPI: dump help to screen directly (opening hlp file in sli and print line-by-line
- From Python
    - print hlp file to screen directly at the Python level without going through SLI `help`
- From Jupyter notebook
    - As for Python, but display in subwindow, just as the help magic in Jupyter does
Concerning `nest.helpdesk()` from a Jupyter window, can we open a new tab with the NEST helpdesk without spawning a new browser process? After all, we are already in a browser. This would avoid the problem that we cannot spawn when NEST is compiled with MPI.

heplesser:
@steffengraber Would the new help generator of #464 be useful to address this?

steffengraber:
@heplesser Do you have an example how the help should look like for Jupyter?

heplesser:
@steffengraber It can look just as when displayed in a terminal, the problem is where it is displayed. Currently, `nest.help()` calls SLI `help`, which calls `page`, which in turn invokes the pager specified by the user (`more` or `less`). The pager then outputs directly to `stdout`, but in the terminal in which the user invoked `jupyter notebook`, not in the notebook window in the browser.

Ideally, the help text should be displayed just as Python help is displayed in the notebook window, e.g., by creating a sub-window at the bottom of the window. In any case, we need a help function at the SLI level that returns the help text as string to Python instead of paging it. The same is true for `PrintNetwork`.

steffengraber:
[...]

PR #688, opened 2017-03-22
Fix of  issue 246 - Make help and helpdesk more robust

steffengraber:
With this pull request, NEST uses only Python for the help system if you are on the Python level. 
Therefore it doesn‘t matter if it is compiled with MPI or not.

In jupyter notebooks `nest.help(obj)` open a modal window with the help text. 
With `nest.help(obj, pager)`,  you can open a texeditor like gedit from a notebook. If you use an terminaleditor (vi, nano, less, …) as pager, again the modal window open up.

If you use `nest.helpdesk()` in a notebook a new tab is created.

This pull request is a fix for issue #246.


heplesser:
@steffengraber I am testing now and here are some experiences:
- If `NEST_INSTALL_DIR` is not defined, `nest.help()` crashes. It would be better to provide the user with more guidance, something like:
```Python
if 'NEST_INSTALL_DIR' not in os.environ:
    print('NEST help needs to know where NEST is installed. Please source nest_vars.sh or defined NEST_INSTALL_DIR manually.')
    return
```
- Instead of `os.environ['NEST_INSTALL_DIR'] + "/share/doc/nest/help/"` it is better to use `os.path.join()`
- Is it best to use `less` as default parser, or would should we use the even more default `more`? no matter what we use as default, we should check that it exists and if not, either use `cat` or read the help file into Python and `print()` it.
- We should avoid the low-level `subprocess.Popen()` and rather use the high-level functions. Since we need to maintain compatibility back to 2.7, we should use the [Old-style high-level API](https://docs.python.org/3/library/subprocess.html#call-function-trio).
- I got this error: `FileNotFoundError: [Errno 2] No such file or directory: 'less -S'`, which is probably because `subprocess` does not like options combined with the command name in one string (calling from plain python or ipython); the `-S` comes from `.nestrc`, if I remove it there, things work
- `nest.help('abs')` works from notebook
- `nest.helpdesk() does nothing when called from plain python or ipython or notebook

steffengraber:
@janhahne Sorry for the late response. 
How do you compile nest? With MPI on or off? 
Under which operating system?

janhahne:
@steffengraber I compiled nest with MPI on and my operating system is openSUSE 13.1. The problem still persist with your recent commits. Just to prevent any error on my side: `nest.help('iaf_psc_alpha')` should be a correct use of `nest.help`, right?

steffengraber:
@janhahne yes, nest.help('iaf_psc_alpha') is right.
Is python installed, and what version?

janhahne:
@steffengraber I am using python 2.7.6 and usually pynest works fine for me

steffengraber:
@janhahne with Python older than 2.7.8 the help generation did not work correctly. Is it possible for you to update and test ist again?

janhahne:
@steffengraber I updated python to 2.7.13 and did a new clean build, but the error remains the same. I also noticed that if I enter any object that does not exist (as for example `nest.help('testtest')`) nothing happens, so there is no output and no exception.

steffengraber:
@janhahne Could you please try nest.help('iaf_psc_alpha', 'less').

janhahne:
@steffengraber `nest.help('iaf_psc_alpha', 'less')` works fine. The corresponding documentation is displayed and no error occurs.

pynest/nest/lib/hl_api_info.py
```diff
@@ -55,23 +58,23 @@ def authors():
...
-    ----------
-    browser : str, optional
-        Name of the browser to use
+    The system default browser.
+    /helpdesk << /command (firefox) >> SetOptions in ~/.nestrc is now obsolete
     """
-
-    sr("/helpdesk << /command (%s) >> SetOptions" % browser)
-    sr("helpdesk")
+    url = os.path.join(os.environ['NEST_DOC_DIR'] + "/help", "helpindex.html")
```
heplesser: Provide a useful error message if `NEST_DOC_DIR` is undefined, similar as for missing `NEST_INSTALL_DIR`.

pynest/nest/lib/hl_api_info.py
```diff
@@ -55,23 +58,23 @@ def authors():
...
+    if sys.platform[:3] == "win":
+        os.startfile(url)
```
heplesser: Why don't you use `webbrowser` also under Windows?

pynest/nest/lib/hl_api_info.py
```diff
@@ -55,23 +58,23 @@ def authors():
...
+    if sys.platform[:3] == "dar":
+        webbrowser.get('safari').open_new(url)
```
heplesser: Why not `webbrowser.open_new(url)` also here? Ok, due to [this problem in OSX](http://www.andrewjaffe.net/blog/2017/05/python-bug-hunt.html) we need to ask for the browser explicitly under macOS are present (10.12.5). Worse, a failing `webbrowser.open_new(url)` does not even raise a Python exception but returns happily. So I think we need to keep this, but i would be good to document why we do this.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -389,17 +395,122 @@ def broadcast(item, length, allowed_types, name="item"):
...
+    if 'NEST_INSTALL_DIR' not in os.environ:
+        print(
+            'NEST help needs to know where NEST is installed.'
+            'Please source nest_vars.sh or define NEST_INSTALL_DIR manually.')
+        return
+
+    helpdir = os.path.join(os.environ['NEST_INSTALL_DIR'], "share", "doc",
+                           "nest", "help")
+    objname = hlpobj + '.hlp'
+    # @todo more pager
```
heplesser: Isn't the pager list pretty complete?

pynest/nest/lib/hl_api_info.py
```diff
@@ -55,23 +58,23 @@ def authors():
...
+    """Open the NEST helpdesk in browser.
 
-    The default browser is firefox.
-
-    Parameters
-    ----------
-    browser : str, optional
-        Name of the browser to use
+    The system default browser.
+    /helpdesk << /command (firefox) >> SetOptions in ~/.nestrc is now obsolete
```
heplesser: I would drop this line of documentation, since it will confuse most users. The nestrc line is still relevant for the SLI level, so I'd leave it in nestrc.

heplesser:
@steffengraber PEP8 was still not entirely happy:
```
MSGBLD0195: [PEP8] pynest/nest/lib/hl_api_info.py:74:56: W291 trailing whitespace
MSGBLD0195: [PEP8] pynest/nest/lib/hl_api_info.py:80:59: W291 trailing whitespace
```
Otherwise, this looks fine to me.

@janhahne Do you also explicitly approve?

steffengraber:
@heplesser  
@janhane found 2 more issues I have to address:

- exception needed fo something like nest.help('testtest')
- nest.help('iaf_psc_alpha', 'less') works perfect for @janhahne but not nest.help('iaf_psc_alpha').

pynest/nest/lib/hl_api_helper.py
```diff
@@ -472,23 +472,33 @@ def show_help_with_pager(hlpobj, pager):
...
-                pypagers = re.findall(
-                    r'\s?/page\s?<<\s?/command\s?\((.*)\).*', line)
-                if pypagers:
-                    pager = pypagers[0]
-                    break
-                else:
-                    pager = 'less'
-        rc.close()
-
+        # check if .netsrc exist
```
heplesser: .nestrc

pynest/nest/lib/hl_api_helper.py
```diff
@@ -472,23 +472,33 @@ def show_help_with_pager(hlpobj, pager):
...
+        rc_file = os.environ['HOME'] + '/.nestrc'
```
heplesser: You should use `os.path.join(os.environ['HOME'], '.nestrc')` for portable path construction; applies in other places as well.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -472,23 +472,33 @@ def show_help_with_pager(hlpobj, pager):
...
+        if os.path.isfile(rc_file):
+            # open ~/.nestrc
+            rc = open(os.environ['HOME'] + '/.nestrc', 'r')
```
heplesser: `rc = open(rc_file, 'r')`

pynest/nest/lib/hl_api_helper.py
```diff
@@ -472,23 +472,33 @@ def show_help_with_pager(hlpobj, pager):
...
+            for line in rc:
+                rctst = re.match(r'^\s?%', line)
```
heplesser: Could you add a comment explaining/illustrating briefly what this RE looks for; that will make the code more maintainable. Applies also to other REs.

I think it would also be helpful if you could explain what this entire loop is trying to do---I cannot quite get it just from the code.

pynest/nest/lib/hl_api_info.py
heplesser: Here you mix `os.path.join()` with string addition to build a URL. I think since we want a URL, we should not use `os.path.join()` here.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -389,17 +395,137 @@ def broadcast(item, length, allowed_types, name="item"):
...
+                else:
+                    if pager in consolepager:
+                        subprocess.call([pager, objf])
+                        fhlp.close()
+                        break
+                    else:
+                        subprocess.call([pager, objf])
+                        fhlp.close()
+                        break
+    if hlperror == True:
```
heplesser: `if hlperror` for tidier code and to make PEP8 happy.

steffengraber:
@heplesser I added some documentation and PEP8 is happy now.

heplesser:
Release notes:  Improved help in PyNEST
- Help will work even if compiled with MPI
- Help will work in Jupyter Notebooks

pynest/nest/lib/hl_api_info.py
```diff
@@ -55,26 +58,37 @@ def authors():
...
     """
+    if 'NEST_DOC_DIR' not in os.environ:
+        print(
+            'NEST help needs to know where NEST is installed.'
+            'Please source nest_vars.sh or define NEST_DOC_DIR manually.')
+        return
 
-    sr("/helpdesk << /command (%s) >> SetOptions" % browser)
-    sr("helpdesk")
+    url = os.path.join(os.environ['NEST_DOC_DIR'] + "/help", "helpindex.html")
```
heplesser: I still have some issues with this line:
1. This is not a URL, but the name of a local file to be opened by a browser, so the variable name should be `helpfile` or similar, not `url`.
2. Instead of mixing string concatenation and `os.path.join()`, just use `os.path.join()`:
```python
helpfile = os.path.join(os.environ['NEST_DOC_DIR'], 'help', 'helpindex.html')
```

pynest/nest/lib/hl_api_helper.py
```diff
@@ -389,17 +395,146 @@ def broadcast(item, length, allowed_types, name="item"):
...
+    helpdir = os.path.join(os.environ['NEST_INSTALL_DIR'], "share", "doc",
+                           "nest", "help")
+    objname = hlpobj + '.hlp'
+    consolepager = ['less', 'more', 'vi', 'vim', 'nano', 'emacs -nw',
+                    'ed', 'editor']
+
+    # reading ~/.nestrc lookink for pager to use.
+    if pager is None:
+        # check if .netsrc exist
+        rc_file = os.environ['HOME'] + '/.nestrc'
```
heplesser: This should be
```python
rc_file = os.path.join(os.environ['HOME'], '.nestrc')
```

PR #691, opened 2017-03-22
Changed all files to comply to code style checks.

terhorstd:
All files (except one) now comply to the code style checked by VERA++. The only non-compliance is due to a VERA-1.2.1 issue with `pragma parallel for` in nestkernel/connection_manager.cpp, which can be fixed by using VERA-1.3.1.


heplesser:
@terhorstd Thank you for this massive effort!

For review, I suggest we split the effort as follows (a bit random):
- @heplesser Directories librandom, sli 
- @hakonsbm models
- @stinebuu precise, topology
- @gtrensch all else

Does this sound like a plan?

extras/static_code_analysis.sh
```diff
@@ -50,6 +50,9 @@ PEP8_IGNORES="E121,E123,E126,E226,E24,E704"
 PEP8_IGNORES_EXAMPLES="${PEP8_IGNORES},E402"
 PEP8_IGNORES_TOPO_MANUAL="${PEP8_IGNORES_EXAMPLES},E265"
 
+# Drop files that should not be checked (space-separated list).
+FILES_TO_IGNORE="libnestutil/sparsetable.h"
```
gtrensch: Files that are not under the NEST Initiative Copyright should be also added.
- compose.hpp
- hashtable-common.h
- libc_allocator_with_realloc.h
- sparseconfig.h
- sparsetable.h
- template_util.h
- type_traits.h

libnestutil/type_traits.h
```diff
@@ -71,7 +71,7 @@ struct is_floating_point;
 template < class T >
 struct is_pointer;
 // MSVC can't compile this correctly, and neither can gcc 3.3.5 (at least)
-#if !defined( _MSC_VER ) && !( defined( __GNUC__ ) && __GNUC__ <= 3 )
+#if not defined( _MSC_VER ) && not( defined( __GNUC__ ) && __GNUC__ <= 3 )
```
gtrensch: 'not' isn't supported by the preprocessor language. Since 'not' is syntactically correct (it is a valid macro name), the preprocessor will generate a warning in the best case. 'not' should be replaced with '!' or the expression can also be re-written using #ifndef.

libnestutil/type_traits.h
```diff
@@ -102,7 +102,7 @@ template < class T >
 struct remove_pointer;
 template < class T, class U >
 struct is_same;
-#if !defined( _MSC_VER ) && !( defined( __GNUC__ ) && __GNUC__ <= 3 )
+#if not defined( _MSC_VER ) && not( defined( __GNUC__ ) && __GNUC__ <= 3 )
```
gtrensch: 'not' isn't supported by the preprocessor language. Since 'not' is syntactically correct (it is a valid macro name), the preprocessor will generate a warning in the best case. 'not' should be replaced with '!' or the expression can also be re-written using #ifndef.

libnestutil/type_traits.h
```diff
@@ -245,7 +245,7 @@ struct is_pointer< const volatile T > : is_pointer< T >
 {
 };
 
-#if !defined( _MSC_VER ) && !( defined( __GNUC__ ) && __GNUC__ <= 3 )
+#if not defined( _MSC_VER ) && not( defined( __GNUC__ ) && __GNUC__ <= 3 )
```
gtrensch: 'not' isn't supported by the preprocessor language. Since 'not' is syntactically correct (it is a valid macro name), the preprocessor will generate a warning in the best case. 'not' should be replaced with '!' or the expression can also be re-written using #ifndef.

libnestutil/type_traits.h
```diff
@@ -331,7 +331,7 @@ struct is_reference< T& > : true_type
 template < class T >
 struct is_pod : integral_constant< bool,
                   ( is_integral< T >::value || is_floating_point< T >::value ||
-#if !defined( _MSC_VER ) && !( defined( __GNUC__ ) && __GNUC__ <= 3 )
+#if not defined( _MSC_VER ) && not( defined( __GNUC__ ) && __GNUC__ <= 3 )
```
gtrensch: 'not' isn't supported by the preprocessor language. Since 'not' is syntactically correct (it is a valid macro name), the preprocessor will generate a warning in the best case. 'not' should be replaced with '!' or the expression can also be re-written using #ifndef.

libnestutil/type_traits.h
```diff
@@ -535,7 +535,7 @@ struct is_same< T, T > : public true_type
 };
 
 // Specified by TR1 [4.6] Relationships between types
-#if !defined( _MSC_VER ) && !( defined( __GNUC__ ) && __GNUC__ <= 3 )
+#if not defined( _MSC_VER ) && not( defined( __GNUC__ ) && __GNUC__ <= 3 )
```
gtrensch: 'not' isn't supported by the preprocessor language. Since 'not' is syntactically correct (it is a valid macro name), the preprocessor will generate a warning in the best case. 'not' should be replaced with '!' or the expression can also be re-written using #ifndef.

extras/static_code_analysis.sh
gtrensch: I reviewed the files in following directories:
- conngen
- examples
- extras
- libnestutil

Everything looks fine. There is just one comment on the exclusion list, se eabove.

heplesser:
@terhorstd @jougs How do we handle the Vera++ problem in `connection_manager.cpp`?  I first thought about installing Vera 1.3.0 on the test machine manually, but building it requires boost, and that would make the build process even tougher. Should we temporarily put `connection_manager.cpp` on the skip list for static code checks?

jougs:
@heplesser: I am not sure if it is such a good idea, if the authors of this PR are also its reviewers. Why don't we also draw in other people?

How can Travis signal green if there is still a problem? Is the solution in Travis 1.3.0 due to a proper fix or just an updated (Tcl/Tk) rule file? The latter could probably be installed without too much effort. I'm rather against adding such a central file to the skip list, as these lists are usually not regularly updated and the file might be forgotten there...

heplesser:
@jougs I will answer the technical questions first:
- [Travis build 1774 for NEST master](https://travis-ci.org/nest/nest-simulator/builds/213876599), i.e., the one that shows green here, included only some 30 files in static code analysis. I am not sure why this happened, but it must have something to do with how we identify files that have changed. @gtrensch Could you take a look at this and create a new issue? Travis used Vera++ 1.2.1.
- [Travis build 410 for my NEST repo](https://travis-ci.org/heplesser/nest-simulator/builds/214316548), on the other hand, checked more than 250 files (even though I only change a few files in my PR), and that included `connection_manager.cpp`, so that build failed.
- I noticed that the current builds install boost as a Vera++-dependency anyways, so installing boost explicitly and then build Vera++ 1.3.0 manually on the Travis system would not be much extra burden.

Concerning who should review, I would consider this a two-tier thing in this case. This PR is massive, but changes are very mechanical, and have mostly been applied through project-wide regular expression substitution. One tedious but straightforward part of the review therefore is to scroll through heaps of code and check that {} ended up in the right places. I think most files have now been double-checked by people who have not (manually) added the same files, so we have a cross-check, though team-internal. I suggested this because I figured this would allow us to wade through a lot of code fast.

But we should probably invite a one or two high-level reviewers who look at the changes we made to infrastructure (travis.yml, vera++ profile, ...) and checks a smaller selection of files. Any suggestions on who to invite?

terhorstd:
Regarding the the Vera++ problem in `connection_manager.cpp`, I think this goes away with the `#define NEST_PARALLEL_FOR` construct that @jougs and me discussed about. Do we have a PR for that already?

jougs:
@terhorstd: here's the macro you've talked about:
```c
#ifdef _OPENMP
  #define NEST_THREADED_LOOP_BEGIN(t) \
    _Pragma("omp parallel for") \
    { \
      size_t t = kernel().vp_manager.get_thread_id();
#else
  #define NEST_THREADED_LOOP_BEGIN(t) \
    for ( size_t t = 0; t < kernel().vp_manager.get_num_threads(); ++t ) \
    {
#endif

#define NEST_THREADED_LOOP_END }
```

Maybe you can really rename it to `NEST_PARALLEL_FOR` to be closer to the OpenMP original.

heplesser:
@terhorstd With #694, I have disabled the Vera check in master so that we can merge PRs until this PR is merged, without having to apply style fixes to individual files. Could you merge that change from master and then enable the Vera-check again for this PR, so that the Vera-check will become permanently enabled once this PR is merged?

heplesser:
@jougs Would you agree to merging this now? After discussion with @terhorstd we leave the vera check inactive because there is still a problem with one line in `connection_manager.cpp`; this will be addressed either by #668 or #695, but this PR should not have to wait for them to avoid code rot.

gtrensch:
@heplesser @jougs #668 will turn on the VERA check again. I would suggest the following procedure:
1. merging this PR
2. turning off VERA message checking, i.e. set IGNORE_MSG_VERA, in #668
3. merging PR #668  
4. adding connection_manager.cpp to the exclusion list (don't know what the state of that is at the moment)
5. provided that there are no messages, turn off IGNORE_MSG_VERA

heplesser:
@gtrensch I agree with nr 1-3 and eventually 5. Nr 4 may not be necessary, #695 probably will solve this issue in a much more elegant manner.

heplesser:
This PR fixes the formatting of a very large number of C++ files in the NEST code base, especially
- placing all one-line blocks in curly braces
- replacing `!` with `not` in all boolean expressions

gtrensch:
The IGNORE_MSG_VERA flag has been set. See PR #668.

PR #692, opened 2017-03-23
Move flush from finalize to new function post_run_cleanup()

jougs:
This is related to #650 and moves flushing of files from `Cleanup` (i.e. `SimulationManager::finalize()`) to the end of `Run` (i.e. `SimulationManager::run()`).


nestkernel/node.h
```diff
@@ -299,11 +299,21 @@ class Node
   virtual void calibrate() = 0;
 
   /**
+   * Cleanup node after Run.
+   * Override this function if a node needs to "wrap up" things after
+   * a call to Run, i.e., before SimulationManager::run(). Typical
```
heplesser: A bit confused by the *before* `SimulationManager::run()`---why not *after*?

nestkernel/node_manager.cpp
```diff
@@ -851,6 +851,38 @@ NodeManager::prepare_nodes()
   LOG( M_INFO, "NodeManager::prepare_nodes", os.str() );
 }
 
+void
+NodeManager::post_run_cleanup()
+{
+#ifdef _OPENMP
+#pragma omp parallel
+  {
+    index t = kernel().vp_manager.get_thread_id();
```
heplesser: Should be `const`

nestkernel/node_manager.cpp
```diff
@@ -851,6 +851,38 @@ NodeManager::prepare_nodes()
...
+#else // clang-format off
+  for ( index t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
+  {
+#endif // clang-format on
+    for ( size_t idx = 0; idx < local_nodes_.size(); ++idx )
+    {
+      Node* node = local_nodes_.get_node_by_index( idx );
```
heplesser: Couldn't you use the `begin()` and `end()` iterators of `local_nodes_` to avoid lookup by index?

nestkernel/recording_device.cpp
```diff
@@ -581,6 +581,25 @@ nest::RecordingDevice::calibrate()
 }
 
 void
+nest::RecordingDevice::post_run_cleanup()
+{
+  if ( B_.fs_.is_open() )
+  {
+    if ( P_.flush_after_simulate_ )
+      B_.fs_.flush();
```
heplesser: Curly braces missing.

nestkernel/recording_device.cpp
```diff
@@ -581,6 +581,25 @@ nest::RecordingDevice::calibrate()
...
+
+    if ( !B_.fs_.good() )
```
heplesser: Should be `not`.

nestkernel/recording_device.cpp
```diff
@@ -581,6 +581,25 @@ nest::RecordingDevice::calibrate()
...
+    {
+      std::string msg =
```
heplesser: Could be `const`

wschenck:
I have no objections to add. Nice work. I endorse the pull request. :+1:

nestkernel/node.h
jougs: Should have been "before `SimulationManager::run()` returns.". Fixed.

nestkernel/node_manager.cpp
jougs: 1) Yes, it could be. However, I just copied the code from the `finalize()` function and did not want to change or refactor the functionality at this point. The code will also go away when nestio is merged.

nestkernel/node_manager.cpp
jougs: See 1) above.

nestkernel/recording_device.cpp
jougs: 2) Yes. And also in hundreds of other places in this file. Can't I just disable the checker until #691 is merged? I don't really see a reason for fixing this all over again...

nestkernel/recording_device.cpp
jougs: See 2. above.

nestkernel/recording_device.cpp
jougs: See 1. above.

PR #693, opened 2017-03-23
Throw exception when connecting fails

### LINKED ISSUE #578: NEST allows multimeter connections to devices that do not support recording
### Description of problem
NEST allows connections to devices that do not support recording.

### Version-Release or Git commit
Master @ 782d30a

### How reproducible?
Always

### Steps to Reproduce
```IPython
In [3]: scg = nest.Create('step_current_generator', params={'amplitude_times': [10., 50.], 'amplitude_values': [100., 200.]})
In [5]: mm = nest.Create('multimeter', params={'record_from': ['amplitude']})
In [6]: nest.Connect(mm, scg)   # Error message expected here
 In [7]: nest.Simulate(100)
n [8]: nest.GetStatus(mm)
Out[8]: 
({'events': {'amplitude': array([], dtype=float64),
   'senders': array([], dtype=int64),
   'times': array([], dtype=float64)},},)
```

### Actual results
- No error message, `events` dictionary has empty entries after simulation.
- No connection is created (checked by `nest.GetKernelStatus()['num_connections']`)

### Expected results
Error message upon `Connect` expected.

### Additional info
- The same problem also applies to other generators such as `dc_generator` and `spike_generator`.
- NEST issue a proper error message when trying to connect a multimeter with non-existing recordable to `iaf_psc_alpha`.

hakonsbm:
This PR fixes an issue where in some cases an attempt to create a connection fails silently, such as when trying to connect a multimeter to a generator. NEST will now throw an exception in these cases. 

Additionally, a part of the regressiontest issue-211 that now throws an exception has been removed, and tests to make sure that NEST throws an exception to avoid silent failure has been added.

This fixes #578.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -413,7 +413,8 @@ nest::ConnectionManager::connect( index sgid,
     // we do not allow to connect a device to a global receiver at the moment
     if ( not source->has_proxies() )
     {
-      return;
+      throw IllegalConnection( "The models " + target->get_name() + " and "
+        + source->get_name() + " cannot be connected." );
     }
     connect_( *source, *target, sgid, tid, syn, d, w );
   }
@@ -474,7 +475,8 @@ nest::ConnectionManager::connect( index sgid,
     // we do not allow to connect a device to a global receiver at the moment
     if ( not source->has_proxies() )
     {
-      return;
+      throw IllegalConnection( "The models " + target->get_name() + " and "
+        + source->get_name() + " cannot be connected." );
     }
     connect_( *source, *target, sgid, tid, syn, params, d, w );
   }
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -543,7 +543,8 @@ nest::ConnectionManager::connect( index sgid,
     // we do not allow to connect a device to a global receiver at the moment
     if ( not source->has_proxies() )
     {
-      return false;
+      throw IllegalConnection( "The models " + target->get_name() + " and "
+        + source->get_name() + " cannot be connected." );
     }
     connect_( *source, *target, sgid, tid, syn, params );
   }
--- a/testsuite/regressiontests/issue-211.sli
+++ b/testsuite/regressiontests/issue-211.sli
@@ -68,8 +68,8 @@ n2 d4 << /weight 1. >> Connect
 [g5 g5] [n1] Connect
 g5 n3 << /weight 1. >> Connect
 
-% device-device; but g5g5 should be ignored
-[g5 g5] [d4 g5] /one_to_one Connect
+% device-device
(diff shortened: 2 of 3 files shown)


heplesser:
@jakobj Since you wrote the test that is modified here, could you review the change in the `issue-221.sli`?

testsuite/regressiontests/issue-211.sli
```diff
@@ -68,11 +68,6 @@ n2 d4 << /weight 1. >> Connect
 [g5 g5] [n1] Connect
 g5 n3 << /weight 1. >> Connect
 
-% device-device; but g5g5 should be ignored
```
jakobj: why did you remove this whole block? I think the connection g5 -> g4 is still valid, and actually created, so I would suggest just removing the first line (`[g5 g5] [d4 g5] /one_to_one Connect`) and keeping the others. This would imply that you need to remove one `[5 4 0]` from the `target_conn` list.

testsuite/unittests/test_connect.sli
```diff
@@ -177,3 +177,52 @@ ResetKernel
   Last /target get 6 eq assert_or_die
 
 } forall
+
```
jakobj: maybe it would make sense to get all devices from the modeldict and loop through all possible combinations instead of listing them manually here? this would avoid the need to add new tests here upon adding a new device and make sure that we do not forget a certain combination.

nestkernel/connection_manager.cpp
```diff
@@ -413,7 +413,8 @@ nest::ConnectionManager::connect( index sgid,
     // we do not allow to connect a device to a global receiver at the moment
     if ( not source->has_proxies() )
     {
-      return;
+      throw IllegalConnection( "The models " + target->get_name() + " and "
```
jakobj: should we throw a similar error in line546?

hakonsbm:
@jakobj Thank you for you review. Besides reinstating the missing test in issue-211 and adding a `throw IllegalConnection` in the connection manager, I have expanded the test in test_connect.sli to check all combinations of models. The test fails if there is no thrown error, but no connection made.

testsuite/unittests/test_connect.sli
```diff
@@ -180,49 +180,38 @@ ResetKernel
...
 
+/n_connections 0 def
+modeldict
 {
-  generator generator Connect
-} fail_or_die
+  pop  % remove model id
+  /model_a Set
+  /type_a model_a GetDefaults /element_type get def 
+  /m_a model_a Create def
```
heplesser: It looks to me as if you do not need `type_a` anywhere, so I think you can shorten this to
```
pop exch Create /node_a Set
```
and similarly for `b` below.

heplesser:
@jakobj Could you have another look?

heplesser:
@jakobj I will merge as is---still a bit wary of re-writing history after it has become visible in public repositories.

PR #695, opened 2017-03-28
rewrote most omp parallel sections to NEST_PARALLEL_... macro

terhorstd:
This PR unifies many parallel sections in using a single macro definition.

### DIFF
--- a/examples/MyModule/step_pattern_builder.cpp
+++ b/examples/MyModule/step_pattern_builder.cpp
@@ -101,10 +101,7 @@ mynest::StepPatternBuilder::connect_()
 // This code is based on AllToAllBuilder, except that we step
 // by source_step_ and target_step_ except for stepping by 1.
 
-#pragma omp parallel
-  {
-    // get thread id
-    const int tid = nest::kernel().vp_manager.get_thread_id();
+  NEST_PARALLEL_FOR_THREAD(tid)
     try
 
     {
@@ -146,7 +143,7 @@ mynest::StepPatternBuilder::connect_()
       exceptions_raised_.at( tid ) =
         lockPTR< WrappedThreadException >( new WrappedThreadException( err ) );
     }
-  }
+  NEST_PARALLEL_END
 }
 
 nest::GIDCollection::const_iterator&
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -111,11 +111,13 @@ nest::ConnectionManager::initialize()
   assert( n_threads <= MAX_THREAD
     && "MAX_THREAD is a constant defined in allocator.h" );
 
-#pragma omp parallel
-  poormansallocpool[ kernel().vp_manager.get_thread_id() ].init();
+  NEST_PARALLEL_FOR_THREAD(t)
+    poormansallocpool[ t ].init();
+  NEST_PARALLEL_END
 #else
-#pragma omp parallel
-  poormansallocpool.init();
+  NEST_PARALLEL_FOR_THREAD(t)
+    poormansallocpool.init();
+  NEST_PARALLEL_END
 #endif
 #endif
 #endif
@@ -782,9 +784,7 @@ nest::ConnectionManager::data_connect_single( const index source_id,
     return;
   }
 
-#pragma omp parallel private( di_s )
-  {
-    thread tid = kernel().vp_manager.get_thread_id();
+  NEST_PARALLEL_FOR_THREAD(tid)
     DictionaryDatum par_i( new Dictionary() );
 
     size_t n_targets = target_ids.size();
@@ -862,7 +862,7 @@ nest::ConnectionManager::data_connect_single( const index source_id,
         continue;
       }
     }
-  }
+  NEST_PARALLEL_END
 }
 
 bool
@@ -1117,14 +1117,7 @@ nest::ConnectionManager::get_connections(
 
   if ( source == 0 and target == 0 )
   {
-#ifdef _OPENMP
-#pragma omp parallel
-    {
-      thread t = kernel().vp_manager.get_thread_id();
-#else
-    for ( thread t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
-    {
-#endif
+    NEST_PARALLEL_FOR_THREAD(t)
       std::deque< ConnectionID > conns_in_thread;
 
       for ( index source_id = 1; source_id < connections_[ t ].size();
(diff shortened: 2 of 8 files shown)


terhorstd:
The code style checks fail, because the parallel sections are indented, but the `{}` are hidden in the macro. How do we want to handle this?

nestkernel/nest_types.h
```diff
@@ -32,6 +32,19 @@
 // Generated includes:
 #include "config.h"
 
+#ifdef _OPENMP
```
heplesser: - Please add a comment on the purpose of these macros, including an example for their use.
- I think it is strange that these macros are defined in `nest_types.h`, since they have nothing to do with types. Wouldn't `vp_manager.h` be a more logical placement?

topology/connection_creator_impl.h
```diff
@@ -223,11 +223,9 @@ ConnectionCreator::target_driven_connect_( Layer< D >& source,
   }
 
 // sharing specs on next line commented out because gcc 4.2 cannot handle them
-#pragma omp parallel // default(none) shared(source, target, masked_layer,
+// previously pragma  parallel // default(none) shared(source, target, masked_layer,
                      // target_begin, target_end)
```
heplesser: You can remove this comment.

examples/MyModule/step_pattern_builder.cpp
```diff
@@ -101,10 +101,7 @@ mynest::StepPatternBuilder::connect_()
 // This code is based on AllToAllBuilder, except that we step
 // by source_step_ and target_step_ except for stepping by 1.
 
-#pragma omp parallel
-  {
-    // get thread id
-    const int tid = nest::kernel().vp_manager.get_thread_id();
+  NEST_PARALLEL_FOR_THREAD(tid)
```
heplesser: I think that for the sense of code clarity, we should include the parallel section in explicit curly braces, even though they are superfluous. This will make the code more readable. Example:
```c++
NEST_PARALLEL_FOR_THREAD( thread_id )
{
  <code>
}
NEST_PARALLEL_END
```
I would also, as in the example here, suggest to use `thread_id` as variable name for the thread throughout.

jakobj: 👍 for extra curly braces, also to confuse basic text editors less. I'd prefer to stick with `tid` instead of `thread_id`, while the latter is clearer, the former is used way more throughout our codebase:
```bash
$ git grep " thread_id" . | wc -l
11
```

vs.

```bash
$ git grep " tid" . | wc -l
182
```

maybe `tid` -> `thread_id` should be left for another PR?

nestkernel/nest_types.h
```diff
@@ -32,6 +32,19 @@
...
+  #define NEST_PARALLEL_FOR_THREAD(t) \
+    _Pragma("omp parallel") \
+    { \
+      size_t t = kernel().vp_manager.get_thread_id();
```
jakobj: shouldn't the type be `thread`?
I guess you already tried making this `const`, but this probably breaks somewhere in `connect`. Are we sure this does not affect performance? Or am I misunderstanding how macros work?! Could we create two macros, one with and one without const?

nestkernel/connection_manager.cpp
```diff
@@ -1117,14 +1117,7 @@ nest::ConnectionManager::get_connections(
...
   if ( source == 0 and target == 0 )
   {
-#ifdef _OPENMP
-#pragma omp parallel
-    {
-      thread t = kernel().vp_manager.get_thread_id();
-#else
-    for ( thread t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
-    {
-#endif
```
jakobj: this for loop is replaced with `size_t t = 0;` in your macro. doesn't this break the logic? or was the for loop unnecessary anyway?

heplesser: @jakobj I have been wondering for some time why we ever put that loop into our code: If we do not have OpenMP, then we have no threads, and so everything is by default on thread 0. So it is perfectly fine to drop it.

nestkernel/nest_types.h
heplesser: @jakobj You are right, this should definitely be `const`: the thread number should not change inside a thread-parallel block. I believe the reason for declaring this `size_t` was that `get_num_threads()` below returns `size_t` and one then runs into signed/unsigned comparison warnings. In general, I would advocate moving away from NEST-specific renames of PODs, so I suggest we keep `size_t` here.

examples/MyModule/step_pattern_builder.cpp
heplesser: @jakobj OK, let's keep the PR minimal and keep variable names.

nestkernel/connection_manager.cpp
jakobj: Thanks for confirming that. Was also my hunch.

nestkernel/nest_types.h
jakobj: Yes, 100% agree, was just a matter of habit. It's quite annoying to keep in mind `thread` `index` etc.

heplesser:
@terhorstd @jakobj @jougs I think we can do this much simpler and then we do not even need the macro.  We currently have
```c++
#ifdef _OPENMP
#pragma omp parallel
    {
      index t = kernel().vp_manager.get_thread_id();
#else
    for ( index t = 0; t < kernel().vp_manager.get_num_threads(); ++t )
    {
#endif
    ...
    } // end for
   } // end parallel
```
Since we do not need the loop over the threads, we can just replace this by
```c++
#ifdef _OPENMP
#pragma omp parallel
#endif
    {
      const size_t t = kernel().vp_manager.get_thread_id();
      ...
    }  // end parallel
```
This keeps things a bit more explicit. It works because in the absence of OpenMP
- The block just becomes a plain block; this has the decided advantage that scoping rules will be identical with and without OpenMP.
- `get_thread_id()` returns 0, which is consistent with all other code.

nestkernel/connection_manager.cpp
jougs: The `for` loop can indeed go away. The reason it is there in the first place is that when we still supported PThreads, we had to prepare some thread-related data structures serially using the loop, or use OpenMP parallelism if that was enabled. No PThreads, no loop!

nestkernel/nest_types.h
jougs: I'd put them in `nest.h`. From there they are available all over NEST.

jougs:
@heplesser: I totally like your simplified explicit statement and think that we should use this instead of the macro. @terhorstd: can you please update the PR accordingly? Thanks!

heplesser:
@terhorstd @jakobj @jougs Actually, it is even simpler:
```c++
#pragma omp parallel
    {
      const size_t t = kernel().vp_manager.get_thread_id();
      ...
    }  // end parallel
```
will do! The C++ standard, §16.6, explicitly states that "[a]ny pragma that is not recognized by the implementation is ignored."

The standard further states that you can have whitespace left of `#`, so we can indent properly:
```c++
    #pragma omp parallel
    {
      const size_t t = kernel().vp_manager.get_thread_id();
      ...
    }  // end parallel
```

terhorstd:
As discussed in the [2017-06-26-Open-NEST-Developer-Video-Conference](https://github.com/nest/nest-simulator/wiki/2017-06-26-Open-NEST-Developer-Video-Conference) there are some other considerations in the parallelization. Other changes simplify it to below the threshold to justify introducing a macro.
Closing it as unnecessary.

Issue #302, opened 2016-04-11
Move or remove manualtests in testsuites


heplesser:
The main and the topology testsuites currently contain tests that need to be run manually. In practice, this means that they are hardly ever run and quickly become outdated, see comment in #221.

Those tests should preferably be converted to proper unit or regression tests, or, where that would be impossible, removed. Removal appears preferable to keeping dead code around.

jakobj:
one of the outdated manualtests seems to be "testsuite/manualtests/test_stdp_dopa.py", which does not seem to produce the desired behaviour:

```bash
expected weight at T=1000 ms: 28.6125 pA
weight at last event: 0.0 pA
```

heplesser:
Manual topology tests are now included in the testsuite per #705. Other manual test are addressed in the context of #348. I therefore close this issue.

PR #705, opened 2017-04-07
Topology tests into installcheck

hakonsbm:
This PR

- Converts manual tests in topology to MPI and unittests to be used with installcheck.
- Converts all SLI topology connection tests to topology unittests with reference results in each file.
- Changes the installcheck script to include the topology SLI and Python tests.

This addresses parts of #302.


heplesser:
@jougs Could you mainly check the technical side, i.e., installation and running of tests?
@jhnnsnk Could you mainly look at the tests themselves? They have all been around for a long time, but just as manual tests.

Don't be scared by the seemingly very large number of changed files. This is mostly because files have been moved and reference data that used to be in three data files per test are now included in the actual test files. So most of the file changes are just deletions.

heplesser:
@hakonsbm Could you update this PR to resolve the conflict with master?

hakonsbm:
@heplesser Updated.

hakonsbm:
@jhnnsnk Thank you for your review. I have fixed some of the inconsistencies, but others are not that strictly enforced in the testsuite. Addressing your points

- The purpose of this PR is to make the existing topology tests available with installcheck. But I have created an issue, #772, on creating new tests to cover the functionality mentioned.
- While it is good practice to not leave arguments on the stack, this is not strictly enforced in many tests in the testsuite.
- I have added `ResetKernel` to tests lacking this. Leaving namespace may also be good practice, but is not strictly enforced in the testsuite.
- The change of resolution is removed from tests that do not need it.

jhnnsnk:
Thanks for addressing my points.
:+1:

heplesser:
Release note: Added tests for topology module to testsuite.

PR #706, opened 2017-04-12
Minor refactoring in Time as discussed in PR #685

jakobj:
This PR implements the changes proposed during review of #685.
I suggest @heplesser @apeyser or @gtrensch as reviewers

### DIFF
--- a/nestkernel/nest_time.cpp
+++ b/nestkernel/nest_time.cpp
@@ -176,8 +176,7 @@ Time::fromstamp( Time::ms_stamp t )
   // intended ones.
   tic_t n = static_cast< tic_t >( t.t * Range::TICS_PER_MS );
   n -= ( n % Range::TICS_PER_STEP );
-  long s = n * Range::TICS_PER_STEP_INV;
-  double ms = s * Range::MS_PER_STEP;
+  const double ms = n * Range::TICS_PER_STEP_INV * Range::MS_PER_STEP;
   if ( ms < t.t )
   {
     n += Range::TICS_PER_STEP;
@@ -203,11 +202,11 @@ Time::reset_to_defaults()
 
 std::ostream& operator<<( std::ostream& strm, const Time& t )
 {
-  if ( t.tics == Time::LIM_NEG_INF.tics )
+  if ( t.is_neg_inf() )
   {
     strm << "-INF";
   }
-  else if ( t.tics == Time::LIM_POS_INF.tics )
+  else if ( t.is_pos_inf() )
   {
     strm << "+INF";
   }
--- a/nestkernel/nest_time.h
+++ b/nestkernel/nest_time.h
@@ -388,7 +388,22 @@ public:
   bool
   is_neg_inf() const
   {
-    return tics == LIM_NEG_INF.tics;
+    return tics <= LIM_NEG_INF.tics; // currently tics can never
+                                     // become smaller than
+                                     // LIM_NEG_INF.tics. however if
+                                     // LIM_NEG_INF.tics represent
+                                     // negative infinity, any smaller
+                                     // value cannot be larger and
+                                     // thus must be infinity as
+                                     // well. to be on the safe side
+                                     // we use less-or-equal instead
+                                     // of just equal.
+  }
+
+  bool
+  is_pos_inf() const
+  {
+    return tics >= LIM_POS_INF.tics; // see comment for is_neg_inf()
   }
 
   bool
@@ -493,11 +508,11 @@ public:
   double
   get_ms() const
   {
-    if ( tics == LIM_POS_INF.tics )
+    if ( is_pos_inf() )
     {
       return LIM_POS_INF_ms;
     }
-    if ( tics == LIM_NEG_INF.tics )
+    if ( is_neg_inf() )
     {
       return LIM_NEG_INF_ms;
     }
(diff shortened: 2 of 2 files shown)


nestkernel/nest_time.h
```diff
@@ -492,11 +492,11 @@ class Time
   double
   get_ms() const
   {
-    if ( tics == LIM_POS_INF.tics )
+    if ( tics >= LIM_POS_INF.tics )
```
heplesser: I realize this change is based on a suggestion I made, but I think we can do better here. I checked the `Time` class once more and we have checks in place there that ensure that `LIM_NEG_INF.tics <= tics <= LIM_POS_INF.tics` at all times. We also have a `is_neg_inf()` method; adding a corresponding `is_pos_inf()` method, we can then turn this into
```c++
if ( is_pos_inf() )
{
  return LIM_POS_INF.tics;
}
if ( is_neg_inf() )
{
  return LIM_NEG_INF.tics;
}
```
I think this would be even more readable.

nestkernel/nest_time.h
```diff
@@ -506,11 +506,11 @@ class Time
   delay
   get_steps() const
   {
-    if ( tics == LIM_POS_INF.tics )
+    if ( tics >= LIM_POS_INF.tics )
```
heplesser: See comment above.

nestkernel/nest_time.h
gtrensch: I agree with @heplesser.

In my opinion this calls for a macro and/or a function (e.g. `IS_POSINFINITY()` `preserve_infinity()`). But this is beyond the scope of this PR and would require additional changes to the time class to handle the different infinity value checks in a more generic way.

jakobj:
Thanks @heplesser for the suggestion. I implemented that now.

nestkernel/nest_time.cpp
```diff
@@ -198,11 +197,11 @@ Time::reset_to_defaults()
 
 std::ostream& operator<<( std::ostream& strm, const Time& t )
 {
-  if ( t.tics == Time::LIM_NEG_INF.tics )
+  if ( t.tics <= Time::LIM_NEG_INF.tics )
```
heplesser: @jakobj I know this is getting a bit nit-picky, but since we are at it, I think we should beautify this to
```c++
if ( t.is_neg_inf() )
```
and the same below for the positive infinity case.

jakobj:
@heplesser just fixed the last detail ;)

heplesser:
@jakobj This looks fine, but I am a bit confused by the fact that there is only a single commit visible in Github for this PR, dated 12 April, even though you have made changes during the review. Did you amend that commit or squash commits?

nestkernel/nest_time.h
```diff
@@ -387,7 +387,13 @@ class Time
   bool
   is_neg_inf() const
   {
-    return tics == LIM_NEG_INF.tics;
+    return tics <= LIM_NEG_INF.tics;
```
apeyser: Is this supposed to be called after computations that stay in range, but go beyond the limit? I ask to check -- why should tics ever be legally < LIM_NEG_INF or > LIM_POS_INF?

heplesser: @apeyser As far as I can see from the `nest_time.h` code, `tics` can never become smaller than `LIM_NEG_INF.tics`; the only way for this to happen would be a `Time` method calling the protected `Time(tic_t)` constructor. I still had suggested the change above from the perspective of defensive programming: if `LIM_NEG_INF.tics` represent negative infinity, any smaller value cannot be larger and thus must be infinity as well.

On the other hand, one could argue that `LIM_NEG_INF.tics` is a "magic value" (just as Inf for doubles is represented by a specific bit pattern), and that we should match that exactly.

jakobj:
@heplesser yes, I squashed all commits since the form one logical unit. this is why only one commit appears.

nestkernel/nest_time.h
apeyser: @heplesser so, the choices are: leave as is, replace the inequality portion with an assertion, remove the inequality.

heplesser: @apeyser Correct, where 
 - "leave as is" means `<=`, 
- "remove inequality" means `==`,
- and "replace the inequality portion" means `assert(tics >= ...); return ... == ...;`
I would want to avoid the assertion for performance reasons, so maybe we should keep the code as it is currently in master, where it is actually `==` ( @jakobj Sorry for the back-and-forth!).

heplesser:
I think merging #685 introduced some conflicts.

nestkernel/nest_time.h
apeyser: @heplesser 
The assertion should be better performing than leave as is, and should only be worse than remove inequality when the assertion is activated with debug flags. Thus, it would be the best of both worlds.

apeyser: @jakobj I'd remove the inequality and comment, but if you prefer to leave the inequality and comment, that's fine with me.

jakobj: I'd prefer the inequality. I will add a condensed version of this as a comment.

jakobj:
Fixed the conflicts and added a comment on the use of `<=` over `==`.

jougs:
@apeyser: are the changes OK with you now? If yes, please click the 'approve' button. Thanks!

PR #709, opened 2017-04-12
Added time stamp to secondary events

### LINKED ISSUE #707: weight_recorder does not work with connection types which use secondary events
Recording the weights of a connection type that uses secondary events gives an empty list

Example script:
```
import nest
import pylab as plt
import numpy as np

n_in = nest.Create('hh_psc_alpha_gap', 1, params={'I_e': 200.})
n_out = nest.Create('hh_psc_alpha_gap', 1)
wr = nest.Create('weight_recorder')

nest.SetDefaults("gap_junction", {"weight_recorder": wr[0]})

nest.Connect(n_in, n_out, {"rule": "one_to_one", "make_symmetric": True}, {"model": "gap_junction", "weight": 1.})

nest.Simulate(100.)

print nest.GetStatus(wr, "events")[0]["weights"]
```

@weidel-p already found the reason: 

> The reason for this behavior is that the timestamp of `SecondaryEvent`s is not set and therefore always 0 ms. As the `weight_recorder` is not active at a simulation time of 0ms, the event is not recorded (see `models/weight_recorder.cpp`).

janhahne:
This fixes #707. I suggest @jakobj and @weidel-p as reviewers.

### DIFF
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -587,6 +587,11 @@ EventDeliveryManager::deliver_events( thread t )
         kernel().model_manager.get_secondary_event_prototype( synid, t )
           << readpos;
 
+        // set time stamp (used by weight_recorder)
+        kernel()
+          .model_manager.get_secondary_event_prototype( synid, t )
+          .set_stamp( kernel().simulation_manager.get_clock() );
+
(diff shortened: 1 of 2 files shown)


weidel-p:
@janhahne thank you for fixing this issue. 

This PR is very short, contains a regressiontest and looks fine to me. :+1:

PR #711, opened 2017-04-12
Fixed bug in delivery of secondary events

### LINKED ISSUE #708: CopyModel does not work with connection types which use secondary events
The CopyModel functionality does not work properly with connection types which use secondary events.

Example script:
```
import nest
import pylab as plt
import numpy as np

n_in = nest.Create('hh_psc_alpha_gap', 1, params={'I_e': 200.})
n_out = nest.Create('hh_psc_alpha_gap', 2)
vm = nest.Create('voltmeter')

nest.CopyModel("gap_junction", "syn0")
nest.CopyModel("gap_junction", "syn1")
nest.SetStatus(vm, {'interval': 0.1})

nest.Connect(n_in, [n_out[0]], {"rule": "one_to_one", "make_symmetric": True}, {"model": "syn0", "weight": 10.})
nest.Connect(n_in, [n_out[1]], {"rule": "one_to_one", "make_symmetric": True}, {"model": "syn1", "weight": 10.})
nest.Connect(vm, n_in + n_out)

nest.Simulate(100.)

data = nest.GetStatus(vm)
senders = data[0]['events']['senders']
V = data[0]['events']['V_m']
us = np.unique(senders)
for i, s in enumerate(us):
    plt.plot(V[np.where(senders == s)], label=i)
plt.legend()
plt.show()
```

The neuron with is connected with `syn1` does not receive the input from the `gap_junction` and stays at its resting potential.

@jakobj already found the reason:

> The origin of this behaviour seems to be the break statement at in https://github.com/INM-6/nest-rate-connection/blob/master/nestkernel/connector_base.h#L1142. 

The `break` statement was accidentally not removed with the original PR #175 which should already have fixed this issue.

janhahne:
This fixes #708. The `break` statement was accidentally not removed with the original PR #175 which should already have fixed this issue.

Originally there was a one-to-one mapping between `syn_id` and `SecondaryEvent` classes. Meaning only `gap_junction`s used `GapJunctionEvent`s. The property `get_syn_id()` was replaced by `supports_syn_id` with #175 to allow the use of `CopyModel` and due to the introduction of labeled synapses, as both `gap_junction` and `gap_junction_lbl` of course use the `GapJunctionEvent`. 

The here removed `break` statement was (in the days of the one-to-one mapping) used to improve the performance and then by mistake not removed when #175 was introduced.

@heplesser Do you remember the original PR? If yes, could you have a quick look?

### DIFF
--- a/nestkernel/connector_base.h
+++ b/nestkernel/connector_base.h
@@ -1212,13 +1212,12 @@ public:
     const std::vector< ConnectorModel* >& cm )
   {
     // for all secondary connections delegate send to the matching homogeneous
-    // connector only
+    // connectors only
     for ( size_t i = primary_end_; i < size(); i++ )
     {
       if ( e.supports_syn_id( at( i )->get_syn_id() ) )
       {
         at( i )->send( e, t, cm );
-        break;
       }
     }
   }
--- a/testsuite/regressiontests/issue-708.sli
+++ b/testsuite/regressiontests/issue-708.sli
@@ -1,0 +1,94 @@
+/*
+ *  issue-708.sli
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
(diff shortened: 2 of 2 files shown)


jakobj:
Seems good, but could you quickly remind me why after this change, events are not delivered to more than one homogeneous connector? as far as I understand, after the copy gap events support multiple `syn_id`. so I would expect a gap event targeting neuron1 to also be delivered to connections targeting neuron2, precicely because we are only checking for "supports" not equality.

janhahne:
@jakobj With this change events are delivered to more than one (of course still only fitting) connector, that is the fix. Or am I understanding you wrong?

jakobj:
Hmm, I am bit confused. I guess my question is, where is it assured that events via `syn0` are only delivered to `neuron_out1` and from `syn1` only to `neuron_out2` if both events support the same `syn_id`s?

jakobj:
nevermind. I think I've got my mind untangled now. :) so 👍 from me

Issue #521, opened 2016-10-13
*_hpc connections return local id instead of global id of target neuron


tillschumann:
SLI example:
```
0 << /local_num_threads 8 >> SetStatus
/iaf_neuron 10 Create
5 6 /tsodyks2_synapse Connect
5 6 /tsodyks2_synapse_hpc Connect
<< >> GetConnections { dup GetStatus info } forall
```

Output:
```
## Name                     Type                Value
delay                    doubletype          1
receptor                 integertype         0
sizeof                   integertype         48
source                   integertype         5
synapse_model            literaltype         tsodyks2_synapse
target                   integertype         6                  <------------
tau_fac                  doubletype          0
tau_rec                  doubletype          800
U                        doubletype          0.5
u                        doubletype          0.5
weight                   doubletype          1
x                        doubletype          1
## Total number of entries: 12
```
```
## Name                     Type                Value
delay                    doubletype          1
receptor                 integertype         0
sizeof                   integertype         32
source                   integertype         5
synapse_model            literaltype         tsodyks2_synapse_hpc
target                   integertype         0                <------------
tau_fac                  doubletype          0
tau_rec                  doubletype          800
U                        doubletype          0.5
u                        doubletype          0.5
weight                   doubletype          1
x                        doubletype          1
```
I think it is misleading that on the one side you get the global neuron id and on the other side the local neuron id.

heplesser:
Fixed by #714, closing.

PR #714, opened 2017-04-24
testsuite::issue-521 - *_hpc connections return local id instead of g… 

tillschumann:
#521 
testsuite::issue-521 - *_hpc connections return local id instead of global id of target neuron

   *_hpc connections should only return thread_local_target instead of target
   to avoid confusion

Currently, it is not possible to get the target gid from a *_hpc connection.
To get it, we would need to pass the thread-id to the get_status function as an argument.
I think it is easier to leave out "target" as an return entry and return "thread_local_target" instead to avoid confusion.


nestkernel/nest_names.cpp
```diff
@@ -322,6 +322,7 @@ const Name t_ref_remaining( "t_ref_remaining" );
 const Name t_ref_tot( "t_ref_tot" );
 const Name t_spike( "t_spike" );
 const Name target( "target" );
+const Name thread_local_target( "thread_local_target" );
```
heplesser: Name definitions should be strictly alphabetical, please move this down.

nestkernel/nest_names.h
```diff
@@ -377,8 +377,9 @@ extern const Name
 extern const Name t_ref_tot; //!< Total refractory period, iaf_tum_2000
 extern const Name t_spike;   //!< Time of last spike
 extern const Name target;    //!< Connection parameters
-extern const Name target_thread; //!< Connection parameters
-extern const Name targets;       //!< Connection parameters
+extern const Name thread_local_target; //!< Connection parameters
```
heplesser: Please move this down so the order remains strictly alphabetical.

testsuite/regressiontests/issue-521.sli
```diff
@@ -0,0 +1,75 @@
...
+
+is_threaded_with_openmp not { exit_test_gracefully } if
+
+/test_targets
+{
+dup /threads Set
+ResetKernel
+0 << /local_num_threads threads >> SetStatus
+
+/iaf_neuron 20 Create
```
heplesser: `iaf_neuron` will be deprecated soon, please use `iaf_psc_alpha` or similar.

testsuite/regressiontests/issue-521.sli
```diff
@@ -0,0 +1,75 @@
...
+
+13 GetStatus /target_dict Set
+
+5 13 /static_synapse Connect
```
heplesser: Is there any special significance to using neurons 5 and 13? If so, it would be good to explain it briefly.

testsuite/regressiontests/issue-521.sli
```diff
@@ -0,0 +1,79 @@
...
+   FirstVersion: 04/2017
+   Author: Till Schumann
+ */
+
+(unittest) run
+/unittest using
+
+M_ERROR setverbosity
+
+is_threaded_with_openmp not { exit_test_gracefully } if
```
heplesser: Please replace this line with
```
skip_if_not_threaded
```
which has been added to master a few days ago. Travis currently fails without threading because we changed the behavior of `exit_test_gracefully` behind the scenes.

testsuite/regressiontests/issue-521.sli
```diff
@@ -0,0 +1,79 @@
...
+
+/test_targets
+{
+dup /threads Set
+ResetKernel
+0 << /local_num_threads threads >> SetStatus
+
+
+% create set of neurons
+/iaf_neuron 20 Create
```
heplesser: Please use `/iaf_psca_alpha`, `/iaf_neuron` is deprecated.

heplesser:
@kappeld This bug fix PR changes the signature of `ConnectorBase::get_synapse_status()`. Could you check if that will cause problems for your module and report back here?

heplesser:
Release nodes: `GetConnections` work correctly now also for `_hpc` synapses.

Issue #678, opened 2017-03-14
Problems generating help in system with multiple Python installations


heplesser:
André Sevenius reported the following problem on NEST User 14 March 2017:

I get the following error:

```
Extracting help information; this may take a litte while.
Traceback (most recent call last):
  File "/opt/rh/anaconda/lib/python3.6/site-packages/site.py", line 73, in <module>
    __boot()
  File "/opt/rh/anaconda/lib/python3.6/site-packages/site.py", line 3, in __boot
    import os
  File "/opt/rh/anaconda/lib/python3.6/os.py", line 409
    yield from walk(new_path, topdown, onerror, followlinks)
             ^
SyntaxError: invalid syntax
make[2]: *** [generate_help] Error 1
make[1]: *** [doc/CMakeFiles/generate_help.dir/all] Error 2
make: *** [all] Error 2
```

```
NEST Configuration Summary
--------------------------------------------------------------------------------
Build type          : 
Target System       : Linux
Cross Compiling     : FALSE
C compiler          : /bin/cc
C compiler flags    :  -O2 -Wall -fopenmp 
C++ compiler        : /bin/c++
C++ compiler flags  :  -O2 -Wall -fopenmp 
Build dynamic       : ON

Built-in modules    : models;precise;topology
User modules        : None
Python bindings     : Yes (Python 3.6: /opt/rh/anaconda/bin/python3.6m)
       Includes     : /opt/rh/anaconda/include/python3.6m
       Libraries    : /opt/rh/anaconda/lib/libpython3.6m.so

Cython bindings     : Yes (Cython 0.25.2: /opt/rh/anaconda/bin/cython)
Use threading       : Yes (OpenMP: -fopenmp)
Use GSL             : Yes (GSL 1.15)
    Includes        : /usr/include
    Libraries       : /usr/lib64/libgsl.so;/usr/lib64/libgslcblas.so

Use Readline        : Yes (GNU Readline 6.2)
    Includes        : /usr/include
    Libraries       : /lib64/libreadline.so;/lib64/libncurses.so

Use libltdl         : Yes (LTDL 2.4.2)
    Includes        : /usr/include
    Libraries       : /lib64/libltdl.so

Use doxygen         : Yes (/bin/doxygen)
                    : target `doc` available
Use MPI             : No
Use MUSIC           : No
Use libneurosim     : No
```

[...]

Issue #712, opened 2017-04-13
with-Python=OFF still leaves python required paths in Help Generation


kkapanova:
In order to compile nest without Python, 
I had to comment few lines in nest-2.12.0/doc/CMakeLists.txt and more precisely I removed the following part from the file:
```
if ( NOT CMAKE_CROSSCOMPILING )
  add_custom_target( generate_help ALL )
  # Extract help from all source files in the source code, put them in
  # doc/help and generate a local help index in the build directory containing 
  # links to the help files.
  add_custom_command( TARGET generate_help POST_BUILD
    COMMAND python -B generate_help.py "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}"
    COMMAND python -B generate_helpindex.py "${PROJECT_BINARY_DIR}/doc"
    WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/extras/help_generator"
    COMMENT "Extracting help information; this may take a litte while."
    )
  # Copy the local doc/help directory to the global installation
  # directory for documentation.
  install( DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/help"
    DESTINATION "${CMAKE_INSTALL_DOCDIR}"
    )
  # Update the global help index to contain all help files that are
  # located in the global installation directory for documentation.
  install( CODE
    "execute_process(
      COMMAND python -B generate_helpindex.py \"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DOCDIR}\"
      WORKING_DIRECTORY \"${PROJECT_SOURCE_DIR}/extras/help_generator\"
    )"
    )
endif ()
```
Probably there should and extra if Python statement

heplesser:
@kkapanova I assume that you encountered this problem on a system on which Python was not available at all? I am just trying to understand why this problem did not show up in our CI testing.

This is closely related to #678.

kkapanova:
@heplesser Yes you are correct. No Python was available on the system during the installation. Nevertheless, once this part of the code is removed, the compilation proceeds with no problems..

heplesser:
This will be handled as part of solving #678, therefore I close this issue.

heplesser:
@kkapanova Could you test if the fix proposed by @steffengraber in #715 works on your Python-free system? The easiest way to do this is probably to manually apply the change in #715 to your copy of the NEST sources and then test in a new build directory. Please report if it works or any problems in #715.

PR #715, opened 2017-05-05
Fix issue #678  - Skip help generation without python or python older than 2.7.8

steffengraber:
Getting python from CMake variables and using this instead of the hard coded python. Skipping help generation if no python is found or the version is older than 2.7.8.

This is a fix for issue #678 and #712

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -20,8 +20,15 @@
 if ( NOT CMAKE_CROSSCOMPILING )
   add_custom_target( generate_help ALL )
 
+    # Python is needed to generate the help. If Python does not exist, there are
+    # problems with the following.
+    # See https://github.com/nest/nest-simulator/issues/678.
     find_package( PythonInterp )
     if ( PYTHONINTERP_FOUND )
+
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -19,28 +19,36 @@
...
-  install( CODE
-    "execute_process(
-      COMMAND python -B generate_helpindex.py \"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DOCDIR}\"
-      WORKING_DIRECTORY \"${PROJECT_SOURCE_DIR}/extras/help_generator\"
-    )"
-    )
+
+    find_package( PythonInterp )
+    if ( PYTHONINTERP_FOUND )
+      if(${PYTHON_VERSION_STRING} VERSION_GREATER "2.7.7")
```
heplesser: I think it would be useful to add a comment explaining why we require a version > 2.7.8, including a reference to #678. Also, add spaces around the parenthesis `if ( ... )`.

andresni:
Confirmed fix for python 3.6 system.

PR #716, opened 2017-05-05
Ensure that ResetKernel resets off_grid_spiking_

suku248:
The default of the kernel parameter `off_grid_spiking_` is `false`. The parameter can be altered by the user and it is automatically set to `true` if a precise neuron model is created.

So far, `ResetKernel` did not result in a reset of `off_grid_spiking_` to `false`, which I find counterintuitive.

### DIFF
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -69,6 +69,8 @@ EventDeliveryManager::~EventDeliveryManager()
 void
 EventDeliveryManager::initialize()
 {
+  off_grid_spiking_ =
+    false; // ensures that ResetKernel resets off_grid_spiking_
   init_moduli();
   reset_timers_counters();
 }
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -69,8 +69,8 @@ EventDeliveryManager::~EventDeliveryManager()
 void
 EventDeliveryManager::initialize()
 {
-  off_grid_spiking_ =
-    false; // ensures that ResetKernel resets off_grid_spiking_
+  // ensures that ResetKernel resets off_grid_spiking_
+  off_grid_spiking_ = false;
   init_moduli();
   reset_timers_counters();
 }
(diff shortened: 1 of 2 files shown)


nestkernel/event_delivery_manager.cpp
```diff
@@ -69,6 +69,8 @@ EventDeliveryManager::~EventDeliveryManager()
 void
 EventDeliveryManager::initialize()
 {
+  off_grid_spiking_ =
+    false; // ensures that ResetKernel resets off_grid_spiking_
```
heplesser: I think it would be nicer to place the comment before the statement, so that the statement itself fits on a single line.

heplesser:
@suku248 Thanks for detecting this! But somehow this makes `test_spike_generator.sli` fail. That may be due to a weakness in the test.

I also noticed that initialization and reset of `EventDeliveryManager` appears to be implemented inconsistently. From a quick check, it appears that
- `secondary_event_buffer_` is not initialized in the constructor
- `secondary_event_buffer_`, `spike_register_`, `offgrid_spike_register_`, `displacements_` and `comm_marker_` are neither reset in `finalize()` or `initialize()`
- All of those except `comm_marker_` are cleared by `cofigure_spike_buffers()`, but that is called directly only from `reset_network()`, `prepare_simulation()` and `set_status()`.

Could you check that all is well and add a comment why we do not need to reset those members explicitly (or fix the code if we do ...)?

nestkernel/event_delivery_manager.cpp
suku248: Done!

suku248:
The problem with the test was that two of the sub-tests required `precise_times` of the `spike_detector` to be `true` such that `offsets` could be retrieved from the `spike_detector` after the simulation. However, the two sub-tests did neither explicitly set `precise_times` to `true` nor globally enable `off_grid_spiking`. The test worked before as another (earlier) sub-test enables `off_grid_spiking` and ResetKernel at the beginning of the two sub-tests did not reset `off_grid_spiking` to `false`.

I agree that initialization and reset of the `EventDeliveryManager` need some code review but not as part of this PR. I have opened issue #717.

heplesser:
@suku248 Thanks for fixing the test and for opening #717!

jakobj:
Looks good! 👍 from me

Issue #720, opened 2017-05-15
Updates recording of currents provided by stimulating devices


appukuttan-shailesh:
This is a continuation of the discussion towards the end of #663.

The following needs to be updated:
- Setting the recordables: GetDefaults('..._generator')['recordables']
- Appropriate updates to unit test to include above
- For noise generator, set current = 0, when no target neuron connected
- For noise generator, keep the current value constant for the duration of noise_generator's dt
(solutions for above discussed in #663)

One outstanding issue is possibly regarding the current generated by AC source. Consider the following:
ac = nest.Create('ac_generator', 1, params = {'amplitude': 550.0, 'offset': 1000.0, 'frequency': 100.0, 'phase' : 0.0, 'origin' : 2.5, 'start' : 2.5, 'stop' : 40.0})

The values that are computed are:
```
t (ms) <-> I (pA)
4.9    <-> 0.0
5.0    <-> 1034.53478574
5.1    <-> 1068.93327846
5.2    <-> 1103.05972302
```    
min_delay = 1.0
```
t (ms) <-> V (mV)
5.9    <-> -70.0
6.0    <-> -70.0
6.1    <-> -69.58756241
6.2    <-> -69.16415163
```
Shouldn't the value at t = 5.0, be I = 1000.00.... i.e. offset + sin(t=0). 
![ac_jump_v2](https://cloud.githubusercontent.com/assets/24866517/26061402/26583d08-3988-11e7-89c8-25f175749b97.png)
For Brian and NEURON, the current and voltage values are as follows:
(ignoring timestamps owing to differences in simulator execution)
```
Current:
0.0
1000.0
1034.53479
1068.93328
1103.05972

Voltage:
-70.0
-70.0
-69.90049751
-69.79805377
-69.69269695
-69.58446864
```
The absence of this value (1000.0) in NEST, I believe, causes a more abrupt change in both the current and voltage at the start of current injection as seen below, which is absent in the next cycle (starting at 15 ms). The change in Vm is more gradual in case of Brian and NEURON (as seen in the values above).

@heplesser , @gtrensch : Any thoughts on this? My current workaround for this is to compensate for the jump by adjusting the phase appropriately. But I believe there should be a more elegant fix, if this is infact considered an issue with the AC source.

appukuttan-shailesh:
@gtrensch : I have pulled the changes from your repository and updated L348 here:
https://github.com/appukuttan-shailesh/nest-simulator/tree/record-stim

[...]

PR #730, opened 2017-05-22
Fix ac_generator when used with start/stop (addresses #720)

heplesser:
Corrected ac_generator when used with start/stop, improved documentation and added test.

This addresses #720, although with a slightly different logic than apparently assumed by PyNN: In NEST, `start` and `stop` strictly do windowing and `origin` strictly moves this window. Thus, time is always counted from 0 when computing the sinusoid. This was incorrectly implemented until now and is now fixed.

Furthermore, for the parameters given in #720, the `ac_generator` now provides I=1000 pA at t=5 ms as expected. 

@appukuttan-shailesh Would you be willing to review this PR? Please use the "Start a review" feature and the "Review changes" button to submit your comments and verdict. I cannot add you to the "Reviewers" since you currently are not a member of NEST on Github.

### DIFF
--- a/models/ac_generator.cpp
+++ b/models/ac_generator.cpp
@@ -215,15 +215,16 @@ nest::ac_generator::update( Time const& origin, const long from, const long to )
   CurrentEvent ce;
   for ( long lag = from; lag < to; ++lag )
   {
-    S_.I_ = 0.0;
+    // We need to iterate the oscillator throughout all steps, even when the
+    // device is not active, since inactivity only windows the oscillator.
+    const double y_0 = S_.y_0_;
+    S_.y_0_ = V_.A_00_ * y_0 + V_.A_01_ * S_.y_1_;
+    S_.y_1_ = V_.A_10_ * y_0 + V_.A_11_ * S_.y_1_;
 
+    S_.I_ = 0.0;
     if ( device_.is_active( Time::step( start + lag ) ) )
     {
-      const double y_0 = S_.y_0_;
-      S_.y_0_ = V_.A_00_ * y_0 + V_.A_01_ * S_.y_1_;
-      S_.y_1_ = V_.A_10_ * y_0 + V_.A_11_ * S_.y_1_;
       S_.I_ = S_.y_1_ + P_.offset_;
-
       ce.set_current( S_.I_ );
       kernel().event_delivery_manager.send( *this, ce, lag );
     }
--- a/models/ac_generator.h
+++ b/models/ac_generator.h
@@ -37,16 +37,25 @@
    Name: ac_generator - provides AC input current
    Description:
 
-   This device produce an ac-current which are sent by a current event.
+   This device produce an ac-current which are sent by a CurrentEvent. The
+   current is given by
+
+           I(t) = offset + amplitude * sin ( om * t + phi )
+
+   where
+
+       om  = 2 * pi * frequency
+       phi = phase / 180 * pi
+
    The parameters are
+
    amplitude   double -  Amplitude of sine current in pA
    offset      double -  Constant amplitude offset in pA
-   phase       double -  Phase of sine current (0-360 deg)
    frequency   double -  Frequency in Hz
-   4) The
+   phase       double -  Phase of sine current (0-360 deg)
 
-   The currents are updated every time step by exact integration schemes from
-   [1]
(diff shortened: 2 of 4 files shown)


appukuttan-shailesh:
@heplesser : Sure, I will review this. About the membership, I saw that I'm listed here: [https://github.com/nest/nest-simulator/network/members](https://github.com/nest/nest-simulator/network/members). Do I need to do anything else?

appukuttan-shailesh:
@heplesser : I was running this [test code](https://github.com/appukuttan-shailesh/PyNN_testData/blob/master/nest_test_ac.py), and I find that the signal is phase shifted by 180:
ac_180phase
If I compensate the phase shift now by specifying: `'phase' : 180.0`, then it is as expected, with the currents being correctly evaluated as:
```
t (ms) <-> I (pA)
4.9    <-> 0.0
5.0    <-> 1000.00
5.1    <-> 1034.53
5.2    <-> 1068.93
5.3    <-> 1103.05
```

heplesser:
@appukuttan-shailesh The phase is as expected: at 100 Hz, the period is 10 ms, so at 5 ms the sine should cross zero in the negative direction. This looked different previously, because then the sine was not iterated from zero, but only while the device was active.

appukuttan-shailesh:
Ah ok, I now get what you meant by "_slightly different logic than apparently assumed by PyNN_ ...".

appukuttan-shailesh:
Ah ok, I now get what you meant by "_slightly different logic than apparently assumed by PyNN ..._". Out of curiosity... is there an advantage in implementing it like this?

On the downside, I would imagine that if a modeler wished to inject current in a model sometime during a run, this implementation would require him to _be aware_ of the phase at that particular time instant? Whereas if the sinusoid is computed from the time instant it is activated (the earlier implementation?), this would be more predictable.

gtrensch:
@heplesser I have a question: Does the parameter 'origin' should repeat the sequence within 'start' and 'stop' ?  If so, the phase needs to be shifted.

heplesser:
See #505 for some thoughts on this matter in general. Current NEST logic says that `start/stop/origin` provide *strictly* a windowing mechanism. As I point out in #505, this mechanism seems dated, and the transition to NEST 3 would provide an opportunity to replace it with something better. I would propose a two-pronged approach:

- Fixing the errors in the current implementation by merging this PR.
- Discussing in #505 the syntax and semantics of an improved approach. Especially for repeated stimulation, we need to think carefully about how to handle repeated stimulation. 

Just some things to think about: For a sine current one could (as one option) define a fixed phase at the beginning of each repetition and implement this easily. For Poisson and gamma processes etc one needs to handle the first interval properly, and other stimulators may raise yet more issues.

gtrensch:
@heplesser thanks, this sounds reasonable.
I approve this PR.
Note: To complete the story from #663 and continued in #720 another PR is required.

gtrensch:
@appukuttan-shailesh yes, if you pull from master then, you perhaps may get a conflict in ac_generator.cpp.

appukuttan-shailesh:
@heplesser : I was going through the commits again... I believe the changes we made for this issue were restricted to:
1) models/ac_generator.h
2) models/dc_generator.h
3) models/noise_generator.h
4) models/step_current_generator.h
5) pynest/nest/tests/test_current_recording_generators.py
6) models/noise_generator.cpp

So the only potential conflict might be with `models/ac_generator.h`. And here the changes in #730 seem restricted to the comments at the start of the file. So should go through without conflicts, I imagine. Should I proceed with the PR?

appukuttan-shailesh:
@gtrensch  : I was going through the commits again... I believe the changes we made for this issue were restricted to:
1) models/ac_generator.h
2) models/dc_generator.h
3) models/noise_generator.h
4) models/step_current_generator.h
5) pynest/nest/tests/test_current_recording_generators.py
6) models/noise_generator.cpp

So the only potential conflict might be with `models/ac_generator.h`. And here the changes in #730 seem restricted to the comments at the start of the file. So should go through without conflicts, I imagine. Should I proceed with the PR?

gtrensch:
Yes, you are right. It was noise_generator.cpp. Please proceed with your PR.
Many thanks !

heplesser:
Release note: ac_generator will now compute current as sin( om * t + phi ) with t counted from the beginning of the simulation even when using start/stop.

PR #731, opened 2017-05-22
Circumvent negative max_width in Travis log parser

gtrensch:
Bypass Travis issue: ValueError: invalid width -29 (must be > 0)

On Travis column_max_width sometimes returns a negative value causing the parser script to abend. The calculation is based on the terminal width which seems to be zero in that case. This pull requests adds two lines to set the value to 70 if it is negative.


heplesser:
Since this is a trivial but urgent fix to our build system, I merge based on a single review.

PR #733, opened 2017-05-25
Fixes #720 Issues related to current recording from devices

### LINKED ISSUE #720: Updates recording of currents provided by stimulating devices

appukuttan-shailesh:
Implements the following (as discussed in #720 and #663 ):

- Setting the recordables: GetDefaults('..._generator')['recordables']
- Appropriate updates to unit test to include above
- For noise generator, set current = 0, when no target neuron connected
- For noise generator, keep the current value constant for the duration of noise_generator's dt
- (solutions for above discussed in #663)

### DIFF
--- a/pynest/nest/tests/test_current_recording_generators.py
+++ b/pynest/nest/tests/test_current_recording_generators.py
@@ -88,33 +88,21 @@ class CurrentRecordingGeneratorTestCase(unittest.TestCase):
     def test_GetRecordables(self):
         """Check get recordables"""
 
-        try:
-            val = nest.GetDefaults('step_current_generator')['recordables'][0]
-            assert val == 'I', \
-                "Incorrect recordables ( %r ) for step current generator" % val
-        except:
-            raise
+        val = nest.GetDefaults('step_current_generator')['recordables'][0]            
+        self.assertEqual(val, 'I',
+                         'Incorrect recordables ({}) for step current generator'.format(val))
 
-        try:
-            val = nest.GetDefaults('ac_generator')['recordables'][0]
-            assert val == 'I', \
-                "Incorrect recordables ( %r ) for ac generator" % val
-        except:
-            raise
+        val = nest.GetDefaults('ac_generator')['recordables'][0]            
+        self.assertEqual(val, 'I',
+                         'Incorrect recordables ({}) for ac generator'.format(val))
 
-        try:
-            val = nest.GetDefaults('dc_generator')['recordables'][0]
-            assert val == 'I', \
-                "Incorrect recordables ( %r ) for dc generator" % val
-        except:
-            raise
+        val = nest.GetDefaults('dc_generator')['recordables'][0]            
+        self.assertEqual(val, 'I',
+                         'Incorrect recordables ({}) for dc generator'.format(val))
 
-        try:
-            val = nest.GetDefaults('noise_generator')['recordables'][0]
-            assert val == 'I', \
-                "Incorrect recordables ( %r ) for noise generator" % val
-        except:
-            raise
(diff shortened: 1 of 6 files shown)


pynest/nest/tests/test_current_recording_generators.py
```diff
@@ -85,6 +85,37 @@ def setUp(self):
                                          'stop': self.t_stop})
         nest.Connect(self.noise, self.neuron)
 
+    def test_GetRecordables(self):
+        """Check get recordables"""
+
+        try:
+            val = nest.GetDefaults('step_current_generator')['recordables'][0]
+            assert val == 'I', \
+                "Incorrect recordables ( %r ) for step current generator" % val
```
heplesser: Please use  
```
self.assertEqual(val, 'I', 'Incorrect recordables ({}) for step current generator'.format(val))
```
also for the tests below.

appukuttan-shailesh: Just to confirm, I would be replacing
``` python
try:
        val = nest.GetDefaults('step_current_generator')['recordables'][0]
        assert val == 'I', \
                "Incorrect recordables ( %r ) for step current generator" % val
except:
        raise
```
with
``` python
val = nest.GetDefaults('step_current_generator')['recordables'][0]
self.assertEqual(val, 'I', 'Incorrect recordables ({}) for step current generator'.format(val))
```
for each of the sub-tests

heplesser: Correct.

heplesser:
@appukuttan-shailesh Could you pep8-ify the test file? Some lines are too long and there is trailing whitespace.

appukuttan-shailesh:
Done.... looks good to go.

heplesser:
This merge fixes current recording from noise generator and adds the recordables entry to other generators that support current recording.

PR #734, opened 2017-05-26
Fix possible deadlock in neuron update loop in multithreaded code (fixes #718)

### LINKED ISSUE #718: NEST Deadlock if exception thrown in update
It Appears that the following situation leads to a deadlock in the v2.12.0 commit. When an exception is raised on any one thread and not on the other thread in the same simulation slice, the loop for that thread only is terminated by the following condition in the while loop. The rest of the threads then stall on the next barrier leading to a deadlock.

```c++

    } while ( to_do_ > 0 and not exit_on_user_signal_
      and not exceptions_raised.at( thrd ) );

```

On a related note, how feasible would it be to implement error handling across MPI Processes? (For example by sending an additional integer during the allgather to signal the state (Error or not) of the MPI Process and then conditionally (i.e. if any one process is in an error state)  gather the exception message and throw a consistent exception across processes?

heplesser:
This PR fixes a potential deadlock when a neuron throws an exception during multithreaded simulation (#718).  

Since it is difficult to test for deadlocks in unittests, there is no automated test. This manual test is inspired by @maharjun's [deadlock module](https://github.com/IGITUGraz/SpikeDetectorFuse/tree/NESTExceptionDeadlockTest). To test, add the following to the beginning of `iaf_psc_alpha::update()`:
```c++
 if ( this->get_thread() == 2 )
	  throw KernelException("Thread 2");
```
and run the following SLI script from a file
```
0 << /local_num_threads 4 >> SetStatus
/iaf_psc_alpha 10 Create ;
10 Simulate
```
NEST should hang if you use current master; with the correction given here, NEST should terminate with an error message.

@maharjun Would you test and review this fix?

### DIFF
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -644,6 +644,8 @@ nest::SimulationManager::update_()
 
   std::vector< lockPTR< WrappedThreadException > > exceptions_raised(
     kernel().vp_manager.get_num_threads() );
+  bool exception_raised = false; // none raised on any thread
+
 // parallel section begins
 #pragma omp parallel
   {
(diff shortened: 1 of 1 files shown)


maharjun:
I have tested it and can confirm that this works correctly.

heplesser:
Merging as tests pass and two reviewers confirmed that the deadlock no longer occurs.

Issue #722, opened 2017-05-16
NEST Name mechanism is not thread-safe under insertion


heplesser:
NEST uses a global `Name` table to accelerate name lookup and comparison in dictionaries. Inserting into this table in a thread-parallel context can corrupt the name table. This can happen if new names are introduced as strings in parallel context, in particular during `set_status()/get_status()` calls.

Measures:
1. Assert that `Name::insert()` is only executed in thread-serial context.
1. Replace all remaining strings in dictionary accesses by `Name` objects.

PR #736, opened 2017-05-29
Add tests for synapse parameter check (fixes #735)

### LINKED ISSUE #735: Incorrect Parameter Name String in conn_builder.cpp
As of the latest master commit `15b63f2` (master branch), the file `conn_builder.cpp` at line '340' appears to mistakenly contain a space after the `"` in the parameter string `" A_minus"`

heplesser:
This PR removes some code from `ConnBuilder` that was meant to raise an error if the user attempted to set stdp parameters on individual `stdp_dopamine_synapse`s that could not be set individually on those synapses. I removed the entire test because:

- the test was not used at all: if attempting to set such a parameter, e.g., `/A_minus`, the error was caught by the "unused parameter" test
- the test would not have worked at all on copies of `stdp_dopamine_synapse`, because it looked at the synapse name to decide whether the test should be applied
- the test was unsafe by implicitly introducing `Name` objects (see #722)

I added regression test to ensure that attempts to set model-only parameters on individual synapses raise errors.

@maharjun Could you check that this addresses your concerns in #735?

### DIFF
--- a/extras/static_code_analysis.sh
+++ b/extras/static_code_analysis.sh
@@ -252,7 +252,7 @@ if [ "x$c_files_with_errors" != "x" ] || [ "x$python_files_with_errors" != "x" ]
 else
   print_msg "" ""
   print_msg "MSGBLD0220: " "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
-  print_msg "MSGBLD0220: " "+               STATIC CODE ANALYSIS TERMINATED SUCESSFULLY !                 +"
+  print_msg "MSGBLD0220: " "+               STATIC CODE ANALYSIS TERMINATED SUCCESSFULLY !                +"
   print_msg "MSGBLD0220: " "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
   print_msg "" ""  
 fi
--- a/lib/sli/processes.sli
+++ b/lib/sli/processes.sli
@@ -703,7 +703,7 @@ Bugs: No error message is raised, if Flag=1 or Flag=2 and the given
       but it is raised in a child process of SLI, and -after- standard
       output has been redirected. The error message can be read from
       the "mouth"-pipe!
-       "spawn" always returns sucessful if Flag=1 or Flag=2. However,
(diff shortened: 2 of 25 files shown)


maharjun:
TBH I'm not really familiar with the way the connection builders work, the issue #735 was onlly noticed by me when I was inspecting the diffs, and this does indeed remove the offending piece of code. Regarding actual function, it would be better if someone else were to review that.

stinebuu:
I am a bit confused about what the removed test is supposed to do. I don't see the connection between 
the regression test and the removed test, as the removed test check to see if the parameters are in the syn_spec sent in with Connect and the regression test tests errors with SetStatus.

 Also, when testing with master on my computer, I get the "unused parameter" error if I do

```
/n /iaf_psc_alpha Create def
n n << /model /stdp_dopamine_synapse /A_plus 2 >> Connect 
```

This call never enters `ConnBuilder::check_synapse_params_`, so the removed test is never tested here, as you said.

If I do
```
/n /iaf_psc_alpha Create def
[n] [n] << /rule /all_to_all >> << /model /stdp_dopamine_synapse /A_plus 2 >> Connect
```

however, I get 
```
Connect_g_g_D_D [Error]: NotImplemented
    Connect doesn't support the setting of parameter A_plus in stdp_dopamine_synapse. Use SetDefaults() or CopyModel().
```

The same error occurs with `/one_to_one` rule, I did not check the other rules.

This disappears when the test is removed, of course. Are we not supposed to test to see if we are sending in these parameters in syn_spec for the `stdp_dopamine_synapse` anymore? Even though it only appears when we send in a rule?

heplesser:
@stinebuu Thank you for your detective work. Clearly I need to extend the regression test to include the case with the "long" form of the `Connect` call. I am surprised the two forms make such a difference.

heplesser:
@stinebuu The reason that `Connect` in the long form triggered the error message was that `check_synapse_parameters_()` is called when you let `Connect` set synapse parameters (from the `ConnBuilder` constructor). I now put the test back into `check_synapse_parameters_()` and extended the regression test. Unfortunately, that fails. To fix this, we need to fix #737 first, so that the test will be done correctly even if we use a synapse model created with `CopyModel`.

heplesser:
NB: Most of the work was done by #739. This PR just adds an additional test.

PR #739, opened 2017-05-31
Moved check_synapse_params to specific synapse classes 

### LINKED ISSUE #737: Review `ConnBuilder::check_synapse_params_`
This is a follow-up to #735 and #736. `ConnBuilder::check_synapse_params_()` tests whether certain synapse parameters for certain synapse types. There are several problems with this code
- string passed by value, not reference;
- slow comparison with strings instead of Name objects;
- testing based on name of base model: tests will not be applied to copied models
- not sure if there is a unit- or regressiontest for all the tests performed here.

stinebuu:
This PR fixes #737 .

It takes each test in `ConnBuilder::check_synapse_params()` and moves it to a `check_synapse_params()` function in the respective synapse class. To make sure we don't have to add it to all the synapse classes, the `GenericConnectionModel` handles the allocation.

A regressiontest is also added, which make sure all tests fail if we send in the specific parameters in the syn_spec dictionary.


heplesser:
@stinebuu All subtest in `issue-737.sli` that use threading should go into a separate file and be skipped if threading is not available.

heplesser:
@kappeld Could you review this pull request? Since it adds the method `check_synapse_params()` to classes `ConnectorModel` and `Connection`, it may have an effect on your non-generic connector model. I particular, I think your connector model will have to implement the method.

models/cont_delay_connection.h
```diff
@@ -141,6 +141,12 @@ class ContDelayConnection : public Connection< targetidentifierT >
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
+   * Check syn_spec dictionary for parameters that are not allowed for this
+   * connection. Will issue warning or throw error if a parameter is found.
```
heplesser: - this general explanation should be in the base class only (`Connection`)
- here, you could just mention what is illegal for this particular synapse model
- Doxygen comments should begin with a single-line summary followed by details
- "an illegal parameter is found"
- add info that the method does nothing otherwise

models/quantal_stp_connection.h
```diff
@@ -127,6 +127,12 @@ class Quantal_StpConnection : public Connection< targetidentifierT >
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
+   * Check syn_spec dictionary for parameters that are not allowed for this
+   * connection. Will issue warning or throw error if a parameter is found.
```
heplesser: See above.

models/static_connection_hom_w.h
```diff
@@ -143,6 +143,21 @@ class StaticConnectionHomW : public Connection< targetidentifierT >
   }
 
   /**
+   * Check syn_spec dictionary for parameters that are not allowed for this
+   * connection. Will issue warning or throw error if a parameter is found.
+   */
```
heplesser: See above.

models/stdp_dopa_connection.h
```diff
@@ -197,6 +197,12 @@ class STDPDopaConnection : public Connection< targetidentifierT >
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
+   * Check syn_spec dictionary for parameters that are not allowed for this
+   * connection. Will issue warning or throw error if a parameter is found.
+   */
```
heplesser: See above.

nestkernel/conn_builder.cpp
```diff
@@ -83,15 +83,14 @@ nest::ConnBuilder::ConnBuilder( const GIDCollection& sources,
...
-  if ( syn_name != "static_synapse" )
-  {
-    check_synapse_params_( syn_name, syn_spec );
-  }
-
   synapse_model_ = kernel().model_manager.get_synapsedict()->lookup( syn_name );
 
+  // We need to make sure that Connect can process all synapse parameters
+  // specified.
+  ConnectorModel& syn_mod =
```
heplesser: This could be `const`, and would suggest the following renaming:
- `synapse_model_` -> `synapse_model_id_` (because it just is the id number)
- `syn_mod`-> `synapse_model`

nestkernel/connection.h
```diff
@@ -150,6 +150,12 @@ class Connection
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
+   * Check syn_spec dictionary for parameters that are not allowed for this
+   * connection. Will issue warning or throw error if a parameter is found.
+   */
```
heplesser: See above, the base class doc should be here. Mention that the base class implementation does nothing and that classes requiring checks need to override with their own method.

nestkernel/connector_model.h
```diff
@@ -150,6 +150,8 @@ class ConnectorModel
 
   virtual const CommonSynapseProperties& get_common_properties() const = 0;
 
+  virtual void check_synapse_params( const DictionaryDatum& ) const = 0;
```
heplesser: Add documentation.

testsuite/regressiontests/issue-737-threads.sli
```diff
@@ -0,0 +1,96 @@
...
+FirstVersion: May 2017
+SeeAlso: testsuite::issue-735, testsuite::issue-737
+*/
+
+(unittest) run
+/unittest using
+
+skip_if_not_threaded
+
+% Test multithreded c in syn_spec for stdp_dopamine_synapse model
```
heplesser: "a" missing in "multithreaded"; I think that is in other places, too.

kappeld:
At a first glance the changes look good to me, but they will effect our custom connector model as @heplesser suggested. I will try this out and review the changes more carefully as soon as possible.

models/cont_delay_connection.h
```diff
@@ -141,8 +141,7 @@ class ContDelayConnection : public Connection< targetidentifierT >
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
-   * Check syn_spec dictionary for parameters that are not allowed for this
-   * connection. Will issue warning or throw error if a parameter is found.
+   * Checks to see if delay is given in syn_spec.
```
heplesser: This is a bit confusing. I would rather write `Throws exception if delays is given`. Otherwise, the reader may think that the check will throw and exception if the weight is missing.

models/quantal_stp_connection.h
```diff
@@ -127,8 +127,7 @@ class Quantal_StpConnection : public Connection< targetidentifierT >
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
-   * Check syn_spec dictionary for parameters that are not allowed for this
-   * connection. Will issue warning or throw error if a parameter is found.
+   * Checks to see if n or a are given in syn_spec.
```
heplesser: Also here change the wording so that it is clear that n and a are "prohibited".

models/stdp_dopa_connection.h
```diff
@@ -197,8 +197,11 @@ class STDPDopaConnection : public Connection< targetidentifierT >
   void set_status( const DictionaryDatum& d, ConnectorModel& cm );
 
   /**
-   * Check syn_spec dictionary for parameters that are not allowed for this
-   * connection. Will issue warning or throw error if a parameter is found.
+   * Checks to see if illegal parameters are given in syn_spec.
+   *
+   * The illegal parameters are: vt, A_minus, A_plus, Wmax, Wmin, b, tau_c,
+   * tau_n, tau_plus, c and n. The last two only apply if we have more than one
```
heplesser: "only apply" -> "are prohibited only"

nestkernel/connection.h
```diff
@@ -150,6 +150,18 @@ class Connection
...
+   * Check syn_spec dictionary for parameters that are not allowed with the
+   * given connection.
+   *
+   * Will issue warning or throw error if an illegal parameter is found. The
+   * method does nothing if no illegal parameter is found.
+   *
+   * @note Classes requiring checks need to override the function with their own
+   * implementation, as this base class implementation does not do anything.
+   */
+  void check_synapse_params( const DictionaryDatum& d ) const;
```
stinebuu: @jakobj As far as I understand, the connection class is not supposed to have any virtual functions, because we then store pointers to the subclasses so we can use the correct overridden function.  By making this into an empty function, we avoid this memory allocation.

@heplesser does this sound about right?

heplesser: Yes, the technical term is *vtable*: For the 4G kernel we worked hard to avoid virtual functions in `Connection` objects so that these objects do not contain a vtable-pointer, which would cost 8B per synapse; see Kunkel et al (2014) for details.

heplesser:
Before merging, wait for feedback from @kappeld concerning the effect on module implementing its own `ConnectorModel`.

nestkernel/connection.h
jakobj: Ah alright, should've known that. ;) thanks for reminding me.

kappeld:
On a related note: Since the API of NEST has changed now quite a bit since the 2.12.0 release (with this and other changes), @heplesser are you planning to issue a 2.12.1 release soon? We would like to keep the version numbers of our custom module in sync with NEST's version number. Since the changes made to NEST also require a slightly updated version of our module to be compatible, making a new minor release might be more transparent for the users.

heplesser:
@kappeld Thanks for checking! Yes, we are considering a 2.12.1 release since there have been a number of bug fixes (and a few more under review).

heplesser:
Release Notes: Refactored `check_synapse_param()` to better object-oriented design. `Connection` subclasses can now define their own checks.

PR #745, opened 2017-06-06
Rate neuron framework

janhahne:
This PR adds rate models to NEST as described in 

 Hahne, J., Dahmen, D., Schuecker, J., Frommer, A.,
 Bolten, M., Helias, M. and Diesmann, M. (2017).
 Integration of Continuous-Time Dynamics in a
 Spiking Neural Network Simulator.
 Front. Neuroinform. 11:34. doi: 10.3389/fninf.2017.00034

Overview:

- added rate model base classes for rate neurons with input noise (`rate_neuron_ipn`) and outout noise (`rate_neuron_opn`)
- added rate connections (`rate_connection` for connections without delay, `delay_rate_connection` for connections with delay and `diffussion_connection` for connection between siegert neurons)
- added several template based rate neuron models (`lin_rate`, `tanh_rate`, `thresholdlin_rate`)
- added `siegert_neuron` model
- added `SecondaryEvent`s for different rate connection
- adjusted some existing unittests to work with or exclude rate neurons
- added sli and python unittests (functionality, mpi, connections) for the new framework 
- added two python examples

We suggest @heplesser as a reviewer


testsuite/unittests/test_rate_connections.sli
```diff
@@ -0,0 +1,250 @@
...
+    % check that legal connections can be created
+    {
+      ResetKernel
+      model_supported Create /neuron_supported Set
+      [neuron_supported] [neuron_supported]
+      << /rule /one_to_one >> 
+      << /model /diffusion_connection >> 
+      Connect
+    }
+    pass_or_die
```
mschmidt87: Same as above: move out of inner loop.

testsuite/unittests/test_rate_connections.sli
```diff
@@ -0,0 +1,250 @@
...
+  /thresholdlin_rate_ipn
+  /thresholdlin_rate_opn
+  /tanh_rate_ipn
+  /tanh_rate_opn ]
+def
+
+/not_supported
+[ /hh_psc_alpha_gap
+  /siegert_neuron
+  /iaf_psc_neuron]
```
mschmidt87: `iaf_psc_neuron` is not a known model name. The test runs through but I suspect, that the part where you use /iaf_psc_neuron actually fails with `UnknownModelName` instead of throwing an error because of the synapse. 
I assume you just want to test one representative spiking neuron here, so I suggest changing this to 'iaf_psc_delta'.
The same for the other parts below.

models/siegert_neuron.h
```diff
@@ -0,0 +1,327 @@
...
+/* BeginDocumentation
+Name: siegert_neuron
+
+Description:
+
+ siegert_neuron is an implementation of a rate model with the
+ non-linearity given by the gain function of the
+ leaky-integrate-and-fire neuron with delta or exponentially decaying
+ synapses. The model can be used for a mean-field analysis of spiking
+ networks.
```
mschmidt87: Perhaps, a reference to a paper with the actual equations would be useful here?
For instance:
Fourcaud, Nicolas and Brunel, Nicolas, "Dynamics of the firing probability of noisy integrate-and-fire neurons", 14 (2002), pp. 2057--2110.

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+double
+siegert( double tau_m,
+  double tau_syn,
+  double tau_r,
+  double theta,
+  double V_r,
+  double mu,
+  double sigma )
+{
+  double alpha = 2.0652531522312172;
```
mschmidt87: I assume this is the evaluation of the Riemann zeta function. Can you add a line comment explaining this briefly?

pynest/nest/tests/test_rate_neuron.py
```diff
@@ -0,0 +1,123 @@
...
+        self.tstart = 10. * self.neuron_params['tau']
+
+        nest.set_verbosity('M_WARNING')
+        nest.ResetKernel()
+        nest.SetKernelStatus(
+            {'resolution': self.dt, 'use_wfr': False, 'print_time': True})
+
+        # set up rate neuron and devices
+        self.rate_neuron = nest.Create(
+            'lin_rate_ipn', params=self.neuron_params)
```
mschmidt87: This test only includes the input noise neuron. It would be straightforward to extend this to test the output noise neuron as well. Wouldn't this be a good idea?

pynest/nest/tests/test_rate_neuron_communication.py
```diff
@@ -0,0 +1,181 @@
...
+import numpy as np
+
+
+def H(x):
+    return 0.5 * (np.sign(x) + 1.)
+
+
+@nest.check_stack
+class RateNeuronCommunicationTestCase(unittest.TestCase):
+    """Check rate_neuron"""
```
mschmidt87: You could extend this documentation by giving some more details, such as:
This test checks the correct function of rate neuron with input noise and three transfer functions. It also checks whether delay and weight of the rate connections work correctly.

models/lin_rate.h
```diff
@@ -0,0 +1,120 @@
...
+ The model supports connections to other rate models with either zero or
+ non-zero delay, and uses the secondary_event concept introduced with
+ the gap-junction framework.
+
+Parameters:
+
+ The following parameters can be set in the status dictionary.
+
+ rate                double - Rate (unitless)
+ tau                 double - Time constant in ms.
```
mschmidt87: Can you explain the influence of the parameter tau in the model description above? The same for the other transfer functions, tanh and threshold_linear.

janhahne:
@mschmidt87 Thank you for the first review and the valuable suggestions. So far I fixed the typos and the sli unittest. We will have a closer look on the other issues soon.

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+  for ( long lag = from; lag < to; ++lag )
+  {
+    // register rate in buffer
+    new_rates[ lag ] = S_.r_;
+
+    // propagate rate to new time step (exponential integration)
+    double drive = siegert( P_.tau_m_ * 1e-3,
+      P_.tau_syn_ * 1e-3,
+      P_.t_ref_ * 1e-3,
```
heplesser: Why do you need to scale the time values here?

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+      P_.theta_,
+      P_.V_reset_,
+      B_.drift_input_[ lag ],
+      std::sqrt( B_.diffusion_input_[ lag ] ) );
```
heplesser: Why the `std::sqrt()` in the call here? Couldn't you just pass the variance?

models/tanh_rate.h
```diff
@@ -0,0 +1,117 @@
...
+namespace nest
+{
+/* BeginDocumentation
+Name: tanh_rate - rate model with hyperbolic tangent non-linearity
+
+Description:
+
+ tanh_rate is an implementation of a non-linear rate model with either
+ input (tanh_rate_ipn) or output noise (tanh_rate_opn) and gain function
+ Phi(h) = tanh(g * h) and Psi(h) = h for linear_summation = True
```
heplesser: Why can the user only choose the gain, but not the location of the inflection point (tanh(gh-x))?.

nestkernel/event.h
```diff
@@ -247,6 +250,16 @@ class Event
   void set_weight( weight t );
 
   /**
+   * Set drift_factor of the event (see DiffusionEvent).
+   */
+  virtual void set_drift_factor( weight t ){};
```
heplesser: Should we use `drift_weight` and `diffusion_weight` as names, maybe? 
@terhorstd Discuss in VC.

nestkernel/event.h
```diff
@@ -1062,6 +1075,450 @@ class GapJunctionEvent : public SecondaryEvent
   double get_coeffvalue( std::vector< unsigned int >::iterator& pos );
 };
 
+/**
+ * Event for rate model connections without delay.
+ */
+class RateNeuronEvent : public SecondaryEvent
```
heplesser: How much of the code here could be moved to `SecondaryEvent`? There seems quite a bit of overlap with other of these classes.

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+      // update last_y_values for next wfr_update iteration
+      B_.last_y_values[ lag ] = S_.r_;
+    }
+  }
+
+  if ( not wfr_update )
+  {
+    // clear last_y_values
+    B_.last_y_values.clear();
+    B_.last_y_values.resize( quantity, 0.0 );
```
janhahne: Your suggestion gives an compiler error:

> error: invalid initialization of non-const reference of type ‘std::vector<double>&’ from an rvalue of type ‘std::vector<double>’

which make sense since the vector we are swapping with is only a temporary object. Is there another way to use swap here oder should we just stick with our code?

nestkernel/event.h
janhahne: @heplesser Yes, there is actually a lot of overlap and I am not particular happy about that. I already thought about how to reduce this overlap, but this is not that easy. Each class contains important individual static member variables, which are used in almost every function. 

Let me explain in more detail: `GapJunctionEvent`,`RateNeuronEvent`,`DelayRateNeuronEvent` and `DiffusionEvent` all contain around 200 lines of code, but are more or less identical classes: All of them store arbitrary data of type `std::vector<double>`. Therefore it would be much better to have one base class (e.g. `DoubleVectorEvent`) of which all the other classes would be derived.

I could think of two approaches to achieve this:

**1. use derived classes**
The problem here is that derived classes do not have their own instance of the static variables. One possible solution ( http://stackoverflow.com/a/1390922 ) is to add a virtual function `type& get_static()` to the base class and call the static variable only with this function. Then each derived class can define its own static variables and by overwriting the virtual `get_static` function(s) they are also accessible for the functions of the base class.

This approach however does not work for us, as we also have a static function `static void set_syn_id( const synindex synid )` in event classes (and in the potential base class) which calls one of the static variables. Calling this static variable with the `get_static` function yields an `cannot call member function without object`-error - which means that `get_static` would have to be static as well, which is not possible as it is a virtual function.

2. **use a template class**
Employ a `template <typename something> class DoubleVectorEvent` with unused (but unique) template parameters. In this case each template class has its own static members. So one could define the events like `typedef DoubleVectorEvent<GapJunction> GapJunctionEvent` which would be nice as well. One (however solvable) problem here is the initialization of the static members. A much bigger problem is the operator `void DoubleVectorEvent::operator()()` which has to stay in `event.cpp` due to dependencies to `node.h`. A template definition of the class would however require to move the operator to `event.h`.

So for now we saw no solution to reduce the overhead here. If you have any suggestion they are of course very welcome!

models/delay_rate_connection.h
```diff
@@ -0,0 +1,160 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+/* BeginDocumentation
+Name: delay_rate_connection - Synapse type for rate connections with delay.
```
janhahne: @heplesser I discussed this with @jschuecker and @ddahmen and we like and apply your second suggestion for renaming.

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+nest::siegert_neuron::update_( Time const& origin,
+  const long from,
+  const long to,
+  const bool wfr_update )
+{
+  assert(
+    to >= 0 && ( delay ) from < kernel().connection_manager.get_min_delay() );
+  assert( from < to );
+
+  bool done = true;
```
janhahne: @heplesser I agree, that this is easier to understand. I changed `done` to `wfr_tol_exceeded` in all models. I do however still return `done` (meaning `not wfr_tol_exceeded`) from `wfr_update` since in `event_delivery_manager` and the `simulation_manager` I like `done` much better than `wfr_tol_exceeded`.

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+/* ----------------------------------------------------------------
+ * Update and event handling functions
+ */
+
+bool
+nest::siegert_neuron::update_( Time const& origin,
+  const long from,
+  const long to,
+  const bool wfr_update )
```
janhahne: @heplesser Regarding the second bullet point: Agreed, changed it to `called_from_wfr_update`.

Regarding the `const`: This is nothing that we added, but used in every neuron model I checked - so if one decides to change this, this should not be done in this PR.

models/siegert_neuron.cpp
```diff
@@ -0,0 +1,479 @@
...
+  return done;
+}
+
+void
+nest::siegert_neuron::handle( DiffusionEvent& e )
+{
+  size_t i = 0;
+  std::vector< unsigned int >::iterator it = e.begin();
+  // The call to get_coeffvalue( it ) in this loop also advances the iterator it
+  while ( it != e.end() )
```
janhahne: @heplesser I agree with you, that the writing into the communication buffer and the reading from the communication buffer of secondary events is only safe if the functions `read_from_communication_buffer` and `write_to_communication_buffer` respectively the operators `operator<<` and `operator>>` match and do not contain any bugs. This is crucial here, otherwise we almost certainly add up with a segmentation fault. But if this is done correctly, then this code here is perfectly safe (although I see that this is not obvious):

All we do here is reading from the double vector that is included in the corresponding event. Since we read it directly from the receive buffer to avoid a copy of the data it is still saved as a unsigned int vector. However this double vector has a predefined length `coeff_length_`. `coeff_length_` is a static member variable to each event class and is set only once in `NodeManager::check_wfr_use()`. An `assert` in `void set_coeffarray` ensures that every event that is send has a double vector of the right length. As `e.begin()` and `e.end()` are iterators pointing to the beginning and end of the double vector - we do exactly know that we cannot exceed the size of `B_.drift_input_` (because it is the same as `coeff_length_` of the event).

We could add a `assert( e.coeff_length_ == B_.drift_input_.size() )` before the while-loop here if you think this would be helpful?

The same arguments hold for the different event classes and handle functions.

Regarding your last bullet point: Is there any benefit in moving the `while` loop to the event class? If there is I do not see it at the moment...

models/rate_neuron_ipn_impl.h
```diff
@@ -0,0 +1,383 @@
...
+{
+  size_t i = 0;
+  std::vector< unsigned int >::iterator it = e.begin();
+  // The call to get_coeffvalue( it ) in this loop also advances the iterator it
+  while ( it != e.end() )
+  {
+    if ( P_.linear_summation_ )
+    {
+      B_.delayed_rates_.add_value(
+        e.get_delay() - kernel().connection_manager.get_min_delay() + i,
```
janhahne: See discussion on the same issue for the siegert_neuron

models/siegert_neuron.cpp
ddahmen: Thanks for the suggestion. I added an explanation of the variable alpha

models/siegert_neuron.h
jschuecker: We added the reference.

pynest/nest/tests/test_rate_neuron.py
jschuecker: Yes, this is a good idea. We extended the test to output noise.

pynest/nest/tests/test_rate_neuron_communication.py
ddahmen: We updated the documentation according to your suggestions.

models/lin_rate.h
ddahmen: I made the documentation a bit more specific stating that it is the time constant of rate dynamics.

models/siegert_neuron.cpp
ddahmen: Good point. The output of the function siegert is a rate in units of Hz. The NEST convention of time constants in ms would therefore give kHz for rates. We improved the conversion and shifted it to the return statement of the siegert functions.

models/siegert_neuron.cpp
ddahmen: We followed your suggestion and pass the variance and convert it to the standard deviation later.

models/lin_rate.h
```diff
@@ -0,0 +1,120 @@
...
+#include "rate_neuron_ipn.h"
+#include "rate_neuron_ipn_impl.h"
+#include "rate_neuron_opn.h"
+#include "rate_neuron_opn_impl.h"
+
+
+namespace nest
+{
+/* BeginDocumentation
+Name: lin_rate - Linear rate model
```
jschuecker: was discussed in NEST developer meeting and decided to keep abbreviated model names as in other neuron models.

models/lin_rate.h
```diff
@@ -0,0 +1,120 @@
...
+Parameters:
+
+ The following parameters can be set in the status dictionary.
+
+ rate                double - Rate (unitless)
+ tau                 double - Time constant in ms.
+ mean                double - Mean of Gaussian white noise.
+ std                 double - Standard deviation of Gaussian white noise.
+ g                   double - Gain parameter
+ linear summation    boolean - specifies type of non-linearity
```
jschuecker: We would like to keep the name, however, we added an explanation in the notes of the documentation making the name more comprehensible.

models/lin_rate.h
```diff
@@ -0,0 +1,120 @@
...
+
+Notes:
+The boolean parameter linear_summation does not have any
+effect in this linear model.
```
jschuecker: You are right. We removed the parameter from the documentation.

models/modelsmodule.cpp
```diff
@@ -176,6 +184,16 @@ ModelsModule::commandstring( void ) const
 void
 ModelsModule::init( SLIInterpreter* )
 {
+  kernel().model_manager.register_node_model< lin_rate_opn >( "lin_rate_opn" );
```
ddahmen: Thanks for the suggestion. This issue of namings was discussed in the developer VC. We decided to go for abbreviated names as done in other neuron models.

models/modelsmodule.cpp
```diff
@@ -176,6 +184,16 @@ ModelsModule::commandstring( void ) const
...
+  kernel().model_manager.register_node_model< lin_rate_ipn >( "lin_rate_ipn" );
+  kernel().model_manager.register_node_model< tanh_rate_opn >(
+    "tanh_rate_opn" );
+  kernel().model_manager.register_node_model< tanh_rate_ipn >(
+    "tanh_rate_ipn" );
+  kernel().model_manager.register_node_model< thresholdlin_rate_opn >(
```
ddahmen: We added the underscore in the name, but stick to the abbreviation, i.e. threshold_lin_rate_opn.

models/rate_neuron_ipn.h
```diff
@@ -0,0 +1,326 @@
...
+
+  // ----------------------------------------------------------------
+
+  /**
+   * State variables of the model.
+   */
+  struct State_
+  {
+    double r_; //!< Rate
+    double x_; //!< Noise
```
ddahmen: Good suggestion. We changed the names 
r_ -> rate_
x_ -> noise_
y_ -> noisy_rate_

models/tanh_rate.h
ddahmen: We followed your suggestion, but implemented the function 
tanh(g*(h-x))

nestkernel/event.h
ddahmen: We decided in the NEST Open Developer VC to keep the old names.

models/diffusion_connection.h
```diff
@@ -0,0 +1,189 @@
...
+Name: diffusion_connection - Synapse type for instantaneous rate connections
+between neurons of type siegert_neuron.
+
+
+Description:
+ diffusion_connection is a connector to create instantaneous connections
+ between neurons of type siegert_neuron. The connection type is identical to
+ type rate_connection for instantaneous rate connections except for the two
+ parameters "drift_factor" and "diffusion_factor" substituting the
+ parameter "weight".
```
jschuecker: Thank you for the suggestions. We added a detailed explanation. Note that these factor are quite general and not necessarily in the form of (28) and (29) in the paper.

models/lin_rate.h
```diff
@@ -0,0 +1,120 @@
...
+
+Sends: RateNeuronEvent, DelayRateNeuronEvent
+
+Receives: RateNeuronEvent, DelayRateNeuronEvent, DataLoggingRequest
+
+Author: David Dahmen, Jan Hahne, Jannis Schuecker
+SeeAlso: rate_connection, delay_rate_connection
+*/
+
+class gainfunction_lin_rate
```
janhahne: For simplicity we just named the gain functions `gainfunction_[model name]`for all template models. If you agree we would like to keep it this way.

models/siegert_neuron.h
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  void get_status( DictionaryDatum& ) const;
+  void set_status( const DictionaryDatum& );
+
+private:
+  void init_state_( const Node& proto );
+  void init_buffers_();
+  void calibrate();
+  bool update_( Time const&, const long, const long, const bool );
```
janhahne: `update_` is the "real" update function, which is either called from `wfr_update` in which case the boolean parameter is true or from `update` in which case the parameter is false.

Do we really need a comment on this here? This should be instantly clear from looking at `update` and `wfr_update` a couple of lines further down.

janhahne:
@mschmidt87 @heplesser Thank you for the careful and detailed review. We finally addressed all of your comments. Please have another look.

models/lin_rate.h
heplesser: OK

models/siegert_neuron.h
heplesser: Think of a programmer who reads this code without knowing it from before: She would have to search for all places where `update_()` is used (e.g. lines 260 and 272 below), compare them and from that comparison infer the meaning of the last parameter. This takes at best a lot of time. So yes, this method and its parameters need a comment.

models/lin_rate.h
heplesser: OK.

models/siegert_neuron.cpp
heplesser: @janhahne Sorry, it needs to be the other way around:
```c++
std::vector< double >( quantity, 0.0 ).swap( B_.last_y_values );
```

models/siegert_neuron.cpp
heplesser: @janhahne Thank you for your detailed explanation! Is that included somewhere in the code? My main concern is the long-term robustness and maintainability of the code. As you describe, there are quite a number of parts in different code locations that need to fit precisely together for this to be safe. An innocent change in one location (maybe to fix an unrelated issue) may then lead to hard-to-find errors.  By moving the while-loop to the event class, we would insulate this "sensitive" code better and we would also avoid (potentially) avoid duplication of this code across multiple models.

models/siegert_neuron.cpp
janhahne: @heplesser Makes sense, I replaced the uses of `clear` and `resize` by `swap`.

models/siegert_neuron.h
janhahne: @heplesser Of course you are right about this. However my impression was that this is quite easy to understand (the occurrences of update_ in lines 260 and 272 are in fact the only ones) and that the "threshold" for comments in the NEST code in general is rather high.

I added a documentation to `update_`.

models/siegert_neuron.cpp
janhahne: @heplesser Well, that exact explanation from above is not in included in the code, however `read_from_communication_buffer` and `write_to_communication_buffer` in the base class`SecondaryEvent` and the operators `operator<<` and `operator>>` of each event class, which are the most vulnerable parts, have detailed comments. In terms of long-term robustness and maintainability i would say that this is safe: All critical parts mentioned before are concentrated in `event.h` - changes in other locations cannot break this code.

Of course using a event class function `e.get_input` instead of several calls to `e.get_coeffvalue` in a loop over an iterator would concentrate this code even more to `event.h`, but within this loop several neuron model variables are used (e.g. `gain_`, `P_.linear_summation_`,  `B_.delayed_rates_` for `DelayedRateConnectionEvent`). In comparison to all other `handle` functions it seems much more natural to leave the loop in the `handle` function - especially as this loop over the iterator is perfectly safe if the critical part in `event.h` is correct.

heplesser:
@janhahne Thank you for addressing my comments! I am still not entirely happy with the lack of encapsulation in event handling ([see discussion on `siegert_neuron()`](https://github.com/nest/nest-simulator/pull/745#discussion_r126135813)), but I think the code is ready for integration with NEST master. Could you create a follow-up issue to remind us about reviewing the encapsulation issue later?

@jougs Since this is a major addition, I think it makes sense that a non-reviewer merges---would you do the honors?

janhahne:
@heplesser I created a follow-up issue, see #806 

@heplesser @mschmidt87 Thanks again to both of you for the detailed review!

jougs:
Due to this merge, master is currently failing. Probably it is not not related to the merge itself, but to the length of `cppcheck` output. See #798 for an explanation and a fix.

heplesser:
@jougs From the Travis logs, it looks indeed like a tests hitting a resource limit durings static code checking, i.e., the same problem as in #798.

PR #751, opened 2017-06-13
Fix #748: For nest.helpdesk() Python 2.7.8 and later is needed

### LINKED ISSUE #748: Error when nest.helpdesk() is called with Python < 2.7.8
When I install NEST with Python 2.7.6 and then call `nest.helpdesk()` in python I get an error that the file helpindex.html does not exist. If I call `nest.help('iaf_psc_alpha')` (or any other NEST object) i get the - of course somehow right - answer `Sorry, there is no help for 'iaf_psc_alpha'!`

I assume this behavior results from the combination of PR #688 and #715.

When I call `nest.helpdesk()` or `nest.help(object)` with Python 2.7.6 I would expect an information as `The NEST help is not available due to a Python version older than 2.7.8`. Otherwise it is unclear to the user what is going wrong - there is no message at all, neither when the help is accessed via python nor while cmake (which might be useful as well?)

steffengraber:
This fix for #748 checks for the right Python version ( >= 2.7.8) when nest.helpdesk() is called.

### DIFF
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -64,7 +64,8 @@ def helpdesk():
     Use the system default browser.
     """
     if sys.version_info < (2, 7, 8):
-        print("The NEST Helpdesk is only available with Python 2.7.8 or later.")
+        print("The NEST Helpdesk is only available with Python 2.7.8 or "
+              "later. \n")
         return
 
     if 'NEST_DOC_DIR' not in os.environ:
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_info.py
```diff
@@ -63,6 +63,10 @@ def helpdesk():
 
     Use the system default browser.
     """
+    if sys.version_info < (2, 7, 8):
+        print("You need python 2.7.8 or later to run this script.\n")
```
heplesser: I fear the user will be confused by "run this script". I would suggest
```
The NEST Helpdesk is only available with Python 2.7.8 or later.
```

heplesser:
@steffengraber Would you add the "Py 2.7.8 required" message also to `help()`?

PR #753, opened 2017-06-14
Add example script for structural plasticity

jakobj:
Since I've just translated a benchmark script for structural plasticity from PyNEST to sli, I thought it would make sense to integrate this into the codebase to make it available to everyone. The idea is to provide a script that allows for reasonable benchmarking of structural plasticity performance, similar as the `hpc_benchmark.sli` does for standard plasticity.

I would like to hear @sdiazpier opinion on whether this translation looks reasonable and @mschmidt87 sli-benchmark-experienced opinion on whether I messed something up that would cause a significant slowdown. Maybe @suku248 or @heplesser could also have a look regarding best practices of SLI use.

### DIFF
--- a/examples/nest/structural_plasticity_benchmark.sli
+++ b/examples/nest/structural_plasticity_benchmark.sli
@@ -1,0 +1,391 @@
+/*
+ *  structural_plasticity_benchmark.sli
+ *
+ *  This file is part of NEST.
+ *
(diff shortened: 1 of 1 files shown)


sdiazpier:
Hi dear @jakobj, great job! As far as I can see, all the elements regarding structural plasticity are ok and the translation looks reasonable. :+1:

PR #760, opened 2017-06-20
Enable users to remove gid from filename in recording devices

weidel-p:
Hi,

with this PR I'd like to give the user more power over the filename of `recording_devices`. So far, the `GID` of the device is a obligatory part of the filename. This can be annoying during the development of network models as the GID might change quite often.

I introduced a field called `use_gid` as a parameter of the `recording_device` which, if turned off, makes the `recording_device` neglect its GID in the filename. As this can lead to more collisions of filenames, this mechanism only works if the user has set a `prefix` for the filename.

### DIFF
--- a/nestkernel/nest_names.cpp
+++ b/nestkernel/nest_names.cpp
@@ -484,7 +484,7 @@ const Name U_std( "U_std" );
 const Name U_upper( "U_upper" );
 const Name update( "update" );
 const Name update_node( "update_node" );
-const Name use_gid( "use_gid" );
+const Name use_gid( "use_gid_in_filename" );
 const Name use_wfr( "use_wfr" );
 
 const Name V_act_NMDA( "V_act_NMDA" );
--- a/nestkernel/nest_names.cpp
+++ b/nestkernel/nest_names.cpp
@@ -484,6 +484,7 @@ const Name U_std( "U_std" );
 const Name U_upper( "U_upper" );
 const Name update( "update" );
 const Name update_node( "update_node" );
+const Name use_gid( "use_gid" );
 const Name use_wfr( "use_wfr" );
 
 const Name V_act_NMDA( "V_act_NMDA" );
--- a/nestkernel/nest_names.cpp
+++ b/nestkernel/nest_names.cpp
@@ -484,7 +484,7 @@ const Name U_std( "U_std" );
 const Name U_upper( "U_upper" );
 const Name update( "update" );
 const Name update_node( "update_node" );
-const Name use_gid( "use_gid_in_filename" );
+const Name use_gid_in_filename( "use_gid_in_filename" );
 const Name use_wfr( "use_wfr" );
 
 const Name V_act_NMDA( "V_act_NMDA" );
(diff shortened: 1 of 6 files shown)


nestkernel/recording_device.cpp
```diff
@@ -984,9 +989,17 @@ nest::RecordingDevice::build_filename_() const
...
-  basename << "-" << std::setfill( '0' ) << std::setw( gidigits )
-           << node_.get_gid() << "-" << std::setfill( '0' )
-           << std::setw( vpdigits ) << node_.get_vp();
+  if ( P_.use_gid_ and not P_.label_.empty() )
+  {
+    basename << "-" << std::setfill( '0' ) << std::setw( gidigits )
+             << node_.get_gid() << "-" << std::setfill( '0' )
+             << std::setw( vpdigits ) << node_.get_vp();
+  }
+  else
```
abigailm: `not P_.label.empty()` needs to be part of the else branch. Perhaps reverse the order.

abigailm:
👍

nestkernel/nest_names.cpp
```diff
@@ -484,6 +484,7 @@ const Name U_std( "U_std" );
 const Name U_upper( "U_upper" );
 const Name update( "update" );
 const Name update_node( "update_node" );
+const Name use_gid( "use_gid" );
```
jougs: This name should refer to the fact that the gid is used in the filename. As it is now, this does not become clear from the variable name

nestkernel/nest_names.h
```diff
@@ -549,6 +549,7 @@ extern const Name U_std;
 extern const Name U_upper;
 extern const Name update;      //!< Command to execute the neuron (sli_neuron)
 extern const Name update_node; //!< Command to execute the neuron (sli_neuron)
+extern const Name use_gid;     //!< use gid in the filename
```
jougs: See above.

nestkernel/recording_device.cpp
```diff
@@ -226,6 +229,8 @@ nest::RecordingDevice::Parameters_::set( const RecordingDevice& rd,
   updateValue< bool >( d, names::flush_records, flush_records_ );
   updateValue< bool >( d, names::close_on_reset, close_on_reset_ );
 
+  updateValue< bool >( d, names::use_gid, use_gid_ );
```
jougs: This line requires to assign to a temporary first and only after checking if a label is set put it in the global variable. If `use_gid` is set to `true` by the user and no label is set, a `BadProperty` exception should be thrown with an appropriate error message.

weidel-p:
I renamed `use_gid` to `use_gid_in_filename`, extended the docstring and added a `BadPropertyException`.

If a user sets the same label for different devices and sets `use_gid_in_filename` to false, NEST first throws the usual `IOError`:

>     The device file 'spike_detector-0.gdf' exists already and will not be  overwritten. 
>     Please change data_path, data_prefix or label, or set 
>     /overwrite_files to true in the root node.

The data only gets corrupted, if the user sets `overwrite_files` to true. 

I understand your concerns and also see the possible risks in this feature but I think the user is well enough protected by being forced to give a label and by the `IOError`.

jougs:
Sorry to be nagging about this, but: please rename all occurences of `use_gid` to `use_gid_in_filename`. Otherwise this gets really confusing and the function of this variable is not clear outside of the recording context. Thanks!

weidel-p:
@jougs I changed every occurrence of `use_gid` to `use_gid_in_filename`. Additionally, I added two tests to this PR.

PR #765, opened 2017-06-21
Update raster_plot

hcantunc:
Added savefig function


pynest/nest/raster_plot.py
```diff
@@ -360,3 +360,14 @@ def show():
     See also: http://matplotlib.sourceforge.net/faq/howto_faq.html#use-show
     """
     pylab.show()
+
+def savefig(fname=str(uuid.uuid4()), **kwargs):
```
jougs: I think that the random default filename is not a good idea for two reasons:
1. this might leave heaps of unclear files lying around in looped situation, which no one will recall being the raster plots after three months
2. according to [the pylab.showfig() documentation](https://matplotlib.org/devdocs/api/_as_gen/matplotlib.pyplot.savefig.html), the file format is deduced from the file name if no format is given. This does not work for the uuid.

jougs:
Many thanks for the update. We currently have some problems with out Travis CI setup. I will merge this as soon as those are resolved.

heplesser:
@hcantunc Could you merge the most recent changes from Master into your branch? This should fix the problems with failing Travis tests.

hcantunc:
@heplesser I merged the recent changes into my branch but the Travis tests failed again. Am I doing something wrong on my end?

pynest/nest/raster_plot.py
```diff
@@ -24,7 +24,7 @@
 import nest
 import numpy
 import pylab
-
+from pylab import show, savefig
 
 def extract_events(data, time=None, sel=None):
```
jougs: Please add another blank line before the function definition.

jougs:
Many thanks! As this is a minor change I'm merging without second review.

PR #766, opened 2017-06-23
Reset python helper functions to make them work again

hakonsbm:
This PR fixes issues introduced with #688, which made some of the PyNEST helper functions inoperable, and adds a Python regression test for these issues.

### DIFF
--- a/pynest/nest/lib/hl_api_helper.py
+++ b/pynest/nest/lib/hl_api_helper.py
@@ -235,11 +235,11 @@ def stack_checker(f):
         if not get_debug():
             return f(*args, **kwargs)
         else:
-            sr
-            stackload_before = spp
+            sr('count')
+            stackload_before = spp()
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_helper.py
```diff
@@ -564,7 +564,7 @@ def set_verbosity(level):
 
     # Defined in hl_api_helper to avoid circular inclusion problem with
     # hl_api_info.py
-    sr
+    sr("%s setverbosity" % level)
```
heplesser: `%` string substitution is deprecated in Py3. Rather use `'{} setverbosity'.format(level)`

heplesser:
@steffengraber Could you take a look at this? The problematic changes from #688 were probably introduced by accident and I did not catch them when I reviewed your changes then.

PR #769, opened 2017-06-29
removed warn patch (fix #662)

### LINKED ISSUE #662: Wrong "Monkeypatch" in warnings.showwarning
In [``hl_api_helper.py``](https://github.com/nest/nest-simulator/blob/32fdfd8eab48499cfea33b9cd5832f3b00f1e5f1/pynest/nest/lib/hl_api_helper.py#L39), a python builtin function has been replaced with a custom one which has:

* a wrong number of arguments
* wrong argument names

This causes crash in scripts using both ``nest`` and the ``warnings`` module.

I'm not sure replacing builtin functions is a good idea, so the question is: **Is this really necessary?**
If so, I can make a PR to replace it (it is really easy); otherwise I would be in favor of suppressing lines 37 to 41 (for which I can also make a PR, which will be even easier ^^).

Silmathoron:
Removed the warning "monkeypatch" that caused error when combining user warnings and NEST (cf #662)

### DIFF
--- a/pynest/nest/lib/hl_api_helper.py
+++ b/pynest/nest/lib/hl_api_helper.py
@@ -26,26 +26,13 @@ API of the PyNEST wrapper.
 
 import warnings
 import inspect
-import json
 import functools
 import textwrap
-import subprocess
-import os
-import re
-import sys
-
-from string import Template
 
 # These variables MUST be set by __init__.py right after importing.
 # There is no safety net, whatsoever.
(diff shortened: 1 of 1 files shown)


heplesser:
@Silmathoron It seems that almost all Python tests now fail.

Silmathoron:
Stupid mistake on my side, I overwrote a few changes without noticing x) fixing this

Issue #260, opened 2016-03-05
PyNEST: Parameters for GetStructuralPlasticityStatus 


flinz:
This is probably for @sdiazpier:

The function [`GetStructuralPlasticityStatus`](https://github.com/nest/nest-simulator/blob/master/pynest/nest/lib/hl_api_simulation.py#L134) currently takes one neccessary parameter `params`. 

I can however neither pass a string, a list of strings, only an empty dictionary works for me so far. 

Perhaps this should be a optional parameter?
E.g., `GetKernelStatus` takes both strs and list of strs, and all of them are optional.

sdiazpier:
Hi @flinz 
Yes you are right, for the moment a dictionary is expected in this function which is not used and it is not optional. Thank you noticing this. I will reformulate the call format of the function.

jougs:
@sdiazpier: can you please update us on the status? If this was fixed by #775, I'd like to close this issue.

heplesser:
Solved by #775, thus closing.

PR #775, opened 2017-07-02
Change in GetStructrualPlasticityStatus format

sdiazpier:
This PR addresses issue #260 by changing the call format for GetStructuralPlasticityStatus. Now the parameter for the call is optional and can include a string defining the elements of interest from the status dictionary. The call still returns the dictionary containing the status filtered by the parameter string. A small test was also added.

### DIFF
--- a/pynest/nest/lib/hl_api_simulation.py
+++ b/pynest/nest/lib/hl_api_simulation.py
@@ -234,9 +234,14 @@ def SetStructuralPlasticityStatus(params):
 
 
 @check_stack
-def GetStructuralPlasticityStatus(keys = None):
+def GetStructuralPlasticityStatus(keys=None):
     """Get the current structural plasticity parameters for the network
     simulation.
+    
+    Parameters
+    ---------
+    keys : str or list, optional
+        Keys indicating the values of interest to be retrieved by the get call
     """
 
     sps({})
@@ -246,8 +251,10 @@ def GetStructuralPlasticityStatus(keys = None):
         return d
     elif is_literal(keys):
         return d[keys]
+    elif is_iterable(keys):
+        return tuple(d[k] for k in keys)
     else:
-        raise TypeError("keys should be either empty or a string")
+        raise TypeError("keys must be either empty, a string or a list of strings")
 
 
 @check_stack
--- a/pynest/nest/lib/hl_api_simulation.py
+++ b/pynest/nest/lib/hl_api_simulation.py
@@ -234,14 +234,20 @@ def SetStructuralPlasticityStatus(params):
 
 
 @check_stack
-def GetStructuralPlasticityStatus(params):
+def GetStructuralPlasticityStatus(keys = None):
     """Get the current structural plasticity parameters for the network
     simulation.
     """
 
-    sps(params)
+    sps({})
     sr('GetStructuralPlasticityStatus')
-    return spp()
+    d = spp()
+    if keys is None:
+        return d
+    elif is_literal(keys):
+        return d[keys]
+    else:
+        raise TypeError("keys should be either empty or a string")
 
 
 @check_stack
--- a/pynest/nest/lib/hl_api_simulation.py
+++ b/pynest/nest/lib/hl_api_simulation.py
@@ -237,7 +237,7 @@ def SetStructuralPlasticityStatus(params):
 def GetStructuralPlasticityStatus(keys=None):
     """Get the current structural plasticity parameters for the network
     simulation.
-    
+
     Parameters
     ---------
     keys : str or list, optional
@@ -254,7 +254,7 @@ def GetStructuralPlasticityStatus(keys=None):
     elif is_iterable(keys):
         return tuple(d[k] for k in keys)
     else:
-        raise TypeError("keys must be either empty, a string or a list of strings")
+        raise TypeError("keys must be either empty, a string or a list")
 
 
 @check_stack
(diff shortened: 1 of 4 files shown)


pynest/nest/lib/hl_api_simulation.py
```diff
@@ -234,14 +234,20 @@ def SetStructuralPlasticityStatus(params):
 
 
 @check_stack
-def GetStructuralPlasticityStatus(params):
+def GetStructuralPlasticityStatus(keys = None):
```
heplesser: No spaces around `=` in parameter lists.

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -234,14 +234,20 @@ def SetStructuralPlasticityStatus(params):
...
     """Get the current structural plasticity parameters for the network
     simulation.
```
heplesser: Please document the `keys` parameter.

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -234,14 +234,20 @@ def SetStructuralPlasticityStatus(params):
...
 
-    sps(params)
+    sps({})
     sr('GetStructuralPlasticityStatus')
-    return spp()
+    d = spp()
+    if keys is None:
+        return d
+    elif is_literal(keys):
+        return d[keys]
```
heplesser: Here, `keys` can only be a single key, so why the plural? And wouldn't it make more sense to allow a list of keys, as in `GetStatus`?

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -234,14 +234,20 @@ def SetStructuralPlasticityStatus(params):
...
+    else:
+        raise TypeError("keys should be either empty or a string")
```
heplesser: "should" is not a good term here, as it implies a recommendation, not a requirement, while we have strict requirements on the type of `keys`. Use "must" instead.

pynest/nest/tests/test_sp/test_all.py
```diff
@@ -60,9 +62,10 @@ def suite():
     test_suite.addTest(test_growth_curves.suite())
     test_suite.addTest(test_sp_manager.suite())
     test_suite.addTest(test_synaptic_elements.suite())
-    test_suite.addTest(test_disconnect.suite())
+    #test_suite.addTest(test_disconnect.suite())
```
heplesser: Why are you commenting out the disconnection test? Either activate it again or delete the line entirely.

pynest/nest/tests/test_sp/test_get_sp_status.py
```diff
@@ -0,0 +1,90 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+'''
+Structural Plasticity GetStatus Test
+-----------------------
+This test shows how to use the GetStructuralPlasticityStatus function
```
heplesser: The test should rather show that the function works correctly. That it also illustrates how to use the function is rather a (welcome) side-effect.

pynest/nest/tests/test_sp/test_get_sp_status.py
```diff
@@ -0,0 +1,90 @@
...
+	    'Axon_ex': growth_curve,
+	}
+	nodes = nest.Create(neuron_model,
+			           2,
+			           {'synaptic_elements': synaptic_elements})
+
+	all = nest.GetStructuralPlasticityStatus()
+	print (all)
+        assert('structural_plasticity_synapses' in all)
+        assert('structural_plasticity_update_interval' in all)
```
heplesser: Why are you not testing the values here?

pynest/nest/tests/test_sp/test_all.py
sdiazpier: This is a mistake, I will remove the comment it now.

pynest/nest/tests/test_sp/test_get_sp_status.py
sdiazpier: I just want to know that the elements have been gathered by the call to get status but I can test the values too :)

pynest/nest/tests/test_sp/test_get_sp_status.py
sdiazpier: Ok, will correct the text

pynest/nest/lib/hl_api_simulation.py
sdiazpier: I left it plural but allow now a list like in GetStatus :)

pynest/nest/tests/test_sp/test_get_sp_status.py
```diff
@@ -0,0 +1,103 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Structural Plasticity GetStatus Test
+-----------------------
+This test the functionality of the GetStructuralPlasticityStatus
```
abigailm: this tests

PR #776, opened 2017-07-03
Prohibit DataConnect in multithreaded mode.

### LINKED ISSUE #527: Segmentation fault with DataConnect and Multithreading
I was trying DataConnect with multiple threads, and I got a segmentation fault if the number of threads was >1.

I have compiled the current Github version of NEST on Linux Ubuntu 14.04, with gcc-4.8.

This is the pynest script that I have used to reproduce the issue:

```
import numpy as np
import nest

CORES = 1

nest.set_verbosity('M_WARNING')
nest.ResetKernel()
nest.SetKernelStatus({'local_num_threads' : CORES,
                      'total_num_virtual_procs' : CORES})

A=nest.Create('iaf_neuron',1)
B=nest.Create('iaf_neuron',1000)

nest.DataConnect(A,[{'target':np.float64(B),
                         'delay':1.0*np.ones(1000, np.float64),
                         'weight':1.0*np.ones(1000, np.float64) }],'static_synapse')

AB_conn = nest.GetConnections(A,B)
print "Number of A-B synapses: " + str(len(AB_conn))

```

If I set "CORES" to 2, 4 or 8, I get a segmentation fault error, sometimes without any information, sometime with a little more:

> [alberto:28834] **\* Process received signal ***
> [alberto:28834] Signal: Segmentation fault (11)
> [alberto:28834] Signal code: Address not mapped (1)
> ## [alberto:28834] Failing at address: (nil)
> 
> mpirun noticed that process rank 0 with PID 28834 on node alberto exited on signal 11 (Segmentation fault).

I have seen that there are similar problems with deprecated functions as Convergent/Divergent connect, maybe this is related.

Do you suggest to avoid using DataConnect and use something else instead?

stinebuu:
This PR fixes #527. I have added a check to see if we have more than one thread when using `DataConnect`, in which case an exception is thrown. I have also added a simple regressiontest.

### DIFF
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -866,8 +866,7 @@ NestModule::DataConnect_i_D_sFunction::execute( SLIInterpreter* i ) const
 
   if ( kernel().vp_manager.get_num_threads() > 1 )
   {
-    throw KernelException(
-      "DataConnect can not be used with multiple threads" );
+    throw KernelException( "DataConnect cannot be used with multiple threads" );
   }
 
   const index source = getValue< long >( i->OStack.pick( 2 ) );
@@ -931,8 +930,7 @@ NestModule::DataConnect_aFunction::execute( SLIInterpreter* i ) const
 
   if ( kernel().vp_manager.get_num_threads() > 1 )
   {
-    throw KernelException(
-      "DataConnect can not be used with multiple threads" );
+    throw KernelException( "DataConnect cannot be used with multiple threads" );
   }
 
   const ArrayDatum connectome = getValue< ArrayDatum >( i->OStack.top() );
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -864,6 +864,12 @@ NestModule::DataConnect_i_D_sFunction::execute( SLIInterpreter* i ) const
 {
   i->assert_stack_load( 3 );
 
+  if ( kernel().vp_manager.get_num_threads() > 1 )
+  {
+    throw KernelException(
+      "DataConnect can not be used with multiple threads" );
+  }
+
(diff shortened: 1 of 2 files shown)


nestkernel/nestmodule.cpp
```diff
@@ -864,6 +864,12 @@ NestModule::DataConnect_i_D_sFunction::execute( SLIInterpreter* i ) const
 {
   i->assert_stack_load( 3 );
 
+  if ( kernel().vp_manager.get_num_threads() > 1 )
+  {
+    throw KernelException(
+      "DataConnect can not be used with multiple threads" );
```
heplesser: `cannot`

nestkernel/nestmodule.cpp
```diff
@@ -922,6 +928,13 @@ void
 NestModule::DataConnect_aFunction::execute( SLIInterpreter* i ) const
 {
   i->assert_stack_load( 1 );
+
+  if ( kernel().vp_manager.get_num_threads() > 1 )
+  {
+    throw KernelException(
+      "DataConnect can not be used with multiple threads" );
```
heplesser: `cannot`

testsuite/regressiontests/issue-527.sli
```diff
@@ -0,0 +1,74 @@
...
+(unittest) run
+/unittest using
+
+skip_if_not_threaded
+
+/test_threaded_DataConnect
+{
+  dup /threads Set
+  ResetKernel
+  0 << /local_num_threads threads /total_num_virtual_procs threads >> SetStatus
```
heplesser: No need to set `total_num_virtual_procs`, that is set automatically based on `local_num_threads`.

heplesser:
Release notes: `DataConnect` is now prohibited in multi-threaded mode because it is unstable, see #527.

PR #788, opened 2017-07-19
Fix #762: PyNEST test result parsing confused by NEST output

### LINKED ISSUE #762: PyNEST test result parsing confused by NEST output
The [code parsing the output generated by nosetests](https://github.com/nest/nest-simulator/blob/f201ca0ac064c28e05d2b57f132c51c8065f6efa/testsuite/do_tests.sh.in#L722) to provide a progress indication during testing is confused by output generated by NEST, since the parser relies on lines of the form `TestABC ... ok`. If NEST input intervenes before `ok` (or any other result), no progress will be displayed, confusing the user (see also #677).

We thus should revise output parsing, but that should first happen once we have decided how to test in the future #761.

gtrensch:
This PR installs the nosetests capturestderr plugin and corrcets some of the nosetests log messages. For more details please see issue #762.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,9 +83,7 @@ before_install:
   # (Please note that the nose-capturestderr plugin is needed to suppress NEST output to stderr which may interfere 
   #  with the nosetests output hindering correct log-parsing. stdout it captured by default.)
   - sudo apt-get install -y python-nose
-  - git clone https://github.com/sio2project/nose-capturestderr.git capturestderr
-  - cd capturestderr
-  - sudo python setup.py install
+  - sudo pip install "nose-capturestderr==1.2"
 
   # Install VERA++ and pep8 for static code analysis.
   - sudo apt-get install -y vera++ pep8
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,12 @@ before_install:
   - sudo apt-get install -y openmpi-bin libopenmpi-dev
 
   # Install nose framework for the Python testsuite.
+  # (Please note that the nose-capturestderr plugin is needed to suppress NEST output to stderr which may interfere 
+  #  with the nosetests output hindering correct log-parsing. stdout it captured by default.)
   - sudo apt-get install -y python-nose
+  - git clone https://github.com/sio2project/nose-capturestderr.git capturestderr
+  - cd capturestderr
+  - sudo python setup.py install
 
   # Install VERA++ and pep8 for static code analysis.
   - sudo apt-get install -y vera++ pep8
@@ -110,6 +115,9 @@ install:
   - which python
   - python --version
   - which pip
+  - which nosetests
+  - nosetests --version
+  - nosetests --plugins
   - which cmake
   - cmake --version
 
--- a/pynest/nest/tests/test_aeif_lsodar.py
+++ b/pynest/nest/tests/test_aeif_lsodar.py
@@ -249,10 +249,10 @@ class AEIFTestCase(unittest.TestCase):
                                     model, var, diff, di_tol[model][var]))
 
     @unittest.skipIf(not HAVE_GSL, 'GSL is not available')
-    def test_closeness_nest_lsodar(self):
-        '''
-        Compare models to the LSODAR implementation.
-        '''
+    def testClosenessNestLSODAR(self):
+
+        # Compare models to the LSODAR implementation.
+
         simtime = 100.
 
         # get lsodar reference
@@ -281,11 +281,11 @@ class AEIFTestCase(unittest.TestCase):
         self.assert_pass_tolerance(rel_diff, di_tolerances_lsodar)
 
     @unittest.skipIf(not HAVE_GSL, 'GSL is not available')
-    def test_iaf_behaviour(self):
-        '''
-        The models should behave as iaf_cond_* if a == 0., b == 0. and
-        Delta_T == 0.
-        '''
+    def testIafBehaviour(self):
+
+        # The models should behave as iaf_cond_* if a == 0., b == 0. and
+        # Delta_T == 0.
+
         simtime = 200.
         # create the neurons and devices
         refs = {syn_type: nest.Create("iaf_" + syn_type,


.travis.yml
```diff
@@ -80,7 +80,12 @@ before_install:
   - sudo apt-get install -y openmpi-bin libopenmpi-dev
 
   # Install nose framework for the Python testsuite.
+  # (Please note that the nose-capturestderr plugin is needed to suppress NEST output to stderr which may interfere 
+  #  with the nosetests output hindering correct log-parsing. stdout it captured by default.)
   - sudo apt-get install -y python-nose
+  - git clone https://github.com/sio2project/nose-capturestderr.git capturestderr
```
Silmathoron: Given how small this plugin is, do you think it would be possible to include this in the nest repo directly after asking the guy about the license? It would avoid problems if the repo was moved or closed...

pynest/nest/tests/test_aeif_lsodar.py
```diff
@@ -249,10 +249,10 @@ def assert_pass_tolerance(self, rel_diff, di_tol):
                                     model, var, diff, di_tol[model][var]))
 
     @unittest.skipIf(not HAVE_GSL, 'GSL is not available')
-    def test_closeness_nest_lsodar(self):
-        '''
-        Compare models to the LSODAR implementation.
-        '''
+    def testClosenessNestLSODAR(self):
+
+        # Compare models to the LSODAR implementation.
```
Silmathoron: Why would you remove standard docstrings to replace them with comments? Does this have to do with the docstrings being printed somewhere during the test? I'm confused...

.travis.yml
gtrensch: This would indeed be an option. There is also hope that the stderr-plugin will be part of nosetests itself. Unclear is also whether we stay with nosetests or move to another unit test framework.

pynest/nest/tests/test_aeif_lsodar.py
gtrensch: If present, nostests takes the first line of the docstring for reporting on the test cases. This led to something like this:  `The models should behave as iaf_cond_* if a == 0., b == 0. and ... ok`
Most of the test cases make no use of a docstring. In that case the function name is reported. Hence I turned the docstring into line comments to make it clear and consistent.

heplesser: @gtrensch Tests should be commented using docstrings and the first line should give a concise description of the test. We haven't been particularly systematic about this, but changing docstrings to comments, is in my opinion the wrong way to go. In this particular case, the text is also fine as a docstring.

pynest/nest/tests/test_aeif_lsodar.py
```diff
@@ -281,11 +281,11 @@ def test_closeness_nest_lsodar(self):
...
     @unittest.skipIf(not HAVE_GSL, 'GSL is not available')
-    def test_iaf_behaviour(self):
-        '''
-        The models should behave as iaf_cond_* if a == 0., b == 0. and
-        Delta_T == 0.
-        '''
+    def testIafBehaviour(self):
+
+        # The models should behave as iaf_cond_* if a == 0., b == 0. and
+        # Delta_T == 0.
```
heplesser: This should remain a docstring, but with a better text, e.g.
```Python
"""
Ensure consistency with iaf_cond_* models.

The models shall behave as iaf_cond_* if a == b == Delta_T == 0.
"""
```

.travis.yml
heplesser: @gtrensch This solution will only work on Travis and makes us dependent on yet another external repository. Users running the testsuite would need to install the plugin manually, complicating things. See also [my comment on the issue](https://github.com/nest/nest-simulator/issues/762#issuecomment-319042461).

pynest/nest/tests/test_aeif_lsodar.py
gtrensch: @heplesser using the docstring here was one of the confusions. As stated above, nosetests takes the first line, and only the first line of the docstring, to report on the test. I agree, well written docstrings are good coding practice. Just a few test cases make use of it. I'm not sure if it should be of the scope of this PR to do this refactoring task here.

pynest/nest/tests/test_aeif_lsodar.py
gtrensch: @heplesser using the docstring here was one of the confusions. As stated above, nosetests takes the first line, and only the first line of the docstring, to report on the test. I agree, well written docstrings are good coding practice. Just a few test cases make use of it. I'm not sure if it should be of the scope of this PR to do this refactoring task here.

heplesser: I think I misunderstood. So even if nosetest only outputs a single line that still breaks our output parsing? I had though/hoped that our parsing would work as long as nose did no insert any line breaks. We should discuss how to proceed in the next VC.

gtrensch:
@heplesser indeed I'm also not happy with this approach. On the other hand we do similar things in the PIP Cython and Music install sections. I see this as an indermediate solution to fix the bug. If we move to nose2 or perhaps pytest this would be gone.

.travis.yml
lekshmideepu: @gtrensch : Have you tried to use the pypi  package for _nose-capturestderr 1.2_
https://pypi.python.org/pypi/nose-capturestderr/1.2

gtrensch:
@lekshmideepu We can also download and install the tar file but isn't it the same source?

Silmathoron:
It think it is not: using pip to access [Pypi](https://pypi.python.org/pypi) means accessing to a "repository of software for the Python programming language", which means that the package is supposed to be a lot more stable in time, whatever happens to the GitHub repo.

gtrensch:
@lekshmideepu @Silmathoron Thanks, I got the point. Installing from the Python package index means you can simply use "pip install". I didn't know this. This simplifies things a lot and as a nice side effect the plugin also appears in the list of the plugins installed.

pynest/nest/tests/test_aeif_lsodar.py
gtrensch: @heplesser I might be wrong but I think the problem had nothing to do with line breaks contained in the docstring. Nest's outpout to stdout/stderr just simply comes in between, the report on the unit test printed by nosetests and the "... ok" string at the end of single test run. The line breaks came from the Nest output. 
If the first line of the docstring would contain a meaningful test description, things are perfect. We have > 500 nose-based unit tests (I believe) where most of it do not make use of a docstring. There is some effort adding docstrings to all tests. 
.. but yes, let's discuss this in the VC.

lekshmideepu:
@gtrensch : Yes, @Silmathoron is right. Pypi packages are supposed to be stable when compared to installing a package from a GitHub repo.
You could see [here](https://pypi.python.org/pypi/nose-capturestderr/1.2) the following: 
Author: The SIO2 Project Team
Home Page: http://github.com/sio2project/nose-capturestderr

lekshmideepu:
@gtrensch : Plugin capturestderr was also shown in the [earlier case](https://travis-ci.org/nest/nest-simulator/jobs/255143651#L1901) when you were installing from the [GitHub repo](https://travis-ci.org/nest/nest-simulator/jobs/255143651#L813).

heplesser:
@lekshmideepu Have the changes by @gtrensch addressed your concerns? If yes, is the PR ready for approval? If no, which problems remain?

Issue #763, opened 2017-06-21
Testing on multiple python versions.


lekshmideepu:
For some time ago, there was a discussion on testing multiple python versions. I tried the matrix [build ](https://travis-ci.org/lekshmideepu/nest-simulator/builds/245277036)for two different versions (2.x and 3.x) of python.  (issue with the pythonlibs has to be fixed in the build. I am working on this.)
I think testing the latest version of builds from both 2 and 3 series is enough. Could you please add your thoughts here. 
Also, testing could be done in two ways, either using a [tox](https://tox.readthedocs.io/en/latest/) plugin (or [tox-travis](https://github.com/ryanhiebert/tox-travis)) or by using the travis python images.

heplesser:
@lekshmideepu I agree that using the latest from the 2 and the 3 series, respectively, gives a reasonable balance between resources used and coverage.

I know too little about tox or travis images, so I would leave that to you.

terhorstd:
I would agree to testing the last versions only.

heplesser:
Closed since testing with Py2 and Py3 has been introduced with #789.

Issue #759, opened 2017-06-20
Moving the installation of CPPCHECK and CLANG-FORMAT to .travis.yml


lekshmideepu:
As we have[ $xSTATIC_ANALYSIS enabled](https://github.com/nest/nest-simulator/blob/master/.travis.yml#L36) for all our builds, do we need [this](https://github.com/nest/nest-simulator/blob/master/build.sh#L90) ?

Also I think we could move these installations from build.sh to the package installation section in .travis.yml where we install all the other packages.

lekshmideepu:
I have successfully moved the installation to the yaml file and the build is fine except the final build summary table as it depends on the MSGBLDxxxx numbers created by @gtrensch 
Made necessary changes in the build.sh file and have to do the changes in extras/parse_travis_log.py

gtrensch:
Generally, moving all the installation to the yml-file is a good idea. $STATIC_ANALYSIS is always set in the build matrix and can probably be removed too. 

The MSGBLDxxxx message numbers indicate the status of the code analysis tool installation process. They are needed to evaluate the build by the parser script.

gtrensch:
@lekshmideepu A note on the message numbers: Parsing log files or console outputs is always a proplem. From a software engineering perspective it introduces dependencies and a coupling which you actually want to avoid. The previous implementation relied on the message texts itself. Means, fixing a typo could mess up the log parser. So I intruduced the numbers. You may move and change the messages but not the mumbers. If so doing, you need to take care of their dependencies with the python log parser.

Issue #758, opened 2017-06-20
Using container based infrastructure for Trusty


lekshmideepu:
Currently we use the"generic" build of Travis where the python image is provided by Ubuntu and is an old version 2.7.6. This creates some SSL warnings and I have already discussed the same with travis developers (https://github.com/travis-ci/travis-ci/issues/7868).
As we also wanted to speedup our builds by using the container based infrastructure of Trusty image, I have opened the following requests to add some of our packages to the apt-addon list.
https://github.com/travis-ci/apt-package-whitelist/issues/4168
https://github.com/travis-ci/apt-package-whitelist/issues/4167
https://github.com/travis-ci/apt-package-whitelist/issues/4166
https://github.com/travis-ci/apt-package-whitelist/issues/4165
https://github.com/travis-ci/apt-package-whitelist/issues/4164
https://github.com/travis-ci/apt-package-whitelist/issues/4163
https://github.com/travis-ci/apt-package-whitelist/issues/4162
https://github.com/travis-ci/apt-package-whitelist/issues/4161

heplesser:
See also #682.

heplesser:
And how does this relate to #641?

lekshmideepu:
As far as I know, the use of [container](https://docs.travis-ci.com/user/trusty-ci-environment/#Container-based-with-sudo%3A-false) based builds and [docker](https://docs.travis-ci.com/user/docker/) services are little bit different.
As some of the packages are not available in the [white list](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty), I think we need to wait a bit for the container builds.

For trying out the docker service of travis, will meet with @DimitriPlotnikov next week and will try to run the NESTML docker image with travis.

Issue #619, opened 2017-01-05
Add test for python 3 in travis


Silmathoron:
Travis test with Python should run tests for both Python 2 and Python 3 (this would avoid including errors from users developing under Python 2)

heplesser:
@lekshmideepu @jougs A recent merge broke NEST under Python 3 (see #617) and it turned out that there were a number of other Python 3 incompatibilities as well (see #620 for fixes).  This was not discovered by Travis since we only test with Python 2 at the moment.

Could you investigate how to set up CI testing with both Python 2.7 and 3.5?

jougs:
I have this on my agenda since long and can have a look. My idea was be to use virtualenv and only test the maximal configuration after that worked with Python 2.7

heplesser:
Yes, that sounds sensible.

lekshmideepu:
Working on this issue. Even though I added the following lines in .travis.yml file, (which should work in a normal case) , it is building only with default python version 2.7
python:
  - "2.7"
  - "3.5"

Using the virtualenv is a nice idea.

Silmathoron:
Hi @lekshmideepu, I'm a bit surprised that the travis matrix does not work.
Here [one that works](https://github.com/Silmathoron/NNGT/blob/master/.travis.yml) on one of my projects. Alternatively you can make sure that NEST gets the desired PyVersion by exporting an environment variable ``PYV=2`` or 3, which you'll pass to ``-Dwith-python=PYV``

PR #789, opened 2017-07-20
Major change in the installation and testing of packages using multiple python versions

lekshmideepu:
This PR consists of a number of changes in the installation and testing processes:
1. Using multiple python versions. #763 
2. Using the container based whitelist packages from Travis. #758 
3. Explicitly defining the CMake options as the system unable to recognise the python libraries.
4. Some code changes to adapt to both python2 and python3
5. Moved the package (cppcheck and CLANG_FORMAT) installation from build.sh  to yaml. #759 
6. Changes in the message passing numbers for the parse log.


.travis.yml
```diff
@@ -20,114 +20,123 @@
...
-  - ./autogen.sh
-  - ./configure --prefix=$HOME/.cache/music.install
-  - make
-  - make install
-  - cd $HOME/build
-
-  # Change directory back to the NEST source code directory.
-  - cd $SOURCEDIR
+  # terminaltables is required by parse_travis_log.py to create the build summary.
+   - pip install cython scipy matplotlib terminaltables mpi4py
```
jougs: `scipy`, `matplotlib` and `mpi4py` are already installed above in the `apt` block. Is `pip install`ing these packages a requirement so that the test setup works with the different Python versions?

Can you please either remove the packages either here (my preferred solution if that works with testing different Python versions) or above, so they are not installed multiple times? Also, could `cython` and `terminaltables` be installed using `apt-get`?

build.sh
```diff
@@ -87,98 +93,71 @@ NEST_RESULT=$(readlink -f $NEST_RESULT)
...
-    make PREFIX=$HOME/.cache CFGDIR=$HOME/.cache/cfg HAVE_RULES=yes install
-    cd ..
-    echo "MSGBLD0040: CPPCHECK installation completed."
-
-    echo "MSGBLD0050: Installing CLANG-FORMAT."
-    wget --no-verbose http://llvm.org/releases/3.6.2/clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04.tar.xz
-    tar xf clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04.tar.xz
-    # Copy and not move because '.cache' may aleady contain other subdirectories and files.
-    cp -R clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04/* $HOME/.cache
-    echo "MSGBLD0060: CLANG-FORMAT installation completed."
```
jougs: Previously we required this exact version of `clang-format`. Is that not the case anymore? Where is clang-format installed now?

testsuite/regressiontests/ticket-659-copyright.py
```diff
@@ -42,6 +42,7 @@
     'libltdl',
     '.git',
     'CMakeFiles',
+    'music',
```
jougs: See above.

music_install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+ # Install current MUSIC version.
+git clone https://github.com/INCF/MUSIC.git music
+cd music
+./autogen.sh
+./configure --prefix=$HOME/.cache/music.install
+make
+make install
```
jougs: Instead of excluding the `music` directory from copyright header checks below, I think it's cleaner to `rm -rf` the source directory after the `make install` step here.

pynest/nest/tests/test_sp/test_disconnect.py
```diff
@@ -75,26 +77,26 @@ def test_synapse_deletion_one_to_one_no_sp(self):
                 conns = nest.GetConnections(
                     [neurons[0]], [neurons[2]], syn_model)
                 if mpi_test:
-                    connstotal = None
-                    conns = self.comm.allgather(conns, connstotal)
-                    conns = filter(None, conns)
+                    conns = self.comm.allgather(conns)
+                   # conns = filter(None, conns)
```
jougs: Please remove the commented out code. This also applies to the occurences below.

build.sh
lekshmideepu: clang-format --version : clang-format version 3.5.0 (tags/RELEASE_350/final)
$ which clang-format :  /usr/local/clang-3.5.0/bin/clang-format

music_install.sh
lekshmideepu: Thanks for the suggestion. I have added the same

.travis.yml
lekshmideepu: Oh yes, sorry for not removing the same from the `apt` block. I modified this part.
The version of those packages were bit old and it was affecting our tests. That is the reason why I have installed the same including `mpi4py` using pip.
Package `terminaltables` is not yet available in the apt package whitelist. I have already opened an [issue](https://github.com/travis-ci/apt-package-whitelist/issues/4206) for that.

pynest/nest/tests/test_sp/test_disconnect.py
lekshmideepu: @jougs : I have removed the commented out code.

.travis.yml
```diff
@@ -20,114 +20,123 @@
...
-language: minimal
-
-# Required for Ubuntu Trusty (14.04 LTS).
+# Using the container infrastructure for Ubuntu Trusty (14.04 LTS).
 sudo: required
 dist: trusty
+
 #https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2
-group: edge
+group: deprecated-2017Q2
```
Silmathoron: From [that page](https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch), this would only be valid until 9/1/17...  
Is this still related to the python versions problem that was discussed during the VC?
Otherwise can we not fix whatever goes wrong instead of working with the deprecated image?

Silmathoron: Here is a possible fix: [https://gist.github.com/dan-blanchard/7045057](https://gist.github.com/dan-blanchard/7045057)
Otherwise [this might be adapted](https://stackoverflow.com/a/17254588/5962321), I think.

.travis.yml
```diff
@@ -20,114 +20,123 @@
...
+   - xTHREADING=0 xMPI=1 xGSL=0 xLTDL=0 xREADLINE=0 xPYTHON=1 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # Python & MPI
+   - xTHREADING=0 xMPI=0 xGSL=0 xLTDL=0 xREADLINE=0 xPYTHON=1 xMUSIC=0 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # only Python
+   - xTHREADING=1 xMPI=1 xGSL=1 xLTDL=1 xREADLINE=1 xPYTHON=1 xMUSIC=1 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB   # full
+python:
+  - 2.7.13
+  - 3.4.4
 matrix:
-  # Do notify immediately about it when a job of a build fails.
-  fast_finish: true
+    exclude:
```
Silmathoron: This is a lot of lines for little done... would the following not work? Remove 3.4.4 from python on L46 and add
```
matrix:
    include:
        - env xTHREADING=1 xMPI=1 xGSL=1 xLTDL=1 xREADLINE=1 xPYTHON=1 xMUSIC=1 xSTATIC_ANALYSIS=1 CACHE_NAME=JOB
          python 3.4.4
```

.travis.yml
Silmathoron: If scipy is install from pip, I'm not sure I understand why we are using the deprecated version, nor why it is still present on line 77

build.sh
Silmathoron: I don't understand the answer... is the formatting with CLANG 3.5 the same as that of 3.6? If it is not the case, then a lot of tests will fail

music_install.sh
```diff
@@ -0,0 +1,11 @@
...
+cd ..
+rm -rf music
```
jougs: Please add a newline at the end of the file.

build.sh
jougs: I agree with @Silmathoron. Our [C++ Coding Guidelines](https://nest.github.io/nest-simulator/coding_guidelines_c++#tooling) explicitly point out that we require clang-format 3.6.

Can you please double-check if all checks run properly using this old version? One way to check would probably be to manually check all our source code files in a TravisCI debug build.

heplesser:
See also #619

jougs:
@lekshmideepu, @sdiazpier: Would it be possible / make sense to factor out the part fixing the MPI problems in `test_disconnect.py` and create another pull request from them? They are more or less independent of the changes to the CI setup, aren't they? Thanks!

sdiazpier:
@lekshmideepu and @jougs I just started a new pull request with the changes that @lekshmideepu implemented in this PR to solve the issue with the testing of synapse deletion using mpi and pynest. @lekshmideepu could you please remove the modification in this PR?

heplesser:
@lekshmideepu It would be extremely useful to get this PR merged. Could you fix the remaining issues in the near future?

jougs:
@lekshmideepu: can you please give a short update on the status? If you require assistance with this, I might be able to help.

heplesser:
@lekshmideepu I see you are very busily working on this PR at the moment. Would you post a little note here once you want the reviewers to take another look?

music_install.sh
lekshmideepu: done!

lekshmideepu:
sure @heplesser

.travis.yml
lekshmideepu: No more using the deprecated branch.

.travis.yml
lekshmideepu: No more using the deprecated version. Since the version of some packages like scipy is old in the source package whitelist, I am installing those via pip

build.sh
lekshmideepu: I revert it back to clang-format 3.6.2

.travis.yml
lekshmideepu: In fact, I spent a lot of time on the matrix include and exclude combinations. The above one simply doesn't work for some "strange" unknown reason. I also had a chat with travis team few months back and it seems their setup for include is a bit tricky.

Silmathoron:
This is looking good, I think all that's left is an error in ``test_unused_names.py``, correcting the output of ``check_output`` through a function like:

```python
def _to_string(byte_string):
    ''' Convert bytes to string '''
    if isinstance(byte_string, bytes):
        return str(byte_string.decode())
    return byte_string
```

lekshmideepu:
@Silmathoron : I am already in touch with @jougs and this particular test is not compatible with Python3.x. Jochen currently issued a PR against my branch. Travis is running this build now.

lekshmideepu:
https://github.com/lekshmideepu/nest-simulator/pull/1

lekshmideepu:
@jougs @Silmathoron Hope it is fine now. Please have a look.

lekshmideepu:
@Silmathoron : include syntax doesn't trigger a travis build. That is why I changed it.

lekshmideepu:
@Silmathoron : Thanks. Probably I would try the matrix include in another PR once this is merged.

lekshmideepu:
@jougs : Could you please take a look again?

.travis.yml
```diff
@@ -20,142 +20,122 @@
...
-  - make install
-  - popd
-
-  # Change directory back to the NEST source code directory.
-  - cd $SOURCEDIR
-
+  # terminaltables is required by parse_travis_log.py to create the build summary.
+   - pip install cython scipy matplotlib terminaltables mpi4py
+  # Install boost libraries required for VERA++ static code analysis.   
+   - sudo apt-get install -y libboost-filesystem-dev libboost-program-options-dev libboost-regex-dev libboost-wave-dev libboost-python-dev
```
jougs: What is the reason for installing these Python libraries via `pip` and the Ubuntu packages manually using `apt-get` instead of adding all of them to the `addons` block above? Please either move them up and/or comment here on the reason.

.travis.yml
```diff
@@ -172,4 +152,4 @@ deploy:
     branch: master
   local-dir: "$TRAVIS_BUILD_DIR/build/artefacts_upload"
   upload-dir: "$TRAVIS_REPO_SLUG/$TRAVIS_BUILD_NUMBER/$TRAVIS_JOB_NUMBER"
-  acl: bucket_owner_full_control
+  acl: bucket_owner_full_control
```
jougs: Please add a newline to the end of the file.

build.sh
```diff
@@ -240,8 +245,15 @@ echo "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +
 echo "+               R U N   N E S T   T E S T S U I T E                           +"
 echo "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
 echo "MSGBLD0290: Running make installcheck."
-export PYTHONPATH=$HOME/.cache/csa.install/lib/python2.7/site-packages:$PYTHONPATH
-export LD_LIBRARY_PATH=$HOME/.cache/csa.install/lib:$LD_LIBRARY_PATH
+if [ "$TRAVIS_PYTHON_VERSION" == "2.7.13" ]; then
+    export PYTHONPATH=$HOME/.cache/csa.install/lib/python2.7/site-packages:$PYTHONPATH
+    export LD_LIBRARY_PATH=$HOME/.cache/csa.install/lib:$LD_LIBRARY_PATH
+elif [ "$TRAVIS_PYTHON_VERSION" == "3.4.4" ]; then
+    ls /usr/lib/x86_64-linux-gnu/
```
jougs: I assume this line was for debugging. Please remove it.

build.sh
```diff
@@ -240,8 +245,15 @@ echo "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +
...
+    export PYTHONPATH=/usr/lib/x86_64-linux-gnu/:$PYTHONPATH
+    #export PYTHONPATH=$HOME/.cache/csa.install/lib/python3.4/site-packages:$PYTHONPATH
```
jougs: I assume this line was for debugging. Please remove it.

pynest/nest/tests/test_sp/test_disconnect.py
```diff
@@ -21,6 +21,9 @@
 
 import nest
 import unittest
+import numpy as np
+
+__author__ = 'naveau'
```
jougs: Isn't this line duplicated now? Please l.28 see below.

.travis.yml
lekshmideepu: Moved the boost library packages to the list of travis add ons source list which would save the bootstrapping time when compared to the installation from the ubuntu repo using apt-get. As few packages have an older version under the add ons list,  installing those via pip

.travis.yml
lekshmideepu: Done!

build.sh
lekshmideepu: Done!

build.sh
lekshmideepu: Done!

pynest/nest/tests/test_sp/test_disconnect.py
lekshmideepu: Done!

lekshmideepu:
@jougs :Thanks for the comments. Hope I have answered all of them.

PR #804, opened 2017-08-07
Prohibit module installation after CopyModel (fixes #746)

### LINKED ISSUE #746: Segmentation fault if `Install` is called after `CopyModel`
@mhoff and I discovered a strange bug that seems to manifests whenever a custom module is loaded after a call to `CopyModel`. More precisely the following python code fails with a segmentation fault:

```
import nest
nest.CopyModel('stdp_synapse','my_synapse')
nest.Install('mymodule')
```

where `mymodule` is just the example custom module that comes with NEST.

We tested with the latest commit of NEST using `ipython3`on Debian Jessie.

I briefly looked into `ModelManager::register_connection_model_(...)`. It seems to me that the `syn_id` that is assigned to the new synapse models get messed up if the commands are called in that order.

It might make sense to prevent users from calling the commands in that order, but then the function should at least throw an exception in my opinion.

heplesser:
This PR fixes #746. It does not contain a regression test, since that would require building `MyModule`, which is currently not supported by our test harness. I will create a separate issue for that.

### DIFF
--- a/nestkernel/dynamicloader.cpp
+++ b/nestkernel/dynamicloader.cpp
@@ -155,6 +155,13 @@ DynamicLoaderModule::LoadModuleFunction::execute( SLIInterpreter* i ) const
 {
   i->assert_stack_load( 1 );
 
+  if ( kernel().model_manager.has_user_models()
+    or kernel().model_manager.has_user_prototypes() )
+  {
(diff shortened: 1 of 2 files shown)


gtrensch:
Looks good to me.

PR #810, opened 2017-08-21
Let iaf_psc_exp_ps handle simultaneous spikes independently. Fixes #368.

### LINKED ISSUE #368: Incorrect summation of synaptic currents for `iaf_psc_exp_ps`
When two inputs, with equal but opposite weights, converge on an `iaf_psc_exp_ps` neuron with different time constants for the `ex` and `in` synapses, the voltage trace is flat. This problem does not occur with `iaf_psc_exp` or if the inputs arrive at different times.

To reproduce, run the following once with `precise=True`, again with `precise=False`:

``` python
import nest

precise = True
relative_delay = 0.0  # the problem does not occur if this is positive

nest.SetKernelStatus({'off_grid_spiking': precise})

if precise:
    neuron_model = 'iaf_psc_exp_ps'
    filename = "with_ps"
else:
    neuron_model = 'iaf_psc_exp'
    filename = "without_ps"

if relative_delay:
    filename += "_delay"

neuron = nest.Create(neuron_model, 1, {'tau_syn_ex': 12.0, 'tau_syn_in': 4.0, 't_ref': 0.1})

spike_sourceD = nest.Create('spike_generator', 1)
spike_sourceR = nest.Create('spike_generator', 1)
nest.SetStatus(spike_sourceD, 'spike_times', [[40.0]])
nest.SetStatus(spike_sourceR, 'spike_times', [[40.0]])

nest.CopyModel('static_synapse', 'decay_syn', {'weight': 200.0, 'delay': 0.5})
nest.CopyModel('static_synapse', 'rise_syn', {'weight': -200.0, 'delay': 0.5 + relative_delay})

nest.Connect(spike_sourceD, neuron, syn_spec='decay_syn')
nest.Connect(spike_sourceR, neuron, syn_spec='rise_syn')

recorder = nest.Create('multimeter')
nest.SetStatus(recorder, {'withtime': True, 'record_from': ['V_m']})
nest.Connect(recorder, neuron, 'all_to_all')

nest.Simulate(200.0)

data = nest.GetStatus(recorder)[0]['events']
senders = data['senders']
times = data['times']
vm = data['V_m']

for n in neuron:
    plt.plot(times[senders==n], vm[senders==n])

plt.savefig(filename + ".png")
```

_`precise = False`_
without_ps

_`precise = True`_
with_ps

NEST: 2.10.0, Python: 3.4.2

Thanks to Hajime Yamauchi for the original bug report to the NeuralEnsemble Google group.

heplesser:
I intentionally made `accumulate_simultaneous` the second argument of `get_next_spike()`, even though making it the last would have allowed me to use the default value `true`. But I thought it would be cleaner to have the two parameters specifying what to get first, followed by the three parameters passing the return values.

### DIFF
--- a/precise/iaf_psc_alpha_canon.cpp
+++ b/precise/iaf_psc_alpha_canon.cpp
@@ -358,7 +358,7 @@ nest::iaf_psc_alpha_canon::update( Time const& origin,
     bool end_of_refract;
 
     if ( not B_.events_.get_next_spike(
-           T, ev_offset, ev_weight, end_of_refract ) )
+           T, true, ev_offset, ev_weight, end_of_refract ) )
     { // No incoming spikes, handle with fixed propagator matrix.
       // Handling this case separately improves performance significantly
       // if there are many steps without input spikes.
@@ -428,8 +428,8 @@ nest::iaf_psc_alpha_canon::update( Time const& origin,
         V_.y3_before_ = S_.y3_;
         last_offset = ev_offset;
 
-      } while (
-        B_.events_.get_next_spike( T, ev_offset, ev_weight, end_of_refract ) );
+      } while ( B_.events_.get_next_spike(
+        T, true, ev_offset, ev_weight, end_of_refract ) );
 
       // no events remaining, plain update step across remainder
       // of interval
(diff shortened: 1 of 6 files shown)


heplesser:
@apdavison At long last a fix for your issue #368. Would you be willing and able to review it? We would like to include the fix in the NEST 2.12.1 release.

apdavison:
I've installed your branch locally, and I now get the same behaviour for `iaf_psc_exp_ps` and `iaf_psc_exp`, so this looks good to me.

Issue #800, opened 2017-08-04
Record conductance from cond_*_multisynapse models


lucianopaz:
Request
===
I wanted to request an enhancement to nest. In the latest stable release 2.12, it is not possible to connect a multimeter to record from the synaptic conductances of any `cond_*_multisynapse` models. At present, the only way to work around this problem is to simulate chunks of time and call GetStatus for the neurons to manually record `g`. In many situations, it is important to record the synaptic channels evolution over time to understand what drives neurons to fire (for example [decision making with attractor dynamics](http://www.ncbi.nlm.nih.gov/pubmed/16436619)) and not having a built-in method is a serious limitation.

----

Proposal
===

I want to help in any way I can to contribute to solve this issue. I have just started to use nest, so I am not entirely familiar with the source. I know a bit of c++ so I looked through the way data is logged, starting from the `aeif_cond_beta_multisynapse` model, which I am interested in using. From there I saw that the `RecordablesMap` static instance inside the neuron is used for logging data. This class inserts a pointer to a valid function that is a member of the `HostNode` and returns `double`, using template SFINAE. The way in which multiple member functions were instantiated was with templates in the `RecordablesMap::create` function.  This leads to the problem that the number of recordables needs to be known at compile time, and will be static. So apparently, in order to get to record `g` from multisynapse neurons, it will be necessary to rewrite a potentially large part of the kernel.

As I said before, I am not really familiar with nest's source, so I'm putting up for discussion what I think could be done to get to record `g`.

From what I understood (which can be fundamentally wrong, so I would appreciate if any of you point out my mistakes), the minimal list of places to apply changes to be able to record synaptic conductance in multisynapse models are:

[recordables_map.h](https://github.com/nest/nest-simulator/blob/master/nestkernel/recordables_map.h)
 The `RecordablesMap` must be changed at runtime in multisynapse models. This entails many problems. For one, the current template class uses a `DataAccessFct` which is defined as a valid `HostNode` member function that returns double, and is resolved at compile time with template SFINAE. I propose to add a functor at the neuron model level which will be treated as the new DataAccess function. This functor would get the information about the index of the `State.y_` vector to get data from at its construction time. This internal value could then remain `const` for the rest of the functor's lifecycle. The functor's `operator()` would return the recordable's state value. To remain compatible with the rest of nest, we must not touch the standard `RecordablesMap` class template.
[...]

PR #812, opened 2017-08-22
Record conductance from aeif_cond_beta_multisynapse model

lucianopaz:
This pull request fixes issue #800. It implements the DynamicRecordablesMap and a DynamicUniversalDataLogger classes. The only difference with the non-dynamic classes is that the DataAccessFct is a pointer to a Functor instance. This Functor is a public class, whose name is DataAccessFunctor, which is nested inside the HostNode, and has the operator() that returns a double. The DataAccessFunctor class has a state variable that is set upon construction and tells it where to search for any recordable in the HostNode state.

The current implementation only focuses on the aeif_cond_beta_multisynapse model but it could be extended to all multisynapse models. At the model's source, the recordablesMap is no longer static, it is changed when the setState is called on the HostNode, and it includes the receptor port conductance channels "g_1", "g_2", ... , "g_n_receptors".

The unittests/test_aeif_cond_beta_multisynapse.sli was also modified to test that the recordables where responding properly to state changes, and also whether the g channels had beta or alpha function dynamics.


heplesser:
@lucianopaz Thank you for your pull request! I have had a first look and it looks quite promising. Have you done any benchmarks to see whether there is a speed difference between the dynamic and plain Map/Logger? The dynamic approach needs more memory, since each neuron now has to have its own recordables map. But compared to the size of the ring buffer for each synapse, this is rather negligible.

I will get back to you with detailed comments soon. In the meantime, could you send me a signed Contributor License Agreement (http://nest.github.io/nest-simulator/index) by email?

lucianopaz:
@heplesser I sent you the filled contributor license agreement to your email. Sorry, I did not do any benchmarks for the pull request. I will do some now and post the results here.

lucianopaz:
I've run some benchmarks to test how this PR affects time performance and memory usage. I'm uploading a [zip](https://github.com/nest/nest-simulator/files/1284008/benchmarks.zip) with the code I used to run the benchmarks and the results I got.

I compared this PR's performance against the [2.12.0 stable release](https://github.com/nest/nest-simulator/releases/download/v2.12.0/nest-2.12.0.tar.gz). The experiment I ran was to construct a population of aeif_cond_beta_multisynapse of N neurons, with different numbers of n_receptors. All the neurons were independent (i.e. not connected between each other) and received the same poisson train of spikes broadcasted from a parrot neuron to different receptor ports. The receptor ports were taken from a uniform_int distribution. I also connected a single multimeter to all neurons to record from `V_m` and `w` (to be fair with the non PR that is unable to record from any `g`). The multimeter did not write the recordings to disk. I then simulated the network for a total time T.

I compared the build time, the simulation time and the total wallclock time for different N and n_receptors values. T was always 1000ms and I used the same globalRNG seed for all simulations.
As for the memory usage, I measured the peak RSS (Resident Set Size) and Virtual Memory Size (VSZ) using `ps`. Surely, the memory measurements could be improved. I did a single run for each parameter set to get a quick result but I will run the benchmarks averaging over 3 iterations and post the results here later.

I ran all experiments on Ubuntu 16.04 LTS 64bits, kernel 4.10.0-33-generic,  Intel Core i7-5820K CPU 3.30GHz × 12, 31GB RAM and 189MB Swap.

benchmark1

My results show that simulation time is almost not affected (max difference of about ~3%), leading also to an almost negligible difference in total wallclock time. However, buildtime is greatly affected. The PR's build time increases much more rapidly with N and n_receptors than the current release of nest, leading to almost 14 times slower builds for large networks with large numbers of receptors.

As for memory usage both RSS and VSZ are increased for the PR's code. However, the maximum increase in RSS is ~3% and ~0.13% in VSZ, so it does not seem to make a big impact.

heplesser:
@janhahne Would you be able to review this in the near future?

janhahne:
@heplesser I can take a look on thursday. I am however not particular familiar with the code that is changed here.

lucianopaz:
@janhahne Yes, the changes in libnestutil and test_siegert_neuron.py were just unrelated formatting issues that made the automatic style checks fail on my machine.

pynest/nest/tests/test_siegert_neuron.py
```diff
@@ -57,7 +58,6 @@ def setUp(self):
         nest.ResetKernel()
         nest.SetKernelStatus(
             {'resolution': self.dt, 'use_wfr': False, 'print_time': True})
-
```
janhahne: To me it seems odd to to remove this line instead of (old) line 62. The same is true for line 75, where probably line 77 should be removed instead. Could you edit this manually and see if it still passes the sytle check?

lucianopaz:
@janhahne I ran the style check for the changed `test_siegert_neuron.py` file, and it passed successfully. For the initial file version in this PR, I had used [`pep8ify`](https://pypi.python.org/pypi/pep8ify) to make it comply with pep8 styling, but the automatic tool is clearly not perfect.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -454,9 +500,38 @@ aeif_cond_beta_multisynapse::set_status( const DictionaryDatum& d )
   // consistent.
   Archiving_Node::set_status( d );
 
+  /**
+   * Here is where we must update the recordablesMap_ if new receptors
+   * are added!
+   */
```
heplesser: Should be a normal comment, not a Doxygen comment.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -454,9 +500,38 @@ aeif_cond_beta_multisynapse::set_status( const DictionaryDatum& d )
...
+    {
+      size_t elem = aeif_cond_beta_multisynapse::State_::G
+        + receptor * aeif_cond_beta_multisynapse::State_::
+                       NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR;
+      rtmp.insert( get_g_receptor_name( receptor ),
+        aeif_cond_beta_multisynapse::DataAccessFunctor( this, elem ) );
+    }
+  }
+  else if ( ptmp.E_rev.size()
+    < P_.E_rev.size() ) // Number of receptors decreased
```
heplesser: It would be better to have the comparison on one line and the comment on another.

nestkernel/exceptions.cpp
```diff
@@ -408,3 +408,12 @@ nest::NumericalInstability::message() const
       << "updating " << model_ << ".";
   return msg.str();
 }
+
+std::string
+nest::NoEntryToMap::message() const
```
heplesser: The name of the exception seems misleading. Isn' t this a `KeyError`?

nestkernel/exceptions.cpp
```diff
@@ -408,3 +408,12 @@ nest::NumericalInstability::message() const
...
+{
+  std::ostringstream msg;
+  msg << "DynamicRecordablesMap::erase( const Name& n ): "
```
heplesser: With this text, this exception can only be thrown from `DynamicRecordablesMap::erase()`. It is usually not a good idea to constrain exceptions in this way. For greater flexibility, the function throwing the exception can pass its name to the exception.

nestkernel/universal_data_logger.h
```diff
@@ -290,6 +302,273 @@ nest::UniversalDataLogger< HostNode >::DataLogger_::DataLogger_(
   recording_interval_ = req.get_recording_interval();
   recording_offset_ = req.get_recording_offset();
 }
+
+
+/**
+ * Universal data-logging plug-in for neuron models.
+ *
+ * This class provides logging of universal data such as
```
heplesser: You should explain early on the difference between the normal and the dynamic logger.

nestkernel/universal_data_logger.h
```diff
@@ -290,6 +302,273 @@ nest::UniversalDataLogger< HostNode >::DataLogger_::DataLogger_(
...
+ *       implementation is in the companion universal_data_logger_impl.h.
+ *       As a consequence, calls to UniversalDataLogger members should
+ *       only come from cpp files---do not inline them.
+ *
+ * @addtogroup Devices
+ */
+
+
+template < typename HostNode >
+class DynamicUniversalDataLogger
```
heplesser: This class including its internal classes is very closely related to `UniversalDataLogger`. Is it really necessary to implement the class completely from scratch? Can´t it inherit from `UniversalDataLogger` and save almost all the code in the process?

models/aeif_cond_beta_multisynapse.h
```diff
@@ -454,9 +500,38 @@ aeif_cond_beta_multisynapse::set_status( const DictionaryDatum& d )
   // consistent.
   Archiving_Node::set_status( d );
 
+  /**
```
heplesser: This `set_status()` is so large it should not be inlined and should be moved to the cpp-file.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -392,8 +431,15 @@ class aeif_cond_beta_multisynapse : public Archiving_Node
...
+
+  // Access functions for UniversalDataLogger -------------------------------
+
   //! Mapping of recordables names to access functions
-  static RecordablesMap< aeif_cond_beta_multisynapse > recordablesMap_;
+  DynamicRecordablesMap< aeif_cond_beta_multisynapse > recordablesMap_;
+
+  // Utility function that inserts the synaptic conductances to the
+  // recordables map
+  inline void insert_conductance_recordables( size_t first = 0 );
```
heplesser: Remove `inline`! The functions operation is too complex to inline it; plus, modern compilers are quite good at reading our code.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -208,14 +209,62 @@ class aeif_cond_beta_multisynapse : public Archiving_Node
   void get_status( DictionaryDatum& ) const;
   void set_status( const DictionaryDatum& );
 
+  static inline Name
```
heplesser: Do not inline, make private and move implementation to cpp-file.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -208,14 +209,62 @@ class aeif_cond_beta_multisynapse : public Archiving_Node
...
+  get_g_receptor_name( size_t receptor )
+  {
+    std::stringstream receptor_name;
+    receptor_name << "g_" << receptor + 1;
+    return Name( receptor_name.str() );
+  };
+
+  //! Class that reads out state vector elements, used by UniversalDataLogger
+  class DataAccessFunctor
```
heplesser: Does this class need to be public? I am in doubt if it is sensible to declare this class here, useful only for `aeif_cond_beta_multisynapse`. Could one use a template function instead? That would avoid writing a lot of boilerplate code in models.

models/aeif_cond_beta_multisynapse.h
```diff
@@ -208,14 +209,62 @@ class aeif_cond_beta_multisynapse : public Archiving_Node
...
+    };
+
+    DataAccessFunctor( aeif_cond_beta_multisynapse* n, size_t elem )
+      : elem_( elem )
+    {
+      set_parent( n );
+    };
+
+    void
+    set_parent( aeif_cond_beta_multisynapse* parent )
```
heplesser: Drop this method and put the single statement into all places calling this function. And if it were to exist further, it ought to be private!

models/aeif_cond_beta_multisynapse.h
lucianopaz: Done in eb41277

models/aeif_cond_beta_multisynapse.h
lucianopaz: Done in eb41277

models/aeif_cond_beta_multisynapse.h
lucianopaz: Done in eb41277

models/aeif_cond_beta_multisynapse.h
lucianopaz: Done in eb41277

models/aeif_cond_beta_multisynapse.h
lucianopaz: Done in eb41277

models/aeif_cond_beta_multisynapse.h
lucianopaz: Done in eb41277

nestkernel/exceptions.cpp
lucianopaz: Changed to KeyError and made the error message more generic in eb41277

nestkernel/exceptions.cpp
lucianopaz: Changed to KeyError and made the error message more generic in eb41277

models/aeif_cond_beta_multisynapse.h
lucianopaz: You're right. `DataAccessFunctor` should be a class template. I moved its implementation to `recordables_map.h`, because it is used by the `DynamicRecordablesMap` class template.
I also had to change the way in which `DataAccessFunctor` instances were created by the `aeif_cond_beta_multisynapse`, and also how the `DataAccessFunctor` queried the private state of the `HostNode`. Now `aeif_cond_beta_multisynapse` has two new functions `get_data_access_functor` and `get_state_element`. The latter returns the value of the recorded state element. The former creates a `DataAccessFunctor<aeif_cond_beta_multisynapse>`, that is intended as the supplier for the `DynamicRecordablesMap`.

nestkernel/universal_data_logger.h
lucianopaz: This point is getting really complicated for me. Both classes are basically the same. The only difference is that one is constructed from a `RecordablesMap` and the other from a `DynamicRecordablesMap`, and most importantly in what is the type of contents the `node_access_` vector has. When using a `RecordablesMap`, the `node_access_` holds function pointers, and they should be called like `(*(node_access[j])()`. When using a  `DynamicRecordablesMap`, the `node_access_` holds `DataAccessFunctor`s, and they should be called like `node_access()`.
The problem with inheritance is that `DataLogger_` is private and all the important handling of `DataLogger_` should be rewritten for the derived class. I tried with template aliassing but for now without success.

lucianopaz: I spent some time trying to successfully share the code between `UniversalDataLogger` and `DynamicUniversalDataLogger`. I tried two approaches:
1. Make `DynamicUniversalDataLogger` a derived class of `UniversalDataLogger`. In principle this should allow us to use the base class' implementation of `handle`, `recorda_data`, `reset` and `init`. The problem is that all four methods use the member `data_loggers_`. The derived `DynamicUniversalDataLogger` defines its data member `data_loggers_` using its own `DataLogger_` private class, however this only hides the base class `data_loggers_` member. This leads to the problem that in base's `handle` method, the base's empty `data_loggers_` vector is used, instead of the populated derived class' `data_loggers_`. This leads to assertion errors and no recordings.
2. Add an extra template parameter `MapType` to `UniversalDataLogger` that distinguishes between the cases where `RecordablesMap` and `DynamicRecordablesMap` are used. This fails to compile because all neuron models assume that the `UniversalDataLogger` template has only one parameter, instead of two. I failed miserably at making a template specialization alias that forces the `MapType` parameter value. If someone with more experience could help with this, it would be great.

In this situation, I think that code duplication between the `UniversalDataLogger` and `DynamicUniversalDataLogger` classes is not so bad.

nestkernel/universal_data_logger.h
lucianopaz: Done in 81e1178

nestkernel/universal_data_logger.h
heplesser: @lucianopaz Thanks for exploring this! It seems then that we need to live with the code duplication here.

heplesser:
@lucianopaz Your code looks very nice now. Apologies for the late reply. I have one additional request though: could you convert the remaining multisynapse models to the Dynamic logger as well? This would ensure a consistent user experience. With that in place, we would be happy to include the dynamic logger in the NEST 2.16 release.

lucianopaz:
@heplesser, yes of course. I'll start converting the following conductance based models (let me know if I missed one):
- `aeif_cond_alpha_multisynapse`
- `gif_cond_exp_multisynapse`

I wanted to ask what you prefer to do with the current based models:
- `gif_psc_exp_multisynapse`
- `iaf_psc_alpha_multisynapse`
- `iaf_psc_exp_multisynapse`
Should we make the `i_syn_`, and other analogue synaptic current state variables recordable?

heplesser:
@lucianopaz Great! In order to be consistent with what can be recorded from the "plain" `iaf_psc_[exp|alpha]` models, it would be great to make `S_.i_syn_[]` recordable in the current-based multisynapse models. At the same time, I would suggest not to support recording "weighted spikes", since that would require an additional vector to buffer data; I believe that recording option was added to the "plain" versions mostly for debugging purposes.

models/iaf_psc_alpha_multisynapse.cpp
```diff
@@ -44,20 +44,50 @@
...
-RecordablesMap< iaf_psc_alpha_multisynapse >::create()
+DynamicRecordablesMap< iaf_psc_alpha_multisynapse >::create(
+  iaf_psc_alpha_multisynapse& host )
+{
+  // use standard names wherever you can for consistency!
+  insert( names::V_m,
+    host.get_data_access_functor( iaf_psc_alpha_multisynapse::State_::V_M ) );
+
+  insert( names::currents,
+    host.get_data_access_functor( iaf_psc_alpha_multisynapse::State_::I ) );
```
heplesser: I think `currents` was never a good name and it is only used in this one model. I suggest to replace it with `I_syn`.

models/iaf_psc_alpha_multisynapse.cpp
```diff
@@ -44,20 +44,50 @@
...
+
+  host.insert_current_recordables();
+}
+
+Name
+iaf_psc_alpha_multisynapse::get_i_syn_name( size_t elem )
+{
+  std::stringstream i_syn_name;
+  i_syn_name << "i_syn_" << elem + 1;
```
heplesser: For consistency, we should stick to capital I, i.e., `I_syn_x` also for the individual currents.

models/iaf_psc_alpha_multisynapse.h
```diff
@@ -158,6 +162,27 @@ class iaf_psc_alpha_multisynapse : public Archiving_Node
...
+      V_M = 0,
+      I,     // 1
+      I_SYN, // 2
+      STATE_VECTOR_MIN_SIZE
+    };
+
+    static const size_t NUMBER_OF_FIXED_STATES_ELEMENTS = 2; // V_M, I
+    static const size_t NUM_STATE_ELEMENTS_PER_RECEPTOR = 1; // I_SYN
+
+    std::vector< double > y_; //!< neuron state
```
heplesser: Introducing this vector only for the sake of recording seems not a good idea, as it costs both memory and runtime. I would suggest to make the accession functor smarter, along the lines of 
```
if ( elem == V_M )
{ return S_.V_m_ }
else 
{ return S_.y2_syn_[elem/2]; }
```
NB: I am not quite sure if my index arithmetic above is correct, this is just meant as a suggestion.

models/iaf_psc_alpha_multisynapse.h
```diff
@@ -158,6 +162,25 @@ class iaf_psc_alpha_multisynapse : public Archiving_Node
    */
   struct State_
   {
+
+    /**
+     * Enumeration identifying elements in state vector State_::y_.
+     * This enum identifies the elements of the vector. The last element
+     * of this (I_SYN, ) will be repeated n times at the end of the
+     * state vector State_::y with n being the number of synapses.
+     * As a first step this vector will only be used for logging.
```
heplesser: I think the comment would be easier to understand if you gave the expression for the index of the synapses, i.e. (abbreviated), `NUM_FIX_ELEMS + k * NUM_ELEMS_PER_RECEPTOR`.

models/iaf_psc_alpha_multisynapse.h
```diff
@@ -158,6 +162,25 @@ class iaf_psc_alpha_multisynapse : public Archiving_Node
...
+     */
+    enum StateVecElems
+    {
+      V_M = 0,
+      I,     // 1
+      I_SYN, // 2
+      STATE_VECTOR_MIN_SIZE
+    };
+
+    static const size_t NUMBER_OF_FIXED_STATES_ELEMENTS = 2; // V_M, I
```
heplesser: Instead of initializing with the "magic" value 2 here, initialize with `I_SYN` to guarantee consistency. Alternatively, you could replace `I_SYN` in the enum above by `NUM_FIXED_STATE_ELEMENTS`; then you would not need the additional constant here. Maybe one could then also drop the `STATE_VECTOR_MIN_SIZE`.

lucianopaz:
@heplesser, I finished the move to the Dynamic logging for the iaf and aeif multisynapse models, and added some tests for the iaf models to assert that each channel's psc has the correct functional form, in fb5210a. Let me know if you think some extra changes are necessary.

heplesser:
@lucianopaz Thank you very much for your efforts! I will merge now.

lucianopaz:
@heplesser, it was a pleasure to contribute to the project.

PR #815, opened 2017-08-24
Fix #740 by representing times in steps in step_current_generator.

### LINKED ISSUE #740: step_current_generator must require sufficient spacing of amplitude_times
Currently, `amplitude_times` in `step_current_generator` can be set to arbitrary doubles; NEST only checks that they are non-descending. If times are less than resolution apart, the `step_current_generator` will [not advance correctly through the amplitude array](https://github.com/nest/nest-simulator/blob/08e830bfd54943809c76b89e7ddced5f6c8ea5d7/models/step_current_generator.cpp#L245), as it can move at most one array position per time step. 

Given that the output current of the generator can only change at time steps, allowing the user to set `amplitude_times` values that do not coincide with time steps, is misleading.

Therefore, I propose that the generator in the future should only accept `amplitude_times` values that are strictly increasing multiples of the resolution and that are cast to time steps using `Time::ms_stamp()`.

The following example runs once at a high resolution, resolving all amplitude times and once at a low resolution:
```
/run_sim [ /doubletype ]  % resolution, neg power of two
{
  /res Set
  ResetKernel
  0 << /tics_per_ms 1024. /resolution res >> SetStatus
  
  /scg /step_current_generator 
    << /amplitude_times  [ 1. 1.0625 1.125 1.1875 1.25 2. 2.5 ]
       /amplitude_values [ 10. -20.  40.   -80.   160. -320. 640. ] 
    >> Create def
  /nrn /iaf_psc_alpha << /V_th 1e10 /C_m 1.0 /tau_m 1.0
                         /E_L 0.0 /V_m 0.0 /V_reset 0.0 >> Create def
  
  /vm  /voltmeter Create def
  vm << /start 1.75 /stop 4.0 /interval 0.25 >> SetStatus
  
  scg nrn Connect
  vm  nrn Connect

  5 Simulate
  vm [ /events /V_m ] get
}
def

(High resolution: ) =only 2. -5 pow run_sim ==
(Low resolution : ) =only 2. -2 pow run_sim ==
```
yielding
```
High resolution: <. 0.000000e+00 -3.137381e+00 3.294848e+01 6.105218e+01 8.293936e+01 -6.190513e+00 -7.560493e+01 8.268632e+01 2.059637e+02 .>
Low resolution : <. 0.000000e+00 2.211992e+00 -2.701283e+00 -6.527746e+00 -9.507798e+00 -1.182866e+01 -1.363616e+01 1.309476e+02 2.435496e+02 .>
```

heplesser:
@appukuttan-shailesh Could you be the second reviewer for this PR?

### DIFF
--- a/models/step_current_generator.cpp
+++ b/models/step_current_generator.cpp
@@ -28,6 +28,7 @@
 #include "universal_data_logger_impl.h"
 
 // Includes from sli:
+#include "booldatum.h"
 #include "dict.h"
 #include "dictutils.h"
 #include "doubledatum.h"
@@ -51,14 +52,16 @@ RecordablesMap< step_current_generator >::create()
  * ---------------------------------------------------------------- */
 
 nest::step_current_generator::Parameters_::Parameters_()
-  : amp_times_()  // ms
+  : amp_time_stamps_()
   , amp_values_() // pA
+  , allow_offgrid_amp_times_( false )
 {
 }
 
 nest::step_current_generator::Parameters_::Parameters_( const Parameters_& p )
-  : amp_times_( p.amp_times_ )
+  : amp_time_stamps_( p.amp_time_stamps_ )
   , amp_values_( p.amp_values_ )
+  , allow_offgrid_amp_times_( p.allow_offgrid_amp_times_ )
 {
 }
 
@@ -71,8 +74,9 @@ nest::step_current_generator::Parameters_&
     return *this;
   }
 
-  amp_times_ = p.amp_times_;
+  amp_time_stamps_ = p.amp_time_stamps_;
   amp_values_ = p.amp_values_;
+  allow_offgrid_amp_times_ = p.allow_offgrid_amp_times_;
 
   return *this;
 }
@@ -83,13 +87,17 @@ nest::step_current_generator::State_::State_()
 }
 
 nest::step_current_generator::Buffers_::Buffers_( step_current_generator& n )
-  : logger_( n )
+  : idx_( 0 )
+  , amp_( 0 )
+  , logger_( n )
 {
 }
 
 nest::step_current_generator::Buffers_::Buffers_( const Buffers_&,
   step_current_generator& n )
-  : logger_( n )
+  : idx_( 0 )
+  , amp_( 0 )
+  , logger_( n )
 {
 }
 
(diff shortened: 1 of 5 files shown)


appukuttan-shailesh:
Sure, I can take this up.

models/step_current_generator.cpp
```diff
@@ -100,48 +108,126 @@ nest::step_current_generator::Buffers_::Buffers_( const Buffers_&,
...
+  ( *d )[ names::allow_offgrid_times ] = BoolDatum( allow_offgrid_amp_times_ );
+}
+
+nest::Time
+nest::step_current_generator::Parameters_::validate_time_( double t,
+  const Time& t_previous )
+{
+  if ( t == 0.0 )
+  {
+    throw BadProperty( "Amplitude cannot be changed at 0." );
```
appukuttan-shailesh: Preferably update to "Amplitude cannot be changed at t = 0."

appukuttan-shailesh: Preferably update to "Amplitude cannot be changed at t = 0."

heplesser: Fixed.

models/step_current_generator.cpp
```diff
@@ -100,48 +108,126 @@ nest::step_current_generator::Buffers_::Buffers_( const Buffers_&,
...
+    {
+      // In this case, we need to round to the end of the step
+      // in which t lies, ms_stamp does that for us.
+      t_amp = Time::ms_stamp( t );
+    }
+    else
+    {
+      std::stringstream msg;
+      msg << "step_current_generator: Time point " << t
+          << " is not representable in current resolution.";
```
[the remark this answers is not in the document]
heplesser: I decided to combine this with the t==0 check: if t <= 0, the user now gets the following message: "Amplitude can only be changed at strictly positive times (t > 0)."

heplesser:
@jougs One use case would be that the user uses random numbers for the times at which amplitudes change. These would not most likely not fall onto the grid. As a convenience for the user, NEST can round these times, but it should only do this if explicitly requested to avoid subtle errors otherwise. 

Should I put this into the inline documentation, or did you just want an explanation here in the discussion?

jougs:
@heplesser: thanks for the explanation. I think it won't hurt if you added something along those lines to the documentation. Thanks for the nice work!

heplesser:
@jougs Done, hope it is ready to merge now!

PR #822, opened 2017-09-07
Fix problem with iaf_psc_delta_canon and test_dcgen_vs_I_e.sli

heplesser:
This fixes a problem reported by SDSC in which `test_dcgen_vs_I_e.sli` fails for `iaf_psc_delta_canon` for certain hardware/compiler combinations. The failure is due to the fact that that model treated `I_e` and injected current slightly differently. This PR ensures that both forms of current input are treated perfectly symmetrical (always appear as sum `P_.I_e_ + S_.I_`).

It also tidies up the test code and removes calls to `executive`, which put SLI into interactive mode in case of a test failure.

@jougs I will ask someone from SDSC to review as well. Since this is a straightforward fix, I add it to 2.14.

### DIFF
--- a/precise/iaf_psc_delta_canon.cpp
+++ b/precise/iaf_psc_delta_canon.cpp
@@ -58,7 +58,7 @@ template <>
 void
 RecordablesMap< iaf_psc_delta_canon >::create()
 {
-  // use standard names whereever you can for consistency!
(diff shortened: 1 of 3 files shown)


heplesser:
Fixed typos as requested by @hakonsbm .

PR #827, opened 2017-09-19
Moved vacant sp elements sanity check to decay function

sdiazpier:
This PR follows up on #818. The check on the vacant synaptic elements is moved to the decay function of the synaptic elements to ensure that the deletion of elements still takes place correctly by the sp_manager. 

I suggest @sanjayankur31  as reviewer for this.

### DIFF
--- a/nestkernel/synaptic_element.h
+++ b/nestkernel/synaptic_element.h
@@ -172,7 +172,7 @@ public:
   int
   get_z_vacant() const
   {
-    return std::max( std::floor( z_ ) - z_connected_, 0. );
+    return std::floor( z_ ) - z_connected_;
   }
   /*
    * Retrieves the current number of synaptic elements bound to a synapse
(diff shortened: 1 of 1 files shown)


sanjayankur31:
:+1: 

Looks good! That should work much better! Thanks @sdiazpier

heplesser:
@sanjayankur31 @sdiazpier After this change, `Archiving_Node::get_synaptic_elements_vacant()` may return a negative value. Is that sensible?

@sanjayankur31 Please use the "Review changes" button in the "Files changed" tab to approve or disapprove---searching for 👍  in discussions can be tedious.

sanjayankur31:
@heplesser - the way structural plasticity is implemented, `ArchivingNode::get_synaptic_elements_vacant() < 0` implies that connections need to be deleted. It's used in `SPManager::get_synaptic_elements()`. (I hadn't realised this either which is why my earlier fix in #818 was incomplete).

I'll use the review tools. Sorry about that!

PR #829, opened 2017-09-20
Correction in the test of synapse disconnection with mpi

sdiazpier:
This PR solves an issue with the test of synapse deletion using mpi and pynest. Thanks to @lekshmideepu who implemented the solution in PR #789 but we have decided to split this change as this problem is not directly related to the issue in that PR.

### DIFF
--- a/pynest/nest/tests/test_sp/test_disconnect.py
+++ b/pynest/nest/tests/test_sp/test_disconnect.py
@@ -19,10 +19,11 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-__author__ = 'naveau'
-
 import nest
 import unittest
+import numpy as np
+
+__author__ = 'naveau'
 
 try:
     from mpi4py import MPI
@@ -32,6 +33,7 @@ except ImportError:
 else:
     # Test with MPI
     mpi_test = 1
+mpi_test = nest.sli_func("statusdict/have_mpi ::") & mpi_test
 
 
 class TestDisconnectSingle(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


heplesser:
@sdiazpier Could you check why the tests are failing on Travis?

pynest/nest/tests/test_sp/test_disconnect.py
```diff
@@ -19,10 +19,11 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-__author__ = 'naveau'
-
 import nest
 import unittest
+import numpy as np
```
hakonsbm: Numpy isn't used in this test.

pynest/nest/tests/test_sp/test_disconnect.py
```diff
@@ -52,12 +54,6 @@ def setUp(self):
             'stdp_dopamine_synapse_hpc_lbl',
             'gap_junction',
             'gap_junction_lbl',
-            'diffusion_connection',
-            'diffusion_connection_lbl',
-            'rate_connection_instantaneous',
-            'rate_connection_instantaneous_lbl',
-            'rate_connection_delayed',
-            'rate_connection_delayed_lbl'
```
hakonsbm: Removing these makes the test fail because these synapse-models aren't supported by `iaf_neuron`/`iaf_psc_alpha`.

sanjayankur31: Confirming - these need to be checked and added back to the exclude list

sdiazpier:
Dear @hakonsbm and @sanjayankur31 ,  sorry for the delay to address all these comments. I hope I have solved all of them. Please let me know if you have any additional comments or suggested changes.

terhorstd:
@sanjayankur31 can you have another look? Thanks a lot!

PR #831, opened 2017-09-26
Docker image for the NEST simulator

steffengraber:
Dockerfiles for building Docker images with NEST 2.12.0 and jupyter notebook. To get started, follow the instructions in the README.md.


tammoippen:
Hi @steffengraber ,
I really like the idea of a NEST on docker and I have a few suggestions:
* Every `RUN` command creates a new layer on the image - hence increasing the total image size. You should reduce the number of them as far as possible, eg

```Dockerfile
RUN wget https://bootstrap.pypa.io/get-pip.py && \
         python get-pip.py && \
         pip install \
              setuptools \
              path.py \
              pandas \
              cython==0.23.4 && \
        ...
```

See [Best Practice](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers)

* when you install build only dependencies / have intermediate files ( like `get-pip.py`) they should be removed in the same `RUN` command, otherwise they are 'baked' into that layer and increase the image size, even if you remove them in a later `RUN` command.

* `pip` has an option `--no-cache-dir` [here](https://pip.pypa.io/en/stable/reference/pip_install/#caching) that also reduces the image size, as no cache is being kept. ( Also, I think python3 has pip3 installed by default / or there are apt-get pip packages...)

Will the image be available on https://hub.docker.com/ ?

steffengraber:
H @tammoippen  
thanks for your hints. With so many 'RUN' commands, error handling is easier. But you're right,  better to clean up the files.
I am very much in favor of having an own dockerimage on dockerhub.  This would reduce the entry level.

DimitriPlotnikov:
Currently, it doesn't work for me. I'm trying to run a `docker-master` as described in the document, but I get the following error:
```
nestml@nestml-VirtualBox:~/nest-simulator/extras/docker$ docker run -it --rm --user nest --name my_app -v ~/nest_scripts:/home/nest/data -p 8080:8080 nest/docker-master /bin/bash
docker: Error response from daemon: linux spec user: unable to find user nest: no matching entries in passwd file.
```

However, I can login into the container if I omit the `--user` adn `--name` key.

Finally, it is quite error prone to tip all this command by hand. Could you add a more highlevel management script to work with docker  (cf. https://github.com/nest/nestml/blob/master/docker/nestml_docker.sh)?

extras/docker/README.md
```diff
@@ -0,0 +1,77 @@
...
+    
+    You are logged in as user 'nest'. Enter 'python3' and in the 
+    python-shell 'import nest'. A 'nest.help()' should display the main 
+    help page.
+
+## Docker 
+
+-   Delete ALL images (use with caution)
+
+        docker rmi $(docker images -a -q)
```
DimitriPlotnikov: I'm using the following command to really delete everything:
`docker stop $(docker ps -a -q) && docker rm $(docker ps -a -q) && docker rmi $(docker images -q)`

tammoippen: Actually there is a docker command for this:
```
docker system prune -fa --volumes
```

See [here](https://docs.docker.com/engine/reference/commandline/system_prune/) for documentation.

DimitriPlotnikov: @tammoippen 
I haven't ever used this command. It seems reasonable. Apparently I have had problems with active/used/running docker images which were not removed. I'm not sure how the `prune` command deals this scenario. My proposal is, of course, an ultimate solution and should be used with caution. Especially, if there are other images then only nest.

DimitriPlotnikov:
Any updates?

steffengraber:
@DimitriPlotnikov I will work on it in the next days.

steffengraber:
Excuse me, excuse me. I have now made even bigger changes to the docker files. This includes the installation and use of Conda for the Python environment, which increases the size of the image, but allows the user a more flexible use of different programming environments. 
Python 3 is standard now. This causes problems when creating MUSIC from the sources. That's why I used the prebuilt binaries from Ubuntu. 
I tried a lot to reduce the image size. Unfortunately without success.

jougs:
@tammoippen, @DimitriPlotnikov: are you available to review this? If not, please respond briefly so we can find other reviewers.

DimitriPlotnikov:
Hello, I'm currently busy.

extras/docker/nest-2.12.0/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+ARG WITH_MPI
+ARG WITH_GSL
+ARG WITH_MUSIC
+ARG WITH_LIBNEUROSIM
+
+# add user 'nest'
+RUN adduser --disabled-login --gecos 'NEST' --home /home/nest nest && \
+    adduser nest sudo
+
+# Install NEST 2.14.0
```
tammoippen: NEST 2.12.0

extras/docker/nest-2.12.0/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    tar -xvzf nest-2.12.0.tar.gz && \
+    mkdir /home/nest/nest-build && \
+    mkdir /home/nest/nest-install
+
+WORKDIR /home/nest/nest-build
+
+# Install NEST 2.12.0
+
+RUN cmake -DCMAKE_INSTALL_PREFIX:PATH=/home/nest/nest-install \
+    -Dwith-python=2 \
```
tammoippen: Isn't there a python3 environment via conda from nest/docker-master:latest? Why not use that?

extras/docker/nest-2.14.0/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+RUN cmake -DCMAKE_INSTALL_PREFIX:PATH=/home/nest/nest-install \
+    -Dwith-python=3 \
+    -Dwith-mpi:BOOL=$WITH_MPI \
+    -Dwith-gsl:BOOL=$WITH_GSL /usr/local/lib \
+    -Dwith-libneurosim:BOOL=$WITH_LIBNEUROSIM /usr/local/lib \
+    -Dwith-music=$WITH_MUSIC /usr/lib \
+    ../nest-simulator-2.14.0
+
+RUN  make && make install && \
```
tammoippen: Why have this as a separate layer? In the 2.12 file this `RUN` and the previous are joined.

extras/docker/nest-2.14.0/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    echo '. /home/nest/nest-install/bin/nest_vars.sh' >> /home/nest/.bashrc
+
+# RUN su nest -c 'make installcheck'
+
+WORKDIR /home/nest/
+
+USER root
+RUN rm -rf /home/nest/nest-build && rm /home/nest/v2.14.0.tar.gz
```
tammoippen: As this is a new layer in the Dockerfile (each `RUN` is a layer), removing the build artifacts and the tar will not reduce the image size.

extras/docker/master/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    libopenmpi-dev \
+    vera++ \
+    pep8 \
+    libpcre3 \
+    libpcre3-dev \
+    jq \
+    libmusic1v5 \
+    music-bin \
+    libmusic-dev && \
+    apt-get autoremove -y
```
tammoippen: Remove apt cache as well? `&& rm -rf /var/lib/apt/lists/*`

extras/docker/master/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+WORKDIR /tmp
+RUN git clone https://github.com/INCF/libneurosim.git libneurosim
+WORKDIR /tmp/libneurosim
+RUN chmod +x autogen.sh
+RUN ./autogen.sh && \
+    chmod +x configure && \
+    ./configure && make &&  make install
+
+EXPOSE 8080
+RUN apt-get autoremove && rm -rf /tmp/*
```
tammoippen: Duplicate? See line 29. There are no other `apt-get install` calls, or? Also, cleaning commands should happen in the same layer (`RUN` command) where the 'waste' was generated, otherwise the 'waste' is burned into that layer and cannot be removed in later layers.

steffengraber:
Thanks to your tips and some more changes  I was able to reduce the size of the images significantly (each ~1.2GB). Additionally NEST 2.16.0 is now integrated.

steffengraber:
This activity is transfered to a seperate repository.

Issue #488, opened 2016-09-19
Discussion: A new interface for parameterization and spatially structured networks in NEST


heplesser:
## Background

NEST currently has two different interfaces for specifying parameters to NEST functions, one used by `Connect` and `Create`, the other used by `topology.ConnectLayers`. The latter is quite strongly based on concepts in the Topology module. This duplicity is unfortunate and should be removed.

Work on NEST optimization has shown that memory locality is crucial to excellent performance, especially in multithreaded code. Therefore, we have decided to remove `subnet`s from NEST. This will also affect the Topology Module, since it currently manages layers as subnets with metadata.

Both changes will lead to changes in the NEST User Interface. It therefore seems sensible to introduce them with NEST 3.
## The Purpose of Parameters

With parameters, in NEST I here mean quantities used to specify values in neuron models, synapses, connection, or that parameterize connection patterns:
- provide model parameters, such as time constants, conductances, ...
- provide initial values, such as membrane potentials, ...
- provide connection weights and delays
- provide connection probabilities 

Parameters may be given as
- fixed values
- random values
- arrays of values
- functions returning context-dependent values, e.g., weights depending on the distance of neurons

In the most general way, all these types of parameters can be considered as functions taking information about the network as input and returning values.
### Core NEST: `Create()` and `Connect()`

Currently,
- fixed scalars
- random scalars from a fixed set of distributions
- arrays

are supported. Parameters cannot depend on context (e.g., GID), and cannot be combined into expressions. Arrays are indexed sequentially, mapping array locations in the parameter array to corresponding arrays of GIDs.

Random scalars are specified as follows:

``` Python
{'distribution': 'uniform', 'low': 0.8, 'high': 2.5}
```
### Topology: `CreateLayers`

Currently,
- fixed scalars
- random scalars from a fixed set of distributions
- a fixed set of distance-dependent functions

are supported. Functions are specified in a topology-specific syntax, e.g.,

``` Python
{'gaussian': {'p_center': 1.0, 'sigma': 0.5}}
{'linear': {'c': 1.0, 'a': -0.05, 'cutoff': 0.0}}
{'uniform': {'min': 0.5, 'max': 2.0}}
```

Parameters in Topology can be combined to create composite parameters, e.g.,

``` Python
lin = topo.CreateParameter('linear', {'c': 1.0, 'a': -0.05, 'cutoff': 0.0})
uni = topo.CreateParameter('uniform', {'min': 0.5, 'max': 2.0})
[...]

PR #836, opened 2017-10-04
Allow negative exponential distributions (fixes #835)

### LINKED ISSUE #835: Can't build an exponential distribution of inhibitory synaptic weights
Hi!

Not sure if this is a bug, a missing feature, by design or me misunderstanding. If I should direct this somewhere else, don't hesitate to let me know and close the issue.

Some models rely on the synaptic weight being negative in order to consider the synapse inhibitory (see [here](https://github.com/nest/nest-simulator/blob/49d917723e743188a48f187ab07ca8ccc8679b3a/models/aeif_cond_exp.cpp#L558)) but the exponential distribution doesn't allow a negative lambda value (see [here](https://github.com/nest/nest-simulator/blob/15ac89b24fd08e9323dcdfe06061193bb3a6c11b/librandom/exp_randomdev.cpp#L38)). So, how can then an exponential distribution of inhibitory synaptic weights be achieved?

From the documentation ([here](http://www.nest-simulator.org/connection-management/)) I expected to be able to do:
```python
nest.Connect(
    nodes_in, nodes_ex,
    {'rule': 'fixed_indegree', 'indegree': N},
    {'model': 'static_synapse', 'weight': {'distribution': 'exponential', 'lambda': -J}})
```

Cheers,
Luis

luis-rr:
Some models rely on the synaptic weight being negative in order to consider the synapse inhibitory.
However the exponential distribution wasn't allowing a negative lambda value.
With this change we can generate exponential distributions of inhibitory synaptic weights.

Note that the implementation in librandom/exp_randomdev.h already works for negative lambda values, this is just removing the check for negative parameter values.

### DIFF
--- a/librandom/exp_randomdev.h
+++ b/librandom/exp_randomdev.h
@@ -40,7 +40,9 @@ namespace librandom
 Name: rdevdict::exponential - exponential random deviate generator
 Description: Generates exponentially distributed random numbers.
 
-  p(x) = lambda exp(-lambda*x), x >= 0.
+  p(x) =  lambda exp(-lambda*x), x >= 0 if lambda > 0
+  p(x) = -lambda exp(-lambda*x), x <= 0 if lambda < 0
+
 
 Parameters:
  lambda - rate parameter (default: 1.0)
--- a/librandom/exp_randomdev.h
+++ b/librandom/exp_randomdev.h
@@ -39,10 +39,15 @@ namespace librandom
 /*BeginDocumentation
 Name: rdevdict::exponential - exponential random deviate generator
 Description: Generates exponentially distributed random numbers.
+Negative values of lambda are allowed and generate a distribution of negative numbers.
 
-  p(x) =  lambda exp(-lambda*x), x >= 0 if lambda > 0
-  p(x) = -lambda exp(-lambda*x), x <= 0 if lambda < 0
+For lambda > 0:
+  p(x) = lambda exp(-lambda*x), for x >= 0
+  p(x) = 0, for x < 0
 
+For lambda < 0:
+  p(x) = 0, for x > 0
+  p(x) = |lambda| exp ( -|lambda| |x| ), for x <= 0
 
 Parameters:
  lambda - rate parameter (default: 1.0)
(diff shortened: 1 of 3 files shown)


heplesser:
Please also send me a signed contributor license agreement (see http://nest.github.io/nest-simulator/index) by direct email.

librandom/exp_randomdev.h
```diff
@@ -40,7 +40,9 @@ namespace librandom
 Name: rdevdict::exponential - exponential random deviate generator
 Description: Generates exponentially distributed random numbers.
 
-  p(x) = lambda exp(-lambda*x), x >= 0.
+  p(x) =  lambda exp(-lambda*x), x >= 0 if lambda > 0
+  p(x) = -lambda exp(-lambda*x), x <= 0 if lambda < 0
```
heplesser: I would write this differently for clarity: Add a sentence above pointing out explicitly that negative lambda are permitted and then yield negative numbers, and then
```
For lambda < 0:
  p(x) = 0 for x >= 0
  p(x) = |lambda| exp ( -|lambda| |x| ) for x < 0
```
Since we are coercing the usual notation here, we should be pretty explicit. Ideally, we would want to draw a positive x and then multiply that with -1, but that will only become possible in the future (see #488 for ideas under consideration).

librandom/exp_randomdev.h
```diff
@@ -39,8 +39,15 @@ namespace librandom
 /*BeginDocumentation
 Name: rdevdict::exponential - exponential random deviate generator
 Description: Generates exponentially distributed random numbers.
+Negative values of lambda are allowed and generate a distribution of negative numbers.
```
heplesser: @luis-rr This line is longer than 79 characters, so it fails our code formatting checks. Could you fix it?

luis-rr: Just sent another commit.

Out of curiosity, how did you identify the error? I'm looking at Travis' log and all I can see is:
MSGBLD0220: C/C++ files with formatting errors:
MSGBLD0220: ... librandom/exp_randomdev.cpp
MSGBLD0220: ... librandom/exp_randomdev.h

But not the description of the formatting error itself.

PR #844, opened 2017-10-24
Install header files

tammoippen:
This PR make nest install its header files into `${CMAKE_INSTALL_PREFIX}/includes/nest`. This allows the development of user modules, without having the nest sources around (cf. email on NEST_USER: ` [NEST_USER] NEST Homebrew formula: copy headers to install directory`).

For this purpose I modified:

* the output of `nest-config --includes` via the `ALL_INCLUDES_tmp` in the base `CMakeList.txt`
* added `FILTER_HEADERS` cmake function that gets all headers from the `XXX_sources` variables
* added the `install( FILE .... )` instructions for the headers
* added missing header files (`genericmodel_impl.h`) to the sources
* adjusted `examples/MyModule/CMakeLists.txt`, such that installations into none-default locations of user modules is more explicit; with further instructions to make nest aware of them (see `NEST_MODULE_PATH` and `SLI_PATH`)

With the headers installed and travis-ci installation into `results` (within the source path) the regression test `ticket-659-copyright.py` failed for some installed headers. In the source, they are explicitly  ignored (cf. `ticket-659-copyright.py:52ff`), hence I ignored the `results` directory as well.

I tested with cmake 2.8.12, 3.5.1 (ubuntu 16.04) and 3.9.4 (OS X). On the ubuntu system I encountered unrelated compilation issues with gsl and fixed them with 2411f5e. I also tested building and loading `MyModule` installed into a custom location.


tclose:
Thanks Tammo, the changes work great.

One small, tenuously related issue that might be able to be bundled together with these changes is that when building NEST inside a Python 3 virtual env, the help files don't build because they aren't Python 3 compatible. Do you know where in the make files this is specified and whether the command could be explicitly changed to call `python2` instead of `python`

tammoippen:
Hi @tclose ,

Have a look at https://github.com/nest/nest-simulator/blob/master/doc/CMakeLists.txt#L26 .
Here one should require python 2 version, like `find_package( PythonInterp 2.7 REQUIRED )`.
BUT one should also check, that this will not mess up the nest installation in a mixed python 2/3 environment, since the `doc` sub dir is [called first](https://github.com/nest/nest-simulator/blob/master/CMakeLists.txt#L217)... the better solution would be to make the doc-generation python 2 and 3 compatible. I think this should definitely be a separate PR. This is also an interesting use case for #853 .

CMakeLists.txt
```diff
@@ -284,13 +284,7 @@ string( REPLACE ";" " " ALL_LIBS "${ALL_LIBS}" )
...
 # all includes
 set( ALL_INCLUDES_tmp
-  "${PROJECT_SOURCE_DIR}/libnestutil"
-  "${PROJECT_BINARY_DIR}/libnestutil"
-  "${PROJECT_SOURCE_DIR}/librandom"
-  "${PROJECT_SOURCE_DIR}/sli"
-  "${PROJECT_SOURCE_DIR}/topology"
-  "${PROJECT_SOURCE_DIR}/nestkernel"
-  "${PROJECT_SOURCE_DIR}/nest"
+  "${CMAKE_INSTALL_FULL_INCLUDEDIR}/nest"
```
Silmathoron: Why remove these? Shouldn't a variable starting with ``ALL_INCLUDES`` indeed contain all include folders? ^^

examples/MyModule/CMakeLists.txt
```diff
@@ -187,10 +184,14 @@ execute_process(
 # on OS X
 set( CMAKE_MACOSX_RPATH ON )
 
-# Install all stuff to NEST's install directories.
-set( CMAKE_INSTALL_LIBDIR ${NEST_LIBDIR}/nest CACHE STRING "object code libraries (lib/nest or lib64/nest or lib/<multiarch-tuple>/nest on Debian)" FORCE )
-set( CMAKE_INSTALL_DOCDIR ${NEST_INSTALL_DOCDIR} CACHE STRING "documentation root (DATAROOTDIR/doc/nest)" FORCE )
-set( CMAKE_INSTALL_DATADIR ${NEST_INSTALL_DATADIR} CACHE STRING "read-only architecture-independent data (DATAROOTDIR/nest)" FORCE )
+if ( "${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr/local" )
```
Silmathoron: Comments and spacing look weird to me, could you change to

```cmake
# Use the NEST_INSTALL_* variables as CMAKE_INSTALL_*, if not set explicitly.
if ( "${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr/local" )
  set( CMAKE_INSTALL_PREFIX "${NEST_INSTALL_PREFIX}" CACHE STRING "Install path prefix, prepended onto install directories." FORCE )
  set( CMAKE_INSTALL_LIBDIR ${NEST_LIBDIR}/nest CACHE STRING "object code libraries (lib/nest or lib64/nest or lib/<multiarch-tuple>/nest on Debian)" FORCE )
  set( CMAKE_INSTALL_DOCDIR ${NEST_INSTALL_DOCDIR} CACHE STRING "documentation root (DATAROOTDIR/doc/nest)" FORCE )
  set( CMAKE_INSTALL_DATADIR ${NEST_INSTALL_DATADIR} CACHE STRING "read-only architecture-independent data (DATAROOTDIR/nest)" FORCE )
endif ()
```

?

I am also a little skeptical about what happens if ``"${CMAKE_INSTALL_PREFIX}"`` is not ``"/usr/local"``... how do we know all other ``CMAKE_INSTALL_*`` variables are correct? Shouldn't we check all?

examples/MyModule/CMakeLists.txt
```diff
@@ -264,19 +265,19 @@ set_target_properties( ${MODULE_NAME}_lib
...
   add_custom_target( generate_help ALL )
   # Extract help from all source files in the source code, put them in
-  # doc/help and generate a local help index in the build directory containing 
+  # doc/help and generate a local help index in the build directory containing
   # links to the help files.
   add_custom_command( TARGET generate_help POST_BUILD
     COMMAND python -B generate_help.py "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}"
     COMMAND python -B generate_helpindex.py "${PROJECT_BINARY_DIR}/doc"
-    WORKING_DIRECTORY "${CMAKE_INSTALL_PREFIX}/${NEST_INSTALL_DATADIR}/help_generator"
+    WORKING_DIRECTORY "${NEST_INSTALL_PREFIX}/${NEST_INSTALL_DATADIR}/help_generator"
```
Silmathoron: I'm making a note for myself that this is indeed correct AND necessary so that I do not forget later ^^"

sli/CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ target_link_libraries( sli_readline sli_lib nestutil ${READLINE_LIBRARIES} )
 
 # add the executable
 add_executable( sli puresli.cc )
-target_link_libraries( sli sli_lib sli_readline )
+target_link_libraries( sli sli_lib sli_readline ${GSL_LIBRARIES} )
```
Silmathoron: why is this addition necessary?

CMakeLists.txt
tammoippen: The `ALL_INCLUDES` variable is only used in [`nest-config`](https://github.com/nest/nest-simulator/blob/master/extras/nest-config.in#L61). Here we expect to get locations of installed header files. The `${PROJECT_SOURCE_DIR}/...` are only available, as long as the NEST sources are there, but (here) we install all headers to `"${CMAKE_INSTALL_FULL_INCLUDEDIR}/nest"`, hence remove all source related paths and have only the install path.

examples/MyModule/CMakeLists.txt
tammoippen: > ... "${CMAKE_INSTALL_PREFIX}" is not "/usr/local" ...

When `CMAKE_INSTALL_PREFIX` is not set with `cmake -DCMAKE_INSTALL_PREFIX=...` then it is set to `/usr/local`. This is the same reasoning we apply to check if nest will be installed to the default location: https://github.com/nest/nest-simulator/blob/master/CMakeLists.txt#L97 There was also an issue discussing this behaviour in NEST, but i cannot find it anymore ...

The other `CMAKE_INSTALL_*` are set with `include( GNUInstallDirs )` later, see [here](https://cmake.org/cmake/help/v3.4/module/GNUInstallDirs.html)

I will apply the requested changes.

sli/CMakeLists.txt
tammoippen: This is related to:

> I tested with cmake 2.8.12, 3.5.1 (ubuntu 16.04) and 3.9.4 (OS X). On the ubuntu system I encountered unrelated compilation issues with gsl and fixed them with 2411f5e. I also tested building and loading `MyModule` installed into a custom location.

Apparently, cmake 2.8.12 cannot do the transitive library discovery:
```
nestutil requires gsl
sli requires nestutil, hence slightly requires gsl
```
and does not put gsl into linkage with sli...

tclose:
Okay, no worries. A separate PR for building the help pages makes sense. Although I just managed to build NEST in a Python 3 virtualenv on a different machine so I am wondering whether the error I was seeing was due to something strange in my setup (it looked very much like a Python 2/3 error though). Checking that the build is Python 3 compatible is probably a good idea in any case though.

tclose:
Is there anything holding up this PR? I am hoping to prepare a Homebrew formula for my Pype9 package but need these changes in the master.

Silmathoron:
Yep, we're waiting for the review of @jougs  ;P

tclose:
Thanks all, that is great.

Do you have a rough ETA of when these changes will make it into a stable release? Will there be a 2.14.1?

PR #845, opened 2017-10-25
Allow rectangular and box masks to be rotated

stinebuu:
This PR adds the optional parameters `azimuth_angle` and `polar_angle` to the rectangular and box mask dictionaries so that the masks can be rotated.

This have led to some changes in the `boxMask` class, most notably with the `inside(..)` function and when finding the boundary box.  When checking if a point is inside the mask, we have to take rotation into account. If we have rotation, we can no longer use `lower_left` and `upper_right` to define the boundary box, so we have to calculate the minimum and maximum values of the mask. This is done in `create_min_max_values_()`.

The functions check to see if we have rotation as much as possible, so that we can use `lower_left` and `upper_right` and avoid unnecessary rotation calculations if we do not have rotation.

I have also added a test.

### DIFF
--- a/topology/mask.h
+++ b/topology/mask.h
@@ -210,12 +210,14 @@ class BoxMask : public Mask< D >
 public:
   /**
    * Parameters that should be in the dictionary:
-   * lower_left  - Position of lower left corner (array of doubles)
-   * upper_right - Position of upper right corner (array of doubles)
+   * lower_left    - Position of lower left corner (array of doubles)
+   * upper_right   - Position of upper right corner (array of doubles)
+   * azimuth_angle - Rotation angle in degrees from x-axis (double), optional
+   * polar_angle   - Rotation angle in degrees from z-axis (double), optional
    */
   BoxMask( const DictionaryDatum& );
 
-  BoxMask( const Position< D >& lower_left, const Position< D >& upper_right );
+  BoxMask( const Position< D >& lower_left, const Position< D >& upper_right, double azimuth_angle = 0.0, double polar_angle = 0.0 );
 
   ~BoxMask()
   {
@@ -250,8 +252,23 @@ public:
   static Name get_name();
 
 protected:
+  /**
+   *  Calculate the min/max x, y, z values in case of a rotated box.
+   */
+  void create_min_max_values_();
+
   Position< D > lower_left_;
   Position< D > upper_right_;
+
+  Position< D > min_values_;
+  Position< D > max_values_;
+
+  double azimuth_angle_;
+  double polar_angle_;
+  double azimuth_cos_;
+  double azimuth_sin_;
+  double polar_cos_;
+  double polar_sin_;
 };
 
 /**
@@ -692,20 +709,66 @@ BoxMask< D >::BoxMask( const DictionaryDatum& d )
 {
   lower_left_ = getValue< std::vector< double > >( d, names::lower_left );
   upper_right_ = getValue< std::vector< double > >( d, names::upper_right );
+
   if ( not( lower_left_ < upper_right_ ) )
   {
     throw BadProperty(
       "topology::BoxMask<D>: "
       "Upper right must be strictly to the right and above lower left." );
   }
+
+  if ( d->known( names::azimuth_angle ) )
+  {
+    azimuth_angle_ = getValue< double >( d, names::azimuth_angle );
+  }
+  else
+  {
+    azimuth_angle_ = 0.0;
+  }
+
+  if ( d->known( names::polar_angle ) )
+  {
+    if ( D == 2 )
+    {
+      throw BadProperty(
+        "topology::BoxMask<D>: "
+        "polar_angle not defined in 2D." );
+    }
+    polar_angle_ = getValue< double >( d, names::polar_angle );
+  }
+  else
+  {
+    polar_angle_ = 0.0;
+  }
+
+  azimuth_cos_ = std::cos( azimuth_angle_ * numerics::pi / 180. );
+  azimuth_sin_ = std::sin( azimuth_angle_ * numerics::pi / 180. );
+  polar_cos_ = std::cos( polar_angle_ * numerics::pi / 180. );
+  polar_sin_ = std::sin( polar_angle_ * numerics::pi / 180. );
+
+  create_min_max_values_();
 }
 
 template < int D >
 inline BoxMask< D >::BoxMask( const Position< D >& lower_left,
-  const Position< D >& upper_right )
+  const Position< D >& upper_right, double azimuth_angle, double polar_angle )
   : lower_left_( lower_left )
   , upper_right_( upper_right )
+  , azimuth_angle_ ( azimuth_angle )
+  , polar_angle_ ( polar_angle )
+  , azimuth_cos_( std::cos( azimuth_angle_ * numerics::pi / 180. ) )
+  , azimuth_sin_( std::sin( azimuth_angle_ * numerics::pi / 180. ) )
+  , polar_cos_( std::cos( polar_angle_ * numerics::pi / 180. ) )
+  , polar_sin_( std::sin( polar_angle_ * numerics::pi / 180. ) )
 {
+  if ( D == 2 and not( polar_angle_ == 0.0 ) )
+  {
+    throw BadProperty(
+      "topology::BoxMask<D>: "
+      "polar_angle not defined in 2D." );
+  }
+
+  create_min_max_values_();
 }
 
 template <>
--- a/topology/mask.h
+++ b/topology/mask.h
@@ -219,8 +219,8 @@ public:
 
   BoxMask( const Position< D >& lower_left,
     const Position< D >& upper_right,
-    double azimuth_angle = 0.0,
-    double polar_angle = 0.0 );
+    const double azimuth_angle = 0.0,
+    const double polar_angle = 0.0 );
 
   ~BoxMask()
   {
@@ -759,8 +759,8 @@ BoxMask< D >::BoxMask( const DictionaryDatum& d )
 template < int D >
 inline BoxMask< D >::BoxMask( const Position< D >& lower_left,
   const Position< D >& upper_right,
-  double azimuth_angle,
-  double polar_angle )
+  const double azimuth_angle,
+  const double polar_angle )
   : lower_left_( lower_left )
   , upper_right_( upper_right )
   , azimuth_angle_( azimuth_angle )
--- a/topology/mask.h
+++ b/topology/mask.h
@@ -217,7 +217,10 @@ public:
    */
   BoxMask( const DictionaryDatum& );
 
-  BoxMask( const Position< D >& lower_left, const Position< D >& upper_right, double azimuth_angle = 0.0, double polar_angle = 0.0 );
+  BoxMask( const Position< D >& lower_left,
+    const Position< D >& upper_right,
+    double azimuth_angle = 0.0,
+    double polar_angle = 0.0 );
 
   ~BoxMask()
   {
@@ -751,11 +754,13 @@ BoxMask< D >::BoxMask( const DictionaryDatum& d )
 
 template < int D >
 inline BoxMask< D >::BoxMask( const Position< D >& lower_left,
-  const Position< D >& upper_right, double azimuth_angle, double polar_angle )
+  const Position< D >& upper_right,
+  double azimuth_angle,
+  double polar_angle )
   : lower_left_( lower_left )
   , upper_right_( upper_right )
-  , azimuth_angle_ ( azimuth_angle )
-  , polar_angle_ ( polar_angle )
+  , azimuth_angle_( azimuth_angle )
+  , polar_angle_( polar_angle )
   , azimuth_cos_( std::cos( azimuth_angle_ * numerics::pi / 180. ) )
   , azimuth_sin_( std::sin( azimuth_angle_ * numerics::pi / 180. ) )
   , polar_cos_( std::cos( polar_angle_ * numerics::pi / 180. ) )
--- a/topology/mask.h
+++ b/topology/mask.h
@@ -258,7 +258,7 @@ protected:
   /**
    *  Calculate the min/max x, y, z values in case of a rotated box.
    */
-  void create_min_max_values_();
+  void calculate_min_max_values_();
 
   Position< D > lower_left_;
   Position< D > upper_right_;
@@ -753,7 +753,7 @@ BoxMask< D >::BoxMask( const DictionaryDatum& d )
 
   is_rotated_ = azimuth_angle_ or polar_angle_;
 
-  create_min_max_values_();
+  calculate_min_max_values_();
 }
 
 template < int D >
@@ -779,7 +779,7 @@ inline BoxMask< D >::BoxMask( const Position< D >& lower_left,
 
   is_rotated_ = azimuth_angle_ or polar_angle_;
 
-  create_min_max_values_();
+  calculate_min_max_values_();
 }
 
 template <>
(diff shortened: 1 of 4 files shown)


topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        maskdict = {'lower_left': [-1.5, -0.5],
+                    'upper_right': [1.5, 0.5],
+                    'polar_angle': 45.0}
+        with self.assertRaises(nest.NESTError):
+            mask = topo.CreateMask('rectangular', maskdict)
+
+    def test_RotatedBoxMask(self):
+        """Test rotated box mask with azimuth angle."""
+        # Test a 3D layer.
+        pos = [[x*1., y*1., z*1.] for x in range(-2, 3)
```
hakonsbm: Missing whitespace around `*`-operators.

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        # Test rotation with an azimuth angle of 135 degrees.
+        maskdict = {'lower_left': [-1.5, -0.5],
+                    'upper_right': [1.5, 0.5],
+                    'azimuth_angle': 135.0}
+        mask = topo.CreateMask('rectangular', maskdict)
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+
+        self.assertEqual(gid_list, (8, 14, 20,))
+
+        # Test that an error is rased if we send in a polar angle to a 2D mask.
```
hakonsbm: Raised, not "rased".

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        mask = topo.CreateMask('box', maskdict)
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+        sorted_gid_list = sorted(gid_list)
+
+        self.assertEqual(sorted_gid_list, [58, 59, 60, 63, 64, 65, 68, 69, 70])
+
+    def test_RotatedBoxMaskByPolarAngle(self):
+        """Test rotated box mask with polar angle."""
+
+        pos = [[x*1., y*1., z*1.] for x in range(-2, 3)
```
hakonsbm: Missing whitespace around `*`-operators.

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        self.assertEqual(gid_list, (8, 14, 20,))
+
+        # Test that an error is rased if we send in a polar angle to a 2D mask.
+        maskdict = {'lower_left': [-1.5, -0.5],
+                    'upper_right': [1.5, 0.5],
+                    'polar_angle': 45.0}
+        with self.assertRaises(nest.NESTError):
+            mask = topo.CreateMask('rectangular', maskdict)
+
+    def test_RotatedBoxMask(self):
```
hakonsbm: Should perhaps be named `test_RotatedBoxMaskByAzimuthAngle`?

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        maskdict = {'lower_left': [-0.5, -1.5, -1.5],
+                    'upper_right': [0.5, 1.5, 1.5],
+                    'polar_angle': 45.}
+        mask = topo.CreateMask('box', maskdict)
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+        sorted_gid_list = sorted(gid_list)
+
+        self.assertEqual(sorted_gid_list, [33, 38, 43, 59, 64, 69, 85, 90, 95])
+
+        # Test with a polar angle of 135 degrees. This should be opposite of
```
hakonsbm: Perhaps perpendicular, not "opposite".

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        sorted_gid_list = sorted(gid_list)
+
+        self.assertEqual(sorted_gid_list, [38, 39, 40, 63, 64, 65, 88, 89, 90])
+
+        self.assertEqual(sorted_gid_list_1, sorted_gid_list)
+
+    def test_RotatedBoxMaskByAzimuthAndPolarAngle(self):
+        """Test rotated box mask with azimuth and polar angle."""
+
+        pos = [[x*1., y*1., z*1.] for x in range(-2, 3)
```
hakonsbm: Missing whitespace around `*`-operators.

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+        mask = topo.CreateMask('box', maskdict)
+        cntr = [0., 0., 0.]
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+        sorted_gid_list = sorted(gid_list)
+
+        self.assertEqual(sorted_gid_list,
+                         [38, 39, 44, 58, 59, 64, 69, 70, 84, 89, 90])
+
+    def test_RotatedRectangleOutsideOrigo(self):
+        """Test rotated rectangle where mask is outside origo."""
```
hakonsbm: As it is the mask that contains things, not origo, perhaps something along the lines of "where origo is not in mask"?

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+                    'azimuth_angle': 90.0}
+        mask = topo.CreateMask('rectangular', maskdict)
+        gid_list = topo.SelectNodesByMask(layer, cntr, mask)
+
+        self.assertEqual(gid_list, (81, 82, 83, 84, 92, 93, 94, 95,))
+
+    def test_RotatedBoxOutsideOrigo(self):
+        """Test rotated box where mask is outside origo."""
+
+        pos = [[x*1., y*1., z*1.] for x in range(-2, 3)
```
hakonsbm: Missing whitespace around `*`-operators.

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+            5  10 15  20  25            31  36  41  46  51
+            6  11 16  21  26            32  37  42  47  52
+
+            some example connections will be:
+
+            2  -> 28
+
+                          44
+            14 ->     40
+                  36
```
hakonsbm: Looks like there's something missing with these example connections.

topology/pynest/tests/test_rotated_rect_mask.py
```diff
@@ -0,0 +1,406 @@
...
+
+        for conn, conn_ref in zip(connections, ref):
+            conn_list = [conn[0], conn[1]]
+            self.assertEqual(conn_list, conn_ref)
+
+
+def suite():
+    suite = unittest.makeSuite(RotatedRectangularMask, 'test')
+    return suite
+
```
hakonsbm: Two blank lines after the function definition.

topology/mask_impl.h
```diff
@@ -87,40 +87,316 @@ Mask< D >::outside( const Box< D >& b ) const
...
+      ( lower_left_[ 2 ] - cntr_z ) * polar_cos_;
+    const double lower_left_polar_sin =
+      ( lower_left_[ 2 ] - cntr_z ) * polar_sin_;
+    const double upper_right_polar_cos =
+      ( upper_right_[ 2 ] - cntr_z ) * polar_cos_;
+    const double upper_right_polar_sin =
+      ( upper_right_[ 2 ] - cntr_z ) * polar_sin_;
+
+    double rotatedLLLx =
+      ( lower_left_cos[ 0 ] - lower_left_sin[ 1 ] ) * polar_cos_
```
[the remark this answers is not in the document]
stinebuu: @hakonsbm I decided not to do this because in addition to these two, I would also have to precalculate 6 other values ( `( upper_right_cos[ 0 ] - lower_left_sin[ 1 ] ) * polar_cos_`, ` upper_right_cos[ 0 ] - lower_left_sin[ 1 ] ) * polar_sin_`,  `( upper_right_cos[ 0 ] - upper_right_sin[ 1 ] ) * polar_cos_` etc..), and this will lead to a lot of precalculated values in addition to the several that are already in the function. As these particular values are only used twice, I think it is cleaner to leave it as is. If you still think it necessary, I can change it.

topology/mask.h
```diff
@@ -250,8 +255,25 @@ class BoxMask : public Mask< D >
...
 protected:
+  /**
+   *  Calculate the min/max x, y, z values in case of a rotated box.
+   */
+  void calculate_min_max_values_();
+
   Position< D > lower_left_;
   Position< D > upper_right_;
+
+  Position< D > min_values_;
```
heplesser: Could you add a comment what these are and how they relate to lower_left/upper_right?

topology/mask.h
```diff
@@ -210,12 +210,17 @@ class BoxMask : public Mask< D >
 public:
   /**
    * Parameters that should be in the dictionary:
-   * lower_left  - Position of lower left corner (array of doubles)
-   * upper_right - Position of upper right corner (array of doubles)
+   * lower_left    - Position of lower left corner (array of doubles)
+   * upper_right   - Position of upper right corner (array of doubles)
+   * azimuth_angle - Rotation angle in degrees from x-axis (double), optional
```
heplesser: Is it documented somewhere that the angles are in degrees? Maybe also add that the polar angle does not apply to 2D.

topology/mask.h
```diff
@@ -692,20 +714,72 @@ BoxMask< D >::BoxMask( const DictionaryDatum& d )
...
+  , polar_sin_( std::sin( polar_angle_ * numerics::pi / 180. ) )
 {
+  if ( D == 2 and not( polar_angle_ == 0.0 ) )
+  {
+    throw BadProperty(
+      "topology::BoxMask<D>: "
+      "polar_angle not defined in 2D." );
+  }
+
+  is_rotated_ = azimuth_angle_ or polar_angle_;
```
heplesser: Make explicit comparisons to 0!

topology/mask_impl.h
```diff
@@ -87,40 +87,318 @@ Mask< D >::outside( const Box< D >& b ) const
...
+  const double new_x = ( p[ 0 ] - cntr_x ) * azimuth_cos_
+    + ( p[ 1 ] - cntr_y ) * azimuth_sin_ + cntr_x;
+  const double new_y = -( p[ 0 ] - cntr_x ) * azimuth_sin_
+    + ( p[ 1 ] - cntr_y ) * azimuth_cos_ + cntr_y;
+  const Position< 2 > new_p( new_x, new_y );
+
+  // We need to add a small epsilon in case of rounding errors.
+  const double x_length = upper_right_[ 0 ] - lower_left_[ 0 ];
+  const double y_length = upper_right_[ 1 ] - lower_left_[ 1 ];
+  const Position< 2 > eps( x_length * 1e-12, y_length * 1e-12 );
```
heplesser: We can use the `cntr_*` values to determine `eps`, saving some operations.

topology/mask_impl.h
```diff
@@ -87,40 +87,318 @@ Mask< D >::outside( const Box< D >& b ) const
...
+  // If we have a rotated box, we rotate the point down to the unrotated box,
+  // and check if it is inside said unrotated box.
+
+  // The new x, y values are calculated using a rotation matrix:
+  // [new_x, new_y] = R(-azimuth)*[x - x_c, y - y_c]
+  // where R(-t) = [cos(t) sin(t); -sin(t) cos(t)]
+
+  // See https://en.wikipedia.org/wiki/Rotation_matrix for more.
+
+  const double cntr_x = ( upper_right_[ 0 ] + lower_left_[ 0 ] ) * 0.5;
```
heplesser: If I am not entirely off, we need to check quite often if a point is inside a (fixed) box. Wouldn't it make more sense to precompute all values that do not depend on `p` in the box constructor?

topology/mask_impl.h
```diff
@@ -87,40 +87,318 @@ Mask< D >::outside( const Box< D >& b ) const
...
+
+  // The new x, y, z values are calculated using rotation matrices:
+  // [new_x, new_y, new_z] =
+  //       R_y(-polar)*R_z(-azimuth)*[x - x_c, y - y_c, z - z_c]
+  // where R_z(-t) = [cos(t) sin(t) 0; -sin(t) cos(t) 0; 0 0 1] and
+  //       R_y(-t) = [cos(t) 0 -sin(t); 0 1 0; sin(t) 0 cos(t)]
+
+  // See https://en.wikipedia.org/wiki/Rotation_matrix for more.
+
+  const double cntr_x = ( upper_right_[ 0 ] + lower_left_[ 0 ] ) * 0.5;
```
heplesser: Same comments apply as for 2D.

topology/mask_impl.h
```diff
@@ -87,40 +87,318 @@ Mask< D >::outside( const Box< D >& b ) const
...
+    const double cntr_x = ( upper_right_[ 0 ] + lower_left_[ 0 ] ) * 0.5;
+    const double cntr_y = ( upper_right_[ 1 ] + lower_left_[ 1 ] ) * 0.5;
+    const Position< 2 > cntr( cntr_x, cntr_y );
+
+    Position< 2 > lower_left_cos = ( lower_left_ - cntr ) * azimuth_cos_;
+    Position< 2 > lower_left_sin = ( lower_left_ - cntr ) * azimuth_sin_;
+    Position< 2 > upper_right_cos = ( upper_right_ - cntr ) * azimuth_cos_;
+    Position< 2 > upper_right_sin = ( upper_right_ - cntr ) * azimuth_sin_;
+
+    double rotatedLLx = lower_left_cos[ 0 ] - lower_left_sin[ 1 ] + cntr_x;
```
heplesser: I suppose these can call be `const`?

topology/mask_impl.h
```diff
@@ -87,40 +87,318 @@ Mask< D >::outside( const Box< D >& b ) const
...
+      ( lower_left_[ 2 ] - cntr_z ) * polar_sin_;
+    const double upper_right_polar_cos =
+      ( upper_right_[ 2 ] - cntr_z ) * polar_cos_;
+    const double upper_right_polar_sin =
+      ( upper_right_[ 2 ] - cntr_z ) * polar_sin_;
+
+    double rotatedLLLx =
+      ( lower_left_cos[ 0 ] - lower_left_sin[ 1 ] ) * polar_cos_
+      - lower_left_polar_sin + cntr_x;
+    double rotatedLLLy = lower_left_sin[ 0 ] + lower_left_cos[ 1 ] + cntr_y;
```
heplesser: Can't these all be const?

stinebuu:
@heplesser thank you for your review! I have tried to address your comments, please take a look.

I changed the "eps" to be constant `1e-12` for x, y and z. Having it depend on the center of the box will fail when the center is in the origin, as eps then will be zero.

When it comes to the rotation angles, we are missing rotation from the y-axis to the z-axis (the &#947; [here](https://en.wikipedia.org/wiki/Euler_angles#/media/File:Eulerangles.svg) ). Do you want me to add it? We do not have it for the elliptical mask, but I could add it there as well and make a PR of course.

topology/mask.h
stinebuu: I have now added a note about the angles in the python documentation for masks.

topology/mask.h
stinebuu: @heplesser I have added the comment. The min/max values correspond to the lower_left/upper_right of the boundary box around the rotated box. Should I change the names to something like `bbox_lower_left_` and `bbox_upper_right_` to make it more clear?

heplesser:
@stinebuu Thanks for addressing the issues. I think this ready to merge now. Would you create a follow-up issue to add rotation about the third Euler angle?

PR #846, opened 2017-10-25
Fixes for the gif multisynapse and quantal_stp model

drodarie:
This request fixes some problems in the gif_cond_exp_multisynapse model relating to the resizing of the neuron state and a minor bug in the quantal_stp model (using thread id instead of virtual process id).

Marc-Oliver Gewaltig
Dimitri Rodarie
Till Schumann

### DIFF
--- a/models/gif_cond_exp_multisynapse.cpp
+++ b/models/gif_cond_exp_multisynapse.cpp
@@ -127,7 +127,7 @@ nest::gif_cond_exp_multisynapse::Parameters_::Parameters_()
   , V_reset_( -55.0 )  // mV
   , Delta_V_( 0.5 )    // mV
   , V_T_star_( -35 )   // mV
-  , lambda_0_( 1. )    // 1/s
+  , lambda_0_( 0.001 ) // 1/ms
   , t_ref_( 4.0 )      // ms
   , c_m_( 80.0 )       // pF
   , tau_stc_()         // ms
@@ -390,8 +390,8 @@ nest::gif_cond_exp_multisynapse::State_::set( const DictionaryDatum& d,
 {
   updateValue< double >( d, names::V_m, y_[ V_M ] );
   y_.resize( State_::NUMBER_OF_FIXED_STATES_ELEMENTS
-                + State_::NUM_STATE_ELEMENTS_PER_RECEPTOR * p.n_receptors(),
-                0.0 );
+      + State_::NUM_STATE_ELEMENTS_PER_RECEPTOR * p.n_receptors(),
+    0.0 );
 
   sfa_elems_.resize( p.tau_sfa_.size(), 0.0 );
   stc_elems_.resize( p.tau_stc_.size(), 0.0 );
@@ -478,8 +478,10 @@ void
 nest::gif_cond_exp_multisynapse::init_buffers_()
 {
   B_.spikes_.resize( P_.n_receptors() );
-  for (size_t i =0; i < P_.n_receptors(); ++i)
-    B_.spikes_[i].clear();   // includes resize
+  for ( size_t i = 0; i < P_.n_receptors(); ++i )
+  {
+    B_.spikes_[ i ].clear(); // includes resize
+  }
 
   B_.currents_.clear(); //!< includes resize
   B_.logger_.reset();   //!< includes resize
@@ -618,8 +620,8 @@ nest::gif_cond_exp_multisynapse::update( Time const& origin,
 
     for ( size_t i = 0; i < P_.n_receptors(); i++ )
     {
-      S_.y_[ State_::G + ( State_::NUM_STATE_ELEMENTS_PER_RECEPTOR
-                           * i ) ] += B_.spikes_[ i ].get_value( lag );
+      S_.y_[ State_::G + ( State_::NUM_STATE_ELEMENTS_PER_RECEPTOR * i ) ] +=
+        B_.spikes_[ i ].get_value( lag );
     }
 
     if ( S_.r_ref_ == 0 ) // neuron is not in refractory period
--- a/models/gif_cond_exp_multisynapse.cpp
+++ b/models/gif_cond_exp_multisynapse.cpp
@@ -131,7 +131,7 @@ nest::gif_cond_exp_multisynapse::Parameters_::Parameters_()
   , t_ref_( 4.0 )      // ms
   , c_m_( 80.0 )       // pF
   , tau_stc_()         // ms
-  , q_stc_()           // nA
+  , q_stc_()           // pA
   , tau_sfa_()         // ms
   , q_sfa_()           // mV
   , tau_syn_( 1, 2.0 ) // ms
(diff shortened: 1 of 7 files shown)


models/quantal_stp_connection.h
```diff
@@ -212,7 +212,7 @@ Quantal_StpConnection< targetidentifierT >::send( Event& e,
   // Compute number of sites that recovered during the interval.
   for ( int depleted = n_ - a_; depleted > 0; --depleted )
   {
-    if ( kernel().rng_manager.get_rng( vp )->drand() < ( 1.0 - p_decay ) )
+    if ( kernel().rng_manager.get_rng( t )->drand() < ( 1.0 - p_decay ) )
     {
```
mschmidt87: You can now remove line 200 of this file where the virtual process id is retrieved since it is now longer needed.

models/gif_cond_exp_multisynapse.h
```diff
@@ -118,14 +118,14 @@
...
-                                  after each spike emission in nA.
+                                  after each spike emission in pA.
     tau_stc    vector of double - Time constants of stc variables in ms.
     q_sfa      vector of double - Values added to spike-frequency adaptation
                                   (sfa) after each spike emission in mV.
     tau_sfa    vector of double - Time constants of sfa variables in ms.
     Delta_V    double - Stochasticity level in mV.
-    lambda_0   double - Stochastic intensity at firing threshold V_T in 1/s.
-    V_T_star   double - Base threshold in mV
+    lambda_0   double - Stochastic intensity at firing threshold V_T in 1/ms.
```
mschmidt87: In NEST, spike rates are generally defined in spikes/s. Unless I misunderstood the neuron model, lambda_0 is a spike rate. I therefore find it more logical to keep it in 1/s.

models/gif_cond_exp_multisynapse.cpp
```diff
@@ -521,15 +523,6 @@ nest::gif_cond_exp_multisynapse::init_buffers_()
 void
 nest::gif_cond_exp_multisynapse::calibrate()
 {
-
```
mschmidt87: Could you explain your changes at this point in more detail? Do I understand correctly that you can remove this here because the state vectors and buffers are now correctly resized in `set` and `init_buffers_`?

models/gif_cond_exp_multisynapse.cpp
```diff
@@ -143,8 +143,7 @@ nest::gif_cond_exp_multisynapse::Parameters_::Parameters_()
 }
 
 nest::gif_cond_exp_multisynapse::State_::State_( const Parameters_& p )
-  : y_( STATE_VEC_SIZE, 0.0 )
-  , size_neuron_state_( 0 )
+  : y_( STATE_VEC_SIZE + NUMBER_OF_STATES_ELEMENTS_PER_RECEPTOR, 0.0 )
```
mschmidt87: Minor comment: I would shorten this variable name to 'NUM_STATE_ELEMENTS_PER_RECEPTOR' . If you agree, then this should also be changed in 'aeif_cond_alpha_multisynapse' for consistency.

models/gif_cond_exp_multisynapse.h
drodarie: Ok, I will bring lambda's units back to 1/s.

models/gif_cond_exp_multisynapse.cpp
drodarie: I don't understand what you want me to remove. Is it in the calibrate function ?
In my commit, I fixed the initialization of the state vectors and move them to the correct functions.
I have also fixed the spike buffers initialization which was not adapted to multiple synapses.

models/gif_cond_exp_multisynapse.cpp
drodarie: It is not the Adaptive Exponential Integrate and Fire model (aeif) but rather the Generalized Leeky Integrate and Fire model (glif or gif)

models/gif_cond_exp_multisynapse.cpp
mschmidt87: No need to remove anything, I just wanted some more explanation. It is fine for me now.

models/gif_cond_exp_multisynapse.cpp
mschmidt87: Thanks. I am aware which neuron model you have worked on in this PR. What I meant was: If you change the name of this variable to my proposal in this neuron model (`gif_cond_exp_multisynapse`) (which you did), then we should also change the name accordingly in the other multisynapse neuron models, which are `aeif_cond_alpha_multisynapse` and `aeif_cond_beta_multisynapse`, if I am not mistaken (I simply did  `grep NUMBER_OF_` in the models directory). In `aeif_cond_beta_multisynapse` I am referring to the variable `NUMBER_OF_FIXED_STATES_ELEMENTS`, which could be changed to `NUM_FIXED_STATE_ELEMENTS`.

This could also be done in a separate PR, but since the changes are very small, I would opt for simply adding it this PR, though in a separate commit.

drodarie: Sorry, I misunderstood your first comment. I will do the change in the other models.

drodarie:
I understand your point, the code is hence more understandable but then you have to convert lambda_0 at each update (except refractory period).
Should we prefer clarity over optimization ?

mschmidt87:
No, you are right, it is better to do this conversion only when the number is set or retrieved. Other models do it this way, too, so the current solution is better.

stinebuu:
@drodarie could you please run `extras/check_code_style.sh` and fix the problems? This will make travis happy.

models/gif_cond_exp_multisynapse.cpp
```diff
@@ -127,11 +127,11 @@ nest::gif_cond_exp_multisynapse::Parameters_::Parameters_()
   , V_reset_( -55.0 )  // mV
   , Delta_V_( 0.5 )    // mV
   , V_T_star_( -35 )   // mV
-  , lambda_0_( 0.001 ) // 1/ms
+  , lambda_0_( 1. )    // 1/s
```
heplesser: This must remain in milliseconds, since the value is used as
```
const double lambda = P_.lambda_0_ * std::exp( ( S_.V_ - S_.sfa_ ) / P_.Delta_V_ );
...
-numerics::expm1( -lambda * Time::get_resolution().get_ms() )
```
i.e., it is multiplied by a value in milliseconds.

The user provides/retrieves `lambda_0` in 1/s; the value is converted to the internal units of 1/ms in `Parameters_::get()/set()`.  This is the standard approach in NEST.

models/gif_cond_exp_multisynapse.h
```diff
@@ -240,7 +240,7 @@ class gif_cond_exp_multisynapse : public Archiving_Node
     double V_reset_;
     double Delta_V_;
     double V_T_star_;
-    double lambda_0_; /** 1/ms */
+    double lambda_0_; /** 1/s */
```
heplesser: Must remain 1/ms, see above.

heplesser:
@drodarie Thank you!

PR #850, opened 2017-10-31
Make Connector properties and growth factor configurable 

jougs:
This PR replaces #507. For explanations and discussion see the comments there.


nestkernel/nest_names.cpp
```diff
@@ -238,6 +238,7 @@ const Name index_map( "index_map" );
 const Name individual_spike_trains( "individual_spike_trains" );
 const Name inh_conductance( "inh_conductance" );
 const Name init_flag( "init_flag" );
+const Name init_connector_capacity( "init_connector_capacity" );
```
heplesser: To keep alphabetical order, this one should come before `init_flag`.

nestkernel/connection_manager.cpp
```diff
@@ -134,7 +134,7 @@ void
 nest::ConnectionManager::set_status( const DictionaryDatum& d )
 {
   long init_cap = init_conn_capacity_;
-  if ( updateValue< long >( d, "init_connector_capacity", init_cap ) )
+  if ( updateValue< long >( d, names::init_connector_capacity, init_cap ) )
```
heplesser: I think I would change this name to `initial_connector_capacity`.  Just `init_...` evokes too strong associations to the verb "initialize" and may therefore be confusing.

heplesser:
Merging now, since @mschmidt87 had approved precursor PR #507 previously and this PR added only minor fixed. Thanks, @jougs and @alexeyshusharin!

PR #857, opened 2017-11-14
 Improve handling of output vars from FindXXX.cmake

tammoippen:
I encountered a different and consistent issue with handling `XXX_LIBRARIES` and `XXX_INCLUDE_DIRS` in the `FindXXX.cmake` files: they are set to `XXX_LIBRARIES-NOTFOUND` via `find_package_handle_standard_args`, which might result in the error:

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LTDL_LIBRARIES (ADVANCED)
    linked by target "nestkernel" in directory /Users/nest/nest-simulator/nestkernel

-- Configuring incomplete, errors occurred!
```

=> You cannot compile nest.

This PR applies a more consistent handling:
* Always check for singular version of `XXX_LIBRARY`, `XXX_INCLUDE_DIR`
* before calling `find_package_handle_standard_args` assign the `XXX_LIBRARY` to their plural form `XXX_LIBRARIES` - these will be the variables external users (i.e. nest) interact with. 

This way `XXX_LIBRARY` will have the `NOTFOUND` value and the user code can use `XXX_LIBRARIES` without issues (as is recommended here: https://cmake.org/cmake/help/git-master/manual/cmake-developer.7.html#find-modules).

Further add OS X specific information about libltdl in the ConfigureSummary. (Thanks to @tomvierjahn .)


tomvierjahn:
@tammoippen, the "set to NOTFOUND" warning is annoying, indeed. However, setting `XYZ_LIBRARIES` and `XYZ_INCLUDE_DIRS` *before* calling `find_package_handle_standard_args` seems unusual to me. `find_package_handle_standard_args` checks if all the requirements are met. According to the CMake documentation, you set the said variables after the call, if the package has been found correctly.

For `cmake/FindLTDL.cmake` this'll lead to
```CMake
...
find_path( LTDL_INCLUDE_DIR
    NAMES ltdl.h
    HINTS ${LTDL_ROOT_DIR}/include
    )
find_library( LTDL_LIBRARY
    NAMES ltdl
    HINTS ${LTDL_ROOT_DIR}/lib
    )
...
include( FindPackageHandleStandardArgs )
find_package_handle_standard_args( LTDL
  FOUND_VAR
    LTDL_FOUND
  REQUIRED_VARS
    LTDL_LIBRARY
    LTDL_INCLUDE_DIR
  VERSION_VAR
    LTDL_VERSION
    )

if (LTDL_FOUND)
  set( LTDL_INCLUDE_DIRS "${LTDL_INCLUDE_DIR}" )
  set( LTDL_LIBRARIES "${LTDL_LIBRARY}" )
endif()
```

tomvierjahn:
Despite of my previous comment, the output if libtool is missing on Mac looks good to me. Thanks!

tammoippen:
@tomvierjahn This sounds valid! And shows, how documentation and implementation can differ / diverge. When I wrote the changes, I oriented on `FindGSL.cmake` from cmake: they set those variables before the `find_package_handle_standard_args`, see here https://github.com/Kitware/CMake/blob/master/Modules/FindGSL.cmake#L118 . I would suggest to change this in all our files, except the `FindGSL.cmake` to keep similarity to the original.

tammoippen:
@tomvierjahn I have applied the suggested changes. Please have a look.

jougs:
Many thanks to all involved people. I think we can waive the review by @gtrensch given the review by @tomvierjahn. @abigailm, can you please merge if you agree with my assessment?

PR #864, opened 2017-12-06
Fix test_disconnect by converting filter object to list for len

jakobj:
This test fails for Python 3 if filter objects are not converted to lists before calling `len`.

I suggest @jougs as reviewer


heplesser:
Trivial fix to test, merging without further review.

PR #865, opened 2017-12-06
Add STDP synapses with nearest-neighbor spike pairing schemes

aserenko:
Add three types of nearest-neighbour STDP: symmetric (but not to be
confused with PR #218), presynaptic-centered, and restricted
symmetric. These are worth implementing because described in @abigailm's 
review [Morrison A., Diesmann M., and Gerstner W. (2008)
Phenomenological models of synaptic plasticity based on spike timing,
Biol. Cybern. 98, 459--478].

In these models a spike is taken into account in the weight change
rule not with all preceding spikes, but only with some of the
nearest, see the three pairing schemes in Fig. 7 in the aforementioned paper
[Morrison et al., 2008].

The implementation relies on two additional variables - presynaptic
and postsynaptic traces, like in stdp_synapse. However, their update
rules differ from those of the latter. So, the archiving_node
was modified to add one more state variable, nearest_neighbor_Kminus.
It is not stored in the node, just computed on the fly when
requested. To return nearest_neighbor_Kminus I changed the
archiving_node.get_K_values() function, because, as fas as
I can see, this function is not currently used anywhere.

The tests for all three models reside in a single file (but as three
separate tests), and are the following: generate two Poisson spike
sequences as presynaptic and postsynaptic, feed them to NEST to get
the synaptic weight change, then reproduce the weight change
independently outside of NEST and check whether the expected weight
equals the obtained one.

I suggest **I:Internal API** label, because I tamper the archiving_node.get_K_values() function signature (its existence and its name made me think that adding more get_whatever_values() functions for specific models is discouraged). And I would appreciate the reviewers checking me in whether this function has indeed not been used anywhere yet.


aserenko:
By the way, there was probably a mistake in Dr. Morrison's paper, in the equation describing the presynaptic trace update for the presynaptic-centered pairing scheme (page 12 of the article or page 470 of the whole volume, section 4.1.2 "Spike pairing scheme").

Since a postsynaptic spike is paired not with the nearest preceding presynaptic spike, but with all preceding presynaptic ones till the previous postsynaptic one, the presynaptic trace should be increased **by** 1 (like in all-to-all fashion), rather than **to** 1, on a pre-spike occurrence.

I have mailed Dr. Gerstner, the corresponding author, and got a confirmation. However, another check by the adorable @abigailm would do no harm.

aserenko:
I sincerely apologize to all the NEST team and to the reviewers personally for being possibly impolite due to my lack of vocabulary. Thank you for keeping the simulator open to external contributions, even to students with poor language skills trying to do their bit.

jougs:
@aserenko: can you please have a look at the failing tests? Thanks!

aserenko:
Just a couple of complaints by clang-format. Sorry for that, @jougs.

I've also applied the same correction of comparison to zero that was done to other STDP-like plasticity models in PR #924.

nestkernel/archiving_node.cpp
```diff
@@ -113,16 +113,18 @@ nest::Archiving_Node::get_K_value( double t )
...
 nest::Archiving_Node::get_K_values( double t,
   double& K_value,
+  double& nearest_neighbor_K_value,
   double& triplet_K_value )
 {
   // case when the neuron has not yet spiked
   if ( history_.empty() )
   {
     triplet_K_value = triplet_Kminus_;
+    nearest_neighbor_K_value = Kminus_;
```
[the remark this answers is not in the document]
aserenko: If I understand the question right, `nearest_neighbor_K_value` is set the same as `K_value` only if the history is empty. The meaningful case is in line 137, is it?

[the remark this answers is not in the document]
jakobj: ah, OK, I didn't check the context below. thanks for clarifying

aserenko:
I've merged changes introduced into master by 5G connection infrastructure. The reason is, the changes that had been made to stdp_connection.h had to be manually reproduced in the equivalent code of stdp_nn_*_connection.h models. Otherwise, despite the "mergeable" and "tests passed" green ticks, this branch would instantly fail the tests when merged into master because of relying on outdated connector implementations.

aserenko:
Oh, please wait a little before reviewing. I see my own test script fail.

aserenko:
I get an error during unittests/test_common_props_setting.sli:
> CopyModel cannot generate another synapse. Maximal synapse model count of 63 exceeded.

It does not look related to my newly added synapse models. Still, merging the up-to-date master hasn't solved the issue. Could somebody be so kind to advise me what may be happening?

heplesser:
@aserenko NEST is simply running out of space for synapse models. I will create a separate issue to discuss that. As a work-around, in `unittests/test_common_props_setting.sli`, in
```
  synmodels
  {
    /synmod Set 
```
place a `ResetKernel` before `/synmod Set` and in the first and the second test. You can then remove the `ResetKernel` between the tests.

clinssen:
@aserenko : could you look at the PR I made to your branch? Thanks!

jougs:
@aserenko: could you please have a look at/merge the [PR by @clinssen](https://github.com/aserenko/nest-simulator/pull/4) to fix the failing test on OSX? Thanks!

heplesser:
@jougs I think there are enough capable reviewers on this PR so I will withdraw; my earlier comments were on formalities. If is there a special reason you think I should review this one, please add me again.

jougs:
@heplesser: Fine with me. I also un-requested review from @abigailm, as two capable reviewers are already aboard.

terhorstd:
Currently review capacity is limited, but we do not want to delay the next release any further, so I remove this from the 2.18 milestone and we continue after the release.

jougs:
@clinssen, @jakobj, @suku248: could one of you please have a look at this? Many thanks!

jougs:
@aserenko: could you please fix the merge conflict? I'm really sorry that this whole thing took so terribly long, but I can promise to merge it as soon as you fixed the conflict.

jougs:
Can you please also execute the `extras/format_all_c_c++_files.sh` script? Thanks!

aserenko:
Please wait a bit more, @jougs. It is not that straightforward. There are a few formatting changes that, I am afraid, have to be reproduced in my newly-added model files manually.

aserenko:
There still are a few things that concern me.
- Are the reviewers happy with the namings `stdp_nn_pre-centered_synapse`, and so on?
- Is it acceptable that I haven't created the homogeneous `_hom` versions of the models?
- My test passing means just that the models are in line with my understanding of the corresponding STDP rules. Could you, the plasticity experts, confirm that my understanding is correct?

jougs:
@suku248, @jakobj: do you have any comments on this? Many thanks!

jakobj:
I don't, sorry, don't have a lot of experience with STDP implementation in NEST

models/stdp_nn_pre-centered_connection.h
```diff
@@ -0,0 +1,367 @@
...
+    // a postsynaptic spike
+    // (we now know there was at least one between t_lastspike_ and t_spike)
+    // erases the state of the synapse,
+    // and all the preceding presynaptic spikes are forgotten.
+    Kplus_ = 0;
+  }
+
+  // depression due to the latest post-synaptic spike finish->t_
+  // before the current pre-synaptic spike t_spike
+  --finish;
```
jstapmanns: Is shifting this pointer necessary? I think it is not used below, right?

models/stdp_nn_restr_connection.h
```diff
@@ -0,0 +1,357 @@
...
+    assert( minus_dt < -1.0 * kernel().connection_manager.get_stdp_eps() );
+
+    weight_ = facilitate_( weight_, std::exp( minus_dt / tau_plus_ ) );
+  }
+
+  // depression due to the latest post-synaptic spike finish->t_
+  // before the current pre-synaptic spike t_spike
+  if ( start != finish )
+  {
+    --finish;
```
jstapmanns: Same as above. Is shifting this pointer `--finish;` necessary?

models/stdp_nn_pre-centered_connection.h
aserenko: You are right, it is not indeed!

PR #867, opened 2017-12-10
Add pynest version of hpc benchmark

jakobj:
This PR adds a pynest version of `hpc_benchmark.sli` to enable systematic benchmarking also via the Python interface.
I tried to stay as close as possible to the original implementation.
I propose @mschmidt87 and @heplesser as reviewers.

### DIFF
--- a/pynest/examples/hpc_benchmark.py
+++ b/pynest/examples/hpc_benchmark.py
@@ -1,23 +1,23 @@
 # -*- coding: utf-8 -*-
 #
-#  hpc_benchmark.py
+# hpc_benchmark.py
 #
-#  This file is part of NEST.
+# This file is part of NEST.
 #
-#  Copyright (C) 2004 The NEST Initiative
+# Copyright (C) 2004 The NEST Initiative
 #
-#  NEST is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation, either version 2 of the License, or
-#  (at your option) any later version.
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
 #
-#  NEST is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
 #
-#  You should have received a copy of the GNU General Public License
-#  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 
 '''
@@ -42,9 +42,9 @@
 
    This benchmark was originally developed for very large-scale simulations on
    supercomputers with more than 1 million neurons in the network and
-   11.250 incoming synapses per neuron. For such large networks, input to a
-   single neuron will be little correlated and network activity will remain
-   stable for long periods of time.
+   11.250 incoming synapses per neuron. For such large networks, synaptic input
+   to a single neuron will be little correlated across inputs and network
+   activity will remain stable over long periods of time.
 
    The original network size corresponds to a scale parameter of 100 or more.
    In order to make it possible to test this benchmark script on desktop
@@ -89,12 +89,13 @@ params = {
     'simtime': 250.,        # total simulation time in ms
     'presimtime': 50.,      # simulation time until reaching equilibrium
     'dt': 0.1,              # simulation step
-    'record_spikes': True,  # switch to record spikes of excitatory neurons to file
-    'path_name': '',      # path where all files will have to be written
+    'record_spikes': True,  # switch to record spikes of excitatory
+                            # neurons to file
+    'path_name': '',        # path where all files will have to be written
     'log_file': 'log',      # naming scheme for the log files
 }
 
-# -------------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 
 
 def convert_synapse_weight(tau_m, tau_syn, C_m):
@@ -109,8 +110,9 @@ def convert_synapse_weight(tau_m, tau_syn, C_m):
     # time of maximum
     t_max = 1.0 / b * (-lambertwm1(-np.exp(-1.0 / a) / a).real - 1.0 / a)
     # maximum of PSP for current of unit amplitude
-    v_max = np.exp(1.0) / (tau_syn * C_m * b) * ((np.exp(-t_max / tau_m) -
-                                                  np.exp(-t_max / tau_syn)) / b - t_max * np.exp(-t_max / tau_syn))
+    v_max = np.exp(1.0) / (tau_syn * C_m * b) * (
+        (np.exp(-t_max / tau_m) - np.exp(-t_max / tau_syn)) /
+        b - t_max * np.exp(-t_max / tau_syn))
     return 1. / v_max
 
 
@@ -124,12 +126,12 @@ def convert_synapse_weight(tau_m, tau_syn, C_m):
 #     a = tau_m / tau_syn
 #     b = 1.0 / tau_syn - 1.0 / tau_m
 #
-# Inverting this equation numerically leads to tau_syn = 0.32582722403722841 ms,
-# as specified in model_params below.
+# Inverting this equation numerically leads to tau_syn =
+# 0.32582722403722841 ms, as specified in model_params below.
 
 tau_syn = 0.32582722403722841
 
-# -------------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 
 brunel_params = {
     'NE': int(9000 * params['scale']),  # number of excitatory neurons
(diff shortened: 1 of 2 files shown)


pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+   1000 ms simulation time.
+
+   For meaningful use of this benchmark, you should use a scale > 10 and check
+   that the firing rate reported at the end of the benchmark is below 10 spikes
+   per second.
+'''
+
+import numpy as np
+import os
+# from scipy.special import lambertw
```
heplesser: Please remove if not needed.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+import nest.raster_plot
+
+M_INFO = 10
+M_ERROR = 30
+
+
+'''
+ Parameter section
+
+ define all relevant parameters: changes should be made here
```
heplesser: Define

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+
+params = {
+    'nvp': 1,               # total number of virtual processes
+    'scale': 1.,           # scaling factor of the network size,
+                            # total network size = scale*11250 neurons
+    'simtime': 250.,        # total simulation time in ms
+    'presimtime': 50.,      # simulation time until reaching equilibrium
+    'dt': 0.1,              # simulation step
+    'record_spikes': True,  # switch to record spikes of excitatory neurons to file
+    'path_name': './',      # path where all files will have to be written
```
heplesser: I think the default path name (CWD) should be empty, since the expression for "this directory" may differ from `.` on non-Unix systems. The slash should definitely not be included.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+    'log_file': 'log',      # naming scheme for the log files
+}
+
+# -------------------------------------------------------------------------------------
+
+
+def convert_synapse_weight(tau_m, tau_syn, C_m):
+    '''Computes conversion factor for synapse weight from mV to pA. This
```
heplesser: Start docstrings with a brief single-line description, then details on new line. Here, break after "pA."

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+                                                  np.exp(-t_max / tau_syn)) / b - t_max * np.exp(-t_max / tau_syn))
+    return 1. / v_max
+
+
+# Rise time of synaptic currents
+# The synaptic rise time is chosen such that the rise time of the
+# evoked post-synaptic potential is 1.700759 ms.
+# For alpha-shaped postynaptic currents, the rise time of the post-synaptic
+# potential follows from the synaptic rise time as
+
```
heplesser: Add # here to make clear that this is one long comment; same on line 125.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+    Generator for efficient looping over local nodes.
+    '''
+
+    nvp = nest.GetKernelStatus('total_num_virtual_procs')  # step size
+
+    i = 0
+    while i < len(nodes):
+        if nest.GetStatus([nodes[i]], 'local')[0]:
+            yield nodes[i]
+            i += nvp
```
heplesser: This is only correct if we have nodes with proxies, isn't it? That definitely needs to be documented in the docstring.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+   Kunkel et al (2014). Spiking network simulation code for petascale
+     computers. Front. Neuroinform. 8:78
+
+   A note on scaling
+   -----------------
+
+   This benchmark was originally developed for very large-scale simulations on
+   supercomputers with more than 1 million neurons in the network and
+   11.250 incoming synapses per neuron. For such large networks, input to a
+   single neuron will be little correlated and network activity will remain
```
mschmidt87: Correlated across neurons, right?
Perhaps change to "synaptic inputs will have low correlations across neurons"

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+   stable for long periods of time.
```
mschmidt87: stable over long periods of time.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+    'path_name': './',      # path where all files will have to be written
+    'log_file': 'log',      # naming scheme for the log files
+}
+
+# -------------------------------------------------------------------------------------
+
+
+def convert_synapse_weight(tau_m, tau_syn, C_m):
+    '''Computes conversion factor for synapse weight from mV to pA. This
+    function is specific to the used neuron model Leaky
```
mschmidt87: to the used neuron model which is a leaky ...

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+    'eta': 1.685,  # scaling of external stimulus
+    'filestem': params['path_name']
+}
+
+'''
+ FUNCTION SECTION
+'''
+
+
+def build_network(logger):
```
mschmidt87: Add a brief description of the function incl. description of function argument here, something like

'''Build the network including setting of simulation and neuron parameters, creation of neurons and connections.

Parameters
---------------
logger : Instance of Logger class
'''

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+        'overwrite_files': True})
+
+    nest.SetDefaults('iaf_psc_alpha', brunel_params['model_params'])
+
+    nest.message(M_INFO, 'build_network', 'Creating excitatory population.')
+    E_neurons = nest.Create('iaf_psc_alpha', NE)
+
+    nest.message(M_INFO, 'build_network', 'Creating inhibitory population.')
+    I_neurons = nest.Create('iaf_psc_alpha', NI)  # subnet gets own gid
+
```
mschmidt87: What is the purpose of this comment? You are not creating a subnet here, are you?

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+        nest.Connect(local_neurons[:brunel_params['Nrec']], E_detector)
+
+    # read out time used for building
+    BuildEdgeTime = time.time() - tic
+
+    logger.log(str(BuildEdgeTime) + ' # build_edge_time')
+    logger.log(str(memory_thisjob()) + ' # virt_mem_after_edges')
+
+
+def run_simulation():
```
mschmidt87: Same as above: Please add a brief description of the function.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+
+    logger.done()
+
+# ------------------------------------------------------------------------------------
+
+
+def compute_rate():
+    '''compute approximation of average firing rate based on number of
+    local nodes, number of local spikes and total time
+    '''
```
mschmidt87: There should be one common format for docstrings of functions across the script.
Either like this:
'''...
'''

or 
'''
...
'''

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,426 @@
...
+        self.max_rank_cout = 5  # copy output to cout for ranks 0..max_rank_cout-1
+        self.max_rank_log = 30  # write to log files for ranks 0..max_rank_log-1
+        self.line_counter = 0
+
+        if nest.Rank() < self.max_rank_log:
+
+            # convert rank to string, prepend 0 if necessary to make
+            # numbers equally wide for all ranks
+            rank = '{:0' + str(len(str(self.max_rank_log))) + '}'
+            fn = file_name + '_' + rank.format(nest.Rank()) + '.dat'
```
mschmidt87: It would be better to use proper string formatting here. For instance:

`fn = "{}_{}.dat".format(file_name, rank.format(nest.Rank))`

The same for the `log`-Function below.

pynest/examples/hpc_benchmark.py
jakobj: I think it should also work for nodes without proxies. The minimal gid separation is nvp, isn't it? And in case that node is not local, we just start incrementing by one until we find a local node. At least that was the logic I had in mind. Updated the docstring.

jakobj: I see now. You are right. This will only work for nodes with proxies. The minimal step size for nodes without proxies could be one.

pynest/examples/hpc_benchmark.py
jakobj: Yes, I think this means across neurons. This is the text from `hpc_benchmark.sli`. Should I also change it there?

pynest/examples/hpc_benchmark.py
jakobj: (see above)

jakobj:
Thank you for the quick review. I've now implemented almost all suggested changes. I've also changed the logger to a context manager, to make sure file are properly closed also on aborted simulations.

pynest/examples/hpc_benchmark.py
mschmidt87: I would say, yes, though it is a minor point.

jakobj:
Fixed the formatting and adapted doc according to @mschmidt87's suggestions.

pynest/examples/hpc_benchmark.py
```diff
@@ -0,0 +1,487 @@
...
+# evoked post-synaptic potential is 1.700759 ms.
+# For alpha-shaped postynaptic currents, the rise time of the post-synaptic
+# potential follows from the synaptic rise time as
+#
+# def PSP_rise_time(tau_m, tau_syn):
+#     a = tau_m / tau_syn
+#     b = 1.0 / tau_syn - 1.0 / tau_m
+#
+# Inverting this equation numerically leads to tau_syn =
+# 0.32582722403722841 ms, as specified in model_params below.
```
heplesser: @mschmidt87 I am a bit confused here, and since you inserted the corresponding comments/code in the SLI version of the benchmark a while ago, I hope you can clarify what is happening here. My first concern is that in the function `PSP_rise_time()` above a line seems to be missing after the two lines defining `a` and `b`---maybe the line defining `t_max` on line 111 above? As it is written now, there is nothing really to invert.

The other question is why we want to achieve the rise time of 1.700759 ms?

heplesser: @mschmidt87 I am missing a line in the sli and the py version of hpc_benchmark. The `PSP_rise_time` function, as far as I can see, only sets local variables `a` and `b`, but does not compute any rise time and does not return any value. I *assume* that the rise time the function is supposed to compute is the `t_max` expression in function `ConvertSynapseWeight`. Probably someone here planned to split that function up, forgot to copy a line, and then never followed quite through. If you agree with this interpretation, we could change both versions consistently.
The duplicate definition of `/tau_syn` looks like a code duplication introduced by a merge.

mschmidt87: Yes, you are absolutely right, I agree to your interpretation, sorry for being a bit slow-witted here. Line 111 should be moved inside `PSP_rise_time`, the code then be commented in, and lines 108 and 111 can then be replaced by a call of `PSP_rise_time` (`b` is still needed in the rest of the function). Also, `t_max` can be renamed to `t_rise`, or at least we should settle on either "time of maximum" or "rise time".

jakobj:
With the PRs of @hakonsbm and @heplesser merged into my branch, this should be ready to go to master, please have another look

heplesser:
@jakobj There seem to be a few PEP8 issues left, could you take a look?

jakobj:
Done

Issue #868, opened 2017-12-11
Module header not found on compiling NEST with user modules


jgarridoalcazar:
**Description of problem:**
NEST cmake cannot find the header of the external example module. The module installation procedure copies the header in <nest_path>/include/<module_name>/<module_name>.h but the cmake search for <nest_path>/include/<module_name>.h
Compiler: gcc 5.4.0 and cmake 3.5.1

**Version-Release or Git commit:**
NEST 2.14.0 current master (11/12/2017) (3d4d2bad93dd424b29304e56bd2a85bbecc441d1). I could not reproduce it on NEST 2.12.0

**How reproducible: (Always/Sometimes/Unsure)**
Always

**Steps to Reproduce:**
1. Install NEST as usual: cmake -DCMAKE_INSTALL_PREFIX:PATH=<NEST_PREFIX> <nest_source>; make; make install
2. Compile and install myModule example: cmake; make; make install.
3. Call cmake on NEST with external module: cmake -DCMAKE_INSTALL_PREFIX:PATH=<NEST_PREFIX> -Dexternal-modules=my <nest_source>

**Actual results:**
CMake Error at cmake/ProcessOptions.cmake:221 (message):
  Cannot find header for external module 'my'.  Should be
  '/opt/nest/nest_master/nompi/include/mymodule.h'

Expected results:
No error.

Additional info:
During the module installation it indicates this target path.
Installing: <NEST_PREFIX>/include/mymodule/mymodule.h

heplesser:
Fixed by merging #870

PR #870, opened 2017-12-11
Bug fixed on cmake with user modules

jgarridoalcazar:
User module headers are now searched on a specific module folder (<nest_prefix>/include/<module_name>/). It is intended to solve issue #868.


tammoippen:
Good catch. Must have missed that in #844. 👍 from my side.

heplesser:
@jgarridoalcazar Thanks for fixing this!

Issue #869, opened 2017-12-11
User module not linked with NEST in recent GCC versions


jgarridoalcazar:
**Description of problem:**
GCC (versions 4.6 and above) does not link user modules when building NEST with such an option. By default, recent GCC versions do not link against libraries that are not used at compile time (i.e. if no symbols are imported from them, as it is the case of user modules). It is reported [here](https://stackoverflow.com/questions/11631161/force-to-link-against-unused-shared-library). This issue has been reported in the NEST user mailing list.

**Version-Release or Git commit:**
I have reproduced this issue on NEST 2.10, 2.12 and 2.14 (current master branch 3d4d2bad93dd424b29304e56bd2a85bbecc441d1)
Compiler and OS: gcc 5.4.0, cmake 3.5.1, Ubuntu 16.4

**How reproducible: (Always/Sometimes/Unsure)**
Always.

**Steps** to Reproduce:
1. Install NEST as usual: cmake -DCMAKE_INSTALL_PREFIX:PATH=<NEST_PREFIX> <nest_source>; make; make install
2. Compile and install myModule example: cmake; make; make install.
3. Recompile NEST with external module: cmake -DCMAKE_INSTALL_PREFIX:PATH=<NEST_PREFIX> -Dexternal-modules=my <nest_source>
4. Check linked libraries in NEST executable. ldd <NEST_PREFIX>/bin/nest

**Actual results:**
A long list of libraries but not including libmymodule.so.

**Expected results:**
The same list adding libmymodule.so

**Additional info:**
Due to this bug the node and synapsis models including in that module are not available in the list on NEST startup.

heplesser:
Fixed by merging #871.

PR #871, opened 2017-12-11
Fixing issue on linking user module libraries with recent GCC versions

jgarridoalcazar:
This changes are intended to fix issue #869.

It explicitely defines "-Wl,--no-as-needed" linking flag on nest executable, nest library and pynest library. By default, recent versions of GCC set --as-needed flag. This option allows the linker to ignore, i.e., not link against, some of the libraries supplied on its command line if they are not actually used by the shared library that is being created. This option avoids external modules to be linked in NEST even if they are defined with -Dexternal-modules=....

By setting --no-as-needed GCC will link against all the libraries explicitely indicated (including external modules). It may produce slower executable/library startup if more libraries are added than required.

This PR requires PR #870 to be added in order to fix issue #868.

This PR has only been tested with GCC 5.4.0 and Ubuntu 16.04. It should be tested on different compilers and OS.


nest/CMakeLists.txt
```diff
@@ -31,10 +31,16 @@ if ( IS_BLUEGENE AND static-libraries )
         )
 endif ()
 
+set_target_properties( nest
+    PROPERTIES
+    LINK_FLAGS "-Wl,--no-as-needed"
```
hakonsbm: Remove trailing space.

nest/CMakeLists.txt
```diff
@@ -31,10 +31,16 @@ if ( IS_BLUEGENE AND static-libraries )
...
+set_target_properties( nest
+    PROPERTIES
+    LINK_FLAGS "-Wl,--no-as-needed" 
+    )
+
 add_library( nest_lib ${nest_sources} )
 set_target_properties( nest_lib
     PROPERTIES
     OUTPUT_NAME nest
+    LINK_FLAGS "-Wl,--no-as-needed"
```
hakonsbm: Remove trailing space here too.

jgarridoalcazar:
Thanks @hakonsbm for reviewing this PR. The trailling spaces have already been removed.

nest/CMakeLists.txt
```diff
@@ -31,10 +31,16 @@ if ( IS_BLUEGENE AND static-libraries )
         )
 endif ()
 
+set_target_properties( nest
+    PROPERTIES
+    LINK_FLAGS "-Wl,--no-as-needed"
```
heplesser: This flag is not supported by the Apple linker, so you need to protect this by an `if ( NOT APPLE ) ... endif()`.  I am also not sure about the linkers for more special systems, such as BlueGene and K. Maybe the best thing would be to add a test whether `--no-as-needed` is supported by the linker and only apply it if it is. Unfortunately, this does not seem straightforward in CMake, see https://public.kitware.com/pipermail/cmake/2016-January/062623.html; we need to support older CMake versions, back to 2.8.12.

nest/CMakeLists.txt
```diff
@@ -31,10 +31,16 @@ if ( IS_BLUEGENE AND static-libraries )
...
+    )
+
 add_library( nest_lib ${nest_sources} )
 set_target_properties( nest_lib
     PROPERTIES
     OUTPUT_NAME nest
+    LINK_FLAGS "-Wl,--no-as-needed"
```
heplesser: See above.

pynest/CMakeLists.txt
```diff
@@ -46,8 +46,9 @@ if ( HAVE_PYTHON )
     endif ()
     python_add_module( pynestkernel ${pynestkernel_generated_file} pynestkernel.pxd )
     if ( APPLE )
-      set_target_properties( pynestkernel PROPERTIES LINK_FLAGS "-undefined dynamic_lookup" )
+      set_target_properties( pynestkernel PROPERTIES LINK_FLAGS "-undefined dynamic_lookup -Wl,--no-as-needed" )
```
heplesser: See above, leave the Apple-case unchanged.

jgarridoalcazar:
You are right, @heplesser . Removing the --no-as-needed flag in OSX systems should not be an issue since you already tested this bug in OSX and it was not reproducible (see the mailing list). After the last commit the flag is only added to non-Apple systems. It could be very helpful if anyone having access to BlueGene and/or K could test the building in these systems.

jougs:
I have just tested the branch on BlueGene/Q (JUQUEEN) and it worked flawlesly. @heplesser: can you please confirm that it now also works on OS X so we can merge? Thanks!

heplesser:
@jgarridoalcazar Works like a charm on macOS as well now (requires #870 merged, which is in place now). Thanks!

PR #872, opened 2017-12-13
Remove strong dependency on matplotlib

tammoippen:
PyNEST only uses pylab (matplotlib.pylab) in `raster_plot` and `voltage_trace`. In
tests, this matplotlib is mostly optional - except for `test_sp/test_growth_curves.py`.
This PR makes matplotlib completely optional for test.

Making matplotlib optional will greatly simplify maintaining the nest `homebrew` formula in [`homebrew-core`](https://github.com/Homebrew/homebrew-core/pull/21561), as `homebrew-science` will be [deprecated](https://github.com/Homebrew/homebrew-science/issues/6365).

### DIFF
--- a/pynest/nest/tests/test_sp/test_growth_curves.py
+++ b/pynest/nest/tests/test_sp/test_growth_curves.py
@@ -23,10 +23,8 @@ from scipy.integrate import quad
 import math
 import numpy
 from numpy import testing
-import pylab
 import unittest
 import nest
-from nest import raster_plot
 import time
 HAVE_OPENMP = nest.sli_func("is_threaded")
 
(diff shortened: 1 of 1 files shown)


jakobj:
Nice! How about using the full `try,except,else` construct:
```python
try:
    import pylab
except ImportError:
    pass
else:
    <plot stuff>
```

This isolated the exception a bit better ;)
Otherwise 👍  from me

pynest/nest/tests/test_sp/test_growth_curves.py
```diff
@@ -339,22 +339,28 @@ def simulate(self):
...
-            pylab.plot(self.sim_steps, self.se_python[i])
-            pylab.legend(('nest', sei.__class__.__name__))
-            pylab.savefig('sp' + sei.__class__.__name__ + '.png')
-        raster_plot.from_device(self.spike_detector)
-        pylab.savefig('sp_raster_plot.png')
+        try:
+            import pylab
+            from nest import raster_plot
+        except ImportError:
+            pass
```
heplesser: I would replace this with `return` and then have the `else` part as plain code.

pynest/nest/tests/test_sp/test_growth_curves.py
```diff
@@ -339,22 +339,28 @@ def simulate(self):
...
+                pylab.plot(self.sim_steps, self.ca_python[i])
+                pylab.legend(('nest', sei.__class__.__name__))
+                pylab.subplot(1, 2, 2)
+                pylab.title('Synaptic Element')
+                pylab.plot(self.sim_steps, self.se_nest[0, :])
+                pylab.plot(self.sim_steps, self.se_python[i])
+                pylab.legend(('nest', sei.__class__.__name__))
+                pylab.savefig('sp' + sei.__class__.__name__ + '.png')
+            raster_plot.from_device(self.spike_detector)
+            pylab.savefig('sp_raster_plot.png')
```
heplesser: Why should we plot in a test at all?

tammoippen: Look at [this line](https://github.com/tammoippen/nest-simulator/blob/6c71b55ff9b6e3bb154f3ee5683f20039533d356/pynest/nest/tests/test_sp/test_growth_curves.py#L489). Apparently, the author of the test wanted to debug the tests...

pynest/nest/tests/test_sp/test_growth_curves.py
tammoippen: This would be even nicer.

pynest/nest/tests/test_sp/test_growth_curves.py
heplesser: @tammoippen I suppose this is a matter of philosophy, but it makes some sense. To make the code more legible, I would suggest 
- to move the definition of the `plot()` method to right before
- rename it `_plot()` to mark it as internal
- add a docstring like `"Can be called by tearDown() to cross-check tests.`.

tammoippen: @heplesser 
> * to move the definition of the plot() method to right before

Which version are you referring to? The complete `_plot` code in the `try:` block?

tammoippen:
I applied the changed I suggest @heplesser asked me to do. This PR is hindering nest to be accepted in homebrew core since Dec 13, 2017. I get no replies since Jan 15. Can somebody please have a look?

tammoippen:
@heplesser Done as requested.

tammoippen:
Thanks! 🎉 🍰

PR #873, opened 2017-12-14
Add SymmetricBernoulliConnector

jakobj:
This adds a new connector based on BernoulliConnector, which creates symmetric connections between nodes with a fixed probability p. However, instead of iterating over sources for each target and creating a connection with probability p, this connector draws a random indegree for each target and then selects number indegree many random sources. To make sure connections are created symmetrically also in the distributed case, identical indegrees and sources need to be drawn on each virtual process. To achieve this, the connector uses the global rng to seed a local rng on each virtual process with the same seed.

As reviewers I recommend @heplesser, @janhahne and @hannahbos

### DIFF
--- a/pynest/nest/tests/test_connect_symmetric_pairwise_bernoulli.py
+++ b/pynest/nest/tests/test_connect_symmetric_pairwise_bernoulli.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+#
+# test_connect_symmetric_pairwise_bernoulli.py
+#
+# This file is part of NEST.
+#
+# Copyright (C) 2004 The NEST Initiative
+#
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import collections
+import numpy as np
+import unittest
+import scipy.stats
+from . import test_connect_helpers as hf
+from .test_connect_parameters import TestParams
+
+
+class TestSymmetricPairwiseBernoulli(TestParams):
+
+    # sizes of source-, target-population and connection probability for
+    # statistical test
+    N_s = 60
+    N_t = 60
+    # specify connection pattern and specific params
+    rule = 'symmetric_pairwise_bernoulli'
+    p = 0.5
+    conn_dict = {'rule': rule, 'p': p, 'multapses': True, 'autapses': False,
+                 'make_symmetric': True}
+    # Critical values and number of iterations of two level test
+    stat_dict = {'alpha2': 0.05, 'n_runs': 20}
+
+    def testStatistics(self):
+        for fan in ['in', 'out']:
+            expected = hf.get_expected_degrees_bernoulli(
+                self.p, fan, self.N_s, self.N_t)
+
+            pvalues = []
+            for i in range(self.stat_dict['n_runs']):
+                hf.reset_seed(i, self.nr_threads)
+                self.setUpNetwork(conn_dict=self.conn_dict,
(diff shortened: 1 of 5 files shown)


pynest/nest/tests/test_connect_symmetric_pairwise_bernoulli.py
```diff
@@ -0,0 +1,149 @@
...
+        conns = set()
+        for conn in hf.nest.GetConnections():
+            key = tuple(conn[:2])
+            conns.add(key)
+
+        for conn in hf.nest.GetConnections():
+            key = tuple(conn[:2])
+            self.assertTrue(key[::-1] in conns)
+
+
```
janhahne: I am missing an test that actually checks if the all created connections are symmetric. Something like

```py
import nest
import numpy as np
import test_connect_helpers as hf
from mpi4py import MPI

n = nest.Create('hh_psc_alpha_gap',40)

rule = 'symmetric_pairwise_bernoulli'
p = 0.5
conn_dict = {'rule': rule, 'p': p, 'multapses': True, 'autapses': False, 'make_symmetric': True}

nest.Connect(n ,n , conn_dict, {'model':'gap_junction'})

local_matrix = hf.get_connectivity_matrix(n,n)

global_matrix = hf.gather_data(local_matrix)

if MPI.COMM_WORLD.Get_rank() == 0:
  #print global_matrix
  assert(np.allclose(global_matrix, global_matrix.T, atol=1e-8))
```
which is called from e.g. 1,4 and 8 MPI processes. Is there an easy way to include this?

jakobj: Actually, we were testing for symmetric connections in line 128--135.
1) add all pairs of source gid, target gid to a set
2) check whether the reversed pairs are in the set

Shouldn't this do the trick?
```python
        conns = set()
        for conn in hf.nest.GetConnections():
            key = tuple(conn[:2])
            conns.add(key)

        for conn in hf.nest.GetConnections():
            key = tuple(conn[:2])
            self.assertTrue(key[::-1] in conns)
```

I agree that we do not currently test for the distributed case. As far as I can see, this is done in `test_connect_all_patterns.py`. However this test fails on my system. Will create an issue for that and include MPI tests as soon as I've figured out what goes wrong.

jakobj:
Thanks for the quick review @janhahne!

jakobj:
Fixed formatting and updated doc

pynest/nest/tests/test_connect_symmetric_pairwise_bernoulli.py
janhahne: @jakobj Oh sorry, of course you are right, I somehow missed that part and thought that `testMakeSymmetric(self)` only checks that `make_symmetric` must be enabled. I think the current test is sufficient to test the functionality without MPI (also in theory there could be several identical connections in one direction and only one connection in the opposite direction and the test would not detect it).

Once an MPI test is added I am all happy :+1:

hannahbos:
@jakobj this is great work and thanks for the good documentation. I have two rather open remarks.

1) It seems at bit overloaded to introduce yet another key for symmetry. Can the key own_symmetric be removed? It is only required in line 355 of conn_builder.cpp to prevent weight resetting and creation of symmetric connections triggered by the key make_symmetric. I do not really understand why make_symmetric needs to be set to true. Could you leave it at false and drop own_symmetric? 

2) Is the new builder faster than the original Bernoulli builder? Would it be possible to replace the algorithm in the original Bernoulli builder with the proposed algorithm? The Bernoulli builder could have a key ‘symmetric’, which indicates whether connections are created both ways.

jakobj:
Thank you for your comments @hannahbos, sorry for not replying sooner.
1. I agree, it doesn't feel completely right to introduce `own_symmetric`. However, if we set `make_symmetric=false` but the connector creates symmetric connections, doesn't this violate the principle of least surprise quite strongly? Or did I misinterpret your comment?
2. I haven't performed any measurements. To be honest, I also haven't inspected the code of the original bernoulli builder enough to guess whether there might be speed difference. Will look into that.

jakobj:
@hannahbos: I've taken a look at the original `BernoulliBuilder` at it seems that builder is fully parallel. If that is the case, I would not recommend to replace it, since the `SymmetricBernoulliBuilder` is not parallel: all virtual processes use identical random number streams, only considering pairs of neurons relevant for them. This seemed to be the only way to construct symmetric connections in a distributed setting without additional MPI communication.

jakobj:
since this builder is also implemented in 5g, #920 will introduce it to the NEST kernel. I am hence closing this PR in favor of #920.

PR #887, opened 2018-01-30
Fix incorrect error message when connecting with bad delay

### LINKED ISSUE #287: Incorrect error message when connecting with bad delay
NEST may issue an incorrect error message if minimum and maximum delay have been set explicitly and a connection is created with invalid delay

```
SLI ] 0 << /min_delay 2.0 /max_delay 4.0 >> SetStatus
SLI ] /iaf_neuron Create dup 1. 2.5 Connect
SLI ] /iaf_neuron Create dup Connect

Mar 28 15:13:59 Connect_i_i_l [Error]: BadDelay
    Delay value 1 is invalid: Default delay of 'static_synapse' must be 
    between min_delay 0.1 and max_delay 0.1.
SLI [3] 0 [ [ /min_delay /max_delay ] ] get ==
[2.000000e+00 4.000000e+00]
```

Note that the first `Connect` call with delay 2.5 ms works fine, indicating that the delay limits have been set. The second `Connect` calls uses the default delay of 1 ms, provoking an error, but the error message reports min and max delay of 0.1 ms, which are the initial values. The subsequent lookup reports that the correct values were set in the kernel. Note that moving the `get` call before the `Connect` calls will lead to correct data in the error message (except that "ms" is missing).

The exception is raised by

```
void
GenericConnectorModel< ConnectionT >::used_default_delay()
{
  if ( default_delay_needs_check_ )
  {
    try
    {
      if ( has_delay_ )
      {
        kernel().connection_builder_manager.get_delay_checker().assert_valid_delay_ms(
          default_connection_.get_delay() );
      }
    }
    catch ( BadDelay& e )
    {
      throw BadDelay(
        default_connection_.get_delay(),
        String::compose( "Default delay of '%1' must be between min_delay %2 and max_delay %3.",
          get_name(),
          Time::delay_steps_to_ms( kernel().connection_builder_manager.get_min_delay() ),
          Time::delay_steps_to_ms( kernel().connection_builder_manager.get_max_delay() ) ) );
    }
    default_delay_needs_check_ = false;
  }
}
```

so the problem is most likely that the `get_min_delay()`/`get_max_delay()` methods return non-updated values.

hakonsbm:
This PR fixes #287 by updating the saved values of the delay bounds if the delay bounds are updated.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -180,6 +180,11 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
   {
     delay_checkers_[ i ].set_status( d );
   }
+  //  Need to update the saved values if we have changed the delay bounds.
+  if ( d->known( names::min_delay ) or d->known( names::max_delay ) )
+  {
+    update_delay_extrema_();
+  }
 }
 
 nest::DelayChecker&


terhorstd:
approved by open nest video conference

PR #888, opened 2018-01-31
Issue error message when attempting to set an invalid resolution

### LINKED ISSUE #643: Issue error message when attempting to set an invalid resolution
Since #265, simulation time must be a multiple of the resolution. This can lead to confusing error messages upon calls to `Simulate` if the resolution cannot be represented properly in tics. For example,
```
0 << /resolution 0.03125 >> SetStatus  % == 2^-5
1 Simulate
```
will lead to an error on the `Simulate` call, even though the simulation time is 32 times the resolution.
The problem is that with a default tic of 1 microsecond, NEST actually sets the resolution to 0.031, not 0.03125.  `SetStatus` should raise an error in this case.

hakonsbm:
Issues an error message when attempting to change the resolution to something which is not a multiple of the tic size.

Fixes #643.

### DIFF
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -134,6 +134,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
     updateValue< double >( d, names::tics_per_ms, tics_per_ms );
   double resd = 0.0;
   bool res_updated = updateValue< double >( d, names::resolution, resd );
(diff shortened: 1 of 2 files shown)


hakonsbm:
@Silmathoron The idea of the test is to check that you get an error message when trying to set an invalid resolution. Setting the resolution to a valid value is already in multiple other tests.

PR #890, opened 2018-02-06
Add 'install-nodoc' target for make

### LINKED ISSUE #153: Add `install-nodoc` target
Installation of NEST online help takes an irritating amount of time during development work, when one is frequently recompiling and reinstalling. As long as one only changes C++ code or PyNEST Python code, one can use `make install-exec` to skip installation of documentation, but this also skips installation of SLI libraries, the testsuite and examples. Therefore, we should add an `install-nodoc` target that installs everything except the online documentation.

See also #152 .

jougs:
This fixes #153.

Unfortunately, CMake does not support custom code or conditionals in the `install` target. I therefore implemented this as an extension to the help extraction system itself.


CMakeLists.txt
```diff
@@ -361,4 +361,8 @@ install( FILES LICENSE README.md NEWS
     DESTINATION ${CMAKE_INSTALL_DOCDIR}
     )
 
+add_custom_target( install-nodoc
+  COMMAND make NEST_INSTALL_NODOC=true install
```
heplesser: Can I still use `-j N` with this?

jougs: The short answer is yes.

The slightly longer answer is that my changes do not alter the behavior of your call to `make` for any target but `install`. And for that `-j N` did not have an effect anyway as the installation is always performed serially in order to avoid problems with concurrent writes.

jougs:
@heplesser: Thanks for the review. I've addressed your comments in my commit 
2b16a06 and my reply to your question.

heplesser:
@jougs I can use the `install-nodoc` target, but it still installs the help.

heplesser:
@jougs I just sent you a PR adding a required "optional" flag. With this, things work in a way: as long as one only ever calls `make` with the `install-nodoc`, help will not get installed. If invoking a "naked" `make`, `doc/help` will be populated and then installed in all subsequent `make` invocations, even for target `install-nodoc`.

I wonder if a complete solution of this problem would require introduction of proper build and install configurations, such as deploy/develop/debug.

jougs:
@heplesser: I've come up with a reasonably stable solution in my recent commits. Can you please test again? Thanks!

Issue #848, opened 2017-10-31
nest import causes scikit-learn to segfault


maharjun:
The following is the code that can reproduce the error

```python
import nest
from sklearn.svm import LinearSVC

import numpy as np

X_data = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
Y_data = np.array([0, 0, 1])
linsvc = LinearSVC(C=1.0, class_weight={0: 0.001, 1: 2.0}).fit(X=X_data, y=Y_data)
```

This code will segfault. However, if you flip the order of the imports to
```python
from sklearn.svm import LinearSVC
import nest
```

Then everything works just fine.

I have dug into the issue a bit and uncovered this.

scikit-learn imports cython-compiled libraries libsvm and liblinear both of which contain a (non-exposed) function named `set_parameter` which is called when one makes an invocation to `SVC.fit()` or `LinearSVC.fit()` respectively. The problem is that if nest is imported before scikit-learn, then it causes scikit-learn to mess up the dynamic loading somehow (This is the part that I don't know) which leads to the `liblinear.train_wrap` function that is called by `LinearSVC.fit()` to use the `set_parameter` function from libsvm instead of liblinear. The parameters of the two varieties of `set_parameter` are different thus causing a segfault. It appears that this does not happen if nest is imported after sklearn.svm.

This can be seen by setting a breakpoint on the function `set_parameter` when running the python script under GDB.

This is the first case (i.e. nest before sklearn.svm => segfault). The following is the output of gdb when it reaches a function named `set_parameter`
```
Breakpoint 1, set_parameter (param=0x1, svm_type=2, kernel_type=24510720, degree=26374064, gamma=0.0001, coef0=1, nu=0.10000000000000001, cache_size=0, 
    C=1.3852388523421298e-309, eps=0, p=0, shrinking=1000, probability=686439501, nr_weight=-11664, weight_label=0x0, weight=0x0, max_iter=686439501, 
    random_seed=-563176512) at sklearn/svm/src/libsvm/libsvm_helper.c:74
```

This is the second case (i.e. nest after sklearn.svm => no segfault). The following is the output of gdb when it reaches a function named `set_parameter`

```
Breakpoint 1, set_parameter (solver_type=1, eps=0.0001, C=1, nr_weight=2, weight_label=0x10aad20 "", weight=0x160e8d0 "\374\251\361\322MbP?", 
[...]

PR #891, opened 2018-02-13
Updating import instructions for scipy, sklearn - issue #848

jessica-mitchell:
This PR is addressing the issue #848, which notes that you have import sklearn before nest, as well as scipy. This is not a fix but just a documentation change to make people aware of it. I added a comment to the PyNEST tutorial regarding the import order of both scipy and sklearn. Additionally, I added a line to indicate the import order (wrt scipy) in the CampbellSiegert example, since it was the only example that used scipy that did not explicitly mention the order needed.  
There are also a couple of minor changes to the PyNEST tutorial, unrelated to the import issue but brought to my attention as mistakes in the docs that I hope would be ok to add into this PR.


pynest/examples/CampbellSiegert.py
```diff
@@ -34,6 +34,11 @@
 Sven Schrader, Nov 2008, Siegert implementation by Tom Tetzlaff
 '''
 
+'''
+First, we import all necessary modules for simulation and analysis. Scipy
```
Silmathoron: On my computer, putting the two scipy imports after nest does not lead to any kind of trouble, can you post what the error message you get is?
(python 3.6.4, scipy 0.19)

extras/userdoc/md/documentation/part-1-neurons-and-simple-neural-networks.md
```diff
@@ -66,12 +66,22 @@ Python interpreter.
 
     import nest
 
+
+It should be noted, however, that certain external packages must be imported *before* importing nest. 
+These include `sklearn` ([scikit-learn](http://scikit-learn.org/stable/index.html)) and `scipy` ([SciPy](https://www.scipy.org/)):
+
+    from sklearn.svm import LinearSVC
```
Silmathoron: I can confirm the sklearn bug on Python 3.6.4 and SKL 0.19.1, but not with scipy.
Also importing only ``sklearn`` and not the specific submodule does not solve the problem.

extras/userdoc/md/documentation/part-1-neurons-and-simple-neural-networks.md
```diff
@@ -66,12 +66,22 @@ Python interpreter.
...
+    import nest
+
+
 As with every other module for Python, the available functions can be prompted
 for.
 
     dir(nest)
 
-One such command is `nest.Models?`, which will return a list of all the
+One such command is `nest.Models?`, which will return a list of all the
```
jougs: Please remove trailing whitespace

pynest/examples/CampbellSiegert.py
jougs: The original issue we experienced was https://github.com/numpy/numpy/issues/2521, apparently fixed in https://github.com/numpy/numpy/commit/a226daa75816577ac9c734133e600e5eebd9aaa9.

As the suggested import order does not matter in most cases and prevents a crash in the best case, I like the proposed addition although most people will not be hit by the long-fixed issue anymore.

extras/userdoc/md/documentation/part-1-neurons-and-simple-neural-networks.md
```diff
@@ -248,8 +258,7 @@ Now we are ready to display the data in a figure. To this end, we make use of
     pylab.figure(1)
     pylab.plot(ts, Vms)
 
-The second line opens a figure (with the number 1), the third line clears the
-window and the fourth line actually produces the plot. You can’t see it yet
+The second line opens a figure (with the number 1), and the third line actually produces the plot. You can’t see it yet
```
jougs: Please break the like in a similar fashion as before.

extras/userdoc/md/documentation/part-1-neurons-and-simple-neural-networks.md
jougs: @Silmathoron thanks for double-checking. In this case I agree with the suggested changes.

heplesser:
@jougs @Silmathoron Are you happy with the PR now?

PR #895, opened 2018-02-28
New neuronal population model gif_pop_psc_exp

mdeger:
Tilo Schwalger (@schwalger), Hesam Setareh (@hesam-setareh) and I would like to add the model **gif_pop_psc_exp** to NEST by means of this pull request. 

The model represents the pooled activity a population of (any number of) neurons in a single NEST node, accurately treating stochastic finite size effects. The algorithm and a range of simulations of single neuron vs. population activity are published in the journal article [1]: 

Schwalger T, Deger M, Gerstner W (2017) Towards a theory of cortical columns: From spiking neurons to interacting neural populations of finite size. PLoS Comput Biol 13(4): e1005507. https://doi.org/10.1371/journal.pcbi.1005507

The model is related to the flexible single neuron model gif_psc_exp, hence the name gif_pop_psc_exp. Using this population model, one can build models of interconnected populations, etc, as discussed in [1], which can be simulated efficiently.


mdeger:
Somehow the travis builds are failing. As far as I can see, the problem is this:

`/home/travis/build/nest/nest-simulator/models/gif_pop_psc_exp.h:315:5: error: ‘GSL_BinomialRandomDev’ in namespace ‘librandom’ does not name a type`

However, I cannot reproduce this error when building on my machine. Could it be that there is a different version of GSL on the travis-CI build system? Any ideas how to proceed?

Silmathoron:
Models that require GSL need to be registered inside the [GSL guard](https://github.com/nest/nest-simulator/pull/895/files#diff-532c3a773a2b35a8dbeaaade0459b80aR373).
The error does not occur locally because you have installed gsl; however, in the tests, we do some of them without GSL, which is where your code fails.

testsuite/regressiontests/issue-77.sli
```diff
@@ -201,3 +202,4 @@ M_ERROR setverbosity
 assert_or_die
 
 endusing
+
```
hakonsbm: Remove the extra blank line.

testsuite/unittests/test_gif_pop_psc_exp.sli
```diff
@@ -0,0 +1,112 @@
...
+/* BeginDocumentation
+Name: testsuite::test_gif_pop_psc_exp - sli script for test of gif_pop_psc_exp
+
+Synopsis: (test_gif_pop_psc_exp) run -> compare mean and variance of population rate of the model with desired values
+
+Description:
+ 
+ This script simulate an inhibitory population with the population model. After each simulation, it calculates
+ the mean and variance of population rate and compare them with desired values. The simulated values should be 
+ in a acceptable margine.
```
hakonsbm: "margin".

testsuite/unittests/test_gif_pop_psc_exp.sli
```diff
@@ -0,0 +1,112 @@
...
+Author:  December 2017, Setareh, Schwalger
+SeeAlso: gif_pop_psc_exp, testsuite::test_pp_pop_psc_delta
+
+*/
+
+(unittest) run
+/unittest using
+
+1.0 /err_mean Set      % error margine for mean rate (Hz)
+6.0 /err_var Set       % error margine for variance of rate (Hz^2)
```
hakonsbm: "margin"

models/modelsmodule.cpp
```diff
@@ -86,6 +87,8 @@
 #include "gif_psc_exp_multisynapse.h"
 #include "gif_cond_exp.h"
 #include "gif_cond_exp_multisynapse.h"
+#include "gif_pop_psc_exp.h"
```
heplesser: Isn't this the same line as line 81?

models/gif_pop_psc_exp.cpp
```diff
@@ -526,22 +527,18 @@ nest::gif_pop_psc_exp::update( Time const& origin,
...
-            / ( P_.tau_syn_in_ - P_.tau_m_ ) );
+    // membrane update (line 10 of [1])
+    double h_ex_tmpvar_ = ( P_.tau_syn_ex_ * V_.P11_ex_ * ( JNy_ex_ - JNA_ex_ )
+              - V_.P22_ * ( P_.tau_syn_ex_ * JNy_ex_ - P_.tau_m_ * JNA_ex_ ) );
+    double h_in_tmpvar_ = ( P_.tau_syn_in_ * V_.P11_in_ * ( JNy_in_ - JNA_in_ )
+              - V_.P22_ * ( P_.tau_syn_in_ * JNy_in_ - P_.tau_m_ * JNA_in_ ) );
+    double h_ex_ = P_.tau_m_ * ( JNA_ex_ + 
+                    h_ex_tmpvar_ / ( P_.tau_syn_ex_ - P_.tau_m_ ) );
+    double h_in_ = P_.tau_m_ * ( JNA_in_ + 
+                    h_in_tmpvar_ / ( P_.tau_syn_in_ - P_.tau_m_ ) );
```
heplesser: The compiler may be able to optimize better if you declare these four variables `const`. Also, local variable names should not end with `_`, that suffix is only for private members.

models/gif_pop_psc_exp.cpp
```diff
@@ -553,37 +550,37 @@ nest::gif_pop_psc_exp::update( Time const& origin,
...
+    S_.V_m_ = ( S_.V_m_ - P_.E_L_ ) * V_.P22_ + h_tot_; // line 3 of [1]
 
     // compute free adaptation state
-    for ( uint j = 0; j < P_.tau_sfa_.size(); j++ ) // line 4
+    for ( uint j = 0; j < P_.tau_sfa_.size(); ++j ) // line 4 of [1]
     {
-      V_.g_[ j ] = V_.g_[ j ] * V_.Q30_[ j ]
-        + ( 1. - V_.Q30_[ j ] ) * V_.n_[ V_.k0_ ]
-          / ( static_cast< double >( P_.N_ ) * V_.h_ );           // line 5
+      double g_j_tmp_ = ( 1. - V_.Q30_[ j ] ) * V_.n_[ V_.k0_ ]
```
heplesser: As above: declare local helper variables `const` where possible and avoid the trailing `_`.

models/gif_pop_psc_exp.cpp
```diff
@@ -553,37 +550,37 @@ nest::gif_pop_psc_exp::update( Time const& origin,
...
+          / ( static_cast< double >( P_.N_ ) * V_.h_ );
+      V_.g_[ j ] = V_.g_[ j ] * V_.Q30_[ j ] + g_j_tmp_; // line 5 of [1]
       S_.theta_hat_ = S_.theta_hat_ + V_.Q30K_[ j ] * V_.g_[ j ]; // line 6
     }
 
     // compute free escape rate
-    double lambda_tld_ = escrate( S_.V_m_ - S_.theta_hat_ ); // line 8
+    double lambda_tld_ = escrate( S_.V_m_ - S_.theta_hat_ ); // line 8 of [1]
```
heplesser: Also this could be `const`

models/gif_pop_psc_exp.cpp
```diff
@@ -553,37 +550,37 @@ nest::gif_pop_psc_exp::update( Time const& origin,
...
     double P_free_ = 1 - std::exp( -0.5 * ( V_.lambda_free_ + lambda_tld_ )
-                           * V_.h_ / 1000. );         // line 9
+                           * V_.h_ / 1000. );         // line 9 of [1]
```
heplesser: This could probably also be `const`. Replacing division by 1000. with multiplication by 0.001 would improve performance. Or could you even redefine `V_.h_` in a way to avoid the scaling here?

pynest/examples/gif_pop_psc_exp.py
```diff
@@ -31,79 +40,69 @@
...
-At first, we load the necessary modules:
-'''
-
-import matplotlib
-import numpy as np
-import matplotlib.pyplot as plt
-import nest
 
 '''
 Next, we set the parameters of the microscopic model
```
heplesser: "Next" doen't seem to fit well for the very first step in the example.

models/gif_pop_psc_exp.cpp
```diff
@@ -23,20 +23,17 @@
 /* Point process population model with exponential postsynaptic currents and
  * adaptation */
 
-#include "exceptions.h"
+/* [1]: line numbers in comments refer to the algorithm pseudocode in 
+Figures 11 and 12 of the paper
```
heplesser: Please indent all lines so that they line up under "line".

models/gif_pop_psc_exp.cpp
```diff
@@ -0,0 +1,702 @@
...
+  tau_sfa_.push_back( 300.0 ); // ms
+  q_sfa_.push_back( 0.5 );     // mV
+}
+
+nest::gif_pop_psc_exp::State_::State_()
+  : y0_( 0.0 )
+  , I_syn_ex_( 0.0 )
+  , I_syn_in_( 0.0 )
+  , V_m_( 0.0 )
+  , theta_hat_( 0.0 )   // initialization value has no effect for theta_hat_
```
hakonsbm: `theta_hat_` should be initialized after `n_spikes_` to avoid warnings during compilation.

mdeger:
Thank you very much for your comments. I will start looking at the new ones right away.

Concerning the GSL Guard: Please give me some more explicit instructions or a reference, I don't understand what you mean. Is GSL Guard some list where models needing GSL are to be registered? Or something I need to include?

models/modelsmodule.cpp
mdeger: Thanks. I removed line 81 instead.

models/gif_pop_psc_exp.cpp
mdeger: lambda_tld_ is assigned assigned a new value in line 592, so I think it cannot be const.

testsuite/unittests/test_gif_pop_psc_exp.sli
```diff
@@ -38,6 +38,9 @@ SeeAlso: gif_pop_psc_exp, testsuite::test_pp_pop_psc_delta
 
 */
 
+% This test should only run if we have GSL
+statusdict/have_gsl :: not {statusdict/exitcodes/success :: quit_i} if
```
heplesser: Please move this check to after `(unittest) run` and replace it with `skip_if_without_gsl`. This is not only simpler, but it will also mark the test as skipped due to lack of GSL instead of implicitly counting it as passed.

pynest/examples/gif_pop_psc_exp.py
```diff
@@ -0,0 +1,401 @@
...
+and one inhibitory population.
+
+Note that the population model represents the mesoscopic level description of
+the corresponding microscopic network based on the NEST model gif_psc_exp.
+'''
+
+
+'''
+Next, we set the parameters of the microscopic model
+'''
```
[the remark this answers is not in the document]
heplesser: This is in principle correct, but for PyNEST examples, we have different rules: Multi-line strings framed with `'''` are considered documentation and are extracted and converted to Markdown to make the examples accessible on our website; see http://www.nest-simulator.org/py_sample/brunel_alpha_nest/index.html for an example.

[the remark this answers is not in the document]
hakonsbm: My bad, I was not aware of that. Better use the triple quotes for example documentation after all then.

hakonsbm:
@mdeger Thank you for the changes! This looks good now.

Sorry for the erroneous comment. I have created a PR in your fork reinstating the triple quotes. As soon as you merge that, I will approve this PR.

PR #898, opened 2018-03-07
Make sure that Run and Cleanup are called after Prepare

### LINKED ISSUE #659: Add state checks for prepare-run-cleanup loop
PR #650 : after merging, consideration should be given to more carefully checking state.

hakonsbm:
An error will now be raised if `Run()` or `Cleanup()` are called without calling `Prepare()` first.

This fixes #659.

### DIFF
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -379,6 +379,13 @@ nest::SimulationManager::prepare()
 {
   assert( kernel().is_initialized() );
 
+  if ( prepared_ )
+  {
+    std::string msg = "Prepare called twice.";
+    LOG( M_ERROR, "SimulationManager::prepare", msg );
+    throw KernelException();
+  }
+
   if ( inconsistent_state_ )
   {
     throw KernelException(
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -67,6 +67,7 @@ nest::SimulationManager::initialize()
   Time::reset_resolution();
   clock_.calibrate();
 
+  prepared_ = false;
   simulating_ = false;
   simulated_ = false;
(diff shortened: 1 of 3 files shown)


nestkernel/simulation_manager.cpp
```diff
@@ -430,6 +431,7 @@ nest::SimulationManager::prepare()
       * kernel().connection_manager.get_min_delay();
     kernel().music_manager.enter_runtime( tick );
   }
+  prepared_ = true;
```
apeyser: We should check at the top of the function that `! prepared_` and otherwise throw an exception like in the other checks.

nestkernel/simulation_manager.h
```diff
@@ -169,7 +169,8 @@ class SimulationManager : public ManagerInterface
   delay to_step_;         //!< update clock_+from_step<=T<clock_+to_step_
   timeval t_slice_begin_; //!< Wall-clock time at the begin of a time slice
   timeval t_slice_end_;   //!< Wall-clock time at the end of time slice
-  long t_real_;     //!< Accumunated wall-clock time spent simulating (in us)
+  long t_real_;   //!< Accumulated wall-clock time spent simulating (in us)
+  bool prepared_; //!< indicates whether the SimulationManager has already been
```
apeyser: `indicates whether the SimulationManager has already been` seems to be missing a word or two.

testsuite/regressiontests/issue-659.sli
```diff
@@ -0,0 +1,57 @@
...
+  ResetKernel
+  10 Run
+} fail_or_die
+
+{
+  % Cleanup without prepare.
+  ResetKernel
+  Cleanup
+} fail_or_die
+
```
apeyser: And prepare after prepare test.

PR #648, opened 2017-01-30
New iaf_exp precise model that guarantees perfect detection of spikes

krishnanj:
This model implements an additional function is_spike_ that guarantees detection of all spikes in a simulation, in particular the case that the current version of precise iaf_exp_ps is expected to miss.

Conventional spike detection algorithms propagate the initial state forwards in time from t=0 to t=h and see whether it meets the threshold. This function implements an alternative general method to solve the threshold-detection problem for an integrable, affine or linear time evolution by applying geometric analysis. The idea is to propagate the threshold backwards in time from t=0 to t=-h and check whether its trajectory meets the initial state. The threshold now evolves, whereas the initial state is fixed, therefore is_spike can tell occurrence of spike but does not return time of occurrence of spike, t_spike, hence is computationally faster.

In state space spanned by voltage and current as the threshold plane is evolved backward in the time interval [t+dt, t] for a propagation step dt, the spiking region S and non-spiking region NS is bounded by the threshold plane, the envelope curve or propagated threshold, and plane corresponding to the final timestep that clearly separates the spiking regions S and 
non-spiking regions NS where each branch of the is_spike_ function points to a region in the state space.  

This particular ordering of tests for different state space regions is optimized such that conditions for the most visited region is checked for first, followed by others. If a missed spike is detected, the precise time of spike occurrence is calculated by bisection algorithm.

PR #901, opened 2018-03-07
Add iaf_psc_exp_ps_lossless model by Krishnan et al

heplesser:
This PR is a follow-up to the retired #648. Two issues remain at present:

1. The test for the model fails, because the reference data are taken from the non-precise iaf_psc_exp model. We need reference data generated outside of NEST for this model.
1. The test for the model is very simple, just response to DC input; it does not test any of the advanced features, in particular the losslessness.
1. There is some problem with the copyright header, probably a technical problem with the test.


precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+    V_.I_syn_in_before_ = S_.I_syn_in_;
+    V_.y2_before_ = S_.y2_;
+
+    // get first event
+    double ev_offset;
+    double ev_weight;
+    bool end_of_refract;
+
+    if ( not B_.events_.get_next_spike(
+           T, true, ev_offset, ev_weight, end_of_refract ) )
```
Silmathoron: This adds all the simultaneous spikes before a potential threshold crossing, i.e. a reset, meaning that the effect of all spikes that arrived in addition to those sufficient to trigger threshold crossing will be "lost", compared to the "standard" implementation of ``iaf_psc_exp_ps``.

I have not strong preference (at least I did not think long enough about this to have one), I'm just pointing out that this differs from the other models.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+    y2_root = P20 * ( P_.I_e_ + V_.y0_before_ ) + P21_ex * V_.I_syn_ex_before_
+      + P21_in * V_.I_syn_in_before_ + expm1_tau_m * V_.y2_before_
+      + V_.y2_before_;
+  }
+  return root;
+}
+
+double
+nest::iaf_psc_exp_ps_lossless::is_spike_( const double dt )
```
Silmathoron: [disclaimer] I have no time to check this in details... [/disclaimer]

precise/iaf_psc_exp_ps_lossless.h
```diff
@@ -0,0 +1,486 @@
...
+  // ----------------------------------------------------------------
+
+  /**
+   * State variables of the model.
+   */
+  struct State_
+  {
+    double y0_;       //!< External input current
+    double I_syn_ex_; //!< Exc. exponential current
+    double I_syn_in_; //!< Inh. exponential current
```
stinebuu: Maybe add a comment about which values correspond to y1.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+  B_.events_.clear();
+  B_.currents_.clear(); // includes resize
+  B_.logger_.reset();
+}
+
+void
+nest::iaf_psc_exp_ps_lossless::calibrate()
+{
+  B_.logger_
+    .init(); // ensures initialization in case mm connected after Simulate
```
stinebuu: This would be easier to read for me if the comment is placed above the `B_.logger_.init();` so that the actual code is one line.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+  V_.h_ms_ = Time::get_resolution().get_ms();
+
+  V_.expm1_tau_m_ = numerics::expm1( -V_.h_ms_ / P_.tau_m_ );
+  V_.expm1_tau_ex_ = numerics::expm1( -V_.h_ms_ / P_.tau_ex_ );
+  V_.expm1_tau_in_ = numerics::expm1( -V_.h_ms_ / P_.tau_in_ );
+  V_.P20_ = -P_.tau_m_ / P_.c_m_ * V_.expm1_tau_m_;
+  V_.P21_ex_ = -P_.tau_m_ * P_.tau_ex_ / ( P_.tau_m_ - P_.tau_ex_ ) / P_.c_m_
+    * ( V_.expm1_tau_ex_ - V_.expm1_tau_m_ );
+  V_.P21_in_ = -P_.tau_m_ * P_.tau_in_ / ( P_.tau_m_ - P_.tau_in_ ) / P_.c_m_
+    * ( V_.expm1_tau_in_ - V_.expm1_tau_m_ );
```
stinebuu: The expressions for `V_.P20_{ex, in}_` are a bit hard to read. I think you can multiply `-P_.tau_m_ * P_tau_{ex, in}_` by `P_.c_m_` because you are dividing twice.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+  if ( not S_.is_refractory_ )
+  {
+    const double expm1_tau_m = numerics::expm1( -dt / P_.tau_m_ );
+
+    const double P20 = -P_.tau_m_ / P_.c_m_ * expm1_tau_m;
+    const double P21_ex = -P_.tau_m_ * P_.tau_ex_ / ( P_.tau_m_ - P_.tau_ex_ )
+      / P_.c_m_ * ( expm1_tau_ex - expm1_tau_m );
+    const double P21_in = -P_.tau_m_ * P_.tau_in_ / ( P_.tau_m_ - P_.tau_in_ )
+      / P_.c_m_ * ( expm1_tau_in - expm1_tau_m );
+
```
stinebuu: Same as before, I think you can multiply with `P_.c_m_` instead of dividing.

Also, I generally find it a bit confusing with the division in the middle with multiplications around, but I know this is how it would be written on paper (when you have the e^(...) ), so it is fine.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+    const double expm1_tau_ex = numerics::expm1( -root / P_.tau_ex_ );
+    const double expm1_tau_in = numerics::expm1( -root / P_.tau_in_ );
+    const double expm1_tau_m = numerics::expm1( -root / P_.tau_m_ );
+
+    const double P20 = -P_.tau_m_ / P_.c_m_ * expm1_tau_m;
+    const double P21_ex = -P_.tau_m_ * P_.tau_ex_ / ( P_.tau_m_ - P_.tau_ex_ )
+      / P_.c_m_ * ( expm1_tau_ex - expm1_tau_m );
+    const double P21_in = -P_.tau_m_ * P_.tau_in_ / ( P_.tau_m_ - P_.tau_in_ )
+      / P_.c_m_ * ( expm1_tau_in - expm1_tau_m );
+
```
stinebuu: Same as before

testsuite/unittests/test_iaf_psc_exp_ps_lossless.sli
```diff
@@ -0,0 +1,176 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+ /* BeginDocumentation
+    Name: testsuite::test_iaf_psc_exp_ps_lossless - sli script for overall test of iaf_psc_exp model
+
```
stinebuu: iaf_psc_exp_lossless model

precise/iaf_psc_exp_ps_lossless.h
```diff
@@ -0,0 +1,486 @@
...
+Receives: SpikeEvent, CurrentEvent, DataLoggingRequest
+
+SeeAlso: iaf_psc_exp_ps
+*/
+
+namespace nest
+{
+/**
+ * Leaky iaf neuron, exponential PSC synapses, canonical implementation.
+ * @note Inherit privately from Node, so no classes can be derived
```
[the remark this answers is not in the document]
heplesser: Ok, will update.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+  def< double >( d, names::tau_syn_in, tau_in_ );
+  def< double >( d, names::t_ref, t_ref_ );
+}
+
+double
+nest::iaf_psc_exp_ps_lossless::Parameters_::set( const DictionaryDatum& d )
+{
+  // if E_L_ is changed, we need to adjust all variables defined relative to
+  // E_L_
+  const double ELold = E_L_;
```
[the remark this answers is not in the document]
heplesser: Done, for consistency as `E_L_old`.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+  if ( U_reset_ >= U_th_ )
+  {
+    throw BadProperty( "Reset potential must be smaller than threshold." );
+  }
+
+  if ( U_reset_ < U_min_ )
+  {
+    throw BadProperty(
+      "Reset potential must be greater equal minimum potential." );
```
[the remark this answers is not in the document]
heplesser: Done.

precise/iaf_psc_exp_ps_lossless.cpp
heplesser: I will need to look into this and its history.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+void
+RecordablesMap< iaf_psc_exp_ps_lossless >::create()
+{
+  // use standard names whereever you can for consistency!
+  insert_( names::V_m, &iaf_psc_exp_ps_lossless::get_V_m_ );
+  insert_( names::I_syn, &iaf_psc_exp_ps_lossless::get_I_syn_ );
+  insert_( names::I_syn_ex, &iaf_psc_exp_ps_lossless::get_I_syn_ex_ );
+  insert_( names::I_syn_in, &iaf_psc_exp_ps_lossless::get_I_syn_in_ );
+}
+}
```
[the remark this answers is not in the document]
heplesser: We need `namespace nest { ... }` here at the beginning for the *declaration* of the `RecordablesMap< iaf_psc_exp_ps_lossless >::create()` specialisation (which also is a definition); this has to happen in `namespace nest { ... }`.  All else are just definitions of methods that have already been declared in the header file, so we do not need the namespace around. Generally, we want to have as little as possible inside `namespace xyz { }` to avoid that we accidentally introduce names into a namespace.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,671 @@
...
+  const Parameters_& p ) const
+{
+  def< double >( d, names::V_m, y2_ + p.E_L_ ); // Membrane potential
+  def< bool >( d, names::is_refractory, is_refractory_ );
+  def< double >(
+    d, names::t_spike, Time( Time::step( last_spike_step_ ) ).get_ms() );
+  def< double >( d, names::offset, last_spike_offset_ );
+  def< double >( d, names::I_syn_ex, I_syn_ex_ );
+  def< double >( d, names::I_syn_in, I_syn_in_ ); // y1 state
+  def< double >( d, names::I_syn, I_syn_ex_ + I_syn_in_ );
```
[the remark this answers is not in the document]
heplesser: I removed the comment.

precise/iaf_psc_exp_ps_lossless.h
heplesser: I will leave it as it is (but fixed corresponding comments in `Variables_`,  since `y1_` is not used in this model.

precise/iaf_psc_exp_ps_lossless.cpp
heplesser: Absolutely.

precise/iaf_psc_exp_ps_lossless.cpp
heplesser: I think I will leave them as is, corresponding expressions look the same in all other models. Once we generate code from NESTML it will look differently anyways.

precise/iaf_psc_exp_ps_lossless.cpp
heplesser: This is related to #368. In a [comment there](https://github.com/nest/nest-simulator/issues/368#issuecomment-222008250), I actually proposed that one should remove the accumulation altogether. It would simplify code and ensure consistency (in the way discussed in #368) for all precise models. I will work on it.

heplesser:
I encountered a problem with the `issue-368.sli` regression test related to #368. When the model receives an excitatory spike (which does not evoke a spike on its own), and an inhibitory spike at the same time, then neuron will emit a spike in the past. This happens if we do not accumulate incoming spikes, which we should never do, as discussed in #368.

**This needs to be resolved before merging!**

ChristianKeup:
Hi, so I was assigned by Moritz to help completing this pull request.
Since this is my first adventure into the nest code and I don't know SLI, pls be patient with me!
I would like to know what exactly you want me to do:
- fix the unittest 
- expand the unittest to incorporate more features -> which ones?
- try to fix the problem related to #368 just mentioned ?
- more?

Also, do you want me to work directly with (i.e. push to) the heplesser:krishnanj-time-reversal-precise branch, or should I use my own branch that will then in the end be merged into nest:master? (sorry if this is clear)

heplesser:
@ChristianKeup Welcome aboard! You summarized the remaining tasks quite well. Some comments:
- Currently, three tests are failing. We will work on the problem with `ticket-659-copyright`, that seems to be some strange technicality.
- The `issue-77.sli` test is failing due to the problem related to #368; I will also try looking into it, but it would be good if you could take a look as well. As I wrote before, the problem is caused by spikes arriving at the same time with opposite signs.
- `test_iaf_psc_exp_ps_lossless.sli` fails because the reference values stored in the file are not correct. Only those values need to be updated and replaced by correct values that have been obtained differently. Since in that test the model is only driven by constant input, it should be possible to obtain the correct solution analytically.
- What is missing are tests that test whether the "new stuff" in the model works, i.e., the guaranteed detection of spikes also in corner cases. Does the paper contain examples of such cases we could use for the test? If you prefer, you could implement the extended tests in Python. 

Concerning the practical procedure: This PR is from heplesser:krishnanj-time-reversal-precise, i.e., branch `krishnanj-time-reversal-precise` in my Github fork of NEST (https://github.com/heplesser/nest-simulator). You should check out that branch, commit your changes to your local copy of the branch, the push it to your NEST fork on Github and create a PR against my branch (not master). I can then integrate your changes and they will end up here in the PR against master.

stinebuu:
@heplesser concerning `ticket-659-copyright`, there seems to be a whitespace after [`*  test_iaf_psc_exp_ps_lossless.sli`](https://github.com/heplesser/nest-simulator/blob/b39e3a25533113db22cc0771a699f3da5dc9e5bc/testsuite/unittests/test_iaf_psc_exp_ps_lossless.sli#L2) in the test file. Removing the whitespace makes the copyright error disappear.

ChristianKeup:
`test_iaf_psc_exp_ps_lossless.sli` seems to fail also because  `/iaf_psc_exp_ps_lossless` is not in the modeldict. 
I am trying to search for the reason by comparing to the implementation of the similar `/iaf_psc_exp_ps`, but if someone has an immediate idea about where this problem is likely originating from, the hint would be appreciated.

heplesser:
@stinebuu Thanks for the detective work!

ChristianKeup:
I have found a problem where the lossless neuron is behaving for small inputs close to threshold like a normal unprecise ias_psc_exp neuron. I'm currently still looking into it.

heplesser:
@ChristianKeup I have just pushed a change that solves the problem with issue-77.sil by explicitly skipping state propagation and spike check if two spikes arrive at the same time, so that `ministep==0`. I don't know if that will help the issue you are working on. The problem with the copyright check is also fixed.

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -414,18 +414,20 @@ nest::iaf_psc_exp_ps_lossless::update( const Time& origin,
...
+          const double spike_time_max = is_spike_( ministep );
 
-        if ( not numerics::is_nan( spike_time_max ) )
-        {
-          emit_spike_( origin, lag, V_.h_ms_ - last_offset, spike_time_max );
+          if ( not numerics::is_nan( spike_time_max ) )
+          {
+            emit_spike_( origin, lag, V_.h_ms_ - last_offset, spike_time_max );
+          }
         }
```
ChristianKeup: @heplesser  thanks!
If I see it correctly, the issue of a `ministep == 0` is handled slightly differently in `iaf_psc_exp_ps.cpp`, where this is done by checking that `dt != 0` in `propagate_( dt )` and `emit_spike( ... , dt )`.
I think it does the same, but just wanted to mention this in case you didn't do it differently on purpose

heplesser: @ChristianKeup Ah, I overlooked that. I will update my code.

ChristianKeup:
The problem I mentioned turned out to be due to external currents, as from a /dc_generator, not being taken into account in the spike detection method `is_spike( dt )`. It only took into account the external current given the model parameter `P_.I_e_`.
In the process I checked the equations implemented in `is_spike( dt )`. They are in fact correct and working. However, it turned out that eq. (49) in Krishnan et al. 2018  is not correct. Also, in the spike detection algorithms 1 and 2 in the paper, the functions g and f are interchanged. (I have mailed with Luca about this)
I changed to code to be less misleading and added a short comment in case someone else should ever try to understand the implementation.
Now the model is working nicely. I fixed the current unittest and also added the exact same unittest for the lossy `/iaf_psc_exp_ps` since none yet existed. In this simple test, both models behave exactly the same, as they should. I will write an extension testing for the region of previously missed spikes next week.
@heplesser I've created a PR with the changes against your branch, hopefully in the correct way.

heplesser:
@Silmathoron @stinebuu I believe this PR is now in order. Could you have another look? For background on some of the changes and design decisions, please see also the discussion on https://github.com/heplesser/nest-simulator/pull/22

precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -0,0 +1,710 @@
...
+
+      const double spike_time_max = is_spike_( V_.h_ms_ );
+      if ( not numerics::is_nan( spike_time_max ) )
+      {
+        emit_spike_( origin, lag, 0, spike_time_max );
+      }
+    }
+    else
+    {
+      // We only get here if there is at least on event,
```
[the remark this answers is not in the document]
heplesser: @stinebuu Fixed.

PR #903, opened 2018-03-07
Support linking against topology when building extension modules

### LINKED ISSUE #841: Linking error when registering masks or kernels during external module installation on macOS
Dear NEST users, 

I am trying without success to install in nest-2.12.0 the example external module 'MyModule' extended with the 'EllipticMask' class (as described in the Topology user manual) and experience a linking error related to the registration of the mask with TopologyModule when compiling MyModule on Mac OS. I would greatly appreciate if anyone could help!

You can find the module and installation scripts necessary to replicate the issue, as well as the command line outputs of the nest and module installation on my machine at the following address: https://drive.google.com/drive/folders/0B0gyPoMB36PFLVNYNzBSU0MtNjA?usp=sharing

I am working on macOS Sierra and installed nest manually using the following commands:

```
cmake ${WORKING_DIR}/nest-2.12.0-src \
    -DCMAKE_INSTALL_PREFIX:PATH=${WORKING_DIR}/nest-2.12.0 \
    -DCMAKE_C_COMPILER=/usr/local/bin/gcc-7 \
    -DCMAKE_CXX_COMPILER=/usr/local/bin/g++-7 \
    -Dwith-python=3 \
    -Dwith-optimize=ON \
make -j8
make install
source ${WORKING_DIR}nest-2.12.0/bin/nest_vars.sh
```

Nest works nicely at that point, but I run into an error when I try installing my module using the commands below (as described in the Developer manual):
```
# From the directory containing MyModule
mkdir bld
cd bld
cmake ../MyModule \
    -Dwith-nest=$NEST_CONFIG_PATH  # path to nest-config.sh from nest install
make
make install
```

The configuration step runs fine but here is the error returned during the 'make' command:
```
[ 50%] Linking CXX shared library libmymodule.dylib
Undefined symbols for architecture x86_64:
  "nest::TopologyModule::mask_factory_()", referenced from:
      mynest::MyModule::init(SLIInterpreter*)     in mymodule.cpp.o
  "nest::Mask<2>::minus_mask(nest::AbstractMask const&) const", referenced from:
      vtable for mynest::EllipticMask in mymodule.cpp.o
  "nest::Mask<2>::union_mask(nest::AbstractMask const&) const", referenced from:
      vtable for mynest::EllipticMask in mymodule.cpp.o
  "nest::Mask<2>::intersect_mask(nest::AbstractMask const&) const", referenced from:
      vtable for mynest::EllipticMask in mymodule.cpp.o
  "nest::Mask<2>::inside(std::vector<double, std::allocator<double> > const&) const", referenced from:
      vtable for mynest::EllipticMask in mymodule.cpp.o
  "nest::Mask<2>::outside(nest::Box<2> const&) const", referenced from:
      vtable for mynest::EllipticMask in mymodule.cpp.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make[2]: *** [libmymodule.dylib] Error 1
make[1]: *** [CMakeFiles/mymodule_lib.dir/all] Error 2
make: *** [all] Error 2
```
You can find the full console output on the google drive.

The error is related to the registration of the new mask with in the module init and disappears if I comment out the corresponding line in mymodule.cpp ( 
  `` nest::TopologyModule::register_mask<EllipticMask>("elliptic"); ``). In which case I can install the module using nest.Install('mymodule') and use the neuron and synapse models defined in the module.

A few more elements to the puzzle:
- I encounter the same type of error ('undefined symbols for architecture x86_64' for a bunch of references) when trying to register a kernel with the nest::TopologyModule::register_parameter() function (if I try to add the affine2d kernel following the Topology User manual)
- I can install, import and use the example MyModule directory provided without this error (so the registration of neuron and synapse models works fine)
- I tried installing the module using the ``-Dexternal-modules`` cmake flag during nest installation without success.
- I tried the same installation process on a different mac machine without success.

Thank you so much for your help !

Tom Bugnon

heplesser:
This PR fixes #841 by adding topology to `ALL_LIBS`, the CMake variable listing all libraries against which to link when, e.g., building extension modules. 

I assume it worked on some systems because there linkers delayed look-up until runtime. Since topology was loaded on NEST startup, the necessary symbols are available when loading the module at runtime.

@TomBugnon Could you test if this fix solves your problem?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,7 @@ set( ALL_LIBS
   "-lrandom"
   "-lsli"
   "-lnestkernel"
+  "-ltopology"
   "${OpenMP_CXX_FLAGS}"
   "${LTDL_LIBRARIES}"
   "${READLINE_LIBRARIES}"


TomBugnon:
That works for both masks and kernels, thank you so much. Good bye docker!

jougs:
Many thanks for addressing the issue. Given two positive reviews, I merge.

PR #909, opened 2018-03-12
Reformat and clean up Name definitions

jougs:
This PR removes the (mostly wrong and useless) comments from the headers that define `Name` objects, cleans up inconsistent definitions/declarations and removes unused `Name`s.

It also adds a regression test to check consistent definition/declaration and also that all `Name`s defined are actually used.


testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,99 @@
...
+        else:
+            names_defined.add(h)
+
+names_defined = list(names_defined)
+names_defined.sort()
+
+
+from subprocess import check_output
+
+names_used_raw = check_output(["grep", "-ro", "names::[A-Za-Z0-9_]*", source_dir])
```
jougs: This could be refactored using `os.walk`, iterating files and performing regex matching directly in Python. I settled on the `subprocess`/`grep` solution, as it requires less code. However, I'm open to discuss this decision and implement the pure Python solution if the reviewer thinks that that a self-contained implementation is better.

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+#
+# unused_names.py
```
heplesser: I would suggest `test_unusued_names.py` as name and placing this in unittests, as the test is not directly motivated by a reported issue.

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
...
+          "the source tree you want to check!")
+    sys.exit(EXIT_NO_SOURCE)
+
+# Base names of files that contain const Name definitions
+# (i.e. filenames without filename extension)
+names_files = [
+    "librandom/librandom_names",
+    "nestkernel/nest_names",
+    "topology/topology_names",
+]
```
heplesser: I don't think hard-coding the names is a good idea. Why not 
```
name_files = [n[:-2] for n in glob.glob('*/*_names.h')]
```

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
...
+for names_file in names_files:
+    fname = os.path.join(source_dir, names_file)
+
+    names_header = []
+    with open(fname + ".h") as names_file:
+        lines = names_file.readlines()
+        for line in lines:
+            if "extern const Name" in line:
+                names_header.append(line.split("Name ")[1].split(";")[0])
+    names_header.sort()
```
heplesser: Could you turn this into a function or loop to avoid code duplication? I also wonder if it would not be safer and more efficient to test with a regular expression, especially to avoid a that a single extra space confuses the test. Something like `extern\s+const\s+Name\s+([_\w]+)\s*;` and then get the name from the match group.

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
...
+
+# Base names of files that contain const Name definitions
+# (i.e. filenames without filename extension)
+names_files = [
+    "librandom/librandom_names",
+    "nestkernel/nest_names",
+    "topology/topology_names",
+]
+
+names_defined = set()
```
heplesser: I would make this a list. Then you can test later if `names_defined == unique(names_defined)`. This catches cases where automated merges duplicate lines---has happened a few times.

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
...
+            print(msg + "{} != {}".format(h, s))
+            sys.exit(EXIT_NAME_H_CPP_MISMATCH)
+        else:
+            names_defined.add(h)
+
+names_defined = list(names_defined)
+names_defined.sort()
+
+
+from subprocess import check_output
```
heplesser: Please collect all imports at the start of the file.

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
...
+
+grep_cmd = ["grep", "-ro", "names::[A-Za-Z0-9_]*", source_dir]
```
heplesser: Would you add a comment pointing out why you use Unix grep here instead of Python?

testsuite/regressiontests/unused_names.py
```diff
@@ -0,0 +1,101 @@
...
+for line in names_used_raw.split("\n"):
+    if "::" in line:
+        name = line.split("::")[1]
+        if name != "":
+            names_used.add(name)
+
+names_used = list(names_used)
+names_used.sort()
+
+for d, u in zip(names_defined, names_used):
```
heplesser: I think you could do this more efficiently as `set(names_defined) - set(names_used)`.

nestkernel/nest_names.h
```diff
@@ -31,613 +31,503 @@ namespace nest
...
- * Node::get_status and Node::set_status to make data exchange
- * more efficient and consistent. Creating a Name from a std::string
- * is in O(log n), for n the number of Names already created. Using
- * predefined names should make data exchange much more efficient.
+ * Node::get_status and Node::set_status to make data exchange more
+ * efficient and consistent. Creating a Name from a std::string is in
+ * O(log n), for n the number of Names already created. Using
+ * predefined names makes data exchange much more efficient as it
+ * uses integer comparisons instead of string comparisons internally.
  */
```
heplesser: Please extend this comment to explain the rules of ordering in this file.

nestkernel/nest_names.cpp
```diff
@@ -214,29 +197,21 @@ const Name histogram_correction( "histogram_correction" );
...
 const Name I_e( "I_e" );
-const Name I_ex( "I_ex" );
 const Name I_h( "I_h" );
-const Name I_in( "I_in" );
 const Name I_KNa( "I_KNa" );
-const Name I_L( "I_L" );
 const Name I_NaP( "I_NaP" );
 const Name I_stc( "I_stc" );
-const Name I_std( "I_std" );
-const Name I_syn( "I_syn" );
```
heplesser: Please keep `I_syn`, #812 will use it.

testsuite/regressiontests/unused_names.py
jougs: Sounds reasonable. I've renamed/moved the file in 69456a5a.

testsuite/regressiontests/unused_names.py
jougs: Isn't that a bit dangerous? I can imagine other files to also end in `_names.h` without actually defining `Name`s in the sense that the checker should check for. The current list of files was stable over quite some years and I assume that to be the case for some more...

testsuite/regressiontests/unused_names.py
jougs: I've turned it into a function using regex matching in 45a67f7c86.

testsuite/regressiontests/unused_names.py
jougs: I've refactored this together with the check for unused names in 5a58f65026e and think that catching errors with automatic merges (and other unrelated things) should not be covered by this test if it does not come for free.

testsuite/regressiontests/unused_names.py
jougs: Done in ac0306468827.

testsuite/regressiontests/unused_names.py
jougs: Done in ac0306468827.

testsuite/regressiontests/unused_names.py
jougs: I've refactored the test logic for unused `Name`s in 5a58f65026e.

nestkernel/nest_names.h
jougs: I've extended the comment in 50b99d1ed62.

nestkernel/nest_names.cpp
jougs: I've re-added it in f6bbc9305.

testsuite/regressiontests/unused_names.py
heplesser: OK.

testsuite/regressiontests/unused_names.py
heplesser: Ok.

heplesser:
The Open NEST Developer Meeting on 19 March has no objections to removing the comments from the Names files.

Issue #894, opened 2018-02-23
Direct comparison of floating point numbers in STDP


suku248:
Floating point numbers are tested for equality without using an error margin in `ArchivingNode` and all STDP models, which results in spikes being included/excluded by mistake in the STDP update.

The problem is shown [in this commit](https://github.com/suku248/nest-simulator/commit/76f7f82bb8fa13a7b63321739e40bd75e3007e9e) when running `hpc_benchmark.sli`.

In the long run, we need to use NEST's standard way of representing spike times as integer time stamps (and double offsets in case of off-grid spiking). This needs testing of possible implementations. For example, we need to check whether storing spikes as integer time stamps and double offsets in `ArchivingNode` instead of just storing double spike times results in performance penalties.
I am working on this but, for now, we might want to incorporate this [temporary fix](https://github.com/suku248/nest-simulator/commit/c0d88698b440f6492644aa37434f101fb2b3ccea).

heplesser:
@suku248 Your "temporary fix" seems sensible, but how robust would it be when precise neurons are combined with STDP? Would you create a PR based on the fix?

heplesser:
@suku248 I will close this as we have a fix per #924. Would you create a follow-up issue for a more thorough solution to the underlying problem?

Issue #861, opened 2017-11-27
Strange handling of zero-delta-t spike pairs in STDP in stdp_facetshw - innocent spikes get thrown away


aserenko:
This is a discussion issue about the implementation of the FACETS-STDP model.

[The following code in stdp_connection_facetshw_hom.h](https://github.com/nest/nest-simulator/blob/a1416d8e7707ec3cab97d7c315578cf5c84701d3/models/stdp_connection_facetshw_hom.h#L496) is said to implement symmetric nearest-neighbour spike pairing scheme:
```
  if ( start != finish ) // take only first postspike after last prespike
  {
    minus_dt = t_lastspike - ( start->t_ + dendritic_delay );
  }

  if ( start != finish ) // take only last postspike before current spike
  {
    --finish;
    plus_dt = ( finish->t_ + dendritic_delay ) - t_spike;
  }

  if ( minus_dt != 0 )
  {
    a_causal_ += std::exp( minus_dt / cp.tau_plus_ );
  }

  if ( plus_dt != 0 )
  {
    a_acausal_ += std::exp( plus_dt / cp.tau_minus_ );
  }
```

If a postsynaptic spike happens at exactly the same moment as the presynaptic one (leading to `minus_dt = 0`), such a pair is excluded from accounting. This, as far as I know, is done by convention (see, for example, [letter by Maximilian Schmidt on August 16, 2017 in the mailing list](http://mail.nest-initiative.org/cgi-bin/mailman/private/nest_user/2017/005131.html).

What I am complaining about is that we then do not come back to pick the next postsynaptic spike. For example, imagine presynaptic spikes at times, say, 10 and 30 and postsynaptic spikes at 10 and 20. Processing the update at time 30, we request the history of post-spikes accumulated so far, and get [10, 20]. Then we take the first spike at 10, omit it, and that's all, no facilitation at all during this update. Should we have proceeded next to the post-spike at 20 and accounted the pair post20-pre10 in this case?

aserenko:
As far as I understand, it should indeed be resolved now.

As soon as Archiving_Node::get_history() does not return events with time equal to its start time argument t1, events that would lead to accounting coinciding (zero-delta-t) spike pairs can never be returned by get_history. So, the comparison `if ( minus_dt != 0 )` just checks if `minus_dt`
or `plus_dt` retain their initial values, which can occur only if no events are returned by get_history().

PR #924, opened 2018-04-09
Numerical imprecision stdp

suku248:
This PR is a temporary fix for issue #894


models/stdp_dopa_connection.h
```diff
@@ -565,8 +567,9 @@ STDPDopaConnection< targetidentifierT >::send( Event& e,
     process_dopa_spikes_( dopa_spikes, t0, start->t_ + dendritic_delay, cp );
     t0 = start->t_ + dendritic_delay;
     minus_dt = t_last_update_ - t0;
-    if ( start->t_ < t_spike ) // only depression if pre- and postsyn. spike
-                               // occur at the same time
+    // only depression if pre- and postsyn. spike occur at the same time
+    if ( t_spike - start->t_ > 1.0
+        * kernel().connection_manager.get_stdp_eps() )
```
heplesser: I would drop the multiplication with `1.0`. I also don't understand the comment: the comment talks about pre- and post-spike being at the same time, but then you facilitate only if there is a non-zero time difference. Your code does the same thing as the original code, so my confusion also applies to the comment in the original code.

nestkernel/archiving_node.cpp
```diff
@@ -81,7 +88,9 @@ Archiving_Node::register_stdp_connection( double t_first_read )
   // For details see bug #218. MH 08-04-22
 
   for ( std::deque< histentry >::iterator runner = history_.begin();
-        runner != history_.end() && runner->t_ <= t_first_read;
+        runner != history_.end()
+          && ( t_first_read - runner->t_ > -1.0
+                 * kernel().connection_manager.get_stdp_eps() );
```
heplesser: Could you try to get the -1.0 on the same line as `kernel()....`? I think that would be good for legibility.

nestkernel/archiving_node.cpp
```diff
@@ -100,7 +109,8 @@ nest::Archiving_Node::get_K_value( double t )
   int i = history_.size() - 1;
   while ( i >= 0 )
   {
-    if ( t > history_[ i ].t_ )
+    if ( t - history_[ i ].t_ > 1.0
+        * kernel().connection_manager.get_stdp_eps() )
```
heplesser: Could you try to get the -1.0 on the same line as `kernel()....`? I think that would be good for legibility.

nestkernel/archiving_node.cpp
```diff
@@ -159,12 +170,16 @@ nest::Archiving_Node::get_history( double t1,
   else
   {
     std::deque< histentry >::iterator runner = history_.begin();
-    while ( ( runner != history_.end() ) && ( runner->t_ <= t1 ) )
+    while ( ( runner != history_.end() )
+      && ( t1 - runner->t_ > -1.0
+                * kernel().connection_manager.get_stdp_eps() ) )
```
heplesser: Could you try to get the -1.0 on the same line as `kernel()....`? I think that would be good for legibility.

nestkernel/archiving_node.cpp
```diff
@@ -159,12 +170,16 @@ nest::Archiving_Node::get_history( double t1,
...
     {
       ++runner;
     }
     *start = runner;
-    while ( ( runner != history_.end() ) && ( runner->t_ <= t2 ) )
+    while ( ( runner != history_.end() )
+      && ( t2 - runner->t_ > -1.0
+                * kernel().connection_manager.get_stdp_eps() ) )
```
heplesser: Could you try to get the -1.0 on the same line as `kernel()....`? I think that would be good for legibility.

nestkernel/archiving_node.cpp
```diff
@@ -126,7 +136,8 @@ nest::Archiving_Node::get_K_values( double t,
   int i = history_.size() - 1;
   while ( i >= 0 )
   {
-    if ( t > history_[ i ].t_ )
+    if ( t - history_[ i ].t_ > 1.0
+        * kernel().connection_manager.get_stdp_eps() )
```
heplesser: Drop multiplication by 1.0?

nestkernel/connection_manager.h
```diff
@@ -419,6 +423,10 @@ class ConnectionManager : public ManagerInterface
 
   //! Capacity growth factor to use beyond the limit
   double large_connector_growth_factor_;
+
+  //! Maximum distance between (double) spike times in STDP that is
+  //! still considered 0
```
heplesser: I think it would be useful to refer to the issue here.

nestkernel/nestmodule.cpp
```diff
@@ -1699,6 +1699,24 @@ NestModule::DisableStructuralPlasticity_Function::execute(
   i->EStack.pop();
 }
 
+/**
+ * Set epsilon that is used for comparing spike times in STDP.
+ * Spike times in STDP synapses are currently represented as double
+ * values. The epsilon defines the maximum distance between spike
+ * times that is still considered 0.
```
heplesser: Maybe add a `Note: See issue ...`?

nestkernel/nestmodule.cpp
```diff
@@ -1794,6 +1812,9 @@ NestModule::init( SLIInterpreter* i )
     "GetStructuralPlasticityStatus", &getstructuralplasticitystatus_function );
   i->createcommand( "Disconnect", &disconnect_i_i_lfunction );
   i->createcommand( "Disconnect_g_g_D_D", &disconnect_g_g_D_Dfunction );
+
+  i->createcommand( "SetStdpEps", &setstdpeps_dfunction );
```
heplesser: Should we have a `GetStdpEps` function, too?  Or maybe "hide" this in the kernel status dict after all?

models/stdp_dopa_connection.h
```diff
@@ -452,7 +452,8 @@ STDPDopaConnection< targetidentifierT >::process_dopa_spikes_(
   // process dopa spikes in (t0, t1]
   // propagate weight from t0 to t1
   if ( ( dopa_spikes.size() > dopa_spikes_idx_ + 1 )
-    && ( dopa_spikes[ dopa_spikes_idx_ + 1 ].spike_time_ <= t1 ) )
+    && ( t1 - dopa_spikes[ dopa_spikes_idx_ + 1 ].spike_time_ > -1.0
+           * kernel().connection_manager.get_stdp_eps() ) )
```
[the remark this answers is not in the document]
suku248: I am afraid the coding style guidelines do not permit this.

models/stdp_dopa_connection.h
```diff
@@ -468,7 +469,8 @@ STDPDopaConnection< targetidentifierT >::process_dopa_spikes_(
     // process remaining dopa spikes in (t0, t1]
     double cd;
     while ( ( dopa_spikes.size() > dopa_spikes_idx_ + 1 )
-      && ( dopa_spikes[ dopa_spikes_idx_ + 1 ].spike_time_ <= t1 ) )
+      && ( t1 - dopa_spikes[ dopa_spikes_idx_ + 1 ].spike_time_ > -1.0
+                * kernel().connection_manager.get_stdp_eps() ) )
```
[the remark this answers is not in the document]
suku248: Again, not possible due to coding style guidelines.

nestkernel/archiving_node.cpp
suku248: Again, not possible due to coding style guidelines.

nestkernel/archiving_node.cpp
suku248: Again, not possible due to coding style guidelines.

nestkernel/nestmodule.cpp
suku248: Yes, good point. Done.

nestkernel/nestmodule.cpp
suku248: No, I don't think that we want a `GetStdpEps`. As this is a temporary fix, I would not like to make the variable visible in the kernel dict. I have added an info message to `SetStdpEps` though, which states the new value of `stdp_eps` and I have added two checks.

nestkernel/archiving_node.cpp
suku248: Yes, thanks, done.

nestkernel/connection_manager.h
suku248: Yes, you are right. Done.

nestkernel/archiving_node.cpp
suku248: It's a +1.0 and I have removed it.

nestkernel/archiving_node.cpp
suku248: Again, not possible due to coding style guidelines.

models/stdp_dopa_connection.h
suku248: I agree and have removed the 1.0.

The comment is misleading as it's about synaptic depression. I have adjusted it such that it explains that facilitation is skipped in case of 0 temporal difference between pre- and post spike.

models/stdp_dopa_connection.h
```diff
@@ -565,8 +567,9 @@ STDPDopaConnection< targetidentifierT >::send( Event& e,
     process_dopa_spikes_( dopa_spikes, t0, start->t_ + dendritic_delay, cp );
     t0 = start->t_ + dendritic_delay;
     minus_dt = t_last_update_ - t0;
-    if ( start->t_ < t_spike ) // only depression if pre- and postsyn. spike
-                               // occur at the same time
+    // faciltate only in case of post- after presyn. spike
```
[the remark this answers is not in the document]
suku248: Thanks, fixed!

models/stdp_pl_connection_hom.h
```diff
@@ -58,6 +58,8 @@
 
 // C++ includes:
 #include <cmath>
+#include <iomanip>
+#include <limits>
```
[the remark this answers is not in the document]
suku248: That's correct. I have removed them.

nestkernel/archiving_node.cpp
```diff
@@ -29,6 +29,13 @@
 
 #include "archiving_node.h"
 
+// C++ includes:
+#include <iomanip>
+#include <limits>
```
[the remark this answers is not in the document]
suku248: That's correct. I have removed them.

suku248:
Thank you @hakonsbm for the comments!

aserenko:
@suku248, could you possibly be so kind to have a look at https://github.com/nest/nest-simulator/blob/79bbbb08d0ef209f0c8295339abc3c137a321137/models/stdp_connection_facetshw_hom.h#L507 ?

Firstly, there is also a direct comparison to zero that should be corrected in the same way.

The second question is a bit deeper. As far as I understand, now that you have changed nestkernel/archiving_node.cpp, the Archiving_Node::get_history() does not return events with time equal to its start time argument t1. And so, events that would lead to accounting coinciding (zero-delta-t) spike pairs can never be returned by get_history.
That way, the behaviour of this very model that I was wanting to discuss in issue #861 seem to have been changed to the way I argued for in that issue. Do you too think it has?

suku248:
Yes, you are right. I have simply overlooked `stdp_connection_facetshw_hom.h`.

I am not really familiar with this model. Do you have some experience with model and do you think you could correct the comparisons of double values for this model and then generate a PR? That would be great.

I think, it's possible that this will solve #861

aserenko:
Sorry, mostly a false alarm. There was an initialization a few lines higher:
```
  double minus_dt = 0;
  double plus_dt = 0;
```
So, now (as soon as the current PR has been merged) the direct comparison to zero here just checks if `minus_dt` or `plus_dt` retain their initial values, which basically means the same as checking `if ( start != finish )`.

Still, I could not resist the temptation to mess with the code and opened a PR (#952) anyway. However, it is not that necessary, and I won't mind if it is rejected.

PR #929, opened 2018-04-17
Added test for spatial kernels in topology (depends on #823).

heplesser:
This PR adds test for spatial kernels to NEST, based on code from Daniel Hjertholm's master thesis.

It can only be merged once #823 is merged, since it requires the gamma kernel.

@babsey Could you also review?

### DIFF
--- a/topology/pynest/tests/test_spatial_kernels.py
+++ b/topology/pynest/tests/test_spatial_kernels.py
@@ -123,7 +123,7 @@ class SpatialTester(object):
         self._gauss = lambda D: (self._params['c'] +
                                  self._params['p_center'] *
                                  math.exp(-(D - self._params['mean']) ** 2 /
-                                        (2. * self._params['sigma'] ** 2)))
+                                          (2. * self._params['sigma'] ** 2)))
         self._gamma = lambda D: (D ** (self._params['kappa'] - 1) /
                                  (self._params['theta'] **
                                   self._params['kappa'] *
@@ -503,7 +503,7 @@ class SpatialTester(object):
         variance_num_targets = sum([p * (1. - p) for p in ps])
 
         if variance_num_targets == 0:
-            return math.nan, 1.0
+            return np.nan, 1.0
         else:
             sd = math.sqrt(variance_num_targets)
             z = abs((num_targets - expected_num_targets) / sd)
@@ -593,7 +593,7 @@ if PLOTTING_POSSIBLE:
             plt.plot(x, y, color='black', linewidth=3, label='Theory',
                      zorder=1)
             plt.hist(self._target_dists, bins=bins, histtype='step',
-                     linewidth=1, density=True, color='red',
+                     linewidth=1, normed=True, color='red',
                      label='Empirical', zorder=2)
             plt.ylim(ymin=0.)
 
--- a/topology/pynest/tests/test_spatial_kernels.py
+++ b/topology/pynest/tests/test_spatial_kernels.py
@@ -54,8 +54,7 @@ try:
 except:
     PLOTTING_POSSIBLE = False
 
-# If False, tests will be run; otherwise, as single case will be
-# plotted.
+# If False, tests will be run; otherwise, a single case will be plotted.
 DEBUG_MODE = False
 
 # Constant defining sensitivity of test (minimal p-value to pass)
--- a/topology/pynest/tests/test_spatial_kernels.py
+++ b/topology/pynest/tests/test_spatial_kernels.py
@@ -31,6 +31,7 @@ structured neural networks, MSc thesis, Norwegian University of
 Life Science, 2013. http://hdl.handle.net/11250/189117.
 """
 
+import math
 import numpy as np
 import numpy.random as rnd
 import scipy.integrate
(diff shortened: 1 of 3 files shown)


topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,811 @@
...
+
+        Parameters
+        ----------
+            D: Distance in interval [0, max_dist].
+
+        Return values
+        -------------
+            Unnormalized PDF at distance D.
+        '''
+
```
stinebuu: This function is quite long and cluttered. I had a very quick look in the thesis, but when I couldn't find a total definition of the PDF I gave up. Looking at this function, it does make sense that the definition of the PDF is scattered or not written down explicitly in the thesis, as it has so many different constraints/conditions, but I have to admit that I have not looked at the math here in detail. Seems to be working though, as the test pass! Some comments might be helpful, but it might just make the function more dense.

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,811 @@
...
+        '''
+        Construct a test object.
+
+        Parameters
+        ----------
+            seed         : Random seed for test
+            dim          : Dimensions (2 or 3)
+            L: Side length of area / volume.
+            N: Number of nodes.
+            open_bc      : Network with open boundary conditions
```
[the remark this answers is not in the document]
heplesser: Done.

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,811 @@
...
+import matplotlib.pyplot as plt
+
+# Constants defining sensitivity of test (minimal p-value to pass)
+P_MIN = 0.1
+SEED = 1234567
+
+
+class SpatialTester(object):
+    '''Tests for spatially structured networks.'''
+
```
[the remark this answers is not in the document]
heplesser: Removed the subclass.

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,811 @@
...
+        self._build()
+        self._connect()
+
+        self._target_dists = sorted(self._target_distances())
+        self._all_dists = self._all_distances()
+
+    def _reset(self, seed):
+        '''Reset simulator and seed PRNGs.'''
+
+        pass
```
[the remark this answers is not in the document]
heplesser: All gone.

topology/pynest/tests/test_spatial_kernels.py
heplesser: I don't have more details at the moment, but I will add documentation when it becomes available (as publication). I hope that is good enough for now.

heplesser:
@stinebuu I have restructured and simplified the code. I decided to keep the plotting in for now (but split it out into a separate subclass), since code placement would otherwise have become a bit problematic when supporting a test without plotting and an example with plotting. I have also merged all recent changes from master, so #823 with the gamma kernel is now included.

babsey:
Thank you for asking me to review it. 
I had a great opportunity to review 790 lines. Please forgive me for taking some time. :smile: 

Some technical points concerning me are:
- DEBUG_MODE is on. I assume it should be off.
- the code has functions (sqrt,exp) from math and some from numpy
- np.sum could be faster than sum.

I do not consider the correct outcome of the spatial testing because I trust @heplesser  and @stine that this testing suite is working.

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,780 @@
...
+
+import nest
+import nest.topology as topo
+
+
+try:
+    # for debugging
+    from mpl_toolkits.mplot3d import Axes3D
+    import matplotlib.pyplot as plt
+    # make sure we can open a window; DISPLAY may not be set
```
stinebuu: What does DISPLAY refer to?

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,780 @@
...
+    fig = plt.figure()
+    plt.close(fig)
+    PLOTTING_POSSIBLE = True
+except:
+    PLOTTING_POSSIBLE = False
+
+# If False, tests will be run; otherwise, as single case will be
```
stinebuu: I assume you mean 'a single case...' and not 'as single case..'!

heplesser:
@babsey I have now also explicitly added as a reviewer, so you get a nice "Submit review" button with "comment/require changes/approve" radio buttons to click :).

- `DEBUG_MODE` is off now, I accidentally pushed a version with it on first.
- I will use `math` everywhere except where I really need NumPy. The complicated case distinctions in `_pdf` operate on scalars only, and for them `math` versions are faster than `np` versions of mathematical functions.
- Since the operands to `sum()` are lists, not arrays, I think `np.sum()` would not make much of a difference. Anyways, the main optimization potential for these tests is in `_pdf` and `_cdf`, but that will need to wait till another time.

I will push in a little while.

topology/pynest/tests/test_spatial_kernels.py
babsey: It's an environment variable for X11 programs. You can check yours in terminal.
`echo $DISPLAY

In this case matplotlib needs to have DISPLAY environment for opening a window.

topology/pynest/tests/test_spatial_kernels.py
heplesser: Fixed.

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,779 @@
...
+            '''
+
+            plt.figure()
+            x = np.linspace(0, self._max_dist, 1000)
+            area = scipy.integrate.quad(self._pdf, 0, self._max_dist)[0]
+            y = np.array([self._pdf(D) for D in x]) / area
+            plt.plot(x, y, color='black', linewidth=3, label='Theory',
+                     zorder=1)
+            plt.hist(self._target_dists, bins=bins, histtype='step',
+                     linewidth=1, density=True, color='red',
```
babsey: the argument density works with matplotlib 2.2 but for older version it would give an error. Older versions has this argument but it called normed.

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,779 @@
...
+        self._kernel = kernels[kernel_name]
+
+        default_params = {
+            'constant': 1.,
+            'linear': {'a': -math.sqrt(2) / self._L, 'c': 1.0},
+            'exponential': {'a': 1.0, 'c': 0.0, 'tau': -self._L /
+                            (math.sqrt(2) * math.log((.1 - 0) / 1))},
+            'gaussian': {'p_center': 1., 'sigma': self._L / 4.,
+                         'mean': 0., 'c': 0.},
+            'gamma': {'kappa': 3., 'theta': self._L / 4.}}
```
babsey: I tested it, an adequate parameter set for gamma would be 
` 'gamma': {'kappa': 4., 'theta': self._L / 20.}}`

topology/pynest/tests/test_spatial_kernels.py
```diff
@@ -0,0 +1,779 @@
...
+
+        num_targets = len(self._target_dists)
+
+        ps = ([max(0., min(1., self._kernel(D)))
+               for D in self._all_dists])
+        expected_num_targets = sum(ps)
+        variance_num_targets = sum([p * (1. - p) for p in ps])
+
+        if variance_num_targets == 0:
+            return math.nan, 1.0
```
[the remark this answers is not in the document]
heplesser: Right, `math.nan` was introduced with Py 3.5 only.

topology/pynest/tests/test_spatial_kernels.py
heplesser: Ok, reverted to `normed` for backward compatibility.

topology/pynest/tests/test_spatial_kernels.py
heplesser: In which sense adequate? Wouldn't essentially any parameter set with kappa != 1 be useful?

babsey: You are right, it is an personal view.

Issue #414, opened 2016-06-30
Remove topology manual figures from git


heplesser:
This is a follow-up to #408. The figure files created while building the Topology User Manual should be deleted after the PDF is built and should not be committed to the Git repo.

heplesser:
This will fixed by #930.

heplesser:
Fixed by #930, therefore closing.

PR #930, opened 2018-04-18
Updated Topology User Manual for NEST 2.16.

heplesser:
The only changes (I know of) is that the version number is updated and that the gamma kernel is now also in the PDF, thus bringing documentation up to date after merging #823.

In the review, please check that all code examples are in the text, they have been missing at times.

This PR also removes the PNG-figure files included in the PDF from Github, no need to commit them. See #414.


topology/doc/make_manual.sh
```diff
@@ -42,5 +43,6 @@ ${MKIDX}  ${fname}
 ${LATEX}  ${fname}
 ${LATEX}  ${fname}
 
-# remove temporary files
+# remove temporary files and figures
 rm -f ${fname}.{aux,bbl,blg,idx,ilg,ind,ins,lof,log,lot,synctex.gz,toc}
+rm -rf user_manual_figures
```
hakonsbm: End the last line of the file with a newline.

heplesser: Done.

PR #931, opened 2018-04-19
Fix parsing of ~/.nestrc in show_help_with_pager()

physicalist:
This fixes a bug in the code parsing the pager command in ~/.nestrc which
could lead to a TypeError exception when calling subprocess.call with
pager=None if ~/.nestrc is an empty file (0 bytes).
I discovered this when @babsey and I discussed his changes in #928 and #913.
In addition, the code now opens ~/.nestrc safely using `with` and uses
only one loop and regular expression for parsing and returns the last
`/page << /command` entry instead of the first it finds.

### DIFF
--- a/pynest/nest/lib/hl_api_helper.py
+++ b/pynest/nest/lib/hl_api_helper.py
@@ -504,37 +504,27 @@ def show_help_with_pager(hlpobj, pager):
     consolepager = ['less', 'more', 'vi', 'vim', 'nano', 'emacs -nw',
                     'ed', 'editor']
 
-    # reading ~/.nestrc lookink for pager to use.
+    # check for pager command in ~/.nestrc if none provided in function call
     if pager is None:
-        # check if .netsrc exist
-        rc_file = os.path.join(os.environ['HOME'], '.nestrc')
-        if os.path.isfile(rc_file):
-            # open ~/.nestrc
-            rc = open(rc_file, 'r')
-            # The loop goes through the .nestrc line by line and checks
-            # it for the definition of a pager. Whether a pager is
-            # found or not, this pager is used or the standard pager 'more'.
-            for line in rc:
-                # the re checks if there are lines beginning with '%'
-                rctst = re.match(r'^\s?%', line)
-                if rctst is None:
-                    # the next re checks for a sth. like
-                    # '/page << /command (more)'
-                    # and returns the given pager.
-                    pypagers = re.findall(
-                        r'^\s?/page\s?<<\s?/command\s?\((\w*)', line)
-                    if pypagers:
-                        for pa in pypagers:
-                            if pa:
-                                pager = pa
-                            else:
-                                pager = 'more'
-                        break
-                    else:
-                        pager = 'more'
-            rc.close()
-        else:
-            pager = 'more'
+        # default to `more`
+        pager = 'more'
+
+        # Try to open ~/.nestrc safely
+        try:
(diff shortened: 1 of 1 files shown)


pynest/nest/lib/hl_api_helper.py
```diff
@@ -504,37 +504,27 @@ def show_help_with_pager(hlpobj, pager):
...
+                    # definition of a pager using a regular expression
+                    # which matches lines of the format
+                    # ' /page << /command (more)'
+                    # but ignores comments (via the `(?!%)` part)
+                    for pa in re.findall(
+                            r'^\s?(?!%)\s?\/page\s?<<\s?\/command\s?\((\w*)',
+                            line):
+                        pager = pa
+        except:
+            print('Failed to open ~/.nestrc, using default pager')
```
heplesser: I would rather write "parse" than "open", since in principle an exception could occur during reading as well.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -504,37 +504,27 @@ def show_help_with_pager(hlpobj, pager):
...
+            pass
```
heplesser: `pass` is not needed here, please remove it.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -504,37 +504,27 @@ def show_help_with_pager(hlpobj, pager):
...
```
jougs: Why are you parsing the `nestrc` file manually? It is run by SLI upon startup, so you can just do something like
```
nest.sli_func("/page GetOptions")["command"]
```
to get the command in a much safer way. To have a default pager available in case `nestrc` does not define one, you can use
```
try:
    pager = nest.sli_func("/page GetOptions")["command"]
except:
    pager = 'more'
````

File: pynest/nest/lib/hl_api_helper.py [hunk 4]
```diff
@@ -504,37 +504,27 @@ def show_help_with_pager(hlpobj, pager):
...
+                    # Go through the .nestrc line by line to check for the
+                    # definition of a pager using a regular expression
+                    # which matches lines of the format
+                    # ' /page << /command (more)'
+                    # but ignores comments (via the `(?!%)` part)
+                    for pa in re.findall(
+                            r'^\s?(?!%)\s?\/page\s?<<\s?\/command\s?\((\w*)',
+                            line):
+                        pager = pa
+        except:
```
heplesser: [PEP8](https://www.python.org/dev/peps/pep-0008/#id51) advises against bare except clauses.  `except IOError` seems appropriate here.

pynest/nest/lib/hl_api_helper.py
heplesser: @jougs The parsing code was introduced over a year ago in bdc87b7 and merged with #688. I reviewed then, and as now entirely overlooked the fact that we know the parser once NEST is loaded. So @physicalist, the solution proposed by @jougs is the way to go.

pynest/nest/lib/hl_api_helper.py
physicalist: True, I meant to look up the most general but still appropriate exception class but forgot to do that :)

jougs: For the above solution the exception class would simply be `NESTError`.

heplesser:
Superseeded by #948.

Issue #329, opened 2016-05-02
Multiple events for a single spike in hh_cond_exp_traub when increasing time resolution


astoeckel:
When running a simulation with the hh_cond_exp_traub neuron model, an increased time resolution (a time step smaller than the default 0.1ms), causes single output spikes in the HH model to be issued multiple times. The cause of issue should be

``` c++
void
nest::hh_cond_exp_traub::calibrate()
{
  B_.logger_.init(); // ensures initialization in case mm connected after Simulate
  V_.RefractoryCounts_ = 20;
  V_.U_old_ = S_.y_[ State_::V_M ];
}
```

in `hh_cond_exp_traub.cpp` where the number of ticks in `V_.RefractoryCounts_` is initialised with the constant `20` instead of calculating the number of ticks depending on the time step.

However, I'm not convinced that introducing an artificial, time-dependent refractoriness to the Hodgkin-Huxley model is actually the right way to go (actually, this causes bug #330). In my implementation of the HH model [1] I simply set a boolean `in_refrac` to true once a spike is issued (at the maximum membrane potential above a certain threshold) and reset this variable once the membrane potential drops below the threshold potential. This way no assumption about timings is made.

[1] https://github.com/hbp-sanncs/cinder/blob/ede53f1070b2b0a4eeed90278b61eea6db4c775d/cinder/models/neurons/hodgkin_huxley.hpp#L197

Silmathoron:
I agree with @astoeckel that it seems the way refractoriness is considered in the ``hh_cond_exp`` model is rather weird: on hybrid models such as the ``iaf`` or ``aeif``, the voltage is clamped during the whole refractory period so there is no risk of missing spike. Such a feature can be necessary to get closer to real neural behaviour because these simple models have no intrinsic equivalent of refractoriness.

On the contrary, the ``hh`` model has a current that induces to a "real" refractory period, so implementing an "artificial" refractory period which is not related to the intrinsic parameters __and__ does not clamp the potential seems a bit strange... and indeed seems to lead to problems.

Wouldn't it be better to propose either no such refractory period or one that (as in the other models) clamps the potential?

heplesser:
@astoeckel Thank you for reporting this and #330. I suggest we discuss both issues here.
@Silmathoron Thanks for the reference to the PyNN for Brian implementation.

[...]

PR #944, opened 2018-04-27
Better documentation of hh_cond_exp_traub and Brette-2007-compatible refractory time

heplesser:
When fixing #473 to make t_ref configurable in hh_cond_exp_traub, we configured the model with `t_ref==0` as default. This does not work properly, since the spike detection mechanism relies on an absolute refractory peroid. This PR sets the default `t_ref=2`ms as used in the Brette et al (2007) review and documents the relation of this model to the review more clearly, as well as differences from the Traub&Miles (1991) model.

See https://github.com/nest/nest-simulator/issues/329#issuecomment-385411618 for a thorough discussion.

### DIFF
--- a/doc/quickref.html
+++ b/doc/quickref.html
@@ -485,15 +485,15 @@ properties.</p>
     </tr>
     <tr>
       <td width="25%"><a href="help/cc/hh_psc_alpha.html">hh_psc_alpha</a></td>
-      <td>Hodgkin Huxley neuron model.</td>
+      <td>Hodgkin-Huxley neuron model.</td>
     </tr>
     <tr>
       <td width="25%"><a href="help/cc/hh_psc_alpha_gap.html">hh_psc_alpha_gap</a></td>
-      <td>Hodgkin Huxley neuron model with gap-junction support.</td>
+      <td>Hodgkin-Huxley neuron model with gap-junction support.</td>
     </tr>
     <tr>
       <td width="25%"><a href="help/cc/hh_cond_exp_traub.html">hh_cond_exp_traub</a></td>
-      <td>Hodgkin Huxley based model, Traub modified.</td>
+      <td>Hodgkin-Huxley model for Brette et al (2007) review.</td>
     </tr>
   </tbody>
 </table>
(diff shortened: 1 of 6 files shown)


heplesser:
@Silmathoron This neuron model is used for the "hh_coba" benchmark in the Brette et al (2007) simulator review, so we ought to keep it in NEST. Unfortunately, the paper text does not describe spike or reset mechanisms at all. Benchmark code for different simulators [is available on ModelDB](https://senselab.med.yale.edu/ModelDB/ShowModel.cshtml?model=83319&file=/destexhe_benchmarks/#tabs-2), although it is not easy to infer how spike detection is done by the different simulators. The most transparent is Brian:
```Python
P=NeuronGroup(4000,model=eqs,
              threshold=EmpiricalThreshold(threshold=-20*mV,refractory=3*ms),
              implicit=True,freeze=True,compile=False)
```
I.e., a fixed threshold at -20 mV and an absolute refractory period of 3 ms. Since no reset is specified, I assume that without the refractory period, Brian would emit spikes for each time step until membrane potential drops below -20 mV.

I have not been able to find threshold conditions for other simulators, although for Neuron this piece of code
```
proc connect2target() { //$o1 target point process, $o2 returned NetCon
  soma $o2 = new NetCon(&v(1), $o1)
  $o2.threshold = -10 //*	
}
```
looks like there may be a spike threshold of 10 mV.

Quite a mess, but maybe compatibility with the Brian implementation would not be a bad idea?

heplesser:
See also #329.

heplesser:
@Silmathoron @diesmann I have updated the PR after discussions, so default t_ref now is 2 ms. I also made the connection to Brette et al (2007) clearer in the documentation. Could you take a look?

Silmathoron:
@heplesser so all the changes mentioned in #329 including the change in spike detection would be for NEST 3?

heplesser:
@Silmathoron Yes, they would all be for NEST 3 and then in a new model, while the current `hh_cond_exp_traub` would be kept primarily to keep the Brette et al (2007) examples runnning.

PR #960, opened 2018-05-30
Support for gap junctions in Topology

### LINKED ISSUE #904: [feature request] Create multiple synapses among neuron pairs with a single tp.ConnectLayers
Dear all,

I would like to connect the neurons in two layers according to a given spatial profile and using multiple receptors (GABA_A & GABA_B, or AMPA & NMDA in my case). That is to say, I would like to use nest.Topology to draw the pre/post-synaptic neuron pairs, and connect each of those pairs multiple times with different types of synapses (one for each receptor type in my case). This seems rather natural to have this feature since this is what happens in the brain.
 
However, as of today I couldn't think of a way to do so easily using `tp.ConnectLayers`. If I were to call `tp.ConnectLayers()` multiple times with different synapse models, the pre/post-synaptic pairs would be different between the two calls. The only option I can think of right now would be to gather the connections with `nest.GetConnections()` and connect each pair another time with a different synapse model. This is certainly doable but...
Wouldn't it be great if the `synapse_model` value in the connection parameters could be a list of synapse models. ConnectLayers() could then create a single synapse for each model in that list between each of the drawn neuron pairs.

There might be a good reason why this is not already an option that I am not aware of, but is this a feature that you would otherwise consider incorporating in future versions? 
Thanks!

### LINKED ISSUE #939: [feature request] Support gap junctions in tp.ConnectLayers
Hi all,

I am noticing that gap junctions currently can't be created via the topology module's `ConnectLayers` function.  It would make sense to have support for this as gap junctions -like any other connection- are topologically organized in the brain, so I thought I would just raise the issue... (also, it would help me in my research :) )

Two immediate problems are that ConnectLayers doesn't seem to support the `make_symmetric` flag (this would be a very nice upgrade in itself as well, by the way) and uses a default value for the `delays` parameter although gap junctions don't have any. Are there more fundamental reasons why this is not supported?

In the meantime, the workaround proposed by Hans [in this issue](https://github.com/nest/nest-simulator/issues/904) of connecting layers with ConnectLayers using labelled synapses, querying the connections by label and reconnecting with `nest.Connect` does the trick here as well (although very inefficiently): 

```
l = topo.CreateLayer({'rows': 2, 'columns': 3, 'elements': 'hh_psc_alpha_gap'})

# Create dummy topological connections with labels and null weights
nest.SetDefaults('static_synapse_lbl', {'synapse_label': 1})
topo.ConnectLayers(l, l, {'connection_type': 'convergent',
                              'mask': {'circular': {'radius': 0.2 }},
                              'synapse_model': 'static_synapse_lbl',
                              'weights': 0.})

# Create gap junctions between the units connected by the dummy ConnectLayers call
conns = nest.GetConnections(source=nest.GetChildren(l)[0], synapse_label=1)
srcs, tgts, _, _, _ = zip(*conns)
nest.Connect(srcs, tgts, {'rule': 'one_to_one', 'make_symmetric': True}, {'model': 'gap_junction', 'weight': 1.0})
```

Cheers,
Tom

hakonsbm:
This PR adds support for gap junctions in `ConnectLayers`. It also adds a `make_symmetric` option, as this is needed by the gap junction synapse model.

Fixes #939.

### DIFF
--- a/testsuite/unittests/test_gap_junction.sli
+++ b/testsuite/unittests/test_gap_junction.sli
@@ -204,5 +204,83 @@ M_ERROR setverbosity
   and
 } assert_or_die
 
+% Check that we can use gap junction with target-driven connect in topology
+{
+  ResetKernel
+  /lr << /rows 2
+         /columns 3
+         /elements /hh_psc_alpha_gap      
+      >> CreateLayer def
+  
+  lr lr << /connection_type (convergent)
+           /synapse_model /gap_junction
+           /make_symmetric true
+           /weights 0.5
+        >> ConnectLayers
+} pass_or_die
+
+% Check that we can use gap junction with source-driven connect in topology
+{
+  ResetKernel
+  /lr << /rows 2
+         /columns 3
+         /elements /hh_psc_alpha_gap      
+      >> CreateLayer def
+  
+  lr lr << /connection_type (divergent)
+           /synapse_model /gap_junction 
+           /make_symmetric true
+           /weights 0.5
+        >> ConnectLayers
+} pass_or_die
+
+% Check that we can use gap junction with convergent connect in topology
+{
+  ResetKernel
+  /lr << /rows 2
+         /columns 3
+         /elements /hh_psc_alpha_gap      
+      >> CreateLayer def
+  
+  lr lr << /connection_type (convergent)
+           /synapse_model /gap_junction 
+           /make_symmetric true
+           /weights 0.5
+           /number_of_connections 3
+        >> ConnectLayers
+} pass_or_die
+
+
+% Check that we can use gap junction with divergent connect in topology
+{
+  ResetKernel
+  /lr << /rows 2
+         /columns 3
+         /elements /hh_psc_alpha_gap      
+      >> CreateLayer def
+  
+  lr lr << /connection_type (divergent)
+           /synapse_model /gap_junction 
+           /make_symmetric true
+           /weights 0.5
+           /number_of_connections 3
+        >> ConnectLayers
+} pass_or_die
+
+% Check that non-symmetric gap junctions cannot be created in topology
+{
+  ResetKernel
+  /lr << /rows 2
+         /columns 3
+         /elements /hh_psc_alpha_gap      
+      >> CreateLayer def
+  
+  lr lr << /connection_type (convergent)
+           /synapse_model /gap_junction 
+           /make_symmetric false
+           /weights 0.5
+        >> ConnectLayers
+} fail_or_die
+
 endusing
 
--- a/testsuite/unittests/test_gap_junction.sli
+++ b/testsuite/unittests/test_gap_junction.sli
@@ -204,79 +204,93 @@ M_ERROR setverbosity
   and
 } assert_or_die
 
-% Check that we can use gap junction with target-driven connect in topology
+% Check that using gap-junction in topology gives the correct connections.
+{
+  /tgt_driven << /connection_type (convergent)
+                 /synapse_model /gap_junction
+                 /make_symmetric true
+                 /weights << /uniform << /min 0.3 /max 0.6 >> >>
+              >> def
+  /src_driven << /connection_type (divergent)
+                 /synapse_model /gap_junction
+                 /make_symmetric true
+                 /weights << /uniform << /min 0.3 /max 0.6 >> >>
+              >> def
+  /convergent << /connection_type (convergent)
+                 /synapse_model /gap_junction
+                 /make_symmetric true
+                 /weights << /uniform << /min 0.3 /max 0.6 >> >>
+                 /number_of_connections 3
+              >> def
+  /divergent << /connection_type (divergent)
+                /synapse_model /gap_junction
+                /make_symmetric true
+                /weights << /uniform << /min 0.3 /max 0.6 >> >>
+                /number_of_connections 3
+             >> def
+  
+  true
+  [ tgt_driven src_driven convergent divergent  ]
+  {
+    /connspec Set
+    ResetKernel
+    
+    /src_lr << /rows 2
+           /columns 3
+           /elements /hh_psc_alpha_gap
+        >> CreateLayer def
+    /tgt_lr << /rows 2
+           /columns 3
+           /elements /hh_psc_alpha_gap
+        >> CreateLayer def
+    
+    src_lr tgt_lr connspec ConnectLayers
+    
+    % Storing connection info to array.
+    /nodes src_lr GetLocalLeaves def
+    << /source nodes >> GetConnections
+    dup length exch 
+    {
+      dup /source get exch
+      dup /target get exch
+      dup /synapse_model get exch
+      /weight get
+      4 arraystore exch
+    } forall
+    arraystore
+    /source_array Set
+    
+    % With flipped source and target position,
+    % this array should contain the same elements as the source array.
+    /nodes tgt_lr GetLocalLeaves def
+    << /source nodes >> GetConnections
+    dup length exch 
+    {
+      dup /target get exch
+      dup /source get exch
+      /gap_junction exch
+      /weight get
+      4 arraystore exch
+    } forall
+    arraystore
+    /target_array Set
+    
+    source_array length target_array length eq
+    source_array { target_array exch MemberQ and } Fold
+    and
+  } Fold
+} assert_or_die
+
+% Check that non-symmetric gap-junctions cannot be created in topology
 {
   ResetKernel
   /lr << /rows 2
          /columns 3
-         /elements /hh_psc_alpha_gap      
+         /elements /hh_psc_alpha_gap
       >> CreateLayer def
   
   lr lr << /connection_type (convergent)
            /synapse_model /gap_junction
-           /make_symmetric true
-           /weights 0.5
-        >> ConnectLayers
-} pass_or_die
-
-% Check that we can use gap junction with source-driven connect in topology
-{
-  ResetKernel
-  /lr << /rows 2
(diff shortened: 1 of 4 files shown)


testsuite/unittests/test_gap_junction.sli
```diff
@@ -204,5 +204,83 @@ M_ERROR setverbosity
...
+         /elements /hh_psc_alpha_gap      
+      >> CreateLayer def
+  
+  lr lr << /connection_type (convergent)
+           /synapse_model /gap_junction 
+           /make_symmetric false
+           /weights 0.5
+        >> ConnectLayers
+} fail_or_die
+
```
janhahne: The tests are fine, but i am wondering if we should add an additional test, that actually tests if the obtained results are valid. Maybe an mpi test similar to `testsuite/mpitests/test_gap_junctions_mpi.sli`?

topology/connection_creator.h
```diff
@@ -208,10 +216,28 @@ ConnectionCreator::connect_( index s,
...
+  const index& syn,
+  const double& d,
+  const double& w )
+{
+  const index s_thread = kernel().vp_manager.get_thread_id();
+  Node* const s_node = kernel().node_manager.get_node( s, s_thread );
+  kernel().connection_manager.connect(
+    target->get_gid(), s_node, s_thread, syn, d, w );
+}
+
```
janhahne: Is this also working for parallel simulations? I am asking since connection rules as `fixed_total_number` etc. cannot be implemented with symmetric connections in their current version (which is why they are not allowed with gap junctions)

topology/connection_creator_impl.h
```diff
@@ -41,6 +41,14 @@ template < int D >
 void
 ConnectionCreator::connect( Layer< D >& source, Layer< D >& target )
 {
+  if ( kernel().model_manager.connector_requires_symmetric( synapse_model_ )
+    and not( make_symmetric_ ) )
```
heplesser: Drop `()` around `make_symmetric_`.

topology/connection_creator.cpp
```diff
@@ -58,6 +59,11 @@ ConnectionCreator::ConnectionCreator( DictionaryDatum dict )
 
       allow_multapses_ = getValue< bool >( dit->second );
     }
+    else if ( dit->first == names::make_symmetric )
+    {
+
```
heplesser: Remove empty line.

topology/connection_creator.cpp
```diff
@@ -131,8 +137,17 @@ ConnectionCreator::ConnectionCreator( DictionaryDatum dict )
   }
   if ( not delay_.valid() )
   {
-    delay_ =
-      TopologyModule::create_parameter( ( *syn_defaults )[ names::delay ] );
+    if ( synapse_model_
+      == static_cast< index >( kernel().model_manager.get_synapsedict()->lookup(
+           "gap_junction" ) ) )
```
heplesser: Look-up by name is unsafe when users copy models or new gap-type models are added. `syn_defaults[names::has_delay]` is false for connections that have no delay.

testsuite/unittests/test_gap_junction.sli
heplesser: This test is too weak on several accounts: (i) It only tests that the code runs, not that the resulting connections are correct; (ii) in the first two cases it creates all-to-all connections, so one cannot check that forward and backward connections are symmetric; (iii) all connections have the same weight, so one cannot check whether forward and backward connection have identical weight; (iv) the comments for test cases 3 and 4 are the same as for 1 and 2, respectively.

And I agree with @janhahne that we should also have a test ensuring that correct symmetric connections are create when using MPI and threads.

heplesser:
@TomBugnon Do you want to give this PR a try? Adding `'make_symmetric': True` to the connection specification shall now produce symmetric connections.

testsuite/unittests/test_gap_junction.sli
```diff
@@ -204,79 +204,93 @@ M_ERROR setverbosity
...
+    dup length exch 
+    {
+      dup /source get exch
+      dup /target get exch
+      dup /synapse_model get exch
+      /weight get
+      4 arraystore exch
+    } forall
+    arraystore
+    /source_array Set
```
heplesser: You can simplify the code here quite a bit using `Map`, and I think you do not need to select by `/source`:
```
<< >> GetConnections 
{ /conn Set
  [ /source /target /synapse_model /weight ] { conn exch get } Map
} Map
```
and the same thing again with the position of source and target swapped.

testsuite/unittests/test_gap_junction.sli
```diff
@@ -204,79 +204,93 @@ M_ERROR setverbosity
...
+      4 arraystore exch
+    } forall
+    arraystore
+    /target_array Set
+    
+    source_array length target_array length eq
+    source_array { target_array exch MemberQ and } Fold
+    and
+  } Fold
+} assert_or_die
```
heplesser: I would suggest to change the loop order, so that if something fails, we know which of the four variants failed, something like
```
[ tgt_driven src_driven convergent divergent  ]
{
   /connspec Set
  {
     ResetKernel 
     ...
  } assert_or_die
}
forall
```

TomBugnon:
Hi,
Sorry for the late response. 
I have a couple issues and remarks about that PR.

1 - Some tests fail at install for `hakonsbm:topology_gap_junction`. The traceback for `make installcheck` ends with:
```
  Running test testSingleThread...
  Running test test_NoGid...
  Running test test_WithGid...
  File "/Users/Tom/NEST/nest-simulator-topology_gap_junction/share/nest/extras/do_tests.py", line 66, in <module>
    result_topo = runner.run(nest.topology.tests.suite())
  File "/Users/Tom/NEST/nest-simulator-topology_gap_junction/lib/python3.6/site-packages/nest/topology/tests/test_all.py", line 54, in suite
    suite.addTest(test_spatial_kernels.suite())
AttributeError: module 'nest.topology.tests.test_spatial_kernels' has no attribute 'suite'

  PyNEST tests:
     Passed:
     Skipped:
     Failed:
/Users/Tom/NEST/nest-simulator-topology_gap_junction/share/nest/extras/do_tests.sh: line 790: 291 +  : syntax error: operand expected (error token is " ")
make[3]: *** [CMakeFiles/installcheck] Error 1
make[2]: *** [CMakeFiles/installcheck.dir/all] Error 2
make[1]: *** [CMakeFiles/installcheck.dir/rule] Error 2
make: *** [installcheck] Error 2
```
I have the same error with `hakonsbm:603090f146bc2bb16c26f577cc92b7303625dfac`, just before the commits of this PR. So it's not directly related to the changes to topology.


2- I have run a couple tests anyway with `hakonsbm:topology_gap_junction` and the kernel dies during simulation when gap junctions are created with tp.ConnectLayers **if local_num_threads > 1** with the following errors:
```
...

Jun 14 16:55:25 SimulationManager::wfr_update [Warning]:
    Maximum number of iterations reached at interval 23-24 ms

Jun 14 16:55:25 SimulationManager::wfr_update [Warning]:
    Maximum number of iterations reached at interval 24-25 ms

Jun 14 16:55:25 SimulationManager::wfr_update [Warning]:
    Maximum number of iterations reached at interval 25-26 ms
python3.6(50251,0x70000a50f000) malloc: *** error for object 0x7f94aaa10498: incorrect checksum for freed object - object was probably modified after being freed.
```
 This doesn't happen if identical gap junctions are created using nest.Create().
You'll find a jupyter script that should reproduce the issue here: https://drive.google.com/drive/folders/1W5wx3ISrql4NwQ4XLqDkHSGZQkJgrNma?usp=sharing

3- FInally a little remark: due to the `make_symmetric` argument, the `allow_multapses` and `allow_autapses` become a bit misleading since at most two gap junctions rather than one can be created between units.

Apart from these issues the PR works great, thanks a lot for implementing this!

heplesser:
@TomBugnon Thanks for testing! Concerning problem 1, could you look at the beginning of the Python tests section in the installcheck output. Does is say `Using nosetests.` or does it say `Nosetests unavailable. Using ...`? There may be a little error in a test script so that it passes with nosetests but runs into trouble with plain unittest.

Concerning item 3, I think we need to live with the notation we have, as gap junctions are a bit of a corner case.

Concerning number 2, @hakonsbm and @janhahne could you take a look? I guess a first step would be to use the two different ways of connecting (`ConnectLayers` with `make_symmetric` vs `ConnectLayers` followed by `GetConnections` and explicit `Connect`, i.e., the work-around solution) and then use `GetConnections` to check that we really get the same connections. The failing example by @TomBugnon should go into the testsuite.

hakonsbm:
@heplesser @janhahne I've compared connections from the two ways of connecting and they are equal, at least as far as one can tell from using `GetConnections` and `GetStatus`. So it must be a more subtle difference in the connections. I will keep digging and see if I can find it.

TomBugnon:
@heplesser indeed the output reads:
```
Phase 7: Running PyNEST tests.
------------------------------

  Nosetests unavailable. Using fallback test harness.
  Fewer tests will be excuted.
```

I put the full command line output and the `report` subdirectory of the build dir in https://drive.google.com/drive/folders/1iGL6Wn2fM94_zuNJZU6oWtHSiQcrnhKN?usp=sharing
Let me know if I should file an issue.

heplesser:
@TomBugnon It turns out that this approach ("auf meine Mist gewachsen", as we say in German) does not work in parallel situations—I had been rather too fast. The problem is that it is always that target process that needs to create the connection, and one cannot simply switch source and target neuron. The source turns into target and will in most cases be on a different virtual process. So we have to go back to the drawing board. I am therefore closing the PR.

Issue #961, opened 2018-05-30
Assertion error during ConnectLayers execution


keiko-fujii:
When I call topology.ConnectLayers(), I got an assertion error.

> python: /home/MY_PATH/nest-simulator/topology/ntree_impl.h:523:
> nest::Ntree<D, T, max_capacity, max_depth>::iterator nest::Ntree<D, T,
max_capacity, max_depth>::insert(nest::Position<D, double>, const T&) [with
int D = 3, T = long unsigned int, int max_capacity = 100, int max_depth =
10]: Assertion `( pos >= lower_left_ ) && ( pos < lower_left_ + extent_ )'
failed.

I inserted printf() in ntree_impl.h to get 'pos', 'lower_left_', and 'extent_' values.
- pos = [-27.783000, -26.453000, 13.511000]
- lower_left_= [-28.775250, -26.535305, 12.565406]
- extent_ = [0.992250, 1.263586, 1.047117]
(In this case, x-value of (lower_left_+extent_) is equal to pos.)

I generated a simplified code, which replicates the error.
The file 'neuron_pos.dat'  is generated by Matlab and contains 350220 x 3 double data (I don't know how to share it). I tried to replicate the error with smaller XYZ position data, but so far I couldn't.

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

import nest
import nest.topology as tp
import numpy as np

nest.ResetKernel()
nest.ResetNetwork()

# --- Load neuron position data, and get brain size
# --- Matrix size = [N x 3]. 1st_col=x, 2nd_col=y, 3rd_col=z. N = num of XYZ positions
pos_filename = 'neuron_pos.dat'
pos = np.loadtxt(open(pos_filename, 'r'))
size_x = np.max(pos, axis=0)[0] - np.min(pos, axis=0)[0]
size_y = np.max(pos, axis=0)[1] - np.min(pos, axis=0)[1]
size_z = np.max(pos, axis=0)[2] - np.min(pos, axis=0)[2]

# --- Create a layer
num_overlap = 30
layer_property = {
    'positions': pos.tolist(),
    'extent': [size_x*2, size_y*2, size_z*2],  # I guess brain_size*2 is wide enough
    'elements': ['ht_neuron', num_overlap],  # 30 neurons share one XYZ position
    'edge_wrap': False
}
cortex_layer = tp.CreateLayer(layer_property)

# --- Create connections
receptors = nest.GetDefaults('ht_neuron')['receptor_types']
[...]

PR #962, opened 2018-06-04
Improve documentation on setting extent and center (addresses #961)

heplesser:
Added paragraphs to sections on setting extent and center use of values that are exact binary numbers to avoid roundoff.

Addresses #961, which cannot be fixed in a perfect and general way: there is simply no general way in which we can avoid rounding errors here (at least not one that would be relatively straightforward to implement), and NEST needs to terminate if rounding errors would need to inconsistent placement of nodes in the Topology-internal data structures.


heplesser:
@jessica-mitchell Yes, I mean the "2-element tuple of floats". I will make this more consistent. I will also integrate your other suggestions.

PR #963, opened 2018-06-04
Fixed a bug in the SecondaryEvent delivery

### LINKED ISSUE #955: SecondaryEvent data type smaller than unsigned int
It looks like the logic in `write_to_comm_buffer` and `read_from_comm_buffer` (https://github.com/nest/nest-simulator/blob/master/nestkernel/event.h#L885 ff) for secondary events assumes that the events which will be written/read need to be larger than or of equal size as unsigned ints (`c + i * sizeof( unsigned int )`). Otherwise one might run into unallocated memory. @janhahne is that correct and if so, how would you introduce secondary events which are smaller than 4bytes?

janhahne:
This PR fixes #955. The origin of the error is explained in the corresponding issue. 

All SecondaryEvents which exist at the moment use `double` as data type. In this cases the sizes in question are the same (this is actually the case for most data types that I checked), which is why this error remained undetected until now (and cannot be tested with the current master). 

I have created a branch (based on 5g) where I created an additional SecondaryEvent with `bool` data (see https://github.com/janhahne/nest-simulator/tree/5g_bool_sec_event). After merging the fix everything works fine there now.

### DIFF
--- a/nestkernel/event.h
+++ b/nestkernel/event.h
@@ -1057,7 +1057,9 @@ public:
           i != coeffarray_as_d_end_;
           i++ )
     {
-      write_to_comm_buffer( *i, pos );
+      // we need the static_cast here as the size of a stand-alone variable
+      // and a std::vector entry may differ (e.g. for std::vector< bool >)
+      write_to_comm_buffer( static_cast< DataType >( *i ), pos );
     }
     return pos;
   }


nestkernel/event.h
```diff
@@ -1053,11 +1053,13 @@ class DataSecondaryEvent : public SecondaryEvent
...
     write_to_comm_buffer( *( supported_syn_ids_.begin() ), pos );
     write_to_comm_buffer( sender_gid_, pos );
+    DataType elem;
     for ( typename std::vector< DataType >::iterator i = coeffarray_as_d_begin_;
           i != coeffarray_as_d_end_;
           i++ )
     {
-      write_to_comm_buffer( *i, pos );
+      elem = *i;
+      write_to_comm_buffer( elem, pos );
```
heplesser: I would make this even more compact as 
```c++
const DataType elem = *i;
```
And one should probably be able to do it as a one-liner:
```c++
write_to_comm_buffer( static_cast< DataType >(*i), pos );
```
I would prefer that, keeping things as tight as possible.

I would also strongly suggest adding a comment here why this extra effort is necessary.

janhahne:
@heplesser You are right, a `static_cast` is more compact. Please have another look.

PR #964, opened 2018-06-06
Nest developer space: Description for conversion to 5g

steffengraber:
5g introduces a few changes to model code. In order to help users and developers to update their custom models to NEST 2.16 this text describes the necessary changes to neuron models, synapse models and devices.
See https://github.com/jakobj/nest-simulator/issues/72

### DIFF
--- a/index.md
+++ b/index.md
@@ -72,16 +72,16 @@ can either be static or implement synaptic plasticity rules such as
 * [Synapses in NEST: An overview](synapses_overview)
 * [Developing synapse models](synapse_models)
 
-## Updating models from 2.14 or prior to 2.16
+## Updating Models for the 5g Kernel
 
 With the introduction of the new connection infrastructure of the [5g kernel](https://www.frontiersin.org/articles/10.3389/fninf.2018.00002/full), 
-neuron, synapse and device models need to be slightly adapted from prior 
+rate neuron, synapse and device models need to be slightly adapted from prior 
 versions to be compatible with the latest release (2.16). In the following we describe 
 all necessary changes:
 
 * [Updating models from 2.14 or prior to 2.16](model_conversion_5g.md)
 
-## Updating models for NEST 2.4 or prior to 2.6 or later
+## Updating Models for the 4g Kernel
 
 If you find models written for NEST version 2.4 and prior not
 working anymore in newer versions, it is most likely due to recent
--- a/index.md
+++ b/index.md
@@ -72,23 +72,23 @@ can either be static or implement synaptic plasticity rules such as
 * [Synapses in NEST: An overview](synapses_overview)
 * [Developing synapse models](synapse_models)
 
-## Updating Models for the 5g Kernel
+## Updating models to 2.16
 
 With the introduction of the new connection infrastructure of the [5g kernel](https://www.frontiersin.org/articles/10.3389/fninf.2018.00002/full), 
 rate neuron, synapse and device models need to be slightly adapted from prior 
 versions to be compatible with the latest release (2.16). In the following we describe 
 all necessary changes:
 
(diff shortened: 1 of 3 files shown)


index.md
```diff
@@ -72,6 +72,17 @@ can either be static or implement synaptic plasticity rules such as
...
 * [Developing synapse models](synapse_models)
 
+## Updating models from 2.14 or prior to 2.16
+
+With the introduction of the new connection infrastructure of the [5g kernel](https://www.frontiersin.org/articles/10.3389/fninf.2018.00002/full), 
+neuron, synapse and device models need to be slightly adapted from prior 
+versions to be compatible with the latest release (2.16). In the following we describe 
+all necessary changes:
+
+* [Updating models from 2.14 or prior to 2.16](model_conversion_5g.md)
```
heplesser: I like that the details are in a separate file, but I am not sure about the format of the link here: (i) Does a single-item bullet list make sense? (ii) Could the reader be confused because the link text is the same as the section heading? How about replacing the text from "In the following ..." with something like `[Please see these instructions for adapting models to NEST 2.16](model_conversion_5g.md)`?
@jessica-mitchell What do you think?

index.md
```diff
@@ -72,6 +72,17 @@ can either be static or implement synaptic plasticity rules such as
 * [Synapses in NEST: An overview](synapses_overview)
 * [Developing synapse models](synapse_models)
 
+## Updating models from 2.14 or prior to 2.16
+
+With the introduction of the new connection infrastructure of the [5g kernel](https://www.frontiersin.org/articles/10.3389/fninf.2018.00002/full), 
+neuron, synapse and device models need to be slightly adapted from prior
```
heplesser: @jakobj @suku248  Should we make this "rate neuron, synapse, and device models"?

index.md
jessica-mitchell: I think the format matches quite nicely with the rest of the page (i.e. links are bullet lists), which is good for consistency and making the link stand out. However, I agree that the link text and title might be better if different from each other. What do you think of something like removing the headings "Updating models from  ..."  and putting both texts and links under a heading "Updating Models for the 5g Kernel"  or "Updating Models for Newest Release"?
Note: I am getting a 404 on the page for "updating models from 2.4 or prior to 2.6".

steffengraber: I followed @jessica-mitchell  suggestion and changed it accordingly. Jessica, as you may have noticed, has all internal links point to 404, you would have to explicitly point to the file with the extension '.md', which is a problem for the HTML parser,

index.md
steffengraber: Done.

index.md
```diff
@@ -72,16 +72,16 @@ can either be static or implement synaptic plasticity rules such as
 * [Synapses in NEST: An overview](synapses_overview)
 * [Developing synapse models](synapse_models)
 
-## Updating models from 2.14 or prior to 2.16
+## Updating Models for the 5g Kernel
```
heplesser: @steffengraber "5g kernel" is a term we are using only internally, so it may confuse users, so we should use 2.14 and 2.16 here, and similarly below for "4g". To avoid having the same line twice, once as heading and once as bullet point, I'd suggest to change the text in the bullet point to "How to update models from 2.14 or prior to 2.16" and corresponding for 4g. Sorry I didn't make an explicit suggestion earlier.

steffengraber: Ok, now it is:

    ## Updating models to 2.16
    * [How to update models from NEST 2.14 or prior to 2.16](model_conversion_5g)

    ## Updating models to 2.6 or later
    * [How to update models from NEST 2.4 or prior to 2.6 or later](model_conversion_3g_4g)

heplesser:
@suku248 Could you double-check that the content of these instructions is okay?

PR #966, opened 2018-06-06
Fix #825: Improve sign check

### LINKED ISSUE #825: vogels_sprekeler_synapse / sign check
The Vogels-Sprekeler synapse implements a constant weight decay for each pre-synaptic spike. 
For sufficient pre-synaptic spikes, this can result in the weight becoming 0, which is expected and desired behaviour . However, if the initial weight was negative, nest throws an error if subsequently the synapse properties are changed (e.g. when the learning rate is changed). 

MWE: (kinda)

```python 
import nest

a, b = [[gid] for gid in nest.Create('iaf_psc_alpha', 2)]
nest.Connect(a, b,
             conn_spec={'rule':'one_to_one'},
             syn_spec={'model'  :'vogels_sprekeler_synapse',
                       'weight' :-0.,
                       'Wmax'   :-1.})
```

I would appreciate any pointers on how to work around this issue.

sanjayankur31:
Fixes #825

### DIFF
--- a/models/vogels_sprekeler_connection.h
+++ b/models/vogels_sprekeler_connection.h
@@ -298,8 +298,7 @@ VogelsSprekelerConnection< targetidentifierT >::set_status(
   updateValue< double >( d, names::Kplus, Kplus_ );
 
   // check if weight_ and Wmax_ has the same sign
-  if ( not( ( ( weight_ >= 0 ) - ( weight_ < 0 ) )
-         == ( ( Wmax_ >= 0 ) - ( Wmax_ < 0 ) ) ) )
+  if ( weight_ != 0 && not( ( weight_ > 0 ) == ( Wmax_ > 0 ) ) )
   {
     throw BadProperty( "Weight and Wmax must have same sign." );
   }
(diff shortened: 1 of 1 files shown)


sanjayankur31:
Eh - I seem to have broken a whole lot of tests with this. Checking up.

models/vogels_sprekeler_connection.h
```diff
@@ -298,8 +298,7 @@ VogelsSprekelerConnection< targetidentifierT >::set_status(
   updateValue< double >( d, names::Kplus, Kplus_ );
 
   // check if weight_ and Wmax_ has the same sign
-  if ( not( ( ( weight_ >= 0 ) - ( weight_ < 0 ) )
-         == ( ( Wmax_ >= 0 ) - ( Wmax_ < 0 ) ) ) )
+  if ( weight_ != 0 && not( ( weight_ > 0 ) == ( Wmax_ > 0 ) ) )
```
heplesser: @sanjayankur31 The comment above the check is on entirely in sync with the code. Also, the code needs quite a bit of thought to understand it. Since we now allow C++11 constructs in NEST, I would suggest
```c++
weight != 0 and ( std::signbit( weight_ ) != std::signbit( Wmax_ ) )
```

PR #970, opened 2018-06-11
Fixes #967: Correct parameter in documentation

### LINKED ISSUE #967: Documentation error for spike_generator.
From spike_generator.h : line 77: 

```
  /allow_offgrid_times   default: false
     If false, spike times will be rounded to the nearest step if they are
     less than tic/2 from the step, otherwise NEST reports an error.
     If true, spike times are rounded to the nearest step if within tic/2
     from the step, otherwise they are rounded up to the *end* of the step.
```

But when wants to set the allow_offgrid_times parameter it is not found. The correct name for it is
allow_offgrid_spikes as evident in spike_generator.cpp line 101: 

```
 ( *d )[ names::allow_offgrid_spikes ] = BoolDatum( allow_offgrid_spikes_ );
```

sanjayankur31:
Fixes #967

[skip ci]

### DIFF
--- a/models/spike_generator.h
+++ b/models/spike_generator.h
@@ -74,7 +74,7 @@ namespace nest
      receiving the spike train can handle precise timing information. In this
      case, the other two options are ignored.
 
-  /allow_offgrid_times   default: false
+  /allow_offgrid_spikes   default: false
      If false, spike times will be rounded to the nearest step if they are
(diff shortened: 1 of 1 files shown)


heplesser:
Bring parameter documentation into accordance with behavior. Merging without further review.

PR #980, opened 2018-06-27
Added parameter to get/set function in tanh_rate model

janhahne:
The tanh_rate model contains a parameter `theta` that cannot be set at the moment. This PR fixes this small bug.


heplesser:
Since this is a minor ans straightforward correction, I merge right away.

Issue #990, opened 2018-08-01
Ensure NEST compiles with libc++ from LLVM in C++11 mode


heplesser:
Since we have permitted the use of C++11 features in NEST, and as some compilers require `-std=c++11` switches to compile code with C++11 features (only gcc adds them "automagically" unless prohibited), some constructs no longer work. This is especially true for toolchains that do not use libstdc++ from GNU, but libc++ from LLVM.

A useful test setup currently is using Clang 6.0 with the `-std=c++11` switch. This results currently in at least two issues: `LONG_LONG_{MAX,MIN}` is replaced by `LLONG_{MAX,MIN}` and multisynapse models fail to compile with the following error message:
```
In file included from /Users/plesser/NEST/code/src/models/aeif_cond_alpha_multisynapse.cpp:23:
In file included from /Users/plesser/NEST/code/src/models/aeif_cond_alpha_multisynapse.h:38:
In file included from /Users/plesser/NEST/code/src/nestkernel/archiving_node.h:41:
In file included from /Users/plesser/NEST/code/src/nestkernel/node.h:35:
In file included from /Users/plesser/NEST/code/src/nestkernel/event.h:36:
In file included from /Users/plesser/NEST/code/src/nestkernel/exceptions.h:30:
In file included from /Users/plesser/NEST/code/src/sli/name.h:30:
/usr/local/Cellar/llvm/6.0.1/include/c++/v1/map:629:15: error: object of type 'std::__1::pair<Name, const nest::DataAccessFunctor<nest::aeif_cond_alpha_multisynapse> >'
      cannot be assigned because its copy assignment operator is implicitly deleted
        {__nc = __v.__cc; return *this;}
              ^
/usr/local/Cellar/llvm/6.0.1/include/c++/v1/__tree:1645:35: note: in instantiation of member function 'std::__1::__value_type<Name, const
      nest::DataAccessFunctor<nest::aeif_cond_alpha_multisynapse> >::operator=' requested here
                __cache->__value_ = *__first;
                                  ^
/usr/local/Cellar/llvm/6.0.1/include/c++/v1/__tree:1575:9: note: in instantiation of function template specialization 'std::__1::__tree<std::__1::__value_type<Name, const
      nest::DataAccessFunctor<nest::aeif_cond_alpha_multisynapse> >, std::__1::__map_value_compare<Name, std::__1::__value_type<Name, const
[...]

PR #991, opened 2018-08-01
Make test for "long long max" C++11-safe on all systems

heplesser:
This is a partial fix for #990, making the selection between `LLONG_{MIN,MAX}` and `LONG_LONG_{MIN,MAX}` robust across compilers and C++ standards.


heplesser:
Closing, will be replaced right away with new PR fixing all issues in #990.

PR #993, opened 2018-08-02
Fix 990 / Compilation issues with C++11 and libc++

heplesser:
Note that the issues fixed here **do not occur with gcc and libstdc++**, they only occur with LLVM/Clang compilers using libc++ in C++11 mode and with some other compilers that do not use libstdc++.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,7 @@ include( CPackVariables )
 include( CheckExtraCompilerFeatures )
 include( ConfigureSummary )
 include( GetTriple )
+include( DefaultCompilerFlags )
 
 # get triples arch-vendor-os
 get_host_triple( NEST_HOST_TRIPLE NEST_HOST_ARCH NEST_HOST_VENDOR NEST_HOST_OS )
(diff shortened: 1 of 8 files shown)


heplesser:
@ikitayama @carlosengutierrez This PR should solve the `operator=()` compilation issues with C++11/libc++. Could you check?

ikitayama:
@carlosengutierrez are you finished building the 5g branch with FCC natively? Here it's been running for 2 hours, but still going.

carlosengutierrez:
Hi @ikitayama , 
   I am trying to build the branch -->>  heplesser:fix-990-c++11-issues    (that is the branch to test, right?)
   I have to install it by interactive jobs, last night I could reach 97% but connection was broken, now trying again.
Thanks, C.

ikitayama:
You could use screen so that your build job don't get terminated.

heplesser:
@ikitayama Excellent, thank you!

heplesser:
I have added another commit, which automatically includes the `-std=c++11` compiler flag to `CMAKE_CXX_FLAGS` before all other flags are processed. This flag appears to be standard for all compilers I checked (gcc, clang, intel, xlc, pgi, fujitsu). While gcc will compile NEST with c++11 features even without this flag, other compilers need it to accept C++11 features.

carlosengutierrez:
@heplesser I could compile the updated nest master branch. PYNEST is running.

PR #997, opened 2018-08-07
Fix version display during startup

jougs:
This PR addresses two issues:
* remove the confusing printout of the git branch and hash during the run of `cmake`
* make `nest` and `sli` print the correct version information or git branch and hash during startup instead of always the last released version number.

As these issues were initially discussed by @steffengraber and @heplesser on the mailing list, I suggest them to be reviewers for the PR.

### DIFF
--- a/extras/create_rcsinfo.sh
+++ b/extras/create_rcsinfo.sh
@@ -28,6 +28,5 @@ mkdir -p $sli_libdir
 
 # create rcsinfo.sli
 echo "statusdict /rcsinfo ($version) put" > $sli_libdir/rcsinfo.sli
-echo $version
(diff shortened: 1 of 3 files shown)


jougs:
For a release it will show the tag, i.e. something like `v2.14.0`.

steffengraber:
There is the 'pynest/nest/pynest-init.sli'. Does this file needs to be customized for the message in Pynest?

jougs:
@steffengraber: good catch. The `pywelcome` implementation completely slipped my attention. I have adapted it in 898ab1b to match the implementation of the normal `welcome` function.

PR #999, opened 2018-08-08
Adjust send_recv_count_target_data_

suku248:
`send_recv_count_target_data_` should be adjusted in `MPIManager::increase_buffer_size_target_data()`.

Calling `set_buffer_size_target_data()` ensures this.

@jakobj and @wschenck could you please review this?

### DIFF
--- a/nestkernel/mpi_manager.h
+++ b/nestkernel/mpi_manager.h
@@ -597,12 +597,14 @@ MPIManager::increase_buffer_size_target_data()
     if ( buffer_size_target_data_ * growth_factor_buffer_target_data_
       < max_buffer_size_target_data_ )
     {
-      buffer_size_target_data_ = static_cast< size_t >(
-        floor( buffer_size_target_data_ * growth_factor_buffer_target_data_ ) );
+      // this adjusts also send_recv_count_target_data_per_rank_
+      set_buffer_size_target_data( static_cast< size_t >( floor(
+        buffer_size_target_data_ * growth_factor_buffer_target_data_ ) ) );
     }
     else
     {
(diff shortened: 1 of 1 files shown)


heplesser:
@hakonsbm Could you also check if this change has an effect on benchmarks with more than one MPI process?

nestkernel/mpi_manager.h
```diff
@@ -597,12 +597,14 @@ MPIManager::increase_buffer_size_target_data()
     if ( buffer_size_target_data_ * growth_factor_buffer_target_data_
       < max_buffer_size_target_data_ )
     {
-      buffer_size_target_data_ = static_cast< size_t >(
-        floor( buffer_size_target_data_ * growth_factor_buffer_target_data_ ) );
+      // this adjusts also send_recv_count_target_data_per_rank_
```
hakonsbm: Should be "also adjusts".

nestkernel/mpi_manager.h
```diff
@@ -597,12 +597,14 @@ MPIManager::increase_buffer_size_target_data()
...
+      set_buffer_size_target_data( static_cast< size_t >( floor(
+        buffer_size_target_data_ * growth_factor_buffer_target_data_ ) ) );
     }
     else
     {
-      buffer_size_target_data_ = max_buffer_size_target_data_;
+      // this adjusts also send_recv_count_target_data_per_rank_
```
hakonsbm: Same as above.

nestkernel/mpi_manager.h
suku248: Agreed!

PR #1004, opened 2018-08-14
Fix compatibility issues with Python 2

hakonsbm:
This PR contains some fixes for compatibility issues when using Python 2 when running examples and model_details scripts.

Python 2 doesn't handle complex numbers as elegantly as Python 3, which led to an error when running `HillTononiModels.ipynb` using Python 2. The problematic variable is now explicitly converted to a complex number if needed. There were also some minor issues in the PyNEST examples.


doc/model_details/HillTononiModels.ipynb
```diff
@@ -1261,6 +1261,11 @@
     "        self.hp = ht_params\n",
     "        \n",
     "    def m_inf(self, D):\n",
+    "        # Workaround for handling complex numbers in Python 2\n",
```
heplesser: Hm, I am not really happy with this work-around, it is a lot of distracting code for a notebook. And can `D` really become negative? From the math, we should always have `D>0` strictly.

pynest/examples/brunel_alpha_evolution_strategies.py
```diff
@@ -64,6 +64,7 @@
 
 '''
 
+from __future__ import print_function
```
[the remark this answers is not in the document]
hakonsbm: Because of the `end=' '` here:
https://github.com/nest/nest-simulator/blob/4b9227814595259da12653dd65745d1f8514b16a/pynest/examples/brunel_alpha_evolution_strategies.py#L558

doc/model_details/HillTononiModels.ipynb
hakonsbm: I agree that it's not an elegant work-around. `m_inf` is called from `voltage_clamp` like this
```python
m_old = channel.m_inf(DT_V_seq[0][1])
``` 
where `channel` is `iDK` and `DT_V_seq` is the list in
```python
nr, cr = voltage_clamp(iDK, [(500, -65.), (500, -35.), (500, -25.), (500, 0.), (5000, -70.)],
                      nest_dt=1.)
```

pynest/examples/brunel_alpha_evolution_strategies.py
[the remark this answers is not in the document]
heplesser: OK.

doc/model_details/HillTononiModels.ipynb
heplesser: So I guess the problem is that in
```python
    def compute_I(self, t, V, m0, h0, D0):
        self.D = si.odeint(self.dD, D0, t, args=(V,))
        self.m = self.m_inf(self.D)
        return - self.hp['g_peak_KNa'] * self.m * (V - self.hp['E_rev_KNa'])
```
`self.D` may contain negative values (numerical error from `odeint`). But we know that `D` must always be strictly positive, so we should (i) check that we only get numerical noise below zero and then (ii) replace negative values in the result of `odeint()` with the smallest possible positive number.

hakonsbm: I think you misunderstood my previous comment. `voltage_clamp()` is called with a channel, and a list of interval-voltage pairs referred to as `DT_V_seq`. In `voltage_clamp` we have
```python
# Control part
t_old = 0.
try:
    m_old = channel.m_inf(DT_V_seq[0][1])
except NotImplementedError:
    m_old = None
```
where `DT_V_seq[0][1]` is the voltage of the first interval-voltage pair, i.e. `-65.0`.

This is before it calls `compute_I()`. And I checked the values for `self.D`, they are all positive.

PR #1007, opened 2018-08-16
Clean up modification of dynamic recordables map in models

### LINKED ISSUE #996: Clean up model modification of dynamic recordables map
Some models update the dynamic recordables map at the end of their `set_status()` by creating a copy of `recordablesMap_`, make changes to the copy, then assign the copy back to `recordablesMap_`. See for example https://github.com/nest/nest-simulator/blob/b69389b918b1a8a6e0f7651231997722de2cb46e/models/aeif_cond_alpha_multisynapse.cpp#L696

The reason for this is presumably to prevent changes to `recordablesMap_` in case something goes wrong. However, if something goes wrong here, it will be too serious to recover from. Therefore there is no reason to create a copy here, and this should be cleaned up by making the changes directly to  `recordablesMap_`.

Models affected are those using `DynamicRecordablesMap`, i.e.
- `aeif_cond_alpha_multisynapse`
- `aeif_cond_beta_multisynapse`
- `iaf_psc_alpha_multisynapse`
- `iaf_psc_exp_multisynapse`

Not creating a temporary copy of `recordablesMap_` should also allow `DataAccessFunctor` in `DynamicRecordablesMap` to be `const` again (which was changed in #990 after Clang complained when compiling with the `-std=c++11` flag).

hakonsbm:
Do changes directly to `recordablesMap_`, not to a temporary variable. Also reinstate constant  in `DynamicRecordablesMap`. Fixes #996.

### DIFF
--- a/models/aeif_cond_alpha_multisynapse.cpp
+++ b/models/aeif_cond_alpha_multisynapse.cpp
@@ -693,8 +693,6 @@ aeif_cond_alpha_multisynapse::set_status( const DictionaryDatum& d )
    * Here is where we must update the recordablesMap_ if new receptors
    * are added!
    */
-  DynamicRecordablesMap< aeif_cond_alpha_multisynapse > rtmp =
-    recordablesMap_;                         // temporary copy in case of errors
   if ( ptmp.E_rev.size() > P_.E_rev.size() ) // Number of receptors increased
   {
     for ( size_t receptor = P_.E_rev.size(); receptor < ptmp.E_rev.size();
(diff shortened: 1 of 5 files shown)


hakonsbm:
@clinssen Yes, if any properties in the dictionary `d` are wrong, it throws `BadProperty` without actually changing anything. `Parameters` and `State` aren't changed before we know that all properties are correct. However, it is not possible to recover from errors arising from changing the `DynamicRecordablesMap`.

heplesser:
To clarify: inserting into the recordables map can only fail if we run out of memory, and then preserving an old state makes no sense.  Setting parameters or state, in contrast, will quite frequently raise exceptions when a user provides an illegal parameter value.

Issue #1006, opened 2018-08-15
libreadline is assume to exist and ready to be linked when building NEST with -Dwith-readline=OFF


ikitayama:
@heplesser assumes there's something wrong with the current CMake files under the cmake directory.

heplesser:
More specifically, as described in #994, nest seems to be linked against libreadline even if `-Dwith-readline=OFF` is given, which leads to problems if no (statically linked) libreadline is available.

gtrensch:
@ikitayama I did not succeed in reproducing the problem.

Perhaps it's a version-dependent CMake issue. However, please can you check the NEST Configuration Summary and provide more information?

For comparison: If `-Dwith-readline=OFF` is set  ...
```
--------------------------------------------------------------------------------
NEST Configuration Summary
--------------------------------------------------------------------------------
Build type          : 
Target System       : Linux
Cross Compiling     : FALSE
C compiler          : /usr/bin/cc
C compiler flags    :  -O2 -Wall -fopenmp -fdiagnostics-color=auto
C++ compiler        : /usr/bin/c++
C++ compiler flags  : --static -std=c++11 -O2 -Wall -fopenmp -fdiagnostics-color=auto
Build dynamic       : OFF

Built-in modules    : models;precise;topology
User modules        : None
Python bindings     : No
Use threading       : Yes (OpenMP: -fopenmp)
Use GSL             : Yes (GSL 2.1)
    Includes        : /usr/include
    Libraries       : /usr/lib/x86_64-linux-gnu/libgsl.a;/usr/lib/x86_64-linux-gnu/libgslcblas.a

Use Readline        : No  <--- !!!
Use libltdl         : No
Use doxygen         : Yes (/usr/bin/doxygen)
                    : target `doc` available
    `dot` available : Yes (/usr/bin/dot)
                    : target `fulldoc` available
Use MPI             : No
Use MUSIC           : No
Use libneurosim     : No
Use Boost           : No
```
... `Use Readline  : No` should be reported and `-lreadline` should not show up at the linker options.

SLI kernel:
```
/usr/bin/c++ --static -std=c++11 -O2 -Wall -fopenmp -fdiagnostics-color=auto   CMakeFiles/sli.dir/puresli.cc.o  -o sli -rdynamic libsli.a libsli_readline.a -Wl,-Bstatic -lgsl -lgslcblas libsli.a ../libnestutil/libnestutil.a -lgsl -lgslcblas -Wl,-Bdynamic`
```
[...]

PR #1011, opened 2018-08-20
do not search for libreadline when with-readline=OFF requested by user

clinssen:
This PR addresses issue #1006.

I checked with `ldd` that there is no more dependency on libreadline after building with `with-readline=OFF`, for all of the generated .so files. (Note that this issue was irrespective of static vs. dynamic linking.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,11 @@ nest_process_tics_per_ms()
 nest_process_tics_per_step()
 nest_process_with_ps_array()
 nest_process_with_libltdl()
-nest_process_with_readline()
+
+if ( with-readline )
+  nest_process_with_readline()
+endif ()
+
 nest_process_with_gsl()
 nest_process_with_python()
 nest_process_with_openmp()


heplesser:
@ikitayama Could you verify that this PR fixes the problem you reported in #1006?

CMakeLists.txt
```diff
@@ -144,7 +144,11 @@ nest_process_tics_per_ms()
 nest_process_tics_per_step()
 nest_process_with_ps_array()
 nest_process_with_libltdl()
-nest_process_with_readline()
+
+if ( with-readline )
```
gtrensch: This is better done in nest_process_with_readline() ... and it is already there.

gtrensch:
I performed some tests and think that the issue #1006, which should be addressed in this PR, needs further investigation. See my comprehensive comment in #1006.

ikitayama:
@heplesser -- I'll do check when @gtrensch comes up with a patch for us after investigating further on the issue #1006.

gtrensch:
@ikitayama you can have look at the comments I made on #1006. I have requested some additional information.

heplesser:
Closing this, see #1006 for further discussion of ways to address this matter.

PR #1012, opened 2018-08-21
Add skip_if_without_music to nest dictionary

stinebuu:
In PR #1005, `skip_if_without_music` was added to some music examples. There was previously a `statusdict/have_music :: not {statusdict/exitcodes/success :: quit_i} if` in these examples, and calling `skip_if_without_music` is cleaner. `skip_if_without_music` is only defined in the SLI `unittest` namespace however, so when trying to run the music examples we now get

```
SLIInterpreter::execute [Error]: DictError
    Key '/skip_if_without_music' does not exist in dictionary.
```

even if we have compiled NEST with music. In this PR I therefore add `skip_if_without_music` to the SLI NEST namespace to get the examples to run. It is still also defined in the `unittest` namespace.

### DIFF
--- a/lib/sli/nest-init.sli
+++ b/lib/sli/nest-init.sli
@@ -1453,12 +1453,6 @@ Description:
 } bind def
 
 
-/skip_if_without_music
-{
-  statusdict/have_music :: not {statusdict/exitcodes/success :: quit_i} if
-} def
-
-
 % Install modules in environment variable NEST_MODULES. Modules have
 % to be separated by colon.
 (NEST_MODULES) getenv { (:) breakup { Install } forall } if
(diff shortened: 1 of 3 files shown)


lib/sli/nest-init.sli
```diff
@@ -1453,6 +1453,12 @@ Description:
 } bind def
 
 
+/skip_if_without_music
+{
+  statusdict/have_music :: not {statusdict/exitcodes/success :: quit_i} if
```
heplesser: I would suggest spaces around the curly braces.

stinebuu:
@heplesser I have now moved all the `/skip_if_*` definitions to `sli-init.sli`!

PR #1017, opened 2018-08-28
fix for interpretation of `verbosity` and `debug` command line arguments as input script filename

### LINKED ISSUE #1016: --verbosity option does not work
In current NEST master (1aa9c69eb) and thus most likely also in 2.16.0, the `--verbosity` switch fails as follows:
```
Hanss-MacBook-Pro:~ plesser$ nest --verbosity=ERROR
[INFO] [2018.8.28 8:31:1 /Users/plesser/NEST/code/src/nestkernel/rng_manager.cpp:238 @ Network::create_rngs_] : Creating default RNGs
[INFO] [2018.8.28 8:31:1 /Users/plesser/NEST/code/src/nestkernel/rng_manager.cpp:284 @ Network::create_grng_] : Creating new default global RNG
NEST 1aa9c69eb (C) 2004 The NEST Initiative

Aug 28 08:31:01 file [Error]: FileOpenError
    Could not open the following file for reading: "--verbosity=ERROR.sli".
```

Interestingly, the verbosity value is read an respected during startup, but then the option is *also* interpreted as a script to execute.

clinssen:
This fixes #1016

### DIFF
--- a/sli/slistartup.cc
+++ b/sli/slistartup.cc
@@ -282,7 +282,6 @@ SLIStartup::SLIStartup( int argc, char** argv )
   for ( int i = 0; i < argc; ++i )
   {
     StringDatum* sd = new StringDatum( argv[ i ] );
-    ad.push_back( Token( sd ) );
 
     if ( *sd == "-d" || *sd == "--debug" )
     {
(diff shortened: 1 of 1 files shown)


jougs:
After giving this another thought (or two), I now think the cleanest solution is to add a new `ArrayDatum` to the `SLIStartup` class, which only contains the non-option arguments (i.e. those not starting with `--`) and only iterate that during SLI's `start` procedure in `sli-init.sli`. This way the `targs` array can contain the full list of command line arguments, while not having to parse strings in `start`.

In agreement with @clinssen, I'll implement this together with regression tests for #1016 and #776 and start a new PR once I'm ready. I'm thus closing this PR for now.

PR #1018, opened 2018-08-28
Make test_refractory.py run with make installcheck

stinebuu:
`test_refractory.py` is currently not being run when doing `make installcheck` when `nosetests` is available. The reason seems to be that the test use a decorator to test all neurons available in NEST, and `make installcheck` does not like it. I don't know why. So instead of using the decorator I loop through all models in NEST with a for-loop.

Since this test has not been run in a while there were several models that made the test fail. This was mostly because the models in question do not use `t_ref`, and so the test would fail when trying to set this variable. I have put all these models in the `ignore_model` list. In addition to this, `iaf_psc_exp_ps_lossless` caused the test to fail because it use precise times for spikes, and so the test would not detect that spiking time corresponded to the same times as when a voltmeter would detect a spike ([found here in the test](https://github.com/nest/nest-simulator/blob/master/pynest/nest/tests/test_refractory.py#L204)). I tried to round the `spike_times[1]` variable, but I got into trouble with floating point accuracy, so the test would still fail. I therefore put `iaf_psc_exp_ps_lossless` into `ignore_list` as well.

Furthermore, the `siegert_neuron` model caused the test to fail because it could not be connected to a `voltmeter`. I don't know if this is as it should be?

The `gif_pop_psc_exp` model caused the test to fail because `V_reset` (which is 0.0) is not close to the potential recorded from the voltmeter. We get a python `IndexError` [here](https://github.com/nest/nest-simulator/blob/master/pynest/nest/tests/test_refractory.py#L211), and I do not know enough about the model to know if the test is supposed to pass.

### DIFF
--- a/pynest/nest/tests/test_refractory.py
+++ b/pynest/nest/tests/test_refractory.py
@@ -43,21 +43,39 @@ trigger immediate spiking
 Untested models
 ---------------
 * ``aeif_cond_alpha_RK5``
+* ``erfc_neuron``
+* ``gauss_rate_ipn``
+* ``gif_pop_psc_exp``
 * ``ginzburg_neuron``
 * ``hh_cond_exp_traub``
 * ``hh_psc_alpha``
 * ``hh_psc_alpha_gap``
-* ``ht_neuron``
 * ``iaf_chs_2007``
 * ``iaf_chxk_2008``
+* ``iaf_psc_exp_ps_lossless``
 * ``iaf_tum_2000``
 * ``izhikevich``
+* ``lin_rate_ipn``
+* ``lin_rate_opn``
 * ``mcculloch_pitts_neuron``
 * ``parrot_neuron``
 * ``parrot_neuron_ps``
 * ``pp_pop_psc_delta``
 * ``pp_psc_delta``
+* ``rate_transformer_gauss``
+* ``rate_transformer_lin``
+* ``rate_transformer_sigmoid``
+* ``rate_transformer_sigmoid_gg_1998``
+* ``rate_transformer_tanh``
+* ``rate_transformer_threshold_lin``
 * ``sli_neuron``
+* ``siegert_neuron``
+* ``sigmoid_rate_gg_1998_ipn``
+* ``sigmoid_rate_ipn``
+* ``tanh_rate_ipn``
+* ``tanh_rate_opn``
+* ``threshold_lin_rate_ipn``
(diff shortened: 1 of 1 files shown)


stinebuu:
There is something a bit strange happening with Travis because of `mat2_psc_exp` and `amat2_psc_exp`. It fails on and off again. The commit called *mat2_psc_exp* passed here for instance, but the exact same branch failed on Travis on my repository. I have now put them both in the `ignore_model` list.

heplesser:
@stinebuu The reason for the inconsistent behavior may be the randomization of the refractory time
```
t_ref = resolution * np.random.randint(min_steps, max_steps)
```
which will mean that different runs use different refractory times. There are several things we need to fix here:

1. Use a fixed `t_ref` in this test, but maybe a slightly unusual one, e.g., 1.7 ms.
1. Find out for which `t_ref` the `(a)mat` models fail the test (maybe just write a loop over all possible values generated by randint, i.e., 0.1 to 20 ms), then understand why the test sometimes fails.
1. The list of excluded models is rather long. It really should only contain models which do not have an absolute refractory time implemented, and it should be possible to identify those automatically because they have no `t_ref` parameter.

clinssen:
Just to check on the status: @stinebuu fixed the script so that it tests _some_ models, where before it tested _no_ models. Rather than wait for full coverage, I propose to OK the PR as-is.

The Travis CI build failure seems a separate issue: did you ever find out why the build failed for certain values of `t_ref`? Has this issue been fixed?

stinebuu:
@clinssen Yes, the test did not run with make installcheck before, so no models were tested. Now it runs, and tests most models. The models it does not test is outside the scope of the test-script. So it does not check models that do not use `t_ref` for instance, as checking `t_ref` is the object of the test. The models in the `ignore_model` list also makes the test fail, because of how they are defined.

The Travis problem happened because the test used a random variable for `t_ref` and because `mat2_psc_exp` and `amat2_psc_exp` both need a very high current to spike. The current they were given before was too low to be able to trigger a spike for some `t_ref`s. What happened was that some of the Travis runs generated `t_ref`s that hit the too small values. I fixed it now by giving the two models a higher input current. I have checked that all working models work for all `t_ref`s between 0.1 and 20 ms, but as this makes the test slow, I have removed it from the PR and used a fixed value for `t_ref`, namely 1.7 ms.

PR #1019, opened 2018-08-28
Make testsuite run when nosetests is unavailable

stinebuu:
When the testsuite is run, but `nosetests` is **not** available, do_test.sh crash before finishing all the tests and give `AttributeError: module 'nest.topology.tests.test_spatial_kernels' has no attribute 'suite'`. This comes from the fact that when we do not have `nosetests` available, we go through the suites defined in the `test_all.py` files, and run the corresponding tests. Topology's `test_all.py` call the `suite()` function of `test_spatial_kernels.py`, but `test_spatial_kernels.py` do not have a `suite()` function defined. So in this PR I have added the function to make the testsuite run to completion.

I have also added a bunch of missing test-files to the different `test_all`, to make more tests run when `nosetests` is unavailable, and added `suite()` functions to the files that were missing them.

NB!! Even though this PR pass on Travis, the testsuite will report one failure if you run `make installcheck` on your local computer and you do not have `nosetests`. The failing test is `test_refactory.py` which, as it turns out, has not been run in a while. I have made a separate PR for this issue, #1018, as that failure has nothing to do with `nosetests`. Once this PR and #1018 is merged though, all test will run and pass with or without `nosetests`.

### DIFF
--- a/pynest/nest/tests/test_all.py
+++ b/pynest/nest/tests/test_all.py
@@ -24,74 +24,108 @@ import nest
 
 from . import compatibility
 
-from . import test_errors
-from . import test_stack
-from . import test_create
-from . import test_status
-from . import test_onetooneconnect
+from . import test_aeif_lsodar
+from . import test_connect_all_patterns
 from . import test_connect_all_to_all
+from . import test_connect_array_fixed_indegree
+from . import test_connect_array_fixed_outdegree
+from . import test_connect_distributions
 from . import test_connect_fixed_indegree
 from . import test_connect_fixed_outdegree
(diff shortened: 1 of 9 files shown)


clinssen:
Hi @stinebuu, thanks for the PR! I am agreed but have just one question: should we check the exit code of `do_test.sh` and error out completely if the script fails? This to make things more future-proof.

stinebuu:
@clinssen Thanks for your review! I am not quite sure what you mean with your comment though, if some of the tests fail, or the entire `do_test.sh` script fail, we already get an error message (as far as I know at least).

PR #840, opened 2017-10-13
Fix handling of PyNEST and SLI commandline arguments

jougs:
In particular, this makes the following changes:
    * Replace the the command line argument --sli-debug by --debug only right before loading the pynestkernel
    * Make all given command line arguments available in the statusdict
    * introduce --sli-debug to avoid confusion with Python's --debug command line argument
    * Remove mutation of argv by PyNEST

This fixes #779 and fixes #784.

Issue #779, opened 2017-07-05
Importing nest in python mutates sys.argv


maharjun:
The following lines of code

```
echo "import sys; print(sys.argv)" > temppy.py
python temppy.py testarg
```

results in the the following output
```
['temppy.py', 'testarg']
```

which is as expected. However, the following lines of code:

```
echo "import sys; import nest; print(sys.argv)" > temppy.py
python temppy.py testarg
```

result in the following:
```
[INFO] [2017.7.5 17:47:44 /home/arjun/RandomRepos/nest-simulator/nestkernel/rng_manager.cpp:238 @ Network::create_rngs_] : Creating default RNGs
[INFO] [2017.7.5 17:47:44 /home/arjun/RandomRepos/nest-simulator/nestkernel/rng_manager.cpp:233 @ Network::create_rngs_] : Deleting existing random number generators
[INFO] [2017.7.5 17:47:44 /home/arjun/RandomRepos/nest-simulator/nestkernel/rng_manager.cpp:238 @ Network::create_rngs_] : Creating default RNGs
[INFO] [2017.7.5 17:47:44 /home/arjun/RandomRepos/nest-simulator/nestkernel/rng_manager.cpp:284 @ Network::create_grng_] : Creating new default global RNG

              -- N E S T --

  Copyright (C) 2004 The NEST Initiative
  Version 2.12.0 Jul  5 2017 11:36:46

This program is provided AS IS and comes with
NO WARRANTY. See the file LICENSE for details.

Problems or suggestions?
  Visit http://www.nest-simulator.org

Type 'nest.help()' to find out more about NEST.
['pynest', 'testarg']
```

Here we clearly see that the importing of the nest module results in `sys.argv[0]` being changed to `pynest`

COMMIT ID: a721e62116fd57349439af4cf863835d596918b3
Cmake flags: `-Dwith-python=3 -Dwith-mpi=OFF -Dwith-optimize=ON -Dwith-debug=ON ../../nest-simulator/`

terhorstd:
That is very interesting. Is there a particular problem with this you need to have solved?

maharjun:
When using python, it is useful in many cases to have access to the command used to execute it (for example, I would like to store the command used to execute a run so that I can reproduce it in the future). If the name of the main file (`sys.argv[0]`) is overwritten then this becomes impossible.

[...]

Issue #1022, opened 2018-09-04
`run_examples.sh` could use wrong python executable


steffengraber:
On systems with different Python versions installed there is a problem when using `share/doc/nest/run_examples.sh`. With `runner=python` in line 70 only the standard Python version of the system is used, but if NEST was built with a different one, this could end in an error. 

Isn't it better to use something like that:

    PYRUNNER=$($NEST_INSTALL_DIR/bin/nest-config --python-executable)

and then:

    runner=$PYRUNNER

jougs:
I like this suggestion. Can you please review #1024?

Issue #972, opened 2018-06-13
`make installcheck` reaches wrong testing code after sourcing `nest_vars.sh` for a different version


TomBugnon:
Hi all,

I notice that when installing a given version of NEST using the usual routine:

```
rm -r bld
mkdir bld
cd bld
cmake ${SOURCE_DIR} -DCMAKE_INSTALL_PREFIX:PATH=${INSTALL_DIR}
make
make install
make installcheck
```

... the testing suite that is used for the install checks is not that of the NEST version that was just built, but rather that of the NEST version referenced by the environment variables from `nest_vars.sh`.

This caused me quite some headaches as I was trying to install nest versions with different test suites although the same `nest_vars.sh` script had been sourced in my .zshrc the whole time..

Adding the line `source "${INSTALL_DIR}/bin/nest_vars.sh"` before `make installcheck` seems to solve the issue.

terhorstd:
@jougs any updates on this?

jougs:
Basically, the code is ready, but depends on the merge of #1039. I'll start a PR once that's done.

clinssen:
Is this not already fixed by #1561?

jougs:
In principle yes (and please feel free to the magic *fixed* word to #1561).

The cleanest solution from my point of view is, however, to remove the problematic exports altogether in order to also prevent possible problems for people not running `make installcheck` or `do_tests.sh`, but
1. sourced `nest_vars.sh` previously for NEST installation _A_
2. manually re-set `PATH` and `PYTHONPATH` later on to make another NEST installation _B_ available (unknowingly that there are other variables set)
3. run into problems with _B_, as it still uses part of _A_.

And above that, as I laid out #1578, none of the exports is really needed for the proper working of NEST, as all exported variables are set to their defaults anyway. Actually, I don't see a good reason why one would want to use NEST *A* with the documentation or SLI library of NEST *B* and am considering to remove the mechanism for *NEST_DOC_DIR*/*NEST_DATA_DIR* altogether.

heplesser:
@jougs I agree that a complete clean-up is useful. Would you create a new issue for that, so this one can be safely closed when #1561 is merged?

PR #1023, opened 2018-09-06
Prevent execution of --debug and --verbosity options as files

### LINKED ISSUE #1016: --verbosity option does not work

jougs:
This is a replacement for #1017, which fixes #1016, while keeping the fix for #779 from #840 intact.

It also adds a regression test that uses a combination of `verbosity` and `debug` flags and makes sure they get not used as files and at all arguments still end up properly in the `argv` array in the `statusdict`.

### DIFF
--- a/testsuite/regressiontests/issue-779-1016.py
+++ b/testsuite/regressiontests/issue-779-1016.py
@@ -29,28 +29,22 @@ This is a regression test for GitHub issues 779 and 1016.
 from subprocess import check_output, STDOUT
 from os.path import join
 from tempfile import mktemp
-from sys import exit
+from sys import exit, version_info
 
 EXIT_SUCCESS = 0
 EXIT_FAILURE = 126
 EXIT_SKIPPED = 206
 
-try:
-    import nest
-except:
-    exit(EXIT_SKIPPED)
+decode = lambda x: x if version_info < (3,) else x.decode("utf8")
 
 nestscript = mktemp(".sli")
-nestcmd = [join(nest.sli_func("statusdict/prefix ::"), "bin", "nest"),
-           "-d",
-           "--verbosity=ALL",
-           nestscript]
+nestcmd = ["nest", "-d", "--verbosity=ALL", nestscript]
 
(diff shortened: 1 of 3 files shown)


jougs:
It seems that Travis does not have the `PYTHONPATH` set correctly during the execution of the test suite. Maybe this is a good excuse to first tackle #1022, #972 and #345?

testsuite/regressiontests/issue-779-1016.py
```diff
@@ -0,0 +1,60 @@
...
+nestcmd = [join(nest.sli_func("statusdict/prefix ::"), "bin", "nest"),
+           "-d",
+           "--verbosity=ALL",
+           nestscript]
+
+with open(nestscript, "w") as f:
+    f.write("statusdict/argv :: ==")
+
+raw_output = check_output(nestcmd, stderr=STDOUT)
+output = [x for x in raw_output.split("\n") if x != ""]
```
heplesser: According to Travis, this is not Py3 compatible.

heplesser:
@clinssen The magic 12 is actually the length of `--verbosity=` ;).

PR #1030, opened 2018-09-14
Use reverse iterator in get_history

jarsi:
This pull request replaces the iterator in get_history in the archiving_node with a reverse iterator. This reverse iterator runs through the spike history starting at the back rather than in the front. The change speeds up finding the relevant range in stdp synapses, as this range tends to be located in the back. A comparison of this change with nest-master (commit 620bcf3) using the hcp_benchmark.sli with 32 nodes and 8 threads on JURECA reveals a a speedup of 2.3 %.

### DIFF
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -164,14 +164,14 @@ nest::Archiving_Node::get_history( double t1,
   else
   {
     std::deque< histentry >::reverse_iterator runner = history_.rbegin();
-    while ( ( runner != history_.rend() ) 
+    while ( ( runner != history_.rend() )
       and ( t2 - runner->t_ < - 1.0
                 * kernel().connection_manager.get_stdp_eps() ) )
     {
       ++runner;
     }
     *finish = runner.base();
-    while ( ( runner != history_.rend() ) 
+    while ( ( runner != history_.rend() )
       and ( t1 - runner->t_ < - 1.0
                 * kernel().connection_manager.get_stdp_eps() ) )
     {
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -163,22 +163,22 @@ nest::Archiving_Node::get_history( double t1,
   }
   else
   {
-    std::deque< histentry >::iterator runner = history_.begin();
-    while ( ( runner != history_.end() )
-      and ( t1 - runner->t_ > -1.0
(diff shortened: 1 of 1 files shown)


nestkernel/archiving_node.cpp
```diff
@@ -163,22 +163,22 @@ nest::Archiving_Node::get_history( double t1,
   }
   else
```
heplesser: I would suggest to remove the `else` here. Since the `if` ends with a return, we don't really need it.

nestkernel/archiving_node.cpp
```diff
@@ -163,22 +163,22 @@ nest::Archiving_Node::get_history( double t1,
...
   {
-    std::deque< histentry >::iterator runner = history_.begin();
-    while ( ( runner != history_.end() )
-      and ( t1 - runner->t_ > -1.0
+    std::deque< histentry >::reverse_iterator runner = history_.rbegin();
+    while ( ( runner != history_.rend() )
+      and ( t2 - runner->t_ < -1.0
                 * kernel().connection_manager.get_stdp_eps() ) )
```
heplesser: If I understand the code right, then only `runner->t_` changes in the comparison. Would it then not be tidier and possibly more efficient to write
```c++
const double t2_lim = t2 + kernel().connection_manager.get_stdp_eps();
while ( ... and ( runner->t_ > t2_lim ) )
  ...
```
and the same in the loop below. At the very least we should re-organize the expression to avoid the multiplication with `-1.0`.

heplesser: I also think that you need to use `<=` in both cases to ensure identical behavior as before. Consider the case where `t1 + stdp_eps == runner->t_` for some spike. Then, in the old version, the t1-loop is not executed any more, the t2-loop is entered and the access-counter for that particular spike increased. In the new version the counter would not be increased in that case, unless we have `<=`.

nestkernel/archiving_node.cpp
```diff
@@ -163,22 +163,22 @@ nest::Archiving_Node::get_history( double t1,
...
     }
-    *start = runner;
-    while ( ( runner != history_.end() )
-      and ( t2 - runner->t_ > -1.0
+    *finish = runner.base();
+    while ( ( runner != history_.rend() )
+      and ( t1 - runner->t_ < -1.0
                 * kernel().connection_manager.get_stdp_eps() ) )
     {
       ( runner->access_counter_ )++;
```
heplesser: Are the parentheses here necessary?

Issue #268, opened 2016-03-09
postscript-sli-mode in emacs leads to unusable keys


mschmidt87:
When using postscript-sli-mode with the provided lisp files in emacs, some keys are no longer usable.
In particular, typing 'd', '[', and ']' leads to the same error, which reads
`Symbol's value as variable is void: last-command-char`.

Produced with Emacs 24.4.1.

heplesser:
We suggest using the standard `postscript-mode`, which is regularly maintained by an active external community.

ruedigerkupper:
Ohmygod!!! SLI-mode is gone??? Shame, shame, shame!!!
(Go ahead, cheers, Rüdiger)

PR #1033, opened 2018-09-17
Remove emacs support for SLI

### LINKED ISSUE #322: Update Emacs extension to use Git instead of svn.
The Emacs extension for sli, [`sli.el.in`](https://github.com/nest/nest-simulator/blob/master/extras/emacs/sli.el.in), still uses the SVN extension [`psvn.el`](https://github.com/nest/nest-simulator/blob/master/extras/emacs/psvn.el). The task in this issue is to update the `sli.el` to use the [Git extension](https://www.emacswiki.org/emacs/Git).

jougs:
This PR removes the SLI support files for emacs from the repository and modifies all corresponding files accordingly. The rationale behind this removal is that the extension is basically unmaintained and starts rotting (for an example, see #322).

I'm suggesting @mschmidt87 as additional reviewer, as he was a user of this extension once and might have an opinion here. I'm also mentioning @ruedigerkupper just for the fun of it ;-)

Fixes #322.


mschmidt87:
Hi, I am fine with this change and think that it makes sense. Although I use emacs, I actually never used the sli-mode, partly because of #268 . In #268 , @heplesser had already suggested to use the sli-mode instead.

The changes makes sense to me and I didn't find any remaining traces of emacs in the code, therefore :+1:

ruedigerkupper:
Jeees, I remember coding away days and nights with this mode. Still dreaming in reverse polish now and then...

/all_of_you state /good eq hope

Rüdiger

PR #1036, opened 2018-09-25
Add documentation in Topology User Manual

stinebuu:
Added documentation of elliptical/ellipsoidal masks and the rotation of masks in the Topology UserManual. Have also made it possible to plot the rotated rectangular mask.

The change is found under ch. 3.3 in the User Manual.


topology/doc/Topology_UserManual.tex
```diff
@@ -706,32 +706,52 @@ \subsection{Masks for 2D layers}\label{sec:free_masks}
...
+
+\scriptfig{conn2_b}{0.9}{Masks for 2D layers}%
+{The same masks as in Fig.~\ref{fig:conn2_a}, but
+ centered about $(-1.5,-1.5)$, $(-2,0)$, $(1.5,1.5)$ and $(2, -1)$,
+ respectively, using the \lstinline!'anchor'! parameter.}
+
+\noindent It is possible to rotate the $\textbf{rectangular}$ and
+$\textbf{elliptical}$ masks, see Fig~\ref{fig:conn2_c}. To do so, add an
+\lstinline!'azimuth_angle'!\index{azimuth\_angle} entry in the
+specific mask dictionary. The azimuth\_angle is measured in degrees and is the
```
heplesser: azimuth_angle should be lstinline here as well

topology/pynest/hl_api.py
```diff
@@ -2087,6 +2087,7 @@ def PlotKernel(ax, src_nrn, mask, kern=None, mask_color='red',
     import matplotlib
     import matplotlib.pyplot as plt
     import numpy as np
+    from math import cos, sin, pi
```
heplesser: Importing "naked" functions is not a good idea, in particular since we also have `numpy` here. I would use `import math` and then `math.cos()` etc below.

topology/doc/Topology_UserManual.tex
```diff
@@ -740,10 +760,18 @@ \subsection{Masks for 3D layers}\label{sec:3d_masks}
...
 As in the 2D case, you can change the location of the mask relative to the
-driver node by specifying a 3D vector in the
-\lstinline!'anchor'!\index{anchor} entry in the mask dictionary.
+driver node by specifying a 3D vector in the \lstinline!'anchor'!\index{anchor}
+entry in the mask dictionary. If you want to rotate the box or ellipsoidal
+masks, you can add an \lstinline!'azimuth_angle'!\index{azimuth\_angle} entry
+in the specific mask dictionary for rotation from the x-axis towards the y-axis,
+or an \lstinline!'polar_angle'!\index{polar\_angle} entry, specifying the
+rotation angle in degrees from the z-axis. You can specify both at once of
+course. NEST currently do not support rotation in all three directions.
```
heplesser: Could you be more explicit about which axis you cannot rotate at present?

stinebuu:
@heplesser I have added references to when elliptical masks and rotation was added, as well as tried to describe which rotational angle is missing. Please have a look and see if it makes sense.

I have also added a sentence about `SelectNodesByMask()` as this function was not documented either.

PR #1038, opened 2018-09-29
Create iaf_cond_beta

dnao:
I have created a new neuron model based on iaf_cond_alpha. This model uses a beta function to model synaptic conductance, with different time constants for the rise and decay time for both excitatory and inhibitory conductance:

tau_ex_rise
tau_ex_decay
tau_in_rise
tau_in_decay

The model is outlined in Rotter & Diesmann, Biol Cybern 81:381 (1999) and Roth and van Rossum, Ch 6, in De Schutter, Computational Modeling Methods for Neuroscientists, MIT Press, 2010.

I have also added two new parameters to specific a constant input conductance: a more realistic alternative to constant current inputs that has been used in some modelling studies. The new variables are:

G_ex
G_in

I have tested the model and the results are consistent with my expectations. I have also compiled the code and confirmed that it is running without issues. I have updated my fork with the latest changes from the upstream repo so the only changes should be the new models files and the new names for the new variables and model.


dnao:
Scratching my head a bit here. I ran extras/format_all_c_c++_files.sh on iaf_cond_beta.cpp and iaf_cond_beta.h but am still getting fail.. but only on two builds?

dnao:
OK I think I've fixed all the formatting issues.

Silmathoron:
Hey, sorry I'm a bit late.
I'm very busy until Friday 16 so I'll have a look at the end of next week, hope it's fine.

heplesser:
@Silmathoron No problem!

dnao:
Any updates? @Silmathoron

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+  assert( pnode );
+  const nest::iaf_cond_beta& node =
+    *( reinterpret_cast< nest::iaf_cond_beta* >( pnode ) );
+
+  // y[] here is---and must be---the state vector supplied by the integrator,
+  // not the state vector in the node, node.S_.y[].
+
+  // The following code is verbose for the sake of clarity. We assume that a
+  // good compiler will optimize the verbosity away ...
+  const double I_syn_exc =
```
Silmathoron: since they are used several time, I would consider declaring the variables in ``const double &`` to make code more readable

```suggestion
  const double& V    = y[ S::V_M ];
  const double& g_ex = y[ S::G_EXC ];
  const double& g_in = y[ S::G_INH ];

  const double I_syn_exc =
```

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+    ( y[ S::G_EXC ] + node.P_.g_ext_ex ) * ( y[ S::V_M ] - node.P_.E_ex );
```
Silmathoron:
```suggestion
    ( g_ex + node.P_.g_ext_ex ) * ( V - node.P_.E_ex );
```

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+  const double I_syn_inh =
+    ( y[ S::G_INH ] + node.P_.g_ext_in ) * ( y[ S::V_M ] - node.P_.E_in );
```
Silmathoron:
```suggestion
    ( g_in + node.P_.g_ext_in ) * ( V - node.P_.E_in );
```

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+  const double I_leak = node.P_.g_L * ( y[ S::V_M ] - node.P_.E_L );
```
Silmathoron:
```suggestion
  const double I_leak = node.P_.g_L * ( V - node.P_.E_L );
```

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+  // dV_m/dt
+  f[ 0 ] = ( -I_leak - I_syn_exc - I_syn_inh + node.B_.I_stim_ + node.P_.I_e )
+    / node.P_.C_m;
+
+  // d dg_exc/dt, dg_exc/dt
+  f[ 1 ] = -y[ S::DG_EXC ] / node.P_.tau_Edecay;
+  f[ 2 ] = y[ S::DG_EXC ] - ( y[ S::G_EXC ] / node.P_.tau_Erise );
```
Silmathoron:
```suggestion
  f[ 2 ] = y[ S::DG_EXC ] - ( g_ex / node.P_.tau_Erise );
```

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+  // d dg_exc/dt, dg_exc/dt
+  f[ 3 ] = -y[ S::DG_INH ] / node.P_.tau_Idecay;
+  f[ 4 ] = y[ S::DG_INH ] - ( y[ S::G_INH ] / node.P_.tau_Irise );
```
Silmathoron:
```suggestion
  f[ 4 ] = y[ S::DG_INH ] - ( g_in / node.P_.tau_Irise );
```

models/iaf_cond_beta.cpp
```diff
@@ -0,0 +1,521 @@
...
+      {
+        throw GSLSolverFailure( get_name(), status );
+      }
+    }
+
+    // refractoriness and spike generation
+    if ( S_.r )
+    { // neuron is absolute refractory
+      --S_.r;
+      S_.y[ State_::V_M ] = P_.V_reset; // clamp potential
```
Silmathoron: this line must be inside the while: the voltage must stay clamped during all the evolution of the ODE, not just "appear" clamped to the user recording it; only the decrement of ``S_.r`` should be outside

dnao:
Thank you for the review! Regarding the G_ex/G_in parameters, I thought this would be more prevalent but I have only been able to find a few examples of it in the literature. My belief was that, if modeling a small part of a larger neural network, one could simulate the inputs from the 'external' neurons as a constant conductance to each neuron as this would be more realistic than a constant current which really only occurs when clamping. However, it seems that the convention in the literature is to simulate such realistic external inputs as poisson spike trains, so perhaps I should follow this with my model.

The only other case I could find where this might apply is simulating a [dynamic clamp](https://nemenmanlab.org/~ilya/images/1/1a/Prinz-etal-04.pdf) which seems like a very niche usage.

models/iaf_cond_beta.cpp
dnao: This section is identical to the existing model [iaf_cond_alpha.cpp](https://github.com/nest/nest-simulator/blob/master/models/iaf_cond_alpha.cpp) so i am not too familiar with it. if there are any issues here they should also be corrected there:

dnao:
OK @Silmathoron, I have tested the model using poisson inputs and have confirmed that it reproduces the same behavior as with external constant conductance input. Hence I have removed them from the model. I have also incorporated your suggestion of replacing replacing y[S::V_M] with a variable V.

dnao:
OK I've renamed the constants as per your suggestion:

`tau_rise_[ex/in]` and `tau_decay_[ex/in]`

models/iaf_cond_beta.cpp
janhahne: @Silmathoron I think this is fine either way, as `S_.y[ State_::V_M ]` has no influence on the other ODEs in the system. Therefore the results will be the same, as long as it is reset properly at the end of the clamping

models/iaf_cond_beta.h
```diff
@@ -0,0 +1,428 @@
...
+the Fluctuation- Driven Regime. Jneurosci 24(10) 2345-2356
+
+Rotter & Diesmann, Biol Cybern 81:381 (1999)
+
+Roth and van Rossum,
+Ch 6, in De Schutter, Computational Modeling Methods for Neuroscientists,
+MIT Press, 2010.
+
+
+Author: Schrader, Plesser
```
janhahne: Please put your name here instead. (I suppose you modified a model that was written by Schrader, Plesser?)

models/iaf_cond_beta.cpp
heplesser: @Silmathoron @dnao Thanks for finding this! For the `iaf_cond_{alpha,beta}` models it makes indeed no difference as @janhahne points out, since `V_m` only enters into the ODE for `V_m`. But to do it really systematically, one needs to clamp inside the `..._dynamics` function. See `ht_neuron` for an example, where correct clamping is indeed required due to non-linear currents. 

@clinssen Would you take note of this in connection with NESTML?

Silmathoron: @janhahne @heplesser unfortunately that is not true: we are dealing with a conductance-based model here, so in case the ``while ( t < B_.step_ )`` is done in several steps (rather often in my experience), then the synaptic currents are wrong. It will arguably be a small error, but still...

janhahne: @Silmathoron I am sorry, I still dont see it: Could you point me to any line, where a (wrong) value of `S_.y[ State_::V_M ]` enters any variable or property that is used outside of the ODE that describes `S_.y[ State_::V_M ]` (specified by `f[0]`)? Of course the intermediate values of `S_.y[ State_::V_M ]` that are computed during the `while`-loop will be slightly wrong, but as long as they are not used for anything and do not influence any other variables the results at the NEST grid points will be the same as with clamping inside the while-loop. 

However, I completely agree that it would be better to do this systematically inside the `..._dynamics` function, as it becomes important once `S_.y[ State_::V_M ]` is also a part of other ODEs in the ODE-system.

Silmathoron: No, I did not mean outside of the ODE, but the computation of [``I_syn_ex``](https://github.com/nest/nest-simulator/pull/1038/files/9e982021d3d9684f77c75944e304c9ca5f9dd149#diff-76587cc1da7f6a4cffaf9ac00c1e944aR100) will be wrong, which will add up to the ``V`` error in the derivative.
However, as you said, these errors should remain small, but the integration is, strictly speaking, incorrect.
Regardless, I think this should be handled in another PR so we can ignore it for now, I'll open a new issue.

dnao:
Great, thanks everyone!

dnao:
Hi @pablomc88, thank you for the correction. It looks like I wrote it so that the integral of the conductance was 1.0 nS, rather than the peak conductance. For consistency with ht_neuron and aeif_cond_beta_multisynapse I believe it should be adjusted as you suggest. I will create a new pull request to do so.

heplesser:
@pablomc88 Thanks for catching this! Indeed, synaptic input functions in NEST should be normalized to a peak amplitude of 1 nS.

@dnao Thanks for your willingness to create a PR with a fix!

PR #1039, opened 2018-09-29
Make the check for copyright headers and unused Names part of the CI run

### LINKED ISSUE #1015: Tests require sources to be present
At least the two tests:

* [`regressiontests/ticket-659-copyright.py`](https://github.com/nest/nest-simulator/blob/master/testsuite/regressiontests/ticket-659-copyright.py)
* [`unittests/test_unused_names.py`](https://github.com/nest/nest-simulator/blob/master/testsuite/unittests/test_unused_names.py)

both look at the nest sources during `make installcheck`. In some cases the sources are not available during testing, e.g. packaging systems like homebrew. In the first one, if the env-var `NEST_SOURCE` is set to `SKIP`, the [test is skipped](https://github.com/nest/nest-simulator/blob/01ab7e2c1bbb5735c76a008a6c66bba358844c16/testsuite/regressiontests/ticket-659-copyright.py#L85). In the second, this is not respected. 

There should be a way to skip sources checking tests, or have them outside of `make installcheck` (aka "check the installation") as separate processes run before or after in CI (similar as static linters) .

At least for homebrew, this will hinder a formula for v2.16.0, as no patches are allowed for new formulas ([homebrew-core](https://github.com/Homebrew/homebrew-core/pull/24908#discussion_r180680447)).

jougs:
This PR runs the check for copyright headers and unused Names as part of the CI instead of
running them as part of the testsuite.

Fixes #1015.


extras/check_copyright_headers.py
```diff
@@ -119,22 +126,24 @@
...
                 if (extension == 'py' and
                         line_src.strip() == '#!/usr/bin/env python'):
                     line_src = source_file.readline()
                 line_exp = template_line.replace('{{file_name}}', fname)
                 if line_src != line_exp:
-                    print("Incorrect copyright header in '{0}':".format(tested_file))           # noqa
-                    print("    expected -> '{0}', actual -> '{1}'\n".format(line_exp.strip(),   # noqa
-                                                                            line_src.strip()))  # noqa
+                    fname = os.path.relpath(tested_file)
+                    eprint("[COPY] {0}: expected '{1}', found '{2}'.".format(
```
heplesser: I do not quite understand why some things are printed with `print()`, others with `eprint()`.  Could you document this in the code?

extras/check_unused_names.py
```diff
@@ -96,7 +103,9 @@ def get_names(fname, pattern):
 
 names_unused = names_defined - names_used
 if len(names_unused) != 0:
-    print("Unused Name definition(s): " + ", ".join(names_unused))
+    msg = "unused Name definition(s): " + ", ".join(names_unused)
+    eprint("[NAME] " + msg)
+    print("... {}\\n".format(msg))
```
heplesser: Explain somewhere `eprint()` vs `print()`.

extras/static_code_analysis.sh
```diff
@@ -109,6 +109,14 @@ if $RUNS_ON_TRAVIS; then
   fi
 fi
 
+
+export NEST_SOURCE=$PWD
+print_msg "MSGBLD0100: " "Running check for copyright headers"
```
heplesser: I don't remember the point of the numbering in the end. Is it correct that all messages here end with a `0100`?

extras/check_copyright_headers.py
jougs: I've added a docstring for `eprint()` and the general working of the script in 93258e33d8.

extras/check_unused_names.py
jougs: I've added a docstring for `eprint()` and the general working of the script in 93258e33d8.

jougs:
Actually, all tests have [succeeded](https://travis-ci.org/jougs/nest-simulator/builds/439124195). The previous failures were caused by package fetching timeouts on Travis, not by errors in the code.

extras/static_code_analysis.sh
gtrensch: The basic idea of message numbers is the decoupling of the message text from the log parser, e.g., correcting a typo in a message may cause the log parser `extras/parse_travis_log.py` to fail. There is of course still a logical coupling by the message numbers, but this is much easier to handle than arbitrary text.

MSGBLD0100 is not a good choice as it is already used for retrieving the list of changed files.I would suggest using MSGBLD1x00 while x should be chosen so that some numbers between 1040 - 1x00 are reserved. This simplifies possible later extensions of the script.

gtrensch: MSGBL0100 is already in use. See also the comment below.

gtrensch:
@jougs nice work! Please see my comment above. If you could change the message number, I would be very happy to approve this PR.

jougs:
@gtrensch: I've changed the message in 7d731fc1f

Issue #564, opened 2016-11-25
Autocomplete for PyNEST API functions in editors


zbarni:
Currently autocomplete features are not possibly with PyNEST in editors such as PyCharm. Reason for this are the dynamic imports in __init.py__:

    for name in os.listdir(os.path.join(os.path.dirname(__file__), "lib")):
        if name.endswith(".py") and not name.startswith('__'):
            exec("from lib.{0} import *".format(name[:-3]))`

One solution is to add imports manually, possibly also after the loop,

    from lib.hl_api_connections import *
    from lib.hl_api_helper import *
etc.
Future modules would require manual addition in the script, but at least the base functions are there.

heplesser:
@apeyser Could you take a look at this?

apeyser:
It shouldn't be exec, but it shouldn't be static either, but  should use importlib module (which probably doesn't fix PyCharm, but it's still cleaner than this).

heplesser:
Closing, the work in #1150 goes a way to addressing this, a perfect solution is not possible.

PR #1046, opened 2018-10-11
changed dynamic imports using exec to static imports

clinssen:
This addresses https://github.com/nest/nest-simulator/issues/564

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -248,16 +248,15 @@ def test():
 
     hl_api.set_debug(debug)
 
-from .pynestkernel import *         # noqa
-from .lib.hl_api_helper import *    # noqa
-
-# We search through the subdirectory "lib" of the "nest" module
-# directory and import the content of all Python files therein into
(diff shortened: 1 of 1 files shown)


terhorstd:
this one is related to #1058 and should be checked for compatibility

jougs:
Is Topology still working with this fix in place? The reason for the dynamic import was to support extension modules that install an optional Python component. I don't see how they'd be working with a purely static import.

clinssen:
Abandoning this PR due to on-point issues raised by @jougs! 

We could potentially switch to a static import of the existing api files and a dynamic import in a special directory containing only user modules, but this is probably better handled later/as part of a different PR, because of active refactoring in PyNEST.

babsey:
> static import of the existing api files and a dynamic import in a special directory containing only user modules

I like the idea of having static and dynamic imports.

apeyser:
It shouldn't be exec, but it shouldn't be static either, but should use importlib module (which probably doesn't fix PyCharm, but it's still cleaner than this).

Should be a different PR on this.

PR #1047, opened 2018-10-11
New container for connections

hakonsbm:
With 5g, connections are all put in one big vector. As the number of connections becomes large, memory consumption from vector growth becomes increasingly problematic. Some attempts were made to limit this problem (see jakobj/nest-simulator#116) and we also tried using std::deque as a container, both with limited success. Memory consumption is still bad, and random access with std::deque is inefficient and therefore slowed down simulations.

This PR implements a specialized container, named BlockVector, holding data in a vector of vectors. The size of each block (inner vector) is fixed to a number that is a power of 2, allowing use of bitwise operators to map an index to the right block and position in that block. Because of its fragmented data-structure, memory is better managed than when using a single vector, and using bitwise operators makes random access close to as efficient as using vector.

Additionally, when sorting connections, the `exchange_()` function is replaced with `std::swap()`, which somewhat reduces sorting time.

### DIFF
--- a/libnestutil/block_vector.h
+++ b/libnestutil/block_vector.h
@@ -1,5 +1,5 @@
 /*
- *  seque.h
+ *  block_vector.h
  *
  *  This file is part of NEST.
  *
@@ -20,8 +20,8 @@
  *
  */
 
-#ifndef SEQUE_H_
-#define SEQUE_H_
+#ifndef BLOCK_VECTOR_H_
+#define BLOCK_VECTOR_H_
 
 #include <cmath>
 #include <vector>
@@ -30,37 +30,37 @@
 #include <cassert>
 
 template < typename value_type_ >
-class Seque;
+class BlockVector;
 template < typename value_type_, typename ref_, typename ptr_ >
-class seque_iterator;
+class bv_iterator;
 
 constexpr int block_size_shift = 10; //!< max_block_size = 2^block_size_shift
 constexpr int max_block_size = 1L << block_size_shift;
 constexpr int max_block_size_sub_1 = max_block_size - 1;
 
 /**
- * @brief A Seque::iterator.
+ * @brief A BlockVector::iterator.
  * @tparam value_type_ Type of element.
  * @tparam ref_ Type of reference to the element.
  * @tparam ptr_ Type of pointer to the element.
  *
- * Seque holds one of this internally, marking the end of the valid range.
+ * BlockVector holds one of this internally, marking the end of the valid range.
  */
 template < typename value_type_, typename ref_, typename ptr_ >
-class seque_iterator
+class bv_iterator
 {
-  friend class Seque< value_type_ >;
+  friend class BlockVector< value_type_ >;
 
   // Making all templated iterators friends to allow converting
   // iterator to const_iterator.
   template < typename, typename, typename >
-  friend class seque_iterator;
+  friend class bv_iterator;
 
 private:
   template < typename cv_value_type_ >
-  using iter_ = seque_iterator< value_type_, cv_value_type_&, cv_value_type_* >;
+  using iter_ = bv_iterator< value_type_, cv_value_type_&, cv_value_type_* >;
 
-  const Seque< value_type_ >* seque_; //!< Seque to which this iterator points
+  const BlockVector< value_type_ >* block_vector_; //!< BlockVector to which this iterator points
   size_t block_index_; //!< Index of the current block in the blockmap
   //! Iterator pointing to the current element in the current block.
   typename std::vector< value_type_ >::const_iterator block_it_;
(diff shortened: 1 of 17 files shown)


ikitayama:
@hakonsbm - do you have some performance data collected on a supercomputer?

heplesser:
Some benchmark results attached, including from Jureca.
[Bench_214_sq_db70dcc1edb-2.pdf](https://github.com/nest/nest-simulator/files/2505835/Bench_214_sq_db70dcc1edb-2.pdf)

libnestutil/block_vector.h
```diff
@@ -0,0 +1,642 @@
...
+}
+
+template < typename value_type_ >
+inline typename BlockVector< value_type_ >::iterator
+BlockVector< value_type_ >::erase( const_iterator first, const_iterator last )
+{
+  assert( first.block_vector_ == this );
+  assert( last.block_vector_ == this );
+  assert( last < finish_ or last == finish_ );
+  // TODO: Except for some special cases, this method may kill performance as
```
jakobj: should these todos be addressed before merging?

nestkernel/source_table_position.h
```diff
@@ -46,6 +48,9 @@ struct SourceTablePosition
   template < typename T >
```
jakobj: remove outdated function

nestkernel/connector_base.h
```diff
@@ -174,6 +174,7 @@ class ConnectorBase
    * Sort connections according to source gids.
    */
   virtual void sort_connections( std::vector< Source >& ) = 0;
```
jakobj: remove outdated function

nestkernel/connector_base.h
```diff
@@ -451,7 +450,7 @@ class Connector : public ConnectorBase
   void
   reserve( const size_t count )
   {
-    C_.reserve( count );
+    //    C_.reserve( count );
```
jakobj: maybe add a comment why this is commented out, or remove function entirely if not used?

libnestutil/sort.h
```diff
@@ -180,7 +282,14 @@ quicksort3way( std::vector< T1 >& vec_sort,
  */
 template < typename T1, typename T2 >
 void
-sort( std::vector< T1 >& vec_sort, std::vector< T2 >& vec_perm )
+sort( std::vector< T1 >& vec_sort, BlockVector< T2 >& vec_perm )
```
jakobj: necessary? (see above)

libnestutil/sort.h
```diff
@@ -68,7 +70,7 @@ median3_( const std::vector< T >& vec,
 template < typename T1, typename T2 >
 void
 insertion_sort( std::vector< T1 >& vec_sort,
```
jakobj: why do we need the combination of `vector` and `BlockVector`? shouldn't the sorting always be called on two `BlockVector`s? there's heavy code duplication here, so either this function should be removed or integrated with the sort on two `BlockVector`s

testsuite/cpptests/test_sort.h
```diff
@@ -49,51 +49,102 @@ is_sorted( std::vector< size_t >::const_iterator begin,
...
 
 /**
  * Tests whether two arrays with randomly generated numbers are sorted
  * correctly by a single call to sort.
  */
-BOOST_AUTO_TEST_CASE( test_random )
+BOOST_AUTO_TEST_CASE( test_random_vec_seq )
 {
   const size_t N = 20000;
   std::vector< size_t > vec0( N );
```
jakobj: why test the combination of `vector` and `BlockVector`?

libnestutil/block_vector.h
hakonsbm: The function isn't that much used anyway, and making it more efficient would require quite a bit of mental work. So I removed the TODO.

nestkernel/source_table_position.h
hakonsbm: Removed!

nestkernel/connector_base.h
hakonsbm: Removed!

nestkernel/connector_base.h
hakonsbm: I removed the function entirely, as one can't (and shouldn't) reserve memory for the new container.

libnestutil/sort.h
hakonsbm: Removed it!

hakonsbm:
@jakobj Thanks for your review! I have updated the PR and addressed your comments.

libnestutil/sort.h
```diff
@@ -23,32 +23,34 @@
...
 /**
- * Exchanges elements i and j in vector vec.
+ * Calculates the median of three elements.
+ * See http://algs4.cs.princeton.edu/23quicksort/QuickX.java.html.
  */
 template < typename T >
-inline void
-exchange_( std::vector< T >& vec, const size_t i, const size_t j )
+inline size_t
+median3_( const std::vector< T >& vec,
```
jakobj: this can probably also be removed

hakonsbm:
@jakobj I agree, `reserve_connections` in `GenericConnectorModel` can be removed. And how about [this one](https://github.com/hakonsbm/nest-simulator/blob/098fc61b2a6bc3e46206fda2ee246ee58402b3bc/nestkernel/connection_manager.cpp#L1453) in `connection_manager`?

jakobj:
I would remove those and the corresponding calls from `conn_builder.cpp` as well. Otherwise we will wonder soon why these function exist ;)

heplesser:
@hakonsbm I agree with the removal of the `reserve()` calls for now. But could you create a branch or tag (at least in your repo) that preserves them---maybe we will find a way of using them for optimization in the future.

ikitayama:
@hakonsbm I'd like to test the latest development of your branch if possible.

hakonsbm:
@heplesser I have created a [tag in my repo](https://github.com/hakonsbm/nest-simulator/releases/tag/before_removing_reserve).

@ikitayama Please do! :+1: 
https://github.com/hakonsbm/nest-simulator/tree/new_container

ikitayama:
I’ll share the results with you and the usual suspects off the list alright?

hakonsbm:
That's okay for me.

Issue #1043, opened 2018-10-09
NEST is running out of space for synapse models


heplesser:
NEST currently supports only 63 different synapse models. The current master has already 55 synapse models, so that only very few additional synapse models can be created with `CopyModel`.  Adding more synapse models makes the situation even worse, see #865. 

The number of synapse models available by default is as large as 55, because many synapse models exist in multiple versions (plain, `_lbl`, `_hpc`).

I believe we need to discuss increasing the space for synapse models to at least 255 and reconsider if all model variants should be created even if they are never used: `_lbl` is used by PyNN, which then does not use the plain version, and models using `_hpc` variants do not use the plain variant simultaneously.

jakobj:
yes, the fact that one can only create few additional synapse models is certainly an issue. here are my thoughts:
1) we should explicitly discourage people from use copy model for identical models with different parameters settings as this not only exhausts the maximal number of synapse types but also can incur runtime penalties
2) we should reduce the default synapse model count by only registering the "_lbl" and "_hpc" synapses when explicitly requested during compile time
3) we should add a cmake flag that allows users to increase the number of possible synapses at the expense of maximal mpi ranks/threads/number of neurons. the choice of concrete numbers will have to be discussed (if we increase the number of possible synapse models without introducing additional constraints, we will increase the memory usage of NEST)

jougs:
@jakobj: I was assigned to this mainly for mental support, not for the coding ;-)

First, let me reply to [your comment](https://github.com/nest/nest-simulator/issues/1043#issuecomment-429878910):
* Discouraging the use of `CopyModel` probably won't help much if NEST already brings 55 types and the user anyway only has 8 types left for customization.
* Not registering the special types probably leads to quite some user complaints as PyNN needs the `_lbl` types, some examples and existing code use the `_hpc` types and people generally don't expect a recompilation to be necessary for such things to work.

The current split of the `Target` union into the different fields allows 1,048,576 ranks (20 bit), 1,024 threads per rank (10 bit), 64 synapse ids (6 bit), 134,217,728 local synapses (27 bit) and a flag for knowing if the synapse has been processed. These numbers seem to be optimized for the super-large scale, while creating problems for _normal_ users in interactive sessions, where 8 custom synapse types may quickly be used up during exploration (again, NEST already populates most of the space and NESTML will make the creation of new  synapse types much easier soon...)

[...]

PR #1049, opened 2018-10-15
Add CMake flag to enable/disable hpc synapses

jakobj:
under the assumption that most users do not use the hpc versions of the synapses, this PR adds a cmake option to enable/disable(default) hpc synapse models. this allows for more custom models to be registered before running out of possible synapse models. partly addresses #1043 

@heplesser @suku248


ikitayama:
Could you keep it enable by default? I’m one of those HPC users.

jakobj:
thanks for the comment @ikitayama. i think this is a matter of taste. since most users will not use hpc synapses, and those who do, including you, know how to adjust cmake 
options to their desire, I thought it would be more reasonable to set the default to off, but I am open to discuss this.

ikitayama:
@jakobj - I suppose a new CMake variable would be easily overlooked thus ended up a binary that is built to not load the HPC-themed models. And an HPC user might waist his/her couples of hours wondering at to why the models aren't loaded at run time. This is the scenario I worry about.

Given that we'd let many modules piled up in the code base, I'd like to oppose introducing module
load/unload switch in a ad-hoc fashion.

jakobj:
closing, see #1043

PR #1057, opened 2018-10-23
build using clang and gcc compilers

lekshmideepu:
Added two new matrix builds for gcc4.x and clang5.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -236,11 +236,6 @@ before_install:
      - pip install cython scipy matplotlib terminaltables
 
 install:
-   - echo $CXX
-   - echo $CC
-   #- eval "${MATRIX_EVAL}"
-   #- echo $CXX
-   #- echo $CC
    - which cython
    - cython --version
    - which python
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,13 +66,17 @@ matrix:
 
     include:
     - language: cpp
-      compiler: gcc      
+      compiler: gcc-8
+      env:
+        - C_COMPILER=gcc-8
+        - CXX_COMPILER=g++-8       
       addons:
         apt:
           sources:
             - r-packages-trusty
             - ubuntu-toolchain-r-test
           packages:
+            - g++-8
             - build-essential
             - cmake
             - libltdl-dev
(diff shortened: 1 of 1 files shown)


heplesser:
@lekshmideepu Thanks for this PR. I have a few questions:
- Aren't we using gcc 4.8.4 as default, and wouldn't it make more sense to test with a newer gcc (8.x)?
- Why clang 5 and not the current clang 7?
- Python 2.7 is nearing end-of-life. Shouldn't we set up all new tests with Python 3?

lekshmideepu:
@heplesser : I fully agree with you regarding the versions.

- Yes, GCC 4.8.4 is the default. In fact I tried to use a newer version of  GCC and for some strange reason it was not recognised by the Travis build system. I also had a discussion with their support team two months back. Their suggestions didn't worked out.  Package whitelist contains only [GCC6.0](https://github.com/travis-ci/apt-package-safelist/blob/master/ubuntu-trusty#L1301)
- Same with [clang](https://github.com/travis-ci/apt-package-safelist/blob/master/ubuntu-trusty#L482)
- Regarding Python, in my opinion we could may be keep just 1 Python2.x build as the sunset date for Python2.x will be on January 1st, 2020

lekshmideepu:
@heplesser : I would contact the Travis team again for the issues when using the latest version of the compilers. Probably we could solve this once we have the NEST docker builds.

lekshmideepu:
https://github.com/travis-ci/travis-ci/issues/10292

lekshmideepu:
@heplesser With [this](https://travis-ci.org/nest/nest-simulator/jobs/446037505#L1132) export line under the install section, it seems like the build is using the user specified compilers. 
As of now, GCC8.0 is used for the extra GCC build and clang7.0 for the specific clang build. For other matrix builds, it is using the default versions.

lekshmideepu:
@heplesser : sure. I already have a sample build [here](https://travis-ci.org/lekshmideepu/nest-simulator/builds/450909094) . Only the OSX build failed here.

heplesser:
@lekshmideepu Could you comment on @jougs' question?

lekshmideepu:
Thanks @jougs and @heplesser  I think it is better to merge this.

PR #1059, opened 2018-10-24
Fix issues in siegert_neuron

AlexVanMeegen:
This is the PR for some issues detected and fixed by @mhelias in the `siegert_neuron`:
1. The absence of input was caught with `std::abs( mu - 0. ) < 1e-12`, which is only true for no / low noise levels. Fix: Use `( theta_shift - mu ) > 6. * sigma` based on Brunel 2000, eq. (22).
2. The relative and absolute errors for the numerical GSL integration routine were fixed to `0.1` for both `siegert1` and `siegert2`. However, their absolute errors differ considerably. Fix: Use zero absolute error and relative error of `1.49e-8` for both.
3. The GSL workspace was allocated on every call. Fix: Only allocate the workspace on creation of the `siegert_neuron`.

A test relied on the first issue to create a `siegert_neuron` modelling the external drive. This is also fixed by adjusting `theta` for this neuron.

### DIFF
--- a/models/siegert_neuron.cpp
+++ b/models/siegert_neuron.cpp
@@ -280,8 +280,7 @@ nest::siegert_neuron::siegert1( double theta_shift,
   }
 
   gsl_integration_qags(
-    &F, lower_bound, upper_bound, 0.0, 1.49e-8, 1000, gsl_w_, &result, &error
-  );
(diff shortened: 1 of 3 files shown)


heplesser:
@janhahne 2^-26 = 1.4901161193847656e-08, i.e., this is machine precision for single-precision floats and a common limit used in several packages.

AlexVanMeegen:
Thanks for the reviews!

Issue #971, opened 2018-06-12
allow_offgrid_times vs allow_offgrid_spikes


heplesser:
This is a follow-up to #967.

`spike_generator` has parameter `allow_offgrid_spikes` that controls if spike times can be set to times not multiples of the step; such times are rounded up to the next step.

`inhomogeneous_poisson_generator` and `step_current_generator` have parameter `allow_offgrid_times` that similarly control the setting of times at which poisson rate or output current are changed to new values.

Does it really make sense to have these two almost equal names? Should we standardize to one, which in this case would have to be `allow_offgrid_times`, since `..._spikes` makes no sense for the current generator and would be misleading for the poisson generator.

abigailm:
It doesn't make sense. We should standardize to `allow_offgrid_times` or simply `allow_offgrid`

heplesser:
@stinebu Ping!

stinebuu:
@heplesser  I have just fixed this in the nest-3 branch, see f2242bc2576863bb5e514af07f275da6c84f1634. Should I close this, or are we waiting for the branch to be merged into master?

heplesser:
@stinebuu I think we decided to leave the issues open until NEST-3 is merged.

Issue #679, opened 2017-03-14
Replacement needed for PrintNetwork


heplesser:
With subnets about to be removed by #480, the old `PrintNetwork` function needs replacement. There are three issues:

1. Since the simulator no longer represents network hierarchy, we need another way to present the network. 
1. The name `PrintNetwork` was always a bit of a misnomer, since it prints only the nodes.
1. `PrintNetwork` is hard-coded to write to `cout`, which breaks at the very latest when calling the command from Python.

Here is a proposal for a replacement for PrintNetwork.

## Content
Output should present the network information available from Modelrange Manager, i.e., GID ranges for different models. I would propose an output format like this:
```
  1 .. 100 iaf_psc_alpha
101 .. 102 poisson_generator
103        spike_detector
```
One could debate if 
- the last line should be `103 .. 103`
- model ID should be shown
- there should be a table header (similar to the format of dictionary output)
Column width is chosen based on the maximum GID in the network

## Command Name
The command name should reflect that the command prints node information only. Also, we want a command without any arguments for immediate output, but we may need one taking an output stream for redirecting output (see Python below). I would suggest (in SLI)
- `PrintNodes` as command without arguments
- `PrintNodesToStream` as command with ostream  argument
I think we cannot safely overload these two with a trie.

## Python
In Python, we have only `PrintNodes()`. This calls `PrintNodesToStream` with an `ostringstream` argument and then uses Python `print()` to output the result. This ensures that the command also works properly in Jupyter Notebooks.

stinebuu:
The changes for this is implemented in the [compneuronmbu SubnetFreeTopology](https://github.com/compneuronmbu/nest-simulator/tree/fix-481-SubnetFreeTopology) branch.

stinebuu:
@heplesser As this is implemented in the [nest-3 branch](https://github.com/compneuronmbu/nest-simulator/tree/nest-3) do we want to close this issue to be able to close the [Remove subnets](https://github.com/nest/nest-simulator/projects/2#card-2069522) project?

heplesser:
Closing as this is fixed in the NEST 3 branch.

PR #1060, opened 2018-10-25
Added deprecation warnings and changed format of the warnings

stinebuu:
As we in NEST 3.0 will replace `allow_offgrid_spikes` with `allow_offgrid_times` in the `spike_generator`, #971, I have added `allow_offgrid_times` to `spike_generator` and added a deprecation warning for `allow_offgrid_spikes`  which is shown the first time a user change the value for `allow_offgrid_spikes`. I added the `allow_offgrid_times` flag so that users can change to this value now if they want to.

I have also changed the way deprecation warnings are displayed on Python level. When a deprecation warning is shown in Python, the deprecation warning is displayed, but also
```
warnings.warn('\n' + text)    # add LF so text starts on new line
```
([this line](https://github.com/nest/nest-simulator/blob/master/pynest/nest/lib/hl_api_helper.py#L91)). So I have added a `format_Warning` function so we just see the actual warning. In addition, deprecation warnings on model types will now only be shown the first time they are created.

Lastly, I have added a deprecation warning to `PrintNetwork` as this will be replaced by `PrintNodes` in NEST 3.0, #679. To let users change to `PrintNodes` if they wish, I have added this function, which right now just call `PrintNetwork()`, but is implemented in the NEST 3 branch.


pynest/nest/lib/hl_api_helper.py
```diff
@@ -41,8 +41,17 @@
 sps = spp = sr = pcd = kernel = None
 
 # These flags are used to print deprecation warnings only once. The
-# corresponding functions will be removed in the 2.6 release of NEST.
-_deprecation_warning = {'BackwardCompatibilityConnect': True}
+# corresponding functions will be removed in a future release of NEST.
```
jougs: `<3` removing the version number from this comment ;-)

models/spike_generator.h
```diff
@@ -249,18 +250,22 @@ class spike_generator : public DeviceNode
     bool precise_times_;
 
     //! Allow and round up spikes not on steps; irrelevant if precise_times_
+    bool allow_offgrid_times_;
+    //! Same as allow_offgrid_times_, here for backward compatibility
```
heplesser: Keep only `allow_offgrid_times_` internally. We need backward compatibility of the *user interface*, not of the implementation.

models/spike_generator.h
```diff
@@ -249,18 +250,22 @@ class spike_generator : public DeviceNode
...
     bool allow_offgrid_spikes_;
 
     //! Shift spike times at present to next step
     bool shift_now_spikes_;
 
+    bool deprecation_warning_issued_;
```
heplesser: We may have deprecation warnings also on other parameters in other models in the future. We should find a general way of handling them, so we don't need to implement it in each node class that has a deprecated parameter. Maybe a general class for handling deprecations (essentially a map from names to flags), and then a member of this class in the `Node` base class.

models/spike_generator.cpp
```diff
@@ -178,16 +183,43 @@ nest::spike_generator::Parameters_::set( const DictionaryDatum& d,
...
   const Time& now )
 {
-  const bool flags_changed =
+  const bool allow_offgrid_spikes_changed = updateValue< bool >(
+    d, names::allow_offgrid_spikes, allow_offgrid_spikes_ );
+  const bool allow_offgrid_times_changed =
+    updateValue< bool >( d, names::allow_offgrid_times, allow_offgrid_times_ );
+
+  if ( allow_offgrid_spikes_changed and allow_offgrid_times_changed
+    and not allow_offgrid_spikes_ == allow_offgrid_times_ )
```
heplesser: We should simply forbid using old an new parameter name in the same dictionary.

jougs:
@heplesser: I think we should not make the mechanisms for handling deprecations much more sophisticated than they are. In my view most people are anyway not reading them and only change things once they break. Insofar, my take would even be to relax our stability policy for NEST3 and allow breaking changes between releases and only keep things backward compatible in special cases.

stinebuu:
@jougs @heplesser thanks for the reviews! I have now removed most of the references to `allow_offgrid_spikes` internally.

When it comes to the deprecation warnings, I have started working on it, but I am dealing with some linking problems. I haven't been able to work on it that much lately, because of all the work with the benchmarks, so I don't know when I'll be able to work on it again. I'll let you discuss if it is needed. When deprecating a parameter like I have done here, you only need to add a `deprecation_warning_issued_` parameter to the model (and write the actual warning), so it is not that difficult to deprecate parameters with the setup we already have (especially since we already have the `M_DEPRECATED` error code).

jougs:
@stinebuu @heplesser: it's probably best to discuss the exact policy at the hackathon in a wider group. Maybe we can even come up with a small write-up to circulate among the developers

jougs:
@stinebuu, @heplesser: what's the status of this? Thanks!

heplesser:
@jougs @stinebuu Thanks for the work and discussions! I agree that we can stay with the current framework. We can think about general solutions in context of code generation. I will approve and merge.

PR #1041, opened 2018-10-05
New User Documentation for NEST

steffengraber:
This pull request reorganizes the documentation files. It also sets up sphinx and readthedocs as the default user-level documentation generator and platform, respectively.

The documentation is structured under the doc/ folder instead of extras/.../.
The python documentation generator, sphinx, builds the documentation from restructured text files and the configuration file (conf.py).

Several new files (/topic_folder/index.rst) are created for indexing and generating a logical table of contents.
The markdown files will eventually be removed but remain for the time being as we ensure
the new set up works properly with restructured text.

Changes to the content have begun but **this is a work in progress** and we plan to  go through the documentation to make necessary improvements (such as fixing apis and example python scripts, adding navigation links, editing content, formatting the layout etc).

PR #1062, opened 2018-10-25
Source code comment blocks refactored

clinssen:
The comment blocks describing each C++-class and SLI function were refactored. These comment blocks are all of the same style, with headers such as "Name", "Parameters", "Authors", "SeeAlso", etc. Their placement was, however, inconsistent, such that doxygen could not link them to the class they belong to. For example, the following construct was common:

    /** @BeginDocumentation
    Name: aeif_cond_alpha_RK5 - Conductance based exponential integrate-and-fire
                            neuron model according to Brette and Gerstner (2005)
    [...]
    */
    namespace nest
    {
    class aeif_cond_alpha_RK5 : public Archiving_Node
    {
    [...]

and was refactored by moving the comments block to just before the class declaration. The upshot of this is that doxygen then picks up on the comments and includes them in the generated HTML documentation.

The `generate_help.py` script in `extras/doc` parses these same comment blocks for the purposes of generating the "NEST helpdesk" HTML pages. The script was also refactored so that it continued to work and became a bit more robust. E.g.:
- HTML entities are now escaped ("<" becomes "&amp;lt;", and so on);
- CSS is used for formatting instead of `<pre>`;
- removed stray whitespace from HTML output; all sections now look the same;
- things are in general less fragile, e.g. parsing a "Parameters" section where the user has created additional sub-headings like "Membrane parameters:" will now work.

Other minor changes:
- The doxygen index page now renders README.md from the repository root. This ensures that this page is always up-to-date and dissolves earlier questions about what to do with the doxygen index page.
- The keyword "BeginDocumentation" was changed globally to "@BeginDocumentation" because this allows doxygen to ignore it but the help generator script to pick up on it.
- A few files contained duplicate comment blocks: once in the style described above, and another, separate block for doxygen. Often they contained complementary information, and we have, for each file, merged them without losing any information overall.

With mad props to @jessica-mitchell!


sli/processes.h
```diff
@@ -973,9 +962,8 @@ SeeAlso: available, ignore
 */
 
 
-/*
-BeginDocumentation
-
+/**
+@BeginDocumentation
```
steffengraber: There should be no line break between '/**' and  '@BeginDocumentation'

steffengraber:
Modifications missing in `conngen/conngenmodule.cpp`.

clinssen:
Just a comment: with https://github.com/nest/nest-simulator/pull/1041/commits/73e02ca2ebde9d3ad1429086795dd0990c34d5f0, all MarkDown files have been replaced by reStructuredText. Unfortunately doxygen cannot parse rst, so the index page of the documentation shows up blank instead of containing the HTML-rendered README.md/.rst. I don't see any immediate/obvious fix (especially without introducing new dependencies on e.g. conversion scripts), but fortunately this is not a show-stopper for this PR.

clinssen:
Hi @jougs, thanks for your feedback! After talking with @jessica-mitchell, I understand that the plan is to keep the main nest-simulator README as an .md for now, because GitHub also has some trouble rendering .rst. So the doxygen fix can stay in place.

As for the copyright headers, as far as I can see they're already there! I merged your other suggestion using the web interface—thanks for that.

PR #1065, opened 2018-10-30
32-bit compatibility (64-bit gid on 32-bit architecture)

### LINKED ISSUE #1031: 'success' in 'struct nest::StaticAssert...' does not name a type
I'm getting a failed build with GCC 8.2.1 on Fedora 30 systems for the nest 2.16.0 release. I'm still looking at the cause, but here is the error message.

```
/builddir/build/BUILD/nest-simulator-2.16.0/nestkernel/target.h:135:48: error: 'success' in 'struct nest::StaticAssert<false>' does not name a type
 typedef StaticAssert< sizeof( Target ) == 8 >::success success_target_size;
                                                ^~~~~~~
```

The compiler flags are:

```
C compiler flags    : -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m32 -march=i686 -mtune=generic -msse2 -mfpmath=sse -mstackrealign -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -O2 -Wall -fopenmp -fdiagnostics-color=auto

C++ compiler flags  : -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m32 -march=i686 -mtune=generic -msse2 -mfpmath=sse -mstackrealign -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -std=c++11 -O2 -Wall -fopenmp -fdiagnostics-color=auto
```

Cheers!

gtrensch:
This pull request resolves #1031.

This pull request fixes an issue with compiling and running NEST on a 32-bit platform. It has been tested on a 32-bit ARM Raspberry Pi.

The source code changes do the following: 
- Let CMake identify when compiling for a 32-bit architecture and then include the appropriate headers. Even on 32-bit platforms there is usually an `uint64_t` available in `stdint.h`.
- Define the `gid`explicitly as 64-bit. Both, `unsigned long` and `unsigend long long` are not appropriate for this. They are platform-dependent data types!
- The 64-bit gid on 32-bit required a few more changes down to sli-types, i.e., overloading constructors and ensuring that this does not conflict with data types on 64-bit platforms, i.e., `uint64_t` might be equivalent to `unsigned long` preventing overloading.

I found that it is not needed for the moment to change the types of `index` and `thread` as they are compiled from a subset of the bits still fitting 32-bit. Making this `uint64_t` breaks sli-types and will require more changes. Since we decided to follow the C++11 standard, I also introduced `constexpr` for the constant-definitions to ensure their evaluation at compile-time.


nestkernel/nest_types.h
```diff
@@ -75,10 +85,10 @@ namespace nest
  * different architectures (e.g. 32 or 64 bit).
  */
 
+
```
apeyser: Extra space

nestkernel/source.h
```diff
@@ -40,25 +40,25 @@ namespace nest
...
 private:
-  unsigned long gid_ : 62; //!< gid of source
-  bool processed_ : 1;     //!< whether this target has already been moved
-                           //!to the MPI buffer
+  uint64_t gid_ : 62;  //!< gid of source
+  bool processed_ : 1; //!< whether this target has already been moved
+                       //!to the MPI buffer
   bool primary_ : 1;
-  static const size_t disabled_marker_ = 4611686018427387904 - 1; // 2 ** 62 - 1
+  static constexpr uint64_t GID_DISABLED = 0x3FFFFFFFFFFFFFFF; // 2 ** 62 - 1
```
apeyser: Much better! (And checked)
But maybe even better would be (1 << 62) - 1 ?

nestkernel/target.h
```diff
@@ -41,31 +41,31 @@ namespace nest
...
+  static constexpr uint8_t SHIFT_TID = 47;
+  static constexpr uint8_t SHIFT_SYN_ID = 57;
+  static constexpr uint8_t SHIFT_PROCESSED = 63;
 
   // maximal sizes are determined by bitshifts
-  static const int max_lcid_ = 134217728; // 2 ** 27
-  static const int max_rank_ = 1048576;   // 2 ** 20
-  static const int max_tid_ = 1024;       // 2 ** 10
-  static const int max_syn_id_ = 64;      // 2 ** 6
+  static constexpr int MAX_LCID = 134217728; // 2 ** 27
```
apeyser: Maybe these should be in hex to make it clearer?
0x8000000
0x10000000

and so forth?

Or even better,
1 << 27, 1 << 20, 1 << 10, 1 << 6 ?

nestkernel/target.h
```diff
@@ -165,73 +165,73 @@ inline Target::Target( const thread tid,
...
 {
-  assert( tid < max_tid_ );
-  data_ = ( data_ & ( ~tid_mask ) )
-    | ( static_cast< unsigned long >( tid ) << tid_shift );
+  assert( tid < MAX_TID );
+  data_ =
+    ( data_ & ( ~MASK_TID ) ) | ( static_cast< uint64_t >( tid ) << SHIFT_TID );
 }
 
 inline thread
```
apeyser: Just realized: anyone checked that the `data_` top bit and the masks following guarantee either that the top bit is 0 after masking, or that the return cast throws away any top 1s created by right shifting?

sli/token.cc
```diff
@@ -69,6 +69,13 @@ Token::Token( unsigned long value )
   p = new IntegerDatum( value );
 }
 
+#ifdef HAVE_32BIT_ARCH
```
apeyser: Need to check whether `sizeof(long) != 64` -- implementations even of 32 bits could, in principle, have a 64 bit long which would cause a compilation error.

nestkernel/source.h
gtrensch: Yes. This comment has now induced me to refactor the entire class, despite the initial intention to just make it work again on 32-bit. I am going to push a new version.

nestkernel/target.h
gtrensch: A refactored version of the code will follow.

nestkernel/target.h
gtrensch: I am not sure if I understand your comment correctly. I think the author's intention is to set the thread-id (the bits defined by the bit-mask) without touching anything else. The high-order "processed-bit" is toggled elsewhere and must not be changed here.

sli/token.cc
gtrensch: I'll add this to the CMake scripts to be on the safe side. I am not aware of any 32-bit system with a 64-bit long. Doesn't it violate the ILP32 data model?

apeyser: Neither am I aware of that -- but it's just better to fail violently for corner cases that may appear and be difficult to identify later.

nestkernel/target.h
apeyser: Checked: remote_target_id_ is unsigned.

nestkernel/target.h
```diff
@@ -75,163 +114,197 @@ class Target
...
+    return ( TARGET_ID_PROCESSED );
+  }
+  return ( TARGET_ID_UNPROCESSED );
 }
 
 inline bool
 Target::is_processed() const
 {
-  return ( data_ & processed_mask ) >> processed_shift;
+  if ( Target::get_status() == TARGET_ID_PROCESSED )
```
apeyser: Guess it's a question of taste, I prefer the conciseness of `return Target::get_status() == TARGET_ID_PROCESSED`

gtrensch: Yes, much better!

cmake/CheckIncludesSymbols.cmake
```diff
@@ -53,6 +53,28 @@ if ( UINT16_T_SIZE GREATER 0 )
...
+check_type_size( uint64_t UINT64_T_SIZE )
+set( CMAKE_EXTRA_INCLUDE_FILES OFF )
+if ( UINT64_T_SIZE GREATER 0 )
+    set( HAVE_UINT64_T ON )
+endif ()
+
+if( CMAKE_SIZEOF_VOID_P EQUAL 4 )
+    set( CMAKE_EXTRA_INCLUDE_FILES "stdint.h" )
+    check_type_size( "long" LONG_SIZE )
+    if( NOT LONG_SIZE EQUAL 8 )
```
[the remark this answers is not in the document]
gtrensch: I agree, it is much better to terminate CMake with a proper error message. I changed the script accordingly.

gtrensch:
@apeyser thanks for your comments! I have included your suggestions in the code.

cmake/CheckIncludesSymbols.cmake
```diff
@@ -53,6 +53,30 @@ if ( UINT16_T_SIZE GREATER 0 )
...
+set( CMAKE_EXTRA_INCLUDE_FILES OFF )
+if ( UINT64_T_SIZE GREATER 0 )
+    set( HAVE_UINT64_T ON )
+endif ()
+
+if( CMAKE_SIZEOF_VOID_P EQUAL 4 )
+    set( CMAKE_EXTRA_INCLUDE_FILES "stdint.h" )
+    check_type_size( "long" LONG_SIZE )
+    if( NOT LONG_SIZE EQUAL 4 )
+        message( FATAL_ERROR "Platform not supported: None-ILP32 data model for 32 bit architecture." )
```
sanjayankur31:
```suggestion
        message( FATAL_ERROR "Platform not supported: Non-ILP32 data model for 32 bit architecture." )
```

cmake/CheckIncludesSymbols.cmake
```diff
@@ -53,6 +53,30 @@ if ( UINT16_T_SIZE GREATER 0 )
   set( HAVE_UINT16_T ON )
 endif ()
 
+set( CMAKE_EXTRA_INCLUDE_FILES "sys/types.h" )
```
sanjayankur31: I'm wondering of this some comments should be added here just to make it easier for devs in the future to understand why this is here?

sanjayankur31:
Still getting build failures on the 2 32 bit architectures supported in Fedora: i686 and armv

7hl:

- i686: https://koji.fedoraproject.org/koji/taskinfo?taskID=30897603
- armv7hl: https://koji.fedoraproject.org/koji/taskinfo?taskID=30897605

I'm also uploading the build logs. These are "scratch" builds and will be deleted in few hours.

[armv7hl-build-log.txt](https://github.com/nest/nest-simulator/files/2585037/armv7hl-build-log.txt)
[i686-build-log.txt](https://github.com/nest/nest-simulator/files/2585038/i686-build-log.txt)

nestkernel/exceptions.h
```diff
@@ -939,6 +939,34 @@ class KeyError : public KernelException
   std::string message() const;
 };
 
+/**
+ * Exception to be thrown if an internal error occures.
+ * @ingroup KernelExceptions
+*/
+class InternalError : public KernelException
```
jakobj: why do we need this new exception? wouldn't a `KernelException` by sufficient?

sanjayankur31:
I'm afraid all the travis tests machines seem to be amd64 machines, so the tests don't really check these code changes. The code doesn't build in i686 and armv7hl for me (as commented above).

gtrensch:
@sanjayankur31 Ich checked the logs. The "constructor ambiguity" issue was one of the problems I also went into when testing on the 32-bit ARM Raspberry. It should be resolved. One of the warnings shows:
```
warning: unsigned conversion from 'long long int' to 'size_t' {aka 'unsigned int'} changes value from '4611686018427387903' to '4294967295' [-Woverflow]
   static const size_t disabled_marker_ = 4611686018427387904 - 1; // 2 ** 62 - 1
```
I removed that raw number from the code. Can you please check whether you ran your tests on the correct branch?

sanjayankur31:
@gtrensch ah, I think I may have run the build before those commits. I'll re-run with the latest one in your branch and report back.

sanjayankur31:
The latest builds for i686 and armvhl7 completed successfully:

- https://koji.fedoraproject.org/koji/taskinfo?taskID=30993779
- https://koji.fedoraproject.org/koji/taskinfo?taskID=30993781

heplesser:
@gtrensch Could you resolve the conflicts that have occurred (I assume #1062 may be at fault), then this one should be ready to be merged.

nestkernel/exceptions.h
gtrensch: Are not all exceptions derived from the base class `KernelException`? The enumeration that I introduced in `targets.h` as well as the `InternalError` exception should make the code more robust and easier to extend - just in case that there is more than a single status bit needed in future.

gtrensch:
The conflict is resolved.

cmake/CheckIncludesSymbols.cmake
gtrensch: This macro needs to be set up before `check_type_size()` and loses scope after execution of `check_type_size()`. This is CMake magic. I think it is clear what this part of the CMake script does. The rest is somehow CMake-technical  and I tend not to begin documenting it.

Issue #556, opened 2016-11-21
Define kernel states and transitions between them properly


heplesser:
As @apeyser pointed out during the review of #542, which addresses #311, kernel states are currently labeled by a collection of boolean flags with no formally defined transition rules.

The task is to defined a finite set of states, preferable as enumeration, and to precisely define the transition rules between those states.

See also #554 and #555.

PR #1069, opened 2018-11-08
Refactor registration of models in modelsmodule

clinssen:
This PR refactors the code that registers models into the kernel in `modelsmodule.h/.cpp`, into something that is functionally equivalent but eliminates about half of the lines of code that pertain to adding synapses. This is because each synapse type was explicitly added twice: each with its own index data type (`TargetIdentifierPtrRport` and `TargetIdentifierIndex`). I moved out all the `register_connection_model()` calls out to a templated function, and then call the function twice, once with each data type given above.

Also removed a large amount of source code comments that did not belong here.


models/modelsmodule.h
```diff
@@ -49,6 +51,10 @@ class ModelsModule : public SLIModule
 
   const std::string name( void ) const;
   const std::string commandstring( void ) const;
+
+protected:
+  template < typename ConnectionT >
+  void register_connection_models( std::string name_postfix = "" );
```
heplesser: Please add Doxygen documentation for the method. 
And is this method specific to modelsmodule? Wouldn't it be useful also for other modules defining synapses and should thus rather be provided by `ModelManager`?

models/modelsmodule.cpp
```diff
@@ -297,72 +297,6 @@ ModelsModule::init( SLIInterpreter* )
     "volume_transmitter" );
 
   // Create voltmeter as a multimeter pre-configured to record V_m.
-  /*BeginDocumentation
```
heplesser: Where does the `voltmeter` documentation go?

models/modelsmodule.cpp
```diff
@@ -432,48 +366,12 @@ ModelsModule::init( SLIInterpreter* )
     "music_message_in_proxy" );
 #endif
 
-  // register synapses
-
-  /* BeginDocumentation
-     Name: static_synapse_hpc - Variant of static_synapse with low memory
```
heplesser: Where does this documentation go?

models/modelsmodule.cpp
```diff
@@ -432,48 +366,12 @@ ModelsModule::init( SLIInterpreter* )
...
-      "static_synapse" );
-  kernel()
-    .model_manager
-    .register_connection_model< StaticConnection< TargetIdentifierIndex > >(
-      "static_synapse_hpc" );
-
+  // register all connection models: once for the normal indexing type and
+  // once for the more efficient "HPC" indexing type
+  register_connection_models< TargetIdentifierPtrRport >();
+  register_connection_models< TargetIdentifierIndex >( "_hpc" );
```
heplesser: This makes me almost wonder if we shouldn't place the suffix information in the `TargetIdentifier...` classes, to make sure that they are consistent throughout. With the current setup, users could create any kind of suffix-TI combinations.

clinssen:
The NEST Open Developer's Meeting of November 12th, 2018, decided that the "voltmeter" is a useful concept (a kind of "syntactic sugar") and will remain supported. Where the initialization and documentation (in the form of a source code comment block) should go, however, is not yet clear. Will await further comments from @jougs on this issue while I address the other comments made by @heplesser.

models/modelsmodule.h
```diff
@@ -49,6 +51,10 @@ class ModelsModule : public SLIModule
 
   const std::string name( void ) const;
   const std::string commandstring( void ) const;
+
+protected:
```
gtrensch: Shouldn't it be private? There is no inheritance, all happens in the ModelsModule class.

jougs:
@clinssen: what is the status of this? Thanks!

models/voltmeter.cpp
```diff
@@ -0,0 +1,60 @@
+/*
+ *  multimeter.cpp
```
heplesser: `voltmeter.cpp`

models/voltmeter.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+// Includes from nestkernel:
+#include "event_delivery_manager_impl.h"
+
+namespace nest
+{
+Voltmeter::Voltmeter()
+  : Multimeter()
+{
+    //record_from_.push_back( names::V_m );
```
heplesser: Remove commented code.

models/voltmeter.cpp
```diff
@@ -0,0 +1,60 @@
...
+  ( *vmdict )[ names::record_from ] = ad;
+  set_status( vmdict );
+}
+
+Voltmeter::Voltmeter( const Voltmeter& n )
+  : Multimeter( n )
+{
+  /*DictionaryDatum vmdict;
+  n.get_status(vmdict);
+  assert(vmdict[names::record_from] == names::V_m.toString());*/
```
heplesser: Remove commented code.

models/voltmeter.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+/*nest::Voltmeter::Parameters_::Parameters_( const Parameters_& p )
+  : interval_( p.interval_ )
+  , offset_( p.offset_ )
+  , record_from_( p.record_from_ )
+{
+  assert (p.record_from_.size() == 1 
+          && p.record_from_[0] == names::V_m );
+  interval_.calibrate();
+}*/
```
heplesser: Remove commented code.

models/voltmeter.h
```diff
@@ -0,0 +1,123 @@
+/*
+ *  multimeter.h
```
heplesser: `voltmeter.h`

models/modelsmodule.h
```diff
@@ -27,6 +27,8 @@
 #include <string>
 
 // Includes from sli:
+#include "connector_model.h"
+#include "genericmodel.h"
```
heplesser: From the changes in this file, I don't see why you would need to add these two includes here. Remove if not needed, or move to `modelsmodule.cpp` if needed there. They are also misplaced here under heading `Includes from sli`.

models/modelsmodule.h
heplesser: - As it is a private method, add `_` at the end of the method name.
- Concerning the docstring:
    - No need to write that it is for internal use, the fact it is `private` makes that clear.
    - Docstring should start with a single line short description
    - Add a Note to make clear that for any new connection model, registration must be added in this method.

models/modelsmodule.cpp
heplesser: If this is challenging to implement, I suggest to create a new issue to remind us to consider further refactoring later and for now use the code you propose here.

jougs:
@clinssen and I will work on this when I'm back in Jülich in approx. two weeks.

jougs:
As discussed offline, the `enum` type safety problems can be solved by following [this approach](http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/). @clinssen will add this as a general macro-based solution in `libnestutil`. Probably something like this can also be used by @heplesser to express the simulator states in #556. @gtrensch and @terhorstd are also interested in this.

ikitayama:
Hi,
If you guys can share the NEST3 branch that has the redactor codebase,
would you point me to it?

clinssen:
Dear @ikitayama, you may find yourself in the wrong thread. Could you open a separate issue (or email to nest_user) for this issue? Thanks!

clinssen:
Another potential implementation of bitfields (due to @terhorstd) for consideration as an alternative: https://gist.github.com/terhorstd/e882e6cef744f10b44f87473343b4b4f

ikitayama:
> Dear @ikitayama, you may find yourself in the wrong thread. Could you open a separate issue (or email to nest_user) for this issue? Thanks!

Well, I was actually interested in your refactoring work discussed in this PR #1069. If this isn't
merged to the mainline, in the meantime, I'll do testing on your branch. Thanks!

nestkernel/model_manager.h
```diff
@@ -179,26 +180,26 @@ class ModelManager : public ManagerInterface
...
-    const bool requires_clopath_archiving = false );
+  template < template < typename > class ConnectorModelT >
+  void register_connection_model(const std::string& name,
+    const bool register_hpc=true,
+    const bool register_lbl=true,
+    const bool is_primary=true,
+    const bool has_delay=true,
+    const bool supports_wfr=false,
+    const bool requires_symmetric=false,
+    const bool requires_clopath_archiving=false);
```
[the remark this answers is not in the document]
gtrensch: The point with Boolean arguments is that they are an indicator that a function does more than one thing, which should be avoided for a clean coding style.

clinssen:
@gtrensch: with gcc8 and other fixes having been merged into Travis CI (#1267, #1310), this PR is now ready for re-review. If you have a spare moment, would you have a look? I will be happy to walk through it together. The main change since last time is that a C++ unit test has been added for the bitfield enum class.

jougs:
@gtrensch, can you please have another look at this one? Thanks!

models/voltmeter.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "voltmeter.h"
+
+// Includes from nestkernel:
+#include "event_delivery_manager_impl.h"
+
+namespace nest
+{
+Voltmeter::Voltmeter()
+  : Multimeter()
```
jougs: Since the merge of #1345, both `Multimeter` and `Voltmeter` start with a lower case letter. Please adapt all occurrences accordingly.

nestkernel/model_manager.h
```diff
@@ -178,32 +179,24 @@ class ModelManager : public ManagerInterface
...
+   * different receptor types. Otherwise identical to non-hpc version.
+   *
    * @param name The name under which the ConnectorModel will be registered.
    */
-  template < typename ConnectionT, template < typename > class ConnectorModelT >
+  template < template < typename targetidentifierT > class ConnectionT >
   void register_connection_model( const std::string& name,
-    const bool requires_symmetric = false,
-    const bool requires_clopath_archiving = false );
+    const Register_Connection_Model_Flags flags = default_connection_model_flags );
```
jougs: Could `Register_Connection_Model_Flags` please be just CamelCase without the underbars?

models/modelsmodule.cpp
jougs: @heplesser: the documentation is now [here](https://github.com/nest/nest-simulator/pull/1069/files#diff-4d73c5bddaa99f5c5da31c447c96dc33R185).

models/modelsmodule.cpp
```diff
@@ -380,214 +293,39 @@ ModelsModule::init( SLIInterpreter* )
...
+  // register secondary connection models
+  register_secondary_connection_model< GapJunction >( "gap_junction",
+    Register_Connection_Model_Flags::REQUIRES_SYMMETRIC | Register_Connection_Model_Flags::SUPPORTS_WFR );
+
+  register_secondary_connection_model< RateConnectionInstantaneous >(
+    "rate_connection_instantaneous", Register_Connection_Model_Flags::SUPPORTS_WFR );
+  register_secondary_connection_model< RateConnectionDelayed >(
+    "rate_connection_delayed", Register_Connection_Model_Flags::HAS_DELAY );
+  register_secondary_connection_model< DiffusionConnection >(
+    "diffusion_connection", Register_Connection_Model_Flags::SUPPORTS_WFR );
```
jougs: :heart: :heart: :heart:

clinssen:
Thanks for the praise! The new commits should address all of your comments.

jougs:
@heplesser: could you please indicate if the recent changes address your concerns? Thanks!

models/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ set( models_sources
     mip_generator.h mip_generator.cpp
     modelsmodule.h modelsmodule.cpp
     multimeter.h multimeter.cpp
+    voltmeter.h voltmeter.cpp
```
heplesser: Please put this in proper alphabetical order.

libnestutil/enum_bitfield.h
```diff
@@ -0,0 +1,127 @@
...
+{
+
+/**
+ * These methods support type-safe bitfields using C++'s enum classes.
+ *
+ * Define the bitfield flags as follows:
+ *
+ * .. code-block:: C++
+ *
+ *    enum class My_Flags : unsigned
```
heplesser: I am not too fond of mixing camelcasing and underscores. How about `MyFlags`?

libnestutil/enum_bitfield.h
```diff
@@ -0,0 +1,127 @@
...
+ *        static const bool enable = true;
+ *    };
+ *
+ * Finally, instantiate a bitfield and do some operations on it, for example:
+ *
+ * .. code-block:: C++
+ *
+ *    My_Flags my_flags = My_Flags::FIRST_FLAG | My_Flags::FOURTH_FLAG;
+ *    my_flags |= My_Flags::THIRD_FLAG;
+ *    if (enumFlagSet(my_flags, My_Flags::FOURTH_FLAG))
```
heplesser: Spaces inside the `()`.

libnestutil/enum_bitfield.h
```diff
@@ -0,0 +1,127 @@
...
+ *        THIRD_FLAG = 1 << 2,
+ *        FOURTH_FLAG = 1 << 3
+ *    };
+ *
+ * Then, enable the bitfield operators for this enum class:
+ *
+ * .. code-block:: C++
+ *
+ *    template <>
+ *    struct EnableBitMaskOperators< My_Flags >
```
heplesser: Could you add some more information on what this does, why it is needed? Why can't I go straight from the first code block above to the last one below?

Is there any risk that use of `static` gobal variables could cause problem in threaded situations? I assume not, but would like it confirmed.

libnestutil/enum_bitfield.h
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * These methods support type-safe bitfields using C++'s enum classes.
+ *
+ * Define the bitfield flags as follows:
+ *
+ * .. code-block:: C++
+ *
+ *    enum class My_Flags : unsigned
+ *    {
+ *        FIRST_FLAG = 1 << 0,
```
heplesser: I could use any names for the flags here, correct?

models/modelsmodule.h
```diff
@@ -29,10 +29,17 @@
 // Includes from sli:
 #include "slimodule.h"
 
+#include "nest.h"
```
heplesser: Add  `// Includes from nestkernel:` on line above and move section before includes from SLI

nestkernel/model_manager_impl.h
```diff
@@ -79,54 +81,59 @@ ModelManager::register_preconf_node_model( const Name& name,
   return register_node_model_( model, private_model );
 }
 
-template < typename ConnectionT, template < typename > class ConnectorModelT >
+// we're templating this function with a class template, e.g.
+// BernoulliConnection< targetidentifierT >
```
heplesser: Could you move this information to the Doxygen comment for the method in the h-file?

nestkernel/nest.h
```diff
@@ -55,6 +56,45 @@ void reset_network();
 void enable_dryrun_mode( const index n_procs );
 
 void register_logger_client( const deliver_logging_event_ptr client_callback );
+
+enum class RegisterConnectionModelFlags : unsigned
```
heplesser: I would prefer `unsigned int` to just `unsigned`

libnestutil/enum_bitfield.h
jougs: Oh! I must have missed this. This definitely needs to go away!

libnestutil/enum_bitfield.h
clinssen: The magic happens entirely in the compiler! I'm not sure whether the EnableBitMaskOperators struct is ever even instantiated at runtime.

libnestutil/enum_bitfield.h
clinssen: Correct, we could alternatively do `FOO_FLAG`, `XYZZY_FLAG`, ...

clinssen:
Thanks for the feedback! The new commits should resolve all your remaining comments.

clinssen:
Whoops! I amended the last commit and did a force push, to keep the binary blob out of the repository's history.

Issue #415, opened 2016-06-30
CLI: Setting verbosity does not work


astoeckel:
According to `nest --help` the command line switch `--verbosity=QUIET` turns off all messages. However, executing

```
echo "(Hello World) =" | nest --verbosity=QUIET -
```

yields

```
[INFO] [2016.6.30 21:54:48 /home/andreas/source/nest-simulator/nestkernel/rng_manager.cpp:226 @ Network::create_rngs_] : Creating default RNGs
[INFO] [2016.6.30 21:54:48 /home/andreas/source/nest-simulator/nestkernel/rng_manager.cpp:221 @ Network::create_rngs_] : Deleting existing random number generators
[INFO] [2016.6.30 21:54:48 /home/andreas/source/nest-simulator/nestkernel/rng_manager.cpp:226 @ Network::create_rngs_] : Creating default RNGs
[INFO] [2016.6.30 21:54:48 /home/andreas/source/nest-simulator/nestkernel/rng_manager.cpp:272 @ Network::create_grng_] : Creating new default global RNG
Hello World
```

Note the log messages with severity "INFO".  The same happens when trying to set the verbosity to "FATAL". Tested with the current git version.

heplesser:
@astoeckel The messages above are issued when the `RNGManager` is first created during startup, before any command line arguments are read in `SLIStartup`. The code checking for verbosity settings in that constructor is rather weak, actually. So we probably need to re-organize handling of command-line arguments. This is not trivial, especially when interacting with PyNEST and/or Music.

astoeckel:
@heplesser This sounds like an understandable limitation. Wouldn't it be sufficient to just cache the log messages in the logger until a verbosity level has been set?

I stumbled across this while interfacing with NEST via SLI and stdin/stdout. While it surely isn't hard to ignore log messages in the output, it adds another parsing step. So alternatively, it would be sufficient for me if log messages were written to stderr, where they are put by most UNIX programs.

heplesser:
@jougs Could you take a look?

heplesser:
I am closing this in favour of #700, which came with an offer of a fix.

PR #1071, opened 2018-11-11
[WIP] Refactor testsuite setup

### LINKED ISSUE #345: do_tests should get all arguments from the commandline

### LINKED ISSUE #700: Suppress all NEST startup output

1. NEST's welcome message

### LINKED ISSUE #972: `make installcheck` reaches wrong testing code after sourcing `nest_vars.sh` for a different version

### LINKED ISSUE #1130: Phase 5 Tests all fail with ' Open MPI does not support recursive calls of mpirun'
I'm building `nest-stimulator` directly from a fork of the repo. things are largely working ok. I ran cmake with:

> cmake -DCMAKE_INSTALL_PREFIX:PATH=/home/eric/nest_dev /home/eric/nest_dev/nest-simulator -Dwith-python=3 -Dwith-mpi=ON

While `make` and `make install` ran ok,  phase 5 of the tests (for mpi) all failed with something like:

> **********************************************************
> Open MPI does not support recursive calls of mpirun
> **********************************************************
> Feb 22 10:01:50 unittest::assert_or_die [Fatal]: 
>     Assertion failed: unittest::distributed_collect_assert_or_die: collected 
>     results.
> Feb 22 10:01:50 unittest::assert_or_die [Fatal]: 
>     Exiting with code 1.
> Failed: missed SLI assertion
> Running  /opt/nest/bin/sli /opt/nest/share/doc/nest/topology/mpitests/topo_mpi_test_divergent.sli
> SLI v2.16.0 (C) 2004 The NEST Initiative
> N_MPI: 1

I've attached TEST-core.phase_5.xml (note changed extension to `txt` otherwise it wouldn't let me attach). [TEST-core.phase_5.txt](https://github.com/nest/nest-simulator/files/2894640/TEST-core.phase_5.txt)

I am in Ubuntu 16/Python 3.7 (Anaconda) and have Open MPI 1.10.2.

jougs:
This PR refactors the test runner scripts of NEST and PyNEST and makes them independent of configuration-time variables previously replaced into the scripts.

Changes in detail:
* `do_tests.sh` now gets all paths and executables on the commandline (fixes #345).
* `nest_vars.sh` does not export paths anymore, that could have confused NEST(fixes #972).
* Tests are now run in parallel with as many processors as are available (kudos to @terhorstd).
* Test results fro SLI tests are not written as JUnit/XML anymore.
* The info output from the `RandomManager` initialization is gone (fixes #700 and #415).
* The PyNEST testsuite was
   * moved outside of the `nest` module directory
   * converted to using `pytest` instead of `nosetests`
   * modified so that PyNEST tests do not include MPI anymore (fixes #1130).
* Increased legibility by adding an informative header and removing redundant paths from the testsuite printouts.


heplesser:
@jougs In the Python-free test configurations a `*.py` regression test fails.

jougs:
@stinebuu: can you please check if [what I did to your script from #1120](https://github.com/nest/nest-simulator/pull/1071/files#diff-e8570780ae94f31e498d275238e8083e) is good (or reasonable, at least :wink:)?

@heplesser, @gtrensch: I think I've sorted out all remaining issues with failing tests, formatting and such. As a bonus for the long wait, @terhorstd added full parallelization of the testsuite runs to this PR. Feel free to start reviewing at any time now.

pynest/nest/tests/test_sp/test_mpitests.py
```diff
@@ -19,45 +19,41 @@
...
-import unittest
-import nest
 import os
-from subprocess import call
+import subprocess
+import unittest
 
-HAVE_MPI = nest.ll_api.sli_func("statusdict/have_mpi ::")
+cmd = ["nest", "-c", "statusdict/have_mpi :: =only"]
+HAVE_MPI = subprocess.check_output(cmd) == "true"
```
stinebuu: @jougs I think `subprocess.check_output()` returns data as encoded bytes for Python 3 and as regular strings for Python 2. The test will only run if I use Python 2 and when I use Python 3 I get the message `skipped 'NEST was compiled without MPI'`, even though I have compiled with MPI. The same is true for `test_connect_all_patterns`. The two tests are also skipped during `make installcheck`.

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+print_paths () {
+    indent="`printf '%23s'`"
+    echo "$1" | sed "s/:/\n$indent/g" | sed '/^\s*$/d'
+}
+
+echo "================================================================================"
+echo
+echo "  NEST testsuite"
+echo "  Date: $(date -u)"
+echo "  Sysinfo: $(uname -s -r -m -o)"
```
heplesser: `uname`  on OSX does not support `-o`, I'd suggest to remove it.

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+export TEST_BASEDIR="${PREFIX}/share/doc/nest"
+
+# Gather some information about the host
+INFO_ARCH="$(uname -m)"
+INFO_HOME="$(/bin/sh -c 'echo ~')"
+INFO_HOST="$(hostname -f)"
+INFO_OS="$(uname -s)"
+INFO_USER="$(whoami)"
+INFO_VER="$(uname -r)"
+NPROCS="$(cat /proc/cpuinfo | grep processor | wc -l)"
```
heplesser: `/proc/cpuinfo` does not exist on OSX. Under OSX, `sysctl -n hw.physicalcpu` returns the number of processors. 
Will cpuinfo work on more exotic systems?

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+
+print_paths () {
+    indent="`printf '%23s'`"
+    echo "$1" | sed "s/:/\n$indent/g" | sed '/^\s*$/d'
```
heplesser: `sed` works differently on OSX since it is inherited from BSD, see eg https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed

One solution is to require the user to install GNU sed as `gsed` (e.g. via `brew install gnu-sed`) and defining the `sed` program to use as a variable depending on OS.

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+
+        echo "  C++ tests : ${CPP_TEST_TOTAL}"
+        echo "     Passed : ${CPP_TEST_PASSED}"
+        echo "     Failed : ${CPP_TEST_FAILED}"
+    else
+	echo "  Not running C++ tests because NEST was compiled without Boost."
+    fi
+}
+
+phase_one
```
heplesser: Should we add to the name what the tests are doing? E.g. `phase_one_selftests`?

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+phase_three() {
+    echo
+    echo "Phase 3: Integration  tests"
+    echo "---------------------------"
+
+    #for test_ext in sli py ; do
+    #    for test_dir in "unittests/" "topology/unittests/" ; do
+    #        for test_name in $(ls "${TEST_BASEDIR}/${test_dir}" | grep ".*\.${test_ext}\$") ; do
+    echo "TEST_BASEDIR $TEST_BASEDIR"
+    echo "PWD $(pwd)"
```
heplesser: Remove these two output lines and the commented-out code above.

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+
+echo
+echo "NEST Testsuite Summary"
+echo "----------------------"
+echo "  NEST Executable: $(which nest)"
+echo "  SLI Executable : $(which sli)"
+echo "  Total number of tests: $(( $(cat "${TEST_TOTAL}" | wc -l) + $CPP_TEST_TOTAL + $PYNEST_TEST_TOTAL))"
+echo "     Passed: $(( $(cat "${TEST_PASSED}" | wc -l) + $CPP_TEST_PASSED + $PYNEST_TEST_PASSED ))"
+echo "     Skipped: $(( $(cat "${TEST_SKIPPED}" | wc -l) + $CPP_TEST_SKIPPED + $PYNEST_TEST_SKIPPED )) ${PYNEST_TEST_SKIPPED_TEXT:-}"
+echo "     Failed: $(( $(cat "${TEST_FAILED}" | wc -l) + $CPP_TEST_FAILED + $PYNEST_TEST_FAILED )) ${PYNEST_TEST_FAILED_TEXT:-}"
```
heplesser: - Could one report total time used here?
- Could one provide info on which tests failed?

testsuite/run_test
```diff
@@ -0,0 +1,175 @@
...
+
+
+param_script="$1"
+param_success="$2"
+param_skipped="$3"
+param_failure="$4"
+
+msg_error=
+
+script_name="$( echo "${param_script}" | sed 's/^.*\/\([^/]\+\)$/\1/' )"
```
heplesser: Does not work on OSX, must use `gsed` or modify expressions.

testsuite/run_test
```diff
@@ -0,0 +1,175 @@
...
+#   Additionally this script needs TEST_PASSED, TEST_SKIPPED, TEST_FAILED, and
+#   TEST_TOTAL to be defined in the environment. Their values are treated as
+#   filenames where one line is added for the passed/skipped/failed test and
+#   one line in total.
+#
+# measure runtime of command
+#
+time_cmd()
+{
+    start=`date +%s%N`
```
heplesser: This will give the number of seconds with `N` appended on OSX and then arithmetic below works no more. `date +%s` gives just the seconds since the epoch, so by dropping the division below, one gets a coarse measure. Not sure how to get ns-resolution time on OSX.

nestkernel/rng_manager.cpp
```diff
@@ -235,8 +235,6 @@ nest::RNGManager::create_rngs_()
     rng_.clear();
   }
 
-  LOG( M_INFO, "Network::create_rngs_", "Creating default RNGs" );
-
```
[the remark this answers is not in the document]
jougs: See below.

nestkernel/rng_manager.cpp
```diff
@@ -280,8 +278,6 @@ nest::RNGManager::create_rngs_()
 void
 nest::RNGManager::create_grng_()
 {
-  // create new grng
-  LOG( M_INFO, "Network::create_grng_", "Creating new default global RNG" );
```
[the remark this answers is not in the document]
jougs: The output in this line and the one mentioned above make `regressiontests/issue-779-1016.py` fail on my machine (the test captures output and compares it to some reference).

As the error only started since the changes in this PR I thought that problem might also affect other users and removed them (they were not that helpful, were they?).

pynest/do_tests.py
```diff
@@ -66,7 +69,10 @@ def _setup(arg):
 result_topo = runner.run(nest.topology.tests.suite())
 
 total = result.testsRun + result_topo.testsRun
-failed = len(result.failures) + len(result.errors) + len(result_topo.failures) + len(result_topo.errors)
+failed = len(result.failures) + \
```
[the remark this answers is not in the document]
jougs: Resolved by in c0935ab5c

testsuite/do_tests.sh
jougs: I think that we should actually have some kind of associative array mapping phase names to phase functions and then automatic command line arguments along the lines of `--list-phases`and `--skip-phases`. Ideally the phase functions would be in the corresponding directories, accompanied by a metadata file. I would, however, find this to be out of the scope of this PR.

@terhorstd: what is your take on this?

jougs:
@terhorstd: Can you please look into commented-out-code and portability-issues of the code you contributed? Also the one-report-per-test seems to come from your parallelization work.

Here are some additional hints:
* [SO#45181115](https://stackoverflow.com/questions/45181115) has a portable solution for getting the number of processors
* `do_tests.sh` contained a `portable_inplace_sed` function before your commit  f99d345d019dc7f815b1da3b378609dffedf1a71 maybe that helps in solving the portability problems.

testsuite/do_tests.sh
heplesser: @jougs I like your idea! It would be useful to be able to selectively run only certain groups of tests.

jougs: @heplesser: thanks for liking the idea. Should this also go into this PR or should it be its own?

nestkernel/rng_manager.cpp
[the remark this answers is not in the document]
heplesser: @jougs This is a bit strange. The test ([see here](https://github.com/nest/nest-simulator/blob/2c9d73f2306d391902719c8bc88c851811869b7f/testsuite/regressiontests/issue-779-1016.py#L49)) should print the argv from statusdict after the kernel has completely started, so the CreateRNG messages should come first. Why do they come later on your machine? Interestingly, they cannot even be suppressed using `--verbosity=FATAL`, since they apparently are executed before verbosity has been handled.

But maybe these lines here should go, after all, `Create()` doesn't print a message "new neurons created" either.

testsuite/do_tests.sh
heplesser: @jougs There is probably still enough to do in this PR, so I wouldn't want to enlarge. Create a separate one for the named and flexible phases.

nestkernel/rng_manager.cpp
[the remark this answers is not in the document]
jougs: Argh, you are right!

I was pointing you to the wrong test and even got the context all wrong. I meant [test_connect_all_patterns.py](https://github.com/nest/nest-simulator/pull/1071/files#diff-5e7934b0517f89b433502dd81359c5f6R27) and [test_mpitests.py](https://github.com/nest/nest-simulator/pull/1071/files#diff-e8570780ae94f31e498d275238e8083eR27). I the linked lines I need to get the output of NEST used in pure command-mode. And as you correctly observed, these messages cannot be silenced at all, because they are printed from some constructor long before the verbosity setting is actually taken into account.

See also #415 and #700 (which I've now added to the description of this one).

jougs:
@stinebuu: Thanks for catching the Python 3 issue and that the test was not actually run and thus could not fail. I've fixed this in b1bcbf7. Can you please re-evaluate?

jougs:
I'm afraid that I have to be more invasive to complete this. I've added the `[WIP]` label to the title of the PR in order to make people aware that this is still worked on. @gtrensch, @stinebuu, @heplesser: I'll notify you once I think the code is ready for re-review. Sorry for the noise.

jougs:
Getting there! Please note that although Travis is convinced that everything is good now, it isn't. The Pytest MPI tests are still failing and I have messed up error detection in the runner script. I hope to get this fixed tomorrow. Stay tuned!

jougs:
@stinebuu, @heplesser: As this got way bigger than expected, I now think that it might be a much better idea to merge this into `nest-3` instead of master. Do you have any objections against that? If not, I will close this PR and create a new one once I'm done with the rebasing. Thanks!

heplesser:
@jougs I am wary of changing the test setup at the same time as we make a massive change to NEST. How about first merging `nest-3` into master and then add this to master 2-3 weeks later, but definitely before releasing NEST3?

BTW, there is one feature I have been longing for for a long time: A concise overview over the tests (SLI or Py) that failed at the end, instead of the current "1 out of 900 test failed. Enjoy scrolling back to find which one!" situation. Any hope for such luxury?

jougs:
Alright. Back in business on this one :-)

As I have decided to rebase the whole thing onto nest-3 to make everything easier down the line, I will now close this PR and open a new one against that branch. Stay tuned!

jougs:
This PR was originally meant to refactor the test runner scripts of NEST and PyNEST to make them independent of configuration-time variables previously replaced into the scripts. Unfortunately, it suffered (and eventually died) from a severe case of _feature creep_. Mostly for the record, the following summary re-iterates the detailed list of changes in this PR, with annotations that describe the current state of affairs:

* `do_tests.sh` now gets all paths and executables on the commandline (fixes #345).
   * **fixed by replacement #1686**
* `nest_vars.sh` does not export paths anymore, that could have confused NEST(fixes #972).
   * **fixed by replacements #1561 and  #1578**
* Tests are now run in parallel with as many processors as are available (kudos to @terhorstd).
* Test results fro SLI tests are not written as JUnit/XML anymore.
   * **JUnit/XML is used to gather information from tests since #1587, rendering this change inappropriate** 
* The info output from the `RandomManager` initialization is gone (fixes #700 and #415).
   * **fixed by replacement #1577**
* The PyNEST testsuite was
   * moved outside of the `nest` module directory
   * converted to using `pytest` instead of `nosetests`
   * modified so that PyNEST tests do not include MPI anymore (fixes #1130).
* Increased legibility by adding an informative header and removing redundant paths from the testsuite printouts.
   * **fixed by replacement #1686**

jougs:
for the record: The only un-merged thing from this PR is the parallelization of test runs with as many processors as are available.

testsuite/run_test
```diff
@@ -0,0 +1,175 @@
+#!/bin/sh
+
+# from http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -eu
+#set -o pipefail
```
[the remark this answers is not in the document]
terhorstd: The "from" line is a proper comment that describes the following line. I strongly vote for leaving that in.
Actually the line should be a single enabled line:
```suggestion
set -euo pipefail
```
Some constructs may however be sensitive to `pipefail`, so this needs an explicit look at the code to enable without surprises.

testsuite/do_tests.sh
```diff
@@ -0,0 +1,684 @@
...
+# The test suite consists of SLI and Python scripts that use the
+# language's native `unittest` library to assert certain invariants
+# and thus ensure a correctly working installation of NEST.
+#
+# For commandline options, see the the function usage() below.
+#
+
+# from http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -eu
+#set -o pipefail
```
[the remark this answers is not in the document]
terhorstd: same as above.

testsuite/do_tests.sh
terhorstd: Would `-a` work on OSX?
```suggestion
echo "  Sysinfo: $(uname -a)"
```
I'm not sure if there was a reason against that…

testsuite/do_tests.sh
terhorstd: I think the standard coreutils command for this is
```suggestion
NPROCS="$(nproc)"
```
Are coreutils available in OSX?

PR #1088, opened 2018-12-11
Allow compile-time configuration of bit-field sizes in Target class

### LINKED ISSUE #1043: NEST is running out of space for synapse models

jakobj:
This PR implements a flexible choice for the size of various bitfields used in the `Target` class. In particular the new default setting allows for more synapse types. In addition it makes the sizes of bitfields used in various objects consistent with this global choice.
I suggest @jougs @Silmathoron @heplesser as reviewers.

fixes #1043


jakobj:
the test that currently fails (ticket-573.sli) checks for an error message upon exceeding the maximal number of synapses via `CopyModel`. It uses a hardcoded number for this. How should we proceed here? Let this test be generated by CMAKE with the correct max number?

Silmathoron:
This seems to be the a reasonable solution (to be honest I don't really see another).

jougs:
In the light of #1071, which makes the testsuite more independent of the NEST version it came from, I don't like the idea of creating a test that is tied that tight to the NEST version it came from.

What about making `MAX_SYN_ID` available in the kernel status dictionary and querying the value from within the test script?

Silmathoron:
Ah, I did not see this PR... then this is indeed a problem. I guess that if we make the name quite explicit, like ``"max_synaptic_models"``, it should not be a problem if it shows up in the status dict...

jakobj:
ok, thanks, i've implemented your suggestion

jakobj:
Talking about this a bit more with @suku248 we have the impression that providing the possibility to customize the bit widths of the `Target` fields does not make much sense: these sizes interact with the sizes of other variables in various places (e.g., https://github.com/nest/nest-simulator/blob/master/nestkernel/syn_id_delay.h#L36, https://github.com/nest/nest-simulator/blob/master/nestkernel/spike_data.h#L52) and getting all of these consistent is not trivial. Finding a working custom split is hence quite involved and to reduce frustration for the user we should maybe only provide splits that actually work.

Related: should we move all bit width definitions that now are spread over various files to a central location, e.g., `nest_types.h`?

heplesser:
@jakobj All in favor of collecting the widths in one place! And only providing settings that work also makes sense.

jakobj:
sorry for taking so long, i've just pushed an updated version where i've moved all size definitions to `nest_types.h` and removed the option of providing a custom split.

jakobj:
curently the `sli_neuron` test breaks as `Name::insert` throws an exception after failing an `assert( not omp_in_parallel() )`. this is called from `ModelManager::get_status` after calling `calibrate` of the sli neuron (`GetResolution` triggers a cascade of `get_status` calls, see `test_sli_neuron.sli:101`). this is strange since the sli call in `sli_neuron::calibrate` is protected by an `omp critical`. @heplesser any idea what is going wrong here?

hakonsbm:
@jakobj `omp_in_parallel()` returns `true` when called from within a region executing in parallel, including nested regions that are serialized (e.g. `critical`, `single` or `master` regions). So to avoid the assert failing, it has to be outside the outermost parallel region. However, I suggest adding `max_num_syn_models` to nest_names, avoiding the assert altogether.

jougs:
@jakobj: ping!

jakobj:
thanks for the hint @hakonsbm, wasn't sure about `omp_in_parallel` from within an `omp_critical`. seems like it returns true. fixed it now by following your suggestions. please have another look.

jakobj:
looks like everything is in order @hakonsbm :)

nestkernel/nest_types.h
```diff
@@ -89,10 +126,8 @@ const index invalid_index = SIZE_MAX;
 /**
  *  Unsigned char type for enumerations of synapse types.
  */
-typedef unsigned char synindex;
-// const synindex invalid_synindex = UCHAR_MAX;
-const synindex invalid_synindex =
-  63; // number of synapse types limited by size of syn_id in target.h
+typedef unsigned int synindex;
```
hakonsbm: Change the documentation comment above to match the changed type.

nestkernel/source.h
```diff
@@ -40,12 +40,10 @@ namespace nest
 class Source
 {
 private:
-  uint64_t gid_ : 62;  //!< gid of source
-  bool processed_ : 1; //!< whether this target has already been moved
-                       //!to the MPI buffer
+  uint64_t gid_ : NUM_BITS_GID;  //!< gid of source
+  bool processed_ : 1;           //!< whether this target has already been moved
+                                 //!to the MPI buffer
```
hakonsbm: I think this line should also start with `//!<` for doxygen to parse it correctly.

nestkernel/spike_data.h
```diff
@@ -33,6 +33,14 @@
 namespace nest
 {
 
+enum enum_status_spike_data_id
```
hakonsbm: Is this used anywhere?

nestkernel/spike_data.h
```diff
@@ -42,16 +50,16 @@ namespace nest
...
-  const static unsigned int end_marker_ = 1;
-  const static unsigned int complete_marker_ = 2;
-  const static unsigned int invalid_marker_ = 3;
 
-  index lcid_ : 27;         //!< local connection index
-  unsigned int marker_ : 2; //!< status flag
-  unsigned int lag_ : 14;   //!< lag in this min-delay interval
-  thread tid_ : 10;         //!< thread index
-  synindex syn_id_ : 8;     //!< synapse-type index
+  static constexpr int MAX_MARKER =
```
hakonsbm: Is this used anywhere?

nestkernel/spike_data.h
```diff
@@ -128,9 +136,13 @@ class SpikeData
   double get_offset() const;
 };
 
+//!< check legal size
```
hakonsbm: Doxygen documentation placed before the thing they document should start with `//!` only.

nestkernel/spike_data.h
```diff
@@ -128,9 +136,13 @@ class SpikeData
...
+typedef StaticAssert< sizeof( SpikeData ) == 8 >::success
+  success_spike_data_size;
```
hakonsbm: Since we use C++11 now, maybe use type alias instead? So
```c++
typedef StaticAssert< sizeof( SpikeData ) == 8 >::success success_spike_data_size;
```
would become
```c++
using success_spike_data_size = StaticAssert< sizeof( SpikeData ) == 8 >::success;
```

nestkernel/spike_data.h
```diff
@@ -269,6 +281,10 @@ class OffGridSpikeData : public SpikeData
   double get_offset() const;
 };
 
+//!< check legal size
```
hakonsbm: Doxygen documentation placed before the thing they document should start with `//!` only.

nestkernel/spike_data.h
```diff
@@ -269,6 +281,10 @@ class OffGridSpikeData : public SpikeData
...
+typedef StaticAssert< sizeof( OffGridSpikeData ) == 16 >::success
+  success_offgrid_spike_data_size;
```
hakonsbm: Maybe use C++11 type alias instead?

nestkernel/syn_id_delay.h
```diff
@@ -105,6 +105,11 @@ struct SynIdDelay
     return disabled;
   }
 };
+
+//!< check legal size
```
hakonsbm: Doxygen documentation placed before the thing they document should start with `//!` only.

nestkernel/syn_id_delay.h
```diff
@@ -105,6 +105,11 @@ struct SynIdDelay
...
+typedef StaticAssert< sizeof( SynIdDelay ) == 4 >::success
+  success_syn_id_delay_data_size;
```
hakonsbm: Maybe use C++11 type alias instead?

nestkernel/target.h
```diff
@@ -39,34 +39,34 @@ namespace nest
...
+ *  |   +---- synapse-type id (syn_id)
+ *  |   |
+ *  ||-----||---thread----||---------rank------------||----local connection id (lcid)----|
+ *  0000 0000  0000 0000  0000 0000  0000 0000  0000 0000  0000 0000  0000 0000  0000 0000
+ *  |       |  |       |  |       |  |       |  |       |  |       |  |       |  |       |
+ *  63      56 55      48 47      40 39      32 31      24 23      16 15      8  7       0
+ *
+ * Other custom layouts can be chosen by providing a list of 5
+ * numbers, representing the bits required for rank, thread, synapse
+ * id, local connection id and processed flag, repectively. The number
```
hakonsbm: "repectively" to "respectively"

nestkernel/target.h
```diff
@@ -39,34 +39,34 @@ namespace nest
...
+ * of bit needs to sum to 64. The processed flag must always use one
```
hakonsbm: Should it be "bits"?

nestkernel/target.h
```diff
@@ -78,19 +78,17 @@ class Target
...
 
   static constexpr uint8_t BITPOS_LCID = 0U;
   static constexpr uint8_t BITPOS_RANK = NUM_BITS_LCID;
   static constexpr uint8_t BITPOS_TID = BITPOS_RANK + NUM_BITS_RANK;
   static constexpr uint8_t BITPOS_SYN_ID = BITPOS_TID + NUM_BITS_TID;
   static constexpr uint8_t BITPOS_PROCESSED_FLAG =
     BITPOS_SYN_ID + NUM_BITS_SYN_ID;
+
+  typedef StaticAssert< NUM_BITS_PROCESSED_FLAG == 1U >::success
+    bits_for_processed_flag;
```
hakonsbm: Maybe use C++11 type alias instead?

nestkernel/target.h
```diff
@@ -180,6 +173,9 @@ class Target
   double get_offset() const;
 };
 
+//!< check legal size
```
hakonsbm: Start with `//!` only.

nestkernel/target.h
```diff
@@ -180,6 +173,9 @@ class Target
...
+typedef StaticAssert< sizeof( Target ) == 8 >::success success_target_size;
```
hakonsbm: Maybe use C++11 type alias instead?

nestkernel/target_data.h
```diff
@@ -74,6 +76,10 @@ class TargetDataFields
   synindex get_syn_id() const;
 };
 
+//!< check legal size
+typedef StaticAssert< sizeof( TargetDataFields ) == 8 >::success
+  success_target_data_fields_size;
```
hakonsbm: Fix doxygen documentation and maybe use C++11 type alias instead?

nestkernel/target_data.h
```diff
@@ -124,6 +130,10 @@ class SecondaryTargetDataFields
   synindex get_syn_id() const;
 };
 
+//!< check legal size
+typedef StaticAssert< sizeof( SecondaryTargetDataFields ) == 8 >::success
+  success_secondary_target_data_fields_size;
```
hakonsbm: Fix doxygen documentation and maybe use C++11 type alias instead?

nestkernel/target_data.h
```diff
@@ -150,6 +160,14 @@ SecondaryTargetDataFields::get_syn_id() const
   return syn_id_;
 }
 
+enum enum_status_target_data_id
```
hakonsbm: Is this used anywhere?

nestkernel/target_data.h
```diff
@@ -163,16 +181,21 @@ class TargetData
...
+  static constexpr uint8_t NUM_BITS_IS_PRIMARY = 1U;
+
+  static constexpr int MAX_LID = generate_max_value( NUM_BITS_LID );
+  static constexpr int MAX_MARKER = generate_max_value( NUM_BITS_MARKER );
+
+  unsigned int source_lid_ : NUM_BITS_LID; //!< local id of presynaptic neuron
+  //!< thread index of presynaptic neuron
+  unsigned int source_tid_:  NUM_BITS_TID;
+  unsigned int marker_ : NUM_BITS_MARKER;
+   //!< TargetData has TargetDataFields else SecondaryTargetDataFields
```
hakonsbm: Doxygen documentation with `//!<` points to the variable that precedes it, so these should be `//!` only.

nestkernel/target_data.h
```diff
@@ -163,16 +181,21 @@ class TargetData
...
-  unsigned int marker_ : 2;
-  bool is_primary_ : 1; //!< TargetData has TargetDataFields
-                        //!< else has SecondaryTargetDataFields
+  static constexpr uint8_t NUM_BITS_LID = 19U;
+  // NUM_BITS_TID set via cmake
+  static constexpr uint8_t NUM_BITS_MARKER = 2U;
+  static constexpr uint8_t NUM_BITS_IS_PRIMARY = 1U;
+
+  static constexpr int MAX_LID = generate_max_value( NUM_BITS_LID );
+  static constexpr int MAX_MARKER = generate_max_value( NUM_BITS_MARKER );
```
hakonsbm: Is `MAX_MARKER` used anywhere?

nestkernel/spike_data.h
jakobj: yes, e.g., https://github.com/nest/nest-simulator/blob/a406662974b459d875aedf8fdea749b367ce09ae/nestkernel/spike_data.h#L145

nestkernel/spike_data.h
jakobj: no, removed

nestkernel/spike_data.h
jakobj: is this best practice now?

nestkernel/target_data.h
jakobj: yes, e.g., l229

nestkernel/target_data.h
jakobj: no, removed

jakobj:
thanks for the review, I've implemented all you suggestions except the C++11 alias since I'm not familiar with it. if this is best practice i can implement it of course, just let me know. regarding formatting, i'm fighting with the correct local setup.

hakonsbm:
@jakobj Great, thanks for the changes! I think C++11 type alias is easier to read than `typedef`.

jakobj:
ok, done. still didn't manage to get clang-format-3.6 to work, do you have a working version @hakonsbm ?

hakonsbm:
@jakobj Yes, I'll fix the formatting.

Strangely, when trying to get the changed files Travis fails with 
```
jq: error (at <stdin>:4): Cannot index string with string "filename"
```
and reports no formatting errors. I've seen it with other PRs too, which has led to formatting errors sneaking into master.

CMakeLists.txt
```diff
@@ -80,6 +80,8 @@ option( k-computer "Enable K computer." OFF )
 set( disable-timing OFF CACHE BOOL "Disable measurements via Stopwatches. [default OFF]" )
 set( disable-counts ON CACHE BOOL "Disable measurements of call counts. [default ON]" )
 
+set( target-bits-split "standard" CACHE STRING "Split of the 64bits of the Target class. 'standard' is recommended for most users. If running on more than 262144 MPI processes or more than 512 threads, change to 'hpc'. [default standard]" )
```
gtrensch: The description can perhaps be more specific: instead of "64bits of the Target class" I would suggest "64-bit target neuron identifier type".

cmake/ProcessOptions.cmake
```diff
@@ -541,6 +541,31 @@ function( NEST_PROCESS_WITH_BOOST )
...
+    else()
+      message( FATAL_ERROR "Invalid target-bits-split selected." )
+    endif()
+    set( NUM_BITS_RANK ${NUM_BITS_RANK}U PARENT_SCOPE )
+    set( NUM_BITS_TID ${NUM_BITS_TID}U PARENT_SCOPE )
+    set( NUM_BITS_SYN_ID ${NUM_BITS_SYN_ID}U PARENT_SCOPE )
+    set( NUM_BITS_LCID ${NUM_BITS_LCID}U PARENT_SCOPE )
+    set( NUM_BITS_PROCESSED_FLAG ${NUM_BITS_PROCESSED_FLAG}U PARENT_SCOPE )
+  endif()
+endfunction()
```
gtrensch: Exposing this in CMake creates an unnecessary coupling of the NEST code with CMake. For the sake of clarity, the better place for the constants and toggling between them ("standard" and "hpc") is, for example, nest_types.h.

nestkernel/source.h
```diff
@@ -144,13 +142,13 @@ Source::is_primary() const
...
 {
-  gid_ = GID_DISABLED;
+  gid_ = MAX_GID;
 }
 
 inline bool
 Source::is_disabled() const
 {
-  return gid_ == GID_DISABLED;
+  return gid_ == MAX_GID;
```
gtrensch: This is counterintuitive. MAX_GID (2^62 -1 ) is not a valid GID, correct? If "high-value" fields are indicate something, for clarity, this should be reflected in the source code.

gtrensch:
@jakobj thanks for your code changes, I very much like that raw numbers are removed from the code. I have some minor suggestions, concerning the coupling with CMake and the readability of the code. Please see my comments?

nestkernel/source.h
jakobj: i am not sure i understand. do you mean we should do something `constexpr uint64_t DISABLED = MAX_GID;` and then use that to indicated disabled sources?

jakobj:
thanks @gtrensch for the helpful comments. i've addressed all except one where i'd need some input. please have another look.

nestkernel/nest_types.h
```diff
@@ -59,6 +59,50 @@ namespace nest
...
+constexpr uint8_t NUM_BITS_MARKER_SPIKE_DATA = 2U;
+constexpr uint8_t NUM_BITS_LAG = 14U;
+constexpr uint8_t NUM_BITS_DELAY = 21U;
+constexpr uint8_t NUM_BITS_GID = 62U;
+
+/*
+ * Maximally allowed values for bitfields
+ */
+constexpr uint64_t MAX_LCID = generate_max_value( NUM_BITS_LCID );
+constexpr int MAX_RANK = generate_max_value( NUM_BITS_RANK );
```
gtrensch:
```suggestion
constexpr uint64_t MAX_RANK = generate_max_value( NUM_BITS_RANK );
```

nestkernel/nest_types.h
```diff
@@ -59,6 +59,50 @@ namespace nest
...
+constexpr int MAX_TID = generate_max_value( NUM_BITS_TID );
```
gtrensch:
```suggestion
constexpr uint64_t MAX_TID = generate_max_value( NUM_BITS_TID );
```

nestkernel/nest_types.h
```diff
@@ -59,6 +59,50 @@ namespace nest
...
+constexpr int MAX_SYN_ID = generate_max_value( NUM_BITS_SYN_ID );
```
gtrensch:
```suggestion
constexpr uint64_t MAX_SYN_ID = generate_max_value( NUM_BITS_SYN_ID );
```

nestkernel/source.h
gtrensch: @jakobj The name of the constant 'MAX_GID' expresses that it represents the highest valid GID, but we test for `gid < MAX_GID`. We hide information here and make the code less readable. Better would be to have two constants.
`GID_DISABLED` and `MAX_GID = GID_DISABLED - 1`. Then we can test for `gid <= MAX_GID` which is what I would expect.

gtrensch:
@jakobj Many thanks! Some minor suggestions: please can you have a look at my comments and suggested code changes.

nestkernel/nest_types.h
gtrensch: @jakobj The constexpr function now returns an uint64_t. Since MAX_nnn is only used in the assertions uint64_t is the better choice.

jakobj:
thanks for the comments @gtrensch, i've now incorporated your changes. unfortunately `MAX_RANK` and `MAX_TID` both need be signed because at least for the `thread` type `-1` is a magic number and hence the type `thread` needs to be signed. variables of this type will be compared to `MAX_TID` which hence needs to be signed as well. I'll look into making both unsigned in a consistent fashion, but this should be a separate PR.

PR #1093, opened 2018-12-13
Make GetConnection filter by synapse label when only target attribute is given

### LINKED ISSUE #1085: `GetConnections` does not seem to filter by synapse label when only the `target` attribute is given.
Description of problem: `GetConnections` does not seem to filter by synapse label when only the `target` attribute is given.

Version-Release or Git commit: 2.16.0. This works with 2.14.0

How reproducible: Always

Steps to Reproduce:
```python
import nest

a = nest.Create("iaf_cond_exp", n=5)

nest.Connect(a, a, 'all_to_all', {"model": "static_synapse_lbl", "synapse_label": 123})
nest.Connect(a, a, 'all_to_all', {"model": "static_synapse_lbl", "synapse_label": 456})

print("a", len(nest.GetConnections(source=a, target=a, synapse_label=123)))
print("b", len(nest.GetConnections(target=a, synapse_label=123)))
print("c", len(nest.GetConnections(source=a, synapse_label=123)))
```

Actual results:
```
a 25
b 50
c 25
```

Expected results:
```
a 25
b 25
c 25
```

stinebuu:
This PR fixes #1085.

I have added a check for synapse label in `ConnectionManager::get_connections` for the case when `GetConnection` is called with only the `target` attribute.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1232,7 +1232,8 @@ nest::ConnectionManager::get_connections(
           for ( size_t i = 0; i < source_lcids.size(); ++i )
           {
             if ( synapse_label == UNLABELED_CONNECTION
-                 or connections->get_syn_label( source_lcids[ i ] ) == synapse_label )
+              or connections->get_syn_label( source_lcids[ i ] )
+                == synapse_label )
             {
               conns_in_thread.push_back( ConnectionDatum( ConnectionID(
                 source_table_.get_gid( tid, syn_id, source_lcids[ i ] ),
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1231,13 +1231,18 @@ nest::ConnectionManager::get_connections(
 
           for ( size_t i = 0; i < source_lcids.size(); ++i )
           {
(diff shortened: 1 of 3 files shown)


heplesser:
@apdavison Could you review this as it should fix your #1085?

nestkernel/connection_manager.cpp
```diff
@@ -1231,13 +1231,19 @@ nest::ConnectionManager::get_connections(
...
-              source_table_.get_gid( tid, syn_id, source_lcids[ i ] ),
-              *t_gid,
-              tid,
-              syn_id,
-              source_lcids[ i ] ) ) );
+            if ( synapse_label == UNLABELED_CONNECTION
+              or connections->get_syn_label( source_lcids[ i ] )
+                == synapse_label )
+            {
+              conns_in_thread.push_back( ConnectionDatum( ConnectionID(
```
jakobj: why don't we use `ConnectorBase::get_connection` here? this would also take care of filtering by label and would make this if branch consistent with the others. in addition we wouldn't need to introduce the helper functions below. is speed an issue here?

stinebuu:
I am closing this in favor of #1099, which, in addition to fixing #1085, also improves performance.

PR #1098, opened 2019-01-05
Add new neuron model hh_cond_beta_gap_traub

dnao:
I have created a new neuron model by modifying the hh_psc_alpha_gap and hh_cond_exp_traub models. It uses beta-function synapses and gap junction coupling. Though I have attempted to follow the existing models closely, feedback would be appreciated as there may be some interactions between the dynamics and synaptic/gap junction conductance which I have not considered.


models/iaf_cond_beta.cpp
```diff
@@ -373,7 +373,7 @@ nest::iaf_cond_beta::get_normalisation_factor( double tau_rise,
   double tau_decay )
 {
   // Factor used to normalise the synaptic conductance such that
-  //
+  // incoming spike causes a conductance of amplitude 1 nS.
```
janhahne: This seems unrelated to this PR, but its probably fine to fix it here

dnao:
Done, though I haven't had much luck using clang-format with the config file in the nest directory. I find that the CI requires some additional formatting different to what the script produces. I'm using clang 7.0 so perhaps I should switch to 3.6? Would it be a bad idea to squash and rebase all these commits to make this PR less cluttered?

janhahne:
@dnao Thank you! With clang 3.6 the automatic formatting works (at least for me) without any need for further adjustments, so it is probably a good idea to use version 3.6. Regarding your second question: It is not at all a bad idea, it is very welcome (however not mandatory).

models/hh_cond_beta_gap_traub.h
```diff
@@ -0,0 +1,521 @@
...
+  port handles_test_event( GapJunctionEvent&, rport );
+
+  void
+  sends_secondary_event( GapJunctionEvent& )
+  {
+  }
+
+  /**
+   * Return membrane potential at time t.
+potentials_.connect_logging_device();
```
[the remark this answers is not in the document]
dnao: Hi @janhahne could you clarify which lines should be removed? Do you mean lines 194-197?

dnao:
@janhahne I have added the unit tests you suggested and also fixed the normalisation so incoming spikes cause a peak conductance of 1nS as with other models using beta-function synapses. I will need to fix the same normalisation issue in the iaf_cond_beta model but perhaps it is better if I make a new PR for that?

models/hh_cond_beta_gap_traub.h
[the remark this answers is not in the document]
janhahne: @dnao Originally I only meant line 197, but after a short investigation I think that the functions `set_potential` and `get_potential` can be removed completely (see also #1121).

janhahne:
@dnao Thanks, the test looks fine! I agree that it would be better to create a new PR for the fix of `models/iaf_cond_beta`. Maybe just undo the (outdated) changes to `models/iaf_cond_beta.cpp` here and apply the correct normalisation in a new PR. 

The indent in `models/CMakeLists.txt` is still wrong, please have another look there. Also see my inline comment about `set_potential` and `get_potential`. Otherwise I am happy with this PR!

@heplesser Did you meant to approve this PR when asking for the NEST Contributor Agreement or do you want to have another look?

dnao:
Thanks @janhahne . I've removed the functions `set_potential` and `get_potential`. If these changes are OK now then I will squash them into a single commit.

jougs:
@janhahne, @heplesser: can this be merged now? Thanks!

models/hh_cond_beta_gap_traub.cpp
```diff
@@ -0,0 +1,787 @@
...
+  // normalization factor g0
+  // Both denom1 and denom2 are null if tau_decay = tau_rise, but they
+  // can also be null if tau_decay and tau_rise are not equal but very
+  // close to each other, due to the numerical precision limits.
+  // In such case the beta function reduces to the alpha function,
+  // and the normalization factor for the alpha function should be used.
+  double denom1 = tau_decay - tau_rise;
+  double denom2 = 0;
+  double normalisation_factor = 0;
+  if ( denom1 != 0 )
```
clinssen: I suggest replacing the condition with `|denom1| < std::numeric_limits< double >::epsilon()`

Also: why is denom2 calculated inside the first if block? Can we use the return-early pattern here?

models/hh_cond_beta_gap_traub.cpp
```diff
@@ -0,0 +1,787 @@
...
+      // sending spikes: crossing 0 mV, pseudo-refractoriness and local
+      // maximum...
+      // refractory?
+      if ( S_.r_ > 0 )
+      {
+        --S_.r_;
+      }
+      else
+        // (    threshold    &&     maximum       )
+        if ( S_.y_[ State_::V_M ] >= P_.V_T + 30.
```
clinssen: Where does the value 30 come from? Can we give this a name and move it to a private const member variable?

heplesser: It is exactly the same code as in `hh_cond_exp_traub`, so I suggest to leave this as it its.

models/hh_cond_beta_gap_traub.cpp
heplesser: I agree with the suggestion to check ```std::abs(denom1) < std::numeric_limits< double >``` makes sense, also applies to the check for `denom2`.

heplesser:
#329 discusses a problem with spike generation in the `hh_cond_exp_traub` model, which most likely also applies here. So we need to update this model once we have a solution for #329.

heplesser:
@dnao Using Github's web-based tools, I just resolved a merge conflict that had occurred due to merging #1095 into master, see commit bacfb05. Could you take a look at my comments above? If you could respond to them within a few days, we should be able to merge this PR in time for the next NEST release.

models/hh_cond_beta_gap_traub.cpp
dnao: Hi, this section of the code is the same as aeif_cond_beta_multisynapse, so I suggest these changes should also apply there.

heplesser:
@dnao All looks good now, except that one regression test fails. It is expected that this test fails for `hh_cond_beta_gap_traub`. In order to exclude your model from being tested, please go to file `testsuite/regressiontests/ticket-421.sli` and add `hh_cond_beta_gap_traub` to the list of excluded models in the proper alphabetical position:
```
/exclude_models [/aeif_cond_exp /aeif_cond_alpha /aeif_cond_alpha_RK5 /a2eif_cond_exp /a2eif_cond_exp_HW 
                 /aeif_cond_alpha_multisynapse /aeif_psc_delta_clopath
                 /aeif_psc_exp /aeif_psc_alpha /aeif_psc_delta /aeif_cond_beta_multisynapse /hh_cond_exp_traub
                 /hh_psc_alpha /hh_psc_alpha_clopath /hh_psc_alpha_gap /ht_neuron /ht_neuron_fs
                 /iaf_cond_exp_sfa_rr /izhikevich] def
```

PR #1099, opened 2019-01-07
Refactor GetConnections to improve performance

### LINKED ISSUE #1085: `GetConnections` does not seem to filter by synapse label when only the `target` attribute is given.

### LINKED ISSUE #1096: Strange performances of nest.GetConnections()
_Originally posted by @alberto-antonietti in https://github.com/nest/nest-simulator/pull/920#issuecomment-409323965_

Follow-up of my comment on #920

hakonsbm:
This PR refactors parts of the `GetConnections` function to improve performance when specifying target nodes.

In the following benchmarks there are two populations of 5000 neurons each, connected with `all_to_all`. `GetConnections` is then called specifying target neurons, i.e. calling
```
<< /target target_nrns >> GetConnections
```
Times are given for the call to `GetConnections` only, in seconds.

| threads | v2.14 | ad5ef5c | refactor | v2.14 / refactor | ad5ef5c / refactor | 
|---------|---------|-------------|-----------------|-------------|-------------|
| 1       | 319.14  | 843.65      | 35.03           | 9.11        | 24.08       |
| 2       | 136.6   | 352.71      | 22.48           | 6.08        | 15.69       |
| 4       | 60.85   | 164         | 16.38           | 3.71        | 10.01       |
| 8       | 31.37   | 118.51      | 13.39           | 2.34        | 8.85        |
| 16      | 31.98   | 117.52      | 16.25           | 1.97        | 7.23        |

With the same setup, but now specifying both source and target neurons, i.e. calling
```
<< /source source_nrns /target target_nrns >> GetConnections
```

| threads | v2.14 | ad5ef5c | refactor | v2.14 / refactor | ad5ef5c / refactor |
|---------|---------|-------------|-----------------|-------------|-------------|
| 1       | 326.34  | 665.25      | 43.28           |  7.54       | 15.37       |
| 2       | 137.24  | 347.76      | 39.99           |  3.43       | 8.70        |
| 4       | 61.46   | 254.3       | 96.32           |  0.64       | 2.64        |
| 8       | 32.42   | 103.9       | 19.07           |  1.70       | 5.45        |
| 16      | 34.38   | 127.29      | 21.6            |  1.59       | 5.89        |

In both setups the refactored `GetConnections` is in almost all cases many times faster compared to both master (ad5ef5c) and v2.14. However with 4 threads the time for the refactored `GetConnections` is surprisingly long, but still an improvement on ad5ef5c.

Fixes #1096. Fixes #1085.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1221,23 +1221,23 @@ nest::ConnectionManager::get_connections(
       ConnectorBase* connections = connections_[ tid ][ syn_id ];
       if ( connections != NULL )
       {
+        const size_t num_connections_in_thread = connections->size();
+        for ( index lcid = 0; lcid < num_connections_in_thread; ++lcid )
+        {
+          const index source_gid = source_table_.get_gid( tid, syn_id, lcid );
+          connections->get_connection_with_specified_targets( source_gid,
+            target_neuron_gids,
+            tid,
+            lcid,
+            synapse_label,
+            conns_in_thread );
+        }
+
         for ( std::vector< index >::const_iterator t_gid =
                 target_neuron_gids.begin();
               t_gid != target_neuron_gids.end();
               ++t_gid )
         {
-          std::vector< index > source_lcids;
-          connections->get_source_lcids( tid, *t_gid, source_lcids );
-
-          for ( size_t i = 0; i < source_lcids.size(); ++i )
-          {
-            conns_in_thread.push_back( ConnectionDatum( ConnectionID(
-              source_table_.get_gid( tid, syn_id, source_lcids[ i ] ),
-              *t_gid,
-              tid,
-              syn_id,
-              source_lcids[ i ] ) ) );
-          }
           // target_table_devices_ contains connections both to and from
           // devices. First we get connections from devices.
           target_table_devices_.get_connections_from_devices_(
@@ -1305,18 +1305,12 @@ nest::ConnectionManager::get_connections(
             }
             else
             {
-              for ( std::vector< index >::const_iterator t_gid =
-                      target_neuron_gids.begin();
-                    t_gid != target_neuron_gids.end();
-                    ++t_gid )
-              {
-                connections->get_connection( source_gid,
-                  *t_gid,
-                  tid,
-                  lcid,
-                  synapse_label,
-                  conns_in_thread );
-              }
+              connections->get_connection_with_specified_targets( source_gid,
+                target_neuron_gids,
+                tid,
+                lcid,
+                synapse_label,
+                conns_in_thread );
             }
           }
         }
--- a/testsuite/regressiontests/issue-1085.sli
+++ b/testsuite/regressiontests/issue-1085.sli
@@ -1,0 +1,87 @@
+/*
+ *  issue-1085.sli
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+ /** @BeginDocumentation
+Name: testsuite::issue-1085
+
+Synopsis: (issue-1085) run -> NEST exits if test fails
+
+Description:
+This test checks that GetConnection filter by synapse label.
+
+Author: Stine Brekke Vennemo
+FirstVersion: December 2018
+SeeAlso:
+*/
+
+(unittest) run
+/unittest using
+
+M_ERROR setverbosity
+
+{
+  ResetKernel
+
+  /nodes /iaf_psc_alpha 5 Create def
+  /nodes [nodes] Range def
+  
+  nodes nodes << /rule /all_to_all >>
+              << /model /static_synapse_lbl /synapse_label 123 >> Connect
+  nodes nodes << /rule /all_to_all >>
+              << /model /static_synapse_lbl /synapse_label 456 >> Connect
+  
+  << /source nodes /synapse_label 123 >> GetConnections length_a 25 eq
+}
+assert_or_die
+
+{
+  ResetKernel
+
+  /nodes /iaf_psc_alpha 5 Create def
+  /nodes [nodes] Range def
+  
+  nodes nodes << /rule /all_to_all >>
+              << /model /static_synapse_lbl /synapse_label 123 >> Connect
+  nodes nodes << /rule /all_to_all >>
+              << /model /static_synapse_lbl /synapse_label 456 >> Connect
+  
+  << /source nodes /target nodes /synapse_label 123 >> GetConnections length_a 25 eq
+}
+assert_or_die
+
+{
+  ResetKernel
+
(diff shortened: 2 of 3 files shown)


hakonsbm:
@stinebuu It does! Can you create a PR to my branch with the regressiontest?

alberto-antonietti:
@hakonsbm Thanks for taking care of this.

I tried your fix with previous (`ad5ef5c`) and current `master` (`201e722`) and I obtained the same improvements. I also got the same unexpectedly higher time.

However, I have seen that, even with your fix, the times needed by GetConnections changes with different connectivity rules used to generate the connections. With `v14 ` the time needed was more similar.

For example, creating 1.3M synapses between neurons A and B with a single thread:

```
A=nest.Create('iaf_cond_exp',650000)
B=nest.Create('iaf_cond_exp',2)

```
if connected with **all_to_all**:
`nest.Connect(A,B,'all_to_all', 'static_synapse')`
GetConnections takes **249s** (!!) with `201e722+PR` and 0.99s with `v2.14`

if connected with **fixed_indegree**:
`nest.Connect(A,B,{'rule': 'fixed_indegree','indegree':650000, 'multapses': False, 'autapses': False}, 'static_synapse')
`
GetConnections takes 2s with `201e722+PR` and 1.99s with `v2.14`

if connected with **fixed_outdegree**:
`nest.Connect(A,B,{'rule': 'fixed_outdegree','outdegree':2, 'multapses': False, 'autapses': False}, 'static_synapse')
`
GetConnections takes 1.7s with `201e722+PR` and 0.99s with `v2.14`

Are you experiencing the same?
Do you have any hint about the reason why exactly the same connectivity behaves differently?

hakonsbm:
@alberto-antonietti `GetConnections` after connecting all to all is slower for me as well, by a factor of ~10. With 5G (v2.16) connections have to be sorted when getting connections. Sorting the connections after connecting all to all takes longer than after connecting with fixed out- or indegree, probably because using all to all leaves the connection list in a state which takes longer to sort than using fixed out- or indegree.

heplesser:
@hakonsbm So is the factor 10 entirely down to the sorting? Since we are using our own implementation of quicksort, maybe we can figure out why things go wrong? I assume it may have something to do with quicksort's dislike of sorted data.

hakonsbm:
@heplesser The large increase in runtime for the all to all case is solely because of the sorting. Simulating for a short period before calling `GetConnections`, `GetConnections` takes about as long as in v2.14, as the connections are already sorted from simulating.

The difference between the fixed indegree case and the all to all case when sorting connections is number of swaps performed, mainly here

https://github.com/nest/nest-simulator/blob/386fb6ed77615a197d6c16fa39c738d10804a484/libnestutil/sort.h#L142-L150

Debugging with Valgrind, connecting 100000 -> 2 neurons with fixed indegree, sorting swaps elements 237277 times. Meanwhile connecting with all to all, elements are swapped 5075239 times.

Connections are sorted by the GIDs of the presynaptic neurons, and the initial order of connections is of course based on the connection rule. Fixed outdegree iterates source neurons, so connections are already sorted. Fixed indegree iterates target neurons and picks random source neurons, so connections are more or less random when sorting.

All to all iterates target neurons, then for each target neuron iterates source neurons (at least in this benchmark setup). This means that sources of the connections are the sequence of source neurons, repeated for each target neuron, e.g. [1, 2, ..., N, 1, 2, ..., N]  for N sources and 2 targets. This form is apparently a bad case for the sorting algorithm.

#### Is using std::sort a better alternative?
I benchmarked the NEST quicksort implementation versus `std::sort`, connecting 1M -> 2 neurons for fixed in- and outdegree, and 200000 -> 2 neurons for all to all, with the following results (all in seconds)

| Connection rule | NEST quicksort | `std::sort` |
|-----------------|----------------|-----------|
| Fixed outdegree | 0.0394041      | 0.0546829 |
| Fixed indegree  | 0.24474        | 0.187029  |
| All to all      | 5.82502        | 15.9669   |

There doesn't seem to be much to gain by using `std::sort` in these cases, especially as there would be additional overhead from adaptions to apply operations from `std::sort` on a second vector (which the NEST quicksort does natively). However, when connecting 30000 -> 3000 neurons all to all:

| Connection rule | NEST quicksort | `std::sort` |
|-----------------|----------------|-----------|
| All to all      | 55.5934        | 4.47839   |

Here the NEST quicksort implementation struggles, while `std::sort` is more than 12 times as fast.

jakobj:
have you tried it with 30000 -> 3000 +- a few? the issue seems to be how the pivot element is chosen in an array with regular structure. for a badly chosen pivot element, quicksort needs an unnecessary high number of swapping operations as it unevenly divides the elements. usually one would protect against these kinds of performance issues by first shuffling the original array (complexity linear in N) which we have failed to do here. lesson learned. seems like we have three options:
1. implement a shuffle operation which keeps sources and connections in the correct relative order and apply it before sorting the array
2. find a smarter way to choose the pivot
3. find a library implementation of quicksort that is already battle tested and can do what we need to replace this implementation

hakonsbm:
@jakobj I have implemented an alternative sort method using Boost sort. Their algorithm avoids this pivot problem, and is quite a lot faster when sorting random numbers as well. There will be a PR soon.

jougs:
@jakobj: would you be willing to serve as reviewer here? I have the feeling you are more into this than I am... Thanks!

nestkernel/connection_manager.cpp
```diff
@@ -1221,23 +1221,23 @@ nest::ConnectionManager::get_connections(
...
+        {
+          const index source_gid = source_table_.get_gid( tid, syn_id, lcid );
+          connections->get_connection_with_specified_targets( source_gid,
+            target_neuron_gids,
+            tid,
+            lcid,
+            synapse_label,
+            conns_in_thread );
+        }
+
```
jakobj: as far as I can tell this loop uses at most `n_conns * n_targets` comparisons (loop + linear complexity of `std::find` in `get_connection_with_specified_targets`). the previous implementation also should use the same amount of comparisons in the worst case. your new implementation is much faster, so where did my reasoning fail?

testsuite/regressiontests/issue-1085.sli
```diff
@@ -0,0 +1,87 @@
+/*
```
jakobj: should this be part of this PR?

nestkernel/connection_manager.cpp
hakonsbm: In the previous implementation both the vector of specified targets and the vector of local connections are always completely iterated, so there are always `n_conns * n_targets` comparisons, like you say. This implementation iterates over the connections first, so it only has to check if the target of the connection is among the specified targets, and therefore doesn't have to check all of them for every connection. Also `std::find` may do some hardware-near optimizations.

testsuite/regressiontests/issue-1085.sli
hakonsbm: Yes. @stinebuu suggested as both this PR and #1093 fixes #1085, #1093 should be closed in favor of this. Therefore this PR got the regressiontest.

jakobj: ok

nestkernel/connection_manager.cpp
jakobj: ok, I think I understand. while the worst case runtime is identical for both implementations, the average case is much better in the new one. :+1:

jakobj:
looks good! thanks for your effort :+1:

PR #1101, opened 2019-01-10
local_spike_counter takes multiplicity into account

### LINKED ISSUE #1100: local_spike_counter does not handle multiplicity
`local_spike_counter` does not take multiplicity into account when counting the spikes. By running the example below, you'll see that the `local_spike_counter` and `n_events` differ, but they should be the same. If we look at the delivered spikes (if we connect a `spike_detector`), we see that all spikes are delivered, so it is just the `local_spike_counter` that is the problem. I'll make a PR shortly.

```
/s_to /parrot_neuron 4000 Create def
/t_to /parrot_neuron 10 Create def

/sources 1 s_to cvgidcollection def
/targets s_to 1 add t_to cvgidcollection def

/pg /poisson_generator << /rate 1.0 >> Create def
/sd /spike_detector Create def

pg pg cvgidcollection sources Connect
sources targets Connect

sources sd sd cvgidcollection Connect
targets sd sd cvgidcollection Connect

10.0 Simulate

0 GetStatus /local_spike_counter get ==
sd GetStatus /n_events get ==
```

stinebuu:
With this PR, the `local_spike_counter` will also add the multiplicity so that we get the correct spike count.

This fixes #1100.

### DIFF
--- a/nestkernel/event_delivery_manager_impl.h
+++ b/nestkernel/event_delivery_manager_impl.h
@@ -63,7 +63,8 @@ EventDeliveryManager::send< SpikeEvent >( Node& source,
   e.set_sender_gid( source_gid );
   if ( source.has_proxies() )
   {
-    ++local_spike_counter_[ tid ];
+    local_spike_counter_[ tid ] += e.get_multiplicity();
+
     e.set_stamp(
       kernel().simulation_manager.get_slice_origin() + Time::step( lag + 1 ) );
     e.set_sender( source );
--- a/testsuite/regressiontests/issue-1100.sli
+++ b/testsuite/regressiontests/issue-1100.sli
@@ -27,8 +27,9 @@ Name: testsuite::issue-1100
 Synopsis: (issue-1100) run -> NEST exits if test fails
 
 Description:
-This test ensures that local_spike_counter also counts the spikes with
-multiplicity.
+Creates a parrot neuron, which receives one spike with multiplicity two from a
+spike generator. The parrot neuron should then emit a spike with multiplicity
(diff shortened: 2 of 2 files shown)


stinebuu:
@suku248 The test in this PR creates 4000 source parrot neurons and 10 target parrot neurons and the source neurons receive spikes from a `poisson_generator`, not a `spike_generator`. With the `poisson_generator` some of the spikes will have multiplicity greater than one. Should I update the test to just have two parrot neurons and use a `spike_generator` with multiplicity two? It would make the test more explicit. Or would you be ok with keeping the test as is, and having this description:

Description:
Creates 4000 source parrot neurons and 10 target parrot neurons. The source parrots receive spikes from a poisson generator, some of which will have multiplicity greater than one. The source neurons should then emit a spike with multiplicity greater than one to the target neurons, which should be counted as multiple spikes by the local spike counter.

suku248:
@stinebuu As this is a regression test, I'd prefer it to be as minimalist as possible. A minimalist test that exposes the bug (spikes of parrot neurons with multiplicity > 1 are not counted multiple times) is the one that I described above: One spike generator is connected to one parrot neuron. The spike generator feeds one spike with multiplicity two to the parrot neuron, which then emits a spike with multiplicity two. In the current master, the local spike counter only counts one spike, which is incorrect :bug:. With this PR, the local spike counter should count two spikes :butterfly:.

stinebuu:
@suku248 the emojis are making me so happy! I have now simplified the test, which had the added bonus of loosing the scattered `cvgidcollection`s, @jougs :grin:

PR #1103, opened 2019-01-14
Throw error if creating too many connections on one thread

### LINKED ISSUE #1102: Too many connections on one thread leads to erroneous results
Creating more than 2²⁷ = 134217728 connections with one synapse type on one thread leads to erroneous results when simulating. This is because the `Target` class uses 27 bits for storing the local connection index, `lcid`. New connections beyond those 134217728 therefore get the wrong `lcid`, leading to wrong results.

There are asserts in the `Target` class that check if `lcid` gets too high, but they never trigger, and no other messages that something might go wrong are given when trying to create more connections than can be handled.

See also #1043.

hakonsbm:
There is currently a limit of 2²⁷ connections per synapse per thread, due to having 27 bits for the `lcid` in the `Target` class. With this PR an exception is thrown if attempting to create more connections than `Target` can handle.

Fixes #1102.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -685,6 +685,10 @@ nest::ConnectionManager::increase_connection_count( const thread tid,
     num_connections_[ tid ].resize( syn_id + 1 );
   }
   ++num_connections_[ tid ][ syn_id ];
+  if ( num_connections_[ tid ][ syn_id ] >= 1 << 27 )
+  {
+    throw KernelException("Too many connections per thread for synapse.");
+  }
 }
 
 nest::index
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -687,7 +687,10 @@ nest::ConnectionManager::increase_connection_count( const thread tid,
   ++num_connections_[ tid ][ syn_id ];
   if ( num_connections_[ tid ][ syn_id ] >= ( 1 << 27 ) - 1 )
   {
-    throw KernelException( "Too many connections per thread for synapse." );
+    throw KernelException( String::compose(
+      "Too many connections: at most %1 connections supported per virtual "
+      "process and synapse model.",
+      ( 1 << 27 ) - 1 ) );
   }
 }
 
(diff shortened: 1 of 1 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -685,6 +685,10 @@ nest::ConnectionManager::increase_connection_count( const thread tid,
     num_connections_[ tid ].resize( syn_id + 1 );
   }
   ++num_connections_[ tid ][ syn_id ];
+  if ( num_connections_[ tid ][ syn_id ] >= 1 << 27 )
```
clinssen: Can the expression (1 << 27) be changed to `MAX_LCID` after merging #1088? See: https://github.com/nest/nest-simulator/pull/1088/files#diff-42615d7378f1abcdbe8ff2af37737101R78 Thanks!

hakonsbm: Yes, I'll update it once #1088 is merged.

heplesser: Is this comparison strictly safe, i.e., will everything work with 2^27 connections, or is the limit rather 2^27-1?

nestkernel/connection_manager.cpp
```diff
@@ -685,6 +685,10 @@ nest::ConnectionManager::increase_connection_count( const thread tid,
...
+  {
+    throw KernelException( "Too many connections per thread for synapse." );
```
heplesser: I think this message is not very informative for the end user. How about something like `"Too many connections: at most ??? connections supported per virtual process and synapse model."`? The number can be inserted using `compose()` from `libnestutil`.

nestkernel/connection_manager.cpp
hakonsbm: @heplesser You are right, I have changed it to 2^27-1.

nestkernel/connection_manager.cpp
hakonsbm: I have made the error message more informative.

heplesser:
@hakonsbm One important thing: Once #1088 is in place, the hard-coded `27` in the code introduced here needs to be replaced with the proper compile-time variable. Can you create a PR against jakobj:feature/cmake-target-customization to ensure this?

hakonsbm:
@heplesser I'll take care of it when this is merged.

terhorstd:
marked this as blocked till #1088 is in.

PR #1105, opened 2019-01-15
Added mpi multiplicity test

stinebuu:
As far as we can tell, there doesn't seem to be an MPI test for correct number of spikes when multiplicity > 1, so here I have made one.


stinebuu:
Thanks for the reviews @jougs and @suku248! I have now updated the test. :)

PR #1111, opened 2019-01-30
Conditional travis builds using build stages

lekshmideepu:
Added four build stages: 
          * Static code analysis check
          * MPI and Threading (5 different jobs)
          * Full builds for Python 3.4.4 and Python 2.7.13
          * cpp builds for GCC 8 and Clang 7 (2 jobs)


jougs:
@lekshmideepu: nice PR number, btw :smile_cat:

Issue #1087, opened 2018-12-10
Weird behavior (probably numerics) in iaf_psc_exp_ps for tau_m=tau_syn


healther:
It seems as if the `iaf_psc_exp_ps` model has problems integrating synaptic input for `tau_m=tau_syn_*`. We think it is a problem with the implementation of the alpha-function, as it works for even slight differences between `tau_m` and `tau_syn`.

The following code produces the 4 attached plots (executed with nest 2.14.0 and reproduced by @jakobj on current master), bug found and reproducer code by @JulianGoeltz
```
import matplotlib.pyplot as plt
import nest
def doit(ratio, model):
    print("MWE_plot_{}_{}.svg".format(model, ratio))
    tau_syn = 10.
    nest.ResetKernel()
    nest.ResetNetwork()
    x = nest.Create(model, 1, {'tau_m': ratio * tau_syn, 'tau_syn_ex': tau_syn, 'tau_syn_in': tau_syn,
                                       'V_th': -10., 'E_L': -20., 'V_m': -15.})
    y = nest.Create('spike_generator', 1, {'spike_times': [53., 55., 57.]})
    nest.Connect(y, x, syn_spec={'weight': 240.})
    vm = nest.Create('voltmeter')
    nest.Connect(vm, x)
    nest.Simulate(100.)
    ev = nest.GetStatus(vm, 'events')[0]
    fig = plt.figure()
    plt.plot(ev['times'], ev['V_m'])
    fig.savefig("MWE_plot_{}_{}.svg".format(model, ratio))

for model in ['iaf_psc_exp', 'iaf_psc_exp_ps']:
    for ratio in [1., 2.]:
        doit(ratio, model)
```

[MWE_plot_iaf_psc_exp_1.0.pdf](https://github.com/nest/nest-simulator/files/2664051/MWE_plot_iaf_psc_exp_1.0.pdf)
[MWE_plot_iaf_psc_exp_2.0.pdf](https://github.com/nest/nest-simulator/files/2664052/MWE_plot_iaf_psc_exp_2.0.pdf)
[MWE_plot_iaf_psc_exp_ps_1.0.pdf](https://github.com/nest/nest-simulator/files/2664053/MWE_plot_iaf_psc_exp_ps_1.0.pdf)
[MWE_plot_iaf_psc_exp_ps_2.0.pdf](https://github.com/nest/nest-simulator/files/2664054/MWE_plot_iaf_psc_exp_ps_2.0.pdf)

[...]

PR #1112, opened 2019-02-01
Fix stability of iaf_psc_exp_ps for equal membrane and synaptic time constants

jakobj:
as discussed in #1087 

I suggest @heplesser and @healther as reviewers

### DIFF
--- a/libnestutil/propagator_stability.h
+++ b/libnestutil/propagator_stability.h
@@ -23,6 +23,8 @@
 #ifndef PROPAGATOR_STABILITY_H
 #define PROPAGATOR_STABILITY_H
 
+// Propagators to handle similar tau_m and tau_syn_* time constants.
+// For details, please see doc/model_details/IAF_neurons_singularity.ipynb.
 double propagator_31( double tau_syn, double tau, double C, double h );
 double propagator_32( double tau_syn, double tau, double C, double h );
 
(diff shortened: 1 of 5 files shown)


clinssen:
@jakobj: thanks for the patch! I've established that this is also an issue in the NESTML generated code (see nest/nestml#460).

As for this PR, I was initially at a loss for documentation of the `propagator_32()` functions. After some digging I found that things are quite nicely described in the Jupyter notebook in https://github.com/nest/nest-simulator/blob/master/doc/model_details/IAF_neurons_singularity.ipynb. Could you please update the documentation link in https://github.com/nest/nest-simulator/blob/73b0b530c112cc7715dfd73fe593c7c43750f067/precise/iaf_psc_exp_ps.h#L100 and https://github.com/nest/nest-simulator/blob/73b0b530c112cc7715dfd73fe593c7c43750f067/precise/iaf_psc_alpha_presc.h#L89 and https://github.com/nest/nest-simulator/blob/73b0b530c112cc7715dfd73fe593c7c43750f067/precise/iaf_psc_alpha_canon.h#L112 and add the same ipynb link to libnestutil/propagator_stability.h?

As a final question, do you know whether there might be other models that are also vulnerable to this edge case in their current implementation?

heplesser:
@clinssen All models with `tau_m` and `tau_syn*` parameters are candidates if they are integrated using exact integration. So `cond`  models are "safe", because they are integrated using standard ODE solvers.

jakobj:
thanks for the comments, i've updated the doc.

PR #1116, opened 2019-02-07
Add ostream operator template for std::vectors

terhorstd:
Streaming `std::vector`s just streams all elements (given that the elements support `cout << obj;`). This is very useful in many circumstances and should not be implemented each time anew.


jougs:
@terhorstd: you can either fix the format of the code you're proposing or start another PR to change the formatting guidelines as discussed two weeks ago and to make your code pass the tests. Both options would be fine for me at this point in time.

nestkernel/streamers.h
```diff
@@ -0,0 +1,47 @@
...
+		if (not first)
+		{
+			os << ", ";
+		}
+		os << element;
+	}
+	return os << "]";
+}
+
+#endif //ndef STREAMERS_H
```
heplesser: Why `ndef`?

nestkernel/streamers.h
```diff
@@ -0,0 +1,47 @@
+/*
```
heplesser: `libnestutil` would be a better location for this file.

nestkernel/streamers.h
```diff
@@ -0,0 +1,47 @@
...
+ *  streamers.h
```
heplesser: I don't like the name `streamers`.  How about `prettyprint`?

nestkernel/streamers.h
jougs: +1

nestkernel/streamers.h
terhorstd: The function does not print anything, so I would not call it "*print". The function is an operator with any ostream object as first argument. The term "stream" in this context is C++ convention. ...and just dumping an array to a stream is not "pretty" in my opinion. Would you rather like a different pre-/postfix than "vector[" and "]"?

nestkernel/streamers.h
terhorstd: I agree

nestkernel/streamers.h
terhorstd: identifies the expected #if statement. we don't seem to have a consistent convention for `#endif`s. I'll leave this for now. We should set up a convention for this.

libnestutil/streamers.h
```diff
@@ -0,0 +1,49 @@
...
+{
+  os << "vector[";
+  bool first = true;
+  for ( const auto& element : vec )
+  {
+    if ( not first )
+    {
+      os << ", ";
+    }
+    os << element;
```
heplesser: @terhorstd Don't you need to set `first` to false somewhere?

PR #1118, opened 2019-02-08
Connection sorting using Boost's sorting function

hakonsbm:
This PR implements a helper iterator, `iterator_pair`, to be able to sort contents of two containers (based on values from the first, with permutations applied to the second) using library functions, specifically using the Boost.Sort library. If NEST is not compiled with Boost support, it falls back on the current sorting implementation. The motivation for using a library sorting function is unexpectedly higher sorting times in certain cases, as discussed in #1099.

Note that Boost version >=1.58.0 is required, so if an older version is installed, NEST is installed without Boost.

#### Benchmarks
Addressing the issues discussed in #1099, connecting many neurons to few neurons with `all_to_all` (sorting times given in seconds):

| Connection     | master | Boost.Sort | Speedup |
|----------------|--------|------------|---------|
| 100000->2      | 3.16   | 0.06       | 52.64   |
| 10000->1000    | 5.80   | 0.46       | 12.72   |

Using hpc_benchmark, timing only the sorting:

| Scale | master | Boost.Sort | Speedup |
|-------|--------|------------|---------|
| 8     | 11.26  | 5.42       | 2.08    |
| 12    | 22.82  | 10.61      | 2.15    |
| 16    | 32.92  | 14.19      | 2.32    |
| 20    | 43.88  | 18.58      | 2.36    |

So using the Boost.Sort library both solves the problems discussed in #1099, and sorts (mostly) random connections considerably faster than the current implementation.


jakobj:
the timing looks very convincing, thanks! haven't looked at it in details, but is doesn't seem to include the fix suggested (https://github.com/nest/nest-simulator/pull/1099#issuecomment-459834317) for the sorting if boost is absent. could you please include this? would be great to see how the timings compare after applying this fix.

jakobj:
after talking to @suku248 I can express another concern of mine more clearly: the boost implementation does not reduce complexity by replacing the custom sort implementation (necessary as a fallback), but rather increases complexity by introducing a significant amount of new loc via the new `IteratorPair` class. The custom sort and the boost library calls and helpers will hence both need to be maintained. If the performance difference is not an order of magnitude or so, I would hence refrain from increasing complexity and rather have only one implementation.

heplesser:
@hakonsbm @jakobj Since using the boost sort gives us a factor 2 in realistic cases such as hpc_benchmark, I suggest we use the boost sort if available. At the same time, we should implement alternative pivoting in `nest_quicksort` to avoid the worst-case behavior shown by the two first benchmark results above.

hakonsbm:
@heplesser I have added the changes proposed by @jakobj.

libnestutil/block_vector.h
```diff
@@ -76,7 +76,10 @@ class bv_iterator
   using value_type = value_type_;
   using pointer = ptr_;
   using reference = ref_;
-  using difference_type = size_t;
+  using difference_type = long int;
+
```
gtrensch:
```suggestion
```

gtrensch:
@hakonsbm thanks for this contribution! I also agree with @heplesser , a factor 2 is worth to include boost here. I just deleted an empty line (see the suggested code change above). I approve this PR.

testsuite/cpptests/test_sort.h
```diff
@@ -35,18 +35,10 @@
 namespace nest
 {
 
-const bool
-is_sorted( std::vector< size_t >::const_iterator begin,
```
jakobj: where did this function go?

hakonsbm: We only sort `BlockVector`s, so this function is never used.

PR #1119, opened 2019-02-08
Improve connection performance over threads

stinebuu:
This PR contains two new changes that improve the connection performance, especially over increasing number of threads and for large networks:

1. Instead of all threads writing to `has_primary_connections_` or `secondary_connections_exist_` when we add a new connection, we check if the individual thread has primary or secondary connections, and only write to the variable if the thread has not previously updated it. We also make sure that the threads don't try to write to the variable all at once.
2. Instead of creating a new empty dummy parameter dictionary every time we connect two nodes, we create a static dummy dictionary once, and then use this dummy parameter every time we need an empty dictionary.

These two fixes improves the performance of master quite a bit, but we are not on the level with the NEST version used in the 5g paper (f104c6d), see the plots below. The difference between `5g` and this updated master (`master-update`) comes from the BlockVector.

Here I have run the `hpc_benchmark` with a scale of 10, which equals 112500 neurons and 1.265738e+09 connections, on 1 node on the Piz Daint supercomputer, on an increasing number of threads per MPI process (strong scaling over threads). The number of threads are 3, 6, 9, 18, and 36. The figure shows the connection time vs threads.

strong-scaling-threads-pr

The figure below  also shows the connection time over increasing number of threads, but the number of VPs are fixed to 36. Here, the scale is 20 (almost filling the node), that is 225000 neurons and 2531476000 connections. As you can see, connection time slightly increases as the threads increases, which is strange, but at least we are on not that worse off than with 2.14.

connection-times-threads-pr

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -70,7 +70,9 @@ nest::ConnectionManager::ConnectionManager()
   , have_connections_changed_( true )
   , sort_connections_by_source_( true )
   , has_primary_connections_( false )
+  , check_primary_connections_()
   , secondary_connections_exist_( false )
+  , check_secondary_connections_()
   , stdp_eps_( 1.0e-6 )
 {
 }
@@ -92,6 +94,9 @@ nest::ConnectionManager::initialize()
   connections_.resize( num_threads );
   secondary_recv_buffer_pos_.resize( num_threads );
   sort_connections_by_source_ = true;
+
+  check_primary_connections_.resize( num_threads, false );
+  check_secondary_connections_.resize( num_threads, false );
 
 #pragma omp parallel
   {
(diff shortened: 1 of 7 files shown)


stinebuu:
Here are some other benchmarks: [PR1119-checks.pdf](https://github.com/nest/nest-simulator/files/2850254/PR1119-checks.pdf)

nestkernel/connection_manager.cpp
```diff
@@ -633,13 +638,17 @@ nest::ConnectionManager::connect_( Node& s,
 
   increase_connection_count( tid, syn_id );
 
-  if ( is_primary )
+  if ( not check_primary_connections_[ tid ] and is_primary )
```
jakobj: why do we need `check_primary_connections_` for every thread? wouldn't one for the whole process suffice since threads only read this variable?

stinebuu: @jakobj Do you mean checking `has_primary_connections_` instead? I did try without `check_primary_connections_` and did the following
```
if ( not has_primary_connections_ and is_primary )
{
#pragma omp atomic write
    has_primary_connections_ = true;
}
```
but that led to worse performance on Piz Daint. I don't quite understand why.

jakobj: ok, maybe add this as a comment to avoid that someone will "improve" this bit?

PR #1122, opened 2019-02-13
hpc_benchmark.sli is now invariant under different MPI-process/thread splits with fixed VP number

### LINKED ISSUE #247: Make hpc_benchmark.sli example invariant under mpi/thread splits
The `hpc_benchmark.sli` example is currently not invariant under different MPI/Thread-splits of a fixed number of virtual processes. This has historic reasons and is not a problem in itself, but I think it would make much sense to revise the example to be invariant.

The key issues are
- Recording from a fixed number of neurons per _mpi process_. The total number of recorded neurons should be fixed.
- Membrane potential randomization based on _mpi processes_. Randomization should be based on virtual processes.

[This Gist](https://gist.github.com/heplesser/e78bed61478c25f7047c) is a modified version of the benchmark as a first step (cf #237). In that version, membrane potential is not randomized.

stinebuu:
This PR fixes #247.

I have added a new example, `hpc_benchmark_invariant.sli`, a mpi/thread-split invariant version of `hpc_benchmark.sli`. To make it invariant under mpi/thread splits, we
- record from a fixed set of neurons (the Nrec first exitatory), instead of a fixed number of local neurons
- randomize the membrane potential based on the virtual processes, not the mpi processes.

With this new version, we get the following average rates (also averaged over processes). I include the rates from hpc_bencmark as well for comparison, scale = 1.0.

| Num_vps | Split | hpc_benchmark_invariant | hpc_benchmark |
|:------------:|:------:|:-----------------------------------:|:---------------------:|
|        1       | 1 mpi proc., 1 thread   |  11.832  |        11.832         |
|        4       | 1 mpi proc., 4 threads |    13.5    |        11.812         |
|        4       | 2 mpi proc., 2 threads |    13.5    |         29.06          |
|        4       | 4 mpi proc., 1 thread   |    13.5    |         54.58          |

### DIFF
--- a/examples/nest/hpc_benchmark_invariant.sli
+++ b/examples/nest/hpc_benchmark_invariant.sli
@@ -1,0 +1,512 @@
+/*
+ *  hpc_benchmark_invariant.sli
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/*
+   This script produces a balanced random network of scale*11250 neurons in
+   which the excitatory-excitatory neurons exhibit STDP with
+   multiplicative depression and power-law potentiation. A mutual
+   equilibrium is obtained between the activity dynamics (low rate in
+   asynchronous irregular regime) and the synaptic weight distribution
+   (unimodal). The number of incoming connections per neuron is fixed
+   and independent of network size (indegree=11250).
+
+   This is the standard network investigated in:
+   Morrison et al (2007). Spike-timing-dependent plasticity in balanced random
+     networks. Neural Comput 19(6):1437-67
+   Helias et al (2012). Supercomputers ready for use as discovery machines for
+     neuroscience. Front. Neuroinform. 6:26
+   Kunkel et al (2014). Spiking network simulation code for petascale 
+     computers. Front. Neuroinform. 8:78
+
+
+   The difference between this script and hpc_benchmark.sli is that this version
+   is invariant under mpi/thread splits. This affects
+     - Randomization of the membrane potential (here it is based on the virtual
+       processes)
+     - Neurons that are being recorded from (here we always record from the
+       first Nrec neurons).
+ 
+ 
+   A note on scaling
+   -----------------
+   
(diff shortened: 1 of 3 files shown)


jakobj:
sounds good, but could you elaborate why the rates increases so much for the old hpc_benchmark? this is hopefully just a measurement artifact?

stinebuu:
@jakobj Yes, it has to do with the neurons we record from. If I use the record strategy from hpc_benchmark_invariant in the hpc_benchmark, that is, always record from the Nrec first neurons and not let recording be based on the mpi processes, the increase in rate is much lower. Then the rate is 11.812 for 1 mpi proc. 4 threads, 14.68 for 2 mpi procs. 2 threads and 13.49 for 4 mpi procs. 1 thread.

heplesser:
@stinebuu Nice, but I would suggest to also correct the firing rate calculation in the original `hpc_benchmark.sli`, so that one gets sensible results.

Should we in addition to the average rate also report 
- the total spike count
- the number of nodes
- the number of connections?

examples/nest/hpc_benchmark_invariant.sli
```diff
@@ -0,0 +1,512 @@
...
+  I_neurons I_neurons << /rule (fixed_indegree) /indegree CI /autapses false /multapses true >>
+                      << /model /syn_in >> Connect
+
+
+  record_spikes true eq
+  {
+    E_neurons size Nrec lt
+    {
+      M_ERROR (BuildNetwork)
+      (Networks is smaller than number of neurons to record from!) message
```
jakobj: since your only checking the size of the E population, the error message should reflect this

examples/nest/hpc_benchmark_invariant.sli
```diff
@@ -0,0 +1,512 @@
...
+
+  ResetKernel
+
+  memory_thisjob cvs ( # virt_mem_0) join logger /log call
+
+  BuildNetwork
+
+  tic
+
+  Prepare
```
jakobj: why use prepare, run, cleanup instead of simulate? do the advantages outweigh the change wrt the hpc_benchmark?

stinebuu: I think I must have been inspired by another benchmark. Tested quickly now, and it does not seem to have an effect, will change to Simulate.

stinebuu:
@heplesser In #247 it was decided that we should keep the original `hpc_benchmark.sli` as constant as possible as it is used in many publications. If I change the firing rate calculations the benchmark will change a bit, should I still change the benchmark? I am also wondering if we would then really need the invariant version? The difference between the invariant version and the original hpc_benchmark is basically just the firing rate calculations and the detector scheme, so if we change one we might as well change both.

I'll add the total spike count, number of nodes and number of connections to `hpc_benchmark.sli`, it is already reported in `hpc_benchmark_invariant.sli`. :)

heplesser:
@stinebuu I agree that the changes are so small that we probably can just update `hpc_benchmark.sli`, but I would be interested in @suku248's take on this.

I also think we should move the definitions of `/max_rank_cout` and `/max_rank_log`, which are now hidden in the definition of the `/logger` function forward to the parameter section to make them visible and accessible to the user.

stinebuu:
@jakobj @jarsi @heplesser I have now renamed the original hpc_benchmark to be `hpc_benchmark_2_16.sli` and called my invariant one for `hpc_benchmark.sli`. I have added a comment in `hpc_benchmark_2_16.sli` about it being the original one, so if you want to reproduce your old results, you should use that one.

Issue #459, opened 2016-08-11
Inhibitory weight distribution


lepmik:
Wen connecting two populations with pyNEST the sign of synaptic weights define if synapses are inhibitory or excitatory. Presently it does not seem to be possible to explicitly select an inhibitory weight distribution. Below is an example of a workaround and a suggestion for a fix.

```python
Connect(nodes_in, nodes_ex,
       {'rule': 'pairwise_bernoulli', 'p': p_in},
       syn_spec={'model': 'static_synapse',
                 'weight': {'distribution': 'lognormal',
                            'mu': mu_in,
                            'sigma': sigma_in},
                  "delay":delay})
conns = GetConnections(nodes_in, nodes_ex)
conn_vals = GetStatus(conns, 'weight')
SetStatus(conns, [{'weight': -1.*val} for val in conn_vals])
```

Comment from mailing list by Hans Ekkehard Plesser:
> At present, this is not possible in NEST, but I think it should be reasonably easy to implement. Take a look at the RandomParameter class defined in conn_parameter.{h,cpp}.
> 
> I we added an extra flag 'sign' with allowed values -1 and +1, the constructor could extract it from the dict defining the distribution and store it in a local variable sign_. The value_{int, double}() methods would then multiply the return value with sign_ before returning.
> 
> Note that any clipping of values would be applied to the original values before multiplying with sign_.

heplesser:
This issue follows up ticket 887 in the old NEST trac.

heplesser:
A more general approach would be to support arbitrary mathematical expressions for parameters. The question is how to implement this efficiently. Topology has *undocumented* support for this on the SLI level:
```
topology using
/p1 << /uniform << /min 0.5 /max 2.0 >> >> CreateParameter def
/p2 << /constant << /value 0.5 >> >> CreateParameter def
/pprod p1 p2 mul def
```
This is undocumented since we have not come up with a good solution for lifting this to the Python level yet. Something like
```python
p1 = nest.parameter.uniform(min=0.5, max=2.0)
p2 = nest.parameter.constant(0.5)
p3 = p1 * p2
```
might be feasible, but the question is how to allow for efficient evaluation of the resulting expression.

mschmidt87:
[...]

PR #1126, opened 2019-02-18
Add Allen Institute GLIF neuron models

binghuangcai:
GLIF (Generalized Leaky Integrate and Fire) Models for NEST Simulator


binghuangcai:
We also have a question about where the Copyright Transfer Agreement form should be sent to. Thanks.

heplesser:
> We also have a question about where the Copyright Transfer Agreement form should be sent to. Thanks.

@binghuangcai Thank you for your contribution! Please send the CTA to [hans.ekkehard.plesser@nmbu.no](mailto:hans.ekkehard.plesser@nmbu.no).

heplesser:
@binghuangcai It appears that your code format does not agree with the NEST code formatting standards. Could you run `extras/check_code_style.sh` from your source code directory and fix problematic files with `clang-format-3.6 -i ...`?

heplesser:
@binghuangcai Thanks for the formatting fix. But it looks it may have broken the code: it does not compile properly on Travis. Could you check the log at 
https://travis-ci.org/nest/nest-simulator/jobs/508126166 ?

binghuangcai:
@heplesser Thanks. I have made some organizing changes to the conductance based models. The most recent version has passed the compile check. However, there are some problems with the Mpi-threading-python test on Travis. See https://travis-ci.org/nest/nest-simulator/builds/508594819. Are these something easy for you to fix, or we should fix them by ourselves? Thanks.

clinssen:
@binghuangcai : it looks like the failing test is regression test "ticket-459" (see the file `testsuite/regressiontests/ticket-459.sli`). According to its description:

    Description:
    For all models having a reversal potential E_L, we change this potential and check that all
    other parameters and properties of the model remain unchanged. #459 reports that the 
    order in which E_L and V_th are set determines which values are actually set. This is a
    direct consequence of incorrect handling of changes to E_L, which show up as changes in,
    e.g. V_m and V_th as a side effect of changing E_L. This happened in those neuron models
    that internally represent voltages relative to E_L and did not adjust those values on
    changes to E_L.

Could you comment on whether this holds for the newly added models, or if not, why not (we can add an exclusion list to the regression test)?

binghuangcai:
@heplesser It probably this line [V_m_ = p.E_L_ ](https://github.com/AllenInstitute/nest-simulator/blob/6e1a7d0be781a5284d9a6a3aa94421b624e68e9f/models/glif_lif.cpp#L144) in the [nest::glif_lif::State_::set(...)](https://github.com/AllenInstitute/nest-simulator/blob/6e1a7d0be781a5284d9a6a3aa94421b624e68e9f/models/glif_lif.cpp#L138) function causes the failure of ticket-459. We use absolute value but no relative value for V_m_, but we need to initialize the V_m_ to p.E_L_. I have tried a few other options to initialize the V_m_, e.g., in [nest::glif_lif::State_::State_(...)](https://github.com/AllenInstitute/nest-simulator/blob/6e1a7d0be781a5284d9a6a3aa94421b624e68e9f/models/glif_lif.cpp#L77), which ends up initializing V_m_ to the default value (-77.4) of E_L_ but not the one set when creating neurons. The current code works well for initialization of V_m_, but it may cause the failure for ticket-459. Is there other way I can initialize V_m_ to E_L_ but not cause problem for ticket-459? Thanks.

heplesser:
@binghuangcai In your code
```c++
  updateValue< double >( d, names::V_m, V_m_ );
  V_m_ = p.E_L_;
```
the first line changes `V_m_` to the value given by the user, while the second line forces the value to `p.E_L_`.  This means the user cannot set the membrane potential to arbitrary values. Why not? The `State_::set()` method is used by `SetStatus` and `SetDefaults` directly, and behind the scenes by `Create`. If setting `V_m_` to user-defined values should not be possible, then there should be no `State_::set()` method. But users generally expect to be able to modify the membrane potential, so it would be interesting to understand why this might not make sense for the glif models.

binghuangcai:
@heplesser glif models would allow user to set value to V_m_. The problem is whether it is possible to set the default initial `V_m_` to `p.E_L_`, which set in `Create`. As mentioned in previous comment, a few tries in `State_::State_()` seems not work. If it is not possible to do this automatically inside Nest, the only way would be letting user to set the initial `V_m_` if needed. Is this the common practice? Thanks.

heplesser:
@binghuangcai You can pass the `Parameters_` object to the `State_` constructor:
```c++
nest::glif_lif::State_::State_( const Parameters_& p )
  : V_m_( p.E_L ) // mV
  , I_( 0.0 )     // pA
{
}
```
This will ensure consistent initialization on construction, even if the user changes the default value of `E_L`.

If the user later changes `E_L` by calling `SetStatus`, in my opinion `V_m` should not change automatically. If the user wants to change `V_m` as well, she can pass it in explicitly.

binghuangcai:
@heplesser I did try that (i.e., passing `Parameters_` to `State_` constructor) before and just checked it one more time, but it was not working. I explicitly print the `E_L` out. It looks `E_L` was set to the value set in `Create` and it was then set back to the default value. Could `SetStatus` be called first before `SetDefaults` is called in `Create`?

models/glif_lif_asc.h
```diff
@@ -0,0 +1,285 @@
...
+    //! Logger for all analog data
+    nest::UniversalDataLogger< glif_lif_asc > logger_;
+  };
+
+  struct Variables_
+  {
+    double t_ref_remaining_; // counter during refractory period in ms
+    double t_ref_total_;     // total time of refractory period in ms
+    int method_; // voltage dynamics solver method flag: 0-linear forward euler;
+                 // 1-linear exact
```
heplesser: Since the linear exact method is available, is there any reason to include forward Euler? I think the linear exact method might be improved even more by integrating all currents using exact integration; see `mat2_psc_exp` for an example and the appendix of [Heiberg et al (2018)](http://dx.doi.org/10.1007/s10827-018-0693-9) for details.

If you want a `method_` switch, I would suggest to use an `enum` instead of a plain `int`.

models/glif_lif_asc.h
```diff
@@ -0,0 +1,285 @@
...
+    Parameters_();
+
+    void get( DictionaryDatum& ) const;
+    void set( const DictionaryDatum& );
+  };
+
+
+  struct State_
+  {
+    double V_m_;                       // membrane potential in mV
```
heplesser: In almost all other linear neuron models, we work internally with `U = V_m - E_rev`. This makes the numerics slightly simpler and avoids any problems when changing `E_rev`. The `State::get()/set()` methods convert between `U` and `V_m`.

models/glif_lif_asc.cpp
```diff
@@ -0,0 +1,351 @@
...
+ * Update and spike handling functions
+ * ---------------------------------------------------------------- */
+
+void
+nest::glif_lif_asc::update( Time const& origin, const long from, const long to )
+{
+  const double dt = Time::get_resolution().get_ms();
+  double v_old = S_.V_m_;
+  double tau = P_.G_ / P_.C_m_;
+  double exp_tau = std::exp( -dt * tau );
```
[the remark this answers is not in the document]
binghuangcai: Solved. No this problem in psc and cond models.

models/glif_lif_asc.h
```diff
@@ -0,0 +1,285 @@
...
+  void handle( nest::CurrentEvent& );
+  void handle( nest::DataLoggingRequest& );
+
+  nest::port handles_test_event( nest::SpikeEvent&, nest::port );
+  nest::port handles_test_event( nest::CurrentEvent&, nest::port );
+  nest::port handles_test_event( nest::DataLoggingRequest&, nest::port );
+
+  bool is_off_grid() const // uses off_grid events
+  {
+    return true;
```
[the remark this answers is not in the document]
binghuangcai: The model was designed to be able to emit both grid and precise spike times based on users' setting for <spike_detector>. By default, it emits precise spike time. By setting <precise_times> <params> for <spike_detector> to <False>, it should emit grid spike time. The models do have codes to generate/compute <spike_offset> in the  <update()> function. I have tested this at the beginning of implementing these models.The above function <is_off_grid()> needs to be return <true> to get precise spike times from connected <spike_detector>. Let me know if I am wrong or the Nest has some updates on this. Thanks.

models/glif_lif_asc.h
binghuangcai: <method_> has removed. The models are using linear exact solutions.

models/glif_lif_asc.cpp
```diff
@@ -0,0 +1,351 @@
...
+void
+nest::glif_lif_asc::Parameters_::set( const DictionaryDatum& d )
+{
+  updateValue< double >( d, names::V_th, V_th_ );
+  updateValue< double >( d, names::g, G_ );
+  updateValue< double >( d, names::E_L, E_L_ );
+  updateValue< double >( d, names::C_m, C_m_ );
+  updateValue< double >( d, names::t_ref, t_ref_ );
+  updateValue< double >( d, names::V_reset, V_reset_ );
+  updateValue< std::vector< double > >( d, Name( "asc_init" ), asc_init_ );
```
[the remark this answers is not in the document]
binghuangcai: all glif variable names have been updated into "nest_names". Note: we are working on a new branch and will merge it to this PR branch once it is done.

models/glif_lif_asc.h
binghuangcai: Models have been updated to use relative membrane potential internally. We will update this branch shortly.

binghuangcai:
All the glif models have been combined to one single model for psc-based and cond_based and updated in this branch. The model files are `glif_psc.h`, `glif_psc.cpp`, `glif_cond.h`, `glif_cond.cpp`.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+#include "event.h"
+#include "nest_types.h"
+#include "ring_buffer.h"
+#include "universal_data_logger.h"
+
+#include "dictdatum.h"
+
+/* BeginDocumentation
+Name: glif_psc - Current-based generalized leaky integrate and fire (GLIF)
+models
```
heplesser: I would suggest the singular `model` here for consistency, even though this model come with five variants.

@jessica-mitchell What do you think from the documentation perspective?

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+
+Description:
+
+glif_psc is implementations of five generalized leaky integrate and
```
heplesser: "is implementations of" -> "provides"

"integrate and fire" -> "integrate-and-fire"

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+fire (GLIF) models [1] with alpha-function shaped synaptic currents.
+Incoming spike events induce a post-synaptic change of current modeled
+by an alpha function. The alpha function is normalized such that an event of
+weight 1.0 results in a peak current of 1 pA at t = tau_syn. On the postsynapic
```
heplesser: While the text here is based on `iaf_psc_[alpha,exp]_multisynapse`, it is not really good in the original. So I would suggest to replace everything from "On the postsynaptic ..." to " ... in the connectors." with the following paragraph

"By default, glif_psc has a single synapse that is accessible through receptor_port 1. An arbitrary number of synapses with different time constants can be configured by setting the desired time constants as tau_syn array. The resulting synapses are addressed through receptor_port 1, 2, 3, ...."

@jessica-mitchell I am not entirely happy with my text either, any suggestions? Once we have a good text, we also need to update `iaf_psc_[alpha,exp]_multisynapse` correspondingly.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+                    is the time constant of the voltage-dependent component
+                    of the threshold in 1/ms (bv in Equation (4) in [1]).
+tau_syn    double vector - Rise time constants of the synaptic alpha function
+                           in ms.
+glif_model string - glif model type:
+                    "lif"/"glif_lif"/"1" - GLIF model 1,
+                    "lif_r"/"glif_lif_r"/"2" - GLIF model 2,
+                    "lif_asc"/"glif_lif_asc"/"3" - GLIF model 3,
+                    "lif_r_asc"/"glif_lif_r_asc"/"4" - GLIF model 4, or
+                    "lif_r_asc_a"/"glif_lif_r_asc_a"/"5" - GLIF model 5.
```
heplesser: I would suggest to allow only the "lif_..." names. The "glif_" is redundant and numbers are uninformative and will make simulation scripts difficult to read in the long term. I am also wondering if we could drop "lif" from the names, since it is redundant.

Maybe we should consider a different scheme: How about having a parameter "mechanism" (I am open for better suggestions) which is an array of strings, expressing which mechanisms to incorporate:

- 1: `/mechanisms []`
- 2: `/mechanisms [ /spike_dependent_threshold ]`
- 3: `/mechanisms [ /after_spike_currents  ]`
- 4: `/mechanisms [ /spike_dependent_threshold /after_spike_currents  ]`
- 5: `/mechanisms [ /spike_dependent_threshold /after_spike_currents /adapting_threshold ]`
 
This is more verbose, but I think it will make reading any script using the models much more comprehensible a few years down the line and will also avoid user errors by confusing, e.g., model 3 and 4. Internally in the model this could be mapped to model numbers 1-5 and unsupported combinations could be forbidden.

For an example of how to read out such a list of labels, see the code handling `record_to` in [recording_device.cpp](https://github.com/nest/nest-simulator/blob/8adec3cd2bb8025ca07c91f8e281ecb71f13658f/nestkernel/recording_device.cpp#L241); the `/` are in SLI only, in Python this would be strings.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+
+  nest::port handles_test_event( nest::SpikeEvent&, nest::port );
+  nest::port handles_test_event( nest::CurrentEvent&, nest::port );
+  nest::port handles_test_event( nest::DataLoggingRequest&, nest::port );
+
+  void get_status( DictionaryDatum& ) const;
+  void set_status( const DictionaryDatum& );
+
+  typedef std::string model_type;
+  std::map< std::string, long > model_type_lu = { { "lif", 1 },
```
heplesser: This should probably be revised, see comment above. If not, it should be `const` and private.
What is the meaning of `_lu` at the end of the name?

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+  void update( nest::Time const&, const long, const long );
+
+  // The next two classes need to be friends to access the State_ class/member
+  friend class nest::RecordablesMap< glif_psc >;
+  friend class nest::UniversalDataLogger< glif_psc >;
+
+
+  struct Parameters_
+  {
+    double G_;               // membrane conductance in nS
```
heplesser: Please make all comments doxygen comments by starting them with `//!<`

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+    std::vector< double > asc_init_; // initial values of ASCurrents_ in pA
+    std::vector< double > k_;        // predefined time scale in 1/ms
+    std::vector< double > asc_amps_; // in pA
+    std::vector< double > r_;        // coefficient
+    std::vector< double > tau_syn_;  // synaptic port time constants in ms
+
+    // boolean flag which indicates whether the neuron has connections
+    bool has_connections_;
+
+    model_type glif_model_;
```
heplesser: Doxygen comment missing

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+The following parameters can be set in the status dictionary.
+
+V_m        double - Membrane potential in mV
+V_th       double - Instantaneous threshold in mV.
+g          double - Membrane conductance in nS.
+E_L        double - Resting membrane potential in mV.
+C_m        double - Capacitance of the membrane in pF.
+t_ref      double - Duration of refractory time in ms.
+V_reset    double - Reset potential of the membrane in mV (glif 1-3).
+a_spike    double - Threshold addition following spike in mV.
```
heplesser: Which parameter is this in Table 1 in [1]? Should be explicit for all parameters.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+
+    void get( DictionaryDatum& ) const;
+    double set( const DictionaryDatum& );
+  };
+
+
+  struct State_
+  {
+    double U_;                         // relative membrane potential in mV
+    std::vector< double > ASCurrents_; // after-spike currents in pA
```
heplesser: I would suggest to collect all vectors in one place after the scalar variables.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "glif_psc.h"
+
+// C++ includes:
+#include <limits>
+#include <iostream>
```
heplesser: Please keep includes in alphabetical order within each block.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+namespace nest
+{
+// Override the create() method with one call to RecordablesMap::insert_()
+// for each quantity to be recorded.
+template <>
+void
+RecordablesMap< nest::glif_psc >::create()
+{
+  insert_( names::V_m, &nest::glif_psc::get_V_m_ );
```
heplesser: Should the total threshold (Theta_inf + Theta_s + Theta_v) or Theta_s and Theta_v be recordable as well?

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+}
+
+/* ----------------------------------------------------------------
+ * Default constructors defining default parameters and state
+ * ---------------------------------------------------------------- */
+
+nest::glif_psc::Parameters_::Parameters_()
+  : G_( 9.43 )                                           // in nS
+  , E_L_( -78.85 )                                       // in mV
+  , th_inf_( 27.17 )                                     // in mv, rel to E_L_, 51.68 - E_L_
```
heplesser: You can make this explicit as
```C++
, th_inf_( 51.68 - E_L_ )
```
since `E_L_` comes before `th_inf_` in the initializer list.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  , C_m_( 58.72 )                                        // in pF
+  , t_ref_( 3.75 )                                       // in ms
+  , V_reset_( 0.0 )                                      // in mV, rel to E_L_, -78.85 - E_L_
```
heplesser: See comment on `th_inf_`

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  , a_spike_( 0.37 )                                     // in mV
+  , b_spike_( 0.009 )                                    // in 1/ms
+  , voltage_reset_a_( 0.20 )                             // deterministic
```
heplesser: "deterministic" gives no meaning here.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  , voltage_reset_b_( 18.51 )                            // in mV
+  , a_voltage_( 0.005 )                                  // in 1/ms
+  , b_voltage_( 0.09 )                                   // in 1/ms
+  , asc_init_( std::vector< double >( 2, 0.0 ) )         // in pA
+  , k_( std::vector< double >{ 0.003, 0.1 } )            // in 1/ms
+  , asc_amps_( std::vector< double >{ -9.18, -198.94 } ) // in pA
+  , r_( std::vector< double >( 2, 1.0 ) )                // deterministic
```
heplesser: See above

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  updateValue< std::vector< double > >( d, names::k, k_ );
+  updateValue< std::vector< double > >( d, names::asc_amps, asc_amps_ );
+  updateValue< std::vector< double > >( d, names::r, r_ );
+
+  updateValue< std::vector< double > >( d, names::tau_syn, tau_syn_ );
+  updateValue< model_type >( d, names::glif_model, glif_model_ );
+
+  // check after ASC parameters size
+  if ( not(
+         ( asc_init_.size() == k_.size() ) && ( k_.size() == asc_amps_.size() ) && ( asc_amps_.size() == r_.size() ) ) )
```
heplesser: Please use `and` and `or` instead of `&&` and `||` for boolean operators.

I also think the code would be clearer if you chose one value as reference and then compared all others to it, e.g.,

```c++
const size_t asc_size = asc_init_.size();
if ( not ( k_.size() == asc_size and asc_amps._size() == asc_size and ... ) )
```

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+    }
+    for ( size_t i = 0; i < tau_syn_.size(); ++i )
+    {
+      if ( tau_syn_[ i ] <= 0 )
+      {
+        throw BadProperty( "All synaptic time constants must be strictly positive." );
+      }
+    }
+  }
+  return delta_EL;
```
heplesser: Insert empty line above.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  if ( updateValue< double >( d, names::V_m, U_ ) )
+  {
+    U_ -= p.E_L_;
+  }
+  else
+  {
+    U_ -= delta_EL;
+  }
+
+  updateValue< std::vector< double > >( d, names::ASCurrents, ASCurrents_ );
```
heplesser: Since you allow setting the ASCurrents, would it also make sense to allow setting Theta_s and Theta_v?
Should there be a protection against setting ASC or Theta for model variants not supporting them?

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  updateValue< double >( d, names::a_voltage, a_voltage_ );
+  updateValue< double >( d, names::b_voltage, b_voltage_ );
+
+  updateValue< std::vector< double > >( d, names::asc_init, asc_init_ );
+  updateValue< std::vector< double > >( d, names::k, k_ );
+  updateValue< std::vector< double > >( d, names::asc_amps, asc_amps_ );
+  updateValue< std::vector< double > >( d, names::r, r_ );
+
+  updateValue< std::vector< double > >( d, names::tau_syn, tau_syn_ );
+  updateValue< model_type >( d, names::glif_model, glif_model_ );
```
heplesser: Should there be any restriction on changing the model type? E.g., you can only change the model type before the simulation starts?

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+}
+
+/* ----------------------------------------------------------------
+ * Update and spike handling functions
+ * ---------------------------------------------------------------- */
+
+void
+nest::glif_psc::update( Time const& origin, const long from, const long to )
+{
+  // glif_lif_r_asc_a
```
heplesser: This comment makes no sense here.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  const double dt = Time::get_resolution().get_ms();
+
+  // get model type
```
heplesser: This should be done only once when the model type is changed.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  std::string model_str = P_.glif_model_;
+  std::transform( model_str.begin(), model_str.end(), model_str.begin(), ::tolower );
```
heplesser: We should not support carelessness on part of the user. The user should use exact spelling, so no automatic "tolower" or similar. But see my suggestion on handling this in a different way in a separate comment in the h-file.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+  if ( nest::glif_psc::model_type_lu.find( model_str ) == nest::glif_psc::model_type_lu.end() )
+  {
+    throw BadProperty( "Bad glif model type string." );
+  }
+  long model_type = nest::glif_psc::model_type_lu[ model_str ];
```
heplesser: `model_type` should be a member of Parameters, and I would strongly suggest an enumeration.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+  double v_old = S_.U_;
+  double spike_component = 0.0;
+  double voltage_component = 0.0;
+  double th_old = S_.threshold_;
+
+  for ( long lag = from; lag < to; ++lag )
+  {
+
+    // update threshold via exact solution of dynamics of spike component of
```
heplesser: Please format comment more tidily.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+    // threshold
+    // exact solution of dynamics of spike component of threshold for glif2/4/5
+    // models with "R"
+    if ( model_type == 2 || model_type == 4 || model_type == 5 )
```
heplesser: I would suggest to introduce a flag `has_theta_s_` that is set for those model variants that have it, so this can be just
```c++
if ( has_theta_spike_ )
```
That is faster and more expressive.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+    {
+      spike_component = V_.last_spike_ * std::exp( -P_.b_spike_ * dt );
```
heplesser: Exponentials are expensive and the value is constant, so `std::exp( - P_.b_spike_ * dt )` should be pre-computed in `calibrate()` and stored in `V_`.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+    }
+    V_.last_spike_ = spike_component;
```
heplesser: Why not simply (see comment in h-file for change in variable names and assignment to S_ vs V_)
```c++
if ( has_theta_spike_ )
{
  S_.theta_spike_ *= V_.theta_spike_decay_rate_;
}
```

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+      // While neuron is in refractory period count-down in time steps (since dt
+      // may change while in refractory) while holding the voltage at last peak.
+      V_.t_ref_remaining_ -= dt;
+      if ( V_.t_ref_remaining_ <= 0.0 )
+      {
+        // Neuron has left refractory period, reset voltage and after-spike
+        // current
+
+        // Reset ASC_currentsfor glif3/4/5 models with "ASC"
+        if ( model_type == 3 || model_type == 4 || model_type == 5 )
```
heplesser: See above, I suggest a `has_asc_` flag.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+        {
+          for ( std::size_t a = 0; a < S_.ASCurrents_.size(); ++a )
+          {
+            S_.ASCurrents_[ a ] =
+              P_.asc_amps_[ a ] + S_.ASCurrents_[ a ] * P_.r_[ a ] * std::exp( -P_.k_[ a ] * V_.t_ref_total_ );
```
heplesser: Should comment here why you use `V_.t_ref_total_`, it is not obvious. I also don't quite manage to match this to Eqs 3 and 7 in [1], could you explain?

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+          }
+        }
+
+        // Reset voltage
+        if ( model_type == 1 || model_type == 3 )
```
heplesser: See above, consider boolean flag.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+          // rest the global threshold (voltage component of threshold: stay the
+          // same)
+          S_.threshold_ = V_.last_spike_ + V_.last_voltage_ + P_.th_inf_;
+        }
+        // Check if bad reset
+        // TODO: Better way to handle?
+        if ( S_.U_ > S_.threshold_ )
+        {
+          printf(
```
heplesser: Throw an exception if this happens, e.g., `NumericalInstability`.

And can this actually happen? Could it make sense to allow for a minimal numerical error and in that case just set U to threshold?

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+      }
+    }
+    else
+    {
+      // Integrate voltage and currents
+
+      // Calculate new ASCurrents value using expoential methods
+      S_.ASCurrents_sum_ = 0.0;
+      // for glif3/4/5 models with "ASC"
+      if ( model_type == 3 || model_type == 4 || model_type == 5 )
```
heplesser: Boolean flag.

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+      {
+        for ( std::size_t a = 0; a < S_.ASCurrents_.size(); ++a )
+        {
+          S_.ASCurrents_sum_ += S_.ASCurrents_[ a ];
```
heplesser: Why do you sum before updating?

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+          S_.ASCurrents_[ a ] = S_.ASCurrents_[ a ] * std::exp( -P_.k_[ a ] * dt );
```
heplesser: Precompute exponential in `calibrate()` and store in `V_`

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+      S_.I_syn_ = 0.0;
+      for ( size_t i = 0; i < P_.n_receptors_(); i++ )
+      {
+        S_.U_ += V_.P31_[ i ] * S_.y1_[ i ] + V_.P32_[ i ] * S_.y2_[ i ];
+        S_.I_syn_ += S_.y2_[ i ];
+      }
+
+      // Calculate exact voltage component of the threshold for glif5 model with
+      // "A"
+      if ( model_type == 5 )
```
heplesser: Boolean flag like `has_adapting_threshold_`

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+      {
+        double beta = ( S_.I_ + S_.ASCurrents_sum_ ) / P_.G_;
+        double phi = P_.a_voltage_ / ( P_.b_voltage_ - P_.G_ / P_.C_m_ );
```
heplesser: Both doubles can be `const`

models/glif_psc.cpp
```diff
@@ -0,0 +1,578 @@
...
+        voltage_component = phi * ( v_old - beta ) * std::exp( -P_.G_ * dt / P_.C_m_ )
```
heplesser: Why need for `voltage_compontent`? Why not update `S_.theta_voltage_` directly (see comment in h file for change in names).

Also, precompute all exponentials and divisions if at all possible. Can terms be rearranged for simpliciation and numerical stability?

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+    nest::RingBuffer currents_;              //!< Buffer incoming currents through delay,
+
+    //! Logger for all analog data
+    nest::UniversalDataLogger< glif_psc > logger_;
+  };
+
+  struct Variables_
+  {
+    double t_ref_remaining_; // counter during refractory period, in ms
+    double t_ref_total_;     // total time of refractory period, in ms
```
heplesser: `t_ref_ ` is already in `P_`, why repeat it here? It would make sense if it was the refractory time rounded to steps and the expressed in ms again, see comment on t_ref_remaining_.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+    std::vector< nest::RingBuffer > spikes_; //!< Buffer incoming spikes through delay, as sum
+    nest::RingBuffer currents_;              //!< Buffer incoming currents through delay,
+
+    //! Logger for all analog data
+    nest::UniversalDataLogger< glif_psc > logger_;
+  };
+
+  struct Variables_
+  {
+    double t_ref_remaining_; // counter during refractory period, in ms
```
heplesser: Why is this in ms and not in steps, since the model places it spikes on the time grid? Should be steps.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+    double t_ref_total_;     // total time of refractory period, in ms
+    double last_spike_;      // threshold spike component in mV
+    double last_voltage_;    // threshold voltage component in mV
```
heplesser: To bring code closer to the paper, I would suggest renaming these two to `theta_s_` and `theta_v_`, for example. They should be part of `S_`, since they are state variables; roughly speaking, any variable that appears in an equation defining the model on the left-hand side of an ODE is a state variable and should go into `S_`.

models/glif_psc.h
```diff
@@ -0,0 +1,354 @@
...
+    std::vector< double > P31_; // synaptic/membrane current evolution parameter
+    std::vector< double > P32_; // synaptic/membrane current evolution parameter
+
+    /** Amplitude of the synaptic current.
+              This value is chosen such that a post-synaptic current with
+              weight one has an amplitude of 1 pA.
+    */
+    std::vector< double > PSCInitialValues_;
+
+    unsigned int receptor_types_size_;
```
heplesser: Wouldn't `num_receptors_` or `num_synapses_` be a better name?

models/glif_psc.cpp
binghuangcai: ASC checking has been added.

models/glif_psc.cpp
binghuangcai: This line is for ASC rest, i.e., Eq. 7 in [1] with f_j = exp ( - k_j * delta_t) (see texts right below Eq. (6) in the Supplementary Materials of [1]). delta_t is the total time of spike cut length (refractory period) as the after spike currents also decay during such a period. V_.t_ref_total_ in codes is delta_t.

models/glif_psc.cpp
binghuangcai: The ASC sum is used to compute the Vm value of next step for the dynamics. The ASC updating is in the same manner to compute the ASC for the next step.

models/glif_psc.h
binghuangcai: I have created three bool parameters: spike_dependent_threshold, after_spike_currents and adapting_threshold, for model mechanisms setting. This should be much straightforward and more verbose for users. Checking of combination of these three parameters has also been added for only supporting the five glif models. Other combinations of input will throw BadProperty. Updates will be pushed once done with other comments.

models/glif_psc.h
binghuangcai: converted to steps

models/glif_psc.h
binghuangcai: converted to steps. t_ref_remaining_ has been removed.

models/glif_psc.h
binghuangcai: <S_.threshold_spike_> and <S_.threshold_voltage_> have been added and replace related original variables.

models/glif_psc.h
binghuangcai: It is a redundant variable, has been removed.

models/glif_psc.cpp
binghuangcai: See comments on new state variables for spike and voltage components of threshold above. All exponential and divisions have been precomputed in <calibrate()>

models/glif_psc.cpp
binghuangcai: This problem not longer exists after reorganized the reset part of code.

models/glif_psc.cpp
binghuangcai: thta_s and theta_v are settable in the codes now.

pynest/examples/glif_psc_neuron.py
```diff
@@ -0,0 +1,322 @@
...
+# with units being converted from SI units( i.e., V, S( 1 / Ohm ), F, s, A ) to
+# NEST used units( i.e., mV, nS( 1 / GOhm ), pF, ms, pA ) and values being
+# rounded to appropriate digits for simplification.
+# Note that the rounded values of the Level 5 GLIF model of such a cell are
+# the default parameter values of ``glif_psc`` neuron.
+# Each function includes effected parameters for each level of GLIF model.
+# Also, the initial membrane potential V_m is set to E_L.
+
+
+def create_lif_psc(mechs, syn_tau):
```
heplesser: The `mechs` parameter makes no sense: Since this function must create a LIF model, you require that `mechs==[False, False, False]`. If it is not, you get a different model back, which would be very confusing. The same holds for the other `create_...()` functions.

Also, it seems as if you set almost every parameter here. Is that strictly necessary? The example should introduce the user to the model, so all the code setting parameters may distract. Would it be possible to just use the default parameters?

Maybe then one could drop the `create_...()` functions entirely?

pynest/examples/glif_psc_neuron.py
```diff
@@ -0,0 +1,322 @@
...
+# ``glif_psc`` node.The node is created by calling the above defined function
+# for the picked model.We store the returned handle in variable for later
+# reference.
+
+glif_model = "lif"  # "lif", "lif_r", "lif_asc", "lif_r_asc", or "lif_r_asc_a"
+
+if glif_model not in model_mech_dict.keys():
+    glif_model = 'lif'
+
+mechs = model_mech_dict[glif_model]
```
heplesser: I think the example would be more instructive if you created one neuron of each variant and the showed how they react to the same input, i.e., something like
```python
n_lif = nest.Create('glif_psc', 'spike_dependent_threshold': False, 'after_spike_currents': False, 'adapting_threshold': False)
n_lif_r = nest.Create('glif_psc', 'spike_dependent_threshold': True, 'after_spike_currents': False, 'adapting_threshold': False)
...
neurons = n_lif + n_lif_r + ...
```

pynest/examples/glif_psc_neuron.py
```diff
@@ -0,0 +1,322 @@
...
+###############################################################################
+# We pick one of the five GLIF model levels to be tested from
+# ("lif", "lif_r", "lif_asc", "lif_r_asc", "lif_r_asc_a").
+# Test "lif" model if name other than the five above names is set.
+# For the picked GLIF model, model mechanisms are retrieved to create
+# ``glif_psc`` node.The node is created by calling the above defined function
+# for the picked model.We store the returned handle in variable for later
+# reference.
+
+glif_model = "lif"  # "lif", "lif_r", "lif_asc", "lif_r_asc", or "lif_r_asc_a"
```
heplesser: The example should not require the user to make changes. See also comment below.

models/glif_psc.h
```diff
@@ -0,0 +1,404 @@
...
+adapting_threshold        bool - flag whether the neuron has a voltage dependent
+                                 threshold component.
+
+Remarks:
+
+GLIF model mechanism setting is based on three parameters
+(spike_dependent_threshold, after_spike_currents, adapting_threshold).
+The settings of these three parameters for the five GLIF models are list below.
+Other combinations of these parameters will not be supported.
+GLIF Model 1 - (False, False, False),
```
heplesser: Repeat the LIF, LIF_R, ... labels in this table.

models/glif_psc.cpp
```diff
@@ -0,0 +1,632 @@
...
+  if ( not( ( not has_theta_spike_ and not has_asc_ and not has_theta_voltage_ ) or // glif1
+         ( has_theta_spike_ and not has_asc_ and not has_theta_voltage_ ) or        // glif2
+         ( not has_theta_spike_ and has_asc_ and not has_theta_voltage_ ) or        // glif3
+         ( has_theta_spike_ and has_asc_ and not has_theta_voltage_ ) or            // glif4
+         ( has_theta_spike_ and has_asc_ and has_theta_voltage_ )                   // glif5
+         ) )
+  {
+    throw BadProperty(
+      "Incorrect model mechanism combination setting."
+      "See documents for setting of model mechanism parameters:"
```
heplesser: documents -> documentation

models/glif_psc.cpp
heplesser: Ok. You sum the currents at the beginning of the time step and take that value as constant throughout the time step; and then you update the current values to the value at the end of the time step, ready for the next time step. You may want to comment this. Strictly speaking, the ASC should be included in the exact integration scheme as well, but that would be challenging given that the number of currents can be changed by the user. Are the time constants reasonably long to limit integration error?

binghuangcai:
@heplesser  Thanks. I will work on your comments on the current based glif. Also, I am still working on an issue of the conductance-based glif. I will push the update and let you know once both are done.

binghuangcai:
> @binghuangcai Thank you for your efforts, the code looks much better now (I have not checked the `cond` version in detail yet). I think the example could be improved to make it even more instructive for users. Test [test_refractory.py](https://github.com/nest/nest-simulator/blob/74bc23ac7e67e77dfdd5251daca8a30a2796d282/pynest/nest/tests/test_refractory.py) currently fails. This test drives the neuron with a strong input current to provoke to spikes that are just separated by the refractory time of the neuron. Could you try to find out why it does not work for glif? The test setup has ways to provide extra parameters for some neurons if needed. Only if the models cannot be made to work at all for this test should they be put on the exclusion list.

@heplesser I just run the test_refractory.py locally for the glif_psc model. It passed the test. Can you point me to the log of the failure of such a test? It seems not in Travis CI log. Thanks.

pynest/examples/glif_psc_neuron.py
binghuangcai: @heplesser The reason for the `create_...()` function is that users can see what parameters are associated with the selected level of models. All the parameters are repeating the default parameters in the model. Do you think this is necessary to help user get familiar with the parameter set of each level of glif model? If so, I will remove such function and change the example to the one you mentioned in the next comment.

binghuangcai: Examples have been updated.

binghuangcai:
@heplesser For ASC: a factor has been added to use the exact ASC sum within the time step for the voltage dynamics computation. Comments on how the ASC sum is computed and used have been added in the code.

models/glif_psc.cpp
```diff
@@ -0,0 +1,644 @@
...
+  , th_spike_decay_( 0.009 ) // in 1/ms
+  , voltage_reset_fraction_( 0.20 )
+  , voltage_reset_add_( 18.51 )                          // in mV
+  , th_voltage_index_( 0.005 )                           // in 1/ms
+  , th_voltage_decay_( 0.09 )                            // in 1/ms
+  , asc_init_( std::vector< double >( 2, 0.0 ) )         // in pA
+  , asc_decay_( std::vector< double >{ 0.003, 0.1 } )    // in 1/ms
+  , asc_amps_( std::vector< double >{ -9.18, -198.94 } ) // in pA
+  , asc_r_( std::vector< double >( 2, 1.0 ) )
+  , tau_syn_( 1, 2.0 ) // in ms
```
clinssen: I'm a little confused over why there are two synapses by default. Is this just by means of example?

Unrelated to that, I'm surprised that an std::vector can be initialised this way. Could you make it `tau_syn_( std::vector< double >(1, 2) )` for consistency?

binghuangcai:
> glif_cond.h reads: "We recommend the setting of the parameters of these three models to follow the
> condition of (E_L + voltage_reset_fraction * ( V_th - E_L ) - voltage_reset_add)"
> 
> whereas glif_psc.h reads: "We recommend the setting of the parameters of these three models to follow the
> condition of (E_L + voltage_reset_fraction * ( V_th - E_L ) + voltage_reset_add)"
> 
> Note the difference in sign in front of voltage_reset_add. Is this correct? Can this condition be automatically checked when setting (updating) the parameter values, for example in an assertion?

@clinssen I have corrected the typo in the glif_cond.h from - to +. For your question, we have a discussion with @heplesser in an email stream a while ago. This inequality is a recommending condition to set the the related parameters to avoid voltage resetting to above the threshold leading the neuron to spike until the end of the simulation. @heplesser suggested to put the note in documentation rather than using assertion to terminate the simulation since there is no absolute rule that neurons need to reset below threshold.

models/glif_cond.h
```diff
@@ -0,0 +1,488 @@
...
+               and after-spike currents (LIF_R_ASC) model,
+GLIF Model 5 - Leaky integrate and fire with biologically defined reset rules,
+               after-spike currents and a voltage dependent threshold
+               (LIF_R_ASC_A) model.
+
+Parameters:
+
+The following parameters can be set in the status dictionary.
+
+V_m        double - Membrane potential in mV
```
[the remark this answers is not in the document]
binghuangcai: It is absolute value. I have added '(absolute value)' at the end of this line.

models/glif_psc.cpp
binghuangcai: For glif_psc, it has one default synapse with time constant as 2.0ms. For glif_cond, it has two default synapses, one as excitation and one as inhibition. For both models, users can set as many synapses as they needed. I have changed the `tau_syn_` stating.

models/glif_cond.cpp
```diff
@@ -0,0 +1,838 @@
...
+  , E_rev_( std::vector< double >{ 0.0, -85.0 } )        // in mV
+  , has_connections_( false )
+  , has_theta_spike_( false )
+  , has_asc_( false )
+  , has_theta_voltage_( false )
+{
+}
+
+nest::glif_cond::State_::State_( const Parameters_& p )
+  : threshold_( -51.68 - p.E_L_ ) // in mV
```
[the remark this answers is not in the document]
binghuangcai: The state variable `threshold_` to be initialize as instantaneous threshold parameter `p.th_inf_`. I have change the value to `p.th_inf_`. Would this look reasonable to you?

models/glif_psc.h
```diff
@@ -0,0 +1,405 @@
...
+
+  nest::port handles_test_event( nest::SpikeEvent&, nest::port );
+  nest::port handles_test_event( nest::CurrentEvent&, nest::port );
+  nest::port handles_test_event( nest::DataLoggingRequest&, nest::port );
+
+  void get_status( DictionaryDatum& ) const;
+  void set_status( const DictionaryDatum& );
+
+private:
+  //! Reset parameters and state of neuron.
```
heplesser: This line documents a function that does not exist, please remove.

models/glif_cond.h
```diff
@@ -0,0 +1,488 @@
...
+
+  nest::port handles_test_event( nest::SpikeEvent&, nest::port );
+  nest::port handles_test_event( nest::CurrentEvent&, nest::port );
+  nest::port handles_test_event( nest::DataLoggingRequest&, nest::port );
+
+  void get_status( DictionaryDatum& ) const;
+  void set_status( const DictionaryDatum& );
+
+private:
+  //! Reset parameters and state of neuron.
```
heplesser: This line comments a method that does not exist, please remove it.

models/glif_cond.h
```diff
@@ -0,0 +1,488 @@
...
+    Parameters_();
+
+    void get( DictionaryDatum& ) const;
+    double set( const DictionaryDatum& );
+  };
+
+
+  struct State_
+  {
+
```
heplesser: Please remove this empty line.

pynest/nest/tests/test_all.py
```diff
@@ -46,6 +46,7 @@
 from . import test_events
 from . import test_facetshw_stdp
 from . import test_getconnections
+from . import test_glif_psc
```
heplesser: Why is test_glif_cond not imported in the same way?

pynest/nest/tests/test_all.py
```diff
@@ -101,6 +102,7 @@ def suite():
     suite.addTest(test_events.suite())
     suite.addTest(test_facetshw_stdp.suite())
     suite.addTest(test_getconnections.suite())
+    suite.addTest(test_glif_psc.suite())
```
heplesser: Why is test_glif_cond.suite() not added in the same way?

heplesser:
@jessica-mitchell @lekshmideepu @terhorstd Could you help me and @binghuangcai to fix a final problem with this PR. The situation is as follows:

1. All tests pass, except the OSX test. That fails because `Python.h` is not found when building PyNEST, most likely due to some setup problem.
1. Master has not been merged into the `feature/glif_aibs` branch for a while, I assume that with current setup files the OSX problems would disappear.
1. I have merged current master into the `feature/glif_aibs` branch on my machine and pushed to my repo on Github (https://github.com/heplesser/nest-simulator/tree/glif_aibs_hep). The merge when through without conflicts and the code builds and tests well on my Mac.
1. Testing on Travis (https://travis-ci.org/heplesser/nest-simulator/jobs/615539879) fails then due to PEP8 problems in some files:
```
doc/examples/hpc_benchmark.py
doc/examples/rate_neuron_dm.py
doc/examples/structural_plasticity.py
doc/examples/testiaf.py
doc/tutorials/topology/hill_tononi_Vp.py
extras/help_generator/conf.py
pynest/examples/hpc_benchmark.py
pynest/examples/rate_neuron_dm.py
pynest/examples/structural_plasticity.py
pynest/examples/testiaf.py
pynest/nest/lib/hl_api_connections.py
pynest/nest/ll_api.py
pynest/nest/tests/test_regression_issue-1034.py
pynest/nest/voltage_trace.py
```
1. Problems seem to be too many empty lines and documentation comment blocks that are not indented, e.g., inside dictionaries [as here](https://github.com/nest/nest-simulator/blob/8f21d4cef44ae4a1b96566b280e07d8c199b55f3/doc/examples/hpc_benchmark.py#L163). Note that these problems also exist in Master (see also #1328).
1. Concerning fixes for this:
    1. Is it safe to indent the doc-comment-blocks according to PEP8, or do they need to stay in column 1 to make our doc generator work?
    1. One cannot use `pep8ify` to fix the files, since that would destroy carefully crafted comments such as [here](https://github.com/nest/nest-simulator/blob/8f21d4cef44ae4a1b96566b280e07d8c199b55f3/doc/examples/hpc_benchmark.py#L94).

Could you advise how to proceed?

stinebuu:
@heplesser We have fixed the OSX missing `Python.h` file with #1342. That is now merged into master, so merging master into this branch should fix that problem. The PEP8 problems I don't know about.

heplesser:
@binghuangcai Could you merge the master branch into your branch? This should work without conflict and after that all tests should pass, see [Travis results for my fork of your branch + master](https://travis-ci.org/heplesser/nest-simulator/builds/616296487?utm_medium=notification) (I merged master per efac6d2d7).

binghuangcai:
@heplesser I have merged master into feature/glif_aibs branch. All the tests have passed. Thank you very much.

heplesser:
@binghuangcai @kaeldai And then it is merged. Thank you and your colleagues at the Allen Institute for your contribution!

PR #1127, opened 2019-02-19
Fix broken PyNEST, remove traces of ResumeSimulation

stinebuu:
PRs #1094 and #1058 have been worked on simultaneously. With #1058 we provide explicit indices to all  PyNEST functions in `__all__`, but as #1094 removed `ResumeSimulation`, PyNEST now breaks because `ResumeSimulation` is not found.


jougs:
Awesome! Many thanks for catching, analyzing and fixing this so quickly. :dancing_men:

PR #1133, opened 2019-02-23
Fix code formatting issue introduced by merging #1060

heplesser:


heplesser:
Merging without further review as it only fixes a minor code-formatting issue introduced by merging #1060.

PR #1137, opened 2019-02-27
Fix postsynaptic trace edge case

### LINKED ISSUE #1034: `get_K_value()` can return 0 erroneously in edge case
To compute the postsynaptic trace, the function `Archiving_Node::get_K_value(double t)` finds the last known spike in the history, and decays exponentially from that point on (https://github.com/nest/nest-simulator/blob/0c993abc963c709e7632b9cb5fc701d155d8ab40/nestkernel/archiving_node.cpp#L98).

An edge case can arise when presynaptic spike times precisely coincide with postsynaptic spike times. To replicate the issue, generate a list of pre and a list of post spike times, e.g.:

    pre_st = [  2.   5.   7.   8.  10.  11.  15.  17.  20.  21.  22.  23.  26.  28.]
    post_st = [  3.   7.   8.  10.  12.  13.  14.  16.  17.  18.  19.  20.  21.  22.]

Note that there are no duplicates in each individual list. (I'm not sure if this is something that should be supported--note that the SpikeGenerator supports this behaviour.) 

Hook up an STDP synapse between two parrot neurons which are parroting SpikeGenerators with the above spike times.

Log shows the returned value by `get_K_value()`. The fall-through case (https://github.com/nest/nest-simulator/blob/0c993abc963c709e7632b9cb5fc701d155d8ab40/nestkernel/archiving_node.cpp#L115) is marked by asterisks ("*** 0 ***"):

    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 2) = 0
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 5) = 0.951229
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 7) = 0.860708
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 8) = *** 0 ***
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 10) = 2.59689
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 11) = 2.47023
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 15) = 4.69725
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 17) = *** 0 ***
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 20) = 7.1538
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 21) = *** 0 ***
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 22) = *** 0 ***
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 23) = *** 0 ***
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 26) = 9.40356
    * In Archiving_Node(this = 0x5641c062dac0)::get_K_value(t = 28) = 8.50869

Regardless of whether we take the trace value just before or just after the incoming pre spike, the value returned should not be a flat zero.

If I comment out the increment of the history access counter (https://github.com/nest/nest-simulator/blob/0c993abc963c709e7632b9cb5fc701d155d8ab40/nestkernel/archiving_node.cpp#L174), the issue disappears:

    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 2) = 0
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 5) = 0.951229
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 7) = 0.860708
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 8) = 0.818731
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 10) = 2.59689
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 11) = 2.47023
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 15) = 4.69725
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 17) = 5.15509
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 20) = 7.1538
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 21) = 7.75613
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 22) = 8.32909
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 23) = 8.87411
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 26) = 8.49872
    * In Archiving_Node(this = 0x5601aad15340)::get_K_value(t = 28) = 7.68996

I am not yet sure what the best way to address this would look like.

clinssen:
This fixes #1034.

With props to my fellow sleuth @suku!

Background:
--------

The synapse, when handling an incoming presynaptic spike at time `t_pre_spike` in `send()`, requests postsynaptic history within the interval `t_prev_pre_spike - dendritic_delay` < t ≤ `t_pre_spike - dendritic_delay`. Spikes in this interval will have their access counters incremented. The postsynaptic neuron, in `Archiving_Node::set_spiketime()`, will purge all spikes from the buffer which have been accessed by all connecting synapses.

A minimal reproducing example of the bug has presynaptic spike times at [2, 3, 10] and postsynaptic at [1, 2, 3] (all times in ms). All delays (axonal, dendritic, synaptic) are equal to 1. The failure occurs as follows: The first postsynaptic spike at t = 2 is added to the postsynaptic buffer. The first presynaptic spike at t = 3 causes the synapse to access the postsynaptic buffer. This increments the access counter of the spike at t = 2, marking it for deletion. An extra postsynaptic spike at t = 3 is necessary because the history always needs to have at least two entries if a spike is to be deleted. Next, the postsynaptic neuron spikes at t = 4, which causes the spike to be deleted from the buffer. When another presynaptic spike then arrives at t = 4, the `get_K_value()` function is called on the postsynaptic partner to obtain the trace at t = 3, but the necessary spike at time 2 has now been deleted. Consequently, zero is erroneously returned for the trace value.

How this PR fixes things:
--------

Recall that spike buffer pruning occurs upon firing of the neuron at time `t_sp` in `Archiving_Node::set_spiketime()`. A new condition for pruning is added: only remove a spike if there is another, later spike, that is strictly more than `delay` away from `t_sp` (https://github.com/clinssen/nest-simulator/blob/76a71602edc929116a464d4e8ab34c8cc00b5242/nestkernel/archiving_node.cpp#L210). This ensures that when `get_K_value()` is called to obtain the trace at time `t - delay`, a spike strictly earlier than this is still in the buffer.

To obtain the maximum delay value across all synapses, a new private member variable `max_delay_` is added in `Archiving_Node`, which is updated whenever `Archiving_Node::register_stdp_connection()` is called.

Notes for reviewers:
--------

* The trace value as calculated by `get_K_value()` is cached in the `Archiving_Node` under the new private member variable `trace_`. This is mostly for debugging purposes. If this affects performance, I suggest tying it to the "with-debug" NEST build flag.

* I am not happy about the large tolerance that seems to be necessary when comparing the NEST trace values with the reference values obtained through Python. See: https://github.com/clinssen/nest-simulator/blob/76a71602edc929116a464d4e8ab34c8cc00b5242/pynest/nest/tests/test_post_trace.py#L35


clinssen:
Hi Håkon, thanks for your comments! The new commits should address almost all of these. Just one question: could you share with me the rationale for moving the "testbench-internal" functions you list in your second bullet point out of the TestCase class?

I'm also still getting a failing Travis-CI test due to an issue with `@check_stack`. I don't presently understand why this leads to a failure in this testbench but not others. This is on the agenda for me and Jochen at the start of next week.

hakonsbm:
@clinssen My idea with moving the "testbench-internal" functions is to make the code a bit more structured. What I had in mind was to have a class with the helper functions, which can be initialized with the tolerance and other common arguments (spike times, resolution, etc.). The class can then be used in `test_post_trace`, possibly like this:
```python
def test_post_trace(self):

    # generate spike times, other variables

    ptc = PostTraceClass(init_arguments...)
    self.assertTrue(ptc.compare_post_trace_to_reference())
```

For further tidying, parts of `run_post_trace_test_nest_` and especially `nest_trace_matches_ref_trace` can be refactored into separate methods to cut down on duplicate code and `for`/`if` branches.

With #1058 `nest.check_stack` was moved to `nest.ll_api.check_stack`, which is why Travis complains.

hakonsbm:
@clinssen Take a look at [test_spatial_kernels.py](https://github.com/nest/nest-simulator/blob/master/topology/pynest/tests/test_spatial_kernels.py) for an example of what I had in mind.

jakobj:
maybe @jarsi would be a better reviewer? i am not too familiar with the details of the STDP implementation

clinssen:
Gentlemen, many thanks for your excellent suggestions. The new series of commits should address all of the points you raised. Additionally, I've been able to tighten the (absolute and relative) tolerances from 1E-2 to 1E-3 by making some modifications in how the reference timeseries is generated.

heplesser:
@clinssen 

> * I'm not sure if we need a copyright notice in notebooks (there aren't any in the others)

I would leave the copyright notice out of the notebook for now, in accordance with the existing notebooks.

testsuite/regressiontests/issue-1034.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+import nest
+import numpy as np
+import scipy as sp
+import scipy.stats
+import unittest
+
+
+class PostTraceTester(object):
```
heplesser: Please provide at least brief docstrings for this class and its methods here.

hakonsbm:
@clinssen Thanks for the changes! There are a couple comments in the python test you may have overlooked:
- L131, use parentheses to break lines here
- Can the if-tests on lines L192 and L207 be merged together?

I'll give the thumbs up when these comments are resolved.

heplesser:
@clinssen Thank you for the notebook. Could you add some more text in the final part of the notebook explaining the figures, especially the synaptic trace from NEST?  Is the behaviour shown in the graphs the behaviour before or after this fix has been applied? If it is the behaviour after the fix, then why do we still see this difference?

clinssen:
The latest commits should address all of your comments. @hakonsbm: the if-tests cannot be merged, because I'm using the pattern:

    valid = test_if_valid1()
    if not valid:
        valid = test_if_valid2()
    if not valid:
        ...

clinssen:
@heplesser: I posted my previous comment on a stale version of this page; please hang tight!

PR #1142, opened 2019-03-07
Fixed creation of inhomogeneous poisson generator with parameters

### LINKED ISSUE #1140: Segmentation fault in inhomogeneous poisson generator
We occur this issue when we create inhomogeneous poisson generator with parameters. However, it works when we create this stimulator without parameters and then set parameter values.

To view the issue, we implemented two examples. This will crash:
```
nest.Create('inhomogeneous_poisson_generator', params={
  'rate_times': [10.0, 110.0, 210.0], 
  'rate_values': [400.0, 1000.0, 200.0]
})
```

Interestingly, this code works:
```
ipg = nest.Create('inhomogeneous_poisson_generator')
nest.SetStatus(ipg, {
    'rate_times': [10.0, 110.0, 210.0],
    'rate_values': [400.0, 1000.0, 200.0]
})
```

We used ipython (6.2.1 - 6.4.0) for Python 3.6.5 with NEST 2.16.0.

I created the issue for these reasons: 
- We want to ask if anyone also occur the same problem.
- We were desperate to find a suitable documentation as well as an example about it.

hakonsbm:
This PR fixes an issue with the inhomogeneous poisson generator where it could not be created with empty arrays for the `rate_times` and `rate_values` parameters. This is required when creating a node with specified parameters.

Fixes #1140.

### DIFF
--- a/models/inhomogeneous_poisson_generator.cpp
+++ b/models/inhomogeneous_poisson_generator.cpp
@@ -144,15 +144,19 @@ nest::inhomogeneous_poisson_generator::Parameters_::set(
     throw BadProperty( "Rate times and values must be reset together." );
   }
 
-  // if empty parameters given, return here
+  // if neither times or rates are given, return here
   if ( not( times or rates ) )
   {
     return;
   }
 
-  // from here on we are sure the passed arguments are not empty
   const std::vector< double_t > d_times =
     getValue< std::vector< double_t > >( d->lookup( names::rate_times ) );
+
+  if ( d_times.empty() )
+  {
+    return;
+  }
 
   if ( d_times.size() != rate_values_.size() )
   {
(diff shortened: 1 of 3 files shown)


heplesser:
@hakonsbm Thanks for the PR! "Test 7" is a regression test for #1142, isn't it? Does it cover all possibilities that could trigger this segfault? I also wonder if we shouldn't put that test in a separate regressiontest file.

hakonsbm:
@heplesser I agree that it would be more at home in a separate regressiontest file. The test covers the cause of this particular segfault, but implicitly. Calling the SLI `Create` function with parameters triggers setting parameters with empty arrays, here:
https://github.com/nest/nest-simulator/blob/ffb0b4ab637f304b80c3821528d48b5b13cc0c4b/lib/sli/nest-init.sli#L182
I can make a more explicit test as well in the new regressiontest file.

PR #1143, opened 2019-03-07
Implement breathe extension and improve generation/format of model documentation for website

jessica-mitchell:
This pull request implements the sphinx extension [breathe](https://breathe.readthedocs.io/en/latest/index.html), which allows doxygen-generated xml to be extracted and rendered in a restructured text format.
The configuration (conf.py and doc/Doxyfile) is set up so Doxygen only looks at the models directory and only extracts xml.
This allows the model documentation to auto generate (the current state of things on ReadtheDocs is that the model documentation was manually created and static).

How the models are structured:
In the _models/_ folder, I created a file called `group_models.h` that defines the groups neurons, synapses and devices and each model is assigned to one of those groups.
In _doc/models/_  each group if called with a restructured text directive (doxygengroup) to display the content of each group in a separate page.

How the models are formatted:
_breathe_ can interpret some Doxygen mark-up so to fix some inconsistencies and render math nicely,  I went through and included bullets for the parameters, so they appear as lists, included a verbatim rst section for references and formatted them to be consistent and included the doxygen markup around equations. 
 
**Please take a look at the output here** (https://nest-test.readthedocs.io/en/extract_models_breathe/models/index.html).

Some remaining questions:
* Should the models be subdivided further? (E.g. Conductance-based, Current-based neuron models)
 The list for each model type is rather long and only in alphabetical order.
* Most examples in the documentation are in SLI (if there is an example at all).
 Should PyNEST examples be added or replace the SLI examples OR should we simply link pynest/examples to corresponding models? Should all models have examples?


doc/conf.py
```diff
@@ -132,7 +136,11 @@ def __getattr__(cls, name):
 #    'backreferences_dir': False
 # }
 
+read_the_docs_build = os.environ.get('READTHEDOCS', None) == 'True'
+
+if read_the_docs_build:
```
Silmathoron: why only on readthedocs? shouldn't users who want to build it locally also have the models documentation?
I would rather wrap it to prevent errors if doxygen is not installed

doc/models/neurons.rst
```diff
@@ -0,0 +1,8 @@
+Model Documentation - Neurons
+==============================
+
+.. doxygengroup:: Neurons
+   :content-only:
+
+
```
Silmathoron: not really important, but two of the blank lines should be removed

models/aeif_cond_alpha.h
```diff
@@ -84,51 +86,51 @@ This implementation uses the embedded 4th order Runge-Kutta-Fehlberg solver with
 adaptive step size to integrate the differential equation.
 
 The membrane potential is given by the following differential equation:
-C dV/dt= -g_L(V-E_L)+g_L*Delta_T*exp((V-V_T)/Delta_T)-g_e(t)(V-E_e)
-                                                     -g_i(t)(V-E_i)-w +I_e
+@f[ C dV/dt= -g_L(V-E_L)+g_L*Delta_T*exp((V-V_T)/Delta_T)-g_e(t)(V-E_e)
```
Silmathoron: should we switch the equations to full LaTeX style? (``\Delta``) for instance? And possibly fractions, or should we keep it as readable as possible in the raw file?

clinssen: In the NESTML model documentation, we have decided to move towards full LaTeX (as part of reStructuredText). For now, we have not yet addressed converting all the existing documentation, but any new documentation added should conform to this format.

clinssen:
@jessica-mitchell: looks great, cheers!

Minor comment: could you add a few lines of documentation to `group_models.h` so that a user coming across the file will know what it's there for? Can we potentially move this file outside of the main NEST codebase (e.g. to `doc/`)?

models/aeif_cond_alpha.h
Silmathoron: yeah, this sounds good, I don't see why we should trade off documentation readability to favor the raw file

heplesser: I agree, full LaTeX would be great!

models/aeif_cond_alpha.h
```diff
@@ -138,9 +140,13 @@ Sends: SpikeEvent
 
 Receives: SpikeEvent, CurrentEvent, DataLoggingRequest
 
-References: Brette R and Gerstner W (2005) Adaptive Exponential
-            Integrate-and-Fire Model as an Effective Description of Neuronal
-            Activity. J Neurophysiol 94:3637-3642
+References:
+
+\verbatim embed:rst
+.. [1] Brette R and Gerstner W (2005). Adaptive Exponential
```
heplesser: Could we add DOIs with links to the references?

models/aeif_cond_alpha.h
```diff
@@ -84,51 +86,51 @@ This implementation uses the embedded 4th order Runge-Kutta-Fehlberg solver with
...
-                                                     -g_i(t)(V-E_i)-w +I_e
+@f[ C dV/dt= -g_L(V-E_L)+g_L*Delta_T*exp((V-V_T)/Delta_T)-g_e(t)(V-E_e)
+                                                     -g_i(t)(V-E_i)-w +I_e @f]
 
 and
 
-tau_w * dw/dt= a(V-E_L) -W
+@f[ tau_w * dw/dt= a(V-E_L) -W @f]
 
 Parameters:
```
heplesser: In the rendered version, the headings and subheadings don't seem to be handled as headings, but rather as normal text. Is there any way to mark them up as proper headings at different levels?

models/aeif_cond_alpha.h
```diff
@@ -84,51 +86,51 @@ This implementation uses the embedded 4th order Runge-Kutta-Fehlberg solver with
...
 The following parameters can be set in the status dictionary.
 
 Dynamic state variables:
-  V_m        double - Membrane potential in mV
-  g_ex       double - Excitatory synaptic conductance in nS.
-  dg_ex      double - First derivative of g_ex in nS/ms
-  g_in       double - Inhibitory synaptic conductance in nS.
-  dg_in      double - First derivative of g_in in nS/ms.
-  w          double - Spike-adaptation current in pA.
+-  V_m        double - Membrane potential in mV
```
heplesser: Any way to turn this into even nicer tables when rendered?

The `double` data type information here makes little sense. Since almost all users will be working in Python, if we give it at all it should be `float`.

heplesser:
@jessica-mitchell This is a great step forward, thank you! Some comments
- Do we need the group_models.h file, or could that be some other file type, e.g., `.rst` and could we place it in the `doc` folder hierarchy? It is really a doc file, not a code file.
- I think the models should be grouped more.
- Is there a way to get a table of models with short descriptions, so one can get an overview?
- LaTeX equations would be great.
- The documentation still betrays its C++ origin, referring to C++ classes. Is there any way to suppress this?

models/aeif_cond_alpha.h
jougs: +1 for full LaTeX.

Also, I saw that some of the equations are now wrongly rendered. E.g. `tau_max` in the [docs of the correlation_detector`](https://nest-test.readthedocs.io/en/extract_models_breathe/models/devices.html#_CPPv3N4nest20correlation_detectorE) should be "tau<sub>max</sub>" and currently is "tau<sub>m</sub>ax". I thus think we should fix documentation rather soon (either in this PR or very quickly after)

models/aeif_cond_alpha.h
jougs: I think we can more easily address this when generating the parameter tables from NESTML. C.f. https://github.com/nest/nestml/pull/446 for the work-in-progress PR doing that.

jougs:
Many thanks for this really nice PR. I have several general comments:

* I also would like to see a more fine-grained grouping of the models. Neurons could be spiking, binary or rate and there are further subdivisions for those. For the devices, we clearly have recorders and stimulators at least.
* We really need some kind of overview, or a table of contents at least.
* Taking NESTML into the picture, I am not sure if going via doxygen is really the best way to go forward. We probably can generate better RST from the NESTML model description directly, instead of going via a doxagen-style comment in the generated C++ and extracting it from there.

Silmathoron:
@jougs while doxygen may become unnecessary for most neuron (and synapse) models, it is still relevant for the devices, isn't it?

jougs:
@Silmathoron: yes!

jougs:
Regarding the examples, I think that we should remove both the SLI and the PyNEST snippets from the model documentation and rather automatically/dynamically link to all examples that use a specific model in the documentation of that model. If we did this for SLI and PyNEST, we would have a much better coverage while keeping the model documntation on a more conceptual level.

jougs:
@jessica-mitchell, @steffengraber: we will have a NEST hackathon here in Jülich in week 15 of 2019. @abigailm will be joining and might be able to clean up model documentation then (convert equations to pure LaTEX, check references, correct speling, englishify things, ...)

doc/conf.py
steffengraber: Yes, I think these two lines are obsolete.

models/aeif_cond_alpha.h
jessica-mitchell: I've changed the lists into tables now for parameters - (https://nest-test.readthedocs.io/en/extract_models_breathe/models/neurons.html) how does that look? Should the double column be completely removed or replaced with float?

models/aeif_cond_alpha.h
jessica-mitchell: I went through and updated the model formulas to convert them to full latex. Please double check that they are correct.  https://nest-test.readthedocs.io/en/extract_models_breathe/models/neurons.html

models/aeif_cond_alpha.h
```diff
@@ -84,63 +85,87 @@ This implementation uses the embedded 4th order Runge-Kutta-Fehlberg solver with
 adaptive step size to integrate the differential equation.
 
 The membrane potential is given by the following differential equation:
-C dV/dt= -g_L(V-E_L)+g_L*Delta_T*exp((V-V_T)/Delta_T)-g_e(t)(V-E_e)
-                                                     -g_i(t)(V-E_i)-w +I_e
+@f[ C dV/dt= -g_L(V-E_L)+g_L*\Delta_T*\exp((V-V_T)/\Delta_T)-g_e(t)(V-E_e) \\
```
Silmathoron:
```suggestion
@f[ C_m \frac{dV}{dt} = -g_L(V-E_L)+g_L\Delta_T\exp\left(\frac{V-V_{th}}{\Delta_T}\right) - g_e(t)(V-E_e) \\
```

models/aeif_cond_alpha.h
```diff
@@ -84,63 +85,87 @@ This implementation uses the embedded 4th order Runge-Kutta-Fehlberg solver with
...
+                                                     -g_i(t)(V-E_i)-w +I_e @f]
 
 and
 
-tau_w * dw/dt= a(V-E_L) -W
+@f[ \tau_w * dw/dt= a(V-E_L) -W @f]
```
Silmathoron:
```suggestion
@f[ \tau_w \frac{dw}{dt} = a(V-E_L) - w @f]
```

models/aeif_cond_alpha.h
jessica-mitchell: Unfortunately, [breathe](https://github.com/michaeljones/breathe) replaces the heading tags with just emphasized text. e.g. source file:
`## Parameters` 
generated HTML: 
` <p> <em> Parameters: </em> </p>`   
So to make headings stand out we could add bold or italics.

models/aeif_cond_alpha.h
Silmathoron: Thanks for the update! I changed a few things to make fractions more readable and make parameters names in the equation coherent with the C++ counterparts

models/aeif_cond_alpha.h
Silmathoron: is this also true when using rst-style headers? like 

```
title
=====

subtitle
--------

subsubtitle
+++++++++++
```

jessica-mitchell: Yes. I've tried a few things but nothing worked. The XML file still has the heading tags  `<para><heading level="1">Parameters: </heading> </para>` but this is removed in HTML.

jessica-mitchell:
I've now added several more 'groups' to subdivide the models (see the group_models.h). The output currently looks like this https://nest-test.readthedocs.io/en/extract_models_breathe/models/index.html. A couple of things to note that may need discussion/changes 
1)  the Hill Tononi is separated out as its own group (it's the synapse and neuron models together) 
2) From its documentation, I couldn't determine any specific subgroup the Izhikevich model would fit into, so its only tagged with 'Neuron'. 
3) the mat, gif and iaf neurons are all in one group for generalized and leaky integrate and fire neurons 
4) None of the groups are actually children of a parent group. 

Regarding bibliography info:
* All the references should now include a working link for the DOI. There are a couple of books which don't have any links

jessica-mitchell:
I've now updated the data types and cleaned up the organization. Many thanks to @jougs @clinssen and @Silmathoron for all your input and help 
 
Changes for data types were basically as follows:
 double → replace  with appropriate type (ms, Hz, mV etc _or_ real). 
vector of double → list of 'type' (same as above)
int → integer
long → integer
bool → boolean 
vector long → list of integers 


A couple of data types, I was not sure of:
the  _'rate'_ is described slightly differently in different models and I'm wondering if this could be made more consistent:
`lin_rate`/ `threshold_lin_rate `/`sigmoid_rage_gg_1998`/`tanh_rate`: rate is unitless
`sinusoidal_poisson_generator`: `spikes/second`
`ppd_sup_generator`: `Hz`
`gamma_sup_generator`: `s^-1`

In the `correlation_detector`:
 is the parameter `histogram` == real?

In `spike_generator`: 
the `spike_weights` and `spike_mulitplicities` are array and integer array, respectively?

Silmathoron:
Hey @jessica-mitchell!
After a brief discussion, the NESTival committee decided that ``spikes/s`` should be the preferred unit for rates.
From the [``correlation_detector`` documentation](https://nest-simulator.org/helpindex/cc/correlation_detector.html), the entries in ``histogram`` are squared synaptic weights (for which the unit depends on the model...).
For ``spike_generator``, ``spike_weights`` is similarly dependent on the neuron model, so just mention them as "synaptic weights" (optionally specifying that this is model dependent).
Multiplicities are just integers.

Could you also do a ``git pull your_nest_remote_name master`` to clear the file conflicts by making them up-to-date with master?

jessica-mitchell:
@Silmathoron  And what about the rate-based models that state the rate is unitless?

Silmathoron:
Yeah, in that case the rate is indeed unitless (so just ``real``)

doc/models/aeif.rst
```diff
@@ -0,0 +1,5 @@
+Phenomological (AEIF) neuron models
```
Silmathoron:
```suggestion
Adaptive Exponential Integrate and Fire (aEIF) neuron models
```

doc/models/aeif.rst
```diff
@@ -0,0 +1,5 @@
...
+======================================
```
Silmathoron:
```suggestion
============================================================
```

doc/models/conductance.rst
```diff
@@ -0,0 +1,5 @@
+Conductance-based models
```
Silmathoron:
```suggestion
Neuron models with conductance-based synapses
```
We cannot use "conductance-based models" because this is the dedicated expression for HH-like models (see [Scholarpedia article](http://www.scholarpedia.org/article/Conductance-based_models)

doc/models/current-based.rst
```diff
@@ -0,0 +1,5 @@
+Current-based models
```
Silmathoron:
```suggestion
Neuron models with current-based synapses
```
to be consistent with conductance-based synapses

doc/models/hill_tononi.rst
```diff
@@ -0,0 +1,5 @@
+Hill Tononi models
```
Silmathoron:
```suggestion
Hill-Tononi models
```

doc/models/hill_tononi_neuron.rst
```diff
@@ -0,0 +1,6 @@
+Hill and Tononi neuron model
```
Silmathoron:
```suggestion
Hill-Tononi neuron model
```

doc/models/hill_tononi_synapse.rst
```diff
@@ -0,0 +1,5 @@
+Hill and Tononi synapse model
```
Silmathoron:
```suggestion
Hill-Tononi synapse model
```

doc/models/iaf.rst
```diff
@@ -0,0 +1,7 @@
+Integrate and fire neuron models
```
Silmathoron:
```suggestion
Integrate-and-Fire (IF) neuron models
```

doc/models/aeif.rst
jessica-mitchell:
```suggestion
Adpaptive exponential integrate and fire (AEIF) neuron models
```

Just adjusting the capitalization

doc/models/iaf.rst
jessica-mitchell: I think integrate-and-fire is usually abbreviated iaf ?
Also for consistency, I'd make 'fire' lower case

doc/models/aeif.rst
Silmathoron:
```suggestion
Adaptive Exponential Integrate-and-Fire (aEIF) neuron models
```
Unfortunately, though it may look surprising, "aEIF" is the proper way to write it (see e.g. [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3325187/?tool=pmcentrez&report=abstract)).
Similarly, "Integrate-and-Fire" should be uppercase (and I forgot the hyphens) and is abbreviated IF.

doc/models/iaf.rst
Silmathoron: This is the proper way to write and abbreviate it: see [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3325187/?tool=pmcentrez&report=abstract) and above response.
One can drop the uppercase to "fire", though.
And IF is also used for consistency with other IF models (aEIF, GIF, EIF...)

doc/models/aeif.rst
jessica-mitchell: Ok, well it looks to me that integrate-and-fire is written in the literature both ways. If the preference is uppercase, that's fine, but this needs to be consistent in nest (I think another PR in the future is in order for proper naming of things).  As for IF instead IAF, this is a bit confusing because I see IAF in nest a lot, and I dont think I've ever seen IF used.

doc/models/iaf.rst
jessica-mitchell: I do agree we should use the commonly accepted terms, so I don't have a problem committing these changes. We just may need to consider how we use If and IAF in nest.

Silmathoron: you're right... I never payed attention to this issue... IAF definitely looks wrong, but I may have seen IaF used some times... I'm more used to IF though, but we might have to do a proper survey

doc/models/aeif.rst
Silmathoron: I have no real preference for the uppercase in general, I just though it was the normal way to go when we are defining abbreviations. Agreed for the proper naming PR

Silmathoron: @jessica-mitchell so are you OK with my 2nd proposal for the title?

jessica-mitchell: Yes, sorry I think I skipped over this by accident

Silmathoron:
I'm on my phone and cannot find the review button, but I approve this PR ^^"
:+1:

jessica-mitchell:
@clinssen 

The way I build this is with sphinx.
At minimum, I think you'd need to the following installed :
 sphinx 
 breathe                                                                                                                                                                                                                                                                                                                                                                             
 mock                                                                                                                                
 recommonmark 
doxygen                                                                                                                        
 sphinx-rtd-theme             (optional but it will output same style as doc webpage)      

You should then be able to run `make html`    from the doc/ folder  and view the docs in browser (Models are found in Model Directory in sidebar)

PR #1144, opened 2019-03-08
Add PyNEST function to synchronize MPI processes

hakonsbm:
The SLI function `SyncProcesses` blocks until all processes have reached the function, and is used to debug MPI issues. This PR adds a PyNEST binding. 

The PyNEST function is accessible from the `hl_api` submodule, i.e. as `nest.hl_api.SyncProcesses()`, to not encourage users to control MPI on the Python level.


terhorstd:
minimal change

PR #1145, opened 2019-03-11
Use relative path to logo in documentation

steffengraber:
With the relative path always the correct logo is inserted.


heplesser:
Travis failed due to a time-out. Once Travis has gone through successfully, this PR can be merged without further review.

PR #1146, opened 2019-03-12
Create masked layer earlier when using topology's convergent connect

hakonsbm:
If not all MPI processes have targets, an MPI deadlock would occur when calling `source.get_global_positions_vector()`, because it creates a `MaskedLayer`, requiring MPI communication. By essentially splitting up what `source.get_global_positions_vector()` does,  creating the `MaskedLayer` outside of the target loop, this deadlock is no longer possible.


heplesser:
The original reporter of the underlying problem has confirmed that this fix works for him.

jougs:
@stinebuu: could you maybe review this? @jhnnsnk is currently unavailable and we'd like to release soon.

PR #1147, opened 2019-03-12
Fix MPI synchronization problem in presence of very small layers

hakonsbm:
When connecting layers on multiple MPI processes, the processes may end up with different values for the bool `ConnectionManager::have_connections_changed_` if not all of the processes actually create connections. This is a problem in cases where updating connection infrastructure only occurs when connections are changed, as `update_connection_infrastructure()` requires MPI communication. Processes then go out of sync.

This is not a problem when connecting normally, as `have_connections_changed_` is set to `true` even if no connections are created. In topology however, `have_connections_changed_` is not set until a connection is made. With this PR it is set to `true` earlier, even if no connections are created.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -936,6 +936,7 @@ nest::ConnectionManager::data_connect_single( const index source_id,
 bool
 nest::ConnectionManager::data_connect_connectome( const ArrayDatum& connectome )
 {
+  kernel().connection_manager.set_have_connections_changed( true );
   for ( Token* ct = connectome.begin(); ct != connectome.end(); ++ct )
   {
     DictionaryDatum cd = getValue< DictionaryDatum >( *ct );
@@ -964,7 +965,6 @@ nest::ConnectionManager::data_connect_connectome( const ArrayDatum& connectome )
     Node* source_node = kernel().node_manager.get_node( source_gid );
     connect_( *source_node, *target_node, source_gid, thr, syn_id, cd );
   }
-  kernel().connection_manager.set_have_connections_changed( true );
   return true;
 }
 
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -964,6 +964,7 @@ nest::ConnectionManager::data_connect_connectome( const ArrayDatum& connectome )
     Node* source_node = kernel().node_manager.get_node( source_gid );
     connect_( *source_node, *target_node, source_gid, thr, syn_id, cd );
   }
+  kernel().connection_manager.set_have_connections_changed( true );
   return true;
 }
 
--- a/topology/topology.cpp
+++ b/topology/topology.cpp
@@ -210,8 +210,8 @@ connect_layers( const index source_gid,
   const index target_gid,
   const DictionaryDatum& connection_dict )
 {
-  kernel().connection_manager.set_have_connections_changed(true);
-  
+  kernel().connection_manager.set_have_connections_changed( true );
+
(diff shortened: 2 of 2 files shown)


hakonsbm:
@heplesser It doesn't look like it's a problem with `data_connect_single`, all processes go into `ConnectionManager::connect` and set `have_connections_changed_ = true`. And when I try to provoke the same issue with `data_connect_connectome`, it throws an `IllegalConnection` exception.

heplesser:
@hakonsbm  `data_connect_connectome()` seems to call `connect_()` only, so it would never set `have_connections_changed_`. Could you double check this? We probably don't have an MPI test fully checking `data_connect_connectome()`, so problems here may have gone undetected so far. I also find the logic of `data_connection_single()` too complex to feel entirely safe that the flag will eventually be set in all cases.

hakonsbm:
@heplesser That's true, `have_connections_changed_` is actually never set for `data_connect_connectome()`! It won't cause processes to diverge, but it may cause other problems. Should I fix it in this PR, or create a new one?

 I can't see any places for `data_connection_single()` where processes may diverge before the `have_connections_changed_` flag is set. In `ConnectionCreator::convergent_connect_()` processes may diverge because it runs over local targets only. But `data_connection_single()` runs over specified targets from the SLI level, which is the same for all processes.

heplesser:
@hakonsbm 

> @heplesser That's true, `have_connections_changed_` is actually never set for `data_connect_connectome()`! It won't cause processes to diverge, but it may cause other problems. Should I fix it in this PR, or create a new one?

I would suggest that you fix it in this PR, as it should be easy to do.

> I can't see any places for `data_connection_single()` where processes may diverge before the `have_connections_changed_` flag is set. In `ConnectionCreator::convergent_connect_()` processes may diverge because it runs over local targets only. But `data_connection_single()` runs over specified targets from the SLI level, which is the same for all processes.

Ok.

jougs:
@stinebuu: could you maybe review this? @jhnnsnk is currently unavailable and we'd like to release soon.

nestkernel/connection_manager.cpp
```diff
@@ -964,6 +964,7 @@ nest::ConnectionManager::data_connect_connectome( const ArrayDatum& connectome )
     Node* source_node = kernel().node_manager.get_node( source_gid );
     connect_( *source_node, *target_node, source_gid, thr, syn_id, cd );
   }
+  kernel().connection_manager.set_have_connections_changed( true );
```
heplesser: @hakonsbm For consistency with the other cases, I would suggest to move this call to the beginning of the method. In that way the flag is set even if an exception should occur after the first connection has been created.

jhnnsnk:
Thanks for the fix. Just a small question: Why is `have_connections_changed_` initialized to `true`? Shouldn't it be set to `true` only inside the connect calls?

heplesser:
@jhnnsnk If `ConnectLayers()` is called on layers with fewer neurons than the number of MPI ranks, then some ranks would not create any connections. This does in practice happen if you have very small subpopulations in a large network spread across many compute nodes. In that case, `have_connections_changed_` will be set on some, but not all ranks, and the next MPI communication will lead to desynchronization as not all ranks participate. Setting the flag early avoids this.

jhnnsnk:
@heplesser The question I have is whether this initialization to `true` is needed:
https://github.com/nest/nest-simulator/blob/6e64c37c52d559e4dbd17033ef473842ef8a8d17/nestkernel/connection_manager.cpp#L70
Otherwise I don't have any comments and will approve.

Issue #210, opened 2016-01-20
Request: Provide Anaconda package for NEST


apdavison:
Currently, installing NEST on OS X is painful, and requires Macports or Homebrew.

Many Python users on OS X now use Anaconda, since it makes installing large packages such as Numpy and Matplotlib much easier.

Having a NEST package in the Anaconda Cloud repository would make installing NEST on OS X as simple as:

```
conda install -c https://conda.anaconda.org/nest-simulator nest
```

apeyser:
Another reason for this: #209
Apparently, Anaconda is quite popular, but will require some package wrapping because of dynamic libraries, which will be blamed on nest. Entry point for how this is done: http://conda.pydata.org/docs/build_tutorials/pkgs.html

apeyser:
Anaconda seems to have come from the issue of building numpy correctly, with a large number of non-python dependencies, for people who would like to move a bit faster than their distributions do --- and some distributions have quite poor numpy performance because of their choice of packages to provide those dependencies (such as using netlib blas). 

Apparently by search interest, google says that python anaconda software is 1/10 as popular as the song by Nicki Minaj. The question is, how much bother is supporting an anaconda recipe?

apeyser:
If someone is using anaconda and is willing to share their recipe, this would be a great place to put the gist.

jougs:
@tammoippen: Could you check if this gets any simpler using your cmake port of our build system? Thanks!

tammoippen:
Hi,

I can confirm the loading problem (#209) on a fresh Linux (Lubuntu 15.10). I first installed Anaconda (2.4.1), then all NEST dependancies. When calling `python -c "import nest"` i get the same error:

    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/test/Downloads/nest2.10.build/install/lib/python2.7/site-packages/nest/__init__.py", line 52, in <module>
        from . import pynestkernel as _kernel
    ImportError: /home/test/anaconda2/bin/../lib/libreadline.so.6: undefined symbol: PC

Calling it with debug `LD_DEBUG=libs python -c "import nest"`, reveals that python loads its readline library using the `RPATH`. Another readline will not be loaded, even if you set `LD_LIBRARY_PATH`, as `RPATH` will be searched first.

    find library=libreadline.so.6 [0]; searching
       search path=/home/test/anaconda2/bin/../lib		(RPATH from file python)
[...]

PR #1153, opened 2019-03-20
Added test_multiarea_model_siegert_rates

AlexVanMeegen:
Added a test that uses the Siegert neurons to compute stationary rates of the multi area model (see [Schuecker et al. 2017](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005179)) as requested by @terhorstd. Motivated by the finding that, before PR #1059, different ways of compiling NEST with different conda environments lead to numerically inconsistent results - see issue #210.

Three pickled numpy arrays are necessary: indegrees, weights and stationary rates of the multiarea model. For the time being, they are in a subfolder `test_multiarea_model_siegert_rates/*.npy` - is this the correct location?


AlexVanMeegen:
Dear @heplesser, thanks for reviewing this. However, I am not sure if I understand your suggestions correctly. 

What do you refer to with 'reference data' - the resulting mean firing rates (`mf_rates.npy`) or the rates and the full connectivity (`mf_rates.npy, J.npy, K.npy`)? Storing `K.npy` or `J.npy` in the test file would mean to store 254x255 floats, for `mf_rates.npy` it's 254 floats.

Also, I am not sure what you mean with 'the code that generated the reference data'. Essentially, the test is a piece of the multiarea model code that is cut out and adapted: all pre-processing (generating the connectivity) is left out and replaced by `K.npy` and `J.npy`. To get rid of these would mean to add the full multiarea model as a dependency as far as I can see.

Finally, since #1059 I don't have a setup to test the test anymore. I discussed this with @terhorstd and he suggested to include it anyways because the calculation was useful in the past to detect inconsistencies in the environment. @terhorstd, please correct me if there is a deeper motivation that I missed here.

terhorstd:
Yes, I do think it's a good idea to keep this mechanism around, but as discussed we would rather seek for a test with an actual single library call, than include a megabyte of data. I have opened an issue #1164 for this. This PR still contains all files and changes necessary to reproduce the behaviour (even though probably only given #1059 is reverted).

So, as we agreed, I close this PR for now.

PR #1157, opened 2019-03-27
Simplify dynamic import of PyNEST submodules

apeyser:
Feature/dynamic import #1150

Since @jougs liked the brevity of the old solution, I shrunk this as much as possible -- instead of doing module member transfers by hand, it recreates the old exec, but using an ast.

The advantage of this over `exec('string')` is that you won't get weird errors for garbage filenames in the directory (and the code can be simply re-used elsewhere)

For example, with the old string approach, if a file were dropped with the name of `'lib/x a .py` your error would be an unclear syntax error -- you can make arbitrarily bad 'file names' that could do anything, given sufficient uncleveriness.

### DIFF
--- a/pynest/nest/hl_api.py
+++ b/pynest/nest/hl_api.py
@@ -41,7 +41,7 @@ def _import_libs(mod_file, mod_dict, path,
       level: relative level (1 is relative to path, 2 is ../path, ...)
              (default 1)
       """
-                 
+
     import ast
     import os
 
--- a/pynest/nest/hl_api.py
+++ b/pynest/nest/hl_api.py
@@ -23,35 +23,31 @@
 High-level API of PyNEST Module
 """
 
-import os
-import sys
-
 # We search through the subdirectory "lib" of the "nest" module
 # directory and import the content of all Python files therein into
 # the global namespace. This makes the API functions of PyNEST itself
 # and those of extra modules available to the user.
 
 
-def _pkg_attrs(pkg):
-    # return all package attributes that
-    # from $pkg import *
-    # would put in the module namespace $pkg
-    # see https://docs.python.org/3/reference/simple_stmts.html#import
-    try:
-        # if there's an "__all__", return that
-        return pkg.__all__
-    except AttributeError:
-        # otherwise, return everything at top level that doesn't start with `_`
-        return (attr for attr in pkg.__dict__ if attr[0] != '_')
+def _import_libs(mod_file, mod_dict, path,
+                 prefix=None, ignore=frozenset(), level=1):
+    """
+    _import_libs: construct a relative import of all modules
+      mod_file: name of calling module file (__file__)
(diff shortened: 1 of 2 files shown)


pynest/nest/hl_api.py
```diff
@@ -23,35 +23,31 @@
...
+      level: relative level (1 is relative to path, 2 is ../path, ...)
+             (default 1)
+      """
 
-def _import_names(pkg, mod_dict):
-    # import * names from pkg to the dictionary of mod_dict
-    for attr in _pkg_attrs(pkg):
-        mod_dict[attr] = getattr(pkg, attr)
+    import ast
+    import os
```
babsey: According to PEP 8, imports are always put at the top of the file. (https://www.python.org/dev/peps/pep-0008/#imports)

apeyser: Should I pollute the hl_api namespace with unneeded imports? I think the goal of the pep8 requirement is to put global imports at the top of the file rather than in the middle or at the bottom, for ease of reading (but who knows -- a lot of pep8 is fetishes by weird folks).

Another option is to extract all of this and put it in a submodule to hl_api. Probably cleaner that way.

babsey: I agree with you avoiding hl_api namespaces pollution. It is not necessary to put import on global level. Thus, I approved the PR.

pynest/nest/hl_api.py
```diff
@@ -23,35 +23,31 @@
...
+    """
+    _import_libs: construct a relative import of all modules
+      mod_file: name of calling module file (__file__)
+      mod_dict: globals() of calling module
+      path: relative path to check for modules
+      prefix: string for import prefix (defaults to path)
+      ignore: set of import names ($prefix.$name) to not import
+      level: relative level (1 is relative to path, 2 is ../path, ...)
+             (default 1)
+      """
```
jougs: Can you please convert the docstring to NumPy style to conform with the other functions?

pynest/nest/hl_api.py
```diff
@@ -61,11 +57,17 @@ def _import_libs(mod_file, mod_dict, path, prefix, ignore_modules=frozenset()):
             continue  # not a regular python module
 
         pkg_name = "{}.{}".format(prefix, name[:-3])
-        if pkg_name in ignore_modules:
+        if pkg_name in ignore:
             continue  # this package is not to be imported dynamically
 
-        pkg = __import__(pkg_name, mod_dict, locals(), ['*'], 1)
-        _import_names(pkg, mod_dict)
+        # construct from .pkg_name import * ; this is the old 'exec(string)'
```
jougs: Thanks for the hint. I think you can remove the `; this is the old 'exec(string)'` part from this comment now.

apeyser:
Responded to comments, and extracted the code into nest.import_libs

PR #1160, opened 2019-04-03
Introduce time-driven spiking connections

jakobj:
While the event-driven simulation method for synapse updates makes NEST very efficient, it is often non-trivial to formulate an given plasticity rule involving multiple factors in an event-driven algorithm. To support fast prototyping in NEST a time-driven-update mode for synapses would hence be of great value.

This PR implements a simple approach to time-driven synapse updates: the framework for continuous interactions is used to communicate spikes. This results in the corresponding synapses being updated in every time step. As a proof of concept the `iaf_psc_exp` neuron now can be connected via a new `time_diven_static_connection`, which results in identical dynamics as the matching event-driven implementation. For this model both event- and time-driven connections can be used within the same simulation.

Since the technical side is straightforward, I think the decision of accepting this PR is rather a question of how useful this is estimated to be vs. the increase in loc.
I suggest @heplesser @janhahne @suku248 as reviewers, @clinssen may be interested with regards to NESTML.

### DIFF
--- a/models/time_driven_static_connection.h
+++ b/models/time_driven_static_connection.h
@@ -1,0 +1,161 @@
+/*
+ *  time_driven_static_connection.h
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+#ifndef TIME_DRIVEN_STATIC_CONNECTION_H
+#define TIME_DRIVEN_STATIC_CONNECTION_H
+
+/* @BeginDocumentation
+   Name: time_driven_static_synapse - Synapse type for time-driven static
+   connections.
+
+   Description:
+
+   For efficiency reasons spiking connections in NEST are only updated on every
+   presynaptic spike, making it difficult to implement time-driven plasticity
+   rules. This model implements a simple approach to time-driven synapse
+   updates: the framework for continuous interactions is used to communicate
+   spikes. This causes this model to be updated in every time step.
+
+   The event received by the synapse contains a buffer of length h / dt in which
+   non-zero values indicate a spike. For simplicity this static synapse model
+   forwards the entire TimeDrivenSpikeEvent to the receiving node where the
+   buffer is unpacked. To compute updates in dt steps in the synapse object, the
+   event should instead be unpacked in `send()` and a SpikeEvent in every dt
+   step should be forwarded to the receiving node.
+
+   time_driven_static_synapse does not support any kind of plasticity. It simply
+   stores the parameters target, weight, delay and receiver port for each
+   connection.
+
+   FirstVersion: April 2019
+
+   Author: Jakob Jordan
+
+   Transmits: TimeDrivenSpikeEvent
+
+   SeeAlso: static_synapse
+*/
+
+// Includes from nestkernel:
+#include "connection.h"
+
+namespace nest
+{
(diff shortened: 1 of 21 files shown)


ikitayama:
Is this going to be merged in the next major release?

heplesser:
@ikitayama This PR will not be in NEST 2.18, which we will release very soon.

jakobj:
@ikitayama no it was not meant to be included in the next release. in case you are interested in this feature you could use my branch for now

clinssen:
Thanks for the reference implementation! From a NESTML perspective, I am all for this, as it will increase the genericity of models that can be expressed.

models/time_driven_static_connection.h
```diff
@@ -0,0 +1,148 @@
...
+
+/* @BeginDocumentation
+   Name: time_driven_static_synapse - Synapse type for time-driven static
+   connections.
+
+   Description:
+
+   time_driven_static_synapse does not support any kind of plasticity. It simply
+   stores
+   the parameters target, weight, delay and receiver port for each connection.
```
janhahne: I suggest to extend this documentation here with the information that are provided in the first post of this PR. So basically explaining what this new connection is good for and when and how it could be adapted to simplify prototype implementations of new learning rules.

jakobj: good point, i've updated the docstring of the `TimeDrivenStaticSynapse`. please have a look whether that description is understandable.

jakobj:
@janhahne thanks for the review, i've addressed your comment and fixed a few minor other things. please have another look

models/time_driven_static_connection.h
```diff
@@ -0,0 +1,161 @@
...
+
+   The event received by the synapse contains a buffer of length h / dt in which
+   non-zero values indicate a spike. For simplicity this static synapse model
+   forwards the entire TimeDrivenSpikeEvent to the receiving node where the
+   buffer is unpacked. To compute updates in dt steps in the synapse object, the
+   event should instead be unpacked in `send()` and a SpikeEvent in every dt
+   step should be forwarded to the receiving node.
+
+   time_driven_static_synapse does not support any kind of plasticity. It simply
+   stores
```
janhahne: Just a minor formatting issue (probably due to auto-formatting): Maybe move the remaining part of the sentence up to this line. Having a line with a single word within a sentence looks a bit strange.

models/time_driven_static_connection.h
janhahne: Thanks! Looks good to me.

models/time_driven_static_connection.h
jakobj: fixed

jakobj:
@janhahne thanks, i've fixed the strange formatting and squashed all commits

jakobj:
@janhahne @suku248 
please put this on hold for the moment, I'm thinking about another implementation that puts less strain on the communication infrastructure.

sbrodeur:
I've also been giving a try with an implementation a general mechanism for time-driven update of synaptic connections. I'm developing homeostatic plasticities for my research project, and it must be time-driven. It is too expensive with PyNEST to do it outside the simulation kernel (e.g. from Python-side, use _GetStatus_ to get the weights, modify them and update them with _SetStatus_). At the same time, I explored a mean to reset the dynamical states of the synaptic connections. So far, all these new features required only a minimal amount of changes and don't break existing models that do not support those features. You can take look at the draft implementation here: 
https://github.com/NECOTIS/nest-simulator/tree/feature/time-driven-connection-update

From the implementation perspective, I've not using a custom _Event_ class but integrated directly the update of connections in the main loop of the simulation manager. This is very similar to how _Node_ instances are updated, and greatly benefits from OpenMP parallelization. By default, the update methods for connections are called at every time step but I added a global parameter to define an update interval (e.g. every 100 steps) to avoid some computational burden (i.e. homeostatic plasticity can run every 10-100 ms and it is fine).

I haven't planned yet to do a PR since I have just a draft implementation, and my neuron and synapse models are in a custom module. I'm under pressure to finish my experiments, so I can't polish things as much as I would like. But maybe I can hopefully contribute to NEST in the long-term. Is there a plan for time-driven updates of connections and resettable synapse dynamics in NEST 3.0?

heplesser:
@jakobj Since you are working on a different approach, would you consider closing this PR and creating a new PR when you have the new approach ready?

jakobj:
@sbrodeur I had a look at your changes, and they are in line with the approach I was thinking about. I'd be happy to review a PR of the changes implementing a time-driven update of the connections. Of course we would need to think a bit more about (and possibly measure) the impact on simulations not using this feature, but my feeling so far is that there's no drawback. 
Resetting the dynamical state should be handled in a separate PR since the two are logically independent.

@heplesser sure, this PR can be closed.

PR #1166, opened 2019-04-08
Add advice against using NEST in threaded processes; addresses #361.

### LINKED ISSUE #361: Clearly document that threading in NEST should not be combined with multiple threads in Python
Threading with python is bad.
Threading with Nest & python is really bad.
Threading with Nest & python & openmp is really really bad (SEGV bad often).
The question is: should this simply be declared "cave canem", or should we preempt by checking that python is running on the main thread (and to what extent is even that well-defined?) when nest is entered if openmp is active. I'm referring to having multiple threads created from python, while leaving unchanged the openmp threading model of nest.

Or is there some third idea about blowing up in a well-defined way with multiple threads?

heplesser:
This fixes #361.

### DIFF
--- a/doc/guides/parallel_computing.rst
+++ b/doc/guides/parallel_computing.rst
@@ -28,6 +28,11 @@ See `Plesser et al
 information on NEST parallelization and be sure to check the
 documentation on `Random numbers in NEST <random-numbers.md>`__.
 
+**Note:** The NEST simulation kernel assumes that it has full control
+over threads in the NEST process. Combining this with other use of threads
+in the same process that is running the NEST kernel, can lead to unpredictable
+results and **is not supported**. 
+
 Concepts and definitions
 ------------------------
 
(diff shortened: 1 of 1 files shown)


doc/guides/parallel_computing.rst
```diff
@@ -28,6 +28,11 @@ See `Plesser et al
 information on NEST parallelization and be sure to check the
 documentation on `Random numbers in NEST <random-numbers.md>`__.
 
+**Note:** The NEST simulation kernel assumes that it has full control
+over threads in the NEST process. Combining this with other use of threads
+in the same process that is running the NEST kernel, can lead to unpredictable
+results and **is not supported**.
```
jougs:
```suggestion
results and **is not supported**. Examples for this would be the use of Python
threads or multi-threaded modules from within Python.
```

Issue #370, opened 2016-05-26
Create iaf_psc_alpha_ps


heplesser:
We currently have models `iaf_psc_exp_ps` and `iaf_psc_delta_ps`, which are direct precise equivalents of their normal counterparts. Create `iaf_psc_alpha_ps` as direct counterpart to `iaf_psc_alpha`. Currently, we only have `iaf_psc_alpha_canon`, which only has a single synaptic time constant, while `iaf_psc_alpha` has different time constants for E and I synapses. The `_canon` and `_presc` models are counterparts to `iaf_neuron`, not `iaf_psc_alpha`.

Silmathoron:
It's in my [PR](https://github.com/nest/nest-simulator/pull/229), but it's a GSL version...

heplesser:
Since #229 was not merged, we are still lacking these models. See also #368 .

Silmathoron:
Right x)
I completely forgot that... but I guess we should discuss the implementation anyway... I'm usually in favor of GSL but maybe this model is solvable exactly and should be implemented differently...

Btw @heplesser, is there a document somewhere explaining the new labels? The "C", "I", P" letters are not very explicit to me...

heplesser:
@Silmathoron I will post an explanation of the labels on the Wiki shortly. C: Component, I: Interface, P: Progress, S: Severity, T: Type. The main point of having these letters in front is to make sure that the labels for the different "dimensions" always are in the same (alphabetical) order.

heplesser:
@Silmathoron Any news on this?

heplesser:
@Silmathoron Ping!

Silmathoron:
@heplesser I assume this means that we go for the GSL version then ;) I'll try to do that tomorrow

Silmathoron:
Ok, I ended up wasting my day and actually did the propagator version... cheers!

stinebuu:
Fixed by #1169.

PR #1169, opened 2019-04-09
Add iaf_psc_alpha_ps neuron model

Silmathoron:
This PR implements the ``iaf_psc_alpha_ps`` precise model.
Addresses #370 

I'll add a test ASAP.
For formatting, do we really still need to use ``clang-format-3.6`` ?! This doesn't exist anywhere anymore ^^


heplesser:
@Silmathoron We recently decided to name PRs so that the PR names directly turn into user-informative lines for the release notes. They should always start with a verb in present tense and should not contain references to issues. I therefore changed the title.

precise/precisemodule.cpp
```diff
@@ -48,6 +48,7 @@
 #include "model_manager_impl.h"
 
 // Includes from precise:
+#include "iaf_psc_alpha_ps.h"
```
heplesser: Please put the includes in alphabetical order.

precise/precisemodule.cpp
```diff
@@ -91,6 +92,8 @@ void
 PreciseModule::init( SLIInterpreter* )
 {
   // register models
+  kernel().model_manager.register_node_model< iaf_psc_alpha_ps >(
+    "iaf_psc_alpha_ps" );
```
heplesser: Please put registrations in alphabetical order.

pynest/nest/tests/test_iaf_psc_alpha_ps.py
```diff
@@ -0,0 +1,190 @@
...
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import unittest
+
```
heplesser: I'd suggest to remove the empty lines between imports.

pynest/nest/tests/test_iaf_psc_alpha_ps.py
```diff
@@ -0,0 +1,190 @@
...
+import numpy as np
+
+import nest
+
+
+"""
+Comparing the new implementations the precise ``iaf_psc_alpha_ps`` model to the
+previous reference implementation given by ``iaf_psc_alpha_canon``.
+
+The new implementation provides two separate buffer for excitatory and
```
heplesser: buffers

pynest/nest/tests/test_iaf_psc_alpha_ps.py
```diff
@@ -0,0 +1,190 @@
...
+                if np.any(nonzero):
+                    rds = rds[nonzero] / np.abs(reference[record][nonzero])
+                # ignore events around spike times for V if it diverges
+                if record == "V_m" and params["Delta_T"] > 0.:
+                    spiking = (dmm[record] > V_lim)
+                    rds = rds[~spiking]
+                rel_diff[model][record] = np.average(rds)
+        return rel_diff
+
+    def test_closeness_same_timescales(self):
```
heplesser: time constants, not time scales, isn't it?

pynest/nest/tests/test_iaf_psc_alpha_ps.py
```diff
@@ -0,0 +1,190 @@
...
+        '''
+        Compare models for tau_syn_ex == tau_syn_in (should be equal).
+        '''
+        Ie = 370.
```
heplesser: Insert empty line above.

pynest/nest/tests/test_iaf_psc_alpha_ps.py
```diff
@@ -0,0 +1,190 @@
...
+        data_ref = nest.GetStatus(mm_ref, "events")[0]
+        data_new = nest.GetStatus(mm_new, "events")[0]
+
+        self.assertTrue(np.allclose(data_new["V_m"], data_ref["V_m"]))
+
+    def test_unequal_different_timescales(self):
+        '''
+        Check that the behavior differs if ``tau_syn_in != tau_syn_ex``
+        '''
+        Ie = 370.
```
heplesser: Insert empty line above.

pynest/nest/tests/test_iaf_psc_alpha_ps.py
```diff
@@ -0,0 +1,190 @@
...
+        nest.Connect(mm_ref, ref)
+        nest.Connect(mm_new, new)
+
+        nest.Simulate(1000.)
+
+        data_ref = nest.GetStatus(mm_ref, "events")[0]
+        data_new = nest.GetStatus(mm_new, "events")[0]
+
+        # test before 500.
+        keep = data_new["times"] < 500.
```
heplesser: keep seems a strange name. Could you add a comment on why you expect equality before 500, but not thereafter?

precise/iaf_psc_alpha_ps.h
```diff
@@ -0,0 +1,548 @@
...
+
+// Includes from precise:
+#include "slice_ring_buffer.h"
+
+namespace nest
+{
+
+/** @BeginDocumentation
+Name: iaf_psc_alpha_ps - Leaky integrate-and-fire neuron
+with alpha-shape postsynaptic currents (one timescale for excitatory and
```
heplesser: I would remove the text in (), since this model should become the default. Also, in this model we rather have *different* timescales, as it is written now one could think there is just a single for ex *and* in.

precise/iaf_psc_alpha_ps.h
```diff
@@ -0,0 +1,548 @@
...
+V_m          double - Membrane potential in mV
+E_L          double - Resting membrane potential in mV.
+V_min        double - Absolute lower value for the membrane potential.
+C_m          double - Capacity of the membrane in pF
+tau_m        double - Membrane time constant in ms.
+t_ref        double - Duration of refractory period in ms.
+V_th         double - Spike threshold in mV.
+V_reset      double - Reset potential of the membrane in mV.
+tau_syn_ex   double - Rise time of the excitatory synaptic function in ms.
+tau_syn-in   double - Rise time of the inhibitory synaptic function in ms.
```
heplesser: tau_syn_in

precise/iaf_psc_alpha_ps.h
```diff
@@ -0,0 +1,548 @@
...
+    Integration with Continuous Spike Times in Discrete Time Neural Network
+    Simulations. To appear in Neural Computation.
+[2] Rotter S & Diesmann M (1999) Exact simulation of time-invariant linear
+    systems with applications to neuronal modeling. Biologial Cybernetics
+    81:381-402.
+[3] Hanuschkin A, Kunkel S, Helias M, Morrison A & Diesmann M (2010)
+    A general and efficient method for incorporating exact spike times in
+    globally time-driven simulations Front Neuroinformatics, 4:113
+
+Author: Tanguy Fardet (from Diesmann, Eppler, Morrison, Plesser, Straube)
```
heplesser: from -> based on

precise/iaf_psc_alpha_ps.h
```diff
@@ -0,0 +1,548 @@
...
+
+Sends: SpikeEvent
+
+Receives: SpikeEvent, CurrentEvent, DataLoggingRequest
+
+SeeAlso: iaf_psc_alpha, iaf_psc_alpha_presc, iaf_psc_exp_ps
```
heplesser: Why no reference to iaf_psc_alpha_canon?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+  if ( updateValue< double >( d, names::V_m, V_m_ ) )
+  {
+    V_m_ -= p.E_L_;
+  }
+  else
+  {
+    V_m_ -= delta_EL;
+  }
+
+  updateValue< double >( d, names::I_syn_ex, I_ex_ );
```
heplesser: The synaptic currents and their derivatives should not be settable. I don't think they are in any other models.

Silmathoron: They were in [canon](https://github.com/nest/nest-simulator/blob/master/precise/iaf_psc_alpha_canon.cpp#L206) but this is set state, not parameters, so it seems normal to me

heplesser: Your are right about `canon`, but I think that is rather a leftover from debugging. The derivatives are not documented in `canon`, and they cannot be set in the plain `iaf_psc_...` models. So I would argue against introducing them here.

pynest/nest/tests/test_iaf_psc_alpha_ps.py
Silmathoron: You want an empty line between the docstring and ``Ie``?! Damn... I hate that... :'( is it a NEST convention?

precise/iaf_psc_alpha_ps.h
Silmathoron: Should I add it even though we're deprecating it?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+nest::iaf_psc_alpha_ps::calibrate()
+{
+  B_.logger_.init();
+
+  V_.h_ms_ = Time::get_resolution().get_ms();
+
+  V_.psc_norm_ex_ = 1.0 * numerics::e / P_.tau_syn_ex_;
+  V_.psc_norm_in_ = 1.0 * numerics::e / P_.tau_syn_in_;
+
+  V_.gamma_ex_ = 1 / P_.c_m_ / ( 1 / P_.tau_syn_ex_ - 1 / P_.tau_m_ );
```
jakobj: these do not seem numerically stable for `tau_syn_ex_/in_ == tau_m_`, neither do they seem to be in `iaf_psc_alpha_canon`. this should be fixed, but maybe not in this PR?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+    throw BadProperty( "Invalid interpolation order in iaf_psc_alpha_ps." );
+  }
+  return 0;
+}
+
+// finds threshpassing via linear interpolation
+double
+nest::iaf_psc_alpha_ps::thresh_find1_( double const dt ) const
+{
+  double tau =
```
jakobj: `const double`

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+  SpikeEvent se;
+  se.set_offset( S_.last_spike_offset_ );
+  kernel().event_delivery_manager.send( *this, se, lag );
+
+  return;
+}
+
+// finds threshpassing
+inline double
+nest::iaf_psc_alpha_ps::thresh_find_( double const dt ) const
```
jakobj: `double` <-> `const`?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+    return thresh_find3_( dt );
+  default:
+    throw BadProperty( "Invalid interpolation order in iaf_psc_alpha_ps." );
+  }
+  return 0;
+}
+
+// finds threshpassing via linear interpolation
+double
+nest::iaf_psc_alpha_ps::thresh_find1_( double const dt ) const
```
jakobj: `double` <-> `const`?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+{
+  double tau =
+    ( P_.U_th_ - V_.V_m_before_ ) * dt / ( S_.V_m_ - V_.V_m_before_ );
+  return tau;
+}
+
+// finds threshpassing via quadratic interpolation
+double
+nest::iaf_psc_alpha_ps::thresh_find2_( double const dt ) const
```
jakobj: `double` <-> `const`?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -0,0 +1,832 @@
...
+    return tau2;
+  }
+  else
+  {
+    return thresh_find1_( dt );
+  }
+}
+
+double
+nest::iaf_psc_alpha_ps::thresh_find3_( double const dt ) const
```
jakobj: `double` <-> `const`?

precise/iaf_psc_alpha_ps.cpp
Silmathoron: This is checked in the propagator (line 314 and below)

jakobj: hmm, as far as I understand the propagators are selected according to whether `tau_syn_ex_ ~ tau_m_` and then computed such that division by zero is avoided. I am failing to see how the same is done for the gammas? maybe I need another coffee? ;)

Silmathoron: No, obviously you're right! I see it now that I had mine ;)
I wouldn't mind if this was done in another PR, yes ^^"

Silmathoron:
@heplesser @jakobj I corrected spike handling (single buffer like the exp version) and incorporated most of your comments.

Regarding the tests, I removed ``canon`` wherever it seemed reasonable.
There are still a few ``iaf_ps_`` tests where I did not include the new model because ``iaf_psc_exp_ps`` was not included either and adding it would require removing the ``tau_syn`` variable which does not exist in the model. If you think it would be better, I can remove ``tau_syn`` from `` Set`` and include the model in the tests.
These tests are:
* ``test_iaf_ps_psp_accuracy.sli``
* ``test_iaf_ps_psp_poisson_generator_accuracy.sli``
* ``test_iaf_ps_psp_poisson_spike_accuracy.sli``

Silmathoron:
By the way, ``do_tests.py`` (called from ``make installcheck``) errors on current master because it tries to call ``nest.init``, which does not exist (``init`` is part of ``ll_api.py`` but is not imported to the main ``__init__.py``, so)

As far as I'm concerned, I think this is a bug and that ``nest.ll_api.init`` should be called instead. Could you confirm/infirm, and tell me if this is already known or if I should file a new issue?

lekshmideepu:
Please take a look [here](https://travis-ci.org/lekshmideepu/nest-simulator/jobs/521221141#L6860)
` File "/Users/travis/build/lekshmideepu/nest-simulator/result/share/nest/extras/do_tests.py", line 50, in <module>
    nest.init(["--verbosity=WARNING", "--quiet"])
AttributeError: 'module' object has no attribute 'init'
Assertion failed: (kernel_manager_instance_), function get_kernel_manager, file /Users/travis/build/lekshmideepu/nest-simulator/nestkernel/kernel_manager.h, line 206.
/Users/travis/build/lekshmideepu/nest-simulator/result/share/nest/extras/do_tests.sh: line 794: 21235 Abort trap: 6           "${PYTHON}" "${PYTHON_HARNESS}" >> "${TEST_LOGFILE}"
cat: pynest_test_numbers.log: No such file or directory`

jougs:
@Silmathoron: @lekshmideepu and I just found exactly the same problem with `do_tests.py` (see above) and also think that it should be `nest.ll_api.init(...)`. Alternatively we could make `init()` avaliable in `__all__`.

Everything works fine if you have `nosetests` installed.

lekshmideepu:
@Silmathoron : I would just fix it in my PR

Silmathoron:
@lekshmideepu which PR? Can you file an issue at the same time, then?
@jougs I don't think it belongs on the top level, I would just fix the call...

lekshmideepu:
@Silmathoron : sure I would file an issue

lekshmideepu:
#1181

precise/iaf_psc_alpha_ps.cpp
jakobj: ok, then I would suggest to throw an exception when the time constants are too close as a temporary solution. could you additionally please create an issue that describes what should be fixed here?

pynest/nest/tests/test_iaf_psc_alpha_ps.py
heplesser: I don't think we have an explicit convention, I just find it more legible. PEP257 requires an empty line behind class docstrings, but this is not the case here. Btw, PEP257 also wants double quotes instead of single quotes around docstrings ;)

precise/iaf_psc_alpha_ps.h
heplesser: Maybe not.

doc/guides/simulations_with_precise_spike_times.rst
```diff
@@ -72,7 +72,7 @@ delta-shaped post-synaptic currents that employs precise spike times;
 its grid-constrained counterpart is ``iaf_psc_delta``. In this model the
 precise location of an outgoing spike is determined analytically.
 
-``iaf_psc_alpha_canon`` and ``iaf_psc_alpha_presc`` are
+``iaf_psc_alpha_psc`` and ``iaf_psc_alpha_presc`` are
```
heplesser: `..._ps`, not `..._psc`

heplesser:
The NEST Open Developer VC on 3 June approved the deprecation of `iaf_psc_alpha_canon` and its removal in NEST 3.0. 

The documentation of `iaf_psc_alpha_ps` should remark that this model replaced `iaf_psc_alpha_canon` from Morrison el al (2007).

suku248:
In this model the threshold crossing is determined using interpolation as described in Morrison et al. (2007).
A more efficient way of locating the threshold crossing is bisectioning (Hanuschkin et al., 2010) as implemented in `iaf_psc_exp_ps`. I think the interpolation in `iaf_psc_alpha_ps` should also be replaced with bisectioning. If we don't replace it, `iaf_psc_exp_ps` and `iaf_psc_alpha_ps` have similar names but use different methods for finding the threshold crossing.
@heplesser what do you think?

heplesser:
@suku248 Thanks for spotting the difference in method between `_exp_ps` and this `_alpha_ps`.  I agree with you that we should be consistent across different models, so I would suggest adopting bisectioning also for `_alpha_ps`.

Silmathoron:
@heplesser @suku248 I updated the model with the bisectioning method, could you check?

precise/iaf_psc_alpha_ps.h
```diff
@@ -45,15 +45,16 @@ namespace nest
 
 /** @BeginDocumentation
 Name: iaf_psc_alpha_ps - Leaky integrate-and-fire neuron
-with alpha-shape postsynaptic currents.
+with alpha-shape postsynaptic currents and bisectioning method for
+approximation of threshold crossing..
```
heplesser: Double fullstop at end of sentence.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -69,7 +69,8 @@
 # deprecated functions will be registered by the @deprecated decorator.
 _deprecation_warning = {'BackwardCompatibilityConnect': True,
                         'subnet': True,
-                        'aeif_cond_alpha_RK5': True}
+                        'aeif_cond_alpha_RK5': True,
+                        'iaf_psc_alpha_canon': True}
```
heplesser: Also deprecated `iaf_psc_delta_canon`, it should be replaced with `iaf_psc_delta_ps` eventually.

heplesser: Furthermore, deprecate `iaf_psc_alpha_presc`. This model will not be replace with anything else, it was just a temporary test model for the Morrison et al 2007 paper.

precise/iaf_psc_alpha_ps.h
Silmathoron:
```suggestion
approximation of threshold crossing.
```

jougs:
@suku248: Are you OK with merging this? Doumo arigatou!

PR #1170, opened 2019-04-09
CMake option and documentation for using the Intel compiler

### LINKED ISSUE #242: Support Intel compiler suite and MKL
A range of HPC sites provide the Intel compiler suite and MKL. Currently, we provide no instructions on how to build NEST with the Intel compiler suite with pertaining libraries. It seems, though, that the Intel tools may provide significant performance advantages, see #237, so it may be useful to support the Intel tools. #237 also contains the current state of experience with building with Intel tools.

### LINKED ISSUE #1027: Add CMake configuration for Intel Compiler, MPI, MKL
We should add CMake support for Intel compilers. One important point is to set the `-fp-model strict` flag to ensure IEEE754-compatible numerics.

See also #242.

hakonsbm:
This PR adds a CMake option, `with-intel`, that is to be used when compiling with the Intel compiler, and documentation on this option. Adding 
```
-Dwith-intel=ON
```
when configuring NEST adds the compiler flag `-fp-model strict` to ensure that computations obey the IEEE754 standard for floating point numerics. Other flags can also be set with 
```
-Dwith-intel=<compiler-flags>
```

Fixes #242 
Fixes #1027


ikitayama:
Shouldn’t we do that in an arch/platform file instead the general
ProcessOptions?

hakonsbm:
Some changes:
- `-fp-model strict` flag is set by default if the Intel compiler is used.
- This can be overridden with the option `-Dwith-intel-flags=<compiler-flags>`.

@ikitayama I'm not sure if we want a separate file just for compiler flags/options. But I'm happy to refactor if this is the decision we make.

ikitayama:
It’s up to Hans’s call always.

PR #1176, opened 2019-04-11
Code formatting

clinssen:
During the NEST Hackathon 2019, it was unanimously decided to extend the global line length limit from 80 to 120 characters. This holds for C++ as well as Python code, although note that C++ code is reformatted using clang-format, whereas Python code is verified only.

We additionally decided to use the "Stroustrup" bracket code formatting rules for C++ (see https://clang.llvm.org/docs/ClangFormatStyleOptions.html).


clinssen:
Closing this in favour of a follow-up PR that will only change the maximum line length style.

Issue #1152, opened 2019-03-19
Support current versions of clang-format


jakobj:
nest requires clang-format in the exact version 3.6. the current release version in 7. the configuration file should be updated to work with the more recent versions to make life easier for people who are not stuck with, e.g., ubuntu 16.04 lts (https://www.howtoinstall.co/en/ubuntu/xenial/clang-format-3.6).

terhorstd:
This seems to be an intended upstream design decision: https://lists.llvm.org/pipermail/cfe-dev/2017-June/054351.html.
Seems like the only option is to fix the required version...

heplesser:
Closing, since there is no general solution for this due to the nature of clang-format, as pointed out [above](https://github.com/nest/nest-simulator/issues/1152#issuecomment-481333599).

Silmathoron:
@heplesser shouldn't we at least update it to the latest version so that people don't have to download and install a super old llvm and clang but just use one that is directly available on most systems (between 3.9 and 7)?
As somebody who just had to wait for everything to compile, I must say this is quite annoying...

Silmathoron:
We don't have to change the file, just decide that starting 2.18+ we use it with cland-format-X (X>=3.9)

heplesser:
@Silmathoron Version 6 seems sensible, since it is the default on a number of current Ubuntu versions. But we will always be tied to a specific version, since part of the formatting is hard-wired in clang-format. A PR would be welcome, but other issues seem more important.

Silmathoron:
@heplesser it's just that using a version which is not available anywhere anymore is a bit annoying... at least version 6 would let us breathe for a few additional months or years.
The pull request is pretty straightforward: just change a number in the file and run it, but it would mean that all other PRs should be updated accordingly.

PR #1177, opened 2019-04-11
Increase maximum line length in source code formatting style from 80 to 120

clinssen:
This is a follow-up to #1176.

The purpose of this PR is to change the static code checker rules, such that the maximum line length within *.cpp, *.h and *.py source code files is increased from 80 to 120. All *.cpp and *.h files are then reformatted with clang-format, which causes the corresponding patch (48ea8e6) to be quite large. Reviewers might like to start with the other commits in this PR first.

@jessica-mitchell: are there any places in the developer documentation that should be updated with this information?


jakobj:
while I welcome the general spirit of this change, just one small request: please give all PRs prior to this one a chance to be merged before merging this one. we've already had many, many annoying merge issues updating various branches when the 80 character limit was introduced.

clinssen:
A live discussion about this subject will be scheduled for the Open NEST Developer Conference of April 29th.

@jakobj: although I welcome the spirit of your request, given the age of many open PRs, waiting for them all to be merged before merging this one is not feasible. Note that reformatting of all C++ files (.h, .cpp) can be done in a fully automated fashion.

jakobj:
@clinssen though agreeing with your point that one can not wait for all PRs to be merged, I would like to ask that the obviously active PRs should get a chance to be merged since at least one of the maintainers (could it be me?) still has not managed to get `clang-format-3.6` running to automate the reformatting and will hence have to fight with merge issues ;)

Silmathoron:
if this is adopted (hopefully not for the python files), I guess it would be a good idea to move formatting from the antediluvian clang-format-3.6 to a more recent version (e.g 6, as discussed in #1152) in the same PR, to get all the annoying stuff done in one go ;)

clinssen:
@Silmathoron : in response to your comment on clang-format versioning, I was recently faced with the same issue. I ended up copying-and-pasting the following lines from the Travis build script:

https://github.com/nest/nest-simulator/blob/4c1e1f6bbe2d0a894057def7d0fdb76cbf26c802/extras/travis_build.sh#L133

This made the installation process quite painless. We could consider bundling the binary into our own git submodule in nest/nest-simulator, depending on the projected longevity of the llvm.org link. In addition I would recommend adding a note on this in the installation documentation. @jessica-mitchell : do you have any thoughts on what the best place for this note should be? Could we add a "Developer Tools" subheader under User Documentation → Installation Instructions?

jessica-mitchell:
@clinssen Yeah, that sounds like a good idea. As of right now, we have the developer space ghpages for dev-related topics. So I would think the best thing to do would be to add something to the developer space and link to it from the installation page.

Silmathoron:
@clinssen thanks, I just checked it and it seems to work also on non ubuntu-based plateforms, so it looks like a plan :+1:

extras/static_code_analysis.sh
```diff
@@ -53,6 +53,7 @@ IGNORE_MSG_PEP8=${16}         # true or false, indicating whether PEP8 messages
 PEP8_IGNORES="E121,E123,E126,E226,E24,E704"
 PEP8_IGNORES_EXAMPLES="${PEP8_IGNORES},E402"
 PEP8_IGNORES_TOPO_MANUAL="${PEP8_IGNORES_EXAMPLES},E265"
+PEP8_MAX_LINE_LENGTH=120
```
gtrensch:
```suggestion

# PEP8 rules.
PEP8_MAX_LINE_LENGTH=120
```

clinssen:
@jarsi noted that comments are not reformatted (e.g. `// line 1`, `// line 2` into `// line 1 line 2`) . In the future, we could consider setting `ReflowComments` to `true` in `.clang-format`. Leaving it at `false` is the more conservative option and we suggest to stick with `false` for this PR.

clinssen:
@steffengraber noted that function declarations in some files (e.g. `block_vector.h`) are formatted according to `IndentWrappedFunctionNames=true`, even though in our `.clang-format`, it is set to `false`. Possibly, these indentations were added by hand. We could consider setting `IndentWrappedFunctionNames=true` in the future.

PR #1200, opened 2019-05-13
Replace bash by sh to eliminate extra dependency

### LINKED ISSUE #1194: Please use /bin/sh in extras/nest_vars.sh.in
https://github.com/nest/nest-simulator/blob/master/extras/nest_vars.sh.in#L1

bash requires an extra-dependency on BSD platforms.

jougs:
Fixes #1194.

### DIFF
--- a/extras/nest_vars.sh.in
+++ b/extras/nest_vars.sh.in
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # NEST is installed here. When you relocate NEST, change this variable.
 export NEST_INSTALL_DIR=@CMAKE_INSTALL_PREFIX@


apeyser:
Has anyone considered that:
```
export NEST_DATA_DIR=$NEST_INSTALL_DIR/@CMAKE_INSTALL_DATADIR@
```

Is likely to be broken under POSIX sh if there are spaces or other special characters in either `$NEST_INSTALL_DIR` or `@CMAKE_INSTALL_DATADIR@` without quoting? Under bash, the results of the right hand side of an assignment tend to have lots of special rules that don't do field splitting, but the POSIX spec states that field splitting is done after parameter expansion. I'm not sure whether @CMAKE_...@ are quoted by cmake.

Requiring bash may have been covering some sins.

jougs:
@apeyser: thanks for sharing your wisdom. As this is already merged, a fix has to go in another PR.

As #1071 removes the bulk of exports from `nest_vars.sh`, I can probably sneak a fix for the remaining variables into that one.

PR #1225, opened 2019-06-07
Fix uninitialized variable in iaf_psc_exp

jakobj:
Not initializing the presynaptic stepwise constant input current `i_1_` can have disastrous consequences when the memory is not zero by chance as its initial value is directly used (https://github.com/nest/nest-simulator/blob/master/models/iaf_psc_exp.cpp#L339).


heplesser:
@jougs @terhorstd Please include this in 2.18.0 if it is not too late. A tiny but important bugfix.

PR #1241, opened 2019-06-28
Add decorator that can wrap pynest functions to accept plain GIDs

jakobj:
Are your tired of littering your beautiful pynest code with superfluous square brackets?
```python
n = nest.Create('iaf_psc_delta', 2)
nest.SetStatus([n[0]], {'I_e': 400.})
nest.Connect([n[0]], [n[1]])
```
That doesn't look very nice!
Fear no more, now you can provide plain GIDs to pynest functions!
```python
n = nest.Create('iaf_psc_delta', 2)
nest.SetStatus(n[0], {'I_e': 400.})
nest.Connect(n[0], n[1])
```
Much better!

In all seriousness, I feel like this feature existed before but was removed for consistency? I think that for good reasons one can depart from consistency (only arrays!), as long as the departure is intuitive. I would argue that this is the case here.

I've not decorated all functions yet as I first wanted to hear your thoughts @heplesser @suku248 @jougs @stinebuu @hakonsbm

### DIFF
--- a/pynest/nest/lib/hl_api_connections.py
+++ b/pynest/nest/lib/hl_api_connections.py
@@ -107,6 +107,7 @@ def GetConnections(source=None, target=None, synapse_model=None,
     return spp()
 
 
+@possibly_promote_plain_gid_to_iterable(0, 1)
 @check_stack
 def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
     """
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -202,6 +202,7 @@ def message(level, sender, text):
     sr('message')
 
 
+@possibly_promote_plain_gid_to_iterable(0)
 @check_stack
 def SetStatus(nodes, params, val=None):
     """Set parameters of nodes or connections.
@@ -269,6 +270,7 @@ def SetStatus(nodes, params, val=None):
     sr('Transpose { arrayload pop SetStatus } forall')
 
 
+@possibly_promote_plain_gid_to_iterable(0, demote_output=True)
 @check_stack
 def GetStatus(nodes, keys=None, output=''):
     """Return the parameter dictionaries of nodes or connections.
--- a/pynest/nest/ll_api.py
+++ b/pynest/nest/ll_api.py
@@ -75,6 +75,7 @@ __all__ = [
     'check_stack',
     'get_debug',
     'pcd',
+    'possibly_promote_plain_gid_to_iterable',
     'set_debug',
     'sli_func',
     'sli_pop',
@@ -202,6 +203,52 @@ def get_debug():
     """
 
     return __debug
+
+def possibly_promote_plain_gid_to_iterable(*indices, demote_output=False):
+    """Decorator to promote the function arguments at given indices to iterables if
+    they are plain GIDs, i.e., integers.
+
+    Parameters
+    ----------
+    indices: arbitrary number of indices that should be checked for promotion
+             to iterable
+
+    demote_output: whether the output iterable should be reduced to its first
+                   element if one of the inputs was promoted to an iterable
+                   default: False
+
+    Returns
+    -------
+    function
+        Decorated function
+
+    """
+
+    def outer_wrapper(func):
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+
+            new_args = list(args)  # convert to list to make arguments mutable
+            del args  # just to make sure we do not accidentally use it later
+
+            input_promoted = False
+            for idx in indices:
+                if isinstance(new_args[idx], int):
+                    new_args[idx] = (new_args[idx],)
+                    input_promoted = True
+
+            res = func(*new_args, **kwargs)
+
+            # to be consistent we only demote output if input was promoted
+            if demote_output and input_promoted:
+                return res[0]
+            else:
+                return res
+
+        return wrapper
+
+    return outer_wrapper
 
 
 def set_debug(dbg=True):


Silmathoron:
@jakobj thanks for the much needed update!
Could you make it so that it also accepts numpy arrays, there has been a long standing issue of ``Connect`` not accepting them...

hakonsbm:
@jakobj Although this looks pretty nice, we just released 2.18. And the next release will (most probably) be NEST 3, which uses GIDCollections. In NEST 3 the extra brackets are never needed as you're dealing with GIDCollections in stead of plain GIDs, and your second example would work out of the box. Therefore I unfortunately don't really see any reason to add these decorators to the current master.

You can take a look at some examples of the new features in [the NEST 3 branch](https://github.com/nest/nest-simulator/blob/nest-3/doc/nest-3/New%20features%20in%20NEST%203.ipynb), and any suggestions would be appreciated.

Silmathoron:
@hakonsbm this is something that still needs discussing (and I know should make it an item on the VC agenda ASAP): as the number of 3rd party libraries using NEST increases, planned suppression of functions such as ``GetNodes``, as well as making ``GIDCollection``s compulsory are probably not the best idea since they will make life very difficult for the developers.
Thus, I think it would be very useful to provide such a decorator, potentially changing it to a conversion to ``GIDCollection``s internally.

jakobj:
@Silmathoron regarding `Connect` not accepting numpy arrays, could you elaborate a bit please? this code seems to work for me:
```python
nest.ResetKernel()
n = nest.Create('iaf_psc_delta', 2)
n = np.array(n)
nest.Connect(n, n)
```

regarding your last comment, could you elaborate on the use cases you have in mind? are you talking about third-party tool generating neuron ranges that should be connected? in general I agree, we should make it as easy as possible for various libraries to interact with NEST.

@hakonsbm looks amazing! can't wait to use it :+1:  would be even cooler if we could finally do something like `nodes_alpha.V_th = -50.` instead of having to use the setter function explicity, but I guess this is a rather simple extension. the examples however do not seem to cover indexing `GIDCollections`: what happens if I do `nodes_alpha[0]`? is this a `GIDCollection` with just one entry?

heplesser:
@Silmathoron NEST3 will be "the big break" with significant changes to the user interface, also bringing NESTIO and a new random generator interface, making NEST more powerful and easy to use. I would much rather spend time on working with developers of third-party libraries to help them to adapt their code to NEST3 than to spend that valuable time to maintain backward compatibility.  Any 2.18.x releases should be pure bug-fix releases if they should become necessary.

hakonsbm:
@jakobj Happy that you like it! Yes, indexing GIDCollections with a single index returns a new GIDCollection with a single GID.

jougs:
I think that `GIDCollections` really make everything easier, and code like the following will also help third-party extension developers:
```python
n = nest.Create("iaf_psc_alpha", 2)
plain_gids = list(n)                 # pure Python list [1,2] usable outside of nest
gid_coll = nest.GIDCollection([1,2]) # GIDCollection containing 1 and 2
                                     # safe, as it checks if nodes actually exist
```
We might add an implicit conversion to `GIDCollections` to GID-taking functions if this is what third-party developers require.

Silmathoron:
@heplesser libraries that need to access neurons or connections post-creation without access to the ``GIDCollection`` cannot adapt their code: either NEST provides ``GetNodes``/``GetConnections`` or it doesn't ;)
This is more or less the same issue that was discussed several times regarding access to synapse properties or node degrees, and I'm afraid that current deprecation choices are going in the other direction...

@jougs yes, that's what I meant: it doesn't cost anything and does not require users or 3PL to make sure they have converted their ids to ``GIDCollection`` every time

jougs:
@Silmathoron: maybe I misunderstand something, but `GetNodes` was just an accessor function to get you the node IDs of the children in a `subnet`. As `subnet`s were never used anywhere in NEST unless you specifically told it to do so, you're not really loosing anything by its removal. With NEST-3, you still have access to all your nodes in the network plus much more flexibility, because `GIDCollections` allow you to slice and combine lists of GIDs as you widh. Regarding `GetConnections`, that function is not going away. it just also was made more powerful.

Could you maybe post a code snippet to explain what you're afraid of loosing?

Silmathoron:
@jougs up to 2.16, ``GetNodes`` returned all created gids, neurons, or devices, depending on the specified properties.
Without it, there is no way that I know of to obtain this information if you do not already have the existing ``GIDCollection``s

jakobj:
replaced by #1249

jougs:
@Silmathoron: Thanks for your input. I think what we really need then is a new `GetNodes` function that returns a `GIDCollection` containing all nodes that fulfill certain criteria using a full search through the status dicts of all nodes, possibly also involving MPI communication to gather GIDs of matching remote nodes.

@stinebuu, @hakonsbm: could you please have a look at the old `Get{Global,Local}Nodes` functions to see how complex a re-implementation would be?

stinebuu:
@jougs, @Silmathoron Just so I know exactly what we need, we need to re-implement the `GetNodes` function with some sort of properties variable? I have not really used `GetNodes` this way before, so I just want to make sure I understand how it is used :sweat_smile: 

Looking at the docstring of the `GetNodes` function in current master, you can call `GetNodes` with a properties dictionary, so I guess this is what we need to re-implement, just that instead of the subnet, you can pass a `GIDCollection`.

https://github.com/nest/nest-simulator/blob/e4f6b328354eec859b1f953e495f9c133c4464ce/pynest/nest/lib/hl_api_subnets.py#L170-L181

So if we need all nodes that have membrane potential equal to -77 for instance, we pass this to `GetNodes`?

Silmathoron:
@stinebuu now that ``subnets `` are gone, I would say to remove the positional argument altogether so the function becomes:

```python
def GetNodes(properties=None, local_only=False): 
     """Return the global ids of the all nodes of the given subnets. 
  
     Parameters 
     ---------- 
     properties : dict, optional 
         Only global ids of nodes matching the properties given in the 
         dictionary exactly will be returned. Matching properties with float 
         values (e.g. the membrane potential) may fail due to tiny numerical 
         discrepancies and should be avoided.
```

otherwise yes, that would be great, thank you!

stinebuu:
@Silmathoron  Excellent!

PR #1244, opened 2019-07-01
Add support for sending secondary events to and from devices

jakobj:
So far, secondary events could only be exchanged between neurons. This PR adds the necessary changes to make sure that secondary events can be sent to/received from devices.

I suggest @janhahne @suku248 as reviewers

branches off from #1243, so should be merged after


jakobj:
@janhahne thanks for the review, i now rebased on the current master

janhahne:
@jakobj Thanks! This PR now actually allows to remove the proxies for the `step_rate_generator` device which uses rates. I created a PR against your branch. Everything works well, such that I will approve once this is included here :+1:

jakobj:
@janhahne i've integrated your changes. let's give travis another run before this PR is merged

PR #1249, opened 2019-07-02
Add getattr/setattr to GIDCollection

jakobj:
Defining `__getattr__` and `__setattr__` allows more pythonic access of node properties:
```python
import nest

nodes = nest.Create('iaf_psc_alpha', 10)

print('V_m', nodes.V_m)
nodes.V_m = np.random.normal(-60., 1., 10)
print('V_m', nodes.V_m)
```

@hakonsbm 

edit: here's a more interesting example
```python
import nest

nodes = nest.Create('iaf_psc_alpha', 10)
nodes.V_m = np.random.normal(-60., 1., 10)
nodes.I_e = 376.

m = nest.Create('multimeter')
m.record_from = ('V_m',)

nest.Connect(m, nodes[0])

nest.Simulate(100.)

plt.plot(m.events['times'], m.events['V_m'])
plt.show()
```

### DIFF
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -568,16 +568,12 @@ class GIDCollection(object):
         return self.get(attr)
 
     def __setattr__(self, attr, value):
-        # check whether the attribute is a node property; need to manually
-        # exclude _datum since without it, get() does not work and we can not
-        # check whether attr is a node property
-        if attr != '_datum':
-            if attr in self.get().keys():
-                self.set({attr: value})
-                return  # if we have set the property in NEST, we are done
-
-        super().__setattr__(attr, value)  # if attr is not a node property,
-                                          # store it as regular attribute
+        # `_datum` is the only property of GIDCollection that should not be
+        # interpreted as a property of the model
+        if attr == '_datum':
+            super().__setattr__(attr, value)
+        else:
+            self.set({attr: value})
 
 
 class ConnectomeIterator(object):
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -564,6 +564,21 @@ class GIDCollection(object):
         """
         return self.tolist().index(gid)
 
+    def __getattr__(self, attr):
+        return self.get(attr)
+
+    def __setattr__(self, attr, value):
+        # check whether the attribute is a node property; need to manually
+        # exclude _datum since without it, get() does not work and we can not
+        # check whether attr is a node property
+        if attr != '_datum':
+            if attr in self.get().keys():
+                self.set({attr: value})
+                return  # if we have set the property in NEST, we are done
+
+        super().__setattr__(attr, value)  # if attr is not a node property,
+                                          # store it as regular attribute
+
 
 class ConnectomeIterator(object):
     """


pynest/nest/lib/hl_api_types.py
```diff
@@ -547,6 +547,21 @@ def set(self, params, val=None):
...
+        # check whether the attribute is a node property; need to manually
+        # exclude _datum since without it, get() does not work and we can not
+        # check whether attr is a node property
+        if attr != '_datum':
+            if attr in self.get().keys():
+                self.set({attr: value})
+                return  # if we have set the property in NEST, we are done
+
+        super().__setattr__(attr, value)  # if attr is not a node property,
+                                          # store it as regular attribute
```
stinebuu: @jakobj This is super neat!

I don't think I quite understand why we need this last point tough. It is there so that we can set whatever we want on the GIDCollection if it is not a node property, do you have an example of what that could be? In principle I think this sounds a little unsafe, but it can be that I don't see all the things it can be used for.

jakobj: hmm, I thought it would still be nice to be able to set instance variables like so
```python
nodes = nest.Create('iaf_psc_alpha', 10)
nodes.my_variable = 'some value'
```
in order for this to work we need to use the `__setattr__` on the superclass as a far as I understand. however, you might be right that it could potentially lead to confusion as no error message would be raised if someone had a typo like `nodes.Vm = np.random.normal(-60., 1., 10)`.  this would just store create a new instance variable `Vm` but not actually affect the node properties in NEST.

jougs: @jakobj: I quite like the idea of allowing the user to store additional properties for a `GIDCollection` purely on the Python level. However, I also see the danger you mention (that was the original reason for adding [access checks](https://github.com/nest/nest-simulator/blob/8c6843c50e8944e829b72a1c1555133c0f9164fe/sli/dict.h#L168) for SLI dictionaries). We should at least discuss this in a wider group and I recommend to remove it from this PR for now.

jakobj:
removing this check `if attr in self.get().keys():` to allow NEST to throw an error upon encountering an unknown attribute now generates a strange error message: `nest.ll_api.ArgumentType: ('ArgumentType in Take: The type of the first parameter did not match the argument(s) of this function.', 'ArgumentType', 'Take', ': The type of the first parameter did not match the argument(s) of this function.')` as during `Create` and attribute-value pair is passed: `spatial None` @stinebuu do you know what might cause this?

stinebuu:
@jakobj Yes, the PyNEST GIDCollection class has a spatial parameter for layer metadata in case the GIDCollection is a layer. This spatial parameter is defined a little clumsily, and this affects this PR. I have updated it now in the nest-3 branch, I got super inspired by you PR, so thanks for that! Could you pull in the newest version of the branch? I think that should fix your problem.

jakobj:
happy to help :) thanks, i've now rebased on your new version, it seems to work. do you think we could handle `spatial` in a similar way as we handle `_datum` separately? or do you maybe have a better idea for handling `_datum`?

jakobj:
i needed to treat this separately to handle the setting of `_datum` internally, see https://github.com/nest/nest-simulator/blob/55881d7b9247d7d1de8dcb07e678e0e2843b24a2/pynest/nest/lib/hl_api_types.py#L307 ff. i think it's clear from the naming scheme (leading underscore) that users should not modify this variable since it will breaks things.

stinebuu:
@jakobj Ok, that makes sense.

There seems to be a problem with python 2 compatibility so the tests fails. I had this problem as well, could you add `GIDCollection, self` to the `super()` call:
```
super(GIDCollection, self).__setattr__(attr, value)
```
`super()` takes an extra variable for Py2.

jougs:
I'm not sure if we should actually fix this Python 2 thing, or rather create a PR against the nest-3 branch removing all Python 2 builds from Travis and all Python 2/3 hacks from the code. Python 2 will anyway die when NEST-3 starts to live...

heplesser:
@jougs A good point. NEST3 == Python3. I will put it on the agenda for the Open Developer Meeting on Monday.

jougs:
@heplesser: no need to discuss this again. We reached critical mass agreement on this at the last hackathon.

jakobj:
ok, so i don't need to fix anything in this PR?

jougs:
Most likely no. We just wait until @clinssen removed Python 2 support, you merge `nest-3` and we're good to go.

stinebuu:
A problem has occurred on the nest-3 branch that (I think) would be fixed pretty nicely with the `__setr__` from this PR. I therefore very much want to merge it, and suggest that we remove the Python 2 testing from Travis for the nest 3 branch, so that no tests fail because of Python 2. @clinssen can then move all other Python 2 support once he has time, but we can merge this PR right after we have removed the Python 2 check in Travis.

Does anyone have any objections to this? If not, I will go ahead and remove Python 2 from our nest-3 Travis setup.

jougs:
@stinebuu, I am fine with your proposal.

PR #1250, opened 2019-07-02
Add warning about invalid connection descriptors when new connections are created

### LINKED ISSUE #1224: Connection descriptors are invalidated after adding new connections, but still accepted by GetStatus
1. When a new connection is created, the connection descriptors (`source-gid, target-gid, target-thread, synapse-id, port`) are rearranged, and ids of the existing connections may change. In this case descriptors returned by GetConnections() before become incorrect.

2. Such incorrect descriptors are still accepted by GetStatus(), because the latter seems to look only at the synapse id, ignoring the fact that the other components of the descriptor are inconsistent. This may lead to quite confusing output.

A minimal reproducing example:
```
other_neurons_ids = nest.Create('parrot_neuron', 1)  # gets id 1
neurons_ids = nest.Create('parrot_neuron', 1)  # gets id 2

# note that we first use the neuron with higher id, this is the cause!
nest.Connect(
	pre=neurons_ids,
	post=neurons_ids,
	conn_spec='one_to_one',
	syn_spec={'weight': 0}
)

connections_ids = nest.GetConnections()
print(
	connections_ids,
	nest.GetStatus(connections_ids, 'weight')
)
```
outputs:
```
(array('l', [2, 2, 0, 0, 0]),)
{'weight': 0.0, 'source': 2, 'target': 2, ...}
```

Now we create one more connection.
```
nest.Connect(
	pre=other_neurons_ids,
	post=other_neurons_ids,
	conn_spec='one_to_one',
	syn_spec={'weight': 2}
)
print(nest.GetConnections())
```
Now the descriptors are all different:
`(array('l', [1, 1, 0, 0, 0]), array('l', [2, 2, 0, 0, 1]))`
Trying to query the synapse by a descriptor stored in advance
```
print(nest.GetStatus(connections_ids))
```
outputs the nonsensical
```
{'weight': 2.0, 'source': 2, 'target': 1, ...}
```
where the 'weight' and 'target' properties are the ones of the synapse which now has id 0, and 'source' has remained intact from the query.

If this is not a bug, I would have suggested the following:
1. Document that storing the return values of GetConnections() for future use is discouraged;
2. Fail GetStatus() if the query descriptor is inconsistent, i.e. its other fields do not match its id.

jakobj:
fixes #1224

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -426,13 +426,7 @@ nest::ConnectionManager::connect( const index sgid,
 {
   kernel().model_manager.assert_valid_syn_id( syn_id );
 
-  // Need to check if have_connections_changed_ has already been set, because if
-  // we have a lot of threads and they all try to set the variable at once we get
-  // performance issues on supercomputers.
-  if ( not have_connections_changed_[ target_thread ] )
-  {
-    have_connections_changed_.set( target_thread, true );
-  }
+  set_have_connections_changed( target_thread );
 
   Node* const source = kernel().node_manager.get_node( sgid, target_thread );
   const thread tid = kernel().vp_manager.get_thread_id();
@@ -509,13 +503,7 @@ nest::ConnectionManager::connect( const index sgid,
 
   const thread tid = kernel().vp_manager.get_thread_id();
 
-  // Need to check if have_connections_changed_ has already been set, because if
-  // we have a lot of threads and they all try to set the variable at once we get
-  // performance issues on supercomputers.
-  if ( not have_connections_changed_[ tid ] )
-  {
-    have_connections_changed_.set( tid, true );
-  }
+  set_have_connections_changed( tid );
 
   if ( not kernel().node_manager.is_local_gid( tgid ) )
   {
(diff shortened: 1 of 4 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -1570,3 +1570,14 @@ nest::ConnectionManager::check_secondary_connections_exist()
 {
   secondary_connections_exist_ = kernel().mpi_manager.any_true( secondary_connections_exist_ );
 }
+
+void
+nest::ConnectionManager::set_have_connections_changed( const bool changed )
+{
+  if ( ( not have_connections_changed_ ) and changed )
+  {
+    std::string msg = "New connections created, old connection descriptors from GetConnections are now invalid.";
```
jougs:
```suggestion
    std::string msg = "New connections created, connection descriptors previously obtained using 'GetConnections' are now invalid.";
```

terhorstd:
@jakobj what do you think of @jougs suggestion?

jakobj:
looks good @jougs @terhorstd 

should i squash the commits before merging?

jougs:
@jakobj: The change I suggested apparently lead to a line with more than 120 characters. Could you please break/change it as suggested by clang-format? I'll approve and merge once this is fixed and the tests succeed. Thanks!

jakobj:
i've now updated this PR, mainly needed to integrate #1276 

@clinssen @jougs please have another look

PR #1251, opened 2019-07-02
Fix GetConnections to correctly get connections where the target is a global receiving device

### LINKED ISSUE #1190: GetConnections() does not work with volume_transmitter as target
When retrieving the connections to a volume_transmitter, `GetConnections()` behaves inconsistently.

```
A=nest.Create("iaf_cond_exp")
VT=nest.Create("volume_transmitter")
nest.Connect(A,VT)
nest.GetConnections(A,VT)
```
returns: ()
```
nest.GetConnections()
```
returns: (array('l', [1, 2, 0, 0, 0]),)

In `788e93e` , with `NEST 2.14` GetConnections() is working fine.

I will try to locate the source of the problem, but maybe someone familiar with the function (e.g., @hakonsbm, who refactored GetConnections in #1099 ) could identify the issue immediately.

hakonsbm:
Fixes an issue with GetConnections, where it would not get the connection if the specified target is a global receiving device. 

In jakobj#104, to improve the performance of GetConnections, targets are split into vectors containing neurons and devices. However, even though the global receiving device target is a device, it is connected in the same way as a normal neuron node. This means that the connection is never added to `TargetTableDevices`, and the connection isn't found.

This PR changes the splitting function to put global receiving devices in the neuron vector, which allows GetConnections to correctly get the connections. Two tests are also added, checking that GetConnections with specified source and target always gets the connection.

Fixes #1190.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1091,7 +1091,11 @@ nest::ConnectionManager::split_to_neuron_device_vectors_( const thread tid,
   for ( size_t t_id = 0; t_id < gid_token_array->size(); ++t_id )
   {
     const index gid = gid_token_array->get( t_id );
-    if ( kernel().node_manager.get_node( gid, tid )->has_proxies() )
+    const auto node = kernel().node_manager.get_node( gid, tid );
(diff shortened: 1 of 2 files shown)


heplesser:
@suku248 Could you review this PR in the near future? It is one of two PRs for 2.18.1 still open.

Issue #1242, opened 2019-06-28
[stdp_dopamine_synapse] Multi-threading results on different dopamine concentrations


gui-miotto:
**Description of problem**: The dopamine concentration seems not to be correctly calculated when running on multi-thread.

**Version-Release or Git commit**: 2.16.0 (0b07c84) and 2.18.0 (2ee42f9)

**How reproducible**: Always

**Steps to Reproduce**: Just run the following script.
```python
import nest
import numpy as np
import matplotlib.pyplot as plt

def mwe(n_threads):
    # kernel configuration    
    nest.ResetKernel()
    nest.SetKernelStatus({'local_num_threads' : n_threads })

    # create nodes
    neurons = nest.Create('iaf_psc_delta', 2)
    vt = nest.Create('volume_transmitter')
    DA_spikes = nest.Create('spike_generator', params={'spike_times' : [999.]})

    # Connections
    nest.CopyModel('stdp_dopamine_synapse', 'dopa_syn', params={
        'n' : .005, 
        'vt' : vt[0],
        })
    nest.Connect(DA_spikes, vt)
    nest.Connect(neurons, neurons, syn_spec='dopa_syn')
    synapse = nest.GetConnections(neurons[:1], neurons[1:])

    # Simulate and store dopamine concentration
    delta_t = 5.
    time = np.arange(0., 2000., delta_t)
    DA_n = []
    for _ in time:
        DA_n.append(nest.GetStatus(synapse, 'n'))
        nest.Simulate(delta_t)
    
    # plot dopamine concentration
    plt.plot(time, DA_n)
    plt.suptitle(f'Running with {n_threads} local thread(s)')
    plt.ylabel('dopamine concentration')
    plt.xlabel('time (ms)')
    plt.show()

mwe(2)
mwe(1)
```
**Actual results**:
![2threads](https://user-images.githubusercontent.com/9371512/60347045-c89acb80-99bc-11e9-9bad-e31feb3155e2.png)

**Expected results**:
![1thread](https://user-images.githubusercontent.com/9371512/60347099-e2d4a980-99bc-11e9-8590-e670f47e7596.png)

**Additional info**: Related to #410. But this issue has been long closed.

gui-miotto:
And apparently there is a test for that: [testsuite::issue-410](https://www.nest-simulator.org/helpindex/sli/testsuite::issue-410.html)

gui-miotto:
Just tested with 2.18 
Issue persists

[...]

PR #1254, opened 2019-07-12
Fix logical error in elseif-statements in connect functions

janhahne:
Issue #1242 is a symptom of this logical error in the elseif-state in the connect functions in `connection_manager.cpp`. Currently the case `not source->has_proxies() and not target->has_proxies()` is handled before the case `not target->has_proxies() and not target->local_receiver()` which has a subcase `if ( not source->has_proxies() )` which therefore is not reachable.

This results in wrong behavior in a case where `not source->has_proxies() and not target->has_proxies() and not target->local_receiver()` holds (as for example in the reproducer of #1242). In addition a source code comment indicates that this case is not implemented at all at the moment. Therefore this PR also adds  a `throw IllegalConnection` here, to prevent a silent error where the connect call is executed without creating any connection.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -464,6 +464,18 @@ nest::ConnectionManager::connect( const index sgid,
   {
     connect_from_device_( *source, *target, target_thread, syn_id, params, delay, weight );
   }
+  // globally receiving devices, e.g. volume transmitter
+  else if ( not target->has_proxies() and not target->local_receiver() )
+  {
+    // we do not allow to connect a device to a global receiver at the moment
+    if ( not source->has_proxies() )
+    {
+      throw IllegalConnection( "We do not allow to connect a device to a global receiver at the moment" );
+      return;
+    }
+    target = kernel().node_manager.get_node( target->get_gid(), tid );
+    connect_( *source, *target, sgid, tid, syn_id, params, delay, weight );
+  }
   // normal devices -> normal devices
   else if ( not source->has_proxies() and not target->has_proxies() )
   {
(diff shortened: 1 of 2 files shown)


jakobj:
thanks for catching this @janhahne. could you also please add a test based on #1242 ?

janhahne:
I added a regressiontest. However this can only ensure that such a connection cannot be created. If we decide to implement such connections at some point the reproducer from #1242 would be a good candidate to ensure consistent results with different numbers of threads.

nestkernel/connection_manager.cpp
```diff
@@ -547,28 +548,29 @@ nest::ConnectionManager::connect( const index sgid,
...
-    }
-  }
   // globally receiving devices, e.g. volume transmitter
   else if ( not target->has_proxies() and not target->local_receiver() )
   {
     // we do not allow to connect a device to a global receiver at the moment
     if ( not source->has_proxies() )
     {
+      throw IllegalConnection( "We do not allow to connect a device to a global receiver at the moment" );
       return false;
```
stinebuu: Same as above with the `return` statement.

nestkernel/connection_manager.cpp
```diff
@@ -464,6 +464,18 @@ nest::ConnectionManager::connect( const index sgid,
...
     connect_from_device_( *source, *target, target_thread, syn_id, params, delay, weight );
   }
+  // globally receiving devices, e.g. volume transmitter
+  else if ( not target->has_proxies() and not target->local_receiver() )
+  {
+    // we do not allow to connect a device to a global receiver at the moment
+    if ( not source->has_proxies() )
+    {
+      throw IllegalConnection( "We do not allow to connect a device to a global receiver at the moment" );
+      return;
```
stinebuu: Maybe remove the `return` as I don't think it makes sense to return after an exception has been thrown?

testsuite/regressiontests/issue-1242.sli
```diff
@@ -0,0 +1,53 @@
...
+FirstVersion: July 2019
+*/
+
+(unittest) run
+/unittest using
+
+M_ERROR setverbosity
+
+% Check if such a connections results in an exception
+{
```
stinebuu: Connection should be singular (so no s)

janhahne:
@stinebuu Thanks for the review! I addressed your comments. Regarding your question: As this is explicitly caught as a "special" case (before calling the regular connect) I just assumed that this cannot be the solution here, but I will give it a try.

PR #1255, opened 2019-07-17
Files needed to build Debian packages (.deb)

steffengraber:
This PR provides all files needed to build local Debian packages or to integrate into a "Personal Package Archive" (PPA) on launchpad.net. Launchpad automatically builds binary packages for different Ubuntu versions and simplifies the installation of NEST.


debian/nest.postinst
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+rm -f /etc/profile.d/nest-simulator.sh
+touch /etc/profile.d/nest-simulator.sh
+cat >>/etc/profile.d/nest-simulator.sh <<EOF
+export NEST_INSTALL_DIR='/usr'
+export NEST_DATA_DIR='/usr/share/nest'
+export NEST_DOC_DIR='/usr/share/doc/nest'
+EOF
```
jougs: Can't we just source `nest_vars.sh` here? That would probably be more stable in the long run.

debian/source/format
```diff
@@ -0,0 +1 @@
+3.0 (quilt)
```
jougs: Please add a newline to the end of the file.

debian/rules
```diff
@@ -1,105 +1,40 @@
...
+override_dh_auto_build:
+override_dh_auto_install:
+    # This is a little bit tricky
+    # @todo Check if this really is problem or only in building the packages
+	sed -i 's/site-packages/dist-packages/g' ./cmake/ProcessOptions.cmake
+	sed -i 's/NEST_INSTALL_PREFIX \"\/\"/\"\/usr\/"/g' ./sli/slistartup.cc
+	cmake $(CMAKE_FLAGS) .
+	make
+	make install
+	rm -f $(DESTDIR)/usr/bin/nest_serial $(DESTDIR)/usr/bin/nest_indirect $(DESTDIR)/usr/bin/sli $(DESTDIR)/usr/bin/nest_vars.sh
```
jougs: Why are you deleting these? At least `sli` is useful to get the helpdesk shown if you compile with MPI.

debian/nest.postinst
steffengraber: No. NEST was created on another system within another folder structure, so  `nest_vars.sh` does not contain the correct values.

debian/rules
steffengraber: You're totally right. This line can go away.
Instead I have inserted a line in which `nest_wars.sh` is adjusted and the correct paths are set. Then `nest_vars.sh` is sourced in `etc/profile.d/nest-simulator.sh` as you suggested.

Since NEST is installed in the default paths on the system, it will work after installation. 
If you have problems viewing the documentation, the paths must be set explicitly. There are now two simple ways to do this:

1. log off and on again, or
2. source `nest_vars.sh`  with `. nest_ vars.sh`

PR #1257, opened 2019-07-23
Add Urbanczik-Senn synapse and supported neuron model

jstapmanns:
This PR introduces the `urbanczik_synapse` described in Urbanczik and Senn 2014 to NEST. The PR includes

- the synapse model `urbanczik_synapse`
- an extended archiving node `Urbanczik_Archiving_Node` that provides
the archiving that is needed for the Urbanczik synapse
- a new two-compartment neuron models `pp_cond_exp_mc_urbanczik`
that uses `Urbanczik_Archiving_Node` and  supports the new synapse model
- an additional property `requires_urbanczik_archiving` for synapses that ensures that Urbanczik
synapses can only be created with supported neuron models
- an unit tests `test_urbanczik_synapse.py` and an example `urbanczik_synapse_example.py` for the new synapse
- some required adaptations of other unit tests


terhorstd:
The last few commits settled last discussion. Reviewers can now commence.

clinssen:
@jstapmanns @janhahne Thanks for running me through the code and model! My comments are collected below.

- get_urbanczik_history() is currently a member of the Node class, but might belong better in Urbanczik_Archiving_Node. Same goes for get_C_m(), get_g_L() etc. Consider using dynamic_cast<> to cast from Node to Urbanczik_Archiving_Node in cases where you know for certain that your Node* is actually a Urbanczik_Archiving_Node*. (dynamic_cast<> works also for pointers!)
- Why is Urbanczik_Archiving_Node templated with the urbanczik_parameters class? In case there is more than one set of parameters that might apply (urbanczik_parameters1, urbanczik_parameters2, ...), consider using a purely virtual base class, and inheriting from this. The Urbanczik_Archiving_Node would then simply have a pointer to the base class. Possibly, dynamic_cast<> will be helpful here as well.
- The magic number 0.03 in the unit test seems a little big, although the absolute value of the weight is also rather large (~100). If you are convinced that 0.03 is sufficiently strict, consider making this a relative rather than absolute error margin.

janhahne:
@clinssen Thank you for the review! We adapted the unit test and got rid of the magic number, please have another look.

> get_urbanczik_history() is currently a member of the Node class, but might belong better in Urbanczik_Archiving_Node. Same goes for get_C_m(), get_g_L() etc. Consider using dynamic_cast<> to cast from Node to Urbanczik_Archiving_Node in cases where you know for certain that your Node* is actually a Urbanczik_Archiving_Node*. (dynamic_cast<> works also for pointers!)

Currently this is not possible, as Urbanczik_Archiving_Node is a template class and you would have to provide the template instantiation, which is not available in nestkernel (and can also not be determined there).

> Why is Urbanczik_Archiving_Node templated with the urbanczik_parameters class? In case there is more than one set of parameters that might apply (urbanczik_parameters1, urbanczik_parameters2, ...), consider using a purely virtual base class, and inheriting from this. The Urbanczik_Archiving_Node would then simply have a pointer to the base class. Possibly, dynamic_cast<> will be helpful here as well.

I actually started to implement this interesting suggestion, but then ran into the problem that the size of  `std::deque< histentry_extended > urbanczik_history_[ urbanczik_parameters::NCOMP - 1 ];` depends on the number of compartments of the used neuron model (that is saved in the template class `urbanczik_parameters`). The idea is that the `Urbanczik_Archiving_Node` should automatically work with different neuron models with a different number of compartments. At the moment I do not see any way to achieve this without the template, unless you also have an idea for that?

rcfduarte:
Thank you @jstapmanns for clarifying my doubts and walking me through the code and sorry for the delay in providing my review. The model implementation is very elegant and the code is equally well-written and documented. I see no major problem and so my review will be short. Below, I send some comments/suggestions.

* In the paper, all units are dimensionless. While checking the conversions you introduced, I realized that you already provide all the corrected values in `urbanczik_synapse_example.py`. However, running the example with these values leads to incorrect results where the signals don’t have the right amplitudes (see attached figures). I couldn’t figure out what was causing this difference, but it’s likely some incorrect scaling parameter. Even though this is only a small example case, it might be worth checking

no-units
units

* Since the neuron model has a single dendritic compartment, it seems confusing to refer to it as “proximal”. If the model was composed of proximal and distal dendrites, then this segregation would make more sense, but having a single dendritic compartment, I think it would be more intuitive (from the user perspective) to separate the parameters as “soma” and “dendrite”. Of course, this doesn’t really make any difference, it’s just a suggestion.

jstapmanns:
Thank you @rcfduarte for reviewing our code so carefully and pointing us to the two issues which we addressed in the updated version of this pull request.

> In the paper, all units are dimensionless. While checking the conversions you introduced, I realized that you already provide all the corrected values in urbanczik_synapse_example.py. However, running the example with these values leads to incorrect results where the signals don’t have the right amplitudes (see attached figures). I couldn’t figure out what was causing this difference, but it’s likely some incorrect scaling parameter. Even though this is only a small example case, it might be worth checking

We resolved this problem by adjusting the learning rate `eta`. Now the learning should work also with the original units used in the paper.
> Since the neuron model has a single dendritic compartment, it seems confusing to refer to it as “proximal”. If the model was composed of proximal and distal dendrites, then this segregation would make more sense, but having a single dendritic compartment, I think it would be more intuitive (from the user perspective) to separate the parameters as “soma” and “dendrite”. Of course, this doesn’t really make any difference, it’s just a suggestion.

You are absolutely right that the naming of the dendritic compartment is not intuitive. Therefore, we changed all variable names `proximal` &rarr; `dendritic` and `PROX` &rarr; `DEND` in the code. Moreover, we modified the documentation in the comments accordingly.

heplesser:
Travis tests on [the merge of this branches' 4c55c9f into master at e018635](7f62b4be56b646ecbf6dfa967a485fd6ee582950) has passed [all checks on Travis](https://travis-ci.org/github/nest/nest-simulator/builds/688927829), but somehow this information has not propagated back to Github here, showing tests still pending. But since all is fine, I will proceed and merge.

PR #1260, opened 2019-07-31
Add initial membrane potential options to microcircuit example

HanjiaJiang:
To avoid the transient firing burst and fluctuation in the first 10 ms of simulation, options of optimized initial membrane potentials are added for both sli and pynest microcircuit example code. The optimized membrane potential distributions were measured in the stationary state (100th to 200th ms) for all neuron populations in the model.
In the network parameter file, the user can choose either the original initial values from the previous version, or the optimized initial values. A warning message is printed in the terminal or the log if the user misspells one of the options.

This is the same content as the pull request #1206, which had a wrong git history due to a corrupted rebase. The current pull request has a clean git history.

### DIFF
--- a/examples/nest/Potjans_2014/network_params.sli
+++ b/examples/nest/Potjans_2014/network_params.sli
@@ -101,8 +101,40 @@ def
 /neuron_model /iaf_psc_exp def  % neuron model. For PSP-to-PSC conversion to
                                 % be correct, synapses should be current-based
                                 % with an exponential time course
-/Vm0_mean -58.0 def             % mean of initial membrane potential (mV)
-/Vm0_std 10.0 def               % std of initial membrane potential (mV)
+
+% Initial conditions for the membrane potential, options are:
+% (original): uniform mean and std for all populations
+% (optimized): population-specific mean and std, allowing a reduction of the
+% initial activity burst in the network
+/Vm0_type (optimized) def        % choose either (original) or (optimized)
+Vm0_type (optimized) eq
+{
+    % optimized
+    % mean of initial membrane potential (mV)
+    /Vm0_mean [[-68.28   % 2/3e
+                -63.16]  % 2/3i
+               [-63.33   % 4e
+                -63.45]  % 4i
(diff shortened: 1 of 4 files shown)


examples/nest/Potjans_2014/network_params.sli
```diff
@@ -101,8 +101,40 @@ def
...
                                 % be correct, synapses should be current-based
                                 % with an exponential time course
-/Vm0_mean -58.0 def             % mean of initial membrane potential (mV)
-/Vm0_std 10.0 def               % std of initial membrane potential (mV)
+
+% Initial conditions for the membrane potential, options are:
+% (original): uniform mean and std for all populations
+% (optimized): population-specific mean and std, allowing a reduction of the
+% initial activity burst in the network
+/Vm0_type (optimized) def        % choose either (original) or (optimized)
```
stinebuu: I think that the default value should be set to `original`, as users might not be aware that the example has changed.

pynest/examples/Potjans_2014/network_params.py
```diff
@@ -195,12 +195,22 @@ def get_std_PSP_matrix(PSP_rel, number_of_pop):
     # Relative standard deviation of the delay of excitatory and
     # inhibitory connections (in relative units).
     'rel_std_delay': 0.5,
+    # Initial conditions for the membrane potential, options are:
+    # 'original': uniform mean and std for all populations.
+    # 'optimized': population-specific mean and std, allowing a reduction of
+    # the initial activity burst in the network.
+    # Choose either 'original' or 'optimized'.
+    'V0_type': 'optimized',
```
stinebuu: Again, I think the default value should be 'original'

HanjiaJiang:
> Thanks for this! My only objection is that I think the default value should be the 'original' one, and not the 'optimized' one, see my comments below. This is because I think users should be able to run the example as they always have done and get the same results, or else it might lead to confusions.

I agree, thanks a lot for this opinion. I revised it accordingly.

PR #1262, opened 2019-08-02
Add new options for installation and improve instructions for NEST

jessica-mitchell:
This pull request incorporates several changes to the 'onboarding and installation-related'  pages of documentation.
This includes new instructions for several 'package managers', a troubleshooting section, and a modified getting-started page. **See output here:**  https://nest-test.readthedocs.io/en/improve_installation/installation/index.html 
1) Added tabs so users can switch between desired install procedure 
2) Currently the tabs are split into two main groups, standard install and advanced
   2.1) Standard install include the Ubuntu PPA, docker, Conda and the live media (windows)
-   I do not have the brew instructions explicitly here because they have not been updated since 2.14, and we've had problems installing nest on Mojave with this method

   2.2) Advanced install includes a 
-  link to the instructions to compile nest on Linux, 
-   the mac installation instructions (these instructions may need to be checked because of difficulties installing with this method on Mojave)
-    link to GitHub and the workflow for GitHub
-    link to HPC instructions (these need to be assessed if they are still valid) 
3) The installation page links to two main further sections - the [getting-started page,](https://nest-test.readthedocs.io/en/improve_installation/getting_started.html) which shows a simple simulation in nest, along with a more detailed breakdown of that example so new users can get a sense of the commands in nest. Thanks to  @AlexVanMeegen for some of his tutorial material.
There is also a [troubleshooting page](https://nest-test.readthedocs.io/en/improve_installation/troubleshooting.html) that tries to cover the main issues people  have when installing nest, namely environment and package issues. 
We regularly have questions from users because of misunderstandings with installation environments. The idea with the troubleshooting page is to have a place to point people when these questions arise, that hopefully provide some useful tips, so they can fix their environment.
@terhorstd and I have discussed this, and we  think the troubleshooting page could be a separate entity from a FAQ page, where the FAQ is focused on NEST specific questions and the troubleshooting page is aimed at  system or environment issues.
4) The [download page ](https://nest-test.readthedocs.io/en/improve_installation/download.html) has been updated with link to 2.18 versions


AlexVanMeegen:
Looks great! Thanks a lot for this. Minor comments: 

1. To me,
```
python
import nest
```
is a bit confusing; I'd prefer to split the two commands to make it explicit that the latter is run in the python shell.

2. In the first sentence in the `getting_started` page is a duplicate question mark: "Have you already installed NEST??"

3. Are there somewhere instructions how to get the virtual machine fired up on windows? Just to give a bit more than the plain download link.

PR #1266, opened 2019-08-09
Allow specifying arrays of values in the synapse specification when connecting with fixed_total_number

hakonsbm:
This PR makes it possible to specify arrays of values in the synapse specification when connecting with a fixed total number of connections in PyNEST. This is already possible from SLI, so there are only some small changes on the PyNEST level.

### DIFF
--- a/pynest/nest/lib/hl_api_connections.py
+++ b/pynest/nest/lib/hl_api_connections.py
@@ -194,12 +194,16 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
     except for 'receptor_type' which must be initialised with an integer.
 
     Parameter arrays are available for the rules 'one_to_one',
-    'all_to_all', 'fixed_indegree' and 'fixed_outdegree':
+    'all_to_all', 'fixed_total_number', 'fixed_indegree' and
+    'fixed_outdegree':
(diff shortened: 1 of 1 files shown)


pynest/nest/lib/hl_api_connections.py
```diff
@@ -303,12 +303,21 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, model=None):
                                     "scalar or a dictionary.")
                             else:
                                 syn_spec[key] = value
+                        elif rule == 'fixed_total_number':
+                            if ('N' in conn_spec
+                                    and value.shape[0] != conn_spec['N']):
```
stinebuu: Can you use `fixed_total_number` without giving `N`? I don't think you need to check if `N` is in `conn_spec`.

hakonsbm: You're right in that `N` is required when using `fixed_total_number`, but I did it this way to make the error consistent. If we don't check if `N` is specified here, we would get a Python `KeyError`, while usually when using `fixed_total_number` without `N` we would get a `NESTErrors.DictError`.

stinebuu: @hakonsbm Ok!

Issue #1246, opened 2019-07-01
Static code checks should be run after all other tests


jakobj:
As far as I can see no other test is run if the static code checks fail. This seems suboptimal to me as the more important information early in the review process is whether the changes break existing functionality and not whether some curly brace is on the correct line. Would it be possible to change this order?

@lekshmideepu

Silmathoron:
The reason it is performed like this is to minimize the amount of wasted CPU time, I would strongly advise to keep it that way...

jakobj:
could you elaborate what you mean by "wasted CPU time"? my naive approach would be that code style is the last thing that should be checked before the PR is merged, not the first thing? is the CPU time wasted because travis shows me that my code is leading to unexpected results if my formatting is off?

Silmathoron:
Suppose we change the order and your code is fine:
- Travis runs the tests (CPU1)
- Travis checks the style (CPU2) and breaks
You will fix the style and push.
- Travis runs the tests (CPU1)
- Travis checks the style (CPU2) and validate
So we ran 2*CPU1+2*CPU2 which is roughly 2*CPU1

With our current setup, the first CPU1 goes away so we divide CPU consumption by 2.

clinssen:
At present, a full build of NEST is performed in the "static check" stage. Perhaps it would be possible to dramatically speed things up by eliminating the build from the static checker stage. I cannot see the results from the build being used in this stage, but I could be overlooking something. See https://github.com/nest/nest-simulator/blob/0ac2385db24bdf76c43f310ee5f7da10d274c457/extras/travis_build.sh#L216 

@lekshmideepu : perhaps you can shed some light on this!

Silmathoron:
@clinssen where do you see the build?

jakobj:
@Silmathoron ah, I see now what you mean. i'm not intimately familiar with travis but would it be possible to have two different kinds of modes of operation
1) only the testsuite is run
2) first the style check is run, then the testsuite is run

after creating a PR the first mode would be activated by default, and one of the maintainers could switch to the second mode after the review is complete, as a final verification just before the PR is merged?

Silmathoron:
@jakobj I don't know but it looks pretty complicated... especially compared to its alternative: running a cppcheck before clicking on the PR button ;)

clinssen:
[...]

PR #1267, opened 2019-08-13
Fast static check in Travis CI

clinssen:
See discussion for issue #1246. Quoting:

> I have experimented a bit with speeding up the static check. Without doing a full NEST-simulator build, the static check takes about 3m30s, against 10m30s before.


lekshmideepu:
https://travis-ci.org/nest/nest-simulator/jobs/571439073#L1324

clinssen:
I think this error with `jq` was a glitch in Travis-CI. `jq` processes JSON from a GitHub API web request, which might have errored before, but seems to have gone through OK after I restarted the job.

lekshmideepu:
@clinssen Thanks for initiating this PR. Unfortunately I couldn't successfully run the build on Travis where cppcheck detected some formatting[ issues](https://travis-ci.org/lekshmideepu/nest-simulator/jobs/575691410#L1337).

extras/travis_build.sh
```diff
@@ -168,25 +168,15 @@ if [ "$xSTATIC_ANALYSIS" = "1" ]; then
...
-     do
-       for single_file_name in $file_names
-       do
-         echo "MSGBLD0095: File changed: $single_file_name"
-       done
-     done
-    echo "MSGBLD0100: Retrieving changed files completed."
-    echo
-
+    #file_names=`find . -name "*.h" -o -name "*.c" -o -name "*.cc" -o -name "*.hpp" -o -name "*.cpp" -o -name "*.py"`
```
lekshmideepu: Please remove the commented out lines.

clinssen: I added a comment explaining the purpose (check *all* files instead of just the ones that have changed); this might come in handy for testing but should not be the default on Travis CI due to excessive runtimes. Note that over the course of time, some files slipped through, i.e. they are in nest-master but fail the static check. Some of these I fixed in this PR (these are the seemlingly unrelated changes in the diff).

clinssen:
Request to those with the power to click the "Merge" button: please squash this mess of a history!

.travis.yml
```diff
@@ -114,51 +114,58 @@ cache:
...
+         if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew --version; brew update; brew tap brewsci/science; brew tap brewsci/bio; brew install coreutils gsl open-mpi automake autoconf libtool; fi
+         pip install -U pip
+         pip install -U setuptools
+         # Installing additional packages using pip as they only have
+         # outdated versions in the Travis package whitelist.
+         # terminaltables is required by parse_travis_log.py to create
+         # the build summary.
+         pip install cython scipy matplotlib terminaltables
+         pip install numpy==1.13.3
+      fi
```
gtrensch:
```suggestion
     fi
```

extras/travis_build.sh
```diff
@@ -117,7 +117,7 @@ echo $NEST_VPATH
 mkdir "$NEST_VPATH" "$NEST_RESULT"
 mkdir "$NEST_VPATH/reports"
 
-if [ "$xSTATIC_ANALYSIS" = "1" ]; then
+if [ "$xSTATIC_ANALYSIS" == "1" ]; then
```
gtrensch: The script uses 3 different variants of the bash conditionals syntax.
if [ "$xSTATIC_ANALYSIS" = "1" ]; then ..
if [ "$xSTATIC_ANALYSIS" == "1" ]; then ..
if [[ "$xSTATIC_ANALYSIS" == "1" ]]; then ..
The first one, I believe, is the standard.

extras/travis_build.sh
```diff
@@ -215,82 +219,83 @@ if [ "$xSTATIC_ANALYSIS" = "1" ]; then
     "$IGNORE_MSG_VERA" "$IGNORE_MSG_CPPCHECK" "$IGNORE_MSG_CLANG_FORMAT" "$IGNORE_MSG_PEP8"
 else
     echo "MSGBLD0225: Static code analysis skipped due to build configuration."
-fi  # End of Static code analysis.
```
gtrensch: If this is moved to the bottom of the script, execution of the build will depend on the setting of xSTATIC_ANALYSIS. Wouldn't it be better to leave control to the build matrix in the .travis.yaml file and add another variable to the matrix and encapsulate the build process?

extras/travis_build.sh
```diff
@@ -215,82 +219,83 @@ if [ "$xSTATIC_ANALYSIS" = "1" ]; then
...
-  echo "MSGBLD0330: (WARNING) Build artifacts not uploaded to Amazon S3."
-fi
+    if [ "$TRAVIS_REPO_SLUG" != "nest/nest-simulator" ] ; then
+        echo "MSGBLD0320: This build was from a forked repository and not from nest/nest-simulator."
+        echo "MSGBLD0330: (WARNING) Build artifacts not uploaded to Amazon S3."
+    fi
 
-echo "MSGBLD0340: Build completed."
+    echo "MSGBLD0340: Build completed."
+fi
```
gtrensch: Please see my comment above.

extras/travis_build.sh
clinssen: There are subtle differences (especially between `[]` and `[[]]`), and the only reason I didn't address this before is that I'm afraid to break things. E.g. `if [[ $OSTYPE == darwin* ]]` would break if it was changed for single brackets.

gtrensch: If I'm right, **==** is an alias for **=** and **[[ ... ]]** allows to omit the quotes while empty strings are handled properly. However, I'm fine with all variants. The point I wanted to make is that for the same thing we should use the same syntax, i.e. when checking the build-matrix variables. Of course, I would leave it as it is for the OSTYPE.

extras/travis_build.sh
clinssen: There seems to be a variable `xRUN_TESTSUITE` that already follows exactly this pattern (it is 1 if and only if `xSTATIC_ANALYSIS` is 0). Right now it is only wrapping the `make installcheck` command. Does it still make sense to have an extra variable (e.g. `xBUILD_NEST`), or shall we rename `xRUN_TESTSUITE` and make it encapsulate *both* the build and running the test suite?

gtrensch: `xRUN_TESTSUITE` skips `make installcheck` if set to 0. In contrast, `xSTATIC_ANALYSIS` skips code analysis and turns the build on if set to 0. This is not what I would expect when looking at the build matrix in the `.travis.yml` file.

What do you think of the following solution? 
- rename `xRUN_TESTSUITE` into `xRUN_BUILD_AND_TESTSUITE` and change the script ...

```
if [ "$xRUN_BUILD_AND_TESTSUITE" = "1" ]; then
  echo
  echo "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"  
  echo "+               B U I L D   N E S T                                           +"
  echo "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
   echo "MSGBLD0250: Running Make."
  make VERBOSE=1
  echo "MSGBLD0260: Make completed."
  ...
  make install
  ...
  make installcheck
fi
```
Test without build does not make any sense. We can combine it and be more explicit without hiding the logic.

lekshmideepu:
All good from my side 👍

PR #1276, opened 2019-08-27
Fix threading issue when connecting

stinebuu:
A little while ago @heplesser found that when trying to reproduce the figures from Ippen et al (2017), the strong scaling by threads variant behaved badly with master. Luckily, he found the problem as well, which is that a lot of threads are trying to write to `have_connections_changed_` at once. This PR use `CompletedChecker` for `have_connections_changed_` to make sure that all threads do not write to the same variable.

A benchmark can be found here: [Fig_5_results_one_node_noinst.pdf](https://github.com/nest/nest-simulator/files/3545900/Fig_5_results_one_node_noinst.pdf) The benchmark is from July, `0ac2385` was master at the time, and `8fa2e37` contains the fix. With the fix we see that strong scaling by threads and strong scaling by mpi again are equal when we *connect*.

This PR also cleans up some of our thread-safe connection routines. Instead of passing around one static empty dummy dictionary when connecting, we use a list of empty dictionaries, one dict per thread.


nestkernel/connection_manager.cpp
```diff
@@ -427,7 +426,10 @@ nest::ConnectionManager::connect( const index sgid,
 {
   kernel().model_manager.assert_valid_syn_id( syn_id );
 
-  have_connections_changed_ = true;
+  if ( not have_connections_changed_[ target_thread ] )
```
jakobj: do we need this check here? isn't it cheaper to just write in all cases? the logic wouldn't change, would it?

nestkernel/connection_manager.cpp
```diff
@@ -502,10 +504,13 @@ nest::ConnectionManager::connect( const index sgid,
 {
   kernel().model_manager.assert_valid_syn_id( syn_id );
 
-  have_connections_changed_ = true;
-
   const thread tid = kernel().vp_manager.get_thread_id();
 
+  if ( not have_connections_changed_[ tid ] )
```
jakobj: (see above)

nestkernel/connection_manager.cpp
```diff
@@ -605,18 +610,12 @@ nest::ConnectionManager::connect_( Node& s,
...
-  // We do not check has_primary_connections_ and secondary_connections_exist_
-  // directly as this led to worse performance on the supercomputer Piz Daint.
   if ( not check_primary_connections_[ tid ] and is_primary )
   {
-#pragma omp atomic write
-    has_primary_connections_ = true;
     check_primary_connections_.set( tid, true );
   }
-  else if ( not check_secondary_connections_[ tid ] and not is_primary )
+  if ( not check_secondary_connections_[ tid ] and not is_primary )
```
jakobj: was the `else` a logic error or why is it removed here?

nestkernel/connection_manager.cpp
```diff
@@ -694,7 +693,10 @@ nest::ConnectionManager::find_connection( const thread tid, const synindex syn_i
 void
 nest::ConnectionManager::disconnect( const thread tid, const synindex syn_id, const index sgid, const index tgid )
 {
-  have_connections_changed_ = true;
+  if ( not have_connections_changed_[ tid ] )
```
jakobj: (see above)

nestkernel/connection_manager.cpp
```diff
@@ -1562,11 +1565,19 @@ nest::ConnectionManager::resize_connections()
 void
 nest::ConnectionManager::sync_has_primary_connections()
 {
+  has_primary_connections_ = check_primary_connections_.any_true();
+#pragma omp barrier
```
jakobj: is this in a threaded part?

nestkernel/connection_manager.h
```diff
@@ -340,7 +340,7 @@ class ConnectionManager : public ManagerInterface
    * Sets flag indicating whether connection information needs to be
    * communicated.
    */
-  void set_have_connections_changed( const bool changed );
+  void set_have_connections_changed( thread tid, const bool changed );
```
jakobj: `thread tid` could be const

nestkernel/connection_manager.h
```diff
@@ -707,13 +707,13 @@ ConnectionManager::get_remote_targets_of_local_node( const thread tid, const ind
 inline bool
 ConnectionManager::have_connections_changed() const
 {
-  return have_connections_changed_;
+  return have_connections_changed_.any_true();
 }
 
 inline void
-ConnectionManager::set_have_connections_changed( const bool changed )
+ConnectionManager::set_have_connections_changed( thread tid, const bool changed )
```
jakobj: (see above)

nestkernel/connector_base.h
```diff
@@ -262,12 +262,10 @@ class Connector : public ConnectorBase
     C_[ lcid ].set_status( dict, static_cast< GenericConnectorModel< ConnectionT >& >( cm ) );
   }
 
-
-  Connector< ConnectionT >&
+  void
```
jakobj: these changes are logically distinct from the changes that this PR is implementing, aren't they? if so, they should be moved to a separate PR

nestkernel/connector_model_impl.h
```diff
@@ -276,11 +276,8 @@ GenericConnectorModel< ConnectionT >::add_connection_( Node& src,
 
   assert( connector != 0 );
 
-  // TODO: simplify: push_back should not return anything
   Connector< ConnectionT >* vc = static_cast< Connector< ConnectionT >* >( connector );
-  connector = &vc->push_back( connection );
-
-  thread_local_connectors[ syn_id ] = connector;
+  vc->push_back( connection );
```
jakobj: (see above)

nestkernel/sp_manager.cpp
```diff
@@ -210,6 +210,7 @@ SPManager::disconnect_single( index sgid, Node* target, thread target_thread, Di
 {
   if ( kernel().connection_manager.have_connections_changed() )
   {
+    kernel().connection_manager.check_secondary_connections_exist();
```
jakobj: why do we need this here? was this a bug before?

topology/connection_creator.cpp
```diff
@@ -79,17 +74,14 @@ ConnectionCreator::ConnectionCreator( DictionaryDatum dict )
     }
     else if ( dit->first == names::mask )
     {
-
```
jakobj: i'm not sure whether these formatting changes should be part of this PR

topology/connection_creator.cpp
```diff
@@ -25,11 +25,10 @@
 namespace nest
 {
 
-const DictionaryDatum ConnectionCreator::dummy_param_ = new Dictionary;
-
 ConnectionCreator::ConnectionCreator( DictionaryDatum dict )
   : allow_autapses_( true )
   , allow_multapses_( true )
+  , allow_oversized_( false )
```
jakobj: what is this used for?

nestkernel/connection_manager.cpp
stinebuu: It was, I have reverted back to the original, as that might be easier to read.

stinebuu:
@jakobj First of all, sorry for taking way too long to address your comments! I have made some changes based on your comments. Concerning your comments about why I don't set `have_connections_changed_` directly, I want to do some benchmarking again. I feel like there was a reason for this, but as I can't remember it I will test it out, and then either add a comment or fix it.

So no need to do another review yet.

topology/connection_creator.cpp
stinebuu: I think I just added it because I noticed it wasn't initialized. I can remove it again.

heplesser:
@jakobj @jarsi Could you (re-)review this PR in the near future? It is one of two PRs for 2.18.1 still open.

stinebuu:
@heplesser It is hanging on me I'm afraid.

stinebuu:
@jakobj @jarsi The PR is again ready for the review, sorry for the delay.

Somewhere along the way the benchmarks showed worse performance again, and it took some time to detect the problem, but that is fixed now: [benchmark_results.pdf](https://github.com/nest/nest-simulator/files/3973159/benchmark_results.pdf)

topology/connection_creator.cpp
stinebuu: @jakobj If `allow_oversized_` is not initialized here, the test `test_oversized_mask.sli` fails. I don't understand why the test suddenly fails here and not in master though.

nestkernel/connection_manager.cpp
stinebuu: @jakobj I tested on Blaustein, and if we do not check before setting the parameter, then the performance gets a lot worse. You can see it as `4c388f4` [here](https://github.com/nest/nest-simulator/files/3973159/benchmark_results.pdf).  It must have something to do with all the threads trying to access at once and maybe something with the `omp automatic write` in the `set` function. I wrote a comment before every `if`.

nestkernel/conn_builder.h
```diff
@@ -222,8 +222,8 @@ class ConnBuilder
   //! dictionaries to pass to connect function, one per thread
   std::vector< DictionaryDatum > param_dicts_;
 
-  //! empty dictionary to pass to connect function
-  const static DictionaryDatum dummy_param_;
+  //! empty dictionary to pass to connect function, one per thread
```
heplesser: @stinebuu Could you add a note to the comment to explain why we need one dummy dict per thread? Just so we remember ten years from now.

stinebuu: @heplesser done!

nestkernel/completed_checker.cpp
```diff
@@ -64,6 +64,34 @@ CompletedChecker::all_true() const
   return true;
 }
 
+bool
+CompletedChecker::any_false() const
```
jakobj: why not use `std::any_of` since we use c++11 now? we should probably also replace the other functions, maybe in a separate PR though

nestkernel/completed_checker.cpp
```diff
@@ -64,6 +64,34 @@ CompletedChecker::all_true() const
...
+    if ( not a_[ i ] )
+    {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool
+CompletedChecker::any_true() const
```
jakobj: see above

nestkernel/connection_manager.cpp
jakobj: ok,thanks for clarifying

topology/connection_creator.cpp
jakobj: uninitialized is always bad, thanks for fixing this.

heplesser:
@jakobj Thanks for you comments! We cannot use `std::any_of` for two reasons: The underlying data type of `a_` is a C-style array to allow OpenMP atomic updates and we need and OpenMP barrier before the "any" check.

As all reviewers have approved now, I will merge.

Issue #1271, opened 2019-08-21
Attribution Error : module 'nest' has no attribute 'sli_push'


nosratullah:
Hi,
I used this [link](https://www.nest-simulator.org/py_sample/brunel_alpha_nest/) to recreate a balanced netwrok and follow the documentation,
`def LambertWm1(x):
nest.sli_push(x)
nest.sli_run('LambertWm1')
y = nest.sli_pop()
return y`

and get this error :
`module 'nest' has no attribute 'sli_push' `

heplesser:
@nosratullah I can confirm this. A work-around is as follows: Replace the function definition with
```python
def LambertWm1(x): 
     return nest.ll_api.sli_func('LambertWm1', float(x)) 
```

This issue raises several questions:
- In how many other places in examples have we forgotten to update `nest.sli_...` to `nest.ll_api.sli_...`?
- Why haven't we detected this problem with CI? We need automated testing of examples ...
- Should we provide the LambertW function as part of the high-level PyNEST API?

hakonsbm:
It looks like the example file is correct, so for this case it seems the problem is that the html files haven't been generated after the update. 

https://github.com/nest/nest-simulator/blob/4a0b53df2a2fff853d8e05bf70c9ea8819ed7585/pynest/examples/brunel_alpha_nest.py#L74-L78

However, a quick search found that there is still documentation and examples that use ``nest.sli_func()``, ``nest.sli_run()``, and ``nest.spp()``. We should definitely have automated testing of examples, and maybe also of code snippets in documentation.

And I am all for providing PyNEST API for all useful functions. I don't think users should have to use functions from `ll_api` at all. Note however that the Lambert W function is also available from SciPy.

heplesser:
@hakonsbm Thanks for checking! Could you check out the LambertW function in SciPy and see if we can replace the NEST-version with the SciPy version? After all, we also do not provide trig functions through PyNEST. My main concern would be that LambertW may be very new in SciPy and not be available in even slightly older versions.

hakonsbm:
@heplesser The Lambert W function in SciPy gives the same values (or with a difference very close to machine precision) as the NEST-version. It was also [added in the 0.8.0 version of SciPy](https://docs.scipy.org/doc/scipy/reference/release.0.8.0.html#lambert-w-function), which was released in 2010.

PR #1277, opened 2019-08-28
Add updated NEST examples to ReadtheDocs 

jessica-mitchell:
Previously, the examples  were linked to the old website and some scripts needed to be updated to work with newest versions of NEST. This is related to issue #1271.

Here the updated examples were generated via sphinx-gallery (which produces an rst, py and ipynb for each example) and are moved to the doc/examples folder.

**This is an interim solution to have the updated examples available for users** 

Future PRs are planned to automate/test the examples, include the output images in the gallery of examples, and improve the organization of the examples - as it is now, they are all listed together, with related examples next to each other.

Tagging @hakonsbm @heplesser since you were involved in discussions regarding issue #1271 

See output here: https://nest-test.readthedocs.io/en/pyexamples_update/examples/index.html


doc/examples/index.rst
```diff
@@ -1,5 +1,136 @@
...
+these applications `(Parallel computing with NEST) <https://www.nest-simulator.org/parallel_computing/>`__.
+The number of threads (per MPI process) can be chosen by adjusting
+`local_num_threads` in ``sim_params.py``. The number of MPI processes can be
+set by choosing a reasonable value for `num_mpi_prc` and then running the
+script with the following command.
+
+.. code-block:: python
+
+   mpirun -n num_mpi_prc
+   python example.py
```
heplesser: This needs to be on one line, otherwise `mpirun` will fail complaining about not having been given an executable.

doc/examples/index.rst
```diff
@@ -1,5 +1,136 @@
...
+
+The default version of the simulation uses Poissonian input, which is defined
+in the file ``network_params.py`` to excite neuronal populations of the
+microcircuit. If no Poissonian input is provided, DC input is calculated, which
+should approximately compensate the Poissonian input. It is also possible to
+add thalamic stimulation to the microcircuit or drive it with constant DC
+input. This can be defined in the file ``stimulus_params.py``.
+
+Tested configuration: This version has been tested with NEST 2.10.0,
+Python 2.7.12 and 3.5.2, NumPy 1.11.2
```
heplesser: We keep telling people to use current versions of NEST. Could you test if this works with 2.18? I assume all our examples are auto-tested before a release, so I hope this just needs an updated version number (or two) here.

jessica-mitchell: @heplesser The microciricuit example.py ran for me with 2.18. This text was leftover I guess from when this example was originally made. I'd rather avoid writing 'tested versions', and prefer just remove this completely. (Unless we find an automated way of showing this).

As for the formatting errors, with the gap_junctions example- what exactly do you mean? If I need to regenerate the output files, it could take a bit of time

heplesser:
@jessica-mitchell The formatting errors are "blue boxes" that shouldn't be there, e.g. the "1" and "Authors" in the [gap junction example](https://nest-test.readthedocs.io/en/pyexamples_update/examples/gap_junctions_inhibitory_network.html) and multiple lines of normal text offset in boxes in the [multimeter to file example](https://nest-test.readthedocs.io/en/pyexamples_update/examples/multimeter_file.html), see also screenshots below.

Screenshot 2019-09-04 at 22 58 33

Screenshot 2019-09-04 at 22 58 51

jessica-mitchell:
@heplesser The 1 for the reference is correct, it's how references are formatted in restructured text in this theme.  
The see also and authors sections were sections we wanted to include in the examples, but these are not completed yet. I can remove the empty ones  for the time being if that's preferred.

The other sections are mistakes in formatting the python scripts, which I can go through and fix.

heplesser:
@jessica-mitchell Thank you for the explanations! I think it makes sense to remove empty sections to reduce user confusion. Once the sections are filled, shouldn't "See also" and "Authors" be styled the same way, i.e., bold-face large?

Is the formatting of the reference numbers a choice or technical necessity? The current style looks so different from normal reference formatting that I find it rather confusing. I think a plain "[1]" in front of the citation would work better.

jessica-mitchell:
@heplesser I have now modified the files to fix up some formatting issues. I also changed the CSS a bit to make the reference number appear a bit nicer, but they're not in brackets. The authors, if provided, have a subsection similar to see also.

heplesser:
@jessica-mitchell Looks much better now, but in some random checks I noticed a few more problems:

- The reference is incomplete in https://nest-test.readthedocs.io/en/pyexamples_update/examples/brette_gerstner_fig_3d.html
- The 2nd reference looks funny ("2(1,2)") in https://nest-test.readthedocs.io/en/pyexamples_update/examples/precise_spiking.html

Could you check all examples once more to make sure the references are proper?

jessica-mitchell:
@heplesser 
I've updated the references but the reason for this syntax 1(1,2) actually is a back reference to every instance the reference is mentioned in the file.

heplesser:
I wonder how many readers will understand the 1(1,2) syntax before accidentally clicking on the links ;).

PR #1280, opened 2019-09-03
Fix poisson generator ps

### LINKED ISSUE #1278: poisson_generator_ps causes segmentation fault after setting rate from 0 to positive value
This issue is based on https://github.com/NeuralEnsemble/PyNN/issues/650, see also #1279.

The following will cause a segmentation fault:
```python
p = nest.Create('poisson_generator_ps', params={'rate': 100.})
n = nest.Create('parrot_neuron_ps')    
nest.Connect(p, n) 
nest.Simulate(100)
nest.SetStatus(p, {'rate': 0.})  
nest.Simulate(100)
nest.SetStatus(p, {'rate': 10.}) 
nest.Simulate(100)
```

While the segfault occurs in `parrot_neuron_ps::handle(SpikeEvent&)`, it is most likely caused by an incorrect handling of rate change in [`poisson_generator_ps::event_hook(DSSpikeEvent&)`](https://github.com/nest/nest-simulator/blob/8adec3cd2bb8025ca07c91f8e281ecb71f13658f/precise/poisson_generator_ps.cpp#L212). The method has special code for handling the first spike correctly, but not for handling rate changes.

### LINKED ISSUE #1279: poisson_generator_ps does not handle rate change correctly
This issue is based on https://github.com/NeuralEnsemble/PyNN/issues/650, see also #1278.

When the rate of `poisson_generator_ps` is changed after simulating, then the new rate will be applied properly only after the next spike has been fired. When moving from a very low to a high rate, this may take a long time. The solution to this is most likely that `P_::set()` needs to flag that there has been a parameter change, and that [`poisson_generator_ps::event_hook(DSSpikeEvent&)`](https://github.com/nest/nest-simulator/blob/8adec3cd2bb8025ca07c91f8e281ecb71f13658f/precise/poisson_generator_ps.cpp#L212) needs to handle the interval from the moment of rate change to the first subsequent spike using an exponential distribution, as is currently the case for the very first spike fired by the generator at the beginning of the simulation.

hakonsbm:
Adds a `P::rate_changed_` flag to `poisson_generator_ps` which lets the `event_hook()` handle the interval from the rate change to the first subsequent spike. This makes sure that the generator model correctly handles rate changes. A regeressiontest is also added.

Fixes #1278 
Fixes #1279

### DIFF
--- a/precise/poisson_generator_ps.h
+++ b/precise/poisson_generator_ps.h
@@ -125,6 +125,7 @@ private:
   {
     double rate_;      //!< process rate [Hz]
     double dead_time_; //!< dead time [ms]
+    bool rate_changed_; //!< flag for rate change
 
     /**
      * Number of targets.
--- a/precise/poisson_generator_ps.cpp
+++ b/precise/poisson_generator_ps.cpp
@@ -45,6 +45,7 @@
 nest::poisson_generator_ps::Parameters_::Parameters_()
   : rate_( 0.0 )      // Hz
   , dead_time_( 0.0 ) // ms
+  , rate_changed_( false )
   , num_targets_( 0 )
 {
 }
@@ -70,7 +71,9 @@ nest::poisson_generator_ps::Parameters_::set( const DictionaryDatum& d )
     throw BadProperty( "The dead time cannot be negative." );
   }
 
-  updateValue< double >( d, names::rate, rate_ );
+  // Change of rate has to be flagged to let the event_hook handle the
+  // interval from the rate change to the first subsequent spike.
+  rate_changed_ = updateValue< double >( d, names::rate, rate_ );
 
   if ( rate_ < 0.0 )
   {
@@ -223,7 +226,7 @@ nest::poisson_generator_ps::event_hook( DSSpikeEvent& e )
   // introduce nextspk as a shorthand
   Buffers_::SpikeTime& nextspk = B_.next_spike_[ prt ];
 
-  if ( nextspk.first.is_neg_inf() )
+  if ( nextspk.first.is_neg_inf() or P_.rate_changed_ )
   {
     // need to initialize relative to t_min_active_
     // first spike is drawn from backward recurrence time to initialize the
@@ -255,6 +258,7 @@ nest::poisson_generator_ps::event_hook( DSSpikeEvent& e )
     nextspk.first = Time::ms_stamp( spike_offset );
     nextspk.second = nextspk.first.get_ms() - spike_offset;
     nextspk.first += V_.t_min_active_;
+    P_.rate_changed_ = false;
   }
 
   // as long as there are spikes in active period, emit and redraw
--- a/testsuite/regressiontests/issue-1278-1279.sli
+++ b/testsuite/regressiontests/issue-1278-1279.sli
@@ -1,0 +1,72 @@
+/*
+ *  issue-1278-1279.sli
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  NEST is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+ /** @BeginDocumentation
+Name: testsuite::issue-1278-1279
+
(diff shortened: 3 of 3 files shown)


precise/poisson_generator_ps.h
```diff
@@ -123,8 +123,9 @@ class poisson_generator_ps : public DeviceNode
    */
   struct Parameters_
   {
-    double rate_;      //!< process rate [Hz]
-    double dead_time_; //!< dead time [ms]
+    double rate_;       //!< process rate [Hz]
+    double dead_time_;  //!< dead time [ms]
+    bool rate_changed_; //!< flag for rate change
```
heplesser: I am not entirely happy with `rate_changed_` here in `P_`, since it is not a model parameter but an implementation detail. `B_` is probably the best place to put it, since `V_` is for variables recomputed by `calibrate()`.  Would it be difficult to move the flag to `B_`?

hakonsbm:
@heplesser There is now a Python test replacing the SLI regressiontest, which performs the same change of rate, but uses a KS-test on the spike times.

jstapmanns:
@hakonsbm I may have overlooked something but can't you just set `B_.next_spike_` to `Time::neg_inf()` whenever the rate is changed with `set_satus`?

pynest/nest/tests/test_poisson_generator_rate_change.py
```diff
@@ -0,0 +1,85 @@
...
+import numpy as np
+
+
+class TestPgRateChange(unittest.TestCase):
+
+    def test_statistical_rate_change(self):
+        """Statistical test of poisson_generator_ps rate change"""
+
+        # Lower limit of acceptable p_value
+        p_value_lim = 0.3
```
heplesser: p = 0.3 seems an unusual limit.

pynest/nest/tests/test_poisson_generator_rate_change.py
```diff
@@ -0,0 +1,85 @@
...
+
+        def kstest_first_spiketimes(sd, start_t):
+            events = nest.GetStatus(sd)[0]['events']
+            senders = events['senders']
+            times = events['times']
+            min_times = [np.min(times[np.where(senders == s)])
+                         for s in np.unique(senders)]
+            d, p_val = scipy.stats.kstest(
+                min_times, 'expon', args=(start_t + 0.1, 10.))
```
heplesser: I am not really happy with the "magic numbers" 0.1 and 10. here. It would be more robust and transparent to pass the expected rate  as argument. For even more robustness, I'd suggest to set the resolution explicitly and have it in a variable; maybe use 0.25 or so to avoid the numerically unpleasant 0.1 ;).

hakonsbm:
@heplesser I have updated the test, please have another look.

pynest/nest/tests/test_poisson_generator_rate_change.py
```diff
@@ -31,44 +31,56 @@ class TestPgRateChange(unittest.TestCase):
...
-        nest.SetStatus(pg, {'rate': 0.})
+        rate = 0.
+        nest.SetStatus(pg, {'rate': rate})
         # We need to skip a timestep to not receive the spikes from the
         # previous simulation run that were sent, but not received.
-        nest.SetStatus(sd, {'n_events': 0, 'start': 100.1, 'stop': 200.})
-        nest.Simulate(100)
+        nest.SetStatus(sd, {'n_events': 0,
+                            'start': float(sim_time) + resolution,
+                            'stop': 2.*sim_time})
```
heplesser: Spaces around `*`.

pynest/nest/tests/test_poisson_generator_rate_change.py
```diff
@@ -31,44 +31,56 @@ class TestPgRateChange(unittest.TestCase):
...
         # Third simulation, with rate increased back up to 100
-        nest.SetStatus(pg, {'rate': 100.})
-        nest.SetStatus(sd, {'n_events': 0, 'start': 200., 'stop': 300.})
-        nest.Simulate(100)
-        kstest_first_spiketimes(sd, 200.)
+        rate = 100.
+        nest.SetStatus(pg, {'rate': rate})
+        nest.SetStatus(sd, {'n_events': 0,
+                            'start': 2. * sim_time,
+                            'stop': 3.*sim_time})
```
heplesser: Spaces around `*` as on line above.

pynest/nest/tests/test_poisson_generator_rate_change.py
```diff
@@ -31,44 +31,56 @@ class TestPgRateChange(unittest.TestCase):
...
 
-        # Lower limit of acceptable p_value
-        p_value_lim = 0.3
+        p_value_lim = 0.05  # Lower limit of acceptable p_value
+        resolution = 0.25  # Simulation resolution
+        n_parrots = 1000  # Number of parrot neurons
+        sim_time = 100  # Time to simulate
 
-        def kstest_first_spiketimes(sd, start_t):
+        def kstest_first_spiketimes(sd, start_t, rate):
```
heplesser: Code in `ks_first_spiketimes()` depends in `resolution` and `n_parrots` defined in the surrounding scope. That is neither transparent nor safe. I would suggest to pass them as arguments along with `rate`. I'd also suggest to make the function a private method in `TestPgRateChange` instead of defining it as an internal function here.

hakonsbm:
@heplesser The test is updated with your proposed changes now.

PR #1281, opened 2019-09-09
Fix example in nest-3 cheatsheet

jhnnsnk:
The keyword `synapse_model` is now correctly used in the `syn_spec`.


heplesser:
@stinebuu Since this is just a documentation fix, it can be merged with a single positive review.

PR #1283, opened 2019-09-11
NESTio

jougs:


doc/conf.py
```diff
@@ -254,3 +262,31 @@ def setup(app):
...
+    "models/spike_detector",
+    "models/weight_recorder",
+    "nestkernel/recording_backend_ascii",
+    "nestkernel/recording_backend_memory",
+    "nestkernel/recording_backend_screen",
+    "nestkernel/recording_backend_sionlib",
+)
+
+import re
+import os
```
gtrensch: These imports violate PEP8 rule "E402 module level import not at top of file".
`Import os `is furthermore duplicate and can be removed.

gtrensch:
```suggestion
```

PR #1287, opened 2019-09-12
Use shape as argument instead of columns/rows/depth and update topology examples to NEST3

jhnnsnk:
This PR replaces the arguments `columns`/`rows`/`depth` used previously for creating grid layers with a single array named `shape`. `shape` can be of length 2 or 3 for 2D or 3D layers, respectively. The grid mask and its anchor are also adapted.

Topology examples are now updated to NEST3 and moved to pynest/examples/spatial.


pynest/examples/spatial/grid_iaf_irr.py
```diff
@@ -31,30 +31,27 @@
...
 import pylab
 import random
-import nest.topology as topo
 
 nest.ResetKernel()
 
-# generate list of 12 (x,y) pairs
-pos = [[random.uniform(-0.75, 0.75), random.uniform(-0.5, 0.5)]
-       for j in range(12)]
+pos = nest.spatial.free([[random.uniform(-0.75, 0.75), random.uniform(-0.5, 0.5)]
```
stinebuu: maybe use nest.random.uniform here?

Silmathoron: numpy.random.uniform?

stinebuu: @Silmathoron with the new NEST-3 we have made random distributions from NEST accessible from PyNEST, and seeing as this is an example I think we should use the NEST version.

hakonsbm: The problem here is that we want different distributions for different dimensions, which currently can't be done when passing `nest.random` functions to `nest.spatial.free()`. But it shouldn't be difficult to add this, maybe after this PR is merged, then we can update this example.

jhnnsnk: Yesterday, I asked @hakonsbm about this and he recommended to use numpy.random.uniform in this case because with nest.random.uniform it is apparently not possible, yet, to define different min and max for the two dimensions. If there is a better solution, please let me know and I will adjust it.

hakonsbm: Yes, passing a `nest.random` function to `nest.spatial.free()` will currently use the same distribution in all dimensions. But it shouldn't be difficult to add the possibility to have different distributions for different dimensions. Maybe after this PR is merged, then we can update this example?

stinebuu: @jhnnsnk @hakonsbm Ah, I missed that. I think we should be able to have different distributions for different dimensions, but I agree that we can do it after this PR is merged. I will add a note on the project page, and merge this PR into the NEST-3 branch.

PR #1297, opened 2019-09-25
Updated/extended macOS install instructions

heplesser:
I updated the macOS install instructions to a current gcc version number and added troubleshooting information. I also removed the MacPorts instructions, because they have not been tested for a long time and we have, I believe, currently no-one around who uses MacPorts and could test them.

### DIFF
--- a/doc/installation/mac_install.rst
+++ b/doc/installation/mac_install.rst
@@ -1,73 +1,29 @@
-Installation on macOS
-=========================
+Manual Installation on macOS
+===============================
 
-
-On the Mac, you can install NEST either via Homebrew or manually. If you want to use PyNEST, you need to have a version of Python with some science packages installed, see the `section Python on Mac <python-on-mac>`_ for details.
-
-Installation via Homebrew
---------------------------
-
-The easiest way to install NEST on a Mac is to install it via the Homebrew package manager:
-
-*  To install homebrew, follow the instructions at `brew.sh <http://brew.sh/>`_
-
-*  Then, in a terminal
-
-    * Add the homebrew/science tap by running:
-
-       .. code-block:: sh
-
-         brew tap brewsci/science
-
-    * For information on what options NEST has and what will be installed, run:
-
-       .. code-block:: sh
-
-         brew info nest
-
-    * To install nest, execute
-
-       .. code-block:: sh
-
-         brew install nest
-
-Options have to be appended, so for example, to install NEST with PyNEST run:
-
-       .. code-block:: sh
-
-         brew install nest --with-python
-
-This will install the most recent release version of NEST. To build
-NEST from the most recent sources on Github, use:
-
-       .. code-block:: sh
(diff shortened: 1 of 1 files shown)


doc/installation/mac_install.rst
```diff
@@ -49,7 +49,7 @@ Manual installation
 
 The clang/clang++ compiler that ships with OS X/macOS does not support OpenMP threads and creates code that fails some tests. You therefore need to use **GCC** to compile NEST under OS X/macOS.
 
-Installation instructions here have been tested under OS X 10.11 *El Capitan* and macOS 10.12 *Sierra* with `Anaconda Python 2 and 3 <https://www.continuum.io/anaconda-overview>`_ and all other dependencies installed via `Homebrew <http://brew.sh>`_. See below for `Manual installation with dependencies from MacPorts`_.
+Installation instructions here have been tested under macOS 10.14 *Mojave* with `Anaconda Python 3 <https://www.continuum.io/anaconda-overview>`_ and all other dependencies installed via `Homebrew <http://brew.sh>`_. They should also work with earlier versions of macOS.
```
jessica-mitchell: @heplesser Can you be more explicit here in the steps you took to create this environment? 
What about the PyNEST dependencies like numpy, scipy, and matplotlib?
which order did you use (brew install then conda install) ? Do you create a conda environment before installing the dependencies?

heplesser: I have tried different approaches at different times, both with the full Anaconda package and miniconda plus the `conda install` line shown in the Python section. I did not explicitly create conda environments in all cases I tested, it worked with and without for me. Therefore I have't been more specific here.

heplesser:
@jessica-mitchell Should we replace the Homebrew section with one on how to install from Conda? Given that the Homebrew version is outdated and no longer maintained, I believe (@tammoippen ?), replacing it with instructions to install from Conda would make sense. But since #1262 is on its way, maybe for now we could leave it as it is?

I would also suggest leaving the Python section at the end for now to keep it out of the way. There is a forward link at the start of the document, so users should be able to find it.

jessica-mitchell:
@heplesser 
I don't like the location of this link to the Python on Mac section. It assumes the user will read the entire page and not just skip to a section of interest. I'd like to see a link inside manual installation section regarding the python on mac section. 

I would also suggest using a numbered list rather than bullets.

In #1262, the 'pre-packaged' versions (conda, ppa, docker, virtual machine) are separated from the 'compiling from source' directions. So I wouldnt want to add the Conda instructions on this page.  I think we can discuss in that PR if we want to change how it's organized. But in the case of homebrew, if it will not be updated and maintained with our releases, I would be for removing it.

Other than that, I think it's good to go

heplesser:
@jessica-mitchell I have removed the Homebrew instructions now, changed the top-level heading to "manual installation" and changed to an enumerated list.

PR #1299, opened 2019-09-27
Check variables of GSL binomial random generator where they are set

### LINKED ISSUE #1298: Asking for too many connections with FixedTotalNumber leads to an unexpected number of connections
When connecting with `FixedTotalNumber` and setting the number of connections above a certain limit, NEST will not create the correct number of connections when NEST is compiled with GSL. This is because the GSL function random binomial function `gsl_ran_binomial()`, which is used when connecting with a fixed total number, takes an unsigned int as the argument for number of successes, `n`. So if we ask for a number of connections that is greater than what can fit in the unsigned int, we will not get the expected number of connections.

Because this is a detail in GSL, we cannot fix it in NEST and will have to generate an error if the number of connections the user asks for does not fit in an unsigned int.

hakonsbm:
In `GSL_BinomialRandomDev`, checks of new `n` and `p` values are now done in the setter methods, to catch erroneous or error-inducing values used when setting `n` and `p` directly without using `GSL_BinomialRandomDev::set_status()`.

Fixes #1298.

### DIFF
--- a/librandom/gsl_binomial_randomdev.cpp
+++ b/librandom/gsl_binomial_randomdev.cpp
@@ -75,7 +75,7 @@ librandom::GSL_BinomialRandomDev::ldev( RngPtr rng ) const
 }
 
 void
-librandom::GSL_BinomialRandomDev::set_p_n( double p_s, unsigned n_s )
+librandom::GSL_BinomialRandomDev::set_p_n( double p_s, size_t n_s )
 {
   set_p( p_s );
   set_n( n_s );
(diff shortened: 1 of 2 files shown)


librandom/gsl_binomial_randomdev.cpp
```diff
@@ -84,13 +84,27 @@ librandom::GSL_BinomialRandomDev::set_p_n( double p_s, unsigned n_s )
...
 void
-librandom::GSL_BinomialRandomDev::set_n( unsigned int n_s )
+librandom::GSL_BinomialRandomDev::set_n( size_t n_s )
 {
+  // gsl_ran_binomial() takes n as an unsigned int, so it cannot be
+  // greater than what an unsigned int can hold.
+  const auto N_MAX = std::numeric_limits< unsigned int >::max();
+  if ( n_s >= N_MAX )
+  {
+    throw BadParameterValue( String::compose( "Gsl_binomial RDV: N < %1 required.", static_cast< double >( N_MAX ) ) );
```
heplesser: I'd suggest lowercase `gsl_binomial` in error message.

hakonsbm: @heplesser Done.

hakonsbm:
Looks like the formatting check fails on Cppcheck warnings, even though it is supposed to ignore them.

hakonsbm:
I have created #1303 for the Cppcheck issue.

PR #1300, opened 2019-09-30
Skip deployment in Travis CI testsuite when not running a build

clinssen:
This did not show up as an error in the previous PR updating the Travis CI suite (#1267) because the deployment step is always skipped on forks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -169,12 +169,15 @@ script:
      fi
 
 before_deploy:
-   - cd $TRAVIS_BUILD_DIR/build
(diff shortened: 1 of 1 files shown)


lekshmideepu:
Right. I was looking into that.

lekshmideepu:
As this is a very critical PR to fix the issues with the merge build in master, I voluntarily added me as a reviewer :-)

lekshmideepu:
I am merging this without second opinion as it is an obvious change and it fix the issues in the master.

PR #1302, opened 2019-10-01
deploy the artifacts only when the build flag is set in the reply section

lekshmideepu:


lekshmideepu:
Proof of concept build is [here](https://travis-ci.org/nest/nest-simulator/jobs/592081185#L1030)

PR #1306, opened 2019-10-02
Fix cppcheck in Travis CI testsuite

### LINKED ISSUE #1303: Static code check failing because of Cppcheck messages when they should be ignored
The static code check on Travis fails because of Cppcheck messages, even with 
https://github.com/nest/nest-simulator/blob/2f37ee82e8a2b365ca8a8a0f5eaf99df7add8499/extras/travis_build.sh#L207

From [check_code_style.sh](https://github.com/nest/nest-simulator/blob/master/extras/check_code_style.sh) it looks like no messages from any tools can be ignored.

clinssen:
Update C++ standard to C++11, because the old was giving syntax errors on the enum class.

Do not throw "inconclusive" warnings, as there seem to be many non-consequential ones (has this option been discussed before?)

Fixes #1303.

### DIFF
--- a/extras/static_code_analysis.sh
+++ b/extras/static_code_analysis.sh
@@ -171,7 +171,7 @@ for f in $FILE_NAMES; do
       # CPPCHECK
       if $PERFORM_CPPCHECK; then
         print_msg "MSGBLD0150: " "Running CPPCHECK ...: $f"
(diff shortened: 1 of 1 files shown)


hakonsbm:
Regarding the "inconclusive" warnings, my guess is that the idea was to get as much information on the state of the code as possible. But it generates a lot of non-consequential noise instead, so I agree that it's better to leave it out.

hakonsbm:
Merging as it is a minimal change.

PR #1308, opened 2019-10-04
Use an unsigned data type for thread in spike data

jakobj:
thread is typically a signed data type since -1 is used in some cases as a special value for thread. However to be able to use the full range of thread ids that are available with NUM_BITS_TID bits in spike data, we need to use an unsigned data type here.

I suggest @suku248 @clinssen as reviewers.

### DIFF
--- a/nestkernel/spike_data.h
+++ b/nestkernel/spike_data.h
@@ -55,7 +55,7 @@ protected:
   index lcid_ : NUM_BITS_LCID;                       //!< local connection index
   unsigned int marker_ : NUM_BITS_MARKER_SPIKE_DATA; //!< status flag
   unsigned int lag_ : NUM_BITS_LAG;                  //!< lag in this min-delay interval
-  thread tid_ : NUM_BITS_TID;                        //!< thread index
+  unsigned int tid_ : NUM_BITS_TID;                  //!< thread index
   synindex syn_id_ : NUM_BITS_SYN_ID;                //!< synapse-type index
 
 public:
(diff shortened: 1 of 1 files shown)


ikitayama:
Jacob, do we have (prospective) hardware that requires widening the number
of bits for expressing thread id like a 1024 HW threads CPU?

jougs:
@clinssen: The way I understand it is that `-1` can still be used everywhere else in the code. This PR merely allows the field `tid_` to take on all possible numbers representable using the given number of bits instead of splitting the range into an unused negative half and a useable positive one with only half the possible size. @jakobj: if my understanding is correct, I'll approve the change as soon as you fix the indentation of the comment following the code.

jakobj:
@ikitayama no, i don't have any specific hardware in mind
@clinssen what @jougs said :) the special value of -1 is (should) never be communicated via the buffers for which this data type is relevant. if desired i could add an assert just to be sure.
@jougs you are correct, fixing the identation now

jougs:
@jakobj: thanks for your answers. Are there any checks in place that make sure the values actually fit in the bit-restricted fields?

jakobj:
yes, in the `set` function: https://github.com/nest/nest-simulator/blob/ee7b54cd8a4ebba71552de10d66b1106de89a0b7/nestkernel/spike_data.h#L162

jougs:
Then it's probably wise to add an `assert` against `-1` there as well.

jakobj:
that's a good point. fixed.

PR #1310, opened 2019-10-14
Switch to GCC8 as default in continuous integration tests on Travis

### LINKED ISSUE #1309: Update gcc version in Travis CI
While finalising my pull request #1069, I ran into a compiler bug in gcc. When specialising a class template in a different file than the generic class definition, an issue with namespace resolution can arise; see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56480 for more detail.

This requires gcc to be updated to version >= 7 (current being 5.4), because that is where the bug was first fixed. We presently already have a Travis CI stage dedicated to gcc 8.

Are there any preferences (or objections) to specific versions? gcc 9 is out but still seems to require adding a new PPA. The easiest solution would be to use gcc 8 everywhere. This also cuts the Travis CI runtime down a bit (as gcc 8 will no longer need to be a distinct step).

clinssen:
Fixes #1309.

### DIFF
--- a/doc/installation/install_options.rst
+++ b/doc/installation/install_options.rst
@@ -20,9 +20,10 @@ For more detailed information please see below: ``Python3 Binding (PyNEST)``
 Choice of compiler
 ------------------
 
-The default compiler for NEST is GNU gcc/g++, but NEST has also successfully
-been compiled with other compilers, including Intel icc/icpc, Pathscale,
-Portland and IBM compilers.
+The default compiler for NEST is GNU gcc/g++. Version 7 or higher is required
+due to the presence of bugs in earlier versions that prevent the compilation
+from succeeding. NEST has also successfully been compiled with other 
+compilers, including Intel icc/icpc, Pathscale, Portland and IBM compilers.
 
 To select a specific compiler, please add the following flags to your ``cmake``
 line::
--- a/doc/installation/install_options.rst
+++ b/doc/installation/install_options.rst
@@ -22,8 +22,8 @@ Choice of compiler
 
 The default compiler for NEST is GNU gcc/g++. Version 7 or higher is required
 due to the presence of bugs in earlier versions that prevent the compilation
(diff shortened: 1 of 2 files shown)


ikitayama:
May I ask why Clang 7 is still the choice for CI/CD on Travis as Clang 9 is
out in the fields?

clinssen:
@ikitayama: the reason we upgraded GCC was because there was a specific reason to do so: we were running into a compiler bug on the previous version. In general, I believe our policy is a conservative one, and we only upgrade if there is a compelling reason for it. If you can make the case for clang 9, then of course please be invited to do so!

ikitayama:
Then I’d think we can creat an edge CI/CD with the latest stable Clang,
keeping the stable one with Ckang 7, unless the resource is serveky limited.

doc/installation/install_options.rst
```diff
@@ -20,9 +20,10 @@ For more detailed information please see below: ``Python3 Binding (PyNEST)``
 Choice of compiler
 ------------------
 
-The default compiler for NEST is GNU gcc/g++, but NEST has also successfully
-been compiled with other compilers, including Intel icc/icpc, Pathscale,
-Portland and IBM compilers.
+The default compiler for NEST is GNU gcc/g++. Version 7 or higher is required
+due to the presence of bugs in earlier versions that prevent the compilation
+from succeeding. NEST has also successfully been compiled with other 
+compilers, including Intel icc/icpc, Pathscale, Portland and IBM compilers.
```
heplesser: We should update the list of "other compilers" mentioned here. First, we definitely should mention Clang, since we now explicitly test with it. Some of us probably use the Intel compiler occasionally, but I think it is a long time since anyone has tested Pathscale, Portland and probably also IBM, so I would take those out.

ikitayama:
I’m using XL on JURON at JSC.

Issue #1318, opened 2019-10-17
Multitude of warnings when compiling NEST with -Dwith-openmp=OFF


Helveg:
There seem to be a lot of warnings when we try to make NEST 2.18 with the following flags:

```
cmake  -Dwith-mpi=ON -Dwith-python=3 -Dwith-openmp=OFF
```

# Warnings

```
[ 35%] Building CXX object nestkernel/CMakeFiles/nestkernel.dir/archiving_node.cpp.o

In file included from /home/nest-simulator/nestkernel/connection_manager.h:34:0,
                 from /home/nest-simulator/nestkernel/kernel_manager.h:27,
                 from /home/nest-simulator/nestkernel/archiving_node.cpp:33:
/home/nest-simulator/nestkernel/completed_checker.h:108:0: warning: ignoring #pragma omp atomic [-Wunknown-pragmas]
 #pragma omp atomic update
 ^
/home/nest-simulator/nestkernel/completed_checker.h:115:0: warning: ignoring #pragma omp atomic [-Wunknown-pragmas]
 #pragma omp atomic write
 ^
/home/nest-simulator/nestkernel/source_table.cpp:53:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
 #pragma omp parallel
 ^
/home/nest-simulator/nestkernel/source_table.cpp:213:0: warning: ignoring #pragma omp single [-Wunknown-pragmas]
 #pragma omp single
 ^
/home/nest-simulator/nestkernel/source_table.cpp:230:0: warning: ignoring #pragma omp critical [-Wunknown-pragmas]
 #pragma omp critical
 ^
/home/nest-simulator/nestkernel/source_table.cpp:237:0: warning: ignoring #pragma omp barrier [-Wunknown-pragmas]
 #pragma omp barrier
 ^
/home/nest-simulator/nestkernel/source_table.cpp:239:0: warning: ignoring #pragma omp single [-Wunknown-pragmas]
 #pragma omp single
 ^
/home/nest-simulator/nestkernel/target_table_devices.cpp:81:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
 #pragma omp parallel
 ^
/home/nest-simulator/nestkernel/target_table_devices.cpp:93:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
 #pragma omp parallel
 ^
[...]

Issue #1196, opened 2019-05-11
pynestkernel.pyx:689:10: Unreachable code


yurivict:
I'm getting this with python-3.6 on FreeBSD:
```
warning: /usr/ports/science/nest/work/nest-simulator-2.14.0-2100-g788e93e3/pynest/pynestkernel.pyx:689:10: Unreachable code
warning: /usr/ports/science/nest/work/nest-simulator-2.14.0-2100-g788e93e3/pynest/pynestkernel.pyx:689:10: Unreachable code
```

jougs:
We all get to see these warnings.

The reason for them is that Cython gets confused by the [`raise` statement in the line above](https://github.com/nest/nest-simulator/blob/master/pynest/pynestkernel.pyx#L687) although that is clearly in an (even unlikely) `else` clause and thus usually not executed. In the generated code in `pynest/pynestkernel.cxx`, you can see that the "unreachable" statement is instantiated just fine (line 11805 in my case). So no need to worry.

I'm closing this issue as "Wont't fix" and "External bug". Probably someone should [file a bug against Cython](https://github.com/cython/cython/issues).

PR #1322, opened 2019-10-22
Fix [-Wsign-compare] warnings when Cythonizing pynestkernel.pyx

alberto-antonietti:
## Problem

Compiling NEST, when it cythonizes pynestkernel.pyx, the compiler (tested with both gcc 9.1.0 and gcc 8.3.1) generates the following warnings:

```
[100%] Compiling Cython CXX source for pynestkernel...
warning: /home/aantonietti/workspace/nest-simulator/pynest/pynestkernel.pyx:689:10: Unreachable code
warning: /home/aantonietti/workspace/nest-simulator/pynest/pynestkernel.pyx:689:10: Unreachable code
```
Which is an external Cython bug (see #1196) [Won't fix]

And

```
Scanning dependencies of target pynestkernel
[100%] Building CXX object pynest/CMakeFiles/pynestkernel.dir/pynestkernel.cxx.o
/home/aantonietti/workspace/nest-simulator/b/pynest/pynestkernel.cxx: In function ‘PyObject* __pyx_f_12pynestkernel_sli_array_to_object(ArrayDatum*)’:
/home/aantonietti/workspace/nest-simulator/b/pynest/pynestkernel.cxx:11232:35: warning: comparison of integer expressions of different signedness: ‘Py_ssize_t’ {aka ‘long int’} and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
     for (__pyx_temp=0; __pyx_temp < __pyx_v_dat->size(); __pyx_temp++) {
                        ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/aantonietti/workspace/nest-simulator/b/pynest/pynestkernel.cxx:11260:33: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘Py_ssize_t’ {aka ‘long int’} [-Wsign-compare]
   for (__pyx_t_4 = 0; __pyx_t_4 < __pyx_t_2; __pyx_t_4+=1) {
                       ~~~~~~~~~~^~~~~~~~~~~
```
As reported also in #1318 .
## Solution

This PR fixes the two [-Wsign-compare] warning with an explicit cast to `int` of size( ) and of the argument of range( ).

### DIFF
--- a/pynest/pynestkernel.pyx
+++ b/pynest/pynestkernel.pyx
@@ -623,10 +623,12 @@ cdef inline object sli_datum_to_object(Datum* dat):
 
 cdef inline object sli_array_to_object(ArrayDatum* dat):
 
-    # the size of dat has to be explicitly cast to int to avoid compiler warnings (#1318) during cythonization
+    # the size of dat has to be explicitly cast to int to avoid
+    # compiler warnings (#1318) during cythonization
     cdef tmp = [None] * int(dat.size())
 
-    # i and n have to be cast to size_t (unsigned long int) to avoid compiler warnings (#1318) in the for loop below
+    # i and n have to be cast to size_t (unsigned long int) to avoid
+    # compiler warnings (#1318) in the for loop below
     cdef size_t i, n
     cdef Token* tok = dat.begin()
 
--- a/pynest/pynestkernel.pyx
+++ b/pynest/pynestkernel.pyx
@@ -623,12 +623,14 @@ cdef inline object sli_datum_to_object(Datum* dat):
 
 cdef inline object sli_array_to_object(ArrayDatum* dat):
 
-    cdef tmp = [None] * dat.size()
+    cdef tmp = [None] * int(dat.size())
 
-    cdef size_t i
+    cdef size_t i, n
     cdef Token* tok = dat.begin()
 
-    for i in range(len(tmp)):
+    n = len(tmp)
+
+    for i in range(n):
         tmp[i] = sli_datum_to_object(tok.datum())
         inc(tok)
 
--- a/pynest/pynestkernel.pyx
+++ b/pynest/pynestkernel.pyx
@@ -623,8 +623,10 @@ cdef inline object sli_datum_to_object(Datum* dat):
 
 cdef inline object sli_array_to_object(ArrayDatum* dat):
 
+    # the size of dat has to be explicitly cast to int to avoid compiler warnings (#1318) during cythonization
     cdef tmp = [None] * int(dat.size())
 
+    # i and n have to be cast to size_t (unsigned long int) to avoid compiler warnings (#1318) in the for loop below
     cdef size_t i, n
     cdef Token* tok = dat.begin()
 


pynest/pynestkernel.pyx
```diff
@@ -623,12 +623,14 @@ cdef inline object sli_datum_to_object(Datum* dat):
 
 cdef inline object sli_array_to_object(ArrayDatum* dat):
 
-    cdef tmp = [None] * dat.size()
+    cdef tmp = [None] * int(dat.size())
```
jougs: Is `int` big enough or should it rather be cast to `size_t`?

alberto-antonietti: Casting it to `int` was the only solution to the problem. This is probably due to a Cython bug related to the `len( )` function that will be solved in [Cython 3.0.0 ](https://cython.readthedocs.io/en/latest/src/changes.html)

A working alternative can be:
```
    cdef Py_ssize_t i
    cdef Py_ssize_t n = dat.size()

    cdef tmp = [None] * n
    cdef Token* tok = dat.begin()


    for i in range(n):
        tmp[i] = sli_datum_to_object(tok.datum())
        inc(tok)
```
Do you prefer this option?

jougs: I think that's fine. At least I definitely like it better than the idea of the original solution behaving randomly or (IMO actually safer) crashing when converting `ArrayDatum`s of a size larger than what an `int` can hold...

However, it is probably a good idea to also add a short comment explaining why these steps are needed to stop the next someone from "simplifying" it in a year from now ;-)

clinssen: Hi, as far as I'm aware, Python's "int" is not a primitive type, but some clever contraption that has no upper bound. However, I'm not sure whether this line of code will be executed by the Python interpreter, or first converted into C and run there. It might be instructive to look at the generated code.

alberto-antonietti: These are the relative generated codes [before ](https://pastebin.com/wjNKzPx9) and [after ](https://pastebin.com/pBG0FePE) the PR.

alberto-antonietti: > I think that's fine. At least I definitely like it better than the idea of the original solution behaving randomly or (IMO actually safer) crashing when converting `ArrayDatum`s of a size larger than what an `int` can hold...
> 
> However, it is probably a good idea to also add a short comment explaining why these steps are needed to stop the next someone from "simplifying" it in a year from now ;-)

I have added a couple of lines of comment ;-)

clinssen: OK, it looks like neither version is doing any downcasting to machine word lengths! Readability of the generated code could be improved, but I'll leave this verdict up to @jougs. :)

pynest/pynestkernel.pyx
```diff
@@ -623,12 +623,16 @@ cdef inline object sli_datum_to_object(Datum* dat):
 
 cdef inline object sli_array_to_object(ArrayDatum* dat):
 
-    cdef tmp = [None] * dat.size()
+    # the size of dat has to be explicitly cast to int to avoid compiler warnings (#1318) during cythonization
```
jougs: Can you please break the line so it's below 80 characters?

pynest/pynestkernel.pyx
```diff
@@ -623,12 +623,16 @@ cdef inline object sli_datum_to_object(Datum* dat):
...
+    cdef tmp = [None] * int(dat.size())
 
-    cdef size_t i
+    # i and n have to be cast to size_t (unsigned long int) to avoid compiler warnings (#1318) in the for loop below
```
jougs: See above.

alberto-antonietti:
> @alberto-antonietti: I think I'm fine with the changes as they are now and will merge once the comment lines are broken as requested. @clinssen: Improving other people's generated code is probably for another life ;-)

Just to be sure, I saw that with #1177 the max line length increased from 80 to 120 characters.
This file has a .pyx extension, but I assume it should be considered as a .py file?

I notice that there are a few lines (183, 201, 447, 615) that exceeed also the 120 characters limit. Due to the .pyx instead of .py extension, this is not caught by the PEP8 check?

I have then 2 questions:
- Should I stick to the 80 char or 120 char limit?
- Should I break the lines exceeding the 80/120 char limit in this PR?

jougs:
OK, sorry! This opens a whole new can of worms, which was not my intention at all. According to a short Google search, there is no agreed upon way to handle Cython files when it comes to linting (probably flake8 plus some custom config works best). As we currently don't have that kind of  infrastructure in place and thus can't enforce anything, you can leave all other lines as they are.

Please just break the comments in question once so they look less odd above the quite short code lines they refer to. Thanks!

alberto-antonietti:
Agreed! I have broken the long comment lines, as requested. Thanks for having reviewed the PR!

PR #1324, opened 2019-10-29
Update copyright file in debian directory

alexmyczko:
i'd suggest dropping debian/ altogehter, making an official debian package, file an RFP/ITP bug, close it via debian/changelog
and that'll also you help easily have ubuntu copy your work (from debian) (officially distributing binaries of your software)


debian/copyright
```diff
@@ -1,9 +1,19 @@
 Format: http://dep.debian.net/deps/dep5
 Upstream-Name: nest-simulator
-Source: https://github.com/nest/nest-simulator/archive/master.zip
+Source: https://github.com/nest/nest-simulator/
+
+Files: *
+Copyright: 2004-2019 The NEST Initiative <nest_user@nest-initiative.org>
```
jougs: This is the email address of the old mailing list. And even if it were the new one, it would not be very useful, as sending to it requires to be subscribed to the list. I thus suggest to remove it.

debian/copyright
```diff
@@ -1,9 +1,19 @@
...
+License: GPL-2
+
+Files: librandom/*
+Copyright: other people
+License: other license
 
 Files: debian/*
 Copyright: 2019 The NEST Initiative <nest_user@nest-initiative.org>
```
jougs: See above.

debian/copyright
heplesser: As a conscious policy decision, we always use 2004, the year of the first public release, as year of the NEST copyright, even for changes added later. So it should only be 2004 here.

debian/copyright
```diff
@@ -1,9 +1,19 @@
 Format: http://dep.debian.net/deps/dep5
 Upstream-Name: nest-simulator
-Source: https://github.com/nest/nest-simulator/archive/master.zip
+Source: https://github.com/nest/nest-simulator/
+
+Files: *
+Copyright: 2004-2019 The NEST Initiative <nest_user@nest-initiative.org>
+License: GPL-2
```
heplesser: NEST license is GPL v2 or later, by choice of the user.

debian/copyright
heplesser: 2004, not 2019, see comment above.

debian/copyright
```diff
@@ -1,9 +1,19 @@
...
+
+Files: librandom/*
+Copyright: other people
+License: other license
 
 Files: debian/*
 Copyright: 2019 The NEST Initiative <nest_user@nest-initiative.org>
+License: GPL-2
```
heplesser: GPL v2 or later, see above.

debian/copyright
```diff
@@ -1,9 +1,19 @@
...
+
 License: GPL-2
```
heplesser: GPL v2 or later.

debian/copyright
alexmyczko: you can use the initial copyright, but debian uses whenever the source was started being written, to when the last change has been made to it, so 2004-2019 for debian/copyright is perfectly fine. and you shouldn't have that file at all anyways in a source tarball release, see top suggestion

alexmyczko: taken from ./CMakeLists.txt:set( NEST_USER_EMAIL "nest_user@nest-initiative.org" )
So you should remove it from your sources, or update it?

it's also listed in:
./doc/sli_docs/an-introduction-to-sli.rst:     Mailing list: nest_user@nest-initiative.org
./doc/guides/using_nest_with_music.rst:      Mailing list: nest_user@nest-initiative.org
./doc/guides/using_nest_with_music.rst:      Mailing list: nest_user@nest-initiative.org

debian/copyright
```diff
@@ -1,10 +1,20 @@
 Format: http://dep.debian.net/deps/dep5
 Upstream-Name: nest-simulator
-Source: https://github.com/nest/nest-simulator/archive/master.zip
+Source: https://github.com/nest/nest-simulator/
+
+Files: *
+Copyright: 2004-2019 The NEST Initiative <nest_user@nest-initiative.org>
+License: GPL-2+
```
steffengraber: Why do you write` GPL-2+` and not `GPL-2 `? We use the GPL version 2.0, so I think `GPL-2` is correct.

debian/copyright
```diff
@@ -1,10 +1,20 @@
...
+
+Files: librandom/*
+Copyright: other people
+License: other license
```
steffengraber: I think the block can be deleted.

debian/copyright
```diff
@@ -1,10 +1,20 @@
...
 
 Files: debian/*
 Copyright: 2019 The NEST Initiative <nest_user@nest-initiative.org>
-License: GPL-2
+License: GPL-2+
+
+License: GPL-2+
  NEST is free software: you can redistribute it and/or modify
```
steffengraber: See my question about `GPL-2+` above.

debian/copyright
steffengraber: After having read the license again I agree with GPL-2+. Sorry.

debian/copyright
steffengraber: Thanks for bringing this up. The changes have been included in #1346 .

jougs: Sending to the mailing list address will yield a mail containing subscription information as well as a contact path to the mailing list owners, so this is not so bad after all.

jougs:
@alexmyczko: can you please have a look at the comments by @steffengraber and merge upstream to resolve the conflicts? Thanks!

jougs:
@alexmyczko: can you please resolve the conflicts in your branch, so we can merge this PR? Thanks!

debian/copyright
heplesser: @alexmyczko Looking at https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#copyright-field, there seems no requirement to include a mail address on the copyright line, so it should be removed here. I also do not find elsewhere (https://www.debian.org/doc/debian-policy/ch-docs.html#s-copyrightfile, https://www.debian.org/doc/manuals/maint-guide/dreq.en.html#copyright) any explicit policy on copyright dates for Debian. Therefore, we should use 2004, and only 2004 here in keeping with general NEST Copyright policy.

debian/copyright
heplesser: See above, this must also be 2004.

debian/copyright
```diff
@@ -1,10 +1,16 @@
 Format: http://dep.debian.net/deps/dep5
 Upstream-Name: nest-simulator
-Source: https://github.com/nest/nest-simulator/archive/master.zip
+Source: https://github.com/nest/nest-simulator/
+
+Files: *
+Copyright: 2004-2019 The NEST Initiative <nest_user@nest-initiative.org>
```
terhorstd, line 6:
```suggestion
Copyright: 2004 The NEST Initiative
```

debian/copyright
```diff
@@ -1,10 +1,16 @@
...
+License: GPL-2+
 
 Files: debian/*
 Copyright: 2019 The NEST Initiative <nest_user@nest-initiative.org>
```
terhorstd, line 10:
```suggestion
Copyright: 2004 The NEST Initiative
```

jougs:
Many thanks for this contribution and sorry for the messy review. In a group session during our ongoing hackathon, we addressed all open issues with this PR and came to the conclusion that we rather merge it as is instead of having this open for even longer.

PR #1336, opened 2019-11-12
Autogenerate pynest examples from source with sphinx_gallery

jessica-mitchell:
This PR allows sphinx_gallery to autogenerate examples from pynest/examples/ 

The mock module in the conf.py file was removed due to a conflict between it and sphinx_gallery; this change should not affect any part of the documentation generation.

The examples that were previously in doc/ are now removed, so we no longer have the duplicate files
And the paths were updated in the examples/index.rst file
The html output should not change.

@heplesser  Should I add the topology/examples?  I wasn't sure if they need to be updated  for 2.18 - The last update mentioned in the README is 2014 NEST 2.6


heplesser:
@jessica-mitchell Sorry for commenting late on this one. I suggest to treat the topology examples in the same way as the other pynest examples.

PR #1338, opened 2019-11-15
Communication via compressed spikes

jakobj:
In the current ("5g") communication scheme if a neuron has multiple targets on different threads of a target process, spikes for all different threads are generated on the presynaptic side and send to the postsynaptic side via MPI. For example for 48threads, a neuron with targets on all threads will generate 48spikes presynaptically. This leads to unnecessarily large MPI buffers for massively multihreaded simulations (buffer sizes scale ~ #threads per process).

By introducing an additional postsynaptic data structure and the concept of a "compressed spike" this PR modifies the communication infrastructure such that this redundancy is avoided. A neuron with targets on all threads of a process hence only needs to send a single spike which is then multiplexed on the postsynaptic side, significantly reducing the load on the communication network. Initial benchmarks show significant improvements over 5g on specific benchmark instances with many threads (measured and reported by @jarsi @jhnnsnk ).
The current implementation relies on heuristics to select the connections which qualify for using compressed spikes, these should be critically reviewed.

@jhnnsnk @terhorstd

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -69,6 +69,7 @@ nest::ConnectionManager::ConnectionManager()
   , have_connections_changed_()
   , has_get_connections_been_called_( false )
   , sort_connections_by_source_( true )
+  , use_compressed_spikes_( true )
   , has_primary_connections_( false )
   , check_primary_connections_()
   , secondary_connections_exist_( false )
@@ -94,6 +95,7 @@ nest::ConnectionManager::initialize()
   connections_.resize( num_threads );
   secondary_recv_buffer_pos_.resize( num_threads );
   sort_connections_by_source_ = true;
+  compressed_spike_data_.resize( 0 );
 
   have_connections_changed_.initialize( num_threads, false );
   check_primary_connections_.initialize( num_threads, false );
@@ -134,6 +136,7 @@ nest::ConnectionManager::finalize()
   delete_connections_();
   std::vector< std::vector< ConnectorBase* > >().swap( connections_ );
   std::vector< std::vector< std::vector< size_t > > >().swap( secondary_recv_buffer_pos_ );
+  compressed_spike_data_.clear();
 }
 
 void
@@ -159,6 +162,13 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
       "If structural plasticity is enabled, sort_connections_by_source can not "
       "be set to false." );
   }
+
+  updateValue< bool >( d, names::use_compressed_spikes, use_compressed_spikes_ );
+  if ( use_compressed_spikes_ and not sort_connections_by_source_ )
+  {
+    throw KernelException( "Spike compression requires sort_connections_by_source to be true." );
+  }
+
   //  Need to update the saved values if we have changed the delay bounds.
   if ( d->known( names::min_delay ) or d->known( names::max_delay ) )
   {
@@ -183,6 +193,7 @@ nest::ConnectionManager::get_status( DictionaryDatum& dict )
   def< long >( dict, names::num_connections, n );
   def< bool >( dict, names::keep_source_table, keep_source_table_ );
   def< bool >( dict, names::sort_connections_by_source, sort_connections_by_source_ );
+  def< bool >( dict, names::use_compressed_spikes, use_compressed_spikes_ );
 
   def< double >( dict, names::time_construction_connect, sw_construction_connect.elapsed() );
 }
@@ -1460,5 +1471,27 @@ nest::ConnectionManager::unset_have_connections_changed( const thread tid )
   if ( have_connections_changed_[ tid ].is_true() )
   {
     have_connections_changed_[ tid ].set_false();
+  }
+}
+
+
+void
+nest::ConnectionManager::collect_compressed_spike_data( const thread tid )
+{
+  if ( use_compressed_spikes_ )
+  {
+    assert( sort_connections_by_source_ );
+
+#pragma omp single
+    {
+      source_table_.resize_compressible_sources();
+    } // of omp single; implicit barrier
+
+    source_table_.collect_compressible_sources( tid );
+#pragma omp barrier
+#pragma omp single
+    {
+      source_table_.fill_compressed_spike_data( compressed_spike_data_ );
+    } // of omp single; implicit barrier
   }
 }
--- a/nestkernel/connection_manager.h
+++ b/nestkernel/connection_manager.h
@@ -40,6 +40,7 @@
 #include "nest_types.h"
 #include "per_thread_bool_indicator.h"
 #include "source_table.h"
+#include "spike_data.h"
 #include "target_table.h"
 #include "target_table_devices.h"
 
@@ -87,6 +88,8 @@ public:
 
   void compute_target_data_buffer_size();
   void compute_compressed_secondary_recv_buffer_positions( const thread tid );
+  void collect_compressed_spike_data( const thread tid );
+  void clear_compressed_spike_data_map( const thread tid );
 
   /**
    * Add a connectivity rule, i.e. the respective ConnBuilderFactory.
@@ -315,6 +318,8 @@ public:
    */
   bool get_sort_connections_by_source() const;
 
+  bool use_compressed_spikes() const;
+
   /**
    * Sorts connections in the presynaptic infrastructure by increasing
    * source node ID.
@@ -396,6 +401,8 @@ public:
   // public stop watch for benchmarking purposes
   // start and stop in high-level connect functions in nestmodule.cpp and nest.cpp
   Stopwatch sw_construction_connect;
+
+  const std::vector< SpikeData >& get_compressed_spike_data( const synindex syn_id, const index idx );
 
 private:
   size_t get_num_target_data( const thread tid ) const;
@@ -541,6 +548,13 @@ private:
   SourceTable source_table_;
 
   /**
+   * A structure to hold "unpacked" spikes on the postsynaptic side if
+   * spike compression is enabled. Internally arranged in a 3d
+   * structure: synapses|sources|spike data
+   */
+  std::vector< std::vector< std::vector< SpikeData > > > compressed_spike_data_;
+
+  /**
    * Stores absolute position in receive buffer of secondary events.
    * structure: threads|synapses|position
    */
@@ -597,6 +611,14 @@ private:
 
   //! Whether to sort connections by source node ID.
   bool sort_connections_by_source_;
+
+  //! Whether to use spike compression; if a neuron has targets on
+  //! multiple threads of a process, this switch makes sure that only
+  //! a single packet is sent to the process instead of one packet per
+  //! target thread; requires sort_connections_by_source_ = true; for
+  //! more details see the discussion and sketch in
+  //! https://github.com/nest/nest-simulator/pull/1338
+  bool use_compressed_spikes_;
 
   //! Whether primary connections (spikes) exist.
   bool has_primary_connections_;
@@ -781,6 +803,12 @@ ConnectionManager::get_sort_connections_by_source() const
   return sort_connections_by_source_;
 }
 
+inline bool
+ConnectionManager::use_compressed_spikes() const
+{
+  return use_compressed_spikes_;
+}
+
 inline double
 ConnectionManager::get_stdp_eps() const
 {
@@ -824,6 +852,18 @@ inline void
 nest::ConnectionManager::set_has_get_connections_been_called( const bool has_get_connections_been_called )
 {
   has_get_connections_been_called_ = has_get_connections_been_called;
+}
+
+inline const std::vector< SpikeData >&
+ConnectionManager::get_compressed_spike_data( const synindex syn_id, const index idx )
+{
+  return compressed_spike_data_.at( syn_id ).at( idx );
+}
+
+inline void
+ConnectionManager::clear_compressed_spike_data_map( const thread tid )
+{
+  source_table_.clear_compressed_spike_data_map( tid );
 }
 
 } // namespace nest
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -620,17 +620,40 @@ EventDeliveryManager::deliver_events_( const thread tid, const std::vector< Spik
     {
       const SpikeDataT& spike_data = recv_buffer[ rank * send_recv_count_spike_data_per_rank + i ];
 
-      if ( spike_data.get_tid() == tid )
+      se.set_stamp( prepared_timestamps[ spike_data.get_lag() ] );
+      se.set_offset( spike_data.get_offset() );
+
+      if ( not kernel().connection_manager.use_compressed_spikes() )
       {
-        se.set_stamp( prepared_timestamps[ spike_data.get_lag() ] );
-        se.set_offset( spike_data.get_offset() );
+        if ( spike_data.get_tid() == tid )
+        {
+          const index syn_id = spike_data.get_syn_id();
+          const index lcid = spike_data.get_lcid();
+          const index source_node_id = kernel().connection_manager.get_source_node_id( tid, syn_id, lcid );
+          se.set_sender_node_id( source_node_id );
 
+          kernel().connection_manager.send( tid, syn_id, lcid, cm, se );
+        }
+      }
+      else
+      {
         const index syn_id = spike_data.get_syn_id();
-        const index lcid = spike_data.get_lcid();
-        const index source_node_id = kernel().connection_manager.get_source_node_id( tid, syn_id, lcid );
-        se.set_sender_node_id( source_node_id );
+        // for compressed spikes lcid holds the index in the
+        // compressed_spike_data structure
+        const index idx = spike_data.get_lcid();
+        const std::vector< SpikeData >& compressed_spike_data =
+          kernel().connection_manager.get_compressed_spike_data( syn_id, idx );
+        for ( auto it = compressed_spike_data.cbegin(); it != compressed_spike_data.cend(); ++it )
+        {
+          if ( it->get_tid() == tid )
+          {
+            const index lcid = it->get_lcid();
+            const index source_node_id = kernel().connection_manager.get_source_node_id( tid, syn_id, lcid );
+            se.set_sender_node_id( source_node_id );
 
-        kernel().connection_manager.send( tid, syn_id, lcid, cm, se );
+            kernel().connection_manager.send( tid, syn_id, lcid, cm, se );
+          }
+        }
       }
 
       // break if this was the last valid entry from this rank
--- a/nestkernel/source_table.cpp
+++ b/nestkernel/source_table.cpp
@@ -49,12 +49,16 @@ nest::SourceTable::initialize()
   saved_entry_point_.initialize( num_threads, false );
   current_positions_.resize( num_threads );
   saved_positions_.resize( num_threads );
+  compressible_sources_.resize( num_threads );
+  compressed_spike_data_map_.resize( num_threads );
 
 #pragma omp parallel
   {
     const thread tid = kernel().vp_manager.get_thread_id();
     sources_[ tid ].resize( 0 );
     resize_sources( tid );
+    compressible_sources_[ tid ].resize( 0 );
+    compressed_spike_data_map_[ tid ].resize( 0 );
   } // of omp parallel
 }
 
@@ -66,12 +70,16 @@ nest::SourceTable::finalize()
     if ( is_cleared_[ tid ].is_false() )
     {
       clear( tid );
+      compressible_sources_[ tid ].clear();
+      compressed_spike_data_map_[ tid ].clear();
     }
   }
 
   sources_.clear();
   current_positions_.clear();
   saved_positions_.clear();
+  compressible_sources_.clear();
+  compressed_spike_data_map_.clear();
 }
 
 bool
@@ -305,7 +313,7 @@ nest::SourceTable::previous_entry_has_same_source_( const SourceTablePosition& c
     and local_sources[ previous_lcid ].get_node_id() == current_source.get_node_id() );
 }
 
-void
+bool
 nest::SourceTable::populate_target_data_fields_( const SourceTablePosition& current_position,
   const Source& current_source,
   const thread source_rank,
@@ -323,10 +331,36 @@ nest::SourceTable::populate_target_data_fields_( const SourceTablePosition& curr
     next_target_data.set_is_primary( true );
 
     TargetDataFields& target_fields = next_target_data.target_data;
-    // we store the thread index of the source table, not our own tid!
-    target_fields.set_tid( current_position.tid );
     target_fields.set_syn_id( current_position.syn_id );
-    target_fields.set_lcid( current_position.lcid );
+    if ( kernel().connection_manager.use_compressed_spikes() )
+    {
+      // WARNING: we set the tid field here to zero just to make sure
+      // it has a defined value; however, this value is _not_ used
+      // anywhere when using compressed spikes
+      target_fields.set_tid( 0 );
+      auto it_idx = compressed_spike_data_map_.at( current_position.tid )
+                      .at( current_position.syn_id )
+                      .find( current_source.get_node_id() );
+      if ( it_idx != compressed_spike_data_map_.at( current_position.tid ).at( current_position.syn_id ).end() )
+      {
+        // WARNING: no matter how tempting, do not try to remove this
+        // entry from the compressed_spike_data_map_; if the MPI buffer
+        // is already full, this entry will need to be communicated the
+        // next MPI comm round, which, naturally, is not possible if it
+        // has been removed
+        target_fields.set_lcid( it_idx->second );
+      }
+      else // another thread is responsible for communicating this compressed source
+      {
+        return false;
+      }
+    }
+    else
+    {
+      // we store the thread index of the source table, not our own tid!
+      target_fields.set_tid( current_position.tid );
+      target_fields.set_lcid( current_position.lcid );
+    }
   }
   else // secondary connection, e.g., gap junctions
   {
@@ -343,6 +377,8 @@ nest::SourceTable::populate_target_data_fields_( const SourceTablePosition& curr
     secondary_fields.set_recv_buffer_pos( relative_recv_buffer_pos );
     secondary_fields.set_syn_id( current_position.syn_id );
   }
+
+  return true;
 }
 
 bool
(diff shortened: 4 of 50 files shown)


jakobj:
Sketch of the modified postsynaptic infrastructure. Note that spikes can be delivered regularly (green arrow) or via the new data structure (blue arrows) depending on the number of target threads.
infrastructure_5g_compressed_spikedata

nestkernel/event_delivery_manager.cpp
```diff
@@ -571,6 +571,27 @@ EventDeliveryManager::deliver_events_( const thread tid, const std::vector< Spik
 
         kernel().connection_manager.send( tid, syn_id, lcid, cm, se );
       }
+      else if ( spike_data.get_tid() == MAX_TID )  // handle compressed spikes
```
7eia: Is it sensible to use `MAX_TID` to identify compressed spikes? How about a negative number? Are IDs signed?

nestkernel/event_delivery_manager.cpp
```diff
@@ -571,6 +571,27 @@ EventDeliveryManager::deliver_events_( const thread tid, const std::vector< Spik
...
+      {
+        se.set_stamp( prepared_timestamps[ spike_data.get_lag() ] );
+        se.set_offset( spike_data.get_offset() );
+
+        const index syn_id = spike_data.get_syn_id();
+        const index idx = spike_data.get_lcid();  // for compressed spikes lcid holds the index in the compressed_spike_data structure
+        const std::vector< SpikeData >& compressed_spike_data =
+          kernel().connection_manager.get_compressed_spike_data( syn_id, idx );
+        for ( auto it = compressed_spike_data.cbegin(); it != compressed_spike_data.cend(); ++it )
```
7eia: Is type inference used by NEST? I can't see any other instance of `auto` elsewhere. Explicitly state type for consistency and clarity.

nestkernel/source_table.cpp
```diff
@@ -357,9 +362,26 @@ nest::SourceTable::get_next_target_data( const thread tid,
         next_target_data.set_is_primary( true );
         // we store the thread index of the source table, not our own tid!
         TargetDataFields& target_fields = next_target_data.target_data;
-        target_fields.set_tid( current_position.tid );
         target_fields.set_syn_id( current_position.syn_id );
-        target_fields.set_lcid( current_position.lcid );
+        if ( current_source.is_compressed() )
+        {
+          target_fields.set_tid( MAX_TID );  // use MAX_TID as marker for compressed spikes
```
7eia: See https://github.com/nest/nest-simulator/pull/1338/files#r352456115.

nestkernel/source_table.cpp
```diff
@@ -357,9 +362,26 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+          auto it_idx = compressed_spike_data_map_[ current_position.tid ].find( current_source.get_gid() );
```
7eia: Use of `auto` to be avoided?

nestkernel/event_delivery_manager.cpp
heplesser: @7eia NEST is older than the C++-98 standard, our first commit dates back to 1996. We are phasing in C++11 constructs in new or refactored code, but do not have the capacity to refactor everything to C++11.

nestkernel/source_table.cpp
heplesser: @7eia No, use of C++11 constructs is encouraged in new or refactored code.

nestkernel/event_delivery_manager.cpp
7eia: @heplesser Noted.

nestkernel/source_table.cpp
7eia: @heplesser Thank you for clarifying.

7eia:
@jarsi and I have been working to verify that MPI buffer sizes are sensibly sized when compression is active. By maintaining the network size and increasing the number of virtual processes, we find that the size of MPI buffers holding compressed spikes approaches (and eventually equals) the MPI buffer size in the case of no compression. This check demonstrates, as expected, that the original behaviour, i.e. in the case of no compression, is recovered when there is no opportunity for compression.

7eia:
Can someone try running the HPC benchmark with compressed spikes. My runs consistently fail. Configuration: scale set to 10, with 5-10 nodes on Jureca, 24 MPI ranks and 2 OpenMP threads. The same occurs with 1 rank and 48 threads. This configuration is fine without compressed spikes.

jakobj:
@7eia thanks for your comments and testing the branch. could you please specify how your runs fail? are you using the `hpc_benchmark.sli` from the examples folder?

7eia:
@jakobj yes, I'm using `hpc_benchmark.sli`. The runs fail silently after `NodeManager::prepare_nodes [Info]`. Standard error shows segmentation faults and bus errors, so nothing too helpful. This could indicate an OOM issue, but that should still be sorted.

jhnnsnk:
@jakobj @H27CK I just tested `hpc_benchmark.sli`, the microcircuit and some other examples. I compared spike results for NEST2.20 and the branch of this PR. If spike compression is not active [`local_num_threads=1`  and >= 1 MPI process(es)], spike results always coincide. With >1 threads, microcircuit results are also good.
However, if both >1 threads and synapse models created with `CopyModel` are used (as in the case of `hpc_benchmark.sli`), this branch fails with a Segmentation fault.
The reason is that a dimension for different synapse types is missing, most likely in the `compressed_spike_data_map_` (@jakobj already mentioned that this might become an issue). I could try to propose a fix next week.

jhnnsnk:
A fix is now in: https://github.com/jakobj/nest-simulator/pull/133

jakobj:
awesome, thanks you @jhnnsnk :tada: :tada: i've merged your fix and rebased onto the current master, now everything should be up to date.

nestkernel/event_delivery_manager.cpp
jakobj: no, it's not. this should be fixed before merging into master, thanks @H27CK

nestkernel/source_table.cpp
```diff
@@ -400,7 +424,11 @@ nest::SourceTable::get_next_target_data( const thread tid,
     // set the source rank
     source_rank = kernel().mpi_manager.get_process_id_of_node_id( current_source.get_node_id() );
 
-    populate_target_data_fields_( current_position, current_source, next_target_data );
+    if ( not populate_target_data_fields_( current_position, current_source, next_target_data ) )
```
jakobj: this construction is not great, and should maybe be avoided when preparing for merge into master

heplesser:
I removed the WIP-tag from the title and instead turned the PR into Draft state.

terhorstd:
Hi @jakobj , I heard you recently discussed the implementation of the compression. Do you want to continue work on this draft PR, or are you planning to file a new one?

jakobj:
i've finally updated this PR with the solution we have discussed @jhnnsnk @suku248 :rocket: 

locally, this gives me identical spike counts as the current master. more thorough checking of this implementation (spike counts, timing, memory consumption) on some HPC system for the relevant use cases (microcircuit & hpc benchmark?) is necessary in my opinion before this gets merged.

@jhnnsnk could you please put this into your benchmark pipeline (after travis is happy) and update us with the results? :)

nestkernel/connection_manager.cpp
```diff
@@ -157,6 +158,13 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
       "If structural plasticity is enabled, sort_connections_by_source can not "
       "be set to false." );
   }
+
+  updateValue< bool >( d, names::use_compressed_spikes, use_compressed_spikes_ );
+  if ( use_compressed_spikes_ and not sort_connections_by_source_ )
```
heplesser, line 163: Will this check work correctly also if `sort_connections_by_source_` is changed in the same call to `SetKernelStatus`?

heplesser:
@jakobj Many tests fail with the message 
```
SetKernelStatus [Error]: Can not use compressed spikes without sorting connections by source.
```
What is the proper approach to address this? We do want NEST to work correctly with and without compressed spikes, don't we? So the testsuite needs to cover both cases, and for individual tests we need to decide which combinations to use.

nestkernel/connection_manager.cpp
jakobj: yes, it should. the previous 6 lines update the value of `sort_connections_by_source_` so here we can assume that has been set to the desired value, and if that value clashes with the one for `use_compressed_spikes_` we raise an error.

jakobj:
> @jakobj Many tests fail with the message
> 
> ```
> SetKernelStatus [Error]: Can not use compressed spikes without sorting connections by source.
> ```
> 
> What is the proper approach to address this? We do want NEST to work correctly with and without compressed spikes, don't we? So the testsuite needs to cover both cases, and for individual tests we need to decide which combinations to use.

I would've just set `use_compressed_spikes_` to `false` for all tests which explicitly disable sorting by source and would've left it on for all other tests. This way we test both settings.

I'm not sure what the best approach is. If the testsuite setup would (does it?) support it this would be a classic case for parametrized tests in which all tests are run with all possible/meaningful combinations of kernel parameters. Besides software support I'm not sure we have the resources to run such exhaustive testing. :thinking:

heplesser:
> I would've just set `use_compressed_spikes_` to `false` for all tests which explicitly disable sorting by source and would've left it on for all other tests. This way we test both settings.
> 
> I'm not sure what the best approach is. If the testsuite setup would (does it?) support it this would be a classic case for parametrized tests in which all tests are run with all possible/meaningful combinations of kernel parameters. Besides software support I'm not sure we have the resources to run such exhaustive testing. 🤔

@jakobj Your approach makes sense. Parameterized testing would be nice indeed but we don't have proper support for it in SLI, so one would have to do it all by hand, and I wouldn't want to introduce it on the PyNEST side before we have converted properly to PyTest there. With the transition to Github Checks the resource situation is better, but we do want to keep the amount of energy spent on NEST tests in sustainable bounds.

So I suggest that you proceed as you suggested.

jakobj:
alrighty, let's give this another try :D @jhnnsnk

heplesser:
@jakobj `test_multimeter_support.sli` fails with a segfault, most other test failures seem to be conflicts between sorting and compressing. I was also a bit confused by your "Old compressed spikes implementation" commit above—old in which sense?

jakobj:
> @jakobj `test_multimeter_support.sli` fails with a segfault, most other test failures seem to be conflicts between sorting and compressing. I was also a bit confused by your "Old compressed spikes implementation" commit above—old in which sense?

i'll look into the multimeter issue.

my commit refers to porting the implementation based on some old master to the most recent master. the old implementation was benchmarked and found to be beneficial for performance. it hence provides a good starting point. the remaining commits simplify this implementation, throwing out things that are not necessary etc.

jakobj:
fixed the multimeter and related issues. fyi threaded resizing can bite you ;)

ready for benchmarking @jhnnsnk (unless i've overlooked some test failing with segfault)

jhnnsnk:
@jakobj I just benchmarked the microcircuit with your latest commit ( 1539d29 ). Now, the real-time factors show the expected good performance. However, the sum of the `local_spike_counter`s matches the reference master only on one compute node (8 MPI processes, 16 threads per MPI process). For more nodes, the spike compression version counts more spikes than the respective references; the number of additional spikes grows with the number of compute nodes. Results were correct in the earlier version of spike compression with the heuristic commented out.

jakobj:
> @jakobj I just benchmarked the microcircuit with your latest commit ( [1539d29](https://github.com/nest/nest-simulator/commit/1539d29224bfb71be9d979ac8d8ec7477c9ff43d) ). Now, the real-time factors show the expected good performance. However, the sum of the `local_spike_counter`s matches the reference master only on one compute node (8 MPI processes, 16 threads per MPI process). For more nodes, the spike compression version counts more spikes than the respective references; the number of additional spikes grows with the number of compute nodes. Results were correct in the earlier version of spike compression with the heuristic commented out.

thanks for the quick data @jhnnsnk! great to hear the timing is good. i'll try to reproduce the count issue locally and will get back to you.

jakobj:
alright, fixed the issue. locally spike counts match for various combinations of MPI processes and threads. @jhnnsnk

jhnnsnk:
> alright, fixed the issue. locally spike counts match for various combinations of MPI processes and threads. @jhnnsnk

Thanks, @jakobj. Before I start the next test runs, I should probably wait until the branch is updated such that all checks have passed, right?

jakobj:
> > alright, fixed the issue. locally spike counts match for various combinations of MPI processes and threads. @jhnnsnk
> 
> Thanks, @jakobj. Before I start the next test runs, I should probably wait until the branch is updated such that all checks have passed, right?

hmm, not sure what the issues are. some of it seem pretty strange:
2021-04-01-163045_1683x151_scrot

jakobj:
@jhnnsnk i think this is good to be tested again :)

@heplesser i think the error i posted above might indicate an actual problem in the testsuite (`[...] object has no attribute 'fail'`). if test tests succeed it may just be hidden because only the `try` block is executed. could you have a look please?

heplesser:
@jakobj All the tests have passed now, so it seems to have been a temporary glitch in the test system.

jakobj:
> @jakobj All the tests have passed now, so it seems to have been a temporary glitch in the test system.

well, since the tests pass, the line which caused the error is never hit. so even if it would be a bug an not just a temporary glitch, our CI wouldn't catch it. unfortunately i can not reproduce this locally, otherwise i would investigate.

jhnnsnk:
@jakobj Running `make` gives me these warnings related to this PR:
```
[ 58%] Built target randomtest
.../nestkernel/source_table.cpp: In member function ‘void nest::SourceTable::resize_compressible_sources()’:
.../nestkernel/source_table.cpp:455:29: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for ( thread tid = 0; tid < compressible_sources_.size(); ++tid )
                         ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../nestkernel/source_table.cpp: In member function ‘void nest::SourceTable::fill_compressed_spike_data(std::vector<std::vector<std::vector<nest::SpikeData> > >&)’:
.../nestkernel/source_table.cpp:494:29: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for ( thread tid = 0; tid < compressible_sources_.size(); ++tid )
                         ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../nestkernel/source_table.cpp:509:29: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   for ( thread tid = 0; tid < compressible_sources_.size(); ++tid )
                         ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../nestkernel/source_table.cpp:522:53: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
         for ( thread other_tid = tid + 1; other_tid < compressible_sources_.size(); ++other_tid )
                                           ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
It is a minor thing, but maybe it could be fixed easily.

hakonsbm:
The warnings should have been detected by the CI. However, it looks like our GitHub actions setup currently doesn't check for warnings at all. I'll create an issue for this.

heplesser:
> The warnings should have been detected by the CI. However, it looks like our GitHub actions setup currently doesn't check for warnings at all. I'll create an issue for this.

@hakonsbm Could you in this context take a look at #1967? At that point warnings were monitored, but maybe the fix then broke something.

hakonsbm:
@heplesser #1967 was supposed to be fixed by #1980, which modified `parse_travis_log.py`. But because `parse_travis_log.py` was never run by GitHub Actions, the problem still remains.

I've created #2005.

nestkernel/connection_manager.cpp
```diff
@@ -159,6 +162,13 @@ nest::ConnectionManager::set_status( const DictionaryDatum& d )
...
+  {
+    throw KernelException( "Can not use compressed spikes without sorting connections by source." );
```
jhnnsnk, line 169: I would suggest to formulate this in a positive way, e.g., "Spike compression requires connections to be sorted  by source.".

nestkernel/connection_manager.cpp
```diff
@@ -1462,3 +1472,28 @@ nest::ConnectionManager::unset_have_connections_changed( const thread tid )
...
+
+
+void
+nest::ConnectionManager::collect_compressed_spike_data( const thread tid )
+{
+  if ( use_compressed_spikes_ )
+  {
+    if ( not sort_connections_by_source_ )
+    {
+      throw KernelException( "Can not use compressed spikes without sorting connections by source." );
```
jhnnsnk, line 1484: See above.

jhnnsnk (changes requested):
I have here one tiny further suggestion on the Exception message. Apart from that, I would suggest that I perform another test run as soon as the Warnings are fixed.

jhnnsnk:
Thanks, @jakobj!
I have now performed new microcircuit benchmark runs comparing this spike compression branch to its master ancestor on up to 10 nodes, and the results are only positive:
1. The simulations are now correct according to the local spike counter.
2. The simulation times (communication phase) are drastically reduced in line with previous benchmarks.
3. The required total memory is even slightly reduced with spike compression (in contrast to earlier implementations).

Once Travis is happy with this PR, I will approve.

heplesser:
@jakobj Travis seems happy except for PEP8 problems in `pynest/nest/tests/test_spatial/test_basics.py`.

heplesser:
@jakobj Please add documentation for users for the "use_spike_compression" flag to the `SetKernelStatus` docstring in `pynest/nest/lib/hl_api_simulation.py`.

Could you also add at a suitable place (connection_manager.h|cpp) some information for developers about how the compression mechanism works, so that we won't have to decode this from the code five years from now. Or is that written up elsewhere?

nestkernel/connection_manager.cpp
```diff
@@ -1462,3 +1472,28 @@ nest::ConnectionManager::unset_have_connections_changed( const thread tid )
...
+      source_table_.resize_compressible_sources();
+    } // of omp single; implicit barrier
+
+    source_table_.collect_compressible_sources( tid );
+#pragma omp barrier
+#pragma omp single
+    {
+      source_table_.fill_compressed_spike_data( compressed_spike_data_ );
+    }
+  } // of omp single; implicit barrier
```
heplesser, line 1498: Isn't this the closing bracket of `if (use_compressed_spikes_)`?

nestkernel/connection_manager.h
```diff
@@ -315,6 +318,8 @@ class ConnectionManager : public ManagerInterface
    */
   bool get_sort_connections_by_source() const;
 
+  bool get_use_compressed_spikes() const;
```
heplesser, line 321: I think code calling this method will be more readable if we call it just `use_compressed_spikes()`.

nestkernel/source_table.cpp
```diff
@@ -408,3 +448,103 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+  }
+}
+
+void
+nest::SourceTable::collect_compressible_sources( const thread tid )
+{
+  for ( synindex syn_id = 0; syn_id < sources_[ tid ].size(); ++syn_id )
+  {
+    index lcid = 0;
+    while ( lcid < sources_[ tid ][ syn_id ].size() )
```
heplesser, line 469: To make this code more readable, I'd introduce
```c++
auto &syn_sources = sources_[ tid ][ syn_id ];
```
and then use that in what follows.

nestkernel/source_table.cpp
```diff
@@ -408,3 +448,103 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+    {
+      const std::pair< index, SpikeData > source_node_id_to_spike_data =
+        std::make_pair( sources_[ tid ][ syn_id ][ lcid ].get_node_id(), SpikeData( tid, syn_id, lcid, 0 ) );
+      compressible_sources_[ tid ][ syn_id ].insert( source_node_id_to_spike_data );
+
+      // find next source with different node_id (assumes sorted sources)
+      const index old_source_node_id = sources_[ tid ][ syn_id ][ lcid ].get_node_id();
```
heplesser, line 476: If you move this before the `make_pair` line, you can save one call to `get_node_id()` and improve clarity.

nestkernel/source_table.cpp
```diff
@@ -408,3 +448,103 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+}
+
+void
+nest::SourceTable::fill_compressed_spike_data(
+  std::vector< std::vector< std::vector< SpikeData > > >& compressed_spike_data )
+{
+  compressed_spike_data.clear();
+  compressed_spike_data.resize( kernel().model_manager.get_num_synapse_prototypes() );
+
+  for ( thread tid = 0; tid < static_cast< thread >( compressible_sources_.size() ); ++tid )
```
heplesser, line 494: Replace by `#pragma omp parallel`?

nestkernel/source_table.cpp
```diff
@@ -408,3 +448,103 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+  {
+    compressed_spike_data_map_[ tid ].clear();
+    compressed_spike_data_map_[ tid ].resize(
+      kernel().model_manager.get_num_synapse_prototypes(), std::map< index, size_t >() );
+  }
+
+  std::vector< SpikeData > spike_data; // will hold spike data
```
heplesser, line 501: The comment here looks confusing. Put rather a longer comment explaining what is going on in the following ahead of this line, including why we need to loop over threads here.

nestkernel/source_table.h
```diff
@@ -119,11 +120,14 @@ class SourceTable
   bool previous_entry_has_same_source_( const SourceTablePosition& current_position,
     const Source& current_source ) const;
 
-  void populate_target_data_fields_( const SourceTablePosition& current_position,
+  bool populate_target_data_fields_( const SourceTablePosition& current_position,
     const Source& current_source,
     const thread source_rank,
     TargetData& next_target_data ) const;
 
+  std::vector< std::vector< std::map< index, SpikeData > > > compressible_sources_;
```
heplesser, line 128: Add doxygen comments describing purpose and structure of the two new data structures.

nestkernel/source_table.h
```diff
@@ -119,11 +120,14 @@ class SourceTable
   bool previous_entry_has_same_source_( const SourceTablePosition& current_position,
     const Source& current_source ) const;
 
-  void populate_target_data_fields_( const SourceTablePosition& current_position,
```
heplesser, line 122: Add doxygen comment for method.

nestkernel/event_delivery_manager.cpp
```diff
@@ -620,17 +620,39 @@ EventDeliveryManager::deliver_events_( const thread tid, const std::vector< Spik
...
+          kernel().connection_manager.send( tid, syn_id, lcid, cm, se );
+        }
+      }
+      else
+      {
         const index syn_id = spike_data.get_syn_id();
-        const index lcid = spike_data.get_lcid();
-        const index source_node_id = kernel().connection_manager.get_source_node_id( tid, syn_id, lcid );
-        se.set_sender_node_id( source_node_id );
+        const index idx =
```
heplesser, line 641: Move comment ahead of line for better formatting.

heplesser (changes requested):
@jakobj This looks quite fine, I mainly have suggestions for more documentation and some improvements for better code clarity.

jakobj:
> @jakobj Travis seems happy except for PEP8 problems in `pynest/nest/tests/test_spatial/test_basics.py`.

i'm really unsure how to read the massive travis output, but from what i can decipher, these errors have nothing to do with this PR:
```shell
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:42:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:49:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:60:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:89:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:164:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:251:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:287:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:310:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:357:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:375:9: E741 ambiguous variable name 'l'
MSGBLD0195: [PEP8] pynest/nest/tests/test_spatial/test_basics.py:409:9: E741 ambiguous variable name 'l'
```

i guess fixing this should be a separate PR?

jakobj:
> @jakobj Please add documentation for users for the "use_spike_compression" flag to the `SetKernelStatus` docstring in `pynest/nest/lib/hl_api_simulation.py`.
> 
> Could you also add at a suitable place (connection_manager.h|cpp) some information for developers about how the compression mechanism works, so that we won't have to decode this from the code five years from now. Or is that written up elsewhere?

done.

may i point out that the duplication of information between python docstrings and c++ files is suboptimal. maybe there's some way to merge the two in the future?

nestkernel/connection_manager.h
```diff
@@ -540,6 +547,8 @@ class ConnectionManager : public ManagerInterface
    */
   SourceTable source_table_;
 
+  std::vector< std::vector< std::vector< SpikeData > > > compressed_spike_data_;
```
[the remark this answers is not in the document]
jakobj: ok

nestkernel/connection_manager.h
```diff
@@ -598,6 +607,8 @@ class ConnectionManager : public ManagerInterface
   //! Whether to sort connections by source node ID.
   bool sort_connections_by_source_;
 
+  bool use_compressed_spikes_;
```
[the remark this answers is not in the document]
jakobj: ok

nestkernel/connection_manager.cpp
```diff
@@ -1462,3 +1472,28 @@ nest::ConnectionManager::unset_have_connections_changed( const thread tid )
...
+
+void
+nest::ConnectionManager::collect_compressed_spike_data( const thread tid )
+{
+  if ( use_compressed_spikes_ )
+  {
+    if ( not sort_connections_by_source_ )
+    {
+      throw KernelException( "Spike compression requires sort_connections_by_source to be true." );
+    }
```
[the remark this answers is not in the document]
jakobj: ok

nestkernel/connection_manager.h
jakobj: ok

nestkernel/source_table.cpp
jakobj: ok

nestkernel/source_table.cpp
jakobj: ok

nestkernel/source_table.cpp
jakobj: can't, we're in an omp single block https://github.com/jakobj/nest-simulator/blob/enh/com-com/nestkernel/connection_manager.cpp#L1494

nestkernel/source_table.cpp
jakobj: ok

nestkernel/source_table.h
jakobj: ok

nestkernel/source_table.h
jakobj: ok

nestkernel/connection_manager.cpp
jakobj: yes!

heplesser:
@jakobj I just sent you a PR that fixes the "ambiguous l". The problem popped up here because you made an unrelated change to `test_basic.py` and thus exposed the file to PEP8's radar.

nestkernel/source_table.cpp
heplesser: OK!

heplesser (approved):
@jakobj Very nice! I approve, merging should wait until Travis is happy (see also my PR to you fixing the l-problem).

jhnnsnk:
@jakobj I just noticed that `use_compressed_spikes` is not accessible from the kernel status dictionary. Apparently the flag can be set ( `nest.SetKernelStatus({'sort_connections_by_source': False, 'use_compressed_spikes': False}` works ) but not read out ( `nest.GetKernelStatus('use_compressed_spikes')` fails with a KeyError). This behavior is not intended, isn't it?

diesmann:
> @jakobj I just noticed that `use_compressed_spikes` is not accessible from the kernel status dictionary. Apparently the flag can be set ( `nest.SetKernelStatus({'sort_connections_by_source': False, 'use_compressed_spikes': False}` works ) but not read out ( `nest.GetKernelStatus('use_compressed_spikes')` fails with a KeyError). This behavior is not intended, isn't it?

In situations where we have such flags, maybe we should have a test whether the status can be read out anyway.

jessica-mitchell:
> may i point out that the duplication of information between python docstrings and c++ files is suboptimal. maybe there's some way to merge the two in the future?

Hi @jakobj - just piping in here to say that reducing the redundancy in information is something the documentation team is trying to work on. Thanks for pointing out a specific instance of the problem - it's especially helpful to know where in the c++ code we have this issue. A possible way to merge the two in future is that we are planning to put the developer docs on ReadtheDocs, alongside the userdocs (#1844). I think this might give us a better chance to remove some redundancy and make it easier to cross-reference information.

heplesser:
> > may i point out that the duplication of information between python docstrings and c++ files is suboptimal. maybe there's some way to merge the two in the future?
> 
> Hi @jakobj - just piping in here to say that reducing the redundancy in information is something the documentation team is trying to work on. Thanks for pointing out a specific instance of the problem - it's especially helpful to know where in the c++ code we have this issue. A possible way to merge the two in future is that we are planning to put the developer docs on ReadtheDocs, alongside the userdocs (#1844). I think this might give us a better chance to remove some redundancy and make it easier to cross-reference information.

Duplication is one thing, complementarity another. Python docstrings need to tell the user what she needs to know to use NEST features. C++ documentation tells the developer what she needs to know to understand how the code works, what data structures represent and how they interact, to be able to maintain and evolve the code. Parameter names in PyNEST may match variable names in C++ code, but that is not a law of nature, the C++ implementation could change while PyNEST stays the same, so some duplicity of information cannot reasonably be avoided. Also, when working with C++ code, you will not always want to have to look at PyNEST to find information.

jakobj:
alright, i've now implemented getting the kernel setting as suggested by @jhnnsnk and included the pep8 fixes from @heplesser. also, i've rewritten the history to be nice and clean. please have another look.

heplesser:
@jakobj The CI failure in the MPI_ONLY case is due to failing `ticket-459.sli` (the long list of failed tests at the end is a testsuite error, see #1945). I believe this may be due to a variable not properly initialised and have just sent you a PR to fix this.

jakobj:
> @jakobj The CI failure in the MPI_ONLY case is due to failing `ticket-459.sli` (the long list of failed tests at the end is a testsuite error, see #1945). I believe this may be due to a variable not properly initialised and have just sent you a PR to fix this.

thanks for the quick fix! :)

great teamwork, thanks to everyone who has contributed to this PR. everything looks good now, doesn't it?

jhnnsnk:
I will do a final test run as soon as JURECA DC is back from maintenance.

jhnnsnk (approved):
I approve! :+1: 
My final benchmarks were successful: this PR greatly improves simulation times for models similar to the microcircuit. I did not encounter any issues any more.
Thank you, @jakobj, very much for your efforts!

diesmann:
Congratulations and a big Thanks to all of you.

Issue #1337, opened 2019-11-15
Can't install NEST with Python 3.8


Boboz8:
I get Python 3.8 from `apt -get install python 3.8` and i followed this guide here: https://nest-simulator.readthedocs.io/en/latest/installation/linux_install.html#standard-installation , i followed the standard installation process, added `-Dwith-python=3` to the cmake command and in the make process it give me this:

```
[ 99%] Built target nest
[100%] Compiling Cython CXX source for pynestkernel...
warning: /home/bobo/nest-simulator-2.18.0/pynest/pynestkernel.pyx:689:10: Unreachable code
warning: /home/bobo/nest-simulator-2.18.0/pynest/pynestkernel.pyx:689:10: Unreachable code
Scanning dependencies of target pynestkernel
[100%] Building CXX object pynest/CMakeFiles/pynestkernel.dir/pynestkernel.cxx.o`
/home/bobo/nest-simulator-2.18.0-build/pynest/pynestkernel.cxx: In function ‘PyObject* __pyx_f_12pynestkernel_sli_array_to_object(ArrayDatum*)’:
/home/bobo/nest-simulator-2.18.0-build/pynest/pynestkernel.cxx:11220:35: warning: comparison of integer expressions of different signedness: ‘Py_ssize_t’ {aka ‘long int’} and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
     for (__pyx_temp=0; __pyx_temp < __pyx_v_dat->size(); __pyx_temp++) {
                        ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/bobo/nest-simulator-2.18.0-build/pynest/pynestkernel.cxx:11247:33: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘Py_ssize_t’ {aka ‘long int’} [-Wsign-compare]
   for (__pyx_t_4 = 0; __pyx_t_4 < __pyx_t_3; __pyx_t_4+=1) {
                       ~~~~~~~~~~^~~~~~~~~~~
/home/bobo/nest-simulator-2.18.0-build/pynest/pynestkernel.cxx: In function ‘int __Pyx_modinit_type_init_code()’:
/home/bobo/nest-simulator-2.18.0-build/pynest/pynestkernel.cxx:28268:38: warning: ‘_typeobject::tp_print’ is deprecated [-Wdeprecated-declarations]
[...]

PR #1340, opened 2019-11-19
Fix to allow installation with Python 3.8

steffengraber:
This is a fix to allow installation with python 3.8.
After building NEST with python 3.8, an `import nest` returns
the following error. 

```
  File "nest-simulator_1574171714961/test_tmp/run_test.py", line 2, in <module>
    import nest
  File "nest-simulator_1574171714961/lib/python3.8/site-packages/nest/__init__.py", line 30, in <module>
    from .hl_api import *      # noqa
  File "nest-simulator_1574171714961/lib/python3.8/site-packages/nest/hl_api.py", line 40, in <module>
    _il.import_libs(__file__, globals(), 'lib', ignore=_ignore_modules)
  File "nest-simulator_1574171714961/lib/python3.8/site-packages/nest/import_libs.py", line 73, in import_libs
    code = compile(module, mod_file, 'exec')
TypeError: required field "type_ignores" missing from Module
```

There is a further explanation here:
https://bugs.python.org/issue35894


steffengraber:
This PR fixes #1337.

pynest/nest/tests/test_sp/test_growth_curves.py
```diff
@@ -26,6 +26,7 @@
 import unittest
 import nest
 import time
+import sys
 HAVE_OPENMP = nest.ll_api.sli_func("is_threaded")
```
heplesser: Shouldn't there be an empty line after the last import?

pynest/nest/tests/test_sp/test_growth_curves.py
```diff
@@ -306,15 +307,21 @@ def simulate(self):
         self.se_python = numpy.zeros(
             (len(self.se_integrator), len(self.sim_steps)))
 
-        start = time.clock()
+        if sys.version_info >= (3, 8):
+            start = time.process_time()
+        else:
+            start = time.clock()
```
heplesser: Three questions here: 
1. I might be blind, but I don't see that the `start` variable is used anywhere in the remainder of the test. Maybe we can just delete the code entirely?
2. If I am wrong on 1., the code occurs twice and may be relevant in other places as well (benchmarks, ...). Should we wrap this in a helper function?
3. `process_time()` has been available since Python 3.3. I would suggest to change the test to `>= (3, 3)` therefore.

`process_time()` has been available since P

steffengraber: After I made a few test installations, I am happy to simply delete the code. 
If it's needed elsewhere (benchmarks ..) I can't say. 
@jougs do you know more about that?

jougs: I don't know more, but am always happy if code gets deleted.

PR #1274, opened 2019-08-27
Fix pynest api generation for documentation using mock

jessica-mitchell:
This PR provides a solution for the PyNEST api generation on readthedocs.

The problem: pynest apis do not work on readthedocs.
For sphinx to auto generate the apis (sphinx-autodoc) for the website, it requires a running instance of nest. If we try to include nest in the build on ReadtheDocs, it times out because of resource limitations. 

The solution:
Mock nest 
All the credit goes to  @Silmathoron, who provided the scripts/code required to mock nest!

The PyNEST apis functions were cleaned up a bit then indexed, and referenced in the docs 

See output here: https://nest-test.readthedocs.io/en/pynest_mock/ref_material/pynest_apis.html

PR #1343, opened 2019-11-23
Update syntax formatting for terms in documentation

jessica-mitchell:
This PR updates the format of terms for more consistency and creates links for models, functions and glossary terms.
Basically several terms in double back ticks were changed to single back ticks if they were parameters, variables or values.
A script was used (thanks to @terhorstd for helping with this) to replace the models, function and glossary terms into a linkable format in reStructuredText files and the pynest examples.
Links to functions will not work until the PR #1274 is merged

Note: The multimeter name needs to still be updated in the model file because it is the only one that is capitalized.


jessica-mitchell:
Closing in favour of working in nest-3

PR #1344, opened 2019-11-24
Fix broken css style

jessica-mitchell:
This PR removes a custom css line that caused some strange formatting in the html.


jougs:
Thanks!

Issue #1305, opened 2019-10-02
Rounding error in small resolutions


weidel-p:
Setting the resolution to a small value often leads to rounding errors and unexpected behavior.

If I understand correctly, the resolution must be a multiple of the tic length which is by default _0.001 ms_. So a resolution of _0.105 ms_ and _0.102 ms_ should both work.

When I try to set these resolutions, the following happens:

`nest.SetKernelStatus({'resolution': 0.105})` -> `Temporal resolution changed.`

`nest.SetKernelStatus({'resolution': 0.102})` -> `Resolution must be a multiple of the tic length. Value unchanged.`

It's clear that this is a rounding issue as changing the representation of _0.102_ slightly by multiplying _0.0102_ by _10_ fixes the issue.

`nest.SetKernelStatus({'resolution': 0.0102 * 10})`  -> `Temporal resolution changed.`

I assume the responsible line of code is in `simulation_manager.cpp`:

`std::modf( resd * tics_per_ms, &integer_part ) != 0` 

Would it be sufficient to change the comparison with _0_ to a small epsilon or would that lead to other issues?

I'm using the Git commit:
0ac2385db24bdf76c43f310ee5f7da10d274c457

heplesser:
@weidel-p Provided that epsilon is very small indeed (corresponding to the last 1-2 bits), this should be safe. We have some sensitive precision test, they might trigger if my assumption is wrong. Have you had a look at the blame-history to find the issue that introduced the problematic code? That might help us understand better why code is as it is, but in general, comparing floats to 0 exactly makes no sense precisely because of rounding errors.

If you need very high precision/resolution, you may want to consider to change the `tics_per_ms` and `resolution` to powers of 2.

PR #1356, opened 2019-12-05
Fix rounding issue when setting small resolution

weidel-p:
As described in  #1305 there is a problem setting the resolution to a small numbers.
The origin of this issue is a rounding error and the comparison of `modf` to `0` in `simulation_manager.cpp`.

In this PR I fixed this issue by comparing the the result of `modf` to a small epsilon instead of `0`.

### DIFF
--- a/libnestutil/numerics.h
+++ b/libnestutil/numerics.h
@@ -125,5 +125,10 @@ double dround( double );
  */
 double dtruncate( double );
 
+/**
+ * Returns true if n is integer
+ * @return bool
+ */
+bool is_integer( double );
 
 #endif
--- a/libnestutil/numerics.h
+++ b/libnestutil/numerics.h
@@ -126,7 +126,7 @@ double dround( double );
 double dtruncate( double );
 
 /**
- * Returns true if n is integer
+ * Returns true if n is integer up to rounding error.
  * @return bool
  */
 bool is_integer( double );
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -102,6 +102,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
   TimeConverter time_converter;
 
   double time;
+  double eps = 1e-10;
   if ( updateValue< double >( d, names::time, time ) )
   {
     if ( time != 0.0 )
@@ -176,7 +177,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
           "unchanged." );
         throw KernelException();
       }
-      else if ( std::modf( resd * tics_per_ms, &integer_part ) != 0 )
+      else if ( std::modf( resd * tics_per_ms + eps, &integer_part ) > 2 * eps )
       {
         LOG( M_ERROR,
           "SimulationManager::set_status",
@@ -212,7 +213,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
           "unchanged." );
         throw KernelException();
       }
-      else if ( std::modf( resd / Time::get_ms_per_tic(), &integer_part ) != 0 )
+      else if ( std::modf( resd / Time::get_ms_per_tic() + eps, &integer_part ) > 2 * eps )
       {
         LOG( M_ERROR,
           "SimulationManager::set_status",
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -177,7 +177,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
           "unchanged." );
         throw KernelException();
       }
-      else if ( !is_integer( resd * tics_per_ms ) )
+      else if ( not is_integer( resd * tics_per_ms ) )
       {
         LOG( M_ERROR,
           "SimulationManager::set_status",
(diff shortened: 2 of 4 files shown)


nestkernel/simulation_manager.cpp
```diff
@@ -102,6 +102,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
   TimeConverter time_converter;
 
   double time;
+  double eps = 1e-10;
```
heplesser: Please make this `const` and give it a more expressive name (ALL_CAPS). Could you reduce the value even further, around 1e-14?

weidel-p:
Thanks, I applied your suggested changes.

nestkernel/simulation_manager.cpp
```diff
@@ -102,6 +102,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
   TimeConverter time_converter;
 
   double time;
+  const double RESOLUTION_TOLERANCE = 1e-14;
```
jougs: Why can't [`std::numeric_limits<T>::epsilon`](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) be used here? I dislike the idea of hard-coded magic number (especially if they come without any comment)?

weidel-p: Good point. I tried, but I found the error is actually much larger than than this epsilon (2.2e-16).

For example if I try to set the resolution to `0.102` ms and leave the `tics_per_ms` at 1000, it should work, but there is a significant rounding error happening.

`resd / Time::get_ms_per_tic()` should yield a round `102`, but is actually `101.99999999999998578915`.

heplesser: Ah, it is all relative ... For a value around 100, round-off error should be around 100 * eps ~ 10^-14, as is the case here. So I'd suggest

```C++
bool is_integer( double n )
{
  double int_part;
  double frac_part = std::modf( n, &int_part );

  // Since n > 0 and modf always rounds towards zero, a value of n just below an 
  // integer will result in frac_part = 0.99999.... . We subtract from 1 in this case. 
  if ( frac_part > 0.5 )
  {
    frac_part = 1 - frac_part;
  }
  
  // factor 4 allows for two bits of rounding error
  return frac_part < 4 * n * std::numeric_limits<double>::epsilon;
}
```

This should probably go somewhere in `libnestutil`.

weidel-p: Alright, that's a good solution. I hope that's better now.

libnestutil/numerics.h
```diff
@@ -125,5 +125,10 @@ double dround( double );
  */
 double dtruncate( double );
 
+/**
+ * Returns true if n is integer
```
heplesser: Could you expand the documentation to point out that this means "integer up to rounding error"?

nestkernel/simulation_manager.cpp
```diff
@@ -177,7 +177,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
           "unchanged." );
         throw KernelException();
       }
-      else if ( std::modf( resd * tics_per_ms + RESOLUTION_TOLERANCE, &integer_part ) > 2 * RESOLUTION_TOLERANCE )
+      else if ( !is_integer( resd * tics_per_ms ) )
```
heplesser: `not is_integer(...)`

nestkernel/simulation_manager.cpp
```diff
@@ -213,7 +213,7 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
           "unchanged." );
         throw KernelException();
       }
-      else if ( std::modf( resd / Time::get_ms_per_tic() + RESOLUTION_TOLERANCE, &integer_part ) > 2 * RESOLUTION_TOLERANCE )
+      else if ( !is_integer( resd / Time::get_ms_per_tic() ) )
```
heplesser: `not is_integer(...)`

weidel-p:
I added a test and implemented your suggestions.

nestkernel/simulation_manager.cpp
jougs: Awesome!

jougs:
@weidel-p can you please fix the formatting? Thanks!

weidel-p:
Sure, should be correct now.

testsuite/regressiontests/issue-1305.sli
```diff
@@ -0,0 +1,63 @@
...
+M_ERROR setverbosity
+
+% Test setting valid resolution 
+{
+  ResetKernel
+  
+  0 << /resolution 0.102 /tics_per_ms 1000.0 >> SetStatus
+  0 GetStatus
+  /res {/resolution get} def
+  /delta {res 0.102 sub} def
```
heplesser: I find the code here very confusing, it took me a few minutes to understand what is going on ... 

How about
```
/target_resolution 0.102 def
0 << /resolution target_resolution /tics_per_ms 1000.0 >> SetStatus

0 /resolution get 
target_resolution 
sub abs 
1e-15 leq

weidel-p: SLI is not my native language, so sure, everything you want. Changed.

PR #1367, opened 2019-12-13
Ensure number of connections for fixed total number is properly set

### LINKED ISSUE #1366: Problem with fixed_total_number and gsl_binomial
I am having problems with connecting with a `fixed_total_number` of `N=1` when using mpi and GSL.  A very small reproducer script is included:
[test_gsl.py.zip](https://github.com/nest/nest-simulator/files/3959629/test_gsl.py.zip)
It sets `total_num_virtual_procs` to be 4, creates 5 neurons, and tries to connect the neurons with `fixed_total_number` of 1.

If I run

```
mpirun -np 1 python test_gsl.py 
```

I get the following error
```
  File "/home/stine/Desktop/test_gsl.py", line 8, in <module>
    nest.Connect(nodes, nodes, {'rule': 'fixed_total_number', 'N': 1})
  File "/home/stine/NEST/bld_master/install/lib/python3.6/site-packages/nest/ll_api.py", line 246, in stack_checker_func
    return f(*args, **kwargs)
  File "/home/stine/NEST/bld_master/install/lib/python3.6/site-packages/nest/lib/hl_api_connections.py", line 317, in Connect
    sr('Connect')
  File "/home/stine/NEST/bld_master/install/lib/python3.6/site-packages/nest/ll_api.py", line 132, in catching_sli_run
    raise exceptionCls(commandname, message)
nest.lib.hl_api_exceptions.BadParameterValue: ('BadParameterValue in Connect_g_g_D_D: gsl_binomial RDV: n >= 1 required.', 'BadParameterValue', <SLILiteral: Connect_g_g_D_D>, ': gsl_binomial RDV: n >= 1 required.')
-------------------------------------------------------
```

The prblem comes, as the error message suggest, from `gsl_binomial_randomdev.cpp` line 104 where we check the value of `n_` before setting it (`n_` is number of connections in this instance, and should be 1)
https://github.com/nest/nest-simulator/blob/dc212c0b1d055895a31beb3987d4e76e4dac64aa/librandom/gsl_binomial_randomdev.cpp#L104-L108

If I write `n_s` to screen, I get
```
n_s: 1
n_s: 1
n_s: 1
n_s: 0
```
Shouldn't I get `n_s: 1` for all four processes?

stinebuu:
This PR fixes #1366.

For `fixed_total_number`, a loop is used to set the number of connections on all processes. This PR checks if all connections have already been distributed, and if so, exits the loop. This way, if we have created all connections on the other processes, we will not get a problem with gsl binomial, and the number of connections on the last processes will automatically be zero (because all is initialized to zero).

### DIFF
--- a/nestkernel/conn_builder.cpp
+++ b/nestkernel/conn_builder.cpp
@@ -1376,8 +1376,12 @@ nest::FixedTotalNumberBuilder::connect_()
   librandom::BinomialRandomDev bino( grng, 0, 0 );
 #endif
 
-  for ( int k = 0; k < M; k++ )
+  for ( int k = 0; k < M - 1; k++ )
   {
+    if ( N_ == sum_partitions )
+    {
+      break;
+    }
     if ( number_of_targets_on_vp[ k ] > 0 )
     {
       double num_local_targets = static_cast< double >( number_of_targets_on_vp[ k ] );
@@ -1390,6 +1394,9 @@ nest::FixedTotalNumberBuilder::connect_()
     sum_dist += static_cast< double >( number_of_targets_on_vp[ k ] );
     sum_partitions += static_cast< unsigned int >( num_conns_on_vp[ k ] );
   }
+
+  // Number of connections on last process is N_ - connections on other vp's
+  num_conns_on_vp[ M - 1 ] = N_ - sum_partitions;
 
 // end code adapted from gsl 1.8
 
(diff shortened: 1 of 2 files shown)


nestkernel/conn_builder.cpp
```diff
@@ -1376,8 +1376,12 @@ nest::FixedTotalNumberBuilder::connect_()
   librandom::BinomialRandomDev bino( grng, 0, 0 );
 #endif
 
-  for ( int k = 0; k < M; k++ )
+  for ( int k = 0; k < M - 1; k++ )
```
heplesser: @stinebuu You may want to add a comment here to explain why we only need to go to M-1. Otherwise, we will probably wonder about this a year from now.

stinebuu:
@heplesser @suku248 I have gone back to iterating through all processes, and commenting the `if` instead, as this made it easier to read.

I thus no longer set the number of connections on the last process manually.

PR #1358, opened 2019-12-06
Fix Arbor recording backend.

w-klijn:
Updated and now correctly implemented NESTIO arbor back-end. 
Based on the original code from Alex Peyser.

- recording_backend for arbor.
- Directory with test files to run the example, including a pure python Arbor simulator fake

See https://github.com/arbor-sim/arbor/pull/913 for the other side of the interface.

PR #1370, opened 2019-12-17
Fix arbor recording backend

jougs:
This replaces #1358.


jougs:
As I have already reviewed @w-klijn's commits in #1358 andthis one just adds formatting changes, I am merging without waiting for another review.

Issue #1315, opened 2019-10-17
only partially overridden in class ... 


ikitayama:
With a proprietary compiler, I am seeing a lot of warnings like below:

```
"/home/users/go21/go0009/projects/nest-simulator/nestkernel/parameter.h", line 281: warning: overloaded virtual function "nest::Parameter::value" is only partially overridden in class "nest::NormalParameter"
  class NormalParameter : public Parameter
```
Can someone explain as to why it wasn't fully implemented?

heplesser:
This is in the nest-3 branch, isn't it? Any chance this could be related to missing `using` directives?

ikitayama:
Yes, I am on nest-3.

ikitayama:
With an old GCC that comes with RHEL 7.5, I see no warnings.

hakonsbm:
I can confirm getting similar warnings as @ikitayama for `parameter.h` when compiling the nest-3 branch with GCC using the `-Woverloaded-virtual` flag. A warning is issued because a function declaration hides virtual functions from the base class. This can be solved using `using` directives, as noted by @heplesser.

Similar warnings also occur when compiling master, however, for the following member functions:
- `raw_value()` of topology parameter subclasses
- `inside()` of mask subclasses
- `handles_test_event()` of rate_neuron_ipn
- `handles_test_event()` of rate_transformer_node
- `handles_test_event()` of rate_neuron_opn
- `handles_test_event()` of siegert_neuron

heplesser:
@hakonsbm We should fix the missing `using` directives everywhere. I would also suggest to add `-Woverloaded-virtual` to the compiler flags for testing on Travis, maybe just add them to the default compiler flags.

@lekshmideepu @gtrensch Are we counting/reporting compiler warnings in any way? We have the declared aim that NEST should compile without warnings, except where they arise from code beyond our control (Python interface).

lekshmideepu:
@heplesser To my knowledge, we are not counting/reporting compiler warnings.

PR #1377, opened 2020-01-03
Add missing using directives

hakonsbm:
Adds `using` directives in classes where a virtual function from a base class is hidden. The `-Woverloaded-virtual` flag is also added to default compiler flags. This is a partial fix of #1315.


hakonsbm:
Apparently adding `Woverloaded-virtual` to default compiler flags causes a lot of warnings when compiling with MUSIC, which makes Travis terminate the job.

heplesser:
@clinssen Could you review, as this is also relevant for code generation from NESTML?

heplesser:
@hakonsbm Sorry it took me long to get back to this. Could you fix the merge conflicts? then this one should be ready to be merged. Could you create a follow-up issue to add the `-Woverloaded-virtual` flag, to be done once https://github.com/INCF/MUSIC/pull/61 has been merged?

PR #1380, opened 2020-01-09
Implement communication via Alltoallv for secondary events

jakobj:
This PR switches the communication scheme for secondary events from `MPI Alltoall` to `MPI Alltoallv`, i.e., it allows for sending chunks of different size to each MPI process. Since for secondary events we can compute the chunk sizes before the simulation starts, this does not entail any overhead during simulation, while avoiding communicating empty buffer entries for structured, e.g., nearest neighbor, networks in which the amount of data communicated to each process is not uniform.

Work with @suku248.

As reviewers I suggest @janhahne @stinebuu @hakonsbm @jougs


nestkernel/nest_names.cpp
```diff
@@ -74,7 +74,8 @@ const Name b( "b" );
 const Name beta( "beta" );
 const Name beta_Ca( "beta_Ca" );
 const Name binary( "binary" );
-const Name buffer_size_secondary_events( "buffer_size_secondary_events" );
+const Name send_buffer_size_secondary_events( "send_buffer_size_secondary_events" );
+const Name recv_buffer_size_secondary_events( "recv_buffer_size_secondary_events" );
```
stinebuu: `nest_names.{cpp, h}` are sorted alphabetically, could you place these under s and r?

nestkernel/source_table.cpp
```diff
@@ -365,16 +368,15 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
 
-        const size_t recv_buffer_pos = kernel().connection_manager.get_secondary_recv_buffer_position(
-          current_position.tid, current_position.syn_id, current_position.lcid );
-
-        // convert receive buffer position to send buffer position
-        // according to buffer layout of MPIAlltoall
-        const size_t send_buffer_pos =
-          kernel().mpi_manager.recv_buffer_pos_to_send_buffer_pos_secondary_events( recv_buffer_pos, source_rank );
+        // compute the buffer position relative to the first position
+        // the source rank will write to from the absolute position in
```
stinebuu: Maybe change to `the source rank will write the absolute position in...`?

suku248:
@H27CK do you think you could review this, too?
@jakobj just a reminder that this branch needs updating

H27CK:
@suku248 yes, can do.

H27CK:
@jakobj, a couple of queries:
1. do you have any timing measurements for this implementation?
2. what is the level of inhomogeneity in the communication of secondary events? You mention that this avoids communication of empty chunks for structured networks by Alltoall, but how many empty chunks can one expect? This is case dependent. This poses the question: is it always optimal to use Alltoallv with its associated overhead? Is there any reason for this to be introduced as an option at build time to allow users to select Alltoall or Alltoallv as required?

jakobj:
@suku248 right, will rebase onto current master :grimacing: 

@H27CK 
1. there are timing measurements in https://doi.org/10.3389/fninf.2020.00012, but i guess you are asking for comparison between 5g Alltoall vs 5g Alltoallv. i'm sure we measured that at some point, will search for the data as soon as JUWELS is up again. 
2. you are absolutely right, the imbalance and hence number of empty chunks heavily depends on the user-defined network structure. please note however that here we do not introduce overhead for communicating chunk sizes before each Alltoallv call, as this is necessary only a single time at the beginning of the simulation (provided network structure does not change during runtime). it might however be that MPI implementations have heavily optimized Alltoall, but not Alltoallv. hopefully this would be visible in the data mentioned in 1). personally I would prefer to not have another compile-time option that could potentially be messed up, unless we observe huge runtime differences.

H27CK:
@jakobj 

1. Thanks, I will have a look over this.
2. The overhead I refer to is internal to MPI. MPI needs to read, interpret, and act on the information in the 4 additional vectors you provide (even if they do not change), i.e. `send_counts`, `send_displacements`, `recv_counts`, `recv_displacements`, to make some decisions about communication internally. This incurs some cost. I have seen this myself when experimenting with Alltoallv for spike communication. Depending on the degree of heterogeneity, savings are diminished by the need to treat the aforementioned vectors. It is important that the time saved by avoiding communication of empty chunks is not lost to some internal MPI operations. I agree that an additional build option increases complexity, and I'd rather not recommend such a strategy, but I am currently reluctant to say that one solution is optimal for all cases. We must either: 1) ensure that Alltoallv is indeed the preferred choice for the majority of cases (covering a range of networks and systems), and potentially allow the performance of some cases to suffer (I'll read the article in 1. for this); 2) provide the user with an option to apply the communication method that suits their needs; 3) have NEST determine which strategy is best, increasing complexity further. Any other ideas?

jakobj:
>     2. [...snip...] 1) ensure that Alltoallv is indeed the preferred choice for the majority of cases (covering a range of networks and systems), and potentially allow the performance of some cases to suffer (I'll read the article in 1. for this); 2) provide the user with an option to apply the communication method that suits their needs; 3) have NEST determine which strategy is best, increasing complexity further. Any other ideas?

I fully agree. I think for a data-based decision we should look at benchmarks without load imbalance, i.e., optimal regime for Alltoall and quantify how much more expensive the Alltoallv version is. for this I would rely on the benchmark case we also used in our manuscript. an exhaustive analysis of the majority of cases is unfortunately practically not feasible without significant investment of resources. in terms of increasing complexity (2,3): between the two evils I would always opt for increase in code complexity.

jakobj:
rebased, but still need to make sure nothing got lost on the way.

jakobj:
@H27CK here are finally some benchmarks comparing master (38f8a55) and the alltoallv branch (9b3bdf9) for two different networks: one with random connectivity and approx 60 gap junctions per neuron (top), and one with ring topology and exactly 60 gap junctions per neuron (bottom, cf. https://doi.org/10.3389/fninf.2020.00012). while for the random network both implementations achieve similar runtime, the alltoallv branch clearly outperforms the current master when the network topology exhibits a large degree of organization. as discussed before this difference is caused by the current master communicating a large amount of unnecessary data.

something that concerns me a little is the linear increase in build time for both master and alltoallv for the ring topology network. this is using the `all-to-all` connection rule  and hence should scale perfectly. the log data shows similar number of spikes per process for all network sizes, so i'd assume the code does the correct thing. any ideas, maybe NEST3 related? @suku248 @stinebuu @hakonsbm here's the code used for the benchmark: https://pastebin.com/5DeNi1th 

JUWELS_5g_gap_git38f8a55_5g_gap_git9b3bdf9_bympiprocesses
JUWELS_5g_gap_ring_git38f8a55_5g_gap_ring_git9b3bdf9_bympiprocesses

stinebuu:
@jakobj thanks for the benchmarks, I can look into it. We already know of some problems with spatial networks for NEST 3.0, so it might be related to this. I don't suppose you also have the results for NEST 2.20?

nestkernel/nest_names.cpp
```diff
@@ -73,7 +73,8 @@ const Name b( "b" );
 const Name beta( "beta" );
 const Name beta_Ca( "beta_Ca" );
 const Name buffer_size( "buffer_size" );
-const Name buffer_size_secondary_events( "buffer_size_secondary_events" );
+const Name send_buffer_size_secondary_events( "send_buffer_size_secondary_events" );
+const Name recv_buffer_size_secondary_events( "recv_buffer_size_secondary_events" );
```
stinebuu, line 77: `send_buffer_size_secondary_events` should be under s and `recv_buffer_size_secondary_events` should be under r.

jougs:
@jakobj can you please update this and/or comment on the state of this? Thanks!

jakobj:
@jougs  i've updated with the most recent master and fixed the conflicts

jougs:
@H27CK: would you be available to have another look at this? Many thanks!

H27CK:
Currently on vacation until 21/09. Will review on my return.

nestkernel/nest_names.cpp
```diff
@@ -77,7 +77,6 @@ const Name beta( "beta" );
 const Name beta_Ca( "beta_Ca" );
 const Name box( "box" );
 const Name buffer_size( "buffer_size" );
-const Name buffer_size_secondary_events( "buffer_size_secondary_events" );
 const Name buffer_size_spike_data( "buffer_size_spike_data" );
```
stinebuu, line 80: `buffer_size_secondary_events` must also be removed in `nest_names.h` I think this might be why Travis is currently failing :)

stinebuu (changes requested):
Thanks, I'm good with this now, but there is still a tiny problem with `nest_names`. I think this is why Travis is failing.

nestkernel/nest_names.cpp
H27CK: Can we confirm that this is the reason for the failed checks?

jougs:
The rule is that each of the names declared in `nest_names.h` has to be used somewhere throughout the code and defined in `nest_names.cpp`. The files should be sorted alphabetically. You can run the checker locally by invoking `NEST_SOURCE=. python3 extras/check_unused_names.py` in your source directory.

H27CK:
I am happy to approve this once CI checks have passed.

stinebuu:
@jakobj Could you remove the line in `nest_names.h` so we can merge this?

jakobj:
thanks for the reminder. done.

stinebuu (approved):
Thanks! @H27CK Could you take another look?

stinebuu:
@jakobj could you merge master so we get a green check mark? There was a problem with Travis last time.

@H27CK could you have another look?

nestkernel/source_table.cpp
```diff
@@ -243,18 +243,22 @@ nest::SourceTable::compute_buffer_pos_for_unique_secondary_sources( const thread
...
-    std::vector< long > max_uint_count(
-      1, *std::max_element( recv_buffer_position_by_rank.begin(), recv_buffer_position_by_rank.end() ) );
-    kernel().mpi_manager.communicate_Allreduce_max_in_place( max_uint_count );
-    kernel().mpi_manager.set_chunk_size_secondary_events_in_int( max_uint_count[ 0 ] + 1 );
+    // each chunk needs to contain one additional int that can be used
+    // to communicate whether waveform relaxation has converged
+    for ( thread rank = 0; rank < kernel().mpi_manager.get_num_processes(); ++rank )
+    {
+      recv_counts_secondary_events_in_int_per_rank[ rank ] += 1;
+    }
```
hakonsbm, line 258: I think a range-based loop is more readable here.
```suggestion
    for ( auto& recv_count : recv_counts_secondary_events_in_int_per_rank )
    {
      ++recv_count;
    }
```

hakonsbm (changes requested):
Thanks, @jakobj, this looks pretty good!

I think we can at least cut `in_int` from the names to shorten them a bit. Maybe also try to abbreviate some of the words, like @stinebuu suggested:
> Maybe it could be `get_recv_displacement_secondary_events` or even `get_recv_displacement_sec_events`?

jakobj:
> Thanks, @jakobj, this looks pretty good!
> 
> I think we can at least cut `in_int` from the names to shorten them a bit. Maybe also try to abbreviate some of the words, like @stinebuu suggested:
> 
> > Maybe it could be `get_recv_displacement_secondary_events` or even `get_recv_displacement_sec_events`?

dropping the `in_int` would lead to confusion i think, for example both `send_recv_count_spike_data_in_int_per_rank` and `send_recv_count_spike_data_per_rank_` exist, and the former is the latter multiplied by `sizeof( SpikeData ) / sizeof( unsigned int )`. we can certainly replace `secondary` with sec, but this should then be changed across the whole code base. i would leave this to another PR.

nestkernel/mpi_manager.h
```diff
@@ -261,6 +268,17 @@ class MPIManager : public ManagerInterface
...
   bool adaptive_spike_buffers() const;
 
+  void set_recv_counts_secondary_events_in_int_per_rank( const std::vector< int >& recv_counts_in_int_per_rank );
+
+  size_t get_recv_count_secondary_events_in_int( const size_t source_rank ) const;
+  size_t get_recv_displacement_secondary_events_in_int( const size_t source_rank ) const;
+  size_t get_send_count_secondary_events_in_int( const size_t target_rank ) const;
+  size_t get_send_displacement_secondary_events_in_int( const size_t target_rank ) const;
+  size_t get_done_marker_position_in_secondary_events_send_buffer( const size_t target_rank ) const;
+  size_t get_done_marker_position_in_secondary_events_recv_buffer( const size_t source_rank ) const;
```
hakonsbm, line 278: Add some documentation to these.

nestkernel/mpi_manager.h
```diff
@@ -294,6 +309,12 @@ class MPIManager : public ManagerInterface
   unsigned int send_recv_count_spike_data_per_rank_;
   unsigned int send_recv_count_target_data_per_rank_;
 
+  std::vector< int > recv_counts_secondary_events_in_int_per_rank_;
+  std::vector< int > send_counts_secondary_events_in_int_per_rank_;
+
+  std::vector< int > recv_displacements_secondary_events_in_int_per_rank_;
+  std::vector< int > send_displacements_secondary_events_in_int_per_rank_;
```
hakonsbm, line 316: Add some documentation to what these refer to.

hakonsbm (changes requested):
@jakobj So the `in_int` always refers to that the events are counted in units of `unsigned int`, not that the variables are of type `vector<int>`? I think some Doxygen comments should be added to clear up the confusion and more clearly explain the intent.

I see that there are a few other places as well where the naming scheme could be improved, so better to do that another time then.

jakobj:
> @jakobj So the `in_int` always refers to that the events are counted in units of `unsigned int`, not that the variables are of type `vector<int>`? I think some Doxygen comments should be added to clear up the confusion and more clearly explain the intent.

exactly, all elements/events are counted in ints. this way was easier than introducing a custom MPI type, although that may be an idea for the future. i've added some comments now, please have another look.

nestkernel/mpi_manager.h
```diff
@@ -261,6 +268,49 @@ class MPIManager : public ManagerInterface
...
+
+  /**
+   * Returns the recvcounts parameter of Alltoallv for communication of
+   * secondary events, i.e., the number of elements (in ints) to recv
+   * from `source_rank`.
+   */
+  size_t get_recv_count_secondary_events_in_int( const size_t source_rank ) const;
+
+  /**
+   * Returns the rispls parameter of Alltoallv for communication of
```
hakonsbm, line 286: Is `rispls` a typo? Looks like it should be
```suggestion
   * Returns the recvdisplacements parameter of Alltoallv for communication of
```

nestkernel/mpi_manager.h
```diff
@@ -294,6 +341,18 @@ class MPIManager : public ManagerInterface
...
+                                                                    //received from each rank
+  std::vector< int >
+    send_counts_secondary_events_in_int_per_rank_; //!< how many secondary elements (in ints) will be sent to each rank
+
+  std::vector< int > recv_displacements_secondary_events_in_int_per_rank_; //!< offset in the MPI receive buffer (in
+                                                                           //ints) at which elements received from each
+                                                                           //rank will be written
+  std::vector< int > send_displacements_secondary_events_in_int_per_rank_; //!< offset in the MPI send buffer (in ints)
+                                                                           //from which elements send to each rank will
+                                                                           //be read
```
hakonsbm, line 354: For Doxygen to see all documentation lines, they all need to start with `!<`.
```suggestion
  std::vector< int > recv_counts_secondary_events_in_int_per_rank_; //!< how many secondary elements (in ints) will be
                                                                    //!< received from each rank
  std::vector< int >
    send_counts_secondary_events_in_int_per_rank_; //!< how many secondary elements (in ints) will be sent to each rank

  std::vector< int > recv_displacements_secondary_events_in_int_per_rank_; //!< offset in the MPI receive buffer (in
                                                                           //!< ints) at which elements received from each
                                                                           //!< rank will be written
  std::vector< int > send_displacements_secondary_events_in_int_per_rank_; //!< offset in the MPI send buffer (in ints)
                                                                           //!< from which elements send to each rank will
                                                                           //!< be read
```

hakonsbm (changes requested):
@jakobj Thanks for the comments! Just a couple small things left and I'm happy.

nestkernel/mpi_manager.h
jakobj: indeed a typo, should be `rdispls` (https://www.open-mpi.org/doc/v4.0/man3/MPI_Alltoallv.3.php)

jakobj:
> @jakobj Thanks for the comments! Just a couple small things left and I'm happy.

thx, fixed.

hakonsbm (approved):

stinebuu:
Thanks, @jakobj! Sorry for the long process of getting this merged!

PR #1384, opened 2020-01-10
Make MPI errors fatal

heplesser:
Currently, NEST does not systematically check the error status of MPI function calls, so NEST might proceed past MPI errors.

This PR ensures that MPI will treat any errors occurring in MPI routines as fatal.


heplesser:
@hakonsbm Good point! Could you edit your comment to add a link to the source for the text from the standard?

Should we keep this PR to make the default explicit or just drop it?

jougs:
@terhorstd and me came to the conclusion that we'd rather close this for now.

PR #1385, opened 2020-01-10
Remove handling of external POSIX signals sent to NEST.

### LINKED ISSUE #555: How should NEST handle external signals?
At present, NEST can handle signals `SIGINT`, `SIGUSR1` and `SIGUSR2`, provided that NEST is compiled without MPI. If such a signal is received during the simulation update loop, the loop is exited at the end of the next min-delay interval.

Issues
- The interrupt mechanism appears to use some very ancient constructs, see `sli/psignal.*`
- It is not documented nor clear that just the three given signals are supported
- To be useful, signal handling should probably be available under MPI as well
- How to handle remaining simulation time after and interrupt, see #554

heplesser:
This PR fixes #555 by removing the capability to handle POSIX signals sent to NEST. It also removes the `SLIsignalflag` variable. The only side effect is that if `sli-init.sli` is not found in `SLIStartup::init()`, we now exit directly by calling `exit()`, which will terminate Python if we are importing PyNEST; in NEST2, Python shows a rather meaningless error message complaining about an empty stack.


heplesser:
@gtrensch Could you check if this change allows NEST to be compiled under Windows?

Issue #571, opened 2016-11-30
Structural plasticity delivers spikes via connections that were not present at the time of the spike


jakobj:
The `update_` function in NEST contains the following while loop (severely simplified, see https://github.com/nest/nest-simulator/blob/master/nestkernel/simulation_manager.cpp#L588 ff):
```c++
    do
    {
      if ( kernel().sp_manager.is_structural_plasticity_enabled()
        && ( clock_.get_steps() + from_step_ )
            % kernel().sp_manager.get_structural_plasticity_update_interval()
          == 0 )
      {
        [...] // update of structural plasticity
      }

      if ( from_step_ == 0 ) // deliver only at beginning of slice
      {
        kernel().event_delivery_manager.deliver_events( thrd );
        [...]
      }
      [...]

#pragma omp master
      {
        // gather only at end of slice
        if ( to_step_ == kernel().connection_manager.get_min_delay() )
          kernel().event_delivery_manager.gather_events( true );
        [...]
      }
      [...]
    } while ( ( to_do_ != 0 ) && ( !terminate_ ) );
```

As one can deduce from the comments, spikes are always "gathered", i.e., communicated among all processes at the end of a slice, while they are delivered at the beginning of the next slide. However, due to the order of operations in the while loop, structural plasticity can create connections that were not present when spikes were gathered in the previous step. This can lead to the following situation in two subsequent slices:
1. neurons 1 and 2 are unconnected
2. neuron 1 spikes
3. structural plasticity creates a connection from neuron 1 to neuron 2
4. a spike is delivered to neuron 2

At least for me, this is quite counterintuitive, since (for some reason) I expect spikes only to be delivered via connections *that were present at the time of the spike*. However, my intuition might be wrong, so I would like to discuss this issue.
In case we agree this is undesired behaviour, an easy fix (that I have already prepare locally) is to move `deliver` above the structural plasticity update.
@sdiazpier @suku248 @heplesser, please have a look at this.

heplesser:
[...]

PR #1395, opened 2020-01-14
Changed EnableStructuralPlasticity documentation to address issue #571

sdiazpier:
This PR addresses issue #571 by adding additional documentation to the EnableStructuralPlasticity function about the possibility of having spikes delivered via connections that were not present at the time of the spike. Further non-trivial work would need to be done in order to completely remove this possibility.

### DIFF
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -1623,6 +1623,9 @@ NestModule::GetStructuralPlasticityStatus_DFunction::execute( SLIInterpreter* i 
 /**
  * Enable Structural Plasticity within the simulation. This means, allowing
  * dynamic rewiring of the network based on mean electrical activity.
+ * Please note that in the current implementation of structural plasticity
+ * spikes could be delivered via connections that were not present at the 
+ * time of the spike in some cases.
  * @param i
  */
 void
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -1621,11 +1621,11 @@ NestModule::GetStructuralPlasticityStatus_DFunction::execute( SLIInterpreter* i 
 }
 
 /**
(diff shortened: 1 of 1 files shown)


nestkernel/nestmodule.cpp
```diff
@@ -1623,6 +1623,9 @@ NestModule::GetStructuralPlasticityStatus_DFunction::execute( SLIInterpreter* i
 /**
  * Enable Structural Plasticity within the simulation. This means, allowing
  * dynamic rewiring of the network based on mean electrical activity.
+ * Please note that in the current implementation of structural plasticity
+ * spikes could be delivered via connections that were not present at the 
+ * time of the spike in some cases.
```
sarakonradi: Perhaps it would be clearer if you wrote:

`Please note that, in the current implementation of structural plasticity, spikes could occasionally be delivered via connections that were not present at the time of the spike.`

What do you think?

nestkernel/nestmodule.cpp
```diff
@@ -1623,6 +1623,9 @@ NestModule::GetStructuralPlasticityStatus_DFunction::execute( SLIInterpreter* i
 /**
  * Enable Structural Plasticity within the simulation. This means, allowing
  * dynamic rewiring of the network based on mean electrical activity.
```
sarakonradi: `This allows dynamic rewiring of the network based on mean electrical activity.`

nestkernel/nestmodule.cpp
sdiazpier: Great! thanks so much for checking this @sarakonradi

nestkernel/nestmodule.cpp
sdiazpier: Thanks for your suggestion @sarakonradi

sdiazpier:
Thanks @jakobj and @sarakonradi for your comments. I have integrated them now.

PR #1397, opened 2020-01-15
Small typographical updates to user documentation

clinssen:


heplesser:
Minimal typographical corrections to documentation, therefore merging without further review.

PR #1398, opened 2020-01-15
Document purpose of multiplicity-unrolling code in spike delivery

### LINKED ISSUE #126: Create MPI-test for plasticity with pre-synaptic neurons using spike multiplicity
Several neuron models, especially `pp_*` models, use `set_multiplicity()` to convey that several spikes have been fired within a single time step. Since these models have proxies, their spikes are conveyed via `Network::send_remote()`, which converts a `SpikeEvent` with multiplicity n into n  `SpikeEvent`s of multiplicity 1. Thus, it does not matter that `STDPConnection::send()` and other plastic synapses do not heed event multiplicity.

Now, it is not inconceivable that this behavior will change in the future, i.e., that event multiplicity is transmitted. Then, plasticity would not be handled correctly. We should therefore do the following:
- add `assert(e.get_multiplicity() == 1)` to the `send()` methods of all plastic synapse models
- add an MPI test that checks that spikes sent with > 1 multiplicity are handled properly wrt plasticity.

heplesser:
This PR adds developer documentation warning against removing multiplicity unrolling when communicating non-device spikes, as this could invalidate our plasticity handling.

This fixes #126.


heplesser:
@jougs But the comment will not hurt, will it? I wasn't sure if your comment was an :+1: or a :-1: on this PR.

jougs:
Sorry for being unclear. My comment was rather a :-1:, but I wanted to discuss before closing the PR.

I think that the comment is not adding any value and points out something rather obvious: all lines in the code are there for a reason. The same comment (i.e. "do not remove the line below, it does something useful") could be added to each and every line in NEST. I can't see why this case is special.

I'm all in for adding a simple comment just saying *why* the line is there, i.e. "the following loop is unrolling the multiplicity contained in the event".

jougs:
@heplesser: I did not want to scare you away ;-) Can we please discuss this and come to a conclusion? Thanks!

heplesser:
@jougs I have now minimized the comments to straightforward explanations of what the loops do (and why).

jougs:
@hakonsbm: can you please indicate if you still approve? Thanks!

PR #1403, opened 2020-01-17
More informative compiler information in CMake summary

### LINKED ISSUE #1393: Make compiler information in NEST configuration summary more informative
Currently, the configuration summary provided at the end of a CMake run may tell as little as
```
C++ compiler        : /usr/bin/c++
```
about the compiler. In addition to the absolute path to the compiler, the summary should also report the actual compiler and its version, e.g. something like
```
The CXX compiler identification is GNU 8.3.0
```
This information is currently found in the CMake output, but requires scrolling back from the config report.

hakonsbm:
Adds a more informative message on compiler identification and version in the CMake summary.

Example of current message:
```
C compiler          : /usr/bin/gcc-8
C++ compiler        : /usr/bin/g++-8
```
Example of new message:
```
C compiler          : GNU 8.3.0 (/usr/bin/gcc-8)
C++ compiler        : GNU 8.0.1 (/usr/bin/g++-8)
```

Fixes #1393.


ikitayama:
At least is this applicable to Clang as well?
What if we use other vendors compilers?

hakonsbm:
@ikitayama It should apply to any compiler. The summary uses the CMake variable `CMAKE_C_COMPILER_ID`, which on my computer is `GNU` if compiling with GCC, and `Clang` if compiling with Clang.

ikitayama:
Hakon,
Thanks for exposing this to me!

heplesser:
Minimal change to CMake report, merging without second review.

ikitayama:
I just checked the current master with FCC, it wasn't that descriptive like GCC, unfortunately.
But always cosmetics can be added later.

PR #1404, opened 2020-01-17
Refactor testsuite and move from nose to Pytest

### LINKED ISSUE #345: do_tests should get all arguments from the commandline

### LINKED ISSUE #700: Suppress all NEST startup output

### LINKED ISSUE #972: `make installcheck` reaches wrong testing code after sourcing `nest_vars.sh` for a different version

### LINKED ISSUE #1130: Phase 5 Tests all fail with ' Open MPI does not support recursive calls of mpirun'

jougs:
This PR refactors the test runner scripts of NEST and PyNEST and makes them independent of configuration-time variables previously replaced into the scripts.

Changes in detail:
* `do_tests.sh` now gets all paths and executables on the commandline (fixes #345).
* `nest_vars.sh` does not export paths anymore, that could have confused NEST(fixes #972).
   Now a separate PR: #1578
* Tests are now run in parallel with as many processors as are available (kudos to @terhorstd).
* Test results from SLI tests are not written as JUnit/XML anymore.
* The info output from the `RandomManager` initialization is gone (fixes #700 and #415).
   Now a separate PR: #1577
* The PyNEST testsuite was
   * moved outside of the `nest` module directory
   * converted to using `pytest` instead of `nosetests`
   * modified so that PyNEST tests do not include MPI anymore (fixes #1130).
* Increased legibility by adding an informative header and removing redundant paths from the testsuite printouts.

---
This replaces #1071, but with base branch nest-3 instead of master.


heplesser:
@jougs Great work! The test output already looks so much better, and the configuration information at the beginning of the test execution will allow us to support users experiencing failing test much better. I have a few questions and comments:

1. If I understand the test output on Travis right, where up to 2 tests are run in parallel, then there are two output lines for each test, one "Running test_xyz" and one "test_xyz ... Success", but the order of lines for different tests is somewhat unpredictable due to parallel execution. Do we really need the first ("Running ...") line?
1. If tests should fail, will the summary at then end report which tests failed, so one no longer would have to scroll back and search for the failing test?
1. Could we make the maximum number of tests run in parallel configurable, e.g., through a variable in `.nestrc`? This may be useful when running the testsuite on a system used for interactive work.

hakonsbm:
@jougs Nice, this looks a lot better! I have a few remarks though:

1. Running the testsuite on my computer reports 7 failing PyNEST tests. However it is not clear which tests are failing (by process of elimination it should be the seven listed PyNEST MPI tests, but this isn't stated).
1. Should every PyNEST test be listed (instead of dots), to keep output consistent with the SLI tests?
1. Regarding @heplesser's second point, failing tests are not reported at the end of the testsuite. This would be a useful addition.
1. The C++ test phase isn't run. I have compiled with Boost, and the `run_all_cpptests` file is created, but it says `Not running C++ tests because NEST was compiled without Boost`.
1. The todos in the C++ test phase should probably be fixed. I have some work on this in a stash, so I can fix the todos and make a PR towards your branch.

jougs:
@heplesser: Thanks for the praise and your suggestions. I'm very open to implement them.

First, I'd like to get it working, though! Could you maybe help out with the problems on Mac? Apparently @terhorstd's solution for getting the number of available processors via `/proc` and system information using `uname` is not very portable.

lekshmideepu:
@jougs  :  sysctl -a | grep machdep.cpu

Silmathoron:
Just noticed the commit text and came to provide a bit of psychological support!
Cheers <3
(maybe take a break and then come back to it later? ^^)

jougs:
# :heart_eyes:
Thank you so much! This is really a total mess. Everything's running fine locally, while it fails on Travis. And the debugging sessions I can get there are way too short to hunt the problem down. Argh!

jougs:
I'm closing this, as this got way too messy. The code was/will be made into multiple smaller PRs.

jougs:
For a detailed list of which issue was fixed in which PR, see https://github.com/nest/nest-simulator/pull/1071#issuecomment-762513312

PR #1407, opened 2020-01-21
Refactor beta normalization factor calculation into function

### LINKED ISSUE #1199: Refactor beta-function normalization
Once #1197 is merged, we will have code computing the normalization constant for the beta  function in at least three places:
- `aeif_cond_beta_multisynapse::calibrate()`
- `ht_neuron::get_synapse_constant()` (best documentation)
- `iaf_cond_beta::get_normalization_factor()`

We should refactor this to a single `nestutils/beta_normalization_factor.h` function.

hakonsbm:
This PR refactors out calculation of the beta normalization factor from models that each have their own implementation of it.

Fixes #1199.


models/ht_neuron.cpp
```diff
@@ -703,34 +706,7 @@ nest::ht_neuron::init_buffers_()
 double
 nest::ht_neuron::get_synapse_constant( double tau_1, double tau_2, double g_peak )
 {
-  /* The solution to the beta function ODE obtained by the solver is
```
clinssen: Can this comment block be kept and moved to `beta_normalization_factor.h`?

clinssen:
I was looking at the "beta" function in NESTML as well, and found that this function, as formulated in the code and in the two provided references, effectively enforces tau_decay > tau_rise. You can set the taus so as to violate this inequality, but they will be implicitly swapped. I cannot seem to obtain dynamics that would correspond to tau_decay < tau_rise (slow rise, fast decay). Unless I'm missing something obvious, should this be mentioned in the documentation?

Silmathoron:
@clinssen I never thought about that... but I guess if it decays faster than it rises, then it does not look like a beta function at all ^^
But it would not hurt to say it in the documentation

PR #1416, opened 2020-02-06
Fix misuse of outdated variable that led to an impossible condition

jakobj:
This removes a local counter used as an exit condition for a while true loop that was never increased. It is replaced by using the correct functions of send_buffer_positions. While this should not lead to any bugs, it potentially slows down construction of the presynaptic infrastructure quite severely.

Discovered during a code review session with @suku248 @diesmann @heplesser


heplesser:
Reviewed in VC by @suku248, @terhorstd, @heplesser , thus merging without further review.

PR #1419, opened 2020-02-09
Corrects documentation of the stdp_dopamine_synapse model

### LINKED ISSUE #1253: [stdp_dopamine_synapse] A_plus and A_minus documentation
Hi everyone,

The documention for stdp_dopamine_synapse describes the parameters A_plus and A_minus as:

>   A_plus   double	- Amplitude of weight change for **facilitation**  
  A_minus   double	- Amplitude of weight change for **depression**  

Technically that is not true, because this ultimately depends on the value of n-b (actual concentration - baseline) . If n-b is negative, A_plus will control depression and A_minus facilitation.
The reason why that is the case is clear when we examine the equations in section 2.2 of [Potjans et al, 2010](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2996144/). But we can also see that this is in fact NEST's behaviour in the example bellow; it shows that A_plus can cause facilitation or depression depending solely on the baseline value.

My understanding is that A_plus controls weight changes driven by pre-before-post spike pairings and A_minus does the same for post_before_pre. So maybe a bit more accurate description would be be:

>   A_plus   double	- Amplitude of weight change for pre-before-post spike pairings
  A_minus   double	- Amplitude of weight change for post-before-pre spike pairings

``` python
import nest
import numpy as np

def mwe(baseline_shift=0.):
    # timing parameters
    dt = .1
    exp_duration = 20000.

    # kernel configuration
    nest.ResetKernel()
    nest.set_verbosity(20)
    nest.SetKernelStatus({'resolution' : dt})

    # dopamine setup
    DA_spike_times = np.round(np.arange(dt, exp_duration, dt), decimals=1)
    DA_spikes = nest.Create('spike_generator', params={'spike_times' : DA_spike_times})
    vt = nest.Create('volume_transmitter')
    nest.Connect(DA_spikes, vt, syn_spec={'delay' : dt})
    nest.CopyModel('stdp_dopamine_synapse', 'dopa_syn', params={
        'A_minus' : 0., 
        'b' : 1. / dt + baseline_shift,
        'vt' : vt[0],
    })
    
    # neurons
    neuron_spike = nest.Create('spike_generator', params={'spike_times' : [10000.]})
    neurons = nest.Create('parrot_neuron', 2)
    nest.Connect(neuron_spike, neurons[:1])
    nest.Connect(neurons[:1], neurons[1:], syn_spec='dopa_syn')
    synapse = nest.GetConnections(neurons[:1], neurons[1:])
    
    # Simulate 
    print('\nBaseline shift:', baseline_shift)
    print(f'Weight before simulation: {nest.GetStatus(synapse, "weight")[0]:.4f}')
    nest.Simulate(exp_duration)
    print(f'Weight after simulation: {nest.GetStatus(synapse, "weight")[0]:.4f}')

mwe(baseline_shift=0.000)
mwe(baseline_shift=0.001)
mwe(baseline_shift=-.001)
```

For which the output is:

```
Baseline shift: 0.0
Weight before simulation: 1.0000
Weight after simulation: 1.0000

Baseline shift: 0.001
Weight before simulation: 1.0000
Weight after simulation: 0.0952

Baseline shift: -0.001
Weight before simulation: 1.0000
Weight after simulation: 1.9048
```

(Note: I set A_minus to zero just to make clear that the weight changes we observe are dependent only on A_plus. The results are the same regardless of A_minus because there is just a pre-before-post event.)

gui-miotto:
closes #1253

### DIFF
--- a/models/stdp_dopa_connection.h
+++ b/models/stdp_dopa_connection.h
@@ -74,9 +74,28 @@ Parameters:
                    transmitting the spikes to the synapse. A value of
                    -1 indicates that no volume transmitter has been
                    assigned.
- A_plus    real    Amplitude of weight change for facilitation
- A_minus   real    Amplitude of weight change for depression
- tau_plus  ms      STDP time constant for facilitation
+ A_plus    real    Multiplier applied to weight changes caused by
+                   pre-before-post spike pairings. If b (dopamine
+                   baseline concentration) is zero, then A_plus
+                   is simply the multiplier for facilitation (as in the
+                   stdp_synapse model). However, if b is not zero, then
+                   A_plus will be the multiplier for facilitation only
+                   if n - b is positive, where n is the instantenous
+                   dopamine concentration in the volume transmitter.
+                   Conversely, if n - b is negative, A_plus will be
+                   the multiplier for depression.
+ A_minus   real    Multiplier applied to weight changes caused by
+                   post-before-pre spike pairings. If b (dopamine
+                   baseline concentration) is zero, then A_minus
+                   is simply the multiplier for depression (as in the
+                   stdp_synapse model). However, if b is not zero, then
+                   A_minus will be the multiplier for depression only
+                   if n - b is positive, where n is the instantenous
+                   dopamine concentration in the volume transmitter.
+                   Conversely, if n - b is negative, A_plus will be
+                   the multiplier for potentiation.
+ tau_plus  ms      STDP time constant for weight changes caused by
+                   pre-before-post spike pairings.
  tau_c     ms      Time constant of eligibility trace
  tau_n     ms      Time constant of dopaminergic trace
  b         real    Dopaminergic baseline concentration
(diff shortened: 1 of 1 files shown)


models/stdp_dopa_connection.h
```diff
@@ -74,9 +74,28 @@ dopaminergic dynamics is calculated in the synapse itself.
...
+                   baseline concentration) is zero, then A_minus
+                   is simply the multiplier for depression (as in the
+                   stdp_synapse model). However, if b is not zero, then
+                   A_minus will be the multiplier for depression only
+                   if n - b is positive, where n is the instantenous
+                   dopamine concentration in the volume transmitter.
+                   Conversely, if n - b is negative, A_minus will be
+                   the multiplier for potentiation.
+ tau_plus  ms      STDP time constant for weight changes caused by
+                   pre-before-post spike pairings.
```
sarakonradi: Looks good. I would suggest removing ``However`` and ``Conversely``, so that it reads better.

gui-miotto: Thanks for the review. Implemented.
I also just noticed that I used the word "potentiation" once. Although it is regarded as synonym of "facilitation", I changed it to "facilitation" for consistency.

sarakonradi: Sounds good. I agree with the change from "potentiation" to "facilitation".

PR #1421, opened 2020-02-10
Refactor SendBufferPosition, fail on query with rank outside of range

jakobj:
During code review with @suku248 we realized one could query the `SendBufferPosition` with ranks that are outside of the range for which a specific thread is responsible, returning nonsense values due to modulo arithmetic. This PR addresses that issue by adding two specific asserts.

Depends on #1416, should be merged after.

@hakonsbm @heplesser @jougs @stinebuu @jarsi

### DIFF
--- a/nestkernel/send_buffer_position.h
+++ b/nestkernel/send_buffer_position.h
@@ -41,14 +41,17 @@ namespace nest
 class SendBufferPosition
 {
 private:
+  thread begin_rank_;
+  thread end_rank_;
+  thread size_;
+  thread max_size_;
   size_t num_spike_data_written_;
-  std::vector< unsigned int > idx_;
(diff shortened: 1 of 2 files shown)


nestkernel/send_buffer_position.h
```diff
@@ -110,6 +114,8 @@ inline SendBufferPosition::SendBufferPosition( const AssignedRanks& assigned_ran
 inline thread
 SendBufferPosition::rank_to_index_( const thread rank ) const
 {
+  assert( begin_rank_ <= rank );
```
jakobj: these are the important asserts

heplesser:
Reviewed in VC by @suku248, @terhorstd, @heplesser, merging without further review.

PR #1422, opened 2020-02-10
Refactor/collocate target data

jakobj:
This PR refactors some code related to the collocation of buffers for the communication of connections. In particular the notorious `get_next_target_data` and `SourceTablePosition`. I'm happy about any suggestions to clean this up further, it's still a mess.

@suku248 @jarsi @heplesser

### DIFF
--- a/nestkernel/source_table.cpp
+++ b/nestkernel/source_table.cpp
@@ -290,34 +290,24 @@ bool
 nest::SourceTable::next_entry_has_same_source_( const SourceTablePosition& current_position,
   const Source& current_source ) const
 {
-  if ( ( current_position.lcid + 1
-           < static_cast< long >( sources_[ current_position.tid ][ current_position.syn_id ].size() )
-         and sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid + 1 ].get_node_id()
-           == current_source.get_node_id() ) )
-  {
-    return true;
-  }
+  const auto& local_sources = sources_[ current_position.tid ][ current_position.syn_id ];
+  const size_t next_lcid = current_position.lcid + 1;
 
-  return false;
+  return ( next_lcid < local_sources.size()
+	   and local_sources[ next_lcid ].get_node_id() == current_source.get_node_id() );
 }
 
 bool
 nest::SourceTable::previous_entry_has_same_source_( const SourceTablePosition& current_position,
   const Source& current_source ) const
 {
-  // decrease the position without returning a TargetData if the
-  // entry preceding this entry has the same source, but only if
-  // the preceding entry was not processed yet
-  if ( ( current_position.lcid - 1 >= 0 )
-    and ( sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid - 1 ].get_node_id()
-          == current_source.get_node_id() )
-    and ( not sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid - 1 ]
-                .is_processed() ) )
-  {
-    return true;
-  }
+  const auto& local_sources = sources_[ current_position.tid ][ current_position.syn_id ];
+  const long previous_lcid = current_position.lcid - 1;  // needs to be a signed type such that negative
+                                                         // values can signal invalid indices
 
-  return false;
+  return ( previous_lcid >= 0
+	   and not local_sources[ previous_lcid ].is_processed()
+	   and local_sources[ previous_lcid ].get_node_id() == current_source.get_node_id() );
 }
 
 bool
--- a/nestkernel/source_table.cpp
+++ b/nestkernel/source_table.cpp
@@ -274,16 +274,17 @@ nest::SourceTable::resize_sources( const thread tid )
 }
 
 bool
-nest::SourceTable::source_should_be_processed_( const thread rank_start, const thread rank_end, const Source& source ) const
+nest::SourceTable::source_should_be_processed_( const thread rank_start,
+  const thread rank_end,
+  const Source& source ) const
 {
   const thread source_rank = kernel().mpi_manager.get_process_id_of_node_id( source.get_node_id() );
 
-  return not ( source.is_processed()
-	       or source.is_disabled()
-	       // is this thread responsible for this part of the MPI
-	       // buffer?
-	       or source_rank < rank_start
-	       or rank_end <= source_rank );
+  return not( source.is_processed() or source.is_disabled()
+           // is this thread responsible for this part of the MPI
+           // buffer?
+           or source_rank < rank_start
+           or rank_end <= source_rank );
 }
 
 bool
@@ -293,8 +294,8 @@ nest::SourceTable::next_entry_has_same_source_( const SourceTablePosition& curre
   const auto& local_sources = sources_[ current_position.tid ][ current_position.syn_id ];
   const size_t next_lcid = current_position.lcid + 1;
 
-  return ( next_lcid < local_sources.size()
-	   and local_sources[ next_lcid ].get_node_id() == current_source.get_node_id() );
+  return (
+    next_lcid < local_sources.size() and local_sources[ next_lcid ].get_node_id() == current_source.get_node_id() );
 }
 
 bool
@@ -302,12 +303,11 @@ nest::SourceTable::previous_entry_has_same_source_( const SourceTablePosition& c
   const Source& current_source ) const
 {
   const auto& local_sources = sources_[ current_position.tid ][ current_position.syn_id ];
-  const long previous_lcid = current_position.lcid - 1;  // needs to be a signed type such that negative
-                                                         // values can signal invalid indices
+  const long previous_lcid = current_position.lcid - 1; // needs to be a signed type such that negative
+                                                        // values can signal invalid indices
 
-  return ( previous_lcid >= 0
-	   and not local_sources[ previous_lcid ].is_processed()
-	   and local_sources[ previous_lcid ].get_node_id() == current_source.get_node_id() );
+  return ( previous_lcid >= 0 and not local_sources[ previous_lcid ].is_processed()
+    and local_sources[ previous_lcid ].get_node_id() == current_source.get_node_id() );
 }
 
 bool
@@ -347,8 +347,7 @@ nest::SourceTable::get_next_target_data( const thread tid,
       current_position.tid, current_position.syn_id, current_position.lcid,
       next_entry_has_same_source_( current_position, current_source ) );
 
-    // no need to communicate this entry if the previous entry has the
-    // same source
+    // no need to communicate this entry if the previous entry has the same source
     if ( previous_entry_has_same_source_( current_position, current_source ) )
     {
       current_source.set_processed( true ); // no need to look at this entry again
@@ -356,7 +355,8 @@ nest::SourceTable::get_next_target_data( const thread tid,
       continue;
     }
 
-    // reaching this means we found a valid TargetData
+    // reaching this means we found an entry that should be
+    // communicated via MPI, so we prepare to return the relevant data
 
     const auto node_id = current_source.get_node_id();
 
--- a/nestkernel/source_table.cpp
+++ b/nestkernel/source_table.cpp
@@ -274,6 +274,62 @@ nest::SourceTable::resize_sources( const thread tid )
 }
 
 bool
+nest::SourceTable::source_should_be_processed_( const thread rank_start, const thread rank_end, const Source& source )
+{
+  if ( source.is_processed() or source.is_disabled() )
+  {
+    // looks like we've processed this already, let's continue
+    return false;
+  }
+
+  const thread source_rank = kernel().mpi_manager.get_process_id_of_node_id( source.get_node_id() );
+
+  // determine whether this thread is responsible for this part of
+  // the MPI buffer; if not we just continue with the next iteration
+  // of the loop
+  if ( source_rank < rank_start or source_rank >= rank_end )
+  {
+    return false;
+  }
+
+  return true; // if we reach this we should process it
+}
+
+bool
+nest::SourceTable::next_entry_has_same_source( const SourceTablePosition& current_position,
+  const Source& current_source ) const
+{
+  if ( ( current_position.lcid + 1
+           < static_cast< long >( sources_[ current_position.tid ][ current_position.syn_id ].size() )
+         and sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid + 1 ].get_node_id()
+           == current_source.get_node_id() ) )
+  {
+    return true;
+  }
+
+  return false;
+}
+
+bool
+nest::SourceTable::previous_entry_has_same_source( const SourceTablePosition& current_position,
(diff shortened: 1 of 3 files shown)


nestkernel/source_table.cpp
```diff
@@ -273,6 +273,62 @@ nest::SourceTable::resize_sources( const thread tid )
...
+  {
+    return false;
+  }
+
+  return true; // if we reach this we should process it
+}
+
+bool
+nest::SourceTable::next_entry_has_same_source( const SourceTablePosition& current_position,
+  const Source& current_source ) const
```
heplesser: I would suggest to replace the entire code by

```c++
const auto &local_sources = sources_[ current_position.tid ][ current_position.syn_id ];
const size_t next_lcid = current_position.lcid + 1;

return ( next_lcid < local_sources.size() 
  and local_sources[ next_lcid ].get_node_id() == current_source.get_node_id() );
```

nestkernel/source_table.cpp
```diff
@@ -273,6 +273,62 @@ nest::SourceTable::resize_sources( const thread tid )
...
+  {
+    return true;
+  }
+
+  return false;
+}
+
+bool
+nest::SourceTable::previous_entry_has_same_source( const SourceTablePosition& current_position,
+  const Source& current_source ) const
```
heplesser: I would suggest to replace the entire code by

```c++
const auto prev_lcid = current_position.lcid - 1;

if ( prev_lcid < 0 )
{
  return false;
}
else
{
  const auto &prev_source = sources_[ current_position.tid ][ current_position.syn_id ][ prev_lcid ];
  return not prev_source.is_processed() and prev_source.get_node_id() == current_source.get_node_id();
}
```

nestkernel/source_table.cpp
```diff
@@ -296,21 +352,9 @@ nest::SourceTable::get_next_target_data( const thread tid,
     const Source& const_current_source =
```
heplesser: I would change the name to `current_source`, no need to have type information in the variable name. And I see no reason why to have a `const_current_source` here and then a `current_source` below.

nestkernel/source_table.cpp
```diff
@@ -320,67 +364,59 @@ nest::SourceTable::get_next_target_data( const thread tid,
     current_source.set_processed( true );
```
heplesser: I don't think we need this second `current_source` variable here, we have it from above (without const).

And shouldn't we mark it as processed first after we have processed it, not here where we are about to begin processing it?

nestkernel/source_table.cpp
```diff
@@ -320,67 +364,59 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
-    // entry, if existent, has the same source; start by assuming it
-    // has a different source, only change if necessary
-    kernel().connection_manager.set_has_source_subsequent_targets(
-      current_position.tid, current_position.syn_id, current_position.lcid, false );
-    if ( ( current_position.lcid + 1
-             < static_cast< long >( sources_[ current_position.tid ][ current_position.syn_id ].size() )
-           and sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid + 1 ].get_node_id()
-             == current_source.get_node_id() ) )
+    // entry, if existent, has the same source
+    kernel().connection_manager.set_has_source_subsequent_targets( current_position.tid,
```
heplesser: Shouldn't the setter be called `set_source_has_subsequent_targets` to be good English? And maybe `set_source_has_more targets` for legibility?

nestkernel/source_table.cpp
```diff
@@ -320,67 +364,59 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+      current_position.syn_id,
+      current_position.lcid,
+      next_entry_has_same_source( current_position, current_source ) );
+
+    // no need to communicate this entry if the previous entry has the
+    // same source
```
heplesser: Since we allow long lines now, I'd put this comment on one line.

nestkernel/source_table.cpp
```diff
@@ -320,67 +364,59 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
 
-    // decrease the position without returning a TargetData if the
-    // entry preceding this entry has the same source, but only if
-    // the preceding entry was not processed yet
-    if ( ( current_position.lcid - 1 >= 0 )
-      and ( sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid - 1 ].get_node_id()
-            == current_source.get_node_id() )
-      and ( not sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid - 1 ]
-                  .is_processed() ) )
+    // reaching this means we found a valid TargetData
```
heplesser: "valid" in what sense? Data  that we need to communicate?

nestkernel/source_table.cpp
```diff
@@ -320,67 +364,59 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
+
+    // set the source rank
+    source_rank = kernel().mpi_manager.get_process_id_of_node_id( current_source.get_node_id() );
+
+    // set values of next_target_data
+    next_target_data.set_source_lid( kernel().vp_manager.node_id_to_lid( current_source.get_node_id() ) );
+    next_target_data.set_source_tid(
+      kernel().vp_manager.vp_to_thread( kernel().vp_manager.suggest_vp_for_node_id( current_source.get_node_id() ) ) );
+    next_target_data.reset_marker();
```
heplesser: What kind of marker? Could we have a more verbose name?

nestkernel/source_table.cpp
```diff
@@ -320,67 +364,59 @@ nest::SourceTable::get_next_target_data( const thread tid,
...
-    // the preceding entry was not processed yet
-    if ( ( current_position.lcid - 1 >= 0 )
-      and ( sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid - 1 ].get_node_id()
-            == current_source.get_node_id() )
-      and ( not sources_[ current_position.tid ][ current_position.syn_id ][ current_position.lcid - 1 ]
-                  .is_processed() ) )
+    // reaching this means we found a valid TargetData
+
+    // set the source rank
+    source_rank = kernel().mpi_manager.get_process_id_of_node_id( current_source.get_node_id() );
```
heplesser: I'd introduce
```c++
const auto cs_gid =current_source.get_node_id();
```
to simplify the lines below.

nestkernel/source_table_position.h
```diff
@@ -41,14 +41,14 @@ struct SourceTablePosition
...
+
   SourceTablePosition();
   SourceTablePosition( const long tid, const long syn_id, const long lcid );
   SourceTablePosition( const SourceTablePosition& rhs );
 
-  template < typename T >
-  void wrap_position( const std::vector< std::vector< BlockVector< T > > >& sources );
-
-  bool is_at_end() const;
+  void seek_to_next_valid_index( const std::vector< std::vector< BlockVector< Source > > >& sources );
```
heplesser: Please add doxygen documentation for new methods.

nestkernel/source_table_position.h
```diff
@@ -97,15 +106,17 @@ SourceTablePosition::wrap_position( const std::vector< std::vector< BlockVector<
       continue;
     }
 
+    // if we can not find a valid lcid by decreasing synapse or thread
+    // indices, we have read all entries
     assert( tid < 0 );
     assert( syn_id < 0 );
     assert( lcid < 0 );
-    return;
+    return; // reached the end
```
heplesser: I would comment here `reached the end without finding a valid lcid` and then have a (if unnecessary) `return` outside the while loop with a comment `found a valid lcid`

nestkernel/source_table_position.h
```diff
@@ -97,15 +106,17 @@ SourceTablePosition::wrap_position( const std::vector< std::vector< BlockVector<
...
   }
 }
 
 inline bool
-SourceTablePosition::is_at_end() const
+SourceTablePosition::is_invalid() const
 {
```
heplesser: Simplify to
```c++
return tid < 0 and syn_id < 0 and lcid < 0;
```

nestkernel/source_table_position.h
```diff
@@ -117,6 +128,12 @@ SourceTablePosition::is_at_end() const
   }
 }
 
+inline void
+SourceTablePosition::decrease()
+{
+  --lcid;
```
heplesser: I think `SourceTablePosition` should guarantee that `lcid >= -1`, so that in `SourceTable`, we can be sure that `x.lcid + 1 >= 0`. So I'd suggest to add `assert( lcid >= -1 )` after the reduction.

nestkernel/source_table_position.h
```diff
@@ -97,15 +106,17 @@ SourceTablePosition::wrap_position( const std::vector< std::vector< BlockVector<
       continue;
     }
 
+    // if we can not find a valid lcid by decreasing synapse or thread
+    // indices, we have read all entries
     assert( tid < 0 );
     assert( syn_id < 0 );
     assert( lcid < 0 );
```
heplesser: I would tighten the last assert to `lcid == -1`, ensuring that `lcid + 1 >= 0` always.

nestkernel/source_table.h
```diff
@@ -129,6 +129,13 @@ class SourceTable
    */
   bool is_cleared() const;
 
+  bool source_should_be_processed_( const thread rank_start, const thread rank_end, const Source& source );
```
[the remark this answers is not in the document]
jakobj: yes, it should be const, added. added comments. yes all are only for internal use and are now private.

nestkernel/source_table.cpp
```diff
@@ -273,6 +273,62 @@ nest::SourceTable::resize_sources( const thread tid )
   sources_[ tid ].resize( kernel().model_manager.get_num_synapse_prototypes() );
 }
 
+bool
+nest::SourceTable::source_should_be_processed_( const thread rank_start, const thread rank_end, const Source& source )
```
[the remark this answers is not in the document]
jakobj: sounds good.

nestkernel/source_table.cpp
```diff
@@ -273,6 +273,62 @@ nest::SourceTable::resize_sources( const thread tid )
...
+{
+  if ( source.is_processed() or source.is_disabled() )
+  {
+    // looks like we've processed this already, let's continue
+    return false;
+  }
+
+  const thread source_rank = kernel().mpi_manager.get_process_id_of_node_id( source.get_node_id() );
```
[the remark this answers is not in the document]
jakobj: it does not occur that often and for example in https://github.com/nest/nest-simulator/blob/master/nestkernel/source_table.cpp#L251 we need to be able to pass just the GID. so i would leave it as is

nestkernel/source_table.cpp
jakobj: nice!

nestkernel/source_table.cpp
jakobj: why not even more compact, like the next_entry function?

nestkernel/source_table.cpp
jakobj: because we lift the restriction of thread-private data structures the idea was to use `const` to tell the compiler that only read access will follow. might be a matter of premature optimization though, i don't have a strong opinion on it.

nestkernel/source_table.cpp
jakobj: i) see above
ii) yes, it should be marked as processed right before we return it, changed

nestkernel/source_table.cpp
jakobj: i've just added the setter prefix `set_` to the name of the getter `has_source_...` (https://github.com/nest/nest-simulator/blob/master/nestkernel/connection.h#L269) causing this terrible function name. i thought in terms of consistency between getter and setter this would justify the bad language, but i'd be willing to be overruled.

nestkernel/source_table.cpp
jakobj: ok

nestkernel/source_table.cpp
jakobj: yes, changed the comment accordingly

nestkernel/source_table.cpp
jakobj: the marker can signal different things, depending on it's value, e.g., whether more communication rounds are necessary, whether valid data resides in the buffer or whether subsequent entries should be skipped. i can not think of a more verbose name right now.

nestkernel/source_table.cpp
jakobj: ok

nestkernel/source_table_position.h
jakobj: ok

nestkernel/source_table_position.h
jakobj: ok

nestkernel/source_table_position.h
jakobj: nice!

nestkernel/source_table_position.h
jakobj: good point!

nestkernel/source_table_position.h
jakobj: shouldn't we do this for all indices?

jakobj:
thanks a lot for the helpful comments @heplesser, i've incorporated most of them, once the tests pass you can have another look

nestkernel/source_table.cpp
heplesser: For just the single use of the "const" version (on line 377 below), I think the duplicity of a const and a plain version are not worth the extra code complexity.

nestkernel/source_table.cpp
heplesser: I think `source_has_more_targets` would be a much better getter name as well.

heplesser:
@jakobj Something very interesting has suddenly happened: MPI tests now fail with 
```
nest: /home/travis/build/nest/nest-simulator/nestkernel/source_table_position.h:123: void nest::SourceTablePosition::seek_to_next_valid_index(const std::vector<std::vector<BlockVector<nest::Source> > >&): Assertion `tid == -1' failed.
```
Previously, in that location we had `assert( tid < 0 )` and that passed. Doesn't this mean that we seem to have a situation where `tid < -1`? As far as I can see right now, this could happen (only) if we enter `seek_to_...` with `tid == -1`. But should this be possible at all? And if we entered with `tid  == -1`, should we then seek for anything?

jakobj:
that is surprising indeed! i can not reproduce it outside the testsuite though, running `nest test_connect_array_all_to_all_mpi.sli` exits normally so does the same command with `mpirun`. i there a special trick to run the mpi tests outside the testsuite?

jakobj:
thanks @heplesser for the comments, i've integrated them now. the MPI tests failed because one could enter `get_next_target_data` with a position where all indices are -1, seeking then changed tid to -2. in case all indices are -1, the while(true) loop is now never entered. this should not change program behaviour, but be more consistent.

jakobj:
I've fixed the merge conflicts @suku248, you can have another look

PR #1425, opened 2020-02-12
Fix python 3.8.0 attribute error: module 'cgi' has no attribute 'escape'

steffengraber:
When working with python3.8 there is an error creating the help files in `../share/doc/nest/help:`

> AttributeError: module 'cgi' has no attribute 'escape

The python documentation lists the attribute until version 3.7. with the note:

> Deprecated since version 3.2: This function is unsafe because quote is false by default, and therefore deprecated. Use html.escape() instead.

https://docs.python.org/3.7/library/cgi.html#cgi.escape

This PR fix the error, so the help files can be created with python3.8 without problems.


heplesser:
Merging without second review since it is a minimal correction to help generation.

Issue #1394, opened 2020-01-14
Spike loss in repeated simulations of exact same script


jarsi:
I have observed, that under some circumstances the number of recorded spikes can diverge. In these repeated simulations I run the exact same script (a static version of examples/nest/hpc_benchmark.sli) and randomly get different results. The following table summarizes the problem:

| commit  | ntasks | number of threads | jube_wp_iteration | number_of_spikes |
| --------|--------|-------------------|-------------------|------------------|
| v2.14.0 |     32 |                24 |                 0 |           316209 |
| v2.14.0 |     32 |                24 |                 1 |           316209 |
| v2.14.0 |     32 |                24 |                 2 |           316209 |
| v2.14.0 |     32 |                24 |                 3 |           316209 |
| v2.14.0 |     32 |                24 |                 4 |           316209 |
| v2.14.0 |     32 |                24 |                 5 |           316209 |
| v2.14.0 |     32 |                24 |                 6 |           316209 |
| v2.14.0 |     32 |                24 |                 7 |           316209 |
| v2.14.0 |     32 |                24 |                 8 |           316209 |
| v2.14.0 |     32 |                24 |                 9 |           316209 |
| v2.14.0 |    768 |                 1 |                 0 |           316209 |
| v2.14.0 |    768 |                 1 |                 1 |           316209 |
| v2.14.0 |    768 |                 1 |                 2 |           316209 |
| v2.14.0 |    768 |                 1 |                 3 |           316209 |
| v2.14.0 |    768 |                 1 |                 4 |           316209 |
| v2.14.0 |    768 |                 1 |                 5 |           316209 |
| v2.14.0 |    768 |                 1 |                 6 |           316209 |
| v2.14.0 |    768 |                 1 |                 7 |           316209 |
| v2.14.0 |    768 |                 1 |                 8 |           316209 |
| v2.14.0 |    768 |                 1 |                 9 |           316209 |
| v2.16.0 |     32 |                24 |                 0 |           **328277** |
| v2.16.0 |     32 |                24 |                 1 |           **333873** |
| v2.16.0 |     32 |                24 |                 2 |           325318 |
| v2.16.0 |     32 |                24 |                 3 |           **325193** |
| v2.16.0 |     32 |                24 |                 4 |           325318 |
| v2.16.0 |     32 |                24 |                 5 |           325318 |
| v2.16.0 |     32 |                24 |                 6 |           **327136** |
[...]

PR #1428, opened 2020-02-12
Use proper delete[] and atomic read in CompletedChecker

jakobj:
as the title states. just maintenance, i don't think this is related to #1394 

@suku248 @heplesser


heplesser:
Reviewed in VC by @suku248, @terhorstd, @heplesser. Merging now to correct weaknesses in existing code, expecting replacement by new solution as part of solution of #1394 .

PR #1430, opened 2020-02-13
Rename has_source_subsequent_targets -> source_has_more_targets

jakobj:
as suggested by @heplesser in #1422

### DIFF
--- a/nestkernel/syn_id_delay.h
+++ b/nestkernel/syn_id_delay.h
@@ -80,7 +80,7 @@ struct SynIdDelay
   bool
   source_has_more_targets() const
   {
-    return this->more_targets;
+    return more_targets;
   }
 
   /**
--- a/nestkernel/syn_id_delay.h
+++ b/nestkernel/syn_id_delay.h
@@ -34,12 +34,12 @@ struct SynIdDelay
 {
   unsigned int delay : NUM_BITS_DELAY;
   unsigned int syn_id : NUM_BITS_SYN_ID;
-  bool subsequent_targets : 1;
+  bool more_targets : 1;
   bool disabled : 1;
 
   explicit SynIdDelay( double d )
     : syn_id( invalid_synindex )
-    , subsequent_targets( false )
+    , more_targets( false )
     , disabled( false )
   {
     set_delay_ms( d );
@@ -48,7 +48,7 @@ struct SynIdDelay
   SynIdDelay( const SynIdDelay& s )
     : delay( s.delay )
     , syn_id( s.syn_id )
-    , subsequent_targets( s.subsequent_targets )
+    , more_targets( s.more_targets )
     , disabled( s.disabled )
   {
   }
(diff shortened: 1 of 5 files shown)


nestkernel/syn_id_delay.h
```diff
@@ -34,12 +34,12 @@ struct SynIdDelay
 {
   unsigned int delay : NUM_BITS_DELAY;
   unsigned int syn_id : NUM_BITS_SYN_ID;
-  bool subsequent_targets : 1;
+  bool more_targets : 1;
```
heplesser: This should be initialized with `true` since it is a `bool`.

nestkernel/syn_id_delay.h
```diff
@@ -72,15 +72,15 @@ struct SynIdDelay
   }
 
   void
-  set_has_source_subsequent_targets( const bool subsequent_targets )
+  set_source_has_more_targets( const bool more_targets )
   {
-    this->subsequent_targets = subsequent_targets;
+    this->more_targets = more_targets;
```
heplesser: Is the `this->` needed here?

nestkernel/syn_id_delay.h
```diff
@@ -72,15 +72,15 @@ struct SynIdDelay
...
   }
 
   bool
-  has_source_subsequent_targets() const
+  source_has_more_targets() const
   {
-    return this->subsequent_targets;
+    return this->more_targets;
```
heplesser: Is the `this->` needed here?

nestkernel/syn_id_delay.h
jakobj: i think this should be initialized with `false` and this is done in https://github.com/nest/nest-simulator/blob/ebae2906f82fe3893e68ab48d1f123f8865cadc2/nestkernel/syn_id_delay.h#L42. Am I missing something here?

nestkernel/syn_id_delay.h
jakobj: i think so, since the parameter name is the same as the member name and we need to distinguish the two

nestkernel/syn_id_delay.h
jakobj: no, it's not needed here

nestkernel/syn_id_delay.h
heplesser: Sorry, I had misread the code, all is fine.

nestkernel/syn_id_delay.h
heplesser: OK!

heplesser:
Interaction between Github and Travis seems to be in trouble here (@lekshmideepu for info), so even though the Travis build completed successfully a long time ago (https://travis-ci.org/nest/nest-simulator/builds/652502447), Github shows the check as still pending.

Since the test has been completed and the change approved in VC by @suku248, @terhorstd, @heplesser, I will merge.

lekshmideepu:
@heplesser Thanks for the info. I would keep an eye on this issue.

lekshmideepu:
@heplesser I guess that was a temporary issue. Syncing between Github and Travis for the latest build (https://github.com/nest/nest-simulator/pull/1415) seems to be fine.

PR #1438, opened 2020-02-19
Only check for contains list if params is not empty

### LINKED ISSUE #1436: nest.Create() with empty parameters dictionary results in exception
OK:
```
(Pdb) nest.Create('iaf_psc_delta')
NodeCollection(metadata=None, model=iaf_psc_delta, size=1, first=2)                                            
```
Not OK:
```
(Pdb) nest.Create('iaf_psc_delta', params={})                                                                  
*** ValueError: max() arg is an empty sequence
```

Version: NEST3 master@2b3c13e92

jakobj:
not entirely sure what the use case is @clinssen, but fixes #1436 ;)

as reviewers i suggest @stinebuu @hakonsbm

### DIFF
--- a/pynest/nest/lib/hl_api_nodes.py
+++ b/pynest/nest/lib/hl_api_nodes.py
@@ -99,7 +99,7 @@ def Create(model, n=1, params=None, positions=None):
         return layer
 
     params_contains_list = True
-    if isinstance(params, dict):
+    if isinstance(params, dict) and params:
         params_contains_list = [is_iterable(v) or isinstance(v, Parameter)
                                 for k, v in params.items()]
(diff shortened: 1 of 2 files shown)


jakobj:
good point @hakonsbm, I've added a test now

jakobj:
oh and btw: i think https://github.com/nest/nest-simulator/blob/bc7ff0779b831d4b27af9de9b528bc4240139ec2/pynest/nest/lib/hl_api_nodes.py#L102 ff would really benefit from some comments explaining what's going on, I find it hard to follow @hakonsbm @stinebuu

jakobj:
good suggestion @hakonsbm, thanks!

PR #1443, opened 2020-02-26
Count local devices for each thread

### LINKED ISSUE #1409: Nest 3 : Problem of reproducibility and extrem firing rate of neurons
I test Nest 3 and I get strange behavior of the firing rate.
I try to reduce my code in order to isolate the problems.

I get the following result with the version 2.18.0 : 

test_nest_2_18
 
When I use the commit fbebdc0aefa4a06e09d6647f19c695bf8071d640 of Nest 3, I get the following result :

test_nest_3

We can see the result are different for each run.
Moreover, some firing rate are to high for the parameters.

This is the code for reproduce this figure:

[problem_nest.txt](https://github.com/nest/nest-simulator/files/4107914/problem_nest.txt)

stinebuu:
This PR fixes #1409.

When creating devices, the number of local devices are counted. With NEST 3.0, the counting happens in parallell, but not in a thread safe way, so sometimes the threads might overwrite the variable. Then we get a problem with device ids, which in return can mess up spike delivery.

With this PR we now create a vector for the counting instead, so each thread keep track of the number of devices on the thread.

@lionelkusch could you check if this fixes your issue?

### DIFF
--- a/pynest/nest/tests/test_regression_issue-1409.py
+++ b/pynest/nest/tests/test_regression_issue-1409.py
@@ -46,7 +46,7 @@ class MultiplePoissonGeneratorsTestCase(unittest.TestCase):
 
             parrots = nest.Create('parrot_neuron', num_neurons)
             poisson_generator = nest.Create('poisson_generator', num_pg)
-            nest.SetStatus(poisson_generator, 'rate', 2000.)
+            poisson_generator.rate = 2000.
 
             nest.Connect(poisson_generator, parrots, 'all_to_all')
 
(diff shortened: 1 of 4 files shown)


heplesser:
@stinebuu Nice detective work :)!

pynest/nest/tests/test_regression_issue-1409.py
```diff
@@ -0,0 +1,66 @@
...
+        num_spikes = []
+        for i in range(num_iterations):
+            nest.ResetKernel()
+            nest.SetKernelStatus({'local_num_threads': local_num_threads})
+            nest.set_verbosity('M_WARNING')
+            print('num iter {:>5d}/{}'.format(i+1, num_iterations), end='\r')
+
+            parrots = nest.Create('parrot_neuron', num_neurons)
+            poisson_generator = nest.Create('poisson_generator', num_pg)
+            nest.SetStatus(poisson_generator, 'rate', 2000.)
```
hakonsbm:
```suggestion
            poisson_generator.rate = 2000.
```

PR #1429, opened 2020-02-13
Add option to generate connections from NumPy arrays

hakonsbm:
Connections are created one-to-one based on NumPy arrays of source and target node IDs, and a synapse specification containing NumPy arrays of weights and delays. All arrays have to be of equal length. For efficiency, the PyNEST kernel bypasses SLI and passes a pointer for each array to the C++ level.

It can be used as in the following example:

```python
nest.Create('iaf_psc_alpha', 10)
# sources and targets must be NumPy arrays of equal lengths, with integer dtype
sources = np.array([1, 5, 7], dtype=np.uint64)
targets = np.array([2, 4, 6], dtype=np.uint64)
# weights, delays, and receptor type must be NumPy arrays with lengths equal to 
# that of sources and targets, or unspecified
weights = np.ones(len(sources))
delays = np.ones(len(sources))

# conn_spec cannot be specified
nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays,
                                         'synapse_model': 'static_synapse'})
```

This replaces the implementation of `Connect_nonunique`.

Fixes #1423.

PR #1447, opened 2020-02-28
Remove pylab and delay pyplot import

### LINKED ISSUE #1124: Remove pylab from pynest
I'm curiouis how people would feel about removing `pylab` from the docs/examples. It's a bit of a namespace wrecker, and teaches new programmers bad habits (under the hood, it includes `from numpy import *` and `from matplotlib import *`). It is officially deprecated and discouraged by the Matplotlib developers. For instance, in the [matplotlib usage guide](https://matplotlib.org/tutorials/introductory/usage.html#sphx-glr-tutorials-introductory-usage-py):

>  pylab is deprecated and its use is strongly discouraged because of namespace pollution. Use pyplot instead.

How do people feel about starting to replace `import pylab` with more Pythonic, explicit references to `numpy` and `matplotlib`? It is a pretty easy fix. I'm working through the tutorials/examples and replacing pylab every time it comes up. I'd be happy to help with this cleansing if people support it.

Silmathoron:
This PR fixes #1124 and delays import of ``matplotlib.pyplot`` to allow easy backend switch even after NEST is imported.

PS: also corrected a small error [there](https://github.com/nest/nest-simulator/pull/1447/files#diff-9666555a827ac627644d4e15634d6003R1493)


Silmathoron:
Damn, apparently I wrongly included some changes from another branch... either we merge this after #1429 or I try to correct this and do a force push...?

doc/tutorials/pynest_tutorial/part_1_neurons_and_simple_neural_networks.rst
```diff
@@ -438,7 +438,7 @@ a constant input current, and add a second neuron.
 
 ::
 
-    import pylab
+    import plt
```
stinebuu: Shouldn't this be `import matplotlib.pyplot as plt`

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -848,6 +847,8 @@ def SelectNodesByMask(layer, anchor, mask_obj):
 def _draw_extent(ax, xctr, yctr, xext, yext):
     """Draw extent and set aspect ration, limits"""
 
+    import matplotlib.pyplot as plt
```
stinebuu: Could you add a comment about why we import here and not at the top? It would not be obvious for me when looking through the code, and I think it would be easy to convert back if we are not reminded why it is here. Also for the imports below.

doc/tutorials/pynest_tutorial/part_1_neurons_and_simple_neural_networks.rst
Silmathoron: I will actually remove because it is not used

Silmathoron:
@stinebuu thanks for the comments, should be addressed now.
And yeah, let's wait for #1429, please add the PR blocked label :+1:

PR #1448, opened 2020-02-28
Initialize indicator for changes to connectivity to false

suku248:


stinebuu:
I can't see anything wrong with this change, all tests pass so that indicates that everything is ok to me. It seems like the variable was changed from `false` to `true` in ba0ec596cad42614dddb0ff67379ea22ab00fc19, and that commit had something to do with manual disconnects and structural plasticity. I don't know why the initialiser has anything to do with disconnect though, but I could very well be missing something.

PR #1458, opened 2020-03-09
Correctly initialize result vector in music_rate_in_proxy::get_status()

### LINKED ISSUE #1449: GetDefaults occasionally causes bad_alloc if NEST is built with MUSIC
We have in recent weeks seen a number of Travis build failures in stage 7 (the "all in build"), e.g.

- https://travis-ci.org/nest/nest-simulator/jobs/655816865?utm_medium=notification&utm_source=github_status
- https://travis-ci.org/nest/nest-simulator/jobs/655980035?utm_medium=notification&utm_source=github_status

The failure can occur in different tests, but common is that

- a `bad_alloc` exception is triggered
- `GetDefaults` is involved.

Since this issue obstructs testing, I escalate it to critical.

Experimenting on Travis, I have confirmed that the issue occurs if NEST is built with just MPI and MUSIC, but no threading, GSL, LDTL, Python, Readline.

I have also confirmed the issue under OSX building with gcc 9.2.0. There
```
for i in {1..20}; do nest --verbosity=ERROR -c "modeldict keys { ResetKernel GetDefaults ; } forall" ; done
```
will usually evoke in one of the twenty runs
```
Mar 01 17:23:21 GetDefaults_l [Error]: C++Exception
    cannot create std::vector larger than max_size()
```
In all cases I have observed, the last model for which `GetDefaults` was called, was `music_rate_in_proxy`, although a test which only called `GetDefaults` on that model did not trigger the exception  in 200 runs.

The following version calls `GetDefaults` only on the `music_` models and prints the model names. It usually triggers an exception at least once in twenty runs.
```
for i in {1..20}; do 
    nest --verbosity=ERROR -c "modeldict keys { dup cvs 6 Take (music_) eq 
           { ResetKernel dup =only ( ) =only GetDefaults ; } 
           { ; } ifelse 
       } forall ( ) = ";
done
```

hakonsbm:
Corrects the type of the `Buffers_::data_` variable in the MUSIC rate in proxy model.

With the original uninitialised `double`, passing the variable to the `std::vector` constructor may lead to an `std::bad_alloc` error. This can happen when the model tries to copy the vector with
```c++
( *d )[ names::data ] = DoubleVectorDatum( new std::vector< double >( B_.data_ ) );
```

This PR changes the variable to an `std::vector`.

Fixes #1449.

### DIFF
--- a/models/music_rate_in_proxy.h
+++ b/models/music_rate_in_proxy.h
@@ -167,7 +167,7 @@ private:
 
   struct Buffers_
   {
-    double data_;
+    std::vector< double > data_; //!< The buffer for incoming data
   };
 
   // ------------------------------------------------------------
--- a/models/music_rate_in_proxy.h
+++ b/models/music_rate_in_proxy.h
@@ -167,7 +167,7 @@ private:
 
   struct Buffers_
   {
-    std::vector< double > data_; //!< The buffer for incoming data
+    double data_;
   };
 
   // ------------------------------------------------------------
--- a/models/music_rate_in_proxy.cpp
+++ b/models/music_rate_in_proxy.cpp
@@ -150,7 +150,7 @@ nest::music_rate_in_proxy::get_status( DictionaryDatum& d ) const
   P_.get( d );
   S_.get( d );
 
-  ( *d )[ names::data ] = DoubleVectorDatum( new std::vector< double >(1, B_.data_ ) );
+  ( *d )[ names::data ] = DoubleVectorDatum( new std::vector< double >( 1, B_.data_ ) );
 }
 
 void
--- a/models/music_rate_in_proxy.cpp
+++ b/models/music_rate_in_proxy.cpp
@@ -37,6 +37,7 @@
 
 // Includes from nestkernel:
 #include "kernel_manager.h"
+#include "event_delivery_manager_impl.h"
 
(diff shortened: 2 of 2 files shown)


models/music_rate_in_proxy.h
```diff
@@ -167,7 +167,7 @@ class music_rate_in_proxy : public DeviceNode
 
   struct Buffers_
   {
-    double data_;
+    std::vector< double > data_; //!< The buffer for incoming data
```
heplesser: @hakonsbm Based on your [previous analysis](https://github.com/nest/nest-simulator/issues/1449#issuecomment-593950033) this update makes sense, but I wonder if a smaller change, namely changing 
```c++
( *d )[ names::data ] = DoubleVectorDatum( new std::vector< double >( B_.data_ ) );
```
to
```c++
( *d )[ names::data ] = DoubleVectorDatum( new std::vector< double >( 1, B_.data_ ) );
```
so that we correctly create a single-element vector, would be the safer approach here to just fix this problem?

hakonsbm: Done. But we should really have a test of this model.

heplesser: Indeed, would you create a follow-up issue for a proper review of this (and other music) proxies?

PR #1465, opened 2020-03-13
Backport of #1458 to 2.20.1

jasperalbers:


heplesser:
@jasperalbers My mistake! I believed that `music_rate_in_proxy` was available in 2.20.0. Since it wasn't, we should not add it. I therefore close this PR.

PR #1439, opened 2020-02-21
Create documentation workflow page

sarakonradi:
This PR creates a documentation workflow page. 

The page introduces the tools we use to generate and publish documentation (Sphinx, GitHub, Read the Docs).

It explains how to:
- Change the documentation 
- Set up a working environment
- Generate HTML with Sphinx
- Preview the content on Read the Docs

Currently, our User Documentation and Developer Space are separate from one another. @steffengraber and I discussed that, once the latter is available on Read the Docs, this workflow would be moved to the developer section.

PR #1468, opened 2020-03-16
Move topology documentation into doc folder

### LINKED ISSUE #1408: Remove pdf versions of Topology User Manual for NEST 3.0?
There are currently two pdf versions of the Topology User Manual in the source code, one in the topology folder: https://github.com/nest/nest-simulator/blob/master/topology/doc/rendered/Topology_UserManual_tex.pdf and one in the doc foler: https://github.com/nest/nest-simulator/blob/master/doc/guides/Topology_UserManual.pdf. The one in the doc folder is for NEST 2.12, while the one in the topology folder is for 2.16.

There is also a rst version of the file, https://github.com/nest/nest-simulator/blob/master/doc/topology/Topology_UserManual.rst, for [ReadTheDocs](https://nest-simulator.readthedocs.io/en/latest/topology/Topology_UserManual.html). For nest-3, we have updated this rst file to fit with the changes ([here](https://github.com/nest/nest-simulator/blob/nest-3/doc/guides/guide_structurally_spatial_networks.rst)).

The Topology folder also has a [rst file](https://github.com/nest/nest-simulator/tree/master/topology/doc/source), as well as a [SnakeFile](https://github.com/nest/nest-simulator/blob/master/topology/doc/Snakefile), which is used to generate the PDF version from the rst file.

In nest-3, we want to move all documentation out of the topology folder. My question is this:

- Do we still want to keep the pdf versions of the Topology User Manual?
- If yes, where should we keep the Snakefile? I think it should be close to the rst file in the doc folder, as the pdf version should be based on that one.

sarakonradi:
This PR moves the documentation out of the ``topology`` folder and into the ``doc`` folder. 

More specifically, it: 
- Renames the user manual to ``guide_spatially_structured_networks.rst``
- Removes the PDF
- Deletes the Snakefile
- Updates the links to the guide

It fixes #1408.


sarakonradi:
@stinebuu Thanks for checking! I agree that ``guide_spatially_structured_network.rst`` looks nice on Read the Docs and have removed everything that used to be in ``topology/doc``. 

Currently, the topology files can now be found in two locations in the documentation:
- ``guide_spatially_structured_network.rst`` in ``doc/guides``
- ``user_manual_figures`` and ``user_manual_scripts`` in ``doc/topology``

Alternatively, I can move the ``topology`` folder into ``doc/guides`` and include everything there (guide, figures, scripts). What do you think?

doc/tutorials/pynest_tutorial/part_4_topologically_structured_networks.rst
```diff
@@ -46,9 +46,12 @@ In the previous sections, we saw that we can create deterministic or
...
 models that incorporate spatial location and spatial connectivity
-profiles. **NOTE:** Full documentation for usage of the spatial functions
-is present in our :doc:`../../guides/guide_structurally_spatial_networks` [1]_,
-which in the following is referenced as full-source.
+profiles.
+
+.. note::
+   Full documentation for the usage of spatial functions
+   is present in our :doc:`../../guides/guide_spatially_structured_networks` [1]_,
+   which in the following is referenced as full-source.
```
heplesser: I find "is present" a bit cumbersome and "full-source" a little informative name. How about
"For comprehensive documentation of spatial properties and connectivity, see the :doc:`../../guides/guide_spatially_structured_networks` [1]_, which is referenced as *Spatial Network Guide* in the following."

How does the link show in the rendered docs?

Furthermore, the reference [1] seems to point to the PDF version of the manual. But that makes no sense. As long as all the material from the manual is available in the RST `guide_spatially_structured_networks`, there is no need for the PDF and no point in citing it.

stinebuu:
@sarakonradi @heplesser I think then the subdirectory should not be called `topology`, but maybe `spatial` or something.

heplesser:
> @sarakonradi @heplesser I think then the subdirectory should not be called `topology`, but maybe `spatial` or something.

Indeed!

doc/tutorials/pynest_tutorial/part_4_topologically_structured_networks.rst
sarakonradi: Yes, it does sound a bit cumbersome! I changed it to the phrase you suggested with a full stop after ``[1]_``. The link shows up as the title of the guide (``Guide to spatially structured networks``) and points to the RST version of the guide. The PDF has been removed.  

You can preview the documentation locally using Sphinx. Sure you already saw the instructions in #1439 :)

The figures are added in the reStructured text using ``.. figure::`` and the path. The images are located in ``doc/guides/spatial/user_manual_figures``. The code examples are embedded using ``.. literalinclude::`` and can be found in ``doc/guides/spatial/user_manual_scripts``.

sarakonradi:
@stinebuu @heplesser Good idea! I deleted ``doc/topology`` and moved everything to ``doc/guides/spatial``. 

Two more questions:
- Should we rename the PyNEST tutorial from``part_4_topologically_structured_networks.rst`` to ``part_4_spatially_structured_networks.rst``?
- What should we do about the name of the topology module? (see ``guide_spatially_structured_networks.rst``, line 1549)

heplesser:
> * Should we rename the PyNEST tutorial from`part_4_topologically_structured_networks.rst` to `part_4_spatially_structured_networks.rst`?

Yes please!

> * What should we do about the name of the topology module? (see `guide_spatially_structured_networks.rst`, line 1549)

@stinebuu @hakonsbm Could you comment on this one? It concerns adding additional masks at the C++ level. Is that still meaningful at all since you can do everything on the Python level?

stinebuu:
@heplesser I think it is still meaningful. It's about adding custom masks, which I don't think is possible from Python (@hakonsbm?). For masks you need several `inside` functions etc., which sounds difficult from Python.

I don't know how many use custom masks, and I guess people can just read the documentation for MyModule. But, to add a mask, you still need to register it in `nest::TopologyModule`, because we have kept the separation between spatial C++ implementations of NEST and core NEST, and never renamed it at C++ level. This is why we use topology module in line 1549 I think.

Maybe change it to `The mask class must then be registered with the NEST's C++ topology module`? To try to make it clear that we separate the two?

hakonsbm:
@heplesser @stinebuu It's not possible to add custom masks from Python. However, by using the  `nest.logic.conditional()` and `nest.spatial.distance` Parameters it is possible to create a connection probability that behaves like a mask. Of course, this would probably not be as efficient as a propper mask.

heplesser:
> @heplesser @stinebuu It's not possible to add custom masks from Python. However, by using the `nest.logic.conditional()` and `nest.spatial.distance` Parameters it is possible to create a connection probability that behaves like a mask. Of course, this would probably not be as efficient as a propper mask.

@hakonsbm OK, so there is still a use case for C++ masks? Is the documentation at the bottom of `guide_spatially_structured_networks.rst` still up to date or do you have suggestions for how to improve them? And how does one use a self-defined mask in the new `spatial` setup?

hakonsbm:
> @hakonsbm OK, so there is still a use case for C++ masks?

@heplesser  Yes. I just did some benchmarks, and using a mask is significantly more efficient (20-30 times faster) than cutting off the probability with the `nest.logic.conditional()` Parameter. Still, using a Parameter is a lot easier than implementing a custom mask in a C++ module. Should probably also look into if using a Parameter can be made more efficient.

> Is the documentation at the bottom of `guide_spatially_structured_networks.rst` still up to date or do you have suggestions for how to improve them?

No, it looks a bit outdated. MyModule has no `bootstrap.sh` script, for example. I think we should have the guide on how to compile and use the MyModule in a single location, and just reference it here and note the relevant changes and additions.

> And how does one use a self-defined mask in the new `spatial` setup?

According to the guide, which implements an elliptic mask:
```Python
nest.Install('mymodule')
l = nest.Create('iaf_psc_alpha', positions=nest.spatial.grid(rows=11, columns=11, extent=[1., 1.]))
nest.Connect(l, l, {'rule': 'pairwise_bernoulli',
                    'p': 1.0,
                    'mask': {'elliptic': {'r_x': 0.5, 'r_y': 0.25}}})
```
Apart from the outdated `nest.spatial.grid` arguments, this looks correct, but I haven't tested it.

heplesser:
@sarakonradi @hakonsbm We should handle the update of the mymodule-doc in a separate issue, focus on moving files here.

@stinebuu Could you re-review this PR and see if you concerns are addressed?

sarakonradi:
I renamed the PyNEST tutorial to ``part_4_spatially_structured_networks.rst``, as discussed with @heplesser.

> @sarakonradi @hakonsbm We should handle the update of the mymodule-doc in a separate issue, focus on moving files here.

Yes, good idea.

stinebuu:
@sarakonradi Now that `guide_spatially_structured_networks.rst` is placed under `doc/guides/spatial` do we also need to update or removed the link from `doc/guides/index.rst` line 22?

sarakonradi:
> @sarakonradi Now that `guide_spatially_structured_networks.rst` is placed under `doc/guides/spatial` do we also need to update or remove the link from `doc/guides/index.rst` line 22?

@stinebuu We do, thanks for noticing. Just fixed that. Also found another link in the PyNEST tutorial that had to be changed.

Issue #1426, opened 2020-02-12
`less -S`  in `.nestrc` failed


steffengraber:
When I try to call the help in NEST with for example `nest.help('abs')`, I get the error

> Displaying help with pager "less -S" failed. Please define a working parser in file .nestrc in your home directory.

Whereby a test with less -S SOMEFILE in the terminal works.

If I change `less -S` to `less` in `.nestrc`, `nest.help('abs')` will also work.

Tested with the source install of the latest `master` with python3.6 and python3.8.

steffengraber:
Checked on macOS Catalina, with NEST 2.20.0 and python3.7 - same problem.

PR #1474, opened 2020-03-19
Enables display of commands with default settings in the `.nestrc `

steffengraber:
This simple solution separates the pager name and its options into a list. This allows the pager to be recognized correctly with the default settings in `.nestrc ` (`/page << /command (less -S)`).

This PR fix issue #1426


pynest/nest/lib/hl_api_helper.py
```diff
@@ -501,15 +501,16 @@ def show_help_with_pager(hlpobj, pager=None):
         return
 
     try:
-        subprocess.check_call([pager, objf])
+        pagerl = pager.split(" ")
```
terhorstd: do we want to use shlex.split here? That would also handle spaces and quotes correctly, etc.
```suggestion
        import shlex
        pagerl = shlex.split(pager)
```

terhorstd: any opinion on this, anyone?

steffengraber: @terhorstd That looks good. I will be trying it out.

steffengraber: @terhorstd  Works for me. Changed it.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -499,7 +499,8 @@ def show_help_with_pager(hlpobj, pager=None):
         return
 
     try:
-        pagerl = pager.split(" ")
+        # pagerl = pager.split(" ")
```
terhorstd, line 502: remove commented code
```suggestion
```

heplesser:
@steffengraber This one should be good to go once the gitignore conflict is resolved.

PR #1475, opened 2020-03-20
Improve visibility of NEST 3.0 conversion guides

### LINKED ISSUE #1450: Improve visibility of NEST2-to-NEST3 conversion guide
There are two NEST2-to-NEST3 conversion guides in the documentation: 
- A general overview in `doc/guides`, which can be found within "Guides" on Read the Docs: https://github.com/nest/nest-simulator/blob/master/doc/guides/from_nest2_to_nest3.rst
- A detailed functions guide in `doc/ref_material`, which is listed in the main table of contents: https://github.com/nest/nest-simulator/blob/master/doc/ref_material/nest2_vs_3.rst 

This may be confusing to a user who visits the documentation website and is interested in the general "What's new?", but who lands on the detailed functions guide instead.

It made me wonder if we could: 
- List both files under "Guides": https://nest-simulator.readthedocs.io/en/latest/guides/index.html
- Use more meaningful titles, e.g. "NEST 2.x to NEST 3.0: What's new?" and "NEST 2.x to NEST 3.0: Changes to functions" 

@heplesser I realize that you already shared the conversion guides' links in your NEST 2.20 announcement (and possibly elsewhere). 

My question is: Would such changes make sense at this stage?

sarakonradi:
This PR makes the NEST 3.0 conversion guides more visible to users visiting Read the Docs. It also tidies up the folder structure. 

More specifically, it:
- Changes the title of the overview guide to ``NEST 3.0: What's new?``
- Renames the detailed guide to ``NEST 3.0: Detailed transition guide``
- Lists both guides under ``Transition guides to NEST 3.0`` on Read the Docs
- Adds the overview guide to the table of contents
- Moves both files to ``doc/guides/nest2_to_nest3``
- Adds ``_overview`` and ``_detailed_transition guide`` to the file names
- Updates all corresponding links throughout the documentation

It fixes #1450.


doc/guides/nest2_to_nest3/nest2_to_nest3_detailed_transition_guide.rst
```diff
@@ -20,12 +20,12 @@ NEST 2.X vs. NEST 3.0 conversion reference guide
...
 
-  You can find code examples of changes in NEST 3.0 :doc:`in our NEST-3 overview <../guides/from_nest2_to_nest3>`
+  To see code examples of the key changes, check out our :doc:`nest2_to_nest3_overview` guide.
 
 .. _setverbosity:
 
-Supress output on startup
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Suppress output on startup
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
heplesser: Is one ~ missing here at the end?

doc/index.rst
```diff
@@ -1,17 +1,17 @@
-***********************************************
+********************************************
 Welcome to the NEST simulator documentation!
-***********************************************
+********************************************
 
 :orange:`Introducing NEST 3.0!`
 -------------------------------
 
 Our aim with NEST 3.0 is to provide a more intuitive experience with simplified yet versatile handling and manipulation of nodes and connections.
```
heplesser: Be more direct: "NEST 3.0 provide a more intuitive ..."

doc/index.rst
```diff
@@ -1,17 +1,17 @@
...
 
 
-- We have provided a :doc:`guide to NEST 3.0 <guides/from_nest2_to_nest3>` , including examples of the new functionality.
+- We have provided an :doc:`overview of NEST 3.0 <guides/nest2_to_nest3/nest2_to_nest3_overview>`, including examples of the new functionality.
```
heplesser: How about: "Get an :doc:`overview of NEST 3.0 ...`, including ..."? And should we capitalise the name of the guide as a title, i.e., "Overview of NEST 3.0".

doc/index.rst
```diff
@@ -1,17 +1,17 @@
...
 
 
-- You can also checkout our :doc:`ref_material/nest2_vs_3` to get a comparison between old and new syntax.
+- You can also check out our :doc:`detailed transition guide <guides/nest2_to_nest3/nest2_to_nest3_detailed_transition_guide>` for an in-depth comparison between old and new syntax.
```
heplesser: Suggestions: "Read the :doc:`detailed transition guide` ... for an in-depth comparison between old and new syntax and semantics."

doc/index.rst
sarakonradi: Done.

doc/index.rst
sarakonradi: Done.

doc/index.rst
sarakonradi: I agree about making the sentence more direct and capitalizing the title. 

How about ``Visit our What’s new? guide to get an overview of NEST 3.0 and the new functionality``? 

Since ``What's new?`` is the name of the guide.

sarakonradi:
@heplesser Thanks for the suggestions, I added the changes and replied inline.

doc/index.rst
heplesser: Good!

sarakonradi:
> Very nice. I noticed one other thing: When I click on "NEST 3.0: What's new?" I see it twice on the left side in the navigation. Is that supposed to happen?

@steffengraber That happens because ``NEST 3:0: What's new?`` is listed in the main table of contents as well in ``guides/index.rst`` (which is also part of the table of contents).

A possible solution would be to remove it from ``contents.rst`` and to keep it in ``guides/index.rst`` only. This way, clicking it will cause it to appear once in the navigation. 

What do you think?

steffengraber:
> A possible solution would be to remove it from `contents.rst` and to keep it in `guides/index.rst` only. This way, clicking it will cause it to appear once in the navigation.
> 
> What do you think?

But then it does not have the top-level visibility in the navigation, which is actually intended. Right?

sarakonradi:
> But then it does not have the top-level visibility in the navigation, which is actually intended. Right?

@steffengraber Yes, ideally it should have a top-level visibility.

We have three options here: 
1. Removing it from ``contents.rst`` 
Trade-off: No top-level visibility 

2. Removing it from ``guides/index.rst``
Trade-off: It is a guide and should be kept there

3. Keeping everything as it is
Trade-off: It shows twice in the navigation

Which one should we choose?

steffengraber:
Option 1 is ok for me.

PR #1477, opened 2020-03-22
Build NEST with Boost as default

### LINKED ISSUE #1238: boost include path not properly set in NEST 2.18.0
Make fails with the following error with -Dwith-boost=ON or -Dwith-boost=<PATH_TO_BOOST>, even with the correct boost include path being displayed in the NEST Configuration Summary.

nest-simulator-2.18.0/libnestutil/sort.h(36): catastrophic error: cannot open source file "boost/sort/spreadsort.hpp"

This error can be solved by adding
-Dwith-includes=<PATH_TO_BOOST_INCLUDE_DIR>

### LINKED ISSUE #1239: test_connect_after_simulate.sli fails with assertion 'source_lid < MAX_LID' when compiled with boost=ON in NEST 2.18.0
After successfully building with boost=ON by explicitly specifying the boost include path (see #1238)

make installcheck fails with:

<NEST_SOURCE_DIR>/nestkernel/target_data.h:264: void nest::TargetData::set_source_lid(unsigned long): Assertion 'source_lid < MAX_LID' failed.

when executing:

mpirun -np 1 <NEST_INSTALL_PATH>/bin/nest <NEST_INSTALL_PATH>/share/doc/nest/unittests/test_connect_after_simulate.sli

Note:
All installcheck pass when configured without boost

### LINKED ISSUE #1469: Build with boost by default
Currently, the default for NEST builds is `-Dwith-boost=OFF`, so unless users are very observant, they will build without Boost and suffer considerable performance problems.

Make `-Dwith-boost=ON` the default and provide a warning in the CMake summary if Boost is not available (similar to GSL).

**Additional context**
See also #1461, #1238.

niltonlk:
This PR

1. Makes Boost enabled by default (fixes #1469)
2. Display a message on the Configuration summary when Found Version is greater or equal than 1.69.0 (fixes #1239)
3. Fix Boost include path issue (fixes #1238) following suggestions by @clinssen, however with the correct spelling.


heplesser:
This is relevant also in connection with #1489. We need to discuss carefully if we want to make Boost the default. We should definitely *not*  prohibit newer versions of Boost just because they currently cause problems. We rather need to find the root of the problems, fix them, and then possibly admit only newer versions of Boost.

heplesser:
On macOS, using Boost from Homebrew will fail when compiling with g++-9, because g++ assumes abi:cxx11 for unit tests, while Boost's unit test binary does not use the cxx11 ABI (see also https://www.howtobuildsoftware.com/index.php/how-do/kQw/c-unit-testing-boost-linker-abi-boost-unit-test-link-error-abi-mismatch). It can be made to work by adding `-D_GLIBCXX_USE_CXX11_ABI=0` to the compiler flags. No problem using the Clang compiler.

heplesser:
The problems with Boost versions later than 1.68 should be solved once #1502 is merged.

@niltonlk Could you modify this PR so that it accepts any Boost version 1.58.0 or later?

cmake/ConfigureSummary.cmake
```diff
@@ -149,7 +149,11 @@ function( NEST_PRINT_CONFIG_SUMMARY )
     message( "    Libraries       : ${BOOST_LIBRARIES}" )
     message( "" )
   else ()
-    message( "Use Boost           : No" )
+    if ( "${BOOST_VERSION}" VERSION_GREATER_EQUAL "1.69" )
+      message( "Use Boost           : Boost disabled due to version missmatch. Version must be less than 1.69.0 (Found Version: ${BOOST_VERSION})" )
```
clinssen, line 153:
```suggestion
      message( "Use Boost           : Boost disabled due to version mismatch. Version must be less than 1.69.0 (Found Version: ${BOOST_VERSION})" )
```

niltonlk:
Sorry for the late reply. I have disabled Boost upper bound version check as requested by @heplesser and corrected the misspelling suggested by @clinssen

niltonlk:
As I have just commented on PR #1502, but should I change
```
set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIR}" PARENT_SCOPE )
include_directories( ${Boost_INCLUDE_DIR} )
```
 to
```
set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIRS}" PARENT_SCOPE )
include_directories( ${Boost_INCLUDE_DIRS} )
```
with INCLUDE_DIR**S** as commented by @hakonsbm on PR #1502 ?

cmake/ProcessOptions.cmake
```diff
@@ -594,6 +594,8 @@ function( NEST_PROCESS_WITH_BOOST )
       set( BOOST_LIBRARIES "${Boost_LIBRARIES}" PARENT_SCOPE )
       set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIR}" PARENT_SCOPE )
       set( BOOST_VERSION "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" PARENT_SCOPE )
+      
+      include_directories( ${Boost_INCLUDE_DIR} )
```
gtrensch, line 598: @niltonlk This PR conflicts with #1502.

niltonlk:
I have correct accordginly to PR #1502 as requested by @gtrensch .
In addition, to avoid confusion, I also updated line 595 to unify the variable used for the returned boost include path
`set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIRS}" PARENT_SCOPE )`

PR #1481, opened 2020-03-24
Switch from bisection method to regula falsi method in precise models

### LINKED ISSUE #403: Nest running into infinite loop (?)
I'm encountering an error when running trying to run my network. Nest runs into an infinite loop, using 100% cpu of one core but making no progress. I'm using PyNN, but debugging showed that the loop occurs in nest. I have encountered this behaviour multiple times, it is difficult to predict and highly dependent on the neuron parameters. In a similar example, the error did not occur immediately, but only 70 out of 100 seconds into the simulation.
I apologise for the large example, it is the smallest example that was able to reliably produce the error. For understanding, I am trying to reproduce behaviour of NEF with Nest.
I am using PyNN 0.8.1 from git and Nest 2.10.0 from the website, though I have also encountered a similar error with PyNN 0.8.0 and Nest 2.8.0 as well as PyNN 0.7.5 and Nest 2.2.2.

What I find particularly confusing is that I have a different script that produces the exact same network, with the same components and same parameters which does not produce the error. I have attached it as well. Invoke it with `python save_ensemble.py <random-seed>`. The seed I used was 11245.

``` python
# (not so) mwe.py
# produces infinite loop during the first simulation run
import numpy as np
import pyNN.nest as sim

sim.setup(timestep=1.0, min_delay=1.0)

n_neurons = 40
n_steps = 100

sim_time = 1.0

T_REF = 0.001
T_RC = 0.02
V_THRESH = 20.0
V_REST = -70.0
TAU_SYN = 0.001

RATE_LOW = 25
RATE_HIGH = 75

ratio = 1.0

func = lambda x: x * x

x_values = np.linspace(-1, 1, n_steps)
func_values = func(x_values)

def calc_tuning_curves(encoders, gain, bias):
    x_samples = np.linspace(-1, 1, n_steps)
    tuning_curves = np.zeros((n_steps, n_neurons))
    for i in range(n_steps):
        J = gain * np.dot(x_samples[i], encoders) + bias
        j = J - V_THRESH
        tuning_curves[i, j > 0] = 1.0 / (T_REF + T_RC * np.log1p(V_THRESH / j[j > 0]))
    return tuning_curves.astype(np.uint16)

def calc_gain_bias(intercepts, max_rates):
    x = V_THRESH / (1 - np.exp((T_REF - (1.0 / max_rates)) / T_RC))
    gain = (V_THRESH - x) / (intercepts - 1)
    bias = V_THRESH - gain * intercepts
    return gain, bias

def generate_ensemble(seed):
    np.random.seed(seed)
    encoders = np.random.choice([-1,1], n_neurons)
    intercepts = np.random.uniform(-1, 1, n_neurons)
    max_rates = np.random.uniform(RATE_LOW, RATE_HIGH, n_neurons)
    gain, bias = calc_gain_bias(intercepts, max_rates)
    tuning_curves = calc_tuning_curves(encoders, gain, bias)
    return encoders, intercepts, max_rates, tuning_curves

encoders_A, intercepts_A, max_rates_A, tuning_curves_A = generate_ensemble(11245)

gain, bias = calc_gain_bias(intercepts_A, max_rates_A)

decoders_A = np.linalg.lstsq(tuning_curves_A, x_values)[0]

cellparams = {
    'tau_refrac':T_REF * 1000.0,
    'cm':T_RC / gain,
    'v_rest':V_REST,
    'tau_syn_E':TAU_SYN * 1000.0,
    'tau_syn_I':TAU_SYN * 1000.0,
    'tau_m':T_RC * 1000.0,
    'i_offset':bias / gain,
    'v_thresh':V_REST + V_THRESH,
    'v_reset':V_REST
}

A = sim.Population(n_neurons, sim.IF_curr_exp(**cellparams))
A.initialize(v=V_REST)
A.record('spikes')

input_source = sim.Population(2, sim.SpikeSourcePoisson())
weights = np.array([[1.0/ratio] * n_neurons * encoders_A] * 2)
weights[1] *= -1
sim.Projection(input_source, A, sim.AllToAllConnector(),
    sim.StaticSynapse(weight=weights))

decoders = np.linalg.lstsq(tuning_curves_A, func_values)[0]

spikes_A = np.zeros((n_steps, n_neurons))

print "Starting simulation..."

for i in range(n_steps):
    rates = [0.0, 0.0]
    if x_values[i] < 0:
        rates[1] = -x_values[i] / TAU_SYN * ratio
    else:
        rates[0] = x_values[i] / TAU_SYN * ratio
    input_source.set(rate=rates)

    sim.run(sim_time * 1000.0)
    data = A.get_data('spikes', clear=True).segments[0]
    spikes_A[i] = [len(train) / sim_time for train in data.spiketrains]

    print "Step {} complete.".format(i+1)

sim.end()

A_values = np.dot(spikes_A, decoders_A)

plt.figure()
plt.plot(x_values, A_values, label="A")
plt.plot(x_values, x_values, 'k-')
plt.plot(x_values, func_values, 'k-')
plt.legend()
plt.show()
```

``` python
# save_ensemble.py
# does not produce infinite loop error when invoked with seed 11245
import numpy as np
import pyNN.nest as sim
import sys

seed = int(sys.argv[1])

sim.setup(timestep=1.0, min_delay=1.0)
np.random.seed(seed)

n_neurons = 40
n_steps = 100

sim_time = 1.0

T_REF = 0.002
T_RC = 0.02
V_THRESH = 20.0
V_REST = -70.0
TAU_SYN = 0.001

RATE_LOW = 25
RATE_HIGH = 75

ratio = 1.0

x_values = np.linspace(-1, 1, n_steps)

encoders = np.random.choice([-1,1], n_neurons)
intercepts = np.random.uniform(-1, 1, n_neurons)
max_rates = np.random.uniform(RATE_LOW, RATE_HIGH, n_neurons)# * 1000.0/TAU_SYN

x = V_THRESH / (1 - np.exp((T_REF - (1.0 / max_rates)) / T_RC))
gain = (V_THRESH - x) / (intercepts - 1)
bias = V_THRESH - gain * intercepts

cellparams = {
    'tau_refrac':T_REF * 1000.0,
    'cm':T_RC * 1000.0 / gain,
    'v_rest':V_REST,
    'tau_syn_E':TAU_SYN * 1000.0,
    'tau_syn_I':TAU_SYN * 1000.0,
    'tau_m':T_RC * 1000.0,
    'i_offset':bias / gain,
    'v_thresh':V_REST + V_THRESH,
    'v_reset':V_REST
}

pop = sim.Population(n_neurons, sim.IF_curr_exp(**cellparams))
pop.initialize(v=V_REST)
pop.record('spikes')

input_source = sim.Population(2, sim.SpikeSourcePoisson())
weights = np.array([[1.0/ratio] * n_neurons * encoders] * 2)
weights[1] *= -1
sim.Projection(input_source, pop, sim.AllToAllConnector(),
    sim.StaticSynapse(weight=weights))

spikes = np.zeros((n_steps, n_neurons))

print "Network created, simulating..."

for i in range(n_steps):
    rates = [0.0, 0.0]
    if x_values[i] < 0:
        rates[1] = -x_values[i] / TAU_SYN * ratio
    else:
        rates[0] = x_values[i] / TAU_SYN * ratio
    input_source.set(rate=rates)

    sim.run(sim_time * 1000.0)
    data = pop.get_data('spikes', clear=True).segments[0]
    spikes[i] = [len(train) / sim_time for train in data.spiketrains]

sim.end()

print "Simulated. Saving..."

with open('ensemble_{}.npz'.format(seed), 'w') as dump:
    np.savez(dump, encoders=encoders, intercepts=intercepts,
        max_rates=max_rates, spikes=spikes)
```

Again, I apologise for posting a lot of code, it was really quite tricky to reproduce this error.

### LINKED ISSUE #1271: Attribution Error : module 'nest' has no attribute 'sli_push'

### LINKED ISSUE #1272: iaf_psc_alpha_ps model stuck when simulating
When running the `brunel_ps.sli` example with current master (after having changed `tau_syn` to `tau_syn_ex` and `tau_syn_in`), the example never finish simulating. The problem is the `iaf_psc_alpha_ps` model, which seems to be stuck in the loop in the `bisectioning_` function. If I try to use the `iaf_psc_alpha_canon` model, the example runs through without problems.

For a simpler test, one can also look at `test_mini_brunel_ps.sli`, which is currently using the `iaf_psc_alpha_canon` model. But if we instead use `iaf_psc_alpha_ps`, we again get stuck when simulating. I have tried to change the threshold in the `bisection_` function to `1e-6`, and this does not help. If I try to print a counter in the while loop, the counter keeps on increasing and it seems likely that this might be the problem and that we are not able to leave the loop.

stinebuu:
This PR fixes #1272 and fixes #403

The PR
- Replace the `bisectioning_` method with the [regula falsi](https://en.wikipedia.org/wiki/Regula_falsi) method (Illinois algorithm) in `iaf_psc_alpha_ps`, `iaf_psc_exp_ps`, `iaf_psc_exp_ps_lossless`.
- Fix membrane potential function for `iaf_psc_alpha_ps` so we are using the one for `alpha`.
- Switch to use the already defined `propagator_{31, 32}` where they were not already used.
- Use `std::exp()*x` instead `expm1()*x + x` throughout said models.


precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
 }
 
 double
-nest::iaf_psc_alpha_ps::bisectioning_( const double dt ) const
+nest::iaf_psc_alpha_ps::V_m_root_function_( double t_step ) const
```
heplesser: The function returns the difference of V_m and threshold, so I would call it `threshold_distance_`.

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
+
+  double V_m_root = ps_P30 * ( P_.I_e_ + V_.y_input_before_ ) + ps_P31_ex * V_.dI_ex_before_
+    + ps_P32_ex * V_.I_ex_before_ + ps_P31_in * V_.dI_in_before_ + ps_P32_in * V_.I_in_before_
+    + V_.V_m_before_ * std::exp( -t_step / P_.tau_m_ );
+
+  return V_m_root - P_.U_th_;
+}
+
+double
+nest::iaf_psc_alpha_ps::regula_falsi_method_( const double dt ) const
```
heplesser: Just `regular_falsi`.

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
+  int side = 0;
+
+  double a_k = 0.0;
+  double b_k = dt;
+
+  double V_m_func_a_k = V_m_root_function_( a_k );
+  double V_m_func_b_k = V_m_root_function_( b_k );
 
-  double V_m_root = V_.V_m_before_;
+  assert( V_m_func_a_k * V_m_func_b_k <= 0 );
```
heplesser: Maybe better to throw an exception is this fails. There could always be some strange corner case.

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
 
-  double div = 2.0;
+  int max_iter = 500;
```
heplesser: Should be a constant, with capital letters. Also the 1e-14 in the termination criterion.

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
 
-  while ( fabs( P_.U_th_ - V_m_root ) > 1e-14 )
+  for ( int iter = 0; iter < max_iter; ++iter )
   {
-    if ( V_m_root > P_.U_th_ )
-    {
-      root -= dt / div;
-    }
-    else
+    root = ( a_k * V_m_func_b_k - b_k * V_m_func_a_k ) / ( V_m_func_b_k - V_m_func_a_k );
```
heplesser: Do we need to protect against a zero denominator?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
+    V_m_root = V_m_root_function_( root );
```
heplesser: This is not V_m, but the difference from the threshold.

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
+
+    if ( std::abs( V_m_root ) <= 1e-14 )
```
heplesser: I would avoid `<=` if possible, a simple `<` is cleaner and has the same effect here. The limit value should be defined as named constant.

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -578,36 +575,78 @@ nest::iaf_psc_alpha_ps::emit_instant_spike_( Time const& origin, const long lag,
...
+  return V_m_root - P_.U_th_;
+}
+
+double
+nest::iaf_psc_alpha_ps::regula_falsi_method_( const double dt ) const
+{
+  double root;
+  double V_m_root;
+
+  int side = 0;
```
heplesser: I don't understand the logic behind "side" here and below, especially why it starts as zero and then will be -1 or +1 one forever as soon as it has taken on a different value than 0.  Could you explain this better? Possibly use an enumeration to have meaningful names instead of -1/0/+1?

precise/iaf_psc_alpha_ps.h
```diff
@@ -253,13 +253,20 @@ class iaf_psc_alpha_ps : public Archiving_Node
...
   };
 
+
   /**
-   * Localize threshold crossing by bisectioning.
+   * Difference between threshold and membrane potential for given time step.
+   * @param   double time step
+   * @returns difference between updated membrane potential and threshold
+   */
+  double V_m_root_function_( double t_step ) const;
```
heplesser: Name should express what the function returns.

precise/iaf_psc_alpha_ps.h
```diff
@@ -253,13 +253,20 @@ class iaf_psc_alpha_ps : public Archiving_Node
...
+
+  /**
+   * Localize threshold crossing by using Illinois algorithm of regula falsi method.
```
heplesser: Maybe add a reference to the algorithm, it may not be very well known.

precise/iaf_psc_exp_ps.cpp
```diff
@@ -540,36 +540,75 @@ nest::iaf_psc_exp_ps::emit_instant_spike_( const Time& origin, const long lag, c
...
+  const double P21_in = propagator_32( P_.tau_in_, P_.tau_m_, P_.c_m_, t_step );
+
+  double y2_root = P20 * ( P_.I_e_ + V_.y0_before_ ) + P21_ex * V_.y1_ex_before_ + P21_in * V_.y1_in_before_
+    + V_.y2_before_ * std::exp( -t_step / P_.tau_m_ );
+
+  return y2_root - P_.U_th_;
+}
+
+double
+nest::iaf_psc_exp_ps::regula_falsi_method_( const double dt ) const
```
heplesser: Instead of implementing this method for models that need it, could we implement it once i libnestutil? We would need to pass `V_m_root_function_` (with a better name) as a pointer-to-member-function, I think.

testsuite/unittests/test_iaf_ps_dc_accuracy.sli
```diff
@@ -340,8 +340,6 @@ V <- ( mV.) <- endl
 endl
 
 (               h in ms    ) <-
-(        alpha_canon [mV]) <-
```
heplesser: Why do you remove these lines?

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -361,6 +358,7 @@ nest::iaf_psc_alpha_ps::update( Time const& origin, const long from, const long
       // update membrane potential
       if ( not S_.is_refractory_ )
       {
+        // If we do not use V_.expm1_tau_m_*S_.V_m_ + S_.V_m_ here, the accuracy is a bit worse.
```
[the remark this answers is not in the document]
stinebuu: @heplesser This is a very long answer to your question, but the accuracy compared to a theoretical value is higher when using `exp()`. If we use `S_.V_m_ * std::exp( -V_.h_ms_ / P_.tau_m_ )` instead of `V_.expm1_tau_m_ * S_.V_m_ + S_.V_m_` two tests fail, `testsuite/unittests/test_iaf_ps_dc_t_accuracy.sli` and `testsuite/unittests/test_iaf_ps_dc_accuracy.sli` (the two tests I have changed below). The tests "compares spikes generated to the theoretical value for different computation step sizes and interpolation orders". `testsuite/unittests/test_iaf_ps_dc_t_accuracy.sli` writes out information, and if we use `exp()` instead of `expm1()` we get, for interpolation order 3
```
order: 3, exact value of membrane potential after 5 ms is 4.70003629245736 ms.

               h in ms              alpha_presc [ms]      error           [ms]          delta_canon [ms]        error         [ms]
--------------------------------------------------------------------------------------------------------------------------------
                       0          4.70003629245736      8.88178419700125e-16          4.70003629245736      8.88178419700125e-16  
                      -1          4.70003629245736                         0          4.70003629245736                         0  
                      -2          4.70003629245736                         0          4.70003629245736                         0  
                      -3          4.70003629245735      6.21724893790088e-15          4.70003629245736      8.88178419700125e-16  
                      -4          4.70003629245737      1.33226762955019e-14          4.70003629245735      4.44089209850063e-15  
                      -5          4.70003629245736      1.77635683940025e-15          4.70003629245736      8.88178419700125e-16  
                      -6          4.70003629245733      2.48689957516035e-14          4.70003629245735      1.77635683940025e-15  
                      -7          4.70003629245728      7.72715225139109e-14          4.70003629245736      5.32907051820075e-15  
                      -8          4.70003629245749      1.35891298214119e-13          4.70003629245736      8.88178419700125e-16  
                      -9          4.70003629245741      5.24025267623074e-14          4.70003629245736                         0  
                     -10          4.70003629245698      3.77475828372553e-13          4.70003629245736      5.32907051820075e-15  
                     -11          4.70003629245835       9.9475983006414e-13          4.70003629245736      2.66453525910038e-15  
                     -12          4.70003629245631      1.04805053524615e-12          4.70003629245736      5.32907051820075e-15  
                     -13          4.70003629246496      7.60724816473157e-12          4.70003629245733      2.48689957516035e-14  
                     -14          4.70003629246858      1.12256870465899e-11          4.70003629245738       2.1316282072803e-14 
```

When we use `expm1()` as we do now however, we get
```
order: 3, exact value of membrane potential after 5 ms is 4.70003629245736 ms.

               h in ms              alpha_presc [ms]      error           [ms]          delta_canon [ms]        error         [ms]
--------------------------------------------------------------------------------------------------------------------------------
                       0          4.70003629245736      8.88178419700125e-16          4.70003629245736      8.88178419700125e-16  
                      -1          4.70003629245736      8.88178419700125e-16          4.70003629245736                         0  
                      -2          4.70003629245736      8.88178419700125e-16          4.70003629245736                         0  
                      -3          4.70003629245735      1.77635683940025e-15          4.70003629245736      8.88178419700125e-16  
                      -4          4.70003629245735      2.66453525910038e-15          4.70003629245735      4.44089209850063e-15  
                      -5          4.70003629245735      1.77635683940025e-15          4.70003629245736      8.88178419700125e-16  
                      -6          4.70003629245735      2.66453525910038e-15          4.70003629245735      1.77635683940025e-15  
                      -7          4.70003629245736      6.21724893790088e-15          4.70003629245736      5.32907051820075e-15  
                      -8          4.70003629245736      8.88178419700125e-16          4.70003629245736      8.88178419700125e-16  
                      -9          4.70003629245736                         0          4.70003629245736                         0  
                     -10          4.70003629245736      5.32907051820075e-15          4.70003629245736      5.32907051820075e-15  
                     -11          4.70003629245736       3.5527136788005e-15          4.70003629245736      2.66453525910038e-15  
                     -12          4.70003629245736        7.105427357601e-15          4.70003629245736      5.32907051820075e-15  
                     -13          4.70003629245733      2.39808173319034e-14          4.70003629245733      2.48689957516035e-14  
                     -14          4.70003629245738      2.04281036531029e-14          4.70003629245738       2.1316282072803e-14  
```
So for lower `h` the error compared to a theoretical value is lower when using `expm1()`. I can still use `exp()`, but then I will need to change the tolerance in the two tests to make them pass.

Silmathoron:
Hi there, I'm sorry but I won't have time to review this in the foreseeable future.
Furthermore, I only implemented the previous version because nobody else could do it; I'm not used to working with propagators so I would have to redo all the math to check it, which is definitely not happening right now.
I would also like to remind you that I was explicitely asked to change the spike detection from interpolation to bisectioning in the original PR #1169 (see [here](https://github.com/nest/nest-simulator/pull/1169#issuecomment-499239564)), so I do not know why the change in method is now considered necessary, but I hope you made sure that this is indeed what is needed ^^

testsuite/unittests/test_iaf_ps_dc_accuracy.sli
stinebuu: `iaf_alpha_canon` was previously removed from the test as it is deprecated. I noticed the output was still here in the test when working on this PR because the test failed. I can add it again and remove it in another PR?

heplesser:
@Silmathoron Thanks for letting me know, I have assigned another reviewer. I had a look at the discussion in #1169 and I fear we got rather confused in that review (my fault and apologies!). Hanuschkin et al (2010) actually write in their paper that they skip use of bisection as suggested in earlier work by Brette, and just use Newton-Raphson. The "lossless" paper by Krishnan et al (2018) only deals with finding an interval and points out that one then needs to do some root-finding within that interval, but do not discuss any specific algorithm.

Experiments by @stinebuu in connection with this PR indicate that interpolation (regula falsi) requires about 10 times fewer steps than bisection.

testsuite/unittests/test_iaf_ps_dc_accuracy.sli
heplesser: No, fixing it here is fine.

precise/iaf_psc_alpha_ps.cpp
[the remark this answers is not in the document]
heplesser: @stinebuu Thanks for the details! I think we need to take a look at some aspects of that test. The model names in the table header seem wrong (could we take them directly from the list of models tested instead of hardcoding them?), the `iaf_psc_[alpha,delta]_ps` models have no parameter "interpolation order" any more (so no reason to loop over orders), and `iaf_psc_exp_ps` and `iaf_psc_exp_lossless` are not tested.

This leaves the curious matter that the presumably more accurate numerical expression leads to less accurate results. It would be nice to understand that, but that would be a follow-up issue. For now, let us stick to the  V expm1 + V, which gives best performance. That then also should be used in the function computing V during smaller steps and during interpolation. In the comment, maybe point explicitly to this test.

libnestutil/regula_falsi.h
```diff
@@ -0,0 +1,112 @@
...
+
+    if ( threshold_dist_a_k * threshold_dist_root > 0.0 )
+    {
+      // threshold_dist_a_k and threshold_dist_root have the same sign
+      a_k = root;
+      threshold_dist_a_k = threshold_dist_root;
+
+      if ( last_threshold_sign == 1 )
+      {
+        // If threshold_dist_a_k and threshold_dist_root had the same sign in the last time step, we half the value of
```
heplesser: I'd suggest to make the comment lines shorter to improve readibility, I think they should stay within the 80 char limit.

nestkernel/node.h
```diff
@@ -830,6 +830,13 @@ class Node
    */
   virtual index get_local_device_id() const;
 
+  /**
+   * Difference between threshold and membrane potential for given time step.
```
heplesser: Maybe add a note that this is used by precise neurons.

precise/iaf_psc_alpha_ps.h
```diff
@@ -172,6 +172,13 @@ class iaf_psc_alpha_ps : public Archiving_Node
   void get_status( DictionaryDatum& ) const;
   void set_status( const DictionaryDatum& );
 
+  /**
+   * Difference between threshold and membrane potential for given time step.
+   * @param   double time step
+   * @returns difference between updated membrane potential and threshold
+   */
```
heplesser: No need to repeat doxygen comment when overriding a virtual method.

libnestutil/regula_falsi.h
```diff
@@ -0,0 +1,112 @@
...
+ * See https://en.wikipedia.org/wiki/Regula_falsi#The_Illinois_algorithm for details on the algorithm.
+ *
+ * @param   Node   model object that call function
+ * @param   function threshold_distance function
+ * @param   double length of interval since previous event
+ * @returns time from previous event to threshold crossing
+ */
+template < typename CN >
+double
+regula_falsi( CN* node, double ( Node::*threshold_distance )( double ) const, const double dt )
```
heplesser: @stinebuu Nice, and apologies for having complicated things more than necessary, as I had forgotten about the details of pointer to member function (too much Python programming lately ...). In contrast to Python, C++ does not have bound member functions, so we need to point the node anyways, and then we could just pass the node:

```c++
template < typename CN >
double 
regula_falsi( const CN& node, double dt )
...
```

nestkernel/node.h
```diff
@@ -830,6 +830,13 @@ class Node
...
+   * @param   double time step
+   * @returns difference between updated membrane potential and threshold
+   */
+  virtual double threshold_distance( double t_step ) const;
```
heplesser: Actually, I wonder if we need `threshold_distance` as a virtual function in `Node` at all: Since `regula_falsi` has the node class as a template argument, it should work with all node classes providing `threshold_distance`, and for all others, it does not matter.

stinebuu: When doing the pointer to member this seemed necessary, but I will try now with just pointing to the node!

stinebuu:
@heplesser I have removed interpolation from `test_iaf_ps_dc_t_accuracy.sli` and changed the output to be based on the neurons used. If I add `iaf_psc_exp_ps` and `iaf_psc_exp_ps_lossless` to the test, it fails and again it is because of the difference in accuracy between `exp()*x` and `expm1()*x + x`. Do you want me to add the neurons to the tests, and switch back to using `expm1`? Or I can lower the tolerance, or do you want me to see if I can figure out why we have the change in accuracy?

precise/iaf_psc_alpha_ps.h
stinebuu: I have removed the virtual method.

libnestutil/regula_falsi.h
```diff
@@ -30,16 +30,16 @@ namespace nest
...
  *
  * @param   Node   model object that call function
- * @param   function threshold_distance function
  * @param   double length of interval since previous event
  * @returns time from previous event to threshold crossing
  */
 template < typename CN >
 double
-regula_falsi( CN* node, double ( Node::*threshold_distance )( double ) const, const double dt )
+regula_falsi( CN* node, const double dt )
```
heplesser: I think `const CN& node` and a corresponding `regula_falsi(*this)` would be better.

stinebuu: Done! :)

libnestutil/regula_falsi.h
```diff
@@ -39,7 +39,7 @@ namespace nest
  */
 template < typename CN >
 double
-regula_falsi( CN* node, const double dt )
+regula_falsi( CN& node, const double dt )
```
heplesser, line 42: Can't this be a `const` reference? I don't think we change the `node` in any way?

stinebuu: Sorry, I was a bit quick and forgot the `const`.

heplesser:
@clinssen Could you take a look in the near future?

precise/iaf_psc_exp_ps.h
```diff
@@ -164,6 +164,13 @@ class iaf_psc_exp_ps : public Archiving_Node
   void get_status( DictionaryDatum& ) const;
   void set_status( const DictionaryDatum& );
 
+  /**
+   * Difference between threshold and membrane potential for given time step.
```
clinssen, line 168:
```suggestion
   * Based on the current state, compute the value of the membrane potential after taking a timestep of length
   * ``t_step``, and use it to compute the signed distance to spike threshold at that time. The internal state is not
   * actually updated (method is defined const).
```

precise/iaf_psc_alpha_ps.cpp
```diff
@@ -500,37 +503,32 @@ nest::iaf_psc_alpha_ps::handle( DataLoggingRequest& e )
...
-      - dt * V_.gamma_in_ * ps_e_TauSyn_in - dt * V_.gamma_in_;
-    const double ps_P32_in = V_.gamma_in_ * ps_e_Tau - V_.gamma_in_ * ps_e_TauSyn_in;
+    const double ps_P31_ex = propagator_31( P_.tau_syn_ex_, P_.tau_m_, P_.c_m_, dt );
+    const double ps_P32_ex = propagator_32( P_.tau_syn_ex_, P_.tau_m_, P_.c_m_, dt );
+    const double ps_P31_in = propagator_31( P_.tau_syn_in_, P_.tau_m_, P_.c_m_, dt );
+    const double ps_P32_in = propagator_32( P_.tau_syn_in_, P_.tau_m_, P_.c_m_, dt );
 
     S_.V_m_ = ps_P30 * ( P_.I_e_ + S_.y_input_ ) + ps_P31_ex * S_.dI_ex_ + ps_P32_ex * S_.I_ex_ + ps_P31_in * S_.dI_in_
-      + ps_P32_in * S_.I_in_ + ps_e_Tau * S_.V_m_ + S_.V_m_;
+      + ps_P32_in * S_.I_in_ + S_.V_m_ * std::exp( -dt / P_.tau_m_ );
```
clinssen, line 517: Can we rewrite this expression to its equivalent expm1 form for better floating-point precision?

heplesser:
@clinssen 
> * What happens to "canon" models?

There is only `iaf_psc_alpha_canon` left, for historical reasons. #1493 will follow that up.

> * Original `test_iaf_ps_dc_t_accuracy.sli` tests for different interpolation orders, why not any longer?

When we first introduced precise models around 2007, the time of threshold crossing was estimated by linear, quadratic or cubic interpolation and the user could select interpolation order as a model parameter.  We switched to iterative methods for threshold-crossing detection quite some time ago, eliminated the interpolation order parameter and never properly updated the tests.

testsuite/unittests/test_iaf_ps_dc_t_accuracy.sli
```diff
@@ -133,125 +131,78 @@ params begin userdict begin
...
-  Flatten {tolerance lt} Map
- } Map
-
- Flatten true exch {and} Fold
+  cout default 15 setprecision 
+
+  endl endl
+  (The precise spike time is ) <- t <- ( ms.) <- endl endl
+
+  (                  h [ms]) <-
```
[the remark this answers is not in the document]
heplesser: You are right! The table header here should really read `log_2 h` and for simplicities sake I would then just ignore the fact that `h` has units; writing the entirely correct `log_2(h / 1ms)` just seems too confusing. And this is test output ...

precise/iaf_psc_alpha_ps.cpp
heplesser: We have this equation essentially in three places: lines 364f, 516f and 588ff. We should use the same implementation everywhere. In the `test_iaf_psc_dc...accurary.sli` tests, the `V * expm1() + V` curiously leads to smaller errors than `V * exp()` in the presence of pure DC input; so it may be the better choice.

clinssen:
Thanks for the clarifications! In this case, additionally, please remove ``enum interpOrder`` from ``iaf_psc_alpha_ps.h`` and the comments block in front of it.

libnestutil/regula_falsi.h
```diff
@@ -0,0 +1,113 @@
...
+      {
+        threshold_dist_a_k /= 2;
+      }
+      last_threshold_sign = -1;
+    }
+    else
+    {
+      throw NumericalInstability( "regula_falsi: Regula falsi method did not converge" );
+    }
+  }
```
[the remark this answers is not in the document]
stinebuu: Good catch, we would have returned the `root` for the last iteration. I have added a `NumericalInstability`. There are now two right after each other, but I couldn't think of a better way of doing it.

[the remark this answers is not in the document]
heplesser: @stinebu Doesn't the for-loop terminate when `iter == MAX_ITER`? And wouldn't it make the code even clearer to move the `return root;` directly into the `if` on line 72 above instead of `break`, and then just throw the exception here at the end unconditionally?

clinssen:
There are still some mentions of "bisectioning" in ``iaf_psc_exp_ps.h``.

PR #1482, opened 2020-03-24
Remove SubnetExpected exception and other references to subnet.

stinebuu:
As subnets are removed the `SubnetExpected` exception is no longer needed.

### DIFF
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -647,7 +647,7 @@ NestModule::CopyModel_l_l_DFunction::execute( SLIInterpreter* i ) const
 }
 
 /** @BeginDocumentation
-   Name: Create - create a number of equal nodes in the current subnet
+   Name: Create - create nodes
 
    Synopsis:
    /model          Create -> node_ids
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -650,10 +650,10 @@ NestModule::CopyModel_l_l_DFunction::execute( SLIInterpreter* i ) const
    Name: Create - create nodes
 
    Synopsis:
-   /model          Create -> node_ids
-   /model n        Create -> node_ids
-   /model   params Create -> node_ids
-   /model n params Create -> node_ids
+   /model          Create -> NodeCollection
+   /model n        Create -> NodeCollection
+   /model   params Create -> NodeCollection
+   /model n params Create -> NodeCollection
 
    Parameters:
    /model - literal naming the modeltype (entry in modeldict)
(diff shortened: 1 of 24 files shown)


stinebuu:
When removing `SubnetExpected` I also noticed that there where several other references to `subnet` so I removed them as well.

There is still a couple of references to `subnet` in `doc\tutorials\music_tutorial\music_tutorial_sli.rst`, but it looks like the whole music tutorial needs updating, it uses `LayoutNetwork`, `GetGlobalNodes` etc. I have therefore not removed the references to `subnet` from the document because I thought the update of music tutorial documentation should be it's own issue and PR. I can of course update the documents here as well if that is better. Do you have any thoughts on the best way to proceed, @sarakonradi?

heplesser:
@stinebuu I'd suggest to create a separate issue & PR for updating the MUSIC tutorial.

nestkernel/nestmodule.cpp
```diff
@@ -647,7 +647,7 @@ NestModule::CopyModel_l_l_DFunction::execute( SLIInterpreter* i ) const
 }
 
 /** @BeginDocumentation
-   Name: Create - create a number of equal nodes in the current subnet
+   Name: Create - create nodes
 
    Synopsis:
    /model          Create -> node_ids
```
heplesser: Isn't the return value here a NodeCollection?

pynest/nest/tests/test_connect_array_fixed_indegree.py
```diff
@@ -36,16 +36,16 @@ class ConnectArrayFixedIndegreeTestCase(unittest.TestCase):
...
         K = 5   # number of connections per neuron
 
         ############################################
         # test with connection rule fixed_indegree
         ############################################
         nest.ResetKernel()
 
-        net1 = nest.Create('iaf_psc_alpha', N)  # creates source subnet
-        net2 = nest.Create('iaf_psc_alpha', N)  # creates target subnet
+        net1 = nest.Create('iaf_psc_alpha', N)  # creates source nodes
```
heplesser: How about `# create source population` and correspondingly in what follows?

nestkernel/target_table_devices.cpp
```diff
@@ -139,7 +139,7 @@ nest::TargetTableDevices::get_connections_to_device_for_lid_( const index lid,
   if ( target_to_devices_[ tid ][ lid ].size() > 0 )
   {
     const index source_node_id = kernel().vp_manager.lid_to_node_id( lid );
-    // not the root subnet and valid connector
+    // not valid connector
```
sarakonradi: ``not the valid connector``

pynest/nest/tests/test_connect_array_fixed_outdegree.py
```diff
@@ -55,10 +55,10 @@ def test_Connect_Array_Fixed_Outdegree(self):
                     'weight': Warr, 'delay': Darr}
         conn_dict = {'rule': 'fixed_outdegree', 'outdegree': K}
 
-        # connects source to target subnet
+        # connects source to target
         nest.Connect(net1, net2, conn_spec=conn_dict, syn_spec=syn_dict)
 
-        for i in range(N):  # loop on all neurons of source subnet
+        for i in range(N):  # loop on all source neurons of
```
sarakonradi: Remove ``of``

nestkernel/nestmodule.cpp
stinebuu: Good catch!

PR #1490, opened 2020-03-28
Update readthedocs style sheet to improve text contrast

clinssen:
Currently, the text inside the "ethical advertisement" box on https://nest-simulator.readthedocs.io/ is unreadable, because the text colour matches the colour of the background. This PR sets the advertisement container's background colour to the same dark grey as the version button on the bottom left.

Also minor cleanup of the CSS file.


doc/_static/css/custom.css
```diff
@@ -1,3 +1,7 @@
+.keep-us-sustainable {
+   background-color: #272525;
+}
```
steffengraber, line 3: @clinssen I cannot find the class `keep-us-sustainable` in the source code of the RTD site. What exactly is being changed here?

clinssen: Perhaps your ad blocker is blocking you from seeing it? Below the menu on the left, there is a box containing the text "Support Read the Docs! Please help keep us sustainable [...]"

If you inspect the element in your browser, you can see the class:
```html
<div id="rtd-c8kphpaw" class="keep-us-sustainable">
```

clinssen: Just to clear this up: the element with class "keep-us-sustainable" is rendered in particular *when* using an ad blocker. A sigificant portion of our visitors are probably running an ad blocker, and this box contains some kind words by ReadTheDocs to allow their ethical ads, so we should probably allow it to render properly. (Curious that they don't prevent us to set display: none on it... of course that would be unethical.)

Issue #1471, opened 2020-03-17
Basic Instrumentation


terhorstd:
**Is your feature request related to a problem? Please describe.**

This requested feature solves two problems. First, in the research on simulation technology carried out with NEST and the continuous improvement of NEST it is necessary to measure execution times of different phases of the simulation. Across projects it is difficult to communicate how exactly these times are determined and every project has to re-implement the instrumentation. This is inefficient and limits reproducibility. Second, the NEST team regularly receives support requests where NEST is reported to be slow. The timing data provided by the client will speed determination of the origin of the problem.

## Interface present status

At present `GetKernelStatus` gives the following timing information:

- `time`
- `time_collocate`
- `time_communicate`

These variables are no longer filled consistently in the present master.

## Interface desired situation

Proposal of different status dictionary entry entries providing timing information:

* `biological_time` : biological time simulated since the last `ResetKernel` (This is not based on stopwatches but needed to calculate the real-time factor)

These *basic timers* do not overlap and should usually not impact performance:

* `time_construction` : time for constructing the network in the nest kernel, cumulative build time for neurons and connections
* `time_communicate_prepare` : setup of presynaptic connection infrastructure (before actual simulation)
* `time_simulate` : total time of simulation without construction

These *detailed timers* contribute to `time_simulate`:

* `time_communicate` : time for communicating spikes between compute nodes
* `time_update` : time for neuron update
* `time_deliver` : time to deliver events from the MPI recv buffers to their
  local synaptic targets (including synaptic update, e.g. STDP synapses) and to
  the spike ring buffers of the corresponding postsynaptic neurons
* `time_collocate` : time to collocate MPI send buffer from spike register

These timers, except the first cumulating the biological time, should be reset at the start of the program/`ResetKernel` (for network construction) and `Simulate` (for timers concerned with dynamics). The timers that may negatively impact performance (e.g. `time_communicate` and `time_deliver`) should only be active if the compiler flag is set.

The proposed name of the compiler option is `--with-basic-timers` (may become default) and `--with-detailed-timers`.

## Implementation

Use the POSIX timing commands already built into NEST which are the basis of SLI `tic` and `toc` commands and the existing `StopWatch` class. Document the compiler options and the interface in the user-level documentation.

## Reasons

0. increase reproducibility of timing measurements, ease of use
[...]

PR #1373, opened 2019-12-20
Doc kernelstatus

morales-gregorio:
This is the PR trying to solve Issue #1350

I have looked into `nest.help("kernel")` and I found a few issues.
* `initial_connector_capacity`, `large_connector_limit`, `large_connector_growth_factor` are present in the help documentation but they do not seem to be present in the kernel status dictionary from NEST >=2.14. I have removed them from the list.
* `grng_seed` and `rng_seed` are specified as write only, but the GetKernelStatus() does return them, they can also be read (I removed the write only tag from their description).
* There is a long list of keys that are present in the kernel dictionary (NEST 2.18.0) and not documented in `nest.help("kernel")`:
```
 ['adaptive_spike_buffers',
 'adaptive_target_buffers',
 'buffer_size_secondary_events',
 'buffer_size_spike_data',
 'buffer_size_target_data',
 'growth_factor_buffer_spike_data',
 'growth_factor_buffer_target_data',
 'keep_source_table',
 'local_spike_counter',
 'max_buffer_size_spike_data',
 'max_buffer_size_target_data',
 'max_num_syn_models',
 'sort_connections_by_source',
 'structural_plasticity_synapses',
 'structural_plasticity_update_interval',
 'time_collocate',
 'time_communicate']
``` 
I have added them, their types and whether they are read only or can be read and written. However their descriptions should  be filled before this PR can be merged to master. (Marked the descriptions with TODO flags)

Best,
Aitor

PR #1491, opened 2020-03-30
Change definition of realtime factor in print_time

jhnnsnk:
If `print_time` was enabled, the realtime factor was printed out as "biological time simulated / wall-clock time".
However, the inverse definition, "wall-clock time / biological simulation time" is currently used in benchmark measurements of the state propagation.
The latter definition is beneficial because the factor then decreases if the simulation becomes faster.
This PR changes the definition accordingly.

In addition, this PR adds a new section "Monitoring elapsed time" to the documentation guide `running_simulations.rst`.

Note that the realtime factor is not the "speedup" often used in the context of computer architectures.
The new documentation distinguishes between the two.

Let's discuss the following points:
1. Do we prefer to write "realtime factor" or "real time factor"?
--> The result of the discussion is "real-time factor".
2. We currently define the factor in symbols by  `T_sim / T_bio` which might be misleading because the argument to the `Simulate()` call is often named `T_sim` or similar but in fact we refer to the biological time.
--> The result of the discussion is to use `T_wall / T_model`.

It would be great if @sarakonradi could have a look.

### DIFF
--- a/doc/guides/running_simulations.rst
+++ b/doc/guides/running_simulations.rst
@@ -210,8 +210,8 @@ The progress of the simulation can be monitored by setting:
     SetKernelStatus({"print_time": True})
 
 If enabled, a line is printed to screen at every time step of the simulation to
-track the percentage and the absolute elapsed biological simulation time as
-well as the real-time factor, for example:
+track the percentage, the absolute elapsed biological simulation time and the
+real-time factor, for example:
 
 ::
 
@@ -226,11 +226,11 @@ argument to the ``Simulate()`` call):
     q_\text{real} = \frac{T_\text{wall}}{T_\text{sim}}
 
 If the real-time factor is larger than `1` as in the example above, the
-simulation runs more slowly than the wall-clock time passes.
+simulation runs slower than the wall-clock time.
 
-In the case that a simulation script contains multiple ``Simulate()`` calls,
+In case a simulation script contains multiple ``Simulate()`` calls,
 the percentage simulation time is reset to `0%` at the beginning of each call,
-but the abolute simulation time and the real-time factor account for the total
+but the absolute simulation time and the real-time factor account for the total
 elapsed times.
 
 .. note::
--- a/doc/guides/running_simulations.rst
+++ b/doc/guides/running_simulations.rst
@@ -233,6 +233,16 @@ the percentage simulation time is reset to `0%` at the beginning of each call,
 but the absolute simulation time and the real-time factor account for the total
 elapsed times.
 
+The real-time factor should not be confused with the concept of *speedup*.
+Speedup refers to a ratio of wall-clock times, namely the wall-clock time
+needed to solve a problem serially and the wall-clock time needed to solve the
+same problem in parallel (e.g., by distributing the work across multiple
+threads or processes):
+
+.. math::
+
+    q_\text{speedup} = \frac{T_\text{wall, serial}}{T_\text{wall, parallel}}
+
 .. note::
 
     For large, distributed simulations, it is recommended to set
--- a/doc/guides/running_simulations.rst
+++ b/doc/guides/running_simulations.rst
@@ -217,16 +217,16 @@ well as the real-time factor, for example:
 
     [ 25% ] Biological simulation time: 250.0 ms, Real-time factor: 2.6711
 
-The *real-time factor* is defined as the quotient of *real time* (which is also
-known as wall-clock time) and the *biological simulation time* (which is the
+The *real-time factor* is defined as the quotient of *wall-clock time* (which
+is also known as real time) and the *biological simulation time* (which is the
 argument to the ``Simulate()`` call):
 
 .. math::
 
-    q_\text{real} = \frac{T_\text{real}}{T_\text{sim}}
+    q_\text{real} = \frac{T_\text{wall}}{T_\text{sim}}
 
 If the real-time factor is larger than `1` as in the example above, the
-simulation runs more slowly than real time passes.
+simulation runs more slowly than the wall-clock time passes.
 
 In the case that a simulation script contains multiple ``Simulate()`` calls,
 the percentage simulation time is reset to `0%` at the beginning of each call,
@@ -237,6 +237,7 @@ elapsed times.
 
     For large, distributed simulations, it is recommended to set
     ``{"print_time": False}`` to avoid the overhead of the print calls.
-    In these cases, the real-time factor can be computed by measuring the real
-    time manually and dividing by the set biological simulation time.
+    In these cases, the real-time factor can be computed by measuring the
+    wall-clock time manually and dividing by the set biological simulation
+    time.
 
--- a/doc/guides/running_simulations.rst
+++ b/doc/guides/running_simulations.rst
@@ -199,3 +199,44 @@ a Poisson spike train using different seeds and output files for each run:
         nest.Connect(nrn, sd)
     
         nest.Simulate(100)
+
+Monitoring elapsed time
+-----------------------
+
+The progress of the simulation can be monitored by setting:
+
+::
+
+    SetKernelStatus({"print_time": True})
+
+If enabled, a line is printed to screen at every time step of the simulation to
+track the percentage and the absolute elapsed biological simulation time as
+well as the real-time factor, for example:
+
+::
+
+    [ 25% ] Biological simulation time: 250.0 ms, Real-time factor: 2.6711
+
+The *real-time factor* is defined as the quotient of *real time* (which is also
+known as wall-clock time) and the *biological simulation time* (which is the
+argument to the ``Simulate()`` call):
+
+.. math::
+
+    q_\text{real} = \frac{T_\text{real}}{T_\text{sim}}
+
+If the real-time factor is larger than `1` as in the example above, the
+simulation runs more slowly than real time passes.
+
+In the case that a simulation script contains multiple ``Simulate()`` calls,
+the percentage simulation time is reset to `0%` at the beginning of each call,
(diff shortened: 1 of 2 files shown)


heplesser:
@jhnnsnk I noticed that this PR does not include any update to user documentation. Is this factor described properly anywhere? Your suggestion seems to make sense, since most benchmarks plot wallclock time per simulated biological second. And I suppose we need to fix the text printed by SLI when reporting the factor?

jhnnsnk:
@heplesser So far the factor does not seem to be documented anywhere. I could propose some lines and add them to this PR. In addition to a definition, I assume that the documentation should say that the `print_time` option itself may slow down the simulation. For proper (and in particular parallel) benchmarks it should be switched off and the factor should be computed manually.

The printed text is already generated on the C++ level by the function `print_progress_()` in  `simulation_manager.cpp`, just below the change I made here. So the text is the same for PyNEST and SLI. Or do I misunderstand the question?

sarakonradi:
@jhnnsnk How about ``real-time factor``?

It would be great if we could include this factor in the user documentation. The question is: Where would it fit best?

jhnnsnk:
@sarakonradi I'd suggest to add a short section to [Running simulations](https://nest-simulator.readthedocs.io/en/latest/guides/running_simulations.html). Another option would be to put it in the docstring of `SetKernelStatus()` of the PyNEST API because that's where `print_time` is set. But the other entries of the kernel status dictionary are also not documented, yet, which is why I would prefer the former. What do you think?

sarakonradi:
@jhnnsnk That is a good idea. We are actually working on the kernel status documentation at the moment, and there will be a PR for that soon. I think it would make sense to add a short section to Running simulations regardless.

jhnnsnk:
In my latest commit, I edited the line to be printed and added a section "Monitoring elapsed time" to the "Running simulations" guide for documentation. I changed the spelling to "real-time factor" as suggested by @sarakonradi. Please check in particular the symbols used (e.g., I suggested `q_real` for the real-time factor). I distinguished now between biological simulation time `T_sim` and real time `T_real`, inspired by [Wikipedia](https://en.wikipedia.org/wiki/Elapsed_real_time). What do you think?

clinssen:
Hi, I would have a slight preference for "wall clock time", as it is unambiguous—as opposed to "real time", which brings up [all sorts of questions about general relativity and quantum mechanics](https://en.wikipedia.org/wiki/Problem_of_time) (or, falling shy of that, possibly just confuses with "real time computing" terminology).

jhnnsnk:
@clinssen That is a good point, thanks. I changed accordingly and now use the symbol `T_wall` for the "wall-clock time".

doc/guides/running_simulations.rst
```diff
@@ -199,3 +199,45 @@ a Poisson spike train using different seeds and output files for each run:
...
+
+The progress of the simulation can be monitored by setting:
+
+::
+
+    SetKernelStatus({"print_time": True})
+
+If enabled, a line is printed to screen at every time step of the simulation to
+track the percentage and the absolute elapsed biological simulation time as
+well as the real-time factor, for example:
```
sarakonradi, line 214: I would add a comma to list these three items, i.e. ``to track the percentage, the absolute elapsed biological simulation time and the real-time factor, for example:``

doc/guides/running_simulations.rst
```diff
@@ -199,3 +199,45 @@ a Poisson spike train using different seeds and output files for each run:
...
+The *real-time factor* is defined as the quotient of *wall-clock time* (which
+is also known as real time) and the *biological simulation time* (which is the
+argument to the ``Simulate()`` call):
+
+.. math::
+
+    q_\text{real} = \frac{T_\text{wall}}{T_\text{sim}}
+
+If the real-time factor is larger than `1` as in the example above, the
+simulation runs more slowly than the wall-clock time passes.
```
sarakonradi, line 229: Perhaps ``above, then the simulation runs slower than the wall-clock time.``?

doc/guides/running_simulations.rst
```diff
@@ -199,3 +199,45 @@ a Poisson spike train using different seeds and output files for each run:
...
+
+In the case that a simulation script contains multiple ``Simulate()`` calls,
```
sarakonradi, line 231: You could shorten this to ``In case a simulation script contains``

doc/guides/running_simulations.rst
jhnnsnk: Initially, I intended "percentage" to be an adjective and therefore chose the previous formulation. But your suggestion is probably easier to understand and I changed accordingly.

doc/guides/running_simulations.rst
jhnnsnk: Ok.

doc/guides/running_simulations.rst
jhnnsnk: Ok.

jhnnsnk:
@diesmann 
> T_sim is still ambiguous as pointed out by HEP before. I like T_wall best to avoid the discussed confusion with real-time performance and will make sure that we improve consistency also in our benchmark plots.

Please check if you are fine with the latest definition of real-time factor in the guide. We now use `T_wall/T_sim` where `T_sim` refers to the biological simulation time.

> In this regard the documentation would profit from a few sentences on the definition of speed-up as T_wall(M)/T_wall(serial) where M is the degree of parallelization in threads or processes.

I added a definition of speedup to the documentation but put the serial part in the nominator. Please check if you are fine with this.

jhnnsnk:
After an offline discussion with @diesmann, I changed `T_sim` to `T_bio` in the documentation according to his suggestion.  `T_sim` can of course be used for naming a parameter (e.g., the argument to the `Simulate()` call), but as "sim" remains ambiguous, we should use `T_bio` consistently for documentation and for presenting results of benchmark simulations.

heplesser:
@jhnnsnk @diesmann Is `T_bio` really the best choice? After all, NEST is used not only to simulate biological brains, but also in neurorobotics and abstract spike-based learning. What would you think about `T_model`?

diesmann:
I am also happy with this more abstract reference to the time used in the model. In this case I suggest T_mod for brevity because this will often be used in axes labels like T_wall/T_mod. In any case this decision concerns a lot of documentation and code, so we need to find a way to end this discussion.

suku248:
`T_sim` has been used so far (also in manuscripts) to refer to the wall-clock time that a simulation required as opposed to the wall-clock time that the preceding build phase required. I think `T_wall` works well in this context here but in a more general context it is also ambiguous as it's not clear whether it refers to build phase, init phase or actual simulation. What about `T_wall_sim`?

albada:
I favor T_model over T_mod, because the latter intuitively refers to 'modified' for me.

diesmann:
valid point

jhnnsnk:
@diesmann @albada @heplesser @sarakonradi As we now converge towards `T_model` (which I also like), we should change the verbal definition, too. Do you prefer, e.g., "model simulation time" or "simulated model time" or "model time simulated" or anything else?

jhnnsnk:
@suku248 I agree that "wall-clock time" is a very general term, but for this context here I would prefer to keep it as simple as possible. If a study requires the symbol to be more specific for differentiation, I think that it is a good idea to extend the subscript (or add a superscript) accordingly. Similarly, one could discuss whether the `wall` is actually needed for the build time because a `model` time does not exist.

diesmann:
I agree with @jhnnsnk here. We also need to see this discussion in the context of #1471 and #1373. These parts of the documentation and parameter naming should be consistent. In #1471 this works nicely by calling biological_time model_time and renaming time_simulate to time_propagation or time_state_propagation (the time referred to here is without construction time). In this way we can get rid of the overloaded notion of "simulation time" altogether.

albada:
> @diesmann @albada @heplesser @sarakonradi As we now converge towards `T_model` (which I also like), we should change the verbal definition, too. Do you prefer, e.g., "model simulation time" or "simulated model time" or "model time simulated" or anything else?

Out of those, I definitely prefer the first one.

diesmann:
I think such three word terms often leave ambiguity in where the strongest binding is, at least for the non-native speaker. In the sense of barrier-free language, why not looking for a longer phrase like "duration by which the state of the model is advanced in time"?

clinssen:
> @diesmann @albada @heplesser @sarakonradi As we now converge towards `T_model` (which I also like), we should change the verbal definition, too. Do you prefer, e.g., "model simulation time" or "simulated model time" or "model time simulated" or anything else?

Anything with "simulated time" would be the most clear for me. "Model simulation time" is again ambigious with respect to simulated vs. wall-clock time.

diesmann:
anything with "simulated" has the danger of being unclear because it will always have this recursive aspect that the term is part of what we are trying to define

jhnnsnk:
To wrap this discussion up, my last commit changes the symbol to `T_model` and uses @diesmann's suggestion as long verbal definition and just "model time" as short definition in the documentation (the word "simulation" or "simulated" is now avoided).

jhnnsnk:
@sarakonradi @suku248 @albada From my side, everything seems to be done now. Could you please review?

diesmann:
ok

heplesser:
@jhnnsnk @lekshmideepu Can either one of you find a way on Travis to trigger a build for this PR?

lekshmideepu:
> @jhnnsnk @lekshmideepu Can either one of you find a way on Travis to trigger a build for this PR?
@heplesser 
triggerbuild

lekshmideepu:
Just restarted the build which failed some days ago due to the temporary issue in installing the libraries. It looks fine now.

PR #1497, opened 2020-04-03
Allow direct conversion from NodeCollection to numpy

Silmathoron:
This PR enables direct conversion of ``NodeCollection`` objects to numpy arrays through e.g.

    arr = np.array(gids)
    arr[:len(gids)] = gids


hakonsbm:
Thanks for your interest in improving `NodeCollection`s, @Silmathoron! However, I don't really see the use for this, as you can already convert `NodeCollection`s to NumPy arrays easily with `np.array(gids.tolist())`. Additionally, the direct conversion is quite a bit slower than using `tolist()`:
```python
>>> gids = nest.Create('iaf_psc_alpha', 10000)

>>> %timeit np.array(gids.tolist())
83.6 ms ± 5.32 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

>>> %timeit np.array(gids)
1.11 s ± 6.61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Silmathoron:
The idea was to make conversion more seamless, but thanks for checking the time, I did not expect it to be so slow... I'll try to figure out if I can make this faster!

Silmathoron:
@hakonsbm it took me surprisingly long to find it but it turns out all that's needed is an ``__array__`` method ^^"
Should be much better now, sorry for the first ugly hack!

pynest/nest/lib/hl_api_types.py
```diff
@@ -268,16 +267,22 @@ def __eq__(self, other):
...
             raise NotImplementedError()
+
         return not self == other
 
     def __len__(self):
         return sli_func('size', self._datum)
 
+    def __int__(self):
+        # this will automatically raise an error if len is not 1
+        return int(self.get('global_id'))
```
hakonsbm, line 284: Is this function still needed?

pynest/nest/lib/hl_api_types.py
```diff
@@ -442,16 +450,27 @@ def index(self, node_id):
...
         """
         index = sli_func('Find', self._datum, node_id)
+
         if index == -1:
             raise ValueError('{} is not in NodeCollection'.format(node_id))
+
         return index
 
+    def __array__(self, dtype=None):
+        ''' Numpy compatibility function '''
```
hakonsbm, line 460: The docstring should describe what the function does and it should be in triple double quotes, e.g.:
```suggestion
        """Convert the NodeCollection to a NumPy array."""
```

pynest/nest/lib/hl_api_types.py
```diff
@@ -442,16 +450,27 @@ def index(self, node_id):
...
     def __getattr__(self, attr):
         if attr == 'spatial':
             metadata = sli_func('GetMetadata', self._datum)
             val = metadata if metadata else None
             super().__setattr__(attr, val)
             return self.spatial
+
+        # numpy compatibility check
+        if attr.startswith('__array_'):
+            raise AttributeError
```
hakonsbm, line 472: Please add a comment on what this actually does.

pynest/nest/lib/hl_api_types.py
Silmathoron: No, but since it was written, I left it... should I remove it?

hakonsbm: If it has nothing to do with the NumPy array conversion, it should be removed in this PR.

pynest/nest/lib/hl_api_types.py
Silmathoron: done

pynest/nest/lib/hl_api_types.py
```diff
@@ -442,16 +446,30 @@ def index(self, node_id):
...
         """
         index = sli_func('Find', self._datum, node_id)
+
         if index == -1:
             raise ValueError('{} is not in NodeCollection'.format(node_id))
+
         return index
 
+    def __array__(self, dtype=None):
+        """ Convert the NodeCollection to a NumPy array. """
```
hakonsbm, line 456: One-line docstrings should have no blank lines either before or after the docstring (see https://www.python.org/dev/peps/pep-0257/#one-line-docstrings).
```suggestion
        """Convert the NodeCollection to a NumPy array."""
```

pynest/nest/lib/hl_api_types.py
```diff
@@ -442,16 +446,30 @@ def index(self, node_id):
...
+
     def __getattr__(self, attr):
         if attr == 'spatial':
             metadata = sli_func('GetMetadata', self._datum)
             val = metadata if metadata else None
             super().__setattr__(attr, val)
             return self.spatial
+
+        # Numpy compatibility check:
+        # raises AttributeError to tell numpy that interfaces other than
```
hakonsbm, line 467:
```suggestion
        # NumPy compatibility check:
        # raises AttributeError to tell NumPy that interfaces other than
```

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -69,6 +69,23 @@ def test_list_to_NodeCollection(self):
         nc = nest.NodeCollection(node_ids_in)
         self.assertEqual(nc.tolist(), [2, 3, 5, 7, 8])
 
+    def test_NodeCollection_to_numpy(self):
+        """ Conversion from NodeCollection to numpy array """
```
hakonsbm, line 73:
```suggestion
        """Conversion from NodeCollection to NumPy array"""
```

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -69,6 +69,23 @@ def test_list_to_NodeCollection(self):
...
+        if HAVE_NUMPY:
+            n_neurons = 10
+            nc = nest.Create('iaf_psc_alpha', n_neurons)
+
+            # direct array conversion
+            n_arr = np.array(nc)
+
+            self.assertTrue(np.array_equal(n_arr, nc.tolist()))
```
hakonsbm, line 81: To avoid the unhelpful `False is not True` message in the event of failure, either convert the NumPy array to a list and use `self.assertEqual()`, or use [`numpy.testing.assert_equal()`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html#numpy-testing-assert-equal).

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -69,6 +69,23 @@ def test_list_to_NodeCollection(self):
...
+
+            # incorporation to bigger array
+            arr = np.zeros(2*n_neurons, dtype=int)
+            arr[2:12] = nc
+
+            self.assertTrue(np.array_equal(arr[2:12], nc.tolist()))
```
hakonsbm, line 87: Same as above.

pynest/nest/tests/test_NodeCollection.py
Silmathoron, line 81:
```suggestion
            self.assertEqual(n_arr.tolist(), nc.tolist())
```

pynest/nest/tests/test_NodeCollection.py
Silmathoron, line 87:
```suggestion
            self.assertEqual(arr[2:12].tolist(), nc.tolist())
```

Silmathoron:
I applied the changes but for some reason it seems that the test do not start when committing is done from GitHub, @hakonsbm could you restart the tests?

hakonsbm:
@Silmathoron Looks like it's not detected by GitHub, but it is running (it's a nice round build number too :smile:)
https://travis-ci.org/github/nest/nest-simulator/builds/695050000

Silmathoron:
@hakonsbm tests passed so when this is merged I think we can incorporate it to #1539, unblock it, and start the review of the new connect behavior :)

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -69,6 +69,23 @@ def test_list_to_NodeCollection(self):
...
+            nc = nest.Create('iaf_psc_alpha', n_neurons)
+
+            # direct array conversion
+            n_arr = np.array(nc)
+
+            self.assertEqual(n_arr.tolist(), nc.tolist())
+
+            # incorporation to bigger array
+            arr = np.zeros(2*n_neurons, dtype=int)
+            arr[2:12] = nc
```
BSVogler, line 85: This should be `arr[2:2+n_neurons] = nc`

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -69,6 +69,23 @@ def test_list_to_NodeCollection(self):
...
+
+            self.assertEqual(arr[2:12].tolist(), nc.tolist())
```
BSVogler, line 87: Same here, `arr[2:2+n_neurons]`

pynest/nest/tests/test_NodeCollection.py
BSVogler, line 85: This should be `arr[2:2+n_neurons] = nc`

Silmathoron: agreed, done for both

PR #1500, opened 2020-04-05
Correct filename quoting

### LINKED ISSUE #1048: Testsuite error in windows subsystem for linux (ubuntu)
When trying to run the testsuite in WSL, the following error occurs directly in the beginning

```bash
$ make installcheck
[...]
[...]/share/nest/extras/do_tests.sh: 360: export: (x86)/Intel/iCLS: bad variable name
[...]
```
NEST version: 2.16

does anyone have experience with this?

terhorstd:
These changes put correct quotes in many places and allow to install NEST to "some/path with spaces", which broke on different levels before (cmake, sli, bash).

Probably fixes #1048, which was due to wrong split at a space, but haven't tested explicitly on windows.

A design decision was to put single quotes around filenames for calls in shell commands, which inhibits shell variable expansion and treats many special characters more predictable/portable. Any replacements are expected to have been done outside of NEST anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,7 +206,7 @@ file( MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/reports/ )
 if ( HAVE_PYTHON )
   add_custom_target( installcheck
     COMMAND ${CMAKE_COMMAND} -E env
-      NEST_PATH="${CMAKE_INSTALL_FULL_BINDIR}"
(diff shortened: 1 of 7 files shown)


jougs:
Do we need a second review, or do we consider this minimal enough so that mine suffices?

heplesser:
Just signing myself up as reviewer to make sure things work on OSX as well after these changes.

PR #1501, opened 2020-04-06
Move dependencies and fix broken image links

### LINKED ISSUE #1505: Figures are broken on NEST3 transition guide
Firgures not loading on https://nest-simulator.readthedocs.io/en/latest/guides/nest2_to_nest3/nest2_to_nest3_overview.html

rST source code (for example):

  .. image:: ../_static/img/NEST3_23_0.png

Resolves to URL: https://nest-simulator.readthedocs.io/en/latest/guides/nest2_to_nest3/guides/_static/img/NEST3_23_0.png

Should be: https://nest-simulator.readthedocs.io/en/latest/_static/img/NEST3_23_0.png

sarakonradi:
This PR consists of a few minor fixes in the conversion guides. 

More specifically, it: 
- Removes the ``Dependencies`` section from the "NEST 3:0: Detailed transition guide"
- Specifies the ``Python``, ``Matplotlib`` and ``CMake`` versions in the installation instructions instead
- Fixes the broken image links in the "NEST 3.0: What's new?" guide
- Adjusts the two spatial grids to make them equal in size

It fixes #1505.


clinssen:
Closes #1505

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
```diff
@@ -642,8 +642,8 @@ use ``nest.spatial.grid()`` or ``nest.spatial.free``.
                                                          num_dimensions=2))
     nest.PlotLayer(free_nodes);
 
-.. image:: ../_static/img/NEST3_24_0.png
-
+.. image:: ../../_static/img/NEST3_24_0.png
+  :width: 500
```
clinssen, line 646: Should this have a unit, such as px? Why is a width specified for this image but not others?

sarakonradi: > Should this have a unit, such as px?

@clinssen Thanks for spotting this. The image renders well with and without the px unit. Whichever option we vote for, we should ensure all images that specify ``:width:`` consistently use that style. 

> Why is a width specified for this image but not others?

I specified the width of these two grids for now, since one of them was very small and the other one much bigger previously:  
conversion_guide_images_2

I did not touch the other plots, as their original size is suitable. Do you think we should specify width it for all?

clinssen: I think that ideally, we should improve docutils so it auto-detects the width and height of a figure from file, but until that day we will have to specify it manually. Without :width: and :height:, the browser is left guessing about the \ width/height, which is not a good thing.

I'm not however suggesting to make that part of this PR, it would go beyond the scope. Perhaps you have a suitable todo list it can be added to. I would have a slight preference for adding px units, but this is maybe also one for the follow-up.

doc/guides/nest2_to_nest3/nest2_to_nest3_detailed_transition_guide.rst
```diff
@@ -13,11 +13,6 @@ NEST 3.0: Detailed transition guide
    :depth: 2
 
 
-**Dependencies:**
-
- * Python 3.X
- * Matplotlib 3.X.X
```
stinebuu, line 19: Have we written clearly that with NEST 3.0 we will no longer support Python 2? I see that there is a sentence about it in the `nest2_to_nest3_overview` guide, stating `With NEST 3.0, we no longer support Python 2, which reached its end of life on January 1, 2020.`, but I am wondering if we should state this even more clearly? It is a bit hidden under [*Whats changed*](https://nest-simulator.readthedocs.io/en/latest/guides/nest2_to_nest3/nest2_to_nest3_overview.html#what-s-changed) in the guide and users might not spot it. It might be a problem if users are already using NEST and don't look at the install page for dependencies. I agree that it should not be written under *Dependencies* in the transition guide though, but I liked that it was written at the top of the page. I don't know if it fits best under the transition guide or under the overview.

sarakonradi: Hi @stinebuu, that is a very good point. I believe it makes more sense to state this point within a "Note" in the ``NEST 3.0: What's new?`` overview, since this guide has top-level visibility in the navigation. We could also add a bullet point on top of the detailed transition guide, in case a user doesn't see it in the overview guide. How does that sound?

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
sarakonradi: @clinssen You are right about the browser aspect, and we can certainly follow up on the docutils improvement in a separate PR in the future. 

For now, I have added the px units to the ``:width:`` as you suggested.

doc/guides/nest2_to_nest3/nest2_to_nest3_detailed_transition_guide.rst
stinebuu: @sarakonradi This sounds good! I don't know if we need it both places, but lets do it so that we are sure.

sarakonradi:
@stinebuu I added a visible ``Python 3`` note to the top section of the ``nest2_to_nest3_overview_guide.rst``. I also included a brief bullet in the ``nest2_to_nest3_detailed_transition_guide.rst``, just in case.

@clinssen As suggested, I entered the px unit and replied to your comment inline.

Please check!

PR #1502, opened 2020-04-06
Improve BlockVector to fix issues when sorting with recent versions of Boost

### LINKED ISSUE #1239: test_connect_after_simulate.sli fails with assertion 'source_lid < MAX_LID' when compiled with boost=ON in NEST 2.18.0

### LINKED ISSUE #1489: Assertion failed: (source_lid < MAX_LID) for GetConnections in Small Neural networks
**Describe the bug**
Assertion failed: (source_lid < MAX_LID) error querying connectivity with GetConnections for 24 or more connections on Conda installation on Mac os 10.14

**To Reproduce**
To reproduce the behavior, run the following script
import nest
# memory testing
nest.ResetKernel()
n = 5 # number of neurons
epop1 = nest.Create("iaf_psc_alpha",n)
epop2 = nest.Create("iaf_psc_alpha",n)
nest.Connect(epop1,epop2,{'rule': 'all_to_all'})

# we can now query the synapses using GetConnections
conns = nest.GetConnections(epop1,target = epop2)
nest.GetStatus(conns,["target","weight"])

This runs fine for n < 5, but I get the following error with n >=5:

**Expected behavior**
Assertion failed: (source_lid < MAX_LID), function set_source_lid, file /usr/local/miniconda/conda-bld/nest-simulator_1583214474797/work/nestkernel/target_data.h, line 264.
Abort trap: 6

Note that other scripts run fine, including simulations of larger networks as in example script brunel_alpha_numpy.py .

 - OS: Mac OS 10.14
 - Python-Version: Python 3.6
 - NEST-Version: Nest 2.20
 - Installation: Conda

**Additional context**
Reproduced successfully by Hans Ekkehard

hakonsbm:
Until Boost version 1.69.0, `spreadsort` used `std::sort` for small lists (<1000 elements). But with 1.69.0 and onward it uses `pdqsort` from the Boost library instead. Because the `pdqsort` algorithm uses some operators that neither `std::sort` nor `spreadsort` uses, and because of inadequacies in the `BlockVector` implementation, this caused lists to not be sorted.

When sorting with `spreadsort` from the Boost library, iterators for the `BlockVector` to be sorted and the `BlockVector` on which the same operations are performed, are combined. The combination is implemented with the `IteratorPair` iterator, which uses `iterator_facade` from the Boost library. The `iterator_facade` uses a few core functions to infer operators of the iterator. 

The problem arose when `pdqsort` used the operator `operator-=()`. In `IteratorPair`, that operator is inferred from the function
```c++
advance( n ) 
{ 
  sort_iter_ += n; 
  perm_iter_ += n; 
} 
```
Where `sort_iter_` and `perm_iter_` are iterators for the two `BlockVector`s. However, for `operator-=()`, `advance(n)` is called with a negative `n`. The iterator of `BlockVector` didn't take into consideration that its `operator+=(n)` could be called with a negative `n`. This PR improves them to take that into consideration, solving the problem.

In addition, this PR 
- adds additional C++ tests of `BlockVector`
- switches to the single header variant for Boost tests, so that NEST can be installed with Boost without static libraries
- moves to the latest Boost version (1.72.0) on Travis

---

Fixes #1489 
Fixes #1239

### DIFF
--- a/cmake/ProcessOptions.cmake
+++ b/cmake/ProcessOptions.cmake
@@ -462,7 +462,7 @@ function( NEST_PROCESS_WITH_OPENMP )
   if ( NOT TARGET OpenMP::OpenMP_CXX )
     add_library(OpenMP::OpenMP_CXX INTERFACE IMPORTED)
   endif()
- 
+
 endfunction()
 
 function( NEST_PROCESS_WITH_MPI )
@@ -584,8 +584,10 @@ function( NEST_PROCESS_WITH_BOOST )
       set( BOOST_ROOT "${with-boost}" )
     endif ()
 
+    set(Boost_USE_DEBUG_LIBS OFF)  # ignore debug libs
+    set(Boost_USE_RELEASE_LIBS ON) # only find release libs
     # Needs Boost version >=1.58.0 to use Boost sorting
-    find_package( Boost 1.58.0 COMPONENTS unit_test_framework )
+    find_package( Boost 1.58.0 )
     if ( Boost_FOUND )
       # export found variables to parent scope
       set( HAVE_BOOST ON PARENT_SCOPE )
@@ -594,6 +596,7 @@ function( NEST_PROCESS_WITH_BOOST )
       set( BOOST_LIBRARIES "${Boost_LIBRARIES}" PARENT_SCOPE )
       set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIR}" PARENT_SCOPE )
       set( BOOST_VERSION "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" PARENT_SCOPE )
+      include_directories( ${Boost_INCLUDE_DIRS} )
     endif ()
   endif ()
 endfunction()
--- a/extras/travis_build.sh
+++ b/extras/travis_build.sh
@@ -24,7 +24,7 @@
 # It is invoked by the top-level Travis script '.travis.yml'.
 #
 # NOTE: This shell script is tightly coupled to Python script
-#       'extras/parse_travis_log.py'. 
+#       'extras/parse_travis_log.py'.
 #       Any changes to message numbers (MSGBLDnnnn) or the variable name
 #      'file_names' have effects on the build/test-log parsing process.
 
@@ -106,11 +106,14 @@ fi
 if [[ $OSTYPE = darwin* ]]; then
     export CC=$(ls /usr/local/bin/gcc-* | grep '^/usr/local/bin/gcc-\d$')
     export CXX=$(ls /usr/local/bin/g++-* | grep '^/usr/local/bin/g++-\d$')
-    CONFIGURE_BOOST="-Dwith-boost=OFF"
-else
-    CONFIGURE_BOOST="-Dwith-boost=ON"
-fi
- 
+
+# Download Boost
+wget --no-verbose https://dl.bintray.com/boostorg/release/1.72.0/source/boost_1_72_0.tar.gz
+tar -xzf boost_1_72_0.tar.gz
+rm -fr boost_1_72_0.tar.gz
+mv boost_1_72_0 $HOME/.cache
+CONFIGURE_BOOST="-Dwith-boost=$HOME/.cache/boost_1_72_0"
+
 NEST_VPATH=build
 NEST_RESULT=result
 if [ "$(uname -s)" = 'Linux' ]; then
@@ -150,14 +153,14 @@ if [ "$xSTATIC_ANALYSIS" = "1" ]; then
         make PREFIX=$HOME/.cache CFGDIR=$HOME/.cache/cfg HAVE_RULES=yes install
         cd ..
         echo "MSGBLD0040: CPPCHECK installation completed."
-    
+
         echo "MSGBLD0050: Installing CLANG-FORMAT."
         wget --no-verbose http://llvm.org/releases/3.6.2/clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04.tar.xz
         tar xf clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04.tar.xz
         # Copy and not move because '.cache' may aleady contain other subdirectories and files.
         cp -R clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04/* $HOME/.cache
         echo "MSGBLD0060: CLANG-FORMAT installation completed."
-    
+
         # Remove these directories, otherwise the copyright-header check will complain.
         rm -rf ./cppcheck
         rm -rf ./clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04
@@ -167,7 +170,7 @@ if [ "$xSTATIC_ANALYSIS" = "1" ]; then
     export PATH=$HOME/.cache/bin:$PATH
 
     echo "MSGBLD0070: Retrieving changed files."
-      # Note: BUG: Extracting the filenames may not work in all cases. 
+      # Note: BUG: Extracting the filenames may not work in all cases.
       #            The commit range might not properly reflect the history.
       #            see https://github.com/travis-ci/travis-ci/issues/2668
     if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
@@ -193,7 +196,7 @@ if [ "$xSTATIC_ANALYSIS" = "1" ]; then
     # Set the command line arguments for the static code analysis script and execute it.
 
     # The names of the static code analysis tools executables.
-    VERA=vera++                   
+    VERA=vera++
     CPPCHECK=cppcheck
     CLANG_FORMAT=clang-format
     PEP8=pep8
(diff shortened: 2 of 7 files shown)


libnestutil/block_vector.h
```diff
@@ -247,6 +260,9 @@ class BlockVector
    */
   int get_max_block_size() const;
 
+  // TODO: To make BlockVector a complete random access container, it should also implement
+  // max_size(), rbegin(), and rend().
```
heplesser, line 264: This should either be done within this PR if straightforward or turned into a follow-up issue, but not stay as a "TODO" comment. If turned into a follow-up issue, would it be an idea to implement these three methods now to just throw an exception?

libnestutil/block_vector.h
```diff
@@ -606,6 +640,13 @@ operator-( const const_iterator& other ) const
   return ( block_index_ - other.block_index_ ) * max_block_size + ( this_element_index - other_element_index );
 }
 
+template < typename value_type_, typename ref_, typename ptr_ >
+inline typename bv_iterator< value_type_, ref_, ptr_ >::reference bv_iterator< value_type_, ref_, ptr_ >::operator[](
```
heplesser, line 644: I am a bit confused: Stripping the template arguments, this looks to me like `bv_iterator::operator[]`. But do we really want to apply array indexing to the *iterator*? I'd expected it used on the *container*. But I may  be overlooking something here.

testsuite/cpptests/test_block_vector.h
```diff
@@ -32,7 +32,30 @@
...
+  bv_vec_reference_fixture()
+    : N( block_vector.get_max_block_size() + 10 )
+  {
+    for ( int i = 0; i < N; ++i )
+    {
+      block_vector.push_back( i );
+      reference.push_back( i );
+    }
+  }
+  ~bv_vec_reference_fixture()
```
heplesser, line 49: Insert empty line above

testsuite/cpptests/test_block_vector.h
```diff
@@ -260,6 +283,119 @@ BOOST_AUTO_TEST_CASE( test_iterator_compare )
   BOOST_REQUIRE( not( it_b < it_a ) );
```
heplesser, line 283: Would it make sense to add tests for all comparison operators here?

testsuite/cpptests/test_sort.h
```diff
@@ -27,78 +27,151 @@
...
 
-  for ( size_t i = 0; i < N; ++i )
-  {
-    const size_t k = std::rand() % N;
-    bv0[ i ] = k;
-    bv1[ i ] = k;
-  }
+  BOOST_REQUIRE( std::is_sorted( bv_sort.begin(), bv_sort.end() ) );
+  BOOST_REQUIRE( std::is_sorted( bv_perm.begin(), bv_perm.end() ) );
+
```
heplesser, line 121: Could you add a comment explaining why you perform both the `is_sorted` and the `equal` tests?

testsuite/cpptests/test_sort.h
```diff
@@ -27,78 +27,151 @@
 #include <boost/test/unit_test.hpp>
 
 // C++ includes:
+#include <algorithm>
 #include <vector>
 
 // Includes from libnestutil:
 #include "sort.h"
 
 namespace nest
```
heplesser, line 36: Why is everything put in namespace nest?

testsuite/cpptests/test_sort.h
```diff
@@ -27,78 +27,151 @@
...
+  BlockVector< int > bv_perm;
+  std::vector< int > vec_sort;
+};
 
 BOOST_AUTO_TEST_SUITE( test_sort )
 
 /**
  * Tests whether two arrays with randomly generated numbers are sorted
- * correctly by a single call to sort.
+ * correctly when sorting with the built-in quicksort.
```
heplesser, line 113: "built-in" == NEST's own?

testsuite/cpptests/test_sort.h
```diff
@@ -27,78 +27,151 @@
...
+  BOOST_REQUIRE( std::is_sorted( bv_perm.begin(), bv_perm.end() ) );
+
+  BOOST_REQUIRE( std::equal( vec_sort.begin(), vec_sort.end(), bv_sort.begin() ) );
+  BOOST_REQUIRE( std::equal( vec_sort.begin(), vec_sort.end(), bv_perm.begin() ) );
+}
 
-  nest_quicksort( bv0, bv1 );
+/**
+ * Tests whether two arrays with randomly generated numbers are sorted
+ * correctly when sorting with Boost.
```
heplesser, line 143: We know that Boost switches sort algorithms between small and large containers, with a boundary at 1000  elements according to Boost::sort documentation (see `boost::sort::spreadsort::detail::min_sort_size` defined in `boost/sort/spreadsort/detail/constants.hpp`). The tests here should cover both cases.

cmake/ProcessOptions.cmake
```diff
@@ -594,6 +596,7 @@ function( NEST_PROCESS_WITH_BOOST )
       set( BOOST_LIBRARIES "${Boost_LIBRARIES}" PARENT_SCOPE )
       set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIR}" PARENT_SCOPE )
       set( BOOST_VERSION "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}" PARENT_SCOPE )
+      include_directories( ${Boost_INCLUDE_DIRS} )
```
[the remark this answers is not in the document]
hakonsbm: Both `Boost_INCLUDE_DIR` and `Boost_INCLUDE_DIRS` will probably point to the same location. However, based on the examples in [the documentation](https://cmake.org/cmake/help/latest/module/FindBoost.html), and [this answer](https://discourse.cmake.org/t/findboost-difference-between-boost-include-dir-and-boost-include-dirs/855/2) in the CMake forum, `Boost_INCLUDE_DIRS` is the correct variable to use.

libnestutil/block_vector.h
hakonsbm: For iterators, the `operator[]` is the offset dereference operator, which gets the dereferenced element shifted `n` places relative to the iterator position. It is required for random access iterators.

libnestutil/block_vector.h
hakonsbm: I have created function definitions that trow exceptions and removed the "TODO".

hakonsbm:
@heplesser Thanks for your review! I have addressed your comments, please have another look.

extras/travis_build.sh
```diff
@@ -106,11 +106,16 @@ fi
...
- 
+
+# Download Boost
+wget --no-verbose https://dl.bintray.com/boostorg/release/1.72.0/source/boost_1_72_0.tar.gz
+tar -xzf boost_1_72_0.tar.gz
+rm -fr boost_1_72_0.tar.gz
+cp -fr boost_1_72_0 $HOME/.cache
+rm -fr boost_1_72_0
+CONFIGURE_BOOST="-Dwith-boost=$HOME/.cache/boost_1_72_0"
+
```
gtrensch, line 118: The above handles the boost library differently than, for example, music or sionlib. Is there a specific reason for this? If not, wouldn't it be better to implement this the same way? I would suggest to hide the above in a file extras/install_boost.sh, adjust .travis.yaml accordingly and wrap boost with an environment variable, same as it is for all other libraries.

hakonsbm: @gtrensch The idea was to always compile with Boost, like Travis currently does. But I can implement it in a way similar to the other libraries. Should some Travis jobs not compile with Boost?

gtrensch: @hakonsbm The static code check (the first build configuration in the travis build matrix) does not require libboost. I will make a suggestion for a change and create a PR against your branch.

cmake/ProcessOptions.cmake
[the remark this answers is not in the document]
niltonlk: I have just updated PR #1477 where I used,
`include_directories( ${Boost_INCLUDE_DIR} )`

should I change to
`include_directories( ${Boost_INCLUDE_DIRS} )`
and use
`set( BOOST_INCLUDE_DIR "${Boost_INCLUDE_DIRS}" PARENT_SCOPE )`
as commented by @hakonsbm ?

And just to add, BOOST_INCLUDE_DIR is a variable set on the scope above the current (PARENT_SCOPE, [see manual](https://cmake.org/cmake/help/latest/command/set.html)), and the following code would not work as expected, as it will contain the previous value of tha variable.
`include_directories( ${BOOST_INCLUDE_DIR} )`

extras/travis_build.sh
gtrensch: @hakonsbm I have created the PR, please check.

PR #1503, opened 2020-04-06
Remove NEST 3.0 preview warnings

sarakonradi:
Currently, there is a NEST 3.0 preview warning on every page of the documentation: 

> **Warning**
> This is A PREVIEW for NEST 3.0 and NOT an OFFICIAL RELEASE! Some functionality may not be available and information may be incomplete!

This PR removes such warnings from the ``conf.py`` and ``download.rst`` files. 

Ideally, it would be merged not now, but right before the release of NEST 3.0.


heplesser:
A single reviewer (during the release preparation stage) should be sufficient for this one.

sarakonradi:
There is also a "Show version warning" checkbox in the Advanced Settings on Read the Docs which, when disabled, seems to produce the same result in the documentation as this PR. Perhaps @steffengraber could provide more input on whether one or both of these actions (update ``conf.py`` + disable warning) are required.

steffengraber:
@sarakonradi in the Backend this options says: "Show warning banner in non-stable nor latest versions.".  So once master becomes stable version 3, it should no longer be valid.

jougs:
@jessica-mitchell: can you please check if this PR is still doing the right things? That is, could we theoretically just merge it and be done? Thanks!

jessica-mitchell:
@jougs I just took a  look at the output locally and on RtD and it looks good to me!

PR #1521, opened 2020-04-20
Fix default parameters of siegert_neuron

AlexVanMeegen:
It seems the default parameters of the `siegert_neuron` are wrong / inconsistent with the implemented dynamics. Since I believe this is easier to see having the equations at hand, I tried to compile the relevant equations and the corresponding issues in a pdf: [siegert_neuron.pdf](https://github.com/nest/nest-simulator/files/4503304/siegert_neuron.pdf)

Identifying the issue and coming up with the proposed solution as presented in the above pdf is the result of a discussion with @ddahmen and @jasperalbers. All mistakes are my own.


AlexVanMeegen:
PS: In case this is a relevant PR, why does travis fail? Most willing to fix it but I don't even see an error :sweat_smile:

heplesser:
@AlexVanMeegen If you look at the Travis build log (https://travis-ci.org/github/nest/nest-simulator/jobs/677213191), it shows a diff generated by clang-format. So it looks like a formatting problem in `siegert_neuron.cpp`. 

@hakonsbm @jougs @lekshmideepu Any idea why there is no proper report of what went wrong at then end of the log?

AlexVanMeegen:
Thanks @heplesser, indeed `clang-format` removed a few whitespaces. Didn't push yet to preserve the non-informative travis log.

heplesser:
@AlexVanMeegen You can push, Travis keeps the logs of previous runs intact, so with the link above it will remain available for anaylsis.

AlexVanMeegen:
Thanks @heplesser, looks like this was the issue and Travis is happy now.

models/siegert_neuron.h
```diff
@@ -73,21 +75,21 @@ The following parameters can be set in the status dictionary.
...
+ mean  1/s    Additional constant input
 =====  ====== ==============================
 
 The following parameters can be set in the status directory and are
 used in the evaluation of the gain function. Parameters as in
 iaf_psc_exp/delta.
 
-
-=========  ======  =====================================================
+\verbatim embed:rst
```
clinssen, line 85: @sarakonradi: is this statement necessary after #1530?

models/siegert_neuron.h
```diff
@@ -58,8 +58,10 @@ Description
 siegert_neuron is an implementation of a rate model with the
 non-linearity given by the gain function of the
 leaky-integrate-and-fire neuron with delta or exponentially decaying
-synapses [2]_ and [3, their eq. 25]. The model can be used for a
-mean-field analysis of spiking networks.
+synapses [2] and [3, their eq. 25]. The model can be used for a
```
clinssen, line 61:
```suggestion
synapses [2]_ and [3]_ (their eq. 25). The model can be used for a
```

models/siegert_neuron.h
sarakonradi: @clinssen It is not. In fact, this needs to be removed. Otherwise the table won't render properly!

clinssen:
@AlexVanMeegen: could you address the small documentation-related comments of me and @sarakonradi? Then we will merge!

AlexVanMeegen:
@clinssen done, thanks for reminding! @sarakonradi could you have another look? Thanks :)

AlexVanMeegen:
@clinssen @sarakonradi @jstapmanns @ddahmen thanks for the reviews!

PR #1526, opened 2020-04-21
Copy the current Microcuircit documentation for use with Readthedocs 

### LINKED ISSUE #1511: Link .py files in PyNEST microcircuit
This issue follows up on #1444. 

If we publish the [README](https://github.com/nest/nest-simulator/blob/master/pynest/examples/Potjans_2014/README.rst) as it currently is on Read the Docs, the user would see the list of ``.py`` files and not know where these are located:

microcircuit_file_structure

One solution would be to find a way to automatically add the ``:download:`` role to the ``.py`` files in the README when generating documentation with Sphinx. This would allow the user to download the files on Read the Docs. It would not affect the README itself in ``pynest/examples/Potjans_2014``. 

A simpler option would be to re-edit the README, turning the files listed under ``File structure`` into external links that direct the user to the file links on GitHub. The benefit of this option is that the hyperlinks would be rendered on Read the Docs as well as on GitHub: https://github.com/sarakonradi/nest-simulator/blob/cb8b4325c0045675063121495d1e604bdd01a879/pynest/examples/Potjans_2014/README.rst

Input from @steffengraber and @jhnnsnk is very welcome.

### LINKED ISSUE #1524: Publish updated microcircuit documentation on Read the Docs
As discussed, the updated microcircuit documentation found in ``pynest/examples/Potjans_2014`` needs to be published on Read the Docs prior to the NEST 3.0 release. The outdated files in ``doc/microcircuit`` should be removed.

steffengraber:
This PR fix issue #1524 and publish the updated microcircuit documentation on Read the Docs and for local use with sphinx.

fixes #1524 
fixes #1511


steffengraber:
@sarakonradi Thanks for your help!

sarakonradi:
@jhnnsnk Thanks for your detailed feedback. I updated the references and changed the link title, as you suggested in 2. and 3. Please check!

> Currently the panel to the left provides links to the sections of the README which is unnecessary. Here we could consider to get links to the individual .py files of the model, or alternatively, just one link to a landing page of the model (for the latter see #1511) which then contains the links to the files. The files are already converted to .html (they appear in auto_examples/Potjans2014 when building the documentation), but they are not linked, yet, as far as I see

We haven't figured out a proper solution for that yet (see my latest comment in #1511). How I understood it in the VC earlier this week is that #1511 is not part of the NEST 3.0 milestone and that such link enhancements will be examined after the release. 

So our main priority for this PR, I believe, should be to publish the new README on Read the Docs. We can improve the links at a later point in time. What do you think?

sarakonradi:
@jhnnsnk @steffengraber I created a new landing page and added the absolute link for ``reference_data``, as discussed offline. The file links in the landing page still need to be updated and the README image fixed. @steffengraber Could you please have a look?

steffengraber:
@sarakonradi I fixed the links.

sarakonradi:
@steffengraber Thanks! The links work now. 

Could you please check the images too? This is how they render at the moment:

Screenshot from 2020-04-23 12-48-00 (1)

steffengraber:
@sarakonradi @jhnnsnk The problem with the 3 pictures is solved.

sarakonradi:
@steffengraber Thanks!

sarakonradi:
@jhnnsnk I just updated the landing page. Please check!

jhnnsnk:
@sarakonradi Looks great to me now! No further comments from my side (apart from this blank line Travis complains about).

steffengraber:
@jhnnsnk Travis is happy now.

@terhorstd Please review.

doc/examples/pynest_microcircuit_link.rst
```diff
@@ -0,0 +1,2 @@
+.. include:: ../../pynest/examples/Potjans_2014/README.rst
```
terhorstd, line 1: I think this should be copied too, so that we can stay in the doc tree with this link and just point to `Potjans_2014/README.rst`. If the `pynest/examples` is moved, we only want to fix the one source path in `conf.py`

doc/conf.py
```diff
@@ -30,13 +30,11 @@
...
 import sys
 import os
-
 import re
-
 import pip
-
 import subprocess
 
+from shutil import copytree, ignore_patterns, rmtree
```
terhorstd, line 37: rmtree seems not to be used
```suggestion
from shutil import copytree, ignore_patterns
```

doc/examples/pynest_microcircuit_link.rst
steffengraber: If this is also copied, a documentation page is also built from it. Then this page exists twice. With each building process it will be copied and converted by sphinx_gallery. This will be sorted in the documentation.

terhorstd: If the duplication of the readme leads to the page existing twice, it is ok to use the original version.
Thanks for the background!
:+1:

PR #1455, opened 2020-03-06
Backport of #1442 to 2.20.1

jasperalbers:
cherry-picked commits from pull request #1442, resolved a minor merge conflict

PR #1532, opened 2020-04-22
Backport of #1481 to 2.20.1

jasperalbers:
This PR is a part of #1455 and includes all commits from #1481, except the following merges:

e49ac09, Merge pull request #7 from heplesser/stine_1272
03156f9, Merge remote-tracking branch 'origin/issue1272_infinite_loop' into is…
5330a78, Merge remote-tracking branch 'origin/issue1272_infinite_loop' into is...

I don't quite understand these commits, and whether they are supposed to be included here or not as they originate in very old and already deleted branches.

Furthermore, there were a few merge conflicts involved because of the differences between NEST 3 and 2, which I solved to the best of my understanding, but which probably should be looked over (in particular concerning the files`testsuite/unittests/test_iaf_ps_dc_accuracy.sli` and `testsuite/unittests/test_iaf_ps_dc_t_accuracy.sli`).


precise/iaf_psc_exp_ps_lossless.cpp
```diff
@@ -29,6 +29,11 @@
 #include "exceptions.h"
 #include "universal_data_logger_impl.h"
 
+// Includes from libnestutil:
+#include "dict_util.h"
```
stinebuu, line 33: dict_util is part of NEST 3.0 and not part of NEST 2, so this should be removed.

nestkernel/node.h
```diff
@@ -860,7 +860,7 @@ class Node
     assert( false );
     return 0;
   }
-
+
```
stinebuu, line 863: Please remove this whitespace!

testsuite/unittests/test_iaf_ps_dc_accuracy.sli
```diff
@@ -83,19 +84,26 @@ M_ERROR setverbosity
...
- [  500.0           1e-9  ]  % error larger because of accumulation
+  [    5.0           1e-13 ]       
+  [  500.0           1e-9  ]  % error larger because of accumulation
 ] /Tlist Set                 % at very small computation step sizes
 
+% Models to be tested by this test
+[
+  /iaf_psc_alpha_ps 
+  /iaf_psc_delta_ps
+  % other precise models should be added to this list
```
stinebuu, line 95: I think you could add `/iaf_psc_alpha_presc` to the list as well, the model is still part of NEST 2. (just not part of NEST 3.0).

jasperalbers: I agree, that model was there before. I also added `/iaf_psc_alpha_presc` to test_iaf_ps_dc_t_accuracy.sli.

stinebuu:
@jasperalbers Travis is failing because we changed `test_iaf_ps_dc_t_accuracy.sli` and now `/iaf_psc_alpha_presc` make the test fail because the accuracy is too low. I don't remember exactly what was changed in the test, but I don't think it is that important. The underlying functionality is not changed in the model, and I think it would be okay to remove `/iaf_psc_alpha_presc` from the test.

jasperalbers:
> @jasperalbers Travis is failing because we changed `test_iaf_ps_dc_t_accuracy.sli` and now `/iaf_psc_alpha_presc` make the test fail because the accuracy is too low. I don't remember exactly what was changed in the test, but I don't think it is that important. The underlying functionality is not changed in the model, and I think it would be okay to remove `/iaf_psc_alpha_presc` from the test for NEST 2.20.1.

@stinebuu Just to clarify, do you suggest removing `/iaf_psc_alpha_presc` only from test_iaf_ps_dc_t_accuracy.sli or also from test_iaf_ps_dc_accuracy.sli?

stinebuu:
@jasperalbers Lets remove it from both for consistency. Sorry for the back and forth!

PR #1536, opened 2020-04-22
Fix NodeCollection creation to correctly raise error when using non-unique or unsorted node IDs

### LINKED ISSUE #1563: Implicit reordering with array slicing of NodeCollection
**Describe the bug**
In [current master](https://github.com/nest/nest-simulator/commit/10ad3af021fd2ef6d999087bc3f56db6791c5db9), since the merge of #1485, ``NodeCollection`` accept array slicing but does not check for index order.
This means that it automatically reorders the result (because a ``NodeCollection`` is ordered).

**To Reproduce**
```python
import nest
nn  = nest.Create("iaf_psc_alpha", 10)
res = nn[[3, 2, 1]]
```

**Expected behavior**
Either unordered ``NodeCollection`` or error.

**EDIT:** as decided in #1570, ``NodeCollection`` should remain ordered and therefore an error should be raised.

hakonsbm:
All node IDs in a NodeCollections are supposed to be unique. But it is possible to create one where the lowest node ID is non-unique.
```python
>>> n = nest.NodeCollection([1, 1, 2])
>>> print(n)
NodeCollection(metadata=None,
               model=iaf_psc_alpha, size=1, first=1;
               model=iaf_psc_alpha, size=2, first=1, last=2)
```

This PR makes NodeCollections properly check uniqueness, and adds a test for it.

These changes are moved from #1485.

Fixes #1563


nestkernel/node_collection.cpp
```diff
@@ -165,7 +165,7 @@ NodeCollection::create_( const std::vector< index >& node_ids )
 
   std::vector< NodeCollectionPrimitive > parts;
 
-  index old_node_id = 0;
+  index old_node_id = current_first;
   for ( auto node_id = ++( node_ids.begin() ); node_id != node_ids.end(); ++node_id )
   {
     if ( *node_id == old_node_id )
```
heplesser, line 171: This test compares only neighboring entries in `node_ids` and does not enforce that they are strictly increasing. Therefore, ill-formed inputs can lead to unexpected results, e.g.
```Python
In [6]: nest.NodeCollection([1, 2, 3, 1, 5])                                                                                                                                              
Out[6]: 
NodeCollection(metadata=None,
               model=iaf_psc_alpha, size=1, first=1;
               model=iaf_psc_alpha, size=3, first=1, last=3;
               model=iaf_psc_alpha, size=1, first=5)
```
To catch both problems, I'd suggest `*node_id <= old_node_id`.
Btw, I am wondering if `auto& node_id` would save you dereferencing `*`s inside the loop.

hakonsbm: @heplesser I can't seem to reproduce your result, for me it throws an error. The `node_ids` argument in `create_` is expected to be sorted, so it shouldn't be a problem. However the sortedness is never checked, but I can add an `assert` for that.

And `node_id` is an iterator which needs to be dereferenced. But I can rename it to for example `node_id_it` to make it more clear.

heplesser: @hakonsbm My mistake! My example above was based on master, not your branch. Also ignore my comment on the iterator!

But I have another issue: 
- If `create_()` expects its input to be sorted, this should be clearly documented somewhere.
- The other questions is if the input that `create_()` receives should at all be sorted on its way from the Python level down to `create_()`, see also the discussions in #1563 and #1570. I would tend to leave the sorting to the user; that saves the sorting in the NEST code, and we would here just have to have a little and very lightweight check for strict monotony.

hakonsbm: @heplesser I can add documentation on the `create_()` function, and I agree that we can leave the sorting to the user. But is that within the scope of this PR, or would it be better to create a separate one?

heplesser: @hakonsbm Let us set this one on hold until we have come to a conclusion in #1570.

heplesser:
@hakonsbm @Silmathoron Is this PR still relevant now that we have resolved #1570 and in a view to #1539? How do these two PRs relate?

I will turn this PR into a draft until we have clarified its status.

Silmathoron:
I think yes: we decided that ``NodeCollection`` will remain a set and unordered so anything non-unique or unordered should raise an error.
Did I miss something?

stinebuu:
@hakonsbm could you add the `assert` for non-ordered nodes (either in this PR or another) and the documentation about non-unique and sorted nodes? So that we address everything on #1570.

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -842,7 +842,7 @@ def SelectNodesByMask(layer, anchor, mask_obj):
     node_id_list = sli_func('SelectNodesByMask',
                             layer, anchor, mask_datum)
 
-    return NodeCollection(node_id_list)
+    return NodeCollection(sorted(node_id_list))
```
stinebuu, line 845: It might be an idea to add a comment here as well, about NodeCollections needing to be sorted.

hakonsbm: Done!

hakonsbm:
This PR now fixes #1563 as well.

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -372,6 +374,37 @@ def test_correct_len_on_NodeCollection(self):
...
+    def test_from_list_unsorted_raises(self):
+        """Creating NodeCollection from unsorted list raises error"""
+        nest.Create('iaf_psc_alpha', 10)
+
+        with self.assertRaises(nest.kernel.NESTError):
+            nest.NodeCollection([5, 4, 6])
+        with self.assertRaises(nest.kernel.NESTError):
+            nest.NodeCollection([5, 6, 4])
+
+    def test_slice_with_unsorted_raises_raises(self):
```
stinebuu, line 397:
```suggestion
    def test_slice_with_unsorted_raises(self):
```

hakonsbm:
@heplesser Can you take another look at this?

jougs:
@Silmathoron: would you be able to step in as reviewer on this one? @heplesser is currently busy teaching.

Silmathoron:
Yes, I'll review it this evening or tomorrow

Silmathoron (approved):
Looks good to me! :+1:

nestkernel/node_collection.cpp
stinebuu: In #1570 we decided that the nodes should be ordered, documentation has been added, so I am resolving this conversation.

PR #1538, opened 2020-04-22
Handling of out-of-range slicing and indexes for NodeCollections

alberto-antonietti:
When working with lists or arrays, if an index outside of the NodeCollection is used (as implemented in #1485, then the right error is thrown:

```python
n = nest.Create('iaf_psc_alpha', 10) 
n[[1, 5, 10]]
```

> IndexError: pos points outside of the NodeCollection

This does not happen when using integers or slices as indexes:

```python
n = nest.Create('iaf_psc_alpha', 10) 
n[10]
# OR
n[1:12]
```

> ~/workspace/nest-simulator/b/lib/python3.6/site-packages/nest/lib/hl_api_types.py in getitem(self, key)
> 216 step = 1 if key.step is None else key.step
> 217
> --> 218 return sli_func('Take', self._datum, [start, stop, step])
> 219 elif isinstance(key, (int, numpy.integer)):
> 220 return sli_func('Take', self._datum, [key + (key >= 0)])
> 
> ~/workspace/nest-simulator/b/lib/python3.6/site-packages/nest/ll_api.py in sli_func(s, *args, **kwargs)
> 181 sli_push(args) # push array of arguments on SLI stack
> 182 sli_push(s) # push command string
> --> 183 sli_run(slifun) # SLI support code to execute s on args
> 184 r = sli_pop() # return value is an array
> 185
> 
> ~/workspace/nest-simulator/b/lib/python3.6/site-packages/nest/ll_api.py in catching_sli_run(cmd)
> 131
> 132 exceptionCls = getattr(kernel.NESTErrors, errorname)
> --> 133 raise exceptionCls(commandname, message)
> 134
> 135
> 
> NESTErrors.BadParameter: ('BadParameter in Take_g_a: stop <= size() required.', 'BadParameter', 'Take_g_a', ': stop <= size() required.')

in this PR, I have added a reporting of more consistent errors (`IndexError`) when setting `index`, `start`, `stop`, or `step `values that are not allowed.

Plus, I should have fixed a small bug in handling negative stop values.

I added a few more tests that covers negative values of indexing.

@hakonsbm has already volounteered to review this PR 😄


Silmathoron:
Hey there, thanks for the PR!
I have a small comment: currently doing ``n[:20]`` raises ``IndexError`` for a ``NodeCollection`` of size 10, however, in python (and numpy), stop outside range defaults to "end-of-container" value and returns the whole container; isn't this some kind of standard we would like to abide by?

alberto-antonietti:
> Hey there, thanks for the PR!
> I have a small comment: currently doing `n[:20]` raises `IndexError` for a `NodeCollection` of size 10, however, in python (and numpy), stop outside range defaults to "end-of-container" value and returns the whole container; isn't this some kind of standard we would like to abide by?
> 
> To clarify: `n[20]` should indeed spit and `IndexError`, but `n[:20]` or `n[-20:]` should return the whole collection to be consistent with python indexing standard.

I see your point. We need to think about how similar NodeCollection slicing should be with respect to e.g. list slicing. Some differences that are currently present:
- list slicing with bad start/stop/step indexes generates empty lists instead of errors. It generates a `ValueError` only if you try to set a `step=0`.

```python
a = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
In [2]: a[20:]                                                                                                                                                                         
Out[2]: []
```

- list slicing with strictly negative step values is allowed, while NodeCollection slicing can be done only with strictly positive steps.
```python
In [6]: a[10:0:-1]                                                                                                                                                                     
Out[6]: [10, 9, 8, 7, 6, 5, 4, 3, 2]
```

Personally, I would prefer to adopt a more conservative approach in NodeCollection slicing, where errors are thrown when trying to access out-of-range values.
If I try a n[:20] slicing, I would do that because I am expecting to have 20 values, if I get 10 without an error this could pass unnoticed and the problem can be more difficult to be found.

Silmathoron:
I have no strong opinion on both out-of-range start/stop configurations, I just think it should be discussed explicitly...

Not slicing with negative steps seems logical because of the ordered structure of ``NodeCollection``, though, so I see no issue on that side.

heplesser:
@alberto-antonietti We discussed this afternoon and concluded that we should go for consistency with Python/NumPy, i.e., the "to end-of-container" logic. Negative steps will **not** be supported.

alberto-antonietti:
> @alberto-antonietti We discussed this afternoon and concluded that we should go for consistency with Python/NumPy, i.e., the "to end-of-container" logic. Negative steps will **not** be supported.

Fair enough, thanks @Silmathoron for pointing that out.

Now start and stop values out-of-range are capped at the list/NumPy array limits. 
I have added the related tests.

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -156,6 +156,18 @@ def test_slicing(self):
...
+        n_slice_stop_outside = n[:15]
+        n_list_stop_outside = n_slice_stop_outside.tolist()
+        self.assertEqual(n_list_stop_outside, [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+
+        n_slice_start_stop_outside = n[-13:17]
+        n_list_start_stop_outside = n_slice_start_stop_outside.tolist()
+        self.assertEqual(n_list_start_stop_outside, [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+
         with self.assertRaises(IndexError):
             n[::-3]
```
Silmathoron, line 173:
```suggestion
        with self.assertRaises(IndexError):
            n[11]

```

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -103,8 +103,11 @@ def test_indexing(self):
         self.assertEqual(n[4], nc_4)
         self.assertEqual(n[-1], nc_4)
         self.assertEqual(n[-3], nc_2)
-        with self.assertRaises(nest.kernel.NESTError):
-            n[7]
+        self.assertEqual(n[-5], nc_0)
```
alberto-antonietti, line 106: @Silmathoron checking of out-of-range single index is done here

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -103,8 +103,11 @@ def test_indexing(self):
...
+        with self.assertRaises(IndexError):
+            n[5]
+        with self.assertRaises(IndexError):
+            n[-6]
```
alberto-antonietti, line 110: @Silmathoron checking of out-of-range single index is done here

Silmathoron: ah, my bad, I was looking only at the new diff!

Issue #1375, opened 2019-12-31
Implement a registry for populations


jougs:
In order to ease introspection of the network by external tools (think analysis, visualization, NEST Desktop, NRP, TVB, Arbor, ...) it would be very helpful if the NEST kernel provided a registry (or simply a map), where `NodeCollection`s representing population could be enrolled under a textual name. Before rushing into an implementation, I first wanted to discuss this idea.

@stinebuu, @hakonsbm, @heplesser, @abigailm.

stinebuu:
@jougs Just so I understand, it would be a map that would map a name to a `NodeCollection`, and by providing the name you can easily look up the NodeCollection? Would we create the name, or would it be given by the user?

jougs:
Yes, and the name would be given by the user (possibly already in the call to `Create()`).

heplesser:
@jougs This is a very interesting idea. It raises the general question of who should be responsible for maintaining an overview over network structure. Projections between populations are as important as the populations themselves.

For the implementation of the Hill-Tononi network, my approach was to put all information about populations and projections into dictionaries. These could then be passed to the NEST kernel to construct the network, ConnPlotter to visualise it, etc. The dictionaries could also be mapped to other simulators, etc. This placed all structural information at the Python level, while the NEST kernel essentially had "dumb" representation with only the information required to perform simulations stored in the kernel (ignoring the subnets that disappear in NEST3).

Let's discuss this thoroughly, also with the people working on multisimulator interaction and visualisation!

Silmathoron:
During the NEST Community Conference a few years ago, there were a couple of people who wanted to discuss the possibility of accessing the in/out-degrees in addition to the nodes.
Use cases were noise generation (spontaneous release at synapses) and some visualizations with external tools where the network cannot be accessed directly (I don't remember precisely).
This might also be a possible addition to the discussion.

For me, in the context of NNGT, I would need to be able to access at least the nodes, but any additional information would be useful. Information at the Python level only is fine for me, but I think that, since other libraries may want to interact with NEST at C++ level, storing the information there would make more sense.

jougs:
@heplesser: the idea was actually sparked by discussions with people from the [HBP NRP](https://www.neurorobotics.net/), people developing [Insite](https://devhub.vr.rwth-aachen.de/VR-Group/in-situ-pipeline), and the multi-/co-simulation enthusiasts in Jülich.

[...]

PR #1539, opened 2020-04-23
Enhance Connect to work seamlessly with arrays and lists

### LINKED ISSUE #1523: Full support for numpy arrays and lists in Connect
Following [a discussion](https://github.com/nest/nest-simulator/pull/1429#issuecomment-610761021) in #1429, I am adding an issue here regarding the support of lists and arrays in ``Connect``.

Currently, many forms of ``Connect`` are not supported, or require superfluous arguments:
* ``list`` objects are not supported
* using numpy arrays require both inputs to be numpy arrays and the ``syn_spec`` argument

```python
import nest
import numpy as np

# create neurons and devices
neuron = nest.Create("iaf_psc_alpha", 10)

vm = nest.Create("voltmeter", 10)

# set weights
weights = np.ones(len(neuron))
delays = np.ones(len(neuron))

# numpy version
arr_vm  = np.array(vm.tolist(), dtype=int)
arr_nrn = np.array(neuron.tolist(), dtype=int)

# works
nest.Connect(arr_vm, arr_nrn, syn_spec={'weight': weights, 'delay': delays, 'synapse_model': 'static_synapse'})

# does not work (but should)
nest.Connect(arr_vm, neuron.tolist(), syn_spec={'weight': weights, 'delay': delays, 'synapse_model': 'static_synapse'})
nest.Connect(vm.tolist(), neuron.tolist(), syn_spec={'weight': weights, 'delay': delays, 'synapse_model': 'static_synapse'})

nest.Connect(arr_vm, arr_nrn)
nest.Connect(arr_vm, neuron, syn_spec={'weight': weights, 'delay': delays, 'synapse_model': 'static_synapse'})
nest.Connect(arr_vm, neuron)
nest.Connect(vm.tolist(), neuron)
```

All examples shown above should be accepted.

We are currently already checking the type of the input in ``Connect``, thus, the solution to automatically convert to ``NodeCollection`` if the ids are unique would only require a very minor overhead and no additional code complexity (just a ``check_input_nodes`` function that would replace the current test).
I am not sure why ``syn_spec`` is currently required, but, to me, when input nodes are not unique, we should be able to switch to the "data connect" without any additional argument...

Silmathoron:
This PR is a follow up from #1429 and fixes #1523.
It requires either the small changes in #1497 to be merged first or to be incorporated from here.

**Behavior from the user standpoint**
There are technically two cases, but the difference should not even occur to users unless they try to use ``conn_spec != "one_to_one"`` with non-unique elements in their containers:
1. if they provide a ``NodeCollection`` object or a container that is equivalent to it (contains only unique elements) then everything works as it used to in NEST 2, all operations are possible,
2. if they provide something that contains non-unique elements, then they can only do one-to-one connections, anything else raises an error.

**Behavior from the developer standpoint**
as of #1429, there is a ``connect_arrays`` function for "data" connect, so we test input types for the different cases:
1. both inputs to ``Connect`` are ``NodeCollections``, all happens as before, no additional checks performed,
2. both objects are numpy arrays and ``conn_spec`` is "one_to_one", then no checks are performed and ``connect_arrays`` is used
3. otherwise:
   * if ``conn_spec`` is not "one_to_one", then non-``NodeCollection`` objects are checked for uniqueness; if unique, they are converted to ``NodeCollection`` and we move to 1, otherwise an error is raised
   * if ``conn_spec`` is "one_to_one", both objects are converted to numpy arrays and we move to 2.

**Summary of changes**
Compared to the #1429, this modification removes unnecessary ``syn_spec`` parameter from the ``Connect`` calls but requires ``conn_spec="one_to_one"`` to lift the ambiguity between "default" ``Connect``, which uses ``"all_to_all`` and the "data" ``Connect``, which requires ``"one_to_one"``.
(currently, "data connect" requires ``conn_spec`` to be ``None``, then sets it to "one_to_one" internally, which does not seem intuitive).

Through these changes, all behaviors are now consistent with normal ``Connect`` calls using ``NodeCollection`` objects and the uniqueness check is limited to the single case of users performing "non-data" connects without using ``NodeCollection`` objects.

Suggested reviewers: @alberto-antonietti @hakonsbm @stinebuu


Silmathoron:
Examples:

```python
neurons = nest.Create("iaf_psc_alpha", 10)

lst  = [i for i in range(1, 11)]
arr = np.array(lst)
```
contains the gids in 3 different formats.

These 4 calls are all equivalent and will connect neurons all-to-all.
```python
nest.Connect(neurons, neurons)
nest.Connect(lst, neurons)
nest.Connect(arr, arr)
nest.Connect(lst, arr)
```
note that ``lst`` and ``arr`` will be converted to ``NodeCollection`` and the wrong ``dtype`` of ``arr`` is automatically corrected to integers; before conversion, they have been checked to see whether they contain only unique elements.

The new "data connect" is done as follow:
```python
sources = [1, 1, 6, 8, 4, 3, 6, 5, 4, 9]
targets   = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
nest.Connect(sources, targets, conn_spec="one_to_one")
```
sources and targets are converted to numpy arrays but uniqueness is not checked (since it is not required)

```python
nest.Connect(sources, targets)
```
now raises an error since the default ``conn_spec`` is "all-to-all" and is not compatible with non-unique indices in the input.

Silmathoron:
ToDo:

- check that this does not induce too much overhead for "normal" cases (using ``NodeCollection`` and data connect with numpy arrays -- it should not)
- check whether it would be better to use #1536 rather than the ``set`` to check for uniqueness

Silmathoron:
Following the discussion yesterday, I updated the description to explain the behavior from user and developer standpoints, I hope it makes things clear.
From my point of view, this behavior should be intuitive from the user standpoint (they should see no difference from how they used NEST in version 2), but maybe I missed something.

Silmathoron:
Build has errored on Travis but I believe this is a random bug, could someone with maintainer rights restart clang and OSX tests?

jougs:
I agree with @Silmathoron that this should not be a surprising behavior to the users. I thus approve from a purely conceptual stand-point and leave the technical review to @heplesser and @hakonsbm.

heplesser:
@Silmathoron @jougs My position on this PR is a different one. I believe that we should keep semantics plain from a user perspective, even if that in special use cases may require an additional line of code.

### Unambiguous semantics
As @Silmathoron pointed out, current `Connect()` semantics differ depending on whether NodeCollections or arrays are passed. We should resolve this by an explicit connection rule for data-driven connections:
```python
nest.Connect(source_array, target_array, conn_spec='data')
```
Source and target are either NumPy arrays or iterables that will be converted to arrays internally.
If `conn_spec` is `'data'`, only arrays/iterables are permitted and connections are performed one-to-one. The new spec `'data'` is introduced to make explicit that connections here are created based on explicit source and target lists.

### No syntactic sugar
Unless `conn_spec` is `'data'`, both source and target must be given as `NodeCollection`s. Any `conn_spec` except `'data'` may be used in this case, with `'all_to_all'` as default. Since `NodeCollection`s support complex slicing and combination operations, they should be acceptable for almost all use cases—users should use explicit node ids as little as possible. If you want to manually select a few neurons for connection to, e.g., a voltmeter, best practice is
```python
nrns = nest.Create('iaf_psc_exp', 100)
vm = nest.Create('voltmeter')
nest.Connect(vm, nrns[[4, 10, 27, 81]]) 
```
This code is robust under exchange of the `Create()` calls, while
```python
nest.Connect(vm, [4, 10, 27, 81]) 
```
would not be. We shall not lead our users into temptation ;).

Any list of unique node ids can always be converted into a `NodeCollection`, if absolutely necessary, adding at most one line of code. 

The only remaining case then would be that a user would want to connect from a group of sources with duplicates to a group of unique targets (or vice versa). I struggle to come up with a real-life situation where this occurs. And in this case on can generate an array holding the target data.

Therefore, I would suggest to keep it simple and allow only NC+NC or only Array+Array.

Silmathoron:
@jougs thanks for the answer

@heplesser I'm having trouble understanding part of your proposal: I get the part about the ids being different depending on the order of the ``Create`` calls (though if users use explicit numbers, I do not see why they would have trouble getting them right, or why they would not slice an array of gids), however:

* though I called it "data connect", the rule underlying connection with duplicated nodes *is* ``"one_to_one"``, introducing a new ``"data"`` conn_spec for that seems therefore rather weird to me..
* I disagree with your view on ``NodeCollection`` slicing being sufficient as it does not support array slicing with duplicate entries; for many application associated with graph theory, this occurs very often and makes it impossible to work with ``NodeCollection`` objects, making the "syntactic sugar" quite nice

I would like to add that, since we already have to check the type of the input, this syntactic sugar has no additional cost for normal calls and makes life easier for the others.
The decision of converting rather than making a specific case therefore seems more consistent to me as it provides the same logic for all cases, which is also the logic existing users are used to.
Furthermore (though this is not really a strong argument) maintaining this logic would actually break less code since it is compatible with previous NEST 2 code.

For a real life example, all I do on NEST with networks is performed using an array of gids because I need to slice it regularly, often with non-unique slices.

heplesser:
@Silmathoron Thank you for your detailed explanation. I have a few more questions and comments.

1. I understand your need to work with arrays containing duplicates when working with connectivity information from graph theory tools. But then I would expect that you would have both sources and targets as arrays. Where does the need to *mix* arrays and node collections arise?
1. My main concern is with your case 3, i.e., "mixed" input and a connection specification that is not `one_to_one`.  I see two issues here:
    1. Node collections are not only unique, they are also ordered in increasing order of node id. Should conversion raise an error if the array-input is not ordered, or do we accept that nodes will be ordered implicitly upon conversion? See also [my comment on #1536](https://github.com/nest/nest-simulator/pull/1536/files#r414114383) (@hakonsbm).
    1. One could imagine that a user creates the array input in a random fashion, which may result in arrays with duplicates in some realisations and without in others. Thus, the same `Connect()` call would raise errors in some cases and not in others. But maybe I am overly critical here: if a user creates random input, she needs to ensure that the input fulfills requirements in any case.
1. My suggestion to introduce the `data` connection specifier, which does essentially the same as `one_to_one`, is motivated by a belief that this might make scripts more readable by being more explicit: If I use any connection specification except `data`, this indicates that we ask the NEST kernel to generate *rule-based connectivity*, while if we use `data`, we clearly indicate that we tell the NEST kernel explicitly which individual connections to create. Now, both `one_to_one` and `all_to_all`, while abstract rules, are also explicit because they have no degrees of freedom. But the philosophy behind is still different.

Silmathoron:
@heplesser 

1. mixed arrays/NodeCollection is for recordings, but true, this only comes from the fact that I am using the same array as for slicing out of convenience: I could also store the NodeCollection somewhere and use the NodeCollection for that, I'm just not doing it because I do not want to "carry" a duplicate container around
2. i. Fair point, I forgot about that fact... let me think about it...
   ii. I do not think this is an issue.
3. I see your point.

I will think about it and come back to you. I still think the seamless behavior is overall a good thing, but the ordering issue is indeed rather problematic since raising an issue for unordered arrays would make the whole process less intuitive and not raising an issue would hide an implicit sort from the user, which is bad...

@jougs would your previous answer still hold despite this problem?

Silmathoron:
After some (brief) thinking, I realized that, actually, the only case where ordering is an issue is for "one_to_one", isn't it? I think all other rules do not care about the order. We could therefore check for ordering in this case and raise an error.
This is actually speaking in favor of your ``"data"`` conn_spec, since in this case the order should not be checked.

So I would consider the following possibilities:
1. my seamless proposal, with an additional check for order if "one_to_one" is used with something other than NodeCollection (the test is almost identical) and the addition of "data" conn_spec instead of "one_to_one".
2. only using NodeCollections for all calls except for "data", which would accept lists/arrays.

I still think the first proposal is better since it makes things easy without any additional complexity from the user standpoint and (you'll judge from the code) very little complexity or need for maintenance from the developer standpoint.

jougs:
@Silmathoron, @heplesser: many thanks for discussing this in such depth. I am afraid I did initially not fully understand all implications and seem to not be completely up-to-date with respect to the internal details of the current NodeCollection implementation.

@Silmathoron: Your latest proposal indeed sounds like a nice solution to me, but I'm leaving a final decision to @heplesser.

I have one more remark:

> 1. mixed arrays/NodeCollection is for recordings, but true, this only comes from the fact that I am using the same array as for slicing out of convenience: I could also store the NodeCollection somewhere and use the NodeCollection for that, I'm just not doing it because I do not want to "carry" a duplicate container around

I don't see a problem with this, as NodeCollections should usually be pretty lightweight. And maybe an implementation of #1375 can help to make things easier for you.

heplesser:
This discussion is very closely tied to #1570, and I suggest we continue the discussion there. Once we have a decision on #1570, we can continue here. I therefore change this to "Blocked" and "Draft".

Silmathoron:
@heplesser I checked the PR and I think it is fine as is (as long as #1497 is merged first).

There are two little issues remaining, though:
* currently there are two "one_to_one" connect (the "old one" with NodeCollections and the "data one" with numpy arrays
* when using numpy arrays, inputs with an incorrect ``dtype`` are automatically converted to int; since we also accept lists, I don't see any straightforward way of checking type without O(N) operations which seems a bit of a shame...

I don't think these are major problems but I may have overlooked something...

Silmathoron:
@heplesser @hakonsbm I think this PR is now ready for review :)

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -137,9 +144,14 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, connect_np_arr
                             "only be used in conjunction with rules "
                             "'all_to_all', 'fixed_indegree' or "
                             "'fixed_outdegree'.")
+
+        # check that "spynapse_model" is there for data_connect
```
hakonsbm, line 148:
```suggestion
        # check that "synapse_model" is there for data_connect
```

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
     projections = fixdict(projections)
     sps(projections)
     sr('ConnectLayers')
+
+
+def _check_input_nodes(pre, post, conn_spec):
+    '''
+    Check the properties of `pre` and `post` nodes:
```
hakonsbm, line 255: This function does more than checking the properties, it also converts `pre` and `post` if needed. Perhaps it can be renamed to something like `_process_input_nodes`, as that would make it consistent with the helper functions for `syn_spec` and `conn_spec`.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
...
+
+    * if `conn_spec` is "one_to_one", skip uniqueness check for better speed
+      and use data_connect
```
hakonsbm, line 258: This is only true if pre or post is already an array, right?

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
...
+    * if both `pre` and `post` are NodeCollections or can be converted to
+      NodeCollections (i.e. contain unique IDs), then proceed to "normal"
+      connect (potentially after conversion to NodeCollection)
+    * if both `pre` and `post` are arrays and contain non-unique items, then
+      we proceed to "data_connect"
+    * if at least one of them has non-unique items and they have different
+      sizes, then raise an error.
```
hakonsbm, line 265: Only this bullet point has a period. I suggest you use periods and capital first letter for all bullet points.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
...
+    '''
```
hakonsbm, line 266: Docstrings should use tripple double quotes (`""" ... """`).

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
...
+    data_connect = False
+
+    pre_is_nc, post_is_nc = True, True
+
+    if not isinstance(pre, NodeCollection):
+        # skip uniqueness check for data_connect compatible `conn_spec`
+        if conn_spec != 'one_to_one' and len(set(pre)) == len(pre):
```
hakonsbm, line 273: What if `conn_spec` is a dictionary? Isn't it unexpected that calling 
```python
nest.Connect(n, n, conn_spec='one_to_one')
```
behaves in one way, while calling
```python
nest.Connect(n, n, conn_spec={'rule': 'one_to_one'})
```
behaves differently?

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
...
+    if not isinstance(post, NodeCollection):
+        # skip uniqueness check for data_connect compatible `conn_spec`
+        if conn_spec != 'one_to_one' and len(set(post)) == len(post):
+            post = NodeCollection(post)
+        else:
+            post_is_nc = False
+
+    if not pre_is_nc or not post_is_nc:
+        assert len(pre) == len(post), "If `pre` or `post` contain non-unique IDs, then " \
+                                      "they must have the same length."
```
hakonsbm, line 287: Raise an error instead of using an assert.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,61 @@ def fixdict(d):
...
+
+        # convert them to arrays
+        if not pre_is_nc and not isinstance(pre[0], (int, np.integer)):
+            raise NESTErrors.ArgumentType("Connect", " `pre` IDs should be integers.")
+
+        if not post_is_nc and not isinstance(post[0], (int, np.integer)):
+            raise NESTErrors.ArgumentType("Connect", " `post` IDs should be integers.")
+
+        pre = np.asarray(pre, dtype=int)
+        post = np.asarray(post, dtype=int)
```
hakonsbm, line 297: If you use implicit `dtype` in `np.asarray()`, you can check the `dtype` of the arrays afterwards to see if all IDs are integers.

pynest/nest/lib/hl_api_connections.py
```diff
@@ -200,69 +202,55 @@ def Connect(pre, post, conn_spec=None, syn_spec=None,
...
-            raise TypeError("Sources and targets must either both be NodeCollections, "
-                            "or NumPy arrays with conn_spec=None")
-        elif conn_spec is not None:
-            raise ValueError("When connecting two arrays of node IDs, conn_spec cannot be given")
-        elif not (pre.ndim == 1 and post.ndim == 1):
-            raise ValueError("Sources and targets must be 1-dimensional NumPy arrays")
-        else:
-            connect_np_arrays = True
-            conn_spec = 'one_to_one'
+    data_connect, pre, post = _check_input_nodes(pre, post, conn_spec)
```
hakonsbm, line 205: I think the variable name `data_connect` should be changed to better convey that it is a switch to use `connect_arrays()` instead of the normal SLI functions. Perhaps something like `use_connect_arrays` or `connect_arrays_needed`?

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,29 +20,52 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 import unittest
-import nest
 import numpy as np
 
+import nest
+from nest.lib.hl_api_exceptions import NESTErrors
```
hakonsbm, line 26: `NESTErrors` is already available, and used, as `nest.kernel.NESTErrors`. It's ok to import this, but the usage should be consistent.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,29 +20,52 @@
...
+
 nest.set_verbosity('M_WARNING')
 
 
+non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
```
hakonsbm, line 31: Move `non_unique` from global scope into `TestConnectArrays`, and use it as `self.non_unique`.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,29 +20,52 @@
...
+
+
 class TestConnectArrays(unittest.TestCase):
 
     def setUp(self):
         nest.ResetKernel()
 
-    def test_connect_arrays(self):
+    def test_connect_arrays_unique(self):
         """Connecting NumPy arrays of node IDs"""
```
hakonsbm, line 40: Update the docstring to reflect that this tests arrays of unique node IDs.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,29 +20,52 @@
...
+
+        nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays})
+
+        conns = nest.GetConnections()
+
+        self.assertEqual(len(conns), n*n)
+
+        for c in conns:
+            self.assertTrue(np.isclose(c.weight, weights))
+            self.assertTrue(np.isclose(c.delay, delays))
```
hakonsbm, line 56: To avoid the unhelpful error message `False is not True` in event of failure, don't use `self.assertTrue()`. One alternative is to use functions from [`numpy.testing`](https://numpy.org/doc/stable/reference/routines.testing.html) to assert the results.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,29 +20,52 @@
...
 
-        nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays,
-                                                 'synapse_model': syn_model})
+    def test_connect_arrays_nonunique(self):
+        """Connecting NumPy arrays of node IDs"""
```
hakonsbm, line 59: This docstring should also reflect that it's testing nonunique node IDs.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -51,19 +74,29 @@ def test_connect_arrays(self):
             self.assertEqual(c.weight, w)
             self.assertEqual(c.delay, d)
 
+    def test_connect_arrays_no_conn_spec(self):
+        """Connecting NumPy arrays of node IDs"""
```
hakonsbm, line 78: Update the docstring.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -51,19 +74,29 @@ def test_connect_arrays(self):
...
-        targets = np.arange(1, n+1, dtype=np.uint64)
+        targets = non_unique
         weights = np.ones(len(sources))
         delays = np.ones(len(sources))
         syn_model = 'static_synapse'
 
-        nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays,
-                                                 'synapse_model': syn_model})
+        nest.Connect(sources, targets, conn_spec="one_to_one",
+                     syn_spec={'weight': weights, 'delay': delays, 'synapse_model': syn_model})
```
hakonsbm, line 99: Use `'` or `"` for strings consistently throughout the file (docstrings are always with `"`s however).

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -167,44 +193,30 @@ def test_connect_arrays_float_rtype(self):
...
-            nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays, 'synapse_model': syn_model,
-                                                     'receptor_type': receptor_type
-                                                     })
+            nest.Connect(sources, targets, conn_spec="one_to_one",
+                         syn_spec={'weight': weights, 'delay': delays, 'synapse_model': syn_model,
+                                   'receptor_type': receptor_type})
 
     def test_connect_arrays_wrong_dtype(self):
-        """Raises exception when connecting NumPy arrays with wrong dtype"""
+        """Automatically correct NumPy arrays with wrong dtype"""
```
hakonsbm, line 208: This test still checks that an exception is raised.

pynest/nest/lib/hl_api_connection_helpers.py
Silmathoron: no, if one or both are ``NodeCollection`` objects, the whole test is skipped, so of course the uniqueness is also skipped (since we know the uniqueness has to be fulfilled by such an object)

hakonsbm: Yes, but if both are NodeCollections it can never use `connect_arrays`, as both `pre_is_nc` and `post_is_nc` will be `True`. This bullet point makes it seem like if `conn_spec` is "one_to_one", `connect_arrays` will always be used. It's the `and use data_connect` that I find confusing.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,70 @@ def fixdict(d):
...
+    * If at least one of them has non-unique items and they have different
+      sizes, then raise an error.
+    """
+    use_connect_arrays = False
+
+    # check for "one_to_one" conn_spec
+    one_to_one_cspec = (conn_spec == 'one_to_one')
+
+    if isinstance(conn_spec, dict):
+        one_to_one_cspec = (conn_spec.get('rule', 'all_to_all') == "one_to_one")
```
hakonsbm, line 273: Use `'` or `"` for strings consistently.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,55 +20,109 @@
...
+        for c in conns:
+            np.testing.assert_approx_equal(c.weight, weights)
+            np.testing.assert_approx_equal(c.delay, delays)
+
+    def test_connect_arrays_nonunique(self):
+        """Connecting NumPy arrays with non-unique node IDs"""
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = np.arange(1, n+1, dtype=np.uint64)
+        targets = self.non_unique
```
hakonsbm, line 61: It may be better to also have `n` and `np.arange(1, n+1, dtype=np.uint64)` as class variables.

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -80,15 +134,14 @@ def test_connect_arrays_no_delays(self):
         """Connecting NumPy arrays without specifying delays"""
         n = 10
         nest.Create('iaf_psc_alpha', n)
-        sources = np.arange(1, n+1, dtype=np.uint64)
-        targets = np.arange(1, n+1, dtype=np.uint64)
-        weights = np.ones(len(sources))
-        # delays = np.ones(len(sources))
-        syn_model = 'static_synapse'
+        sources = [i for i in range(1, n + 1)]
```
hakonsbm, line 137: It's better to use `list()` here.
```suggestion
        sources = list(range(1, n + 1))
```

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -80,15 +134,14 @@ def test_connect_arrays_no_delays(self):
...
+        targets = self.non_unique
+        weights = np.ones(n)
 
-        nest.Connect(sources, targets, syn_spec={'weight': weights, 'synapse_model': syn_model})
+        nest.Connect(sources, targets, conn_spec='one_to_one', syn_spec={'weight': weights})
```
hakonsbm, line 141: The intention of this test is to check that `connect_arrays` works when specifying only weights. Now, in addition to that, it checks that connecting a list to a NumPy array works. It's probably best to check that in a separate test.

pynest/nest/lib/hl_api_connection_helpers.py
Silmathoron: To me the important bit is to let people reading the docstring know when a unicity test will be performed and when not (for performance reason).
If I understand correctly, your issue is actually with

> and use data_connect

so should I just remove that part and put something like

> if `conn_spec` is "one_to_one", no uniqueness check is performed and regular one_to_one or data_connect is used depending on the input type

pynest/nest/tests/test_connect_arrays.py
Silmathoron: sorry, I don't have time for that ;)

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -81,33 +135,46 @@ def test_connect_arrays_no_delays(self):
...
             self.assertEqual(c.target, t)
             self.assertEqual(c.weight, w)
 
+    def test_connect_array_list(self):
+        """Connecting NumPy array and list"""
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = list(range(1, n + 1))
+        targets = self.non_unique
+        weights = np.ones(n)
```
hakonsbm, line 156: `weights` is unused and should be removed.

Silmathoron:
Ping @heplesser ;)

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
```diff
@@ -356,18 +356,17 @@ This variant of ``Connect()`` will create connections in a one-to-one fashion.
...
+All arrays have to have lengths equal to those of ``sources`` and ``targets``.
 
 ::
 
    weights = np.array([0.5, 0.5, 2., 2.])
    delays = np.array([1., 1., 2., 2.])
-   syn_spec = {'weight': weights, 'delay': delays, 'synapse_model': 'static_synapse'}
-   nest.Connect(sources, targets, syn_spec=syn_spec)
+   syn_spec = {'weight': weights, 'delay': delays}
+   nest.Connect(sources, targets, conn_spec="one_to_one", syn_spec=syn_spec)
```
heplesser, line 369:
```suggestion
   nest.Connect(sources, targets, conn_spec='one_to_one', syn_spec=syn_spec)
```
Use single hyphens for consistency with previous line.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -54,24 +55,30 @@ def _process_conn_spec(conn_spec):
...
+        # for use_connect_arrays, return "static_synapse" by default
+        if use_connect_arrays:
+            return {"synapse_model": "static_synapse"}
+
         return syn_spec
+
     rule = conn_spec['rule']
+
     if isinstance(syn_spec, str):
         return kernel.SLILiteral(syn_spec)
```
heplesser, line 70: Why is this case returning a literal, while all other cases return a dictionary, e.g. on line 61?

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -236,3 +248,71 @@ def fixdict(d):
...
+    * If at least one of them has non-unique items and they have different
+      sizes, then raise an error.
+    """
+    use_connect_arrays = False
+
+    # check for 'one_to_one' conn_spec
+    one_to_one_cspec = (conn_spec == 'one_to_one')
+
+    if isinstance(conn_spec, dict):
+        one_to_one_cspec = (conn_spec.get('rule', 'all_to_all') == 'one_to_one')
```
heplesser, line 274: Setting `one_to_one_cspec` first and then conditionally changing it right afterwards seems confusing. How about
```suggestion
    one_to_one_cspec = ( ( conn_spec if not isinstance(conn_spec, dict) 
                                            else conn_spec.get('rule', 'all_to_all') )
                                          == 'one_to_one' ) 
```
I intentionally put the `if not` in so the plain `conn_spec` comes first.

pynest/nest/lib/hl_api_connections.py
```diff
@@ -143,7 +144,8 @@ def Connect(pre, post, conn_spec=None, syn_spec=None,
     Notes
     -----
     It is possible to connect NumPy arrays of node IDs one-to-one by passing the arrays as `pre` and `post`,
-    with no connection specification, and a synapse specification, `syn_spec`, with at least the synapse model.
+    specifying "one_to_one" for `conn_spec`.
```
heplesser, line 147:
```suggestion
    specifying `'one_to_one'` for `conn_spec`.
```

pynest/nest/tests/test_connect_arrays.py
```diff
@@ -20,55 +20,109 @@
...
             self.assertEqual(c.weight, w)
             self.assertEqual(c.delay, d)
 
+    def test_connect_arrays_nonunique_dict_conn_spec(self):
+        """Connecting NumPy arrays with non-unique node IDs and conn_spec as a dict"""
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = np.arange(1, n+1, dtype=np.uint64)
+        targets = self.non_unique
+        weights = 2*np.ones(n)
```
heplesser, line 82: Spaces around `*`, also next line.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -54,24 +55,30 @@ def _process_conn_spec(conn_spec):
         raise TypeError("conn_spec must be a string or dict")
 
 
-def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, connect_np_arrays):
+def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, use_connect_arrays):
     """Processes the synapse specifications from None, string or dictionary to a dictionary."""
     if syn_spec is None:
+        # for use_connect_arrays, return "static_synapse" by default
+        if use_connect_arrays:
+            return {"synapse_model": "static_synapse"}
```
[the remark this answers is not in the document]
Silmathoron: This does not simplify because of the ``return syn_spec`` below, does it?
It would require an additional ``elif syn_spec is None`` below...
Would it be clearer if I changed ``return syn_spec`` into ``return None``?

pynest/nest/lib/hl_api_connection_helpers.py
Silmathoron: Dictionary is not necessary on line 61 "static_synapse" would be fine... should I change it?
Also line 65 is None, so not a dict either.

pynest/nest/lib/hl_api_connection_helpers.py
Silmathoron: Because I found the original structure more readable but I don't care much so I'll change it

pynest/nest/lib/hl_api_connections.py
Silmathoron: done

pynest/nest/tests/test_connect_arrays.py
Silmathoron: done

hakonsbm:
@Silmathoron @heplesser Connecting arrays is not tested with MPI, and it doesn't look like it works with MPI either. However I don't think fixing that is within the scope of this PR. I can create an issue.

heplesser:
> @Silmathoron @heplesser Connecting arrays is not tested with MPI, and it doesn't look like it works with MPI either. However I don't think fixing that is within the scope of this PR. I can create an issue.

@hakonsbm If we (for now or forever) do not support arrays when using MPI (I can see reasons for this), then we (a) need to provide proper error messages if a user tries to do so, (b) document this limitation clearly and (c) ideally have a test that a proper error is raised when a user tries to use arrays with MPI.

hakonsbm:
@heplesser I have created #1710 to document the bug with `connect_arrays`, where we also can discuss further if connecting with arrays should be supported when using MPI at all. And then I suggest that (a) and (b) is added here to not have this PR rely on resolving #1710 first.

Silmathoron:
MPI check and doc update done.

pynest/nest/lib/hl_api_connection_helpers.py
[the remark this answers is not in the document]
heplesser: @Silmathoron You are right, I had misunderstood the code. Would it then not be clearer to have an explicit `else` clause:
```python
if use_connect_arrays:
    return {"synapse_model": "static_synapse"}
else:
   return syn_spec
```
?

pynest/nest/lib/hl_api_connection_helpers.py
heplesser: I think code is more readable if a function returns the same type in all cases, with `None` as the one exception to signal special cases. Since some other cases further down must return dicts, I would suggest to always return a dict.

pynest/nest/lib/hl_api_connection_helpers.py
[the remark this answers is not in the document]
Silmathoron: I don't think so, the return-early pattern is widely advised as it avoid potential errors with forgotten if/else clauses and reduces the amount of code.
I would really recommend leaving it that way.

pynest/nest/lib/hl_api_connection_helpers.py
Silmathoron: Roger, I'll update the code

Silmathoron:
@heplesser up ;)

heplesser (approved):
@Silmathoron Thanks & well done! Sorry for the long delays in the review.

PR #1553, opened 2020-04-24
Replace Python 3.6 f-strings to achieve Python 3.5 compatibility

clinssen:
There were only two occurrences of Python 3.6+ f-string formatting. As we want to maintain Python 3.5 compatibility at least while 3.5 is officially supported, I have replaced the f-string formatting with good old concatenation (this seems to be the pattern grepping for NESTError exceptions).


heplesser:
Minimal fix, thus merging based on single review.

PR #1561, opened 2020-04-28
Remove redundant setting of environment variables

### LINKED ISSUE #972: `make installcheck` reaches wrong testing code after sourcing `nest_vars.sh` for a different version

clinssen:
Replaces #1545 and fixes #972.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,7 @@ file( MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/reports/ )
 
 add_custom_target( update_environment
   COMMAND bash -c "source ${PROJECT_BINARY_DIR}/extras/nest_vars.sh"
+  COMMENT "Preparing environment..."
 )
 
 if ( HAVE_PYTHON )
@@ -213,7 +214,7 @@ else ()
   )
 endif ()
 
-add_dependencies(installcheck update_environment)
+add_dependencies( installcheck update_environment )
 
 ################################################################################
 ##################        Define Subdirectories here          ##################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,11 +192,6 @@ enable_testing()
 set( TEST_LOGFILE ${PROJECT_BINARY_DIR}/reports/installcheck.log )
 file( MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/reports/ )
 
-add_custom_target( update_environment
-  COMMAND bash -c "source /home/travis/build/nest/nest-simulator/result/bin/nest_vars.sh"
-  COMMENT "Preparing environment (path: ${PROJECT_BINARY_DIR}/extras/nest_vars.sh)..."
-)
-
 if ( HAVE_PYTHON )
   add_custom_target( installcheck
     COMMAND ${CMAKE_COMMAND} -E env
@@ -213,8 +208,6 @@ else ()
     COMMENT "Execute NEST's testsuite...."
   )
 endif ()
-
-add_dependencies( installcheck update_environment )
 
 ################################################################################
 ##################        Define Subdirectories here          ##################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,12 +192,14 @@ enable_testing()
 set( TEST_LOGFILE ${PROJECT_BINARY_DIR}/reports/installcheck.log )
 file( MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/reports/ )
 
+add_custom_target( update_environment
+  COMMAND bash -c "source ${CMAKE_INSTALL_FULL_DATADIR}/bin/nest_vars.sh"
+)
+
 if ( HAVE_PYTHON )
   add_custom_target( installcheck
     COMMAND ${CMAKE_COMMAND} -E env
-      NEST_PATH=${CMAKE_INSTALL_FULL_BINDIR}
       PYTHON="${PYTHON}"
-      NEST_PYTHONPATH="${CMAKE_INSTALL_PREFIX}/${PYEXECDIR}"
       ${CMAKE_INSTALL_FULL_DATADIR}/extras/do_tests.sh --test-pynest
     WORKING_DIRECTORY "${PROJECT_BINARY_DIR}"
     COMMENT "Execute NEST's testsuite...."
@@ -205,12 +207,13 @@ if ( HAVE_PYTHON )
 else ()
   add_custom_target( installcheck
     COMMAND ${CMAKE_COMMAND} -E env
-      NEST_PATH=${CMAKE_INSTALL_FULL_BINDIR}
       ${CMAKE_INSTALL_FULL_DATADIR}/extras/do_tests.sh
     WORKING_DIRECTORY "${PROJECT_BINARY_DIR}"
     COMMENT "Execute NEST's testsuite...."
   )
 endif ()
+
+add_dependencies(installcheck update_environment)
 
 ################################################################################
 ##################        Define Subdirectories here          ##################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,6 +211,13 @@ else ()
   )
 endif ()
 
+# N.B. to ensure "make install" is always run before "make installcheck", we
+# would ideally like to add:
+#   add_dependencies( installcheck install )
+# However, issue in CMake at time of writing (May 2020, see
+# https://gitlab.kitware.com/cmake/cmake/-/issues/8438) precludes us from doing
+# so.
+
 ################################################################################
 ##################        Define Subdirectories here          ##################
 ################################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ set( TEST_LOGFILE ${PROJECT_BINARY_DIR}/reports/installcheck.log )
 file( MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/reports/ )
 
 add_custom_target( update_environment
-  COMMAND bash -c "source ${PROJECT_BINARY_DIR}/extras/nest_vars.sh"
+  COMMAND bash -c "source /home/travis/build/nest/nest-simulator/result/bin/nest_vars.sh"
   COMMENT "Preparing environment (path: ${PROJECT_BINARY_DIR}/extras/nest_vars.sh)..."
 )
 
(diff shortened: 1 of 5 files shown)


clinssen:
(Please see prior discussion in #1545.)

The underlying issue was as follows: the binaries could find the libraries (``libnestkernel.so`` and so on) because their runpath includes ``../lib/nest``.

``pynestkernel.so`` could find the libraries because its runpath includes ``../../../lib/nest``.

But the libraries themselves (e.g. ``libnestkernel.so``) could not find the other libraries (e.g. ``libsli.so``) because ``.`` (or, indirectly, ``../../lib/nest``, which from the perspective of these libraries points back to ``.``), was missing. The latest commit adds the latter to the RPATH.

extras/travis_build.sh
```diff
@@ -282,8 +280,6 @@ echo "MSGBLD0270: Running make install."
 make install
 echo "MSGBLD0280: Make install completed."
 
-source $NEST_RESULT/bin/nest_vars.sh
-
```
jougs, line 286: Instead of removing this line, I would prefer if
* this line were moved to around line 198 and 208 of the toplevel `CMakeLists.txt`
* the setting of `NEST_PATH` and `NEST_PYTHONPATH` were removed from around there
* the explicit setting of `PYTHONPATH` and `PATH` in line 352-356 of `testsuite/do_tests.sh` were removed.

In general, the two path variables should never be set explicitly, but `nest_vars.sh` should be sourced instead, in order to be consistent everywhere.

CMakeLists.txt
```diff
@@ -192,26 +192,30 @@ enable_testing()
...
     COMMAND ${CMAKE_COMMAND} -E env
-      NEST_PATH=${CMAKE_INSTALL_FULL_BINDIR}
       ${CMAKE_INSTALL_FULL_DATADIR}/extras/do_tests.sh
     WORKING_DIRECTORY "${PROJECT_BINARY_DIR}"
     COMMENT "Execute NEST's testsuite...."
   )
 endif ()
 
+add_dependencies( installcheck update_environment )
+
```
jougs, line 218: Beautiful!

heplesser: This makes me wonder if `installcheck` should not also depend on `install`? Currently, if you invoke `make installcheck` before `make install`,  you get a rather incomprehensible error message. And `update_environment` depends on `install`. Can we express this also explicitly here?

gtrensch:
@clinssen 
This change causes `make installcheck` to behave like if it is running on Travis.
`bash: /home/travis/build/nest/nest-simulator/result/bin/nest_vars.sh: No such file or directory`
The user specified install directory seems to be ignored.

clinssen:
I really appreciate all the feedback so far, but we are still in draft mode and I'm still experimenting with possible solutions!

@jougs: regrettably, it seems that the adding a custom dependency option will not work. The environment sourced in the "update_environment" custom CMake target does not carry over to the environment in which ``do_tests.sh`` will be run.

It looks like the invocation in "make installcheck" could be changed from just invoking ``do_tests.sh`` to invoking ``bash -c "source nest_vars.sh && do_tests.sh``. I tested this and it works.

@heplesser: I can still add the dependency of installcheck on install without a problem.

heplesser:
> It looks like the invocation in "make installcheck" could be changed from just invoking `do_tests.sh` to invoking `bash -c "source nest_vars.sh && do_tests.sh`. I tested this and it works.

This looks like a clean solution.
 
> @heplesser: I can still add the dependency of installcheck on install without a problem.

Good.

jougs:
Just a quick heads up: I decided to resect my solution to #345 out of #1404 and make it into a new standalone PR. With that finished, `do_test.sh` will set paths according to what it is given on the command line instead of relying on the environment. This means that this PR does not have to handle that at all.

clinssen:
Thanks Jochen, that simplifies things. Then hopefully you won't mind if I crudely insert a ``source ${CMAKE_INSTALL_FULL_BINDIR}/nest_vars.sh`` in ``do_tests.sh.in`` as a temporary solution.

There is one related issue that I'm now running into. After #1513, we are falsely advertising ``nest_vars.sh`` as being POSIX-compliant. Actually, for the MacOS check, we use ``$OSTYPE`` and the ``[[ ]]`` operators, which are not in ``/bin/sh``. I have replaced this with a more compliant check using ``uname``, so we wouldn't have to "downgrade" ``/bin/sh`` to ``/bin/bash`` in ``nest_vars.sh``, and by extension in ``do_tests.sh``.

@heplesser: adding a dependency of "installtarget" on "install" turns out to be not supported by CMake, because "install" is a built-in target. They've had an issue open for this for over 10 years (https://gitlab.kitware.com/cmake/cmake/-/issues/8438). There are no workarounds that I could find.

jougs:
> Thanks @jougs, that simplifies things. Then hopefully you won't mind if I crudely insert a `source ${CMAKE_INSTALL_FULL_BINDIR}/nest_vars.sh` in `do_tests.sh.in` as a temporary solution.

I do exactly the same in my refactoring branch :smile: 

> There is one related issue that I'm now running into. After #1513, we are falsely advertising `nest_vars.sh` as being POSIX-compliant. Actually, for the MacOS check, we use `$OSTYPE` and the `[[ ]]` operators, which are not in `/bin/sh`. I have replaced this with a more compliant check using `uname`, so we wouldn't have to "downgrade" `/bin/sh` to `/bin/bash` in `nest_vars.sh`, and by extension in `do_tests.sh`.

This sounds good to me.

> @heplesser: adding a dependency of "installtarget" on "install" turns out to be not supported by CMake, because "install" is a built-in target. They've had an issue open for this for over 10 years (https://gitlab.kitware.com/cmake/cmake/-/issues/8438). There are no workarounds that I could find.

Probably this is the reason why it is not implemented. Can you please add a corresponding comment, so we don't have to re-discover this each second year? Thanks!

heplesser:
@clinssen Let me know when this PR can be moved out of Draft state!

clinssen:
@heplesser: it's ready for review! Could you double-check the new MacOS test in ``nest_vars.sh``? Cheers!

CMakeLists.txt
```diff
@@ -197,23 +197,27 @@ file( MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/reports/ )
...
       ${CMAKE_INSTALL_FULL_DATADIR}/extras/do_tests.sh
     WORKING_DIRECTORY "${PROJECT_BINARY_DIR}"
     COMMENT "Execute NEST's testsuite...."
   )
 endif ()
 
+# N.B. to ensure "make install" is always run before "make installcheck", we
+# would ideally like to add:
+#   add_dependencies( installcheck install )
+# However, issue in CMake at time of writing (May 2020, see
```
jougs, line 217:
```suggestion
# However, an issue in CMake at time of writing (May 2020, see
```

cmake/ProcessOptions.cmake
```diff
@@ -221,8 +221,22 @@ function( NEST_PROCESS_STATIC_LIBRARIES )
...
+      # is relative to this origin, so the binary ``bin/nest`` can find the files in the relative location
+      # ``../lib/nest``, and similarly for PyNEST and the other libraries. For simplicity, we set all the possibilities
+      # on all generated objects.
+      #
+      # PyNEST can only act as an entry point; it does not need to be included in the other objects' RPATH itself.
       set( CMAKE_INSTALL_RPATH
+          # for binaries
           "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}/nest"
+          # for libraries (except pynestkernel)
+          "\$ORIGIN/../../${CMAKE_INSTALL_LIBDIR}/nest"
```
jougs, line 239: 1. If this is for our own libraries, shouldn't this always come out to just this?
    ```suggestion
              "\$ORIGIN"
    ```
2. Aren't similar changes required for the APPLE branch of the `if` above?

extras/travis_build.sh
```diff
@@ -108,7 +106,11 @@ if [ "$xLIBNEUROSIM" = "1" ] ; then
     CONFIGURE_LIBNEUROSIM="-Dwith-libneurosim=$HOME/.cache/libneurosim.install"
     chmod +x extras/install_csa-libneurosim.sh
     ./extras/install_csa-libneurosim.sh
-    export LD_LIBRARY_PATH=$HOME/.cache/csa.install/lib:$LD_LIBRARY_PATH
+    if [[ $OSTYPE == darwin* ]]; then
```
jougs, line 109: Shouldn't this also rather use `uname` as above?

extras/travis_build.sh
jougs: The new solution is even better!

cmake/ProcessOptions.cmake
clinssen: 1. Yes! I was slightly weary about adding "." as the search path, because it might be overly broad. This circumlocuitous path amounts to the same, but will for sure only pick up the NEST libraries.
2. You are quite right! Commit incoming.

extras/travis_build.sh
clinssen: Not necessary as travis_build only needs to run on Bash, not pure POSIX shells (/bin/sh). I could replace all the checks, but I'm afraid of feature creep and breaking stuff. However, if you feel it is neater and should be included in this PR I will add it.

jougs: Thanks for the clarification. No need for changes from my side.

cmake/ProcessOptions.cmake
```diff
@@ -213,25 +213,29 @@ function( NEST_PROCESS_STATIC_LIBRARIES )
     # (but later on when installing)
     set( CMAKE_BUILD_WITH_INSTALL_RPATH FALSE PARENT_SCOPE )
 
-    # set the rpath only when installed
+    # set run-time search path (RPATH) so that dynamic libraries in ``lib/nest`` can be located
+
+    # Note: "$ORIGIN" (on Linux) and "@loader_path" (on MacOS) are not CMake variables, but special keywords for the
+    # Linux resp. the MacOS dynamic loader. They refer to the path in which the object is located, e.g.
```
jougs, line 219:
```suggestion
    # Linux resp. the macOS dynamic loader. They refer to the path in which the object is located, e.g.
```

cmake/ProcessOptions.cmake
jougs: Thanks for your reply.

jougs:
@gtrensch, just to be on the safe side: Does this now solve all issues you described and observed?

PR #1564, opened 2020-04-29
Multimeter records at correct time when offset is specified

### LINKED ISSUE #1508: Multimeter sampling with offset parameter
**Description: The multimeter returns unexpected samples for certain `interval` and `offset` parameter combinations**

The multimeter records an unexpected first sample when it is created and connected after some initial simulation period. The multimeter documentation is not complete, but if I remember correctly, we implemented the `offset` parameter #669 to record the first sample relative to `0.`, including a test meant to verify exactly such scenarios with initial simulation time and later recorder creation. However, the test fails to cover some more specific scenarios.

**To Reproduce**
There are 2 slightly different examples of strange recordings, both adapted from the existing [unittest](https://github.com/nest/nest-simulator/blob/v2.20.0/testsuite/unittests/test_multimeter_offset.sli). Steps to reproduce the behavior:

### Scenario 1
```
import nest

simtimeinit = 10
simtime = 20.
recoffs = 1.
recint = 3.
recstart = 10.  # this is not important, remains for comparison with the unittest
recstop = 30.

resolution = 0.1

nest.SetKernelStatus({'resolution': resolution})
neuron = nest.Create('iaf_psc_exp')

mm1 = nest.Create("multimeter", 1)
nest.SetStatus(mm1, {"withtime": True, "record_from": ["V_m"],
                            'interval': recint, 'offset': recoffs,
                            'start': recstart, 'stop': recstop})
nest.Connect(mm1, neuron, syn_spec={'delay': 0.1})

nest.Simulate(simtimeinit)

mm2 = nest.Create("multimeter", 1)
nest.SetStatus(mm2, {"withtime": True, "record_from": ["V_m"],
                            'interval': recint, 'offset': recoffs,
                            'start': recstart, 'stop': recstop})
nest.Connect(mm2, neuron, syn_spec={'delay': 0.1})

nest.Simulate(simtime)
times = nest.GetStatus(mm1)[0]['events']['times']
times2 = nest.GetStatus(mm2)[0]['events']['times']
print(times)
print(times2)
```

**Expected behavior**
Returns
```
[13. 16. 19. 22. 25. 28.]
[10.1 13.  16.  19.  22.  25.  28. ]
```
even though the two outputs should be equal. The sample at `10.1` looks incorrect. Note that this error depends on `simtimeinit`, e.g., it only occurs for `simtimeinit >= 10.0`.

### Scenario 2
```
import nest
resolution = 0.1
interval = 60.
offset = 170.3

initial_sim_time = 110.

nest.SetKernelStatus({'resolution': resolution})
neuron = nest.Create('iaf_psc_exp')
##################
nest.Simulate(initial_sim_time)

multimeter = nest.Create("multimeter", 1)
nest.SetStatus(multimeter, {"withtime": True, "record_from": ["V_m"],
                            'interval': interval, 'offset': offset})
nest.Connect(multimeter, neuron, syn_spec={'delay': 0.1})
nest.Simulate(250.)
times = nest.GetStatus(multimeter)[0]['events']['times']
print(times[0])
```
**Expected behavior**
Returns `230.3`, instead of `170.3` if the offset was relative to `0.` There seems to be a relation between the recording interval, offset, and the initial simulation time before creation. For this case, the first sample if off by exactly 1 interval (60.0 ms), but if we set `initial_sim_time = 121.` (so larger than `2*interval`, the first sample is recorded at `290.3`, so it'd be off by twice the interval.

 - OS: Linux Mint 19.2 Cinnamon
 - Python-Version: 3.6
 - NEST-Version: 2.20.0, 2.18.0
 - Installation: barebone (no conda), with MPI

**Additional context**
I'm not sure the behaviour in the two scenarios are related, or whether this is an issue in the `Multimeter` or the `UniversalDataLogger`. If somehow these are correct behaviours, the documentation should be extended accordingly.

stinebuu:
This PR fixes #1508.

When a `offset` is specified for the `multimeter` we set next recording step to be `offset` and then add `interval` until next recording step has gotten up to current simulation step. This way, if we have specified `offset` and `interval` for the multimeter, and done an initial simulation, the `multimeter` starts recording at the correct step.


nestkernel/universal_data_logger_impl.h
```diff
@@ -117,11 +117,16 @@ nest::DynamicUniversalDataLogger< HostNode >::DataLogger_::init()
...
-  // if offset is not 0, adjust next recording step to account for it by
-  // going one interval step back and adding the offset
+  // If offset is not 0, adjust next recording step to account for it by
+  // first setting next recording step to be offset and then iterating until
+  // the variable is greater than current simulation time.
   if ( recording_offset_.get_steps() != 0 )
   {
-    next_rec_step_ = next_rec_step_ - rec_int_steps_ + recording_offset_.get_steps();
+    next_rec_step_ = recording_offset_.get_steps() - 1; // shifted one to left
+    while ( next_rec_step_ <= kernel().simulation_manager.get_time().get_steps() )
```
heplesser, line 126: Do we need to loop here? Couldn't we just calculate the correct next_rec_step_?

nestkernel/universal_data_logger_impl.h
```diff
@@ -318,11 +323,16 @@ nest::UniversalDataLogger< HostNode >::DataLogger_::init()
...
```
heplesser, line 332: This looks like code duplication with line 120 ff?

nestkernel/universal_data_logger_impl.h
stinebuu: @heplesser I tried for a really long time to calculate the correct `next_rec_step_`, but I never managed to find a calculation that fit for all the different outcomes. If you have a suggestion I could try that of course.

nestkernel/universal_data_logger_impl.h
stinebuu: @heplesser Yes it is, but it line 120 is in another function.

nestkernel/universal_data_logger_impl.h
zbarni: I think it's possible to calculate it in one step for all cases, with something along the lines:
```
next_rec_step_ = recording_offset_.get_steps() - 1; // shifted one to left
tmp_next_step_ = ceil( ( kernel().simulation_manager.get_time().get_steps() - next_rec_step_) /  rec_int_steps_ ) * rec_int_steps_;
if ( tmp_next_step_ > 0. )  // needed as tmp_next_step_ can be negative, in which case no update is required
{
  next_rec_step_ += tmp_next_step_;
}
```
using `ceil()` in case the division is rounded down..

nestkernel/universal_data_logger_impl.h
zbarni: check if the loop can be indeed replaced?

testsuite/unittests/test_multimeter_offset.sli
```diff
@@ -34,6 +34,11 @@ Description:
   2) check if data logger is initialized correctly
   3) ensures offset can only be set before the multimeter is connected to nodes
   4) ensures offset works correctly even if origin, start and stop are modified
+  5) ensures offsets behave correctly when we create a second multimeter after an
+     initial similation
+  6) ensures that offset behave correctly after initial simulation time
+  7) ensures that multimeter with offset behave correctly after initial simulation
+     time longer than offset
 
 Author: Barna Zajzon, 2017-03-10
```
zbarni, line 43: don't forget to add your name, it's teamwork ;)

nestkernel/universal_data_logger_impl.h
stinebuu: @zbarni unfortunately the calculation does not work. Test 2 in [test_multimeter_offset](https://github.com/nest/nest-simulator/pull/1564/files#diff-de974299a71b7711c22611787ba866a0R90) get an assert error. I don't really understand why. Test 5 also fails, and this is because it records at wrong times. The test creates a multimeter with `interval=3, offset = 1` and `start=10`, and then it simulates for `10` ms. Then, it creates a second multimeter, with the same `interval, offset` and `start`, and then simulates for 20 ms. The second multimeter should have a `next_rec_time_` of `129`, because it should start recording at `100` and interval is `30` and offset is `10`, but with the new calculation `next_rec_time_` is `99`, and then the multimeter will start recording before it should. I guess something similar is happening with test 2.
When I tried calculating, I often managed to get the correct `next_rec_time_` for some of the cases, but never for all.

jougs:
I trust @zbarni and @heplesser to handle this without me. I'm thus removing the request for review for myself.

nestkernel/universal_data_logger_impl.h
heplesser: @stinebuu Shouldn't we then factor out this code into a helper method which can be called from both places?

heplesser (changes requested):
@stinebuu Fine, just one little refactoring suggestion left.

stinebuu:
@heplesser I agree that refactoring the duplicate code would absolutely be best. I am having a little problem with it though, because the duplicate code occurs in two different classes(`UniversalDataLogger` and `DynamicUniversalDataLogger`), and they don't inherit from one another. Both have implemented a `DataLogger_` class as a class member, and the duplicate lines occurs in these classes `init()` functions. And I don't quite know how to create a helper function that is able to access the member variables of the individual class members.

https://github.com/nest/nest-simulator/blob/54c206f7b2056ffbf74babebedb69223924ba7b9/nestkernel/universal_data_logger.h#L38-L45

heplesser:
@stinebuu You are right. Actually, the real issue here is that we have two almost identical `DataLogger_` private subclasses in `UniversalDataLogger` and `DynamicDataLogger`. I wonder if one could avoid *that* duplication through some wild templating, but that would definitely be a new PR. So I approve this now.

heplesser (approved):

heplesser:
@zbarni Are you happy as well?

zbarni (approved):
Looks good to me, thanks for taking care of this!

PR #1567, opened 2020-04-29
Update documentation on adding spatial masks in user-defined modules

### LINKED ISSUE #1476: Bring documentation on adding spatial masks in user-defined modules up-to-date
**Describe the bug**
Documentation on how to add spatial masks through user-defined modules is outdated, see https://github.com/nest/nest-simulator/pull/1468#issuecomment-600634502.

**Expected behavior**
Documentation is in sync with current code.

**Additional context**
A PR to fix this should only be created after #1468 has been merged.

hakonsbm:
Fixes #1476.

### DIFF
--- a/doc/guides/spatial/guide_spatially_structured_networks.rst
+++ b/doc/guides/spatial/guide_spatially_structured_networks.rst
@@ -1449,7 +1449,7 @@ Developer Manual. The basic steps required to get started are:
 3. Configure. The configure process uses the script ``nest-config`` to
    find out where NEST is installed, where the source code resides, and
    which compiler options were used for compiling NEST. If
-   ``nest-config`` is not in your path, you need to provided it
+   ``nest-config`` is not in your path, you need to provide it
    explicitly with ``-Dwith-nest`` like this
 
    .. code:: bash
@@ -1478,64 +1478,78 @@ elliptic mask by creating a class called ``EllipticMask``. Note that
 elliptical masks are already part of NEST see
 Sec. \ :ref:`3.3 <sec:conn_masks>`. That elliptical mask is defined in a
 different way than what we will do here though, so this can still be
-used as an introductory example. First, we must include another header
-file:
+used as an introductory example. First, we must include the header
+files for the ``Mask`` parent class:
 
 .. code:: c
 
    #include "mask.h"
+   #include "mask_impl.h"
 
 The ``Mask`` class has a few methods that must be overridden:
 
 .. code:: c
 
-     class EllipticMask : public nest::Mask<2>
+   class EllipticMask : public nest::Mask< 2 >
+   {
+   public:
+     EllipticMask( const DictionaryDatum& d )
+       : rx_( 1.0 )
+       , ry_( 1.0 )
      {
-     public:
-       EllipticMask(const DictionaryDatum& d):
-         rx_(1.0), ry_(1.0)
-         {
-           updateValue<double>(d, "r_x", rx_);
-           updateValue<double>(d, "r_y", ry_);
-         }
+       updateValue< double >( d, "r_x", rx_ );
+       updateValue< double >( d, "r_y", ry_ );
+     }
 
-       using Mask<2>::inside;
+     using Mask< 2 >::inside;
 
-       // returns true if point is inside the ellipse
-       bool inside(const nest::Position<2> &p) const
-         { return p[0]*p[0]/rx_/rx_ + p[1]*p[1]/ry_/ry_ <= 1.0; }
+     // returns true if point is inside the ellipse
+     bool
+     inside( const nest::Position< 2 >& p ) const
+     {
+       return p[ 0 ] * p[ 0 ] / rx_ / rx_ + p[ 1 ] * p[ 1 ] / ry_ / ry_ <= 1.0;
+     }
 
-       // returns true if the whole box is inside the ellipse
-       bool inside(const nest::Box<2> &b) const
-         {
-           nest::Position<2> p = b.lower_left;
+     // returns true if the whole box is inside the ellipse
+     bool
+     inside( const nest::Box< 2 >& b ) const
+     {
+       nest::Position< 2 > p = b.lower_left;
 
-           // Test if all corners are inside mask
-           if (not inside(p)) return false;       // (0,0)
-           p[0] = b.upper_right[0];
-           if (not inside(p)) return false;       // (0,1)
-           p[1] = b.upper_right[1];
-           if (not inside(p)) return false;       // (1,1)
-           p[0] = b.lower_left[0];
-           if (not inside(p)) return false;       // (1,0)
+       // Test if all corners are inside mask
+       if ( not inside( p ) )
+         return false; // (0,0)
+       p[ 0 ] = b.upper_right[ 0 ];
+       if ( not inside( p ) )
+         return false; // (0,1)
+       p[ 1 ] = b.upper_right[ 1 ];
+       if ( not inside( p ) )
+         return false; // (1,1)
+       p[ 0 ] = b.lower_left[ 0 ];
+       if ( not inside( p ) )
+         return false; // (1,0)
 
-           return true;
-         }
+       return true;
+     }
 
-       // returns bounding box of ellipse
-       nest::Box<2> get_bbox() const
-         {
-           nest::Position<2> ll(-rx_,-ry_);
-           nest::Position<2> ur(rx_,ry_);
-           return nest::Box<2>(ll,ur);
-         }
+     // returns bounding box of ellipse
+     nest::Box< 2 >
+     get_bbox() const
+     {
+       nest::Position< 2 > ll( -rx_, -ry_ );
+       nest::Position< 2 > ur( rx_, ry_ );
+       return nest::Box< 2 >( ll, ur );
+     }
 
-       nest::Mask<2> * clone() const
-         { return new EllipticMask(*this); }
+     nest::Mask< 2 >*
+     clone() const
+     {
+       return new EllipticMask( *this );
+     }
 
-     protected:
-       double rx_, ry_;
-     };
+   protected:
+     double rx_, ry_;
+   };
 
 The overridden methods include a test if a point is inside the mask, and
 for efficiency reasons also a test if a box is fully inside the mask. We
@@ -1543,15 +1557,13 @@ implement the latter by testing if all the corners are inside, since our
 elliptic mask is convex. We must also define a function which returns a
 bounding box for the mask, i.e. a box completely surrounding the mask.
 
-.. TODO: how does registering work with mymodule in 3.0?
-
 The mask class must then be registered with the topology module, and this
 is done by adding a line to the function ``MyModule::init()`` in the file
 ``mymodule.cpp``:
 
 .. code:: c
 
-       nest::TopologyModule::register_mask<EllipticMask>("elliptic");
+   nest::TopologyModule::register_mask< EllipticMask >( "elliptic" );
 
 After compiling and installing the module, the mask is available to be
 used in connections, e.g.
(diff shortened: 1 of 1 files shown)


doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1431,46 +1431,45 @@ module, see the section titled `“Writing an Extension Module”
 <http://nest.github.io/nest-simulator/extension_modules>`__ in the NEST
 Developer Manual. The basic steps required to get started are:
 
-1. From the NEST source directory, copy directory examples/MyModule to
+1. From the NEST source directory, copy the directory examples/MyModule to
```
sarakonradi, line 1434:
```suggestion
1. From the NEST source directory, copy the directory ``examples/MyModule`` to
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1431,46 +1431,45 @@ module, see the section titled `“Writing an Extension Module”
...
 
-2. Change to the new location of MyModule and prepare by issuing
-   ``./bootstrap.sh``
+   .. code:: bash
+
+      cp -r <path_to_nest_source>/examples/MyModule mm_src
 
-3. Leave MyModule and create a build directory for it, e.g., mmb next to
-   it
+2. Create a build directory for MyModule, e.g. mm_bld, next to
```
sarakonradi, line 1441:
```suggestion
2. Create a build directory for MyModule, e.g. ``mm_bld``, next to
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1431,46 +1431,45 @@ module, see the section titled `“Writing an Extension Module”
...
+      cd mm_bld
 
-4. Configure. The configure process uses the script ``nest-config`` to
+3. Configure. The configure process uses the script ``nest-config`` to
    find out where NEST is installed, where the source code resides, and
    which compiler options were used for compiling NEST. If
-   ``nest-config`` is not in your path, you need to provided it
-   explicitly like this
+   ``nest-config`` is not in your path, you need to provide it
+   explicitly with ``-Dwith-nest`` like this
```
sarakonradi, line 1453:
```suggestion
   explicitly with ``-Dwith-nest`` as follows:
```

hakonsbm:
> Thanks! Just added some minor suggestions.

@sarakonradi Thanks, I've committed them. :relaxed: 

> @hakonsbm @stinebuu Should we generally replace MyModule with `MyModule`?

I don't really have a strong opinion about this. But my impression is that `MyModule` refers to something in the code, while MyModule refers to the module itself. So to me which one to use would depend on the context. I don't think we use `NodeCollection` generally in the documentation either.

stinebuu:
@sarakonradi @hakonsbm I have no strong opinion on `MyModule` vs MyModule either. I think we discussed going through all documentation and using cursive *NodeCollection* and *SynapseCollection* once, but it was discussed in a PR and we never really followed up. I would be ok with the change.

hakonsbm:
> What do you think about mentioning that you can use `nest.logical.conditional()` and `nest.spatial.distance` to create a connection parameter that behaves like a mask? And then add an example, but state that at the time of writing it is a lot more efficient to create a C++ mask through MyModule. I think the `Adding Masks` section should then be divided into two subsections, one with Parameter and one with MyModule. 

@stinebuu That's a good idea, I'll add it.

> I am a bit torn on my own idea though, on one side I think it is nice to show everything that is possible to the users, but on the other side I don't want them to take the easy way out with the Parameters as it is more efficient with the (slightly more complicated) MyModule..

I think if we show that it's possible to use `nest.logical.conditional()` and `nest.spatial.distance` to get the same behaviour, but say that it is less efficient than creating a C++ module, the users can make an informed decision about which method to use.

sarakonradi:
> @stinebuu That's a good idea, I can add it (unless you have any objections, @sarakonradi?).

I think this is a good idea too. I believe what matters is that we explain the possibility of both options in the documentation and make the trade-off clear. As @hakonsbm suggested, this will then allow the user to decide.

> I don't really have a strong opinion about this. But my impression is that MyModule refers to something in the code, while MyModule refers to the module itself. So to me which one to use would depend on the context. I don't think we use NodeCollection generally in the documentation either.

I agree. It seems that, in this particular context, the focus is on the module itself. Right? In that case, we can keep the format as it is.

sarakonradi:
> @sarakonradi Thanks, I've committed them.

Thank you for the changes!

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1431,46 +1431,45 @@ module, see the section titled `“Writing an Extension Module”
...
+   ``nest-config`` is not in your path, you need to provide it
+   explicitly with ``-Dwith-nest`` as follows:
 
    .. code:: bash
 
-      cmake -Dwith-nest=${NEST_INSTALL_DIR}/bin/nest-config ../MyModule
+      cmake -Dwith-nest=${NEST_INSTALL_DIR}/bin/nest-config ../mm_src
 
-5. MyModule will then be installed to ``\${NEST_INSTALL_DIR}``. This
+4. MyModule will then be installed to ``\${NEST_INSTALL_DIR}``. This
```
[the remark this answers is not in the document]
hakonsbm: Yep, fixed.

hakonsbm:
@sarakonradi @stinebuu I have split the *Adding masks* chapter in two now, describing how to do it with Parameters as well as with MyModule. Please have another look.

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
-.. TODO: adding masks need to be updated for 3.0?
+.. _ch:custom_masks:
+
+Creating custom masks
+---------------------
+
+In some cases, the built-in masks may not meet your needs, and you want to
+create a custom mask. There are two ways to do this: the first is to use Parameters
+to introduce a cut-off to the connection probability. The second is to implement a custom
+mask in C++ as a module.
```
sarakonradi, line 1426:
```suggestion
In some cases, the built-in masks may not meet your needs, and you want to
create a custom mask. There are two ways to do this: 

1. To use Parameters to introduce a cut-off to the connection probability. 
2. To implement a custom mask in C++ as a module.
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
+some additional steps for installation.
 
-.. _ch:extending:
+.. _sec:maskparameter:
+
+Using Parameters to specify connection boundaries
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can use Parameters that represent spatial distances between nodes to create a connection
+probability with mask behaviour. For this you need to create a condition Parameter that describes
```
sarakonradi, line 1444:
```suggestion
probability with mask behaviour. For this, you need to create a condition Parameter that describes
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
+the boundary of the mask. As condition Parameters evaluate to either ``0`` or ``1``, it can be
+used alone, used with the ``nest.logic.conditional()`` Parameter, or multiplied with another
```
sarakonradi, line 1446:
```suggestion
used alone, with the ``nest.logic.conditional()`` Parameter, or multiplied with another
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
+Parameter or value, before passing it as the connection probability.
+
+As an example, we will define an elliptical connection boundary using Parameters. First we define
```
sarakonradi, line 1449:
```suggestion
As an example, we will define an elliptical connection boundary using Parameters. First, we define
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
+   ry = 0.25  # radius in the y-direction
 
-This chapter will show examples of how to extend NEST by adding custom
-masks. Some knowledge of the C++ programming language is needed for this.
-The functions will be added as a part of an extension module which is
-dynamically loaded into NEST. For more information on writing an extension
-module, see the section titled `“Writing an Extension Module”
-<http://nest.github.io/nest-simulator/extension_modules>`__ in the NEST
-Developer Manual. The basic steps required to get started are:
+Next we define the connection boundary. We only want to connect to targets inside an ellipse, so
```
sarakonradi, line 1457:
```suggestion
Next, we define the connection boundary. We only want to connect to targets inside an ellipse, so
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
+::
+
+   x = nest.spatial.distance.x
+   y = nest.spatial.distance.y
+   lhs = x * x / rx**2 + y * y / ry**2
+   mask_param = nest.logic.conditional(lhs <= 1.0, 0.5, 0.0)
+   # Could also have defined it as
+   # mask_param = 0.5*(lhs <= 1.0)
+
+Then we can use the Parameter as connection probability when connecting populations with spatial
```
sarakonradi, line 1476:
```suggestion
Then, we can use the Parameter as connection probability when connecting populations with spatial
```

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,123 @@ center neuron are shown, as well as mask and connection probability.
...
+
+Creating custom masks
+---------------------
+
+In some cases, the built-in masks may not meet your needs, and you want to
+create a custom mask. There are two ways to do this: the first is to use Parameters
+to introduce a cut-off to the connection probability. The second is to implement a custom
+mask in C++ as a module.
+
+Using Parameters is the most accessible option; the entire implementation is done on
```
sarakonradi, line 1428: Could we write ``parameters`` in lowercase throughout the guide? That's the letter case we have also used in the "NEST 3.0: What's new?" guide.

hakonsbm: Done.

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,124 @@ center neuron are shown, as well as mask and connection probability.
...
+
+where :math:`x` and :math:`y` are the distances between the source and target neuron, in x- and
+y-directions, respectively. We use this expression to define the boundary using parameters.
+
+::
+
+   x = nest.spatial.distance.x
+   y = nest.spatial.distance.y
+   lhs = x * x / rx**2 + y * y / ry**2
+   mask_param = nest.logic.conditional(lhs <= 1.0, 0.5, 0.0)
```
stinebuu, line 1473: I think you should have a sentence further up that if `x` or `y` falls inside the boundary, we still only connect with a 50% probablility. So that people understand where `0.5` and `0.0` comes in. :)

doc/guides/spatial/guide_spatially_structured_networks.rst
```diff
@@ -1416,61 +1415,124 @@ center neuron are shown, as well as mask and connection probability.
...
+   # Could also have defined it as
+   # mask_param = 0.5*(lhs <= 1.0)
```
stinebuu, line 1475: So you don't need to use `nest.logic.conditional()` at all when sending down to NEST? It is converted? This we should document elsewhere as well. Not in this PR though!

doc/guides/spatial/guide_spatially_structured_networks.rst
hakonsbm: Done!

doc/guides/spatial/guide_spatially_structured_networks.rst
hakonsbm: In cases where the resulting value of `nest.logic.conditional()` is `0` if the condition evaluates to false, it is possible to just use multiplication with the condition instead. That's because the condition also is a parameter, which actually evaluates to `0` or `1`.

I suppose we should have some documentation on parameter comparators somewhere, probably in the detailed transition guide.

PR #1573, opened 2020-04-30
Add min_rate to specify a non-zero rectfying rate for rate neuron models with input noise (rate_neuron_ipn)

niltonlk:
Rate neuron models with input noise has a flag to rectify output to ZERO when it becomes negative. On the other hand, some work use a non-zero rectifying output [Joglekar et al., 2018].
This PR adds a variable min_rate to be used to rectify the output of rate neuron models with input noise (rate_neuron_ipn). 

Joglekar, M. R., Mejias, J. F., Yang, G. R. & Wang, X-J. (2018). Inter-areal Balanced Amplification Enhances Signal Propagation in a Large-Scale Circuit Model of the Primate Cortex. Neuron, 98(1), 222-234.
 
1. Add min_rate entry to nest_names
2. Add min_rate_ variable in rate_neuron_ipn.h
3. Add codes related to setting initial values of min_rate_, setting and getting its values, and issuing errors when negative values is requested, as well as replace the hard coded ZERO in lines 339 and 341 to min_rate_ in rate_neuron_ipn_impl.h

**Additional discussion**
- I used "min_rate" for the variable name, but would "rectify_rate" be a better choice?
- I have not added this feature in rate neuron models with output noise (rate_neuron_opn) as the rectify_output code was not originally implemented in these models. Shoud I add it to rate_neuron_opn?

**Additional correction**
Corrected a misspelling in the warning message for rate models with output noise (lines 117 and 125 in rate_neuron_opn_impl.h; corrected from rate_neuron_ipn to rate_neuron_opn).


niltonlk:
I just realized that I had to add the description of the added variable to the header files of rate models. When doing so, the use of rectify_rate seemed more natural, so I also changed the parameter names from min_rate to rectify_rate. Description of rectify_output was also corrected accordingly

models/rate_neuron_ipn.h
```diff
@@ -178,6 +178,9 @@ class rate_neuron_ipn : public Archiving_Node
     /** Mean input.*/
     double mu_;
 
+    /** Minimum rate.*/
+    double rectify_rate_;
```
jstapmanns, line 182: I think that l. 191 needs to be changed:
`True: If the rate is negative it is set to zero after each time step.` &rarr; `True: If the rate is smaller than rectify_rate it is set to rectify_rate after each time step.`.

niltonlk: Thanks @jstapmanns . You are right! Your suggestion has been applied!

niltonlk:
Sorry, should I do something?

PR #1577, opened 2020-05-04
Silence startup

### LINKED ISSUE #415: CLI: Setting verbosity does not work

### LINKED ISSUE #700: Suppress all NEST startup output

jougs:
This removes the none-silencable output from `ModelManager` and `RandomManager` at simulator startup and upon calls to `ResetKernel` and thereby fixes #700 and fixes #415. They were non-silencable because they were printed before the user-set verbosity is taken into account.

As pointed out by @heplesser below, another solution would have been to degrade the verbosity of the messages in question to something less than the default level. However, this would have led to a resurfacing of the issue if default level were changed.

### DIFF
--- a/nestkernel/model_manager.cpp
+++ b/nestkernel/model_manager.cpp
@@ -423,12 +423,6 @@ ModelManager::connector_requires_clopath_archiving( const synindex syn_id ) cons
 void
 ModelManager::clear_models_( bool called_from_destructor )
 {
-  // no message on destructor call, may come after MPI_Finalize()
-  if ( not called_from_destructor )
-  {
-    LOG( M_INFO, "ModelManager::clear_models_", "Models will be cleared and parameters reset." );
-  }
-
   // We delete all models, which will also delete all nodes. The
   // built-in models will be recovered from the pristine_models_ in
   // init()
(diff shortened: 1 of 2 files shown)


jougs:
@heplesser: thanks for your approval. I've extended the description to also contain a brief discussion of your alternative approach to fixing this.

Do we actually need to wait for a second review here?

heplesser:
@jougs Thanks! I think we can merge this as is.

PR #1598, opened 2020-05-15
Systematize error messages for Connect failures

### LINKED ISSUE #1400: Improve and systematize error messages for connection failures
This issue is a follow-up to #1339/#1399.

Currently, very similar errors in use of `Connect` can trigger different error messages. E.g., connecting a spike detector to a neuron triggers `UnexpectedEvent`, while connecting a neuron to a voltmeter triggers `IllegalConnection` (status with #1399 in place):
```
SLI ] /n /iaf_psc_alpha Create def
SLI ] /v /voltmeter Create def
SLI ] /d /spike_detector Create def

SLI ] n v Connect
Jan 15 15:09:17 Connect_g_g_D_D [Error]: IllegalConnection
    Creation of connection is not possible because:
    The target node or synapse model does not support spike input.
      Note that volt/multimeters must be connected as Connect(meter, neuron).
SLI [4] clear

SLI ] d n Connect
Jan 15 15:09:32 Connect_g_g_D_D [Error]: UnexpectedEvent
    UnexpectedEvent: Source node does not send output. Note that detectors 
    need to be connected as Connect(neuron, detector).
```

This is confusing for the user. We should systematize exception types and information provided by error messages.

stinebuu:
This PR fixes #1400.

I have mostly looked at systematizing `IllegalConnection`. Other error messages might be thrown when connecting as well.

Some examples of how new changes will look:
```
SLI ] /n /iaf_psc_alpha Create def
SLI ] /v /voltmeter Create def
SLI ] /sd /spike_detector Create def
SLI ] 
SLI ] sd n Connect

May 15 09:32:38 Connect_g_g_D_D [Error]: IllegalConnection
    Creation of connection is not possible because:
    Source node does not send output.
      Note that detectors must be connected as Connect(neuron, detector).
SLI [4] clear
SLI ] 
SLI ] v sd Connect

May 15 09:33:00 Connect_g_g_D_D [Error]: IllegalConnection
    Creation of connection is not possible because:
    The target node or synapse model does not support data logging requests.

SLI ] n sd << /rule /all_to_all >> << /synapse_model /stdp_synapse >> Connect

May 15 09:51:41 Connect_g_g_D_D [Error]: IllegalConnection
    Creation of connection is not possible because:
    The target node does not support STDP synapses.
```

Some Python examples:
```
>>> sn = nest.Create('iaf_psc_alpha', positions=nest.spatial.grid(shape=[5,5]))
>>> sd_spatial = nest.Create('spike_detector', positions=nest.spatial.grid(shape=[1,1]))
>>> v = nest.Create('voltmeter')

>>> nest.Connect(sn, sd_spatial,
                 {'rule': 'pairwise_bernoulli', 'p': 0.5, 'use_on_source': False})
NESTErrors.IllegalConnection: ('IllegalConnection in ConnectLayers_g_g_D: Creation of
connection is not possible because:\nSpatial Connect with pairwise_bernoulli to devices are
not possible.', 'IllegalConnection', 'ConnectLayers_g_g_D', ': Creation of connection is not
possible because:\nSpatial Connect with pairwise_bernoulli to devices are not possible.')

>>> nest.Connect(v, sn, {'rule': 'pairwise_bernoulli', 'p': 0.5})
 NESTErrors.IllegalConnection: ('IllegalConnection in Connect_g_g_D_D: Creation of
connection is not possible because:\nEach multimeter can only be connected once to
a given node.', 'IllegalConnection', 'Connect_g_g_D_D', ': Creation of connection is not
possible because:\nEach multimeter can only be connected once to a given node.')
```

Note that Python errors renders strangely. They can be quite hard to read. I think this is a separate issue though.

**One open question I have:**
Right now we do not give any information about which C++ function fail. This is OK for the users, but for the developers hunting bugs it might be annoying. It is possible to just search for the error message though, so I don't know if it really is a problem. But I can add an option to the `IllegalConnection` `Exception` class where you can send in the name of the function, and then add that to the end of the error message. What do you think?

### DIFF
--- a/nestkernel/conn_builder.cpp
+++ b/nestkernel/conn_builder.cpp
@@ -146,7 +146,7 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
     {
       throw BadProperty(
         "In order to use structural plasticity, both a pre and post synaptic "
-        "element must be specified" );
+        "element must be specified." );
     }
 
     use_pre_synaptic_element_ = false;
@@ -239,7 +239,7 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
   {
     throw KernelException(
       "InvalidNodeCollection: "
-      "sources and targets must be valid NodeCollections" );
+      "sources and targets must be valid NodeCollections." );
   }
 }
 
@@ -1386,7 +1386,7 @@ nest::FixedTotalNumberBuilder::FixedTotalNumberBuilder( NodeCollectionPTR source
     {
       throw BadProperty(
         "Total number of connections cannot exceed product "
-        "of source and targer population sizes." );
+        "of source and target population sizes." );
     }
   }
 
@@ -1831,7 +1831,7 @@ nest::SPBuilder::SPBuilder( NodeCollectionPTR sources,
   // Check that both pre and post synaptic element are provided
   if ( not use_pre_synaptic_element_ or not use_post_synaptic_element_ )
   {
-    throw BadProperty( "pre_synaptic_element and/or post_synaptic_elements is missing" );
+    throw BadProperty( "pre_synaptic_element and/or post_synaptic_elements is missing." );
   }
 }
 
@@ -1866,7 +1866,7 @@ nest::SPBuilder::connect_()
 {
   throw NotImplemented(
     "Connection without structural plasticity is not possible for this "
-    "connection builder" );
+    "connection builder." );
 }
 
 /**
@@ -1879,7 +1879,7 @@ nest::SPBuilder::connect_( NodeCollectionPTR sources, NodeCollectionPTR targets 
 {
   throw NotImplemented(
     "Connection without structural plasticity is not possible for this "
-    "connection builder" );
+    "connection builder." );
 }
 
 void
@@ -1889,8 +1889,7 @@ nest::SPBuilder::connect_( const std::vector< index >& sources, const std::vecto
   // make sure that target and source population have the same size
   if ( sources.size() != targets.size() )
   {
-    LOG( M_ERROR, "Connect", "Source and Target population must be of the same size." );
-    throw DimensionMismatch();
+    throw DimensionMismatch( "Source and Target population must be of the same size." );
   }
 
 #pragma omp parallel
--- a/nestkernel/conn_builder.cpp
+++ b/nestkernel/conn_builder.cpp
@@ -145,8 +145,7 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
     if ( syn_spec->known( names::pre_synaptic_element ) or syn_spec->known( names::post_synaptic_element ) )
     {
       throw BadProperty(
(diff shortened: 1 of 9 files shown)


nestkernel/conn_builder.cpp
```diff
@@ -146,7 +146,7 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
     {
       throw BadProperty(
         "In order to use structural plasticity, both a pre and post synaptic "
-        "element must be specified" );
+        "element must be specified." );
```
hakonsbm, line 149: No need to split the string here.

nestkernel/conn_builder.cpp
```diff
@@ -239,7 +239,7 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
   {
     throw KernelException(
       "InvalidNodeCollection: "
-      "sources and targets must be valid NodeCollections" );
+      "sources and targets must be valid NodeCollections." );
```
hakonsbm, line 242: Same as above.

nestkernel/conn_builder.cpp
```diff
@@ -1386,7 +1386,7 @@ nest::FixedTotalNumberBuilder::FixedTotalNumberBuilder( NodeCollectionPTR source
     {
       throw BadProperty(
         "Total number of connections cannot exceed product "
-        "of source and targer population sizes." );
+        "of source and target population sizes." );
```
hakonsbm, line 1389: And here.

nestkernel/conn_builder.cpp
```diff
@@ -1866,7 +1866,7 @@ nest::SPBuilder::connect_()
 {
   throw NotImplemented(
     "Connection without structural plasticity is not possible for this "
-    "connection builder" );
+    "connection builder." );
```
hakonsbm, line 1869: And here.

nestkernel/conn_builder.cpp
```diff
@@ -1879,7 +1879,7 @@ nest::SPBuilder::connect_( NodeCollectionPTR sources, NodeCollectionPTR targets
...
```
hakonsbm, line 1882: Here as well.

stinebuu:
@hakonsbm I also updated other error strings in `conn_builder` to fit with the line-length.

pynest/nest/tests/test_onetooneconnect.py
```diff
@@ -31,11 +31,13 @@
...
 
+    def setUp(self):
+        nest.ResetKernel()
+
     def test_ConnectPrePost(self):
         """Connect pre to post"""
 
-        # Connect([pre], [post])
-        nest.ResetKernel()
+        # Connect(pre, post)
```
jougs, line 40: For my taste, these comments can go away, as they just repeat the information already given in the docstring of the function right above. If you agree, please also remove all following occurrences.

stinebuu: I agree!

stinebuu:
@jougs thanks for your comments, please have another look!

PR #1605, opened 2020-05-18
Fix subscripts in ginzburg_neuron model documentation

### LINKED ISSUE #1603: Mistake in the model directory (ginzburg neuron)
In the model directory for the ginzburg_neuron (under binary neurons: https://nest-simulator.readthedocs.io/en/stable/models/binary.html), it is stated that the function accepts the parameters "c1", "c2" and "c3". The actual function requires the parameters "c_1", "c_2" and "c_3" instead.

sarakonradi:
This PR changes the parameters ``c1``, ``c2`` and ``c3`` to ``c_1``, ``c_2`` and ``c_3``, as required by the actual function. It fixes #1603.

Should ``c1``, ``c2`` and ``c3`` be updated in the description (lines 48-54) too?


heplesser:
> Should `c1`, `c2` and `c3` be updated in the description (lines 48-54) too?

Since we now use math formatting in our documentation, I would suggest to typeset them properly as c-subscript-1 etc, i.e. something like
```
:math:`c_1`
```

sarakonradi:
I have now updated the subscripts in the description too, ensuring the ``:math:`` renders: 

Screenshot from 2020-05-20 14-23-57 (1)

Not sure if we should use ``:math:`` in the parameter table too, since this would make it inconsistent with all other model tables. Perhaps this should be addressed in a different PR.

models/ginzburg_neuron.h
```diff
@@ -45,13 +45,13 @@ point the total synaptic input h into the neuron is summed up,
 passed through a gain function g whose output is interpreted as
 the probability of the neuron to be in the active (1) state.
 
-The gain function g used here is :math:`g(h) = c1*h + c2 * 0.5*(1 +
-\tanh(c3*(h-\theta)))` (output clipped to [0,1]). This allows to
-obtain affin-linear (c1!=0, c2!=0, c3=0) or sigmoidal (c1=0,
-c2=1, c3!=0) shaped gain functions.  The latter choice
+The gain function g used here is :math:`g(h) = c_1*h + c_2 * 0.5*(1 +
+\tanh(c_3*(h-\theta)))` (output clipped to [0,1]). This allows to
+obtain affin-linear (:math:`c_1!=0, c_2!=0, c_3=0`) or sigmoidal (:math:`c_1=0,
```
clinssen, line 50:
```suggestion
obtain affine-linear (:math:`c_1\neq0, c_2\neq0, c_3=0`) or sigmoidal (:math:`c_1=0,
```

models/ginzburg_neuron.h
```diff
@@ -45,13 +45,13 @@ point the total synaptic input h into the neuron is summed up,
...
+c_2=1, c_3!=0`) shaped gain functions. The latter choice
```
clinssen, line 51:
```suggestion
c_2=1, c_3\neq0`) shaped gain functions. The latter choice
```

models/ginzburg_neuron.h
```diff
@@ -45,13 +45,13 @@ point the total synaptic input h into the neuron is summed up,
...
 corresponds to the definition in [1]_, giving the name to this
 neuron model.
-The choice c1=0, c2=1, c3=beta/2 corresponds to the Glauber
+The choice :math:`c_1=0, c_2=1, c_3=beta/2` corresponds to the Glauber
```
clinssen, line 54:
```suggestion
The choice :math:`c_1=0, c_2=1, c_3=\beta/2` corresponds to the Glauber
```

sarakonradi:
@clinssen Thanks for noticing. :)

heplesser:
> Not sure if we should use `:math:` in the parameter table too, since this would make it inconsistent with all other model tables. Perhaps this should be addressed in a different PR.

No, we shouldn't: in the parameter table, we need the parameter names used in code. In a future iteration, we could consider an additional column in the parameter table: one with the mathematical symbols and one with the parameter names in code. 

Another detail to consider for a future revision is to remove the multiplication symbols (`* `)  from equations now that we have pretty math.

sarakonradi:
> No, we shouldn't: in the parameter table, we need the parameter names used in code. In a future iteration, we could consider an additional column in the parameter table: one with the mathematical symbols and one with the parameter names in code.

Alright, then I will leave the table as it is in this PR. 

> Another detail to consider for a future revision is to remove the multiplication symbols (* ) from equations now that we have pretty math.

That is a very good point. 

I have added both enhancements to the model doc to-do list and will discuss them with @jougs.

PR #1613, opened 2020-05-20
Correct path

jd41:
The path given in the previous version wasn't the right one on my system.
(Furthermore, the new path is consistent with the one given in the cd command one line below.)

### DIFF
--- a/doc/documentation_workflow/user_documentation_workflow.rst
+++ b/doc/documentation_workflow/user_documentation_workflow.rst
@@ -77,7 +77,7 @@ Using Sphinx, you can build documentation locally and preview it offline:
 
    make html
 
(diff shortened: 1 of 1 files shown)


doc/documentation_workflow/user_documentation_workflow.rst
```diff
@@ -77,7 +77,7 @@ Using Sphinx, you can build documentation locally and preview it offline:
 
    make html
 
-3. Preview files. They are then located in ``./docs/_build/html``:
+3. Preview files. They are then located in ``./_build/html``:
 
 .. code-block:: bash
```
sarakonradi, line 82: Line 84 should be consistent with 80.

heplesser:
Minimal developer documentation correction, thus merging without second review.

PR #1614, opened 2020-05-20
Fix Python compatibility issue on Readthedocs

### LINKED ISSUE #1609: Building of documentation in Readthedocs does not work anymore
Redthedocs does not seem to get along with the "ignore_patterns" of the "shutil".  There are problems building the documentation. 

**Possible solutions:**

- Adapting the python version for Readthedoc
- possibly use pathlib instead of shutil

steffengraber:
The existing solution with "dirs_exist_ok=True" in `shutil` does not work until Python 3.8, which also led to the bugs in Readthedocs.  This would also force users locally to switch to Python 3.8 to create the documentation.

The new function should theoretically work with Python version 3.4 and higher - tested so far are local Python 3.6 and Python 3.8 and Readthedocs with Python 3.8.

fixes #1609

@terhorstd Maybe you have a better idea about recursive copying


doc/conf.py
```diff
@@ -282,11 +283,17 @@ def setup(app):
...
+
+# copy all *.png with the right subfolder from source to destination
+for f in source.rglob('*.png'):
+    # ff - files with subfolder
+    ff = destination / Path(*f.parts[4:])
+    # fp - only the path, without file name
+    fp = destination / Path(*f.parts[4:]).parent
+    if not fp.exists():
+        Path(fp).mkdir(parents=True, exist_ok=True)
+    copyfile(f, ff)
```
terhorstd, line 299:
```suggestion
def copytreeglob(source, target, glob='*.png'):
    ''' 
    Recursively copy all files selected by `glob` from `source` to `target` path,
    recreating the sub-folder structure.

    Parameters
    ----------
    source : path, str
        source folder where to recursively search for glob

    target : path, str
        target folder where to recreate the tree of source files

    glob : str
        shell-glob specifying which files to copy
    '''
    source = Path(source)
    target = Path(target)
    for relativename in [x.relative_to(source) for x in source.rglob(glob)]:
        # manually create directory, since shutil.copyfile() does not support
        # the `dirs_exist_ok=True` below Python-3.8
        targetpath = target/relativename.parents[0]
        if not targetpath.exists():
            targetpath.mkdir(parents=True)
        assert targetpath.is_dir(), "Targetpath is obstructed by a non-directory object (maybe a file)"
        copyfile(source/relativename, target/relativename)


copytreeglob("../pynest/examples/Potjans_2014", "examples", '*.png')

```
This seems a bit longer, but it's mostly the docstring. The for loop is much more readable like this, imho. What do you think?

readthedocs.yml
```diff
@@ -2,7 +2,7 @@ build:
   image: latest
 
 python:
-  version: 3.6
+  version: 3.8
```
terhorstd, line 5: is this necessary for anything else? Everything in the docs should become 3.6 compatible with this PR.
Maybe we can leave this in to see if RTD honours this.

doc/conf.py
steffengraber: Thank you very much for your comments. I have adopted and tested it.

readthedocs.yml
steffengraber: Works perfect with 3.6.

PR #1617, opened 2020-05-22
Systematize multiline strings for message text in hl_api

### LINKED ISSUE #1178: Use proper multiline strings for message texts in hl_api
As pointed out by @jougs [here](https://github.com/nest/nest-simulator/pull/1165/files/453b19942debd52b3bcad8933951b1e6a01c7f71#discussion_r274302893) in the review of #1165, the Python hl_api contains multiline strings defined in rather cumbersome ways, e.g.
- https://github.com/nest/nest-simulator/blob/453b19942debd52b3bcad8933951b1e6a01c7f71/pynest/nest/lib/hl_api_connections.py#L248
- https://github.com/nest/nest-simulator/blob/453b19942debd52b3bcad8933951b1e6a01c7f71/pynest/nest/lib/hl_api_helper.py#L119

The message texts should be given as proper multiline strings.

stinebuu:
This PR fixes #1178.

Systematize multiline strings for message text in hl_api_*. Have also cleaned up some texts to make them easier to read when looking at the code.


pynest/nest/lib/hl_api_helper.py
```diff
@@ -514,9 +513,8 @@ def model_deprecation_warning(model):
 
     if model in _deprecation_warning:
         if not _deprecation_warning[model]['deprecation_issued']:
-            text = "The {0} model is deprecated and will be removed in a \
-            future version of NEST, use {1} instead.\
-            ".format(model, _deprecation_warning[model]['replacement'])
+            text = """The {0} model is deprecated and will be removed in a future version of NEST, use {1} instead.
+            """.format(model, _deprecation_warning[model]['replacement'])
```
clinssen, line 517: This will include extra whitespace at the end of the string.

pynest/nest/lib/hl_api_info.py
```diff
@@ -89,9 +90,11 @@ def helpdesk():
     """
 
     if 'NEST_DOC_DIR' not in os.environ:
-        print(
-            'NEST help needs to know where NEST is installed.'
-            'Please source nest_vars.sh or define NEST_DOC_DIR manually.')
+        text = textwrap.dedent("""\
+            NEST help needs to know where NEST is installed.
+            Please source nest_vars.sh or define NEST_DOC_DIR manually
+        """)
```
clinssen, line 96: unnecessary use of dedent()

pynest/nest/lib/hl_api_info.py
```diff
@@ -139,19 +142,18 @@ def help(obj=None, pager=None, return_text=False):
...
+            Type 'nest.helpdesk()' to access the online documentation in a browser.
+            Type 'nest.help(object)' to get help on a NEST object or command.
+
+            Type 'nest.Models()' to see a list of available models in NEST.
+            Type 'nest.authors()' for information about the makers of NEST.
+            Type 'nest.sysinfo()' to see details on the system configuration.
+            Type 'nest.version()' for information about the NEST version.
+
+            For more information visit https://www.nest-simulator.org.
+        """)
```
clinssen, line 155: unnecessary use of dedent and temporary variable

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -82,58 +82,44 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, connect_np_arr
                     elif rule == 'fixed_total_number':
                         if ('N' in conn_spec and value.shape[0] != conn_spec['N']):
                             raise kernel.NESTError(
-                                "'" + key + "' has to be an array of "
-                                "dimension " + str(conn_spec['N']) + ", a "
-                                "scalar or a dictionary.")
+                                "'" + key + "' has to be an array of dimension " + str(conn_spec['N']) +
+                                ", a scalar or a dictionary.")
```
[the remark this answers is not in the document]
clinssen: Not a fan of this option, as then the code formatting becomes rather ugly (string indentation does not match code indentation), or you end up making unnecessary calls to textwrap.dedent() (see below).

No opinion on using str.format() vs. string concatenation.

pynest/nest/lib/hl_api_info.py
stinebuu: @clinssen If I do not use `dedent()` the indenting of the printed text is a bit strange. Do you have any tips on how to get the best results for the printed text? The best I have managed is the one below. As you can see, there are empty spaces before the text starts, but at least the two sentences are on the same line. (This is without dedent())

I have the same problem with indenting for the print below as well.

image

clinssen: Everything between the triple quotes will become part of the string, including indentation and newlines, so if we use triple-quoted strings, we should use 
```Python
        text = """\
NEST help needs to know where NEST is installed.
Please source nest_vars.sh or define NEST_DOC_DIR manually"""
```
instead of
```Python
        text = """\
            NEST help needs to know where NEST is installed.
            Please source nest_vars.sh or define NEST_DOC_DIR manually
        """
```
I find this a bit messy and would suggest to stick with single-quoted strings, as the strings in question aren't that long.

stinebuu:
@hakonsbm, @clinssen Thanks for your comments! I have tried to switch to single-quoted strings and use format. Please take another look! There might be things I have overlooked.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -515,10 +514,9 @@ def model_deprecation_warning(model):
 
     if model in _deprecation_warning:
         if not _deprecation_warning[model]['deprecation_issued']:
-            text = "The {0} model is deprecated and will be removed in a \
-            future version of NEST, use {1} instead.\
-            ".format(model, _deprecation_warning[model]['replacement'])
-            text = get_wrapped_text(text)
+            text = \
+                "The {0} model is deprecated and will be removed in a future version of NEST, use {1} instead.".format(
+                    model, _deprecation_warning[model]['replacement'])
```
hakonsbm, line 519: I'm not sure that breaking the line with `\` here is the right way to go, I think the preferred way is to use parentheses.

clinssen: Parentheses can be avoided:
```Python
            text = "The {0} model is deprecated and will be removed in a future version of NEST, " \
                   "use {1} instead.".format(model, _deprecation_warning[model]['replacement'])
```

hakonsbm: We want to avoid the backslash. From [PEP8:](https://www.python.org/dev/peps/pep-0008)

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

And it looks like `.format()` works with implicitly continued strings.

pynest/nest/lib/hl_api_helper.py
```diff
@@ -117,9 +117,8 @@ def show_deprecation_warning(func_name, alt_func_name=None, text=None):
     if func_name in _deprecation_warning:
         if not _deprecation_warning[func_name]['deprecation_issued']:
             if text is None:
-                text = "{0} is deprecated and will be removed in a future \
-                version of NEST.\nPlease use {1} instead!\
-                ".format(func_name, alt_func_name)
+                text = """{0} is deprecated and will be removed in a future version of NEST.
```
clinssen, line 120: Please replace triple-quoted string by single-quoted string(s).

pynest/nest/lib/hl_api_helper.py
clinssen: Personally I prefer the option that avoids brackets, but PEP8 is of course the standard for NEST Simulator, so I am dropping this point! Tabs versus spaces, however... :)

stinebuu:
@hakonsbm @clinssen I have adressed your comments, please have another look.

There are a couple of places I have replaced `'` with `"` *or* `"` with `'`. I know this looks strange, but it is for consistency *within* the function. If I have changed the quote it is because other strings in the function use the other type. Let me know if you want me to change it back. :+1: 

Also, Travis is complaining about a bare except in `hl_api_nodes`, ` pynest/nest/lib/hl_api_nodes.py:120:9: E722 do not use bare except'`. I have not changed this except, but I have changed the file. Changing the `except` is not really part of this PR, but it is making the test fail. Any thoughts of what is best to do here?

clinssen:
Please clean up the bare except in a "trivial" way for now by replacing ``try..except:`` by ``try..except Exception:``.

stinebuu:
@clinssen This was solved in #1653, which is now merged. I have merged master, so it should be all good now :+1:

PR #1625, opened 2020-05-27
Get distances between sources and targets in SynapseCollection

### LINKED ISSUE #1616: Get distances from SynapseCollection of spatially structured network
I would like to have an easy way to get a list of distances between sources and targets from a `SynapseCollection` of a spatially structured network. Right now there is only the function `Distance()` which requires positions or `NodeCollections` as arguments, but does not accept the sources and targets of a SynapseCollection.

```
syncoll = nest.GetConnections()
distances = syncoll.get('distance')
```

An alternative would be the other way round:
```
distances = nest.Distance(syncoll)
```

**Additional context**
Of course, the distances only need to be computed if required.

stinebuu:
Get the distance between source and target in a `SynapseCollection` as a tuple of same length as the SynapseCollection.

The value at index `indx` is the distance between the source, target pair at index `indx` in the SynapseCollection. I can also make it so that we return a list of lists with the distance between *all* sources and targets if that is better.

To get the distance, do
```
conn = nest.GetConnections()
distance = conn.distance
```

If the source and target pairs are not spatially distributed, zeros are returned.

This fixes #1616

### DIFF
--- a/doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
+++ b/doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
@@ -543,7 +543,7 @@ Setting and getting attributes directly
     as the SynapseCollection.
     
     For :ref:`spatially distributed <topo_changes>` sources and targets, you can access the distance between
-    the source, target pair by calling `distance` on your SynapseCollection
+    the source, target pair by calling `distance` on your SynapseCollection.
     
     >>>  synColl.distance
          (0.47140452079103173,
@@ -1299,7 +1299,7 @@ probability and delay, and random weights from a normal distribution:
 Retrieving distance information
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 If you have a SynapseCollection with connections from a spatially distributed network, you can retrieve the
-*distance* between the source and target pairs by calling ``.distance`` on the SynapseCollection
+*distance* between the source-target pairs by calling ``.distance`` on the SynapseCollection.
 
   ::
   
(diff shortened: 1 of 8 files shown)


jasperalbers:
I created a short script testing the new implementation from the PyNEST perspective, using different random positions in 2 or 3 dimensions, and comparing the output of `conn.distance` to  the euclidian distance measured in python (using numpy.linalg.norm()). 
This confirmed that the results are correct for all tested values.

The documentation I found to be clear and understandable, nothing to add from my side.

From a user perspective, I would find it more natural to receive `None` rather than `0.0` as a return value when positions aren't specified. Otherwise, one could mistake the neurons to really coincide in position.

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
```diff
@@ -541,6 +541,21 @@ Setting and getting attributes directly
 
     If you use a list to set the parameter, the list needs to be the same length
     as the SynapseCollection.
+    
+    For :ref:`spatially distributed <topo_changes>` sources and targets, you can access the distance between
+    the source, target pair by calling `distance` on your SynapseCollection
```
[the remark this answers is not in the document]
stinebuu: Done!

stinebuu:
@jhnnsnk @jasperalbers Thanks for your comments! I agree with your comments about returning `0.0` if the nodes are not spatially distributed, and have made the return value be `nan` if either the source *or* the target is not spatially distributed. I chose `nan` instead of `None` as it is easier to return `nan` from C++.

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
```diff
@@ -543,7 +543,7 @@ Setting and getting attributes directly
     as the SynapseCollection.
     
     For :ref:`spatially distributed <topo_changes>` sources and targets, you can access the distance between
-    the source, target pair by calling `distance` on your SynapseCollection
+    the source, target pair by calling `distance` on your SynapseCollection.
```
jasperalbers, line 546:
```suggestion
    the source-target pairs by calling `distance` on your SynapseCollection.
```
Changing consistently from 'source, target pair' to 'source-target pairs' as pointed out by @jhnnsnk.

stinebuu: Thanks @jasperalbers, didn't notice that one :+1:

jhnnsnk (approved):
Everything looks good to me! :+1:

PR #1626, opened 2020-05-27
Fix Izhikevich model initial conditions

clinssen:
Presently, the Izhikevich neuron model does not start at equilibrium: the initial conditions for V and U are such that there is an initial transient, before the system settles to steady-state.

This PR proposes to change the default initial conditions such that the initial state is stable.

### DIFF
--- a/models/izhikevich.cpp
+++ b/models/izhikevich.cpp
@@ -78,8 +78,8 @@ nest::izhikevich::Parameters_::Parameters_()
 }
 
 nest::izhikevich::State_::State_()
-  : v_( -70.0 )       // membrane potential
-  , u_( 0.2 * -70.0 ) // membrane recovery variable (b * V_m_init)
+  : v_( -65.0 )       // membrane potential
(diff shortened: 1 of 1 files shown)


models/izhikevich.cpp
```diff
@@ -78,9 +78,9 @@ nest::izhikevich::Parameters_::Parameters_()
 }
 
 nest::izhikevich::State_::State_()
-  : v_( -65.0 ) // membrane potential
-  , u_( 0.0 )   // membrane recovery variable
-  , I_( 0.0 )   // input current
+  : v_( -70.0 )       // membrane potential
```
gtrensch, line 81: The model implementation in the [original Izhikevich publication](https://www.izhikevich.org/publications/spikes.pdf) uses V_m_init = -65.0, U_m_init = b * V_m_init. Is there particular reason why we don't following it?

terhorstd:
one reviewer is enough for this change.

PR #1632, opened 2020-05-31
Update description for parameter U in Tsodyks synapses

HanjiaJiang:
The parameter U in the Tsodyks synapses (defined in tsodyks_connection.h, tsodyks2_connection.h, tsodyks_connection_hom.h) is the parameter determining the step increase of the release probability u with each spike, according to the original article of the model (Tsodyks, Uziel, Markram, 2000). Currently the relevant description in the files does not match this definition very well, so here it is updated. It is changed from "Maximum probability of release" to "Parameter determining the increase in u with each spike". 
Along with this, the description for u in tsodyks2_connection.h is also changed from "Maximum probability of release" to "The probability of release", which also fit the definition in the original article better.

### DIFF
--- a/models/tsodyks2_connection.h
+++ b/models/tsodyks2_connection.h
@@ -60,7 +60,8 @@ Parameters
 The following parameters can be set in the status dictionary:
 
 ========  ======  ========================================================
- U        real    Parameter determining the increase in u with each spike (U1) [0,1], default=0.5
+ U        real    Parameter determining the increase in u with each spike
+                  (U1) [0,1], default=0.5
  u        real    The probability of release (U_se) [0,1],
                   default=0.5
  x        real    Current scaling factor of the weight, default=U
--- a/models/tsodyks2_connection.h
+++ b/models/tsodyks2_connection.h
@@ -60,8 +60,8 @@ Parameters
 The following parameters can be set in the status dictionary:
 
 ========  ======  ========================================================
- U        real    Maximum probability of release (U1) [0,1], default=0.5
- u        real    Maximum probability of release (U_se) [0,1],
+ U        real    Parameter determining the increase in u with each spike (U1) [0,1], default=0.5
+ u        real    The probability of release (U_se) [0,1],
                   default=0.5
  x        real    Current scaling factor of the weight, default=U
  tau_fac  ms      Time constant for facilitation, default = 0(off)
--- a/models/tsodyks_connection.h
+++ b/models/tsodyks_connection.h
@@ -101,7 +101,7 @@ The following parameters can be set in the status dictionary:
                   releasable pool [0,1]
  y        real    Initial fraction of synaptic vesicles in the synaptic
                   cleft [0,1]
-========  ======  ======================================================
+========  ======  ========================================================
 
 References
 ++++++++++
--- a/models/tsodyks_connection.h
+++ b/models/tsodyks_connection.h
@@ -92,7 +92,7 @@ Parameters
 The following parameters can be set in the status dictionary:
 
 ========  ======  ======================================================
- U        real    Maximum probability of release [0,1]
+ U        real    Parameter determining the increase in u with each spike [0,1]
  tau_psc  ms      Time constant of synaptic current
  tau_fac  ms      Time constant for facilitation
  tau_rec  ms      Time constant for depression
--- a/models/tsodyks_connection.h
+++ b/models/tsodyks_connection.h
@@ -91,8 +91,9 @@ Parameters
 
 The following parameters can be set in the status dictionary:
 
-========  ======  ======================================================
- U        real    Parameter determining the increase in u with each spike [0,1]
+========  ======  ========================================================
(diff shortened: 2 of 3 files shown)


models/tsodyks2_connection.h
```diff
@@ -60,8 +60,8 @@ Parameters
 The following parameters can be set in the status dictionary:
 
 ========  ======  ========================================================
- U        real    Maximum probability of release (U1) [0,1], default=0.5
- u        real    Maximum probability of release (U_se) [0,1],
+ U        real    Parameter determining the increase in u with each spike (U1) [0,1], default=0.5
```
clinssen, line 63: "tsodyks2" seems to correspond to another paper than "tsodyks". The description in the paper for tsodyks2 (Maass & Markram 2002) for U is "the fraction of synaptic resources that are used for a single spike", and u is described as "how much of the fraction of synaptic efficacy is actually used for a spike". Could you confirm?

HanjiaJiang: Hi, thanks for replying. "tsodyks2_connection" is also linked to the first 2 reference articles (Tsodyks, Markram, 1997; Fuhrman, Segev, Markram, Tsodyks, 2002). Combining these 2 articles, the modeling of u (as U_SE in the 1st and 2nd article) and U (as U1 second article) is the same as in (Tsodyks, Uziel, Markram, 2000). In these articles, u stands for the instantaneous release probability (or the fraction of available synaptic efficacy that is actually used), and U is the constant for modifying u with each spike.

Regarding the 3rd article (Maass & Markram 2002), it can be seen that u and U there are also compatible to the above. The description for u "how much of the fraction R_n+1 of the synaptic efficacy is actually used for the (n+1)th spike" is the same as "The probability of release", because the actually released fraction of synaptic efficacy in a spike is u_n*R_n, where R_n is the available fraction, and u_n is the released fraction of R_n. 

The explanation for U "the fraction of synaptic resources that are used for a single spike" means that it determine the value of u in the case of only 1 spike or 1st spike if there are many (which is why it is named U1 in the 2nd article). However, this U also defines the step increase in u, according to the models in all these articles, including that in the 3rd article: "The factor u_n+1... It is increased from u_n to U(1-u_n) + u_n". I think this point is more important for a simulation, hence I made the pull request thereby.  

I hope this is a clear explanation. If there’s further question or suggestion regarding this, please kindly let me know.

sarakonradi, line 63:
```suggestion
 U        real    Parameter determining the increase in u with each spike
                  (U1) [0,1], default=0.5
```

models/tsodyks_connection.h
```diff
@@ -92,7 +92,7 @@ Parameters
 The following parameters can be set in the status dictionary:
 
 ========  ======  ======================================================
```
sarakonradi, line 94:
```suggestion
========  ======  ========================================================
```

models/tsodyks_connection.h
```diff
@@ -92,7 +92,7 @@ Parameters
...
- U        real    Maximum probability of release [0,1]
+ U        real    Parameter determining the increase in u with each spike [0,1]
```
sarakonradi, line 95:
```suggestion
 U        real    Parameter determining the increase in u with each spike
                  [0,1]
```

models/tsodyks_connection_hom.h
```diff
@@ -88,7 +88,7 @@ Parameters
 ++++++++++
 
 ========  ======  ======================================================
- U        real    Maximum probability of release [0,1]
+ U        real    Parameter determining the increase in u with each spike [0,1]
```
sarakonradi, line 91:
```suggestion
 U        real    Parameter determining the increase in u with each spike
                  [0,1]
```

models/tsodyks_connection_hom.h
```diff
@@ -88,7 +88,7 @@ Parameters
 ++++++++++
 
 ========  ======  ======================================================
```
sarakonradi, line 90:
```suggestion
========  ======  ========================================================
```

HanjiaJiang:
> @HanjiaJiang Thank you for this!
> 
> Just added some minor suggestions regarding the table formatting. Could you please also adjust the bottom characters of the `tsodyks_connection.h ` and `tsodyks_connection_hom.h ` tables? It seems that you only need to add two equal signs. Thanks!

Thanks for fixing this, it looks nice. Are there some lines I have to further edit, or I just have to accept your suggestions by clicking on 'Commit suggestion'?

sarakonradi:
> Thanks for fixing this, it looks nice. Are there some lines I have to further edit, or I just have to accept your suggestions by clicking on 'Commit suggestion'?

You can just accept them! It's easiest if you go on "Files changed". This way, you can add each commit to your batch by clicking "Add suggestion to batch" and commit all suggestions at once. 

After commiting the suggestions, you would need to edit line 104 in ``tsodyks_connection.h`` and line 100 in ``tsodyks_connection_hom.h`` yourself and add two equal signs at the end.

PR #1636, opened 2020-06-04
Fix GetConnections to correctly get connections with devices as sources when specifying only target

### LINKED ISSUE #1431: GetConnections function returns empty tuple if only target is specified
According to the PyNEST documentation the accepted parameters are source nodes, target nodes, synapse model and synapse label. All parameters use None as a default and thus are not required.
`GetConnections()`, `GetConnections(source=source_nodes)` and `GetConnections(source=source_nodes, target=target_nodes)` do all correctly return a tuple with the connections. However, if only the post-synaptic nodes are specified (i.e. `GetConnections(target=target_nodes)`) the function returns an empty tuple, even if connections are set (verified with `GetConnections(source=source_nodes, target=target_nodes)` ). I assume that's not intended or I misunderstand the documentation.

Environment:
NEST: 2.20.0
Python: 3.6
OS: Ubuntu 18.04

hakonsbm:
Getting connections with a device as pre-synaptic endpoint, when specifying only targets, was in an if-statement. The if-statement checked if a connector object had been created for that thread and synapse. Therefore connections with devices as pre-synaptic endpoint would be retrieved only if there were other connections, without devices, present.

Getting connections with devices as pre-synaptic endpoints is with this PR moved out of the if-statement. Additionally, the test of `GetConnections()` is made more robust to be able to catch errors like this.

Fixes #1431

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -926,7 +926,7 @@ nest::ConnectionManager::get_connections( std::deque< ConnectionID >& connectome
 
       // Getting regular connections, if they exist.
       ConnectorBase* connections = connections_[ tid ][ syn_id ];
-      if ( connections != NULL )
+      if ( connections != nullptr )
       {
         const size_t num_connections_in_thread = connections->size();
         for ( index lcid = 0; lcid < num_connections_in_thread; ++lcid )
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -924,6 +924,7 @@ nest::ConnectionManager::get_connections( std::deque< ConnectionID >& connectome
       std::vector< index > target_device_node_ids;
       split_to_neuron_device_vectors_( tid, target, target_neuron_node_ids, target_device_node_ids );
 
+      // Getting regular connections, if they exist.
       ConnectorBase* connections = connections_[ tid ][ syn_id ];
       if ( connections != NULL )
       {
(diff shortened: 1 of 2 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -934,23 +935,23 @@ nest::ConnectionManager::get_connections( std::deque< ConnectionID >& connectome
...
-          // devices. First we get connections from devices.
-          target_table_devices_.get_connections_from_devices_(
-            0, *t_node_id, tid, syn_id, synapse_label, conns_in_thread );
-        }
+      // Getting connections from devices.
+      for ( std::vector< index >::const_iterator t_node_id = target_neuron_node_ids.begin();
+            t_node_id != target_neuron_node_ids.end();
+            ++t_node_id )
+      {
+        // target_table_devices_ contains connections both to and from devices
```
stinebuu, line 945: I think you can remove this comment, it confuses me more than adds anything now that the connections to targets are outside the if

nestkernel/connection_manager.cpp
```diff
@@ -924,8 +924,9 @@ nest::ConnectionManager::get_connections( std::deque< ConnectionID >& connectome
       std::vector< index > target_device_node_ids;
       split_to_neuron_device_vectors_( tid, target, target_neuron_node_ids, target_device_node_ids );
 
+      // Getting regular connections, if they exist.
```
stinebuu, line 927: This also includes connections from devices, is that something in addition to "regular connections", or does it all fall under one?

hakonsbm:
> This is slightly off topic, but do we have GetConnections tests for sending in synapse model? In `test_getconnections` it seems like we test weight, sending in `source` or `target` to `GetConnections`, but you could also send in `synapse_model`. I don't think this should be fixed for this PR; but if we do not have it, we should create an issue.

@stinebuu Doesn't look like we have any tests for that. Can you make an issue?

stinebuu:
> > This is slightly off topic, but do we have GetConnections tests for sending in synapse model? In test_getconnections it seems like we test weight, sending in source or target to GetConnections, but you could also send in synapse_model. I don't think this should be fixed for this PR; but if we do not have it, we should create an issue.

> @stinebuu Doesn't look like we have any tests for that. Can you make an issue?

I have created issue #1669.

jougs:
It seems the "Some check haven't completed yet" is caused by a glitch in the communication between Travis and GitHub. I'm thus merging now.

PR #1637, opened 2020-06-05
Fix NodeCollection equality comparison to ensure that the other NodeCollection is of the right type

### LINKED ISSUE #1610: Equality comparison of a primitive and a composite NodeCollection with same length leads to a segfault
**Describe the bug**
Equality comparison between a primitive NodeCollection and a composite NodeCollection leads to a segfault, if both NodeCollections have the same length.

**To Reproduce**
 * Try comparing a primitive NodeCollection with a composite NodeCollection of the same length:
```python
n_prim = nest.Create('iaf_psc_alpha', 10)
n_comp = nest.Create('iaf_psc_exp', 5) + nest.Create('iaf_psc_alpha', 5)

n_prim == n_comp
```
 * This will lead to a segfault.

**Expected behavior**
The comparison result should be `False`.

 - OS: any
 - Python-Version: Python 3.7.6
 - NEST-Version: master (5599ef4b344416ddd06a8197363223a104a3cf8d)
 - Installation: From source

hakonsbm:
Fixes #1610

### DIFF
--- a/nestkernel/node_collection.h
+++ b/nestkernel/node_collection.h
@@ -729,7 +729,7 @@ inline bool NodeCollectionPrimitive::operator==( NodeCollectionPTR rhs ) const
 {
   auto const* const rhs_ptr = dynamic_cast< NodeCollectionPrimitive const* >( rhs.get() );
 
-  return first_ == rhs_ptr->first_ and last_ == rhs_ptr->last_ and model_id_ == rhs_ptr->model_id_
+  return rhs_ptr and first_ == rhs_ptr->first_ and last_ == rhs_ptr->last_ and model_id_ == rhs_ptr->model_id_
     and metadata_ == rhs_ptr->metadata_;
 }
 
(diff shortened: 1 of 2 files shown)


nestkernel/node_collection.h
```diff
@@ -729,7 +729,7 @@ inline bool NodeCollectionPrimitive::operator==( NodeCollectionPTR rhs ) const
 {
   auto const* const rhs_ptr = dynamic_cast< NodeCollectionPrimitive const* >( rhs.get() );
 
-  return first_ == rhs_ptr->first_ and last_ == rhs_ptr->last_ and model_id_ == rhs_ptr->model_id_
+  return rhs_ptr and first_ == rhs_ptr->first_ and last_ == rhs_ptr->last_ and model_id_ == rhs_ptr->model_id_
```
stinebuu, line 732: Could you insert a comment about why we need to return the pointer? Also below (though there it is the `not rhs_ptr`)

hakonsbm:
@stinebuu I've added some comments. Bear in mind that the pointer is used as a boolean here, which evaluates to `true` if it's a valid pointer, and to `false` for null pointers.

nestkernel/node_collection.h
heplesser: @hakonsbm I wonder if it would not be better to be explicit here and write `rhs_ptr != 0 and ...` and correspondingly below. I think it will make the code more readable.

hakonsbm: @heplesser Done.

PR #1646, opened 2020-06-15
Add "Acknowledgements" section to welcome page

sarakonradi:
This PR adds an "Acknowledgements" section to the welcome page. It contains the funding information as well as the EBRAINS and HBP logos.

The topic was discussed offline with @steffengraber and @heplesser. 

If there is a transparent EBRAINS PNG too, then we can replace the current image.


clinssen:
Chould the images be clickable links?

doc/index.rst
```diff
@@ -116,4 +116,26 @@ NEST is available under the :doc:`GNU General Public License 2 or later <license
...
+the Helmholtz Portfolio Theme "Supercomputing and Modeling for the Human Brain", the Excellence Initiative of the
+German federal and state governments, the Priority Program (SPP 2041 "Computational Connectomics") of the Deutsche
+Forschungsgemeinschaft [S.J. van Albada: AL 2041/1-1], the Helmholtz young investigator's group VH-NG-1028 "Theory of
+multi-scale neuronal networks", and compute time provided by UNINETT Sigma2 - the National Infrastructure for High
+Performance Computing and Data Storage in Norway and its predecessors.
+
+.. image:: _static/img/HBP.png
+  :width: 45 %
+.. image:: _static/img/EBRAINS.png
+  :width: 45 %
```
steffengraber, line 141: Looks good. But the ebrains logo has a light grey background, does that have to be?

steffengraber: On GitHub the ebrains Logo is much bigger than the HBP one. 
https://github.com/nest/nest-simulator/blob/d348fe0f42c2ca38e557fda05be83195ecfc7fad/doc/index.rst
Could that still be adapted?

clinssen: Would suggest not to optimise for rendering on GitHub as (per earlier discussions) their rendering engine is terrible. Importantly, it should look good on ReadTheDocs.

sarakonradi: > Would suggest not to optimise for rendering on GitHub as (per earlier discussions) their rendering engine is terrible. Importantly, it should look good on ReadTheDocs.

I agree that it should render well on Read the Docs.

> Looks good. But the ebrains logo has a light grey background, does that have to be?

Ideally not. If we could find an EBRAINS PNG that is transparent, like the HBP one, that would be optimal. This is how the EBRAINS logo appears at the moment:

Screenshot from 2020-06-15 13-37-55

sarakonradi:
> Thanks! We also need to add DEEP-EST and ICEI grants, I will dig out the information during the day.

@heplesser Sure!

sarakonradi:
Just obtained an SVG version of the logo! Here is the updated preview:

Screenshot from 2020-06-15 15-07-58

terhorstd:
> @sarakonradi Thanks! We also need to add DEEP-EST and ICEI grants, I will dig out the information during the day.

They should all be in Acknowledgements.txt and should probably better be linked instead of duplicating this information. Otherwise we will again have the situation where constantly one file is forgotten when adding a new grant.

heplesser:
> They should all be in [Acknowledgements.md](https://github.com/nest/nest-simulator/blob/master/ACKNOWLEDGMENTS.md) and should probably better be linked instead of duplicating this information. Otherwise we will again have the situation where constantly one file is forgotten when adding a new grant.

@terhorstd Very good point!

@sarakonradi @steffengraber Would there be any way to automatically include the text from Acknowledgements.md in index.rst? Does Github only allow .md, or could one turn the .md file into .rst to make inclusion easier?

steffengraber:
Does the file have to be a markdown file (.md)? Or can it be a restructured text file directly? This would eliminate the need to convert.
The copying works then as with Microcircuit model.
If it has to remain Markdown, pandoc could do the converting.
The question can be quite stupid now: Does the recognition of the documentation always correspond to that of the entire software?

sarakonradi:
> Thanks! We also need to add DEEP-EST and ICEI grants, I will dig out the information during the day.

@heplesser Just added the information!

> Would there be any way to automatically include the text from Acknowledgements.md in index.rst? Does Github only allow .md, or could one turn the .md file into .rst to make inclusion easier?

@heplesser @terhorstd You are right about the duplication. @steffengraber and I have been discussing this and are currently looking into solutions. Given the urgency of this PR and that of #1647, should we perhaps discuss the automatic text inclusion in a follow-up issue?

heplesser:
> Does the file have to be a markdown file (.md)? Or can it be a restructured text file directly? This would eliminate the need to convert.

I hope @terhorstd can answer this.

> The question can be quite stupid now: Does the acknoledgements
> of the documentation always correspond to that of the entire software?

I think that we should make life simple for us by having a single file with acknowledgements covering the entire project, simply to have only one file to maintain.

I discovered now that the DEEP-EST and ICEI projects are already in the `ACKNOWLEDGEMENTS.md` file :), so by sticking to one file, all will be easier.

@sarakonradi I would suggest to modify the `Acknowledgement.md` file somewhat:
- Put the newest things first
- Start with something like "NEST development has been supported by many organisations, programs and individuals since 1993. The following list of support received is therefore necessarily incomplete."
- Then "The NEST developers gratefully acknowledge support and funding received from:"
- Then I would have one line for each of the EC Framework programs (from FP6 onward) and under that a bullet point for each particular grant
- Other entries in roughly chronological order, newest first (Norwegian grant  178892/V30 is probably the oldest.

heplesser:
> Given the urgency of this PR and that of #1647, should we perhaps discuss the automatic text inclusion in a follow-up issue?

@sarakonradi I agree, approving now. Will you create the follow-up issue? I see @clinssen is also a reviewer—is there a deeper reason or can we proceed and merge?

sarakonradi:
> @sarakonradi I agree, approving now. Will you create the follow-up issue? I see @clinssen is also a reviewer—is there a deeper reason or can we proceed and merge?

Sure! Just created the follow-up issue. @clinssen voluntarily had a look and left us some comments. If there are no objections, I propose merging this and resolving the rest in a separate PR. :)

PR #1651, opened 2020-06-19
Add special cases for empty NodeCollections

### LINKED ISSUE #1522: Empty NodeCollection constructor
nest3 encourages the use of NodeCollections instead of neuron id lists 

> One thing you should not do is to use your knowledge about neuron ids to set up loops

When adding several neurons from node collections to a node collection it is **not** possible to create an empty node collection:

```python
neurons = nest.NodeCollection()
neurons += nest.create(…)
neurons += nest.create(…)
```

The workaround `nest.NodeCollection([])` causes a crash when attempting to connect. As this is seemingly an invalid call it should throw an error.

I would like the introduction of the no-parameter constructor.

Alternatives is currently only working like this:
```python
neuronslist = []
neuronslist.extend(nest.create(…).tolist())
neuronslist.extend(nest.create(…).tolist())
neurons = NodeCollection(neuronslist)
```

This leads to code, where you manage node collections and node id lists side-by-side with redundancy.

hakonsbm:
Joining two NodeCollections, where one of them is empty, will return a copy of the non-empty NodeCollection. If both are empty, a single empty NodeCollection is returned.

Connecting where one or both NodeCollections are empty will throw a `BadProperty` error. Printing an empty NodeCollection will now print
```
NodeCollection(<empty>)
```

Fixes #1522

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -358,11 +358,11 @@ nest::ConnectionManager::connect( NodeCollectionPTR sources,
 {
   if ( sources->empty() )
   {
-    throw BadProperty( "Presynaptic nodes cannot be an empty NodeCollection" );
+    throw IllegalConnection( "Presynaptic nodes cannot be an empty NodeCollection" );
   }
   if ( targets->empty() )
   {
-    throw BadProperty( "Postsynaptic nodes cannot be an empty NodeCollection" );
+    throw IllegalConnection( "Postsynaptic nodes cannot be an empty NodeCollection" );
   }
 
   conn_spec->clear_access_flags();
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -356,6 +356,15 @@ nest::ConnectionManager::connect( NodeCollectionPTR sources,
   const DictionaryDatum& conn_spec,
   const DictionaryDatum& syn_spec )
 {
+  if ( sources->empty() )
+  {
+    throw BadProperty( "Presynaptic nodes cannot be an empty NodeCollection" );
+  }
+  if ( targets->empty() )
+  {
+    throw BadProperty( "Postsynaptic nodes cannot be an empty NodeCollection" );
+  }
+
   conn_spec->clear_access_flags();
   syn_spec->clear_access_flags();
 
--- a/nestkernel/node_collection.cpp
+++ b/nestkernel/node_collection.cpp
@@ -294,6 +294,10 @@ NodeCollectionPrimitive::to_array() const
 
 NodeCollectionPTR NodeCollectionPrimitive::operator+( NodeCollectionPTR rhs ) const
 {
+  if ( not valid() or not rhs->valid() )
+  {
+    throw KernelException( "InvalidNodeCollection" );
+  }
   if ( rhs->empty() )
   {
     return NodeCollectionPTR( new NodeCollectionPrimitive( *this ) );
@@ -318,10 +322,7 @@ NodeCollectionPTR NodeCollectionPrimitive::operator+( NodeCollectionPTR rhs ) co
   {
     throw BadProperty( "Can only join NodeCollections with same metadata." );
   }
-  if ( not valid() or not rhs->valid() )
-  {
-    throw KernelException( "InvalidNodeCollection" );
-  }
+
   auto const* const rhs_ptr = dynamic_cast< NodeCollectionPrimitive const* >( rhs.get() );
 
   if ( rhs_ptr ) // if rhs is Primitive
--- a/nestkernel/node_collection.cpp
+++ b/nestkernel/node_collection.cpp
@@ -300,7 +300,19 @@ NodeCollectionPTR NodeCollectionPrimitive::operator+( NodeCollectionPTR rhs ) co
   }
   if ( empty() )
   {
-    return NodeCollectionPTR( rhs );
+    auto const* const rhs_ptr = dynamic_cast< NodeCollectionPrimitive const* >( rhs.get() );
+    if ( rhs_ptr )
+    {
+      // rhs is primitive
+      return std::make_shared< NodeCollectionPrimitive >( *rhs_ptr );
+    }
+    else
+    {
+      // rhs is composite
+      auto const* const rhs_ptr = dynamic_cast< NodeCollectionComposite const* >( rhs.get() );
+      assert( rhs_ptr );
+      return std::make_shared< NodeCollectionComposite >( *rhs_ptr );
+    }
   }
   if ( ( get_metadata().get() or rhs->get_metadata().get() ) and not( get_metadata() == rhs->get_metadata() ) )
   {
--- a/nestkernel/node_collection.cpp
+++ b/nestkernel/node_collection.cpp
@@ -294,6 +294,14 @@ NodeCollectionPrimitive::to_array() const
 
 NodeCollectionPTR NodeCollectionPrimitive::operator+( NodeCollectionPTR rhs ) const
 {
+  if ( rhs->empty() )
+  {
+    return NodeCollectionPTR( new NodeCollectionPrimitive( *this ) );
+  }
+  if ( empty() )
+  {
+    return NodeCollectionPTR( rhs );
+  }
   if ( ( get_metadata().get() or rhs->get_metadata().get() ) and not( get_metadata() == rhs->get_metadata() ) )
   {
     throw BadProperty( "Can only join NodeCollections with same metadata." );
@@ -400,11 +408,17 @@ NodeCollectionPrimitive::NodeCollectionPrimitive::slice( size_t start, size_t st
 void
 NodeCollectionPrimitive::print_me( std::ostream& out ) const
 {
-  std::string metadata = metadata_.get() ? metadata_->get_type() : "None";
-
-  out << "NodeCollection("
-      << "metadata=" << metadata << ", ";
-  print_primitive( out );
+  out << "NodeCollection(";
+  if ( empty() )
+  {
+    out << "<empty>";
+  }
+  else
+  {
+    std::string metadata = metadata_.get() ? metadata_->get_type() : "None";
+    out << "metadata=" << metadata << ", ";
+    print_primitive( out );
+  }
   out << ")";
 }
 
@@ -550,6 +564,10 @@ NodeCollectionComposite::NodeCollectionComposite( const NodeCollectionComposite&
 
 NodeCollectionPTR NodeCollectionComposite::operator+( NodeCollectionPTR rhs ) const
 {
+  if ( rhs->empty() )
+  {
+    return NodeCollectionPTR( new NodeCollectionComposite( *this ) );
+  }
   if ( get_metadata().get() and not( get_metadata() == rhs->get_metadata() ) )
   {
     throw BadProperty( "can only join NodeCollections with the same metadata" );
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -371,7 +371,7 @@ def GetStatus(nodes, keys=None, output=''):
         raise TypeError("The first input (nodes) must be NodeCollection or a SynapseCollection with connection handles")
 
     if len(nodes) == 0:
-        return nodes
+        return
 
     if keys is None:
         cmd = 'GetStatus'
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -371,7 +371,7 @@ def GetStatus(nodes, keys=None, output=''):
         raise TypeError("The first input (nodes) must be NodeCollection or a SynapseCollection with connection handles")
 
     if len(nodes) == 0:
-        return
+        return [] if output == 'json' else ()
 
     if keys is None:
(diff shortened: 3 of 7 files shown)


BSVogler:
Thank you for this fix. If I am not mistaken this does not introduce the empty constructor `nest.NodeCollection()` to the python interface. My proposal in #1522 includes this, therefore it fixes only a part. Therefore, approving this would result in a too early closing of the ticket.

Silmathoron:
@BSVogler currently the syntax is a special case of the normal constructor: the empty ``NodeCollection`` is ``nc = nest.NodeCollection([])``.
I think this is necessary, but it is indeed a good question whether we should also support ``nc = nest.NodeCollection()``. This would make sense to me.

@hakonsbm any opinion?

hakonsbm:
@BSVogler @Silmathoron Having an empty `nest.NodeCollection()` constructor is fine by me.

hakonsbm:
@Silmathoron Because the emptiness is a property of the NodeCollection, I created a function for it instead, `empty()`. Also, setting and getting status will now not do anything, and return `None`. 

Example of updated behaviour:
```python
>>> nc = nest.NodeCollection([])
>>> nc.empty()
True

>>> print(nest.GetStatus(nc))
None
>>> print(nc.get('V_m'))
None
>>> print(nc.V_m)
None

>>> nest.SetStatus(nc, {'V_m': -50.})  # no effect, no error
>>> nc.set({'V_m': -50.}))  # no effect, no error
>>> nc.V_m = -50.  # no effect, no error

>>> nc += nest.Create("iaf_psc_alpha", 2) 
>>> nc.empty()
False
```

BSVogler:
I just realized that any public interface for a check, if the collection is empty, is not pythonic. Lists and dicts in python also do not have an empty method. The colelctions support the pyhon-wide available `len()` method. A check if empty should hence be performed with `len(nc)==0`, like the empty method is using internally. The empty() method should thus be removed. The coded uses the empty method internally, and python does not support visibility keywords, so maybe use the double underscore indicator `__empty()`?

Silmathoron:
I'm afraid sometimes focusing on what is pythonic or not may be different to focusing on what is convenient...
However, it is true that, contrary to numpy, we have no strong reasons for not using the normal boolean test.

Should we then define

```python
def __bool__(self):
    return len(self) > 0
```
?

On the question of whether we should remove all trace of ``empty`` at the Python level, though, I have no strong opinion.
Maybe @jougs? I now you tend to have opinions about these things ;)

BSVogler:
One should note that to keep backward compatibility interfaces can hardly be removed later once introduced. IMHO the check for emptiness is not called so often by the user, that this would justify the `empty()` convenience method.

The introduction of the bool method is reasonable following PEP 8.
> For sequences, (strings, lists, tuples), use the fact that empty sequences are false:

jougs:
I tend to agree that the simple `__bool__` function is enough.

pynest/nest/lib/hl_api_types.py
```diff
@@ -331,7 +331,7 @@ def get(self, *params, **kwargs):
         set
         """
 
-        if self.empty():
+        if not self.__bool__():
```
Silmathoron, line 334: wouldn't just ``if not self`` work?
I don't know if there would be any significant cost to it (I don't think so) but it would look much nicer to me...

pynest/nest/lib/hl_api_types.py
```diff
@@ -400,7 +400,7 @@ def set(self, params=None, **kwargs):
             If the specified parameter does not exist for the nodes.
         """
 
-        if self.empty():
+        if not self.__bool__():
```
Silmathoron, line 403: see above

pynest/nest/lib/hl_api_types.py
```diff
@@ -460,9 +460,9 @@ def index(self, node_id):
 
         return index
 
-    def empty(self):
-        """Returns whether the NodeCollection is empty."""
-        return self.__len__() == 0
+    def __bool__(self):
+        """Converts the NodeCollection to a bool. False if it is empty, True otherwise."""
+        return self.__len__() > 0
```
Silmathoron, line 465: same as above, use
```suggestion
        return len(self) > 0
```
not only does it look better, it is actually faster (though I don't know why): 49 ns vs 62 ns on my machine

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -715,8 +715,8 @@ def test_empty_nc(self):
         with self.assertRaises(nest.kernel.NESTErrors.BadProperty):
             nest.Connect(empty_nc, empty_nc)
 
-        self.assertTrue(empty_nc.empty())
-        self.assertFalse(nodes.empty())
+        self.assertFalse(bool(empty_nc))
```
Silmathoron, line 718: would ``self.assertFalse(empty_nc)`` work? (it should)
if so please switch all tests

pynest/nest/lib/hl_api_types.py
hakonsbm: 23.3 µs for `if not self.__bool__()` vs 33.1 µs for `if not self`. So not really a significant performance hit, and it looks cleaner.

pynest/nest/lib/hl_api_types.py
hakonsbm: On my computer there isn't any difference, both around 22.7 µs (my machine must be getting old). But I agree it looks better.

pynest/nest/tests/test_NodeCollection.py
hakonsbm: `assertFalse()` uses `if expr:`, and `assertTrue()` uses `if not expr:`, so both work.

jougs:
@Silmathoron: Can you please have another look and mark the discussions as resolved and approve the PR if you agree that this fixes your concerns? Thanks!

Silmathoron:
Sorry @jougs, unexpected stuff piled up... I'll try to do it on Thursday

Silmathoron:
Hey there, sorry, I know I'm annoying but I think there are still a couple of things to discuss :s

Currently, for an empty ``NodeCollection``, access to any random (and non-existing) member returns ``None``.
I'm not sure what to make of this... it might be fine but it feels a little weird to me... maybe that should be discussed more openly before a decision is made?

The behavior I would expect would be something more like:
* raise ``AttributeError`` for everything undefined
* have ``get`` and ``GetStatus`` return an empty dict (instead of ``None`` right now)
* have ``set`` and ``SetStatus`` raise an error

BSVogler:
> Hey there, sorry, I know I'm annoying

I think it is fair to raise a point if you feel like it.

> have set and SetStatus raise an error

I don't agree. I think it should do nothing. Imagine you write some function where some value of a container is set and it can happen that your input to this function is an empty collection. Then the programmer is forced to catch this case. Compare it with `np.array([])*2`

I don't have an opinion on the get-case, and AttributeError for everything undefined is somewhat unclear to me.

hakonsbm:
>  raise `AttributeError` for everything undefined

@Silmathoron Can you elaborate on what you mean here?

>  have `get` and `GetStatus` return an empty dict (instead of `None` right now)

When getting the status, a dict is not the only type that can be returned. If you do `nc.get('V_m')` on a NodeCollection with multiple nodes, you expect a list of values. Or, if it contains a single node, you expect a single float. Returning `None` in all cases where the NodeCollection is empty signifies that there are no values, no matter what type is expected.

>  have `set` and `SetStatus` raise an error

I agree with @BSVogler and his reasoning on this.

Silmathoron:
@hakonsbm 
Currently getting ``empty.plop`` returns None, I'm wondering whether that's fine...

For ``get`` and ``GetStatus``, you're right, what I mean is that it should return the equivalent empty object depending on what is queried:
* ``empty.get()`` and ``GetStatus(empty)`` would return ``{}``
* ``empty.get('V_m')`` and ``GetStatus(empty, 'V_m')``would return ``[]`` or ``None`` depending on node number

OK for ``set`` and ``SetStatus``.

hakonsbm:
@Silmathoron All attributes of an empty NodeCollection are undefined because it contains no nodes. There is no way to know which attributes are expected to be valid, and which ones should raise an error. Therefore all attributes are treated as valid for empty NodeCollections.

> `empty.get('V_m')` and `GetStatus(empty, 'V_m')`would return `[]` or `None` depending on node number

But we don't know if the user expects the NodeCollection to contain a single node or several. Therefore we can't know if we should return `[]` or `None`. We could perhaps choose to always return `[]` in this case, but I'm not sure if this might cause trouble for the user's scripts. And that would also mean that any attribute, like `empty.plop`, would return `[]`, which I think makes less sense than returning `None`.

Silmathoron:
You're right, sorry, I'm a bit tired, my previous answer did not make any sense...
Anyway, as you pointed out, it seems complicated to have a consistent implementation unless we choose to raise an error, which then has the drawback of making empty NodeCollection a special case.
I would still prefer the latter, shall we discuss this at the next VC?

hakonsbm:
Sure, we can bring it up at the next VC.

hakonsbm:
Decided in the [Open NEST Developer Meeting on 17 August](https://github.com/nest/nest-simulator/wiki/2020-08-17-Open-NEST-Developer-Video-Conference#in-depth-discussion) that an exception should be thrown when using `.get()` on an empty NodeCollection.

pynest/nest/tests/test_NodeCollection.py
```diff
@@ -701,6 +701,55 @@ def test_array_indexing_bools(self):
...
+        self.assertEqual(nest.GetKernelStatus('num_connections'), n*n)
+        self.assertTrue(nodes_a)
+        self.assertIsNotNone(nodes_a.get())
+        nodes_a.V_m = vm
+        self.assertEqual(nodes_a.V_m, n * (vm,))
+
+        nest.ResetKernel()
+
+        nodes_b = nest.Create('iaf_psc_alpha', n)
+        nodes_b += nest.NodeCollection([])
```
Silmathoron, line 745: Could we have one of the tests use ``nest.NodeCollection()`` just to make sure it always works an no one removes it inadvertently?

hakonsbm: Done!

Silmathoron (approved):
Great, thx!

nestkernel/connection_manager.cpp
```diff
@@ -356,6 +356,15 @@ nest::ConnectionManager::connect( NodeCollectionPTR sources,
   const DictionaryDatum& conn_spec,
   const DictionaryDatum& syn_spec )
 {
+  if ( sources->empty() )
+  {
+    throw BadProperty( "Presynaptic nodes cannot be an empty NodeCollection" );
```
heplesser, line 361: I think `BadProperty` is not the right exception to raise. `BadProperty` is meant, for, e.g. negative time constants. I would suggest `IllegalConnection` instead.

nestkernel/connection_manager.cpp
```diff
@@ -356,6 +356,15 @@ nest::ConnectionManager::connect( NodeCollectionPTR sources,
...
+  }
+  if ( targets->empty() )
+  {
+    throw BadProperty( "Postsynaptic nodes cannot be an empty NodeCollection" );
```
heplesser, line 365: See above.

nestkernel/node_collection.cpp
```diff
@@ -294,6 +294,14 @@ NodeCollectionPrimitive::to_array() const
 
 NodeCollectionPTR NodeCollectionPrimitive::operator+( NodeCollectionPTR rhs ) const
 {
+  if ( rhs->empty() )
+  {
+    return NodeCollectionPTR( new NodeCollectionPrimitive( *this ) );
+  }
+  if ( empty() )
+  {
+    return NodeCollectionPTR( rhs );
```
heplesser, line 303: Is this consistent with all other cases? In my understanding, `operator+()` should always return a new object, and does so in all other cases, but here we just get a new pointer to an existing object. This might lead to subtle problems later. I would suggest
```suggestion
    return NodeCollectionPTR( new NodeCollectionPrimitive( *rhs ) );
```

pynest/nest/lib/hl_api_info.py
```diff
@@ -371,7 +371,7 @@ def GetStatus(nodes, keys=None, output=''):
         raise TypeError("The first input (nodes) must be NodeCollection or a SynapseCollection with connection handles")
 
     if len(nodes) == 0:
-        return nodes
+        return
```
heplesser, line 374: `GetStatus()` always returns a tuple (array for JSON output). The consistent return value for an empty node collection would then be an empty tuple for default output and an empty JSON array for JSON output.

heplesser (changes requested):
@hakonsbm I took another look and found some issues, see below.

pynest/nest/lib/hl_api_info.py
```diff
@@ -371,7 +371,7 @@ def GetStatus(nodes, keys=None, output=''):
         raise TypeError("The first input (nodes) must be NodeCollection or a SynapseCollection with connection handles")
 
     if len(nodes) == 0:
-        return
+        return [] if output == 'json' else ()
```
heplesser, line 374:
```suggestion
        return '[]' if output == 'json' else ()
```
JSON output must be a string.

heplesser (approved):

PR #1653, opened 2020-06-23
Update docstring, add comments, and minor refactoring of Create()

hakonsbm:
This PR adds some comments to the `Create()` function, as @jakobj pointed out was needed in #1438. Also a minor refactoring, and an update of the docstring.

### DIFF
--- a/pynest/nest/lib/hl_api_nodes.py
+++ b/pynest/nest/lib/hl_api_nodes.py
@@ -54,11 +54,13 @@ def Create(model, n=1, params=None, positions=None):
         Name of the model to create
     n : int, optional
         Number of nodes to create
-    params : dict, list or Parameter, optional
+    params : dict or list, optional
         Parameters for the new nodes. A single dictionary, a list of
-        dictionaries with size n or a :py:class:`.Parameter`. If omitted, the model's defaults are used.
+        dictionaries with size n, or a dictionary with lists of values with size n.
+        Values may be :py:class:`.Parameter` objects. If omitted,
+        the model's defaults are used.
     positions: :py:class:`.spatial.grid` or :py:class:`.spatial.free` object, optional
-        Object describing spatial posistions of the nodes. If omitted, the nodes have no spatial attatchment.
+        Object describing spatial positions of the nodes. If omitted, the nodes have no spatial attachment.
 
     Returns
     -------
@@ -70,6 +72,8 @@ def Create(model, n=1, params=None, positions=None):
     NESTError
         If setting node parameters fail. However, the nodes will still have
         been created.
+    TypeError
+        If the positions object is of wrong type.
     """
 
     model_deprecation_warning(model)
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_nodes.py
```diff
@@ -115,10 +124,10 @@ def Create(model, n=1, params=None, positions=None):
 
     node_ids = spp()
 
-    if params is not None and params_contains_list:
+    if params is not None and iterable_or_parameter_in_params:
         try:
             SetStatus(node_ids, params)
-        except:
+        except Exception:
```
terhorstd, line 130: don't we have a more specific exception to catch here? Seems overly inclusive to only write a warning when anything and everything could have gone wrong.

hakonsbm: The error is still raised:
https://github.com/nest/nest-simulator/blob/e287301bd5e0c2ff68e7391e00932a114033aca0/pynest/nest/lib/hl_api_nodes.py#L130-L134

So scripts will exit with an error if something goes wrong. But for interactive work I suppose it can be useful to know that the nodes have been created.

Are there any exceptions for which we don't want to issue a warning if an exception is raised here?

pynest/nest/lib/hl_api_nodes.py
```diff
@@ -92,19 +99,21 @@ def Create(model, n=1, params=None, positions=None):
...
 
         return layer
 
-    params_contains_list = True
-    if isinstance(params, dict) and params:
-        params_contains_list = [is_iterable(v) or isinstance(v, Parameter)
-                                for k, v in params.items()]
-        params_contains_list = max(params_contains_list)
+    # If any of the elements in the parameter dictionary is either an array-like object,
+    # or a NEST parameter, we create the nodes first, then set the status. If not,
```
stinebuu, line 110: Maybe `# or a NEST parameter, we create the nodes first, then set the given values. If not,`?

pynest/nest/tests/test_create.py
```diff
@@ -54,25 +53,34 @@ def test_ModelCreateN(self):
...
+        nest_errors = nest.kernel.NESTErrors
+        params = [(tuple(), TypeError),
+                  ({'V_m': [-50]}, IndexError),
+                  ({'V_mm': num_nodes*[-50.]}, KeyError),
+                  ({'V_m': num_nodes*[-50]}, nest_errors.TypeMismatch),
+                  ({'V_m': -50, 'C_m': num_nodes*[20.]}, nest_errors.TypeMismatch),
+                  ]
+
+        for p, err in params:
+            print(p)
```
stinebuu, line 74: Remove print

pynest/nest/lib/hl_api_nodes.py
terhorstd: You are right. It makes sense to raise a warning and pass the error on.
Currently I can't think of a type of exception where this warning would be overly harmful.
:+1:

PR #1654, opened 2020-06-23
Remove unused CMake options

hakonsbm:
Removes the unused CMake options `disable-timing` and `disable-counts`.


terhorstd:
one reviewer should be enough for this.

PR #1662, opened 2020-06-29
Remove MyModule example and pertaining documentation

### LINKED ISSUE #797: Systematic unit testing for external modules (MyModule)
After having put together an external module recently (https://github.com/nest/nest-simulator/pull/544#issuecomment-319948176), i have noticed that unit testing of these is still a little tricky. 

Currently i am (https://github.com/flinz/nest-module-STDPStructplast/blob/master/.travis.yml) building NEST in travis, then executing my python unit tests via cmake (inspired by https://github.com/IGITUGraz/spore-nest-module). This works, but it currently only checks one possible NEST configuration and does neither run vera++ nor cppcheck, since i am not calling build.sh (i dont want to unit test NEST itself).

It would be for example very helpful to provide a skeleton .travis.yml together with MyModule, that could be used for CI according to the current NEST standards, with given matrices of NEST configurations and optional vera and cppcheck flags. Changes in NEST master (or releases) could also propagate a rebuild of external modules to ensure they still work.

### LINKED ISSUE #805: Include MyModule in CI Testing
Currently, `MyModule` is not tested during CI testing. This risks code rot and also means that issues with and surrounding user-defined modules will not be detected and cannot be regression-tested, e.g. #746/#804. 

Assuming that NEST source code is in `src`, NEST is build in `bld` and installed to `ins`, all in the same directory, `MyModule` can be built as follows:
```
mkdir bld_mm
cd bld_mm
cmake --with-nest=../ins/bin/nest-config ../src/examples/MyModule
make install
```

I am not sure yet how to handle `installcheck`.

See also #797, #56.

### LINKED ISSUE #1028: Factor out MyModule as a standalone repository
We promote `MyModule` as a way to extend NEST without requiring the NEST sources around, I propose to go the full length and make it it's own repository under the roof of the NEST Initiative. This would have the following advantages:
* `MyModule` documentation disentangled from NEST documentation
* better discoverability for `MyModule` code and documentation
* natural way of enabling CI for the code
* ability to see who forked the module 

See also #56 (for a request to revise the built setup of `MyModule` and #805 (for a request to enable CI for `MyModule`).

jougs:
During the past few days, I added the [`MyModule`](https://github.com/nest/nest-simulator/tree/master/examples/MyModule) example extension module as its own repository at https://github.com/nest/nest-extension-module. While the creation of the repository primarily fixes #1028, and by adding Travis CI support for the new repository also fixes #805 and fixes #797, this PR now removes the `MyModule` example from the NEST source tree to complete the move.

To make building on the example module easier for users, I have taken the opportunity to modify the directory structure so that all user-servicable code is now in a new `src` directory. A minimal module will be added later as was suggested in https://github.com/nest/nest-simulator/issues/1028#issuecomment-421957450.

As there was no formal code review upon creation of the new repository, I would like to ask the reviewers to also give the new repository, the source code layout and especially [`travis.yml`](https://github.com/nest/nest-extension-module/blob/master/.travis.yml) and [`travis_build.sh`](https://github.com/nest/nest-extension-module/blob/master/travis_build.sh) a quick sanity check.

@clinssen: please get in touch with me once the review of the move is finished, so the NESTML module template can be updated accordingly. Thanks!


clinssen:
@jougs: could you remind me where the module SLI init file is searched for? Adding its path (with and without ``sli`` subdirectory at the end) to the PATH environment variable does not seem to have any effect.

jougs:
@clinssen: Internally (i.e. in the C++ code of `SLIStatrup`) the variable is called `slilibdir. It is defined as `NEST_INSTALL_PREFIX "/" NEST_INSTALL_DATADIR`. If I'm not mistaken, the SLI code from a custom module is also installed there and should thus be found naturally without any special precautions. Adding something to `PATH`
is not expected to help here.

That said, now that you mention this "feature", I am considering to remove the SLI-related code from the module example completely. I can't think of a good reason to still have it in 2020.

clinssen:
This is an issue if the installation directory is not writeable by the current user (e.g. if NEST was installed in ``/usr``).

+1 for removing the requirement entirely.

jougs:
@clinssen: I have created https://github.com/nest/nest-extension-module/issues/4 and think we can consider this as solved as far as this PR is concerned. Further discussion should happen in the linked issue.

jougs:
@lekshmideepu
> You could safely remove the commented out packages from the .travis.yml file. Also you could[ avoid llvm-toolchain-bionic-9](https://travis-ci.org/github/nest/nest-extension-module/jobs/706694705#L170).

Done in c163926.

jougs:
I have created a branch called [initial_review](https://github.com/jougs/nest-extension-module/tree/initial_review) on the extension module repository, where I will carry out all work that follows from the discussions here. Please see that branch for the current state.

jougs:
@lekshmideepu: can you please approve in case you have no more comments? Thanks!
@clinssen: any news?

clinssen:
- Please fix the NEST Simulator link in README.md
- Link to installation instructions (cmake -Dwith-nest=... && make && make install)
- Out-of-source build does not seem to work: 
  ```
  $ mkdir ~/nest-extension-module-build
  $ cd ~/nest-extension-module-build
  $ cmake -Dwith-nest=/home/archels/nest-simulator-build/bin/nest-config ~/nest-extension-module
  $ ls -a ~/nest-extension-module-build
  .  ..
  ```
- Was https://github.com/nest/nest-simulator/pull/1662#issuecomment-659073225 addressed? If NEST was installed to /usr, but I am building MyModule as a normal user, make install will fail due to insufficient permissions, and I cannot compensate by adding the .so path to the LD_LIBRARY_PATH. (Is this issue fully covered by https://github.com/nest/nest-extension-module/issues/4)?

jougs:
>     * Please fix the NEST Simulator link in README.md
Done in 349926327.
 
>     * Link to installation instructions (cmake -Dwith-nest=... && make && make install)
Done in 113c99327.

>     * Out-of-source build does not seem to work:
>       ```
>       $ mkdir ~/nest-extension-module-build
>       $ cd ~/nest-extension-module-build
>       $ cmake -Dwith-nest=/home/archels/nest-simulator-build/bin/nest-config ~/nest-extension-module
>       $ ls -a ~/nest-extension-module-build
>       .  ..
>       ```
Weird. For me it works flawlessly. 

>     * Was [#1662 (comment)](https://github.com/nest/nest-simulator/pull/1662#issuecomment-659073225) addressed? If NEST was installed to /usr, but I am building MyModule as a normal user, make install will fail due to insufficient permissions, and I cannot compensate by adding the .so path to the LD_LIBRARY_PATH. (Is this issue fully covered by [nest/nest-extension-module#4](https://github.com/nest/nest-extension-module/issues/4))?
https://github.com/nest/nest-extension-module/issues/4 only covers the removal of the SLI components from the example module, not touching the installation of the shared object itself. I think it is more or less expected behavior to get an error if you try to install to a location you don't have permissions for. But why isn't setting `LD_LIBRARY_PATH` a solution for those situations? I thought that was the whole point of our recent efforts.

clinssen:
I did a clean checkout and build of everything (jougs/nest-extension-module:initial_review in combination with jougs/nest-simulator:remove_mymodule), and the CMake issue went away. 

As for the installation directory permissions: I ran into this when setting up the VMs for the OCNS tutorial. Initially, I had installed NEST system-wide, in /usr. But this meant that there was no way that the NESTML API (to_nest(), install_nest()) was ever going to work, because it was being run from a Jupyter notebook, without write access to /usr. So I had to reinstall NEST Simulator to ~/.local. Picking up the .so with LD_LIBRARY_PATH was not the problem, but the .sli init file ("mymodule-init.sli") could not be searched for in a non-standard location.

Would it be easy to make the presence of this "module-name-init.sli" file optional in NEST Simulator, and load the module anyway even if the file is missing?

jougs:
> I did a clean checkout and build of everything (jougs/nest-extension-module:initial_review in combination with jougs/nest-simulator:remove_mymodule), and the CMake issue went away.

Good!

> As for the installation directory permissions: I ran into this when setting up the VMs for the OCNS tutorial. Initially, I had installed NEST system-wide, in /usr. But this meant that there was no way that the NESTML API (to_nest(), install_nest()) was ever going to work, because it was being run from a Jupyter notebook, without write access to /usr. So I had to reinstall NEST Simulator to ~/.local. Picking up the .so with LD_LIBRARY_PATH was not the problem, but the .sli init file ("mymodule-init.sli") could not be searched for in a non-standard location.
>
> Would it be easy to make the presence of this "module-name-init.sli" file optional in NEST Simulator, and load the module anyway even if the file is missing?

Thanks for the clarification. Here are some remarks:

* There was never a good reason to have the SLI files in the modules created by NESTML (the only reason is the laziness of the original authors when it came to removing them from the MyModule template). You wouldn't see any difference if they were gone. Because of this, my plan is to remove the "module-name-init.sli" and pertaining code altogether from NEST/MyModule. This is what https://github.com/nest/nest-extension-module/issues/4 is meant to express, but maybe the description over there is not all that clear.
* If the current Python pipeline of NESTML (i.e. `to_nest()`, `install_nest()`) does not support installation to a non-standard directory, that feature should be added. I think it is a more and more common use-case to work against a global installation of NEST.
* I thought you can set `LD_LIBRARY_PATH` from within a Jupyter Notebook. If I'm mistaken, we might have to look into library search paths for the dynamic loader again once I'm done with my refactoring.

clinssen:
> If the current Python pipeline of NESTML (i.e. to_nest(), install_nest()) does not support installation to a non-standard directory, that feature should be added.

The situation is the reverse: these API functions cannot work if NEST Simulator is installed to /usr, because you need sudo rights to copy the .sli file there.

We can override in case of a non-standard mymodule.so location with LD_LIBRARY_PATH (which works), but we cannot override a non-standard mymodule-init.sli location because there is no analogous environment variable.

Having nest/nest-extension-module#4 implemented would be really great; could this be as simple as removing the following block of code?
```C++
const std::string
mynest::MyModule::commandstring() const
{
  // Instruct the interpreter to load mymodule-init.sli
  return std::string( "(mymodule-init) run" );
}
```

jougs:
> > If the current Python pipeline of NESTML (i.e. to_nest(), install_nest()) does not support installation to a non-standard directory, that feature should be added.
> 
> The situation is the reverse: these API functions cannot work if NEST Simulator is installed to /usr, because you need sudo rights to copy the .sli file there.
> 
> We can override in case of a non-standard mymodule.so location with LD_LIBRARY_PATH (which works), but we cannot override a non-standard mymodule-init.sli location because there is no analogous environment variable.

OK, Then I think I now understood your earlier explanations.

> Having [nest/nest-extension-module#4](https://github.com/nest/nest-extension-module/issues/4) implemented would be really great; could this be as simple as removing the following block of code?
> 
> ```c++
> const std::string
> mynest::MyModule::commandstring() const
> {
>   // Instruct the interpreter to load mymodule-init.sli
>   return std::string( "(mymodule-init) run" );
> }
> ```

There's also some code in `CMakeLists.txt` that needs adapting, but in principle it's that line that caused you trouble when trying to load the module that failed to install its `.sli` file under `/usr`.

The solution I am currently planning to implement is reaching further and will make NEST extension modules distinct from SLI modules by deriving them for their own base class instead of from `SLIModule` and convert the `DynamicLoaderModule` (which is essentially a SLI module able to load other SLI Modules at runtime) into an `ExtensionManager` purely within NEST that has nothing to do with SLI.

jougs:
By the way: the [minimal module example](https://github.com/nest/nest-extension-module/issues/3) is actually already free from such SLI stuff. Can you please try if that would be a better basis for NESTML modules? Maybe you could even build on the `initialize.py` script in the linked branch to get the NESTML module set up.

PR #1671, opened 2020-07-02
Fix wrong connection in STDP NN unit test

clinssen:


hakonsbm:
Merging with only one review as it is a minimal change.

PR #1679, opened 2020-07-06
Add guide about weight normalization

### LINKED ISSUE #1663: Discuss weight normalisation in documentation
Questions about how to normalise the weights vector in plastic networks come up regularly on the mailing list. The issue is: say you have a neuron with incoming synaptic weights w1, w2, ... wn. A plasticity rule might require that the vector norm |w| remains constant.

As the solution of how to do this in NEST requires some explanation, @diesmann suggested to add a page in the documentation describing how to accomplish this. Please feel free to assign me to doing this.

clinssen:
Closes #1663.

I would recommend @jhnnsnk and/or @YounesBouhadjar as reviewers.


YounesBouhadjar:
looks good to me

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,19 @@
+Weight normalization
+====================
+
+Let's say that you have a neuron with incoming synaptic weights :math:`w_1, w_2, \ldots, w_n`. A plasticity rule might require that the vector norm :math:`|w|` remains constant.
```
jhnnsnk, line 4: It would be good to define more precisely what is to be kept constant. From this definition, I'd assume that the Euclidean norm was meant, but the example below sets the sum of the weights to 1.
Could you add a reference of a concrete plasticity rule that makes use of weight normalization?

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,19 @@
...
+
+Because of the way that the data structures are arranged in NEST, normalizing the weights per neuron is a costly operation (in terms of time spent). One has to iterate over all the neurons, then for each neuron fetch all of its connections, calculate the vector norm and perform the actual normalisation, and finally to write back the new weights.
+
+This would look something like:
+
+.. code-block:: python
+
+   for neur in neurons_to_be_normalised:
+       connect = nest.GetConnections(target=[neur], synapse_model="stdp_synapse")
+       w = nest.GetStatus(connect, "weight")
```
jhnnsnk, line 14: Wouldn't it be more NEST3 conform to use `.get()` und `.set()`?

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,19 @@
+Weight normalization
+====================
+
+Let's say that you have a neuron with incoming synaptic weights :math:`w_1, w_2, \ldots, w_n`. A plasticity rule might require that the vector norm :math:`|w|` remains constant.
+
+Because of the way that the data structures are arranged in NEST, normalizing the weights per neuron is a costly operation (in terms of time spent). One has to iterate over all the neurons, then for each neuron fetch all of its connections, calculate the vector norm and perform the actual normalisation, and finally to write back the new weights.
```
jhnnsnk, line 6: "fetch all of its connections" -> "fetch all of its incoming connections"

doc/guides/weight_normalization.rst
YounesBouhadjar: This model [1] (known as the SORN model) makes use of weight normalization, see equation 6:
1. Lazar, A. et al. (2009). SORN: a Self-organizing Recurrent Neural Network. Frontiers in Computational Neuroscience, 3. https://doi.org/10.3389/neuro.10.023.2009 
2. **spiking implementation of 1**: Klos, C. et al. Bridging structure and function: A model of sequence learning and prediction in primary visual cortex. PLoS Computational Biology, 14(6). https://doi.org/10.1371/journal.pcbi.1006187

I agree that we should be more precise in defining the weight normalization procedure, also maybe it's better to use a variable for the total sum, say **w_total** and not define it equal to **1** as it's done in the little script below. See paper [2], equation 9.

doc/guides/weight_normalization.rst
terhorstd: isn't there even a `connect.weight` property?

clinssen: According to the transition guide, https://nest-simulator.readthedocs.io/en/latest/guides/nest2_to_nest3/nest2_to_nest3_detailed_transition_guide.html, nest.GetConnections() is still the suggested approach.

clinssen:
@jhnnsnk @YounesBouhadjar: I have incorporated all your comments, thanks very much! Could you please have another look?

YounesBouhadjar:
looks good

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,51 @@
...
+
+To be formally correct, this normalization should be done at each simulation timestep, but weights typically evolve on a much longer timescale than the timestep that the network is simulated at, so this would be very inefficient. Depending on how fast your weights change, you may want to perform this update, say, every 100 ms of simulated time, or every 1 s (or even less frequently).
+
+To summarize, the basic strategy is to divide the total simulated time into chunks of, say, 100 ms. You simulate for 100 ms, then you pause the simulation and update the weights (using the code above), and then continue simulating the next chunk.
+
+
+References
+----------
+
+.. [1] Lazar, A. et al. (2009). SORN: a Self-organizing Recurrent Neural Network. Frontiers in Computational Neuroscience, 3. https://doi.org/10.3389/neuro.10.023.2009
```
jhnnsnk, line 48: Here and in the other reference, I suggest to make the DOI clickable as in this example (disregarding the automatic rendering):

DOI: `10.1093/cercor/bhs358 <https://doi.org/10.1093/cercor/bhs358>`__.

clinssen: Done!

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,51 @@
+Weight normalization
+====================
+
+Preliminaries
+-------------
+
+Suppose that the incoming synaptic weights of a neuron are given as :math:`w_1, w_2, \ldots, w_n`. A plasticity rule might require that the vector norm :math:`|w|` remains constant. For example, the 1-norm is used in [1]_ [2]_:
```
jhnnsnk, line 7: Please use upper or lower case `W` consistently.

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,51 @@
...
+
+   |W|_1 = \sum_i W_i
+
+Keeping this norm constant is typically done as an extra step after the main weights plasticity step (e.g. after an STDP weight update). First, the norm is computed, and second, all weights :math:`w_1, \ldots, w_n` are updated according to:
+
+.. math::
+
+   W_i &\leftarrow W_{total} \frac{W_i}{|W|_1}
+
+where :math:`w_{total}` is the target total input weight [2]_.
```
jhnnsnk, line 19: I think it would be easier to follow if `w_{total}` was defined above, for instance: 
"... remains constant, e.g., at the target value :math:`w_{total}`."

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,51 @@
...
+Implementation in NEST
+----------------------
+
+Because of the way that the data structures are arranged in NEST, normalizing the weights is a costly operation (in terms of time spent). One has to iterate over all the neurons, then for each neuron fetch all of its incoming connections, calculate the vector norm and perform the actual normalization, and finally to write back the new weights.
+
+This would look something like:
+
+.. code-block:: python
+
+   def normalize_weights(neuron_gids_to_be_normalized):
```
jhnnsnk, line 31: One option here would be to say right from the start that `w_{total}` is assumed to be one or, which I would prefer, to pass `w_total` as an argument to the function such that there is correspondence in code and equation above.

doc/guides/weight_normalization.rst
```diff
@@ -0,0 +1,51 @@
...
+           conn = nest.GetConnections(target=[neur], synapse_model="stdp_synapse")
+           w = np.array(conn.weight)
+           w_norm = w / sum(w)
+           conn.weight = w_norm
+
+To apply normalization only to a certain synapse type, restrict ``GetConnections()`` to that type. For example: ``GetConnections(..., synapse_model="stdp_synapse")``.
+
+To be formally correct, this normalization should be done at each simulation timestep, but weights typically evolve on a much longer timescale than the timestep that the network is simulated at, so this would be very inefficient. Depending on how fast your weights change, you may want to perform this update, say, every 100 ms of simulated time, or every 1 s (or even less frequently).
+
+To summarize, the basic strategy is to divide the total simulated time into chunks of, say, 100 ms. You simulate for 100 ms, then you pause the simulation and update the weights (using the code above), and then continue simulating the next chunk.
```
jhnnsnk, line 42: "update the weights" -> "normalize the weights"
since you are separating the STDP weight updates explicitly from the normalization step.

Also above: "perform this update" -> "perform a weight normalization"

jhnnsnk (changes requested):

clinssen:
Thanks for the great feedback! I have rewritten parts of the document, that should address all of your comments. Could you double-check?

I went with w_{target} rather than w_{total}, to allow for other norms than the L1-norm.

@YounesBouhadjar: could you give a formal green checkmark by clicking the "Review" button on the "Files changed" tab of this PR? Cheers!

jhnnsnk (approved):
Thanks! :+1:

YounesBouhadjar (approved):

PR #1686, opened 2020-07-07
Make testsuite harness independent of variable replacement

### LINKED ISSUE #345: do_tests should get all arguments from the commandline

### LINKED ISSUE #1181: nest.init fails when going to  fallback test harness
As shown [here](https://travis-ci.org/lekshmideepu/nest-simulator/jobs/521221141#L6865),  nest.init(["--verbosity=WARNING", "--quiet"]) fails with AttributeError
```
 Nosetests unavailable. Using fallback test harness.
  Fewer tests will be excuted.
  File "/Users/travis/build/lekshmideepu/nest-simulator/result/share/nest/extras/do_tests.py", line 50, in <module>
    nest.init(["--verbosity=WARNING", "--quiet"])
AttributeError: 'module' object has no attribute 'init'
Assertion failed: (kernel_manager_instance_), function get_kernel_manager, file /Users/travis/build/lekshmideepu/nest-simulator/nestkernel/kernel_manager.h, line 206.
/Users/travis/build/lekshmideepu/nest-simulator/result/share/nest/extras/do_tests.sh: line 794: 21235 Abort trap: 6           "${PYTHON}" "${PYTHON_HARNESS}" >> "${TEST_LOGFILE}"
  PyNEST tests: 
     Passed: 
     Skipped: 
     Failed: 
/Users/travis/build/lekshmideepu/nest-simulator/result/share/nest/extras/do_tests.sh: line 784: 297 +  : syntax error: operand expected (error token is " ")
```

### LINKED ISSUE #1652: Add junitparser to installation requirements
"Since the transition to the new test summary generator, the testsuite requires the Python ``junitparser`` package to be available." (@heplesser)

This should be added to https://github.com/nest/nest-simulator/blob/master/doc/installation/linux_install.rst, but perhaps also to other places, e.g. the requirements of the .deb packages?

What should happen in case junitparser cannot be found: fail tests, or only issue a warning?

Via https://github.com/nest/nest-simulator/pull/1578#issuecomment-647108209

jougs:
This PR mainly fixes #345 by making the testsuite harness take all required arguments from the command line instead of replacing them into the script from within the build system. As a consequence, the main runner was renamed from `do_tests.sh.in` to `do_tests.sh`. By properly checking for `nosetests` and `junitparser` before any tests are run, this also fixes #1652.

Despite my strongest desire to avoid feature creep, I could not fully resist to also address some other things on on the way. In particular, this PR now also
* removes the check for ancient versions of the NEST config file `.nestrc` ([do_tests.sh:114](https://github.com/nest/nest-simulator/pull/1686/files#diff-0fbffe2b216f7758cec555cdb0e80815L114)-[134](https://github.com/nest/nest-simulator/pull/1686/files#diff-0fbffe2b216f7758cec555cdb0e80815L134))
* adds an informational header that is printed before the testsuite starts running
* removes the custom test harness for PyNEST (fixes #1181)
* removes the workaround for Python 2 from regressiontests `issue-779-1016.py`.


jougs:
@clinssen, @gtrensch: friendly ping!

jougs:
The copyright header check will be unbroken, once #1706 is in master and merged here.

testsuite/do_tests.sh
```diff
@@ -76,78 +88,87 @@ while test $# -gt 0 ; do
...
+fi
+mkdir "${REPORTDIR}"
 
-NEST_BINARY=nest_serial
+TEST_BASEDIR="${PREFIX}/share/doc/nest"
+TEST_LOGFILE="${REPORTDIR}/installcheck.log"
+TEST_OUTFILE="${REPORTDIR}/output.log"
+TEST_RETFILE="${REPORTDIR}/output.ret"
+TEST_RUNFILE="${REPORTDIR}/runtest.sh"
+TEST_TMPDIR=$(mktemp -d "${REPORTDIR}/nest.XXXXX")
```
heplesser, line 128: Since we always set up a clean `REPORTDIR`,  is there any need to have a randomly numbered `TMPDIR`?

jougs: Good catch! Actually it seems like we don't even use `TEST_TMPDIR` anywhere. I've thus removed the line in question in 76cf66d4c.

Issue #1685, opened 2020-07-07
hpc_benchmark.py fails with more than 30 mpi processes


jarsi:
**Describe the bug**
For benchmarking purposes I heavily rely on the hpc_benchmark.sli model. I figured I could maybe give the python version nest-simulator/pynest/examples/hpc_benchmark.py a try and maybe it makes sense to switch to it in the long run. While setting up the benchmarking framework today I ran into problems when using more than 30 MPI processes, the jobs won't finish and are stuck.

The problem seems to lie in the [logger class](https://github.com/nest/nest-simulator/blob/8eb3bdb936ec89ceb01608f8b0eb26b73ce24b04/pynest/examples/hpc_benchmark.py#L414)
There is a variable defined which determines the maximal number of mpi processes to be logged:
```
self.max_rank_log = 30
```

Ranks above 30 should not be logged. While this is a sensible choice, it does not work as expected. For all ranks above 30 the following error message
```
AttributeError: 'NoneType' object has no attribute 'log'
```
is thrown and these ranks remain stuck.

**To Reproduce**
1. Run the script with more than 30 mpi processes

**Expected behavior**
Ranks > 30 should not be logged, simulation should move on normally.

**Desktop/Environment (please complete the following information):**
 - module load GCC CMake ParaStationMPI Python SciPy-Stack GSL jemalloc
- Nest at: aab21a8c4665c8643851eed05471f23307c2231c

PR #1688, opened 2020-07-08
Update hpc_benchmark.py

jarsi:
This pull request addresses issue  #1685. If self is returned in all cases (also for ranks >30) all ranks can access the class logger, while only ranks < 30 actually write output. Thus the AttributeError is not thrown anymore. Has been tested on JURECA with upto 128 MPI processes.

### DIFF
--- a/pynest/examples/hpc_benchmark.py
+++ b/pynest/examples/hpc_benchmark.py
@@ -436,7 +436,7 @@ class Logger(object):
 
             self.f = open(fn, 'w')
 
-            return self
+        return self
 
     def log(self, value):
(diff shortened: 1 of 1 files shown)


terhorstd:
Hey, thanks for the fix! That's a good catch!
I'm not sure how the MacOS build on travis now started to fail, maybe @lekshmideepu can look into that. Seems like it finds an old 2.7 somehow.

terhorstd:
The failing Travis build is a bug documented in #1689

stinebuu:
Merging this as it is a simple one-liner.

PR #1699, opened 2020-07-21
Setting a sensible Alltoall count for spike communication

H27CK:
The default behaviour of NEST causes expansion of spike buffers while the number of communication passes during a step is greater than one. This corresponds to an increase in the count argument of `MPI_Alltoall` for spike communication. Buffers are grown by a factor of 1.5, and there is no mechanism for reduction. During relatively high activity, spike buffers can expand to a size that is excessively large. This will result in the transmission of some amount of 'padding' in the spike buffer, which is wasteful when the number of (valid) spikes to be transmitted is much less than what the capacity of spike buffers allows.

This pull requests, which has been designed to maximise use of preexisting code and conventions, introduces a mechanism that communicates, to all ranks, the maximum number of spikes to be transmitted between any two ranks every 1000 steps (arbitrarily chosen to not call `MPI_Allreduce` too often, but to also update counts somewhat regularly). In the case of a well balanced network and homogeneous communication, this maximum provides a reasonable count, which is the argument dictating the number of elements, i.e. spikes, to send to each MPI process with `MPI_Alltoall`. Note, there is no reduction in the size of spike buffers when the maximum number of spikes is less than previously seen, i.e. we do not allow a call to `vector::resize`; only the count parameter of `MPI_Alltoall` is modified to respect this reduction. Although we can reduce the memory footprint by some small amount, it is no worse than what is currently seen, with the benefit that we do not incur a cost due to reallocation.

Cases in which this has been observed as costly:
1. microcircuit model with `original` set for `V0_type`, resulting in an initial transience that causes spike buffers to grow to a size that is far greater than what is required. This is not seen if using the `optimized` condition, in which case this PR offers only slight reductions in simulation time, since spiking activity remains somewhat constant during the simulation.
2. the multi-area model, due to a high variability in spiking activity, benefits from these changes. Tests on JUSUF have shown a ~30% reduction in simulation time for 30 nodes, 4 MPI tasks per node, 32 OpenMP threads per MPI task. Gains in performance are expected to be greater with more nodes, i.e. for larger networks, since the impact of inter-node communication is greater.


jasperalbers (commented):
Unfortunately I am unable to reproduce the performance improvements for MAM simulations. In my tests, the total simulation time and the communication time are both very similar to the current master, if not a bit slower. Maybe I am missing a crucial setting? Here are the details of my test:

- machine: JURECA
- configuration: 6 MPI processes per node, 4 OpenMP threads per node
- amount of nodes: 20, 60, 100, 140
- using current master + this PR + timers: [link to my branch](https://github.com/jasperalbers/nest-simulator/tree/good-count_%231699)
- modules used during build: `GCC CMake ParaStationMPI Python SciPy-Stack GSL Boost/1.69.0-Python-3.6.8`
- compiler flags: `-DCMAKE_INSTALL_PREFIX:PATH=$instdir -DTIMER=1 -Dwith-python=ON -Dwith-boost=ON -Dwith-ltdl=OFF -Dwith-mpi=ON -DCMAKE_CXX_COMPILER=mpic++ -DCMAKE_C_COMPILER=mpicc`
- modules used in simulation: same + `jemalloc`

H27CK:
@jasperalbers I'll take your branch and run some tests myself to investigate. Will get back to you soon.

H27CK:
@jasperalbers I've confirmed your findings and found the issue. Let me sort it out before you continue.

H27CK:
I've opted to go for a less fragile strategy to reduce MPI cost. I'll submit that as another PR.

PR #1690, opened 2020-07-09
Bump Python version for macOS build

jougs:
This fixes #1689. Being an emergency fix for build failures on macOS in our Travis setup, it should be reviewed and merged as soon as possible.

PR #1623, opened 2020-05-26
Autodetect Python include path and library file instead of hardcoding them 

lekshmideepu:
This PR replaces the explicit setting of Python library and include paths in the Travis build script by a semi-smart autodetection (fixes #1600, fixes #1701).

In addition, this also adds caching for the macOS build, which speeds up build times by 10 minutes.

PR #1700, opened 2020-07-21
Update to latest Python on MacOS Travis CI build

jougs:
After my emergency fix in #1690, Travis again changed their Python version. This is another emergency PR that should work for the time being and until  #1623 provides a more stable permanent solution.

Please review and merge quickly!


lekshmideepu:
Since this is an emergency fix, I am going to merge it without waiting for the second reviewer.

lekshmideepu:
#1701

PR #1704, opened 2020-07-30
Restructure build stages to use names instead of individual flags

jougs:
This is a follow-up to #1623 and restructures the build stages in `.travis.yml` to use a descriptive build type variable (`xNEST_BUILD_TYPE`) and leave the meaning of the different values to `extras/travis_build.sh`. It also re-adds a Python 3.6 job and bumps the minimal required Python version in the documentation to 3.6.


extras/travis_build.sh
```diff
@@ -24,114 +24,28 @@
...
     export CXX=$(ls /usr/local/bin/g++-* | grep '^/usr/local/bin/g++-\d$')
 fi
 
-NEST_VPATH=build
-NEST_RESULT=result
-if [ "$(uname -s)" = 'Linux' ]; then
-    NEST_RESULT=$(readlink -f $NEST_RESULT)
-else
-    NEST_RESULT=$(greadlink -f $NEST_RESULT)
+if [ "xNEST_BUILD_COMPILER" = "CLANG" ]; then
```
clinssen, line 40: $ missing

jougs, line 40:
```suggestion
if [ "$xNEST_BUILD_COMPILER" = "CLANG" ]; then
```

jougs:
> The libneurosim build is broken due to expecting a Python 2 interpreter.

Apparently it is even worse :disappointed:

It seems that [previously](https://travis-ci.org/github/nest/nest-simulator/jobs/713516808#L1522) the stage using Clang was only running a minimal configuration. During the work on this PR, I (accidentally) made that into a full build, which now fails to install dependencies like MUSIC, CSA, libneurosim, SIONlib due to missing MPI and OpenMP or because of a different way of setting the Python version.

I think sorting this out and making the Clang stage a full build is beyond the scope of this PR and would rather just try to get the old behavior back with the new flag mechanism for the stages. Would you be fine with this strategy? If you wish, I can also open an issue about the desired behavior.

jougs:
It seems that the code is [fine in principle](https://travis-ci.org/github/nest/nest-simulator/jobs/716547421#L6067), but the macOS job crashes with `The job exceeded the maximum time limit for jobs, and has been terminated`. I suggest to ignore this for now and review independently.

.travis.yml
```diff
@@ -135,22 +129,19 @@ before_install:
...
          cp extras/install_libboost.sh $HOME
          cd $HOME/build
          echo $PATH
-
-         if [ MATRIX_EVAL="CC=clang-7 && CXX=clang++-7" ]; then
-           pyenv install --list
-           pyenv global 3.6.10
+         if [ "$xNEST_BUILD_COMPILER" = "CLANG" ]; then
+             pyenv install --list
+             pyenv global 3.6.10
```
clinssen, line 144: Hard coded version number: potentially an issue?

jougs: Agreed. I tried removing the lines altogether in ec26ee940. Let's see what Travis thinks of this.

jougs: OK. Apparently it [picks up Python 2.7](https://travis-ci.org/github/nest/nest-simulator/jobs/726840827#L1289) without these lines. I have therefore removed the commit again in my last push.

One alternative could be to define the exact version number as an environment variable in [line 60](https://github.com/nest/nest-simulator/pull/1704/files/a6026390f0dd42d7d14e69545201d2087a890085#diff-354f30a63fb0907d4ad57269548329e3R60), but I'm afraid we would then still have the setting twice, once for our purposes in line 60 and once to tell Travis what we want in [line 59](https://github.com/nest/nest-simulator/pull/1704/files/a6026390f0dd42d7d14e69545201d2087a890085#diff-354f30a63fb0907d4ad57269548329e3R59). The advantage would just be code locality. What do you think?

clinssen: Could this trick be used?

```bash
$ python3 -c "import sys; print('%d.%d.%d' % (sys.version_info.major, sys.version_info.minor, sys.version_info.micro))"
3.7.3
```

jougs: Maybe. In 32365933d, I'm trying out a version based on the somewhat shorter `python3 --version | cut -d' ' -f2`.

jougs: OK. After that also [failed](https://travis-ci.org/github/nest/nest-simulator/jobs/727040026#L947), I have now implemented the solution with the additional environment variable described above. This seems to be working fine.

jougs:
@clinssen: are you happy with the recent changes? I think we can just pretend that Travis is, based on the latest build of the branch itself: https://travis-ci.org/github/jougs/nest-simulator/builds/727066507. And we really should get the macOS build times down...

clinssen (approved):
Many great improvements. Approving this so we can merge and use as a basis for further development.

Should tests of existing PRs be re-run, just in case?

For further discussion on our Travis CI, see https://github.com/nest/nest-simulator/pull/1764.

jougs:
Many thanks for the reviews!

PR #1711, opened 2020-08-05
Add tests for GetConnections with synapse_model and synapse_label as arguments

### LINKED ISSUE #1669: Add tests for specifying synapse_model when calling GetConnections
You can specify a `synapse_model` when getting connections from `GetConnections(synapse_model=None)`. This is not tested for however. We have a [test_getconnections.py](https://github.com/nest/nest-simulator/blob/master/pynest/nest/tests/test_getconnections.py), but it has nothing about `synapse_model` or `synapse_label` and should be extended.

Add tests for `GetConnections(synapse_model=...)` and `GetConnections(synapse_label=...)`.

stinebuu:
Add tests to `test_getconnections.py` so we test for `GetConnections()` with `synapse_model` and `synapse_label` as arguments.

This PR fixes #1669.

### DIFF
--- a/pynest/nest/tests/test_getconnections.py
+++ b/pynest/nest/tests/test_getconnections.py
@@ -138,6 +138,39 @@ class GetConnectionsTestCase(unittest.TestCase):
             conns = nest.GetConnections(source=src, synapse_model=synapse_model)
             self.assertEqual(reference_list, conns.synapse_model)
 
+    def test_GetConnectionsSynapseLabel(self):
+        """GetConnections using synapse_label as argument"""
+
+        labeled_synapse_models = [s for s in nest.Models(mtype='synapses') if s.endswith("_lbl")]
+
+        for synapse_model in labeled_synapse_models:
+            nest.ResetKernel()
+
+            src = nest.Create('hh_psc_alpha_gap', 3)
+            trgt = nest.Create('hh_psc_alpha_gap', 5)
+
+            # First create one connection with static_synapse
+            nest.Connect(src[0], trgt[0])
+
+            try:
+                # Connect with specified synapse
+                nest.Connect(src, trgt, syn_spec={'synapse_model': synapse_model, "synapse_label": 123, })
+            except nest.kernel.NESTError:
+                # If we can't connect iaf_psc_alpha with the given synapse_model, we ignore it.
+                continue
+
+            reference_list = [synapse_model] * 3 * 5
+
+            # Call GetConnections with specified synapse_label and test that connections with
+            # corresponding model are returned
+            conns = nest.GetConnections(synapse_label=123)
+            self.assertEqual(reference_list, conns.synapse_model)
+            self.assertEqual([123] * 3 * 5, conns.synapse_label)
+
+            # Also test that it works if we specify source and synapse_label
+            conns = nest.GetConnections(source=src, synapse_label=123)
+            self.assertEqual(reference_list, conns.synapse_model)
+
 
 def suite():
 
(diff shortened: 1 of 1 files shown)


pynest/nest/tests/test_getconnections.py
```diff
@@ -108,6 +111,69 @@ def test_GetConnectionsSourceModels(self):
...
+            self.assertEqual(reference_list, conns.synapse_model)
+
+    def test_GetConnectionsSynapseLabel(self):
+        """GetConnections using synapse_label as argument"""
+
+        labeled_synapse_models = [s for s in nest.Models(mtype='synapses') if s.endswith("_lbl")]
+
+        for synapse_model in labeled_synapse_models:
+            nest.ResetKernel()
+
```
Silmathoron, line 151: I'd suggest adding 

```python
label = 123
num_source = 3
num_target = 5
```

and removing all numbers below.
(and doing the same for the previous test)

pynest/nest/tests/test_getconnections.py
```diff
@@ -108,6 +111,69 @@ def test_GetConnectionsSourceModels(self):
...
+            nest.Connect(src[0], trgt[0])
+
+            try:
+                # Connect with specified synapse
+                nest.Connect(src, trgt, syn_spec={'synapse_model': synapse_model, "synapse_label": 123, })
+            except nest.kernel.NESTError:
+                # If we can't connect iaf_psc_alpha with the given synapse_model, we ignore it.
+                continue
+
+            reference_list = [synapse_model] * 3 * 5
```
Silmathoron, line 165: also add ``label_list = [label]*num_source*num_target``

pynest/nest/tests/test_getconnections.py
```diff
@@ -28,8 +28,11 @@
 
 nest.set_verbosity('M_ERROR')
 
+HAVE_GSL = nest.ll_api.sli_func("statusdict/have_gsl ::")
+
 
 @nest.ll_api.check_stack
+@unittest.skipIf(not HAVE_GSL, 'GSL is not available')
```
Silmathoron, line 35: I'm not sure why this is necessary, I've probably missed something... couldn't we use only models that do not require GSL? I don't think invalid models would appear in ``Models`` if they were not compiled, would they?

If gap junction need to be tested specifically, then using an ``if HAVE_GSL else`` with different models for hh_psc_alpha_gap?

pynest/nest/tests/test_getconnections.py
```diff
@@ -108,6 +111,69 @@ def test_GetConnectionsSourceModels(self):
...
+                continue
+
+            reference_list = [synapse_model] * 3 * 5
+            if synapse_model == 'static_synapse':
+                reference_list += ['static_synapse']
+
+            conns = nest.GetConnections(synapse_model=synapse_model)
+            self.assertEqual(reference_list, conns.synapse_model)
+
+            # Also test that it works if we specify source and synapse_model
```
Silmathoron, line 140: there's no need to check for target?

stinebuu: You are right, I guess I was just lazy, which you should never be when you write tests! I have added it now :)

pynest/nest/tests/test_getconnections.py
stinebuu: You are right, I haven't used `label` for synapses before, so I just followed an example that use `hh_psc_alpha_gap`. Have switched to normal `iaf_psc_alpha` and removed the `HAVE_GSL`, and the tests pass.

hakonsbm (approved):
Looks good, thanks!

PR #1444, opened 2020-02-26
Refactor PyNEST Microcircuit example Potjans_2014

jhnnsnk:
This PR revises code and documentation of the PyNEST Microcircuit and adjusts the example to NEST3.
Summary of the main changes:

- NEST3-specific
  - remove loops on Python level (using NodeCollections and `nest.random` for distributions)
  - use `nodes.set(params)` and `nodes.parameter=value`

- use `'V0_type': 'optimized'` as default initial condition
- file `example.py` is renamed to `run_microcircuit.py`, the folder `data` is now `reference_data`
- revise all functions in detail
  - change the order of functions and parameter definitions in the respective files for a better overview
  - get rid of redundant functions and parameter definitions
  - remove unnecessary loops and nested if conditions
  - revise all function names and update many variable names to be more descriptive
  - store only those parameters as class attributes for which this is needed
  - provide explicit arguments to functions in `helpers.py` instead of dictionaries
  - define only one `tau_syn` instead of `tau_syn_ex` and `tau_syn_in` with the same value (only one `PSC_e_over_PSP_e` was computed anyways)
  - scale number of neurons in raster plot based on `N_scaling`
- new structure of `Network class` with distinction between create and connect and use of private methods
  - disentangle derivation of dependent parameters from create and connect calls by introducing function `__derive_parameters()`
  - synapse numbers are now only once cast as int
- detailed time measurement (total time, initialize, create, connect, simulate, evaluate) on each rank
- update all docstrings, comments and printouts (use `.format()` instead of `%`) as well as file headers (as discussed with @sarakonradi) 
- new reference data generated with default values
- autopep8
- refactor README.rst (as discussed with @sarakonradi)
  - update structure and text (details on parameters and usage)
  - add images
  - add acknowledgments, citation, information about this PyNEST Microcircuit and other versions
  
The following points still need discussions or decisions:

1. The amplitude of DC input for stimulation is multiplied with `K_ext`. Shall we keep that?
2. The number of thalamic neurons is not scaled with `N_scaling`. Is this intended?
3. What is the best practice for naming parameters (e.g., should mean be prefix or suffix)?
4. How were the original values of `full_mean_rates` computed? Shall we replace them by values obtained with this version? If yes, which combination of MPI processes and threads shall we choose?

PR #1715, opened 2020-08-07
Efficient initialization of membrane potentials with original and optimized conditions in Potjans_2014

jhnnsnk:
This PR refactors the initialization of membrane potentials in the PyNEST example Potjans_2014 for the backport version NEST 2.14.1.

* Adds the optimized initial conditions (introduced in NEST 2.20.0). They are now default.
* Revises the implementation of the initialization to be more efficient.
* Starts section in `readme.md` for modifications done in NEST 2.14.1.

### DIFF
--- a/pynest/examples/Potjans_2014/network.py
+++ b/pynest/examples/Potjans_2014/network.py
@@ -238,8 +238,8 @@ class Network:
         """
         Initializes membrane potentials by sampling from a normal distribution.
 
-        One Python random number generator per virtual process (vp) is used.
-        This results in identical initial conditions if the number of vps,
+        One Python random number generator per virtual process (VP) is used.
+        This results in identical initial conditions if the number of VPs,
         meaning the product of MPI processes and the number of OpenMP threads
         per MPI process, is preserved.
 
@@ -260,13 +260,13 @@ class Network:
         mask = list(nest.GetStatus(neurons, 'local'))
         neurons_loc = np.array(neurons)[mask].tolist()
 
-        # get sequence of vps on this MPI process
+        # get sequence of VPs on this MPI process
         num_threads = nest.GetKernelStatus()['local_num_threads']
         local_vps = nest.GetStatus(neurons_loc[:num_threads], 'vp')
 
         for ivp, vp in enumerate(local_vps):
             # use knowledge that the next neuron of the same MPI process with
-            # the same vp is num_threads away
+            # the same VP is num_threads away
             neurons_vp = neurons_loc[ivp::num_threads]
             nest.SetStatus(neurons_vp, 'V_m',
                 self.pyrngs[vp].normal(V0_mean, V0_sd, len(neurons_vp)))
--- a/pynest/examples/Potjans_2014/network.py
+++ b/pynest/examples/Potjans_2014/network.py
@@ -160,6 +160,19 @@ class Network:
                     )
             self.DC_amp_e = compute_DC(self.net_dict, self.w_ext)
 
+        v0_type_options = ['optimized', 'original']
+        if self.net_dict['V0_type'] not in v0_type_options:
+            print(
+                '''
+                '{0}' is not a valid option for 'V0_type',
+                replacing it with '{1}.'
+                Valid options are {2}.
+                '''.format(self.net_dict['V0_type'],
+                           v0_type_options[0],
+                           v0_type_options)
+                )
+            self.net_dict['V0_type'] = v0_type_options[0]
+
         if nest.Rank() == 0:
             print(
                 'The number of neurons is scaled by a factor of: %.2f'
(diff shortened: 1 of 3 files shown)


pynest/examples/Potjans_2014/network.py
```diff
@@ -160,6 +160,19 @@ def create_populations(self):
...
+            print(
+                '''
+                '{0}' is not a valid option for 'V0_type',
+                replacing it with '{1}.'
+                Valid options are {2}.
+                '''.format(self.net_dict['V0_type'],
+                           v0_type_options[0],
+                           v0_type_options)
+                )
+            self.net_dict['V0_type'] = v0_type_options[0]
```
jougs, line 174: I don't like that it automatically chooses a value if the given value is invalid and I'd really prefer it to fail loudly with an error in this case. The way it is now, the network would in any case provide results, but their interpretation might be wrong as the initial conditions might be not the ones intended (if there was a typo, for instance).

I suggest to change it here and also in the master/2.20.1 version of the script, if that's the same.

pynest/examples/Potjans_2014/network.py
```diff
@@ -199,26 +212,65 @@ def create_populations(self):
...
+        V0_mean
+            Mean value of normal distribution (in mV).
+        V0_sd
+            Standard deviation of normal distribution (in mV).
+        """
+        # find neurons local to this MPI process
+        mask = list(nest.GetStatus(neurons, 'local'))
+        neurons_loc = np.array(neurons)[mask].tolist()
+
+        # get sequence of vps on this MPI process
```
jougs, line 263:
```suggestion
        # get sequence of VPs on this MPI process
```

pynest/examples/Potjans_2014/network.py
```diff
@@ -199,26 +212,65 @@ def create_populations(self):
...
+        num_threads = nest.GetKernelStatus()['local_num_threads']
+        local_vps = nest.GetStatus(neurons_loc[:num_threads], 'vp')
+
+        for ivp, vp in enumerate(local_vps):
+            # use knowledge that the next neuron of the same MPI process with
+            # the same vp is num_threads away
```
jougs, line 269:
```suggestion
            # the same VP is num_threads away
```

jougs (changes requested):
Thanks for this backport. I've left you some inline comments and suggestions.

jougs:
The failures on Travis are due to changes in libneurosim master. The solution is most likely to change the installation script of libneurosim to `git checkout acc5b1a6445c82d66fba9bb8681360e00d7a3c9d` after cloning (the hash comes from `git rev-list -n 1 --first-parent --before="2017-10-20 16:33" master`, the date being the release date of NEST 2.14.0). Something similar might be needed for the CSA and MUSIC installations, as they probably also moved on a bit in the past three years.

If you want, I can attempt the required changes in your branch.

pynest/examples/Potjans_2014/network.py
terhorstd: I agree, the warning will absolutely not be seen in the output. The concept should always be "fail noisy, succeed silently". Something unexpected should break the run, otherwise we couldn't trust the results, because we never know how many warnings we didn't see.

jhnnsnk: I totally agree. This and other observations motivated me to [refactor the whole microcircuit](https://github.com/nest/nest-simulator/pull/1444). In my new implementation, an [Exception](https://github.com/nest/nest-simulator/blob/master/pynest/examples/Potjans_2014/network.py#L334) punishes wrong input.
However, since this is a backport, I assumed that it would be useful to include the whole [PR introducing the new initial conditions](https://github.com/nest/nest-simulator/pull/1260) and only change the algorithm.
But if you also think that this warning does more harm than good, I will just update it as well.

jougs: I definitely am all in favour of changing this! Please go ahead!

But quite honestly, I'm not yet convinced of the purpose of backporting these kind of changes in the first place. Whom are we expecting to actually use a three year old release in practice? Let alone update it to get new results with it? In my humble opinion I would only backport real bug fixes, but leave all examples and such unchanged. If a user wants to do such a comparison, they could always do the backport themselves...

jougs: BTW: thanks for linking to the corresponding code locations in such a nice way. It's really refreshing to see some Markdown being put to good use! :heart:

clinssen (approved):
Looks good to me. I'm not sure how much time we can reasonably ask of @jhnnsnk to fix our Travis CI instances from 2017.

pynest/examples/Potjans_2014/network.py
jhnnsnk: Fixed.

pynest/examples/Potjans_2014/network.py
jhnnsnk: Fixed.

pynest/examples/Potjans_2014/network.py
jhnnsnk: Now, a wrong `V0_type` triggers an Exception.

The reason for this backport is that NEST 2.14 serves as reference for benchmarks with models like Potjans_2014.

jougs (approved):
Thanks for addressing my remarks. I'm approving and think we can safely ignore the Travis issues.

jougs:
@terhorstd Who is merging changes to these kind of release branches? Could you?

PR #1718, opened 2020-08-11
Fix GetConnections to not include erroneous connections when using multiple threads

### LINKED ISSUE #1670: Incorrect connections are returned by GetConnections
**Describe the bug**
When multithreading is used, some additional connections are returned by ``GetConnections`` when using the ``source`` keyword (might occur for other keywords in different circumstances).

**To Reproduce**
```python
import nest

nest.SetKernelStatus({"local_num_threads": 6})

nn = nest.Create("iaf_psc_alpha", 1000)

pg = nest.Create("poisson_generator")

nest.Connect(pg, nn)

sd = nest.Create("spike_detector")
nest.Connect(nn, sd)

conn = nest.GetConnections(source=pg)
print(len(conn))

pg_gid = pg[0]

print([c for c in conn if c[0] != pg_gid])
```
prints ``1004`` and

    [array('l', [998, 1002, 2, 15, 0]), array('l', [999, 1002, 3, 15, 0]), array('l', [1000, 1002, 4, 15, 0]), array('l', [997, 1002, 1, 15, 0])]

**Expected behavior**
Should print ``1000`` and ``[]``.

 - OS: Manjaro
 - Python-Version: 3.8
 - NEST-Version: 2.20.1 (fbd272e107a69f5df159595ee5b216bb78e4b193)

hakonsbm:
If a population is connected to devices both as source and target, getting connections with specified sources may give erroneous connections if multiple threads are used.

When calling `GetConnections` with specified source nodes, it will at some point attempt to collect connections where the post-synaptic nodes are devices. While iterating the source nodes, the specified source node ID is then converted to a thread-local ID, `lid`, which is converted back to a node ID later. However, when using multiple threads, the `lid` on one or more threads may not match the specified node ID. Then, if a connection exist where the erroneous node ID is a source, this erroneous connection is included in the returned collection of connections.

This PR adds a simple check of the `lid` when getting connections with devices as post-synaptic nodes, to see if it matches the specified node ID. A regressiontest is also added.

Fixes #1670

### DIFF
--- a/nestkernel/target_table_devices.cpp
+++ b/nestkernel/target_table_devices.cpp
@@ -117,6 +117,10 @@ nest::TargetTableDevices::get_connections_to_devices_( const index requested_sou
   if ( requested_source_node_id != 0 )
   {
     const index lid = kernel().vp_manager.node_id_to_lid( requested_source_node_id );
+    if ( kernel().vp_manager.lid_to_node_id( lid ) != requested_source_node_id )
+    {
+      return;
+    }
     get_connections_to_device_for_lid_( lid, requested_target_node_id, tid, syn_id, synapse_label, conns );
   }
   else
--- a/testsuite/regressiontests/issue-1640.sli
+++ b/testsuite/regressiontests/issue-1640.sli
@@ -1,0 +1,91 @@
+/*
+ *  issue-1640.sli
+ *
+ *  This file is part of NEST.
+ *
+ *  Copyright (C) 2004 The NEST Initiative
+ *
+ *  NEST is free software: you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


jougs:
BTW: the failing checks is just  timeout on macOS. The code is good.

heplesser:
I have restarted the OSX build. According to the log, it got stuck for 10+ minutes in the `ticket-772.sli`. I actually am confused by that test, because as far as I can see, the test should fail: According to the documentation in the test, we want to make sure NEST raises an exception, but the code has just `pass_or_die` tests.

heplesser:
All tests passed now, thus merging.

Issue #1714, opened 2020-08-06
test_connect_all_patterns.py is broken with mpi4py


espenhgn:
**Describe the bug**
test_connect_all_patterns.py test file is broken if mpi4py is correctly installed. 
TravisCI skips this test as mpi4py is not found, see e.g, https://api.travis-ci.org/v3/job/715501240/log.txt, which may explain why the error below hasn't been captured:
```
$ nosetests /Users/ehagen/anaconda3/envs/nest/lib/python3.8/site-packages/nest/tests/test_connect_all_patterns.py

              -- N E S T --
  Copyright (C) 2004 The NEST Initiative

 Version: fix-1678-broken-build@74dfd2cb6
 Built: Aug  6 2020 09:27:47

 This program is provided AS IS and comes with
 NO WARRANTY. See the file LICENSE for details.

 Problems or suggestions?
   Visit https://www.nest-simulator.org

 Type 'nest.help()' to find out more about NEST.

F
======================================================================
FAIL: testWithMPI (nest.tests.test_connect_all_patterns.TestConnectAllPatterns)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/ehagen/anaconda3/envs/nest/lib/python3.8/site-packages/nest/tests/test_connect_all_patterns.py", line 59, in testWithMPI
    self.assertTrue(not failing_tests, 'The following tests failed when ' +
AssertionError: False is not true : The following tests failed when executing with "mpirun -np 2 nosetests [script]": test_connect_all_to_all.py, test_connect_one_to_one.py, test_connect_fixed_indegree.py, test_connect_fixed_outdegree.py, test_connect_fixed_total_number.py, test_connect_pairwise_bernoulli.py
```
The basic issue is that `subprocess.Popen` is attempting to run an MPI job ("mpirun -np 2 nosetest <test file>") from an MPI process (even if it's a single thread), and was introduced by merging https://github.com/nest/nest-simulator/pull/807. This should be fixed before merging https://github.com/nest/nest-simulator/pull/1712 where the issue was encountered. 

**To Reproduce**

1. Compile NEST with mpi4py (see https://github.com/nest/nest-simulator/issues/1678 for details)
2. Run the test script:
`$ nosetests <path to site-packages>/nest/tests/test_connect_all_patterns.py`
3. The underlying issue can be exposed interactively as well:
```
In [1]: import subprocess as sp                                                                                                                                                          

[...]

PR #1720, opened 2020-08-12
fix test_connect_all_patterns.py on TravisCI

espenhgn:
Should fix issue #1714


lekshmideepu:
Exactly. This test was not running under MacOS from the very beginning as the OsX builds were running without MPI. As I  wrote [here](https://github.com/nest/nest-simulator/issues/1714), it was available for linux builds earlier.

PR #1724, opened 2020-08-18
Check if parameter is in model dictionary before setting it

### LINKED ISSUE #1575: Providing key:dictionary pair to SetStatus via PyNEST generates uninformative error message
Calling the following line in pynest (3115cf68bda26e282c7f7a20fc00a7755eeb2b5b)
```python
nest.SetStatus(n, {'test_dict': {}})
```
leads to the following error:
```bash
nest.lib.hl_api_exceptions.DictError: ("DictError in get_d: Key '/test_dict' does not exist in dictionary.", 'DictError', 'get_d', ": Key '/test_dict' does not exist in dictionary.")
```

instead of the more informative response from SLI (`n << /test_dict << >> >> SetStatus`):
```bash
May 01 16:34:40 SetStatus_id [Error]: DictError
    Unused dictionary items:  test_dict
```

This in itself would be just a minor inconvencience, but if I actually implement `test_dict` in the corresponding model
```c++
  DictionaryDatum test_dict;
  if ( d->known( "test_dict" ) )
  {
    updateValue< DictionaryDatum >( d, "test_dict", test_dict );
  }
```

pynest still throws the same error as above, while SLI succeeds.

I'm digging into the pynest api, but maybe @hakonsbm @stinebuu could point me in the right direction?

stinebuu:
This PR fixes #1575. 

If you can `set` a parameter of a model, but not `get` it, an error used to be thrown on PyNEST level when you tried to use `SetStatus` or `set`. This is no longer the case.

Note that if the parameter you want to set is a single value (double/int/float..), you can not send in a list of values the length of the NodeCollection, unless the parameter is also get-able. That is, if we have
```
Double test_val;
if ( d->known( "test_val" ) )
{
  updateValue< double >( d, "test_val", test_val );
}
```
implemented in the `set` function of the model, but not in the `get` function of the model, the following is **not** possible
```
nodes.set('test_val', [3., 2., 1.])
```

### DIFF
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -271,14 +271,13 @@ def SetStatus(nodes, params, val=None):
     if len(nodes) == 0:
         return
 
-    n0 = nodes[0]
     params_is_dict = isinstance(params, dict)
     set_status_nodes = isinstance(nodes, nest.NodeCollection)
     set_status_local_nodes = set_status_nodes and nodes.get('local')
 
     if (params_is_dict and set_status_local_nodes):
 
-        node_params = n0.get()
+        node_params = nodes[0].get()
         contains_list = [is_iterable(vals) and key in node_params and not is_iterable(node_params[key]) for
                          key, vals in params.items()]
 
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -273,7 +273,7 @@ def SetStatus(nodes, params, val=None):
 
     params_is_dict = isinstance(params, dict)
     set_status_nodes = isinstance(nodes, nest.NodeCollection)
-    set_status_local_nodes = set_status_nodes and nodes.get('local')
+    set_status_local_nodes = set_status_nodes and all(nodes.local)
 
     if (params_is_dict and set_status_local_nodes):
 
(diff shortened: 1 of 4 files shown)


pynest/nest/lib/hl_api_info.py
```diff
@@ -274,11 +274,13 @@ def SetStatus(nodes, params, val=None):
     n0 = nodes[0]
```
clinssen, line 274: Small nitpick: could you move this into the if below?

clinssen (changes requested):

pynest/nest/lib/hl_api_info.py
stinebuu: Of course!

pynest/nest/lib/hl_api_info.py
```diff
@@ -271,14 +271,15 @@ def SetStatus(nodes, params, val=None):
     if len(nodes) == 0:
         return
 
-    n0 = nodes[0]
     params_is_dict = isinstance(params, dict)
     set_status_nodes = isinstance(nodes, nest.NodeCollection)
-    set_status_local_nodes = set_status_nodes and n0.get('local')
+    set_status_local_nodes = set_status_nodes and nodes.get('local')
```
hakonsbm, line 276: Now that we use the whole NodeCollection here, we should check that all are local.
```suggestion
    set_status_local_nodes = set_status_nodes and all(nodes.local)
```
Or should we still check only the first node?

hakonsbm (changes requested):

hakonsbm (approved):
Thanks!

clinssen (approved):
Cheers!

PR #1732, opened 2020-08-24
Update hpc_benchmark.py to avoid error when running with multiple processes

### LINKED ISSUE #1729: Running hpc_benchmark.py with multiple processes raises an error
Because of changes to NodeCollections, running hpc_benchmark.py with multiple processes raises an error:
```
nest.lib.hl_api_exceptions.BadProperty: ('BadProperty in Take_g_a: Cannot slice a sliced composite NodeCollection.', 'BadProperty', 'Take_g_a', ': Cannot slice a sliced composite NodeCollection.')
```
This is because `nest.GetLocalNodeCollection()` slices the NodeCollection and returns a composite NodeCollection with step > 1.

The relevant slicing in in hpc_benchmark.py, here:
https://github.com/nest/nest-simulator/blob/7871dd145e8bee74045fd2468fd9be2d0d0cb913/pynest/examples/hpc_benchmark.py#L338-L339

should be rewritten to avoid slicing the sliced composite NodeCollection.

hakonsbm:
When using multiple processes, the `local_neurons` NodeCollection becomes a stepped composite NodeCollection, which cannot be sliced. To make it sliceable, `local_neurons` is rebuilt as a plain composite NodeCollection, which fixes #1729.

This is the simplest solution to #1729. The downside being that using the non-stepped NodeCollection will perform a bit worse when iterated during connection. However, this only affects connection between the selected number of excitatory neurons and spike detectors. Alternatively, passing slicing arguments can be added to `GetLocalNodeCollection` to be able to keep the sliced NodeCollection a stepped composite.

### DIFF
--- a/pynest/examples/hpc_benchmark.py
+++ b/pynest/examples/hpc_benchmark.py
@@ -323,8 +323,9 @@ def build_network(logger):
     if params['record_spikes']:
         if params['nvp'] != 1:
             local_neurons = nest.GetLocalNodeCollection(E_neurons)
-            # Rebuilding local_neurons to get rid of the stepped composite NodeCollection,
-            # to be able to slice it.
+            # GetLocalNodeCollection returns a stepped composite NodeCollection, which
+            # cannot be sliced. In order to allow slicing it later on, we're creating a
+            # new regular NodeCollection from the plain node IDs.
             local_neurons = nest.NodeCollection(local_neurons.tolist())
         else:
             local_neurons = E_neurons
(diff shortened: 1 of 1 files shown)


pynest/examples/hpc_benchmark.py
```diff
@@ -323,6 +323,9 @@ def build_network(logger):
     if params['record_spikes']:
         if params['nvp'] != 1:
             local_neurons = nest.GetLocalNodeCollection(E_neurons)
+            # Rebuilding local_neurons to get rid of the stepped composite NodeCollection,
+            # to be able to slice it.
```
jougs, line 327:
```suggestion
            # GetLocalNodeCollection returns a stepped composite NodeCollection, which
            # cannot be sliced. In order to allow slicing it later on, we're creating a
            # new regular NodeCollection from the plain node IDs.
```

jougs (commented):
I have just a small update to the comment, otherwise I'm fine. Thanks for fixing this!

stinebuu (approved):

jougs (approved):

PR #1740, opened 2020-08-30
Changes to volume transmitter documentation

jd41:
Hi, I was quite confused by the volume transmitter documentation and only understood it after asking someone, particularly because I didn't catch how the user specifies which synapses the thing would be connected to (which I now see is mentioned in the very end of the text, but I didn't look there before - instead, I tried and failed to understand some SLI code in an older version of the documentation, and looked in the code I am working with). I propose some changes.

I

- shifted the crucial information how someone would specify the synapses a VT is connected to closer to where the reader would wonder about it

- split into paragraphs

- small typo(s)

- colon in line 65: Make it clear that the "hybrid structure" is described in the following sentences.

I am unsure about what's the character limit per line in the documentation; I'm happy to rearrange the lines if I got it wrong.

### DIFF
--- a/models/volume_transmitter.h
+++ b/models/volume_transmitter.h
@@ -64,15 +64,16 @@ in the synapses itself.
 The volume transmitter interacts in a hybrid structure with the neuromodulated
 synapses: In addition to the delivery of the neuromodulatory spikes triggered
 by every pre-synaptic spike, the neuromodulatory spike history is delivered
-at regular time intervals. The interval is an integer multiple of the minimal
-synaptic delay ``d_min``.
+at regular time intervals. The interval is equal to deliver_interval * d_min,
+where deliver_interval is an entry in the parameter dictionary and d_min
+is the minimal synaptic delay.
 
 The implementation is based on the framework presented in [1]_.
 
 Parameters
 ++++++++++
 
-- deliver_interval - time interval given in d_min time steps, in which
+- deliver_interval - time interval given in d_min time steps in which
                      the volume signal is delivered from the volume
                      transmitter to the assigned synapses
 
--- a/models/volume_transmitter.h
+++ b/models/volume_transmitter.h
@@ -64,7 +64,7 @@ in the synapses itself.
 The volume transmitter interacts in a hybrid structure with the neuromodulated
 synapses: In addition to the delivery of the neuromodulatory spikes triggered
 by every pre-synaptic spike, the neuromodulatory spike history is delivered
-in discrete time intervals of a manifold of the minimal synaptic delay.
+at regular time intervals. The interval is equal to the minimal synaptic delay.
 
 The implementation is based on the framework presented in [1]_.
 
--- a/models/volume_transmitter.h
+++ b/models/volume_transmitter.h
@@ -65,8 +65,8 @@ The volume transmitter interacts in a hybrid structure with the neuromodulated
 synapses: In addition to the delivery of the neuromodulatory spikes triggered
 by every pre-synaptic spike, the neuromodulatory spike history is delivered
 at regular time intervals. The interval is equal to ``deliver_interval * d_min``,
-where ``deliver_interval`` is an entry in the parameter dictionary and ``d_min``
-is the minimal synaptic delay.
+where ``deliver_interval`` is an (integer) entry in the parameter dictionary and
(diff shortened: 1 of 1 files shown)


jd41:
I am not qualified to do this, but it would also be nice to include a list - or a link to a list - of neuron models that support receiving signals from the `volume_transmitter`.

models/volume_transmitter.h
```diff
@@ -48,23 +48,25 @@ Description
...
+
+It is assumed that the neuromodulatory signal is a function of the
+spike times of all spikes emitted by the population of neurons connected
+to the volume transmitter. The neuromodulatory dynamics is calculated
+in the synapses itself.
+
+The volume transmitter interacts in a hybrid structure with the neuromodulated
+synapses: In addition to the delivery of the neuromodulatory spikes triggered
+by every pre-synaptic spike, the neuromodulatory spike history is delivered
+in discrete time intervals of a manifold of the minimal synaptic delay.
```
clinssen, line 67: "manifold" seems like not the right word to use here.

clinssen (changes requested):

clinssen (approved):
Perfect, thanks!

jd41:
Wait, @clinssen, before this gets merged: Can you confirm that the formulation I just used actually describes the behaviour correctly?

models/volume_transmitter.h
jd41: I changed it, but would like to hear explicitly that I didn't misunderstand the previous formulation and what is now written there is correct.

models/volume_transmitter.h
```diff
@@ -48,23 +48,25 @@ Description
...
+
+It is assumed that the neuromodulatory signal is a function of the
+spike times of all spikes emitted by the population of neurons connected
+to the volume transmitter. The neuromodulatory dynamics is calculated
+in the synapses itself.
+
+The volume transmitter interacts in a hybrid structure with the neuromodulated
+synapses: In addition to the delivery of the neuromodulatory spikes triggered
+by every pre-synaptic spike, the neuromodulatory spike history is delivered
+at regular time intervals. The interval is equal to the minimal synaptic delay.
```
clinssen, line 67:
```suggestion
at regular time intervals. The interval is an integer multiple of the minimal
synaptic delay ``d_min``.
```
@jd41: good call! This seems to be the right interpretation, consistent with the comment a few lines below:
```
- deliver_interval - time interval given in d_min time steps, in which
                     the volume signal is delivered from the volume
                     transmitter to the assigned synapses
```

clinssen (changes requested):

models/volume_transmitter.h
jd41: Alright, thanks! Then I mark it as resolved.

jd41: Oh, I changed it a bit more. Similar to how the reader's question how the user specifies the receiving synapses should be immediately answered (see first PR message), I think that it should also become immediately clear to the reader how that multiple of d_min is determined.

clinssen:
I think the new changes are an improvement, but they remove all mentions of the type of ``deliver_interval``; before, it said "integer multiple of ``d_min``", so ``deliver_interval`` would be known to be an integer ≥ 0. Could you work this into the new description?

jd41:
@clinssen Good point, I changed it. I hope Travis doesn't time out on the Mac OS build this time.

clinssen (approved):
Many thanks!

stinebuu:
@YounesBouhadjar Do you have time to review this?

models/volume_transmitter.h
```diff
@@ -48,30 +48,34 @@ Description
 +++++++++++
 
 The volume transmitter is used in combination with neuromodulated
-synaptic plasticty, plasticity that depends not only on the activity
+synaptic plasticity, plasticity that depends not only on the activity
```
YounesBouhadjar, line 51: maybe you can rephrase this sentence to something like: The volume transmitter is used in combination with neuromodulated synaptic plasticty, this plasticity depends not only on the activity ...

it's just a small thing, I leave you the choice to ignore or accept this

YounesBouhadjar (commented):
Looks good, I just have a small comment. 
Thanks a lot and sorry for the long delay to review this.

YounesBouhadjar (approved):

PR #1741, opened 2020-09-01
Shrink MPI buffers for spike communication

H27CK:
This PR provides a simple mechanism by which spike buffers are decreased in size. If only one gather round is needed to communicate all spikes, we assume that the buffer can be shrunk by a factor of 1.1. This shrinkage does not resize the underlying vector, it only affects the count argument passed to `MPI_Alltoall()`. This should reduce the amount of 'padding' communicated.


H27CK:
@jasperalbers @jakobj this should be more robust and efficient than the previous PR. Jasper, I've ran some of my own tests with the multi-area model on JUSUF. I used `OMP_PROC_BIND=true`. All tests performed with the following modules:

- `Intel/2019.5.281-GCC-8.3.0`
- `ParaStationMPI/5.2.2-1-mt`
- `jemalloc/5.1.0`
- `GSL/2.5`
- `Python SciPy-Stack`

With 12 nodes, 4 MPI processes, 32 OpenMP threads/process:
- current master: > 1 hour (wall-time limit);
- this PR: 46 minutes to complete.

With 6 nodes, 4 MPI processes, 32 OpenMP threads/process:

- current master: > 37 minutes;
- this PR: 18 minutes to complete.

These configurations aren't necessarily the most optimum, but they serve to demonstrate that buffers are excessively large.

jasperalbers:
@H27CK Thanks for this PR. I have a few questions/remarks.
First, did you mix up the numbers, or why do you get a faster simulation time on just 6 nodes as compared to 12?
Second, what exactly is the time you give in minutes, the total simulation time, the state propagation time or the communication time?

I did some tests myself on JURECA, where I found an improvement over current master, although not as drastic as you observed. I have sent the figure to you via email as the GitLab repo is not up yet. The observable reduction in communication time is most prominent on few (~20) nodes as compared to many (~180) nodes. Is this something one would expect?

H27CK:
@jasperalbers even though we've discussed this elsewhere, I thought it would be good to give an update here.

No, there was no mistake. For 12 nodes, with the number of MPI processes and OpenMP threads used, times are worse. Henceforth, I'll use RTF for simulation completion instead of physical timings.

I've done some more work on JUSUF. With suitable pinning/binding and 35 nodes, 2 MPI processes, 32 OpenMP threads, I see RTF 17.42 with this PR. For the same configuration, master gives 64.75 RTF.

terhorstd:
Friendly ping from the Hackathon... any updates on this?

jasperalbers:
@terhorstd Not from my side. I will be testing on JUSUF in the near future to confirm the findings. It is possible that the changes this PR proposes are not as beneficial for JURECA-like systems as they are for JUSUF-like systems as their hardware are substantially different.

jougs:
@terhorstd: did you have a chance to check on JUSUF already?

jasperalbers:
@jougs : I have run some tests with the multi-area model on JUSUF using current nest master and nest master + the new spike buffer adaptations. Roughly speaking, I can confirm @H27CK findings! I simulated for 10 model seconds using 2 MPI tasks per node, 32 threads per node (with socket pinning) and got the following numbers on 70 nodes, which represents the fastest of my simulations (I have also created more detailed plots if you want to see them):

- master: simulation time of 650s (i.e. real time factor of 65)
- shrink-buffers: simulation time of 150s (i.e. real time factor of 15)

Thus, this PR is much more effective on JUSUF than it was on JURECA.
As a validation test I compared the number of spikes in the two simulations, which turned out to be identical (109195016).

nestkernel/event_delivery_manager.cpp
```diff
@@ -143,10 +145,13 @@ EventDeliveryManager::resize_send_recv_buffers_target_data()
 void
 EventDeliveryManager::resize_send_recv_buffers_spike_data_()
 {
-  send_buffer_spike_data_.resize( kernel().mpi_manager.get_buffer_size_spike_data() );
-  recv_buffer_spike_data_.resize( kernel().mpi_manager.get_buffer_size_spike_data() );
-  send_buffer_off_grid_spike_data_.resize( kernel().mpi_manager.get_buffer_size_spike_data() );
-  recv_buffer_off_grid_spike_data_.resize( kernel().mpi_manager.get_buffer_size_spike_data() );
+  if ( kernel().mpi_manager.get_buffer_size_spike_data() > send_buffer_off_grid_spike_data_.size() )
```
jakobj, line 148: i was a bit confused that you use the size of `off_grid_spike_data` here, maybe just use `send_buffer_spike_data`? it's not wrong, i just needed to check twice

nestkernel/event_delivery_manager.cpp
```diff
@@ -384,12 +392,21 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
...
       {
         buffer_size_spike_data_has_changed_ = kernel().mpi_manager.increase_buffer_size_spike_data();
+        decrease_buffer_size_spike_data_ = false;
       }
     }
 #pragma omp barrier
 
   } // of while
 
+#pragma omp single
```
jakobj, line 402: if i understand the algorithm correctly, then the buffers shrink after every communication round if no more than one round was required. doesn't this mean the buffers always shrink below the required size, then are increased again only to continue shrinking? given the choices for increase and decrease factors (for fixed number of spikes) this seems to imply a cycle length of ~5. have you thought about possibilities to reduce/remove this periodic increasing and decreasing of buffer sizes?

jakobj (commented):
awesome, thanks @H27CK for the implementation and @jasperalbers for the data, the timings look great!

i left on small remark and one question.

nestkernel/mpi_manager.h
```diff
@@ -548,6 +556,16 @@ MPIManager::increase_buffer_size_spike_data()
   }
 }
 
+inline void
+MPIManager::decrease_buffer_size_spike_data()
+{
+  assert( adaptive_spike_buffers_ );
+  if ( buffer_size_spike_data_ / shrink_factor_buffer_spike_data_ > 4.0 * get_num_processes() )
```
heplesser, line 563: Could you add a comment to explain the lower limit, especially the factor four?

heplesser:
@jasperalbers @H27CK This is an amazing speedup indeed, more than a factor four through a rather small change. Since you are using the multiarea model to benchmark, I assume the minimal delay is 0.1 ms, so the simulation requires MPI communication on every time step.

Do we understand why we gain so much? Changing buffer size should only affect the MPI communication, so a speedup of >4 would indicate that the current master spends 90% or so of the simulation time on MPI communication? 

I also noticed that @H27CK reported roughly the same real-time factors (17, 64) on 4 Sep as @jasperalbers reports today (15, 65), but @H27CK used 35 nodes on JUSUF while  @jasperalbers used 70. Does that make sense? And why only 32 threads given that the CPUs have 64?

jasperalbers:
@heplesser:
> @jasperalbers @H27CK This is an amazing speedup indeed, more than a factor four through a rather small change. Since you are using the multiarea model to benchmark, I assume the minimal delay is 0.1 ms, so the simulation requires MPI communication on every time step.

I believe that this is correct.

> Do we understand why we gain so much? Changing buffer size should only affect the MPI communication, so a speedup of >4 would indicate that the current master spends 90% or so of the simulation time on MPI communication?

This is also correct. When plotting the time spent in communication, delivery and update, it becomes apparent that ~80% of the simulation time is indeed used for communication (on JUSUF). On JURECA, this percentage was not as high, which could be a reason why we did not see any meaningful improvement there.

> I also noticed that @H27CK reported roughly the same real-time factors (17, 64) on 4 Sep as @jasperalbers reports today (15, 65), but @H27CK used 35 nodes on JUSUF while @jasperalbers used 70. Does that make sense? And why only 32 threads given that the CPUs have 64?
To your first point: I listed the values for 70 nodes as they give the absolute best performance in terms of wall-time. For 35 nodes I recorded real-time factors of (17, 68), thus a better performance/resource ratio.
To your second point: Indeed, we both observe a better performance when _not_ using all available threads. One possible cause that we came up with is that a slowdown may occur when too many cores/threads access the same L3 cache that is shared amongst them. @ackurth observed, in a study of the microcircuit on (nearly?) identical hardware, that this becomes an issue slowing down the simulation (or at least not speeding it up). But up till now it is mainly an observation, if you have any additional thoughts I would be happy to discuss them.

heplesser:
@jasperalbers Thanks for all the explanations, that makes sense. So essentially there is no scaling beyond 35 nodes in either case for this model. Interesting that we exhaust the threads, L3 collisions seem a plausible explanation. Can one use hardware counters to measure L3 misses (not in this PR!)?

So then the next question is why shrinking the buffers (by how much in total?), has such a massive effect on the time spent on MPI communication?

nestkernel/event_delivery_manager.cpp
H27CK: Yes, buffer size will oscillate around some value (assuming network activity is steady). Note, the 'shrink factor' was derived sometime ago, and may not be optimum (although I suspect this is case dependent). The desire is to vary the count argument of Alltoall by tracking network activity; we want to minimise the communication of 'padding' in the buffer (while avoiding use of Alltoallv). There have been many (very different) versions of the logic presented here, but these come with additional overhead associated with synchronisation, or some amount of compute. This version is embarrassingly simple, efficient and robust. Since only the count variable is modified at the end of a communication round (assuming that our communication vectors already have a size that is greater than necessary, otherwise `vector::resize` is invoked), this oscillation does not incur a high cost (we may have some additional communication rounds).

nestkernel/mpi_manager.h
H27CK: `2.0 * get_num_processes()` represents the minimum allowable size (as initialised). I have doubled this to distinguish the minimum buffer size set through shrinkage (initially for debugging purposes), and by:

`kernel().mpi_manager.set_buffer_size_spike_data( 2 * kernel().mpi_manager.get_num_processes() );`

The lower limit through shrinkage can be reduced, but I suggest otherwise. It is unlikely that activity is so low that all ranks will communicate fewer than 4 spikes to any other rank (once buffers have grown beyond `4.0 * get_num_processes()`).

nestkernel/event_delivery_manager.cpp
H27CK: Yes, currently it makes no real difference which is used. That you checked twice was sort of the aim. My logic: if, in the future, the sizes of `send_buffer_off_grid_spike_data_` and `send_buffer_spike_data_` are no longer equal, this line will serve to prompt the developer that some attention is needed. It's a bit ugly, but I thought it added some value.

H27CK:
> @jasperalbers Thanks for all the explanations, that makes sense. So essentially there is no scaling beyond 35 nodes in either case for this model. Interesting that we exhaust the threads, L3 collisions seem a plausible explanation. Can one use hardware counters to measure L3 misses (not in this PR!)?
> 
> So then the next question is why shrinking the buffers (by how much in total?), has such a massive effect on the time spent on MPI communication?

To add something to this: in the case where current master is using a large number of threads and fewer MPI processes (as NEST seems to prefer), e.g. 2 MPI processes/node and 32 threads/process, then each MPI process is responsible for the communication of a greater number of spikes (associated with a larger spike buffer). When larger spike buffers need to be expanded, without shrinkage (as in master), we are likely to communicate more padding over the course of the simulation.

nestkernel/mpi_manager.h
heplesser: @H27CK Could you add this in compact for as a comment in the code, so that we still understand the factor 4 when looking at this code three years from now?

jougs:
@jasperalbers: I think it would be nice if you could attach some plots to this PR, if only for future reference.

Otherwise I get a feeling this is handled well without me and removed my request for review :smile:

H27CK:
I think this is ready to go.

heplesser (approved):

nestkernel/event_delivery_manager.cpp
```diff
@@ -312,6 +317,9 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
 
   const AssignedRanks assigned_ranks = kernel().vp_manager.get_assigned_ranks( tid );
 
+  // Assume a single gather round
+  decrease_buffer_size_spike_data_ = true;
```
jakobj, line 321: sorry to be so late to the party, but would it be possible to turn this member variable into a variable local to this function?

jakobj (approved):
i've added one more question which would make the code a bit easier to understand.

otherwise i'm happy, performance looks great and travis like the changes! :+1:

jougs:
Given that @H27CK is no longer working with us, I suggest to merge this as is and file new PRs for subsequent improvements. @jakobj would you be OK with this?

jougs:
Woohoo! Another section of my PhD thesis invalidated ;-)

PR #1745, opened 2020-09-03
Delete iaf_psc_exp_multisynapse.cpp.orig

jarsi:
This file seems to be a relic from a merge.


hakonsbm (approved):
Nice catch!

hakonsbm:
Merging with a single review as the PR only removes a leftover file from a merge.

PR #1749, opened 2020-09-12
Fix duplicate "models" entry and remove dead "topology" entry from documentation TOC

jd41:
What it says in the title. Currently, the "models" link in the TOC appears twice in the RTD build of "latest". The TOC Tree also contained a dead link to some topology file, which I removed.

### DIFF
--- a/doc/contents.rst
+++ b/doc/contents.rst
@@ -15,8 +15,8 @@ Table of Contents
    getting_started
    tutorials/index
    guides/index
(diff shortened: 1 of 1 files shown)


jd41:
I also exchanged the position of "examples" and "models", now the two "reference" sections (models, pynest_apis) are closer together.

sarakonradi (approved):
Looks good to me!

PR #1750, opened 2020-09-12
Remove SLI neuron

jougs:
This PR removes `sli_neuron` to further reduce the dependencies between the NEST simulation kernel and the SLI interpreter. From its inception, the model was more or less just an academic exercise and to the best of my knowledge it was never really used by anyone, so removing it should also not hurt anyone.

I can add a deprecation warning to the 2.20.1 milestone, if the reviewers think this is necessary. I don't.


stinebuu (approved):
Nice!

I don't think a deprecation warning is necessary.

diesmann:
The purpose of this code was to explore how an environment and domain specific language needs to look like for rapid prototyping of neuron models. Now after some years of research and different attempts know that our answer is NESTML. In addition, independent of whether the interpreter is SLI or Python such code completely serializes the simulation. The solution to this is code generation as provided by the NESTML framework. For these reasons we can retire the code now. The NESTML documentation and probably also a formal paper should refer back to SLI neuron to document and explain the history.

diesmann (approved):

jougs:
Many thanks for the reviews.

I have now fired up our old development VM with Trac and SVN and saw that there were indeed some additional supporting changes to the kernel and interpreter at the time when the model was added. I'll scan them and revert as much of those in follow-up commits.

PR #1751, opened 2020-09-13
Fix math rendering error in documentation

clinssen:
Remove an unnecessary alignment character (&) from a math block, allowing MathJax to render it properly.

Follow-up to #1679.


jougs (approved):
Thanks! Merging without second reviewer as this is really not all that big ;-)

PR #1748, opened 2020-09-11
Dissolve Topology module and move files to nestkernel/spatial

jougs:

PR #1756, opened 2020-09-14
Rename "topology" to "spatial"

hakonsbm:
As a follow-up to #1748, this PR renames occurrences of `topology` to `spatial`, `spatial structure`, or similar, and attempts to update related documentation.

### DIFF
--- a/doc/features.rst
+++ b/doc/features.rst
@@ -11,6 +11,7 @@ Features
     spike-timing dependent plasticity (STDP).
 
 3.  NEST provides the possibility to create spatially-structured networks.
+    :doc:`Guide to spatially-structured networks<guides/spatial/guide_spatially_structured_networks>`
 
 4.  NEST provides many examples that help you getting started with your
     own simulation project.
--- a/doc/features.rst
+++ b/doc/features.rst
@@ -10,9 +10,7 @@ Features
     plasticity (Tsodyks & Markram) and different variants of
     spike-timing dependent plasticity (STDP).
 
-3.  Nest provides a Topology Module for creating complex networks
-    (`Topology Module User Manual
-    (PDF) <https://www.nest-simulator.org/wp-content/uploads/2014/12/NESTTopologyUserManual.pdf>`__)
+3.  NEST provides the possibility to create spatially-structured networks.
 
 4.  NEST provides many examples that help you getting started with your
     own simulation project.
@@ -86,12 +84,12 @@ Synapse models
    2000 <http://neuro.cjb.net/cgi/content/abstract/20/1/RC50>`__)
 -  Neuromodulatory synapses using dopamine
 
-Network models
---------------
+Spatially structured networks
+-----------------------------
 
--  Topology Module for creating complex networks (`Topology Module User
-   Manual
-   (PDF) <https://www.nest-simulator.org/wp-content/uploads/2014/12/NESTTopologyUserManual.pdf>`__)
+-  Node positions in 2D or 3D space
+-  Connection-rules and parameters based on node positions
+-  :doc:`Guide to spatially-structured networks<guides/spatial/guide_spatially_structured_networks>`
 
 Interoperability
 ----------------
--- a/doc/features.rst
+++ b/doc/features.rst
@@ -11,7 +11,7 @@ Features
     spike-timing dependent plasticity (STDP).
 
 3.  NEST provides the possibility to create spatially-structured networks.
-    :doc:`Guide to spatially-structured networks<guides/spatial/guide_spatially_structured_networks>`
+    (:doc:`Guide to spatially-structured networks<guides/spatial/guide_spatially_structured_networks>`)
 
 4.  NEST provides many examples that help you getting started with your
     own simulation project.
--- a/nestkernel/nestmodule.cpp
+++ b/nestkernel/nestmodule.cpp
@@ -2159,10 +2159,10 @@ NestModule::GetPosition_gFunction::execute( SLIInterpreter* i ) const
             layer from_pos to_node_id Displacement -> [double vector]
 
   Parameters:
-  layer       - NodeCollection for layer
-  from_node_id    - int, node_id of node in a spatial layer
-  from_pos    - double vector, position in layer
-  to_node_id      - int, node_id of node in a spatial layer
+  layer           - NodeCollection for layer
+  from_node_id    - int, node_id of node in a spatial NodeCollection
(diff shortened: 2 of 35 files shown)


clinssen (changes requested):
Thanks for the changes!

Just to check: a ``grep -ri topology`` still brings up some topology mentions that can be removed, such as in ``.gitignore``, in ``doc/tutorials/pynest_tutorial/part_4_spatially_structured_networks.rst`` ("See the table of *Topology-specific NEST Parameters* in NUTM for a selection of NEST" — what is "NUTM", does it stand for NEST-something-toplogy-module?) and ``lib/sli/nest-init.sli``.

nestkernel/nestmodule.cpp
```diff
@@ -2065,21 +2065,19 @@ NestModule::Apply_P_gFunction::execute( SLIInterpreter* i ) const
 //
 
 /** @BeginDocumentation
-  Name: topology::CreateLayer - create a spatial layer of nodes
+  Name: nest::CreateLayer - create nodes with spatial information
```
stinebuu, line 2068:
```suggestion
  Name: nest::CreateLayer - create nodes with spatial properties
```

nestkernel/nestmodule.cpp
```diff
@@ -2155,24 +2153,24 @@ NestModule::GetPosition_gFunction::execute( SLIInterpreter* i ) const
...
+  Name: nest::Displacement - compute displacement vector
 
   Synopsis: layer from_node_id to_node_id Displacement -> [double vector]
             layer from_pos to_node_id Displacement -> [double vector]
 
   Parameters:
   layer       - NodeCollection for layer
-  from_node_id    - int, node_id of node in a topology layer
+  from_node_id    - int, node_id of node in a spatial layer
   from_pos    - double vector, position in layer
```
stinebuu, line 2164: I know this doesn't have anything to do with this PR, but could you add some space so all the `-` are aligned?

nestkernel/nestmodule.cpp
```diff
@@ -2155,24 +2153,24 @@ NestModule::GetPosition_gFunction::execute( SLIInterpreter* i ) const
...
-  Name: topology::Displacement - compute displacement vector
+  Name: nest::Displacement - compute displacement vector
 
   Synopsis: layer from_node_id to_node_id Displacement -> [double vector]
             layer from_pos to_node_id Displacement -> [double vector]
 
   Parameters:
   layer       - NodeCollection for layer
-  from_node_id    - int, node_id of node in a topology layer
+  from_node_id    - int, node_id of node in a spatial layer
```
stinebuu, line 2163: Maybe use 'NodeColletion' or 'network' here instead of layer? I think layer made sense when we used 'topology', but now that we change to 'spatial' I think it makes more sense to change the 'layer' as well. Also further down in this file.

nestkernel/nestmodule.cpp
```diff
@@ -2334,11 +2332,11 @@ NestModule::Distance_aFunction::execute( SLIInterpreter* i ) const
   /type - mask type
   dict  - dictionary with mask specifications
 
-  Description: Masks are used when creating connections in the Topology
-  module. A mask describes which area of the pool layer shall be searched
-  for nodes to connect for any given node in the driver layer. This
-  command creates a mask object which may be combined with other mask
-  objects using Boolean operators. The mask is specified in a dictionary.
+  Description: Masks can be used when creating connections between nodes
+  with spatial information. A mask describes which area of the pool layer
```
stinebuu, line 2336: Maybe `spatial distributions` or `spatial parameters`?

nestkernel/spatial/spatial.h
```diff
@@ -46,7 +46,7 @@ namespace nest
 {
 
 /**
- * Class representing metadata for topology layer.
+ * Class representing metadata for layer containing spatial information.
```
stinebuu, line 49: I would use NodeCollection instead of layer here.

testsuite/regressiontests/ticket-433.sli
```diff
@@ -51,9 +51,9 @@ M_INFO setverbosity
 %  ampa --- AMPA receptor ID
 %  retina, retina_gen, Tp --- layers
 
-% Connection function using topology library
+% Connection function using spatial connection function
```
stinebuu, line 54:
```suggestion
% Connection function using spatial parameters
```

testsuite/regressiontests/ticket-433.sli
```diff
@@ -66,9 +66,9 @@ M_INFO setverbosity
 }
 bind def
 
-% Connection function using topology library
+% Connection function using spatial connection function
```
stinebuu, line 69:
```suggestion
% Connection function using spatial parameters
```

stinebuu (changes requested):
Thanks! I suggest changing some of the 'spatial information' to either 'spatial distributions' or 'spatial parameters', but you can have a look and see what you think.

hakonsbm:
@stinebuu Thanks for the review, I've applied your suggestions.

@clinssen Not sure how I missed the ones in `.gitignore`, but I've fixed them, and the "NUTM"s (I think it was supposed to be "NTUM" for "NEST Topology User Manual"). The ones in `nest-init.sli` are in the function `RestoreModels`, which is never tested and not used in any examples, like the `SaveModels` function above it. I'm wondering if we should just remove these functions altogether. I can bring it up at the next plenary session of the hackathon.

doc/features.rst
```diff
@@ -10,9 +10,8 @@ Features
     plasticity (Tsodyks & Markram) and different variants of
     spike-timing dependent plasticity (STDP).
 
-3.  Nest provides a Topology Module for creating complex networks
-    (`Topology Module User Manual
-    (PDF) <https://www.nest-simulator.org/wp-content/uploads/2014/12/NESTTopologyUserManual.pdf>`__)
+3.  NEST provides the possibility to create spatially-structured networks.
+    :doc:`Guide to spatially-structured networks<guides/spatial/guide_spatially_structured_networks>`
```
stinebuu, line 14: It seems like the other references on this page are in parentheses.
```suggestion
    (:doc:`Guide to spatially-structured networks<guides/spatial/guide_spatially_structured_networks>`)
```

stinebuu (commented):
I only have a tiny comment now

stinebuu (approved):
# :+1:

clinssen (approved):
Great, thanks!

PR #1760, opened 2020-09-16
Add test of connecting arrays using MPI

### LINKED ISSUE #1710: Connect with NumPy arrays of node IDs does not work with MPI
#1429 added the possibility to use `Connect` with NumPy arrays of node IDs. However it doesn't work when using multiple processes with MPI. In addition to making that work, a unittest of `Connect` with NumPy arrays of node IDs should be added.

hakonsbm:
Apparently connecting arrays of node IDs using MPI works after all. A test is added, which fixes #1710.

### DIFF
--- a/pynest/nest/tests/test_connect_arrays_mpi.py
+++ b/pynest/nest/tests/test_connect_arrays_mpi.py
@@ -1,0 +1,139 @@
+# -*- coding: utf-8 -*-
+#
+# test_connect_arrays_mpi.py
+#
+# This file is part of NEST.
+#
+# Copyright (C) 2004 The NEST Initiative
+#
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import subprocess as sp
+import unittest
+import nest
+import numpy as np
+
+try:
+    from mpi4py import MPI
+    HAVE_MPI4PY = True
+except ImportError:
+    HAVE_MPI4PY = False
+
+HAVE_MPI = nest.ll_api.sli_func("statusdict/have_mpi ::")
+MULTIPLE_PROCESSES = nest.NumProcesses() > 1
+
+
+class ConnectArraysMPICase(unittest.TestCase):
+    non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
+    comm = MPI.COMM_WORLD
+
+    # With nosetests, only run these tests if using multiple processes
+    __test__ = MULTIPLE_PROCESSES
+
+    def assert_connections(self, expected_sources, expected_targets, rule='one_to_one'):
+        """Gather connections from all processes and assert against expected connections"""
+        conns = nest.GetConnections()
+        projections = [[s, t] for s, t in zip(conns.source, conns.target)]
+        if rule == 'one_to_one':
+            expected_projections = np.array([[s, t] for s, t in zip(expected_sources, expected_targets)])
+        elif rule == 'all_to_all':
+            expected_projections = np.array([[s, t] for s in expected_sources for t in expected_targets])
+        else:
+            self.assertFalse(True, 'rule={} is not valid'.format(rule))
+
+        recv_projections = self.comm.gather(projections, root=0)
+        if self.comm.Get_rank() == 0:
+            # Flatten the projection lists to a single list of projections
+            recv_projections = np.array([proj for part in recv_projections for proj in part])
+            # Results must be sorted to make comparison possible
+            np.testing.assert_array_equal(np.sort(recv_projections, axis=0), np.sort(expected_projections, axis=0))
+        else:
+            self.assertIsNone(recv_projections)
+
+    def setUp(self):
+        nest.ResetKernel()
+
+    def test_connect_arrays_unique(self):
+        """Connecting NumPy arrays of unique node IDs with MPI"""
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = np.arange(1, n+1, dtype=np.uint64)
+        targets = np.arange(1, n+1, dtype=np.uint64)
+        weights = 1.5
+        delays = 1.4
+
+        nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays})
+
+        self.assert_connections(sources, targets, rule='all_to_all')
+
+    def test_connect_arrays_nonunique(self):
+        """Connecting NumPy arrays with non-unique node IDs with MPI"""
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = np.arange(1, n+1, dtype=np.uint64)
+        targets = self.non_unique
+        weights = np.ones(n)
+        delays = np.ones(n)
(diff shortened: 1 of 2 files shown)


pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -0,0 +1,142 @@
...
+import nest
+import numpy as np
+
+try:
+    from mpi4py import MPI
+    HAVE_MPI4PY = True
+except ImportError:
+    HAVE_MPI4PY = False
+
+HAVE_MPI = nest.ll_api.sli_func("statusdict/have_mpi ::")
```
Silmathoron, line 34: @jougs this is unrelated to the PR, but I find this is a rather convoluted way of getting this information... shouldn't it be available to the Python level more directly? Is there a good reason for that? (If not I'd raise an issue to discuss it)

pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -0,0 +1,142 @@
...
+    processes, and is supposed to only be run with multiple processes.
+    If running with nosetests or pytest, this TestCase is ignored when
+    run with a single process, and called from TestConnectArraysMPI using
+    multiple processes.
+    """
+    non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
+    comm = MPI.COMM_WORLD
+
+    # With nosetests, only run these tests if using multiple processes
+    __test__ = MULTIPLE_PROCESSES
```
Silmathoron, line 52: one can have ``MULTIPLE_PROCESSES`` False but ``HAVE_MPI`` True, would there then be some cases where the test would be run with a single process? (e.g. using pytest: I'm not sure why only nosetests is mentioned...)

More generally, one cannot have ``MULTIPLE_PROCESSES`` if ``HAVE_MPI`` is False so I'm not sure why ``HAVE_MPI`` is required.

pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -0,0 +1,142 @@
...
+        nest.ResetKernel()
+
+    def test_connect_arrays_unique(self):
+        """Connecting NumPy arrays of unique node IDs with MPI"""
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = np.arange(1, n+1, dtype=np.uint64)
+        targets = np.arange(1, n+1, dtype=np.uint64)
+        weights = 1.5
+        delays = 1.4
```
Silmathoron, line 84: why provide them if they are not checked?

pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -0,0 +1,142 @@
...
+        n = 10
+        nest.Create('iaf_psc_alpha', n)
+        sources = np.arange(1, n+1, dtype=np.uint64)
+        targets = self.non_unique
+        weights = np.ones(n)
+        delays = np.ones(n)
+        nest.Connect(sources, targets, syn_spec={'weight': weights, 'delay': delays},
+                     conn_spec='one_to_one')
+
+        self.assert_connections(sources, targets)
```
Silmathoron, line 101: the fact that the default value for "rule" in ``Connect`` is "all_to_all" but it is "one_to_one" for ``assert_connections`` seems like pure evil x)
I would make a variable for the rule every time and always specify it explicitly (also removing the default value in the function argument)

pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -0,0 +1,142 @@
...
+        directory = os.path.dirname(os.path.realpath(__file__))
+        script = os.path.realpath(__file__)
+        test_script = os.path.join(directory, script)
+        command = nest.ll_api.sli_func("mpirun", 2, "nosetests", test_script)
+        command = command.split()
+
+        my_env = os.environ.copy()
+        retcode = sp.call(command, env=my_env)
+
+        self.assertEqual(retcode, 0, 'Test failed when run with "mpirun -np 2 nosetests [script]"')
```
Silmathoron, line 138: Wow, that's something xD
Could you please add a class docstring explaining that this is just a workaround?
Otherwise I'm sure it will puzzle quite a few people.

Silmathoron (changes requested):
Thanks for the tests!
I'm made a few comments about rather minor changes to make the code a little more explicit.

pynest/nest/tests/test_connect_arrays_mpi.py
hakonsbm: I've added `pytest` to the comment and removed the `skipIf(HAVE_MPI)`.

pynest/nest/tests/test_connect_arrays_mpi.py
hakonsbm: True, I've removed them.

pynest/nest/tests/test_connect_arrays_mpi.py
hakonsbm: Done!

pynest/nest/tests/test_connect_arrays_mpi.py
hakonsbm: It's not really a workaround. I've added some docstrings to the class and at the top of the file to try to explain how it works.

Silmathoron: Maybe "workaround" is not the right term, but it is rather hacky so the docstring definitely helps ;)

pynest/nest/tests/test_connect_arrays_mpi.py
Silmathoron: Mhh... maybe I should have phrased this differently, sorry ^^"
I meant: shouldn't there be a check to make sure that synaptic properties are handled correctly in this case? (maybe even with array properties)
Or is this unnecessary (maybe the underlying code reuses the same functions as the normal Connect)?

hakonsbm: Ah, right. While implementing testing of weights and delays I found a bug in `connect_arrays()` that should be fixed first. See #1781.

hakonsbm: Now that #1781 is merged, I've added tests of weights and delays.

Silmathoron (approved):
Thanks, I'm happy as soon as Travis is :+1:

pynest/nest/tests/test_connect_arrays_mpi.py
jougs: I agree that this is a bit indirect and am open for further discussion if you opened an issue and mentioned me there.

pynest/nest/tests/test_connect_arrays_mpi.py
jougs: I agree that the comment is helpful. That said, I have a working prototype for a refactoring of PyNEST'S MPI tests which makes individual tests easier and moves a lot of the dirty code into the test runner. I'll start a PR once I find the time to clean it up.

pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -0,0 +1,175 @@
...
+class ConnectArraysMPICase(unittest.TestCase):
+    """
+    This TestCase uses mpi4py to collect and assert results from all
+    processes, and is supposed to only be run with multiple processes.
+    If running with nosetests or pytest, this TestCase is ignored when
+    run with a single process, and called from TestConnectArraysMPI using
+    multiple processes.
+    """
+    non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
+    comm = MPI.COMM_WORLD
```
jougs, line 60: This could potentially leave to problems, as NEST internally uses the same communicator. I suggest to use something like this instead:
```suggestion
    comm = MPI.COMM_WORLD.Clone()
```

jougs (changes requested):
Other than my small suggestion, I'm fine. Thanks for the test and for removing the restriction from the `Connect()` function.

jougs (approved):
Thanks! Happy and merging.

PR #1761, opened 2020-09-16
Update edgecolor to default value in examples and code that use scatterplot

stinebuu:
Use `None` instead of `'none'` for `edgecolor` in scatterplots in examples and plotting functions.

Using `'none'` led to an error message for me, I have Python 3.7.6, but I have tested the proposed changes with Python 3.5 and Python 3.6 as well.

UPDATE after comments: Have removed `edgecolor` as `None` is default, and there is the problem with `'none'`.

### DIFF
--- a/pynest/examples/spatial/test_3d.py
+++ b/pynest/examples/spatial/test_3d.py
@@ -42,7 +42,7 @@ l1 = nest.Create('iaf_psc_alpha', 1000, positions=pos)
 xpos, ypos, zpos = zip(*nest.GetPosition(l1))
 fig = plt.figure()
 ax = fig.add_subplot(111, projection='3d')
-ax.scatter(xpos, ypos, zpos, s=15, facecolor='b', edgecolor='none')
+ax.scatter(xpos, ypos, zpos, s=15, facecolor='b', edgecolor=None)
 
 # full connections in volume [-0.2,0.2]**3
 nest.Connect(l1, l1,
@@ -57,7 +57,7 @@ nest.Connect(l1, l1,
 ctr = nest.FindCenterElement(l1)
 xtgt, ytgt, ztgt = zip(*nest.GetTargetPositions(ctr, l1)[0])
 xctr, yctr, zctr = nest.GetPosition(ctr)
-ax.scatter([xctr], [yctr], [zctr], s=40, facecolor='r', edgecolor='none')
+ax.scatter([xctr], [yctr], [zctr], s=40, facecolor='r', edgecolor=None)
 ax.scatter(xtgt, ytgt, ztgt, s=40, facecolor='g', edgecolor='g')
 
 tgts = nest.GetTargetNodes(ctr, l1)[0]
--- a/pynest/examples/spatial/test_3d.py
+++ b/pynest/examples/spatial/test_3d.py
@@ -42,7 +42,7 @@ l1 = nest.Create('iaf_psc_alpha', 1000, positions=pos)
 xpos, ypos, zpos = zip(*nest.GetPosition(l1))
 fig = plt.figure()
 ax = fig.add_subplot(111, projection='3d')
-ax.scatter(xpos, ypos, zpos, s=15, facecolor='b', edgecolor=None)
+ax.scatter(xpos, ypos, zpos, s=15, facecolor='b')
 
 # full connections in volume [-0.2,0.2]**3
 nest.Connect(l1, l1,
@@ -57,7 +57,7 @@ nest.Connect(l1, l1,
 ctr = nest.FindCenterElement(l1)
 xtgt, ytgt, ztgt = zip(*nest.GetTargetPositions(ctr, l1)[0])
 xctr, yctr, zctr = nest.GetPosition(ctr)
-ax.scatter([xctr], [yctr], [zctr], s=40, facecolor='r', edgecolor=None)
+ax.scatter([xctr], [yctr], [zctr], s=40, facecolor='r')
 ax.scatter(xtgt, ytgt, ztgt, s=40, facecolor='g', edgecolor='g')
 
 tgts = nest.GetTargetNodes(ctr, l1)[0]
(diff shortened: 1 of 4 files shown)


hakonsbm (changes requested):
According to the [documentation](https://matplotlib.org/3.3.1/api/_as_gen/matplotlib.pyplot.scatter.html), the argument should be `edgecolors`. However, it is `None` by default, so maybe they can just be removed?

heplesser:
@hakonsbm I found the following in the documentation you linked to:

```
edgecolors{'face', 'none', None} or color or sequence of color, default: rcParams["scatter.edgecolors"] (default: 'face')
The edge color of the marker. Possible values:

- 'face': The edge color will always be the same as the face color.
- 'none': No patch boundary will be drawn.
- A color or sequence of colors.

For non-filled markers, the edgecolors kwarg is ignored and forced to 'face' internally.
```

The third alternative in the bullet list is probably for `None`. So in this case, `'none'` and `None` are definitely something different here. Similar logic applies, I think, to a number of Matplotlib functions.

clinssen:
Agree with HEP, the semantics of ``None`` and ``'none'`` are different. The latter works fine for me on the latest matplotlib release (3.3.2). What was the error message that you were getting, @stinebuu?

stinebuu:
@clinssen When running for instance [test_3d.py](https://github.com/nest/nest-simulator/blob/master/pynest/examples/spatial/test_3d.py) I got the following message:
```
  File "/home/stine/anaconda3/lib/python3.7/site-packages/matplotlib/backends/backend_qt5.py", line 508, in _draw_idle
    self.draw()
  File "/home/stine/anaconda3/lib/python3.7/site-packages/matplotlib/backends/backend_agg.py", line 388, in draw
    self.figure.draw(self.renderer)
  File "/home/stine/anaconda3/lib/python3.7/site-packages/matplotlib/artist.py", line 38, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/stine/anaconda3/lib/python3.7/site-packages/matplotlib/figure.py", line 1709, in draw
    renderer, self, artists, self.suppressComposite)
  File "/home/stine/anaconda3/lib/python3.7/site-packages/matplotlib/image.py", line 135, in _draw_list_compositing_images
    a.draw(renderer)
  File "/home/stine/anaconda3/lib/python3.7/site-packages/matplotlib/artist.py", line 38, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/stine/anaconda3/lib/python3.7/site-packages/mpl_toolkits/mplot3d/axes3d.py", line 292, in draw
    reverse=True)):
  File "/home/stine/anaconda3/lib/python3.7/site-packages/mpl_toolkits/mplot3d/axes3d.py", line 291, in <lambda>
    key=lambda col: col.do_3d_projection(renderer),
  File "/home/stine/anaconda3/lib/python3.7/site-packages/mpl_toolkits/mplot3d/art3d.py", line 545, in do_3d_projection
    ecs = (_zalpha(self._edgecolor3d, vzs) if self._depthshade else
  File "/home/stine/anaconda3/lib/python3.7/site-packages/mpl_toolkits/mplot3d/art3d.py", line 847, in _zalpha
    rgba = np.broadcast_to(mcolors.to_rgba_array(colors), (len(zs), 4))
  File "<__array_function__ internals>", line 6, in broadcast_to
  File "/home/stine/anaconda3/lib/python3.7/site-packages/numpy/lib/stride_tricks.py", line 182, in broadcast_to
    return _broadcast_to(array, shape, subok=subok, readonly=True)
  File "/home/stine/anaconda3/lib/python3.7/site-packages/numpy/lib/stride_tricks.py", line 127, in _broadcast_to
    op_flags=['readonly'], itershape=shape, order='C')
ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (0,4) and requested shape (1000,4)
```
It didn't give me much information, but when changing the `edgecolor` the error disappear and I am able to plot.

I guess I did not investigate the difference between `'none'` and `None`, but currently the example does not plot the scatter plot at all for me. If it is only on my computer I can of course close this PR and do the changes locally whenever I plot ;)

clinssen:
I can reproduce the error. Looks like an upstream (matplotlib) bug introduced between 2.2.3 and 3.0. We could add ==2.2.3 to requirements.txt as an interim fix.

Silmathoron:
@clinssen I'm not sure we want to require [matplotlib 2.2.3](https://github.com/matplotlib/matplotlib/releases/tag/v2.2.3), it is ancient...

stinebuu:
@clinssen I think we agreed on supporting matplotlib 3.0 with NEST 3.0, but on the quick search I did on nest 2 to 3 I cannot find it. I agree with @Silmathoron that matplotlib 2.2.3 is quite old.

It seems like there are some disagreements about my proposed changes, should I convert this PR to an issue?

hakonsbm:
As the current examples that are using `edgecolor='none'` are creating confusing output and an error, and it's not an option to force use of ancient versions of matplotlib, I think there are two ways we can deal with this:

1. Comment out the calls to `scatter()` that cause errors, with a comment about the bug in matplotlib.
2. Remove `edgecolor='none'` from the calls to `scatter()` to make plotting work again.

I think the second option is better, as that would make the examples work with minimal changes to the resulting output.

@stinebuu @heplesser @clinssen Do you have any preferences?

clinssen:
I can settle for (2) for now!

hakonsbm (approved):
Looks good to me, thanks!

clinssen (approved):
I am not 100% happy with this as we don't know what this does to these plots: will there e.g. now be a default black outline that obscures the blue/red face colour at a certain scale? Could we create an issue requesting a reversal after matplotlib fixes this issue upstream?

Approving because working is better than broken, thanks for the fix @stinebuu in any case!

stinebuu:
@clinssen Thanks for the approval! Currently, there is a blue line around the blue and red around the red, unless we have specified another color:

image

We don't know how this might change in the future of course.
We could make an issue if you wish :+1:

clinssen:
OK, that looks reasonable, let's skip creating an issue then. Thanks for checking!

PR #1764, opened 2020-09-17
Update MacOS caching in Travis CI

clinssen:


lekshmideepu:
https://travis-ci.org/github/lekshmideepu/nest-simulator/jobs/727993387
https://travis-ci.org/github/lekshmideepu/nest-simulator/jobs/727989757
https://travis-ci.org/github/lekshmideepu/nest-simulator/jobs/727986583

clinssen:
Another idea for CI improvement (per the hackathon 2020-9-17): run the full build(s) first; failure is most likely to occur there, so this implies that if failure is inevitable, it will at least occur earlier and waste fewer compute resources.

lekshmideepu (commented):
I don't think this code change saves time as you just moved the cache code block to the cache directory section. Did this actually saved the time?

clinssen:
It doesn't look like it. It was confusing as there seemed to be a very large variability (30 to 50 min) in the exact same environment and configuration, between builds. I have left this PR up in case there was any code that we could reuse, but probably it can be closed as it doesn't offer any obvious advantage.

PR #1765, opened 2020-09-17
Remove obsolete volume mask

### LINKED ISSUE #1759: Remove obsolite volume mask
For 3D layers we have two types of box-masks, `box` and `volume`. It looks like the volume mask is only there for backward compatibility.

https://github.com/nest/nest-simulator/blob/a902eaccc61f76b718d064b62f6770a5619de58a/nestkernel/nestmodule.cpp#L3016-L3017

The volume mask is not documented, we only document the box mask, but it is used in 3 spatial examples, [test_3d.py](https://github.com/nest/nest-simulator/blob/master/pynest/examples/spatial/test_3d.py), [test_3d_exp.py](https://github.com/nest/nest-simulator/blob/master/pynest/examples/spatial/test_3d_exp.py) and [test_3d_gauss.py](https://github.com/nest/nest-simulator/blob/master/pynest/examples/spatial/test_3d_gauss.py)

I do not think we need both masks, as they do exactly the same thing, and suggest we remove the volume box.

stinebuu:
This PR fixes #1759.

### DIFF
--- a/doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
+++ b/doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
@@ -1296,6 +1296,11 @@ probability and delay, and random weights from a normal distribution:
   |                                                                  |                                                                     |
   +------------------------------------------------------------------+---------------------------------------------------------------------+
 
+Masks
+^^^^^
+The ``volume`` mask is removed with NEST 3.0, you can use ``box`` mask instead. The ``volume`` mask was only in NEST
+for backward compatibility, in fact the two masks are identical, except for the name.
+
 Retrieving distance information
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 If you have a SynapseCollection with connections from a spatially distributed network, you can retrieve the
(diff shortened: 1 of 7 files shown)


hakonsbm (approved):
Looks good to me!

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
```diff
@@ -1296,6 +1296,11 @@ probability and delay, and random weights from a normal distribution:
   |                                                                  |                                                                     |
   +------------------------------------------------------------------+---------------------------------------------------------------------+
 
+Masks
+^^^^^
+The ``volume`` mask is removed with NEST 3.0, you can use ``box`` mask instead. The ``volume`` mask was only in NEST
+for backward compatibility, in fact the two masks are identical, except for the name.
```
jougs, line 1302:
```suggestion
In NEST 3.0, the mask ``volume`` got removed, as the same mask was already available under the name ``box``.
The former was only available in NEST for backward compatibility.
```

jougs (approved):
I'm approving, on condition that you consider my suggestion for the wording in the transition guide. Feel free to merge by yourself once that's resolved.

doc/guides/nest2_to_nest3/nest2_to_nest3_overview.rst
terhorstd: LGTM

PR #1767, opened 2020-09-18
Use ring buffer that allows for multiple values per slot

suku248:
Using the new `MultiValueRingBuffer` increases cache performance of access to input buffers during neuronal update and spike delivery. Instead of using one ring buffer per input channel (e.g. exc. synapses, inh. synapses, input current), only a single ring buffer is used that stores the inputs per time slot for all input channels bunched into an array.

This PR only implements the usage of the `MultiValueRingBuffer` for the neuron models `iaf_psc_exp` and `iaf_psc_alpha`.

### DIFF
--- a/models/iaf_psc_alpha.cpp
+++ b/models/iaf_psc_alpha.cpp
@@ -245,9 +245,7 @@ iaf_psc_alpha::init_state_( const Node& proto )
 void
 iaf_psc_alpha::init_buffers_()
 {
-  B_.ex_spikes_.clear(); // includes resize
-  B_.in_spikes_.clear(); // includes resize
-  B_.currents_.clear();  // includes resize
+  B_.input_buffer_.clear(); // includes resize
 
   B_.logger_.reset();
 
@@ -257,6 +255,8 @@ iaf_psc_alpha::init_buffers_()
 void
 iaf_psc_alpha::calibrate()
 {
+  B_.input_buffer_.resize();
+
   // ensures initialization in case mm connected after Simulate
   B_.logger_.init();
 
@@ -321,6 +321,10 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
 
   for ( long lag = from; lag < to; ++lag )
   {
+    // get access to the correct input-buffer entry at the very beginning
+    const index input_buffer_idx = kernel().event_delivery_manager.get_modulo( lag );
+    std::array< double, Buffers_::NUM_INPUT_CHANNELS >& new_input = B_.input_buffer_.get_values( input_buffer_idx );
+
     if ( S_.r_ == 0 )
     {
       // neuron not refractory
@@ -342,7 +346,7 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
 
     // Apply spikes delivered in this step; spikes arriving at T+1 have
     // an immediate effect on the state of the neuron
-    V_.weighted_spikes_ex_ = B_.ex_spikes_.get_value( lag );
+    V_.weighted_spikes_ex_ = new_input[ Buffers_::SYN_EX ];
     S_.dI_ex_ += V_.EPSCInitialValue_ * V_.weighted_spikes_ex_;
 
     // alpha shape EPSCs
@@ -351,7 +355,7 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
 
     // Apply spikes delivered in this step; spikes arriving at T+1 have
     // an immediate effect on the state of the neuron
-    V_.weighted_spikes_in_ = B_.in_spikes_.get_value( lag );
+    V_.weighted_spikes_in_ = new_input[ Buffers_::SYN_IN ];
     S_.dI_in_ += V_.IPSCInitialValue_ * V_.weighted_spikes_in_;
 
     // threshold crossing
@@ -370,7 +374,15 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
     }
 
     // set new input current
-    S_.y0_ = B_.currents_.get_value( lag );
+    S_.y0_ = new_input[ Buffers_::I0 ];
+
+    // reset input in ring buffer at position lag
+    // note: delegating the reset to a function in the MultiValueRingBuffer
+    // would entail additional costs
+    for ( auto it = new_input.begin(); it < new_input.end(); ++it )
+    {
+      ( *it ) = 0.0;
+    }
 
     // log state data
     B_.logger_.record_data( origin.get_steps() + lag );
@@ -382,15 +394,18 @@ iaf_psc_alpha::handle( SpikeEvent& e )
 {
   assert( e.get_delay_steps() > 0 );
 
+  const index input_buffer_idx = kernel().event_delivery_manager.get_modulo(
+    e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ) );
+
   const double s = e.get_weight() * e.get_multiplicity();
 
   if ( e.get_weight() > 0.0 )
   {
-    B_.ex_spikes_.add_value( e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ), s );
+    B_.input_buffer_.add_value( input_buffer_idx, Buffers_::SYN_EX, s );
   }
   else
   {
-    B_.in_spikes_.add_value( e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ), s );
+    B_.input_buffer_.add_value( input_buffer_idx, Buffers_::SYN_IN, s );
   }
 }
 
@@ -399,10 +414,13 @@ iaf_psc_alpha::handle( CurrentEvent& e )
 {
   assert( e.get_delay_steps() > 0 );
 
+  const index input_buffer_idx = kernel().event_delivery_manager.get_modulo(
+    e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ) );
+
   const double I = e.get_current();
   const double w = e.get_weight();
 
-  B_.currents_.add_value( e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ), w * I );
+  B_.input_buffer_.add_value( input_buffer_idx, Buffers_::I0, w * I );
 }
 
 void
--- a/models/iaf_psc_alpha.h
+++ b/models/iaf_psc_alpha.h
@@ -29,7 +29,7 @@
 #include "event.h"
 #include "nest_types.h"
 #include "recordables_map.h"
-#include "ring_buffer.h"
+#include "ring_buffer_impl.h"
 #include "universal_data_logger.h"
 
 namespace nest
@@ -266,10 +266,17 @@ private:
     Buffers_( iaf_psc_alpha& );
     Buffers_( const Buffers_&, iaf_psc_alpha& );
 
-    /** buffers and summs up incoming spikes/currents */
-    RingBuffer ex_spikes_;
-    RingBuffer in_spikes_;
-    RingBuffer currents_;
+    //! Idices for access to different channels of input_buffer_
+    enum
+    {
+      SYN_IN = 0,
+      SYN_EX,
+      I0,
+      NUM_INPUT_CHANNELS
+    };
+
+    /** buffers and sums up incoming spikes/currents */
+    MultiValueRingBuffer< NUM_INPUT_CHANNELS > input_buffer_;
 
     //! Logger for all analog data
     UniversalDataLogger< iaf_psc_alpha > logger_;
--- a/models/iaf_psc_exp.cpp
+++ b/models/iaf_psc_exp.cpp
@@ -243,9 +243,7 @@ nest::iaf_psc_exp::init_state_( const Node& proto )
 void
 nest::iaf_psc_exp::init_buffers_()
 {
-  B_.spikes_ex_.clear(); // includes resize
-  B_.spikes_in_.clear(); // includes resize
-  B_.currents_.clear();  // includes resize
+  B_.input_buffer_.clear(); // includes resize
   B_.logger_.reset();
   Archiving_Node::clear_history();
 }
@@ -253,7 +251,8 @@ nest::iaf_psc_exp::init_buffers_()
 void
 nest::iaf_psc_exp::calibrate()
 {
-  B_.currents_.resize( 2 );
+  B_.input_buffer_.resize();
+
   // ensures initialization in case mm connected after Simulate
   B_.logger_.init();
 
@@ -319,6 +318,9 @@ nest::iaf_psc_exp::update( const Time& origin, const long from, const long to )
   // evolve from timestep 'from' to timestep 'to' with steps of h each
   for ( long lag = from; lag < to; ++lag )
   {
+    const index idx = kernel().event_delivery_manager.get_modulo( lag );
+    std::array< double, 4 >& new_input = B_.input_buffer_.get_values( idx );
+
     if ( S_.r_ref_ == 0 ) // neuron not refractory, so evolve V
     {
       S_.V_m_ =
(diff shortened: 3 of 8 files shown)


jarsi:
A short update on this PR. I have carried out benchmarks of this PR with both the microcircuit and the hpc_benchmark.

The microcircuit profits from these changes as long as communication does not dominate simulation time. Once communication dominates, the simulation time remains unaffected.

The hpc_benchmark is more or less unaffected by these changes.

Combining these two observations, we can probably move from the draft stage to the review stage

models/iaf_psc_alpha.cpp
```diff
@@ -257,6 +255,8 @@ iaf_psc_alpha::init_buffers_()
 void
 iaf_psc_alpha::calibrate()
 {
+  B_.input_buffer_.resize();
```
heplesser, line 258: Why this `resize()` call here? If I am not entirely mistaken, buffer sizes should not change once buffers have been created, i.e., min/max delay cannot change after `Simulate` has been called.

models/iaf_psc_alpha.cpp
```diff
@@ -321,6 +321,10 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
 
   for ( long lag = from; lag < to; ++lag )
   {
+    // get access to the correct input-buffer entry at the very beginning
+    const index input_buffer_idx = kernel().event_delivery_manager.get_modulo( lag );
+    std::array< double, Buffers_::NUM_INPUT_CHANNELS >& new_input = B_.input_buffer_.get_values( input_buffer_idx );
```
heplesser, line 326: Could one replace the complicated type with `auto` here for readability?

heplesser: And is `new_input` the best possible name? How about just `input`?

models/iaf_psc_alpha.cpp
```diff
@@ -370,7 +374,15 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
     }
 
     // set new input current
-    S_.y0_ = B_.currents_.get_value( lag );
+    S_.y0_ = new_input[ Buffers_::I0 ];
+
+    // reset input in ring buffer at position lag
+    // note: delegating the reset to a function in the MultiValueRingBuffer
+    // would entail additional costs
+    for ( auto it = new_input.begin(); it < new_input.end(); ++it )
```
heplesser, line 382: If we made this an inline function, there should be no call overhead and we avoid code duplication.

models/iaf_psc_alpha.h
```diff
@@ -29,7 +29,7 @@
 #include "event.h"
 #include "nest_types.h"
 #include "recordables_map.h"
-#include "ring_buffer.h"
+#include "ring_buffer_impl.h"
```
heplesser, line 32: Why include the `impl` file here? The main point of `impl` files is to avoid circular inclusion problems and code bloat. So unless absolutely unavoidable, I would suggest to include just `ring_buffer.h` here and include `ring_buffer_impl.h` only in the `.cpp` file.

models/iaf_psc_alpha.h
```diff
@@ -266,10 +266,17 @@ class iaf_psc_alpha : public Archiving_Node
     Buffers_( iaf_psc_alpha& );
     Buffers_( const Buffers_&, iaf_psc_alpha& );
 
-    /** buffers and summs up incoming spikes/currents */
-    RingBuffer ex_spikes_;
-    RingBuffer in_spikes_;
-    RingBuffer currents_;
+    //! Idices for access to different channels of input_buffer_
```
heplesser, line 269:
```suggestion
    //! Indices for access to different channels of input_buffer_
```

nestkernel/ring_buffer.h
```diff
@@ -351,7 +352,30 @@ ListRingBuffer::get_index_( const delay d ) const
...
-}
+
+
+template < unsigned int num_values >
+class MultiValueRingBuffer
+{
+public:
+  MultiValueRingBuffer();
+
+  void add_value( const long idx0, const long idx1, const double val );
```
heplesser, line 363: `idx0` and `idx1` are little informative. How about `time_idx` and `channel_idx` or similar?

nestkernel/ring_buffer.h
```diff
@@ -351,7 +352,30 @@ ListRingBuffer::get_index_( const delay d ) const
...
+
+  std::array< double, num_values >& get_values( const long idx0 );
```
heplesser, line 365: Rename `idx0` in a way consistent with the above.

nestkernel/ring_buffer.h
```diff
@@ -351,7 +352,30 @@ ListRingBuffer::get_index_( const delay d ) const
   assert( ( size_t ) idx < buffer_.size() );
   return idx;
 }
-}
+
+
+template < unsigned int num_values >
```
heplesser, line 357: Should we for consistency use `num_channels` instead of `num_values`?

nestkernel/ring_buffer.h
```diff
@@ -351,7 +352,30 @@ ListRingBuffer::get_index_( const delay d ) const
...
+
+  void clear();
+
+  void resize();
+
+  size_t size() const;
+
+private:
+  //! Buffered data
+  std::vector< std::array< double, num_values > > buffer_;
```
heplesser, line 375: Please expand the docstring and explain outer and inner dimension.

nestkernel/ring_buffer_impl.h
```diff
@@ -0,0 +1,80 @@
...
+template < unsigned int num_values >
+void
+nest::MultiValueRingBuffer< num_values >::add_value( const long idx0, const long idx1, const double val )
+{
+  buffer_[ idx0 ][ idx1 ] += val;
+}
+
+template < unsigned int num_values >
+std::array< double, num_values >&
+nest::MultiValueRingBuffer< num_values >::get_values( const long idx0 )
```
heplesser, line 44: Why `long` here if we anyways need to cast to `size_t` below and negative values make no sense?

nestkernel/ring_buffer_impl.h
```diff
@@ -0,0 +1,80 @@
...
+{
+  assert( 0 <= idx0 and ( size_t ) idx0 < buffer_.size() );
```
heplesser, line 46:
```suggestion
  assert( 0 <= idx0 and  static_cast< size_t >( idx0 ) < buffer_.size() );
```

nestkernel/ring_buffer_impl.h
```diff
@@ -0,0 +1,80 @@
...
+  return buffer_[ idx0 ];
+}
+
+template < unsigned int num_values >
+void
+nest::MultiValueRingBuffer< num_values >::resize()
+{
+  size_t size = kernel().connection_manager.get_min_delay() + kernel().connection_manager.get_max_delay();
```
heplesser, line 54:
```suggestion
  const size_t size = kernel().connection_manager.get_min_delay() + kernel().connection_manager.get_max_delay();
```

heplesser (changes requested):
@suku248 Neat! I have a few comments, mostly concerning legibility. The comments on `iaf_psc_alpha` apply correspondingly to `iaf_psc_exp`.

models/iaf_psc_alpha.cpp
suku248: I have changes this to call `clear()` here, which calls `resize()`, which only resizes the buffer if needed. This is consistent with the previous code. If `clear()` (and hence `resize()`) is not required here, this should be consistently changed for all neuron models.

models/iaf_psc_alpha.cpp
suku248: Adapted both and also changed the name of the called function to `get_values_all_channels()`.

suku248: Also note that the function is now `const` as `input` is no longer directly reset in the neuron model.

models/iaf_psc_alpha.cpp
suku248: Done.

suku248: Also made other frequently called member functions inline.

models/iaf_psc_alpha.h
suku248: Thanks, done.

models/iaf_psc_alpha.h
suku248: Done.

nestkernel/ring_buffer.h
suku248: Yes. I have also changed the name to `MultiChannelRingBuffer` to make this more consistent.

nestkernel/ring_buffer.h
suku248: Added more info here.

nestkernel/ring_buffer_impl.h
suku248: Using `index` now.

nestkernel/ring_buffer_impl.h
suku248: Casting is no longer required as `idx0` is now of type `index`, which is a NEST internal equivalent for `size_t`.

nestkernel/ring_buffer_impl.h
suku248: Done.

nestkernel/ring_buffer.h
suku248: They are now called `slot` and `channel`, respectively. Okay?

nestkernel/ring_buffer.h
suku248: Done.

suku248:
@heplesser thanks for the thorough review, please take another look!
@stinebuu I thought reviewing this might be of interest for you as you work on a related solution in a different context.

models/iaf_psc_alpha.cpp
```diff
@@ -340,9 +339,13 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
...
 
+    // get read access to the correct input-buffer slot
+    const index input_buffer_slot = kernel().event_delivery_manager.get_modulo( lag );
+    auto& input = B_.input_buffer_.get_values_all_channels( input_buffer_slot );
+
     // Apply spikes delivered in this step; spikes arriving at T+1 have
     // an immediate effect on the state of the neuron
-    V_.weighted_spikes_ex_ = B_.ex_spikes_.get_value( lag );
+    V_.weighted_spikes_ex_ = input[ Buffers_::SYN_EX ];
     S_.dI_ex_ += V_.EPSCInitialValue_ * V_.weighted_spikes_ex_;
```
heplesser, line 349: I think we recently decided to drop the `weighted_spikes_` recordables. Then, one can move the `input[ ... ]` right into line 349.

models/iaf_psc_alpha.cpp
```diff
@@ -370,7 +373,10 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
     }
 
     // set new input current
-    S_.y0_ = B_.currents_.get_value( lag );
+    S_.y0_ = input[ Buffers_::I0 ];
+
+    // reset all values in the currently processed input-buffer slot
+    B_.input_buffer_.reset_values_all_channels( input_buffer_slot );
```
heplesser, line 379: Pass `lag` here, see above.

models/iaf_psc_alpha.cpp
```diff
@@ -382,15 +388,18 @@ iaf_psc_alpha::handle( SpikeEvent& e )
 {
   assert( e.get_delay_steps() > 0 );
 
+  const index input_buffer_slot = kernel().event_delivery_manager.get_modulo(
+    e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ) );
```
heplesser, line 392: We ought to clean up this eyesore—you decide if you want to do it in this PR or create a new one. Can't we just have an `e.get_delivery_slot()` method and hide the remaining ugliness inside it?

models/iaf_psc_alpha.cpp
```diff
@@ -382,15 +388,18 @@ iaf_psc_alpha::handle( SpikeEvent& e )
...
   {
-    B_.ex_spikes_.add_value( e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ), s );
+    B_.input_buffer_.add_value( input_buffer_slot, Buffers_::SYN_EX, s );
   }
   else
   {
-    B_.in_spikes_.add_value( e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ), s );
+    B_.input_buffer_.add_value( input_buffer_slot, Buffers_::SYN_IN, s );
   }
 }
```
heplesser, line 404: With all the beauty that MultiRingBuffer provides, I am tempted to suggest (`s` has same sign as `e.get_weight()`)
```
B_.input_buffer( e.get_delivery_slot(), s > 0 ? Buffers_::SYN_EX : Buffers_::SYN_IN, s );
```
What do you think?

models/iaf_psc_exp.cpp
```diff
@@ -372,15 +377,16 @@ nest::iaf_psc_exp::handle( SpikeEvent& e )
 {
   assert( e.get_delay_steps() > 0 );
 
+  const index input_buffer_slot = kernel().event_delivery_manager.get_modulo(
+    e.get_rel_delivery_steps( kernel().simulation_manager.get_slice_origin() ) );
+
   if ( e.get_weight() >= 0.0 )
```
heplesser, line 383: Here we are curiously inconsistent with *alpha: >= 0 vs > 0, without it making any difference in practice ;).

heplesser (changes requested):
@suku248 Thank you for your updates! I think the MCRB is a really neat design and in combination with the enum constants makes for very legible code. I added some more suggestions for even further code simplification and prettification—you decide whether to do this here or in a follow-up. Again most comments are on the alpha version, but apply to exp as well.

Someone has earlier argued that our RingBuffer isn't really a ring buffer. How about calling it `MultiChannelInputBuffer`?

nestkernel/ring_buffer_impl.h
```diff
@@ -0,0 +1,58 @@
...
+
+template < unsigned int num_channels >
+void
+nest::MultiChannelRingBuffer< num_channels >::clear()
+{
+  resize(); // does nothing if size is fine
+  // set all elements to 0.0
+  for ( index slot = 0; slot < buffer_.size(); ++slot )
+  {
+    buffer_[ slot ].fill( 0.0 );
```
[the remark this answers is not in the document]
suku248: Okay, done. This potentially introduces an unnecessary `assert()` per call to the public function  `reset_values_all_channels()`, but I think it's not performance relevant.

models/iaf_psc_alpha.cpp
```diff
@@ -340,9 +339,13 @@ iaf_psc_alpha::update( Time const& origin, const long from, const long to )
     S_.I_ex_ = V_.P21_ex_ * S_.dI_ex_ + V_.P22_ex_ * S_.I_ex_;
     S_.dI_ex_ *= V_.P11_ex_;
 
+    // get read access to the correct input-buffer slot
+    const index input_buffer_slot = kernel().event_delivery_manager.get_modulo( lag );
+    auto& input = B_.input_buffer_.get_values_all_channels( input_buffer_slot );
```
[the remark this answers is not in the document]
suku248: As stated above, I think the actual problem is that the `moduli_`-table machinery is not part of the ring-buffer class.

models/iaf_psc_alpha.cpp
suku248: Is there an issue documenting this decision? In any case, this is outside the scope of this PR.

models/iaf_psc_alpha.cpp
suku248: As stated above, I think the actual problem is that the `moduli_`-table machinery is not part of the ring-buffer class.

models/iaf_psc_alpha.cpp
suku248: Done!

models/iaf_psc_exp.cpp
suku248: It's consitsent now. Both neuron models now use the statement with the conditional operator suggested above.

models/iaf_psc_alpha.cpp
heplesser: I fear it was only an informal, undocumented decision. I will create a separate issue for it.

heplesser (approved):
@suku248 Thanks a lot!

PR #1769, opened 2020-09-20
Remove erroneous example in NEST Tutorial

jd41:
`nest.Models?` in IPython would open the documentation of the `nest.Models()` function, rather than returning a list of all models (for which one needs to call `nest.Models()`).


terhorstd (approved):
That is correct. `nest.Models?` is not the same at all.

hakonsbm (approved):
Good catch, thanks!

PR #1774, opened 2020-09-21
Enhance PyNEST error messages from kernel

### LINKED ISSUE #1599: PyNEST error messages from kernel are difficult to read
When an error message is sent from the kernel to PyNEST, the message can be quite difficult to read. Often, the message has a newline in the form of `\n`, which is not taken into account on PyNEST level. The message might also come twice, which would be fine except for the fact that the two come right next to each other, so it is difficult to separate the two. 

See following examples:
```
>>> n = nest.Create('iaf_psc_alpha')
>>> sd = nest.Create('spike_detector')
>>> v = nest.Create('voltmeter')
>>> nest.Connect(sd, n)
NESTErrors.UnexpectedEvent: ('UnexpectedEvent in Connect_g_g_D_D: UnexpectedEvent:
Source node does not send output. Note that detectors need to be connected as Connect(
neuron, detector).', 'UnexpectedEvent', 'Connect_g_g_D_D', ': UnexpectedEvent: Source
node does not send output. Note that detectors need to be connected as Connect(neuron,
detector).')

>>> nest.Connect(v, sd)
NESTErrors.IllegalConnection: ('IllegalConnection in Connect_g_g_D_D: Creation of
connection is not possible because:\nPossible cause: only static synapse types may be
used to connect devices.', 'IllegalConnection', 'Connect_g_g_D_D', ': Creation of connection
is not possible because:\nPossible cause: only static synapse types may be used to connect
devices.')
```

It would be nice if the newlines where rendered so the messages are easier to read. If it was possible to have the second error message either removed (as the two messages are equal), or at least have a new line for it, that would also help a lot.

**Additional context**
I have used commit 6339f2055fbf14de780496df965a43b3a5cd6ada
Python 3.7.6

hakonsbm:
PyNEST classes for error messages that come from the kernel are corrected to correctly display messages, which fixes #1599. 

Previously, error information was passed as arguments to the constructor of the `Exception` class. However, arguments passed to the `Exception` constructor are all passed as `*args` and displayed as the error message without further processing. Passing several arguments would then make the error message a tuple with all the arguments. This PR changes the PyNEST exception classes to pass only the error message as an argument to the `Exception` class constructor.

#### Example:
```python
>>> n = nest.Create('iaf_psc_alpha')
>>> sr = nest.Create('spike_recorder')
>>> v = nest.Create('voltmeter')
>>> nest.Connect(sr, n)
nest.lib.hl_api_exceptions.IllegalConnection: IllegalConnection in SLI function Connect_g_g_D_D: Creation of connection is not possible because:
Source node does not send output.

>>> nest.Connect(v, sr)
nest.lib.hl_api_exceptions.IllegalConnection: IllegalConnection in SLI function Connect_g_g_D_D: Creation of connection is not possible because:
The target node or synapse model does not support data logging requests.
```

### DIFF
--- a/pynest/nest/lib/hl_api_exceptions.py
+++ b/pynest/nest/lib/hl_api_exceptions.py
@@ -82,7 +82,8 @@ class NESTErrors(metaclass=NESTMappedException):
 
             Parameters:
             -----------
-            message: full error message to report.
+            message: str
+                full error message to report.
             """
 
             Exception.__init__(self, message)
--- a/pynest/nest/lib/hl_api_exceptions.py
+++ b/pynest/nest/lib/hl_api_exceptions.py
@@ -101,7 +101,7 @@ class NESTErrors(metaclass=NESTMappedException):
             commandname: command name from SLI.
             errormessage: message from SLI.
             """
-            message = "{} in {}{}".format(errorname, commandname, errormessage)
+            message = "{} in SLI function {}{}".format(errorname, commandname, errormessage)
             NESTErrors.NESTError.__init__(self, message)
 
             self.errorname = errorname
--- a/pynest/nest/lib/hl_api_exceptions.py
+++ b/pynest/nest/lib/hl_api_exceptions.py
@@ -77,23 +77,22 @@ class NESTErrors(metaclass=NESTMappedException):
         """Base exception class for all NEST exceptions.
         """
 
-        def __init__(self, message, *args, **kwargs):
+        def __init__(self, message):
             """Initializer for NESTError base class.
 
             Parameters:
             -----------
             message: full error message to report.
-            *args, **kwargs: passed through to Exception base class.
             """
 
-            Exception.__init__(self, message, *args, **kwargs)
+            Exception.__init__(self, message)
             self.message = message
 
     class SLIException(NESTError):
         """Base class for all exceptions coming from sli.
         """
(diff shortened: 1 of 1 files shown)


jd41:
Great! I recently had such an error message (the one with the "Possible cause"), and I think I would have understood the problem ~30 (or 60? I don't remember anymore) minutes faster with this new version.

stinebuu (approved):
Thanks!

pynest/nest/lib/hl_api_exceptions.py
```diff
@@ -77,34 +77,32 @@ class NESTError(Exception):
         """Base exception class for all NEST exceptions.
         """
 
-        def __init__(self, message, *args, **kwargs):
+        def __init__(self, message):
             """Initializer for NESTError base class.
 
             Parameters:
             -----------
             message: full error message to report.
```
clinssen, line 85:
```suggestion
            message: str
                full error message to report.
```

clinssen: Just to double-check: this should always be a str, not a tuple, correct?

hakonsbm: That's right.

clinssen (approved):
Thanks!

PR #1778, opened 2020-09-24
Timers for Benchmarking (related to issue #1471 "Basic Instrumentation")

wschenck:
UPDATE: Previous TODOs are all done!

This PR is a suggestion how to meet the requirements stated in issue #1471 ("Basic Instrumentation") regarding timers for profiling and benchmarking.

In the current version, the "basic timers" are always enabled (without any switch; this design decision is the result of a discussion between suku248, jarsi and me). The "detailed timers" can be enabled by the cmake option "-Dwith-detailed-timers=ON". To give an example of a cmake command with detailed timers enabled:
`cmake ../.. -DCMAKE_INSTALL_PREFIX:PATH=~/opt/NEST/zam884_nest3-timers -Dwith-mpi=ON -Dwith-detailed-timers=ON`

In contrast to the requirements in issue  #1471, even more detailed timers are provided. The reason is that these timers are required for the dry-run development for 5G to compare real and dry runs of NEST simulations. But in the end, we think that these additional timers are useful for many additional benchmarking purposes. The complete list is:
- `time_construction_create` (basic timer, always enabled)
- `time_construction_connect` (basic timer, always enabled)
- `time_communicate_prepare` (basic timer, always enabled)
  - nested in `time_communicate_prepare`:
    - `time_gather_target_data`
      - nested in `time_gather_target_data`:
        - `time_communicate_target_data`
- `time_simulate` (basic timer, always enabled)
  - nested in `time_simulate`:
    - `time_update`
    - `time_gather_spike_data`
      - nested in `time_gather_spike_data`:
        - `time_collocate_spike_data`
        - `time_communicate_spike_data`
        - `time_deliver_spike_data`


suku248:
@hakonsbm could you please check that all high-level connect functions contain timers?
@jarsi could you please check that time measurements for this PR are comparable to previous measurements?
@terhorstd do you think the issue #1471 is fully addressed by this PR?

nestkernel/event_delivery_manager.cpp
```diff
@@ -356,7 +383,21 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
       // Needs to be called /after/ set_end_and_invalid_markers_.
       set_complete_marker_spike_data_( assigned_ranks, send_buffer_position, send_buffer );
 #pragma omp barrier
+      // TODO: Discuss, if the barrier above should be better placed after if-clause
```
hakonsbm, line 386: A decision should be made here and the todo should be removed.

nestkernel/event_delivery_manager.cpp
```diff
@@ -356,7 +383,21 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
...
+
+#ifdef TIMER_DETAILED
+#pragma omp barrier
+    if ( tid == 0 )
+    {
+      sw_collocate_spike_data_.stop();
+
+      // TODO_SDR: Following function call commented out because it causes real runs to hang!
+      // kernel().mpi_manager.synchronize(); // to get an accurate time measurement
+      // across ranks
```
hakonsbm, line 397: Can this be removed?

nestkernel/event_delivery_manager.cpp
```diff
@@ -620,10 +677,30 @@ EventDeliveryManager::gather_target_data( const thread tid )
...
 #pragma omp barrier
     kernel().connection_manager.clean_source_table( tid );
+
+#ifdef TIMER_DETAILED
+#pragma omp barrier
+    if ( tid == 0 )
+    {
+      // TODO_SDR: Following function call commented out because it causes real runs to hang!
+      // kernel().mpi_manager.synchronize(); // to get an accurate time measurement
+      //                                    // across ranks
```
hakonsbm, line 687: Can this be removed?

hakonsbm (changes requested):
Looks pretty good, thanks for implementing this! There are some TODOs that should be fixed or removed, see my comments below.

@suku248 There are essentially only three connection functions in the C++ API, where all high-level connect functions converge, and these all contain timers. I have also checked that timing works with the high-level functions. However, It would be nice to have a proper unittest of the instrumentation.

jarsi:
Thank you very much for implementing these!

The timers and their placement are comparable with what we have used so far.

I am wondering about some new omp barrriers. I think their purpose is to make sure every thread takes the same amount of time finishing some timed interval. But won't a bunch of new barriers considerably slow down heavily multithreaded (100 + threads) measurements and consequently distort the measurement?

Would it make sense to either not introduce new barriers at all, thus assuming the work is balanced evenly across all threads, or to let every thread have its own stopwatch and taking either the mean or the max across one mpi process as measured time.

nestkernel/event_delivery_manager.cpp
jarsi: I agree, we shouldn't introduce a MPI synchronization point. This would yield significant overhead, which is also not accounted for in any timer at the moment.

nestkernel/event_delivery_manager.cpp
jarsi: I agree, we shouldn't introduce a MPI synchronization point. This would yield significant overhead, which is also not accounted for in any timer at the moment.

nestkernel/event_delivery_manager.cpp
```diff
@@ -356,7 +383,21 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
       // Needs to be called /after/ set_end_and_invalid_markers_.
       set_complete_marker_spike_data_( assigned_ranks, send_buffer_position, send_buffer );
 #pragma omp barrier
+      // TODO: Discuss, if the barrier above should be better placed after if-clause
+    }
+
+#ifdef TIMER_DETAILED
+#pragma omp barrier
```
jarsi, line 390: I guess this barrier is placed here in order to guarantee that all threads take the same amount of time and tid = 0 represents all other threads. But I think every new barrier would introduce significant synchronization overhead and distort the overall measurement. This would be especially severe if we used these timers on architectures with high core counts allowing a huge number of threads.

nestkernel/event_delivery_manager.cpp
```diff
@@ -620,10 +677,30 @@ EventDeliveryManager::gather_target_data( const thread tid )
     kernel().connection_manager.save_source_table_entry_point( tid );
 #pragma omp barrier
     kernel().connection_manager.clean_source_table( tid );
+
+#ifdef TIMER_DETAILED
+#pragma omp barrier
```
jarsi, line 682: Same as above.

nestkernel/simulation_manager.cpp
```diff
@@ -692,10 +736,25 @@ nest::SimulationManager::update_connection_infrastructure( const thread tid )
...
+    sw_gather_target_data_.start();
+  }
+#endif
+
   // communicate connection information from postsynaptic to
   // presynaptic side
   kernel().event_delivery_manager.gather_target_data( tid );
 
+#ifdef TIMER_DETAILED
+#pragma omp barrier
```
jarsi, line 751: Same as above.

nestkernel/simulation_manager.cpp
```diff
@@ -666,6 +704,12 @@ nest::SimulationManager::call_update_()
 void
 nest::SimulationManager::update_connection_infrastructure( const thread tid )
 {
+#pragma omp barrier
```
jarsi, line 707: This barrier is only called once, right? So one extra barrier probably does not have a huge impact on simulation time. So probably not the same as above.

nestkernel/simulation_manager.cpp
```diff
@@ -706,6 +765,12 @@ nest::SimulationManager::update_connection_infrastructure( const thread tid )
     kernel().node_manager.set_have_nodes_changed( false );
   }
   kernel().connection_manager.unset_have_connections_changed( tid );
+
+#pragma omp barrier
```
jarsi, line 769: Probably only called once? So I guess there is not much overhead expected via this barrier.

nestkernel/simulation_manager.cpp
```diff
@@ -883,9 +948,17 @@ nest::SimulationManager::update_()
         }
 
       } // of if(wfr_is_used)
-      // end of preliminary update
+        // end of preliminary update
 
+#ifdef TIMER_DETAILED
+#pragma omp barrier
```
jarsi, line 954: Same as above.

stinebuu:
@wschenck Will you have time to update this PR before Friday? We would like to get the release started.

nestkernel/event_delivery_manager.cpp
suku248: All threads should come to the same result for `gather_completed_checker_.all_true()`.
Hence, the barrier can stay inside the if-clause. Now removed comment.

nestkernel/event_delivery_manager.cpp
suku248: Agreed. Now removed barrier.

nestkernel/event_delivery_manager.cpp
suku248: Agreed. Now removed synch point and comment.

nestkernel/event_delivery_manager.cpp
suku248: Agreed. Now removed barrier.

nestkernel/event_delivery_manager.cpp
suku248: Agreed. Now removed synch point and comment.

nestkernel/simulation_manager.cpp
suku248: Yes, it's not frequently called. Keep this barrier.

nestkernel/simulation_manager.cpp
suku248: Yes, it's not frequently called. Keep this barrier.

nestkernel/simulation_manager.cpp
suku248: Yes, it's not frequently called. Keep this barrier.

nestkernel/simulation_manager.cpp
suku248: Yes, it's not frequently called. Keep this barrier.

suku248:
@wschenk please check the removed barriers and MPI synch points and ping the reviewers ( @hakonsbm @jarsi @terhorstd ) if you are happy with the changes.

suku248:
> @suku248 There are essentially only three connection functions in the C++ API, where all high-level connect functions converge, and these all contain timers. I have also checked that timing works with the high-level functions. However, It would be nice to have a proper unittest of the instrumentation.

@hakonsbm Thanks for checking the connect functions! I don't think it's possible to create a unit test for this, but including the detailed-timers case in CI might make sense.

nestkernel/event_delivery_manager.cpp
```diff
@@ -679,12 +673,8 @@ EventDeliveryManager::gather_target_data( const thread tid )
     kernel().connection_manager.clean_source_table( tid );
 
 #ifdef TIMER_DETAILED
-#pragma omp barrier
```
wschenck: I think meanwhile that it is more logical at this point to put sw_communicate_target_data.start()/stop() into the "omp single" region.

wschenck: Otherwise I am happy with the changes, thank you! (@hakonsbm @jarsi @terhorstd @suku248)

wschenck:
@hakonsbm, @jarsi, @terhorstd, @suku248 : From my perspective, the pull request is ready to go. If you have no further objections, please pull the code into master.

hakonsbm (approved):
Looks good to me now.

heplesser:
@suku248 @jarsi Are you happy with this PR or do you see the need for further changes?

suku248:
We agreed in the last Hackathon that with the fixes to `time_construction_connect` by @wschenck the PR is now ready for the final steps of the review.

heplesser:
@jarsi @jasperalbers Could you review this soon?

jasperalbers:
@heplesser I will be on it soon, hopefully at the end of the week, or latest by the start of next week.

jasperalbers (commented):
I did a test with the multi-area model which confirms that the timing data is reasonably similar to the old timers implemented by @jarsi, thus I don't see any necessary changes in the implementation.
I would suggest to add the info about
- how to set the correct compilation flag
- how to access the timers
- the precise definition of the measured times

somewhere in the documentation (sorry if this already exists, I wasn't able to find it).

suku248:
Thanks @jasperalbers ! There is a separate issue for the documentation #1977. 
If you are happy with this PR, could you please approve?

jasperalbers (approved):

PR #1781, opened 2020-09-25
Fix incrementation of weight and delay in connect_arrays

hakonsbm:
In `connect_arrays`, incrementing the weight and delay pointers did not work, leading all connections to get the weight and delay of the first element if passing arrays of weights and delays. This PR fixes the incrementation and adds a test.


suku248 (approved):
Thanks for fixing this!

stinebuu (approved):
Thanks!

PR #1739, opened 2020-08-29
Simple typos/documentation fixes

jd41:
I fixed some simple typos and changed the interpunction and capitalization (in particular w.r.t. proper names) in some places. Feel free to cherry-pick if you disagree with some of my choices. I also removed some SLI-like notation in a help page.

PR #1784, opened 2020-09-28
Replace all occurrences of "to old" with "too old"

Helveg:
There were some typos using "to old" instead of "too old", I did a _Replace All_ and hope it's fine like this. I stumbled upon it in this warning:

```
--------------------------------------------------------------------------------


ATTENTION!
You are about to compile NEST without the GNU Scientific
Library or your GSL is to old (before v1.11). This means
that conductance-based neuron models and some random number
generators will not be available.

--------------------------------------------------------------------------------
```


Helveg:
And now noticed there's actually legitimate uses of "to old <noun>" in there, which I'll revert ;)

clinssen:
@Helveg: the unit tests should be fixed now if you pull upstream (we just merged #1739 which should cover the issue).

stinebuu (approved):

stinebuu:
Merging this with only 1 reviewer as it is a small change.

Issue #1779, opened 2020-09-25
Compilation instructions don't work immediately in Kubuntu 20.04


jd41:
Update: I am sorry, apparently I looked at the documentation in the stable branch. The latest branch instructs to install the correct packages [here](https://nest-simulator.readthedocs.io/en/latest/installation/linux_install.html?highlight=%22python3-matplotlib%22#installation-from-source)

The instructions for installing older versions [here](https://nest-simulator.readthedocs.io/en/latest/installation/oldvers_install.html?highlight=%22python-matplotlib%22#standard-configuration) still contain the incorrect package names (or does NEST just not support Python 3 at that point?)

Original text:
Following the installation/compilation instructions on Kubuntu 20.04, the system doesn't find `python-matplotlib` and `python-scipy`.

```
(nest-doc) kubuntu@kubuntu:~/Documents/GitHub/nest-simulator$ sudo apt-get install -y cython libgsl-dev libltdl-dev libncurses-dev libreadline-dev python-all-dev python-numpy python-scipy python-matplotlib python-nose openmpi-bin libopenmpi-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package python-matplotlib is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

Package python-scipy is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'python-scipy' has no installation candidate
E: Package 'python-matplotlib' has no installation candidate
```
The right packages seem to be `python3-matplotlib` and `python3-scipy` here. I am not directly submitting a PR because, for all I know, another distribution may still call the things `python-matplotlib` and `python-scipy`... If I don't hear about that for a few days in this thread, I will submit a PR.

jd41:
Update: I am sorry, apparently I looked at the documentation in the stable branch. The latest branch instructs to install the correct packages [here](https://nest-simulator.readthedocs.io/en/latest/installation/linux_install.html?highlight=%22python3-matplotlib%22#installation-from-source)

The "latest" instructions for installing older versions [here](https://nest-simulator.readthedocs.io/en/latest/installation/oldvers_install.html?highlight=%22python-matplotlib%22#standard-configuration) still contain the incorrect package names.

terhorstd:
[...]

PR #1785, opened 2020-09-29
Updated Ubuntu dependency package names for old versions/trailing slashes in installation command

jd41:
Closes what is left of #1779. Also added trailing \ to the multi-line `apt-get` command.

According to the release log [here](https://www.nest-simulator.org/download/), Python 3 support was already added in NEST 2.4, so the installation instructions continue to be useful for some older versions.


terhorstd:
We do not want to maintain documentation back to the end of time. The file was just missed during the clean-up, #1802 fixed this.

PR #1786, opened 2020-09-29
Disabling the build artefacts upload on to S3

lekshmideepu:
Immediate fix for the S3 access key issue which prevents the build artefacts upload on to S3 and thereby failing the merge commit builds.  As we would have a new upload setup soon, it is better to disable the S3 deployment at the moment.


lekshmideepu:
Since this is a minor change which would fix the issues with our merge commit builds, I am going to merge it.

PR #1788, opened 2020-10-01
Update documentation of gif models

sarakonradi:
This PR updates the links, references and equations of the following files: 
- ``gif_cond_exp.h``
- ``gif_cond_exp_multusynapse.h``
- ``gif_pop_psc_exp.h``
- ``gif_psc_exp.h``
- ``gif_psc_exp_multisynapse.h``

### DIFF
--- a/models/gif_pop_psc_exp.h
+++ b/models/gif_pop_psc_exp.h
@@ -54,13 +54,13 @@ This model simulates a population of spike-response model neurons with
 multi-timescale adaptation and exponential postsynaptic currents, as
 described in Schwalger et al. (2017) [1]_.
 
-The single neuron model is defined by the hazard function
+The single neuron model is defined by the hazard function:
 
 .. math::
 
  lambda_0 * exp[ ( V_m - E_sfa ) / Delta_V ]
 
-After each spike the membrane potential V_m is reset to V_reset. Spike
+After each spike, the membrane potential V_m is reset to V_reset. Spike
 frequency
 adaptation is implemented by a set of exponentially decaying traces, the
 sum of which is E_sfa. Upon a spike, all adaptation traces are incremented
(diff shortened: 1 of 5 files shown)


models/gif_pop_psc_exp.h
```diff
@@ -52,15 +52,15 @@ Description
...
 multi-timescale adaptation and exponential postsynaptic currents, as
-described in [1]_.
+described in Schwalger et al. (2017) [1]_.
 
-The single neuron model is defined by the hazard function
+The single neuron model is defined by the hazard function:
 
 .. math::
 
  lambda_0 * exp[ ( V_m - E_sfa ) / Delta_V ]
```
clinssen, line 61:
```suggestion
 \lambda_0 * \exp\left(( V_m - E_{sfa} ) / \Delta_V\right)
```

models/gif_psc_exp_multisynapse.h
```diff
@@ -86,28 +86,28 @@ where :math:`V_T(t)` is a time-dependent firing threshold:
 
  V_T(t) = V_{T_star} + \gamma_1(t) + \gamma_2(t) + \ldots + \gamma_m(t)
```
clinssen, line 87:
```suggestion
 V_T(t) = V_{T_{star}} + \gamma_1(t) + \gamma_2(t) + \ldots + \gamma_m(t)
```

models/gif_pop_psc_exp.h
terhorstd: Good point, @clinssen, we could maybe think about a general rule which we can apply to all models. This one I would leave as it is now and fix it if necessary after the discussion.

terhorstd (approved):
LGTM :+1:

models/gif_pop_psc_exp.h
clinssen: We cannot leave it as-is, as it wouldn't render right; would have to escape the underscores in that case and wrap variable names in \texttt{}.

clinssen (approved):
Thanks for the fast updates!

PR #1793, opened 2020-10-02
Fix SetKernelStatus examples in documentation

clinssen:
The ``min_delay`` and ``max_delay`` API seems to have changed at one point; could someone confirm that the new method (using GetKernelStatus) is correct?

### DIFF
--- a/doc/faqs/faqs.rst
+++ b/doc/faqs/faqs.rst
@@ -53,12 +53,12 @@ Where does data get stored
 By default, the data files produced by NEST are stored in the directory
 from where NEST is called. The location can be changed by changing the
 property ``data_path`` of the root node using
-``nest.SetKernelStatus("data_path", "/path/to/data")``. This property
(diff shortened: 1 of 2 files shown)


hakonsbm (commented):
Looks good, thanks for catching and fixing this. There are also some examples of the old version in the FAQ:
https://github.com/nest/nest-simulator/blob/a1623206604023bbba22d486e48997d09bd93d76/doc/faqs/faqs.rst#L56
https://github.com/nest/nest-simulator/blob/a1623206604023bbba22d486e48997d09bd93d76/doc/faqs/faqs.rst#L61

Can you fix those too?

clinssen:
@hakonsbm: done! Thanks for pointing that out.

hakonsbm (approved):
Great, thanks!

stinebuu:
Merging with only one reviewer as this is a minimal change. Thanks!

PR #1796, opened 2020-10-08
Fix compiler warnings

hakonsbm:
This PR fixes most compiler warnings generated by GCC 10 and Clang 10 when using the `-pedantic` and `-Wextra` flags in addition to the usual `-Wall` flag.

Warning flags triggered and the changes made to fix them are as follows:

- `-pedantic`: Removed extra semicolons.
- `-Wtype-limits`: Removed asserts where, because of type-constraints, the expression could never be false, and if-tests where the expression could never be true.
- `-Wunused-parameter`: Removed unused parameters in inherited class functions.
- `-Wunused-private-field`: Removed unused class variable.
- `-Wdeprecated-copy`: With C++11, the implicitly declared copy assignment operator is deprecated when a user-declared constructor or copy-constructor is defined (because of the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming))). In many cases the user-defined copy-constructor could be `default`ed, but in some cases a copy assignment operator had to be defined.
- `-Wimplicit-int-float-conversion`: Added `static_cast`s to avoid implicit casts from `long` and `long long` to `double`.
- `-Wmaybe-uninitialized`: Added Position copy assignment operator with vector argument to avoid implicit conversion from vector to Position. The implicit conversion would create a temporary Position which would trigger the warning.
- `-Wsign-compare`: Changed variable type when iterating threads from `index` to `thread`.


jougs (commented):
Amazing! Many thanks for this very nice cleanup!

I don't have a single comment on the code itself, as this looks just perfect :heart: Just one question: would it make sense to use `-pedantic` and `-Wextra` when building in the stage of Travis CI where only static code checks are performed? This way, regressions could be detected more easily. I could even imagine that we now made the check for warnings in [`extras_parse_travis_log.py`](https://github.com/nest/nest-simulator/blob/master/extras/parse_travis_log.py) fail the build. Any opinions?

And a quick heads up to @clinssen: Can you please also have a look at this, as it seems that some of the changes in the model classes should be ported to the NESTML templates.

hakonsbm:
@jougs Doesn't the static code check stage on Travis skip compiling entirely? We could add the flags to the compiling stages, though. It looks like the `extras_parse_travis_log.py` script already counts warnings, so adding a simple limit to catch regressions should be easy.

jougs:
@hakonsbm: yes, you're right! Sorry for mixing this up. I'd appreciate a change like the one you describe. Thanks!

jougs (approved):
Nice! I think for now this is a safe solution, but we should strive for getting rid of the remaining ones. @clinssen and I discussed this and we'll discuss a proposal in the next developer VC.

hakonsbm:
Not all warnings are fixable. Some warnings are configuration-dependent, such as `-Wunused-parameter` when for example compiling without MPI. Some warnings are raised in external headers, like in the OpenMP header or MUSIC headers. The `-Wimplicit-fallthrough` warning is raised in `sli/scanner.cc`, but is not fixable because the switch here is by design supposed to fallthrough.

jougs:
@gtrensch: could you have a look at this, please? Thanks!

stinebuu:
@gtrensch ping from hackathon!

extras/parse_travis_log.py
```diff
@@ -360,6 +360,19 @@ def makebuild_summary(log_filename, msg_make_section_start,
...
+        expected_warnings = 263
+    elif environ['xNEST_BUILD_TYPE'] == 'OPENMP_ONLY':
+        expected_warnings = 8
+    elif environ['xNEST_BUILD_TYPE'] == 'FULL':
+        expected_warnings = 4006
+    elif environ['xNEST_BUILD_TYPE'] == 'FULL_NO_EXTERNAL_FEATURES':
+        expected_warnings = 8
+    else:
+        expected_warnings = 0  # Set to 0 if none of the above build-types, to not crash the script
+
```
gtrensch, line 375: From a code design point of view, wouldn't it be better to pass the content of `NEST_BUILD_TYPE` as a parameter to `parse_travis_log.py`? This would expose the logical dependency between the `.travis.yml` file and the log parser and reduce the risk of someone changing the name and accidentally breaking the log parser.

hakonsbm: Good idea, I've changed it.

gtrensch (approved):
Thanks for this nice contribution!

Issue #1790, opened 2020-10-02
Extension modules cannot find ``spatial`` C++ header files


clinssen:
When building an extension module against the latest master, compilation fails because the header files in the ``spatial`` subdirectory cannot be found; or, alternatively, because they are not copied into the right place in the installation directory.

In ``include/nest/nestmodule.h``, the following lines:
```C++
// Includes from spatial:
#include "spatial/ntree.h"
#include "spatial/position.h"
```
can be replaced by
```C++
// Includes from spatial:
#include "ntree.h"
#include "position.h"
```
for an easy fix.

Questions that need to be addressed before we can move to a fix:
- Should the files be installed to a ``spatial`` subdirectory, or should the include paths be altered to remove the ``spatial`` subdirectory?
- There was discussion of reorganising the ``nestkernel`` directory, which might already make this issue go away.

Related NESTML issue: https://github.com/nest/nestml/issues/581

hakonsbm:
@clinssen Was this fixed by #1798?

clinssen:
Yes, thanks for pointing this out! I wrote "Fixes #1790" in a reply in that comments thread, but apparently GitHub doesn't trust that as I'm not the OP.

Issue #1175, opened 2019-04-11
Source directory layout reorganisation


clinssen:
As discussed during the Jülich NEST Hackathon of April 2019.

- [ ] Create a new `/src` directory and move all C++ and sli source files into here. E.g. `/nestkernel`, `/sli`, etc. are moved to `/src/nestkernel`, `/src/sli`, etc.
- [ ] PyNEST/SLI/C++ examples should be combined in `/examples`
- [ ] Create new subdirectories in the `/src/models` directory according to:
  - `stimulating_devices`
  - `recording_devices`
  - `spiking_neurons` (including precise variants)
  - `synapses` (maybe we need this also with `spiking_`, `rate_`, `binary_`)
  - `binary_neurons`
  - `rate_neurons`
  - gap junctions? dopamine? structural platicity 
- [ ] PyNEST test suite should not be part of NEST module

jougs:
This is closely related to #291.

jougs:
In addition to cleaning things up, this also needs to take into account the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) and commonly respected practice in the major distributions.

heplesser:
I generally consider this proposal sensible sensible, although I can never decide if I really like the `src` directory. Most of our examples are also code of a kind. Topology is going to disappear as a separate entity in NEST3.

clinssen:
Removed topology module from the list.

hakonsbm:
Wait until #1282 is merged!

gtrensch:
@clinssen #1282 is merged.

gtrensch:
@clinssen Are there any updates on this?

jougs:
After recent discussions with @terhorstd and others, I am not sure anymore if the first and third point actually have a majority of the developers on their side. We should definitely bring this up again in one of the video conferences. I have a working prototype of the fourth point and the second should also be relatively easy to fix. As I'm working on this already, I'm re-assigning to myself.

jougs:
The fourth point is now [described as the solution](https://github.com/nest/nest-simulator/issues/1130#issuecomment-679318588) to #1130.

jougs:
As this is more controversial than expected and several points were/are addressed in separate issues, I'm closing this ad *wontfix*.

PR #1798, opened 2020-10-09
Fix location for 'spatial' headers

jougs:
This fixes a problem where the header files under `spatial` have not been found when linking them from extension modules. The fix for now is to move the files to the `nestkernel` directory directly and postpone restructuring of that directory. As this is an emergency fix that only moves files around without touching actual code, I'm only nominating a single reviewer for sanity checking.

This is a follow-up to #1748.


clinssen:
Thinking back to @terhorstd's between-headers dependency analysis of the ``nestkernel`` directory at the recent hackathon, wasn't the ``spatial`` subdirectory mostly (and almost perfectly) the only self-contained section of the nestkernel sources? If we are considering moving towards more subdirectories in the ``nestkernel`` directory, not fewer, then perhaps we could use something like the following CMake code instead of removing the directory as a temporary measure:

```cmake
function( FILTER_HEADERS2 in_list out_list )
    set( tmp_list ${in_list} )
    list( FILTER tmp_list EXCLUDE REGEX "^.*spatial.*$")
    set( ${out_list} ${tmp_list} PARENT_SCOPE )
endfunction ()

FILTER_HEADERS("${nestkernel_sources}" install_headers )
FILTER_HEADERS2("${install_headers}" install_headers )
install( FILES ${install_headers}
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/nest )

install( DIRECTORY "${CMAKE_SOURCE_DIR}/nestkernel/spatial"
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/nest )
```

This comes at the end of ``nestkernel/CMakeLists.txt``.

The "spatial" files have to be explicitly excluded using the FILTER_HEADERS2 function, because the CMake script builds a long list of *all* source files, and the subdirectory is lost when they are copied. So we first exclude them from the regular ``install( FILES ...)`` and copy them separately using an ``install( DIRECTORY ...)``. A better solution would loop over a list containing, for now, the string "spatial" as the only element (but to which other subdirectories could be added).

clinssen:
Fixes #1790.

clinssen:
Related issue that was closed earlier as "wontfix": https://github.com/nest/nest-simulator/issues/1175

Perhaps discussion on the ``nestkernel`` directory layout organisation could take place there after this PR is merged.

hakonsbm (approved):
Looks fine as an emergency fix.

PR #1800, opened 2020-10-14
Skip setting of parameter if test is skipped

### LINKED ISSUE #1799: test_connect_arrays_mpi.py fail even though HAVE_MPI4PY is False
**Describe the bug**
The python test `test_connect_arrays_mpi.py` currently fails on my computer, with the error message `NameError: name 'MPI' is not defined`
```
  File "../nest-simulator/pynest/nest/tests/test_connect_arrays_mpi.py", line 54, in <module>
    class ConnectArraysMPICase(unittest.TestCase):
  File "../nest-simulator/pynest/nest/tests/test_connect_arrays_mpi.py", line 64, in ConnectArraysMPICase
    comm = MPI.COMM_WORLD.Clone()
NameError: name 'MPI' is not defined
```

I do not have mpi4py, and `HAVE_MPI4PY` is thus False. So, the test should be skipped, but it isn't.
https://github.com/nest/nest-simulator/blob/1e0ce51e17e48933a95463cc35a5c2c5d99ca366/pynest/nest/tests/test_connect_arrays_mpi.py#L40-L51

I think the problem is that `@unittest.skipIf(not HAVE_MPI4PY, 'mpi4py is not available')` have to be placed before the [individual tests](https://docs.python.org/3/library/unittest.html), and not before the entire test-class? But I am not sure, does anyone know?

 - OS: Ubuntu-16.04
 - Python-Version: Python 3.7.6
 - NEST-Version: 6997cae6c893470257a0654fe48edb8b799169a0

stinebuu:
This fixes #1799 .

Only set `mpi4py.MPI.COMM_WORLD.Clone()` if `HAVE_MPI4PY` is `True`.

### DIFF
--- a/pynest/nest/tests/test_connect_arrays_mpi.py
+++ b/pynest/nest/tests/test_connect_arrays_mpi.py
@@ -57,7 +57,11 @@ class ConnectArraysMPICase(unittest.TestCase):
     multiple processes.
     """
     non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
-    comm = MPI.COMM_WORLD.Clone()
+
+    # We will enter the test class if HAVE_MPI4PY is false (though all tests are then skipped), so we need to only
+    # set the variable if we have mpi4py.
+    if HAVE_MPI4PY:
+        comm = MPI.COMM_WORLD.Clone()
 
     # With pytest or nosetests, only run these tests if using multiple processes
     __test__ = MULTIPLE_PROCESSES
--- a/pynest/nest/tests/test_connect_arrays_mpi.py
+++ b/pynest/nest/tests/test_connect_arrays_mpi.py
@@ -58,8 +58,10 @@ class ConnectArraysMPICase(unittest.TestCase):
     """
     non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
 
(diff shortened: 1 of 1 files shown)


hakonsbm (approved):
Thanks, looks good!

pynest/nest/tests/test_connect_arrays_mpi.py
```diff
@@ -57,7 +57,11 @@ class ConnectArraysMPICase(unittest.TestCase):
     multiple processes.
     """
     non_unique = np.array([1, 1, 3, 5, 4, 5, 9, 7, 2, 8], dtype=np.uint64)
-    comm = MPI.COMM_WORLD.Clone()
+
+    # We will enter the test class if HAVE_MPI4PY is false (though all tests are then skipped), so we need to only
+    # set the variable if we have mpi4py.
```
jougs, line 62: I find "enter" a bit general here. Here's a suggestion that works better for me (in case I understand the mechanism correctly, anyway):
```suggestion
    # The test class is instantiated by the unittest framework regardless of the value of
    # HAVE_MPI4PY, even though all tests will be skipped in case it is False. In this
    # situation, we have to manually prevent calls to MPI in order to avoid errors during
    # the execution.
```

jougs (commented):
Thanks! I have a minor comment on the wording of the comment but am otherwise fine with the change.

jougs (approved):

PR #1801, opened 2020-10-14
Add PyNEST version dunder

### LINKED ISSUE #1348: Pynest version retrieval not standard
Hello!

I was trying to retrieve the nest version from pynest and noticed that this is done by calling a function `nest.version()`. 

[PEP396 recommends](https://www.python.org/dev/peps/pep-0396/)  adding it as a module [dunder](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names):  `nest.__version__`. Even though PEP396 is deferred (not sure why or by which other PEP entry), to me this is more intuitive and feels quite standard for all other python packages I have used, some examples:
```
>>> numpy.__version__
'1.17.3'
>>> scipy.__version__
'1.3.1'
>>> matplotlib.__version__
'3.1.2'
>>> pandas.__version__
'0.25.3'
>>> seaborn.__version__
'0.9.0'
```

Also the version should probably just be a string of the version number and not contain the package name in it like it is now, e.g. `'Nest 2.18.0'` --> `'2.18.0'`.

I encountered this issue together with issue #1347 and they can probably be solved together.

hakonsbm:
Adds a version dunder and removes `NEST` from the version string.
```python
>>> import nest
>>> print(nest.__version__)
'2.20.0'
```
Fixes #1348.

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -44,7 +44,7 @@ try:
 except ImportError:
     pass
 
-__version__ = version()
+__version__ = ll_api.sli_func("statusdict /version get")
 
 
 def test():
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -44,6 +44,8 @@ try:
 except ImportError:
     pass
 
+__version__ = version()
+
 
 def test():
     """Runs all PyNEST unit tests."""
--- a/pynest/nest/hl_api.py
+++ b/pynest/nest/hl_api.py
@@ -102,5 +102,4 @@ __all__ = [
     'message',
     'set_verbosity',
     'sysinfo',
-    'version',
 ]
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -43,7 +43,6 @@ __all__ = [
     'SetStatus',
     'set_verbosity',
     'sysinfo',
-    'version',
 ]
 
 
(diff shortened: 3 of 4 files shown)


Silmathoron (approved):

jougs (changes requested):
Could `__version__` maybe be turned into a `property` and the `version()` function removed as I requested in #1348? I find it a bit confusing to have both...

hakonsbm:
@jougs Sure, I can remove `version()`. But what should `__version__` be a `property` of?

jougs:
@hakonsbm: hmmm, well, can it be a property of the module? If that's not possible, forget all I said and I'll immediately approve and merge :wink:

hakonsbm:
@jougs Looks like there is no way to set a `property` of a module, it has to be of a class. However, with Python 3.7 it's possible to define a `__getattr__()` for modules. But that seems needlessly complicated just for setting the version string... 

I still think `version()` should still be removed, though.

pynest/nest/__init__.py
```diff
@@ -44,6 +44,8 @@
 except ImportError:
     pass
 
+__version__ = version()
```
jougs, line 47:
```suggestion
__version__ = ll_api.sli_func("statusdict /version get")
```
If this works, `version()` could be safely removed, but there might be help texts that point to it and need to be adapted as well.

jougs (commented):
Thanks for the clarification! I've added a new idea in the form of a direct code suggestion.

pynest/nest/__init__.py
hakonsbm: It works! I've removed `version()` and adapted the help text.

jougs (approved):
Nice!

PR #1812, opened 2020-10-23
Return all values when using get() on a SynapseCollection with more than one synapse

stinebuu:
If you call `get` on a `SynapseCollection` with more than one synapse, and you use a `key` that is only applicable for one (or some) of the synapses, NEST throws a keyword error, like in the example below:
```
nrns = nest.Create('iaf_psc_alpha', 4)

nest.Connect(nrns, nrns, 'one_to_one')
nest.Connect(nrns, nrns, 'one_to_one', {'synapse_model': 'stdp_synapse', 'alpha': 2.})

conns = nest.GetConnections()
conns.get('alpha')
```
If you call `get()` with no arguments, the parameter is not included. This should not be the case though, and this PR fixes it. We now return `None` for the synapses that don't have the parameter, and the value for the synapses that do. This is consistent with how it is done for `NodeCollection`.


pynest/nest/lib/hl_api_helper.py
```diff
@@ -528,7 +528,17 @@ def restructure_data(result, keys):
     int, list or dict
     """
     if is_literal(keys):
-        final_result = result[0] if len(result) == 1 else list(result)
+        if len(result) != 1:
+            all_keys = sorted({key for result_dict in result for key in result_dict})
```
jougs, line 532: # :exploding_head:

jougs (approved):
Looks good to me! @Silmathoron, could you please also have a look? Thanks!

pynest/nest/lib/hl_api_helper.py
Silmathoron: yeah, it also took me a good 30 s to figure out all that was happening in this single line...

pynest/nest/lib/hl_api_types.py
```diff
@@ -737,16 +737,15 @@ def get(self, keys=None, output=''):
...
+        # Return empty tuple if we have no connections or if we have done a nest.ResetKernel()
         num_conn = GetKernelStatus('num_connections')
         if self.__len__() == 0 or num_conn == 0:
             return ()
 
         if keys is None:
             cmd = 'GetStatus'
         elif is_literal(keys):
-            cmd = 'GetStatus {{ /{0} get }} Map'.format(keys)
+            cmd = 'GetStatus'
```
Silmathoron, line 748: sorry, I don't understand why this line changed

Silmathoron (commented):
There's one change I did not understand, please see comment below

pynest/nest/lib/hl_api_types.py
stinebuu: When we have a `SynapseCollection` with 2 or more synapses, for instance `static_synapse` and `stdp_synapse`, and ask for a parameter that is only part of one of them, like `alpha`, this will throw an error, because alpha is not part of the `static_synapse` dictionary. So now we get all dictionaries and look for the parameter on PyNEST level.

pynest/nest/lib/hl_api_helper.py
stinebuu: Do you want me to break it up into more steps?

Silmathoron: no no, it's fine ^^

pynest/nest/lib/hl_api_types.py
Silmathoron: ok, of course, it goes with the other change!
maybe a comment here would be useful though, just saying that it will be done in ``restructure_data``?

stinebuu: Done!

stinebuu:
@Silmathoron thanks for looking through. I have added the comment, have another look!

Silmathoron (approved):
All good for me! :+1:

PR #1815, opened 2020-10-27
Fix measurement of presimulation time in PyNEST Potjans_2014

jhnnsnk:
This PR fixes the measurement of the presimulation time which accidentally included the time to connect.


stinebuu (approved):
Thanks! I am merging this based on one review as it is a one-liner.

PR #1819, opened 2020-10-28
Add table to spike_recorder documentation

sarakonradi:
This PR adds a configuration table to the ``spike_recorder`` documentation, as discussed offline in the NEST Doc Meeting. It also includes a link to the "Recording from simulations" guide.

### DIFF
--- a/models/spike_recorder.h
+++ b/models/spike_recorder.h
@@ -64,9 +64,6 @@ spike creation rather than that of their arrival.
 The call to ``Connect`` will fail if the connection direction is
 reversed (i.e., connecting *sr* to *neurons*).
 
-To learn more about recording devices in NEST, please refer to the
-:doc:`../guides/recording_from_simulations` guide.
-
 Properties
 ++++++++++
 
@@ -91,6 +88,9 @@ All recorders have a set of common properties that can be set using
               deactivation time in ms, relative to origin. The value of stop
               must be greater than or equal to start
 ===========  ======================================================================
+
+To learn more about recording devices in NEST, please refer to the
+:doc:`../guides/recording_from_simulations` guide.
 
 EndUserDocs */
 
--- a/models/spike_recorder.h
+++ b/models/spike_recorder.h
@@ -64,6 +64,9 @@ spike creation rather than that of their arrival.
 The call to ``Connect`` will fail if the connection direction is
 reversed (i.e., connecting *sr* to *neurons*).
 
+To learn more about recording devices in NEST, please refer to the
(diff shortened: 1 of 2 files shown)


models/spike_recorder.h
```diff
@@ -64,6 +64,34 @@ spike creation rather than that of their arrival.
...
 reversed (i.e., connecting *sr* to *neurons*).
 
+Properties
+++++++++++
+
+All recorders have a set of common properties that can be set using
+``SetDefaults`` on the model class or ``SetStatus`` on a device instance:
+
+===========  ======================================================================
+ label        A string (default: `“”`) specifying an arbitrary textual label for
```
terhorstd, line 74: I'm all for UTF-8, but it's probably safer to stick to ascii here.
```suggestion
 label        A string (default: `""`) specifying an arbitrary textual label for
```

models/spike_recorder.h
```diff
@@ -64,6 +64,34 @@ spike creation rather than that of their arrival.
...
+              the device. Recording backends might use the label to generate
+              device specific identifiers like filenames and such.
+ n_events     The number of events that were collected by the recorder can be
+              read out of the `n_events` entry. The number of events can be
+              reset to 0. Other values cannot be set.
+ origin       A positive floating point number (default : `0.0`) used as the
+              reference time for `start` and `stop`.
+ record_to    A string (default: `“memory”`) containing the name of the recording
```
terhorstd, line 82: same here
```suggestion
 record_to    A string (default: `"memory"`) containing the name of the recording
```

terhorstd (commented):
I still have to check if the renderer needs new-lines between table rows...

jougs (commented):
I'm sorry, but I don't like this change as it duplicates the information in the [Guide on recording from simulations](https://nest-simulator.readthedocs.io/en/latest/guides/recording_from_simulations.html#common-recorder-properties) and is bound to get inconsistent over time.

I think the proper solution is to add a `ref`-style link to that section in a way that would be rendered as a link to the guide if the device documentation is displayed [standalone](https://nest-simulator.readthedocs.io/en/latest/models/spike_recorder.html) or as just an anchor-link for the version that is embedded in the guide.

I am well aware that this might require a more dynamic processing of the file by the doc extractor in cases where the file is used in different contexts.

sarakonradi:
During the NEST Hackathon, we agreed that all devices should contain a link to the common properties. @jougs suggested addressing this, among other "Recording from simulations" guide improvements, in a separate PR.

I would thus propose closing the current PR.

PR #1820, opened 2020-10-30
Create StructuralPlasticityNode separate from Archiving_Node

clinssen:
This is a refactoring that helps to separate concerns between Archiving_Node, and a new class, Structural_Plasticity_Node. The former manages spiking history and internal state in a "history" buffer, typically for the sake of STDP synapses. The latter manages a calcium concentration and a map of "synaptic elements".

Because we want to make NESTML generated code independent of the Archiving_Node, we would like NESTML neurons to not inherit from Archiving_Node any longer, but directly from Node. The replacement functionality would be generated by NESTML. However, we are not concerned with the structural plasticity code, and would like to leave this undisturbed. For this reason, it helps a lot if these are separated into two classes, and this is easily possible because they don't have many mutual dependencies.

So we propose to change the inheritance diagram from ``Node ← Archiving_Node`` to ``Node ← Structural_Plasticity_Node ← Archiving_Node``. NESTML neurons would then inherit directly from Structural_Plasticity_Node, thereby avoiding Archiving_Node. No further changes are required to the existing NEST models, which continue to inherit (typically) from Archiving_Node.

@sdiazpier: I know that you were working on a refactoring of the structural plasticity system into a user module. I hope that this PR does not cause a lot of overhead for you in doing that. I've strived for the minimal change to take the structural plasticity components out, but welcome your feedback!


heplesser:
@clinssen Two comments:
- If you no longer inherit from archiving node, how will the NESTML neurons support STDP?
- If you refactor this, I'd much appreciate if you'd use the opportunity to replace our hybrid camel-underscore-naming with pure camel: `ArchivingNode`, `StructuralPlasticityNode`.

clinssen:
Thanks for your comments! NESTML will generate its own code for this part of the infrastructure. Archiving_Node does not offer any flexibility for different types of plasticity: it is not templatised with a histentry type, even though different types of neurons have different archiving requirements (e.g. whether a triplet-STDP type trace is needed, a vanilla all-to-all STDP trace, extra traces for the Clopath synapse, etc.) See also https://github.com/nest/nest-simulator/blob/master/nestkernel/histentry.h.

The NESTML generated code will make Archiving_Node obsolete, as well as the histentry structs, and we should probably remove the virtual declarations that are specific to STDP, Clopath and Urbancyk plasticity from the Node class as well (see from https://github.com/nest/nest-simulator/blob/a1623206604023bbba22d486e48997d09bd93d76/nestkernel/node.h#L672 onwards). But that's for a future PR.

The underscores I will address as soon as our Travis MacOS environment is fixed.

heplesser:
@clinssen Thanks for the explanation, that makes a lot of sense! Does this mean one could make a compile-time decision that I never want plasticity and so can build all neuron models without support for it? Would the same also apply for structural plasticity? I guess I am just wondering why we no longer need a "plasticity node" class (aka Archiving_Node) but still need a "structural plasticity node" class with your new NESTML-based design.

clinssen:
Indeed, if no plasticity is used in the network, then no spike archiving is needed and will not be compiled in (that is, the code will not even be generated in the neuron C++ file). This will tie in well with the just-in-time build strategy that we proposed at the last NEST Conference (but already applies even without JIT, that is, when NESTML is invoked manually).

Structural plasticity is a rather different beast, and for NESTML we would like to remain agnostic to this for now. Hence, we propose to split off the code related to structural plasticity from the ArchivingNode, so that it can be left untouched as much as possible. I believe @sdiazpier has longer-term plans to turn this into a user module, which would then ultimately also replace the StructuralPlasticityNode.

nestkernel/structural_plasticity_node.cpp
```diff
@@ -0,0 +1,270 @@
...
+// Includes from nestkernel:
+#include "kernel_manager.h"
+
+// Includes from sli:
+#include "dictutils.h"
+
+namespace nest
+{
+
+// member functions for Structural_Plasticity_Node
```
heplesser, line 34: This comment is superfluous.

nestkernel/structural_plasticity_node.cpp
```diff
@@ -0,0 +1,270 @@
...
+nest::Structural_Plasticity_Node::clear_history()
+{
+  Ca_minus_ = 0.0;
+  Ca_t_ = 0.0;
+}
+
+
+/* ----------------------------------------------------------------
+* Get the number of synaptic_elements
+* ---------------------------------------------------------------- */
```
heplesser, line 151: This type of comment is not needed here, remove.

nestkernel/structural_plasticity_node.h
```diff
@@ -0,0 +1,184 @@
...
+#include "dictdatum.h"
+
+#define DEBUG_ARCHIVER 1
+
+namespace nest
+{
+
+/**
+ * \class Structural_Plasticity_Node
+ * a node which archives spike history for the purposes of
```
heplesser, line 47: Single line summary starting with capital letter. Details can follow in a separate paragraph.

nestkernel/structural_plasticity_node.h
```diff
@@ -0,0 +1,184 @@
...
+ * timing dependent plasticity
+ */
+class Structural_Plasticity_Node : public Node
```
heplesser, line 50: Please change to `StructuralPlasticityNode`—no need to repeat errors of the past ;).

nestkernel/structural_plasticity_node.h
```diff
@@ -0,0 +1,184 @@
...
+  void set_spiketime( Time const& t_sp, double offset = 0.0 );
+
+  /**
+   * \fn void clear_history()
+   * clear spike history
+   */
+  void clear_history();
+
+private:
+  double last_spike_;
```
heplesser, line 145: Looks to me as if we have `last_spike_` now in `StructuralPlasticityNode` *and*  in `Archiving_Node`. Please make sure that it appears only one, presumably `Archiving_Node`.

nestkernel/structural_plasticity_node.h
```diff
@@ -0,0 +1,184 @@
...
+
+  /*
+   * Structural plasticity
+   */
```
heplesser, line 149: This comment is no longer relevant, since we are now in SPN.

nestkernel/structural_plasticity_node.h
```diff
@@ -0,0 +1,184 @@
...
+
+  // Time of the last update of the Calcium concentration in ms
```
heplesser, line 151: Turn comment and subsequent comments into proper Doxygen comments.

heplesser (changes requested):
@clinssen Overall, this looks good to me overall, but see my inline comments, especially on the class name and `last_spike_`.

clinssen:
Thanks for the detailed review! The new commits should address all your comments. Of course renaming the Archiving_Node bumped us up from 5 to 125 changed files. ``test_jonke_synapse.py`` was corrected for DOS to Unix line endings in the process, which is why it shows up in the diff.

nestkernel/structural_plasticity_node.h
```diff
@@ -142,40 +141,44 @@ class Structural_Plasticity_Node : public Node
...
-
-  // Time of the last update of the Calcium concentration in ms
   double Ca_t_;
 
-  // Value of the calcium concentration [Ca2+] at Ca_t_. Intracellular calcium
-  // concentration has a linear factor to mean electrical activity of 10^2,
-  // this means, for example, that a [Ca2+] of 0.2 is equivalent to a mean
-  // activity of 20Hz.
+  /**
+   * Value of the calcium concentration [Ca2+] at Ca_t_. Intracellular calcium
```
heplesser, line 150: Split into single-line short description followed by details.

heplesser (approved):
@clinssen Thanks for even fixing historical flaws ;). I have one little thing I'd like to see fixed in a comment, but will approve now already.

sdiazpier (approved):
Dear @clinssen  thanks a lot for this PR. It looks great from my side and clean cut. No further comments from my side at this moment.

jougs:
There is still [on failing MPI test](https://travis-ci.org/github/nest/nest-simulator/jobs/742663955#L7816).

clinssen:
@jougs: thanks for being so observant! Pulling latest master fixed that issue, and we also spun off #1837 to point out that the number of warnings generated by the C++ build itself is now checked against a hard-coded number. It should be ready for merge now.

PR #1823, opened 2020-10-31
Update PyNEST template code to reference current nest 3 code.

BSVogler:
When searching for the original code I noticed that this template contains outdated pynest code. I hereby copy the current docstrings to this template. The method name is also capitalized as in the used code and written in the PyNEST README.md.


stinebuu (approved):
Looks good to me!

sarakonradi (approved):
Thanks for updating the template code!

Issue #1755, opened 2020-09-14
Dump and Load


BSVogler:
This issue is intended as a starting point for a discussion in the [hackathon](https://github.com/nest/nest-simulator/wiki/NEST-Hackathon-14.09-18.09-2020).

Following suggestion #1618, the problem of supporting iterative runs remains unsolved.
The current iterative workflow looks like this:

![suggestion for dump nest impl](https://user-images.githubusercontent.com/801818/93072372-9472ff80-f681-11ea-912b-de4492bee2d7.png)

The problem with the suggestion of a partial reset is that the point of a reset call is not defined.
The shared weights across episodes are one example of this workflow. There are other workflows where the starting conditions are changed instead of loading the weights.
### Suggestion
By specifying it with the introduction of an explicit dump command it could save a serialization in memory. This serialization could then later be loaded again.
The new workflow would then be reduced in computing time and requires less user code.
![suggestion for dump nest impl2](https://user-images.githubusercontent.com/801818/93073213-d94b6600-f682-11ea-899d-aa1542fff778.png)

### Specifics in dump and load
For this use case, the neurons and connections (connectome) are of importance as the network is fully reset each episode. The dumping could be incrementally extended to allow the dump to include different parts of the simulation.
The dump command could then look like this `nest.dump(["neuros,", "synapses", "RNG"])`.
`nest.load()` would only overwrite the loaded parameters, potentially throwing an error.
The parameters in the dump command make it explicit to the user which values are loaded.

### Further benefits
Calls to `ResetNetwork` prevent the usage of nest `prepare`+`run` resulting in a lot of overhead. If dumping and loading overwrites everything relevant to the computation the workflow could be further increased to 
Use `Prepare`, `Load`+`Run`, `Load`+`Run`…, `Cleanup`.

This requires that dumping is almost complete.
![suggestion for dump nest impl3](https://user-images.githubusercontent.com/801818/93073230-dea8b080-f682-11ea-818c-bb85d47686f4.png)

Depending on the serialization format it would also enable inspection or sharing of generated networks.

jougs:
Many thanks to all for the (at least in my opinion) very nice discussion during the hackathon. @BSVogler can you update the issue as you go, i.e. add comments on the initial plan and progress and questions? Thanks!

BSVogler:
Thank you for considering this suggestion. Below I want to sumamrize some discussion points.

[...]

PR #1828, opened 2020-11-04
PyNest implementation of Dump and Load

BSVogler:
Implements a first version of #1755.

It supports the dumping and loading of nodes and connections. Because of circular imports `GetKernelStatus` needed to be replaced with the SLI function.


jougs:
@daphnecor: I seem to remember that you also showed interest in this feature some time ago. May I ask you to also have a look and tell us what you think about the proposed implementation? Many thanks for your considerations!

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -460,3 +468,109 @@ def DisableStructuralPlasticity():
...
+    Dict:
+        the created nest obejcts and all synapses
+
+    See Also
+    ---------
+
+    Dump : Function to obtain a structured memory dump.
+    """
+    created = dict()
+    sli_func("({}) M_WARNING message".format("Loading is currently only supported for a single process."))
```
Silmathoron, line 539: see above

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -460,3 +468,109 @@ def DisableStructuralPlasticity():
...
+
+
+    See Also
+    ---------
+
+    Load : Function to construct the net from the data.
+
+    """
+
+    sli_func("({}) M_WARNING message".format("Dumping is currently only supported for a single process."))
```
Silmathoron, line 498: I don't think that warning is a good idea : if it does not work it should crash and burn in a very explicit way ;)

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -460,3 +468,109 @@ def DisableStructuralPlasticity():
...
+    dumpdata = dict()
+    numnetwork = GetKernelStatus("network_size")
+
+    # nothing added
+    if numnetwork < 2:
```
Silmathoron, line 503: I think a more explanatory comment would help here... I've no idea where this magic two comes from...

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -460,3 +468,109 @@ def DisableStructuralPlasticity():
...
+        return dumpdata
+
+    nodes = NodeCollection(range(1, numnetwork + 1))
```
Silmathoron, line 506: Why not use ``GetNodes``?

pynest/nest/lib/hl_api_types.py
```diff
@@ -686,8 +685,7 @@ def get(self, keys=None, output=''):
 
         # Return empty tuple if we have no connections or if we have done a
         # nest.ResetKernel()
-        num_conn = GetKernelStatus('num_connections')
-        if self.__len__() == 0 or num_conn == 0:
+        if self.__len__() == 0 or sli_func('GetKernelStatus /num_connections get') == 0:
```
Silmathoron, line 688: why was that necessary?

pynest/nest/lib/hl_api_types.py
```diff
@@ -743,10 +741,8 @@ def set(self, params=None, **kwargs):
             If the specified parameter does not exist for the connections.
         """
 
-        # This was added to ensure that the function is a nop (instead of,
-        # for instance, raising an exception) when applied to an empty
-        # SynapseCollection, or after having done a nest.ResetKernel().
-        if self.__len__() == 0 or GetKernelStatus()['network_size'] == 0:
+        # Do nothing if we have no connections or if we have done a nest.ResetKernel()
+        if self.__len__() == 0 or sli_func('GetKernelStatus /num_connections get') == 0:
```
Silmathoron, line 745: idem

Silmathoron (changes requested):
Thanks for the proposal.

I think this may require further discussion and specification of what we want behavior-wise.
E.g. do we want the possibility of saving only specific nodes (not devices for instance)?

Regarding MPI, I'm not sure there is any significant issue with its implementation, so I wonder whether it should not be included right away for dump using a simple gather. I don't think it requires any additional change for load. If it is indeed a problem, then an error should be raised.

Finally, I'm a bit unsure about the reason for changes in ``hl_api_types``, could you explain them?

pynest/nest/lib/hl_api_simulation.py
BSVogler: it seems that an empty network already has network size 1. but I agree that this should be added in a comment.

pynest/nest/lib/hl_api_types.py
BSVogler: this allows the removal of one import and solves the problem of circular imports

pynest/nest/lib/hl_api_simulation.py
BSVogler: I think if you run it with MPI enabled on several machines it won't crash but just modify the local machine. Do you think it should check for that case explicitly?

BSVogler:
> do we want the possibility of saving only specific nodes (not devices for instance)?

Either we add more filters to the method signature or the filtering must be performed after the dump is obtained.
The OOP way would be to add the method dump to the neuroncollection class.

> dumps may be memory expensive, do we want to provide a file version that is written progressively without memory impacts?

For the use-cases I had in mind no writing to files are needed, however this is a another common use-case of serialisation and is in danger of consuming too much RAM when performed in two steps. Therefore I would be open to add another optional parameter to the method to pass a file, which, if set, will be written to. The return of the method will then be void.

BSVogler:
The high level code in python is not optimized for an iterative writing to a file. Every node and connection is obtained in one call. I would therefore delay an implementation until the dumping&loading is moved to lower level C++.
I fixed and commented the other points. As I am now working in a rather unrelated role full time and work on nest happens in my free time, further extensive work on this issue from my side would be probably delayed.

The interface could look like this:
`def Dump(selections: List = ("nodes", "synapses"), tofile: str="") -> Union[Dict, None]:`

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -460,3 +468,115 @@ def DisableStructuralPlasticity():
...
+
+    # nothing added
+    if numnetwork == 0:
+        return dumpdata
+
+    nodes = GetNodes()
+    if "nodes" in selections:
+        dumpdata["nodes"] = GetStatus(nodes)
+
+    syn_ids = GetConnections(source=nodes)
```
heplesser, line 512: `source=nodes` has no effect here, since `nodes` are all nodes in the network.

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -460,3 +468,115 @@ def DisableStructuralPlasticity():
...
+                source, target = conn["source"], conn["target"]
+                # remove unused, copy to prevent side effects
+                specs = copy.copy(conn)
+                specs.pop("port")
+                specs.pop("receptor")
+                specs.pop("synapse_id")
+                specs.pop("target_thread")
+                specs.pop("source")
+                specs.pop("target")
+                Connect([source], [target], syn_spec=specs)
```
heplesser, line 574: Array connect possible?

heplesser:
@BSVogler Thank you for your efforts towards a Dump/Load feature for NEST. After careful consideration, I believe that your approach will not and essentially cannot work sufficiently reliably. I will therefore close this PR. If you disagree after reading my arguments below, please re-open as a draft to invite further discussion.

The key problem is that the result of `GetStatus()` and `SetStatus()` depends entirely on the neuron or synapse model in question and is thus not well defined. In particular, there are no clear rules on which state variables are gettable/settable from different neuron models. `iaf_psc_alpha`, e.g., provides only the membrane potential, but neither value nor derivative of the postsynaptic currents. Thus, the result of a Dump/Load cycle would depend considerably on the particular models involved. While this can be useful for an individual user with a specific use case, it will lead to confusion and, in the worst case, to simulations that do different things than the user believes they are doing, resulting in misinterpretation of results. The tests currently included in this PR indeed do not check in detail what is reconstructed on load.

Therefore, a general Dump/Load creates considerable risks and is not compatible with our aim of providing NEST as a highly reliable research tool. In the long run, there may be possibilities for a consistent approach based on NESTML.

NEST3 provides many new features that make storing and reloading select features of the network expliclity, so users and any code readers know exactly what is stored and restored. Here a simple example, which probably is also faster because iterations are moved to the C++ level:

```python
n = nest.Create('iaf_psc_alpha', n=100, params={'V_m': nest.random.uniform(-80., -60.)})
nest.Connect(n, n, {'rule': 'fixed_indegree', 'indegree': 10}, {'weight': nest.random.uniform(1., 2.)})

n_dump = n.get('V_m', output='pandas')
c_dump = nest.GetConnections().get(('source', 'target', 'weight', 'delay'), output='pandas')

# store / load dumps using pickle or pandas

nest.ResetKernel()

n = nest.Create('iaf_psc_alpha', n=len(n_dump), params={'V_m': n_dump.V_m.values})
nest.Connect(c_dump.source.values, c_dump.target.values, 'one_to_one', 
             {'weight': c_dump.weight.values, 'delay': c_dump.delay.values})
```

BSVogler:
@heplesser  The model is specified in the dump and considered in the loading.
Is your point that some neuron models may depend on variables that are not exposed? Then I would agree that this is problematic and opens up undefined behavior in some cases.

I'd like to see this functionality reconsidered again with NESTML.

heplesser:
@BSVogler Yes, the problem is that a number of variables are not exposed.

Issue #1717, opened 2020-08-10
[DESIGN] Refactoring the extension module system


jougs:
The SLI Interpreter encapsulates a certain portion of its extended functionality in *modules*, which consist of some C++ code, implementing new commands, and (optionally) some bundled `.sli` file(s) for type checking and handling function overloading. All such modules derive from `SLIModule` and are added to the interpreter at compile-time by calling `SLIInterpreter::addmodule()`.

At some point in time, the `DynamicLoaderModule` was introduced, which let users load modules dynamically at run-time. The main use case for dynamically loaded modules is adding new neuron and synapse models, connection builders, or recording and (soon, see #1456) stimulation backends to NEST. The `DynamicLoaderModule` as well as the extension modules themselves are derived from `SLIModule` and thus are, at least technically, a part of SLI rather than of NEST.

As SLI's modules were originally not meant to be loaded dynamically, but rather for encapsulation and to provide new commands to the SLI Interpreter, the API they (and by extension also the `DynamicLoaderModule`) rely on are lacking some functionality like the possibility to reset a module and read/pass parameters from/to them.

NEST's mechanism for encapsulation are *managers*. A manager is responsible for a certain aspect of the simulation kernel (e.g. Models, MPI, Threading, RNGs, ...). In a sense, managers serve a similar purpose for NEST (encapsulation) as the modules originally did for SLI, however the implementations are somewhat orthogonal and the managers provide a much more complete API when it comes to dynamic parameter changes at run-time.

A closer look at the existing modules reveals the following situation:
1. Some modules (e.g. the `modelsmodule` providing the standard neuron, device and synapse models to NEST) just use the module mechanism for encapsulation, while they are actually not interacting with the SLI Interpreter, but rather only with NEST.
1. Other modules (e.g. the SLI language modules) are mainly providing new commands to the SLI Interpreter (that may still call functions of the NEST kernel internally)

This basically means that we have two distinct classes of modules:
1. modules that extend the SLI interpreter (think Python modules, but linked statically at compile-time)
2. modules that extend NEST (dynamically loaded at run-time or statically linked at compile-time)

The current use of modules derived from `SLIModule` to extend both SLI and NEST statically (i.e. at compile-time) and dynamically (i.e. at run-time) constitutes a breach of abstraction and a stronger than necessary dependency of the NEST simulation kernel on the SLI Interpreter. This leads to a number of downsides:

[...]

PR #1830, opened 2020-11-07
Refactor conngen module to be a ConnBuilder

jougs:
This PR refactors the Connection Generator interface for NEST to be a regular `ConnBuilder` called `conngen` instead of being an extension module called `conngen`, which used to provide the `CGConnect` function. This addresses one of the action items of  #1717.


nestkernel/conn_builder_conngeninterface.cpp
```diff
@@ -20,158 +20,133 @@
...
 
-      kernel().connection_manager.connect( ( *source_node_ids )[ source ],
-        target_node,
-        target_thread,
-        synmodel_id,
-        dummy_params,
-        params[ d_idx ],
-        params[ w_idx ] );
+      delay_ = new ScalarDoubleParameter( params[ d_idx ], num_threads );
+      weight_ = new ScalarDoubleParameter( params[ w_idx ], num_threads );
```
jougs, line 118: @hakonsbm: I was hoping that setting `weight_` and `delay_` here would have the effect that the subsequent call to `single_connect` will just use these values for establishing the connections. Unfortunately, this seems to not be the case and the connections are just created with defaults.

A completely orthogonal idea would of course be to directly call `ConnectionManager::connect()` here and not use any `Parameters` and `single_connect()` at all. For that I'd probably would need to get some values (e.g. the synapse model) from the *syn_spec*.

So, the point is that, the whole mechanism for `Parameters` and how to implement a `ConnBuilder` has apparently become a bit too complex for me and I'm a bit at loss here. Could you maybe give me some advice? I suggest to have a dedicated session on this during the hackathon next week in which we could also discuss a bit about how to document all this.

nestkernel/conn_builder_conngeninterface.cpp
```diff
@@ -20,158 +20,133 @@
...
+  librandom::RngPtr rng = kernel().rng_manager.get_grng();
 
   int source;
   int target;
-  const int num_parameters = cg->arity();
+  const int num_parameters = cg_->arity();
   if ( num_parameters == 0 )
   {
+    weight_ = delay_ = 0;
+
```
jougs, line 76: Remove to use defaults

doc/installation/compilation_options.rst
```diff
@@ -137,6 +133,21 @@ already contains an example implementation that should work with the
 
 For more details, see :doc:`configuration`.
 
+.. _compile_with_libneurosim:
+
+Support for libneurosim
+-----------------------
+
+In order to allow NEST to create connections using external libraries,
+it provides an interface for the Connection Generator Interface from
```
jougs, line 142:
```suggestion
it provides support for the Connection Generator Interface from
```

nestkernel/conn_builder.cpp
```diff
@@ -58,9 +58,10 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
   , exceptions_raised_( kernel().vp_manager.get_num_threads() )
   , use_pre_synaptic_element_( false )
   , use_post_synaptic_element_( false )
+  , parameters_requiring_skipping_()
+  , synapse_model_id_( kernel().model_manager.get_synapsedict()->lookup( "static_synapse" ) )
```
jougs, line 62: @stinebuu: Not sure why I have this one. Maybe it should go away?

pynest/examples/csa_spatial_example.py
```diff
@@ -28,25 +28,26 @@
...
 References
-~~~~~~~~~~~~
+~~~~~~~~~~
 
-.. [1] Djurfeldt M, Davison AP and Eppler JM (2014). Efficient generation of connectivity in neuronal networks
-       from simulator-independent descriptions. Front. Neuroinform.
+.. [1] Djurfeldt M, Davison AP and Eppler JM (2014). Efficient generation of
+       connectivity in neuronal networks from simulator-independent
+       descriptions. Front. Neuroinform.
        http://dx.doi.org/10.3389/fninf.2014.00043
```
jougs, line 44:
```suggestion
       https://doi.org/10.3389/fninf.2014.00043
```

nestkernel/conn_builder.cpp
jougs: Removed in e40c193

nestkernel/conn_builder_conngen.cpp
```diff
@@ -20,158 +20,140 @@
...
+
+    if ( syn_specs[ 0 ]->known( names::weight ) or syn_specs[ 0 ]->known( names::delay ) )
+    {
+      throw BadProperty( "Properties weight and delay cannot be specified in syn_spec if the "
+			 "ConnectionGenerator has values." );
+    }
   }
-  const index synmodel_id = static_cast< index >( synmodel );
-  DictionaryDatum dummy_params = new Dictionary();
+}
```
stinebuu, line 65: You should probably also check if `syn_specs.size() > 1` and throw an error if this is the case. Collocation of synapses should not be possible if we use the `conngen` rule.

nestkernel/conn_builder_conngen.cpp
```diff
@@ -20,158 +20,140 @@
...
+
+    for ( Dictionary::iterator it = params_map_->begin(); it != params_map_->end(); ++it )
+    {
+      it->second.set_access_flag();
+    }
+
+    if ( syn_specs[ 0 ]->known( names::weight ) or syn_specs[ 0 ]->known( names::delay ) )
+    {
+      throw BadProperty( "Properties weight and delay cannot be specified in syn_spec if the "
+			 "ConnectionGenerator has values." );
```
stinebuu, line 62: Is the indentation right here? It looks a bit weird, but it might just be me :)

nestkernel/conn_builder_conngen.cpp
```diff
@@ -20,158 +20,140 @@
...
 
-    const bool w_idx_is_0_or_1 = ( w_idx == 0 ) or ( w_idx == 1 );
     const bool d_idx_is_0_or_1 = ( d_idx == 0 ) or ( d_idx == 1 );
+    const bool w_idx_is_0_or_1 = ( w_idx == 0 ) or ( w_idx == 1 );
     const bool indices_differ = ( w_idx != d_idx );
-    if ( not( w_idx_is_0_or_1 and d_idx_is_0_or_1 and indices_differ ) )
+    if ( not( d_idx_is_0_or_1 and w_idx_is_0_or_1 and indices_differ ) )
     {
-      throw BadProperty( "w_idx and d_idx have to differ and be either 0 or 1." );
+      throw BadProperty( "d_idx and w_idx have to be either 0 or 1 and cannot be the same." );
```
stinebuu, line 105: It might be better to use `delay index and weight index` instead of `d_idx and w_idx` in the message.

stinebuu (changes requested):
Nice! I have some comments and suggestions, but they shouldn't be too big.

examples/nest/conngen_interface.sli
```diff
@@ -22,19 +22,17 @@
 
 
 /** @BeginDocumentation
-   Name: testsuite::test_conngen_interface - Test runability of all CGConnect variants
+Name: testsuite::test_conngen_interface - Test runability of Connect variant with conngen rule
 
-   Synopsis: (test_conngen_interface) run
+Synopsis: (test_conngen_interface) run
```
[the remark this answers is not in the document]
jougs: Once upon a time, in a land far far away, we all believed that a C++ implementation of the CSA is on its way and will become available *very soon*. It was at that time that I created this file, which was meant to be a *test*. Then reality came (in total contrast to that C++ implementation) and I moved it to `examples` to keep it (hope dies last and all that...).

So much for history (and fairy tales). I have pushed a slightly updated version of the file in f844959be.

nestkernel/conn_builder_conngen.cpp
jougs: What about
```suggestion
      throw BadProperty( "The indices for weight and delay have to be either 0 or 1 and cannot be the same." );
```

nestkernel/conn_builder_conngen.cpp
jougs: Fixed in c2b4eade0.

nestkernel/conn_builder_conngen.cpp
jougs: Ah, yes. Now that you say it, I remember that we talked about this! Fixed in 824caee56. Could you maybe provide me with a code snipped that creates collocated synapses, so I can write a test that checks if the error reporting works? Thanks!

pynest/nest/tests/test_csa.py
```diff
@@ -167,44 +113,61 @@ def test_CSA_OneToOne_synmodel(self):
             self.assertEqual(len(conns), 1)
             self.assertEqual(conns[0]["target"], targets[i].get('global_id'))
```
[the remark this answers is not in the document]
jougs: Much nicer, thanks! Fixed in ccfae0d05.

nestkernel/conn_builder_conngen.cpp
stinebuu: Looks good!

examples/nest/conngen_interface.sli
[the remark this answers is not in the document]
stinebuu: Makes sense, thanks for updating :)

nestkernel/conn_builder_conngen.cpp
stinebuu: A simple example would be:
```
node = nest.Create('iaf_psc_alpha', 3)
nest.Connect(node, node, syn_spec=nest.CollocatedSynapses({'weight': -2.}, {'weight': 2.}))
```

stinebuu: If you want to test without weight, you could use other parameters than weights, like `synapse_model` and other parameters that can be set for the individual synapse types.
```
node = nest.Create('iaf_psc_alpha', 3)
nest.Connect(node, node, syn_spec=nest.CollocatedSynapses({'synapse_model': 'stdp_synapse', 'alpha': 3.}, {'synapse_model': 'static_synapse'}))
```

sarakonradi (changes requested):
Thanks! I've added some documentation-related suggestions inline.

doc/guides/connection_management.rst
```diff
@@ -194,10 +194,51 @@ Example:
...
+~~~~~~~
+
+.. admonition:: Availability
+
+   This connection rule is only available if NEST was compiled with
+   :ref:`support for libneurosim <compile_with_libneurosim>`.
+
+To allow the generation of connectivity by means of an external
+library, NEST supports the Connection Generator Interface (`Djurfeldt
+et al., 2014 <https://doi.org/10.3389/fninf.2014.00043>`_).
```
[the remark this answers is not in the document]
jougs: Good idea. Done in d5802c9a5.

nestkernel/conn_builder_conngen.cpp
jougs: I've added a test in 500a7412b.

jougs:
@stinebuu, @sarakonradi: I've addressed all your points in my recent commits. Not sure if we really need the review by @hakonsbm. If you think we're good without, feel free to say so or remove the request of review.

stinebuu (approved):
Thank you!

sarakonradi (approved):
LGTM!

PR #1836, opened 2020-11-10
Improve printing of SynapseCollections

### LINKED ISSUE #1829: Print a SynapseCollection
Currently, when printing a SynapseCollection, the output are two rows of source and target id's, with some borders to separate the two. This is something I implemented a long time ago, mostly as a quick fix so we could actually see the nodes in the SynapseCollection. It is not very pretty, and becomes even less pretty when we have many connections:
```
*--------*----------------------------------------------------------------------------------------------------------------*
| source | 1, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6, 10, 10, ... 98, 98, 98, 98, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99, |
*--------*---------------------------------------------------------------------------------------------------------------------------------*
| target | 2, 7, 9, 8, 8, 7, 9, 9, 8, 7, 7, 8, 9, 100, 101, ... 105, 108, 103, 110, 104, 105, 102, 107, 106, 108, 100, 103, 101, 109, 110, |
*--------*---------------------------------------------------------------------------------------------------------------------------------*
```

Today, @jhnnsnk mentioned the [prettytable](https://pypi.org/project/prettytable/) package, and I have also looked at the [tabulate](https://pypi.org/project/tabulate/) package. I think one of these would be a much better alternative. But, you would then need to install the package to be able to print a SynapseCollection. The backages prints the sources and targets as columns instead of the rows, and also centers. We could also add weights, synapse model, or other information if that is usefull.

Examples of how it will look (I have here inserted the `...` for SynapseCollections with more than 35 connections, so we do not get an extremely long list for some cases).
**PrettyTable:**
```
+--------+--------+
| source | target |
+--------+--------+
|   1    |   2    |
|   3    |   7    |
|   3    |   9    |
|   3    |   8    |
|   4    |   8    |
|   4    |   7    |
|   4    |   9    |
|   5    |   9    |
|   5    |   8    |
|   5    |   7    |
|   6    |   7    |
|   6    |   8    |
|   6    |   9    |
|   10   |  100   |
|   10   |  101   |
|   .    |   .    |
|   .    |   .    |
|   .    |   .    |
|   98   |  105   |
|   98   |  108   |
|   98   |  103   |
|   98   |  110   |
|   99   |  104   |
|   99   |  105   |
|   99   |  102   |
|   99   |  107   |
|   99   |  106   |
|   99   |  108   |
|   99   |  100   |
|   99   |  103   |
|   99   |  101   |
|   99   |  109   |
|   99   |  110   |
+--------+--------+
 ```
**Tabulate:**
```
 source    target
--------  --------
   1         2
   3         7
   3         9
   3         8
   4         8
   4         7
   4         9
   5         9
   5         8
   5         7
   6         7
   6         8
   6         9
   10       100
   10       101
   .         .
   .         .
   .         .
   98       105
   98       108
   98       103
   98       110
   99       104
   99       105
   99       102
   99       107
   99       106
   99       108
   99       100
   99       103
   99       101
   99       109
   99       110
```

You can also do other cool stuff, but I don't know if that is needed.

Does anyone have an opinion on what looks best? Or a thought about whether it is okay that users then would need to install one of the packages. 

As a side note, both packages are available through pip and conda forge.

We could also keep a version of the old print for those who do not have the package installed? Or just not allow printing if the package is not installed.

### PR DESCRIPTION
**Pull request by stinebuu**
This fixes #1829.

Printing will now look like:
```
# single element SynapseCollection:
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
   1        2     static_synapse    1.0      1.0  

# Some connections in the SynapseCollection:
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
   1        2     static_synapse    1.0      1.0  
   3        7     static_synapse    1.0      1.0  
   3        9     static_synapse    1.0      1.0  
   3        8     static_synapse    1.0      1.0  
   4        8     static_synapse    1.0      1.0  
   4        7     static_synapse    1.0      1.0  
   4        9     static_synapse    1.0      1.0  
   5        8     static_synapse    1.0      1.0  
   5        7     static_synapse    1.0      1.0  
   5        9     static_synapse    1.0      1.0  
   6        7     static_synapse    1.0      1.0  
   6        8     static_synapse    1.0      1.0  
   6        9     static_synapse    1.0      1.0  

# large SynapseCollection:
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
   1        2     static_synapse    1.0      1.0  
   3        7     static_synapse    1.0      1.0  
   3        9     static_synapse    1.0      1.0  
   3        8     static_synapse    1.0      1.0  
   4        8     static_synapse    1.0      1.0  
   4        7     static_synapse    1.0      1.0  
   4        9     static_synapse    1.0      1.0  
   5        9     static_synapse    1.0      1.0  
   5        8     static_synapse    1.0      1.0  
   5        7     static_synapse    1.0      1.0  
   6        7     static_synapse    1.0      1.0  
   6        8     static_synapse    1.0      1.0  
   6        9     static_synapse    1.0      1.0  
   10      100    static_synapse    1.0      1.0  
   10      101    static_synapse    1.0      1.0  
   .        .            .           .        .   
   .        .            .           .        .   
   .        .            .           .        .   
   98      105    static_synapse    1.0      1.0  
   98      108    static_synapse    1.0      1.0  
   98      103    static_synapse    1.0      1.0  
   98      110    static_synapse    1.0      1.0  
   99      104    static_synapse    1.0      1.0  
   99      105    static_synapse    1.0      1.0  
   99      102    static_synapse    1.0      1.0  
   99      107    static_synapse    1.0      1.0  
   99      106    static_synapse    1.0      1.0  
   99      108    static_synapse    1.0      1.0  
   99      100    static_synapse    1.0      1.0  
   99      103    static_synapse    1.0      1.0  
   99      101    static_synapse    1.0      1.0  
   99      109    static_synapse    1.0      1.0  
   99      110    static_synapse    1.0      1.0

# Different synapses in the SynapseCollection:
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
   1        4     static_synapse    1.0      1.0  
   2        4     static_synapse    2.0      1.0  
   1        3      stdp_synapse     4.0      1.0  
   1        4      stdp_synapse     3.0      1.0  
   2        3      stdp_synapse     3.0      1.0  
   2        4      stdp_synapse     2.0      1.0
```

### CODE REVIEW

**Comment by heplesser** [comment]
@stinebuu Looks neat! But I wonder if it wouldn't be even more legible if columns were right-aligned, those with decimal point aligned on the point if possible.

**Comment by stinebuu** [comment]
@heplesser This is how it will look with right alignment. What do you think?
I don't know how to align with the decimal points, because now we go through all the elements in the source, target, synapse_model, weight, delay lists and I do not keep any information about the last variable.

I will also ask the other participants at the hackathon for their input.
```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
       1        4  static_synapse      1.0     1.0
       2        4  static_synapse      2.0     1.0
       1        3    stdp_synapse      4.0     1.0
       1        4    stdp_synapse      3.0     1.0
       2        3    stdp_synapse      3.0     1.0
       2        4    stdp_synapse      2.0     1.0

# Many connections
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
       1        2  static_synapse      1.0     1.0
       3        7  static_synapse      1.0     1.0
       3        9  static_synapse      1.0     1.0
       3        8  static_synapse      1.0     1.0
       4        8  static_synapse      1.0     1.0
       4        7  static_synapse      1.0     1.0
       4        9  static_synapse      1.0     1.0
       5        9  static_synapse      1.0     1.0
       5        8  static_synapse      1.0     1.0
       5        7  static_synapse      1.0     1.0
       6        7  static_synapse      1.0     1.0
       6        8  static_synapse      1.0     1.0
       6        9  static_synapse      1.0     1.0
      10      100  static_synapse      1.0     1.0
      10      101  static_synapse      1.0     1.0
       .        .               .        .       .
       .        .               .        .       .
       .        .               .        .       .
      98      105  static_synapse      1.0     1.0
      98      108  static_synapse      1.0     1.0
      98      103  static_synapse      1.0     1.0
      98      110  static_synapse      1.0     1.0
      99      104  static_synapse      1.0     1.0
      99      105  static_synapse      1.0     1.0
      99      102  static_synapse      1.0     1.0
      99      107  static_synapse      1.0     1.0
      99      106  static_synapse      1.0     1.0
      99      108  static_synapse      1.0     1.0
      99      100  static_synapse      1.0     1.0
      99      103  static_synapse      1.0     1.0
      99      101  static_synapse      1.0     1.0
      99      109  static_synapse      1.0     1.0
      99      110  static_synapse      1.0     1.0

# distributed weights
 source   target   synapse model        weight        delay 
-------- -------- --------------- ------------------ -------
       1        3  static_synapse 1.5607061497867107     1.0
       1        4  static_synapse  1.527867006137967     1.0
       1        5  static_synapse 2.5598286911845207     1.0
       2        3  static_synapse 1.2253916095942259     1.0
       2        4  static_synapse  2.689947947859764     1.0
       2        5  static_synapse  1.581260945647955     1.0
```

Just for a comparison, this is how the last example looks with centered alignment:
```
 source   target   synapse model        weight        delay 
-------- -------- --------------- ------------------ -------
   1        3     static_synapse  1.5607061497867107   1.0  
   1        4     static_synapse  1.527867006137967    1.0  
   1        5     static_synapse  2.5598286911845207   1.0  
   2        3     static_synapse  1.2253916095942259   1.0  
   2        4     static_synapse  2.689947947859764    1.0  
   2        5     static_synapse  1.581260945647955    1.0 
```

**Comment by heplesser** [comment]
@stinebuu Thank you, I prefer the right-aligned version. Should we also right-align the headers of right-aligned columns? Is there any way to control the number of decimals, e.g. three decimals as default, and the possibility to modify? I guess that would have to be done as an option setting in the `SynapseCollection` class.  The most flexible way would be to allow the user to provide a format string, either one for all columns or even a dict mapping column names to format strings.

**Comment by stinebuu** [comment]
@heplesser we discussed showing just some decimals as default previously this week, and decided it would be confusing to round the values, because the user would not necessarily know about it beforehand. I think we could add a way to control the number of decimals, but I think the default should be to show all decimals.

I can right align the column headers as well, yes. Do you think all of them should be right aligned, or just the ones with decimal points?

There are other ways of inspecting the SynapseCollection as well, using `get()` for instance, so I don't know if having the user providing a format string is that necessary? If the user is sufficient enough to provide the format, they can format the output from `get()` as they wish. Printing is more a quick overview of the collection in my opinion.

**Comment by heplesser** [comment]
@stinebuu I think the header should be aligned in the same way as the content. But may keeping them centred is fine as well.

Concerning the decimals, I disagree with the decision earlier (sorry I wasn't in that discussion). I assume that users will print a `SynapseCollection` to get a quick overview over synapses. Focus should be on compact and readable representation.  For other purposes, e.g., further analysis, users will extract the data in other ways. At least from a physicists perspective, showing too many digits is (almost) as bad as showing too few.

In that situation, I think showing many decimals distracts. The only point where many decimals would be relevant are very small weights, where using, e.g., three decimals would give the impression that a weight is zero, where it is not. One solution to that could be to use fixed decimal notation for all rows where it fits and scientific notation for smaller entries.

Given the possibility that users can always get the data with `get()` and format themselves, I agree that we don't need fancy formatting options in `SynapseCollection`. But it might be an idea to allow the user to choose when a table is "large" and shown in large notation, and how many lines to keep.

**Comment by stinebuu** [comment]
@heplesser We discussed this during the hackathon (sorry for the late update). We agree with you, printing should give an overview over the connections, and we can thus print only 3 decimals. I have now implemented this. I have not yet implemented scientific notation where we have very small weights and delays, I do not know how difficult that will be yet.

The weights and delays are rounded, and in the places where the 3rd decimal is zero, we only display 2 decimals. It makes it a bit more difficult to read, but I do not know if it is worth the effort in adding the zero. I do not know how to do it at the moment. What do you think? I have added an example below for how it looks now.

The values are right aligned, but I have shifted them one character to the left to make them aligned under the headers.

I have also added a `print_full` to the SynapseCollection. This is only relevant for SynapseCollections with more than 36 elements. To print all, you would do
```
conns = nest.GetConnections()
conns.print_all = True
print(conns)
```

```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
      1        3  static_synapse    1.225   0.184
      1        4  static_synapse    1.581   0.334
      1        5  static_synapse    1.533    0.18
      2        3  static_synapse     2.69   0.179
      2        4  static_synapse    1.445   0.329
      2        5  static_synapse    1.432   0.129

```

File: pynest/nest/lib/hl_api_types.py [hunk 1]
```diff
@@ -614,36 +617,72 @@ def __getitem__(self, key):
...
+            dlay = dlay[:15] + [u'\u22EE '] + dlay[-15:]
+            s_model = s_model[:15] + [u'\u22EE '] + s_model[-15:]
+
+        headers = f'{src_h:^{src_len}} {trg_h:^{trg_len}} {sm_h:^{sm_len}} {w_h:^{w_len}} {d_h:^{d_len}}' + '\n'
+        boarders = '-'*src_len + ' ' + '-'*trg_len + ' ' + '-'*sm_len + ' ' + '-'*w_len + ' ' + '-'*d_len + '\n'
+        output = ''.join('{:>{}}'.format(str(s), src_len-1) +
+                         ' {:>{}}'.format(str(t), trg_len) +  # need to indent because of space between columns
+                         ' {:>{}}'.format(str(sm), sm_len) +
+                         ' {:>{num}.{deci}}'.format(w, num=w_len, deci=4) +
+                         ' {:>{num}.{deci}}'.format(d, num=d_len, deci=4) +
```
heplesser, line 682: The following should fix the "missing leading zero" problem and give exponential notation automatically:
```suggestion
                         ' {:>#{num}.{deci}g}'.format(w, num=w_len, deci=4) +
                         ' {:>#{num}.{deci}g}'.format(d, num=d_len, deci=4) +
```

heplesser (changes requested):
Very nice, I just added a suggestion on how to fix the remaining formatting problems.

pynest/nest/lib/hl_api_types.py
stinebuu: This will not work when we have many connections, because then we insert the unicode for vertical ellipsis (see [here](https://github.com/nest/nest-simulator/pull/1836/files#diff-bdf50826d5c778982ced8b68c069efde8bda3349c0a88e4888cd14346e42f769R672-R673)), and both `#` and `g` only works on numbers. We therefore get the following error when we print a SynapseCollection with many connections: `ValueError: Unknown format code 'g' for object of type 'str'` (If we do not use `g` we still get the error, but with `#` instead)

heplesser: How about a helper such as
```python
def fmt_row_(s, t, sm, w, d, sl, tl, sml, wn, wd, dn, dd):
    try:
        return f'{s:>{sl}d} {t:>{tl}d} {sm:>{sml}s} {w:>#{wn}.{wd}g} {d:>#{dn}.{dd}g}'
    except ValueError:
        return f'{s:>{sl}} {t:>{tl}} {sm:>{sml}} {w:>{wn}.{wd}} {d:>{dn}.{dd}}'
```
and then just
```python
'\n'.join(fmt_row_(s, t, sm, w, d, ...) 
           for s, t, sm, w, d in zip(srcs, trgt, s_model, wght, dlay))
```
If you define `fmt_row_()` locally in the method, I think you don't even need to pass the length parameters.

heplesser (changes requested):
I came up with a slightly more elaborate solution that hopefully will allow flexible formatting.

pynest/nest/lib/hl_api_types.py
stinebuu: Smart! I will take a look.

stinebuu:
@heplesser Thanks for your comments, the helper function did the trick! This is how different SynapseCollections looks like now when printing:

``` 
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
      1        2  static_synapse    1.000   1.000
```
```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
      1        2  static_synapse    1.000   1.000
      3        7  static_synapse    1.000   1.000
      3        9  static_synapse    1.000   1.000
      3        8  static_synapse    1.000   1.000
      4        8  static_synapse    1.000   1.000
      4        7  static_synapse    1.000   1.000
      4        9  static_synapse    1.000   1.000
      5        8  static_synapse    1.000   1.000
      5        7  static_synapse    1.000   1.000
      5        9  static_synapse    1.000   1.000
      6        7  static_synapse    1.000   1.000
      6        8  static_synapse    1.000   1.000
      6        9  static_synapse    1.000   1.000
```
```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
      1       95  static_synapse    2.892  0.2970
      1       98  static_synapse    2.873  0.2870
      1      101  static_synapse    1.843  0.3330
      1       91  static_synapse    1.225  0.1840
      1      100  static_synapse    2.256  0.3940
      1       94  static_synapse    1.627  0.1460
      1       99  static_synapse    2.207  0.1250
      1       96  static_synapse    2.040  0.1620
      1       92  static_synapse    2.134  0.2270
      1       93  static_synapse    2.982  0.1720
      1       97  static_synapse    1.686  0.3430
      2       97  static_synapse    1.720  0.1430
      2       96  static_synapse    2.649  0.2750
      2       93  static_synapse    1.472  0.1540
      2       95  static_synapse    2.875  0.1940
     ⋮        ⋮               ⋮        ⋮       ⋮ 
     89       98  static_synapse    1.253  0.1500
     89       92  static_synapse    1.244  0.1550
     89       91  static_synapse    1.675  0.3420
     89      101  static_synapse    2.978  0.3730
     90      100  static_synapse    1.774  0.1620
     90       96  static_synapse    1.722  0.1420
     90       95  static_synapse    2.052  0.1870
     90       91  static_synapse    2.416  0.3850
     90       94  static_synapse    1.411  0.2600
     90       98  static_synapse    1.877  0.3650
     90       92  static_synapse    2.911  0.1960
     90       97  static_synapse    1.934  0.1140
     90       93  static_synapse    2.731  0.1700
     90       99  static_synapse    1.366  0.3690
     90      101  static_synapse    2.418  0.1570
```
```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
      1        3  static_synapse    12.25  0.1840
      1        4  static_synapse    15.81  0.3340
      1        5  static_synapse    15.33  0.1800
      2        3  static_synapse    26.90  0.1790
      2        4  static_synapse    14.45  0.3290
      2        5  static_synapse    14.32  0.1290
```
```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
      1        4  static_synapse    1.000   1.000
      2        4  static_synapse    2.000   1.000
      1        3    stdp_synapse    4.000   1.000
      1        4    stdp_synapse    3.000   1.000
      2        3    stdp_synapse    3.000   1.000
      2        4    stdp_synapse    2.000   1.000
```

heplesser (approved):
@stinebuu Excellent!

stinebuu:
@jougs do you have time to look at this? Let me know if you are too busy, I will then find another reviewer ;)

jougs (approved):
I think the code and the printout really look gorgeous now :smile_cat:

Many thanks!

PR #1840, opened 2020-11-12
Add STDP synapse unit testing

clinssen:
This adds a unit test for the STDP synapse, by comparing the weight evolution due to a barrage of pre- and postsynaptic spikes between NEST and a Python reference implementation.

Test stimuli:
- [x] randomly generated (Poisson distribution) spike trains
- [x] hard-coded edge cases such as simultaneous pre and post spikes

Neuron models:
- [x] iaf_psc_exp
- [x] iaf_cond_exp
- [ ] iaf_psc_exp_ps

Todo:
- [ ] Unit test code cleanup
- [ ] Add "documentation" example

### DIFF
--- a/pynest/nest/tests/test_stdp_synapse.py
+++ b/pynest/nest/tests/test_stdp_synapse.py
@@ -1,0 +1,382 @@
+# -*- coding: utf-8 -*-
+#
+# test_stdp_synapse.py
+#
+# This file is part of NEST.
+#
+# Copyright (C) 2004 The NEST Initiative
+#
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import nest
+import unittest
+from math import exp
+import numpy as np
+
+@nest.ll_api.check_stack
+class STDPSynapseTest(unittest.TestCase):
+    """
+    XXX TODO
+    """
+
+    def init_params(self):
+        self.resolution = 0.1  # [ms]
+        self.simulation_duration = 1e+3 #20. # [ms]
+        self.synapse_model = "stdp_synapse"
+        self.presynaptic_firing_rate = 20.  # [ms^-1]
+        self.postsynaptic_firing_rate = 20.  # [ms^-1]
+        self.tau_pre = 16.8
+        self.tau_post = 33.7
+        self.init_weight = .5
+        self.synapse_parameters = {
+            "synapse_model": self.synapse_model,
+            "receptor_type": 0,
+            "delay": self.dendritic_delay,
+            # STDP constants
+            "lambda": 0.01,
+            "alpha": 0.85,
+            "mu_plus": 0.0,
+            "mu_minus": 0.0,
+            "tau_plus": self.tau_pre,
+            "Wmax": 15.0,
+            "weight": self.init_weight
+        }
+        self.neuron_parameters = {
+            "tau_minus": self.tau_post,
+        }
+        '''self.hardcoded_trains_length = 15
+        self.hardcoded_pre_times = (1, 5, 6, 7, 9, 11, 12, 13)
+        self.hardcoded_post_times = (2, 3, 4, 8, 9, 10, 12)'''
+        self.hardcoded_trains_length = 15
+        self.hardcoded_pre_times = np.array([1, 5, 6, 7, 9, 11, 12, 13.])
+        self.hardcoded_post_times = np.array([2, 3, 4, 8, 9, 10, 12.])
+
+    def do_nest_simulation_and_compare_to_reproduced_weight(self, fname_snip):
+        if True:
+            pre_spikes, post_spikes, t_weight_by_nest, weight_by_nest = self.do_the_nest_simulation()
+
+            import matplotlib.pyplot as plt
+
+            fig, ax = plt.subplots(nrows=3)
+            ax1, ax3, ax2 = ax
+
+            n_spikes = len(pre_spikes)
+            for i in range(n_spikes):
+                ax1.plot(2 * [pre_spikes[i]], [0, 1], linewidth=2, color="blue", alpha=.4)
+
+            n_spikes = len(post_spikes)
+            for i in range(n_spikes):
+                ax3.plot(2 * [post_spikes[i]], [0, 1], linewidth=2, color="red", alpha=.4)
+
+            ax2.plot(t_weight_by_nest, weight_by_nest, marker="o", label="nestml")
+            #ax2.plot(t_hist, w_hist_ref, linestyle="--", marker="x", label="ref")
+
+            ax2.set_xlabel("Time [ms]")
+            ax1.set_ylabel("Pre spikes")
+            ax3.set_ylabel("Post spikes")
+            ax2.set_ylabel("w")
+            ax2.legend()
+            for _ax in ax:
+                _ax.grid(which="major", axis="both")
+                _ax.grid(which="minor", axis="x", linestyle=":", alpha=.4)
+                _ax.minorticks_on()
+                _ax.set_xlim(0., self.simulation_duration)
+            fig.savefig("/tmp/nest_stdp_synapse_test" + fname_snip + ".png", dpi=300)
+
+            t_weight_reproduced_independently, weight_reproduced_independently = self.reproduce_weight_drift(
+                pre_spikes, post_spikes,
+                self.init_weight,
+                fname_snip=fname_snip)
+
+
+            # ``weight_by_nest`` containts only weight values at pre spike times, ``weight_reproduced_independently`` contains the weight at pre *and* post times: check that weights are equal only for pre spike times
+            assert len(weight_by_nest) > 0
+            for idx_pre_spike_nest, t_pre_spike_nest in enumerate(t_weight_by_nest):
+                idx_pre_spike_reproduced_independently = np.argmin((t_pre_spike_nest - t_weight_reproduced_independently)**2)
+                np.testing.assert_allclose(t_pre_spike_nest, t_weight_reproduced_independently[idx_pre_spike_reproduced_independently])
+                print("testing equal t = " + str(t_pre_spike_nest))
+                print("\tweight_by_nest = " + str(weight_by_nest[idx_pre_spike_nest]))
+                print("\tweight_reproduced_independently = " + str(weight_reproduced_independently[idx_pre_spike_reproduced_independently]))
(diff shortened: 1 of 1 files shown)


clinssen:
``test_stdp_multiplicity`` has been fixed. The condition for test failure was not appropriate and rewritten. Performance is much better now. The following figure shows how, for smaller and smaller timestep, the final obtained weights of a "normal" versus a "precise" single STDP synapse converge.

**Before (current master):**

image

**After (code in this PR):**

image

models/stdp_synapse.h
```diff
@@ -217,7 +217,7 @@ inline void
 stdp_synapse< targetidentifierT >::send( Event& e, thread t, const CommonSynapseProperties& )
 {
   // synapse STDP depressing/facilitation dynamics
-  const double t_spike = e.get_stamp().get_ms();
+  const double t_spike = e.get_stamp().get_ms() - e.get_offset();
```
heplesser, line 220: @clinssen This change makes me wonder if we have the same error in other synapse models as well.

clinssen:
The failure and fix in ``test_visualization.py`` is completely unrelated, but was apparently triggered by matplotlib figures that were created in the new test, and still open when ``test_visualisation.py`` is run afterwards, which then starts drawing into an existing figure.

models/stdp_synapse.h
clinssen: Yes, it does:
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/clopath_synapse.h#L215
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/ht_synapse.h#L189
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/jonke_synapse.h#L300
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/quantal_stp_synapse.h#L195
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_dopamine_synapse.h#L529
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_nn_pre_centered_synapse.h#L235
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_nn_restr_synapse.h#L229
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_nn_symm_synapse.h#L231
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_pl_synapse_hom.h#L244
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_synapse_hom.h#L281
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/stdp_triplet_synapse.h#L243
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/tsodyks2_synapse.h#L209
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/tsodyks_synapse_hom.h#L256
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/urbanczik_synapse.h#L207
https://github.com/nest/nest-simulator/blob/720f4c55fec83b2925792dcd936e2d18eb076938/models/vogels_sprekeler_synapse.h#L198

heplesser: @clinssen So this means that until now, we have not taken spike time offsets into account in plasticity mechanisms, doesn't it? It does not have any effect on the actual timing of the spike transmitted, since the offset information is passed on to the receiving neuron in the Event object. Could you create a follow-up issue for this? We should involve @abigailm in the discussion.

pynest/nest/tests/test_stdp_multiplicity.py
```diff
@@ -194,33 +200,51 @@ def run_protocol(self, pre_post_shift):
...
+        post_weights = self.run_protocol(pre_post_shift=pre_post_shift)
         w_plain = np.array(post_weights['parrot'])
         w_precise = np.array(post_weights['parrot_ps'])
 
-        assert all(w_plain == w_plain[0]), 'Plain weights differ'
-        dw = w_precise - w_plain
-        dwrel = dw[1:] / dw[:-1]
-        assert all(np.round(dwrel, decimals=3) ==
-                   0.5), 'Precise weights do not converge.'
+        assert all(w_plain == w_plain[0]), 'Plain weights should be independent of timestep, but differ!'
```
heplesser, line 237: I find the error message confusing: "independent" on the one side and "differ" on the other—requires some mental gymnastics to connect the two.

pynest/nest/tests/test_stdp_multiplicity.py
```diff
@@ -194,33 +200,51 @@ def run_protocol(self, pre_post_shift):
...
+        abs_err = np.abs(w_precise - w_plain)
+        assert abs_err[-1] < max_abs_err
+        assert np.all(np.diff(abs_err) < 1), 'Error should decrease with smaller timestep!'
```
heplesser, line 240: Should this be `< 0` then?

pynest/nest/tests/test_stdp_synapse.py
```diff
@@ -0,0 +1,360 @@
...
+import unittest
+from math import exp
+import numpy as np
+
+try:
+    import matplotlib as mpl
+    import matplotlib.pyplot as plt
+    DEBUG_PLOTS = True
+except Exception:
+    DEBUG_PLOTS = False
```
heplesser, line 32: I am not sure that producing plots just because matplotlib is available is a good idea. We do not need plots when running CI, only when debugging. Wouldn't it be better to either make this switch manual (by editing the file)?

See also the "plotting possible" test in `test_visualisation`.

pynest/nest/tests/test_stdp_synapse.py
```diff
@@ -0,0 +1,360 @@
...
+            "mu_plus": 0.0,
+            "mu_minus": 0.0,
+            "tau_plus": self.tau_pre,
+            "Wmax": 15.0,
+            "weight": self.init_weight
+        }
+        self.neuron_parameters = {
+            "tau_minus": self.tau_post,
+        }
+        self.hardcoded_trains_length = 15
```
heplesser, line 69: Is this the sum of the pre- and post-lengths? If yes, why not compute it from the arrays. Otherwise, comment.

pynest/nest/tests/test_stdp_synapse.py
```diff
@@ -0,0 +1,360 @@
...
+        self.hardcoded_pre_times = np.array([1, 5, 6, 7, 9, 11, 12, 13.])
```
heplesser, line 70: I am not really happy about having just one entry with a decimal point to make it a float array. I'd either give all a decimal point or add a `dtype=float` explicitly.

pynest/nest/tests/test_stdp_synapse.py
```diff
@@ -0,0 +1,360 @@
...
+        nest.SetKernelStatus({'resolution': self.resolution})
+
+        neurons = nest.Create(
+            self.nest_neuron_model,
+            2,
+            params=self.neuron_parameters)
+        presynaptic_neuron = neurons[0]
+        postsynaptic_neuron = neurons[1]
+
+        generators = nest.Create(
```
heplesser, line 114: You can use tuple unpacking here directly.

pynest/nest/tests/test_stdp_synapse.py
```diff
@@ -0,0 +1,360 @@
...
+
+        # While the random sequences, fairly long, would supposedly
+        # reveal small differences in the weight change between NEST
+        # and ours, some low-probability events (say, coinciding
+        # spikes) can well not have occured. To generate and
+        # test every possible combination of pre/post precedence, we
+        # append some hardcoded spike sequences:
+        # pre: 1       5 6 7   9    11 12 13
+        # post:  2 3 4       8 9 10    12
+        (
```
heplesser, line 135: The python magic beginning here is rather too tough for my little head. Are you somehow shifting spike times along the time axis? Why is that necessary, and couldn't you operate directly on the numpy arrays instead of the acrobatics here?

pynest/nest/tests/test_stdp_synapse.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def test_stdp_synapse(self):
+        self.dendritic_delay = float('nan')
+        self.init_params()
+        for self.dendritic_delay in [1., self.resolution]:
+            self.init_params()
+            for self.nest_neuron_model in ["iaf_psc_exp",
+                                           "iaf_cond_exp",
+                                           "iaf_psc_exp_ps"]:
+                fname_snip = "_[nest_neuron_mdl=" + self.nest_neuron_model + "]"
```
heplesser, line 345: What is this?

heplesser (changes requested):
@clinssen This looks quite good, but some of the code should be simplified or better documented; tests need to be comprehensible ;).

pynest/nest/tests/test_stdp_multiplicity.py
clinssen: Replaced with 'Plain weights should be independent of timestep!'

heplesser:
@clinssen Ping :)

pynest/nest/tests/test_stdp_synapse.py
clinssen: A snippet inserted into the filename when saving debug plots to disk.

pynest/nest/tests/test_stdp_synapse.py
clinssen: The "plotting possible" test in ``test_visualization`` does essentially the same thing. matplotlib is not installed on CI, as far as I understood, that is by design.

clinssen:
I removed all changes related to support for precise spike timing in synapses, and test coverage for precise neurons. This feature discussion will be continued in #2035.

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -0,0 +1,346 @@
...
+        self.dendritic_delay = float('nan')
+        self.init_params()
+        for self.dendritic_delay in [1., self.resolution]:
+            self.init_params()
+            for self.nest_neuron_model in ["iaf_psc_exp", "iaf_cond_exp"]:
+                fname_snip = "_[nest_neuron_mdl=" + self.nest_neuron_model + "]"
+                self.do_nest_simulation_and_compare_to_reproduced_weight(fname_snip=fname_snip)
+
+
+def suite():
```
heplesser, line 335: Since we now use pytest, anything from here on down is no longer needed.

testsuite/pytests/test_stdp_nn_synapses.py
```diff
@@ -66,6 +66,18 @@ def setUp(self):
             "tau_minus": 33.7
         }
 
+        # While the random sequences, fairly long, would supposedly
```
heplesser, line 69: Could you shorten this comment, just something like "Hard-coded spike sequences testing every possible combination of pre/post precedence"? Although I am not 100% sure what is meant by "precedence" here—just "order"?

heplesser (changes requested):
@clinssen Just a few more comments.

PR #1839, opened 2020-11-12
Update NumPy installation on Mac OS X Travis build

lekshmideepu:
Fixing the issues with cython,numpy and matplotlib on OsX under Travis after the upgrade of Python to 3.9.0
Fix #1822

PR #1842, opened 2020-11-12
Softer checks for compiler warnings in CI

heplesser:
The recently introduced check for the number of compiler warnings appears too strict, since in particular the inclusion of external libraries can lead to large numbers of warnings with the exact number depending on details of header file inclusion. This leads to failing CI tests for rather obscure reasons and is thus counterproductive.

This PR addresses this in two ways:
1. The number of warnings is only checked when building without external dependencies. In this case we have 8 warnings at present; the test now accepts any number of warnings <= 8.
1. NEST is built (on Travis) with warnings=OFF if built with MPI, MUSIC or libneurosim as these cases trigger a large number of warnings, leading to unnecessarily bloated build logs.

@lekshmideepu This will solve the problem causing #5437.5 to fail for PR #1839.

### DIFF
--- a/extras/parse_travis_log.py
+++ b/extras/parse_travis_log.py
@@ -336,7 +336,10 @@ def msg_summary_pep8(log_filename, msg_pep8_section_start,
 
 def makebuild_summary(log_filename, msg_make_section_start,
                       msg_make_section_end):
-    """Read the NEST Travis CI build log file and return the number of build
+    """
+    Parse NEST Travis CI build log and summarize result.
+
+    Read the NEST Travis CI build log file and return the number of build
     error and warning messages as well as dictionaries summarizing their
     occurrences.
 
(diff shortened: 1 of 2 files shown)


jougs:
@heplesser: thanks for the fix.

We discussed this issue extensively during the hackathon today and came to a similar but still somewhat different conclusion. The approach we agreed on is to adapt the detection of warnings even if using external libraries so that we only catch those coming from the NEST source code proper. I have already started to implement this. I will now rebase my work onto your branch, so your efforts are not completely lost. Once I'm ready, I'll be sending a PR towards your branch.

In order to prevent a merge of this PR, I'm converting to draft for now.

heplesser:
@jougs I did wonder if you had not discussed it already ;). Turing warnings off for the full build, the log shrinks from 30000 to 10000 lines and test pass, although one does get the slightly confusing message
```
Warnings: 289 (1000000000 expected)
```
BTW, I accidentally restarted the .5 build on Travis. Until that is finished, you can see the result in the build against my repo (https://travis-ci.org/github/heplesser/nest-simulator/jobs/743277720).

heplesser:
Closing this one, #1852 is a much better solution.

Issue #1612, opened 2020-05-20
Collect dependencies in central file 


sarakonradi:
As @terhorstd pointed out in #1515, we should have a central file where dependencies are collected, as we do for the documentation (conda environment). 

The "Dependencies" section in ``doc/installation/linux_install.rst`` could then have an additional sentence like "For the full list of requirements, check here".

terhorstd:
This is also interesting in the context of #1589.

terhorstd:
The development dependencies  (#1757) maybe are different from the build dependencies.

terhorstd:
For the documentation we may consider loading the `environment.yml` in `conf.py` and exposing the dependencies as variables. Then the maintenance of version numbers in the documentation becomes much easier. Have a look at this [stackoverflow post](https://stackoverflow.com/a/33845358/14085246).

heplesser:
Using one central `environment.yml` file seems to make a lot of sense. We should also use it in our CI build scripts then, both to avoid duplicity and to test the file itself. One point I wonder about is about the relation between `pip` and `conda` here. I know conda can handle `environment.yml` files that have a separate "pip" section for packages not available from conda, but I don't know how it is the other way around. It would be good if conda users could use the `environment.yml` file as well.

heplesser:
It seems @jougs pretty much just answered my question in #1843.

jougs:
I think that it could be a bit confusing to have a set of requirement files in each location where we need them and expect users to infer their meaning based on location, i.e.
* top-level `environment.yml` and `requirements.txt` for runtime requirements
* `extras/environment.yml` and `extras/requirements.txt` for build requirements
* `doc/environment.yml` and `doc/requirements.txt` for documentation requirements
* `testsuite/environment.yml` and `testsuite/requirements.txt` for testing requirements 

Alternatively, we could have
* `extras/conda-environment-nest-{doc,test,build,run}.yml`
* `extras/pyenv-requirements-{doc,test,build,run}.txt`

heplesser:
I quite like @jougs' alternative suggestion with respective files under `extra`.

jougs:
In a video call with @heplesser today, we came to the conclusion that the requirements for building, testing and running NEST are sufficiently similar and lightweight that those can be in a single requirements file, while the ones for doc building are much different and considerably heavier. If we wanted to increase maintainability and reduce the risk of colliding with an infamous thermostat, we could thus get away with only

[...]

PR #1843, opened 2020-11-12
Fix doc buildsystem

### LINKED ISSUE #1660: Integrate Sphinx documentation generation with CMake
Currently, the Sphinx documentation is built using a standalone `Makefile` in the `doc` directory. This has several downsides:
1. Running that `Makefile` requires to `cd` into the `doc` directory, which is at least inconvenient.
1. The `Makefile` for the documentation is hard to find, as it is somewhat uncommon to have two parallel build systems.
1. The extracted documentation ends up in the source directory instead of in the build directory, where it should actually be. Here are some points why this is bad:
   - It requires keeping a list of build artifacts in the `.gitignore` file. If that's incomplete or outdated, `git status` and `git checkout` will get confused.
   - It complicates refactorings, as many search matches will be in generated files that should not need to be touched.
   - Although `git clean -dX` will remove the generated files in most cases, it might fail and lead to further confusion.

I thus propose to integrate the documentation extraction workflow under a new `doc` or `sphinx` target in the regular CMake-based build system.

jougs:
This fixes #1660 by removing the custom `Makefile` for building the Sphinx documentation and integrates the documentation build process into the standard CMake build under the target `html`. I took the liberty to also update some pertaining files on the way. I still need to check if the changes still result in a successful build on *Read the Docs*.


steffengraber (commented):
The PyNEST examples are not created.
This is because the `.rst` files created by `sphinx_gallery` are no longer part of the source directory.  
So the links in `.../doc/examples/index.rst` can not work anymore.

For the Models Directoy the same applies as for the `PyNEST` examples. The directory with the generated `.rst` files is not in the source directory. `Sphinx` cannot find it and generate the html files.

How about copying the `doc/` directory with `cmake` into the build directory and using this as source for `sphinx-build`?

jougs:
@steffengraber: I don't think I fully understand what you are saying (i.e. what do you mean by "not created"?).

Running `make html` locally, I see all PyNEST examples in `<build_dir>/doc/html/examples/index.html` just fine. The same is true for the model directory in `<build_dir>/doc/html/models/index.html`.

My recent commits also make the new `conf.py` work well on Read the Docs, and to me the [result there also looks good](https://nest-simulator-jougs.readthedocs.io/en/fix_doc_buildsystem).

Can you please be a bit more specific about the problem you are referring to?

steffengraber:
In my case, both folders remained empty, even after several attempts. I will test your updates immediately.

steffengraber:
Even after the updates I have no success and the two folders remains empty.
I tried it with a `conda` environment and without. Maybe others should try it as well. 

@sarakonradi @terhorstd Could you test this?

jougs:
That's really strange...

I get the following message at the end of the `make html` run, maybe this is related to your problem?
```
Sphinx-gallery gallery_conf["plot_gallery"] was False, so no examples were executed.
```

jougs:
@steffengraber: with my recent changes, Sphinx should print `doc_path`, `root_path` and `build_path` right at the start. Could you please check if they contain reasonable values?

steffengraber:
@jougs The paths are looking good:

```
build_path <myprefix>/build-nest
root_path <myprefix>/nest-simulator
doc_path <myprefix>/nest-simulator/doc
```

I will try again later in a clean system. The problem seems to be with me, if everything is ok on your system. Which `python` and `sphinx` versions do you have in use?

jougs:
Alright. I'm running Python 3.8.6 and sphinx-build 3.2.1, both installed via `apt` from Ubuntu 20.10.

jougs:
Any updates on this?

sarakonradi:
Hi! I tested this using Ubuntu 18.04. After running CMake in the ``nest-build`` directory, I was unable to ``make html``. Sphinx threw ``KeyError: 'READTHEDOCS'``. 

Anyone else experiencing this?

jougs:
@sarakonradi: c419b2c should fix the problem you [experienced](https://github.com/nest/nest-simulator/pull/1843#issuecomment-738037384). Please let me know if this solves the problem and if building the documentation works for you.

sarakonradi:
@jougs Thanks for fixing the problem! Tested this with the ``nest-doc`` conda environment and was able to build the documentation.

The [auto_examples](https://github.com/nest/nest-simulator/blob/master/doc/examples/index.rst) were generated but not displayed in the PyNEST examples index when rendered:

pynest_examples_index

jougs:
@sarakonradi: I suspect a problem with conda here. As already pointed out by @terhorstd elsewhere, the `doc` directory currently has a standard Python `requirements.txt` file and additionally the conda  environment file `nest_doc_conda_env.yml`. Naturally, these file are badly out of sync (and probably always will be).

What I did was to run `pip3 install -r doc/requirements.txt` after which `make html` worked flawlessly for my out-of-tree configuration. More or less the same (but in-tree) is done by RtD and also leads to [nicely extracted examples](https://nest-simulator-jougs.readthedocs.io/en/fix_doc_buildsystem/examples/index.html) there.

To reduce confusion and solve this issue, I propose to simply discouraging the use of conda for building the docs by removing the corresponding environment file from the repository. Alternatively, someone who is in the know about that technology needs to look into updating that file and write instructions on how to use that workflow. I myself have never used conda and don't feel confident enough to do this.

heplesser:
As an active conda user, I'd prefer to keep the conda environment file around. We should try to find someone who could dig into these issues, I will not have time for a while.

jougs:
After an enlightening video call with @terhorstd, I have now
* updated `doc/requirements.txt` to provide a complete and up-to-date list of dependencies
* renamed the conda environment file `doc/nest_doc_conda_env.yml` to the standard name `doc/environment.yml`
* updated the conda environment file to use the standard requirements file to avoid redundancy
* updated `.readthedocs.yml` to use the conda environment for building the documentation
* updated `doc/README.rst` to reflect all these changes.

It seems that [the documentation on Read the Docs](https://nest-simulator-jougs.readthedocs.io/en/fix_doc_buildsystem) still works nicely with these changes in place.

@sarakonradi, @steffengraber: can you please confirm that the problems you saw earlier are now also gone? Please make sure to recreate/update the conda environment along the lines of what is written in the new `doc/README.rst`.

doc/environment.yml
```diff
@@ -0,0 +1,6 @@
+name: nest-doc
+dependencies:
+  - python=3.6
```
heplesser, line 3: Why do you require Py 3.6 specifically?

jougs: This is what we had in the previous incarnation of the `readthedocs.yml` file because we use 3.6 features in the documentation extraction pipeline. As we could not express a _minimal_ requirement in that file, I put in the fixed version. I'm open to replacing the `=` by `>=` now and just go with whatever the local conda installation provides.

doc/README.rst
```diff
@@ -3,52 +3,54 @@
...
+This folder contains the static parts of the documentation for NEST,
+as well as the infrastructure and tools for extracting documentation
+dynamically from the source code and building a unified HTML site
+using Sphinx.
 
-Build the docs locally
------------------------
-
-You will need the NEST source code in your project::
+For each of the releases, the fullly built documentation is available
```
sarakonradi, line 11:
```suggestion
For each of the releases, the fullly built documentation is available
```
```suggestion
For each of the releases, the fully built documentation is available
```

doc/README.rst
```diff
@@ -3,52 +3,54 @@
...
+at https://nest-simulator.readthedocs.io/
 
+Build the user documentation locally
+------------------------------------
 
-    cd nest-simulator/doc/
+If you intend to work on the documentation, or if you want to obtain a
+local version of it for other reasons, you can build the documentation
+by simply running
```
sarakonradi, line 20:
```suggestion

::

```

doc/README.rst
```diff
@@ -3,52 +3,54 @@
...
- nbsphinx
- breathe
- sphinx-gallery
- ipython
- mock
+    make html
 
+in the build directory of NEST (i.e. the directory where you ran
+``cmake``). To install the dependencies for building the
+documentation, you can either directly run
```
sarakonradi, line 26:
```suggestion

::

```

doc/README.rst
```diff
@@ -3,52 +3,54 @@
...
 
-You can now build the docs using this command (make sure you are in the doc folder)::
+    pip install -r <nest_source_dir>/doc/requirements.txt
 
-    make html
+or alternatively install the dependencies into a conda environment by
+running
```
sarakonradi, line 31:
```suggestion

::

```

sarakonradi (changes requested):
Thanks for the extensive series of updates!

Upon testing this, I can confirm that it works nicely for me. The PyNEST examples are also displayed.

I have added a few minor comments regarding the ``doc/README.rst``. We can update the [User-level documentation workflow](https://github.com/nest/nest-simulator/blob/master/doc/documentation_workflow/user_documentation_workflow.rst) accordingly in a follow-up PR (see #1896).

steffengraber:
My problem with creating the examples and the models still exists. Since everything is ok with Sarah, I suspect the error with me.  Will probably have to set up everything again from scratch.

jougs:
@sarakonradi: thanks for testing again and your suggestions. I've applied them all.
@steffengraber: I also assume a problem on your side. Sorry for breaking things for you.

@terhorstd: could you also check, or directly merge? Thanks!

doc/README.rst
```diff
@@ -3,52 +3,60 @@
...
+    pip install -r <nest_source_dir>/doc/requirements.txt
 
-You can now build the docs using this command (make sure you are in the doc folder)::
+or alternatively install the dependencies into a conda environment by
+running
 
-    make html
+::
 
+    conda env create -f <nest_source_dir>/doc/environment.yml -p conda
```
jessica-mitchell, line 38: While this conda command works, I have to activate the environment from the source directory or use the absolute path name. This isn't a big deal but all my other conda environments are stored along a different path and can be activated from anywhere via their ENVNAME. Is this a potential area of confusion for users?

jessica-mitchell (commented):
I tested the build locally and on RtD and everything seems to be working fine for me. I just have one minor comment regarding the conda activation (see comment inline). Once that's addressed I can approve and the PR can be merged imo.

doc/README.rst
jougs: Hi @jessica-mitchell! Nice to see you around :-)

If you leave out the `-p conda` in the line above, it will create the environment *globally* with the name `nest-doc` (c.f. [here](https://github.com/nest/nest-simulator/pull/1843/files#diff-178af50da4e5ad380ede922c14d44bc72486a29805cac3a1f56dfa0ae7aef21aR1)). I just thought that co~rona~nda-skeptics like me might prefer to have it in an obvious local directory, so I can be sure it'd just go away whenever I delete the build directory.

Would it help to have both variants explained in the `README.md`?

heplesser: I agree with @jessica-mitchell here and would remove the `-p conda`, since it quite strongly breaks with conda logic and user experience. The default for conda is a user-local installation, i.e., the complete conda installation lives in the users home directory.  Global installations for all users of a system are rare for all I know. So installing without `-p ...` will install for the user only, and then automatically in the standard location for environments that allows easy activation. So we should drop the `-p conda`.

doc/environment.yml
```diff
@@ -0,0 +1,6 @@
+name: nest-doc
```
heplesser, line 1: Since we may (will) have different conda environment files included with NEST (for  working on docs, for running NEST, ...), we should have a more specific file name for this yml file, e.g. `nest-doc-conda-env.yml`.  This would also require corresponding changes in the readme and possibly other places.

doc/README.rst
jougs, line 38: Alright.
```suggestion
    conda env create -f <nest_source_dir>/doc/environment.yml
```

doc/environment.yml
jougs: I have discussed this with @terhorstd and our conclusion was that we'd rather stick with the default name. However, I agree that it could be a bit confusing to have a set of those files in each location where we need them and expect users to infer their meaning based on location, i.e.
* top-level `environment.yml` and `requirements.txt` for runtime requirements
* `extras/environment.yml` and `extras/requirements.txt` for build requirements
* `doc/environment.yml` and `doc/requirements.txt` for documentation requirements
* `testsuite/environment.yml` and `testsuite/requirements.txt` for testing requirements 

Alternatively, we could have
* `extras/conda-environment-nest-{doc,test,build,run}.yml`
* `extras/pyenv-requirements-{doc,test,build,run}.txt`

That said, I'd rather prefer to leave this for discussion/fixing in the context #1612. I've cross-posted this comment there, so we have everything in one place.

doc/README.rst
jougs: Fixed in b855286.

jessica-mitchell (approved):
thanks @jougs looks good to me

.gitignore
```diff
@@ -1,67 +1,55 @@
...
-pynest/pynestkernel.cpp
+__pycache__/
+build/
+install/
+reports/
 
-*.o
-*.lo
+# compilation artefacts
+*.cxx
```
heplesser, line 8: I think excluding *.cxx files in a C++ project is dangerous, so I suggest to remove this. What did it apply to?

heplesser (changes requested):
I am requesting changes here because of my concerns about gitignoring `cxx` files.

Another question in this context: In the past, all autogenerated rst-files went flat into doc. Are they now put into a separate subdirectory that then can be easily gitignored?

jougs:
@sarakonradi, @steffengraber: Can you please check out the newest version? I think I found and fixed the problem in f74aa6d.

.gitignore
jougs: This is meant to exclude the C++ file generated by Cython and I agree that the pattern is way too unspecific. What about the following change?
```suggestion
pynest/pynestkernel.cxx
```

jougs:
As a bonus, I've just added installation of the user documentation. This is essential in the context of #1904 and I think it fits quite well within this PR. I would be grateful if you could give this one certain priority in your review queue. Thanks!

.gitignore
heplesser: @jougs I like you suggestion to give the full path!

steffengraber:
No change on my side. That's what I did on a fresh installed Ubuntu 20.04 Virtual Box image: 
```
sudo apt update && sudo apt upgrade
sudo apt install git
git clone https://github.com/jougs/nest-simulator.git
cd nest-simulator/
git checkout fix_doc_buildsystem 
sudo apt install python3-pip
pip3 install -r doc/requirements.txt
# maybe logout/login for the env 
mkdir ~/nest-build
mkdir ~/nest-install
cd ~/nest-build/
cmake -DCMAKE_INSTALL_PREFIX:PATH=../nest-install ../nest-simulator/
make html
```

No example and models are created as html in the doc folder.

heplesser:
I just tried this on my M1 machine in a fresh conda environment based on `environment.yml`. In the `build/doc/html/examples` directory I only find `index.html`, `cortical_microcircuit.html` and `cortical_microciruit_index.html`. I do not see a `html/models` directory, only a `models` directory containing many rst files.

At the same time, I find these and many more similar warnings in the Sphinx output:

```
/Users/plesser/NEST/src/doc/examples/index.rst:4: WARNING: toctree contains reference to nonexisting document 'auto_examples/one_neuron'
/Users/plesser/NEST/src/doc/examples/index.rst:4: WARNING: toctree contains reference to nonexisting document 'auto_examples/one_neuron_with_noise'
/Users/plesser/NEST/src/doc/examples/index.rst:4: WARNING: toctree contains reference to nonexisting document 'auto_examples/twoneurons'
```

jessica-mitchell:
It still works for me. I created anew conda env as described in the README as well as an older conda environment I had.

heplesser:
I don't know if this helps, but comparing the output of `make html` on @jessica-mitchell's machine (examples are in place) and mine (no examples), one difference I found was the following

### @jessica-mitchell 
```
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 5 changed, 0 removed
reading sources... [ 20%] auto_examples/index
reading sources... [ 40%] guides/spatial/guide_spatially_structured_networks
reading sources... [ 60%] guides/spatial/pynest
reading sources... [ 80%] installation/index
reading sources... [100%] ref_material/pynest_apis
```

### @heplesser 
```
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 8 changed, 0 removed
reading sources... [ 12%] contents
reading sources... [ 25%] examples/cortical_microcircuit
reading sources... [ 37%] examples/index
reading sources... [ 50%] guides/recording_from_simulations
reading sources... [ 62%] guides/spatial/guide_spatially_structured_networks
reading sources... [ 75%] guides/spatial/pynest
reading sources... [ 87%] installation/index
reading sources... [100%] ref_material/pynest_apis
```

Digging a bit deeper reveals in @jessica-mitchell's log
```
NESTDIR/doc/auto_examples/index.rst:1138: WARNING: image file not readable: auto_examples/microcircuit.png
NESTDIR/doc/auto_examples/index.rst:1138: WARNING: image file not readable: auto_examples/microcircuit.png
NESTDIR/doc/auto_examples/index.rst:1140: WARNING: image file not readable: auto_examples/reference_data/raster_plot.png
NESTDIR/doc/auto_examples/index.rst:1140: WARNING: image file not readable: auto_examples/reference_data/raster_plot.png
NESTDIR/doc/auto_examples/index.rst:1142: WARNING: image file not readable: auto_examples/reference_data/box_plot.png
NESTDIR/doc/auto_examples/index.rst:1142: WARNING: image file not readable: auto_examples/reference_data/box_plot.png
```

and in mine

```
NESTDIR/doc/examples/index.rst:23: WARNING: toctree contains reference to nonexisting document 'auto_examples/clopath_synapse_small_network'
NESTDIR/doc/examples/index.rst:23: WARNING: toctree contains reference to nonexisting document 'auto_examples/clopath_synapse_spike_pairing'
NESTDIR/doc/examples/index.rst:23: WARNING: toctree contains reference to nonexisting document 'auto_examples/evaluate_quantal_stp_synapse'
...
```

So something seems to be wrong about the `auto_examples` extraction on my system.

sarakonradi:
@heplesser I am experiencing a similar issue when testing this with a fresh conda environment. 

Examples:
```
NESTDIR/doc/examples/index.rst:4: WARNING: toctree contains reference to nonexisting document 'auto_examples/one_neuron'
NESTDIR/doc/examples/index.rst:4: WARNING: toctree contains reference to nonexisting document 'auto_examples/one_neuron_with_noise'
NESTDIR/doc/examples/index.rst:4: WARNING: toctree contains reference to nonexisting document 'auto_examples/twoneurons'
...
```

Models:
```
NESTDIR/doc/contents.rst:8: WARNING: toctree contains reference to nonexisting document 'models/index'
```

jougs:
I have meanwhile found the root of the problem and will create a fix asap. Sorry for making you all test and fail.

jougs:
The recent commits should fix the remaining issues by
* containing all user documentation content (but not the files for *building* the documentation) in `doc/userdoc`
* copying this directory to the build directory in case an out-of-tree build is issued
* Fix and simplify paths (again!), so that everything just works<sup>TM</sup>
* running `sphinx-build` with `PYTHONDONTWRITEBYTECODE=1`, so it does not taint the source directory with a `__pycache__` directory

With these changes in place, one should be able to run `make html` anywhere in either the source or build directory and get the full documentation built into `topdir/doc/userdoc/html`, with `topdir` being the top-level source or build directory. `make install` should install the documentation to `<nest_install_dir>/share/doc/nest/html`.

Because I also moved the SLI user documentation to its own directory (`doc/slidoc`) along the way, it might be that I lost the SLI index page along the way. On the other side, I'm not even sure that was ever linked from somewhere in the user documentation and we need to revisit the generation and presentation of SLI documentation anyway as one of the next steps.

heplesser:
@jougs Very nice, seems to work out-of-tree on my test machine (with a currently hand-crafted conda yml file to set up a build nest & doc environment with as little PIP-involvement as possible). One thing we may want to mention in the installation instructions: If you want to run `make install` at some point, you probably want to run
```
make -jX all
make doc
make install
```
Otherwise, `make install` after `make doc` will trigger a single-process `make all`. 

A general question: building the documentation does not require a running NEST instance to generate figures, does it?

jessica-mitchell:
This works for me except with regard to " one should be able to run `make html` anywhere in either the source of build directory" 
 `make html` works in the build folder where I ran `cmake` and in the build/doc/ folder that is created after `cmake`
Maybe I misunderstand something, but I also dont see that as a problem.
Also, in the old set up, I used to be able to run `make clean` in the doc folder which would remove all the built html files. This was handy when testing output in docs to ensure that I wasnt looking at outdated output. Is there a possible similar thing to do with this setup?
@heplesser regarding your question, no it shouldnt. I have a conda environment that is nest-free, and sphinx runs fine

steffengraber (approved):
Works for me out of the box now. Nice!
I used a fresh  installed VirtualBox image with Ubuntu 20.04, environment installed with pip.

This now also requires an adjustment of the documentation workflow, but this has to be handled with in another PR.

heplesser:
@jessica-mitchell 

> `make html` works in the build folder where I ran `cmake` and in the build/doc/ folder that is created after `cmake`
> Maybe I misunderstand something, but I also dont see that as a problem.

This is indeed expected and unavoidable behaviour: To be able to at all work in a build directory, we need the information about paths etc collected by cmake. This may require some updates to the doc workflow documentation.

heplesser:
@jougs How do you want to coordinate this with #1612? Should I send you a PR with a proposal for conda-ish yml files?

sarakonradi:
The examples and model directory are working for me, but the color roles are not. I tested this again after moving ``colorize.rst`` into ``doc/userdoc``, and the problem seems to be fixed. Anyone else experiencing this?

doc/README.rst
```diff
@@ -3,52 +3,60 @@
...
 
-You will need the NEST source code in your project::
+If you intend to work on the documentation, or if you want to obtain a
+local version of it for other reasons, you can build the documentation
+by simply running
 
+::
 
-    cd nest-simulator/doc/
+    make html
```
heplesser, line 23: This won't work any more, one now needs to run cmake first.

doc/README.rst
```diff
@@ -3,52 +3,60 @@
...
+running
 
-    make html
+::
 
+    conda env create -f <nest_source_dir>/doc/environment.yml
+    conda activate conda/
 
-Once this complete, you can open `_build/html/index.html` to view the docs!
+Once this complete, you can open `doc/html/index.html` to view the docs.
```
heplesser, line 41:
```suggestion
Once this complete, you can open `doc/userdoc/html/index.html` to view the docs.
```

doc/README.rst
```diff
@@ -3,52 +3,60 @@
...
 
 ::
 
-    xdg-open _build/html/index.html
+    xdg-open doc/html/index.html
```
heplesser, line 45: I think `xdg-open` is rather system specific, on macOS a simple `open` will do. Shouldn't we just tell the users to open it in their browser?

And should we mention the `make install` here?

doc/README.rst
```diff
@@ -3,52 +3,60 @@
...
 
+Developer documentation
+-----------------------
+    
 For **developer documentation**, we use `doxygen <http://doxygen.org/>`__
 comments extensively throughout NEST.
 After installing NEST, you can extract comments from the source code with
 ``make doc`` and a doxygen folder with html files will be generated in the doc
```
heplesser, line 53: Should we rename that target to something like `make dev-doc`?

Might also want to point out where the doc will be located.

doc/README.rst
```diff
@@ -59,56 +67,11 @@ index via the command line
...
-
-   conda deactivate
-   conda remove --name nest_doc --all
+ The command ``helpdesk()`` needs to know which browser to launch in
+ order to display the help pages. The browser is set as an option of
+ helpdesk. Please see the file ``~/.nestrc`` for an example setting
+ firefox as browser.  Please note that the command helpdesk does not
+ work if you have compiled NEST with MPI support, but you have to
+ enter the address of the helpdesk
+ (file://<nest_install_dir>/share/doc/nest/index.html) manually into
```
heplesser, line 76: This is for the overall helpdesk. How does one get from there to SLI? The SLI command index is in `<install_dir>/share/doc/nest/help/helpindex.html`-

doc/conf.py
```diff
@@ -72,16 +68,12 @@
 
     fout.write(mockedmodule)
 
-# The pynestkernel_mock has to be imported after it is created.
-import pynestkernel_mock  # noqa
+import pynestkernel_mock
```
heplesser, line 71:
```suggestion
import pynestkernel_mock  # noqa
```

heplesser (changes requested):
@jougs Very nice! Works, even the doxygen part, which I hadn't run for a very long time. Just the documentation of the documentation workflow now needs a little update. And one detail on which PEP8 choked.

heplesser:
@sarakonradi 
> The examples and model directory are working for me, but the color roles are not. I tested this again after moving `colorize.rst` into `doc/userdoc`, and the problem seems to be fixed. Anyone else experiencing this?

On my system, the front-page headline is (in bright orange)
```
:orange:`Introducing NEST 3.0`
```
but I guess the `:orange:` shouldn't be there. Is that what you meant?

steffengraber:
> Anyone else experiencing this?

@sarakonradi I also found this on my home page. But only there. Should be deleted.

sarakonradi:
> but I guess the ``:orange:`` shouldn't be there. Is that what you meant?

@heplesser Yes, exactly. The problem also appears in the "NEST 3.0: Detailed transition guide". 

Moving ``colorize.rst`` into ``doc/userdoc`` should fix the issue.

heplesser:
> Moving `colorize.rst` into `doc/userdoc` should fix the issue.

Indeed, it does! So @jougs just needs to move it in his branch.

steffengraber:
> > Moving `colorize.rst` into `doc/userdoc` should fix the issue.
> 
> Indeed, it does! So @jougs just needs to move it in his branch.

If this is the only thing it does, and the only two places where it is included, shouldn't we then delete it, and remove the `:orange:`?
Does this only remove the `:orange:`, or does something else happen? 
Do we need it otherwise?

steffengraber:
Why is colorize.rst actually still there? It should be gone - see #1906.

sarakonradi:
> Why is colorize.rst actually still there? It should be gone - see #1906.

@steffengraber ``colorize.rst`` allows the color roles to work, so that we can have the orange subtitle in the welcome page and green text in the detailed transition guide. 

The file was re-added in #1909. Sorry for the confusion...

jougs:
> @jougs How do you want to coordinate this with #1612? Should I send you a PR with a proposal for conda-ish yml files?

What would that PR contain exactly? I thought the idea now is to have the conda `environment.yaml` file just as a shell for the `requirements.txt` file to cater to more people (i.e. plain `pip` users,`pyenv` users, and conda users). Wouldn't a more conda-ish YAML file be counterproductive to this aim?

heplesser:
> What would that PR contain exactly? I thought the idea now is to have the conda `environment.yaml` file just as a shell for the `requirements.txt` file to cater to more people (i.e. plain `pip` users,`pyenv` users, and conda users). Wouldn't a more conda-ish YAML file be counterproductive to this aim?
> 
> I had no intention to fix #1612 at all. My hope was that the doc team would collect these dependencies first and then pour them in `requirements.txt`s. Moving the dependency-files contained in this PR to their new location should then be simple enough.

In my experience, the delegation to `requirements.txt` does not work. On my M1-Mac, installing the PIP requirements actually fails, and even it it would work, it takes a long time because much needs to be built locally, while Conda ships everything pre-built (this may change if PyPi at some point should provide M1-built binary packages). So I think we need "conda-native" environment files which use the pip section only for (small) packages not available via Conda, such as `junitparser`. I think I have a pretty good idea of what needs to go into these Conda environment files and could contribute them to this PR.

doc/README.rst
jougs: Yes, we should. However, I'd like to leave that to #1905.

doc/README.rst
jougs: Agreed. That's why the next lines say:
> in the build directory of NEST (i.e. the directory where you ran ``cmake``).

doc/README.rst
jougs: Fixed in b0f735498.

doc/README.rst
jougs: Fixed in b0f735498.

doc/README.rst
jougs: With the present changes, the SLI helpdesk is actually [not installed anymore](https://github.com/nest/nest-simulator/pull/1843/files/ee754d1254d32a64aeae4c90427ae973bb265ed2#diff-652f011e4f45934b3684d4596670e3847750ec3289b5092d4f8847b1ca6a75edL62-L64). Stay tuned for a follow-up PR, refactoring the build and install for that.

jougs (commented):
Thanks for your reviewing again and the nice words. I think I've fixed or at least commented on everything.

jougs:
Argh!

I only found out now that there's *yet another* documentation "README" in the form of the page [User-level documentation workflow](https://nest-simulator.readthedocs.io/en/latest/documentation_workflow/user_documentation_workflow.html) which is largely redundant with `doc/README.rst`. I suggest to merge the content of `README.md` into the workflow guide, update that and just remove the `README.rst`. Any objections?

### ☠⚡🔪🌩🌀💣‼

doc/README.rst
heplesser: Ah, good! But assuming that readers will be eager an try things as soon as they read them, I'd suggest to move the sentence up. Github does't let me create a multiline suggestion here, but like this:
```
by simply running in the NEST build directory (where you ran ``cmake``)

::
    make html

To install ...

**Comment by heplesser** [comment]
@jougs I am definitely for avoiding redundancy, so your suggestions sounds sensible, but I am not entirely certain which files you are talking about, especially the `README.md` you mention left me confused.

File: doc/README.rst (hunk 9 above)
Comment (jougs) [reply doc/README.rst:24]: Agreed. Fixed in 3a94c83e3.

**Comment by jougs** [comment]
Sorry for causing confusion. What I meant is not that we have literally two `README` files (hence the quotes in my comment), but that we currently have similar/overlapping documentation workflow descriptions in two places:
1. [`doc/README.rst`](https://github.com/jougs/nest-simulator/blob/fix_doc_buildsystem/doc/README.rst)
2. [`doc/userdoc/documentation_workflow/user_documentation_workflow.rst`](https://github.com/jougs/nest-simulator/blob/fix_doc_buildsystem/doc/userdoc/documentation_workflow/user_documentation_workflow.rst)

In my opinion this redundancy should be resolved by first merging the (updated) content of 1. into 2., updating 2. to iron out remaining inconsistencies, and then remove 1., or at least strip it down to just a link to 2. and convert it to MarkDown so it renders more nicely on GitHub. Then happiness.

Could you please also indicate how you feel about my answers to your review comments by either resolving them, or by telling me what's still missing? Thanks!

**Review by heplesser — APPROVED** [review]

**Comment by heplesser** [comment]
@jougs I fully agree with your suggestion on integrating 1 and 2. Keeping a minimal README.md with a link to the user_documentation_workflow document (straight to readthedocs?) seems very sensible to me.

And thanks for all the hard work!

**Comment by sarakonradi** [comment]
Hi! I have two more minor comments:

* The microcircuit image on the welcome page (``doc/userdoc/index.rst``) is broken
* The README link on the microcircuit landing page (``doc/userdoc/examples/cortical_microcircuit_index.rst``) is not working

@jougs Do you think you could have a look? Thanks for the many changes!

**Review by jessica-mitchell — APPROVED** [review]
Thanks @jougs I think this looks good

**Comment by jougs** [comment]
@sarakonradi: I've replaced the `copytreeglob` mechanism in `doc/conf.py` by explicit calls to `shutil.copyfile` in f0d9ffa. This solves the problem with the figures and the invisible `README.rst` at least for me. Can you please re-check? Thanks!

@heplesser, @jessica-mitchell: I've updated the documentation workflow documents and the `README.rst` in df5e4ea (as it was  [discussed](https://github.com/nest/nest-simulator/pull/1843#issuecomment-779820037) [above](https://github.com/nest/nest-simulator/pull/1843#issuecomment-779833759)). Can you please have a look? More changes to the workflow documents are to come in #1926.

**Comment by heplesser** [comment]
@jougs This looks pretty good, I just discovered two glitches:
- If I click on the "README" link on `.../install/share/doc/nest/html/examples/cortical_microcircuit_index.html`, I just get to an empty page (`.../install/share/doc/nest/html/examples/cortical_microcircuit.html`)
- The image on the "User-level documentation workflow" page (``) does not work. I do not understand why this happens because the file is there and I can open it with other software, just not with the browser (neither Safari nor Firefox). I wonder if macOS here is overly protective, so we can probably ignore this.

File: doc/userdoc/documentation_workflow/user_documentation_workflow.rst [hunk 15]
```diff
@@ -44,28 +46,23 @@ workflow below!
...
 
 .. code-block:: bash
 
-    cd </path/to/nest_source>/doc
+    pip3 install -r <nest_source_dir>/doc/requirements.txt
 
-3. Create and activate the environment:
+or alternatively install the dependencies into a conda environment (we
+recommend `miniconda <https://docs.conda.io/en/latest/miniconda>`_ for
+this). After installing coda, you can simply run
```
jessica-mitchell, line 58:
```suggestion
this). After installing conda, you can simply run
```

jessica-mitchell (changes requested):
Just a tiny typo, otherwise looks good. Many thanks @jougs

jessica-mitchell (approved):

sarakonradi (approved):

jougs:
Apparently, something went wrong in the last commits. The rendered version on [Read the Docs](https://nest-simulator.readthedocs.io/en/latest/) is currently empty. I'll investigate and create a fix for that now.

hakonsbm:
@jougs Something went wrong with the `.gitignore` file too. Some of my dot-directories (such as `.settings` for Eclipse and `.vscode` for VS Code) are no longer ignored.

heplesser:
@jougs Things look good on my Mac, building both user and developer docs.

jougs:
@hakonsbm: This is caused by my [removal of `.*` from `.gitignore`](https://github.com/nest/nest-simulator/pull/1843/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947L10). I found that way too unspecific and having it prevented Git from picking up changes to `.readthedocs.yml`, `.travis.yml` and such. Would you please create a PR explicitly adding the things you want ignored to the corresponding section in the new `.gitignore` file? Thanks!

@heplesser: Yes, locally it works well. Unfortunately, the [utterly](https://github.com/nest/nest-simulator/pull/1843/files#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R41) [black](https://github.com/nest/nest-simulator/pull/1843/files#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R42) [path](https://github.com/nest/nest-simulator/pull/1843/files#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R45) [magic](https://github.com/nest/nest-simulator/pull/1843/files#diff-652f011e4f45934b3684d4596670e3847750ec3289b5092d4f8847b1ca6a75edR87-R94) to make Read the Docs, in-tree and out-of-tree documentation builds work from the same `conf.py` seems to be more fragile as I was hoping for. I'm working on a fix right now and will create another PR as soon as I have something.

Sorry for the inconveniences.

PR #1845, opened 2020-11-13
Fix NEST3 transition guide formatting

clinssen:
A few tiny updates, mostly to make the green font colour render properly.

Suggesting @sarakonradi as reviewer.


sarakonradi (approved):
Just tested it, thanks for the fix!

Issue #1811, opened 2020-10-22
Fast and safe Siegert implementation


AlexVanMeegen:
**Is your feature request related to a problem? Please describe.**
The `siegert_neuron` requires a numerically tricky integration where an `exp(s**2)` and and `1 + erf(s)` interact in the integrand. Currently, this problem is solved using a substitution that leads to an infinite integral which is better behaved but still tricky. The problem and the [currently implemented solution](https://github.com/nest/nest-simulator/blob/ad5549a7aa0dae6d47d38d8898e7652704fda61b/models/siegert_neuron.cpp#L339) is documented [here](https://www.frontiersin.org/articles/10.3389/fninf.2017.00034/full#h10).

**Describe alternatives you've considered**
Instead of the above substitution, one can use the scaled complementary error function to boil down the integrals to a single, well-behaved one. Here are the details: [siegert_numerics.pdf](https://github.com/nest/nest-simulator/files/5421828/siegert_numerics.pdf). In all my tests so far this works pretty well.

**Additional context**
This idea is approve by @mhelias; I also asked @tomtetzlaff for his opinion but his answer is pending. I can provide a fully Python based implementation in case someone wants to test it.

clinssen:
@AlexVanMeegen: this is a very in-depth description, and I'm not sure I understand the implications without further discussing this in detail. Could we ask you to take care of filing a pull request for this, if you (and Tom) indeed decide that this confers sufficient advantages?

AlexVanMeegen:
@clinssen, I am not sure I see why a full PR is needed to understand the implications. Could you be a bit more specific what you are missing? Then I can try to answer before creating a full blown PR.

Nonetheless, I am of course most willing to make the PR if desired!

clinssen:
I don't understand the domain well enough, to see the complete implications. If you are proposing to implement a mathematical re-formulation that yields the same results, but more accurately or with more numerical stability, this would I think be very welcome! And could you also say how it might impact runtime?

AlexVanMeegen:
Yes, precisely: This is all about the numerical evaluation of a tricky integral (Eq. 2 in my notes). My proposal is simply to use a different mathematical reformulation of the integral than the one that is currently used which is, to my best current knowledge, better behaved.

Naively, I think this will also improve runtime without affecting the accuracy, but this of course needs a careful benchmark.

AlexVanMeegen:
[...]

PR #1849, opened 2020-11-16
Fast and safe Siegert implementation

AlexVanMeegen:
This is the pull request for #1811. It yields consistent results with the old implementation for the cases that I checked (including a complex one with a network of 256 siegert neurons).

There is one TODO left: How to document the implementation properly. Currently, the steps involved to transform the integral are merely documented in the [pdf](https://github.com/nest/nest-simulator/files/5421828/siegert_numerics.pdf) that I included in #1811. Any ideas how to do this properly?

Furthermore, it has been a while since I coded C++. Please also check for stupid errors!

Cheers,
Alex


AlexVanMeegen:
PS: @clinssen, I kept the testing issue in mind. However, it seemed to me that the testsuite for the siegert neuron quite generally needs an extension, thus I kept it out of this PR and created #1850.

AlexVanMeegen:
Anecdotal evidence for the speed improvement: for the examples that I checked, the new implementation is 2 - 5 times faster; the difference is largest in the regimes where the numerics are problematic.

Furthermore, the old implementation eventually breaks for very small sigma while the new implementation tolerates this quite well.

AlexVanMeegen:
Travis fails on only on the macOS test with "The job exceeded the maximum time limit for jobs, and has been terminated." Is this my fault / is there anything I can do?

heplesser:
@AlexVanMeegen We just updated parts of the Travis build process, especially wrt macOS. Could you pull master into this PR and see if the macOS test then also passes?

AlexVanMeegen:
Rebased on current master and force-pushed, hope that didn't mess up everything.. Let's see what Travis says.

AlexVanMeegen:
@heplesser, now one Linux test fails with
```
E: The repository 'http://security.ubuntu.com/ubuntu bionic-security Release' no longer has a Release file.

N: Updating from such a repository can't be done securely, and is therefore disabled by default.

N: See apt-secure(8) manpage for repository creation and user configuration details.

The command "sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"" failed and exited with 100 during .
```
On the other tests, the same command runs flawlessly as far as I can judge. Can I do something about this?

heplesser:
@AlexVanMeegen Strange, that looks like a hick-up on the Ubuntu server side. I have restarted the job manually. If the problem should persist, @lekshmideepu is our Travis expert.

AlexVanMeegen:
Thanks @heplesser, seems like this was the issue. Travis looks happy now.

clinssen:
@AlexVanMeegen: re the documentation TODO, we support reStructuredText rendering in doxygen documentation blocks in the source file, as well as in general in the documentation (the pages on https://nest-simulator.readthedocs.org/). Our current pattern is to place an abbreviated version in the documentation block in the C++ header itself, perhaps containing only the end results of the derivation, whereas the document containing the detailed derivation itself should go into [doc/model_details](https://github.com/nest/nest-simulator/tree/master/doc/model_details). If you have a rST or MarkDown source version of the PDF, that might be preferable to the rendered PDF.

cc'ing @sarakonradi in case I missed anything. Thanks!

models/siegert_neuron.cpp
```diff
@@ -235,114 +210,62 @@ nest::siegert_neuron::~siegert_neuron()
...
+  F.function = &erfcx;
+  F.params = &erfcx_scale;
+
+  // Evaluate integral of exp( s^2 ) * ( 1 + erf( s ) ) from y_r to y_th
+  // depending on the sign of y_th and y_r. Uses the scaled complementary
+  // error function erfcx( s ) = exp( s^2 ) * erf( s ).
+  if ( y_r > 0. )
   {
-    return 0.;
+    gsl_integration_qags( &F, y_r, y_th, 0.0, 1.49e-8, 1000, gsl_w_, &result, &error );
```
clinssen, line 250: Could you define magic constants as const doubles instead?

clinssen (changes requested):

models/siegert_neuron.cpp
AlexVanMeegen: Thanks, that's much clearer. Better now?

AlexVanMeegen:
@clinssen, awesome, I did not know about doc/model_details. That seems perfect for this purpose. Does the documentation look good like this?

AlexVanMeegen:
Apologies for the force push - the macOS test failed with a pip dependency issue, thus I rebased on master in case this was fixed in the meantime.

AlexVanMeegen:
Rebase didn't help, the macOS test fails and I think the relevant line is
```
pip._vendor.pkg_resources.DistributionNotFound: The 'pip==20.3' distribution was not found and is required by the application
```
Any ideas?

doc/model_details/siegert_neuron_integration.ipynb
```diff
@@ -0,0 +1,272 @@
...
+    "\n",
+    "$$\\mathrm{erfcx}(s)=\\frac{1}{s\\sqrt{\\pi}}\\sum_{n=0}^{\\infty}(-1)^{n}\\frac{(2n-1)!!}{(2s^{2})^{n}}$$\n",
+    "\n",
+    "leads to the analytically solvable integral\n",
+    "\n",
+    "$$I(\\tilde{V}_{\\mathrm{th}},\\tilde{V}_{\\mathrm{r}})=\\int_{|\\tilde{V}_{\\mathrm{th}}|}^{|\\tilde{V}_{\\mathrm{r}}|}\\mathrm{erfcx}(s)ds\\approx\\frac{1}{\\sqrt{\\pi}}\\int_{|\\tilde{V}_{\\mathrm{th}}|}^{|\\tilde{V}_{\\mathrm{r}}|}\\frac{1}{s}ds=\\frac{1}{\\sqrt{\\pi}}\\log\\frac{|\\tilde{V}_{\\mathrm{r}}|}{|\\tilde{V}_{\\mathrm{th}}|}.$$\n",
```
clinssen, line 172:
```suggestion
    "$$I(\\tilde{V}_{\\mathrm{th}},\\tilde{V}_{\\mathrm{r}})=\\int_{|\\tilde{V}_{\\mathrm{th}}|}^{|\\tilde{V}_{\\mathrm{r}}|}\\mathrm{erfcx}(s)ds\\approx\\frac{1}{\\sqrt{\\pi}}\\int_{|\\tilde{V}_{\\mathrm{th}}|}^{|\\tilde{V}_{\\mathrm{r}}|}\\frac{1}{s}ds=\\frac{1}{\\sqrt{\\pi}}\\log\\frac{\\left|\\tilde{V}_{\\mathrm{r}}\\right|}{\\left|\\tilde{V}_{\\mathrm{th}}\\right|}.$$\n",
```

doc/model_details/siegert_neuron_integration.ipynb
```diff
@@ -0,0 +1,272 @@
...
+    "\n",
+    "Inserting this into $\\phi(\\mu,\\sigma)$ and using $\\tilde{V}_{\\mathrm{th}}\\approx\\frac{V_{\\mathrm{th}}-\\mu}{\\sigma}, \\tilde{V}_{\\mathrm{r}}\\approx\\frac{V_{\\mathrm{r}}-\\mu}{\\sigma}$ yields\n",
+    "\n",
+    "$$\\phi(\\mu,\\sigma)\\approx\\begin{cases}\n",
+    "0 & \\mu\\le V_{\\mathrm{th}}\\\\\n",
+    "\\frac{1}{\\tau_{\\mathrm{ref}}+\\tau_{\\mathrm{m}}\\log\\frac{\\mu-V_{\\mathrm{r}}}{\\mu-V_{\\mathrm{th}}}} & \\mu>V_{\\mathrm{th}}\\end{cases}$$\n",
+    "\n",
+    "[as it should](https://neuronaldynamics.epfl.ch/online/Ch8.S3.html). Thus, as long as the numerical solution of the integral $\\frac{1}{\\sqrt{\\pi}}\\int_{|\\tilde{V}_{\\mathrm{th}}|}^{|\\tilde{V}_{\\mathrm{r}}|}\\frac{1}{s}ds$ is precise, the deterministic limit is also numerically safe.\n",
+    "\n",
+    "Looking at the difference between $\\mathrm{erfcx}(s)$ and the first order asymptotics, we see that the noise free limit is already relevant for $|\\tilde{V}_{\\mathrm{th}}|\\approx O(100)$:"
```
clinssen, line 182:
```suggestion
    "Looking at the difference between $\\mathrm{erfcx}(s)$ and the first order asymptotics, we see that the noise free limit is already relevant for $\\left|\\tilde{V}_{\\mathrm{th}}\\right|\\approx O(100)$:"
```

clinssen: "we see that the noise free limit is already relevant for O(100)" sorry for not understanding this, doesn't the plot show that the approximation is good for (say) argument > 100, and it breaks down for smaller values? Perhaps I don't understand how this conclusion about the noise-free limit follows from the goodness of the approximation.

doc/model_details/siegert_neuron_integration.ipynb
```diff
@@ -0,0 +1,272 @@
...
+    "s = np.linspace(0.1, 100, 1000)\n",
+    "plt.plot(s, 1/(np.sqrt(np.pi)*s)-erfcx(s), c='black')\n",
+    "plt.xlim(s[0], s[-1])\n",
+    "plt.xscale('log')\n",
+    "plt.yscale('log')\n",
+    "plt.title(r'Absolute error of first order asymptotics of $\\mathrm{erfcx}(s)$')\n",
+    "plt.show()"
+   ]
+  },
+  {
```
clinssen, line 243: remove empty cell

clinssen (changes requested):
Looks good, I just have some minor comments on the notebook.

As for unit testing, ``pynest/nest/tests/test_siegert_neuron.py`` seems to test only one case, whereas we now have three case distinctions for various firing rate regimes. I don't want to feature creep this, but I wanted to ask if it would be relatively easy to add two more test cases, with different presynaptic rates, just to check if the firing rate predictions would still be correct.

AlexVanMeegen:
Thanks @clinssen for the thorough review. I tried to address your comments regarding the documentation, is it clearer now?

Regarding the tests: Unfortunately, one cannot cover all cases by simply changing the rate, in particular the inhibitory case with `mu < 0` is not possible because `mu` is proportional to the rate and the latter has to be positive. To address this I replaced the Poisson input in the test with white noise where `mu` can be arbitrary. Based on this, I covered all cases that seemed relevant. However, this is a rather large refactoring of the test. In particular, it increases its runtime from <0.5s to 4s on my machine. On the positive side, I think this addresses #1850.

clinssen (approved):
This is really great, thanks very much! The extra few seconds of test time shouldn't be anything to worry about. The MacOS stage is still failing on an unrelated installation issue, which is being worked on separately. I am therefore approving.

AlexVanMeegen:
Thanks, @clinssen! Two final questions:
1. I fixed the seed in the test, is this current best practice?
2. Do I have to register the additional tests somewhere?

clinssen:
Regarding 1, I don't think a current best-practice is defined, I will raise point in our next Open NEST Simulator meeting. For now I think it's alright.

Regarding 2: nothing more should be needed, the test is automatically picked up and run. However, your comment did lead me to realise that something was amiss, which led me to find the bug #1875—thanks again! No changes to this PR are needed as a consequence.

AlexVanMeegen:
Thanks, for checking and also for the nice review process. Went through one last time and noticed a misleading variable name, thus the new commit.

clinssen:
@AlexVanMeegen: apologies for the CI issues on MacOS. They should be solved now with the merge of #1889. Could we ask you to pull master and check that the build works OK now?

pnbabu (approved):
Looks good to me.

PR #1851, opened 2020-11-16
Fix Wundefined-var-template warnings

hakonsbm:
Moved definitions of static members from layer_impl.h to layer.h to fix `Wundefined-var-template` warnings when compiling with Clang.

### DIFF
--- a/nestkernel/layer.h
+++ b/nestkernel/layer.h
@@ -512,6 +512,7 @@ protected:
   MaskDatum mask_;
 };
 
+// Must define these static members here to avoid linker errors and Wundefined-var-template warnings.
 template < int D >
 std::shared_ptr< Ntree< D, index > > Layer< D >::cached_ntree_;
 template < int D >
--- a/nestkernel/layer.h
+++ b/nestkernel/layer.h
@@ -512,6 +512,11 @@ protected:
   MaskDatum mask_;
 };
 
+template < int D >
+std::shared_ptr< Ntree< D, index > > Layer< D >::cached_ntree_;
+template < int D >
+std::vector< std::pair< Position< D >, index > >* Layer< D >::cached_vector_ = 0;
+
 inline void
 AbstractLayer::set_node_collection( NodeCollectionPTR node_collection )
 {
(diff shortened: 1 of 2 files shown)


heplesser:
@hakonsbm Thanks, it compiles without warning with Apple Clang 12 as well now. The only remaining warnings are from pynestkernel/Cython-generated code, so not really our fault:
```
[ 99%] Building CXX object pynest/CMakeFiles/pynestkernel.dir/pynestkernel.cxx.o
/Users/plesser/NEST/code/bld_clang_mpi/pynest/pynestkernel.cxx:28238:3: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
  0, /*tp_print*/
  ^
/Users/plesser/opt/miniconda3/envs/nest_dev/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/Users/plesser/opt/miniconda3/envs/nest_dev/include/python3.8/pyport.h:515:54: note: expanded from macro 'Py_DEPRECATED'
#define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
```
Since we only compile cython-generated code in pynestkernel, it made me wonder if we can drop most of the compiler warning flags when building pynestkernel. But that would be an other issue, I suppose.

heplesser (approved):
Approving conditional on all tests passing.

heplesser (approved):
@hakonsbm Could you take a look at https://github.com/nest/nest-simulator/pull/1852/commits/e322613d56265fa8a698218e33abf09e8106f82f by @jougs solving this problem in a slightly different (and I think even better) way? It is already merged, so this PR may be obsolete now.

heplesser (changes requested):
See previous post, I meant to block until you have compared with the change in #1852.

hakonsbm:
@heplesser I think it would have been cleaner to have the definitions of the static members in the same header file as the class definition. But seeing as e322613 is already merged, I'll close this.

jougs:
Sorry for fixing this differently in #1852, I was not actually aware of this PR when I did that.

PR #1853, opened 2020-11-19
Refactor ModelsModule to not be a SLIModule anymore

jougs:
This refactors the `ModelsModule` to just be a shared library with a single function `register_builtin_models()` instead of being a dynamically or statically linked `SLIModule`. This addresses an action item in #1717.

I took the liberty of also deleting the `INSTALL` file that was more or less outdated and anyway just repeating the content of `doc/installation/hpc_install.rst`


nestkernel/kernel_manager.cpp
```diff
@@ -73,6 +77,7 @@ nest::KernelManager::KernelManager()
       &io_manager,
       &node_manager } )
```
clinssen, line 78: It should be possible to get guarantees about the order of initialisation here—could we put model_manager last to avoid having to use the ``builtin_models_registered_`` flag?

models/builtin_models.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace nest
+{
+
+void register_builtin_models();
```
clinssen, line 26: Should it be in its own namespace (under the ``nest`` namespace)?

clinssen (commented):
Thanks for paving the way! Just for my understanding: should we (in principle) be able to apply this same change to the extension module (``MyModule``), so it no longer derives from ``SLIModule``, but can still be dynamically loaded? Can this already be done before the next step, which is to make ``MyModule`` inherit from a new base class ``NESTComponent`` (or something similar)? What would the call to ``nest.Install()`` look like in that case?

nestkernel/kernel_manager.cpp
jougs: The sequence in the file is the result of quite a bit of trial and error. Every change I've ever tried lead to segmentation faults somewhere, so I think the current solution is the only one feasible right now.

The proper solution would of course be to assert the sequence by a clear programmatic definition of the dependencies between the different managers and maybe the architecture in general. However, I'm afraid that that's for another life (aka NEST 4 :wink:)

models/builtin_models.h
jougs: I don't think so. It's just a naked function that belongs to NEST.

jougs:
@clinssen: as mentioned earlier, this PR is only a first step in the direction of the new Extension Module System as outlined in #1717. I don't think a plain `.so` would be sufficiently flexible for being used as an extension module, so the short answer to your question is *no*. We will have to refactor this again to make it adhere to the new interface, once we have that.

What can of course be discussed here, is to close this PR and only refactor once, when the new system is in place. This change is the least important of the ones mentioned in #1717 and does not contain any user-visible changes.

README.md
```diff
@@ -107,9 +107,9 @@ later.
 
 ## Installing NEST
 
-Generic installation instructions can be found in the
-[INSTALL](https://github.com/nest/nest-simulator/blob/master/INSTALL) file that
-you received in the NEST sources.
+Generic installation instructions can be found in the [NEST
+documentation](https://nest-simulator.org/installation/) on Read the
+Docs.
```
heplesser, line 112: Given that we generate the ReadTheDocs doc from material in this repo, I'd include a pointer to the corresponding file here, just in case someone takes a memory stick with NEST sources along to a lonely island without network access ;).

doc/installation/hpc_install.rst
```diff
@@ -52,7 +52,7 @@ as in::
 BlueGene/Q and PyNEST
 ~~~~~~~~~~~~~~~~~~~~~
 
-Building PyNEST on BlueGene/Q requires you to compile dynamically, i.e. 
+Building PyNEST on BlueGene/Q requires you to compile dynamically, i.e.
```
heplesser, line 55: Shouldn't we remove the BG/Q instructions? Are there any BG/Qs left in operation?

doc/installation/hpc_install.rst
```diff
@@ -171,17 +171,19 @@ On the K Computer:
       make -j4
       make install
 
-  The compilation can take quite some time compiling the file ``models/modelsmodule.cpp``
-  due to generation of many template classes. To speed up the process, you can
-  comment out all synapse models you do not need.
-  The option ``-Kfast`` on the K computer enables many different options::
+  The compilation of the built-in models can take quite some time due
```
heplesser, line 174: I suppose this was exclusively relevant for K, which has been decomissioned. Should be replaced with a section for Fugaku, shouldn't it? But that (and BG/Q above) are rather topics for a separate issue.

models/builtin_models.h
heplesser, line 26: Please add a doxygen comment.

nestkernel/kernel_manager.h
```diff
@@ -204,6 +204,8 @@ class KernelManager
 private:
   std::vector< ManagerInterface* > managers;
   bool initialized_; //!< true if all sub-managers initialized
+
+  bool builtin_models_registered_;
```
heplesser, line 208: Please add doxygen comment.

models/builtin_models.h
heplesser: I am also a bit uncertain about the naming scheme. In one way, the scheme proposed is clear: all that is in `models` is built-in and nothing else can be built-in. But how will this scheme generalize? We might at some point soon have a large part of models auto-generated by NESTML and a few still based on hand-written code. I'd expect that there would be one models directory for each "source". And we need a clear naming for the corresponding extension module function as well. I think a naming scheme that can handle all situations consistently would be good to have.

heplesser (changes requested):
@jougs Nice, thank you! I have a few comments and mainly a concern about naming, see below.

doc/installation/hpc_install.rst
jougs: I agree that this information is pretty outdated (cf. the removal of INSTALL in the top directory). I still would rather refer a decision about this to @sarakonradi and the Docteam, as it is not really in the scope of this PR.

doc/installation/hpc_install.rst
jougs: I agree, but for details, please see my answer to the comment below.

jougs:
@clinssen, @heplesser: Your comments resonate well with my own concerns about this PR. I am meanwhile not so convinced anymore that this (more or less brutal) refactoring should actually be merged, but maybe it should rather just be seen as a learning experience about the dependencies and requirements for model/module registration.

Maybe what we should really do, is take `ModelsModule` as our first real-world example for the new extension system and let it guide our refactoring in the spirit of #1717?

heplesser:
> Maybe what we should really do, is take `ModelsModule` as our first real-world example for the new extension system and let it guide our refactoring in the spirit of #1717?

@jougs I like this idea. An important advantage is that it keeps the number of different approaches with use down, by having only "normal" SLIModules and Extension Modules, but not "builtin something modules" as a third variant.

PR #1859, opened 2020-11-30
Remove stray occurrences of double_t

heplesser:
This PR removes a few stray `double_t` occurrences in the code. They worked on most systems so far because `math.h` often defines this type, but it should not be used in NEST code—use just plain `double`.

### DIFF
--- a/models/erfc_neuron.h
+++ b/models/erfc_neuron.h
@@ -128,10 +128,10 @@ class gainfunction_erfc
 {
 private:
   /** threshold of activation function */
-  double_t theta_;
+  double theta_;
 
   /** 1/sqrt(2pi) x inverse of the maximal slope of gain function */
-  double_t sigma_;
+  double sigma_;
 
 public:
   /** sets default parameters */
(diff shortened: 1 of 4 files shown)


jougs (approved):
I'm approving the changes and propose to merge even though the MacOS build is failing again due to some stupid version mismatch.

jougs:
One more thing: if our policy is to not use these `_t` variants, shouldn't we add a test for this during the static code checking phase? Probably a simple `grep` could do most of the trick.

heplesser:
@lekshmideepu We seem to have yet another problem on OSX: Now, the `pip install --upgrade pip setuptools` seems to cause some version conflict between pip, see line 526ff in https://travis-ci.org/github/nest/nest-simulator/jobs/746767558. Do we strictly need the upgrade? And why to we mix on [lines 158ff in .travis.yml](https://github.com/nest/nest-simulator/blob/6848dbe10ecfd7f4f5285b6d298c4bf16e23b159/.travis.yml#L158) uses of `pip` and `pip3`?

heplesser:
> One more thing: if our policy is to not use these `_t` variants, shouldn't we add a test for this during the static code checking phase? Probably a simple `grep` could do most of the trick.

Good idea @jougs, but I'd move that to a follow-up issue.

jougs:
> > One more thing: if our policy is to not use these `_t` variants, shouldn't we add a test for this during the static code checking phase? Probably a simple `grep` could do most of the trick.
> 
> Good idea @jougs, but I'd move that to a follow-up issue.

Fine by me.

hakonsbm (approved):

heplesser:
Since tests only failed on macOS and I have tested manually on macOS, I will merge this PR without waiting for a fix of the pip-version-problem on Travis/macOS.

PR #1867, opened 2020-12-07
Remove unused lines from the part 1 tutorial document.

pnbabu:
import of SciPy and scikit-learn are not required for the part 1 tutorial, removing them.


clinssen:
As these are examples of other packages that *can* (but do not have to) be imported, would it be better to change the formulation of the text, rather than remove the import statements?

For example, instead of saying

"It should be noted, however, that certain external packages must be imported before importing nest. These include scikit-learn and SciPy."

could we say something along the lines of

"In case other packages are required, such as scikit-learn and SciPy, these need to be imported *before* importing nest."

doc/tutorials/pynest_tutorial/part_1_neurons_and_simple_neural_networks.rst
```diff
@@ -73,14 +73,13 @@ functionality into the Python interpreter.
 
     import nest
 
-It should be noted, however, that certain external packages must be
-imported *before* importing nest. These include `scikit-learn <http://scikit-learn.org/stable/index.html>`_
-and `SciPy <https://www.scipy.org/>`_.
+In case other python packages are required, such as `scikit-learn <http://scikit-learn.org/stable/index.html>`_
```
jessica-mitchell, line 76:
```suggestion
In case other Python packages are required, such as `scikit-learn <http://scikit-learn.org/stable/index.html>`_
```

doc/tutorials/pynest_tutorial/part_1_neurons_and_simple_neural_networks.rst
```diff
@@ -73,14 +73,13 @@ functionality into the Python interpreter.
...
+and `SciPy <https://www.scipy.org/>`_, they need to be imported before importing nest.
```
jessica-mitchell, line 77:
```suggestion
and `SciPy <https://www.scipy.org/>`_, they need to be imported *before* importing NEST.
```

jessica-mitchell (changes requested):
I just have a couple of minor changes - see inline for suggestions.

jessica-mitchell (approved):
Thanks @pnbabu! looks good

PR #1645, opened 2020-06-12
Create connections with lists of synapse dictionaries

stinebuu:
With this PR it will be possible to send in several synapse dictionaries when connecting. This means that if a source is connected to a target, we will actually create several connections, one per synapse dictionary in the syn_spec `nest.CollocatedSynapses()` argument specified. So
```
src = nest.Create('iaf_psc_alpha')
trgt = nest.Create('iaf_psc_alpha')

nest.Connect(src, trgt,
             syn_spec=nest.CollocatedSynapses({'weight': 3.}, {'synapse_model': 'stdp_synapse', 'delay': 1.3}))
```
will create **2** connections between `src` and `trgt`, one with a weight of `3`., the other with `stdp_synapse` and a delay of `1.3`.

The way I have implemented this is that instead of the syn_spec `dictionary` we used to pass along on C++ level, we are now sending around a `list` of dictionaries.
The user will still be able to pass a `syn_spec dictionary` on PyNEST and SLI level, this will be converted to a single-element list on C++ level.

I am making this a draft PR, as it is not quite ready yet, but I wanted you to see it so you could add your thoughts/input.

The following still needs to be done:
- [x] Check that it works for spatial networks
- [x] Update documentation on ReadTheDocs -> created issue #1804 
- [x] Decide what to do with structural plasticity: currently it is not possible with list of syn_spec dictionaries and structural plasticity. If someone need's `CollocatedSynapses` with structural plasticity, we can create an issue and PR.

@heplesser please have a look.

This PR will fix #904

PR #1868, opened 2020-12-09
Fix performance issues for weak scale benchmarks caused by creation of DictionaryDatums

stinebuu:
When running weak scale benchmarks for #1549 we discovered some connection creation performance issues for weak scale benchmarks. I have done some bisecting, and found the problem comes from cfa213d0586afa8224dd2cc932283ee597d500d4, and the fact that all threads try to write to `param_dict` at the same time.

Luckily we already have a list of dictionaries for each thread, `dummy_param_dicts_`, so now we fill the dictionary on the thread if the connection has synapse properties, and just use this variable throughout. It might be an idea to change the name of the variable, I just didn't come up with a good name.

I also had to make it `protected`, as it is used by `conngen` rule.

Some benchmarks:

hpc_benchmark_weak_scaling_connection_fix
hpc_benchmark_weak_scaling_fix

I still need to figure out why it is not finishing for `N_VP=1152`, but that is another issue that should not be addressed in this PR.

### DIFF
--- a/nestkernel/conn_builder.cpp
+++ b/nestkernel/conn_builder.cpp
@@ -111,7 +111,7 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
   dummy_param_dicts_.resize( kernel().vp_manager.get_num_threads() );
 #pragma omp parallel
   {
-    dummy_param_dicts_[ kernel().vp_manager.get_thread_id() ] = new Dictionary();
+    dummy_param_dicts_[ kernel().vp_manager.get_thread_id() ] = Dictionary();
   }
 
   // If make_symmetric_ is requested call reset on all parameters in order
@@ -301,14 +301,18 @@ nest::ConnBuilder::disconnect()
   }
 }
 
-void
+Dictionary&
 nest::ConnBuilder::create_param_dict_( index snode_id,
   Node& target,
   thread target_thread,
   librandom::RngPtr& rng,
   index indx )
 {
-  if ( not param_dicts_[ indx ].empty() ) // indicates we have synapse params
+  if ( param_dicts_[ indx ].empty() ) // indicates we have no synapse params
+  {
+    return dummy_param_dicts_[ target_thread ];
+  }
+  else
   {
     assert( kernel().vp_manager.get_num_threads() == static_cast< thread >( param_dicts_[ indx ].size() ) );
 
@@ -318,18 +322,18 @@ nest::ConnBuilder::create_param_dict_( index snode_id,
       {
         // change value of dictionary entry without allocating new datum
         IntegerDatum* id = static_cast< IntegerDatum* >(
-          ( ( *param_dicts_[ indx ][ target_thread ] )[ synapse_parameter.first ] ).datum() );
+          ( param_dicts_[ indx ][ target_thread ][ synapse_parameter.first ] ).datum() );
         ( *id ) = synapse_parameter.second->value_int( target_thread, rng, snode_id, &target );
       }
       else
       {
         // change value of dictionary entry without allocating new datum
         DoubleDatum* dd = static_cast< DoubleDatum* >(
-          ( ( *param_dicts_[ indx ][ target_thread ] )[ synapse_parameter.first ] ).datum() );
+          ( param_dicts_[ indx ][ target_thread ][ synapse_parameter.first ] ).datum() );
         ( *dd ) = synapse_parameter.second->value_double( target_thread, rng, snode_id, &target );
       }
     }
-    dummy_param_dicts_[ target_thread ] = param_dicts_[ indx ][ target_thread ];
+    return param_dicts_[ indx ][ target_thread ];
   }
 }
 
@@ -343,12 +347,12 @@ nest::ConnBuilder::single_connect_( index snode_id, Node& target, thread target_
 
   for ( size_t indx = 0; indx < synapse_model_id_.size(); ++indx )
   {
-    create_param_dict_( snode_id, target, target_thread, rng, indx );
+    Dictionary& param_dict = create_param_dict_( snode_id, target, target_thread, rng, indx );
 
     if ( default_weight_and_delay_[ indx ] )
     {
       kernel().connection_manager.connect(
-        snode_id, &target, target_thread, synapse_model_id_[ indx ], dummy_param_dicts_[ target_thread ] );
+        snode_id, &target, target_thread, synapse_model_id_[ indx ], param_dict );
     }
     else if ( default_weight_[ indx ] )
     {
@@ -356,7 +360,7 @@ nest::ConnBuilder::single_connect_( index snode_id, Node& target, thread target_
         &target,
         target_thread,
         synapse_model_id_[ indx ],
-        dummy_param_dicts_[ target_thread ],
+        param_dict,
         delays_[ indx ]->value_double( target_thread, rng, snode_id, &target ) );
     }
     else if ( default_delay_[ indx ] )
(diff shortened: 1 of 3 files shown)


heplesser:
@stinebuu Thanks, I will take a look at it shortly.

nestkernel/conn_builder.cpp
```diff
@@ -335,10 +329,8 @@ nest::ConnBuilder::create_param_dict_( index snode_id,
         ( *dd ) = synapse_parameter.second->value_double( target_thread, rng, snode_id, &target );
       }
     }
-    param_dict = param_dicts_[ indx ][ target_thread ];
+    dummy_param_dicts_[ target_thread ] = param_dicts_[ indx ][ target_thread ];
```
heplesser, line 332: If I understand the entire context right, then `dummy_param_dicts_` should be an empty dictionary if for a given synapse spec there are no synapse parameters (except weight and delay, which are handled explicitly). With the code as it is now, I think that if we have collocated synapses with specs A and B, where A contains additional parameters and B not, then this code will set `dummy_params_dicts` to a non-empty dictionary when processing A, but not do anything when processing B, so B will be created with the parameter dict from A still in `dummy_param_dicts_`, which is wrong. Or am I overlooking something? Do we have a test for that case?

nestkernel/conn_builder.cpp
```diff
@@ -301,20 +301,14 @@ nest::ConnBuilder::disconnect()
   }
 }
 
-DictionaryDatum
+void
 nest::ConnBuilder::create_param_dict_( index snode_id,
   Node& target,
   thread target_thread,
   librandom::RngPtr& rng,
   index indx )
```
heplesser, line 309: The `indx` parameter name is uninformative, could you find a more specific one?

nestkernel/conn_builder.cpp
```diff
@@ -351,19 +343,19 @@ nest::ConnBuilder::single_connect_( index snode_id, Node& target, thread target_
 
   for ( size_t indx = 0; indx < synapse_model_id_.size(); ++indx )
```
heplesser, line 344: I am confused by `synapse_model_id_` (and the little informative `indx`). This loop iterates over the different synapse specifications provided in a collocated synapse object, doesn't it? Can we find better names to make this clearer?

nestkernel/conn_builder.cpp
```diff
@@ -351,19 +343,19 @@ nest::ConnBuilder::single_connect_( index snode_id, Node& target, thread target_
...
   {
-    DictionaryDatum param_dict = create_param_dict_( snode_id, target, target_thread, rng, indx );
+    create_param_dict_( snode_id, target, target_thread, rng, indx );
```
heplesser, line 346: Maybe `prepare_...`?

nestkernel/conn_builder.cpp
```diff
@@ -351,19 +343,19 @@ nest::ConnBuilder::single_connect_( index snode_id, Node& target, thread target_
...
 
     if ( default_weight_and_delay_[ indx ] )
     {
-      kernel().connection_manager.connect( snode_id, &target, target_thread, synapse_model_id_[ indx ], param_dict );
+      kernel().connection_manager.connect( snode_id, &target, target_thread, synapse_model_id_[ indx ], dummy_param_dicts_[ target_thread ] );
```
heplesser: It seems confusing that all `connect()` calls pass a `dummy_...` argument. We should think more about names here.

heplesser (changes requested):
@stinebuu I have one concern about correctness (see inline comment), but am mainly wondering if we should not rethink our design more carefully. I assume that the threading problems arise from the single-pragmas in the lock/unlock mechanism of the `DictionaryDatum` objects (https://github.com/nest/nest-simulator/blob/4ad971a2aff0dd25523597074978aa0250ff91da/libnestutil/lockptr.h#L141-L155). But I wonder if we at all need `DictionaryDatum`s here and could not just do with (pointer to) plain `Dictionary`s. If `dummy_param_dicts_` and `param_dicts_` would just contain `Dictionary`, then `create_param_dict_` (maybe renamed to `prepare_...`) could return a reference to the proper dictionary, and then in `single_connect()` we can have 
```c++
Dictionary& param_dict = prepare_param_dict_(...);
```
as a thread-local variable. That would probably also improve performance. I think I built all this on `DictionaryDatum` once upon a time, not sure why I did not use just `Dictionary`  then.

stinebuu:
@heplesser Thanks for your review! I will take a look shortly, I am currently looking into why the benchmark does not finish for `N_VP=1152`. I will take a look at the design first, to make sure it works, and then address your other comments. :)

stinebuu:
@heplesser Could you have a look at 36ccfa5d4b9b2a4e5c2a2899717f37b23474e12a and see if this is what you meant? It does not fix the performance problems.

hpc_benchmark_weak_scaling_connection_fix

I thought the problem was that all threads are trying to write to `param_dict` in `single_connect()`, and this leads to caching problems (this might be why we have `dummy_param_dicts_` to begin with). This is why I changed the dictionary in `dummy_param_dicts_`, and use this directly for this PR. This is closer to what we did before #1645 (before that PR we used `dummy_param_dicts_` if we had no synapse parameters, and  `param_dicts_` if we had parameters, and with #1645 I changed it to first write to `param_dict` and then just using this). 

https://github.com/nest/nest-simulator/blob/2869b6364ccd85edf69246deb12c03c05e4a3b49/nestkernel/conn_builder.cpp#L354

I do think it makes sense that it has something to do with the lock/unlock mechanism. Though both `dummy_param_dicts_` and `param_dicts_` both used `DictionaryDatum` for a long time without it causing problems.

heplesser:
@stinebuu This is strange. What confuses me is that in the original code, we execute
```c++
DictionaryDatum param_dict = create_param_dict_( ... )
```
in a thread-local context, so each thread should have its own `param_dict` instance and writing to it should not cause conflicts. At the same time, I have also started to wonder how wise my advice to drop the `...Datum` was, since it may introduce a memory leak when the `Dictionary` is not properly deleted. I am still concerned that your solution does not clear the "dummydict" properly if it has been used to actually carry parameters. Is there any way to profile this with VTune or similar to understand what is going on? I'd really like to understand the cause of the problems.

hakonsbm:
@heplesser @stinebuu
I have run the connection step of hpc_benchmark with VTune using the current master (2869b63), and it gives some insight into what's going on. When connecting, `create_param_dict_()` is called in `single_connect_()` for every connection created. Every time `create_param_dict_()` is called, a new `DictionaryDatum` is created, a `Dictionary` is copied into the datum (or converted to a `DictionaryDatum`), and the datum is copied as the return value of the function. This makes the function inefficient, especially because it has to create a new `DictionaryDatum` every time (which under the hood uses a `Dictionary` smart pointer which calls `new` on initialization). It seems like this is the main issue, not a conflict between the threads.

I have also looked at the current commit in this branch (36ccfa5). Here `create_param_dict_` returns a reference to the `Dictionary` directly, without creating a new `DictionaryDatum` to put it in. The returned `Dictionary` from `create_param_dict_()` is then used in `kernel().connection_manager.connect()`. But that requires a `DictionaryDatum` as argument, making it implicitly convert the `Dictionary` to a `DictionaryDatum` when called. So while 36ccfa5 fixes the problem with `create_param_dict_()`, the problem is more or less just moved to `single_connect_()` which is why it doesn't fix the performance problems.

heplesser:
@hakonsbm Thanks a lot for the analysis! Then we can think now about how to avoid unnecessary creation. The idea behind the entire setup originally was to avoid creating any more dictionaries than absolutely necessary, just updating values as needed. We should probably take a step back to get a bigger picture.

stinebuu:
@heplesser @hakonsbm I have tested the last commit in this PR, a9f994b4b019c2e4de146986c3c4a1bdd28cd1e3, and it shows as good as performance as we had previously. 
 
hpc_benchmark_weak_scaling_connection_fix

heplesser:
@stinebuu Excellent news! This looks very nice now except for the one comment I had on variable names.

stinebuu:
@heplesser I have changed `indx` to `synapse_indx`, is this more intuitive?

heplesser (approved):
👍

hakonsbm (approved):
Looks good now!

PR #1869, opened 2020-12-10
Remove contact invitation from HPC systems tab

sarakonradi:
This PR removes the contact invitation from the ``HPC systems`` tab in the NEST installation instructions. 

As discussed offline, it suffices to provide our contact information in the Community, Troubleshooting and Welcome pages.


heplesser (approved):
@sarakonradi Thanks for the PR!

terhorstd:
we should consider to also back-port this to 2.20.2?

heplesser:
> we should consider to also back-port this to 2.20.2?

Yes.

PR #1876, opened 2020-12-15
Add synapse_label and receptor_type support to spatial connections 

hakonsbm:
Implements connection with specified receptor type or synapse label when using spatial parameters, also for collocated synapses.


stinebuu:
@hakonsbm `receptor_type` already work with `CollocatedSynapses` do they not?

hakonsbm:
@stinebuu Yes, this is only for spatial connections. However, I see that I forgot the test for `receptor_type` with `CollocatedSynapses`, so I'll add that.

stinebuu:
We should wait with this until the problems in #1868 is addressed.

stinebuu (approved):
I am happy with these changes. I think we in the future should change the spatial connect to take all synapse specific parameters, like is done in `conn_builder`, but that is not something that should be done in this PR.

stinebuu:
@hakonsbm could you merge master to make Travis happy?

heplesser (commented):
@hakonsbm The code looks correct to me, but confusing, due to the double-up of code for individual and collocated synapses. I think the code would become clearer by first extracting all parameters relevant to the connection rule, then to check if there is a `synapse_params` (i.e., collocated synapses) and handling that, and then handling the case of a single synapse specification. Maybe extracting the synapse params can then be delegate to a helper function to avoid code duplication. Something along the line of (very pseudocode here)
```c++
if d->known("synapse_params")
{
  param_dicts_.resize(...);
  for ( auto sp : synapse_params )
      extract_params(sp);
}
else
{
  param_dicts_.resize(1);
  extract_params(d);
}
```

hakonsbm:
@heplesser Done!

nestkernel/connection_creator.cpp
```diff
@@ -39,165 +39,70 @@ ConnectionCreator::ConnectionCreator( DictionaryDatum dict )
...
-      {
-        ( *param_dicts_[ 0 ].at( kernel().vp_manager.get_thread_id() ) )[ names::receptor_type ] =
-          getValue< long >( dit->second );
-      }
+      throw BadProperty( "Number of connections cannot be less than zero." );
     }
-    else if ( dit->first == names::synapse_parameters )
+
+    number_of_connections_ = number_of_connections;
+  }
```
heplesser: I think with `updateValue()` this can be simplified to the following. I also exploit here that we are in the constructor, so we can change the member variable directly, because the object will be destroyed if the value is negative, no need for a temporary local variable.

```c++
if ( updateValue< long >(dict, names::number_of_connections, number_of_connections_ )
{
  if ( number_of_connections_ < 0 )
  {
    throw BadProperty( "Number of connections cannot be less than zero." );
  }
} 
```

heplesser (changes requested):
@hakonsbm Thanks, looks much better now. Just to small remarks.

nestkernel/connection_creator.cpp
hakonsbm: But `number_of_connections_` is of type `size_t`, so updating it directly will make it impossible to detect if a negative number is specified.

nestkernel/connection_creator.cpp
```diff
@@ -39,165 +39,70 @@ ConnectionCreator::ConnectionCreator( DictionaryDatum dict )
   Name connection_type;
   long number_of_connections( -1 ); // overwritten by dict entry
 
-  param_dicts_.resize( 1 );
-  param_dicts_[ 0 ].resize( kernel().vp_manager.get_num_threads() );
-#pragma omp parallel
+  // Lambda function to store bool value to specified variable if it exists in dict.
+  auto store_bool_if_known = [&dict]( bool& target, const Name& name ) -> void
```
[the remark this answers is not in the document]
hakonsbm: I suspected that there was a function for this. I've updated the calls now.

nestkernel/connection_creator.cpp
heplesser: Good point, but difficult to see from the code. Could you add a brief comment?

hakonsbm: @heplesser Done.

heplesser (approved):
@hakonsbm Nice work! Approving conditional on Travis passing.

Issue #1879, opened 2020-12-18
Typo in test_sp/test_mpitests.py


niltonlk:
**Describe the bug**
A typo in line 41 of test_sp/test_mpitests.py that reads "python" instead of "nosetests" causes the test to fail.
Note that the error message that should be displayed in case of failure in line 50 reads "nosetests".
Test pass by fixing this typo.

**To Reproduce**
1. (Minimal) reproducing example
make installcheck with MPI ON
2. Used parameters
cmake -DCMAKE_INSTALL_PREFIX:PATH=/opt/ohpc/pub/lib/nest/master -Dwith-mpi=ON -Dwith-boost=ON -DPYTHON_EXECUTABLE=`which python3` -DCYTHON_EXECUTABLE`which cython` ../src
3. Command to run
make
make install
make installcheck

**Expected behavior**
All tests should pass

**Screenshots**

**Desktop/Environment (please complete the following information):**
 - OS: CentOS 7.6
 - Shell: bash
 - Python-Version: Python 3.6.8
 - NEST-Version: nest-2.20.1 and master@2869b63
 - Installation: from source with parameters described above

PR #1880, opened 2020-12-18
Fix typo in test_sp/test_mpitests.py

niltonlk:
Fix issue #1879 
A typo in line 41 of test_sp/test_mpitests.py that reads "python" instead of "nosetests" causes the test to fail.
Note that the error message that should be displayed in case of failure in line 50 reads "nosetests".
Test pass by fixing this typo.


stinebuu:
@jougs Could you take a look at this, it seems fine to me, but I was suddenly afraid it might break something.

jougs (approved):
LGTM. Thanks!

Merging without formal second review, as @stinebuu indicated approval and the change is pretty small.

PR #1886, opened 2021-01-11
Fix for printing empty SynapseCollection

### LINKED ISSUE #1884: Error raised when printing an empty SynapseConnection
**Describe the bug**
Printing a SynapseCollection with no connections raises a `TypeError`.

**To Reproduce**
```python
import nest
print(nest.GetConnections())
```

Raises the error
```
  File "<string>", line 1, in <module>
  File "path/to/nest/install/lib/python3.7/site-packages/nest/lib/hl_api_types.py", line 650, in __str__
    srcs = params['source']
TypeError: tuple indices must be integers or slices, not str
```

**Expected behavior**
Print output representing an empty SynapseCollection.

hakonsbm:
Printing an empty SynapseCollection will no longer raise a `TypeError`, but print
```
 source   target   synapse model   weight   delay 
-------- -------- --------------- -------- -------
```
I'm also open for more explicit representations of an empty SynapseCollection.

Fixes #1884

### DIFF
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -647,18 +647,12 @@ class SynapseCollection(object):
         MAX_SIZE_FULL_PRINT = 35  # 35 is arbitrarily chosen.
 
         params = self.get()
-        if len(params) == 0:
-            srcs = []
-            trgt = []
-            wght = []
-            dlay = []
-            s_model = []
-        else:
-            srcs = params['source']
-            trgt = params['target']
-            wght = params['weight']
-            dlay = params['delay']
-            s_model = params['synapse_model']
+
+        srcs = params['source'] if len(params) != 0 else []
+        trgt = params['target'] if len(params) != 0 else []
+        wght = params['weight'] if len(params) != 0 else []
+        dlay = params['delay'] if len(params) != 0 else []
+        s_model = params['synapse_model'] if len(params) != 0 else []
 
         if isinstance(srcs, int):
             srcs = [srcs]
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -648,11 +648,14 @@ class SynapseCollection(object):
 
         params = self.get()
 
-        srcs = params['source'] if len(params) != 0 else []
-        trgt = params['target'] if len(params) != 0 else []
-        wght = params['weight'] if len(params) != 0 else []
-        dlay = params['delay'] if len(params) != 0 else []
-        s_model = params['synapse_model'] if len(params) != 0 else []
+        if len(params) == 0:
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_types.py
```diff
@@ -667,11 +674,18 @@ def format_row_(s, t, sm, w, dly):
...
+            trg_len = len(trg_h) + 2
+            sm_len = len(sm_h) + 2
+            w_len = len(w_h) + 2
+            d_len = len(d_h) + 2
+        else:
+            src_len = max(len(src_h) + 2, floor(log(max(srcs), 10)))
+            trg_len = max(len(trg_h) + 2, floor(log(max(trgt), 10)))
+            sm_len = max(len(sm_h) + 2, len(max(s_model, key=len)))
+            w_len = len(w_h) + 2
+            d_len = len(d_h) + 2
```
stinebuu, line 688: Maybe something like this instead, to shorten somewhat?
```suggestion
        else:
            src_len = max(len(src_h) + 2, floor(log(max(srcs), 10)))
            trg_len = max(len(trg_h) + 2, floor(log(max(trgt), 10)))
            sm_len = max(len(sm_h) + 2, len(max(s_model, key=len)))
        w_len = len(w_h) + 2
        d_len = len(d_h) + 2
```

pynest/nest/lib/hl_api_types.py
```diff
@@ -647,11 +647,18 @@ def format_row_(s, t, sm, w, dly):
...
-        dlay = params['delay']
-        s_model = params['synapse_model']
+        if len(params) == 0:
+            srcs = []
+            trgt = []
+            wght = []
+            dlay = []
+            s_model = []
+        else:
+            srcs = params['source']
```
stinebuu, line 657: Maybe use `srcs = params['source'] if len(params) != 0 else []` to shorten? And the same for all the others. Now we have 2 `if` sentences to define `srcs` etc. (There is also an `if isinstance (srcs, int)` just below this.

stinebuu (changes requested):
Thanks, I have a couple of suggestions

stinebuu:
Also, I remember fixing the `boarder` thing, but apparently I never committed that, good we get to fix it now!

stinebuu (approved):
Thanks!

jougs (commented):
I'm fine with the code as is, but wonder if we instead should print something more friendly like
```
The synapse collection does not contain any connections!
```
@sarakonradi, @jessica-mitchell: what does the docteam think about this?

heplesser:
> I'm fine with the code as is, but wonder if we instead should print something more friendly like
> 
> ```
> The synapse collection does not contain any connections!
> ```

I agree with @jougs, but would replace the exclamation mark with a full stop :).

jougs (approved):
Thanks!

PR #1889, opened 2021-01-12
Change Travis macOS build to MINIMAL to avoid timeouts (temporary).

heplesser:
All macOS builds on Travis are currently failing due to timeouts. This PR temporarily changes the build config for macOS to `MINIMAL` to ensure that tests can complete again.

This should be reverted once we have a more powerful macOS setup.

I think a single reviewer on this one is ok.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ jobs:
       os: osx
       osx_image: xcode12.2   # Python 3.9.0 running on macOS 10.15.7
       language: shell      # language 'python' results in errors on macOS
-      env: CACHE_NAME=JOB xNEST_BUILD_TYPE=FULL_NO_EXTERNAL_FEATURES
+      env: CACHE_NAME=JOB xNEST_BUILD_TYPE=MINIMAL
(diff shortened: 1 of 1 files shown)


heplesser:
In a passing build in my repo (https://travis-ci.org/github/heplesser/nest-simulator/builds/754094096), I noticed that for macOS, the `before_install.3` stage takes almost 1500s (25 minutes!), while it runs in about one minute under Linux. The minimal testsuite itself (293 tests) takes about 200s on macOS vs just 40s on Linux. This does not reflect normal runtimes on macOS on local machines.

jougs (approved):
The slowness of macOS builds is consistent with earlier observations. Thanks anyway for extracting the numbers. My suspicion is that we're sharing these machines with many people, as they are the ones that actually cost them real money and thus really can't do much about it.

As this is a single line change using a feature I myself have implemented and other PRs are blocked by this, I'm releasing @lekshmideepu from her reviewership and approve and merge immediately. Thanks!

PR #1890, opened 2021-01-15
Avoid misleading warning when connecting

### LINKED ISSUE #1887: Excessive and misleading warning when creating the first connection
**Describe the bug**
In the current master (be21fc9a6b4b12998b80483f56a5fa794f099370), when creating the first connection or the first connection after calling `ResetKernel()`, the following warning will consistently be raised:
```
Jan 11 15:10:13 ConnectionManager [Warning]: 
    New connections created, connection descriptors previously obtained using 
    'GetConnections' are now invalid.
```
It is always raised, regardless of having called `GetConnections()` or not, which can be confusing. And because it is raised again after every `ResetKernel()`, the amount of warnings can be obstructive when for example running unittests.

**To Reproduce**
Running the following will raise the warning:
```python
import nest

n = nest.Create('iaf_psc_alpha')
nest.Connect(n, n)  # Warning is raised here.
```

Then, if calling `ResetKernel()` and connecting again, the warning is raised once more:
```python
nest.ResetKernel()
n2 = nest.Create('iaf_psc_alpha')
nest.Connect(n2, n2)  # Warning is raised again here.
```

**Expected behavior**
I think the warning should only be raised if `GetConnections()` has been called.

**Additional context**
The warning is created here:
https://github.com/nest/nest-simulator/blob/be21fc9a6b4b12998b80483f56a5fa794f099370/nestkernel/connection_manager.cpp#L1433-L1438

Based on `have_connections_changed_`, which is initialized here:
https://github.com/nest/nest-simulator/blob/be21fc9a6b4b12998b80483f56a5fa794f099370/nestkernel/connection_manager.cpp#L97

But it may require an additional variable to keep track of calls to `GetConnections()`.

hakonsbm:
This PR adds a flag to keep track of whether `GetConnections` has been called, to know when to warn about invalidation of connection descriptors.

Fixes #1887.


nestkernel/connection_manager.cpp
```diff
@@ -1450,3 +1457,21 @@ nest::ConnectionManager::unset_have_connections_changed( const thread tid )
...
+nest::ConnectionManager::set_has_get_connections_been_called( const thread tid )
+{
+  if ( has_get_connections_been_called_[ tid ].is_false() )
+  {
+    has_get_connections_been_called_[ tid ].set_true();
+  }
+}
+
+void
+nest::ConnectionManager::unset_has_get_connections_been_called( const thread tid )
```
stinebuu, line 1471: This function does not seem to be used?

nestkernel/connection_manager.cpp
```diff
@@ -1432,9 +1435,13 @@ nest::ConnectionManager::set_have_connections_changed( const thread tid )
...
   {
-    std::string msg =
-      "New connections created, connection descriptors previously obtained using 'GetConnections' are now invalid.";
-    LOG( M_WARNING, "ConnectionManager", msg );
+    if ( has_get_connections_been_called_[ tid ].is_true() )
+    {
+      std::string msg =
+        "New connections created, connection descriptors previously obtained using 'GetConnections' are now invalid.";
+      LOG( M_WARNING, "ConnectionManager", msg );
+      has_get_connections_been_called_[ tid ].set_false();
```
stinebuu, line 1443: So we will only get a warning the first time `Connect()` is called after `GetConnections()`?

nestkernel/connection_manager.cpp
hakonsbm: It's not, it's mostly there to make the interface consistent with that of `have_connections_changed_`. Maybe I should remove it?

nestkernel/connection_manager.cpp
hakonsbm: Yes.

nestkernel/connection_manager.cpp
jougs: I also think that this unused function should best be removed.

nestkernel/connection_manager.h
```diff
@@ -582,6 +593,9 @@ class ConnectionManager : public ManagerInterface
   //! simulate.
   PerThreadBoolIndicator have_connections_changed_;
 
+  //! True if GetConnections has been called.
+  PerThreadBoolIndicator has_get_connections_been_called_;
```
jougs, line 597: Do we strictly need this to be a `PerThreadBoolIndicator`? Wouldn't it be sufficient to have a simple `bool` (initialized `false`) and set `true` just before returning the `result` in [`ConnectionManager::get_connections()`](https://github.com/nest/nest-simulator/blob/master/nestkernel/connection_manager.cpp#L834) instead of doing this in the parallel context? That way, we'd still have less parallel operations and the flag still unset even if any of the exceptions later on in the function would trigger.

jougs (changes requested):
Thanks! I was also already a bit annoyed by the many printouts even though I never actually called `GetConnections`.

nestkernel/connection_manager.h
hakonsbm: I agree, I've converted it to a `bool` now.

stinebuu (approved):
Thanks!

hakonsbm:
Apparently, for 80fe45b, the MPI test `test_multiple_synapses_spatial_networks.sli` [failed](https://travis-ci.org/github/nest/nest-simulator/jobs/755508537#L7419), but I can't see why it should have failed. Also, the testsuite results shows one MPI test as failing, which is correct, but lists a lot of additional tests under [`The following tests failed`](https://travis-ci.org/github/nest/nest-simulator/jobs/755508537#L7508). Is there a bug in `summarize_tests.py`?

stinebuu:
There is something going on with `test_multiple_synapses_spatial_networks.sli`, it fails on Travis sometimes, but I have no idea how or why. If Travis is rerun on the same commit for instance, the test usually pass. So I do not think it has anything to do with this PR. We should probably create an issue about the test, but it is difficult when it only fails randomly.

heplesser:
I had similar observations as @stinebuu and @hakonsbm, e.g. with [build 5661.3](https://travis-ci.org/github/nest/nest-simulator/builds/755491511), which disappeared on re-running. There are two questions here:
1. Why the intermittent failures
2. Why the long list of failed tests reported

A potential explanation for 1 could be that the communication of results between processes in NEST MPI tests is a bit fragile  (and should be replaced by something better). Why the log parsing apparently fails I don't know. We would need a log from a failing case ...

nestkernel/connection_manager.cpp
```diff
@@ -1432,9 +1437,13 @@ nest::ConnectionManager::set_have_connections_changed( const thread tid )
...
   {
-    std::string msg =
-      "New connections created, connection descriptors previously obtained using 'GetConnections' are now invalid.";
-    LOG( M_WARNING, "ConnectionManager", msg );
+    if ( has_get_connections_been_called_ )
+    {
+      std::string msg =
+        "New connections created, connection descriptors previously obtained using 'GetConnections' are now invalid.";
+      LOG( M_WARNING, "ConnectionManager", msg );
+      set_has_get_connections_been_called( false );
```
jougs, line 1445: Am I right in that this line is here to make sure the warning is only printed once? I wonder if we should have a second flag `invalid_connection_descriptor_warning_printed_` for this, as the flag is technically in a wrong state after this statement. With the current set of changes, I'm pretty sure that nothing bad can happen, but if people start relying on `has_get_connections_been_called_` for other purposes in the future, this might blow up and lead to errors that could potentially be hard to debug. But maybe I'm being overly paranoid here...

@stinebuu: any opinions? How are you dealing with this in the deprecation warning framework?

nestkernel/connection_manager.h
```diff
@@ -582,6 +587,9 @@ class ConnectionManager : public ManagerInterface
   //! simulate.
   PerThreadBoolIndicator have_connections_changed_;
 
+  //! True if GetConnections has been called.
```
jougs, line 590:
```suggestion
  //! true if GetConnections has been called.
```

nestkernel/connection_manager.h
```diff
@@ -807,6 +815,12 @@ ConnectionManager::set_source_has_more_targets( const thread tid,
   connections_[ tid ][ syn_id ]->set_source_has_more_targets( lcid, more_targets );
 }
 
+inline void
+nest::ConnectionManager::set_has_get_connections_been_called( const bool has_get_connections_been_called )
```
jougs, line 819: Should we instead have `set_...` and `unset_...` (without arguments) here for additional abstraction?

jougs (commented):
Thanks for considering my last round of comments. I think the changeset looks much cleaner now. I have two more comments, but am happy to forget about them if you and @stinebuu think they're rubbish.

nestkernel/connection_manager.cpp
stinebuu: @jougs For the deprecation warnings I have a map between the name of deprecated functions and whether they are deprecated. I also set them to `false` after the warning is issued, but it is clear that the bool is connected to a deprecation warning. 
https://github.com/nest/nest-simulator/blob/47b7e43f72317911b3e285daa9ecd4f45688fa5d/nestkernel/deprecation_warning.cpp#L37-L45

I think you are right and that `has_get_connections_been_called_` might be used for other purposes in the future, but it might be enough with just one bool if we rename it? Because it is only used for this now, so I do not want to have a bool laying around that is not used at all.

@hakonsbm I think you should also add a comment after `set_has_get_connections_been_called( false )` just to point out why it is there ;)

nestkernel/connection_manager.h
hakonsbm: I don't think that's necessary. It's just a bool being set to `true` or `false`. And this way we don't need to add an additional function to the already pretty crowded `ConnectionManager`.

nestkernel/connection_manager.cpp
hakonsbm: I also think that if the flag starts being used for other things it should be renamed. Until then, I've added some comments to clarify the usage of `set_has_get_connections_been_called()`.

jougs: Fine by me!

nestkernel/connection_manager.h
jougs: Alright. I was just wondering...

jougs (approved):
Thanks!

PR #1899, opened 2021-01-29
Restructure recorder documentation to contain all properties.

jougs:
This PR is a replacement for #1819 and in essence adds the complete table of recorder properties to each recorder by including a documentation snippet extracted from `nestkernel/recording_device.h`, where the `RecordingBackend` base class is defined.

On the infrastructure side, I have added a line of code to handle a new magic `NOINDEX` keyword to the user doc extractor, which excludes the generated documentation from any indices in the model directory, and replaced the direct inclusion of recorder and recording backend documentation into the guide on recording from simulations by links to get the size of that page down.

This PR implements the changes discussed amongst the author and the suggested reviewers on December 12, 2020.

@jhnnsnk: please feel free to send me the example for switching recording backends you have presented at that meeting as a PR against my branch.

### DIFF
--- a/doc/extractor_userdocs.py
+++ b/doc/extractor_userdocs.py
@@ -372,6 +372,9 @@ def rst_index(hierarchy, current_tags=[], underlines='=-~', top=True):
             underlines = underlines[1:]
 
     for tags, items in sorted(hierarchy.items()):
+
+        if "NOINDEX" in tags:
+            continue
         if isinstance(tags, str):
             title = tags
         else:
--- a/doc/guides/connection_management.rst
+++ b/doc/guides/connection_management.rst
@@ -1,4 +1,4 @@
-.. _connection_mgnt:
+.. _connection_management:
 
 Connection Management
 =====================
--- a/doc/guides/recording_from_simulations.rst
+++ b/doc/guides/recording_from_simulations.rst
@@ -7,11 +7,11 @@ from neurons and synapses.
 
 To determine what happens to recorded data, each recording device can
 specify a *recording backend* in its ``record_to`` property. The
-default backend is *memory*, which just stores the recorded data in
-memory for later retrieval. Other backends write the data to file or
-to the screen. The different backends and their usage is explained in
-detail in the section about :ref:`Recording Backends
-<recording_backends>`.
+default backend is *memory*, which stores the recorded data in memory
+for later retrieval. Other backends write the data to file or to the
+screen, or stream it to other applications via the network. The
+different backends and their usage are explained in detail in the
+section about :ref:`Recording Backends <recording_backends>`.
 
 Recording devices can fundamentally be subdivided into two groups:
 
@@ -111,11 +111,11 @@ Recorders for every-day situations
 Where does data end up?
 -----------------------
 
-After a recording device has sampled or collected data, the *recording backends* are
-responsible for how the data are processed.
+After a recording device has sampled or collected data, the data is
+handed to a dedicated *recording backend*, set for each recorder.
+These are responsible for how the data are processed.
 
-
-Theoretically, recording backends are not restricted in what they do
+Theoretically, recording backends are completely free in what they do
 with the data. The ones included in NEST can collect data in memory,
 display it on the terminal, or write it to files.
 
--- a/doc/guides/recording_from_simulations.rst
+++ b/doc/guides/recording_from_simulations.rst
@@ -35,67 +35,9 @@ temporal aspects of the simulation loop.
    of thread execution, events are not necessarily recorded in
    chronological order.
 
-The time span in which a recorder is actually recording, can be specified
-using the additional properties ``start`` and ``stop``. These define
-activation and inactivation times of the device in ms. An additional
-property ``origin`` is available to shift the recording window by a
-certain amount, which can be useful in simulation protocols with
-repeated simulations. The following rules apply:
-
-- Collectors collect all events with timestamps `T` that fulfill
-
-  ::
-
-     start < T <= stop.
-
-  Events with timestamp `T == start` are not recorded.
 
 
-- Sampling devices sample at times *t = nh* (*h* being the simulation
-  resolution) with
 
-  ::
-
-     start < t <= stop
-     (t-start) mod interval == 0
-
-
-Common recorder properties
---------------------------
-
-All recorders have a set of common properties that can be set using
-``SetDefaults`` on the model class or ``SetStatus`` on a device
-instance:
-
-.. glossary::
-
- label
-   A string (default: `""`) specifying an arbitrary textual label for
-   the device.  Recording backends might use the label to generate
-   device specific identifiers like filenames and such.
-
- n_events
-   The number of events that were collected by the recorder can be
-   read out of the `n_events` entry. The number of events can be reset
-   to 0. Other values cannot be set.
-
- origin
-   A positive floating point number (default : `0.0`) used as the
-   reference time for `start` and `stop`.
-
- record_to
-   A string (default: `"memory"`) containing the name of the recording
-   backend where to write data to. An empty string turns all recording
-   of individual events off.
-
- start
-   A positive floating point number (default: `0.0`) specifying the
-   activation time in ms, relative to `origin`.
-
- stop
-   A floating point number (default: `infinity`) specifying the
-   deactication time in ms, relative to `origin`. The value of `stop`
-   must be greater than or equal to `start`
 
 Recorders for every-day situations
 ----------------------------------
--- a/doc/guides/recording_from_simulations.rst
+++ b/doc/guides/recording_from_simulations.rst
@@ -13,17 +13,6 @@ screen, or stream it to other applications via the network. The
 different backends and their usage are explained in detail in the
 section about :ref:`Recording Backends <recording_backends>`.
 
-Recording devices can fundamentally be subdivided into two groups:
-
-- **Collectors** collect events sent to them. Neurons are connected to
-  collectors and the collector collects the events emitted by the
-  neurons connected to it.
-
-- **Samplers** actively interrogate their targets at given time
-  intervals. This means that the sampler must be connected to the
-  neuron (not the neuron to the sampler), and that the neuron must
-  support the particular type of sampling.
-
 Recording devices can only reliably record data that was generated
 during the previous simulation time step interval. See the guide about
 :doc:`running simulations <running_simulations>` for details about the
@@ -35,8 +24,16 @@ temporal aspects of the simulation loop.
    of thread execution, events are not necessarily recorded in
    chronological order.
 
+Recording devices can fundamentally be subdivided into two groups:
 
+- **Collectors** collect events sent to them. Neurons are connected to
+  collectors and the collector collects the events emitted by the
+  neurons connected to it.
 
+- **Samplers** actively interrogate their targets at given time
+  intervals. This means that the sampler must be connected to the
+  neuron (not the neuron to the sampler), and that the neuron must
+  support the particular type of sampling.
 
 
 Recorders for every-day situations
--- a/doc/guides/recording_from_simulations.rst
+++ b/doc/guides/recording_from_simulations.rst
@@ -100,11 +100,9 @@ instance:
 Recorders for every-day situations
 ----------------------------------
 
-.. include:: ../models/multimeter.rst
-
-.. include:: ../models/spike_recorder.rst
-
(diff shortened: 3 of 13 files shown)


nestkernel/recording_device.h
```diff
@@ -42,6 +42,62 @@
 namespace nest
 {
 
+/* BeginUserDocs: NOINDEX
+
+Recording time window
++++++++++++++++++++++
+
+The time span during which the recorder actually records, can be
+specified using the properties ``start`` and ``stop``. These define
```
sarakonradi, line 51:
```suggestion
The time span during which the recorder actually records can be
specified using the properties ``start`` and ``stop``. These define
```

nestkernel/recording_device.h
```diff
@@ -42,6 +42,62 @@
...
+   of individual events off.
+
+ start
+   A positive floating point number (default: `0.0`) specifying the
+   activation time in ms, relative to `origin`.
+
+ stop
+   A floating point number (default: `infinity`) specifying the
+   deactication time in ms, relative to `origin`. The value of `stop`
+   must be greater than or equal to `start`
```
sarakonradi, line 97:
```suggestion
   must be greater than or equal to `start`.
```

nestkernel/recording_device.h
sarakonradi: Can we perhaps replace the word "actually" with "actively"?

sarakonradi (changes requested):
Thanks for restructuring the documentation! Just added a few comments.

doc/guides/recording_from_simulations.rst
```diff
@@ -35,87 +24,35 @@ temporal aspects of the simulation loop.
...
-   reference time for `start` and `stop`.
+Recording devices can fundamentally be subdivided into two groups:
 
- record_to
-   A string (default: `"memory"`) containing the name of the recording
-   backend where to write data to. An empty string turns all recording
-   of individual events off.
+- **Collectors** collect events sent to them. Neurons are connected to
+  collectors and the collector collects the events emitted by the
+  neurons connected to it.
```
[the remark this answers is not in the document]
jougs: Good idea.
```suggestion
- **Collectors** gather events sent to them. Neurons are connected to
  collectors and the collector gathers the events emitted by the
  neurons connected to it.
```

nestkernel/recording_device.h
jougs: Yes.
```suggestion
The time span during which the recorder actively records, can be
specified using the properties ``start`` and ``stop``. These define
```

doc/guides/connection_management.rst
```diff
@@ -1,4 +1,4 @@
-.. _connection_mgnt:
+.. _connection_management:
```
terhorstd, line 1: Is this required at all? I thought Sphinx automatically creates anchors for titles?

models/spike_recorder.h
```diff
@@ -64,6 +66,8 @@ spike creation rather than that of their arrival.
 The call to ``Connect`` will fail if the connection direction is
 reversed (i.e., connecting *sr* to *neurons*).
 
+.. include:: ../models/recording_device.rst
+
 EndUserDocs */
```
terhorstd, line 71: Add a See-also section to actually refer to other recorders
```suggestion
.. include:: ../models/recording_device.rst

See also
++++++++

EndUserDocs */
```

models/weight_recorder.h
```diff
@@ -71,6 +73,8 @@ synapses that fulfill the given criteria.
 
    >>> nest.Connect(pre, post, syn_spec="stdp_synapse_rec")
 
+.. include:: ../models/recording_device.rst
+
 EndUserDocs */
```
terhorstd, line 78: Add a See-also section to actually refer to other recorders
```suggestion
.. include:: ../models/recording_device.rst

See also
++++++++

EndUserDocs */
```

models/multimeter.h
```diff
@@ -110,6 +112,19 @@ fail if carried out in the wrong direction, i.e., trying to connect the
...
+ record_from
+   A list (default: `[]`) of parameters and state variables to sample
+   from the nodes, the multimeter is connected to. Potential
+   recordables are given in the corresponding model documentation.
+
+ interval
+   A float (default: `1.0`) specifying the interval in ms, at which
+   data is collected from the nodes, the multimeter is connected to.
+
 EndUserDocs */
```
terhorstd, line 128: Add a See-also section to actually refer to other recorders
```suggestion

See also
++++++++

EndUserDocs */
```

doc/extractor_userdocs.py
```diff
@@ -372,6 +372,9 @@ def mkitem(t):
             underlines = underlines[1:]
 
     for tags, items in sorted(hierarchy.items()):
+
+        if "NOINDEX" in tags:
+            continue
```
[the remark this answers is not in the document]
terhorstd: Just checked. Currently the included files do not have a "See also" section, so they do not link to the "NOINDEX" index. So all is fine here, until we decide to automatically add "See also" sections where missing (which we probably shouldn't anyway)

terhorstd (changes requested):
Thanks for the restructuring, it improves the recorders docs significantly!
Two small comments:
1. Maybe add a sentence to the "Recorder Properties" section, that this is what can be given in the `params` argument. There is no clear link between parameters and properties on the semantic level. An inline example including a parameter would probably be enough; details are in the "recording from simulations guide" anyway.

2. Add the "See also", sections in the relevant places to have the links to keyword indices, e.g. where suggested.
 
With those small additions it would be fine for me.

doc/guides/connection_management.rst
jougs: I think it does. However, I'd rather rely on a static name for references, instead of one that changes when the title changes...

doc/extractor_userdocs.py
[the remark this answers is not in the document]
jougs: Thanks for the clarification.

jougs:
@terhorstd: Thanks for your review. Can you please have another quick look?

@jhnnsnk: Any change to include your NumPy example in this one? If not, we can also do that in another PR.

jougs:
@terhorstd: here's an observation related to this issue, I'd like to have your opinion on:

I am still convinced that we need the compositionality of documentation files via the 'NOINDEX' tag and `.. include::` that is added in this PR. However, this now bites us a bit in the context of #1926, where the model documentation now just shows a useless include instead of the model properties. Do you think it would be smart to dissolve includes (worst case manually) before actually letting Sphinx create HTML from the RST? This way the resulting RST file would be more self-contained and completee and as a bonus there simply wouldn't be anything to be added or skipped from any index. Thanks!

jougs:
@clinssen created #1930 to describe the naming issue more concisely and suggest a solution also for the problem described [above](https://github.com/nest/nest-simulator/pull/1899#issuecomment-780485133).

clinssen:
Potentially, we could render the rST files (which contain the include directives) to another plaintext format, such as MarkDown or man-page (which should be supported by docutils). This could either happen offline (during installation) or online (when the "help" command is invoked).

jougs:
@clinssen: I have just tried several different tools (some based on [Docutils](https://docutils.sourceforge.io/), some on [pandoc](https://pandoc.org/)) for converting our `.rst` files to MarkDown. They alll failed. Either they replaced unknown roles (e.g. links to other documentation pages [`:doc:`]) by a `Not supported` error message, or they just skipped pertaining text blocks completely. Totally unacceptable. :-(

Then I tried [something different](https://stackoverflow.com/a/47130801). Sphinx apparently has a Markdown builder since some time and that somehow works better, but totally messes up any tables in the `.rst` file and also skips references and foot notes and at least in some cases equations seem to be missing from the output.

As Sphinx was my best bet in this game, I'm afraid that our best choice at this point is really to manually replace the includes and present the source `.rst`, as that is the only one guaranteed to be complete.

jougs:
@terhorstd, I know you're busy, but I'd really very much appreciate your pandoc-foo expertise in these matters. Come join us on the dark dirty scripting side ;-)

heplesser:
@jougs After discussions in recent days, what are the plans now for including material in the online help?

jougs:
Here's a summary of my recent activities:
* I've merged upstream master in 0e380e0 to resolve conflicts
* I've added a new PyNEST example in d590e7aa4 that shows how to use different recording backends;
  kudos to @jhnnsnk for the initial version
* regarding the issue with the includes, I have added a simple include-resolving script in 7c56528
   @terhorstd: could you please have a quick look?

jougs:
Could all of you please re-review? I think I have all points addressed.

jougs:
Thanks!

nestkernel/recording_device.h
```diff
@@ -42,6 +42,62 @@
 namespace nest
 {
 
+/* BeginUserDocs: NOINDEX
+
+Recording time window
++++++++++++++++++++++
+
+The time span during which the recorder actively records, can be
```
sarakonradi, line 50:
```suggestion
The time span during which the recorder actively records can be
```

pynest/examples/recording_demo.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*-
+#
+# recording_demo.py
```
sarakonradi, line 3: We should add this to ``doc/userdoc/examples/index.rst``.

sarakonradi (changes requested):
Many thanks @jougs! Here are two minor suggestions.

pynest/examples/recording_demo.py
jougs: Good idea! Done in af7a69bfc.

jougs:
@sarakonradi: I've just addressed all your concerns.

jougs:
@terhorstd: could you please have a quick look at the include-resolving script added in 7c56528? Thanks!

sarakonradi (approved):

doc/userdoc/guides/recording_from_simulations.rst
```diff
@@ -35,87 +24,35 @@ temporal aspects of the simulation loop.
...
 
 .. _recording_backends:
 
 Where does data end up?
 -----------------------
 
-After a recording device has sampled or collected data, the *recording backends* are
-responsible for how the data are processed.
-
+After a recording device has sampled or collected data, the data is
```
ackurth, line 51: Minor point: Maybe `collected and sampled data` is better than `sampled and collected data` to preserve the order in which the groups of recording devices are introduced.

nestkernel/recording_device.h
```diff
@@ -42,6 +42,62 @@
...
 {
 
+/* BeginUserDocs: NOINDEX
+
+Recording time window
++++++++++++++++++++++
+
+The time span during which the recorder actively records can be
+specified using the properties ``start`` and ``stop``. These define
+the activation period of the device in ms. An additional property
```
ackurth, line 52: If find `activation period` to be a bit ambiguous in this case since to my knowledge in a Neuroscientific context it's usually related to neural activation. Maybe `recording period` is better?

nestkernel/recording_device.h
```diff
@@ -42,6 +42,62 @@
...
+   backend where to write data to. An empty string turns all recording
+   of individual events off.
+
+ start
+   A positive floating point number (default: `0.0`) specifying the
+   activation time in ms, relative to `origin`.
+
+ stop
+   A floating point number (default: `infinity`) specifying the
+   deactication time in ms, relative to `origin`. The value of `stop`
```
ackurth, line 96:
```suggestion
   deactivation time in ms, relative to `origin`. The value of `stop`
```

ackurth (commented):
I think the documentation is very comprehensible, just added some minor comments.

doc/userdoc/guides/recording_from_simulations.rst
jougs, line 51:
```suggestion
After a recording device has collected or sampled data, the data is
```

nestkernel/recording_device.h
jougs, line 52:
```suggestion
the recording period of the device in ms. An additional property
```

jougs (commented):
Just adding suggestions to address @ackurth's review comments.

ackurth (approved):

clinssen (approved):
Looks good!

terhorstd (approved):
The include script is simple but efficient.
Looks good to me now.

Issue #1891, opened 2021-01-18
Catch bad_alloc in SLI main loop and print a friendly error message


jougs:
In case common C++ exceptions are raised somewhere in the NEST code, they are caught by the SLI main loop, and presented to users by printing only the name of the exception. For most exceptions this is probably an appropriate course of action. One where it probably isn't, is `std::bad_alloc`, which signals an out-of-memory condition. Recently, there were multiple reports on the mailing list related to that (e.g. [here](https://www.nest-simulator.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/IBK2L5DWO7OTEPB7OK4GKCJ45ECY4IQL/) and [here](https://www.nest-simulator.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/BFL2QQEL7SEUNN3CAS6WYKPN3K65OGIU)), indicating that users are indeed not able to associate the name of the exception with the cause.

I suggest to add a specific `catch` clause for `std::bad_alloc` to [`sli/interpret.cc`](https://github.com/nest/nest-simulator/blob/master/sli/interpret.cc#L1295-L1311) and print a more friendly error message that ideally also points to `memory_thisjob`, which can be used to measure the memory requirements of NEST from within a simulation script.

pnbabu:
```
import nest; 
nest.Create("iaf_psc_exp", 1000000000)
```
When executed on a Linux machine, the above code does raise an exception which is of the type `std::exception` with the `exception::what()` string set to `std::bad_alloc`. So adding a new catch block to specifically catch the exception of type`std::bad_alloc` is not helpful here as the control will never go to that block when there is an out of memory error. 

Executing the same code on Mac does not produce any exception with the current implementation. Mac OS seems to kill the process without displaying any error stack trace and only outputs the message `Killed: 9`.

So, we could retain the code as it is and maybe add documentation somewhere indicating that the `std::bad_alloc` error means an out of memory has occurred. Any suggestions on what is the right place to add it?

jougs:
@pnbabu: many thanks for your efforts and the thorough investigation. I think in this case, we should rather close this and go for better documentation indeed.

@sarakonradi: do we already have a list of error messages and what they actually mean? If not, we should! Could you maybe start such a list based on the content of [`nestkernel/exceptions.h`](https://github.com/nest/nest-simulator/blob/master/nestkernel/exceptions.h)? Thanks!

jougs:
@sarakonradi: friendly ping!

[...]

PR #1903, opened 2021-02-01
Add a catch clause for std::bad_alloc exception

pnbabu:


pnbabu:
Fixes #1891

clinssen:
Some initial observations:

- ``import nest; nest.Create("iaf_psc_exp", 1000000000)`` reliably triggers the exception.

- There is ``SLIInterpreter::execute_()`` and ``SLIInterpreter::execute_debug_()``. Your patch is only for the latter; please apply it also to the former.

- In these methods, there is a nested try..catch block, and the bad_alloc exception seems to come from the inner block, and is not re-thrown to be caught by the outer block. Thus, it looks like the bad_alloc catch should also be in the inner block, that is to say, here: https://github.com/nest/nest-simulator/blob/c2bd9d970927744e2c557e98aa81e153a032821c/sli/interpret.cc#L1291

heplesser:
This will be useful for Linux, even if corresponding errors are not caught on macOS because of a different approach to handling excess memory requests.

pnbabu:
@heplesser I agree that it would be useful for Linux, but as mentioned [here](https://github.com/nest/nest-simulator/issues/1891#issuecomment-772610925), the exception thrown is of the type `std::exception` and not `std::bad_alloc`, and the former is already caught and handled in the code. Since the specific exception, `std::bad_alloc` is not thrown, adding a specific catch block for it will not help. Instead, we could document this error somewhere. Do you have any other ideas?

clinssen:
I propose to close this PR, as we already have issue #1891 open, and the conclusion was drawn there that it should be a documentation update rather than a code update.

jougs:
@clinssen: I agree with you and will thus close this issue. @pnbabu: many thanks again for your investigations and work!

PR #1906, opened 2021-02-02
Remove colorize.rst from docs

jessica-mitchell:
This PR removes the unnecessary colorize from the docs. From the work of @jd41


sarakonradi (approved):
Just tested it. Thanks!

PR #1908, opened 2021-02-02
Add PyNEST function to obtain local VPs

heplesser:
This PR adds `nest.GetLocalVPs()`, which returns a `range()` object representing the VPs local to a given MPI rank. This can be useful in cases where users want to create VP-specific RNGs on the Python level (examples forthcoming in #1549). The information is not included in `GetKernelStatus()` because `range` objects cannot be serialised to JSON without turning them into potentially huge lists.

The function is entirely implemented on the Python level to be able to make use of a compact `range()` representation without introducing a new SLI datatype. To ensure synchrony with the kernel, a test checks that the returned VPs are correct. This test must run with multiple MPI processes, which is achieved using the `test_mpitest.py` setup from `test_sp`, which this PR lifts to the top level test directory.

The PR has high priority and milestone NEST 3 because it is needed to avoid some ugly constructs in random-number recommendations when using Python RNGs.


pynest/nest/lib/hl_api_parallel_computing.py
```diff
@@ -114,3 +115,13 @@ def SyncProcesses():
     """
 
     sr("SyncProcesses")
+
+
+@check_stack
+def GetLocalVPs():
+    """Return iterable representing the virtual processes local to the MPI rank."""
+
+    # Compute local VPs as range here based on round-robin logic in VPManager::get_vp().
```
jougs, line 124:
```suggestion
    # Compute local VPs as range based on round-robin logic in VPManager::get_vp().
```

pynest/nest/tests/mpitest_get_local_vps.py
```diff
@@ -0,0 +1,60 @@
...
+
+import nest
+import sys
+
+
+class LocalVPsTestCase():
+    """
+    Test local_vps field of kernel status.
+
+    This test ensure that the PyNEST-generated local_vps information
```
jougs, line 34:
```suggestion
    This test ensures that the PyNEST-generated local_vps information
```

jougs (changes requested):
Other than my two small suggestions, I'm fine with this in principle.

I still want to add a little side-note: In addition to moving the MPI test harness from the `test_sp` directory up to the top-level test directory, the file could be merged with [`test_connect_all_patterns.py`](https://github.com/nest/nest-simulator/blob/master/pynest/nest/tests/test_connect_all_patterns.py). I would not object, if that would happen in this PR.

heplesser:
@jougs Thanks for the quick review. I have accepted you suggestions and created #1917 as a follow-up for refactoring the MPI-dependent tests. I think a proper fix for that will take ever so slightly more time than is good for this PR.

jougs (approved):
Thanks!

heplesser:
@gtrensch Ping

heplesser:
I have added this and #1948 to the agenda of the Open Developer VC on 15 March.

gtrensch (approved):

PR #1909, opened 2021-02-03
Fix broken colors in documentation

jessica-mitchell:
Re-adding colorize.rst into the doc/ folder after removing in PR #1906. Without this file the color roles on Read the Docs do not work.


sarakonradi (approved):
Thanks for fixing the color roles! Tested this locally and on RtD, and everything seems to be working fine for me.

PR #1912, opened 2021-02-03
Update NEST 2.20.x CMake configuration to support OpenMP fully on OSX

heplesser:
This is a backport from NEST3 to make sure threaded NEST compiles fine with Clang on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,9 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>
 
-cmake_minimum_required( VERSION 3.1 )
+# Version range from minimum required (3.12 for macOS OpenMP)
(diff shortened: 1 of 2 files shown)


jougs (approved):
LGTM. Merging directly as this is only a backport.

jougs:
@heplesser: Is the failure on MacOS expected (e.g. because of backporting), or is this related to this change? Thanks!

heplesser:
@jougs The macOS failure occurs while installing NumPy, definitely not relate to this backport. We may need to backport some more changes to our Travis setup. I will check.

jougs:
We had very similar issues when doing the 2.20.1 release and decided it's not worth backporting the whole CI fixes when everything is in fact fine on any normal user's machine. I would thus be happy to merge without any further actions applied. @abigailm: what do you think?

abigailm:
Then I will go ahead and merge it and we worry about MacOS and NumPy at a suitable juncture.

PR #1914, opened 2021-02-04
Add a Python script reporting Python imports from the NEST code base.

heplesser:
Just a little and imperfect helper, but useful to create requirements files, I believe.

### DIFF
--- a/extras/find_imports.py
+++ b/extras/find_imports.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
 # find_imports.py
@@ -25,20 +26,37 @@ import sys
 import re
 from collections import defaultdict
 
-
 """Script to report all Python modules imported in NEST
 
-Run this script in the NEST source directory to obtain a
-listing of external modules imported by NEST and the part 
-importing them for those not part of the NEST core.
+This script parses the NEST code base to find all external
+modules imported by NEST, i.e., modules not included in the
+Python Standard Library or as part of NEST itself.
+
+For scripts not included by core parts of NEST (PyNEST
+itself, its examples and tests), the script also reports
+which part of the NEST codebase requires the module.
+
+The script scans the directory given by the NEST_SOURCE
+environment variable if defined and the current working
+directory otherwise.
 
 This may be useful to define requirements.
 
 This is still work in progress and may return false positives.
 """
 
-# directory to parse
-source_dir = '.'
+if len(sys.argv) == 1:
+    if 'NEST_SOURCE' in os.environ:
+        source_dir = os.environ['NEST_SOURCE']
+    else:
+        print('I do not know which directory to parse: ')
+        print('No directory given on the command line and NEST_SOURCE undefined.')
+        sys.exit(1)
+elif len(sys.argv) == 2:
+    source_dir = sys.argv[1]
+else:
+    print(f'\nUsage: {sys.argv[0]} [path to scan]\n')
+    sys.exit(2)
 
 # subdirectories to ignore
 exclude_dirs = [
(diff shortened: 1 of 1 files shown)


extras/find_imports.py
```diff
@@ -0,0 +1,136 @@
...
+listing of external modules imported by NEST and the part 
+importing them for those not part of the NEST core.
+
+This may be useful to define requirements.
+
+This is still work in progress and may return false positives.
+"""
+
+# directory to parse
+source_dir = '.'
```
jougs, line 41: Most other checkers (e.g. [here](https://github.com/nest/nest-simulator/blob/master/extras/check_unused_names.py#L64), [here](https://github.com/nest/nest-simulator/blob/master/extras/check_copyright_headers.py#L58)) use `os.environ['NEST_SOURCE']` to get the path from the outside to make them independent of the working directory when calling. Can you do something similar here? If yes, please also update the docstring above.

extras/find_imports.py
```diff
@@ -0,0 +1,136 @@
...
+import os
+import sys
+import re
+from collections import defaultdict
+
+
+"""Script to report all Python modules imported in NEST
+
+Run this script in the NEST source directory to obtain a
+listing of external modules imported by NEST and the part
```
jougs, line 32: I don't quite understand what `part` here refers to. Can you please clarify? Thanks!

extras/find_imports.py
```diff
@@ -0,0 +1,136 @@
...
+import_re = re.compile('\s*(import|from)\s+(\w+)')
+
+imports = defaultdict(set)
+for dirpath, _, fnames in os.walk(source_dir):
+
+    if any([exclude_dir in dirpath for exclude_dir in exclude_dirs]):
+        continue
+
+    for fname in fnames:
+        if not fname.endswith('.py'):
```
jougs, line 109: Do we also have to scan [`pynestkernel.pxd`](https://github.com/nest/nest-simulator/blob/master/pynest/pynestkernel.pxd) and [`pynestkernel.pyx`](https://github.com/nest/nest-simulator/blob/master/pynest/pynestkernel.pyx)?

extras/find_imports.py
```diff
@@ -0,0 +1,136 @@
...
+             f'{source_dir}/pynest/tests', f'{source_dir}/testsuite'}
+
+# server package
+server_dir = f'{source_dir}/pynest/nest/server'
+
+# Match for import lines, picks up only top-level package and ignores relative imports
+# We need the \s* at the beginning because some imports are nested in try-blocks. This
+# can lead to false positives if a comment line or multiline string line begins with
+# "import" or "from".
+import_re = re.compile('\s*(import|from)\s+(\w+)')
```
jougs, line 100: Would this also match `import modulea, moduleb`?

extras/find_imports.py
heplesser: Improved.

extras/find_imports.py
heplesser: They do not include anything not included in Cython or "the base" and I assume they will stay very stable, so I left them out to avoid complicating the test here.

extras/find_imports.py
heplesser: It would only catch `modulea`, but it is bad style anyways. I just searched the code base manually and didn't find a single occurrence of this pattern.

extras/find_imports.py
jougs: Alright.

extras/find_imports.py
jougs: Fine by me. It might still be good to add this information to the comment, so we don't have to rediscover it manually later on.

heplesser: Done.

jougs (approved):

steffengraber (approved):
Really very useful!

PR #1921, opened 2021-02-09
Clarify the post-pre convention used in plot-weight-matrices example documentation

jessica-mitchell:
This PR is in response to a mailing list question asking to clarify the notation in the weight matrices example. I basically took what Renato said and added that to the documentation.


terhorstd (approved):
LGTM :+1:

Issue #1905, opened 2021-02-02
Sort out and document CMake targets


jougs:
We currently have a bunch of different `make` targets, which are partly the result of a) using CMake defaults, b) carrying over defaults from the time when the build system still used autotools, and c) custom targets (some of which have bad names for historic reasons).

### `all`
* compile all source files and link static and dynamic libraries, as well as the executables
* cythonize PyNEST and compile it;

### `install`
* run the [SLI help extractor](https://github.com/nest/nest-simulator/tree/master/extras/help_generator)

install everything to directories under the `<nest_install_dir>`
* [SLI standard library](https://github.com/nest/nest-simulator/tree/master/lib/sli), [SLI examples](https://github.com/nest/nest-simulator/tree/master/examples/nest), extracted SLI help files
* dynamic libraries, executables (`nest`, `sli`) and helpers (`nest-config`, `nest-vars.sh`)
* PyNEST and [its examples](https://github.com/nest/nest-simulator/tree/master/pynest/examples)
* the [testsuite](https://github.com/nest/nest-simulator/tree/master/testsuite)

### `install-nodoc`
* same as `install`, but without running the SLI help extractor

### `installcheck`
run the testsuite (essentially by calling [`do_tests.sh`](https://github.com/nest/nest-simulator/blob/master/testsuite/do_tests.sh)) using the installed executables and the test files from `<nest_install_dir>/share/doc/nest`

### `doc` and `fulldoc`
run Doxygen to extract the C++-level user documentation in either
* a [minimal configuration](https://github.com/nest/nest-simulator/blob/master/doc/normaldoc.conf.in) (`doc`), or in
* a [extended configuration](https://github.com/nest/nest-simulator/blob/master/doc/fulldoc.conf.in) (`fulldoc`) including class inheritance diagrams and all bling.

### `html`
run the Sphinx-based user-level documentation generation pipeline. This also runs the doc extractor that gets the model documentation snippets from the C++ files that implement them.

## Proposal

There is some (well justified) confusion between what these targets actually do. A proposal for a more sane naming and task split would probably be

* **`all`** can stay
* **`install`** would do the same as before minus running the doc-extractor
* **`userdoc`** would do what `html` currently does, plus run the SLI doc extractor
* **`devdoc`** would do what `fulldoc` currently does
* **`install-userdoc`** would install what **`userdoc`** has produced
* **`install-devdoc`** would install what **`devdoc`** has produced

This issue is also related to #291.

[...]

PR #1926, opened 2021-02-17
Fix PyNest help and helpdesk commands

### LINKED ISSUE #1904: PyNEST's help() function does not display model documentation anymore
Since the dawn of the *doc extractor* and new RST-based model documentation (#1355, #1530), the model documentation is no longer accessible from the PyNEST function `nest.help()` because that only wraps around the SLI command `help` and thus does not "see" the new documentation files.

To solve this issue, the PyNEST help function has to be updated to display the new help texts for models and ideally provide a way to let users decide if they want the documentation as plain RST or rendered HTML.

jougs:
This fixes #1904 by making `helpdesk()` open a link to the Sphinx-generated HTML help index page and the `help()` function operate on the reStructuredText files extracted from the models. I've also simplified the handling of the pager by just using the one provided by `pydoc` instead of from SLI. All pertaining documentation files have been updated.

This PR is based on the branch behind #1843 and should thus not be merged before that. For the same reason, I'm opening this in draft mode, as the respective reviews should happen independently and this one is currently a bit too cluttered for this.

@babsey: can you maybe also quickly give this a try in the context of NEST Desktop? Thanks!


babsey:
> can you maybe also quickly give this a try in the context of NEST Desktop? Thanks!

It works with curl and in NEST Desktop. That is great. Thanks!
In both cases, I am able to fetch documentation text via nest-server:
`localhost:5000/api/help?return_text=true&obj=iaf_psc_alpha`.

However, I have to execute command `make html` before `make install`.
I figured out that several models have no help text: voltmeter, static_synapse.

Is it on purpose?

jougs:
@babsey: Thanks for testing so quickly and for your feedback :-)

We might have to point out the need for `make html` more visibly (but things are in flux, see also #1905). We could probably just return a corresponding error message instead of the actual help text from `help()` in case no help text is found.

Regarding the missing help files:
* internally, the `voltmeter` is now just a pre-configured version of a `multimeter` and therefore does not have its own documentation (anymore). We should probably create a minimal help text for it that points users in the right direction.
* the problem with synapse documentation is more fundamental: our [doc extractor](https://github.com/nest/nest-simulator/blob/master/doc/extractor_userdocs.py) creates the `.rst` file with the same base name as the C++ header file, which is in turn based on the name of the class inside of it. This is unfortunate, because synapse model classes are called `*Connection`, but get registered as `*_synapse`. Long story short: Currently, you will get the help the help using `nest.help('static_connection')`, but that should definitely change.

@terhorstd: What's your opinion on the last point here? Rename all files (and live with file names that are inconsistent with the contained classes' names) or add enable the doc extractor to process a filename hint in the `BeginUserDocs` block?

pynest/nest/lib/hl_api_info.py
```diff
@@ -66,14 +66,24 @@ def authors():
...
+    Please note that the help pages will only be available if you ran
+    ``make html`` prior to installing NEST. For more details, see
+    :ref:`documentation_workflow`.
 
     """
 
-    nestdocdir = sli_func("statusdict/prgdocdir ::")
-    helpfile = os.path.join(nestdocdir, 'help', 'helpindex.html')
+    docdir = sli_func("statusdict/prgdocdir ::")
+    help_fname = os.path.join(docdir, 'html', 'index.html')
```
heplesser, line 81: You changed the filename here to `help_fname`, but on lines 81, 86 and 88 below it is still `helpfile`.

On macOS, the problem mention on line 83f has been fixed long ago, so we no longer need to treat macOS special. On the other hand, on my machine it seems that passing just the filename to `open_new()` has no effect, it has to be a proper URL beginning with `file://`. Documentation for `open_new()` also states that the argument should be a URL (Python 3.8.5).

heplesser (commented):
@jougs Thanks! I will wait with a full review until #1843 is merged, since that should reduce the delta quite a bit. I added one comment below.

And then a general question: Most users will be online most of the time. Why not open the helpdesk at readthedocs? The one downside is that `open_new()` does not tell you if the browser was able to open the webpage, one would have to check that in some other way.

jougs:
@clinssen created #1930 to describe the naming issue more concisely and suggest a solution also for the problem described in the second bullet point [above](https://github.com/nest/nest-simulator/pull/1926#issuecomment-780482157).

pynest/nest/lib/hl_api_info.py
jougs: Too many code paths to test... Sorry. Fixed in 5a19331.

I've also added `file://` to the path and as this also seems to work on Linux, I've removed the conditional check on the operating system.

jougs:
Replying to @heplesser:
> And then a general question: Most users will be online most of the time. Why not open the helpdesk at readthedocs? The one downside is that `open_new()` does not tell you if the browser was able to open the webpage, one would have to check that in some other way.

I would think that opening the browser worked reliably enough (at least if we do it right, i.e. using the `file://` prefix you pointed out). It's part of the standard library after all. The real downside I see with opening the version on the web is that that would get users the help for some release version, while the local one is guaranteed to be for the version they are actually running. So I would rather not change this.

Replying to @babsey:
> However, I have to execute command make html before make install. I figured out that several models have no help text: voltmeter, static_synapse. Is it on purpose?

I've added yet another suggestion to #1905, which would remove the requirement for calling `make userdoc` (aka `html`) and still leave enough flexibility for developers. Any input on that would be welcome!

pynest/nest/lib/hl_api_info.py
```diff
@@ -84,18 +84,15 @@ def helpdesk():
...
         raise FileNotFoundError(msg)
-
+    
+    helpdesk_url = "file://" + help_fname
+    
     # Under Windows systems webbrowser.open is incomplete
     # See <https://bugs.python.org/issue8232>
     if sys.platform[:3] == "win":
-        os.startfile(helpfile)
+        os.startfile(helpdesk_url)
```
heplesser, line 93: @jougs I just tested  under Win10 and `webrowser.open_new()` works even there (Py 3.7.9), so I suggest we remove the special case for Windows here. It is anyways academic since we currently don't support NEST on Windows, and this entire mechanism requires a running NEST.

heplesser:
> I would think that opening the browser worked reliably enough (at least if we do it right, i.e. using the `file://` prefix you pointed out). It's part of the standard library after all. The real downside I see with opening the version on the web is that that would get users the help for some release version, while the local one is guaranteed to be for the version they are actually running. So I would rather not change this.

Good point, let's leave it as it is.

pynest/nest/lib/hl_api_info.py
jougs: OK. Thanks for testing. I've removed the workaround in 9798864.

jessica-mitchell:
@jougs I just took at look at the help and it seems to work well for me too. Thanks for all your work on that.

I know there is still ongoing discussion about how `cmake` will be organized, but I agree with what you said above regarding a useful error message, if a person doesnt `make html` before `make install`. Also, to make sure that the sequence of steps is clearly visible on the CLI before you start with  `make`

On a slightly different note, when one prints nest.help(), you get a basic list of a few commands, but I was wondering if it would be helpful to add at least a specific example for getting the model info: `nest.help('model_name') `and the api info:   `help(nest.Create)` because they have two different syntaxes. And I assume those are probably the pieces of information most users want.

jougs:
@jessica-mitchell: I've implemented your suggestion in [97dc9c3](https://github.com/nest/nest-simulator/pull/1926/files#diff-9e6ec63e0be9fc1328483d098df93d6023d8058c66eeb9c68c225597694b0d0fR22-R37). Thanks!

jougs:
Since #1843 was merged, this PR is based on the branch behind #1955. I'll undraft when that is merged.

jessica-mitchell:
@jougs I have been testing this and I can't get the CLI help to work anymore - I keep getting "file not found error" for nest.helpdesk() and nest.help('model'). After `cmake` I tried running `make`, `make html`, `make install` as well as `make html`, `make`, `make install`, but neither worked. Am I missing something?
(Version: fix-pynest-help@b82dad02e).

heplesser:
I observe the same problem as @jessica-mitchell. It seems that the correct `help_fname` is built and that `html` is built in the build directory, but not installed (applies to master and your fix-pynest-help branch):

```
In [13]: ls /Users/plesser/NEST/bld_foo2/install/share/doc/nest
ConnPlotter/     README.md        help/            model_details/   mpitests/        regressiontests/ selftests/
LICENSE          examples/        logos/           mpi_selftests/   musictests/      run_examples.sh  unittests/
```

I am quite certain that local html-installation worked with #1843. Could #1955 have broken something?

jougs:
Grrr! Yes, #1955 broke it. df7a476 should fix it.

heplesser:
@jougs Works nicely for me now :). I sent you a PR with some touch-ups for this one, hope you'll like it.

jessica-mitchell (commented):
@jougs Thanks, it's working nicely. I like the output on the nest.help(). My only concern is that it's not obvious you need to `make html` to get help working on the CLI (unless you didn't run it, and the error pops up) but maybe we should add something to the installation instructions and the  output after cmake

jessica-mitchell:
Also @jougs,  @stinebuu pointed out in #1961 that there is a broken image link (right before the first note) in the user-level documentation workflow page. Maybe that can be addressed here since it should be an easy fix?

jougs:
@jessica-mitchell: the broken link is fixed by commit d7e20064e. This PR thus also fixes #1961.

jougs:
@lekshmideepu: sorry to bother you with this, but could you please have a look at the failing GitHub Action jobs? It seems that some of the sources for `apt` are not working anymore. #1899 has more failing builds, in case you need more samples... Thanks!

sarakonradi:
Just a brief update: @jessica-mitchell and I agreed to address the image links in a separate PR so that the graphics can work sooner rather than later. #1979 fixes them.

heplesser (approved):

jessica-mitchell (approved):
I have no further changes, thanks

PR #1927, opened 2021-02-17
Using Github actions as CI

lekshmideepu:
Introducing Github actions as the CI tool for NEST builds.

### DIFF
--- a/extras/gha_build.sh
+++ b/extras/gha_build.sh
@@ -1,0 +1,352 @@
+#!/bin/bash
+
+# gha_build.sh
+#
+# This file is part of NEST.
+#
+# Copyright (C) 2004 The NEST Initiative
+#
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+# This shell script is part of the NEST Travis CI build and test environment.
+# It is invoked by the top-level Travis script '.travis.yml'.
+#
+# NOTE: This shell script is tightly coupled to 'extras/parse_travis_log.py'.
+#       Any changes to message numbers (MSGBLDnnnn) or the variable name
+#      'file_names' have effects on the build/test-log parsing process.
+
+
+# Exit shell if any subcommand or pipline returns a non-zero status.
+set -ex
+
+env
+if [ "$xNEST_BUILD_COMPILER" = "CLANG" ]; then
+    export CC=clang-7
+    export CXX=clang++-7
+fi
+
+NEST_VPATH=build
+mkdir -p "$NEST_VPATH/reports"
+
+if [ "$xNEST_BUILD_TYPE" = "STATIC_CODE_ANALYSIS" ]; then
+    echo "+ + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +"
+    echo "+               S T A T I C   C O D E   A N A L Y S I S                       +"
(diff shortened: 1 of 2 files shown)


extras/gha_build.sh
```diff
@@ -0,0 +1,352 @@
...
+#       Any changes to message numbers (MSGBLDnnnn) or the variable name
+#      'file_names' have effects on the build/test-log parsing process.
+
+
+# Exit shell if any subcommand or pipline returns a non-zero status.
+set -ex
+
+env
+if [ "$xNEST_BUILD_COMPILER" = "CLANG" ]; then
+    export CC=clang-7
```
heplesser, line 36: clang-7 seems rather old, the current version is 11. Is there a specific reason to use clang-7?

heplesser:
@lekshmideepu Thanks a lot for this! I am just wondering about one thing (probably something I have not quite understood yet ...): Why do I only see the Travis checks running, but no Github actions? Where would they appear?

lekshmideepu:
> @lekshmideepu Thanks a lot for this! I am just wondering about one thing (probably something I have not quite understood yet ...): Why do I only see the Travis checks running, but no Github actions? Where would they appear?

I was also wondering the same. Ideally it should run under Actions of nest-simulator. I could see that it runs on my repo and not here. Not sure why it is happening. Strange. I am trying to figure this out.

extras/gha_build.sh
lekshmideepu: No specific reason. Thanks for pointing this out. I just added a clang-11 build.

lekshmideepu:
> @lekshmideepu Thanks a lot for this! I am just wondering about one thing (probably something I have not quite understood yet ...): Why do I only see the Travis checks running, but no Github actions? Where would they appear?

I figured this out. I forgot to add the keyword "on pull request". Sorry.  It is now working as expected :-)

heplesser:
@lekshmideepu Great! I have just three suggestions concerning the build matrix before we merge this: 
- We should return to a full build on macOS again (without MUSIC, I think it is still difficult to install on macOS).
- To keep resource use within bounds, I would drop the macOS gcc build.
- I would also consider to drop the clang build on Ubuntu, given that we test against clang on macOS. That is unless there were specific reasons for introducing this combination. 
The last two suggestions are just meant to keep the CO2-footprint of NEST development bounded ;).

lekshmideepu:
Valid suggestions. I agree.

heplesser:
@lekshmideepu I see that the macOS full build fails due to OpenMP problems. Fixing this is a separate issue, so I would suggest that we for this PR stick to the macOS Minimal build after all and create a follow-up for re-instating a full macOS case.

lekshmideepu:
Ok. I think that would be the best solution. Will revert it back.

heplesser (approved):

extras/gha_build.sh
```diff
@@ -0,0 +1,352 @@
...
+    CONFIGURE_MPI="-Dwith-mpi=ON"
+else
+    CONFIGURE_MPI="-Dwith-mpi=OFF"
+fi
+
+if [ "$xPYTHON" = "1" ] ; then
+    export PYTHON_INCLUDE_DIR=`python3 -c "import sysconfig; print(sysconfig.get_path('include'))"`
+    export PYLIB_BASE=lib`basename $PYTHON_INCLUDE_DIR`
+    export PYLIB_DIR=$(dirname `sed 's/include/lib/' <<< $PYTHON_INCLUDE_DIR`)
+    export PYTHON_LIBRARY=`find $PYLIB_DIR \( -name $PYLIB_BASE.so -o -name $PYLIB_BASE.dylib \) -print -quit`
```
jougs, line 230: I'm impressed to see that our detection mechanism is even stable enough to work on a completely different CI system!

:smile_cat:

jougs (approved):
Although I think that our internal variables (e.g. `xNEST_BUILD_TYPE`) would be fine without the leading `x` and others (e.g. `xGSL`) should better be called `NEST_BUILD_FLAG_...` I think we should merge this quickly and improve on such things later on.

In this sense: full approval and again many, many thanks for your hard work and the uncounted iterations this one took you.

extras/gha_build.sh
lekshmideepu: > Although I think that our internal variables (e.g. `xNEST_BUILD_TYPE`) would be fine without the leading `x` and others (e.g. `xGSL`) should better be called `NEST_BUILD_FLAG_...` I think we should merge this quickly and improve on such things later on.
> 
> In this sense: full approval and again many, many thanks for your hard work and the uncounted iterations this one took you.

Thank you very much! :-)

Issue #1928, opened 2021-02-17
Wrong cmake argument with-python=3 and PyThreadState has no member named exc_traceback


dizcza:
I'm building NEST from Younes' branch https://github.com/YounesBouhadjar/nest-simulator/tree/stdsp_synapse to check his project. But the same issue is in the master branch (see part 2).

## Part 1

**Describe the flaw**
NEST docs state that if you want to build NEST with Python 3 support, you need to pass `-Dwith-python=3` in a cmake build command line. Doing so breaks the cmake.

**Suggest an improvement**
Remove this note from the documentation. Python support is managed by default in NEST install, according to https://github.com/nest/nest-simulator/blob/master/INSTALL, and there is nothing about the version of a Python. I suppose, cmake handles this implicitly and the users don't need to care about which Python version they are using as long as cmake can find the python libs.
Not specifying anything about the python I'm using allowed me to successfully run the cmake command.

**Supporting materials**
Cmake command that didn't work.
<details>

```
$ cmake -Dwith-python=3  -DCMAKE_INSTALL_PREFIX=. ..
-- The CXX compiler identification is GNU 7.5.0
-- The C compiler identification is GNU 7.5.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Doxygen: /usr/bin/doxygen (found version "1.8.13") found components: doxygen dot 
-- Looking for include file inttypes.h
-- Looking for include file inttypes.h - found
-- Looking for include file mach-o/dyld.h
-- Looking for include file mach-o/dyld.h - not found
-- Looking for include file mach/mach.h
-- Looking for include file mach/mach.h - not found
-- Looking for include file memory.h
-- Looking for include file memory.h - found
-- Looking for include file stdint.h
-- Looking for include file stdint.h - found
-- Looking for include file sys/types.h
-- Looking for include file sys/types.h - found
-- Looking for C++ include istream
-- Looking for C++ include istream - found
-- Looking for C++ include ostream
-- Looking for C++ include ostream - found
-- Looking for C++ include sstream
-- Looking for C++ include sstream - found
[...]

PR #1932, opened 2021-02-19
Enforce Python and Cython version requirements during configuration

hakonsbm:
Because of f-strings, NEST now requires Python version >=3.6. This PR enforces this requirement during configuration. 

Additionally, if Python version >=3.7 is used, a suitably recent Cython version must also be used. This is because the C api is changed with Python 3.7 (see also #1928). 

@jessica-mitchell @sarakonradi Should the Cython version requirement also be mentioned in the installation guide?

### DIFF
--- a/cmake/ProcessOptions.cmake
+++ b/cmake/ProcessOptions.cmake
@@ -398,7 +398,7 @@ function( NEST_PROCESS_WITH_PYTHON )
   elseif ( ${with-python} STREQUAL "ON" )
 
     # Localize the Python interpreter
-    find_package( PythonInterp 3 REQUIRED )
+    find_package( PythonInterp 3.6 REQUIRED )
 
     if ( PYTHONINTERP_FOUND )
       set( PYTHONINTERP_FOUND "${PYTHONINTERP_FOUND}" PARENT_SCOPE )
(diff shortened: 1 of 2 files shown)


heplesser:
@hakonsbm @jougs @terhorstd This PR makes me wonder if we should hike the version requirement for NEST to Python 3.7, especially because of the C-API-change in Cython. Python 3.7 was released in June 2018, so it is definitely not cutting edge and support will end in Dec 2021. I'd even suggest to go straight to requiring Python 3.8, released in November 2019, because it would allow us to use assignment expressions. Or would that cause problems on some systems, e.g. Fugaku. I'll add it to the agenda for the developer VC on 1 March.

jessica-mitchell:
@hakonsbm yes, I think in the instructions for installing from source we should provide the minimum requirement

jougs:
@heplesser: I think we can safely go with requiring 3.8 as you suggest. Special systems might need manual compilations then, but those are anyway only used by experts, which normally know their way around. For NESTML, we also require 3.8 since https://github.com/nest/nestml/pull/623 was merged.

hakonsbm:
Closing as this PR is superseded by #1954.

PR #1937, opened 2021-02-23
Hike Boost minimum version to 1.69 

heplesser:
This is due to change in Boost sort algoirhtm with version 1.69, see #1502. 

CI uses 1.72 since #1502 and current version is 1.75, so 1.69 is a conservative choice.


cmake/ProcessOptions.cmake
```diff
@@ -596,7 +596,8 @@ function( NEST_PROCESS_WITH_BOOST )
     set(Boost_USE_DEBUG_LIBS OFF)  # ignore debug libs
     set(Boost_USE_RELEASE_LIBS ON) # only find release libs
     # Needs Boost version >=1.62.0 to use Boost sorting, JUNIT logging
-    find_package( Boost 1.62.0 )
+    # Require Boost versoin >=1.69.0 due to change in Boost sort
```
hakonsbm, line 599:
```suggestion
    # Require Boost version >=1.69.0 due to change in Boost sort
```

hakonsbm (changes requested):

hakonsbm (approved):

terhorstd (approved):
LGTM :+1:

PR #1938, opened 2021-02-23
Remove neuronview example

heplesser:
NEST examples contain a "neuronview" directory providing some GUI based on Glade. As far as I can see, it has been around since at least NEST 2.6.1, but has not been maintained since. With the advent of NEST Server and Desktop, this examples seems outdated and this PR removes it.


jessica-mitchell (approved):
This has never been added to the userdocs, so it wasnt very discoverable but if it's not needed then I am fine with removing it.

jougs (approved):
I was the one who made this. A long time ago. But this is indeed superseded by NEST Desktop.

R.I.P. :coffin: :rose:

Issue #1648, opened 2020-06-16
Automatically include "Acknowledgments" text in doc/index.rst


sarakonradi:
This issue follows up on the discussion between @terhorstd, @heplesser and @steffengraber in #1646. 

Here, we can discuss possible solutions for automatically including the text from ``ACKNOWLEDGMENTS.md`` in the last subsection of the welcome page.

heplesser:
Fixed by #1940.

PR #1940, opened 2021-02-24
Include ACKNOWLEDGMENTS.md directly into main page of ReadtheDocs 

jessica-mitchell:
Currently, the acknowledgments is copy pasted into the userdocs, and ideally we would like a single source for content.
I removed the hard-coded text in doc/index.rst and added an include directive to the ACKNOWLEDGMENTS.md file
This resolves issue #1648


terhorstd (approved):
Discussed in doc team and seems the most viable solution without duplicating the list.

sarakonradi (changes requested):
@jessica-mitchell Thanks! 

Just tested this, and the acknowledgments do not appear on the welcome page. This probably has to do with #1843.

jessica-mitchell:
@sarakonradi thanks for pointing this out
 #1843 breaks this pretty badly actually- I thought only the path needed updating, but it appears that the include directive cannot find files outside of userdoc/ - (if I move ACKNOWLEDGMENTS into userdoc/ the  include works fine)
I think this has to do with how the build directory is set up because the ACKNOWLEDGMENTS file isn't actually in there. @jougs do you have an idea of how to solve this?

jessica-mitchell:
@steffengraber @terhorstd @sarakonradi 
I checked and the ACKNOWLEDGMENTS does show up on ReadtheDocs, even though it does not work on a local build (see comment above). I have a question regarding how we want to deal with this issue because this means that the local build no longer reflects the output from RtD.  For this PR, should we approve since RtD works and the change is rather small?

jougs:
> @sarakonradi thanks for pointing this out
> #1843 breaks this pretty badly actually- I thought only the path needed updating, but it appears that the include directive cannot find files outside of userdoc/ - (if I move ACKNOWLEDGMENTS into userdoc/ the include works fine)
> I think this has to do with how the build directory is set up because the ACKNOWLEDGMENTS file isn't actually in there. @jougs do you have an idea of how to solve this?

I see two options:
- copy the file from within [`doc/CMakeList.txt`] to copy the file to `doc/userdoc` (along the lines of line 95)
- copy the file to the right location from within `doc/userdoc/conf.py` as we already do for the files of the microcircuit example

I'd prefer if this were fixed before merging this PR.

steffengraber:
@jessica-mitchell in conf.py thre is `source_suffix = '.rst'` (line 50) and ACKNOWLEDGMENTS is `.md`, maybe that is the problem.

jessica-mitchell:
@jougs I copied the location in the conf.py and that seems to work locally and on RtD, thanks for the help.  I think (if I understand correctly) if I used the CMakelists.txt, we'd still have the same problem because in that case I have to update the path for ACKNOWLEDGMENTS in the include directive, so it's in userdoc/ once we run `cmake` but then it would fail when building on RtD because that's not the source location.

jougs (approved):
LGTM! Thanks!

sarakonradi (approved):
Thanks for fixing this!

PR #1942, opened 2021-02-25
fix setting synapse parameters when running connect_arrays multithreaded

### LINKED ISSUE #1934: Running Connect() multithreaded fails when ``receptor_type`` is specified
Minimal reproducer:
```Python
import numpy as np
import nest

nest.ResetKernel()
nest.SetKernelStatus({'local_num_threads': 4})
neurons = nest.Create("iaf_psc_exp", 30)
source = 1 + np.arange(30)
target = 1 + np.arange(30)
weights = len(source) * [2.]
nest.Connect(source, target,
            'one_to_one',
#            {'weight': weights})       # WORKS OK
            {'receptor_type': 0, 'weight': weights})   # SEGFAULTS
```
Sometimes works, but usually segfaults. Could not reproduce segfaults when omitting ``receptor_type`` from the parameters dictionary. Could not reproduce segfaults with ``receptor_type`` specified, but number of threads set to 1.

NEST version: nest-master @ 84ea0f7

stinebuu:
This fixes #1934.

The datum types of the synapse param dictionaries need to be specified before we enter the parallel session, and updated correctly later on.

Work done by @hakonsbm and me.

@clinssen could you check and see if your problem persists?


YounesBouhadjar:
Upon @clinssen request, I tested this and I'm getting another kind of error. It's throwing BadParameter error for some network realizations.

See this example:

```
import numpy as np
import nest

nest.ResetKernel()
nest.SetKernelStatus({'local_num_threads': 4})
neurons = nest.Create("iaf_psc_exp", 10)

source = [7, 8, 8, 2, 3, 2, 4, 7, 2, 3]
target = [6, 2, 10, 3, 7, 1, 9, 5, 8, 4]
weight = [2., 2., 2., 2., 2., 2., 2., 2., 2., 2.]

nest.Connect(source, target,
            'one_to_one',
#            {'weight': weight})       # WORKS OK
            {'receptor_type': 0, 'weight': weight})   # BadParameter in SLI function connect_arrays
```
I found this specific ordering of the neuron IDs of the source and target running another script, which creates a random network, stores and restores the connectivity, and basically, I just took the above source and target IDs from a failed run of this script.

YounesBouhadjar:
  File "test_multithreaded_connect.py", line 16, in <module>
    nest.Connect(source, target,
  File "/home/bouhadjar/Documents/repo/nest-sim-doc/build/fix_1934_connect_arrays/lib/python3.8/site-packages/nest/ll_api.py", line 228, in stack_checker_func
    return f(*args, **kwargs)
  File "/home/bouhadjar/Documents/repo/nest-sim-doc/build/fix_1934_connect_arrays/lib/python3.8/site-packages/nest/lib/hl_api_connections.py", line 255, in Connect
    connect_arrays(pre, post, weights, delays, synapse_model, syn_param_keys, syn_param_values)
  File "pynestkernel.pyx", line 360, in pynestkernel.NESTEngine.connect_arrays
nest.lib.hl_api_exceptions.BadParameter: BadParameter in SLI function connect_arrays

stinebuu:
@YounesBouhadjar Thanks so much for testing, I have updated the PR, could you check again?

YounesBouhadjar:
@stinebuu I just tested the PR and it is working properly, thanks a lot.

nestkernel/nest.cpp
```diff
@@ -273,26 +295,33 @@ connect_arrays( long* sources,
...
+          auto* param = param_pointer_pair.second + index_counter;
+
           // Receptor type must be an integer.
           if ( param_pointer_pair.first == names::receptor_type )
           {
-            const auto int_cast_rtype = static_cast< size_t >( *param_pointer_pair.second );
-            if ( int_cast_rtype != *param_pointer_pair.second )
+            const auto int_cast_rtype = static_cast< size_t >( *param );
+
+            if ( std::abs( *param - int_cast_rtype ) > 1e-20 ) // To avoid rounding errors
```
clinssen, line 306: The IEEE 754 float64 ("double precision") format can represent integers from −2^53 to 2^53 (−9007199254740992 to 9007199254740992) exactly. Above those values, rounding to nearest multiples of 2, 4, etc. occurs. The integer type is here defined as ``size_t`` but we are also adding an ``int`. Do you know if the result (i.e. ``*param``) is always 2^32-1 or less? In that case, this test can be made exact. If the values of ``*param`` can be higher, this test will trigger even though the user supplied integer numbers (because of the loss of precision).

heplesser: Integer dictionary entries are `long`, so I would cast to `long` here instead of `size_t`. For 64-bit `long` this still does not solve the problem @clinssen referred to. In practice, receptor numbers will always be much smaller than 2^32. To be totally safe, we could have a double test: throw the exception if the value is either `> 1L>>31 or std::abs(...) > 0`. We can compare directly to 0 here instead of 1e-20, since for proper integers we will have exact zero here. And maybe change the variable name to `rtype_as_long`.

stinebuu: @clinssen @heplesser Just so I understand, I should change to
```
const auto rtype_as_long = static_cast< long >( *param );

if ( std::abs( *param - rtype_as_long ) > 1L>>31 or std::abs( *param - rtype_as_long ) > 0 )
```

heplesser:
```
if ( *param > 1L << 31 or std::abs( *param - rtype_as_long ) > 0 )
```

stinebuu: I have updated the `if`, please have another look.

heplesser (approved):
@stinebuu This looks fine to me now. Could you create a follow-up issue for a refactoring of connect_arrays()? nest.cpp should only contain small wrapper functions passing work on to the various managers and connect_arrays() with its 150 lines is far too large.

clinssen (approved):
Cheers!

stinebuu:
Thanks for the approvals!

@heplesser I have created the follow-up issue #1960.

PR #1943, opened 2021-02-25
fix citation for NEST 2.20.1 and add citation for NEST 2.20.2

clinssen:
Fix one citation and add another.

This is already shown correctly in the 2.20.1 docs (https://nest-simulator.readthedocs.io/en/nest-2.20.1/citing-nest.html) but not yet on the 3.0 docs (https://nest-simulator.readthedocs.io/en/latest/citing-nest.html); this PR simply updates the latter to the former.

@jougs, @terhorstd: could you include this step in the release pipeline?

I assume we will not update ``citing-nest.rst`` in older release versions of the documentation (for example, the 2.20.0 docs will never be updated to include citations for later—2.20.1 and up—releases).

### DIFF
--- a/doc/citing-nest.rst
+++ b/doc/citing-nest.rst
@@ -11,7 +11,9 @@ For the full citation and additional formats follow the doi Zenodo link.
 
 Cite the version you used in your work:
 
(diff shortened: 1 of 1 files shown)


jougs (approved):
Great! I made a note for the release tools and am merging directly due to smallness.

Issue #1948, opened 2021-02-26
Add CMake flag for explicit Python installation


Helveg:
**Is your feature request related to a problem? Please describe.**
The current CMake approach fails to deliver a functional Python package correctly isolated in the targetted Python environment.

(TL:DR; there's a summary of the problem halfway through and a bullet list of the solution at the bottom)

The best approach Python users - that follow Python best practices and use venvs - with CMake configuration needs have right now is to set `CMAKE_INSTALL_PREFIX` to the root of their virtual env, to set `WITH_PYTHON=ON`, set `PYTHON_EXECUTABLE` and in the worst case (like on Travis CI or when using `pyenv`) also set `PYTHON_INCLUDE_DIR` and `PYTHON_LIBRARIES`. An error prone process that often gets one of my collaborators stuck for hours figuring out why the PythonLibs aren't picked up 😛 (it's a typo, it's always a typo)

Using this most optimal configuration we end up with nest installed under our virtual env and a `nest` folder snuck into the `site-packages`, pretending to be a Python package, but actually falling short. It lacks the required metadata (see [PEP314](https://www.python.org/dev/peps/pep-0314/), [PEP345](https://www.python.org/dev/peps/pep-0345/), [PEP566](https://www.python.org/dev/peps/pep-0566/)) and is just a folder with some functional Python code in. However, since that `site-packages` folder is on the Python interpreter's path it will be picked up as an **importable module**, but since it has no package metadata none of Python's stdlib packaging modules such as `distutils`, `setuptools` or `pkg_resources` will pick up on its existence. 

As a downstream consequence the `nest` folder cuckoo is entirely invisible to all Python's packaging tools like `pip`. This is a bad user experience for Python folk. Since nest isn't pip installable yet that doesn't lead to too many problems but it still makes handling the ninja folder difficult. It also complicates building Python tools for nest, for example I can't reliably check if the user of my Python package has nest installed: 
 * Did they just create a folder called nest in `site-packages`?
 * Did they maybe have https://pypi.org/project/nest/ installed?
 * Is their `PYTHONPATH` set incorrectly and will I use the nest module and install of another Python?

This `site-packages` folder is not that well known to all Python users either, and for optimal results it needs to be manually specified **indirectly** as `CMAKE_INSTALL_PREFIX` of the 2nd .. or 3rd parent directory?

I'll summarize the core issues again:

[...]

PR #1954, opened 2021-03-01
Hike and enforce Python version requirement to 3.8

### LINKED ISSUE #1953: Hike Python requirements to Python 3.8
The Open Developer VC on 1 March 2021 decided to raise the minimum Python requirement for NEST to Python 3.8, released Nov 2019.

In solving this issue, all documentation should be checked for mentions of Python version numbers.

I suggest to close #1932 and create a new PR for this issue.

hakonsbm:
The minimum requirement for Python version is raised to 3.8, and enforced at CMake configuration. Additionally, for Cython, a version of 0.28.3 or later is enforced because of a change in the Python C api with Python 3.7.

Finding of Python is refactored from using the deprecated CMake functions `FindPythonInterp` and `FindPythonLibs` to use `FindPython`.

All documentation mentioning minor Python version numbers is updated.

This PR supersedes #1932 and fixes #1953.


heplesser:
@hakonsbm Works nicely on my Apple Silicon Mac.

hakonsbm:
On Travis CMake seems to not be able to find the Python library and include directories. This can be because of some mismatched paths when setting `-DPYTHON_LIBRARY` and `-DPYTHON_INCLUDE_DIR`, or possibly because it's using a slightly older CMake version (3.12.4). Should I try to fix it, or should we ignore Travis here because the GitHub Actions jobs are passing?

heplesser:
@lekshmideepu @jougs @terhorstd What do you think about the Python problems on Travis? I would tend to ignore them, given that the Python setup on Travis has always been fickle. If one of you sees a straightforward fix for Travis, that would be good, obviously. #1948 also looks relevant here.

Helveg:
Travis is not the only env where the Python libs and include dirs aren't found. Also the 15% of Python users that use `pyenv` have to manually look for their libs and include dirs.

hakonsbm:
It works fine for me with `pyenv`, using CMake 3.16.3. What CMake version and CMake command do you use, @Helveg?

Helveg:
Oh sorry, I got things mixed up, using the new `FindPython3` CMake module works

Helveg:
@hakonsbm CMake version 3.16.3 WSL2 Ubuntu 20, using pyenv Python 3.8.6:

```
cmake ..
CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find Python (missing: Python_EXECUTABLE Python_INCLUDE_DIRS
  Interpreter Development) (found suitable version "3.8", minimum required is
  "3.8")
  /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.16/Modules/FindPython/Support.cmake:2214 (find_package_handle_standard_args)
  /usr/share/cmake-3.16/Modules/FindPython.cmake:304 (include)
  cmake/ProcessOptions.cmake:401 (find_package)
  CMakeLists.txt:145 (nest_process_with_python)
```

Any idea?

hakonsbm:
Thanks for testing again, @Helveg. I've updated FindCython.cmake and UseCython.cmake files, but I still can't reproduce your errors with `pyenv`, neither with Python 3.9.1 or 3.8.6. It's also strange that they are treated differently by CMake for you.

As you say, `FindPythonLibs` is needed to get the `python_add_module` function. But the variables set by it is not used, so I have removed the `REQUIRED` argument. `FindPythonLibs` is deprecated, however, and I can't find an alternative to `python_add_module`. Maybe we should just copy the 30 lines of the `python_add_module` function from the source so we can remove `FindPythonLibs` completely.

Helveg:
Thanks for going over the cython files; I think vendoring the `python_add_module` code is a good solution. If you're feeling generous and open-sourcey, you could also open a PR on the cmake repo to include it in `FindPython` :)

gtrensch (approved):
Thank you for this contribution. 👍 I have successfully tested it on Ubuntu.

clinssen (approved):
I tested this on Debian 10 with CMake 3.20.0 and Python 3.9.

PR #1956, opened 2021-03-01
Reduce amount of data exchanged in test to ensure stable operation.

### LINKED ISSUE #1944: Stabilize test_multiple_synapses_spatial_networks.sli
`mpitests/test_multiple_synapses_spatial_networks.sli` fails from time to time during CI testing, both on Travis and Github. Since this only occurs from time to time, I assume that the problem is due to the way our test setup collects data from multiple processes. This process is somewhat fragile if the amount of data communicated becomes too large. 

Could you review the test and see if the amount of data exchanged could be reduced?

heplesser:
This should fix #1944. It limits each individual value exchanged between individual MPI processes and the monitoring process to three symbols and thus avoids overflooding of the communication mechanism.

### DIFF
--- a/testsuite/mpitests/test_multiple_synapses_spatial_networks.sli
+++ b/testsuite/mpitests/test_multiple_synapses_spatial_networks.sli
@@ -61,7 +61,7 @@ def
 % Set number of sources and targets according to number of positions
 /num_src pos length def
 /num_trgt pos length def
-/num_conns 4 def % we use one_to_one
+/num_conns 4 def % connections per target
 
 % to reduce data to be transported, reduce every value to max three chars
 /:compact_form
--- a/testsuite/mpitests/test_multiple_synapses_spatial_networks.sli
+++ b/testsuite/mpitests/test_multiple_synapses_spatial_networks.sli
@@ -38,20 +38,37 @@
 (unittest) run
 /unittest using
 
-
 /collocated_synapse_params
   [<< /weight -3. /delay 1.0 /synapse_model /static_synapse >>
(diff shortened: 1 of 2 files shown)


testsuite/mpitests/test_multiple_synapses_spatial_networks.sli
```diff
@@ -38,21 +38,38 @@
...
+  1 Flatten
+def
 
-% There are 9 positions in positions vector, so 9 neurons per source and target
-/num_src 9 def
-/num_trgt 9 def
+% Set number of sources and targets according to number of positions
+/num_src pos length def
+/num_trgt pos length def
 /num_conns 4 def % we use one_to_one
```
stinebuu, line 64: Could you also remove the `% we use one_to_one` comment, it is a copy-paste leftover from `test_multiple_synapses.sli`. And maybe update the comment to state that it is number of connections per target.

stinebuu (commented):
Thanks! There is also a `test_multiple_synapses.sli` test, that basically does the same, but does not have spatial connections. Maybe we should decrease the data here as well? There are fewer connections in that test, so that is why it usually pass I think, but just in case something happens in the future?

testsuite/mpitests/test_multiple_synapses_spatial_networks.sli
heplesser: Done.

heplesser:
@stinebuu I also updated `test_multiple_synapses.sli` as you suggested.

stinebuu (approved):
Thanks!

hakonsbm (approved):
Looks good to me!

PR #1966, opened 2021-03-06
Fixed a number of missing initialisations detected by Eclipse.

heplesser:
This PR fixes a number of missing initialisations that Eclipse alerted me to (in combination with clang). In a few cases, comments are inserted marking the warning as not relevant.


models/aeif_psc_delta.cpp
```diff
@@ -271,7 +273,9 @@ nest::aeif_psc_delta::State_::set( const DictionaryDatum& d, const Parameters_&,
   updateValueParam< double >( d, names::w, y_[ W ], node );
 }
 
-nest::aeif_psc_delta::Buffers_::Buffers_( aeif_psc_delta& n )
+
+nest::aeif_psc_delta::Buffers_::Buffers_( // @suppress("Class members should be properly initialized")
```
clinssen, line 277: Is this ``// @suppress`` Eclipse syntax? Is it desirable to have these comments in upstream?

This also raises the question why the directive is here (what are we suppressing and why)?

heplesser: Fixed, I removed this again, it is too Eclipse-specific. The warning is about missing initialisation, but that is handled in `init_buffers()`, and this again explained in a constructor comment.

nestkernel/sparse_node_array.h
```diff
@@ -60,7 +60,11 @@ class SparseNodeArray
 public:
   struct NodeEntry
   {
-    NodeEntry() = default;
+    NodeEntry()
+      : node_( nullptr )
+      , node_id_( 0 )
+    {
+    }
```
hakonsbm, line 67: Same here, the defaulted constructor will zero-initialize the variables.

hakonsbm (changes requested):

clinssen (approved):

libnestutil/block_vector.h
```diff
@@ -79,7 +79,11 @@ class bv_iterator
   using reference = ref_;
   using difference_type = typename BlockVector< value_type >::difference_type;
 
-  bv_iterator() = default;
+  bv_iterator()
+    : block_vector_( nullptr )
+    , block_index_( 0 )
+  {
+  }
```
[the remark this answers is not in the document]
heplesser: But based on the Python mantra of "explicit is better than implicit" and to make Eclipse happy it does not hurt, does it? So I would like to keep it as is.

nestkernel/sparse_node_array.h
heplesser: See above.

libnestutil/block_vector.h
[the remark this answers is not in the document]
hakonsbm: Fair enough.

hakonsbm (approved):

PR #1972, opened 2021-03-08
Make file, C++ class and model names consistent for synapses

### LINKED ISSUE #1930: Unify nomenclature of synapse names
Neuron models have a consistent naming: the class name (for example, `gif_cond_exp`) is the same as the file name (`gif_cond_exp.{h,cpp}"`) and the name under which it is registered in the `ModelsModule` (`gif_cond_exp`).

For synapses, this is, however, not the case. For example, the class might be called `STDPConnection`, the filename is `stdp_connection.{h,cpp}` and the user-facing name it is given in `ModelsModule` is `stdp_synapse`.

The suggestion is to rename the classes and file names to the name under which the model is registered. Mostly, this will end in "_synapse". Only the "rate" and "diffusion" connections will remain with suffix "_connection". Note that this means that class names will not be in camel case (as is currently already the case for neuron/node models).

As a secondary issue, the keyword `synapses` in the call `nest.Models(mtype, ...)` should be changed to `connections`.

Pertaining documentation and in particular the 2🠒3 transition guide (see also #1929) has to be adapted accordingly.

clinssen:
Fixes #1930, with the feature-creep countermeasure device set to 11.

The terms "connection" and "synapse" are used somewhat interchangeably. It would be great to continue discussion on making this more consistent across the code and docs.


models/modelsmodule.cpp
```diff
@@ -307,30 +307,30 @@ ModelsModule::init( SLIInterpreter* )
 #endif
 
   // register all connection models
-  register_connection_model< BernoulliConnection >( "bernoulli_synapse" );
-  register_connection_model< ClopathConnection >(
+  register_connection_model< bernoulli_synapse >( "bernoulli_synapse" );
```
stinebuu, line 310: Should the comment be `// register all synapse models`?

models/modelsmodule.cpp
```diff
@@ -115,37 +115,37 @@
...
-#include "jonke_connection.h"
-#include "quantal_stp_connection.h"
-#include "quantal_stp_connection_impl.h"
+#include "ht_synapse.h"
+#include "jonke_synapse.h"
+#include "quantal_stp_synapse.h"
+#include "quantal_stp_synapse_impl.h"
 #include "rate_connection_delayed.h"
 #include "rate_connection_instantaneous.h"
 #include "spike_dilutor.h"
```
stinebuu, line 131: Shouldn't the `spike_dilutor.h` be imported under `// Stimulation devices` or `// Neuron models`?

models/stdp_dopamine_synapse.h
```diff
@@ -20,8 +20,8 @@
  *
  */
 
-#ifndef STDP_DOPA_CONNECTION_H
-#define STDP_DOPA_CONNECTION_H
+#ifndef STDP_DOPA_SYNAPSE_H
```
stinebuu, line 23: `STDP_DOPAMINE_SYNAPSE_H` for consistency.

pnbabu (commented):
Looks fine to me. Just a minor comment.

models/stdp_synapse_facetshw_hom_impl.h
```diff
@@ -123,7 +123,7 @@ void
 STDPFACETSHWHomCommonProperties< targetidentifierT >::calc_readout_cycle_duration_()
 {
   readout_cycle_duration_ = int( ( no_synapses_ - 1.0 ) / synapses_per_driver_ + 1.0 ) * driver_readout_time_;
-  // std::cout << "stdp_connection_facetshw_hom::debug: readout cycle duration
+  // std::cout << "stdp_synapse_facetshw_hom::debug: readout cycle duration
```
stinebuu, line 126: I think we can just remove this debugging comment. People can just add it again if they need it.

pynest/nest/tests/test_stdp_nn_synapses.py
```diff
@@ -195,7 +195,7 @@ def reproduce_weight_drift(self, _pre_spikes, _post_spikes,
                     # with t_previous_pre, thus due to the restricted
                     # pairing scheme we do not account it.
                 if (self.synapse_parameters["synapse_model"] == "stdp_nn_symm_synapse" or
-                        self.synapse_parameters["synapse_model"] == "stdp_nn_pre-centered_synapse"):
+                        self.synapse_parameters["synapse_model"] == "stdp_nn_pre_centered_synapse"):
```
stinebuu, line 198: Do you think we need to document this change in synapse name in the NEST 2x -> NEST 3 documentation? It is very minor, so I do not know if it is necessary.

pynest/nest/tests/test_stdp_triplet_synapse.py
```diff
@@ -29,7 +29,7 @@
 
 @nest.ll_api.check_stack
 class STDPTripletConnectionTestCase(unittest.TestCase):
```
stinebuu, line 31: I think the class name should change to `STDPTripletSynapseTestCase`

stinebuu (changes requested):
Thanks! It looks good, I just have some smaller comments.

stinebuu:
> The terms "connection" and "synapse" are used somewhat interchangeably. It would be great to continue discussion on making this more consistent across the code and docs.

@clinssen should we add this as a point to the next hackathon?

models/modelsmodule.cpp
clinssen: Added under devices, as the spike_dilutor class inherits from DeviceNode.

pynest/nest/tests/test_stdp_nn_synapses.py
clinssen: Yes, that would be great to mention, cheers.

clinssen:
> > The terms "connection" and "synapse" are used somewhat interchangeably. It would be great to continue discussion on making this more consistent across the code and docs.
> 
> @clinssen should we add this as a point to the next hackathon?

That would be great! Does the hackathon already have a wiki page?

stinebuu:
@clinssen Yes! You can add it [here](https://github.com/nest/nest-simulator/wiki/NEST-Hackathon-19.04-23.04-2021). Feel free to add it to the plan :smile:

stinebuu (approved):
Thanks, looks good to me!

pnbabu (approved):

Issue #1961, opened 2021-03-05
Wrong path and link in user_documentation_workflow


stinebuu:
**Describe the flaw**
In the user_documentation_workflow on ReadTheDocs, https://nest-simulator.readthedocs.io/en/latest/documentation_workflow/user_documentation_workflow.html, it says that when you install documentation the rendered text can be found in `<nest_install_dir>/share/doc/nest/html/`

![image](https://user-images.githubusercontent.com/21308850/110089948-06a5e080-7d97-11eb-9016-5b3fbe15205a.png)

This folder is empty for me for newest master (for a clean build, and after doing `make html` and `make install`). I think the documentation can now be found in `<nest_install_dir>/doc/userdoc/html/`? So the path should be updated.

Also, there seems to be a broken link to an image further up on the page:
![image](https://user-images.githubusercontent.com/21308850/110090759-ede9fa80-7d97-11eb-8387-b855ec2349ea.png)

jessica-mitchell:
@stinebuu I think the paths will be correct as of  #1926. When I last built nest from that branch, the install directory path for docs was <install>/share/doc/nest/html/ and in the build directory the path was <build>/doc/userdoc/html/. 
The link though does need to be fixed, thanks!

jougs:
The broken link is fixed by #1926 (commit d7e20064e in particular).

PR #1979, opened 2021-03-15
Fix broken images in documentation

sarakonradi:
This PR fixes the broken images in the documentation by updating ``_static`` to ``static`` in the corresponding paths. It is a follow-up to #1843.

### DIFF
--- a/doc/userdoc/documentation_workflow/user_documentation_workflow.rst
+++ b/doc/userdoc/documentation_workflow/user_documentation_workflow.rst
@@ -21,7 +21,7 @@ with each release of NEST having its own documentation.
 
 This workflow aims for the concept of **user-correctable documentation**.
 
-.. image:: ../_static/img/documentation_workflow.png
+.. image:: ../static/img/documentation_workflow.png
   :width: 500
   :alt: Alternative text
 
--- a/doc/userdoc/getting_started.rst
+++ b/doc/userdoc/getting_started.rst
@@ -71,7 +71,7 @@ Display the voltage graph from the voltmeter:
 
 You should see the following image as the output:
 
-.. image:: _static/img/output_getting_started.png
+.. image:: static/img/output_getting_started.png
    :align: center
 
 **And that's it! You have performed your first neuronal simulation in NEST!**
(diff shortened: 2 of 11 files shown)


jessica-mitchell (approved):
Looks good! thanks @sarakonradi

jougs:
Interesting and confusing.

I made exactly the same change yesterday and found that it _breaks_ rather than fixes all images because Sphinx seems to move all static files to a directory called `_static` in its `html` output directory - regardless of what it is called in the source directoy (where it is indeed called `static` since #1843). I suspect this is happening because of setting [`html_static_path` in `conf.py`](https://github.com/nest/nest-simulator/blob/master/doc/userdoc/conf.py#L165). See [this image's URL](https://nest-simulator.readthedocs.io/en/latest/_static/img/oscilloscope.png) for an example.

According to my analysis, the one broken image reported by @stinebuu in #1961 was caused by a `../` missing in the image path.

Can you please reconfirm that this PR is really needed?

jessica-mitchell:
@jougs When I built the documentation with the changes applied here the images worked - I did notice that the images in the homepage (userdoc/index.rst) that are embeded html still had the _static path (and still worked)-and found what you mentioned this _static directory in html/

jougs:
Hmmm... I suggest to check again with a fresh clone, a clean installation directory and a wiped browser cache. On my installation, I can't see a directory `static` under `html` (only `_static` is there), so I don't see how the proposed changes could work. If images are taken from `doc/userdoc/static` directly (instead of from `html/_static`, that is), that would most likely cause problems on Read the Docs (which copies the `html` directory to some public web space after running `sphinx_build`) and locally (if the source and build directories are deleted at one point).

jessica-mitchell:
@jougs I went got a fresh clone of the branch, had a clean build environment and I still get the same result. What I can see is there are two paths in the build directory for 'static' as I think you also have:  `/doc/userdoc/static` and `doc/userdoc/html/_static`
But if I look at the image location after building it's in` build/doc/userdoc/html/_images/`

I dont know what's happening but shouldn't the path in the reStructuredText file point to the source directory image path at least?

jougs:
@jessica-mitchell: thanks for trying!

I guess this means that image paths should indeed be specified as per the source directory layout and Sphinx takes care of mangling them so everything works.

The situation with `_static` in the image paths then probably just works by accident: Sphinx does not recognize, and, consequently, not mangle those, but we end up with working images nonetheless, because we already anticipated the mangling ourselves.

jougs (approved):
Given the recent analysis of the situation, I am happy with this. Thanks!

PR #1980, opened 2021-03-16
Adapt warning parser to output format on GitHub Actions

### LINKED ISSUE #1967: Discrepancy between Github and Travis test results / Warnings count during compilation
On Travis, this minimal build failed

https://travis-ci.org/github/nest/nest-simulator/jobs/761653390

while it passed on Github

https://github.com/nest/nest-simulator/pull/1966/checks?check_run_id=2044070265

Also the other tests (MPI_ONLY, OPENMP_ONLY, FULL) fail on Travis but pass on Github.

The reason for the failure on Travis is that the warnings count for `sli/scanner.cc` is 3 instead of the expected 0. The compiler emits warnings on Travis and Github, but for some reason they are not registered properly by our result handler.

jougs:
This fixes #1967. All information about the problem is nicely described there.

### DIFF
--- a/extras/parse_travis_log.py
+++ b/extras/parse_travis_log.py
@@ -393,7 +393,7 @@ def makebuild_summary(log_filename, msg_make_section_start,
                 # Only count warnings originating in NEST source files
                 warning_match = nest_warning_re.match(line)
                 if warning_match is not None:
-                    warning = warning_match.group(0)
+                    warning = warning_match.group(1)
                     if warning not in known_warnings:
                         file_name = warning.split(':')[0]
                         if file_name not in warning_summary:
--- a/extras/parse_travis_log.py
+++ b/extras/parse_travis_log.py
@@ -367,7 +367,7 @@ def makebuild_summary(log_filename, msg_make_section_start,
     # with some warnings, this would be a good point to re-set the
     # expected_warnings variable conditionally for that build_type.
 
-    nest_warning_re = re.compile(f'{build_dir}.*: warning:')
+    nest_warning_re = re.compile(f'.* ({build_dir}.*: warning:.*)')
     known_warnings = [
         f'{build_dir}/sli/scanner.cc:642:13: warning: this statement may fall through [-Wimplicit-fallthrough=]',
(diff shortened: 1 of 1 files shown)


extras/parse_travis_log.py
```diff
@@ -392,12 +392,14 @@ def makebuild_summary(log_filename, msg_make_section_start,
...
                 # Only count warnings originating in NEST source files
                 warning_match = nest_warning_re.match(line)
-                if warning_match is not None and line.strip() not in known_warnings:
-                    file_name = line.split(':')[0]
-                    if file_name not in warning_summary:
-                        warning_summary[file_name] = 0
-                    warning_summary[file_name] += 1
-                    number_of_warning_msgs += 1
+                if warning_match is not None:
+                    warning = warning_match.group(0)
```
heplesser, line 396: `group(0)` is the full match.
```suggestion
                    warning = warning_match.group(1)
```

extras/parse_travis_log.py
```diff
@@ -367,7 +367,7 @@ def makebuild_summary(log_filename, msg_make_section_start,
     # with some warnings, this would be a good point to re-set the
     # expected_warnings variable conditionally for that build_type.
 
-    nest_warning_re = re.compile(f'{build_dir}.*: warning:')
+    nest_warning_re = re.compile(f'.* ({build_dir}.*: warning:.*)')
     known_warnings = [
```
heplesser, line 371: Wondering if this should be a set instead of a list. For a 4-element collection it probably does not make a difference and list may even be faster since it is simpler—and we don't want this list to become long in any case. So maybe leave as is.

heplesser (changes requested):
@jougs Thanks, but I think it needs one little fix.

extras/parse_travis_log.py
jougs: Thanks for sharing your thoughts. I'll leave it as it is then ;-)

heplesser (approved):

gtrensch (approved):
Looks good to me.

PR #1949, opened 2021-02-26
Provide links to model_details in relevant model documentation

jessica-mitchell:
This PR resolves #1568  
Links to the notebooks were provided to the matching models. However I was not sure about test_post_trace notebook. @heplesser @clinssen Do you have an idea where this notebook can be linked from?
All aeif models are linked to the aeif_models_implementation notebook
For the IAF_neurons_singularity notebook, it states that it's for "integrate-and-fire model neurons with alpha- or exponentially-shaped current, " so I added a link to all models of iaf_psc_alpha* and iaf_psc_exp*. Let me know if that's correct.
Additionally a few fixes to the format were made.

PR #1981, opened 2021-03-17
Remove hill_tononi_Vp example

### LINKED ISSUE #982: outdated example `/topology/examples/hill_tononi_Vp.py`
Some of the description(s) in this example file is out of data. Accd. to a note it's supposed to produce 5Hz oscillations, but connecting a spike recorder to the system reveals fast oscillations:
image

Now that the Hill&Tononi neuron model (`ht_neuron`) is built into NEST, I think the script should probably use that model as well, instead of the presently used `iaf_cond_alpha` neuron

On a related note, @andresni is working on reproducing the model from the paper in detail and has set up model generation from YAML in NEST at https://github.com/andresni/NYP, but the corresponding paper(s) have not been published. Maybe in some future version of NEST his model setup could be integrated closer...

heplesser:
This PR removes the example and adds instead links to the scientifically validated implementation of the full Hill-Tononi model by Ricardo Murphy and colleagues.

This fixes #982.


heplesser:
@jessica-mitchell I have a technical problem here: https://github.com/heplesser/nest-simulator/blob/001e65a623cd0a5341c21b41ac8c1a85e2ff321e/models/ht_neuron.h#L98-L99 .

With the line break as it is now, our clang-format is happy, but Sphinx does not render the link properly. If I put everything on one line, Sphinx renders the link properly, but clang-format complains. Any ideas how to solve this?

jessica-mitchell:
@heplesser When you have a bullet list, the second line has to line up with the first - so it looks like this
```
* `very long text for link
  <link>`_
```

heplesser:
> @heplesser When you have a bullet list, the second line has to line up with the first - so it looks like this
> 
> ```
> * `very long text for link
>   <link>`_
> ```


Thanks for the hint, fixed!

doc/model_details/HillTononiModels.ipynb
```diff
@@ -14,7 +14,9 @@
     "\n",
     "This description is based on the original publication and publications cited therein, an analysis of the source code of the original Synthesis implementation kindly provided by Sean Hill, and plausiblity arguments.\n",
     "\n",
-    "In what follows, I will refer to the original paper as [HT05]."
+    "In what follows, I will refer to the original paper as [HT05].\n",
```
jessica-mitchell, line 17:
```suggestion
    "In what follows, we will refer to the original paper as [HT05].\n",
```

jessica-mitchell (changes requested):
@heplesser Thanks for updating this. I noticed the links to the jupyter notebooks in model_details are broken, which is relevant to this PR. It would be great if that could be fixed here - two things have to change:  re-add extension 'nbsphinx' to conf.py (somewhere along the way since #1949  'nbsphinx' it was removed)  and move the model_details directory to userdoc/.

heplesser:
@jessica-mitchell Done! I noticed that in earlier versions of `conf.py`, in addition to `nbsphinx` there was also `breathe`, which is no longer there. Is it no longer needed?

heplesser:
@jessica-mitchell I struggle with ensuring proper installation of the model_details directory after moving it to userdocs. Could you give me a pointer to a version where model_details was in userdocs and got installed? I just pushed an update that solves some problems but **does not install the notebooks**.

jessica-mitchell:
@heplesser I'm not sure I fully understand what you mean with not installing the notebooks? From your latest commit, I can open the notebooks in the documentation so they're visible on the browser, and I can open them from the console in jupyter-notebook (from `install/share/doc/nest/html/model_details/`).  I moved `model_details `because it seemed more logical to put it into `userdoc/` and the relative path link in the models would be correct. But it could very well be that I broke something - I am still trying to understand all the consequences of this new build workflow for documentation. 

as for breathe - I'm pretty sure we are not using it, that  was replaced by Dennis' extractor script for models.

heplesser:
@jessica-mitchell Sorry, my mistake. I had forgotten to run `make html` before `make install`. All fine now :).

jessica-mitchell (approved):
I think this looks good, thanks @heplesser !

heplesser:
Documentation change, therefore merging after single review.

PR #1983, opened 2021-03-18
Correctly connect MUSIC output proxies when using multiple MPI processes in NEST

### LINKED ISSUE #1974: Too many connections to MUSIC output proxies created when NEST runs MPI-parallel
The following code should connect each parrot neuron exactly once to the MUSIC output proxy:
```
/nrns /parrot_neuron 2 Create def
/out /music_event_out_proxy << /port_name (out_spikes) >> Create def
nrns out Connect
```
Running the code on a single process, this works fine, and the following code 
```
Rank 0 eq { (                         Src Tgt Prt Rcpt) = } if
(Rank: ) Rank cvs join ( --> Connections: ) join
<< >> GetConnections GetStatus { [[ /source /target /port /receptor ]] get } Map pcvs join =
```
prints this output:
```
                         Src Tgt Prt Rcpt
Rank: 0 --> Connections: [[1 3 0 0] [2 3 0 0]]
```
This is the same output we get if `out` were a spike recorder.

Running on two MPI processes, one would expect that each rank has one connection, from the neuron it holds to the local device. But we get **two** connections on each rank:
```
                         Src Tgt Prt Rcpt
Rank: 0 --> Connections: [[2 3 0 0] [2 3 1 0]]
Rank: 1 --> Connections: [[1 3 0 0] [1 3 1 0]]
```
If `out` is a spike recorder, we get one connection per rank as expected.

The consequence of this is that spikes are sent "double up". The example here leaves out the setting of `music_channel` to keep the reproducer minimal; it only affects the value of `rport`.

This bug must have existed for a while and most likely affects all MUSIC out proxies; I am not sure about MUSIC in proxies.

heplesser:
This fixes #1974.

This is still a draft because I still need to add regression tests.


jougs:
@heplesser: Is there any progress on this? Would it be an option to close and re-open once you get around to it?

jougs:
Friendly ping!

heplesser:
@jougs @mdjurfeldt I finally added the regression test (much easier than I had feared ...), so this one is ready for review. Several files in the diff (esp in models) only pop up here due to formatting corrections.

jougs (approved):
I only have a small spelling fix, but am fine otherwise. @mdjurfeldt could you please quickly review and indicate if you're fine with this? Thanks!

mdjurfeldt (approved):
This looks fine to me.

PR #1999, opened 2021-04-08
Add missing include file to CMakeLists

clinssen:
This fixes an issue that we were running into with a NEST extension module not being able to ``#include "dict_util.h"``.

There are several other header files in the ``libnestutil`` directory; should these be added to the list as well for consistency? Or would this pollute the namespace too much?

### DIFF
--- a/libnestutil/CMakeLists.txt
+++ b/libnestutil/CMakeLists.txt
@@ -29,8 +29,10 @@ set( nestutil_sources
     logging.h
     numerics.h numerics.cpp
     propagator_stability.h propagator_stability.cpp
+    regula_falsi.h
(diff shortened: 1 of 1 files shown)


heplesser (approved):
Looks good to me.

hakonsbm:
I see only two additional header files that are not included: `regula_falsi.h` and `streamers.h`. I think it would make sense to include them as well.

clinssen:
> I see only two additional header files that are not included: `regula_falsi.h` and `streamers.h`. I think it would make sense to include them as well.

Done!

hakonsbm (approved):
Thanks!

PR #2012, opened 2021-04-13
Parse build output of GitHub Actions

### LINKED ISSUE #2005: Compilation warnings not detected by GitHub Actions
The output from GitHub Actions CI is never parsed, so a CI job can only fail if any of the commands return a non-zero exit code (e.g. the build fails or the testsuite fails). 

The `parse_travis_log.py` script checks both the output of static checks, and for warnings during compilation. The static checks are currently still handled by Travis, but because the build log is not parsed by GitHub Actions, we are currently not detecting warnings during compilation.

The current GitHub Actions commands should be changed to store the compilation output, then parse it to detect warnings raised during compilation.

As a reference, compare the old Travis commands:
https://github.com/nest/nest-simulator/blob/63efc439338fcdf3b398700652b2ec8590de3a1d/.travis.yml#L197-L205

to the current GitHub Actions commands:
https://github.com/nest/nest-simulator/blob/6c0602df984a05455a867915ade5a47bcad47620/.github/workflows/nestbuildmatrix.yml#L101-L110

hakonsbm:
GitHub actions will now parse the output of the build, give a proper summary, and fail if there are new compiler warnings raised, like it used to with Travis. This fixes #2005.

Also, the `parse_travis_log.py` script is renamed to `parse_build_log.py`, and trailing whitespace is cleaned up in the modified files.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -13,7 +13,7 @@ jobs:
          os: ["ubuntu-20.04", macos-latest]
          cpp_compiler: ["gcc", "clang"]
          xNEST_BUILD_TYPE: ["MINIMAL", "MPI_ONLY", "OPENMP_ONLY", "FULL"]
-         exclude: 
+         exclude:
            - xNEST_BUILD_TYPE: "MPI_ONLY"
              cpp_compiler: "clang"
              os: macos-latest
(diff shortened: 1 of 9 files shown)


heplesser (commented):
@hakonsbm Thanks, this looks good. But I noticed that for the logs I checked, it shows "0 warnings (0 expected)" in all cases, including FULL. Don't we have some cases with a small number of warnings? Or are they so thoroughly handled by the parser that they do not count as expected?

I also noticed that in the summary at the end of build and tests it still says "Travis" in the header. Could you fix this here?

hakonsbm:
@heplesser 
> I noticed that for the logs I checked, it shows "0 warnings (0 expected)" in all cases, including FULL. Don't we have some cases with a small number of warnings? Or are they so thoroughly handled by the parser that they do not count as expected?

Most of that was fixed in #1852, and the rest in #1830. So, except for a few warnings from SLI that are handled by the parser, we are expecting zero warnings in all cases. I've removed it from the summary.

> I also noticed that in the summary at the end of build and tests it still says "Travis" in the header. Could you fix this here?

Fixed!

heplesser (approved):

gtrensch (approved):
Thanks for this contribution. Looks good to me :+1:.

PR #2014, opened 2021-04-19
Set `have_connections_changed` correctly when we have fewer connections than MPI ranks

### LINKED ISSUE #1957: GetConnection hangs if some MPI ranks have no (new) connections
If at least one MPI rank did not create a connection, a subsequent `GetConnections` call will hang. This happens because `ConnectionManager::have_connections_changed_` will not be set on MPI ranks not creating connections. On those ranks, `ConnectionManager::get_connections()` will not call `SimulationManager::update_connection_infrastructure()`, causing NEST to hang.

A minimal reproducer is given below. When run on two MPI processes, one will print `Got conns: []` and the other nothing. 

```
(Start) ==
/n /iaf_psc_alpha Create def
(Created) ==
n n Connect
(Connected) ==

<< >> GetConnections
(Got conns: ) =only ==
```

The simplest solution is probably to set the `have_connections_changed_` flag as soon as `Connect()` is called. A more difficult situation occurs when structural plasticity is used. Then, the only safe approach is most likely to assume that any call to `Simulate` will cause changes in connections. While this approach could cause false positives in rare cases, it would at the same time simplify our code since the change-flag could now be set at the rank level.

stinebuu:
This PR fixes #1957.

Always set `have_connections_changed_` when calling `Connect()`.
When we have structural plasticity we set `have_connections_changed` in the `update_structural_plasticity()` function, to make sure it is set when connections are changed.


nestkernel/sp_manager.cpp
```diff
@@ -417,7 +422,7 @@ SPManager::update_structural_plasticity( SPBuilder* sp_builder )
  * @param post_n number of available synaptic elements in the post node
  * @param sp_conn_builder structural plasticity connection builder to use
  */
-void
+bool
```
hakonsbm, line 425: Should probably add a `@return` entry after the `@param` entries above, giving a short description of the return value.

hakonsbm (changes requested):
Looks pretty good. In places where `set_have_connections_changed()` now only sets it for the main thread, it would be useful to have a comment explaining why we need to set it for only the main thread. Alternatively maybe refactor out the thread argument from `set_have_connections_changed()`.

I have a couple questions:
- Does `disconnect()` have this problem as well? In that case that should also be fixed.
- And what about `connect_arrays()`?

nestkernel/connection_manager.cpp
```diff
@@ -395,6 +395,8 @@ nest::ConnectionManager::connect( NodeCollectionPTR sources,
   ConnBuilder* cb = connbuilder_factories_.at( rule_id )->create( sources, targets, conn_spec, syn_specs );
   assert( cb != 0 );
 
+  set_have_connections_changed( kernel().vp_manager.get_thread_id() );
```
heplesser, line 398: This should move down to right before the `cb->connect()` call to make the code more understandable. Probably it would be good to add a comment like "Set flag before calling cb->connect() in case exception is thrown after some connections have been created."

nestkernel/sp_manager.cpp
```diff
@@ -449,6 +454,8 @@ SPManager::create_synapses( std::vector< index >& pre_id,
 
   // create synapse
   sp_conn_builder->sp_connect( pre_id_rnd, post_id_rnd );
+
+  return ( not pre_id_rnd.empty() );
```
heplesser, line 458: Do we need the parentheses here?

nestkernel/spatial.cpp
```diff
@@ -447,11 +447,7 @@ minus_mask( const MaskDatum& mask1, const MaskDatum& mask2 )
 void
 connect_layers( NodeCollectionPTR source_nc, NodeCollectionPTR target_nc, const DictionaryDatum& connection_dict )
 {
-  const thread num_threads = kernel().vp_manager.get_num_threads();
-  for ( thread tid = 0; tid < num_threads; ++tid )
-  {
-    kernel().connection_manager.set_have_connections_changed( tid );
-  }
+  kernel().connection_manager.set_have_connections_changed( kernel().vp_manager.get_thread_id() );
```
heplesser, line 450: I'd move this to right before `source->connect()` with a similar comment as suggested above.

heplesser (changes requested):
Looks good to me, just minor suggestions to make code more readable.

stinebuu:
@heplesser @hakonsbm Thanks for your comments. I have now addressed them, including setting the flag for `disconnect()` and `connect_arrays()` (good catch).

I have also added a test, but it won't fail if we revert some of the changes, it will just hang. Is it possible to actually make it fail? Do we have a test-mechanism for this, basically testing that a test does not hang?

testsuite/mpitests/issue-1957.sli
```diff
@@ -0,0 +1,51 @@
...
+  /nrns /iaf_psc_alpha Create def
+
+  nrns nrns /all_to_all Connect
+
+  << >> GetConnections { cva 2 Take }  Map /conns Set
+
+  nrns nrns << /rule /all_to_all >> << /synapse_model /static_synapse >> Disconnect_g_g_D_D
+  << >> GetConnections  { cva 2 Take }  Map conns join
+
+} distributed_process_invariant_collect_assert_or_die
```
hakonsbm, line 51: Missing newline at end of file.
```suggestion
} distributed_process_invariant_collect_assert_or_die

```

hakonsbm (changes requested):
Thanks! Tests that hang when failing are bad, but better than no tests I guess. I've used C++ tests for checking hanging, using `boost::unit_test::timeout()`. But I don't know if that works with MPI. For an example, see
https://github.com/nest/nest-simulator/blob/9968c63f2c19f9c9411cedf691accce54e2ee288/testsuite/cpptests/test_parameter.h#L47

nestkernel/connection_manager.cpp
```diff
@@ -609,6 +608,9 @@ nest::ConnectionManager::connect_arrays( long* sources,
     }
   };
 
+  // Set flag before entering parallel section in case we have less connections than ranks.
```
heplesser, line 611:
```suggestion
  // Set flag before entering parallel section in case we have fewer connections than ranks.
```

stinebuu:
I want to test this on the supercomputers to make absolute sure performance is not affected. I will let you know when I have results and you can review :)

stinebuu:
@heplesser @hakonsbm Appologies for taking so long. I have now tested on JUSUF and Piz Daint, and it seems like performance is unaffected by converting `have_connections_changed_` to a bool. Have added a couple of benchmarks below. You can therefore review again.

image
image

heplesser (commented):
@stinebuu Thanks for the benchmarks, this looks good! But I have one more request to further improve code readability: could you rename `have_connections_changed` to `connections_have_changed` and `has_get_connection_been_called` to `get_connection_has_been_called`? I would also remove the `set_has_get_connections_been_called` method since it is only used from inside the `ConnectionManager` class, so changing the corresponding field directly is no problem.

stinebuu:
@heplesser Done!

heplesser (approved):

heplesser:
@hakonsbm Ping :).

hakonsbm (approved):

PR #2018, opened 2021-04-21
Remove unused event_prototypes_ member

heplesser:
As the title says ...


jougs (approved):
Given that this change is rather small and the code still compiles and tests pass, I'll merge without second review. Thanks!

PR #2024, opened 2021-04-23
Automatic conversion of integers to doubles when setting parameters

### LINKED ISSUE #2009: Allow integer values where doubles are expected in NEST

NEST does not convert integers to doubles even in cases where this would be unproblematic. The only exception is the `Simulate` call. Furthermore, the error message emitted if users use integers where double is required is often difficult to comprehend because it does not indicate which entry in a dictionary causes the error, as in [this example from the NEST Mailing list](https://www.nest-simulator.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/BOELLXM5QHJF473FWVO2O3WIWGBLET5G/):

```
 File "test/binary_gdf.py", line 9, in <module>
    kk=tp.CreateLayer({'extent': [1, 1],'rows': 20, 'columns': 20,'elements' : 'izhikevich','edge_wrap':True})
  File "/home/neurobit/local/nest-py3/lib64/python3.6/site-packages/nest/topology/hl_api.py", line 650, in CreateLayer
    return topology_func('{ CreateLayer } Map', specs)
  File "/home/neurobit/local/nest-py3/lib64/python3.6/site-packages/nest/topology/ll_api.py", line 61, in topology_func
    return nest.ll_api.sli_func(slifunc, *args)
  File "/home/neurobit/local/nest-py3/lib64/python3.6/site-packages/nest/ll_api.py", line 182, in sli_func
    sli_run(slifun)      # SLI support code to execute s on args
  File "/home/neurobit/local/nest-py3/lib64/python3.6/site-packages/nest/ll_api.py", line 132, in catching_sli_run
    raise exceptionCls(commandname, message)
nest.lib.hl_api_exceptions.TypeMismatch: ('TypeMismatch in CreateLayer_D: Expected datatype: doubletype\nProvided datatype: integertype', 'TypeMismatch', <SLILiteral: CreateLayer_D>, ': Expected datatype: doubletype\nProvided datatype: integertype')
```

where the problem lies with the integer values for `extent`.

1. NEST should silently convert integers to doubles in keeping with rules for most common programming languages.
2. We should consider whether to accept integer-values doubles where integers are expected (e.g., receptor types)
3. When a type error message is emitted while reading a dictionary entry, the name of the entry should be included in the error message.

ackurth:
Fixes #2009.
The implementation silently converts integers to doubles (both for simple and vector data) where doubles are expected.
We removed the tests that became obsolete due to this new behavior.


heplesser:
@ackurth Could this PR be move from Draft to Open?

ackurth:
> @ackurth Could this PR be move from Draft to Open?

The PR as it is right now does address the issue of passing integers where doubles are expected in cases like ```{'V_m' : 1}```.
However, if a double-vector is expected - e.g. ```{'extent' : [-1, 1]}``` the conversion does not work yet. 
This especially means that the example from the issue is not addressed yet.
I reached out to @hakonsbm  to resolve this.

stinebuu:
@ackurth There seems to be some formatting errors, could you fix this?

sli/tokenarray.cc
```diff
@@ -127,13 +127,21 @@ TokenArray::toVector( std::vector< double >& a ) const
...
-    DoubleDatum* targetid = dynamic_cast< DoubleDatum* >( idx->datum() );
-    if ( targetid == NULL )
+    DoubleDatum* targetdd = dynamic_cast< DoubleDatum* >( idx->datum() );
+    if ( targetdd )
+    {
+      a.push_back( targetdd->get() );
+    }
+    else if ( IntegerDatum* targetid = dynamic_cast< IntegerDatum* >( idx->datum() ) )
+    {
+      a.push_back( ( double ) targetid->get() );
```
heplesser, line 137: Use the explicit cast syntax as recommended by Stroustrup:
```suggestion
      a.push_back( static_cast< double >( targetid->get() ) );
```

sli/tokenutils.cc
```diff
@@ -73,14 +73,23 @@ template <>
...
+  DoubleDatum* dd = dynamic_cast< DoubleDatum* >( t.datum() );
+  if ( dd )
+  {
+    return dd->get();
   }
-  return id->get();
+  IntegerDatum* id = dynamic_cast< IntegerDatum* >( t.datum() );
+  if ( id )
+  {
+    return ( double ) id->get();
```
heplesser, line 85:
```suggestion
    return static_cast< double >( id->get() );
```

sli/tokenutils.cc
```diff
@@ -97,13 +106,20 @@ template <>
 float
 getValue< float >( const Token& t )
```
heplesser, line 107: I don't think we use float anywhere in NEST, could you try to remove this function and see if anything breaks?

sli/tokenutils.cc
```diff
@@ -97,13 +106,20 @@ template <>
...
 {
-  DoubleDatum* id = dynamic_cast< DoubleDatum* >( t.datum() );
-  if ( id == NULL )
-  { // we have to create a Datum object to get the name...
-    DoubleDatum const d;
-    throw TypeMismatch( d.gettypename().toString(), t.datum()->gettypename().toString() );
+  DoubleDatum* dd = dynamic_cast< DoubleDatum* >( t.datum() );
+  if ( dd )
+  {
+    return ( float ) dd->get();
```
heplesser, line 112: Turn into proper static cast if kept.

sli/tokenutils.cc
```diff
@@ -97,13 +106,20 @@ template <>
...
+  }
+  IntegerDatum* id = dynamic_cast< IntegerDatum* >( t.datum() );
+  if ( id )
+  {
+    return ( float ) id->get();
```
heplesser, line 117: See above.

sli/tokenutils.cc
```diff
@@ -97,13 +106,20 @@ template <>
...
   }
-  return ( float ) id->get();
+
+  // we have to create a Datum object to get the name...
+  DoubleDatum const d;
+  throw TypeMismatch( d.gettypename().toString(), t.datum()->gettypename().toString() );
 }
```
heplesser, line 124: Se if `setValue< float >` below can be removed as well.

heplesser (changes requested):
Thanks! I included some comments in the code below.

sli/tokenutils.cc
```diff
@@ -73,14 +73,23 @@ template <>
...
+    return dd->get();
   }
-  return id->get();
+  IntegerDatum* id = dynamic_cast< IntegerDatum* >( t.datum() );
+  if ( id )
+  {
+    return ( double ) id->get();
+  }
+
+  // we have to create a Datum object to get the name...
```
stinebuu, line 88: This is a bit nitpicky, but could you use capital W for this and the other comments?

sli/tokenutils.cc
ackurth: I removed this function from the code and everything seems to work fine, see  c21eef40b26d45b801cfa606d6bdb6917a99d219.

sli/tokenutils.cc
ackurth: Same as above.

sli/tokenutils.cc
ackurth: See above.

sli/tokenutils.cc
ackurth: See above.

sli/tokenutils.cc
ackurth: I addressed this in b1a81009d4d0115d5debb601e1ff9b175fc28ae5.

stinebuu (approved):
Thanks!

heplesser (approved):

PR #2030, opened 2021-04-27
Add functions to access information about simulation time

marcelkrueger:
This PR adds three new functions to expose information about the simulation time of the current simulation.
Thus, it will allow to access information about the end of the current run, the time when the current run was started and the end of the overall simulation (multiple runs accumulated).

Example: A simulation run in the following way:
```
with nest.RunManager():
   i = 10
   for _ in range(3):
      nest.Run(i)
      i = i + 10
```

would return ( Run(10) / Run(20) / Run(30) ):
```
get_number_of_cycles_in_run():	100 / 200 / 300
get_start_cycle_in_run():	0 / 100 / 300
get_end_cycle_in_simulation():	100 / 300 / 600
```

We would like to use this information in the [insite-pipeline](https://github.com/VRGroupRWTH/insite) to give users the ability to access information about the start and end time of the current simulation to allow for a) buffer pre-allocation b) incorporating this information into visualization clients.

Refactored approach of PR #2006.

### DIFF
--- a/nestkernel/simulation_manager.h
+++ b/nestkernel/simulation_manager.h
@@ -150,19 +150,19 @@ public:
    * Get the number of total cycles in the current run.
    * Can be used to calculate the simulation time of the current run.
    */
-  delay get_number_of_cycles_in_run() const;
+  delay get_number_of_steps_in_run() const;
 
   /**
    * Get the start cycle of the current run.
    * Can be used to calculate the start time of the current run.
    */
-  delay get_start_cycle_in_run() const;
+  delay get_start_step_in_run() const;
 
   /**
    * Get the number of cycles that the current simulation will have been simulated for.
    * Can be used to calculate the end time of the simulation (only considering the runs that have be called so far).
    */
-  delay get_end_cycle_in_simulation() const;
+  delay get_end_step_in_simulation() const;
 
   //! Return start of current time slice, in steps.
   // TODO: rename / precisely how defined?
--- a/nestkernel/simulation_manager.h
+++ b/nestkernel/simulation_manager.h
@@ -146,6 +146,24 @@ public:
   // TODO: Precisely how defined? Rename!
   Time const& get_clock() const;
 
+  /**
+   * Get the number of total cycles in the current run.
+   * Can be used to calculate the simulation time of the current run.
+   */
+  delay get_number_of_cycles_in_run() const;
+
+  /**
+   * Get the start cycle of the current run.
+   * Can be used to calculate the start time of the current run.
+   */
+  delay get_start_cycle_in_run() const;
+
+  /**
+   * Get the number of cycles that the current simulation will have been simulated for.
+   * Can be used to calculate the end time of the simulation (only considering the runs that have be called so far). 
+   */
+  delay get_end_cycle_in_simulation() const;
+
   //! Return start of current time slice, in steps.
   // TODO: rename / precisely how defined?
   delay get_from_step() const;
@@ -247,6 +265,26 @@ inline Time const&
 SimulationManager::get_clock() const
 {
   return clock_;
+}
+
+inline delay 
+SimulationManager::get_number_of_cycles_in_run() const
+{
+  return to_do_total_;
+}
+
+inline delay 
+SimulationManager::get_start_cycle_in_run() const
+{
+  assert( not simulating_ ); //implicit due to using get_time()
+  return get_time().get_steps() - (to_do_total_ - to_do_);
+}
+
+inline delay 
+SimulationManager::get_end_cycle_in_simulation() const
+{
+  assert( not simulating_ ); //implicit due to using get_time()
+  return get_time().get_steps() + to_do_;
 }
 
 inline delay
--- a/nestkernel/simulation_manager.h
+++ b/nestkernel/simulation_manager.h
@@ -160,7 +160,7 @@ public:
 
   /**
    * Get the number of cycles that the current simulation will have been simulated for.
-   * Can be used to calculate the end time of the simulation (only considering the runs that have be called so far). 
+   * Can be used to calculate the end time of the simulation (only considering the runs that have be called so far).
    */
   delay get_end_cycle_in_simulation() const;
 
@@ -267,23 +267,23 @@ SimulationManager::get_clock() const
   return clock_;
 }
 
-inline delay 
+inline delay
 SimulationManager::get_number_of_cycles_in_run() const
 {
   return to_do_total_;
 }
 
-inline delay 
+inline delay
 SimulationManager::get_start_cycle_in_run() const
 {
-  assert( not simulating_ ); //implicit due to using get_time()
-  return get_time().get_steps() - (to_do_total_ - to_do_);
+  assert( not simulating_ ); // implicit due to using get_time()
+  return get_time().get_steps() - ( to_do_total_ - to_do_ );
 }
 
-inline delay 
+inline delay
 SimulationManager::get_end_cycle_in_simulation() const
 {
-  assert( not simulating_ ); //implicit due to using get_time()
+  assert( not simulating_ ); // implicit due to using get_time()
   return get_time().get_steps() + to_do_;
 }
 
--- a/nestkernel/simulation_manager.h
+++ b/nestkernel/simulation_manager.h
@@ -147,22 +147,19 @@ public:
   Time const& get_clock() const;
 
   /**
-   * Get the number of total cycles in the current run.
-   * Can be used to calculate the simulation time of the current run.
+   * Get the simulation duration in the current call to run().
    */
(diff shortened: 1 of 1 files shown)


jougs:
Thanks for this PR and the refreshingly nice and comprehensive description :smile_cat:

Can you please fix the [formatting errors](https://travis-ci.org/github/nest/nest-simulator/builds/768557623#L1104) either manually (most of them just seem to be trailing whitespaces), or by running `extras/format_all_c_c++_files.sh` (for which you need version 3.6 of `clang-format`)? Thanks!

marcelkrueger:
I fixed the formatting errors and Travis seems to be happy. 

I should have checked the formatting beforehand and will make sure to do that beforehand next time. 
Thanks!

jougs:
Don't worry, you're [in](https://github.com/nest/nest-simulator/search?q=fix+style&type=commits) [good](https://github.com/nest/nest-simulator/search?q=pep8&type=commits) [company](https://github.com/nest/nest-simulator/search?q=formatting&type=commits) :stuck_out_tongue_winking_eye: 
Plus I think there's not a lot you could have done about it, as I first had to manually enable CI for you as a first-time contributor.

jougs (changes requested):
I am fine with the code, but would like to have all occurrences of `cycle` replaced by `step`.

jougs (approved):
LGTM. @heplesser: can you please also have a quick look?

nestkernel/simulation_manager.h
```diff
@@ -146,6 +146,24 @@ class SimulationManager : public ManagerInterface
...
+   * Get the start cycle of the current run.
+   * Can be used to calculate the start time of the current run.
+   */
+  delay get_start_step_in_run() const;
+
+  /**
+   * Get the number of cycles that the current simulation will have been simulated for.
+   * Can be used to calculate the end time of the simulation (only considering the runs that have be called so far).
+   */
+  delay get_end_step_in_simulation() const;
```
heplesser, line 165: Why is this method called `..._simulation()` while the other two are called `..._run()`? Shouldn't they be consistent?

heplesser (changes requested):
@marcelkrueger Looks fine except for a naming issue. Could you take a look at it?

nestkernel/simulation_manager.h
marcelkrueger: Compared to the other two functions, which will return the start/end time in their respective run, this function returns the end time over multiple runs that were started in the same RunManager (e.g. when calling Run(10) three times then the _simulation function would return 10/20/30 when being called during the respective runs). I thought of the different runs which are in the same RunManager as "the simulation" and thus named the function ..._simulation(). The other two return information that are local to their respective run.

I have to admit that I wasn't 100% sure what names would align the best with the rest of the NEST codebase to model this distinction between a run and what I assumed to be "the overall simulation".

heplesser: The `get_start_...()` function also returns the "full" time at the beginning of each `run()`,  call, in your example above 0, 100 and 300, i.e., it is always the sum of the total number of steps completed in the past. In the same way, `get_end_...()` returns the corresponding end times, 100, 300 and 600. So the `get_start_...()` and `get_end_...()` methods are symmetrical and should have the corresponding names; I also think the correct preposition here is "of".

I will add a suggestion for changes in the comment and code below. Corresponding name changes are required in the definition of the methods.

nestkernel/simulation_manager.h
```diff
@@ -249,6 +267,26 @@ SimulationManager::get_clock() const
...
+
+inline delay
+SimulationManager::get_start_step_in_run() const
+{
+  assert( not simulating_ ); // implicit due to using get_time()
+  return get_time().get_steps() - ( to_do_total_ - to_do_ );
+}
+
+inline delay
+SimulationManager::get_end_step_in_simulation() const
```
heplesser, line 284:
```suggestion
SimulationManager::get_end_step_of_run() const
```

heplesser (changes requested):
@marcelkrueger I have one principal question and some comments in the code. The principal question is why you return the times in numbers of steps. Is there a specific reason for this from the insite project perspective? I would consider the time steps more of a NEST-internal detail and biological time simulated as an external observable. But since insite interacts with NEST at a technical level, steps might be the right thing.

In comments below, I changed "a run" to "a call to run()", since the concept of a "run" is not defined in NEST as such.

marcelkrueger:
> @marcelkrueger I have one principal question and some comments in the code. The principal question is why you return the times in numbers of steps. Is there a specific reason for this from the insite project perspective? I would consider the time steps more of a NEST-internal detail and biological time simulated as an external observable. But since insite interacts with NEST at a technical level, steps might be the right thing.

@heplesser 
Insite is only interested in the biological time (at least at the moment) and not the steps.
I decided to expose these information as steps as I thought that this is the more universal case and there might be other use cases where this internal view is required and it is not a big problem to convert to biological time in insite.  Therefore, I didn't want to introduce "external" requirements into the simulation manager and just wanted to expose already available information in a minimal fashion.

So from the insite perspective it would be totally sufficient or even nicer to get these information directly in biological time and not in steps. Thus, it would totally be okay for me to change it to return biological time instead of steps if that is the more appropriate interface from your perspective.

marcelkrueger:
@jougs @heplesser 
Just out of curiosity and to get a better understanding. 
Both of you referenced/commented that `steps` would fit better compared to `cycles` : 

In the beginning I used the word `cycle` in the function name and comments to be consistent with the comments that are already in the simulation_manager.h. Namely:
https://github.com/nest/nest-simulator/blob/dafe70e88070a03e18f444e5a02297dfbf6b0cb3/nestkernel/simulation_manager.h#L179-L180

as I am basically just exposing these information. 

Is there a conceptual difference why they are described as `cycles` in the above mentioned occurrences compared to `steps` that will be used in the introduced functions in this PR?

While I also wanted to use `steps` from the beginning, I chose to use `cycles` instead in the first draft to be more  consistent with the existing documentation.

heplesser:
> While I also wanted to use `steps` from the beginning, I chose to use `cycles` instead in the first draft to be more consistent with the existing documentation.

It would be interesting to know how the term entered the documentation here. It should really be "steps" there as well—could you add the fix to this PR? We use "cycle" in a few places throughout the code, but not in a consistent way and as far as I saw during a quick check not for simulations steps.

marcelkrueger:
Okay, I will change the functions to return time and rename/document them accordingly.
I will also change cycles to step in the documentation of the to_do_ variables in the PR.

marcelkrueger:
While refactoring I just encountered the following problem: Since I want to use the information for pre-allocation purposes it would probably be the best idea to call it in the pre_run hook as it is nothing that should change once the run is started. However, the pre_run hook is called before the to_do variables are set: https://github.com/nest/nest-simulator/blob/794d7ebb3ddfd3b2ab4cf5c6ef79954f90058e44/nestkernel/simulation_manager.cpp#L537-L550

So my question is what are the my options or the best approach to get these information before the run starts ? 

These are the possible obvious workarounds that I have in mind: 
a) would it be fine to push the pre_run_hook down and call it after the to_do is set ?
b) would it be fine to pull the to_do_ assignment up and place it before the pre_run_hook ?
c) introduce a new callback that sits weirdly between the pre_run_hook and the actual run

Is there another option that I am currently overlooking? 
If not, would either option a) or b) be okay for you or would that break some assumptions/use-cases regarding the pre_run_hook  ?

heplesser:
@jougs Could you comment on the place where the `io_manager.pre_run_hook()` is called? I think it would make sense to call the hook immediately before actually running, i.e., pushing it further down than it is now. But I don't know enough about the IO Manager ;).

nestkernel/simulation_manager.h
```diff
@@ -176,8 +191,8 @@ class SimulationManager : public ManagerInterface
 
   Time clock_;                     //!< SimulationManager clock, updated once per slice
   delay slice_;                    //!< current update slice
-  delay to_do_;                    //!< number of pending cycles.
-  delay to_do_total_;              //!< number of requested cycles in current simulation.
+  delay to_do_;                    //!< number of pending steps.
```
heplesser, line 194:
```suggestion
  delay to_do_;                    //!< number of pending steps
```

heplesser (commented):
@marcelkrueger Just a minimal suggestion.

nestkernel/simulation_manager.h
marcelkrueger: Fixed in 8f270b4

heplesser (approved):

heplesser:
@jougs I'll leave it to you whether to merge this PR as is and move the pre-hook in a separate PR or wait until that part is discussed.

jougs:
In the interest of having this in the NEST 3.o release, I'm merging as is and we'll look into the position of the hooks later on.

PR #2036, opened 2021-05-13
Remove brackets around NodeCollections in tests

### LINKED ISSUE #2034: Some `fail_or_die` tests fail because of leftover NEST 2 code, not because of the thing that should fail
When working on #2025, I noticed that we have `fail_or_die` tests that are failing, but they are failing because of leftover NEST 2 code, and not because of the actual thing that should fail.

I noticed this for [issue-737.sli](https://github.com/nest/nest-simulator/blob/master/testsuite/regressiontests/issue-737.sli), but the same might happen other places as well.

The tests in `issue-737.sli` use brackets around the nodes when connecting, like we used to have to do. This is not possible anymore, so the tests fail. However, they should fail because we are connecting with parameters that are not allowed. This is now never tested.

https://github.com/nest/nest-simulator/blob/dafe70e88070a03e18f444e5a02297dfbf6b0cb3/testsuite/regressiontests/issue-737.sli#L43-L50

We have to inspect all the sli test, especially those with `fail_or_die`, and make sure we do not use the brackets anymore. I do not think this is a problem for the PyNEST tests, but we should check that as well.

stinebuu:
This fixes #2034.

Remove brackets around NodeCollections when connecting to make tests fail from the actual mistake.

The removed test should actually pass, so I have removed it. It used to fail because we got the node id's and tried to connect them, and this does not work. Connecting two `hh_psc_alpha_gap` neurons does and should work though.

### DIFF
--- a/testsuite/regressiontests/issue-1242.sli
+++ b/testsuite/regressiontests/issue-1242.sli
@@ -27,7 +27,7 @@ Name: testsuite::issue-1242
 Synopsis: (issue-1242) run -> NEST exits if test fails
 
 Description:
-Tests that a connection from a node without proxies to 
+Tests that a connection from a node without proxies to
 a node without proxies that has global targets throws an error
 (instead of failing silently or resulting in inconsistent results).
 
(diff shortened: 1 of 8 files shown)


jougs (changes requested):
Many thanks! Mostly fine. Just one little correction for an oversight.

As the changes are pretty much stereotypical replacements, I reckon that we don't need a second review and am thus removing the second request.

jougs (approved):
LGTM!

PR #2037, opened 2021-05-14
Fix failing assert when connecting spatial populations

hakonsbm:
When connecting spatial populations with a mask, in a corner case it would trigger an assert:
```
ntree_impl.h:486: nest::Ntree<D, T, max_capacity, max_depth>::iterator 
nest::Ntree<D, T, max_capacity, max_depth>::insert(nest::Position<D>, const T&) ... :
Assertion `( pos >= lower_left_ ) && ( pos < lower_left_ + extent_ )' failed.
```

When connecting using a mask on a population with more than 100 nodes, the population is split into 4 (in 2D) or 8 (in 3D) sub-regions. Each region is split recursively until each final region contains at most 100 nodes, or until it reaches a specified maximum depth.

The calculations performed when splitting a region may introduce round-off errors, which causes issues when trying to sort a position that is placed on one of the borders of the new sub-regions. Consequently this causes some nodes to be put in the wrong sub-regions, which triggers the assertion.

The round-off errors are handled by introducing an epsilon value when comparing positions.

### DIFF
--- a/nestkernel/ntree_impl.h
+++ b/nestkernel/ntree_impl.h
@@ -23,6 +23,8 @@
 #ifndef NTREE_IMPL_H
 #define NTREE_IMPL_H
 
+#include <limits>
+
 #include "ntree.h"
 
 // Includes from spatial:
@@ -391,7 +393,8 @@ Ntree< D, T, max_capacity, max_depth >::subquad_( const Position< D >& pos )
   int r = 0;
   for ( int i = 0; i < D; ++i )
   {
-    r += ( 1 << i ) * ( pos[ i ] < lower_left_[ i ] + extent_[ i ] / 2 ? 0 : 1 );
+    r += ( 1 << i )
+      * ( ( ( lower_left_[ i ] + extent_[ i ] / 2 ) - pos[ i ] ) > std::numeric_limits< double >::epsilon() ? 0 : 1 );
   }
 
   return r;
@@ -478,7 +481,13 @@ Ntree< D, T, max_capacity, max_depth >::insert( Position< D > pos, const T& node
   if ( leaf_ )
   {
 
-    assert( ( pos >= lower_left_ ) && ( pos < lower_left_ + extent_ ) );
+#ifdef NDEBUG
+    for ( int i = 0; i < D; ++i )
+    {
+      assert( ( pos - lower_left_ )[ i ] > -std::numeric_limits< double >::epsilon()
+        && ( lower_left_ + extent_ - pos )[ i ] > -std::numeric_limits< double >::epsilon() );
+    }
+#endif
 
     nodes_.push_back( std::pair< Position< D >, T >( pos, node ) );
 
--- a/nestkernel/ntree_impl.h
+++ b/nestkernel/ntree_impl.h
@@ -396,8 +396,8 @@ Ntree< D, T, max_capacity, max_depth >::subquad_( const Position< D >& pos )
     // Comparing against an epsilon value in case there are round-off errors.
     // Using a negative epsilon value because the round-off error may go both ways
     // and the difference we check against may therefore be +/- 10^-16.
-    r += ( 1 << i )
-      * ( ( ( lower_left_[ i ] + extent_[ i ] / 2 ) - pos[ i ] ) > -std::numeric_limits< double >::epsilon() ? 0 : 1 );
+    const bool in_left_half = ( ( lower_left_[ i ] + extent_[ i ] / 2 ) - pos[ i ] ) > -std::numeric_limits< double >::epsilon();
+    r += ( 1 << i ) * ( in_left_half ? 0 : 1 );
   }
 
   return r;
@@ -490,7 +490,7 @@ Ntree< D, T, max_capacity, max_depth >::insert( Position< D > pos, const T& node
       // Using a negative epsilon value because the round-off error may go both ways
       // and the difference we check against may therefore be +/- 10^-16.
       assert( ( pos - lower_left_ )[ i ] > -std::numeric_limits< double >::epsilon()
-        && ( lower_left_ + extent_ - pos )[ i ] > -std::numeric_limits< double >::epsilon() );
+        and ( lower_left_ + extent_ - pos )[ i ] > -std::numeric_limits< double >::epsilon() );
     }
 
     nodes_.push_back( std::pair< Position< D >, T >( pos, node ) );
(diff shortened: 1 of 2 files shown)


nestkernel/ntree_impl.h
```diff
@@ -478,7 +481,11 @@ Ntree< D, T, max_capacity, max_depth >::insert( Position< D > pos, const T& node
   if ( leaf_ )
   {
 
-    assert( ( pos >= lower_left_ ) && ( pos < lower_left_ + extent_ ) );
+    for ( int i = 0; i < D; ++i )
+    {
+      assert( ( pos - lower_left_ )[ i ] > -std::numeric_limits< double >::epsilon()
```
stinebuu, line 486: Could you add a comment about why we need the negative limit here?

stinebuu (changes requested):
Thanks! Is it possible to add a small test as well?

nestkernel/ntree_impl.h
hakonsbm: Done!

nestkernel/ntree_impl.h
```diff
@@ -478,7 +482,14 @@ Ntree< D, T, max_capacity, max_depth >::insert( Position< D > pos, const T& node
   if ( leaf_ )
   {
 
-    assert( ( pos >= lower_left_ ) && ( pos < lower_left_ + extent_ ) );
+    for ( int i = 0; i < D; ++i )
+    {
+      // Comparing against an epsilon value in case there are round-off errors.
+      // Using a negative epsilon value because the round-off error may go both ways
+      // and the difference we check against may therefore be +/- 10^-16.
+      assert( ( pos - lower_left_ )[ i ] > -std::numeric_limits< double >::epsilon()
```
stinebuu, line 490: If it is because of the +/- difference, I think it is better to use `std::abs( ( pos - lower_left_ )[ i ] ) > std::numeric_limits< double >::epsilon() ... `

Do we need to do the same in the case above? Or is it only needed here?

stinebuu (commented):
Thanks! I have a comment and question. :)

nestkernel/ntree_impl.h
hakonsbm: Using `(pos - lower_left) > -epsilon` is not the same as `abs(pos - lower_left) > epsilon`.  The first becomes `true` for `pos - lower_left` values that are higher than `-epsilon`, while the second is `true` for values higher than `+epsilon` or lower than `-epsilon`, which is not what we want. We want to check that the `pos - lower_left` value is higher than zero, but because of round-off errors it may be slightly below zero as well.

In the case above I suppose the value may also be slightly below zero too. I've added a negative sign to the epsilon there.

stinebuu: Ah, very true!

stinebuu (approved):

nestkernel/ntree_impl.h
```diff
@@ -478,7 +484,14 @@ Ntree< D, T, max_capacity, max_depth >::insert( Position< D > pos, const T& node
...
+        && ( lower_left_ + extent_ - pos )[ i ] > -std::numeric_limits< double >::epsilon() );
```
heplesser, line 493:
```suggestion
        and ( lower_left_ + extent_ - pos )[ i ] > -std::numeric_limits< double >::epsilon() );
```

heplesser: This is an expensive assert, but for now we should keep it.

nestkernel/ntree_impl.h
```diff
@@ -391,7 +393,11 @@ Ntree< D, T, max_capacity, max_depth >::subquad_( const Position< D >& pos )
   int r = 0;
   for ( int i = 0; i < D; ++i )
   {
-    r += ( 1 << i ) * ( pos[ i ] < lower_left_[ i ] + extent_[ i ] / 2 ? 0 : 1 );
+    // Comparing against an epsilon value in case there are round-off errors.
+    // Using a negative epsilon value because the round-off error may go both ways
+    // and the difference we check against may therefore be +/- 10^-16.
+    r += ( 1 << i )
+      * ( ( ( lower_left_[ i ] + extent_[ i ] / 2 ) - pos[ i ] ) > -std::numeric_limits< double >::epsilon() ? 0 : 1 );
```
heplesser, line 400: To make this more readable, I'd suggest 
```suggestion
    const bool in_left_half = ( ( lower_left_[ i ] + extent_[ i ] / 2 ) - pos[ i ] ) > -std::numeric_limits< double >::epsilon();
    r += ( 1 << i ) * ( in_left_half ? 0 : 1 );
```

I think as a temporary fix this is fine, but we should revisit the issue again later. My main concern is that this test will work as long as coordinates are close to 1, but of they are much smaller or larger, we need to consider relative error. I would feel more comfortable with an approach where we explicitly pad boundaries in some consistent way.

heplesser (changes requested):
@hakonsbm Thanks! I have one suggestion for clarity, and please create a follow up issue to remind us to consider an even more robust solution later, see also my comment below.

heplesser:
@hakonsbm I have thought a bit more about the rounding problems we face. I believe they result from the fact that we do arithmetic on corners, extents and positions (and anchors) in various places in a way that leads to internal inconsistencies. I believe now that by creating definite (canonical) values and then only using comparison operations between them, we can avoid any internal inconsistencies.

I am thinking of the following:
1. There is exactly one method that shifts positions according to anchors.
1. There is exactly one method that maps a position into the extent according to periodic boundary conditions.
1. The previous two points ensure that we have "canonical" position values that will always be the same for the same input position. Only this canonical position will be used in all further operations.
1. Each ntree quad gets both a `lower_left_` and and `upper_right_` member. Upon splitting, we only compute the lower-left value of the right part of the split and then set upper-right value of the left part to exactly the same value. 
1. Assuming left-closed, right-open intervals, using `(ll <= x) and (x < ur)` comparisons then will give consistent results, where `x` is a canonical position.
1. If `x < ur` comparisons are explicitly required (instead of doing everything with `ll <= x`, then positions located on the right edge of the extent would always be excluded. We might want to define it like this on principle to avoid problems with periodic boundary conditions, where positions on the right edge of the extent would be also on its left edge.

What do you think?

heplesser (approved):

heplesser:
@hakonsbm Thanks for the fixes. I have approved the PR now and will merge it. Could you create a follow-up issue for a more through fix and include my ideas from above?

heplesser:
@hakonsbm Travis is complaining about a formatting error in `ntree_impl.h`. Could you fix that before I merge?

hakonsbm:
@heplesser Sure, I can create an issue on a more thorough fix. However, I'm not sure about your suggestions, but we can take that discussion in the new issue.

PR #2039, opened 2021-05-18
Fix issue with building Sphinx documentation with recent Jinja version

### LINKED ISSUE #2038: Generating Sphinx documentation fails with an error after recent Jinja update
Following the [user documentation workflow](https://nest-simulator.readthedocs.io/en/latest/documentation_workflow/user_documentation_workflow.html) to build the Sphinx documentation fails with the following error:
```
Notebook error:
UndefinedError in model_details/HillTononiModels.ipynb:
'nbformat.notebooknode.NotebookNode object' has no attribute 'tags'
make[3]: *** [doc/CMakeFiles/html.dir/build.make:72: doc/CMakeFiles/html] Error 2
make[2]: *** [CMakeFiles/Makefile2:566: doc/CMakeFiles/html.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:573: doc/CMakeFiles/html.dir/rule] Error 2
make: *** [Makefile:231: html] Error 2
```

When creating a Conda environment with `extras/conda-nest-simulator-dev.yml` Jinja is installed as a dependency. Jinja was recently updated to version 3.0, which breaks the Sphinx build. It is likely that this will make building the documentation on Read the Docs fail as well. 

Jinja 3.0 support will be added with Sphinx 4.1.0, but it is still not released. Using any Jinja version before version 3.0 fixes the problem, so this requirement should be added to the yml file, and possibly also to `extras/nest-simulator-doc-requirements.txt` for RTD.

hakonsbm:
Fixes #2038 by adding a requirement to keep Jinja version below 3 when creating a Conda environment. Also adds the requirement for Read the Docs.

### DIFF
--- a/extras/conda-nest-simulator-dev.yml
+++ b/extras/conda-nest-simulator-dev.yml
@@ -32,12 +32,12 @@
 name: nest-simulator-dev
 channels:
   - conda-forge
-  
+
 dependencies:
   # Basics ------------------
   - python >= 3.8
   - pip
-  
+
   # Building NEST -----------
   - cmake >= 3.12
   - cython
@@ -78,6 +78,7 @@ dependencies:
   - sphinx-tabs
   - sphinx >= 2
   - sphinx_rtd_theme
+  - jinja2 < 3
   - tqdm
   - yamllint
 
--- a/extras/nest-simulator-doc-requirements.txt
+++ b/extras/nest-simulator-doc-requirements.txt
@@ -23,6 +23,7 @@ sphinx>=2
 sphinx_autobuild
 sphinx_gallery
 sphinx_rtd_theme
+jinja2<3
 statsmodels
 tqdm
 yamllint


terhorstd (approved):
I am very unhappy with adding transitional dependencies to the nest-dev environment. Now it looks like NEST depends on an old version of Jinja2, where actually this is a bug in the Sphinx package not stating its dependencies correctly and NEST itself doesn't have any direct interest in Jinja2 at all.

For the time being I agree to add it, in order to prevent everyone's doc generation workflow to break. I could imagine this gets fixed upstream in sphinx/nbformat soon and we should remove the whole jinja2 dependency again. (please create an issue)

hakonsbm:
@terhorstd I agree that it's not optimal. Like I wrote in the issue, Jinja 3.0 support will be added with Sphinx 4.1.0, which, according to their milestones, should be finished June 12.

jessica-mitchell (approved):
Thanks for catching this @hakonsbm. I just want to point out, at least up to now there have been no problems building on RtD, but since a fresh install of the conda environment includes jinja2 3.0 I agree with @terhorstd that this should be approved to not break anyone's workflow.

PR #2042, opened 2021-05-18
Make sure `init_state_` is called for models that need it, and remove redundant ones

stinebuu:
This PR makes sure `init_state_` is called for models that need it. At the moment, the function is not called at all.

I have refactored the `init_state()` and `init_buffers()` functions in `node.cpp` to be one function, `init()`, which first calls `init_state_()` and then `init_buffers_()`.

I have removed `init_state_()` in the models which only copies the state from the prototypes, as this really does nothing. I have then also removed alot of `State_& operator=( const State_& )`.

As far as I can see, all the detectors and generators still have a `init_state_`, and they just call `device_.init_state( pr.device_ );` This function is a virtual function, so it might be overloaded. I have therefore kept them.
See for instance:
https://github.com/nest/nest-simulator/blob/794d7ebb3ddfd3b2ab4cf5c6ef79954f90058e44/models/correlation_detector.cpp#L233-L237

### DIFF
--- a/models/ac_generator.cpp
+++ b/models/ac_generator.cpp
@@ -163,11 +163,9 @@ nest::ac_generator::ac_generator( const ac_generator& n )
  * ---------------------------------------------------------------- */
 
 void
-nest::ac_generator::init_state_( const Node& proto )
+nest::ac_generator::init_state_()
 {
-  const ac_generator& pr = downcast< ac_generator >( proto );
-
-  device_.init_state( pr.device_ );
+  device_.init_state();
 }
 
 void
--- a/models/ac_generator.cpp
+++ b/models/ac_generator.cpp
@@ -163,15 +163,6 @@ nest::ac_generator::ac_generator( const ac_generator& n )
  * ---------------------------------------------------------------- */
 
 void
-nest::ac_generator::init_state_( const Node& proto )
-{
-  const ac_generator& pr = downcast< ac_generator >( proto );
-
-  device_.init_state( pr.device_ );
-  S_ = pr.S_;
-}
-
-void
 nest::ac_generator::init_buffers_()
 {
   device_.init_buffers();
--- a/models/ac_generator.cpp
+++ b/models/ac_generator.cpp
@@ -163,6 +163,14 @@ nest::ac_generator::ac_generator( const ac_generator& n )
  * ---------------------------------------------------------------- */
 
 void
+nest::ac_generator::init_state_( const Node& proto )
+{
+  const ac_generator& pr = downcast< ac_generator >( proto );
+
+  device_.init_state( pr.device_ );
+}
+
+void
 nest::ac_generator::init_buffers_()
 {
   device_.init_buffers();
--- a/models/dc_generator.cpp
+++ b/models/dc_generator.cpp
@@ -134,6 +134,11 @@ nest::dc_generator::dc_generator( const dc_generator& n )
 /* ----------------------------------------------------------------
  * Node initialization functions
  * ---------------------------------------------------------------- */
+void
+nest::dc_generator::init_state_()
+{
+  device_.init_state();
+}
 
 void
 nest::dc_generator::init_buffers_()
--- a/models/dc_generator.cpp
+++ b/models/dc_generator.cpp
@@ -136,15 +136,6 @@ nest::dc_generator::dc_generator( const dc_generator& n )
  * ---------------------------------------------------------------- */
 
 void
-nest::dc_generator::init_state_( const Node& proto )
-{
-  const dc_generator& pr = downcast< dc_generator >( proto );
-
-  device_.init_state( pr.device_ );
-  S_ = pr.S_;
-}
-
-void
 nest::dc_generator::init_buffers_()
 {
   device_.init_buffers();
--- a/nestkernel/node.cpp
+++ b/nestkernel/node.cpp
@@ -46,7 +46,7 @@ Node::Node()
   , thread_( 0 )
   , vp_( invalid_thread_ )
   , frozen_( false )
-  , buffers_initialized_( false )
+  , state_buffers_initialized_( false )
   , node_uses_wfr_( false )
 {
 }
@@ -60,7 +60,7 @@ Node::Node( const Node& n )
   , vp_( n.vp_ )
   , frozen_( n.frozen_ )
   // copy must always initialized its own buffers
-  , buffers_initialized_( false )
+  , state_buffers_initialized_( false )
   , node_uses_wfr_( n.node_uses_wfr_ )
 {
 }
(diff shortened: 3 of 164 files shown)


nestkernel/node.h
```diff
@@ -238,33 +238,29 @@ class Node
...
-  void init_state();
-
-  /**
-   * Initialize buffers of a node.
-   * This function initializes the Buffers of a Node, e.g., ring buffers
-   * for incoming events, buffers for logging potentials.
-   * This function is called before Simulate is called for the first time
+   * that change during Node::update(). After calling init(), the state
+   * variables should have the same values that they had after the node
+   * was created. In practice, they will be initialized to the values
```
clinssen, line 247: Calls to init() after the first call will have no effect due to ``if ( state_buffers_initialized_ ) return;`` — I cannot find a spot where we do ``state_buffers_initialized_ = False``, for instance at the end of set_status().

nestkernel/node_manager.cpp
```diff
@@ -633,7 +621,7 @@ NodeManager::prepare_node_( Node* n )
 {
   // Frozen nodes are initialized and calibrated, so that they
   // have ring buffers and can accept incoming spikes.
-  n->init_buffers();
```
clinssen, line 636: We still need this call, no?

stinebuu: @clinssen this function is renamed to `init()`, and `init_buffers_()` is called from here.

clinssen: Ah yes, I see now! I suppose the original idea was for each model to have all its state variables in the State_/S_, and all its RingBuffers in the Buffers_/B_ member struct. Buffers_ now also seems to be home to things like the GSL solver, so I'm not sure this separation between Buffers_ and State_ still makes sense or can be made cleanly. Note that there's also Variables_ ("internal variables of the model"), the delineation of which is also somewhat unclear. Do we want to maintain Buffers_, State_ and Variables_ as separate member structs? (The fourth one being Parameters_.) This also with an eye to potential future "save-and-load" mechanisms.

nestkernel/node.h
heplesser: The comment here is misleading/outdated. It described intended behaviour at a time when we had `ResetNetwork`. Now, `init()` only ensures that state variables which require dynamic configuration and buffers are configured properly. `SetStatus` can later change the values of state variables, but does not require `init()`.

nestkernel/node_manager.cpp
heplesser: I will need to dig out the description of the P-S-B-V logic from deep archives and post it. Briefly:
- Parameters: part of the neuroscientific model, do not change with time during simulation
- State: part of the neuroscientific model, evolve with time during simulation
- Buffers: implementation details, persist from one call to Simulate to the next
- Variables: auxiliary values precomputed for efficiency
When introduced, this separation significantly improved our ability to think systematically about the different parts of the model, so we should make changes only with great care.

nestkernel/node.h
heplesser: Here a suggestion for a new version of this comment, taking into account that we dropped `ResetNetwork`.

----------

Initialize node prior to first simulation after node has been created.

init() allows the node to configure internal data structures prior to being simulated. The method has an effect only the first time it is called on a given node, otherwise it returns immediately. init() calls virtual functions init_state_() and init_buffers_(). 

--------

Here `init_state_()` and `init_buffers_()` should ideally be links to the pertaining functions. 

Suggestions for their respective docstrings:

init_state

---------

Configure state variables depending on runtime information.

Overload this method if the node needs to adapt state variables prior to first simulation to runtime information, e.g., the number of incoming connections.

------------

init_buffers

----------

Configure persistent internal data structures.

Let node configure persistent internal data structures, such as input buffers or ODE solvers, to runtime information prior to first simulation.

--------------

nestkernel/node.cpp
```diff
@@ -46,7 +46,7 @@ Node::Node()
   , thread_( 0 )
   , vp_( invalid_thread_ )
   , frozen_( false )
-  , buffers_initialized_( false )
+  , state_buffers_initialized_( false )
```
heplesser, line 49: Should we call it just `initialized_`, would fit better with `init()`.

nestkernel/node.cpp
```diff
@@ -69,30 +69,26 @@ Node::~Node()
...
 {
-  if ( buffers_initialized_ )
+  if ( state_buffers_initialized_ )
   {
     return;
   }
 
+  Model const* const model = kernel().model_manager.get_model( model_id_ );
+  assert( model );
+  init_state_( model->get_prototype() );
```
heplesser, line 87: The only reason to pass the prototype was to be able to copy values from it as part of `ResetNetwork`.  This is no longer relevant and `init_state_()` will only be called on nodes that have not been "touched" since creation, except possibly for SetStatus calls. Therefore, the node has all information internally and we should drop the argument to `init_state_()` entirely.

heplesser (changes requested):
@stinebuu Thanks, nice work! I have a few comments below.

models/dc_generator.cpp
```diff
@@ -135,15 +135,6 @@ nest::dc_generator::dc_generator( const dc_generator& n )
  * Node initialization functions
  * ---------------------------------------------------------------- */
 
-void
-nest::dc_generator::init_state_( const Node& proto )
-{
-  const dc_generator& pr = downcast< dc_generator >( proto );
-
-  device_.init_state( pr.device_ );
```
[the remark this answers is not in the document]
stinebuu: good catch, fixed!

nestkernel/node.h
stinebuu: I have updated the comments.

stinebuu:
@heplesser @clinssen Thanks for your comments! I have updated the PR, please have another look.

models/ac_generator.cpp
```diff
@@ -163,11 +163,9 @@ nest::ac_generator::ac_generator( const ac_generator& n )
  * ---------------------------------------------------------------- */
 
 void
-nest::ac_generator::init_state_( const Node& proto )
+nest::ac_generator::init_state_()
 {
```
clinssen, line 167: Should there also be an explicit call (in this and similar cases below) to the superclass init_state_() (here: ``DeviceNode::init_state_()``) as a first line of the function?

stinebuu: @clinssen `DeviceNode` does not have an `init_state_()`. Calling `DeviceNode::init_state_()` implicitly calls `Node::init_state_()` (as `Node` is the parent of `DeviceNode`), and this function is empty. `Device::init_state_()` exists, and is also empty, but I guess this is handled with the `device_.init_state_()` call.

heplesser (approved):

stinebuu:
I noticed that quite a lot of the `State_` docstrings included a sentence about the assignment operator, so I removed it.

heplesser (approved):

nestkernel/node_manager.cpp
clinssen: > I will need to dig out the description of the P-S-B-V logic from deep archives and post it. Briefly:
> 
>     * Parameters: part of the neuroscientific model, do not change with time during simulation
> 
>     * State: part of the neuroscientific model, evolve with time during simulation
> 
>     * Buffers: implementation details, persist from one call to Simulate to the next
> 
>     * Variables: auxiliary values precomputed for efficiency
> 
> 
> When introduced, this separation significantly improved our ability to think systematically about the different parts of the model, so we should make changes only with great care.

@heplesser: what about the last sentence in your comment—should this PR's changes still be discussed in this context?

heplesser: @clinssen I am not entirely sure I understand what you mean by "still be discussed in this context"? My suggestion is that within this PR, we clean up in the sense of making sure that both `init_state_()` and `init_buffers_()` are called once on each node, via `node::init()`. Any further discussion should be moved to a follow-up issue.

stinebuu: @heplesser @clinssen I have created issue #2050 for further discussion.

clinssen (approved):
Cheers!

PR #2048, opened 2021-05-25
C++ implementations of spatial distribution functions

hakonsbm:
When connecting using certain spatial distributions, calculation of the distribution may dominate the connection time. This is because the spatial distribution functions are defined as combinations of NEST parameters. As the expression for a distribution function becomes more complicated, there is a growing overhead for each calculation. Additionally, defining a combination of NEST parameters makes the compiler unable to do optimizations on the expression.

This PR adds C++ implementations of spatial distribution functions, which leads to the following improvements in connection time involving the different spatial distributions:

- 17% reduced connection time with `exponential` distribution
- 30% reduced connection time with `gamma` distribution
- 51% reduced connection time with `gaussian` distribution
- 72% reduced connection time with `gaussian2d` distribution

### DIFF
--- a/nestkernel/parameter.cpp
+++ b/nestkernel/parameter.cpp
@@ -342,8 +342,7 @@ Gaussian2DParameter::value( RngPtr rng,
 GammaParameter::GammaParameter( const DictionaryDatum& d )
   : p_( getValue< ParameterDatum >( d, "x" )->clone() )
   , kappa_( getValue< double >( d, "kappa" ) )
-  , theta_( getValue< double >( d, "theta" ) )
-  , inv_theta_( 1.0 / theta_ )
+  , inv_theta_( 1.0 / getValue< double >( d, "theta" ) )
   , delta_( std::pow( inv_theta_, kappa_ ) / std::tgamma( kappa_ ) )
 {
   parameter_is_spatial_ = true;
--- a/nestkernel/parameter.cpp
+++ b/nestkernel/parameter.cpp
@@ -281,6 +281,12 @@ ExpDistParameter::ExpDistParameter( const DictionaryDatum& d )
   , inv_beta_( 1.0 / getValue< double >( d, "beta" ) )
 {
   parameter_is_spatial_ = true;
+
+  const auto beta = getValue< double >( d, "beta" );
+  if ( beta <= 0 )
+  {
+    throw BadProperty( "beta > 0 required for exponential distribution parameter, got beta=" + std::to_string( beta ) );
+  }
 }
 
 double
@@ -298,6 +304,12 @@ GaussianParameter::GaussianParameter( const DictionaryDatum& d )
   , inv_two_std2_( 1.0 / ( 2 * getValue< double >( d, "std" ) * getValue< double >( d, "std" ) ) )
 {
   parameter_is_spatial_ = true;
+
+  const auto std = getValue< double >( d, "std" );
+  if ( std <= 0 )
+  {
+    throw BadProperty( "std > 0 required for gaussian distribution parameter, got std=" + std::to_string( std ) );
+  }
 }
 
 double
@@ -325,6 +337,25 @@ Gaussian2DParameter::Gaussian2DParameter( const DictionaryDatum& d )
                                          * getValue< double >( d, "std_x" ) * getValue< double >( d, "std_y" ) ) )
 {
   parameter_is_spatial_ = true;
+
+  const auto rho = getValue< double >( d, "rho" );
+  const auto std_x = getValue< double >( d, "std_x" );
+  const auto std_y = getValue< double >( d, "std_y" );
+  if ( rho >= 1 or rho <= -1 )
+  {
+    throw BadProperty(
+      "-1 < rho < 1 required for gaussian2d distribution parameter, got rho=" + std::to_string( rho ) );
+  }
+  if ( std_x <= 0 )
+  {
+    throw BadProperty(
+      "std_x > 0 required for gaussian2d distribution parameter, got std_x=" + std::to_string( std_x ) );
+  }
+  if ( std_y <= 0 )
+  {
+    throw BadProperty(
+      "std_y > 0 required for gaussian2d distribution parameter, got std_y=" + std::to_string( std_y ) );
+  }
 }
 
 double
@@ -346,6 +377,16 @@ GammaParameter::GammaParameter( const DictionaryDatum& d )
   , delta_( std::pow( inv_theta_, kappa_ ) / std::tgamma( kappa_ ) )
 {
   parameter_is_spatial_ = true;
+
+  if ( kappa_ <= 0 )
+  {
+    throw BadProperty( "kappa > 0 required for gamma distribution parameter, got kappa=" + std::to_string( kappa_ ) );
+  }
+  const auto theta = getValue< double >( d, "theta" );
+  if ( theta <= 0 )
+  {
+    throw BadProperty( "theta > 0 required for gamma distribution parameter, got theta=" + std::to_string( theta ) );
+  }
 }
 
 double
--- a/nestkernel/parameter.cpp
+++ b/nestkernel/parameter.cpp
@@ -20,6 +20,8 @@
  *
  */
 
+#include <cmath>
+
 #include "node_collection.h"
 #include "node.h"
 #include "spatial.h"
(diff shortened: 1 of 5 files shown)


nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+    , py_( p.py_->clone() )
+    , mean_x_( p.mean_x_ )
+    , mean_y_( p.mean_y_ )
+    , std_x_( p.std_x_ )
+    , std_y_( p.std_y_ )
+    , rho_( p.rho_ )
+    , std_x2_( p.std_x2_ )
+    , std_y2_( p.std_y2_ )
+    , std_xy_( p.std_xy_ )
+    , two_sub_2rho2_( p.two_sub_2rho2_ )
```
heplesser, line 1879: Do we need to keep all parameters around, e.g., `std_x_`, `std_y_`, `rho_`? It should be enough to keep just the precomputed values we actually use in the calculations, shouldn't it? Could also apply to other of these parameters.

heplesser (changes requested):
@hakonsbm Thanks, looks good and speedup is impressive. I added some more suggestions that might provide even more speed ;).

pynest/nest/tests/test_spatial/test_spatial_distributions.py
```diff
@@ -131,6 +132,8 @@ def __init__(self, seed, dim, L, N, spatial_distribution, distribution_params=No
                             (math.sqrt(2) * math.log((.1 - 0) / 1))},
             'gaussian': {'p_center': 1., 'sigma': self._L / 4.,
                          'mean': 0., 'c': 0.},
+            'gaussian2d': {'p_center': 1., 'sigma_x': self._L / 4., 'sigma_y': self._L / 4.,
+                           'mean_x': 0., 'mean_y': 0., 'rho': 0., 'c': 0.},
```
stinebuu, line 136: I think we should test with some other variables than the default ones, in case we for some reason have used the default values instead of parameter names or something similar.

nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+  double
+  value( RngPtr, Node* ) override
+  {
+    throw BadParameterValue( "Exponential distribution parameter can only be used when connecting." );
+  }
+
+  double
+  value( RngPtr, index, Node*, thread ) override
+  {
+    throw KernelException( "Exponential distribution parameter can only be used when using ConnectLayers." );
```
stinebuu, line 1768: I think this should be changed to something like `Exponential distribution parameter can only be used when connecting spatially distributed nodes`. Users will not understand what `ConnectLayers` is. Also for all similar error messages below.

nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+    delete p_;
+  }
+
+  /**
+   * @returns the value of the parameter.
+   */
+  double
+  value( RngPtr, Node* ) override
+  {
+    throw BadParameterValue( "Exponential distribution parameter can only be used when connecting." );
```
stinebuu, line 1762: Is this documented somewhere? It should probably be stated in the docstring for the exponential distribution. The same holds for all the new ones.

nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+  {
+    throw BadParameterValue( "Gaussian distribution parameter can only be used when connecting." );
+  }
+
+  double
+  value( RngPtr, index, Node*, thread ) override
+  {
+    throw KernelException( "Gaussian distribution parameter can only be used when using ConnectLayers." );
+  }
+  double value( RngPtr rng,
```
stinebuu, line 1832: Add a line here.

nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+  {
+    throw BadParameterValue( "Exponential distribution parameter can only be used when connecting." );
+  }
+
+  double
+  value( RngPtr, index, Node*, thread ) override
+  {
+    throw KernelException( "Exponential distribution parameter can only be used when using ConnectLayers." );
+  }
+  double value( RngPtr rng,
```
stinebuu, line 1770: empty line

nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+  {
+    throw BadParameterValue( "Gaussian 2D parameter can only be used when connecting." );
+  }
+
+  double
+  value( RngPtr, index, Node*, thread ) override
+  {
+    throw KernelException( "Gaussian 2D parameter can only be used when using ConnectLayers." );
+  }
+  double value( RngPtr rng,
```
stinebuu, line 1904: New line.

nestkernel/parameter.h
```diff
@@ -1724,6 +1724,275 @@ class DimensionParameter : public Parameter
...
+  {
+    throw BadParameterValue( "Gamma distribution parameter can only be used when connecting." );
+  }
+
+  double
+  value( RngPtr, index, Node*, thread ) override
+  {
+    throw KernelException( "Gamma distribution parameter can only be used when using ConnectLayers." );
+  }
+  double value( RngPtr rng,
```
stinebuu, line 1976: And here :)

nestkernel/parameter.cpp
```diff
@@ -273,4 +275,92 @@ RedrawParameter::value( RngPtr rng,
...
+  , py_( getValue< ParameterDatum >( d, "y" )->clone() )
+  , mean_x_( getValue< double >( d, "mean_x" ) )
+  , mean_y_( getValue< double >( d, "mean_y" ) )
+  , std_x_( getValue< double >( d, "std_x" ) )
+  , std_y_( getValue< double >( d, "std_y" ) )
+  , rho_( getValue< double >( d, "rho" ) )
+  , std_x2_( std_x_ * std_x_ )
+  , std_y2_( std_y_ * std_y_ )
+  , std_xy_( std_x_ * std_y_ )
+  , two_sub_2rho2_( 2. * ( 1. - rho_ * rho_ ) )
```
stinebuu, line 326: I do not think I understand what the first `2` (not the `two` at the beginning) in the name comes from. The last is the `rho_` squared, but what is the first one? Shouldn't it be `two_sub_rho2` (or `two_one_sub_rho2`, but that is a bit too much..)

stinebuu (changes requested):
Nice work! That is a huge performance gain. I have some small comments and suggestions.

heplesser:
@hakonsbm Thanks for turning divisions into multiplications—I am curious about the performance gain we'll see from that. I wonder if we should add checks for the validity of parameter values. Some cannot be zero, others must be (strictly) positive. Right now, we risk division by zero errors and possibly strange behavior if negative numbers are used where they make no sense. The constructors could just throw exceptions for invalid values.

nestkernel/parameter.cpp
hakonsbm: The idea for the name was from the expanded expression: `2*(1 - rho_*rho_) = 2 - 2*rho_**2`. But I've changed the naming scheme now.

hakonsbm:
@heplesser Apparently GCC with `-O3` figured out those changes on its own, so the performance is exactly the same. But it could be that not all compilers manage to make the same efficiency improvements as GCC. And I can add some additional checks for invalid values.

heplesser:
Compilers are amazing indeed. The `const`-ness of the member variables probably allowed the compiler to do this. I now wonder if my suggestions in that light were even a bit counter-productive, because we cannot have `inv_a_(1/a)` initialisers if we want to avoid division-by-zero errors: we need to check first, then assign in the body of the constructor, but then `inv_a_` can no longer be `const`.  One way out here could be to use the following initialiser:
```c++
inv_a_( 1 / a ? a != 0 : numerics::nan )
```
followed by a check in the constructor body. That would give us the best of both worlds.

Another question: Why do we need to `p->clone()`? It would be good to comment the reason somewhere. Could the `params*` be decorated with some const-ness? We probably only use `const` methods  on the parameters, so `params const * const` might work (or do we run into trouble with destruction when we have so much const?).

heplesser (commented):
Thanks for the tests. But do we not risk division by zero errors before the exceptions are reached?

Why is `parameter_is_spatial_` set by assignment and not initializer?

hakonsbm:
@heplesser 

> But do we not risk division by zero errors before the exceptions are reached?

I'm not sure if that's a problem. Dividing a float by zero should become `inf` without a crash. It will then continue to the checks in the constructor body. The checks are of the passed argument, not the result of the division, so that should be fine.

> Why is `parameter_is_spatial_` set by assignment and not initializer?

`parameter_is_spatial_` is a variable in the base class `Parameter`, so it cannot be initialized with an initializer in the subclasses. It could alternatively be initialized in a `Parameter` constructor, but I went with declaring it as `protected` and setting it with assignment to try to reduce the boilerplate code.

> Why do we need to p->clone()?

The `clone()` method was included from when the NEST parameters were adapted from the Topology parameters. I think that without it, a parameter could be destroyed while a pointer to it is still in use by another parameter. Unless there are other reasons for it, I think they can be converted to smart pointers for all parameters, which removes the need for both cloning and destroying the `Parameter` objects. If that doesn't work I can try to document the need for `clone()` somewhere. But that's a job for another PR I think.

> Could the params* be decorated with some const-ness?

The stored parameter cannot be const because the method `value()` cannot be const. That is because some parameters store random distributions or other variables that change when `value()` is called. The pointer can be const though, so I'll add that.

heplesser:
> > But do we not risk division by zero errors before the exceptions are reached?
> 
> I'm not sure if that's a problem. Dividing a float by zero should become `inf` without a crash. It will then continue to the checks in the constructor body. The checks are of the passed argument, not the result of the division, so that should be fine.

Could you test this?

> > Why is `parameter_is_spatial_` set by assignment and not initializer?
> 
> `parameter_is_spatial_` is a variable in the base class `Parameter`, so it cannot be initialized with an initializer in the subclasses. It could alternatively be initialized in a `Parameter` constructor, but I went with declaring it as `protected` and setting it with assignment to try to reduce the boilerplate code.

Would it really cause so much boilerplate code to have a `Parameter(bool is_spatial=false)` constructor? But that would be a matter for a follow-up issue.
> 
> > Why do we need to p->clone()?
> 
> The `clone()` method was included from when the NEST parameters were adapted from the Topology parameters. I think that without it, a parameter could be destroyed while a pointer to it is still in use by another parameter. Unless there are other reasons for it, I think they can be converted to smart pointers for all parameters, which removes the need for both cloning and destroying the `Parameter` objects. If that doesn't work I can try to document the need for `clone()` somewhere. But that's a job for another PR I think.

That sounds a lot like we should replace them with smart pointers. But that would be a follow-up issue as well.

> > Could the params* be decorated with some const-ness?
> 
> The stored parameter cannot be const because the method `value()` cannot be const. That is because some parameters store random distributions or other variables that change when `value()` is called. The pointer can be const though, so I'll add that.

OK.

hakonsbm:
> Could you test this?

Both GCC and Clang follow the IEEE 754 standard, and testing them by providing a zero for the denominator results in `inf`, as expected. The Intel compiler also follows the IEEE 754 standard.

> Would it really cause so much boilerplate code to have a `Parameter(bool is_spatial=false)` constructor? But that would be a matter for a follow-up issue.
> (...)
> That sounds a lot like we should replace them with smart pointers. But that would be a follow-up issue as well.

I'll create a follow-up fixing these issues.

heplesser (approved):
Fine from my side now.

pynest/nest/tests/test_spatial/test_spatial_distributions.py
stinebuu: No, I was more suggesting using other values in this test, not for `sigma`, but for `mean_x` and `rho` so they are not equal to the default.

stinebuu (commented):
I just have the tiny comment about the test, but other than that I am happy.

pynest/nest/tests/test_spatial/test_spatial_distributions.py
hakonsbm: Done.

stinebuu (approved):
Thanks!

PR #2055, opened 2021-05-28
Refactor the `Parameter` classes

hakonsbm:
This PR refactors the `Parameter` classes by

- Adding options to the base class constructor, to avoid having to set `parameter_is_spatial_` and `parameter_returns_int_only_` directly.
- Going from handling parameters as dumb pointers internally to handling them as `shared_ptr`s, which avoids the need to clone parameters and `delete` pointers manually.
  - Parameter operators (`min`, `cos`, `redraw`, etc.) are moved out of the `Parameter` class and are now independent functions because all Parameter arguments must be passed as `shared_ptr`s.


nestkernel/parameter.h
```diff
@@ -61,6 +65,11 @@ class Parameter
   {
```
heplesser, line 65: This constructor seems odd since it does nothing, it does not even set `parameter_is_spatial_` or `parameter_returns_int_only_`.

nestkernel/parameter.h
```diff
@@ -84,88 +93,6 @@ class Parameter
     return value( rng, nullptr );
```
heplesser, line 93: This and the previous `value()` method lack Doxygen comments and even the comment for the first does not explain the arguments.

nestkernel/parameter.h
```diff
@@ -49,9 +49,13 @@ class Parameter
 {
 public:
   /**
-   * Creates an Parameter with default values.
+   * Creates a Parameter with default values.
    */
-  Parameter() = default;
+  Parameter( bool is_spatial = false, bool returns_int_only = false )
+    : parameter_is_spatial_( is_spatial )
+    , parameter_returns_int_only_( returns_int_only )
```
heplesser, line 56: Why not drop `parameter_` from the field names here, it is rather obvious ...

heplesser (changes requested):
@hakonsbm Looks good, just some small comments. Is the `ConverseParameter` removed with this PR? Have you performed any benchmarks to see if this PR affects construction time?

nestkernel/parameter.h
hakonsbm: True, I have removed it.

nestkernel/parameter.h
hakonsbm: Done.

hakonsbm:
@heplesser I've incorporated your suggestions and done some further clean-up. And yes, `ConverseParameter` is removed because it was unused. 

The performance looks as good as unchanged, also with a large number of combined `Parameter`s.

|   num. `Parameter`s |   master (4add1f626) |    this PR |  
|-------------:|-------:|-------:|
|            1 |   4.28 |   4.38 |
|            5 |   4.75 |   4.78 |
|           10 |   4.81 |   4.87 |

heplesser (approved):
Looks good!

nestkernel/parameter.h
```diff
@@ -1974,142 +1415,147 @@ class GammaParameter : public Parameter
...
+  const DictionaryDatum& d );
 
-inline Parameter*
-Parameter::pow( const double exponent ) const
-{
-  return new PowParameter( *this, exponent );
-}
+/**
+ * Create a parameter that chooses between two other parameters,
+ * based on one a given condition parameter. The resulting value of the condition parameter
```
stinebuu, line 1496:
```suggestion
 * based on a given condition parameter. The resulting value of the condition parameter
```

stinebuu (commented):
There is a tiny typo, otherwise all looks very nice.

stinebuu (approved):

PR #2056, opened 2021-06-01
Fix catching build warnings in GitHub Actions

hakonsbm:
Warnings when building in GitHub Actions are currently not caught. This has caused some additional warnings to slip through. This PR fixes the regex to catch the warnings. It also fixes warnings that have been building up and not been caught.

### DIFF
--- a/extras/parse_build_log.py
+++ b/extras/parse_build_log.py
@@ -367,7 +367,7 @@ def makebuild_summary(log_filename, msg_make_section_start,
     # with some warnings, this would be a good point to re-set the
     # expected_warnings variable conditionally for that build_type.
 
-    nest_warning_re = re.compile(f'.* ({build_dir}.*: warning:.*)')
+    nest_warning_re = re.compile(f'.*({build_dir}.*: warning:.*)')
     known_warnings = [
         f'{build_dir}/sli/scanner.cc:642:13: warning: this statement may fall through [-Wimplicit-fallthrough=]',
         f'{build_dir}/sli/scanner.cc:674:19: warning: this statement may fall through [-Wimplicit-fallthrough=]',
--- a/extras/parse_build_log.py
+++ b/extras/parse_build_log.py
@@ -412,10 +412,10 @@ def makebuild_summary(log_filename, msg_make_section_start,
 
     if in_make_section:
         # 'make' was not completed.
-        return False, None, None, None, None
+        return False, None, None, None, None, None
     else:
         # There is no 'make' section at all.
-        return None, None, None, None, None
+        return None, None, None, None, None, None
 
 
 def testsuite_results(log_filename, msg_testsuite_section_start,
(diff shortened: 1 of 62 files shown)


extras/parse_build_log.py
```diff
@@ -412,10 +424,10 @@ def makebuild_summary(log_filename, msg_make_section_start,
 
     if in_make_section:
         # 'make' was not completed.
-        return False, None, None, None, None
+        return False, None, None, None, None, None
     else:
         # There is no 'make' section at all.
-        return None, None, None, None, None
+        return None, None, None, None, None, None
```
jougs, line 430: Thanks for catching this. Maybe you could add a comment explaining what the return values mean? That could potentially prevent such oversights later on.

models/aeif_cond_alpha.cpp
```diff
@@ -167,6 +167,16 @@ nest::aeif_cond_alpha::State_::State_( const State_& s )
...
+nest::aeif_cond_alpha::State_& nest::aeif_cond_alpha::State_::operator=( const State_& s )
+{
+  r_ = s.r_;
+  for ( size_t i = 0; i < STATE_VEC_SIZE; ++i )
+  {
+    y_[ i ] = s.y_[ i ];
+  }
+  return *this;
+}
+
```
jougs, line 179: This re-adds `operator=()` that was removed in #2042 (for example [here](https://github.com/nest/nest-simulator/pull/2042/files#diff-7572728594d53870157ce7b8b13f1122ec6a1bde22908071f022dfb70c2a1e04L170)). Why?

The same comment applies to many of the changes below.

@stinebuu: what do you think about this?

heplesser (approved):
@hakonsbm This looks fine to me.

models/aeif_cond_alpha.cpp
heplesser: @jougs @hakonsbm @stinebuu I think we actually need the assignment operators for `Parameters_` and `State_` for all models to support this pattern:
```c++
inline void
aeif_psc_alpha::set_status( const DictionaryDatum& d )
{
  Parameters_ ptmp = P_;     // temporary copy in case of errors
  ptmp.set( d, this );       // throws if BadProperty
  State_ stmp = S_;          // temporary copy in case of errors
  stmp.set( d, ptmp, this ); // throws if BadProperty
  ArchivingNode::set_status( d );
  P_ = ptmp;
  S_ = stmp;
}
```

I fear we overlooked this with #2042, good to have caught this (my bad ...). This also seems to indicate that we have no tests picking up on this. For models with reasonably plain or smart members, I suppose the automatically constructed assignment operators do the trick, but not for those with C-style arrays as this one. But we should we rely on default operators even where they work?

heplesser:
@jougs @hakonsbm @stinebuu I elevated this to critical and added it to the NEST 3 MS, because I fear #2042 may have broken some neuron models which are fixed by the re-introduced assignment operators here.

extras/parse_build_log.py
hakonsbm: Done!

heplesser:
Two approvals and all passed, so I will merge :).

PR #2064, opened 2021-06-09
Add user-level documentation for built-in timers

suku248:

PR #2065, opened 2021-06-09
Correct documentation of local_spike_counter.

heplesser:
The spike counter has, at least since 2.20.1, been reset on every call to `Simulate()`. This PR describes this correctly in the PyNEST API doc; see also #2064.


jessica-mitchell (approved):
@heplesser seeing this is a very small change, I didn't build the docs and trust that it works. Otherwise looks good to me!

jougs (approved):

PR #2073, opened 2021-06-10
pip install nest-simulator

Helveg:
😏 

Will need to call with a CI person to integrate a new workflow that assembles the package ever so slightly more professional, but this draft shows how easy it is to package the whole NEST repo as a source distribution, run the default CMake and have pip "install" just the Python files that were already CMade there.

All you need to do is run `pypkg.sh` and a functional `sdist` is produced that can be directly pip installed on any machine.


Helveg:
@gtrensch @terhorstd Can you check whether the CI that I added succeeds and produces an installable source distribution?

The `--mpi=ON` option doesn't seem to work yet, but it was a proof of concept; We should think of a full list of CMake flags that we want to expose for `pip install`. It has to be done through a very awkward syntax though: `pip install nest-simulator --install-option="--mpi=ON"` so any strings that need quotes would have to escape those etc. It's not great UX so best to just have those users use CMake.

Helveg:
Also @terhorstd it still produces bad NEST versions, you mentioned something about a fix during the release party?

jessica-mitchell:
Hi @Helveg thanks for this PR!

There still needs to be instructions for installing nest via pip in the docs.

Could you please add a few brief lines about how to install nest via pip to the following file: doc/userdoc/installation/index.rst 
Use the the Ubuntu install instructions as a guide.

Note that the installation options are currently split into reStructuredText tabs. Please add a section called `.. tab:: pip` followed by the text _before_ the Ubuntu section, since I think this would be one of the most popular ways to install nest and should be first.

To check the HTML output, you can build the docs locally
See: https://nest-simulator.readthedocs.io/en/stable/documentation_workflow/user_documentation_workflow.html#setting-up-your-environment)

Message me if you have any questions, thanks!

stinebuu:
@Helveg Apologies for taking ages before testing this. I have tested now though, and I do not think I am able to run with MPI, even though I specify that I want to install with MPI.

I installed with
`pip install nest-on-square-wheels --install-option="--mpi=ON" --install-option="--jobs=4"`,
and I am able to import nest, and run some examples very nicely. But, when I try to run the [Potjans 2014](https://github.com/nest/nest-simulator/tree/master/pynest/examples/Potjans_2014) example with MPI the example runs, just without MPI.

I ran the example as `mpirun -np 4 python <path-to-nest-src>/pynest/examples/Potjans_2014/run_microcircuit.py`, and got that `Total number of virtual processes: 1`. When simulating, NEST tells me that I am running without MPI:
potjans_mpi

Do you know what might be wrong?

Helveg:
Thanks for testing!

I think that the last uploaded version of `nest-on-square-wheels` might have been based on commits before https://github.com/Helveg/nest-simulator/commit/bfe2b622f87c4198e5e0cdbac07d74e36819b49f and the flags weren't properly passed. I uploaded a new version, could you try again? (make sure that it's fetching the new 3.0.2). It compiles with `make install -j` which might slow your computer down, but I added another flag `--jobs` (e.g. `pip install nest-simulator --install-option="--mpi=ON" --install-option="--jobs=4"`) to control that.

During the last NEST hackathon we decided to give a go at binary wheels instead of the source distribution that this PR produces, so I'm converting it to a draft. 

@jougs @heplesser @gtrensch What do you think about keeping this workflow around and in addition to wheels also provide this source distribution which'll basically just do the CMake workflow locally with `pip install nest-simulator --no-binary`? I'm in favor of not providing anything else than "super-vanilla" through `pip`, advanced use cases should use CMake directly.

stinebuu:
@Helveg Am I still supposed to install with `nest-on-square-wheels`? When I try with `pip install nest-simulator --install-option="--mpi=ON" --install-option="--jobs=4"` I get
```
UserWarning: Disabling all use of wheels due to the use of --build-option / --global-option / --install-option.
  cmdoptions.check_install_build_global(options)
ERROR: Could not find a version that satisfies the requirement nest-simulator
ERROR: No matching distribution found for nest-simulator
```

I have tried installing again with `nest-on-square-wheels` (so `pip install nest-on-square-wheels --install-option="--mpi=ON" --install-option="--jobs=4"`, and it looks like it is using version `3.0.2`, I get a message stating `Successfully installed nest-on-square-wheels-3.0.2`, but I am still not able to run with MPI.
Just as when I try with `pip install nest-simulator` I get a warning stating
```
UserWarning: Disabling all use of wheels due to the use of --build-option / --global-option / --install-option.
  cmdoptions.check_install_build_global(options)
```
Is this the problem?

Helveg:
Sorry, I copy pasted the wrong command. The warning is OK, it's warning you that if you use `--install-option` that you can't get a binary wheel and the setup has to be run locally, which is normal. The MPI thing should work though, but I don't have time to look into it and when I do it'll be solved by providing wheels :)

stinebuu:
Ok! Then I will wait with testing again until the wheels are here ;)

PR #2077, opened 2021-06-14
Fix build log parser failed test list

### LINKED ISSUE #1945: Improved robustness of build log parser
On certain occasions, the build log parser will present a long list of failed tests even if only a single test failed. For an example, see https://github.com/nest/nest-simulator/runs/1980183102?check_suite_focus=true, where only `mpitests/test_multiple_synapses_spatial_networks.sli` failed but 68 tests are listed as failing below the summary table.

hakonsbm:
Resolves #1945 by updating the `summarize_tests.py` script for `junitparser` version 2.0. With version 2.0 of `junitparser` [`case.result` is changed to a list](https://junitparser.readthedocs.io/en/latest/#note-on-version-2). This caused skipped tests to not be detected as skipped, and therefore added to the list of failing tests which is printed if any other tests fail.

### DIFF
--- a/testsuite/summarize_tests.py
+++ b/testsuite/summarize_tests.py
@@ -33,6 +33,9 @@ import os
 import sys
 
 
+assert int(jp.version.split('.')[0]) >= 2, 'junitparser version must be >= 2'
+
+
 def parse_result_file(fname):
 
     results = jp.JUnitXml.fromfile(fname)
--- a/testsuite/summarize_tests.py
+++ b/testsuite/summarize_tests.py
@@ -38,7 +38,7 @@ def parse_result_file(fname):
     results = jp.JUnitXml.fromfile(fname)
 
     failed_tests = ['.'.join((case.classname, case.name)) for case in results
-                    if case.result and not isinstance(case.result, jp.junitparser.Skipped)]
+                    if case.result and not isinstance(case.result[0], jp.junitparser.Skipped)]
 
     return {'Tests': results.tests,
             'Skipped': results.skipped,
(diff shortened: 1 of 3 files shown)


heplesser (commented):
@hakonsbm Thanks for this! A few questions:

- Should we add a version requirement to junitparser in the right places?
- Have you tested this by breaking a few tests?
- Are we sure we will only ever need element [0] from the list?

hakonsbm:
@heplesser 

>  Should we add a version requirement to junitparser in the right places?

Done!

>  Have you tested this by breaking a few tests?

Yes, to find out what caused this I had to break a few ~eggs~ tests. See for example [this output](https://github.com/hakonsbm/nest-simulator/runs/2817966318?check_suite_focus=true#step:13:7701) with a failing test.

>  Are we sure we will only ever need element [0] from the list?

Right now it seems like each testcase always give at most one result and skipped tests give exactly one result.

testsuite/summarize_tests.py
```diff
@@ -38,7 +38,7 @@ def parse_result_file(fname):
     results = jp.JUnitXml.fromfile(fname)
 
     failed_tests = ['.'.join((case.classname, case.name)) for case in results
-                    if case.result and not isinstance(case.result, jp.junitparser.Skipped)]
+                    if case.result and not isinstance(case.result[0], jp.junitparser.Skipped)]
```
heplesser, line 41: To protect against future changes and make sure we do not underreport failures, I would suggest to add an assertion here by putting the following right ahead of this list comprehension:
```python
assert all(len(case.result) == 1 for case in results if case.result), "Case result has unexpected length > 1"
```

heplesser (commented):
Just one suggestion for an additional protection against additional result elements.

heplesser (approved):

heplesser:
Minimal but important fix to test reporting, therefore merging based on single review.

PR #2084, opened 2021-06-29
Warn users they are using lastest on ReadtheDocs 

### LINKED ISSUE #2083: Fix RtD warning text about version 3.0 to reference master
**Describe the flaw**
A warning is displayed on RtD in the version 'latest' (master) that warns users that version 3.0 is still incomplete.  This is no longer true.

**Suggest an improvement**
The warning can stay in place but the text should refer to the fact that the version being seen is master and thus it is actively being developed and tested.

jessica-mitchell:
This PR alters the text in the Warning that appears above each page to warn users they are in the latest version of NEST.

Resolves #2083


heplesser (approved):

heplesser:
Important and straightforward doc update, thus merging with single review.

PR #2088, opened 2021-06-30
Remove undocumented weighted_spikes_{ex,in} from iaf_psc_{alpha,exp} recordables

### LINKED ISSUE #2013: Remove undocumented weighted_spikes_{ex,in} recordables?
`iaf_psc_alpha` and `iaf_psc_exp` (and only those) offer recordables `weighted_spikes_{ex,in}`. They have been available at least since NEST 2.6.1, but are not documented on illustrated in any example.

These recordables seem problematic to me because they show instantaneous summed spike weights at a given time step. As long as the recording interval of the multimeter is equal to the time step, this is fine, but if, e.g., time step is 0.1 ms and recording interval 1.0 ms, one would only get every 10th value. For recordables such as `V_m` and `I_{ex,in}`, this is less problematic because they are integrated (smoothed) over time. This property is also not documented.

We need to discuss whether to keep this feature.

heplesser:
This fixes #2013.


jougs (commented):
It seems like there are tests that use the removed variables. Please check throughout the code base.

stinebuu (approved):
Thanks!

jougs (approved):
Very nice! Thanks!

PR #2094, opened 2021-07-01
Replace pass-by-value with pass-by-const-reference in Vose c'tor.

### LINKED ISSUE #1546: Topology/Vose constructor should take const reference argument
The [`Vose()`](https://github.com/nest/nest-simulator/blob/1d10715575ab49416c93c03ca698f67c37b6d96e/topology/vose.h#L70) constructor currently creates a copy of the distribution distribution it receives as argument. This should be replaced by call-by-reference.

heplesser:
Fixes #1546 by passing input distribution by const reference instead of copy.

### DIFF
--- a/nestkernel/vose.cpp
+++ b/nestkernel/vose.cpp
@@ -28,7 +28,7 @@
 
 namespace nest
 {
-Vose::Vose( std::vector< double > dist )
+Vose::Vose( const std::vector< double >& dist )
 {
   assert( not dist.empty() );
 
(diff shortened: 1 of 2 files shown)


stinebuu (approved):

nestkernel/vose.cpp
```diff
@@ -38,30 +38,34 @@ Vose::Vose( std::vector< double > dist )
 
   // We accept distributions that do not sum to 1.
   double sum = 0.0;
-  for ( std::vector< double >::iterator it = dist.begin(); it != dist.end(); ++it )
+  for ( std::vector< double >::const_iterator it = dist.begin(); it != dist.end(); ++it )
```
hakonsbm, line 41: Can this be replaced with `auto it`? Or maybe even a range-based loop?

nestkernel/vose.cpp
```diff
@@ -38,30 +38,34 @@ Vose::Vose( std::vector< double > dist )
...
   std::vector< BiasedCoin >::iterator small = dist_.begin();
   std::vector< BiasedCoin >::iterator large = dist_.end();
 
   index i = 0;
 
-  for ( std::vector< double >::iterator it = dist.begin(); it != dist.end(); ++it )
+  const double pivot = sum / n;
+  const double scale = 1.0 / pivot;
+
+  for ( std::vector< double >::const_iterator it = dist.begin(); it != dist.end(); ++it )
```
hakonsbm, line 55: Same here.

hakonsbm (changes requested):
Looks pretty good, I have just a suggestion about the for-loops.

Not related to this PR, but related to Vose: Now with the new RNG setup, is it possible (and desirable) to replace our custom `Vose` implementation with an [`std::discrete_distribution`](https://en.cppreference.com/w/cpp/numeric/random/discrete_distribution)?

nestkernel/vose.cpp
heplesser: In principle, yes, but in view of your comment above I'd want to minimise changes here.

nestkernel/vose.cpp
heplesser: See above.

heplesser:
> Looks pretty good, I have just a suggestion about the for-loops.
> 
> Not related to this PR, but related to Vose: Now with the new RNG setup, is it possible (and desirable) to replace our custom `Vose` implementation with an [`std::discrete_distribution`](https://en.cppreference.com/w/cpp/numeric/random/discrete_distribution)?

That sounds like a good idea. Could you create a follow-up issue and maybe give it a test to see if performance is ok?

heplesser:
Closing because it is obsoleted by #2141.

PR #2095, opened 2021-07-01
Implement time limit for duration of update cycles.

### LINKED ISSUE #2033: Automatically stop Simulate() if update time becomes too long

NEST simulations can slow down significantly if network activity becomes high, even more so if neuron models use adaptive-stepsize solvers. Currently, there is no way to interrupt this simulation except killing the entire NEST process.

The user sets an upper limit for the duration of a single update time slice (length: minimum delay), e.g.,

```python
nest.SetKernelStatus({'max_update_duration': 1.0})
```

to set a limit of 1 second wall-clock time for a single min-delay-length update step. If a step takes longer, simulation terminates.

**Additional context**
See also this mailing list discussion: https://www.nest-simulator.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/FQ6VDQ7EDSSKRIHMMRETIAIXK43A5GCZ/

heplesser:
This PR uses the `sw_simulate_` stopwatch to monitor the duration of individual update loops in the NEST kernel. If an update time exceeds the user-configurable `update_time_limit`, an exception is raised to terminate the simulation. This is mainly provided to avoid burning many CPU hours when batch jobs slow down due to unusually high activity. The default time limit is +inf. With this PR, kernel status also reports `{min,max}_update_time`, the durations (wall-clock, in seconds) of the shortest and longest update cycles observed to give users an orientation for choosing a limit.

Fixes #2033.


pynest/nest/lib/hl_api_simulation.py
```diff
@@ -375,6 +375,15 @@ def SetKernelStatus(params):
         Whether missed dictionary entries are treated as errors
     keep_source_table : bool
         Whether to keep source table after connection setup is complete
+    max_update_time: double, read only
```
jougs, line 378:
```suggestion
    min_update_time: double, read only
```

jougs (approved):
Other than my small suggestion, this looks good to me. Many thanks!

jarsi (approved):
Thanks for implementing this, it's a really nice feature to have. The code looks good to me, I wasn't aware of the `<limits>` header which seems to be pretty useful here. Thanks for the work!

pynest/nest/lib/hl_api_simulation.py
heplesser: Thanks for noticing, fixed.

jougs (approved):

PR #2098, opened 2021-07-02
Remove superfluous OpenMP barriers

### LINKED ISSUE #1441: Superfluous omp barriers in `gather_spike_data_`
In `gather_spike_data_` these two barriers seem to be unnecessary, as they appear directly before a `gather_completed_checker_.all_true()` that contains a barrier (https://github.com/nest/nest-simulator/blob/master/nestkernel/completed_checker.cpp#L56):

https://github.com/nest/nest-simulator/blob/master/nestkernel/event_delivery_manager.cpp#L381
https://github.com/nest/nest-simulator/blob/master/nestkernel/event_delivery_manager.cpp#L390

i think they could be removed. this is just to remind us of the issue, we should discuss this carefully.

btw, the same repetition seems to appear in `gather_target_data_`. at least we were consistent. ;)

@suku248 @heplesser

stinebuu:
This PR fixes #1441.

Tested on Piz Daint and JUSUF.

### DIFF
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -426,10 +426,6 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
     const bool deliver_completed = deliver_events_( tid, recv_buffer );
     gather_completed_checker_[ tid ].logical_and( deliver_completed );
 
-// Exit gather loop if all local threads and remote processes are
-// done.
-#pragma omp barrier
-
 #ifdef TIMER_DETAILED
     if ( tid == 0 )
     {
@@ -446,7 +442,6 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
         decrease_buffer_size_spike_data_ = false;
       }
     }
-#pragma omp barrier
 
   } // of while
 
@@ -706,7 +701,6 @@ EventDeliveryManager::gather_target_data( const thread tid )
     if ( gather_completed_checker_.all_true() )
     {
       set_complete_marker_target_data_( assigned_ranks, send_buffer_position );
-#pragma omp barrier
     }
     kernel().connection_manager.save_source_table_entry_point( tid );
 #pragma omp barrier
@@ -726,7 +720,6 @@ EventDeliveryManager::gather_target_data( const thread tid )
 
     const bool distribute_completed = distribute_target_data_buffers_( tid );
     gather_completed_checker_[ tid ].logical_and( distribute_completed );
-#pragma omp barrier
 
     // resize mpi buffers, if necessary and allowed
     if ( gather_completed_checker_.any_false() and kernel().mpi_manager.adaptive_target_buffers() )
@@ -736,7 +729,6 @@ EventDeliveryManager::gather_target_data( const thread tid )
         buffer_size_target_data_has_changed_ = kernel().mpi_manager.increase_buffer_size_target_data();
       }
     }
-#pragma omp barrier
   } // of while
 
   kernel().connection_manager.clear_source_table( tid );


stinebuu:
@suku248 @wschenck Would you have time to review this this week? Or should we try to find other reviewers?

heplesser:
@stinebuu I think we should wait for the original reviewers on this one, even if that should delay it to NEST 3.2.

jougs:
@suku248, @wschenck: ping!

jougs:
@suku248, @wschenck: could you please indicate if you are available for reviewing this or not? Thanks!

terhorstd:
Last call for passengers @suku248 and @wschenck …

nestkernel/event_delivery_manager.cpp
```diff
@@ -426,10 +426,6 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
     const bool deliver_completed = deliver_events_( tid, recv_buffer );
     gather_completed_checker_[ tid ].logical_and( deliver_completed );
 
-// Exit gather loop if all local threads and remote processes are
-// done.
-#pragma omp barrier
-
```
wschenck, line 432: With barrier, we measure for deliver_spike_data the runtime of the slowest thread; without barrier, we measure most likely the average runtime over all threads. I think, both variants are ok.

nestkernel/event_delivery_manager.cpp
```diff
@@ -446,7 +442,6 @@ EventDeliveryManager::gather_spike_data_( const thread tid,
         decrease_buffer_size_spike_data_ = false;
       }
     }
-#pragma omp barrier
```
wschenck, line 445: I agree that we don't need this barrier since gather_completed_checker_.any_false() also invokes a barrier.

nestkernel/event_delivery_manager.cpp
```diff
@@ -706,7 +701,6 @@ EventDeliveryManager::gather_target_data( const thread tid )
     if ( gather_completed_checker_.all_true() )
     {
       set_complete_marker_target_data_( assigned_ranks, send_buffer_position );
-#pragma omp barrier
     }
```
wschenck, line 704: Agree here: Superfluous barrier because of preceding invocation of gather_completed_checker_.all_true().

nestkernel/event_delivery_manager.cpp
```diff
@@ -726,7 +720,6 @@ EventDeliveryManager::gather_target_data( const thread tid )
 
     const bool distribute_completed = distribute_target_data_buffers_( tid );
     gather_completed_checker_[ tid ].logical_and( distribute_completed );
-#pragma omp barrier
```
wschenck, line 723: Agree here: Superfluous barrier because of following invocation of gather_completed_checker_.any_false().

nestkernel/event_delivery_manager.cpp
```diff
@@ -736,7 +729,6 @@ EventDeliveryManager::gather_target_data( const thread tid )
         buffer_size_target_data_has_changed_ = kernel().mpi_manager.increase_buffer_size_target_data();
       }
     }
-#pragma omp barrier
```
wschenck, line 739: Agree here: Superfluous barrier because of preceding invocation of gather_completed_checker_.any_false().

nestkernel/event_delivery_manager.cpp
wschenck, line 432: With barrier, we measure for deliver_spike_data the runtime of the slowest thread; without barrier, we measure most likely the average runtime over all threads. I think, both variants are ok.

wschenck (approved):
I think the removal of these barriers is sound and reasonable with minimal changes to the measured runtime of deliver_spike_data.

wschenck:
Sorry for not answering earlier, I completely missed that some of the large number of emails on the nest-simulator mailing list contained my name. I will change my email filter settings...

nestkernel/event_delivery_manager.cpp
heplesser: Code provisions made for benchmarking should not reduce speed under production conditions, so I think it is sensible to remove this barrier.

heplesser:
> Sorry for not answering earlier, I completely missed that some of the large number of emails on the nest-simulator mailing list contained my name. I will change the settings of my email program to avoid this in the future.

Hi @wschenck! Thank you for your review!

nestkernel/event_delivery_manager.cpp
terhorstd: Would this fall into the category of performance instrumentation that was discussed some time ago? We could make this barrier dependent on a cmake flag, but it might also confuse more than it helps. In any case, for sure this important difference in measurement must also be expressed in the documentation of the kernel timers!
@stinebuu, could you check if this difference is clear enough in the docs?

Would this be a problem for comparing benchmarks between NEST versions with/without this barrier?

jougs:
Honestly, I don't think that it does not get much better than it is with the review by @wschenck. Together with the tests that @stinebuu carried out already earlier on Piz Daint and JUSUF, I think we should once [the question by @terhorstd above](https://github.com/nest/nest-simulator/pull/2098#discussion_r788521118) is resolved.

@wschenck: how bad would the difference between old and new measurements be for the timers you mention above?

wschenck:
@jougs: I expect slight differences only for the timer deliver_spike_data. From a max operation (before) to a mean operation (after code change in PR) over threads. The resulting difference depends on the variance between threads. I *guess* that this variance is very small so that the measurement differences would be neglible.

jarsi (approved):
Thanks for your work @stinebuu. I agree, it appears to be save to remove all these barriers.

PR #2102, opened 2021-07-04
Extend data combination routine for NEST Server

jougs:
This PR refines and extends the data combination routine for NEST Server. @babsey: can you maybe check against NEST Desktop to make sure I did not break anything on the way? Thanks!

PR #2103, opened 2021-07-04
 Extend data combination routine for NEST Server

jougs:
This PR refines and extends the data combination routine for NEST Server. @babsey: can you maybe check against NEST Desktop to make sure I did not break anything on the way? Thanks!

[new version of #2102]


babsey (approved):
LGTM.

I can only review the serialization.
It works with NEST Desktop.

PR #2107, opened 2021-07-07
Declared security support for new 3.0 release

Helveg:


jougs:
@terhorstd: This is another one we need to watch out. I've added it to our README for the release tools.

PR #2010, opened 2021-04-13
Link PyNEST functions in documentaton

jessica-mitchell:
This PR provides links to the API functions in the documentation, specifically restructuredtext files in doc/userdoc.
There are still files to update in the nest-3 guides, but waiting for #1929  to be merged, so this is still incomplete.

Note: set() and get() methods are in both SynapseCollection and NodeCollection class, so the class name needs to be included in the link.  (This is mentioned in styleguide  #1986 )

Resolves #1878

PR #2114, opened 2021-07-24
Add kernel attributes and get/set functions to PyNEST module

jougs:
This PR adds the two new functions `get()` and `set()` on the PyNEST module level. The functions are (for now) just forwarding to `GetKernelStatus()` and `SetKernelStatus()`, respectively. We could consider deprecating the functions with the long names in the long run. In addition to adding the functions, this PR also adapts all tests and examples to actually use them.

@jessica-mitchell: what is the future plan for the NEST 2->3 transition guides? Am I right in assuming that all information contained therein is already present in other documentation pages and guides, and they should be removed/replaced by new ones for the upcoming NEST 3.0 -> 3.1 transition? Probably a single *What's new?* page is sufficient in the future due to the shorter release cycles and subsequently smaller amount of changes/breakage per release. 

As this PR already includes the changes proposed in #2108 and should thus be only merged after that. I will un-draft this one, when that happened.


jougs:
@pnbabu, @clinssen: I'm mentioning you, because we talked about this and you also might want to have a look.

jessica-mitchell:
@jougs I figured that the individual pages on the nest 3 features would be indexed (and moved to) the guides/ folder, but I hadn't thought much yet of when this should happen. I am not sure for how long we want to highlight the new features of NEST 3.  I don't think anything aside from the nest 3 index pages themselves should be removed.
I agree, a what's new section that highlights the major changes of the latest release would be best, along with a better organization of the guides so it's clear where to find the info you're looking for.

I have a question about the functions added in this PR: considering we have get() and set() in the NodeCollections and SynapseCollection classes, is having these function named the same in any way confusing or potentially misleading to what the expected input/output will be?

jougs:
@jessica-mitchell: I have created #2115 to restructure the transition documentation.

jessica-mitchell (commented):
In general this looks good, but I want to point out that in #2010, the syntax to link the API functions has been added. I suggest where you add `get()` in text to change it to the :py:func:\`.get\`  (I'm not sure the exact syntax). 
However, I'd still like to know if using the same names (get and set) as in NodeCollection and SynapseCollection could be in anyway confusing. This may be a non-issue I just want to know if anything needs to be said explicitly about the difference.

doc/userdoc/guides/built-in_timers.rst
```diff
@@ -4,11 +4,14 @@ Built-in timers
 Basic timers
 ------------
 
-Basic built-in timers keep track of the time NEST spent for network construction and actual simulation (propagation of the network state). These timers are active in all simulations with NEST, and the measured times can be queried using ``GetKernelStatus``. For example:
+Basic built-in timers keep track of the time NEST spent for network
+construction and actual simulation (propagation of the network
+state). These timers are active in all simulations with NEST, and the
+measured times can be queried using ``get()``. For example:
```
jougs, line 10:
```suggestion
measured times can be queried using :py:func:`.get`. For example:
```

doc/userdoc/guides/connection_management.rst
```diff
@@ -70,11 +70,11 @@ to :py:func:`.Connect` and will just take on the default value.
 
 After your connections are established, a quick sanity check is to
 look up the number of connections in the network, which can be easily
-done using :py:func:`.GetKernelStatus`:
+done using py:func:`.get`:
```
jougs, line 73:
```suggestion
done using :py:func:`.get`:
```

jougs (commented):
I've annotated most of the Python functions in 50dedabb8. Here, I'm suggesting the annotation of one more occurence and a fix for another one.

I don't see a potential for confusion with `get()` and `set()` of NodeCollections and SynapseCollections. I rather would think that this change makes everything more consistent and easier to understand.

jougs:
Since #2115 has meanwhile been merged, I would like to add a short section to the 3.1 release notes, before considering this ready. I'm thus converting to draft.

jougs:
I've added transition information in 319684485.

Helveg:
`.get("size", "models")` still feels a bit "querying data from IBM mainframe, please wait...". Don't they work better as attributes on the `nest` module? If the concept of a "kernel" has to continue existing in client facing code (I don't see why since it's global state and not an object) then replace all occurences of `nest.` below with `nest.kernel.`:

```python
print(nest.network_size)
nest.network_size = 120
> AttributeError: 'network_size' is read-only
```

If someone has to pay their scientists by the keystroke and/or batch `get` or `set` are absolute necessities:

```python
from nest import kernel as k

l, n = k.local_num_threads, k.network_size
k.set(local_num_threads=l, network_size=n)
```

jougs:
@Helveg: I totally hear you w.r.t. making this an attribute. Unfortunately this does not seem to be possible at the moment.

First, you can only overload `__getattr__` and not `__setattr__` on the module level. Second, I tried doing the former based on [so.org#2447353](https://stackoverflow.com/questions/2447353/getattr-on-a-module), which lead me to recursive hell and after a day or so I decided that the effort is not worth it and `get()`/`set()` are also nice. But maybe I'm also too old and just like the mainframe feeling more than I should ;-)

That said, the fact that @clinssen also already suggested the `.kernel` metaphor totally independently of you makes me wonder if that is a better way to go. I definitely won't have time for doing that in the near future, though.

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,28 @@
...
+on the level of the ``nest`` Python module. For now, they are just
+shortcuts for :py:func:`.GetKernelStatus` and
+:py:func:`.SetKernelStatus`. As the new functions might replace the
+older ones in a future version of NEST, we advise to use the shortcuts
+already now.
+
++--------------------------------------------+----------------------------------+
+| NEST 3.0                                   | NEST 3.1                         |
++============================================+==================================+
+| nest.GetKernelStatus('network_size')       | nest.get('network_size)          |
```
jougs, line 25:
```suggestion
| nest.GetKernelStatus('network_size')       | nest.get('network_size')         |
```

Helveg:
What about the `sys.modules` trick, it is endorsed by the Python devs:

```python
import types

class NestModule(types.ModuleType):
  def __getattr__(self, attr):
    pass

  def __setattr__(self, attr, value):
    pass

sys.modules[__name__] = NestModule(__name__)
```

Dynamically generating descriptors for the attributes would be preferred over `__g/setattr__` functions (due to subtle packaging and inspection pains when these are used on `ModuleType`s, but they can be overcome).

If you want I can whip up a draft and you can make up whether you like what you see.

jougs:
This is almost exactly what I tried.

However, I could not make our many relative imports work properly and the resolution of kernel attributes via the "module class" always lead to problems, when a function was calling another function of `nest` (recursion errors, lookup errors, younameit). Feel free to give it a go yourself. I'll happily review/accept a PR ;-)

jessica-mitchell:
@jougs The `:py:func:.get` doesn't produce a working link, I think because sphinx doesn't look in `__init__.py` for functions. As I understand it, this `get` just redirects to GetKernelStatus, so a possible solution could be to do :py:func:\`get() <.GetKernelStatus>\`  so it links to GetKernelStatus in the API docs. If that is undesirable,  you could remove the :py:func: completely so it no longer appears like a link.

jougs:
@jessica-mitchell: thanks for testing. I think we should wait for the solution, @Helveg offered. Then we can sort out linking and other documentation-related issues.

Helveg:
With the eye on documentation etc I think generating seperate _descriptors_ for each attribute would also offer an advantage as they can be docstringed.

@jougs What is your Python code dev cycle? The best I can come up with atm is to modify the installed code in place and then copy back those changes into the repo Python code :/ 

@jougs again: how can I best test for the errors you encountered?

jougs:
My development routine is pretty much the opposite of yours: I modify the code in the repository, call `make install`, and test the installed Python module.

The first problems I encountered prevented an import of `nest` due to cyclic imports. I guess if you can import and run the testsuite without any problems, you're good.

Helveg:
Is there a smaller target than `install` for just the Python code; On WSL I/O is rather slow so all the docs and tests take long to install?

Helveg:
@jougs Any idea how `pynestkernel.pyx` works and why it doesn't find the `Parameter` that is there at runtime? Is there some sort of build time analysis/cythonization going on?

```
  File "<stdin>", line 1, in <module>
AttributeError: module 'nest' has no attribute 'spatial'
>>> import nest.spatial
  File "<stdin>", line 1, in <module>
  File "/home/robin/.pyenv/versions/nest-dev/lib/python3.9/site-packages/nest/spatial/__init__.py", line 22, in <module>
    from .hl_api_spatial import *  # noqa
  File "/home/robin/.pyenv/versions/nest-dev/lib/python3.9/site-packages/nest/spatial/hl_api_spatial.py", line 86, in <module>
    distance = DistanceParameter()
  File "/home/robin/.pyenv/versions/nest-dev/lib/python3.9/site-packages/nest/spatial/hl_api_spatial.py", line 50, in __init__
    distance_parameter = CreateParameter('distance', {})
  File "/home/robin/.pyenv/versions/nest-dev/lib/python3.9/site-packages/nest/lib/hl_api_types.py", line 112, in CreateParameter
    return sli_func('CreateParameter', {parametertype: specs})
  File "/home/robin/.pyenv/versions/nest-dev/lib/python3.9/site-packages/nest/ll_api.py", line 165, in sli_func
    r = sli_pop()        # return value is an array
  File "pynestkernel.pyx", line 244, in pynestkernel.NESTEngine.pop
  File "pynestkernel.pyx", line 523, in pynestkernel.sli_datum_to_object
  File "pynestkernel.pyx", line 582, in pynestkernel.sli_array_to_object
  File "pynestkernel.pyx", line 541, in pynestkernel.sli_datum_to_object
AttributeError: module 'nest' has no attribute 'Parameter'
>>> nest.Parameter
<class 'nest.lib.hl_api_types.Parameter'>
```

jougs:
Phew... Maybe @hakonsbm knows something?

Helveg:
Somehow these `.pyx` modules managed to keep a reference to the original module object, instead of usingthe `sys.modules` object; import twilight zone. Fixed by unpacking the factorized module into the original module at the end of the file 🤷 

I have what I believe to be a working proof of concept now, do I open a new PR? We'd need to have a quick call because I'm unfamiliar with A) how to run the tests locally B) how to organize the code, I just put all required elements right there in `__init__.py` with lots of explanatory comments, that might need some cleanup :) .

jougs:
Thinking more about it, I might have gotten similar errors. IIRC, the reason was that after the conversion of the module to a class and the addition of some form of `getattr`, some attempts to resolve things in the `nest` module ended up in that function instead of finding the real thing within the module. I'm not sure this is your problem here, though.

Helveg:
@jougs ping, I expanded my previous post, I have a proposal ready

jougs:
I sent you an invite to a video call service via email

jougs:
@Helveg: I'm mostly through with updating all examples and tests. All in all, I'm really really happy with the new look :smile_cat:

Here's two points I'd like your opinion on:

* The properties `min_delay` and `max_delay` have to be set together. For now, I left the pertaining code locations like this:
  ```python
  nest.SetKernelStatus({"min_delay": 0.5, "max_delay": 2.5})
  ```
  As this seems to be the only co-dependent pair of properties, I'm inclined to just introduce a new attribute `delay_extrema` and change calls to one of
  ```python
  nest.delay_extrema = (0.5, 2.5)                            # short and concise
  nest.delay_extrema = {"min_delay": 0.5, "max_delay": 2.5}  # more explicit. better?
  ```
* For now, I changed calls like
  ```python
  nest.SetKernelStatus({
      'use_wfr': True,
      'wfr_comm_interval': 1.0,
      'wfr_tol': 0.0001,
      'wfr_max_iterations': 15,
      'wfr_interpolation_order': 3
  })
  ```
  to
  ```python
   nest.use_wfr = True
   nest.wfr_comm_interval = 1.0,
   nest.wfr_tol = 0.0001
   nest.wfr_max_iterations = 15
   nest.wfr_interpolation_order = 3
  ```
  and I am not unhappy with this despite the potential performance hit.

However, both of these points make me wonder if we should add a `set()` shortcut for `SetKernelStatus()` after all. What do you think about this? Or is there a magic way to collect the arguments of subsequent calls to the same function and only after the last call SLI's `SetKernelStatus` function? Any thoughts are most welcome!

Helveg:
afaik there's no magic for last calls in a module. You might be able to get there relatively cheaply with the `inspect` module since Python is an interpreted language and every module is parsed at runtime but it'll be hardly worth the effort and results may ultimately be disappointing and fragile. 

If it's `set(**kwargs)` and not `set({...})` then my Python spidey senses are fine with it 😅. Out of curiosity, what overheads should I imagine with the 2 innocent looking SLI lines executed by the descriptor?

The `delay_extrema` looks like a good solution

jougs:
@Helveg: It seems that 160 tests are failing right now. A preliminary analysis shows that it's lines like this that fail:
```python
n = nest.Create('iaf_psc_alpha', positions=nest.spatial.grid([2, 2]))
```
The exact error message and traceback (obtained locally) looks like this
```null
----> 1 n = nest.Create('iaf_psc_alpha', positions=nest.spatial.grid([2, 2]))

~/tmp/nest-simulator.build/install/lib/python3.9/site-packages/nest/ll_api.py in stack_checker_func(*args, **kwargs)
    226     def stack_checker_func(*args, **kwargs):
    227         if not get_debug():
--> 228             return f(*args, **kwargs)
    229         else:
    230             sr('count')

~/tmp/nest-simulator.build/install/lib/python3.9/site-packages/nest/lib/hl_api_nodes.py in Create(model, n, params, positions)
     85     if positions is not None:
     86         # We only accept positions as either a free object or a grid object.
---> 87         if not isinstance(positions, (nest.spatial.free, nest.spatial.grid)):
     88             raise TypeError('`positions` must be either a nest.spatial.free object or nest.spatial.grid object')
     89         layer_specs = {'elements': model}

AttributeError: module 'nest' has no attribute 'spatial'
```
I assume that this is also caused by `spatial` being looked up in the original module instead of in the replacement class. Could it be a solution to *also* `import spatial` there, so that the lookup works? Or do you see a way to inject the spatial module into it somehow.

Helveg:
It surprises me that a Python module like `ll_api` and `hl_api_nodes` can obtain a reference to a fully formed `nest` module before `__init__.py` is fully executed and our `NestModule` object is placed in `sys.modules`, without raising circular import errors. Perhaps it's to do with the `ast` module and `compile`/`exec` being used in `import_libs` that breaks the interpreter's trace of the imports?

jougs:
I have just tried several ways of importing `spatial` into `hl_api_nodes` to no avail and I'm afraid I have no clue here. This is all a bit too far out for me...

jougs:
Alright. My recent commits fix most of the problems by removing indirections via `hl_api` and adding an explicit lookup of `spatial` from the `NestModule` class. I think we can live with this.

Unfortunately setting co-dependent attributes (`min_delay`+`max_delay`, `tics_per_ms`+`resolution`, and `use_compressed_spikes`+`sort_connections_by_source`) does not work anymore via `SetKernelStatus` because that now also goes via the `KernelAttribute`s, which set each property individually. In the interest of backwards compatibility that function should behave the same as before.

@Helveg: I might only find time to look into this tonight. If you come up with a nice solution before that, I'd happily look at it!

Helveg:
I think the best solution would be to leave `Set`\`Get``KernelStatus` unchanged and to provide the `nest.` attributes as a new feature. ie, `cp` instead of `mv` the 2 lines of SLI code 😋

doc/userdoc/faqs/faqs.rst
```diff
@@ -50,15 +50,13 @@ Installation
...
-prefix for all data files can be set using the property ``data_prefix``
-of the root node by calling
-``nest.SetKernelStatus({"data_prefix": "prefix"})`` or setting the
+By default, the data files produced by NEST are stored in the
+directory from where NEST is called. The location can be changed by
+running ``nest.data_path = "/path/to/data"``. In scripts, this
+property can be set via the environment variable ``NEST_DATA_PATH``.
+Please note that the directory ``/path/to/data`` has to exist and
+will not be created. A common prefix for all data file names can be
+ set by running ``nest."data_prefix" = "prefix"`` or by setting the
```
terhorstd, line 59:
```suggestion
 set by running ``nest.data_prefix = "prefix"`` or by setting the
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -84,9 +84,8 @@ jobs:
            fetch-depth: 0
        - run: |
            echo "CHANGED_FILES<<EOF" >> $GITHUB_ENV
-           echo "CHANGED_FILES=$(git diff --name-only ${{ github.event.before }}..${{ github.event.after }})" >> $GITHUB_ENV
+           echo "$(git diff --name-only ${{ github.event.before }}..${{ github.event.after }})" >> $GITHUB_ENV
            echo 'EOF' >> $GITHUB_ENV
-
        - name: Dependencies MacOS
          if: runner.os =='macOS'
```
terhorstd, line 90:
```suggestion
         if: runner.os == 'macOS'
```

jougs:
@Helveg: I like your suggestion and would be happy if you could work on it.

pynest/nest/lib/hl_api_nodes.py
```diff
@@ -83,12 +83,13 @@ def Create(model, n=1, params=None, positions=None):
     model_deprecation_warning(model)
 
     if positions is not None:
+        spatial = getattr(nest.NestModule, "spatial")
```
Helveg, line 86: I'm surprised this works, I'd expect this to return `<property object 'spatial' at 0x...>`

pynest/nest/lib/hl_api_nodes.py
```diff
@@ -83,12 +83,13 @@ def Create(model, n=1, params=None, positions=None):
...
         # We only accept positions as either a free object or a grid object.
-        if not isinstance(positions, (nest.spatial.free, nest.spatial.grid)):
-            raise TypeError('`positions` must be either a nest.spatial.free object or nest.spatial.grid object')
+        if not isinstance(positions, (spatial.free, spatial.grid)):
+            raise TypeError('`positions` must be either a spatial.free object or spatial.grid object')
```
Helveg, line 89: I would leave `nest.spatial.` in the error message since this is the correct module path for users.

Helveg (changes requested):
I'll look into the `spatial` thing

pynest/nest/lib/hl_api_nodes.py
jougs: This was one of the more desperate tries and I was also surprised. If you find something better, go ahead, but I'm also fine with this. Maybe it should get a comment if it stays...

pynest/nest/lib/hl_api_nodes.py
jougs: Agreed.
```suggestion
            raise TypeError('`positions` must be either a nest.spatial.free or a nest.spatial.grid object')
```

pynest/nest/lib/hl_api_nodes.py
Helveg: Fine with it, You could maybe link to #2134 in the comment.

jessica-mitchell:
@Helveg Here is the stack trace: 

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/cmd/build.py", line 280, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/application.py", line 343, in build
    self.builder.build_update()
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 293, in build_update
    self.build(to_build,
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 307, in build
    updated_docnames = set(self.read())
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 414, in read
    self._read_serial(docnames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 435, in _read_serial
    self.read_doc(docname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 475, in read_doc
    doctree = read_doc(self.app, self.env, self.env.doc2path(docname))
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/io.py", line 188, in read_doc
    pub.publish()
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/core.py", line 217, in publish
    self.document = self.reader.read(self.source, self.parser,
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/io.py", line 108, in read
    self.parse()
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/readers/__init__.py", line 77, in parse
    self.parser.parse(self.input, document)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/parsers.py", line 100, in parse
    self.statemachine.run(inputlines, document, inliner=self.inliner)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 170, in run
    results = StateMachineWS.run(self, input_lines, input_offset,
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/statemachine.py", line 241, in run
    context, next_state, result = self.check_line(
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/statemachine.py", line 459, in check_line
    return method(match, context, next_state)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 2769, in underline
    self.section(title, source, style, lineno - 1, messages)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 327, in section
    self.new_subsection(title, lineno, messages)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 393, in new_subsection
    newabsoffset = self.nested_parse(
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 281, in nested_parse
    state_machine.run(block, input_offset, memo=self.memo,
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 196, in run
    results = StateMachineWS.run(self, input_lines, input_offset)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/statemachine.py", line 241, in run
    context, next_state, result = self.check_line(
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/statemachine.py", line 459, in check_line
    return method(match, context, next_state)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 2342, in explicit_markup
    nodelist, blank_finish = self.explicit_construct(match)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 2354, in explicit_construct
    return method(self, expmatch)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 2096, in directive
    return self.run_directive(
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/docutils/parsers/rst/states.py", line 2146, in run_directive
    result = directive_instance.run()
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/ext/autodoc/directive.py", line 162, in run
    documenter.generate(more_content=self.content)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 979, in generate
    self.document_members(all_members)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 860, in document_members
    documenter.generate(
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 1769, in generate
    return super().generate(more_content=more_content,
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 969, in generate
    self.add_directive_header(sig)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-test/envs/add-init-pynest-docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 1649, in add_directive_header
    if hasattr(self.object, '__orig_bases__') and len(self.object.__orig_bases__):
TypeError: object of type 'type' has no len()
```

To run the documentation locally, you need the requirements for NEST documentation found in:` extras/conda-nest-simulator-dev.yml`. You can then run `make html` from the build directory after `cmake` and open browser to see output. The configuration file for Sphinx is in `/doc/userdoc/conf.py `.

We can also discuss at hackathon this week

thanks!

doc/userdoc/guides/running_simulations.rst
```diff
@@ -87,11 +88,13 @@ the easiest way to assert its integrity is to not change its size after
...
 to ``Simulate()``, the required boundaries of delays can be set manually
-using :py:func:`.SetKernelStatus`:
+using :py:func:`.set`:
 
 ::
 
-    SetKernelStatus({"min_delay": 0.5, "max_delay": 2.5})
+    # For co-dependent properties, we have to use `set()`
+    # instead of kernel attributes
+    nest.set(min_delay=0.5, may_delay=2.5)
```
jougs, line 97:
```suggestion
    nest.set(min_delay=0.5, max_delay=2.5)
```

doc/userdoc/guides/simulations_with_gap_junctions.rst
```diff
@@ -107,11 +107,11 @@ general the following parameters can be set via kernel parameters:
 
 .. code:: python
 
-    nest.SetKernelStatus({'use_wfr': True,
-                          'wfr_comm_interval': 1.0,
-                          'wfr_tol': 0.0001,
-                          'wfr_max_iterations': 15,
-                          'wfr_interpolation_order': 3})
+    nest.use_wfr = True
+    nest.wfr_comm_interval = 1.0,
```
jougs, line 111:
```suggestion
    nest.wfr_comm_interval = 1.0
```

Helveg:
From the NEST hackathon:

* [ ] Switch to `dataclasses`:
  * No need to handroll `KernelAttribute` class from scratch
  * Improved typing
  * Improved RST docstring
* [ ] Correct RST parsing

Helveg:
@terhorstd It seems explicitly forbidden to alter `Field`s so a `dataclass` seems unfit to hold custom descriptors:

> Field objects describe each defined field. These objects are created internally, and are returned by the fields() module-level method (see below). Users should never instantiate a Field object directly. Its documented attributes are:

We either have to monkey patch `dataclasses.Field` so that the `field` function returns `KernelAttribute`s or whatever has the right `___get__` and `__set__` functionality, or do what `field()` does to `Field` objects ourselves on the `KernelAttribute`s that will inherit from `Field`. Either approach is undesirable and ultimately fragile.

 There is no mention of how to deal with descriptors in the docs or online either. It might not be the intended scope of `dataclasses` or I might be misunderstanding. I'll leave it up to you if you want to take a look into it and figure out a `dataclass`y way. The `KernelAttribute` class isn't unPythonic at all, it is a standard descriptor implementing `__get__` and `__set__` that defer to SLI. 
 
 Using Python's typing is fancy, but sphinx parses the typing annotations and a well formatted docstring into the same HTML code ultimately.

I see how it could be an improvement but it's currently too big of a task for me.

jougs:
@Helveg: thanks for your analysis. I would be totally fine with the current implementation, if https://github.com/jougs/nest-simulator/pull/58 would just fix the creation of the restructured text for the docstrings.

heplesser (commented):
Very nice work! I added a few comments on details. Have I overlooked the code that turns the kernel attributes into RST for the documentation, or is it not here yet?

One suggestion for the documentation: I think it would be a good idea to give one example of "when setting multiple parameters, closely related params, you can use three approaches

```python
nest.x = 1
nest.y = 2

nest.set({'x': 1, 'y': 2})

params = {'x': 1, 'y': 2} 
nest.set(**params)
```

pynest/nest/__init__.py
```diff
@@ -36,31 +36,401 @@
...
+
+    return property(lazy_loader)
+
+
+class NestModule(types.ModuleType):
+    """
+    A module class for the `nest` root module to control the dynamic generation
+    of module level attributes such as the KernelAttributes and lazy loading
+    some submodules.
+    """
```
jougs, line 91:
```suggestion

    """
```

pynest/nest/__init__.py
```diff
@@ -36,31 +36,401 @@
...
+        if len(args) == 1:
+            return self.GetKernelStatus(args[0])
+        else:
+            return self.GetKernelStatus(args)
+
+    def __dir__(self):
+        return list(set(vars(self).keys()) | set(self.__all__))
+
+
+# Instantiate a NestModule
```
jougs, line 409:
```suggestion
# Instantiate a NestModule to replace the nest Python module. Based on
# https://mail.python.org/pipermail/python-ideas/2012-May/014969.html
```

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -198,224 +198,35 @@ def ResetKernel():
 
 @check_stack
 def SetKernelStatus(params):
-    r"""Set parameters for the simulation kernel.
+    """Set parameters for the simulation kernel.
+
+    See the nest module attribute documentation for a valid list of params.
```
jougs, line 203:
```suggestion
    See the :ref:`PyNEST API documentation <kernel-attributes>` for a valid list of params.
```

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -198,224 +198,35 @@ def ResetKernel():
...
     GetKernelStatus
 
     """
-    # Resolve if missing entries should raise errors
-    raise_errors = params.get('dict_miss_is_error')
-    if raise_errors is None:
-        raise_errors = GetKernelStatus('dict_miss_is_error')
-
-    # Check validity of passed parameters
+    import nest    # noqa
```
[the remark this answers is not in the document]
jougs: If I understand it correctly, the `import` only makes the name available and the position of the line is chosen as to restrict the availability to the function's scope. The code below is only executed when the module is fully initialized, so this always works.

@Helveg?

testsuite/pytests/test_sp/test_disconnect.py
```diff
@@ -72,22 +69,17 @@ def test_synapse_deletion_one_to_one_no_sp(self):
...
                 nest.Connect(neurons[0], neurons[2], "one_to_one", syn_dict)
                 nest.Connect(neurons[1], neurons[3], "one_to_one", syn_dict)
 
                 # Delete existent connection
-                conns = nest.GetConnections(
-                    neurons[0], neurons[2], syn_model)
-                if mpi_test:
+                conns = nest.GetConnections(neurons[0], neurons[2], syn_model)
+                if test_with_mpi:
                     print("rim with mpi")
```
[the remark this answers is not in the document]
jougs: In #2127 you changed "rim" to "Run" and I honestly did not bother much, as I'm anyway expecting that to be merged first and me to get some conflict in this file. I thus consider this to be resolved.

testsuite/pytests/test_sp/test_disconnect.py
```diff
@@ -97,15 +89,15 @@ def test_synapse_deletion_one_to_one_no_sp(self):
...
 
-                # Assert that one can not delete a non existent connection
+                # Assert that one cannot delete a non existent connection
                 conns1 = nest.GetConnections(
                     neurons[:1], neurons[1:2], syn_model)
-                if mpi_test:
+                if test_with_mpi:
                     conns1 = self.comm.allgather(conns1.get('source'))
                     conns1 = list(filter(None, conns1))
                 assert len(conns1) == 0
```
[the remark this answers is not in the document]
jougs: I honestly think we should rather remove this test instead of trying to fix it...

testsuite/pytests/test_sp/test_disconnect_multiple.py
```diff
@@ -200,9 +200,7 @@ def test_single_synapse_deletion_sp(self):
                     'pre_synaptic_element': 'SE1',
                     'post_synaptic_element': 'SE2'
                 }
-                # nest.SetKernelStatus(
-                #   {'structural_plasticity_synapses': {'syn1': syn_dict}}
-                # )
+                # nest.structural_plasticity_synapses = {'syn1': syn_dict}
```
jougs, line 203:
```suggestion
```

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+
+
+Access to kernel properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With NEST 3.1 it is easier than ever to access properties of the NEST
+simulation kernel. In addition to the traditional access methods
+:py:func:`.GetKernelStatus` and :py:func:`.SetKernelStatus`,
+properties can now also be set and retrieved via *direct kernel
+attributes*.
```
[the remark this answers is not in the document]
jougs: This is the transition guide from 3.0 to 3.1. I thought that the sentence you deleted would help people who already know the old way to understand immediately what kernel attributes are. Also here: @jessica-mitchell?

jougs:
@heplesser: The code for the doc generation is [here](https://github.com/nest/nest-simulator/pull/2114/files#diff-ad45fe539c61468c5813685f7f54e9bebdeb7d3fd172493d518fef7a2c5c0296R286-R297)

doc/userdoc/release_notes/v3.1/index.rst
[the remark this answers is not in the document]
heplesser: Then maybe "Instead of the traditional ..." rather than "In addition to the traditional ..."?

testsuite/pytests/test_sp/test_disconnect.py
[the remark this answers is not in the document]
heplesser: A test for disconnect seems a sensible test to me. Is it tested anywhere else?

pynest/nest/ll_api.py
```diff
@@ -271,6 +272,54 @@ def check_stack(thing):
...
+            + ("." if default is None else f", defaults to ``{default}``.")
+            + ("\n\n" if readonly or localonly else "") + ", ".join(
+                c for c in (
+                    readonly and "**read only**",
+                    localonly and "**local only**",
+                )
+                if c
+            )
+            + f"\n\n:type: {typehint}"
+        )
```
heplesser, line 295: It took me quite some time to understand this code as it mixes quite a number of constructs (and this type of class-building is not a part of Python I have much experience with ...; I even learned something new about `and` in Python from this code). I tried to rewrite the code a little, including renaming `doc` to `description`, with the aim to make the code more expressive:

```suggestion
        self.__doc__ = "".join((
            description,
            "." if default is None else f", defaults to ``{default}``.",
            "\n\n" if readonly or localonly else "",
            ", ".join(c for c in (readonly and "**read only**",
                                          localonly and "**local only**") if c),
            f"\n\n:type: {typehint}"))
```

I don't like the format of the docstring we get here, with a "loose" "read only, local only" line, and the default somewhat hidden in the textual description, which may be long. Ideally, I would include default/ro/lo information in the title line, e.g.,

```
resolution [default: 0.1]

<description>

biologial_time [read-only]

<description>
```

If that is not possible in Sphinx, I'd suggest to put the line with default/ro/lo information directly below the title.

pynest/nest/__init__.py
```diff
@@ -37,30 +37,353 @@
...
```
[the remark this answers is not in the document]
Helveg: Which lines? 359?

pynest/nest/lib/hl_api_simulation.py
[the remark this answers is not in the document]
Helveg: That's correct, we need the nest module to be fully initialized in order to access the `_kernel_attr_names` and `_readonly_kernel_attrs` and `hl_api_simulation` is imported via `hl_api` during initialization so we can't put the import on the module level

pynest/nest/ll_api.py
jougs: Could we maybe postpone the formatting discussion to a follow-up PR? The whole PyNEST API page needs a serious overhaul anyway and having the information like it is now is already a huge improvement over what we had previously.

This PR already contains a lot of changes on the side and we want to release in two weeks.

pynest/nest/__init__.py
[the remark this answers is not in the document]
heplesser: I meant the entire "Instantiate a NestModule" section that follows here.

pynest/nest/lib/hl_api_simulation.py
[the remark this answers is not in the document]
heplesser: Could you add a comment here explaining why nest is imported here, essentially what you wrote in your comment above? When we look at the code three years from now in wonder this discussion here will be far away.

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+.. contents:: On this page you'll find
+   :local:
+   :depth: 1
+
+
+Access to kernel properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With NEST 3.1 it is easier than ever to access properties of the NEST
+simulation kernel. In addition to the traditional access methods
```
jougs, line 24:
```suggestion
simulation kernel. Instead of using the traditional access methods
```

pynest/nest/lib/hl_api_simulation.py
```diff
@@ -198,224 +198,35 @@ def ResetKernel():
...
```
jougs, line 217:
```suggestion
    # We need the nest module to be fully initialized in order to access the _kernel_attr_names
    # and _readonly_kernel_attrs. As hl_api_simulation is imported via hl_api during initialization,
    # we can't put the import on the module level, but have to have it on the function level.
    import nest    # noqa
```

testsuite/pytests/test_sp/test_disconnect.py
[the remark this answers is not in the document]
jougs: After some more off-line discussions, I have now implemented your suggestion.

pynest/nest/ll_api.py
jougs: :+1:

heplesser: @jougs While I agree that we should not discuss formatting of the docstring here, I would still like you to consider my code suggestion above. I think the `join()` approach is clearer than the long sequence of string additions.

heplesser (commented):
Almost good to go from my side, but see my comment on the `__doc__` construction.

jougs:
@heplesser: sorry for the oversight. I have just pushed some new commits including your suggestion.

heplesser (approved):
Great work :+1:

doc/userdoc/guides/running_simulations.rst
```diff
@@ -157,18 +160,19 @@ using
...
 ::
 
-    nest.SetKernelStatus({'rng_type': 'mt19937'})
-    nest.SetKernelStatus({'rng_type': 'mt19937', 'rng_seed': 12234})
+    nest.rng_type = 'mt19937'
+    nest.rng_seed = 12234
 
-In the first case, the ``rng_seed`` set previously (or the default seed) is used,
-otherwise the seed specified.
+In the example, the generator is initialized with the defalt seed (or with a
```
jessica-mitchell, line 173:
```suggestion
In the example, the generator is initialized with the default seed (or with a
```

doc/userdoc/guides/running_simulations.rst
```diff
@@ -157,18 +160,19 @@ using
...
+previously defined seed, if one was set). The second line re-sets the seed to
```
jessica-mitchell, line 174:
```suggestion
previously defined seed, if one was set). The second line sets the seed to
```
I think the 're' might be a bit confusing

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+
+.. contents:: On this page you'll find
+   :local:
+   :depth: 1
+
+
+Access to kernel properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With NEST 3.1 it is easier than ever to access properties of the NEST
```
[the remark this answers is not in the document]
jessica-mitchell: @jougs @heplesser  Generally speaking the marketing language should be avoided in the documentation, but this is also the release information so I see it as a bit of an advertisement for the newest version. But maybe we should save this language for the press releases and user group announcements.

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+simulation kernel. Instead of using the traditional access methods
+:py:func:`.GetKernelStatus` and :py:func:`.SetKernelStatus`,
+properties can now also be set and retrieved via *direct kernel
+attributes*.
+
+Where you previously had ``nest.SetKernelStatus({"resolution": 0.2})``
+in your simulation script, you can now just write ``nest.resolution =
+0.5``. Kernel attributes are first-class citizens of the NEST Python
```
jessica-mitchell, line 31: Is "first-class citizen" a term that's used in Python?

It seems a bit sketchy to me.

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+  |     })                                          |                                             |
+  +-------------------------------------------------+---------------------------------------------+
+
+.. admonition:: Deprecation info
+
+      The use of the access functions :py:func:`.GetKernelStatus` and
+      :py:func:`.SetKernelStatus` is now deprecated and they will be
+      removed in a future version of NEST. To avoid porting trouble
+      later on, we suggest you switch to using the new shortcuts
+      already now.
```
jessica-mitchell, line 62:
```suggestion
      now.
```

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+  |     nest.SetKernelStatus({                      |                                             |
+  |         'min_delay': 0.2,                       |     nest.set(min_delay=0.2, max_delay=2.0)  |
+  |         'max_delay': 2.0                        |                                             |
+  |     })                                          |                                             |
+  +-------------------------------------------------+---------------------------------------------+
+
+.. admonition:: Deprecation info
+
+      The use of the access functions :py:func:`.GetKernelStatus` and
+      :py:func:`.SetKernelStatus` is now deprecated and they will be
```
jessica-mitchell, line 59:
```suggestion
      :py:func:`.SetKernelStatus` is now deprecated, and they will be
```

jessica-mitchell (changes requested):
@jougs I have a few minor comments in line.
Also, when I built this branch on RtD (which works now for me :tada: ) I don't get list of attributes, instead I only get this:
image

Does this work for you?

jougs (commented):
@jessica-mitchell: I've set up a [RtD build](https://nest-simulator-jougs.readthedocs.io/en/getset/ref_material/pynest_apis.html) for the branch and can reproduce the problem you reported

jougs:
Hmmm... weird... Now I see the same problem with the documentation also locally.

I assume this has something to do with the changes proposed by @heplesser (bcac4d81130bd32e36d18469233487bfa9847de1, ce978251fdd3c57742af345c75f00033f6d8e31e). Checking now...

doc/userdoc/release_notes/v3.1/index.rst
Helveg: Some might even say you're not a real Python user until you've complained not being a first-class citizen somewhere  😂 

Maybe:

```
Where you previously had ``nest.SetKernelStatus({"resolution": 0.2})``
in your simulation script, you can now write ``nest.resolution = 0.5``. 
Kernel attributes are documented and can make use of tab completion.
```

jougs:
I'm giving up for now. It seems that the documentation generation only works once NEST is actually installed. Cf.
```bash
rm -rf install; make install html install
```
and
```bash
rm -rf install; make html install
```
As discussed with @terhorstd and @jessica-mitchell , this probably has to do with the mocking in [`doc/userdoc/conf.py`](https://github.com/nest/nest-simulator/pull/2114/files#diff-65301f43a60ab906deea399b3f44bdc08580509ee8fe30d5ce000637c03c51f0R82-R88) and I don't understand enough of this to fix it. And I wonder why [this instance](https://nest-simulator-helveg.readthedocs.io/en/getset/ref_material/pynest_apis.html) (based on the code from https://github.com/jougs/nest-simulator/pull/58) built so nicely.

@Helveg: if you have any idea, that would be greatly appreciated. If not, I can try harder on Monday.

Helveg:
The project settings on RtD perhaps? These are mine, no redirects, translations, subprojects, only github integration, no env vars, no  automation rules.

image

For the mocking: when trying different RST commands I noticed in some far fetched attempts that it would render `MagicMock.__getitem__` when I was asking it to render `nest.*` things that didn't exist, meaning that `nest` must've been a mocked module. After explicitly importing it in `conf.py` this problem was alleviated, I then add the `NestModule` to the object to reference it from the docs, but `nest.__class__` might work just as well.

I did also see the `alias <classname>` rendering pass by a couple of times locally. 

Have you tried wiping the build environment on RtD?

image

jougs:
The non-building of kernel attribute documentation seems to have been a side-effect of #2124, which introduced a version check based on CMake variables and renamed PyNEST's `__init__.py` to a CMake template `__init__.py.in`. As CMake is not run on Read the Docs, the template replacement was never performed there and thus no documentation from `__init__.py` could have been extracted.

In my recent commits, I fixed the problem by renaming the initializer file back and moving the version checking mechanism to a [new optional Python file](https://github.com/nest/nest-simulator/pull/2114/commits/046c0baf2484769b815a42cc73bf4fb515a5a2ba#diff-04222c37b159f67985c83860f7209597122f24afce809c42f65fbbc33335909e).

With these changes in place, the docuemtation builds again nicely on my RtD space and also locally using Python 3.9.5 and Sphinx 3.5.4.

Helveg:
Awh I feel bad, I had commented about that on that PR but not realized the changes made it in here already :/

jougs:
@Helveg: don't feel bad, just press the `approve` button ;-)

Helveg (approved):
😜

doc/userdoc/release_notes/v3.1/index.rst
jessica-mitchell: @jougs My suggestion would simply be to cut it so the sentence would read something like "Kernel attributes now ome with their own docstrings and even tab-completion works for them!"

jessica-mitchell (approved):
Just tested this and it also works for me on RtD. Thanks! I'll approve as the last comment is fairly minor.

doc/userdoc/release_notes/v3.1/index.rst
```diff
@@ -0,0 +1,62 @@
...
+simulation kernel. Instead of using the traditional access methods
+:py:func:`.GetKernelStatus` and :py:func:`.SetKernelStatus`,
+properties can now also be set and retrieved via *direct kernel
+attributes*.
+
+Where you previously had ``nest.SetKernelStatus({"resolution": 0.2})``
+in your simulation script, you can now just write ``nest.resolution =
+0.5``. Kernel attributes are first-class citizens of the NEST Python
+module, which means that they come with their own docstrings and even
+tab-completion works for them!
```
jougs, line 33:
```suggestion
0.5``. Kernel attributes now come with their own docstrings and even
tab-completion works for them!
```

jougs:
@terhorstd: given the three positive reviews for the PR, I'm merging without your final approval. If you have any additional comments, please open a follow-up issue.

PR #2118, opened 2021-08-09
Fix mismatched parenthesis in setup.py.in

hakonsbm:


heplesser (approved):

heplesser:
Merging based on single review since this is a minimal but important syntax fix.

PR #2122, opened 2021-08-20
Remove outdated comment on Python dynamic library flags

Helveg:
With only recent Python versions supported this comment loses its added value


niltonlk:
in `pynest/nest/__init__.py` there is a line checking for Python 2, and the message states that Python should be >= 3.6, however, the minimum requirement is already 3.8. Maybe this version check could also be removed...

heplesser (approved):

heplesser:
@niltonlk Thanks for your comment. Would you mind creating a PR for that? I'd prefer to keep PRs like this small and collect mulitple (if related) fixes.

@Helveg Thanks!

I will merge this minimal PR concerning an outdated comment without second review.

PR #2124, opened 2021-08-21
Update Python version check in __init__.py to check for minor version, and to report correct minimum required version (>= 3.8)

niltonlk:
As commented on #2122, remove Python version check in `__init__.py` as minimum requirement is already enforced to be >= 3.8 (#1954)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -350,6 +350,11 @@ configure_file(
     "${PROJECT_BINARY_DIR}/doc/fulldoc.conf" @ONLY
 )
 
+configure_file(
+    "${PROJECT_SOURCE_DIR}/pynest/nest/__init__.py.in"
+    "${PROJECT_BINARY_DIR}/pynest/nest/__init__.py" @ONLY
+)
+
 
 ################################################################################
 ##################            Install Extra Files             ##################
--- a/pynest/CMakeLists.txt
+++ b/pynest/CMakeLists.txt
@@ -74,8 +74,9 @@ if ( HAVE_PYTHON )
       -D_IS_PYNEST
       )
 
-  install(DIRECTORY  nest/
+  install(DIRECTORY  nest/ ${PROJECT_BINARY_DIR}/pynest/nest/
       DESTINATION ${CMAKE_INSTALL_PREFIX}/${PYEXECDIR}/nest
+      PATTERN "__init__.py.in" EXCLUDE
   )
   install( TARGETS pynestkernel DESTINATION ${PYEXECDIR}/nest/ )
   install( CODE "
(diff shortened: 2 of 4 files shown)


niltonlk:
When runtime version is older than compiletime, besides the RuntimeWarning, import of NEST also fails and execution aborts.
In the opposite situation, when compiletime version is older than runtime, the RuntimeWarning is also displayed but import succeeds.
I have updated the check for exact match with the following Exception message:
Python runtime version does not match 'nest' complietime version. Please use Python @Python_VERSION_MAJOR@.@Python_VERSION_MINOR@.

Helveg (approved):
Very good! This should help people detect when they are loading incorrect NEST/Python versions :)

Unless there are any devs that want to explicitly make loading NEST with different versions of Python a possibility, but then the aforementioned ABI flag should be checked and documented (perhaps it already is).

clinssen (approved):
Great, thank you!

Helveg:
@clinssen won't the docs break if the source dir doesn't contain `.py` files but `.py.in` files, from `doc/userdoc/conf.py`:

```python
source_dir = os.environ.get('NESTSRCDIR', False)
if source_dir:
    source_dir = Path(source_dir)
else:
    source_dir = Path(__file__).resolve().parent.parent.parent.resolve()


if os.environ.get("READTHEDOCS") == "True":
    doc_build_dir = source_dir / "doc/userdoc"
else:
    doc_build_dir = Path(os.environ["OLDPWD"]) / "doc/userdoc"

print("doc_build_dir", str(doc_build_dir))
print("source_dir", str(source_dir))

source_suffix = '.rst'
master_doc = 'contents'

# Create the mockfile for extracting the PyNEST

excfile = source_dir / "pynest/nest/lib/hl_api_exceptions.py"
infile = source_dir / "pynest/pynestkernel.pyx"
outfile = doc_build_dir / "pynestkernel_mock.py"

sys.path.insert(0, str(source_dir))
sys.path.insert(0, str(source_dir / 'doc'))
sys.path.insert(0, str(source_dir / 'pynest'))
sys.path.insert(0, str(source_dir / 'pynest/nest'))
sys.path.insert(0, str(doc_build_dir))
```

niltonlk:
I am not sure about the doc part. How can I check if there are problems with this PR regarding @Helveg last comment? At least `make doc` and `make fulldoc` also resulted in no errors.

Helveg:
There currently should not be a problem, but in an upcoming PR (#2114) exactly the `__init__.py` that you modified is undergoing changes and will be used for the doc, we might notice problems then, so let's wait for it to be merged and I will check the impact on the docs for you!

niltonlk:
Got it! Thank you!

Issue #2125, opened 2021-08-21
Assertion fail in set_connections_have_changed in test_synapse_deletion_one_to_one_no_sp 


niltonlk:
**Describe the bug**
Testsuilt results table does not report failures or erros, however, test_synapse_deletion_one_to_one_no_sp reports assertion failure in current master (NEST version master@8c16625).
The same error was observed in CentOS 7.9.2009 with GCC 8.3.0, OpenMPI 3.1.4, Python 3.8.10, Boost 1.71 and gsl 2.6 as well as in Fedora 34 with GCC 11.2.1, OpenMPI 4.1.0, Python 3.9.6, Boost 1.75 and gsl 2.6.
Besides the difference in package versions, the steps to reproduce are the same.
Error message is as follow:
- CentOS 7.9.2009
```
test_synapse_deletion_one_to_one_no_sp (test_disconnect.TestDisconnectSingle) ... python3.8: /opt/ohpc/pub/share/NEST/nest-master/nestkernel/connection_manager.cpp:1606: void nest::ConnectionManager::set_connections_have_changed(): Assertion `kernel().vp_manager.get_thread_id() == 0' failed.
```
- Fedora 34
```
test_synapse_deletion_one_to_one_no_sp (test_disconnect.TestDisconnectSingle) ... python3.9: /home/nilton/Workspace/git/nest-simulator/src/nestkernel/connection_manager.cpp:1606: void nest::ConnectionManager::set_connections_have_changed(): Assertion `kernel().vp_manager.get_thread_id() == 0' failed.

```

**To Reproduce**
1. Build latest master (master@8c16625) with MPI enabled:
```
$ cmake -DCMAKE_INSTALL_PREFIX:PATH=/opt/ohpc/pub/libs/gnu8/openmpi3/nest/3.0-dev -Dwith-mpi=ON ../nest-master
$ make
$ sudo make install
$ make installcheck
```
2. NEST Configuration Summary
- CentOS 7.9.2009
```
Build type          : 
Target System       : Linux
Cross Compiling     : FALSE
C compiler          : GNU 8.3.0 (/opt/ohpc/pub/compiler/gcc/8.3.0/bin/gcc)
C compiler flags    :  -O2 -Wall -fopenmp   -pthread -fdiagnostics-color=auto
C++ compiler        : GNU 8.3.0 (/opt/ohpc/pub/compiler/gcc/8.3.0/bin/g++)
[...]

PR #2127, opened 2021-08-21
Run Python tests using Pytest, reorganize MPI-based Python tests and improve error reporting.

### LINKED ISSUE #2126: NEST testsuite fails to report crashing test cases
The NEST test suite will not detect Python test cases which lead to a segmentation fault, see #2125 and [also here in #2121](https://github.com/nest/nest-simulator/pull/2121#issuecomment-902520234).

This happens because the return value of `nosetests` is never inspected and now XML-output is written if `nosetests` terminates due to a segmentation fault or failed C++ assertion in the NEST kernel.

heplesser:
This PR implements brute-force detection and reporting of crashes during the PyNEST phase of the testsuite. This fixes #2126 (with bearing on #2125 and #2121).

Briefly, the exit code of `nosetests` is read out and if > 1, and empty XML-file is created for that phase and the summary will report the phase as crashed. Unfortunately, I have not been able to think of a way to find out which test crashed. We know one from #2125 and hopefully when that is fixed there won't be any others.


heplesser:
Strangely, in Github Actions, the test reported failed in #2125 passes as shown [in the log](https://github.com/nest/nest-simulator/pull/2127/checks?check_run_id=3390900258#step:16:33848). It fails on my computer exactly as described in #2125.

heplesser:
> Strangely, in Github Actions, the test reported failed in #2125 passes as shown [in the log](https://github.com/nest/nest-simulator/pull/2127/checks?check_run_id=3390900258#step:16:33848). It fails on my computer exactly as described in #2125.

The reason for not failing on Github seems to be that `mpi4py` is not imported properly on Github and the corresponding test skipped. A further indication for this is that the [issue-1703.py test](https://github.com/nest/nest-simulator/runs/3390900258?check_suite_focus=true#step:16:33172) is also skipped. I will create a follow-up issue for this.

jougs (commented):
It seems like I overlooked another [MPI testing site](https://github.com/nest/nest-simulator/pull/2127/checks?check_run_id=3467061605#step:16:35108) during #2108 and that is still using `nosetests` and now failing. 

Also, I think this would be an excellent opportunity to address the deprecation warnings in the `pytests`.

heplesser:
@jougs Good catch, and I fully agree concerning the warnings, will look at them. Some seem to come from inside flask and I am afraid we need to live with that.

heplesser:
@jougs I just pushed the complete reorganization of MPI-based tests. I hope the symlink-magic will work in the Github runners as well—should know in approx 30 minutes. Two tests are now failing, but that shows that things work and they are fixed elsewhere :).

heplesser:
@jougs @hakonsbm @terhorstd I think my new setup is ready for review now. I will change the title of the PR. The previous Github Checks went through nicely except for the 4-rank Pytests, they should work now after I added `--oversubscribe`. 

I would prefer to fix the warnings from PyTest in a later PR to avoid delaying this one.

heplesser:
All looks good to me now with the following part in the output table:
```
07 pynesttests             658         4         0         1     117.5
35456
07 pynesttests mpi 2       165         0         1         0      29.4
35457
07 pynesttests mpi 4         1         0         0         0       1.0
35458
```
"mpi 2" and "mpi 4" are tests run with 2 and 4 procs, respectively.

The following two tests currently fail:
```
result.share.nest.testsuite.pytests.test_sp.test_disconnect.TestDisconnectSingle.test_synapse_deletion_one_to_one_no_sp
result.share.nest.testsuite.pytests.mpi.2.test_connect_symmetric_pairwise_bernoulli.TestSymmetricPairwiseBernoulli.testMakeSymmetric 
```

The first one will be fixed by #2131, the second needs checking—it was not run in the past.

testsuite/pytests/mpi/README.md
```diff
@@ -0,0 +1,12 @@
+# Python-based MPI tests
+
+This directory contain Python test requiring MPI.
```
hakonsbm, line 3:
```suggestion
This directory contains Python tests requiring MPI. 
```

testsuite/pytests/mpi/README.md
```diff
@@ -0,0 +1,12 @@
+# Python-based MPI tests
+
+This directory contain Python test requiring MPI. 
+
+The name of each subdirectory is the number of MPI processes to be used.
+Tests are to be run as
+
+    mpirun -np N pytest <folder>
```
hakonsbm, line 8: Stick with "directory" for clarity.
```suggestion
    mpirun -np N pytest <directory>
```

testsuite/pytests/mpi/4/test_consistent_local_vps.py
```diff
@@ -0,0 +1,44 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import nest
+import pytest
+
```
hakonsbm, line 24: Unused.
```suggestion
```

testsuite/pytests/mpi/4/test_consistent_local_vps.py
```diff
@@ -0,0 +1,44 @@
...
+    n_vp = 3 * num_procs
+    nest.SetKernelStatus({'total_num_virtual_procs': n_vp})
+
+    local_vps = list(nest.GetLocalVPs())
+
+    # Use thread-vp mapping of neurons to check mapping in kernel
+    nrns = nest.GetLocalNodeCollection(nest.Create('iaf_psc_delta', 2 * n_vp))
+
+    vp_direct = [n.get('vp') for n in nrns]
+    vp_indirect = [local_vps[n.get('thread')] for n in nrns]
```
hakonsbm, line 43: I think using attributes looks cleaner.
```suggestion
    vp_direct = list(nrns.vp)
    vp_indirect = [local_vps[t] for t in nrns.thread]
```

testsuite/pytests/connect_test_base.py
```diff
@@ -0,0 +1,599 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import numpy as np
+import scipy.stats
+import nest
+from scipy.stats import truncexpon
+import unittest
```
hakonsbm, line 26: Unused.
```suggestion
import unittest
```

hakonsbm (changes requested):

testsuite/do_tests.sh
```diff
@@ -154,12 +158,13 @@ echo "  PREFIX ............. $PREFIX"
...
     echo "  Python executable .. $PYTHON (version $PYTHON_VERSION)"
-    NOSE_VERSION="$("${NOSE}" --version | cut -d' ' -f3)"
-    echo "  Nose executable .... $NOSE (version $NOSE_VERSION)"
+    PYTEST_VERSION="$("${PYTEST}" --version 2>&1 | cut -d' ' -f2)"
+    echo "  Pytest executable .. $PYTEST (version $PYTEST_VERSION)"
     echo "  PYTHONPATH ......... `print_paths ${PYTHONPATH:-}`"
 fi
 if test "${HAVE_MPI}" = "true"; then
     echo "  Running MPI tests .. yes"
+    echo "  Mpirun executable .. $MPIRUN"
```
jougs, line 167:
```suggestion
    echo "  MPI launcher ....... $MPIRUN"
```

testsuite/do_tests.sh
```diff
@@ -131,6 +131,10 @@ NEST="nest_serial"
 
 HAVE_MPI="$(sli -c 'statusdict/have_mpi :: =only')"
 
+if test "${HAVE_MPI}" = "true"; then
+  MPIRUN="$(command -v mpirun)"
```
jougs, line 135: For MPICH, this may have to be `mpiexec`, but the user should have this defined in `~/.nestrc`. Maybe it is better to use something like this?
```suggestion
MPIRUN="$(sli -c '1 () () mpirun cst 0 get =only')"
MPIRUN="$(command -v $MPIRUN)"
```

testsuite/do_tests.sh
```diff
@@ -456,23 +461,23 @@ fi
...
-	"${PYTHON}" "${NOSE}" -v --with-xunit --xunit-testsuite-name="${XUNIT_NAME}" \
-		    --xunit-file="${XUNIT_FILE}" "${PYNEST_TEST_DIR}/test_mpitests.py" \
-		    2>&1 | tee -a "${TEST_LOGFILE}" >/dev/null
-	            # "&>FILE" or ">>FILE 2>&1" don't silence the line above. Why?!
+    "${PYTEST}" --verbose --junit-xml="${XUNIT_FILE}" --numprocesses=auto \
+          --ignore="${PYNEST_TEST_DIR}/mpi" "${PYNEST_TEST_DIR}" 2>&1 | tee -a "${TEST_LOGFILE}" 
+  
+    # Run tests in the mpi subdirectories, grouped by number of processes
+    if test "${HAVE_MPI}" = "true" -a "${MPIRUN}" ; then
+       for numproc in 2 4 ; do
```
jougs, line 476: Maybe you can make this iterate over all existing directories? That way, future extensions would really just require to drop a file or symlink into a directory.
```suggestion
       for numproc in $(cd ${PYNEST_TEST_DIR}/mpi/; ls -d */ | tr -d '/'); do
```

testsuite/pytests/test_connect_one_to_one.py
```diff
@@ -21,11 +21,10 @@
 
 import numpy as np
 import unittest
-import test_connect_helpers as hf
-from test_connect_parameters import TestParams
+import connect_test_base as hf
```
jougs, line 24: I find the name `hf` totally not understandable. Why not just
```suggestion
from connect_test_base import ConnectTestBase
```
...

testsuite/pytests/test_connect_one_to_one.py
```diff
@@ -21,11 +21,10 @@
...
 
 
-class TestOneToOne(TestParams):
+class TestOneToOne(hf.ConnectTestBase):
```
jougs, line 27: ... and
```suggestion
class TestOneToOne(ConnectTestBase):
```
?

If you agree, this should be applied for all similar tests.

jougs (changes requested):
Very nice! This is the world's best MPI testing setup we had so far ;-)

I have some suggestions, but am otherwise fine.

testsuite/do_tests.sh
heplesser: You are right. We should actually use this to build the full command further down as well.

testsuite/pytests/mpi/4/test_consistent_local_vps.py
heplesser: Ah, still stuck in the NEST 2 world ...

testsuite/pytests/test_connect_one_to_one.py
heplesser: You are right, will fix it.

testsuite/pytests/test_connect_one_to_one.py
heplesser: Indeed!

heplesser:
@jougs @hakonsbm I have fixed the issues you pointed out now and also got the symmetric Bernoulli test to work under MPI (it had not been written for MPI). I decided to replace the `import connect_test_base as hf` with just `import connect_test_base`, since a number of auxiliary functions are used from the module.

Would you re-review?

testsuite/pytests/mpi/2/test_issue_578_sp.py
```diff
@@ -19,18 +19,20 @@
...
 import sys
+import unittest
 
 HAVE_GSL = nest.ll_api.sli_func("statusdict/have_gsl ::")
 
 
 class TestIssue578():
+    """
+    This test must be run with 2 mpi processes.
+    """
```
jougs, line 33: I think this comment should rather explain what the test is about and the condition should be turned into a `skipIf` to check if we actually have two processes.

jougs (commented):
This looks mostly nice now. Just one more comment.

testsuite/pytests/mpi/2/test_issue_578_sp.py
hakonsbm: If a test in `mpi/2/` is run with anything other than two processes, should it be skipped or raise an error?

jougs: I think that should be up to the test. I can imagine situations, where you would symlink a test into multiple directories to get it run with multiple numbers of processes.

But I agree that violations should probably be treated as error or faillure instead of skip.

heplesser: I have renamed the test to the correct issue number (#576), implemented skipping for insufficient processes, and simplified the test code a little.

jougs (approved):
:+1:

testsuite/pytests/mpi/2/test_issue_576.py
```diff
@@ -0,0 +1,88 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import nest
+import sys
+import unittest
```
hakonsbm, line 25: `sys` is unused.
```suggestion
import unittest
```

testsuite/pytests/mpi/2/test_issue_576.py
```diff
@@ -0,0 +1,88 @@
...
+
+HAVE_GSL = nest.ll_api.sli_func("statusdict/have_gsl ::")
+
+
+@unittest.skipIf(nest.NumProcesses() < 2, "Requires >= 2 MPI process")
```
hakonsbm, line 30: For consistency. I chose to flip the double quotes because single quotes are used the most.
```suggestion
HAVE_GSL = nest.ll_api.sli_func('statusdict/have_gsl ::')


@unittest.skipIf(nest.NumProcesses() < 2, 'Requires >= 2 MPI process')
```

testsuite/pytests/mpi/2/test_issue_576.py
```diff
@@ -0,0 +1,88 @@
...
+        }
+    )
+    # Update the SP interval
+    nest.EnableStructuralPlasticity()
+    nest.SetKernelStatus({
+        'structural_plasticity_update_interval': 1000.,
+    })
+
+    growth_curve = {
+        'growth_curve': "gaussian",
```
hakonsbm, line 53:
```suggestion
        'growth_curve': 'gaussian',
```

testsuite/pytests/mpi/2/test_issue_576.py
```diff
@@ -0,0 +1,88 @@
...
+    }
+    neuronDict = {'V_m': -60.,
+                  't_ref': 5.0, 'V_reset': -60.,
+                  'V_th': -50., 'C_m': 200.,
+                  'E_L': -60., 'g_L': 10.,
+                  'E_ex': 0., 'E_in': -80.,
+                  'tau_syn_ex': 5., 'tau_syn_in': 10.,
+                  'I_e': 220.}
+
+    nest.SetDefaults("iaf_cond_exp", neuronDict)
```
hakonsbm, line 72:
```suggestion
    nest.SetDefaults('iaf_cond_exp', neuronDict)
```

hakonsbm (changes requested):
Just some minor things in the test and I'm happy.

heplesser:
@hakonsbm I accepted your suggestions, hope all is fine now.

hakonsbm (approved):
Thanks! Looks good now.

heplesser:
Good, I will merge as soon as the test are done.

jougs:
I just realized that `pytest` is called `pytest-3` on my machine and the whole new testsuite fails. Is this something we should take care of, or should I just create an alias on my machine and we wait until other people also complain?

heplesser:
@jougs I have just created #2148 to address this. It doesn't work fully yet due to my lack of CMake skills ...

PR #2035, opened 2021-05-13
Move precise spike time offset from Event to Time class

clinssen:

PR #2137, opened 2021-08-30
Add warning about synaptic plasticity mechanisms and precise spike timing

clinssen:
Spin-off from #1840. For now, synaptic plasticity mechanisms will continue to ignore precise spike timing, because changing this is a larger project that raises further questions (to be addressed in #2035). To prevent confusion in the mean time, this PR adds warnings about this behaviour to all synaptic plasticity models that are affected.

### DIFF
--- a/models/clopath_synapse.h
+++ b/models/clopath_synapse.h
@@ -62,7 +62,10 @@ hh_psc_alpha_clopath.
 
 .. warning::
 
-   This synaptic plasticity rule does not take :doc:`precise spike timing <simulations_with_precise_spike_times>` into account. When calculating the weight update, the precise spike time part of the timestamp is ignored.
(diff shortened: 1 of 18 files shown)


jessica-mitchell (changes requested):
@clinssen looks good; Aside from a minor comment about wording,  I noticed that the equations don't seem to render properly. The only thing looking a bit off is the lack of \\ at the end of the first line. Not sure if that will solve it.

clinssen:
- Fixed formatting of equations
- Wrapped long lines
- Merged suggestions for rewording in "jonke" synapses

jessica-mitchell (commented):
Great, looks good, thanks!

jessica-mitchell (approved):

heplesser (approved):

PR #2141, opened 2021-08-31
Replace the `Vose` class with `std::discrete_distribution`

### LINKED ISSUE #1546: Topology/Vose constructor should take const reference argument

### LINKED ISSUE #2097: Replace our custom `Vose` class with `std::discrete_distribution`
Follow-up from #2094. With the modernized RNG setup we may consider to replace the [`Vose`](https://github.com/nest/nest-simulator/blob/bd546f530627137e6280559b4cd2c99dd8569f69/nestkernel/vose.h#L43) class with [`std::discrete_distribution`](https://en.cppreference.com/w/cpp/numeric/random/discrete_distribution), if performance is ok.

hakonsbm:
With the modernized RNG setup we can replace the `Vose` class with `std::discrete_distribution`. See #2097 for discussion and benchmarks. Resolves #2097, resolves #2094.  Resolves #1546.


heplesser:
@hakonsbm Thanks, looks good in general. Could you remind me which connection tests contain cases affected by this change, i.e., spatial connections with fixed in/outdegree?  I just want to be sure we have tests in place that cover this.

hakonsbm:
@heplesser test_connect_layers.py contains tests for connecting spatial NodeCollections with different specifications. Most of the `test_connect_layers_indegree*` and `test_connect_layers_outdegree*` tests there should be affected by this change. See for example https://github.com/nest/nest-simulator/blob/bc5cc6268d42f3c8c829f1cc92ea762832e91449/testsuite/pytests/test_spatial/test_connect_layers.py#L182-L185

jougs (approved):
Just Beautiful! Thanks!

Also good to see how easy adding a new distribution is now :heart:

hakonsbm:
@heplesser I don't think we have such a test.

heplesser (approved):
I delegated stronger tests to a separate issue.

PR #2148, opened 2021-09-02
Detect pytest executable more robustly

heplesser:
Let CMake search for pytest executable to support at least the most common variants.

This does not fully work yet. In the CMake Summary here `PYNEST_FOUND` is not set even if PYNEST has been found: 

https://github.com/heplesser/nest-simulator/blob/2988bca3fd23d405cb94b73947d702a9e193d88a/cmake/ConfigureSummary.cmake#L52

and the `PYNEST` variable also does not "arrive" in `do_tests.sh`. But my CMake-wits are running a bit low right now. @jougs, do you have suggestions?


heplesser:
@jougs @terhorstd I have now fixed the pytest detection. I decided to go for a loop over `command` rather than `type` because the type doc states that hashed paths may lead to type not returning the first occurrence in the path, so command seemed safer.

testsuite/do_tests.sh
```diff
@@ -98,11 +98,18 @@ if test ! "${REPORTDIR}"; then
 fi
 
 if test "${PYTHON}"; then
-    command -v pytest >/dev/null 2>&1 || {
+    # find valid pytest command
+    for ptname in foo pytest pytest3 pytest-3 ; do
```
jougs, line 102: Why the `foo`?

```suggestion
    for ptname in pytest pytest3 pytest-3; do
```

testsuite/do_tests.sh
```diff
@@ -98,11 +98,18 @@ if test ! "${REPORTDIR}"; then
...
+        PYTEST="$(command -v ${ptname})"
+        if test $PYTEST ; then
```
jougs, line 104:
```suggestion
        if test ${PYTEST}; then
```

testsuite/do_tests.sh
```diff
@@ -98,11 +98,18 @@ if test ! "${REPORTDIR}"; then
...
+           break
+        fi
+    done
+
+    if test ! "${PYTEST}" ; then
```
jougs, line 109:
```suggestion
    if test ! "${PYTEST}"; then
```

jougs (changes requested):

heplesser:
@jougs Thanks for your checking and apologies for my sloppiness! All fixed now.

jougs (approved):
LGTM!

heplesser:
There seem to be problems with the Arbor test, which times out. In current master the test was "hiding" and thus seemed to pass. @hakonsbm @gtrensch From my analysis, it seems that something goes wrong with the communication of the min_delay. The exchange happens, but the value received (at least by arbor side) is wrong, typically 0, leading to a division by zero and then an exception on conversion to int. I believe this may be related to a "wild" mix of `double`, `MPI_FLOAT`, `np.float32` in the Arbor communication.

As only one MPI proc crashes, the other gets stuck. to avoid getting stuck, I now added a timeout to pytest, 30s per test. Note that that is per individual test, not per test file, so it should be generous.

heplesser:
I have now disabled the `ARBOR_BACKEND` test stage because that needs more work. This should allow this PR to be merged. I will open a separate issue for the Arbor backend testing.

jougs:
@terhorstd: could you please have a quick look? Thanks!

testsuite/do_tests.sh
```diff
@@ -127,6 +128,8 @@ TEST_OUTFILE="${REPORTDIR}/output.log"
 TEST_RETFILE="${REPORTDIR}/output.ret"
 TEST_RUNFILE="${REPORTDIR}/runtest.sh"
 
+TIME_LIMIT=120  # seconds, for Python tests
```
terhorstd, line 131: Please clarify in the comment if this timeout is per test, or for all tests.
```suggestion
TIME_LIMIT=120  # seconds, for each of the Python tests
```

testsuite/do_tests.sh
```diff
@@ -464,20 +467,26 @@ echo "Phase 7: Running PyNEST tests"
...
+          "${PYNEST_TEST_DIR}/non_concurrent" 2>&1 | tee -a "${TEST_LOGFILE}" 
   
     # Run tests in the mpi subdirectories, grouped by number of processes
     if test "${HAVE_MPI}" = "true" -a "${MPI_LAUNCHER}" ; then
        for numproc in $(cd ${PYNEST_TEST_DIR}/mpi/; ls -d */ | tr -d '/'); do
            XUNIT_FILE="${REPORTDIR}/${XUNIT_NAME}_mpi_${numproc}.xml"
-           PYTEST_ARGS="--verbose --junit-xml=${XUNIT_FILE} ${PYNEST_TEST_DIR}/mpi/${numproc}"
-           $(sli -c "${numproc} (${PYTEST}) (${PYTEST_ARGS}) mpirun =only") 2>&1 | tee -a "${TEST_LOGFILE}"
+           PYTEST_ARGS="--verbose --timeout $TIME_LIMIT --junit-xml=${XUNIT_FILE} ${PYNEST_TEST_DIR}/mpi/${numproc}"
+           $(sli -c "${numproc} (${PYTHON} -m pytest) (${PYTEST_ARGS}) mpirun =only") 2>&1 | tee -a "${TEST_LOGFILE}"
```
terhorstd, line 489: It's so spooky that line 489 actually works…
There is a severe mix of executables and parameters, and different types of quoting mixed with indirect evaluation, that should be cleaned up (not in this PR though). I would not have expected that a call to SLI is necessary to run the pytests, so I'd say this violates our principle of least surprise for me.

Please open an issue to refactor SLI out of the parallelization mechanics. I guess it's only doing config file parsing here to find information that could nowadays probably be found out by other means.

terhorstd (approved):
Only minor changes necessary, and this can be merged when resolved.

testsuite/do_tests.sh
heplesser: @terhorstd I will create an issue. But we do not actually use SLI here to run NEST in parallel. We only use SLI to construct the correct invocation for running MPI jobs. It uses the user-adjusted `mpirun` command defined in `.nestrc`. This ensures that we always use the proper, system-specific way, even away from OpenMPI :).

heplesser:
Merging without waiting for actions to complete since only a single comment was changed.

testsuite/do_tests.sh
jougs: The NEST config file `.nestrc` is a SLI script that can only be read by executing it. I doubt this will change until the point when we disentangle SLI and NEST completely, which I'm not expecting anytime soon. But as this config file is really only needed for getting the MPI launcher command for running the testsuite, the situation is not so bad after all.

terhorstd: Yes, I understand that SLI is not necessary to run pytests in parallel, which is exactly why I think this would be a positive first refactoring step. Requiring SLI just for parsing a string is what bothers me, especially since it already caused a major headache for exactly this wrong way of command line parsing. Just think about all characters that have a special meaning in SLI *or* bash *or* Python, be it a simple space in a path or so.
But probably you're right that we need to look at it in even broader scope. Maybe we find time to look into this at a smallish hackathon… – in a hut waaay out in the depth of nowhere.

heplesser: Please continue this discussion in #2154.

PR #2155, opened 2021-09-07
Remove Arbor recording backend

### LINKED ISSUE #2151: Review and correct Arbor backend test
The Arbor backend test fails currently as shown here: https://github.com/nest/nest-simulator/runs/3510427486?check_suite_focus=true#step:16:8270. Preliminary investigation indicates that MPI communication happens, but returns incorrect results at least for `min_delay`, typically 0. This might be due to a mismatch in datatype.

This issue was discovered in connection with #2148; during the review of #2076, the test was actually never run for a combination of reasons.

The following questions occurred to me in this context:
1. Since we do not test against Arbor, but a Python script imitating Arbor, how certain are we that our script correctly mimics Arbor? Should we at least be explicit about the Arbor version we target here?
2. Building the Arbor backend has no other requirements than building NEST with MPI, so I suggest to include the Arbor backend test in the FULL build instead of having a separate target—otherwise we run ≈1200 tests twice just to run one additional test.
3. I am not happy with the `mpiutil[_impl].h` files: they provide small code only needed for the Arbor backend and I think it would be better to include that code directly in the Arbor backend files. At least the file names should make clear that these files are related to the Arbor backend.

hakonsbm:
Removes the Arbor recording backend and associated tests and examples. Fixes #2151.


heplesser (approved):

jougs (approved):
LGTM. Thanks!

PR #2156, opened 2021-09-07
Update Random123 library to v1.14.0

### LINKED ISSUE #2100: Nest v3.0: bundled Random123 does not support s390x (fixed upstream)
**Describe the bug**
While updating the version of NEST to 3.0 in NeuroFedora, we observed that it includes an older version of Random123 that does not support the s390x architecture.

We sent a patch upstream to fix this, so the version of Random123 on the GitHub does support s390x. Could the included copy in NEST please be updated?

https://github.com/DEShawResearch/random123/blob/main/include/Random123/features/gccfeatures.h

The downstream bug where a Fedora community member submitted a patch is here:
https://bugzilla.redhat.com/show_bug.cgi?id=1806283

The patch is also attached (but it'll be better to use the latest code from upstream if possible).
[Random123-1.13.2-s390x.patch.txt](https://github.com/nest/nest-simulator/files/6759174/Random123-1.13.2-s390x.patch.txt)

If at all possible, since upstream has now made a release and is on GitHub, could an option please be added to allow us to use the system copy of Random123?

hakonsbm:
Fixes #2100.


hakonsbm:
@sanjayankur31 Could you check if this fixes your issue?

@heplesser Now that the Random123 library is git-based and on GitHub, I'm wondering if we should include it as a submodule instead. What do you think?

heplesser (approved):

heplesser:
@hakonsbm Thanks for this! I approved already, but maybe you could add a README in the Random123 directory specifying where the code comes from?

Concerning using a submodule, I know too little about this. One thing I'd be concerned about is if a specific NEST release would then still be well defined. I.e., would a release always point to a specific hash of a submodule? Maybe @jougs or @terhorstd could comment. In any case, I would delay transition to a submodule until after 3.1 is out.

sanjayankur31:
> @sanjayankur31 Could you check if this fixes your issue?

Thanks, that looks good and should fix the s390x build. I've not updated the Fedora package to Nest 3.0 yet since PyNN doesn't support it (See: https://github.com/NeuralEnsemble/PyNN/pull/724) but I'll run a test build in the meantime and report back if it fails.

> @heplesser Now that the Random123 library is git-based and on GitHub, I'm wondering if we should include it as a submodule instead. What do you think?

A submodule would work, but please do refer to a specific release if you do use it. We've seen upstream projects refer to random snapshots which are yet to make it to releases, and that makes it quite hard for downstream consumers (like Linux distributions) to unbundle submodules and use system copies instead since most tend to carry releases (with backports and patches perhaps) rather than unreleased git snapshots.

Also (and this may need verifying in case it has changed), I think GitHub release archives (and other autogenerated archives) do not include the submodules in them, so they will need to be generated manually to include the submodule code. e.g.: https://github.com/flaviotordini/minitube/issues/128

jougs (approved):
The changes look good to me, so I'm approving.

Regarding the question of using submodules, I probably don't have enough experience with them, but get a feeling that that would cause more trouble than just updating these files upon each release of the library. I thus leave that decision to the ones in the know.

Issue #2101, opened 2021-07-04
[NEST-3.0] crashes when built with `GLIBCXX_ASSERTIONS` defined


sanjayankur31:
(Edited for clarity)

**Describe the bug**
When NEST 3.0 is built with `GLIBCXX_ASSERTIONS`, it crashes on running. For example, when running the simple examples from the documentation, one gets:

```
$ python3 recording_demo.py

              -- N E S T --
  Copyright (C) 2004 The NEST Initiative

 Version: nest-3.0
 Built: Jul  3 2021 00:00:00

 This program is provided AS IS and comes with
 NO WARRANTY. See the file LICENSE for details.

 Problems or suggestions?
   Visit https://www.nest-simulator.org

 Type 'nest.help()' to find out more about NEST.

/usr/include/c++/11/bits/stl_vector.h:1123: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::front() [with _Tp = long int; _Alloc = std::allocator<long int>; std::vector<_Tp, _Alloc>::reference = long int&]: Assertion '!this->empty()' failed.
Aborted (core dumped)
```

```
$ python balancedneuron.py

              -- N E S T --
  Copyright (C) 2004 The NEST Initiative

 Version: nest-3.0
 Built: Jul  3 2021 00:00:00

 This program is provided AS IS and comes with
 NO WARRANTY. See the file LICENSE for details.

 Problems or suggestions?
   Visit https://www.nest-simulator.org

 Type 'nest.help()' to find out more about NEST.

Inhibitory rate estimate: 15.00 Hz
/usr/include/c++/11/bits/stl_vector.h:1123: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::front() [with _Tp = long int; _Alloc = std::allocator<long int>; std::vector<_Tp, _Alloc>::reference = long int&]: Assertion '!this->empty()' failed.
Aborted (core dumped)
```

**To Reproduce**
0. Build Nest 3.0 with `-D_GLIBCXX_ASSERTIONS` included in the compilation flags (enabled by default in Fedora)
1. (Minimal) reproducing example: https://nest-simulator.readthedocs.io/en/stable/_downloads/3861b671d578f9f10bb865919eeb2af4/recording_demo.py
2. Used parameters: NA
3. Command to run: `python recording_demo.py`
4. How to see error: crash is shown on the terminal as output

**Expected behavior**
Should not crash

**Screenshots**
NA

[...]

PR #2159, opened 2021-09-08
Fix out of bound access on pp_pop_psc_delta and pulsepacket_generator

niltonlk:
This fix part of #2101 

Installcheck with `-D_GLIBCXX_ASSERTIONS` flag reported out of bound access in pp_pop_psc_delta and pulsepacket_generator (see #2101)

-pp_pop_psc_delta case:
calibration in pp_pop_psc_delta created rhos_ages_ vector 1 unit larger than eta_kernel_ vector.
This caused theta_ages_ vector to be created 1 unit smaller than rhos_ages_ vector in line 348.
This causes out of bound access of theta_ages_ vector in the loop in line 361.

-pulsepacket_generator case:
line 223 was accessing the first element of an empty deque (spiketimes_) due to pop_front in line 221.
This was fixed by checking if spiketimes_ is not empty first (same as line 217)

### DIFF
--- a/models/pp_pop_psc_delta.cpp
+++ b/models/pp_pop_psc_delta.cpp
@@ -296,7 +296,7 @@ nest::pp_pop_psc_delta::calibrate()
       temp = 0;
     }
 
-    for ( int j = 0; j < V_.len_eta_; j++ )
+    for ( int j = 0; j < V_.len_eta_ - 1; j++ )
     {
       S_.age_occupations_.push_back( 0 );
       S_.thetas_ages_.push_back( 0 );
--- a/models/pp_pop_psc_delta.cpp
+++ b/models/pp_pop_psc_delta.cpp
@@ -319,7 +319,7 @@ nest::pp_pop_psc_delta::calibrate()
 void
 nest::pp_pop_psc_delta::update( Time const& origin, const long from, const long to )
 {
-  assert( to >= 0 && ( delay ) from < kernel().connection_manager.get_min_delay() );
+  assert( to >= 0 and ( delay ) from < kernel().connection_manager.get_min_delay() );
   assert( from < to );
 
   for ( long lag = from; lag < to; ++lag )
--- a/models/pp_pop_psc_delta.cpp
+++ b/models/pp_pop_psc_delta.cpp
@@ -296,6 +296,7 @@ nest::pp_pop_psc_delta::calibrate()
       temp = 0;
     }
 
+    // Set all except last state vector elements to zero, then fill last element with initial value
     for ( int j = 0; j < V_.len_eta_ - 1; j++ )
     {
       S_.age_occupations_.push_back( 0 );
--- a/models/pulsepacket_generator.cpp
+++ b/models/pulsepacket_generator.cpp
@@ -220,7 +220,7 @@ nest::pulsepacket_generator::update( Time const& T, const long from, const long 
     long prev_spike = B_.spiketimes_.front();
     B_.spiketimes_.pop_front();
 
-    if ( n_spikes > 0 and not B_.spiketimes_.empty() and prev_spike != B_.spiketimes_.front() )
+    if ( n_spikes > 0 and ( B_.spiketimes_.empty() or prev_spike != B_.spiketimes_.front() ) )
     {
       SpikeEvent se;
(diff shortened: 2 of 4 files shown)


models/pp_pop_psc_delta.cpp
```diff
@@ -296,7 +296,7 @@ nest::pp_pop_psc_delta::calibrate()
       temp = 0;
     }
 
-    for ( int j = 0; j < V_.len_eta_; j++ )
+    for ( int j = 0; j < V_.len_eta_ - 1; j++ )
```
heplesser, line 299: This looks like a very brutal fix. I looked some more at the pp_pop_psc_delta code and noticed (a) that the code is in poor shape and (b) that it has carried since Nov 2016 a deprecation notice referring users to gif_pop_psc_exp. I would therefore suggest to remove the model entirely. @jougs @terhorstd What do you think?

models/pulsepacket_generator.cpp
```diff
@@ -220,7 +220,7 @@ nest::pulsepacket_generator::update( Time const& T, const long from, const long
     long prev_spike = B_.spiketimes_.front();
     B_.spiketimes_.pop_front();
 
-    if ( n_spikes > 0 && prev_spike != B_.spiketimes_.front() )
+    if ( n_spikes > 0 && ( not B_.spiketimes_.empty() && prev_spike != B_.spiketimes_.front() ) )
```
heplesser, line 223: Please replace `&&` with `and` according to our current style guide. I also think that the extra `()` are not needed since we have a sequence of ands.

heplesser (changes requested):

niltonlk:
Have replaced `&&` with `and`
There were also some `||` which was replaced with `or`
pp_pop_psc_delta was also corrected, although it might get deleted.

heplesser:
@niltonlk Interestingly, after your sensible fix, `test_pulsepacket_generator.sli` now fails.

niltonlk:
The conditional expression to deliver the spike event was wrong. The last spike was never sent as it would not pass the if clause in line 223. Sorry for the silly mistake.

heplesser:
> The conditional expression to deliver the spike event was wrong. The last spike was never sent as it would not pass the if clause in line 223. Sorry for the silly mistake.

Good catch! I had to stare at the code for quite some time to understand how spike emission works in this generator.

models/pp_pop_psc_delta.cpp
jougs: I suggest to remove the model in 3.2 and add a corresponding sentence to [the detailed release notes of 3.1](https://github.com/nest/nest-simulator/blob/master/doc/userdoc/release_notes/v3.1/index.rst).

jougs: Just to clarify: please add the deprecation info for the model to the release notes in this PR, and open a new issue with the 3.2 milestone set about the removal of the model.

heplesser: I agree with @jougs. After some more code-staring, I now also understand why the `-1` is correct above. Even though we will remove the model in 3.2, could you add a comment above like
```python
# Set all except last element to zero.
```

heplesser: Given that the code is soon disappearing, I'd keep changes minimal, so your `-1` solution is fine. Just put this comment before the loop:
```python
# Set all except last state vector elements to zero, then fill last element with initial value
```

niltonlk: Got it.

heplesser:
@niltonlk Great, thanks. I created a separate PR for the release notes (#2160), so I will approve this one now.

heplesser (approved):

niltonlk:
Added deprecation message following 'iaf_psc_alpha_canon'.

My question is, is it really necessary to have deprecation message defined in `pynest/nest/lib/hl_api_helper.py` as well? Since in pynest the message defined in `models/modelsmodule.cpp` is also displayed?
Maybe `register_node_model` in `modelsmodule.cpp` could follow the same strategy as in `hl_api_helper.py` to also display the replacement model, and use only this latter method?

jougs (approved):
LGTM.

Also the way deprecation information is added, is fine by me.

PR #2162, opened 2021-09-13
Fix access of first element of an empty vector in pynestkernel

niltonlk:
This PR fix out of bound access in pynestkernel, reported in #2101.
With this fix, the two pynest examples (recording_demo.py and balancedneuron.py) reported in #2101 now succeed.

### DIFF
--- a/pynest/pynestkernel.pyx
+++ b/pynest/pynestkernel.pyx
@@ -620,7 +620,9 @@ cdef inline object sli_vector_to_object(sli_vector_ptr_t dat, vector_value_t _ =
     else:
         raise NESTErrors.PyNESTError("unsupported specialization")
 
-    memcpy(array_data, &vector_ptr.front(), vector_ptr.size() * sizeof(vector_value_t))
+    # skip when vector_ptr points to an ampty vector
+    if vector_ptr.size() > 0:
+        memcpy(array_data, &vector_ptr.front(), vector_ptr.size() * sizeof(vector_value_t))
 
     if HAVE_NUMPY:
         if vector_ptr.size() > 0:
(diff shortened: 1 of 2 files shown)


jougs (approved):
LGTM. I suggest a quick merge as this solves a real problem.

pynest/pynestkernel.pyx
```diff
@@ -620,7 +620,9 @@ cdef inline object sli_vector_to_object(sli_vector_ptr_t dat, vector_value_t _ =
     else:
         raise NESTErrors.PyNESTError("unsupported specialization")
 
-    memcpy(array_data, &vector_ptr.front(), vector_ptr.size() * sizeof(vector_value_t))
+    # skip when vector_ptr points to an ampty vector
```
hakonsbm, line 623:
```suggestion
    # skip when vector_ptr points to an empty vector
```

hakonsbm (changes requested):
Just a tiny typo.

heplesser:
@niltonlk The failure is curious and likely related to using multiple test runners (if several tests write to files of the same name, we will get into trouble). Could you for now change `--numprocesses=auto` to `--numprocesses=1` in these two places:

https://github.com/nest/nest-simulator/blob/12a897ae685082da099d9f1d5c0a7a24efca9691/testsuite/do_tests.sh#L102

https://github.com/nest/nest-simulator/blob/12a897ae685082da099d9f1d5c0a7a24efca9691/testsuite/do_tests.sh#L473

heplesser:
I had another look at the code and am certain that nothing went wrong in the past, because for an empty `vector_ptr`
```c++
memcpy(array_data, &vector_ptr.front(), vector_ptr.size() * sizeof(vector_value_t));
```
would not have copied anything. But it is clearly incorrect to request `front()` of an empty container, to this fix is needed.

hakonsbm (approved):
Thanks! Looks good now.

PR #2163, opened 2021-09-13
Warned about the removal of `nest.hl_api`

Helveg:
#2135 removes the `nest.hl_api` namespace and we wouldn't want to upset our unicorn population so I added a note to the release notes.
image


jougs (approved):
:rofl:

PR #2166, opened 2021-09-16
Handle empty and non-absolute installation prefixes

med-ayssar:
### Changes:

- Warn the user if the given `CMAKE_INSTALL_PREFIX`  is empty and stop the execution of `cmake`.
- Resolve the relative path to an absolute path.


CMakeLists.txt
```diff
@@ -27,6 +27,18 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 project( nest CXX C )
 set( NEST_USER_EMAIL "users@nest-simulator.org" )
 
+# this include for coloring messages
+include( Utils )
+# check if the given CMAKE_INSTALL_PREFIX is not empty
+if( NOT IS_ABSOLUTE ${CMAKE_INSTALL_PREFIX})
```
jougs, line 33: Following the *principle of least surprise*, I think we should just convert a non-absolute path to an absolute one (see [stackoverflow/39027269](https://stackoverflow.com/questions/39027269)).

CMakeLists.txt
```diff
@@ -27,6 +27,18 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 project( nest CXX C )
 set( NEST_USER_EMAIL "users@nest-simulator.org" )
 
+# this include for coloring messages
+include( Utils )
```
jougs, line 31: Maybe rename the file to `ColorMessages.cmake` and remove the comment?

Please also add a new line between this `include` and the next block of code.

jougs (changes requested):
Thanks for your contribution. I especially like the coloring :smile:

cmake/Utils.cmake
```diff
@@ -0,0 +1,89 @@
...
+            message(STATUS "${MSG_COLOR}${MSG_TEXT} ${Reset}") 
+        endif()
+    else()
+        message(STATUS "${MSG_TEXT}") 
+    endif()  
+endfunction()
+
+
+function(printWarning)
+    set(singleValued TEXT)
```
pnbabu, line 62: Could you rename the variable `singleValued` everywhere to `MESSAGE` or something similar?

cmake/Utils.cmake
```diff
@@ -0,0 +1,89 @@
...
+    cmake_parse_arguments(WARN "" "${singleValued}" "" ${ARGN})
+    print(HAS_COLOR MODE "WARNING" TEXT "${WARN_TEXT}" COLOR ${Yellow})
+endfunction()
+
+
+
+function(printError)
```
pnbabu, line 69: I think it is better to define the function arguments with the function name. That way the code is more readable.

med-ayssar: Do you mean to use the function name as prefix for its arguments ?

cmake/Utils.cmake
med-ayssar: Actually, here the keyword "TEXT" is the argument's name, and the user will only use and see the argument "TEXT"

cmake/Utils.cmake
med-ayssar: Ah!, I think you mean within the definition, right?

CMakeLists.txt
```diff
@@ -27,6 +27,25 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 project( nest CXX C )
 set( NEST_USER_EMAIL "users@nest-simulator.org" )
 
+include( ColorMessages )
+
+# check if the given CMAKE_INSTALL_PREFIX is not empty
```
pnbabu, line 32: This comment doesn't actually match with the if condition. Would it make more sense to check for empty string outside the `if( NOT IS_ABSOLUTE ${CMAKE_INSTALL_PREFIX})` since this condition is not a pre-requisite for checking empty string?

cmake/ColorMessages.cmake
```diff
@@ -0,0 +1,81 @@
...
+set(White       	"${Esc}[37m")
+set(BoldRed     	"${Esc}[1;31m")
+set(BoldGreen   	"${Esc}[1;32m")
+set(BoldBlue    	"${Esc}[1;34m")
+set(BoldCyan    	"${Esc}[1;36m")
+set(BoldMagenta 	"${Esc}[1;35m")
+set(BoldYellow  	"${Esc}[1;33m")
+set(BoldWhite   	"${Esc}[1;37m")
+
+function(print)
```
pnbabu, line 43: Could this function also be changed to have function arguments in the function definition?

CMakeLists.txt
```diff
@@ -27,6 +27,24 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
...
+if("${CMAKE_INSTALL_PREFIX}" STREQUAL "")
+  set(PathLength 0)
+  abortMSG("CMAKE_INSTALL_PREFIX cannot be an empty string" "CMAKE_INSTALL_PREFIX is:${CMAKE_INSTALL_PREFIX} [length:${PathLength}]")
+endif()
+
+# handel non-aboslute path
```
jougs, line 38:
```suggestion
# handle relative installation prefixes
```

CMakeLists.txt
```diff
@@ -27,6 +27,24 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
...
+if( NOT IS_ABSOLUTE ${CMAKE_INSTALL_PREFIX})
+    # convert relative path to absolute path
+    printWarning("CMAKE_INSTALL_PREFIX isn't an absolute path")
```
jougs, line 41:
```suggestion
```

CMakeLists.txt
```diff
@@ -27,6 +27,24 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
...
+    get_filename_component(absPath ${CMAKE_INSTALL_PREFIX} ABSOLUTE BASE_DIR ${CMAKE_BINARY_DIR})
+    set(CMAKE_INSTALL_PREFIX ${absPath})
+    printInfo("CMAKE_INSTALL_PREFIX is now set to ${CMAKE_INSTALL_PREFIX}")
```
jougs, line 44:
```suggestion
    printInfo("Relative CMAKE_INSTALL_PREFIX has been converted to absolute path ${CMAKE_INSTALL_PREFIX}")  
```

cmake/ColorMessages.cmake
```diff
@@ -0,0 +1,81 @@
+# Utils.cmake
```
jougs, line 1:
```suggestion
# ColorMessages.cmake
```

cmake/ColorMessages.cmake
```diff
@@ -0,0 +1,81 @@
...
+set(Magenta     	"${Esc}[35m")
+set(Yellow      	"${Esc}[33m")
+set(White       	"${Esc}[37m")
+set(BoldRed     	"${Esc}[1;31m")
+set(BoldGreen   	"${Esc}[1;32m")
+set(BoldBlue    	"${Esc}[1;34m")
+set(BoldCyan    	"${Esc}[1;36m")
+set(BoldMagenta 	"${Esc}[1;35m")
+set(BoldYellow  	"${Esc}[1;33m")
+set(BoldWhite   	"${Esc}[1;37m")
```
jougs, line 41: Please use spaces instead of tabs for indentation and make the indentation consistent

jougs (changes requested):
Please see my inline comments for change requests.

On second thought, it might be better to split the addition of message coloring out into another PR and changes *all* calls to messaging functions to take advantage of that.

CMakeLists.txt
```diff
@@ -27,7 +27,21 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 project( nest CXX C )
 set( NEST_USER_EMAIL "users@nest-simulator.org" )
 
-################################################################################
+# check if the given CMAKE_INSTALL_PREFIX is not empty
+if("${CMAKE_INSTALL_PREFIX}" STREQUAL "")
+  set(PathLength 0)
+  message(FATAL_ERROR "CMAKE_INSTALL_PREFIX cannot be an empty string" "CMAKE_INSTALL_PREFIX is:${CMAKE_INSTALL_PREFIX} [length:${PathLength}]")
```
jougs, line 33:
```suggestion
  message(FATAL_ERROR "CMAKE_INSTALL_PREFIX cannot be set to an empty string")
```
I think the additional text is not that helpful, as the prefix is anyway empty and the length in this situation always 0.

pnbabu: I agree.

pnbabu (approved):
I am fine with merging this after fixing @jougs's comment.

jougs (approved):

PR #2167, opened 2021-09-16
Add an example for the stdp dopamine synapse

YounesBouhadjar:
In this example, I connect two neurons using the `stdp_dopamine_synapse` model. The two neurons are driven by a Poisson generator. The `volume_transmitter` which delivers the dopamine signal is also driven by a Poisson generator.

The spike and weights look as follow:

example_stdp_dopa_synapse

My problem with this example, though, is that I don't show explicitly how the dopamine release affects the weight change. A way out is to record the state variables `c` and `n` of the synapse model, just like the weight, but if I'm not mistaken there are no recorders for synaptic variables other than the weight?

### DIFF
--- a/pynest/examples/stdp_dopamine_synapse.py
+++ b/pynest/examples/stdp_dopamine_synapse.py
@@ -127,6 +127,7 @@ ms_w = 2
 alpha_spikes = 0.4
 alpha_grid = 0.4
 
+# plot data
 fig, ax = plt.subplots(nrows=4)
 
 ax[0].plot(spike_times1, np.ones(len(spike_times1)), '|', ms=ms, mew=mew, color="blue", alpha=alpha_spikes)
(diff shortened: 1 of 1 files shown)


clinssen:
Does it make sense to have an example in addition to the unit test, or does the unit test itself serve as a sufficient example? It it organised in the same manner and also produces a similar output figure.

YounesBouhadjar:
I was wondering about the same question, @heplesser what do you think?

heplesser:
I add @jessica-mitchell as a second reviewer to check the example style. 

In general, I would keep examples and tests separate because they serve different purposes. And which test are you referring to?

Would it make sense to change the example to more clearly show the effect of the presence and absence of dopa-input? E.g., first run without any dopa over a longer period of time (1s?) to show plasticity in absence of dopa and then another second with dopa?

clinssen:
I have a slight preference to create a usage example in the NESTML repository as a Jupyter notebook, and just include the test in NEST. An example/test already exists in NESTML (https://github.com/nest/nestml/blob/master/tests/nest_tests/stdp_neuromod_test.py) but could (and will) be improved further still and converted into a notebook. If you are agreed, this PR can be closed and we can focus on the test (https://github.com/nest/nest-simulator/pull/2168).

YounesBouhadjar:
Alright, sounds good to me as well

PR #2169, opened 2021-09-20
Fix container access in corner cases and add _GLIBCXX_ASSERTIONS checks to all test configurations

hakonsbm:
Partial fix of #2101, fixing the `BlockVector` iterator.

As reported in #2101, some C++ tests fail when compiling with `-D_GLIBCXX_ASSERTIONS`. The tests fail because iteration of  a `BlockVector` uses an index to keep track of the current block. When the iterator goes past the first or last block in the `BlockVector`, it then tries to access a non-existent block. See for example this line in `bv_iterator::operator--()`:
https://github.com/nest/nest-simulator/blob/bb9fab8a3a8796c03d08435815494b9397973107/libnestutil/block_vector.h#L614

This PR refactors the `BlockVector` iterator, `bv_iterator`, to not use an index, `block_index_`, but an iterator to keep track of which block `bv_iterator` is in.

### DIFF
--- a/libnestutil/block_vector.h
+++ b/libnestutil/block_vector.h
@@ -598,8 +598,15 @@ inline bv_iterator< value_type_, ref_, ptr_ >& bv_iterator< value_type_, ref_, p
   if ( block_it_ == current_block_end_ )
   {
     ++block_vector_it_;
-    block_it_ = block_vector_it_->begin();
-    current_block_end_ = block_vector_it_->end();
+    // If we are at end() now, we are outside of the blockmap, which is
+    // undefined. The outer iterator can be in an undefined position that
+    // will compare as safely larger or equal to end(), but we don't set
+    // the inner iterators.
+    if ( block_vector_it_ != block_vector_->blockmap_.end() )
+    {
+      block_it_ = block_vector_it_->begin();
+      current_block_end_ = block_vector_it_->end();
+    }
   }
   return *this;
 }
@@ -613,11 +620,19 @@ inline bv_iterator< value_type_, ref_, ptr_ >& bv_iterator< value_type_, ref_, p
   {
     --block_it_;
   }
-  else
+  else if ( block_vector_it_ != block_vector_->blockmap_.begin() )
   {
     --block_vector_it_;
     current_block_end_ = block_vector_it_->end();
     block_it_ = current_block_end_ - 1;
+  }
+  else
+  {
+    // We are before begin(), which is undefined. We mark this by moving the
+    // outer iterator to an undefined position that will compare as safely smaller
+    // than begin(). According to C++17 Standard, §27.2.6, decrementing an
+    // iterator that is equal to begin() yields undefined behavior.
+    --block_vector_it_;
   }
   return *this;
 }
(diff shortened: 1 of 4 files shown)


heplesser:
@hakonsbm To show that this really fixes the problem, we should add the `-D_GLIBCXX_ASSERTIONS` flag to the full build. Also, somethings seem to go wrong with quicksort on the blockvector now.

hakonsbm:
@heplesser Won't adding the `-D_GLIBCXX_ASSERTIONS` flag here make the other tests listed in #2101 break? This PR fixes only the `BlockVector` tests:
- test_blockvector.test_operator_mm
- test_operator_eq
- test_operator_neq

heplesser:
Yes, but we need to check that this really solves the problem, so I think we need the flag. We will then need to integrate the various PRs solving the different issues to get back to green again.

heplesser:
@hakonsbm Thanks, that shows that we are down to just five failing tests on the full build, which might be closely related. And the C++ tests pass with the assertions in place :).

libnestutil/block_vector.h
```diff
@@ -604,14 +609,14 @@ inline bv_iterator< value_type_, ref_, ptr_ >& bv_iterator< value_type_, ref_, p
...
     --block_it_;
   }
   else
   {
-    --block_index_;
-    current_block_end_ = block_vector_->blockmap_[ block_index_ ].end();
+    --block_vector_it_;
+    current_block_end_ = block_vector_it_->end();
     block_it_ = current_block_end_ - 1;
   }
```
heplesser, line 621: To avoid undefined behavior at the left edge of the blockvector, we need to be careful here. I'd suggest the following:
```c++
else if ( block_vector_it != block_vector_->begin() )
{
  --block_vector_it_;
  current_block_end_ = block_vector_it_->end();
  block_it_ = current_block_end_ - 1;
}
else
{
  // We are before begin(), which is undefined. We mark this by moving the
  // outer iterator to an undefined position that will compare as safely smaller 
  // than begin(). According to C++17 Standard, §27.2.6, decrementing an
  // iterator that is equal to begin() yields undefined behavior.
  --block_vector_it; 
}

**Review by heplesser — CHANGES_REQUESTED** [review]
I added a suggestion for how to handle "walking off the left". Could you also change the title and the description of the PR to cover the corrections to the MPI-communication corner cases as well? Maybe something general "Fix container access in corner cases".

Also, could you as a check run some network model examples (e.g. hpc, microcircuit) once with NEST 3.1 and once with this branch to see if results are equal?

File: libnestutil/block_vector.h (hunk 1 above)
Comment (hakonsbm) [reply libnestutil/block_vector.h]: Done! And I also added an equivalent check in `operator++()`.

**Comment by hakonsbm** [comment]
@heplesser Both HPC benchmark and the microcircuit example give the same results as using NEST 3.1.

**Review by heplesser — APPROVED** [review]

File: libnestutil/block_vector.h [hunk 2]
```diff
@@ -63,7 +63,8 @@ class bv_iterator
   using iter_ = bv_iterator< value_type_, cv_value_type_&, cv_value_type_* >;
 
   const BlockVector< value_type_ >* block_vector_; //!< BlockVector to which this iterator points
```
stinebuu, line 65: This is a bit of a nitpick, but moving this comment to make it consistent with the comments below.
```suggestion
  //! BlockVector to which this iterator points
  const BlockVector< value_type_ >* block_vector_;
```

stinebuu (commented):
I have a tiny little comment, but otherwise it looks good.

stinebuu (approved):

PR #2172, opened 2021-09-27
Remove erroneous check in stimulation device base class

sdiazpier:
This PR solves issue 83 from EBRAINS-cosim. A control had been added by mistake to the stimulation devices during the integration of the current version of NEST I/O. This control did not allow changing the status of the device once it had been initialized.

### DIFF
--- a/nestkernel/stimulation_device.cpp
+++ b/nestkernel/stimulation_device.cpp
@@ -123,10 +123,6 @@ void
 nest::StimulationDevice::set_status( const DictionaryDatum& d )
 {
 
-  if ( kernel().simulation_manager.has_been_prepared() )
-  {
-    throw BadProperty( "Input parameters cannot be changed while inside a Prepare/Run/Cleanup context." );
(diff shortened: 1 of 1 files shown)


sdiazpier:
Dear @jougs and @hakonsbm could you please take a look at this small fix for cosim?

jougs (commented):
Thanks for your contribution. The change in this PR looks good to me.

Given the potential for future problems, I wonder if we should add a test to `testsuite/pytests` that checks the expected behavior for all stimulation devices. Could you please comment on this?

sdiazpier:
Dear @jougs thanks for taking a look at the PR. We can indeed try to design a test suite for stimulation devices. I think the normal functionality without the new IO backends is already quite well tested within other tests but there are some cosim specific features which might be important to keep in check.

jougs:
Merging after an offline discussion with @sdiazpier. Tests will be added in a follow-up PR.

PR #2181, opened 2021-10-08
Distribute node positions over MPI processes

### LINKED ISSUE #2174: Creation and storage of node positions is not distributed over MPI processes
Positions of nodes are not stored only on the process that creates the node. Instead every process keeps the position of all nodes.

For example:

```python 
pos = [[0.1 * x, 0.2 * y] for x in range(2) for y in range(2)]

nodes = nest.Create('iaf_psc_alpha', positions=nest.spatial.free(pos))
print(nodes.spatial['positions'])
```
Running this with `mpirun -n 2 ...` will print the positions for the two processes:
```
((0.0, 0.0), (0.0, 0.2), (0.1, 0.0), (0.1, 0.2))
((0.0, 0.0), (0.0, 0.2), (0.1, 0.0), (0.1, 0.2))
```

The expected output is having positions of each node stored only on the process where that node is created:
```
((0.0, 0.2), (0.1, 0.2))
((0.0, 0.0), (0.1, 0.0))
```

A more serious problem occurs when using random parameters to specify the positions. Positions are then generated independently on every process. *This gives each process a unique set of positions for every node.*

```python 
nodes = nest.Create('iaf_psc_alpha', 4,
                    positions=nest.spatial.free(nest.random.uniform(min=0.0, max=10.0), extent=[10., 10.]))
print(nodes.spatial['positions'])
```
Running this with `mpirun -n 2 ...` prints the following positions for the two processes:
```
((8.02814534936482, 9.016191656392479),
 (6.5348208749375205, 5.340685376430327),
 (5.397721192609427, 7.426365082266452),
 (0.4799942492096431, 7.32943676565794))

((1.8772310713378204, 1.7144695308473847),
 (1.706878151748514, 9.142753859348328),
 (8.89801938544016, 4.133575766823213),
 (0.5622579219786008, 8.15168786096808))
```

hakonsbm:
Node positions are with this PR stored only on the process where the corresponding node is created. This fixes #2174.


heplesser:
@hakonsbm Have you tested this with the 4x4mm model or could @jhnnsnk help with that?

hakonsbm:
@heplesser I've run the 4x4mm model using this branch and master, with 4 nodes, 8 tasks per node and 16 threads per task. Because the model uses parameters for node positions, spike times are different, as expected. But the statistics looks the same. As for performance:

branch | create [s] | connect [s] | presimulate [s] | simulate [s] | MaxRSS [KB]
----------|---------|------------|------------------|-------------|-----
master (c545255) | 33.042 | 554.934 |       8.516 |   15.818 | 11624436
MPI_local_positions | 7.949 | 565.334 |       8.457 |   15.703 | 11346244

Create time is reduced because of the reduced number of positions each MPI process has to create. The MaxRSS is also reduced, as expected.

heplesser:
@hakonsbm Good that the statistics check out!

stinebuu:
@hakonsbm Before I start reviewing, could you just remind me why we are changing the `Take` logic? I am a bit out of the loop here. :)

hakonsbm:
@stinebuu `Take` uses arguments with one-based indexing, while slicing in both Python and C++ uses zero-based indexing. 

In master we adjust the indexing from one- to zero-based when going from Python to SLI, then back again when going to C++. In this PR, to avoid confusion and reduce complexity, I try to keep arguments when slicing consistent all the way from Python to C++. To avoid the change of index base, PyNEST now uses `Take_g_a` directly (which uses zero-based indexing) instead of going through `Take`. Then only `Take` has to make a change from one-based to zero-based indexing before it calls `Take_g_a`.

stinebuu:
@hakonsbm Ok, nice. It then does not directly have anything to do with positions and MPI, right? Other than the slicing of course.

hakonsbm:
@stinebuu Yes, that's right.

nestkernel/node_collection.h
```diff
@@ -363,6 +351,13 @@ class NodeCollection
    */
   virtual long find( const index ) const = 0;
 
+  /**
+   * Returns whether the NodeCollection has contains any nodes with proxies or not.
```
heplesser, line 355: Just for clarification: because of requirements on NCs with metadata, either all or none of the nodes in a NC have proxies, right? We should make that clear in the comment.

nestkernel/node_collection.h
```diff
@@ -490,7 +497,7 @@ NodeCollectionPTR operator+( NodeCollectionPTR lhs, NodeCollectionPTR rhs );
  * The composite type contains a collection of primitives which are not
  * contiguous and homogeneous with each other. If the composite is sliced, it
  * also holds information about what index to start at and which to end at, and
- * the step.
+ * the step. The endpoint is one past the last valid node.
```
heplesser, line 500: If find the documentation a bit confusing. Here you write that the endpoint is one past the last valid, in all comments below it says "... to end at", which sounds more like the "end" is included. Can we formulate this clearer?

nestkernel/node_collection.h
```diff
@@ -617,13 +626,13 @@ inline NodeIDTriple nc_const_iterator::operator*() const
   {
```
heplesser, line 626: I think inlining this operator makes not much sense given how complex it is.

nestkernel/node_collection.h
```diff
@@ -617,13 +626,13 @@ inline NodeIDTriple nc_const_iterator::operator*() const
...
-                  and element_idx_ < composite_collection_->stop_offset_ ) ) )
+      if ( not( part_idx_ < composite_collection_->end_part_
+             or ( part_idx_ == composite_collection_->end_part_
+                  and element_idx_ < composite_collection_->end_offset_ ) ) )
       {
-        throw KernelException( "Invalid NodeCollection iterator (composite element beyond specified stop element)" );
+        throw KernelException( "Invalid NodeCollection iterator (composite element beyond specified end element)" );
       }
     }
     else if ( part_idx_ >= composite_collection_->parts_.size()
```
heplesser, line 638: This test here seems redundant. Can't we always initialise end_offset_ and end_part_ so that the first test will do and we do not need to check them against being zero?

nestkernel/node_collection.h
```diff
@@ -721,6 +730,28 @@ inline nc_const_iterator nc_const_iterator::operator+( const size_t n ) const
   return it += n;
 }
 
+
+inline size_t nc_const_iterator::operator-( const nc_const_iterator& other ) const
```
heplesser, line 734: Do we need this operator? If not, we can remove it. The implementation by explicit iteration is slow. At least we should avoid code duplication by something like this:
```c++
for ( auto it = nc_const_iterator( min( *this, other ) ) ; it < max( *this, other ) ; ++it )
{
  ++i;
}
```

nestkernel/node_collection.h
```diff
@@ -384,6 +379,16 @@ class NodeCollectionPrimitive : public NodeCollection
...
+  bool nodes_have_no_proxies_;         //!< Whether the primitive contains devices or not
+
+  /**
+   * Returns whether the model ids of all nodes in the primitive are the same as the expected model id.
+   *
+   * @param model_id Expected model id.
+   * @return true if model ids of all nodes in the primitive are the same as the expected model id, false otherwise.
+   *
+   */
+  bool all_model_ids_same_as_expected_( index ) const;
```
heplesser, line 391: This seems confusing. How could it ever be false for a primitive and thus homogeneous node collection?

nestkernel/node_collection.cpp
```diff
@@ -413,17 +422,14 @@ NodeCollectionPrimitive::NodeCollectionPrimitive::slice( size_t start, size_t st
   }
 
   NodeCollectionPTR sliced_nc;
-  if ( step == 1 )
+  if ( step == 1 and metadata_ == nullptr )
   {
-    sliced_nc = std::make_shared< NodeCollectionPrimitive >( first_ + start, first_ + stop - 1, model_id_, metadata_ );
+    // Subtract one from the last ID because end is one past the last node.
```
heplesser, line 427:
```suggestion
    // Create primitive NodeCollection passing node IDs.
    // Subtract 1 because "end" is one past last element to take while constructor expects ID of last node.
```

nestkernel/node_collection.cpp
```diff
@@ -240,7 +240,13 @@ NodeCollectionPrimitive::NodeCollectionPrimitive( index first,
   , last_( last )
   , model_id_( model_id )
   , metadata_( meta )
+  , nodes_have_no_proxies_( not kernel().model_manager.get_model( model_id_ )->has_proxies() )
 {
+  if ( not all_model_ids_same_as_expected_( model_id_ ) )
+  {
+    throw BadProperty( "model ids does not match" );
```
heplesser, line 247:
```suggestion
    throw BadProperty( "All nodes must have the same model model id" );
```
Maybe improve further with Compose to inform where it goes wrong. Even better: change name to `assert_consistent_model_ids_()` and throw exceptions directly in there.

nestkernel/node_collection.cpp
```diff
@@ -475,18 +481,33 @@ NodeCollectionPrimitive::overlapping( const NodeCollectionPrimitive& rhs ) const
   return ( ( rhs.first_ <= last_ and rhs.first_ >= first_ ) or ( rhs.last_ <= last_ and rhs.last_ >= first_ ) );
 }
 
+bool
+NodeCollectionPrimitive::all_model_ids_same_as_expected_( index expected_model_id ) const
```
heplesser, line 485: As commented above, turn into `assert_consistent_model_ids_()` and throw exception from here. Use `Compose()` to build error message telling which node ID causes problem.

lib/sli/nest-init.sli
```diff
@@ -215,18 +215,45 @@ def
   /coll Set
   slice length 1 eq
   {
-    slice 0 get dup 1 3 arraystore coll exch Take_g_a
+    slice 0 get dup 1 sub exch 1 3 arraystore coll exch Take_g_a
```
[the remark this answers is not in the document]
hakonsbm: I've fixed the logic, and added tests and comments.

testsuite/unittests/test_node_collection.sli
```diff
@@ -421,7 +421,7 @@ pop pop
   [ << /s 3 /e [ 1 2 3 ] >>
     << /s -3 /e [ 18 19 20 ] >>
     << /s [ 7 12 ] /e [ 7 8 9 10 11 12 ] >>
-    << /s [ -4 -2 ] /e [ 17 18 19 ] >>
```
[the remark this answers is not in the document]
hakonsbm: Not sure, I've reverted it.

testsuite/pytests/test_spatial/test_connect_sliced.py
```diff
@@ -46,88 +46,100 @@ class ConnectSlicedSpatialTestCase(unittest.TestCase):
...
+        for pos in [self.free_pos, self.grid_pos, wrapped_pos, extent_pos]:
+            nest.ResetKernel()
+            nodes = nest.Create('iaf_psc_alpha', positions=pos)
             for nodes_sliced in nodes:
                 for attr in ['edge_wrap', 'extent']:
                     spatial_attr = nodes.spatial[attr]
                     sliced_spatial_attr = nodes_sliced.spatial[attr]
                     self.assertEqual(spatial_attr, sliced_spatial_attr, 'with attr="{}"'.format(attr))
 
-    def test_connect_sliced_spatial_range(self):
```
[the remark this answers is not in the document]
hakonsbm: They were temporarily out of commission, but I've reinstated them now.

nestkernel/nestmodule.cpp
```diff
@@ -1581,23 +1581,16 @@ NestModule::Take_g_aFunction::execute( SLIInterpreter* i ) const
...
+  // If start or stop are counted backwards from the end with negative keys, they must be adjusted.
+  if ( start < 0 )
   {
-    start -= 1; // adjust from 1-based to 0-based indexing
-  }
-  else
-  {
-    start += g_size; // automatically correct for 0-based indexing
+    start += g_size;
+    stop = stop == 0 ? g_size : stop;
```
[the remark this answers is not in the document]
hakonsbm: Slices that result in empty NodeCollections are not allowed, and we must have `start < stop`.

nestkernel/nestmodule.cpp
```diff
@@ -1581,23 +1581,16 @@ NestModule::Take_g_aFunction::execute( SLIInterpreter* i ) const
...
-  if ( stop >= 0 )
-  {
-    // no adjustment necessary: adjustment from 1- to 0- based indexing
-    // and adjustment from last- to stop-based logic cancel
-  }
-  else
+  if ( stop < 0 )
   {
-    stop += g_size + 1; // adjust from 0- to 1- based indexin
+    stop += g_size;
```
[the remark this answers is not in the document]
hakonsbm: The last element is excluded. I've added comments and documentation on this.

nestkernel/free_layer.h
```diff
@@ -317,6 +337,18 @@ FreeLayer< D >::insert_global_positions_vector_( std::vector< std::pair< Positio
   std::sort( vec.begin(), vec.end(), node_id_less< D > );
 }
 
+template < int D >
+index
+FreeLayer< D >::lid_to_position_id_( index lid ) const
+{
+  const auto num_procs = kernel().mpi_manager.get_num_processes();
+  if ( not this->node_collection_->has_proxies() )
```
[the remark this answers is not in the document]
hakonsbm: It is correct for `FreeLayer`. I've fixed `GridLayer`.

nestkernel/node_collection.h
hakonsbm: True, but not relevant for this PR. I've created an issue, #2203.

nestkernel/node_collection.h
hakonsbm: Maybe. I've added it to #2203.

heplesser (changes requested):
Hi @hakonsbm! Thanks a lot! Now I just have a few small suggestions left, then this should be ready :).

testsuite/pytests/test_NodeCollection.py
```diff
@@ -178,21 +178,61 @@ def test_slicing(self):
...
+            self.assertEqual(nodes[::].tolist(), n_list[::])
+
+            # With start values
+            for start in range(-len(nodes), len(nodes)):
+                self.assertEqual(nodes[start:].tolist(), n_list[start:], f'with [{start}:]')
+
+            # With stop values
+            for stop in range(-len(nodes) + 1, len(nodes) + 1):
+                if stop == 0:
+                    continue  # Slicing to an empty NodeCollection is not allowed.
```
stinebuu, line 208:
```suggestion
                    continue  # Slicing an empty NodeCollection is not allowed.
```

testsuite/pytests/test_NodeCollection.py
```diff
@@ -178,21 +178,61 @@ def test_slicing(self):
...
+
+            # With start and step values
+            for start in range(-len(nodes), len(nodes)):
+                for step in range(1, len(nodes)):
+                    self.assertEqual(nodes[start::step].tolist(), n_list[start::step], f'with [{start}::{step}]')
+
+            # With stop and step values
+            for stop in range(-len(nodes) + 1, len(nodes) + 1):
+                if stop == 0:
+                    continue  # Slicing to an empty NodeCollection is not allowed.
```
stinebuu, line 223:
```suggestion
                    continue  # Slicing an empty NodeCollection is not allowed.
```

testsuite/pytests/test_NodeCollection.py
```diff
@@ -178,21 +178,61 @@ def test_slicing(self):
...
+                for step in range(1, len(nodes)):
+                    self.assertEqual(nodes[:stop:step].tolist(), n_list[:stop:step], f'with [:{stop}:{step}]')
+
+            # With start, stop, and step values
+            for start in range(-len(nodes), len(nodes)):
+                for stop in range(start+1, len(nodes) + 1):
+                    if stop == 0 or (start < 0 and start+len(nodes) >= stop):
+                        continue  # Cannot slice to an empty NodeCollection, or use stop <= start.
```
stinebuu, line 231:
```suggestion
                        continue  # Cannot slice an empty NodeCollection, or use stop <= start.
```

stinebuu (commented):
Thanks! I only have some suggestions to changes to some of the comments.

stinebuu (approved):

heplesser (approved):
Thanks a lot @hakonsbm!

PR #2183, opened 2021-10-09
Tidy up SLI startup, especially opening of sli-init.sli.

heplesser:
This PR tidies `slistartup.[h,cc]`, removing unused code, renaming some variables and in particular making the check for the `sli-init.sli` file more robust by not only checking for its existence, but also for readability. This might address confusing problems observed by @steffengraber during container construction.


jougs (approved):
This is indeed much cleaner.

PR #2186, opened 2021-10-11
Make volume transmitter inherit from Node instead of ArchivingNode

clinssen:
This is just a small cleanup to restrict the scope of the class hierarchy. As far as I'm aware, the volume transmitter does not by design ever need to maintain its own spike history, calcium concentration or structural plasticity information. Thus, the volume transmitter does not need to inherit from StructuralPlasticityNode or ArchivingNode.


jougs (approved):
LGTM!

PR #2194, opened 2021-10-20
Show colored messages in cmake output

med-ayssar:
#### Changes:

-  Replace each occurrence of `message([<mode>])`  with the corresponding `print` function, depending on the `mode` keyword.
-  `FATAL_ERROR` is mapped to `printError` function, outputting a **red** colored message.
-  `WARNING` is mapped to `printWarning` function, outputting a **yellow** colored message.
- `STATUS` is mapped to `printInfo` function, outputting a **green** colored message.
- A `message` function without a `mode`  is replaced by the `printInfo` function.
-  The summary phase of CMake is assigned a **BoldCyan** color.
- The `print` function in [ColorMessges](https://github.com/med-ayssar/nest-simulator/blob/fdbe4e2318e69a606c5e7485979171f0d7bb1d8a/cmake/ColorMessages.cmake#L43) can be used to extend new print functions with different colors.
- **Important**:  If the output will be written to a file, it will have escape sequences in it because they are being forced to be emitted always.
#### Before:
- First output lines:
cmake_output_orig_teil_1
- Summary:
camke_output_orig_teil_2

#### After:
- First output lines:
camke_output_teil_1

- Summary:
cmake_output_teil_2

- On Error: Setting `CMAKE_INSTALL_PREFIX` to an empty string!
cmake_abort


jougs:
@med-ayssar: ping!

cmake/ColorMessages.cmake
```diff
@@ -0,0 +1,71 @@
...
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+# Determine NEST version based on git branch
+#
```
jougs, line 21: This comment looks like it was copy-pasted-and-forgotten from another file. Please remove.
```suggestion
```

CMakeLists.txt
```diff
@@ -27,17 +27,19 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 project( nest CXX C )
 set( NEST_USER_EMAIL "users@nest-simulator.org" )
 
+include( ColorMessages )
+
 # check if the given CMAKE_INSTALL_PREFIX is not empty
 if("${CMAKE_INSTALL_PREFIX}" STREQUAL "")
-  message(FATAL_ERROR "CMAKE_INSTALL_PREFIX cannot be an empty string")
+  printERROR("CMAKE_INSTALL_PREFIX cannot be an empty string")
```
jougs, line 34:
```suggestion
  printError("CMAKE_INSTALL_PREFIX cannot be an empty string")
```

cmake/ColorMessages.cmake
```diff
@@ -0,0 +1,71 @@
...
+    print(HAS_COLOR MODE "WARNING" TEXT "Warning: ${TEXT}" COLOR ${Yellow})
+endfunction()
+
+function(printError TEXT)
+    print(HAS_COLOR MODE "FATAL" TEXT "Error: ${TEXT}" COLOR ${Red})
+endfunction()
+
+function(printInfo TEXT)
+    print(HAS_COLOR TEXT "Info: ${TEXT}" COLOR ${BoldGreen} )
+endfunction()
```
jougs, line 71:
```suggestion
endfunction()

```

CMakeLists.txt
jougs: According to [the CMake documentation](https://cmake.org/cmake/help/latest/command/message.html), using `FATAL_ERROR` does not just print the message, but also stops all processing and generation.

As much as I always disliked that side effect, I would not want the behavior changed by this PR. Could you please add a call like `message(FATAL_ERROR "")` to the definition of `printError`?

Or even better (and more explicit) wrap that call into a function `exit_cmake()` and calls to that everywhere right after where we use `printError`?

jougs (changes requested):
Thanks for the update. I have some minor suggestions and a less minor comment.

jougs:
@med-ayssar: could you please have a look at my comments and finalize this? I really think it is a big improvement and should go in instead of sitting here stale and half-ready :-)

CMakeLists.txt
med-ayssar: Actually, the `printError` will eventually call the function `message` with `FATAL_ERROR`, or am I misunderstanding your comment?

cmake/ColorMessages.cmake
med-ayssar: What did you change here?

jougs:
I think I have replied to all your questions. Please have another look.
And please also merge master, so the conflicts go away. Thanks!

jougs (approved):
@med-ayssar: I just took the liberty to merge my suggestions into your branch :grin:

@terhorstd: could you please especially look at differentiating between output to the shell and when redirected to files?

terhorstd:
All common tools strip color codes when the output is not going to a terminal. I'm not sure how the CI handles the output, but all common highlighting seems to be still there so there is either additional highlighting, or trickery with pseudo-ttys. So I'd guess that the `isatty()` / `test -t` type of calls can be used to split to non-highlighting versions when used in a pipe or redirected to a file (see for example this [StackExchange answer](https://unix.stackexchange.com/a/515781)).

heplesser:
@med-ayssar Could you fix the conflicts that have come up?

jougs:
@terhorstd!

jougs:
Due to the vacation of @terhorstd and that a) this is lying around for quite some time now, b) #2379 depends on it, and c) this is minor change that only affects the CMake output, I'm merging without the second review. Any (possible) problems with the rendering of output on the CI can be fixed in follow-up PRs.

PR #2195, opened 2021-10-22
Use `Development.Module` to be compatible with static Python interpreter

Helveg:
The current CMake build requires Python's shared library and header files to be available (`libpython.so` and `Python.h`). These aren't always available, and appearantly NEST builds perfectly fine without them. 

Let's drop this requirement (one line change from `find_package(Python3 Development)` to `Development.Module`) so that NEST can also build and be deployed to environments where Python is built in static mode.

Many non-Ubuntu non-system Python installations do not have these shared libraries. The most relevant use cases where we bump into this issue is on lightweight containers and the manylinux wheel images where these files have been [intentionally removed](https://github.com/pypa/manylinux/issues/255#issuecomment-450761634).

(And another one, often relevant to me and 15% of Python users, is that `pyenv` also builds without shared libraries ;) )

### DIFF
--- a/cmake/ProcessOptions.cmake
+++ b/cmake/ProcessOptions.cmake
@@ -392,7 +392,12 @@ function( NEST_PROCESS_WITH_PYTHON )
   elseif ( ${with-python} STREQUAL "ON" )
 
     # Localize the Python interpreter and ABI
-    find_package( Python 3.8 REQUIRED Interpreter Development.Module )
+    if ( ${CMAKE_VERSION} VERSION_LESS "3.18.0")
+      find_package( Python 3.8 REQUIRED Interpreter Development )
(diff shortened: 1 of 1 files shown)


jougs:
@terhorstd and I have discussed this and are in favor of a switch based on the CMake version available. I.e.,
> * Fall back to Development and emit a warning.

Helveg:
@jougs ready for review

jougs (approved):
Sweet!

jougs:
@steffengraber: could you please re-review (and possibly merge)? Thanks!

steffengraber (approved):
Have tested it in a clean Ubuntu 20.04 installation and found no problems. :+1:

PR #2201, opened 2021-11-03
Fix Clang warnings

heplesser:
This PR fixes a large number of warnings emitted by Apple Clang. Almost all of these warnings were due to explicit template class instantiations for Datum types, which could be removed by moving minimal methods from `_impl.h` files to `.h` files. This allowed removal of several `_impl.h` and `.cc` files.


jougs (approved):
Thanks for this cleanup. I did not follow each move individually, but trust the fact that the CI is happy.

nestkernel/parameter.cpp
```diff
@@ -27,12 +27,9 @@
 #include "spatial.h"
 #include "vp_manager_impl.h"
 
-// includes from sli
-#include "sharedptrdatum.h"
```
hakonsbm, line 30: I think a single blank line here is sufficient.
```suggestion
```

hakonsbm (changes requested):
Looks good. I just have one very minor formatting suggestion.

hakonsbm (approved):
Thanks!

PR #2205, opened 2021-11-09
Update outdated code snippets in documentation

stinebuu:
I noticed some code snippets in the guides that are outdated.


jougs (approved):
LGTM. Merging without second review, as this "only" concerns documentation.

PR #2208, opened 2021-11-12
Clean up the kernel in preparation for the module system refactoring

jougs:
This PR is a bit of everything all over the place.

The main user-facing changes are summarized in the [release notes file](https://github.com/nest/nest-simulator/compare/master...jougs:refactor_module_system?expand=1#diff-6be5a8306fdfd7edbc254cf36d2858491dc7272269ab026781a938a02bf131ea)

To ease review of the many seemingly unconnected changes, here's a high-level list of changes contained:
* Streamline the naming of model containers and access functions. Everything related to node models now has `node_models` in the name (used to be just `models`), everything connection model related has `connection_model` (used to be `synapse_prototype` or just `prototype`)
* Make the SLI-level dictionaries `modeldict`, `synapsedict`, `connruledict`, `growthcurvedict` implementation details of the respective manages by providing proper access functions and removing all external references to them
* Clean up and simplify the handling of the simulation resolution in `SimulationManager` and the setting of numbers of threads and virtual processes in `VPManager`
* Make changing the number of threads less radical by introducing a `change_number_of_threads()` function in the manager interface that allows each manager to decide what to do, instead of calling `finalize()` and `initialize()` on each of them unconditionally
* Clean up the registration and use of `proxynode`s and `dummy_spike_sources` and remove the related flag `private_model` from pertaining functions for registering models


nestkernel/connector_model.h
```diff
@@ -103,7 +103,7 @@ class ConnectorModel
 
   virtual void set_syn_id( synindex syn_id ) = 0;
 
-  virtual std::vector< SecondaryEvent* > create_event( size_t n ) const = 0;
+  virtual SecondaryEvent* create_event() const = 0;
```
heplesser, line 106: Why no longer a vector?

nestkernel/dynamicloader.cpp
```diff
@@ -153,11 +153,6 @@ DynamicLoaderModule::LoadModuleFunction::execute( SLIInterpreter* i ) const
 {
   i->assert_stack_load( 1 );
 
-  if ( kernel().model_manager.has_user_models() or kernel().model_manager.has_user_prototypes() )
```
heplesser, line 156: Why is this check no longer needed?

nestkernel/kernel_manager.cpp
```diff
@@ -132,25 +130,18 @@ nest::KernelManager::reset()
...
 nest::KernelManager::change_number_of_threads( thread new_num_threads )
 {
-  node_manager.finalize();
-  connection_manager.finalize();
-  model_manager.finalize();
-  modelrange_manager.finalize();
-  random_manager.finalize();
+  // For descriptions of the following conditions, see the cascade of
+  // user input checks in VPManager::set_status()
+  assert( node_manager.size() == 0 );
```
heplesser, line 135: Is the meaning of the comment above that any user input is checked and problems properly handled in `VPManager::set_status()`, so we only need the assertions here as a extra safety check? And if a user tries to change the number of threads at a time it cannot be changed they will get a proper error message, not NEST crashing on an assertion?

nestkernel/logging_manager.h
```diff
@@ -79,10 +78,11 @@ class LoggingManager : public ManagerInterface
   /**
    * Create a LoggingEvent.
    *
-   * This function creates a LoggingEvent that will be delivered to all
-   * registered logging clients, if the severity is above the set logging
-   * level. Do not use this function to do actuall logging in the source code,
-   * insted use the LOG() function provided by the logging.h header.
+   * This function creates a LoggingEvent that will be delivered to
+   * all registered logging clients, if the severity is above the set
+   * logging level. Do not use this function to do actuall logging in
```
heplesser, line 83:
```suggestion
   * logging level. Do not use this function to do actual logging in
```

nestkernel/model_manager.cpp
```diff
@@ -306,18 +305,18 @@ ModelManager::set_model_defaults( Name name, DictionaryDatum params )
...
   else if ( not synmodel.empty() )
   {
     id = static_cast< index >( synmodel );
     set_synapse_defaults_( id, params );
+    return true;
   }
   else
   {
-    throw UnknownModelName( name );
+    return false;
```
heplesser, line 318: Why do we no longer throw an exception?

nestkernel/model_manager.cpp
```diff
@@ -492,17 +504,18 @@ ModelManager::calibrate( const TimeConverter& tc )
 bool
 ModelManager::compare_model_by_id_( const int a, const int b )
```
heplesser, line 505: Is the name of this method right? We pass in ints, presumably model IDs, but the sorting criterium is the name.

nestkernel/model_manager.h
```diff
@@ -318,60 +246,39 @@ class ModelManager : public ManagerInterface
...
-   * The list of clean node models. The first component of the pair is a
-   * pointer to the actual Model, the second is a flag indicating if
-   * the model is private. Private models are not entered into the
-   * modeldict.
+   * List of clean built-in node models.
    */
-  std::vector< std::pair< Model*, bool > > pristine_models_;
-
-  std::vector< Model* > models_; //!< List of available models
+  std::vector< Model* > builtin_node_models_;
```
heplesser, line 251: I am not certain that I like `builtin_...`. First, in Python etc builtin functions are used just as other functions, they just come from a specific place. But in NEST, the `builtin_`s are never used for anything except to clone models from it. Second, from the perspective of `nestkernel`,  also `modelsmodule` is just another module, and any user modules will register to `builtin_node_models` just as `modelsmodule`. So there is nothing "built-in" about the models here. 

How about `unmodified_node_models_`or `reference_node_models_` or `raw_node_models_`? Maybe @abigailm has an idea?

nestkernel/modelrange_manager.cpp
```diff
@@ -112,24 +112,7 @@ ModelRangeManager::get_model_id( index node_id ) const
 nest::Model*
 nest::ModelRangeManager::get_model_of_node_id( index node_id )
 {
-  return kernel().model_manager.get_model( get_model_id( node_id ) );
-}
-
-bool
-ModelRangeManager::model_in_use( index i ) const
```
heplesser, line 119: Why is this no longer needed?

nestkernel/music_manager.h
```diff
@@ -42,45 +42,20 @@
...
-std::map< std::string, MusicPortData > music_in_portlist_;
-std::map< std::string, MusicEventHandler > music_event_in_portmap_;
- */
 
 namespace nest
 {
 
+/**
+ * Encapsulate all calls to MUSIC.
+*/
```
heplesser, line 51:
```suggestion
 */
```

nestkernel/nest.cpp
```diff
@@ -270,36 +262,50 @@ copy_model( const Name& oldmodname, const Name& newmodname, const DictionaryDatu
...
-
-  DictionaryDatum dict;
+  try
+  {
+    const index model_id = kernel().model_manager.get_node_model_id( component );
+    return kernel().model_manager.get_node_model( model_id )->get_status();
+  }
+  catch ( UnknownModelName& )
+  {
+  }
```
heplesser, line 291: Maybe add a comment here that this is a "fall-through" construction? Also on line 299-300? Took me a while to understand that ...

nestkernel/node_manager.cpp
```diff
@@ -81,6 +81,13 @@ NodeManager::finalize()
   destruct_nodes_();
 }
 
+void
+NodeManager::change_number_of_threads()
+{
+  // No nodes exist at this point, so nothing to tear down
```
heplesser, line 87: How do we know this?

heplesser (changes requested):
@jougs This is a major improvement. I have so far gotten through the first 58 files, up to and including `nestkernel/node_manager.h`. I still submit the review now so you can take a look and will try to do more later.

nestkernel/dynamicloader.cpp
jougs: I really dug deep to understand why this check was there in the first place. Other than a comment by @mschmidt87 in our old trac that "problems might occur when these are not checked" I could not find a good reason. So I removed them together with the pertaining tests.

nestkernel/connector_model.h
jougs: Because only element 0 of the vector was ever used. I assume this should have been refactored a long time ago...

nestkernel/kernel_manager.cpp
jougs: Yes. There are user-level check in `VPManager::set_status()` and the assertions here are only there to make sure those are properly checking their thing.

doc/userdoc/nest_server.rst
```diff
@@ -247,13 +247,14 @@ Client to execute a simple script on the Server using the
...
 
-    response = nsc.exec_script("print('Hello world!')")
-    print(response['stdout'])                        # 'Hello world!'
+    script = "print('Hello world!')"
+    response = nsc.exec_script(script)
+    print(response['stdout'])          # 'Hello world!'
 
-    response = nsc.exec_script("models=nest.Models()", 'models')
+    script = "models=nest.node_models"
+    response = nsc.exec_script(script, 'models')
```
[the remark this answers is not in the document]
jougs: `'models'` is the variable to return after running the script. Although not strictly part of this PR, I agree and clarified the example in 31ae6f140 by adding the keyword in front of the argument.

nestkernel/model_manager.cpp
jougs: A better name would be `compare_models_given_by_id_by_their_name`, but I'd rather not touch this right now. I actually think the whole `memory_info()` thing together with the custom allocator for models should go away. I reckon that a modern allocator would easily outperform our 18 year old custom solution...

nestkernel/modelrange_manager.cpp
jougs: I don't know. IIRC it has never been used anywhere, so I removed it.

testsuite/pytests/test_status.py
```diff
@@ -97,28 +99,28 @@ def test_SetDefaults(self):
 
         nest.ResetKernel()
 
-        for m in nest.Models():
-            if 'V_m' in nest.GetDefaults(m):
-                v_m = nest.GetDefaults(m)['V_m']
+        for model in nest.node_models + nest.synapse_models:
+            if 'V_m' in nest.GetDefaults(model):
```
hakonsbm, line 103: Are there any synapse models with `V_m`? Same for the other tests below.

nestkernel/node_manager.cpp
jougs: Because of the check in `VPManager::set_status()`. I've extended the comment in a885c2e238851b245.

hakonsbm (changes requested):
Nice! I have just a couple small suggestions.

nestkernel/dynamicloader.cpp
heplesser: OK, no risk of model IDs getting mixed up when we have arbitrary mixes of CopyModel and ModuleInstall?

nestkernel/kernel_manager.cpp
```diff
@@ -132,25 +130,18 @@ nest::KernelManager::reset()
 void
 nest::KernelManager::change_number_of_threads( thread new_num_threads )
 {
-  node_manager.finalize();
-  connection_manager.finalize();
-  model_manager.finalize();
-  modelrange_manager.finalize();
-  random_manager.finalize();
+  // For descriptions of the following conditions, see the cascade of
+  // user input checks in VPManager::set_status()
```
heplesser, line 134:
```suggestion
  // Inputs are checked in VPManager::set_status().
  // Just double check here that all values are legal.
```

nestkernel/kernel_manager.cpp
heplesser: Thanks, I will create a suggestion for a clearer comment.

nestkernel/model_manager.cpp
heplesser: OK!

nestkernel/model_manager.h
heplesser: OK.

pynest/nest/lib/hl_api_models.py
```diff
@@ -38,23 +39,20 @@
 
 @check_stack
 def Models(mtype="all", sel=None):
```
heplesser, line 41: Shouldn't this be marked as deprecated?

pynest/nest/lib/hl_api_models.py
```diff
@@ -95,34 +90,33 @@ def Models(mtype="all", sel=None):
 
 @check_stack
 def ConnectionRules():
```
heplesser, line 92: Also deprecated?

pynest/nest/lib/hl_api_models.py
```diff
@@ -95,34 +90,33 @@ def Models(mtype="all", sel=None):
...
     Parameters
     ----------
     model : str
-        Name of the model
+        Name of the model or recording backend
     params : str or dict
         Dictionary of new default parameter values
     val : str, optional
-        If given, `params` has to be the name of a model property.
+        If given, ``params`` has to be the name of a property.
```
heplesser, line 119:
```suggestion
        If given, ``params`` has to be the name of a parameter.
```

pynest/nest/ll_api.py
```diff
@@ -398,8 +399,17 @@ def init(argv):
         else:
             try:
                 import keyword
```
heplesser, line 401: Why not place with other imports on top?

pynest/nest/ll_api.py
```diff
@@ -398,8 +399,17 @@ def init(argv):
...
+                    "node_models",
+                    "recording_backends",
+                    "rng_types",
+                    "stimulation_backends",
+                    "synapse_models",
+                )
+                for kwl in keyword_lists:
+                    keyword.kwlist += GetKernelStatus(kwl)
             except ImportError:
                 pass
```
heplesser, line 414: Why continue if `from .lib.hl_api_simulation import GetKernelStatus` fails? Doesn't that point to a non-working PyNEST installation and we could quit at once?

sli/slidict.cc
```diff
@@ -248,7 +248,7 @@ DictinfoFunction::execute( SLIInterpreter* i ) const
  Synopsis: dict length_d -> int
 
  Examples: <</a 1 /b 2>> length_d -> 2
-   modeldict length_d --> 34
+   statusdict length_d --> 28
```
heplesser, line 251:
```suggestion
```
Length of `statusdict` could change, so I would drop the example entirely.

testsuite/pytests/test_status.py
```diff
@@ -76,7 +76,9 @@ def test_GetDefaults(self):
 
         nest.ResetKernel()
 
-        for model in nest.Models():
+        node_models = nest.GetKernelStatus("node_models")
+        synapse_models = nest.GetKernelStatus("synapse_models")
+        for model in node_models + synapse_models:
```
heplesser, line 81:
```suggestion
        for model in nest.node_models + nest.synapse_models:
```
Or did I misunderstand something about this test?

testsuite/regressiontests/ticket-396.sli
```diff
@@ -1,123 +0,0 @@
-/*
```
heplesser, line 1: Why is this test no longer relevant?

testsuite/regressiontests/ticket-683.sli
```diff
@@ -1,77 +0,0 @@
...
- *
- *  You should have received a copy of the GNU General Public License
- *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
- *
- */
-
-/** @BeginDocumentation
-
-Name: testsuite::ticket-683 - Ensure that NEST raises error when number of threads changed too late.
-
```
heplesser, line 26: Why is this test no longer relevant?

heplesser (changes requested):
@jougs Thanks for the first batch of fixes. I have now gone through the remaining files and have some more questions and suggestions.

pynest/nest/lib/hl_api_models.py
```diff
@@ -95,34 +90,33 @@ def Models(mtype="all", sel=None):
 
 @check_stack
 def ConnectionRules():
-    """Return a typle of all available connection rules, sorted by name.
+    """Return a tuple of all available connection rules.
```
stinebuu, line 93: Is this function also deprecated?

testsuite/pytests/test_status.py
```diff
@@ -76,7 +76,9 @@ def test_GetDefaults(self):
 
         nest.ResetKernel()
 
-        for model in nest.Models():
+        node_models = nest.GetKernelStatus("node_models")
```
stinebuu, line 79: Why do we not use `nest.node_models` and `nest.synapse_models` directly here?

stinebuu (changes requested):
Very nice @jougs! I have some small comments and suggestions.

nestkernel/dynamicloader.cpp
jougs: I don't see such a risk.

testsuite/regressiontests/ticket-396.sli
jougs: Because now the resolution can in fact be changed after the resolution, something this test would detect as wrong behavior. The reason why this is now possible is that we are not actually `finalizing()` and re-`initializing()` each manager upon changing the number of threads anymore, but instead just do whatever is necessary in each manager and thus a previous setting of the resolution survives that operation (which it did not previously).

testsuite/regressiontests/ticket-683.sli
jougs: Because we can now set model defaults at arbitrary times instead of just before setting the resolution or the number of threads as it used to be. I could not see any bad effect when lifting this restriction.

pynest/nest/ll_api.py
jougs: See below.

pynest/nest/ll_api.py
jougs: I'm not the original author of the code, but think that the `except` is meant to catch a missing `keyword` module rather than to catch the `import` that you pointed out. If my interpretation is right, I think the code is fine as is.

jougs (commented):
@stinebuu, @heplesser: I think I have addressed all of your concerns. Could you please have another look and resolve the conversations if you're happy? And could you please also have a look at @hakonsbm's last comment? Thanks!

jougs:
Sorry for pushing only after submitting the comments and by this ruining the linking. :-/

pynest/nest/ll_api.py
heplesser: But `keyword` has been part of the Python library since at least 2.7 (see https://docs.python.org/2.7/library/keyword.html), so it should always be there.

testsuite/regressiontests/ticket-396.sli
heplesser: I assume you meant "... now the number of threads can ... be changed after the resolution ..."? Do we have a new test that ensures that changing the number of threads before and after changing the resolution works correct (even in presence of MPI)?

testsuite/regressiontests/ticket-683.sli
heplesser: Do we have tests for this as well, i.e., setting defaults at arbitrary times compared to resolution changes and change of number of threads?

pynest/nest/ll_api.py
jougs: Alright. I've moved the keyword import in 36ac587c1.

heplesser:
> @stinebuu, @heplesser: I think I have addressed all of your concerns. Could you please have another look and resolve the conversations if you're happy? And could you please also have a look at @hakonsbm's last comment? Thanks!

@jougs I just commented. Which comment by @hakonsbm do you refer to? It looks to me like you addressed all of his.

jougs:
@heplesser: Thanks for pointing out the missing tests. When adding them, I found that some of the thread changing functions are actually not quite working as expected. I'll investigate and come back asap. Should I convert to draft, or are you fine waiting for a request to re-review as is?

heplesser:
> @heplesser: Thanks for pointing out the missing tests. When adding them, I found that some of the thread changing functions are actually not quite working as expected. I'll investigate and come back asap. Should I convert to draft, or are you fine waiting for a request to re-review as is?

I think it depends on the time perspective you expect. If you think it will be fixed within a day or two, leave as is, if it will take a week, convert to draft.

jougs:
This PR is now also based on #2260 and should not be merged before that.

testsuite/regressiontests/ticket-396.sli
jougs: I've added a (rather minimal) test in 6e843d4.

testsuite/regressiontests/ticket-683.sli
jougs: I've added a (rather minimal) test in 6e843d4.

jougs:
@hakonsbm, @heplesser, @stinebuu: I think I have now addressed all your concerns and comments. Could you please have another look?

The problem of times in models is now addressed in #2260 and this PR already contains the changes from there. It should thus only be merged after that one.

models/correlation_detector.cpp
```diff
@@ -54,13 +56,15 @@ nest::correlation_detector::Parameters_::Parameters_( const Parameters_& p )
...
-  // Check for proper properties is not done here but in the
-  // correlation_detector() copy c'tor. The check cannot be
-  // placed here, since this c'tor is also used to copy to
-  // temporaries in correlation_detector::set_status().
-  // If we checked for errors here, we could never change values
-  // that have become invalid after a resolution change.
-  delta_tau_.calibrate();
+  if ( not delta_tau_.is_step() )
+  {
+    delta_tau_ = 5 * Time::get_resolution();
```
heplesser, line 61: Should this emit a warning?

models/correlation_detector.cpp
```diff
@@ -346,3 +342,23 @@ nest::correlation_detector::handle( SpikeEvent& e )
...
 }
+
+void
+nest::correlation_detector::calibrate_time( const TimeConverter& tc )
+{
+  if ( not P_.delta_tau_.is_step() )
+  {
+    std::string old = String::compose( "(was %1 ms)", P_.delta_tau_.get_ms() );
+    P_.delta_tau_ = 5 * Time::get_resolution();
+    std::string msg = String::compose( "Default value for delta_tau is now %1 ms %2", P_.delta_tau_.get_ms(), old );
```
heplesser, line 353: I find the message construction a bit difficult to follow. Why not store the old value in a double and then build the full string here?

models/correlation_detector.h
```diff
@@ -100,9 +104,8 @@ the
 delta_tau            ms       Bin width. This has to be an odd multiple of
                               the resolution, to allow the symmetry between
                               positive and negative time-lags.
```
heplesser, line 106: I would mention the default value also here, also the default for tau_max (that one below in the tau_max part)

models/correlomatrix_detector.cpp
```diff
@@ -58,13 +60,15 @@ nest::correlomatrix_detector::Parameters_::Parameters_( const Parameters_& p )
...
-  // Check for proper properties is not done here but in the
-  // correlomatrix_detector() copy c'tor. The check cannot be
-  // placed here, since this c'tor is also used to copy to
-  // temporaries in correlomatrix_detector::set_status().
-  // If we checked for errors here, we could never change values
-  // that have become invalid after a resolution change.
-  delta_tau_.calibrate();
+  if ( not delta_tau_.is_step() )
+  {
+    delta_tau_ = 5 * Time::get_resolution();
```
heplesser, line 65: Emit warning?

models/correlomatrix_detector.cpp
```diff
@@ -391,3 +387,23 @@ nest::correlomatrix_detector::handle( SpikeEvent& e )
 
   } // device active
 }
+
+void
+nest::correlomatrix_detector::calibrate_time( const TimeConverter& tc )
+{
+  if ( not P_.delta_tau_.is_step() )
+  {
+    std::string old = String::compose( "(was %1 ms)", P_.delta_tau_.get_ms() );
```
heplesser, line 396: See comment above.

models/correlospinmatrix_detector.cpp
```diff
@@ -58,13 +60,15 @@ nest::correlospinmatrix_detector::Parameters_::Parameters_( const Parameters_& p
...
-  // Check for proper properties is not done here but in the
-  // correlospinmatrix_detector() copy c'tor. The check cannot be
-  // placed here, since this c'tor is also used to copy to
-  // temporaries in correlospinmatrix_detector::set_status().
-  // If we checked for errors here, we could never change values
-  // that have become invalid after a resolution change.
-  delta_tau_.calibrate();
+  if ( not delta_tau_.is_step() )
+  {
+    delta_tau_ = Time::get_resolution();
```
heplesser, line 65: Warning?

models/correlospinmatrix_detector.cpp
```diff
@@ -483,3 +479,25 @@ nest::correlospinmatrix_detector::handle( SpikeEvent& e )
 
   } // device active
 }
+
+void
+nest::correlospinmatrix_detector::calibrate_time( const TimeConverter& tc )
+{
+  if ( not P_.delta_tau_.is_step() )
+  {
+    std::string old = String::compose( "(was %1 ms)", P_.delta_tau_.get_ms() );
```
heplesser, line 488: See comment above.

models/noise_generator.cpp
```diff
@@ -74,9 +75,14 @@ nest::noise_generator::Parameters_::Parameters_( const Parameters_& p )
   , dt_( p.dt_ )
   , num_targets_( 0 ) // we do not copy connections
 {
-  // do not check validity of dt_ here, otherwise we cannot copy
-  // to temporary in set(); see node copy c'tor
-  dt_.calibrate();
+  if ( not dt_.is_step() )
+  {
+    dt_ = 10 * Time::get_resolution();
```
heplesser, line 80: Warning?

models/noise_generator.cpp
```diff
@@ -391,3 +389,19 @@ nest::noise_generator::set_data_from_stimulation_backend( std::vector< double >&
...
 }
+
+void
+nest::noise_generator::calibrate_time( const TimeConverter& tc )
+{
+  if ( not P_.dt_.is_step() )
+  {
+    std::string old = String::compose( "(was %1 ms)", P_.dt_.get_ms() );
+    P_.dt_ = 10 * Time::get_resolution();
+    std::string msg = String::compose( "Default value for dt is now %1 ms %2", P_.dt_.get_ms(), old );
```
heplesser, line 400: See comment above.

nestkernel/simulation_manager.cpp
```diff
@@ -204,13 +204,20 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
       }
       else
       {
+        double old_res = nest::Time::get_resolution().get_ms();
+        tic_t old_tpms = nest::Time::get_resolution().get_tics_per_ms();
```
heplesser, line 208:
```suggestion
        const double old_res = nest::Time::get_resolution().get_ms();
        const tic_t old_tpms = nest::Time::get_resolution().get_tics_per_ms();
```

nestkernel/simulation_manager.cpp
```diff
@@ -204,13 +204,20 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
...
         nest::Time::set_resolution( tics_per_ms, resd );
         // adjust to new resolution
         clock_.calibrate();
         // adjust delays in the connection system to new resolution
         kernel().connection_manager.calibrate( time_converter );
         kernel().model_manager.calibrate( time_converter );
-        LOG( M_INFO, "SimulationManager::set_status", "tics per ms and resolution changed." );
+
+        std::string msg = String::compose(
+          "tics per ms and resolution changed from %1 tics and %2 ms to %3 tics and %4 ms.",
```
heplesser, line 218:
```suggestion
          "Tics per ms and resolution changed from %1 tics and %2 ms to %3 tics and %4 ms.",
```

nestkernel/simulation_manager.cpp
```diff
@@ -233,12 +240,17 @@ nest::SimulationManager::set_status( const DictionaryDatum& d )
       }
       else
       {
+        double old_res = nest::Time::get_resolution().get_ms();
```
heplesser, line 243:
```suggestion
        const double old_res = nest::Time::get_resolution().get_ms();
```

testsuite/pytests/test_changing_tic_base.py
```diff
@@ -36,52 +36,6 @@ class TestChangingTicBase(unittest.TestCase):
     def setUp(self):
         nest.ResetKernel()
 
-    def test_models(self):
```
heplesser, line 39: Why are you removing this test? Or rather, why is it justified to do so?

heplesser (changes requested):
@jougs Looks pretty good, just a few small comments.

testsuite/pytests/mpi/4/test_set_status_resolution_nthreads.py
```diff
@@ -0,0 +1,48 @@
...
+import pytest
+
+
+@pytest.fixture
+def reset():
+    nest.ResetKernel()
+
+
+def testSetStatus_resolution_before_nthreads(reset):
+    """"""
```
hakonsbm, line 32: Should these tests have docstrings?

hakonsbm (changes requested):

testsuite/pytests/test_changing_tic_base.py
jougs: The test was testing that values are the same before and after changing the number of tics per ms and the resolution. This is not the case anymore for the models that are now automatically updating their values. On second thought, I should probably rather create a mechanism that just excludes auto-adjusted values from the check, instead of removing the test altogether.

models/correlation_detector.cpp
jougs: Agreed. Fixed in a62b0dabb.

doc/userdoc/release_notes/v3.2/index.rst
```diff
@@ -19,4 +19,5 @@ ConnPlotter
 ~~~~~~~~~~~
 All files related to ConnPlotter have been removed from NEST 3.2 and 
 moved to a separate repository `connplotter <https://github.com/nest
-/connplotter> _`.
\ No newline at end of file
+/connplotter> _`.
+>>>>>>> cb61289cc5751f68eb42f5eaf6564acc03823252
```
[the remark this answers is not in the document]
jougs: Fixed in 0b6995752. Thanks for being so attentive :-)

models/correlation_detector.cpp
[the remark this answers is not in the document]
jougs: Good idea. Done for all pertaining models in 8a73f7c0d.

models/correlation_detector.cpp
```diff
@@ -54,13 +56,15 @@ nest::correlation_detector::Parameters_::Parameters_( const Parameters_& p )
...
-  // that have become invalid after a resolution change.
-  delta_tau_.calibrate();
+  if ( not delta_tau_.is_step() )
+  {
+    delta_tau_ = 5 * Time::get_resolution();
+  }
+  else
+  {
+    delta_tau_.calibrate();
+  }
```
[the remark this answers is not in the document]
jougs: Agreed. Fixed in 59b119354.

testsuite/pytests/test_changing_tic_base.py
jougs: I've re-added the test extended by an ignore-list for automatically adjusted parameters in 6436b9811.

models/correlation_detector.h
jougs: Done for all changed models in fb8470296.

testsuite/pytests/mpi/4/test_set_status_resolution_nthreads.py
jougs: They should and doctrings have been added in 90e6b6585.

models/correlation_detector.cpp
jougs: I don't think so.

There is already an info message when the resolution is actually changed. The adaptation here only updates the default (to the same value that was already infoed to the user) when models are reinstated from pristine prototypes (that still might have old values) upon changing the number of threads and if the very first model created happens to be one that can adapt default values automatically.

I consider all of these cases technical details the user does not need to know about and which will probably change again in the next installment of the module refactoring series of PRs.

models/correlospinmatrix_detector.cpp
jougs: See above.

models/noise_generator.cpp
jougs: See above.

jougs:
@heplesser, @hakonsbm: many thanks for your detailed reviews. I've addressed all of your points and would be grateful if you could quickly resolve conversations where possible.

!! Please do not merge yet! This should only go into 3.3 and we first need to get 3.2 out.

doc/userdoc/release_notes/v3.3/index.rst
```diff
@@ -0,0 +1,89 @@
...
+~~~~~~~~~~~~~~
+
+The model parameter ``delta_tau`` in the ``correlation_detector``,
+``correlomatrix_detector``, and ``correlospinmatrix_detector``, as
+well as ``dt`` in the ``noise_generator`` are now automatically
+adjusted and made compatible with a newly set simulation resolution to
+avoid errors when those models are instantiated. Moreover, the default
+value for ``delta_tau`` in the ``correlation_detector`` has been
+changed from 1.0 to 5 times the simulation resolution, in order to be
+consistent with the documentation of the device.
```
jougs, line 28:
```suggestion
changed from 1.0 ms to 5 times the simulation resolution, in order to
be consistent with the documentation of the device.
```

models/correlation_detector.cpp
heplesser: OK, thanks for the explanation.

heplesser (approved):
@jougs Thanks a lot, especially for the re-instated and improved test! All fine now from my side.

hakonsbm (approved):
Thanks! Looks good to me now.

heplesser:
> !! Please do not merge yet! This should only go into 3.3 and we first need to get 3.2 out.

Do you want to switch it to draft to avoid accidental merging?

stinebuu (approved):
Looks good to me once formatting/tests pass again.

PR #2214, opened 2021-11-22
Change internal links to ref role in user documentation

jessica-mitchell:
This PR changes nearly all the internal links that use `:doc: `to `:ref:`

The main reason is that the :doc: role requires the relative path, and anytime the source file is moved, all links to the file in the documentation need updating. 
With the :ref: role, no path is required, but each page needs to include a label to use as a reference.

We are anticipating changes to the organization of the doc source files, and this makes it easier to not have to worry about all mentions of a file.

Every reStructuredText file in userdoc/ got a label. Most* :doc: roles are changed to :ref:. 
In addition, many broken links were fixed.
The styleguide is also updated to include the preference for using the :ref: role.

*There are still some :doc: roles in the files, because with everything there are exceptions:
Links to external projects using intersphinx, Python files, Jupyter notebooks, and autogenerated files (models and examples) still use the :doc: role.

During build, Sphinx spits out warnings of "undefined label" when a it can't find the internal reference. These have been checked and should all be resolved.


doc/userdoc/guides/random_numbers.rst
```diff
@@ -45,7 +45,7 @@ How to work with randomness in NEST
 NEST makes randomness easy: Setting a single seed gives you full
 control over randomness throughout a simulation, and you can choose seed values
 at liberty. NEST's powerful mechanism to :ref:`parameterize nodes and synapses <param_ex>`
-and NEST's :ref:`probabilistic connection rules <connection_mgnt>` allow you
+and NEST's :ref:`probabilistic connection rules <connection_management>` allow you
```
med-ayssar, line 48: Won't be better to make the name and the reference have similar meanings.
```suggestion
and NEST's :ref:`probabilistic connection properties <network_with_random_properties>` allow you
```

med-ayssar (approved):

doc/userdoc/guides/random_numbers.rst
jessica-mitchell:
```suggestion
and NEST's :ref:`probabilistic connection rules <conn_rules>` allow you
```

jessica-mitchell: The label has to reference an existing reference. I'd suggest specifying the section on connection rules in the connection management guide.

jessica-mitchell:
@terhorstd I addressed your comments, but since this was updating the double back ticks for some of the terms, I decided to go through and try find all terms that needed double back ticks and add them.

jessica-mitchell:
@pnbabu @med-ayssar  I found some additional broken links in the models directory that I also updated to the ref role

pnbabu (approved):
Looks good to me!

med-ayssar:
Looks fine to me too.

PR #2216, opened 2021-11-22
Add additional reference to Izhikevich model documentation

jessica-mitchell:
This PR adds some text to the model doc mentioning differences between NEST and the original author's model. Includes reference to Pauli et al. paper. 2018.

Not sure if this needs a lot more detail in the documentation or if link suffices.

### DIFF
--- a/models/izhikevich.h
+++ b/models/izhikevich.h
@@ -77,6 +77,13 @@ For all other purposes, it is recommended to use the standard technique for
 forward Euler integration. In this case, ``consistent_integration`` must be set
 to true (default).
 
+Note that the Izhikevich model in NEST differs from the origial publication [1]_.
+The Izhikevich model (as published by the author) has some numerical issues, which the Izhikevich model
+(in the NEST source) does not have. Also, the STDP model used in the original paper does not perfectly
+match the STDP synapse shipped with NEST. For further details see [2]_.
+
+
+
 Parameters
 ++++++++++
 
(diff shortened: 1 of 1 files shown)


clinssen (commented):
Would move to reject (sorry!)

The paragraph about the numerics is redundant with the preceding paragraph, that already covers these issues, and how these are addressed in the NEST model. The comment about STDP does not seem relevant to this model as it is a neuron model without any direct relationship to synaptic plasticity.

models/izhikevich.h
```diff
@@ -77,6 +77,13 @@ For all other purposes, it is recommended to use the standard technique for
 forward Euler integration. In this case, ``consistent_integration`` must be set
 to true (default).
 
+Note that the Izhikevich model in NEST differs from the origial publication [1]_.
+The Izhikevich model (as published by the author) has some numerical issues, which the Izhikevich model
+(in the NEST source) does not have. Also, the STDP model used in the original paper does not perfectly
+match the STDP synapse shipped with NEST. For further details see [2]_.
+
+
```
jougs, line 85:
```suggestion
For a detailed analysis and discussion of the numerical issues in the original publication, see [2]_.
```

jougs (changes requested):
I tend to agree with @clinssen, but still think a reference to *Pauli et al.* would be a good addition to the model documentation.

jougs (approved):
Thanks!

PR #2218, opened 2021-11-25
Fix typo in hpc_benchmark

jarsi:
There was a typo in the hpc_benchmark.py. This PR fixes this.


jougs (approved):
LGTM. Merging without second review due to minor-ness ;-)

PR #2220, opened 2021-11-26
Attribute assignment safeguards for `NestModule` and `SynapseCollection`

### LINKED ISSUE #2219: Prohibit setting of unknown attributes on nest kernel and synapses
As of NEST 3.1, kernel attributes can be set directly as attributes of `nest`. This can lead to subtle errors due to typing errors or incorrectly remembered names. For example,

```python
nest.random_seed = 1234
```

looks like correct code to set the random seed, but has actually no effect on the simulation, because the correct attribute name is

```python
nest.rng_seed = 1234
```

We introduced careful checking of parameter names in dictionaries passed to `SetStatus` over 15 years ago, so not catching this any longer is a serious regression.

Any attempt to set an undefined attribute should raise an error. 

The same problem exists for connections, where

```python
n = nest.Create('iaf_psc_alpha')
nest.Connect(n, n)
c = nest.GetConnections()
c.gewicht = 10
```

passed without error. Only for node attributes  checks seem to be in place, as

```python
n.voltage = -50
```

raises an error.

Helveg:
* Settings unknown attributes on nest module raises an AttributError
* Fixed an always-true condition that allowed setting unknown attributes on SynapseCollection.

Closes #2219.

Just 1 remark: while the network or synapse collection is empty, you _can_ set unknown attributes. Any idea why this was implemented, should it remain that way?

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -100,6 +100,7 @@ class NestModule(types.ModuleType):
         _api = list(k for k in self.__dict__ if not k.startswith("_"))
         _api.extend(k for k in dir(type(self)) if not k.startswith("_"))
         self.__all__ = list(set(_api))
+
         # Block setting of unknown attributes
         type(self).__setattr__ = _setattr_error
 
@@ -413,7 +414,6 @@ def _setattr_error(self, attr, val):
             cls_attr.__set__(self, val)
         else:
             raise err
-
 
 
 def _rel_import_star(module, import_module_name):
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -406,6 +406,16 @@ class NestModule(types.ModuleType):
 
 
 def _setattr_error(self, attr, val):
+    """
+    When attributes on the `nest` module instance are set, check if it exists on the
+    module type and try to call `__set__` on them. Without this explicit check `nest`s
+    `__setattr__` shadows class attributes and descriptors (such as `KernelAttribute`s).
+
+    Once this function exists on the `nest` module, new attributes can only be added using
+    `__dict__` manipulation. It is added onto the module at the end of `__init__`,
+    "freezing" the module.
+    """
+
     err = AttributeError(f"can't set attribute '{attr}' on module 'nest'")
     try:
         cls_attr = getattr(type(self), attr)
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -94,6 +94,9 @@ class NestModule(types.ModuleType):
 
         # Lazy loaded modules. They are descriptors, so add them to the type object
         type(self).spatial = _lazy_module_property("spatial")
+        type(self).raster_plot = _lazy_module_property("raster_plot")
+        type(self).visualization = _lazy_module_property("visualization")
+        type(self).voltage_trace = _lazy_module_property("voltage_trace")
 
         self.__version__ = ll_api.sli_func("statusdict /version get")
         # Finalize the nest module with a public API.
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -403,7 +403,16 @@ class NestModule(types.ModuleType):
 
 
 def _setattr_error(self, attr, val):
-    raise AttributeError(f"module 'nest' has no attribute '{attr}'")
+    err = AttributeError(f"can't set attribute '{attr}' on module 'nest'")
+    try:
+        cls_attr = getattr(type(self), attr)
+    except AttributeError:
+        raise err from None
+    else:
+        if hasattr(cls_attr, "__set__"):
+            cls_attr.__set__(self, val)
+        else:
+            raise err
 
 
 
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -100,6 +100,9 @@ class NestModule(types.ModuleType):
         _api = list(k for k in self.__dict__ if not k.startswith("_"))
         _api.extend(k for k in dir(type(self)) if not k.startswith("_"))
         self.__all__ = list(set(_api))
+        # Block setting of unknown attributes
+        type(self).__setattr__ = _setattr_error
+
 
     def set(self, **kwargs):
         return self.SetKernelStatus(kwargs)
@@ -397,6 +400,11 @@ class NestModule(types.ModuleType):
     _readonly_kernel_attrs = builtins.set(
         k for k, v in vars().items() if isinstance(v, KernelAttribute) and v._readonly
     )
+
+
+def _setattr_error(self, attr, val):
+    raise AttributeError(f"module 'nest' has no attribute '{attr}'")
+
 
 
 def _rel_import_star(module, import_module_name):
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -104,7 +104,6 @@ class NestModule(types.ModuleType):
         # Block setting of unknown attributes
         type(self).__setattr__ = _setattr_error
 
-
     def set(self, **kwargs):
         return self.SetKernelStatus(kwargs)
 
--- a/testsuite/pytests/test_get_set.py
+++ b/testsuite/pytests/test_get_set.py
@@ -54,15 +54,15 @@ class TestNestGetSet(unittest.TestCase):
         kst = nest.get("keep_source_table")
         self.assertEqual(type(nest).keep_source_table._default, kst, "get value not equal to default after ResetKernel")
         self.assertEqual(kst, nest.keep_source_table, 'kernel attribute value not equal to get value')
-        with self.assertRaises(AttributeError, "no AttributeError for unknown attribute"):
+        with self.assertRaises(AttributeError, msg="no AttributeError for unknown attribute"):
             nest.accessAbsolutelyUnknownThingOnNestModule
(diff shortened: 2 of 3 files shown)


pynest/nest/__init__.py
```diff
@@ -100,6 +100,9 @@ def __init__(self, name):
         _api = list(k for k in self.__dict__ if not k.startswith("_"))
         _api.extend(k for k in dir(type(self)) if not k.startswith("_"))
         self.__all__ = list(set(_api))
+        # Block setting of unknown attributes
```
heplesser, line 103: For clarity, I'd insert an empty line above.

pynest/nest/__init__.py
```diff
@@ -399,6 +402,11 @@ def __dir__(self):
     )
 
 
+def _setattr_error(self, attr, val):
+    raise AttributeError(f"module 'nest' has no attribute '{attr}'")
+
+
+
```
heplesser, line 409: Aren't there too many empty lines here?

heplesser (changes requested):
@Helveg Thanks!

- This change seems too drastic, all Pynest tests seem to fail now ;).
- How does this address the problem that unknown synapse collection objects are not detected?
- Please add a regression test, both for kernel and synapse attributes.

heplesser:
> Just 1 remark: while the network or synapse collection is empty, you _can_ set unknown attributes. Any idea why this was implemented, should it remain that way?

I cannot think of any reason for this behavior. One technical aspect might be that the detection of unknown parameter names for node collections largely happens in `SetStatus()` in the kernel, and for an empty collection we have nothing to call that on. @stinebuu @hakonsbm Any comments?

Helveg:
> How does this address the problem that unknown synapse collection objects are not detected?

Where can I find a description of this problem? I thought the problems were related to attribute assignment on `nest` and `SynapseCollection`?

heplesser:
> > How does this address the problem that unknown synapse collection objects are not detected?
> 
> Where can I find a description of this problem? I thought the problems were related to attribute assignment on `nest` and `SynapseCollection`?

Sorry, my comment was unclear. What I meant is: How does this change ensure that

```python
c = nest.GetConnections()
c.gewicht = 10
```

will raise an exception because `gewicht` is not a known synapse parameters. It should have been "parameter" or "attribute", not "object" in my original question.

Helveg:
There already was a `__setattr__` implementation that defers to `SynapseCollection.set`, which goes into SLI and errors graciously if the key isn't known. The problem was that the condition to defer to `.set` was always false, and instead `super().__setattr__(attr, value)` was called which would just set the attribute on `SynapseCollection`. You can see in e87ae1c6958dd1ac4f2caac5a3c0c5d2a16b6850 how I remedied that.

pynest/nest/__init__.py
```diff
@@ -399,6 +405,19 @@ def __dir__(self):
     )
 
 
+def _setattr_error(self, attr, val):
```
heplesser, line 408: @Helveg If I understand the code here right, this checks if there is a class attribute of the given name and raises an error only if that is not the case. Could you add a comment describing the mechanism here, so we remember it five years from now?

testsuite/pytests/test_get_set.py
```diff
@@ -41,6 +41,35 @@
...
 
+@nest.ll_api.check_stack
+class TestNestGetSet(unittest.TestCase):
+    """nest module get/set tests"""
+
+    def setUp(self):
+        nest.ResetKernel()
+
+    def test_get(self):
+        # TestCase.setUp calls ResetKernel so kernel attributes should be equal to their
```
heplesser, line 52: Please turn it into a docstring for the test.

testsuite/pytests/test_get_set.py
```diff
@@ -41,6 +41,35 @@
...
+        # defaults. Test will error if there is a problem with the `.get` mechanism.
+        kst = nest.get("keep_source_table")
+        self.assertEqual(type(nest).keep_source_table._default, kst, "get value not equal to default after ResetKernel")
+        self.assertEqual(kst, nest.keep_source_table, 'kernel attribute value not equal to get value')
+        with self.assertRaises(AttributeError, msg="no AttributeError for unknown attribute"):
+            nest.accessAbsolutelyUnknownThingOnNestModule
+        with self.assertRaises(KeyError, msg="no KeyError for unknown get key"):
+            nest.get("accessAbsolutelyUnknownKernelAttribute")
+
+    def test_set(self):
```
heplesser, line 62: Could you add a brief docstring describing what the test checks?

heplesser (changes requested):
@Helveg Thanks for the fixes! Could you improve comments a little, see inline remarks?

testsuite/pytests/test_get_set.py
Helveg: It's specific to what just the first lines do, I added a docstring for the test but left this as a comment.

heplesser (approved):

hakonsbm:
> > Just 1 remark: while the network or synapse collection is empty, you _can_ set unknown attributes. Any idea why this was implemented, should it remain that way?
> 
> I cannot think of any reason for this behavior. One technical aspect might be that the detection of unknown parameter names for node collections largely happens in `SetStatus()` in the kernel, and for an empty collection we have nothing to call that on. @stinebuu @hakonsbm Any comments?

@heplesser Yes, but the behaviour isn't consistent for NodeCollections and SynapseCollections. 

For empty NodeCollections, trying to get a non-existent attribute results in an `AttributeError`, while trying to set an attribute will just return `None` without setting anything.

An empty SynapseCollection will return an empty tuple when trying to get an unknown attribute, while setting an attribute will set the attribute *on the SynapseCollection itself*. Interestingly, setting a valid attribute, like `weight`, on a non-empty SynapseCollection will also set it as an attribute of the SynapseCollection, not as a property of the connections.

testsuite/pytests/test_get_set.py
```diff
@@ -41,6 +41,49 @@
...
+        """
+
+        # TestCase.setUp calls ResetKernel so kernel attributes should be equal to their
+        # defaults. Test should also error if there is a problem in general with the
+        # `.get` mechanism.
+        kst = nest.get("keep_source_table")
+        self.assertEqual(type(nest).keep_source_table._default, kst, "get value not equal to default after ResetKernel")
+        self.assertEqual(kst, nest.keep_source_table, 'kernel attribute value not equal to get value')
+        # Getting the value of unknown attributes should error. Should error if there is
+        # a problem with possible `__getattr__` implementations.
```
stinebuu, line 64:
```suggestion
        # Getting the value of unknown attributes should error. The test should also error if there is
        # a problem with possible `__getattr__` implementations.
```

stinebuu (commented):
This looks good to me, I only have a small comment on a comment.

Regarding the different behavior for NodeCollection and SynapseCollection, I cannot remember why there is a difference. Maybe something we should look at further in another issue?

stinebuu (approved):
:+1:

jougs:
@Helveg: The merge of this PR leads to the following error on Read the Docs.
```
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/envs/latest/lib/python3.8/site-packages/sphinx/cmd/build.py", line 276, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/envs/latest/lib/python3.8/site-packages/sphinx/application.py", line 216, in __init__
    self.config = Config.read(self.confdir, confoverrides or {}, self.tags)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/envs/latest/lib/python3.8/site-packages/sphinx/config.py", line 173, in read
    namespace = eval_config_file(filename, tags)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/envs/latest/lib/python3.8/site-packages/sphinx/config.py", line 342, in eval_config_file
    raise ConfigError(msg % traceback.format_exc()) from exc
sphinx.errors.ConfigError: There is a programmable error in your configuration file:

  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/envs/latest/lib/python3.8/site-packages/sphinx/config.py", line 329, in eval_config_file
    exec(code, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/checkouts/latest/doc/userdoc/conf.py", line 86, in <module>
    nest.NestModule = type(nest)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/checkouts/latest/pynest/nest/__init__.py", line 423, in _setattr_error
    raise err from None
AttributeError: can't set attribute 'NestModule' on module 'nest'
```

@steffengraber, @jessica-mitchell, @terhorstd

terhorstd:
This breaks the documentation generation: https://readthedocs.org/api/v2/build/15469804.txt
We're looking into it.

terhorstd:
This was also one of the Python trickery `nest.NestModule = type(nest)` in Sphinx's `conf.py`, that is now not possible anymore. Where would this best to be fixed? Here or in `conf.py`?

Helveg:
1. Check if the trickery is still required, the line may just be removed (check if the KernelAttribute doc page is still generated correctly)
2. Change it to `nest.__dict__["NestModule"] = type(nest)`

PR #2147, opened 2021-09-01
Remove Connplotter from the NEST code base

pnbabu:
Fixes #1971 

This PR removes the files related to `ConnPlotter` and are now moved to a new repository [nest-connplotter](https://github.com/nest/nest-connplotter).

The PR in the `nest-connplotter` repo with the initial changes - https://github.com/nest/nest-connplotter/pull/1

PR #2221, opened 2021-11-26
Removed `object` base from class argument lists

Helveg:
In Python 3 all classes have `object` added into their bases implicitly. There is no need to declare it explicitly. I left it in in the `ConnPlotter` files as not to conflict with #2147


heplesser (approved):
Thanks for cleaning up!

heplesser:
Since this is an absolutely straightforward clean-up, I merge without second review.

PR #2226, opened 2021-12-01
Remove some mergeconflict artifacts

jarsi:
This PR removes a couple of `*.orig` files. Judging from their file endings they appear to be merge artifacts.

Is it possible to have some kind of git hook or ci feature which automatically detects files with these endings and refuses to merge these?


clinssen:
Thanks for the PR! I would suggest adding ``*.orig`` to ``.gitignore``.

jougs:
I agree with @clinssen and will merge as soon as that's happened.

jarsi:
Thanks for your suggestions, it sounds very reasonable. I updated the `.gitignore` file.

PR #2231, opened 2021-12-07
Fix compiler warning issued by newest Clang

heplesser:
This PR fixes a compiler warning introduced by the last version of Clang (see https://windowsquestions.com/2021/08/12/is-the-class-name-in-same-scope-as-the-out-of-line-definition-of-class-templates-destructor/), which made all macOS tests fail (zero warnings expected).

I am setting this to critical with Milestone 3.2 because all tests will fail in macOS until this is fixed.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -201,27 +201,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest]
-        cpp_compiler: ["gcc", "clang"]
-        xNEST_BUILD_TYPE: ["MINIMAL", "MPI_ONLY", "OPENMP_ONLY", "FULL"]
-        exclude:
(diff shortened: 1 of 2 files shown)


heplesser:
I reduced to MINIMAL testing on macOS again. For some reason, Cmake does not find how to compile with OpenMP on the test runner, even though that should work with `-Xclang -fopenmp` compiler switches and the runner has the same Clang version as I do—all works locally.

jougs (approved):
LGTM! Many thanks for taking care!

hakonsbm (approved):
Looks good to me! However, I can also add that I found this stackoverflow post which seems to suggest that the warning is due to a defect in the wording of the standard:
https://stackoverflow.com/questions/68751682/is-a-class-templates-name-in-scope-for-a-qualified-out-of-line-destructors-def

PR #2232, opened 2021-12-07
Format error messages in `jonke_synapse` and `stdp_dopamine_synapse`

### LINKED ISSUE #2044: Check if synapse parameters can be set during connection
There are, after #2043 is merged, 4 synapse models that use `check_synapse_params` when connecting. These are: `cont_delay_synapse`, `jonke_synapse`, `static_synapse_hom_w` and `stdp_dopamine_synapse`.

`check_synapse_params` checks if a parameter can be set while connecting, or if it needs to be set before with `SetDefaults`. For some of the models, we still might not be able to set the parameters when connecting, but we should go through all and see if some of the parameters can actually be set during connection, like was the case for `quantal_stp_synapse` (see #2043).

Some of the error messages in the functions should at any rate be rewritten, see for instance `stdp_dopamine_synapse`, where the error message will not give the parameter that cannot be set, it will just display `Connect doesn't support the setting of parameter param_arr[ n ] in stdp_dopamine_synapse. Use SetDefaults() or CopyModel().`
https://github.com/nest/nest-simulator/blob/dafe70e88070a03e18f444e5a02297dfbf6b0cb3/models/stdp_dopamine_synapse.h#L398-L409

See also #2025.

stinebuu:
This PR fixes #2044.

As far as I can see, the parameters still need to be set with `SetDefaults()` or `CopyModel()`, because they are part of their own class containing the common properties (`JonkeCommonProperties` and `STDPDopaCommonProperties`). This PR thus only change the formatting of the error messages.

### DIFF
--- a/models/jonke_synapse.h
+++ b/models/jonke_synapse.h
@@ -420,9 +420,9 @@ jonke_synapse< targetidentifierT >::check_synapse_params( const DictionaryDatum&
   {
     if ( syn_spec->known( param_arr[ n ] ) )
     {
-      throw NotImplemented(
-        "Connect doesn't support the setting of parameter param_arr[ n ]"
-        "in jonke_synapse. Use SetDefaults() or CopyModel()." );
+      std::string msg = "Connect doesn't support the setting of parameter " + param_arr[ n ]
+        + " in jonke_synapse. Use SetDefaults() or CopyModel().";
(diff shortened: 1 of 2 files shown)


heplesser:
Since this is a minimal formatting change, I merge without second review.

PR #2238, opened 2021-12-09
Update links for live media image

jessica-mitchell:
This PR updates the links for the ova files. 
Older versions are removed and links to 2.20.2, 3.1 and latest (will be 3.2) are now available.

Reasoning is that we can always replace the file to nest-latest.ova to the most recent release without having to update the link.
We also wanted to keep only a couple of older versions, as we have capacity limits, and we want the versions that have the important fixes in the code. This means that now only these versions are available as ova.


steffengraber (approved):
:+1:

PR #2244, opened 2021-12-12
Fix quirks with import machinery in Python 3.8

### LINKED ISSUE #2241: Lazy loading modules for PyNEST
On initializing (`import nest`) nest doesn't import modules like `spatial`, `visualization`, `raster_plot` and `voltage trace`.
Instead it uses lazy loading approach but the first call of one of these modules shows `AttributeError`.

To solve this issue I added the lines to re-import called module in `/pynest/nest/__init__.py` (#2239, line 470).
I think this solution is not optimal. Maybe anyone can find a better approach.

1. Just call `nest.spatial` in Python Interpreter.

It raises `AttributeError` which disturbs the construction approach.

image

### LINKED ISSUE #2242: Start uwsgi with nest.server failed
Since `server` module is not imported in `pynest/nest/__init__.py` the command `nest-server` failed to start the server instance.
Thus I re-implemented importing `server` module -> #2239 

But it is not a long-term solution, I am open for alternative method for this issue when nest-server is working.

Helveg:
Apparently `importlib.import_module` sets an attribute on the parent module in python 3.8 (no longer does this in 3.9 it seems, or it was a bug that was fixed?), this didn't play nice with #2220.  On top of that, the error of `_setattr_error` during the `importlib.import_module` was silenced but would lead to an `AttributeError` for `nest.<lazy_module>` even though by all means the attribute was declared, likely because `_lazy_module_property` uses `delattr` right before `import_module` which would then error out, dissapearing the attribute.

I also added `nest.server` as a lazy loaded module.

closes #2241, closes #2242, closes #2239

### DIFF
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -416,19 +416,20 @@ def _setattr_error(self, attr, val):
     `__dict__` manipulation. It is added onto the module at the end of `__init__`,
     "freezing" the module.
     """
-    # Allow import machinery to set imported modules on `nest`
     if isinstance(val, types.ModuleType):
-        return super(type(self), self).__setattr__(attr, val)
-    err = AttributeError(f"can't set attribute '{attr}' on module 'nest'")
-    try:
-        cls_attr = getattr(type(self), attr)
-    except AttributeError:
-        raise err from None
+        # Allow import machinery to set imported modules on `nest`
+        self.__dict__[attr] = val
     else:
-        if hasattr(cls_attr, "__set__"):
-            cls_attr.__set__(self, val)
+        err = AttributeError(f"can't set attribute '{attr}' on module 'nest'")
+        try:
+            cls_attr = getattr(type(self), attr)
+        except AttributeError:
+            raise err from None
         else:
-            raise err
+            if hasattr(cls_attr, "__set__"):
+                cls_attr.__set__(self, val)
+            else:
+                raise err
 
 
 def _rel_import_star(module, import_module_name):
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -429,7 +429,7 @@ def _setattr_error(self, attr, val):
             if hasattr(cls_attr, "__set__"):
                 cls_attr.__set__(self, val)
             else:
-                raise err
+                raise err from None
 
 
 def _rel_import_star(module, import_module_name):
--- a/pynest/nest/__init__.py
+++ b/pynest/nest/__init__.py
@@ -93,8 +93,9 @@ class NestModule(types.ModuleType):
         _rel_import_star(self, ".lib.hl_api_types")
 
         # Lazy loaded modules. They are descriptors, so add them to the type object
-        type(self).spatial = _lazy_module_property("spatial")
         type(self).raster_plot = _lazy_module_property("raster_plot")
+        type(self).server = _lazy_module_property("server")
+        type(self).spatial = _lazy_module_property("spatial")
         type(self).visualization = _lazy_module_property("visualization")
         type(self).voltage_trace = _lazy_module_property("voltage_trace")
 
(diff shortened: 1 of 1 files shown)


babsey:
Thank you for fast fixing.

I tested nest-server on commit `6361af6` but I occur another problem to start `nest-server` using lazy importing.

```
spreizer@Latitude-5300:~$ uwsgi --module nest.server:app -s 5000
*** Starting uWSGI 2.0.19.1 (64bit) on [Sun Dec 12 16:17:46 2021] ***
compiled with version: 9.3.0 on 14 May 2021 13:30:03
os: Linux-5.11.0-41-generic #45~20.04.1-Ubuntu SMP Wed Nov 10 10:20:10 UTC 2021
nodename: Latitude-5300
machine: x86_64
clock source: unix
detected number of CPU cores: 8
current working directory: /home/spreizer
detected binary path: /usr/local/bin/uwsgi
!!! no internal routing support, rebuild with pcre support !!!
*** WARNING: you are running uWSGI without its master process manager ***
your processes number limit is 62736
your memory page size is 4096 bytes
detected max file descriptor number: 1024
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uwsgi socket 0 bound to UNIX address 5000 fd 3
Python version: 3.8.10 (default, Sep 28 2021, 16:10:42)  [GCC 9.3.0]
*** Python threads support is disabled. You can enable it with --enable-threads ***
Python main interpreter initialized at 0x555603fff880
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
mapped 72920 bytes (71 KB) for 1 cores
*** Operational MODE: single process ***

              -- N E S T --
  Copyright (C) 2004 The NEST Initiative

 Version: HEAD@6361af614
 Built: Dec 12 2021 16:04:57

 This program is provided AS IS and comes with
 NO WARRANTY. See the file LICENSE for details.

 Problems or suggestions?
   Visit https://www.nest-simulator.org

 Type 'nest.help()' to find out more about NEST.

***
*** WARNING: NEST Server runs without a RestrictedPython trusted environment.
***
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 979, in _find_and_load_unlocked
  File "/home/spreizer/opt/nest-server/lib/python3.8/site-packages/nest/__init__.py", line 421, in _setattr_error
    return super(type(self), self).__setattr__(attr, val)
AttributeError: can't set attribute
unable to load app 0 (mountpoint='') (callable not found or import error)
*** no app loaded. going in full dynamic mode ***
*** uWSGI is running in multiple interpreter mode ***
spawned uWSGI worker 1 (and the only) (pid: 147756, cores: 1)
```

@Helveg: Can you test it on Python3.8? If you also reproduce the same issue, what would be your solution?

babsey:
I confimed that the issue of `nest.spatial` etc. is fixed in Python 3.8.

Helveg:
Very unexpected behavior, `super().__setattr__` shouldn't error out, and it doesn't in 3.9 :( I switched the implementation to use `__dict__` to sidestep this, but doing so we stray 1 step further from the light ;p

babsey:
Great! NEST Server is working again.  :+1: :partying_face: 

I am confident with this pull instead of #2239. 
Thank you @Helveg!

babsey (approved):
LGTM!

jougs (approved):
This looks pretty clean to me. Many thanks for addressing the issue. I also tested that
```
mpirun --oversubscribe -np 4 install/bin/nest-server-mpi
```
works and it does. Thus I'm approving.

jougs:
@terhorstd: Given the criticality of this, can you please also have a quick look? Thanks!

babsey (approved):

jougs:
@steffengraber: I will merge this now. Can you please comment here, if there are any problems on Read the Docs caused by the merge? Thanks!

PR #2245, opened 2021-12-12
Fix CMAKE_INSTALL_PREFIX in switching between environments

med-ayssar:


Helveg:
Thanks for the ping! I don't fully understand why the prefix is being set to `"${CMAKE_CURRENT_BINARY_DIR}/nest_install"`? The decided on behavior was that the user either has to be in a Python virtual env, or HAS to explicitly set `CMAKE_INSTALL_PREFIX`, which would indicate that they understood the consequences of the install prefix. With this PR we'd force everyone, even expert users to install into a virtual env, wouldn't we?

med-ayssar:
The `unset` command in cmake doesn't seem to do the trick by removing the `CMAKE_INSTALL_PREFIX` from the `CMAKECache.txt`. So, either the user has to delete the cache file or rerun `cmake -DCMAKE_INSTALL_PREFIX=update_path` to force changing the value of `CMAKE_INSTALL_PREFIX` in the cache.

With the current update, the following might happen:

Assuming you run `cmake` in `/path/to/build` and without specifying the `CMAKE_INSTALL_PREFIX`. 

- First case: Running `cmake` not in a virtual env will force the `cmake` to set the  `CMAKE_INSTALL_PREFIX` to `/path/to/build/nest_install` (avoid `sudo` permission).

- Second case: Running `cmake` in a virtual env behaves like the first case.

- Third Case: Running `cmake` not in a virtual env, and assuming there were some errors, and you rerun `cmake` in a new virtual env, then the build will always continue happening at the same location. 

The user has then to explicitly set `CMAKE_INSTALL_PREFIX`, or the `cmake` built library will be generated in the subdirectory under the current folder where the `cmake` is being executed.

However, a more simple solution can be done by deleting the `CMakeCache.txt` on each run of `cmake`.

@jougs what do you think?

Helveg:
Deleting the `CMakeCache.txt` might be a bit violent, it will also clear any other flags that the user had set. Maybe it can be modified though so that all things `CMAKE_INSTALL_PREFIX` are fresh again?

heplesser:
What precisely is the problem you want to address here? My approach with fail cmakes is to delete everything in the build directory (or the entire build directory) and start with a clean slate. Instead of trying to create complicated constructions that will work under all circumstances, maybe we could just add "clean up an try again" recommendations to the documentation?

Helveg:
New users often run into installation problems by using their system Python, so it was decided to error out when they don't use virtual envs. To allow expert users to retain the same level of control over the installation setting a `CMAKE_INSTALL_PREFIX` overrides this safeguard.

But, when a user runs into the safeguard, and appropriately switches to a virtual env, the stale cache keeps the incorrect prefix that we just warned them not to use. It doesn't seem right to expect new users who just want a smooth copy-paste command install to understand how CMake or its cache works.

heplesser:
@Helveg Thanks for the explanation. But instructions like "If you run into trouble, delete the entire build directory, create it again and try again with correct options" seems relatively straightforward to me and does not require users to understand much of CMake. I mean this just as a fallback in case an "automagic" solution should be too difficult or fickle.

Helveg:
Right, understood :) I think a simple, not so fickle, solution might still be possible if we unset `CMAKE_INSTALL_PREFIX` when the error occurs. I believe `GNUInstallDirs` is called after we process the Python options, so no derivatives of `CMAKE_INSTALL_PREFIX` should be made anywhere yet.

Helveg:
After some more discussion it might be an added maintenance burden to fight against the CMake cache, and instructions to clean the build dir in the error, plus a note in the install docs about CMake caches are preferred.

PR #2248, opened 2021-12-16
Fix path to connection diagram for HPC benchmark example

steffengraber:
The used path works only in the local preview of the HTML files. 
The path inside the generated `.rst` file must point to the image in the directory `userdoc/examples`, only then it will be moved to the correct path by make html.


jarsi:
Thank you @steffengraber for working out the path magic. It indeed fixes the appearance of the figure, so it could be merged.

PR #2250, opened 2021-12-29
Version shared objects

sanjayankur31:
While updating PyNN in Fedora, I remembered that it needs to link the NEST extensions it uses to various libraries that NEST provides in `/usr/lib{,64}/nest`. Previously, I've treated these NEST libraries as "internal" libraries in Fedora and carried them in without any soversions. However, given that they can be used to develop extensions against, they're not technically private libraries. 

If this is the case, I guess the current best practice is to version these libraries, as explained here:

https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

This pull request adds the necessary CMake bits to do this. I've set the "build version" (`VERSION`) to `NEST_VERSION` and the ABI version (`SOVERSION`) to 3 (could be automated to extract the major version from `NEST_VERSION`, but that's not necessary since release version and ABI version do not need to be the same):

https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html?highlight=soversion


sanjayankur31:
Note that if the libraries are not private, they also do not need to be installed into a `nest` subfolder in `LIBDIR` (and for normal usage, this means no RPATH is required since ld will be able to find the shared objects in standard locations)

jougs:
@sanjayankur31: Sorry for the long delay. Can you please merge `master` again to trigger a re-run of the CI? Thanks!

sanjayankur31:
> @sanjayankur31: Sorry for the long delay. Can you please merge `master` again to trigger a re-run of the CI? Thanks!

Done :)

jougs (approved):
Looks good to me.

One question: does this require additional changes in [NESTML](https://github.com/nest/nestml) and/or the [NEST extension module repository](https://github.com/nest/nest-extension.module) to make sure those are only built against the correct version of the NEST libraries?

sanjayankur31:
> Looks good to me.
> 
> One question: does this require additional changes in [NESTML](https://github.com/nest/nestml) and/or the [NEST extension module repository](https://github.com/nest/nest-extension.module) to make sure those are only built against the correct version of the NEST libraries?

NESTML does not link against the shared objects at all, so I don't think it needs any tweaking to build/run. It will need to track PyNEST if that's what it uses using the python versioning system etc. (But I haven't looked into NESTML in detail, so this is based on a quick look at setup.py). 

The extension modules seem to link against the shared objects. So nothing needs to be changed in their compilation/build process, but once they've been built, they will need the same version of the shared objects to run because the linker will now look for the versioned shared objects. Each time the NEST shared object soversion changes in a backwards incompatible way, the extension modules will need to be re-built to run (in the same way as we need to recompile bits when we update to a new version of gcc or boost and whatnot).

jougs:
So in short, this means that modules are now safer, as they can't accidentally dlopen the wrong version of the library? Thanks!

sanjayankur31:
> So in short, this means that modules are now safer, as they can't accidentally dlopen the wrong version of the library? Thanks!

I'll have to check to see if dlopen allows opening arbitrary library versions but the system linker (`ld`?) will certainly not open a different version that the one it needs.

gtrensch (approved):
Looks good to me. :+1:

PR #2257, opened 2022-01-10
More detailed explanation of why C++ ADL requires nest::apply()

heplesser:
Extended the comment based on information given in https://github.com/llvm/llvm-project/issues/53084.


heplesser:
Since this is a minimal change of a technical comment, I merge without review.

PR #2260, opened 2022-01-12
Automatically adjust model defaults upon changes to resolution

jougs:
Previous to this PR, the four models `correlation_detector`, `correlomatrix_detector`, `correlospinmatrix_detector` and `noise_generator` contained parameters as `nest::Time` objects that were not properly updated when the resolution was re-set, which could lead to problems upon instantiation of these models when a non-default resolution was set. This PR adds automatic conversions to the corresponding variables to avoid such problems.

The general old behavior was like this:
* a parameter was initialized to the resolution or a fixed value in ms in the default constructor
* the copy constructor of `Parameters` copied the values from the original node instance
* node constructors threw an exception if the value was incompatible with the current resolution.
 
The new behavior is like this:
* a parameter is initialized to the resolution or a fixed value in ms in the default constructor
* the copy constructor of `Parameters` and `calibrate_time()` update the value to a reasonable value based on the newly set resolution *if* the stored value becomes incompatible 
* the node constructors don't throw exceptions anymore upon incompatible values as those are impossible now.

The parameter `delta_tau_` in the `correlation_detector` was changed to `5 * Time::get_resolution()` from the previous value of `Time(1.0)` in the default constructor to fulfill the condition "odd multiple of the resolution" that is mentioned in the documentation.

The changes are explained in a user-understandable fashion in the release notes for NEST 3.3.

### DIFF
--- a/doc/userdoc/release_notes/v3.3/index.rst
+++ b/doc/userdoc/release_notes/v3.3/index.rst
@@ -1,0 +1,28 @@
+All about NEST 3.3
+==================
+
+This page contains a summary of all breaking and non-breaking changes
+from NEST 3.2 to NEST 3.3. In addition to the `auto-generated release
+notes on GitHub <https://github.com/nest/nest-simulator/releases/>`_,
+this page also contains transition information that helps you to
+update your simulation scripts when you come from an older version of
+NEST.
+
+If you transition from a version earlier than 3.0, please see our
+selection of earlier :doc:`transition guides <release_notes/index>`.
+
+.. contents:: On this page you'll find
+   :local:
+   :depth: 1
+
+Model defaults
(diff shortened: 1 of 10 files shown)


heplesser:
@jougs I commented on #2208, I believe the changes here are the same as there? Would it have made more sense to comment here rather than in there?

jougs:
Yes, the changes are the same and the comments should indeed rather go here. As these changes are independent of those in #2208, I thought it would be easier to create a smaller PR that can be reviewed separately...

jougs:
Alternatively, I can also close this one and we're solely considering #2208.

jougs:
Further discussion will take place in #2208.

PR #2261, opened 2022-01-14
Fix issues encountered by the Fujitsu compiler

### LINKED ISSUE #2207: Refactor variable initialization to avoid C++17 feature for type deduction
In `stimulation_backend_mpi.cpp` we define and initialize a variable `data` like this:
https://github.com/nest/nest-simulator/blob/3da76ddaed8fbb6fde81e9a8241641c7df512a8f/nestkernel/stimulation_backend_mpi.cpp#L206

This should only work with C++17, where the type of an expression like `auto x{3}` deduces to `int` (see [N3922](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3922.html)). With C++11 the same expression deduces to `std::initializer_list<int>`. So, with C++11 the type for `data` in the above line deduces to a `std::initializer_list`, which causes trouble when we try to use it as a C-style array of pointers (the C-style array should also be changed to something more robust).

With Intel's C++ compiler the compilation fails, but for some reason we don't get the same problem when compiling with GCC or Clang.

hakonsbm:
This PR fixes issues that surfaced when compiling with the Fujitsu compiler on Fugaku.
- The CMake configuration and C++ includes are updated for the case where compiling and running the Random123 library test fails during configuration.
- A C-style array in stimulation_backend_mpi.cpp is replaced with a more robust `std::vector`, which fixes #2207.
- Removes the template qualifier added with #2231, and adds the warning produced by the missing template qualifier to expected warnings.


cmake/CheckExtraCompilerFeatures.cmake
```diff
@@ -259,7 +259,11 @@ function( NEST_CHECK_RANDOM123 )
   if ( ${COMPILE_RESULT} AND ${RUN_RESULT} EQUAL 0 )
     set( HAVE_RANDOM123 ON )
   else ()
-    message( ${RUN_OUTPUT} )
+    if ("${RUN_OUTPUT}" STREQUAL "")
+      message( "Test output is empty." )
```
heplesser, line 263: Why do we need this message? Would it not be better to call `message()` only if `RUN_OUTPUT` is not empty?

heplesser (commented):
Looks good, just got one question, see below.

jougs (approved):
LGTM!

cmake/CheckExtraCompilerFeatures.cmake
hakonsbm: `RUN_OUTPUT` is empty only when the compilation of the Random123 tests failed, so I've added a message for that instead.

heplesser (approved):

PR #2262, opened 2022-01-14
Make `plot_weight_matrices.py` run

### LINKED ISSUE #2138: `plot_weight_matrices.py` should be rewritten
**Describe the bug**
The PyNEST example [plot_weight_matrices.py](https://github.com/nest/nest-simulator/blob/master/pynest/examples/plot_weight_matrices.py) _only_ contains a function for plotting weight matrices, you have to provide a network yourself. 

https://github.com/nest/nest-simulator/blob/bc5cc6268d42f3c8c829f1cc92ea762832e91449/pynest/examples/plot_weight_matrices.py#L59-L63

When providing a network, the script fails with the following error message
```
/NEST/nest-simulator/pynest/examples/plot_weight_matrices.py in plot_weight_matrices(E_neurons, I_neurons)
    105              a_EI_trg[idx] - min(E_neurons)] += c_EI[idx]
    106     for idx in range(len(a_IE)):
--> 107         W_IE[a_IE_src[idx] - min(E_neurons),
    108              a_IE_trg[idx] - min(I_neurons)] += c_IE[idx]
    109     for idx in range(len(a_II)):

TypeError: '<' not supported between instances of 'NodeCollection' and 'NodeCollection'
```

As stated in this [NEST user e-mail](https://www.nest-simulator.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/VCA4SVGOD3IRTZL2EWSOD635C2RNUJFB/), changing to `min(E_neurons.tolist())` and `min(I_neurons.tolist())` should solve the problem.

The example should also be rewritten to include an actual network, and then calling the function. This way, we will detect problems with the example in the future when we run all examples, as the reason this was not detected until now was that the function in the example has not been called in a while.

stinebuu:
This PR fixes #2138.

### DIFF
--- a/pynest/examples/plot_weight_matrices.py
+++ b/pynest/examples/plot_weight_matrices.py
@@ -29,7 +29,6 @@ these values in weight matrices for further analysis and visualization.
 
 All connection types between these populations are considered, i.e.,
 four weight matrices are created and plotted.
-
 """
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


hakonsbm (approved):
Works and looks good :+1:

pnbabu (approved):
Looks good to me!

PR #2272, opened 2022-01-20
Fix deprecation warnings in testsuite

jougs:


jougs:
This is now part of #2273 and can be merged from there.

PR #2273, opened 2022-01-20
Clean up OpenMP pragmas

jougs:
This PR removes empty loops in case of single threading and generally cleans up and comments the OpenMP pragmas. On the way, i took the liberty to clean up some deprecation warnings that originated from the Python tests.


heplesser (approved):
Thanks for cleaning up!

testsuite/pytests/test_events.py
```diff
@@ -62,7 +62,7 @@ def test_EventsSpikes(self):
 
         d = nest.GetStatus(sr, 'events')[0]
 
-        self.assert_(len(d['times']) > 0)
+        self.assertTrue(len(d['times']) > 0)
```
hakonsbm, line 65:
```suggestion
        self.assertGreater(len(d['times']), 0)
```

hakonsbm (changes requested):
Thanks! I have just a tiny suggestion.

hakonsbm (approved):

PR #2274, opened 2022-01-21
Fix version warning

steffengraber:
On Read the docs the version specified as "latest" is in some places ambiguously pointing to "latest release" and "latest on main branch". This leads to some confusion when referring to "latest". Removing this from the code and relying only on the "show version warning" option on readthedocs, we can enable and disable this warning from the outside, without having a potentially wrong warning in a release that is not removable.


terhorstd (commented):
LGTM

Issue #2276, opened 2022-01-26
Inconsistent formatting since clang-format 9 merge


mlober:
**Describe the bug**
Found files with inconsistent formatting: 
```
MSGBLD0220: + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +
MSGBLD0220: +                 STATIC CODE ANALYSIS DETECTED PROBLEMS !                    +
MSGBLD0220: + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + + +

MSGBLD0220: C/C++ files with formatting errors:
MSGBLD0220: ... libnestutil/block_vector.h
MSGBLD0220: ... libnestutil/enum_bitfield.h
MSGBLD0220: ... libnestutil/lockptr.h
MSGBLD0220: ... libnestutil/sort.h
MSGBLD0220: ... nestkernel/connection_creator.cpp
MSGBLD0220: ... nestkernel/connection_manager.cpp
MSGBLD0220: ... nestkernel/event.h
MSGBLD0220: ... nestkernel/grid_layer.h
MSGBLD0220: ... nestkernel/nest_time.h
MSGBLD0220: ... nestkernel/node_collection.h
MSGBLD0220: ... nestkernel/ntree.h
MSGBLD0220: ... nestkernel/position.h
MSGBLD0220: ... nestkernel/stimulation_backend_mpi.cpp
MSGBLD0220: ... nestkernel/target_data.h
MSGBLD0220: ... sli/dict.cc
MSGBLD0220: ... sli/dict.h
MSGBLD0220: ... sli/tarrayobj.h
MSGBLD0220: ... sli/token.h
MSGBLD0220: ... sli/tokenarray.h
Error: Process completed with exit code 1.
```

**To Reproduce**
checkout to master and run clang-format on any of the files above. 
in terminal: `diff -u nestkernel/node_collection.h /dev/stdin <<<$(clang-format nestkernel/node_collection.h)`

**Expected behavior**
No difference when running clang-format on the (untouched) files above.

**Desktop/Environment (please complete the following information):**
 - OS Ubuntu 20.04
 - Python-Version: 3.9
 - NEST-Version: master
 - Installation: nest-simulator conda environment; apt-installed clang-format version 10 (although this is the wrong version, formatting for example nestkernel/node_collection.h with clang-format 10 fixed the issue for this file.)

**Additional context**
[...]

PR #2278, opened 2022-01-26
Update `operator+=` logic for composite NodeCollection

mlober:
The operator+= of a node collection composite was changed in order to avoid the usage of the very inefficient operator++ which was previously called by operator+=. Instead of using the operator++ to loop through a NCO composite until the desired index is reached, one can now directly "jump" to this index in the new operator+=. 
This change is necessary to avoid extremely large nest.connect-times when the source is a node collection composite object.


nestkernel/node_collection.h
```diff
@@ -717,9 +718,34 @@ nc_const_iterator::operator+=( const size_t n )
...
+      {
+        part_idx_ = composite_collection_->end_part_;
+        element_idx_ = composite_collection_->end_offset_;
+      }
+    }
+    else if ( part_idx_ >= composite_collection_->parts_.size() )
+    {
+      auto end_of_composite = composite_collection_->end();
+      part_idx_ = end_of_composite.part_idx_;
+      element_idx_ = end_of_composite.element_idx_;
```
hakonsbm, line 748: Except for the different steplength in line 721, this is a duplication of the code from `nc_const_iterator::operator++()`. I think it would therefore be better to put this code in a separate helper function (a private function within `nc_const_iterator`), perhaps with the steplength as argument, and call that function from both `operator++()` and `operator+=()`.

hakonsbm (changes requested):
Thanks for this improvement!

nestkernel/node_collection.h
mlober: That's true, it would make the code much nicer. Thanks! I will get to it in the next days.

mlober:
I now put the duplicated code into a separate function "cond_update_idx". 
The static checks fail due to formatting issues in functions that I have not touched. I reported this issue here: https://github.com/nest/nest-simulator/issues/2276

hakonsbm (changes requested):
Thanks for the update! I have a couple more suggestions.

hakonsbm:
@mlober There are some differences between different clang-format versions. We are enforcing formatting with clang-format 9 now, so you can try formatting node_collection.h with that and see if that fixes the problem.

hakonsbm (approved):
Thanks!

stinebuu (approved):
Looks good to me, thanks!

PR #2279, opened 2022-01-27
Re-add version number for clang-format

jougs:
This avoids problems when multiple versions are around and fixes #2276.

PR #2281, opened 2022-01-31
Fix GHA clang format issue

hakonsbm:
Fixes an issue with the static code checks where clang-format failed because version 9 was not installed (it is required since #2279). That led to every formatting check of C++ files failing with the following message:
```
./build_support/static_code_analysis.sh: line 90: clang-format-9: command not found
```
This PR adds clang-format-9 to system dependencies, which fixes the problem. Additionally, the static code analysis script aborts if any of the requested tools are not found.


heplesser (approved):
Thanks!

PR #2290, opened 2022-02-09
Accelerate node lookup in presence of many nodes without proxies

heplesser:
This PR revises the node-lookup heuristics of `SparseNodeArray` to provide fast lookup also in cases where many nodes without proxies (devices) are present. With this PR in place, good performance will be achieved if all devices are created either before or after all nodes with proxies.

The attached
[sra_demo.sli.gz](https://github.com/nest/nest-simulator/files/8030498/sra_demo.sli.gz) script illustrates this effect. Call it as
```
nest --userargs=num_sg,indegree,num_threads
```
The script creates 100.000 neurons and `num_sg` spike generators. It then connects neurons to neurons and generators to neurons with the given fixed `indegree`, multapses permitted, running on `num_threads`. The interesting point is to see performance differences for small (eg 1) and large (eg 25.000) `num_sg` for serial and thread-parallel cases. Current master (cc15d0590) performs very poorly on multiple threads, much worse than serial, because the linear lookup heuristics becomes practically useless when nodes distributed across processes are mixed with a large number of non-distributed devices. This PR (split-sparse-array@a9cc70dce) notes where in the SparseNodeArray the boundary between normal nodes and devices is and uses separate linear lookup in each part. To keep complexity within bounds, it only allows a single boundary, so neurons and devices should be created en bloc, not intermingled.

Benchmark results for indegree==100 from my laptop (macOS 12.2, 2 GHz Intel Core i5, 4 cores, Clang 13), where the rightmost column is most important:
 
|  | | 1 SG, 1 Thread | 1 SG 4 Threads | 25k SGs, 1 Thread | 25k SGs, 4 Threads |
|--|--|--:|--:|--:|--:|
|Master| N->N| 2.1s | 0.4s | 2.1s |18.3s|
|           | G->N| 0.8s |  0.3s | 4.7s|21.4s|
|This PR|N->N|2.6s |0.4s |2.1s |0.5s |
|            |G->N| 1.1s| 0.3s| 4.7s|2.1s | 

To do:
- Review wether to keep adjustment of scale based on lookup experience
- Review field names
- Add user documentation pointing out the need to not intermingle creation of neurons and devices


nestkernel/sparse_node_array.cpp
```diff
@@ -24,81 +24,114 @@
 
 // Includes from nestkernel:
 #include "exceptions.h"
+#include "kernel_manager.h"
 #include "node.h"
+#include "vp_manager_impl.h"
+
+
+#include <iomanip>
+#include <iostream>
```
stinebuu, line 33: Are these two used? I might be overlooking something. If they are used, they should have `// C++ includes:` as a header and be placed over `// Includes from nestkernel:`

nestkernel/sparse_node_array.cpp
```diff
@@ -24,81 +24,114 @@
...
 
-  // node_id must exceed max_node_id_
-  assert( node_id > max_node_id_ );
+  // mark array inconsistent until set_max_node_id() called
+  global_max_node_id_ = 0;
 
-  // all is consistent, register node and update auxiliary variables
-  nodes_.push_back( NodeEntry( node, node_id ) );
-  if ( local_min_node_id_ == 0 ) // only first non-zero
+  // Setup when first node is added
```
stinebuu, line 85: All other comments here start with a lower case
```suggestion
  // setup when first node is added
```

nestkernel/sparse_node_array.cpp
```diff
@@ -24,81 +24,114 @@
...
+  if ( local_min_node_id_ == 0 )
   {
     local_min_node_id_ = node_id;
+    left_side_has_proxies_ = node.has_proxies();
+
+    // We now know which scale applies on which side of the split
```
stinebuu, line 91: Same here
```suggestion
    // we now know which scale applies on which side of the split
```

stinebuu (changes requested):
Looks good! I mostly have comments about comments.

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ *
+ * 1. Entries are sorted by strictly increasing node ID (nid).
+ * 2. All entries with index i < split_idx_ belong to the left part of the array,
+ *    all remaining entries to the right part.
+ * 3. All entries with node ID nid < split_node_id_ belong to the left part of the array,
+ *    all remaining entries to the right part.
+ * 4. nodes_[0].get_node()->has_proxies() == nodes_[i].get_node()->has_proxies() for 0 <= i < lookup_split_idx_
+ *
+ * @note
+ * - Invariant 5 simply means that all nodes in the left part of the array have the same value of has_proxies().
```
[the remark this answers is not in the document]
heplesser: Yes.

pynest/nest/lib/hl_api_nodes.py
```diff
@@ -48,6 +48,12 @@ def Create(model, n=1, params=None, positions=None):
     given, a single node is created. Note that if setting parameters of the
     nodes fail, the nodes will still have been created.
 
+    Note
+    ----
+    During network construction, create all nodes representing model neurons first, then all nodes
+    representing devices (generators, recorders or meters). Otherwise, network connection can be
```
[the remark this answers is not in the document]
heplesser: Yes, fixed.

nestkernel/sparse_node_array.cpp
heplesser: OK.

nestkernel/sparse_node_array.cpp
heplesser: OK.

heplesser:
@stinebuu Thanks for the review, all should be fixed now.

BTW, I created a branch https://github.com/heplesser/nest-simulator/tree/split-sparse-array-with-tracing with a version that prints additional tracing info (steps taken in search) when NEST terminates. That is not to be merged, just for exploration/debugging.

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
- * This class is a container providing lookup of local nodes (as Node*)
- * based on node IDs.
+ * SparseNodeArray maps node IDs of thread-local nodes to Node*. It provides
+ * a const iterator interface for iteration over all local nodes, lookup by
+ * node ID and lookup by numeric index into local nodes. The latter is provided
+ * to support HPC synapses using TargetIdentifierIndex representation.
  *
- * Basically, this array is a vector containing only pointers to local nodes.
- * For M MPI processes, we have
+ * For efficient lookup, all normal nodes (with proxies) and all devices need
```
jessica-mitchell, line 49: Does normal nodes mean nodes that represent neurons?

```suggestion
 * For efficient lookup, all normal nodes (with proxies) need to be created together 
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ * to be created en bloc. It does not matter which are created first. If nodes
```
jessica-mitchell, line 50:
```suggestion
 * and all devices need to be created together. It does not matter which are created first. If nodes
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
- *   node ID  %  M  --> rank
- *   node ID div M  --> index on rank
+ * As nodes are added to an initially empty SparseNodeArray, the array tracks
+ * whether nodes have proxies or not. Array lookup is split at the point of the first
+ * proxy/no-proxy (or no-proxy/proxy) transition between nodes. In the no-proxy
+ * region, where all nodes are represented locally, Node IDs are mapped directly
+ * to array indices. In the proxy region, they are scaled by 1/n_vp.
+ *
+ * To reliably reject requests for node IDs beyond the globally maximal node ID, the
+ * latter must be set explicit. A SparseNodeArray is said to be in *consistent state*
```
jessica-mitchell, line 61:
```suggestion
 * latter must be set explicitly. A SparseNodeArray is said to be in *consistent state*
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ * if the global maximal node ID has been set. Once add_local_node() is called, the
+ * array is not in consistent state until the global maximal node ID is set again. This
```
jessica-mitchell, line 63:
```suggestion
 * array is not in a consistent state until the global maximal node ID is set again. This
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ * is indicated by setting the max_node_id_ == 0. Looking up nodes while the
+ * array is not in consistent state triggers an assertion.
```
jessica-mitchell, line 65:
```suggestion
 * array is not in a consistent state triggers an assertion.
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ *
+ * To also support cases in which users interleave creation of nodes with and
+ * without proxies or use nodes with special behavior (music nodes, e.g.), we
```
jessica-mitchell, line 68:
```suggestion
 * without proxies or use nodes with special behavior (e.g., MUSIC nodes), we
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ * to array indices. In the proxy region, they are scaled by 1/n_vp.
+ *
+ * To reliably reject requests for node IDs beyond the globally maximal node ID, the
+ * latter must be set explicit. A SparseNodeArray is said to be in *consistent state*
+ * if the global maximal node ID has been set. Once add_local_node() is called, the
+ * array is not in consistent state until the global maximal node ID is set again. This
+ * is indicated by setting the max_node_id_ == 0. Looking up nodes while the
+ * array is not in consistent state triggers an assertion.
+ *
+ * To also support cases in which users interleave creation of nodes with and
```
jessica-mitchell, line 67:
```suggestion
 * To also support cases in which users alternate creation of nodes with and
```

nestkernel/sparse_node_array.h
```diff
@@ -39,52 +39,100 @@ namespace nest
...
+ * without proxies or use nodes with special behavior (music nodes, e.g.), we
+ * perform a linear search from the estimated location of the node in the array.
+ *
+ * The following invariants hold when the array is in consistent state:
```
jessica-mitchell, line 71: Is the term 'invariant' fairly well understood and used among developers? Would something like "the following points hold true / remain unchanged" be more clear?

pynest/nest/lib/hl_api_nodes.py
```diff
@@ -48,6 +48,13 @@ def Create(model, n=1, params=None, positions=None):
     given, a single node is created. Note that if setting parameters of the
     nodes fail, the nodes will still have been created.
 
+    Note
+    ----
+    During network construction, create all nodes representing model neurons first, then all nodes
+    representing devices (generators, recorders or meters), or all devices first and then all neurons.
```
jessica-mitchell, line 54:
```suggestion
    representing devices (generators, recorders, or detectors), or all devices first and then all neurons.
```

jessica-mitchell (changes requested):
I have a few minor suggestions, but looks good.

nestkernel/sparse_node_array.h
heplesser: Yes, "normal" means (with minimal exceptions) nodes representing neurons.

nestkernel/sparse_node_array.h
heplesser: Yes, invariants are standard terminology in algorithm analysis.

jessica-mitchell: ok thanks

jessica-mitchell (approved):
Thanks, I think there is still one outstanding suggestion (an 'a' in front of consistent) that needs to be resolved, but I'll approve as it's very minor.

stinebuu (approved):
Looks good to me, thanks!

mlober:
I tested the example script and can confirm the acceleration of this PR compared to the master version.

mlober (approved):

Issue #2017, opened 2021-04-21
Replacing deprecated model `iaf_psc_alpha_canon` with `iaf_psc_alpha_ps` leads to significant time increase


stinebuu:
`iaf_psc_alpha_canon` is deprecated, and should thus be removed. It can be replaced with `iaf_psc_alpha_ps`

https://github.com/nest/nest-simulator/blob/f7562a6d457adc83559395839e09b178b03de054/pynest/nest/lib/hl_api_helper.py#L70-L73

`iaf_psc_alpha_canon` is not used much in the code, it is used in one test and two examples, so it's not difficult to remove it. However, when trying to remove the model, I noticed that the two examples (`ArtificialSynchrony.sli` and `brunel_ps.sli`) get quite slow.

With `iaf_psc_alpha_canon` `ArtificialSynchrony.sli` took about 9 minutes to complete, with `iaf_psc_alpha_ps` it took about 18 minutes.
For `brunel_ps.sli` the simulation time with  `iaf_psc_alpha_canon` is 185.63 sec on my computer, but with `iaf_psc_alpha_ps` simulation time is 490.83 sec.

I wanted to create an issue where we can discuss this a bit before actually removing the model, as this is a significant time difference.

heplesser:
@stinebuu Interesting! From a quick look at the code, there are two potential sources of this difference: alpha_ps allows different time constants for E and I synapses, canon has only one. This requires twice as many exponential functions to be evaluated. The other point is that alpha_ps determines threshold crossings precisely with regula falsi, while alpha_ps by default uses linear interpolation.

Could you run one of the examples with a profiling tool to see what takes most time?

I would still suggest to remove _canon, but wonder what @abigailm and @diesmann think?

stinebuu:
@heplesser I will have a look with a profiling tool, yes!

Maybe interesting to also note that you can set the interpolation order for `iaf_psc_alpha_canon`, and for `ArtificialSynchrony.sli` the model use interpolation order 3.

stinebuu:
@heplesser After having a look on VTune it looks like it is indeed the many exponential functions. It is mostly the exponentials in the propagators, `propagator_31` and `propagator_32` that contribute to the time increase.
Just as a test I replaced the exponent functions in the propagators with `std::expm1() + 1`, and this decreased the time a bit (`brunel_ps.sli` then use 325.82).

heplesser:
@stinebuu Thanks for testing this!

[...]

PR #2294, opened 2022-02-15
Remove deprecated models iaf_psc_alpha_canon and pp_pop_psc_delta

jougs:
Here's a small and hopefully non-controversial one removing deprecated models. I also cleaned up the handling of connection model flags a bit to get shorter and more readable lines.


examples/nest/brunel_ps.sli
```diff
@@ -196,10 +196,10 @@ ResetKernel      % clear all existing network elements
 tic % start timer on construction
 
   (Creating excitatory population.) =  % show message
-  /E_neurons /iaf_psc_alpha_canon NE Create def  % create exitatory neurons
+  /E_neurons /iaf_psc_alpha_ps NE Create def  % create exitatory neurons
 
   (Creating inhibitory population.) =  % show message
-  /I_neurons /iaf_psc_alpha_canon NI Create def  % create inhibitory neurons
+  /I_neurons /iaf_psc_alpha_ps NI Create def  % create inhibitory neurons
```
stinebuu, line 202: You also need to change `tau_syn` in line 240.

examples/nest/ArtificialSynchrony.sli
```diff
@@ -163,7 +161,7 @@ userdict begin
       /overwrite_files true          % overwrite previous output
   >> SetKernelStatus
 
-  sim 1 eq { /neurons /iaf_psc_alpha_canon nr Create def }
+  sim 1 eq { /neurons /iaf_psc_alpha_ps nr Create def }
```
stinebuu, line 164: This example is not running for me, you probably need to change a `tau_syn` somewhere!

stinebuu (changes requested):
Looks good to me, but I tested the two examples and they do not run for me. I think you need to change `tau_syn` to `tau_syn_{ex, in}` somewhere.

examples/nest/brunel_ps.sli
jougs: Done in fc1357d

examples/nest/ArtificialSynchrony.sli
jougs: Done in fc1357d

stinebuu (approved):
Looks good to me! The examples take a bit more time to run (`brunel_ps` have a simulation time of 267.26 s with `iaf_psc_alpha_ps` and 94.99 s with the master version, likewise `ArtificialSyncrony` is also slower with the precise model), but I guess this is to be expected with a precise model? See also issue #2017 (the times there are longer, if I remember correctly I ran the examples during a hackathon, and had video running simultaneously, slowing everything down).

models/modelsmodule.cpp
```diff
@@ -309,8 +303,8 @@ ModelsModule::init( SLIInterpreter* )
 
   // register all synapse models
   register_connection_model< bernoulli_synapse >( "bernoulli_synapse" );
-  register_connection_model< clopath_synapse >(
-    "clopath_synapse", default_connection_model_flags | RegisterConnectionModelFlags::REQUIRES_CLOPATH_ARCHIVING );
+  auto flags = default_connection_model_flags | RegisterConnectionModelFlags::REQUIRES_CLOPATH_ARCHIVING;
```
clinssen, line 306: Not a fan of introducing a temporary variable here to hold the flags, but if you prefer it this way, could you move the declaration to the top of the method?

clinssen (approved):
Approving on the condition that we are OK with the runtime performance regression. I wasn't privy to the discussion about why these models were marked as deprecated in the first place.

jougs:
@stinebuu: many thanks for measuring execution times. I'm actually not sure if this slow-down is to be expected, as the deprecated (and now removed) model `iaf_psc_alpha_canon` was also a precise variant.

When replacing `iaf_psc_alpha_canon` with `iaf_psc_alpha_ps` I just followed [the documentation of the former](https://github.com/nest/nest-simulator/blob/6b9acf4bb056764fe1906328d7c9b37266fa456f/models/iaf_psc_alpha_canon.h#L56-L59).

@suku248 can you please quickly comment on this?

models/modelsmodule.cpp
jougs: I might indeed have gone a bit too far here and have reverted to the original code.

suku248:
Regarding @stinebuu 's observation of slower runtimes when using `iaf_psc_alpha_ps` instead of the deprecated `iaf_psc_alpha_canon`: The implementations differ in the way they determine the precise location of a threshold crossing. While `iaf_psc_alpha_canon` used an interpolation method (precision depended on simulation resolution), `iaf_psc_alpha_ps` uses an iterative method (no dependence on sim res). [Hanuschkin et al. (2010)](https://doi.org/10.3389/fninf.2010.00113) did not observe any significant differences in runtime when comparing the two implementations for the IAF neuron model with exponential PSCs (see Fig. 5). Therefore, I would not have expected to see a substantial slowdown here. Maybe the iterative method doesn't converge as fast as it should be?

jougs:
@suku248, many thanks for these insights.

I'm really not sure how to proceed here. @heplesser, @abigailm, @diesmann: any thoughts on this? Is there anyone currently working on the precise models who could investigate further?

diesmann:
Thanks HEP. Whether _ps or _canon are faster depends on the computation time step and the requested accuracy. The general observation is that for a given accuracy _ps is faster than _canon because _ps can reach this accuracy already at larger computation time steps. With a fixed computation time step and a low required accuracy _canon might be faster because there are fewer operations to be carried out. The Hanuschkin paper has the details on this relationship.

stinebuu:
I have been doing some digging. It seams like the problem is the use of `propagator_31` and `propagator_32` in [`iaf_psc_alpha_ps`' `propogate_` function](https://github.com/nest/nest-simulator/blob/master/models/iaf_psc_alpha_ps.cpp#L497). In canon, the propogate variables are calculated directly, while ps use `propagator_31` and `propagator_32`.

When I use canon's way of doing things in the ps model (see the change [here](https://github.com/nest/nest-simulator/compare/master...stinebuu:iaf_psc_alpha_ps?expand=1)) I get the following times:

| type | time |
|---------------------------------|--------------|
| `brunel_ps.sli` with ps model   | 101.99 sec   |
| `brunel_ps.sli with canon       | 93.96 sec    |
| `ArtificialSyncrony` with ps    | 5 min 59 sec |
| `ArtificialSyncrony` with canon | 6 min 39 sec |

(`brunel_ps` with original ps model takes 227.94 sec).

I will try updating `propagator_31` and `propagator_32` now so they are part of a class, where constants are pre-calculated, similar to what is done with canon.

jougs:
Merging this still depends on more discussion about the performance implications. Therefore removing the milestone.

terhorstd:
Hi everyone, has more discussion lead to new insights? – Just a friendly ping.

heplesser:
I have now re-run benchmarks for the `brunel_ps.sli` model using the propagator implementation provided in #2385 using `iaf_psc_alpha_canon` (deprecated, to be removed) and `iaf_psc_alpha_ps`. Experiments were done on a MacBook Pro with 2 GHz Quad Core Intel Core i5 with Clang 14. Code is from https://github.com/heplesser/nest-simulator/commit/10f643477bf4a2d8d408cb68067d518a7169a130, which provides a new `IAFPropagator` class, except for the `..._alpha_ps` version of `brunel_ps.sli`, which is taken from this PR. Simulation times are minima of three runs. For comparison, also results for the current master https://github.com/nest/nest-simulator/commit/e49be2ae64b92b6b6d7c451b2ed8027c0975e480 with the old propagator implementation are given. The `alpha_canon` model uses linear interpolation to find the precise spike time.

Code | Neuron model | Simulation time 
-- | -- | --:
heplesser@10f6434 | `iaf_psc_alpha_canon` | 41.4 s
heplesser@10f6434 | `iaf_psc_alpha_ps` | 55.3 s
master@e49be2a | `iaf_psc_alpha_ps` | 78.9 s

The `ps` variant is thus about 25% slower than the `canon` variant, but for two good reasons:
1. `ps` uses an iterative method to find the spike time, while `canon` uses linear interpolation
2. `ps` supports different excitatory and inhibitory synaptic time constants

Notice also that the new `IAFPropagator` class provided by #2385 significantly improves performance for the `ps` model compared to current master.

I therefore believe that the qualities of the `ps` model justify removal of `canon` once #2385 is merged.

diesmann:
In

Morrison A., Straube S., Plesser HE., Diesmann M. (2007) Exact Subthreshold Integration with Continuous Spike Times in Discrete-Time Neural Network Simulations. Neural Computation 19:47–79.
DOI: [10.1162/neco.2007.19.1.47](https://doi.org/10.1162/neco.2007.19.1.47).

We argue that stating the time required to propagate the implementation of a particular neuron model has no meaning. If the integration error is not specified an implementation can be arbitrarily fast, for example by doing nothing.  Therefore one first needs to specify an accuracy goal and then select the implementation which achieves this in the shortest time. In the case of spiking neuron models we have two measures: the precision of the spike times and the number of missed spikes. The _canon model was included in the code base for NEST not for production but for documenting  the results of the publication above and as a reference. When comparing _ps implementations among each other and with grid-constrained methods it also needs to be considered that for modern _ps algorithms the computation time step h can be increased because these models jump from incoming spike to incoming spike anyway.

Later results are summarized in:

Hanuschkin A, Kunkel S, Helias M, Morrison A and Diesmann M (2010) A general and efficient method for incorporating precise spike times in globally time-driven simulations. Front. Neuroinform. 4:113
 
Krishnan J, Porta Mana P, Helias M, Diesmann M and Di Napoli E (2018) Perfect Detection of Spikes in the Linear Sub-threshold Dynamics of Point Neurons. Front. Neuroinform. 11:75

heplesser:
The figure below shows the median spike time error for 288 spikes recorded over 10 s from a single neuron driven by precise Poisson spike trains mimicking a `brunel_ps` network with mean rate 28.6 sp/s. Error is computed relative to a reference solution obtained with the `iaf_psc_alpha_ps` model at resolution $h=2^{-12}$ms. This is similar to the approach of Hanuschkin et al (2010). Errors were obtained for the following models:
- `grid`: `iaf_psc_alpha`
- `can/lin`: `iaf_psc_alpha_canon` with linear interpolation
- `can/cub`: `iaf_psc_alpha_canon` with cubic interpolation
- `ps`: `iaf_psc_alpha_ps`

To achieve the same accuracy obtained with `ps` for $h=0.5$ms, one needs to run `can/cub` with $h=2*-9$ms. Then, runtimes are (best of three runs, same machine as above)

Neuron model | Simulation time
--- | ---:
`iaf_psc_alpha_canon` (cubic) | 73.1 s
spike_time_error

heplesser:
For reference, attached is the notebook used to create the plot above.
[NEST-PR2294-PrecisionComparison.ipynb.zip](https://github.com/nest/nest-simulator/files/10277736/NEST-PR2294-PrecisionComparison.ipynb.zip)

PR #2298, opened 2022-02-16
Generalize double and integer synapse parameters in `connect_arrays()`

### LINKED ISSUE #2285: Can't specify synapse label when using array indices and one-to-one connection rule
**Describe the bug**

It's not possible to use pass a `synapse_label` parameter in the `syn_spec` dictionary, in conjunction with the `one_to_one` rule and node ids passed as arrays. It works when nodes are passed as `NodeCollection`, but this limits some use cases where the same node should appear multiple times.

**To Reproduce**

```
import nest
neurons = nest.Create('iaf_psc_exp', 2)
nest.Connect([1], [2], conn_spec='one_to_one', syn_spec={'synapse_model': 'static_synapse_lbl', 'synapse_label': 1})
```

gives the error 
```
>> nest.lib.hl_api_exceptions.TypeMismatch: TypeMismatch in SLI function connect_arrays
```
However, the following works:
```
neurons = nest.Create('iaf_psc_exp', 2)
nest.Connect(nest.NodeCollection([1]), nest.NodeCollection([2]),
             conn_spec='one_to_one', syn_spec={'synapse_model': 'static_synapse_lbl', 'synapse_label': 1})
```
Removing the synapse_label parameter also works.

**Expected behavior**

Connection should be successful.

[cmake_summary.txt](https://github.com/nest/nest-simulator/files/7990166/cmake_summary.txt)

 - OS: Linux Mint 19.3, kernel 5.4.0-050400-generic
 - Python-Version: 3.9.0
 - NEST-Version: 3.2
 - Installation: manual install inside Miniconda env, see attached file.

hakonsbm:
Synapse parameters in `connect_arrays()` are now checked against the default values for that synapse model, and stored in a SLI `Dictionary` with the correct type. Fixes #2285.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -578,7 +578,7 @@ nest::ConnectionManager::connect_arrays( long* sources,
   sw_construction_connect.start();
 
   // Mapping pointers to the first parameter value of each parameter to their respective names.
-  // The bool is whether the value is an integer or not, which is determined at a later point.
+  // The bool indicates whether the value is an integer or not, and is determined at a later point.
   std::map< Name, std::pair< double*, bool > > param_pointers;
   if ( p_keys.size() != 0 )
   {
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -578,17 +578,21 @@ nest::ConnectionManager::connect_arrays( long* sources,
   sw_construction_connect.start();
 
   // Mapping pointers to the first parameter value of each parameter to their respective names.
-  std::map< Name, double* > param_pointers;
+  // The bool is whether the value is an integer or not, which is determined at a later point.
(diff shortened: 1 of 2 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -578,39 +578,51 @@ nest::ConnectionManager::connect_arrays( long* sources,
   sw_construction_connect.start();
 
   // Mapping pointers to the first parameter value of each parameter to their respective names.
-  std::map< Name, double* > param_pointers;
+  // The bool is whether the value is an integer or not, which is determined at a later point.
```
stinebuu, line 581:
```suggestion
  // The bool indicates whether the value is an integer or not, and is determined at a later point.
```

mlober (approved):
Sorry for approving so late, I did not realize I was a reviewer until now. Everything looks good to me.

PR #2304, opened 2022-02-22
Make sure node parameters can be iterables or Parameter when creating spatial nodes

### LINKED ISSUE #2251: Dependency between `params` and `positions` in `nest.Create`
**Describe the bug**
Providing `positions` to the `nest.Create()` function together with the `params`, forces the `params` to contain only single values and thus no list will be allowed
**To Reproduce**
1. `params = {"V_th":[-54.0, -55.0, -56,0}`
2. `positions = nest.spatial.grid(shape=[1,3])`
3. `x = nest.Create('iaf_psc_alpha', positions=positions, params=params)`
    *   An error will be thrown: 
            `TypeMismatch in SLI function CreateLayer_D_D: Expected datatype: doubletype
             Provided datatype: arraytype`

**Expected behavior**
`PyNest` or `SLI` should be able to infer the size of the population from the `spatial.grid` object.

 - OS: Ubuntu-20.04
 - NEST-Version: v3.1

stinebuu:
This fixes #2251

### DIFF
--- a/pynest/nest/lib/hl_api_nodes.py
+++ b/pynest/nest/lib/hl_api_nodes.py
@@ -82,6 +82,13 @@ def Create(model, n=1, params=None, positions=None):
 
     model_deprecation_warning(model)
 
+    # If any of the elements in the parameter dictionary is either an array-like object,
+    # or a NEST parameter, we create the nodes first, then set the given values. If not,
+    # we can pass the parameter specification to SLI when the nodes are created.
+    iterable_or_parameter_in_params = True
+    if isinstance(params, dict) and params:  # if params is a dict and not empty
+        iterable_or_parameter_in_params = any(is_iterable(v) or isinstance(v, Parameter) for k, v in params.items())
+
     if positions is not None:
         # Explicitly retrieve lazy loaded spatial property from the module class.
         # This is needed because the automatic lookup fails. See #2135.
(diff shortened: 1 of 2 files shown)


med-ayssar:
Thank you @stinebuu, it works pretty well. However, I would like addressing another issues regarding the `positions` parameter.
Once the `positions`  are given to the `create` function, we will have the following `NodeCollection` instance: `x = NodeCollection(metadata=spatial, model=iaf_psc_alpha, size=6, first=19, last=24)`, but calling `y = ǹest.NodeCollection(x.tolist())` return another `NodeCollection` instance with the `metatdata` set to `None`, `y = NodeCollection(metadata=None, model=iaf_psc_alpha, size=6, first=19, last=24)`, and I think that the instance `y` must also somehow have its `metadata` equal to the `metadata` of `x`, as they both point to the same collection.

stinebuu:
Thanks @med-ayssar. Could you create another issue about this? I do not think this should be solved in this PR as it is a separate issue. The problem is a bit more complex for this second issue, I am not sure there is an easy fix, we would need to look into it. The metadata needs to be set on Kernel level, and we would need to find a way to check if the node ids have metadata. It would be great if you could say a little about when you would need this in the issue as well. We are aware of the issue already, and would be happy to look into it if there is a real use case, but up to now there hasn't really been any use cases.

med-ayssar (approved):

hakonsbm (approved):
Thanks!

PR #2283, opened 2022-02-01
Restructure the userdoc directory

jessica-mitchell:
This PR reorganizes the documentation to improve findability of different topics, both in the source on GitHub and on Read the docs.

In the userdoc/ directory the guides/ have been removed in favour of more descriptive folders
New folders:

- neuron_docs/  -- docs about neurons
- synapse_docs/ -- docs about synapses and connections
- device_docs/ -- docs about devices
- coupling_nest_docs/  -- docs about other tools that connect with nest
- nest_behavior_docs/ -- docs about how nest works
- hpc_docs/ -- docs about hpc setup

A new table of contents was also created for highlighting topics on Read the Docs:
You can see the structure output here: https://nest-test.readthedocs.io/en/new-doc-folder-structure/contents.html

In addition, I moved the developer docs back into userdoc/ because
1. I want to build the dev docs with Sphinx and that's done easiest if it's in userdoc/
2. I want to include all docs that relate to maintaining, contributing or developing the documentation or code in anyway in devdoc/.  (Thus all contribution guidelines, styleguides, workflows, go in here) 
3. Note that the actual doxygen output is not rendered on Rtd yet

See the output table of contents for the developer area here: https://nest-test.readthedocs.io/en/new-doc-folder-structure/devdoc/index.html

Please comment if you have suggestions for alternatives!

PR #2305, opened 2022-02-22
Add documentation about HPC in NEST

jessica-mitchell:
This PR adds documentation regarding using optimizing HPC systems to use with NEST, including pinning threads.

Setting as draft for a few changes still needed.

Needs #2283  to be merged.


jessica-mitchell:
Removing to fix up

PR #2306, opened 2022-02-22
Remove CMake options and other references to BlueGene and K-computer

### LINKED ISSUE #2249: C-make options and other references to Bluegene and K-computer still exists
It is still possible to build NEST with Bluegeene and the K-computer, even though the machines do not exist anymore. See for instance

https://github.com/nest/nest-simulator/blob/06f9276ecef401b0e5cf2cfef8a6f028470ba1c1/cmake/ProcessOptions.cmake#L143-L146

When searching through the code base for `bluegene` and `k-computer` quite a lot of references show up. All the references to the machines should probably be removed.

stinebuu:
This fixes #2249


cmake/ProcessOptions.cmake
```diff
@@ -71,11 +71,7 @@ endfunction()
 function( NEST_PROCESS_WITH_WARNING )
   if ( with-warning )
     if ( with-warning STREQUAL "ON" )
-      if ( NOT k-computer STREQUAL "ON" )
-        set( with-warning "-Wall" )
-      else()
-        set( with-warning "" )
-      endif()
+      set( with-warning "" )
```
heplesser, line 74:
```suggestion
      set( with-warning "-Wall" )
```

nest/neststartup.cpp
```diff
@@ -127,27 +127,15 @@ neststartup( int* argc, char*** argv, SLIInterpreter& engine, std::string module
  * The following section concerns shared user modules and is thus only
  * included if we built with libtool and libltdl.
  *
- * On BlueGene, we need to link user modules statically, but for convenience
+ * On some machines, we need to link user modules statically, but for convenience
```
heplesser, line 130:
```suggestion
 * One may want to link user modules statically, but for convenience
```

nest/neststartup.cpp
```diff
@@ -127,27 +127,15 @@ neststartup( int* argc, char*** argv, SLIInterpreter& engine, std::string module
...
  */
 #ifdef HAVE_LIBLTDL
   // dynamic loader module for managing linked and dynamically loaded extension
   // modules
   nest::DynamicLoaderModule* pDynLoader = new nest::DynamicLoaderModule( engine );
-
-// initialize all modules that were linked into at compile time
-// these modules have registered via calling DynamicLoader::registerLinkedModule
-// from their constructor
-#ifndef IS_BLUEGENE
```
heplesser, line 145: Since this is an `#ifndef`, I believe we need to just remove the guard but keep the code.

nestkernel/mpi_manager_impl.h
```diff
@@ -57,10 +54,9 @@ extern MPI_Comm comm;
 /* ------------------------------------------------------
    The following datatypes are defined here in communicator_impl.h
    file instead of as static class members, to avoid inclusion
-   of mpi.h in the .h file. This is necessary, because on
-   BlueGene/L mpi.h MUST be included FIRST. Having mpi.h in
-   the .h file would lead to requirements on include-order
-   throughout the NEST code base and is not acceptable.
+   of mpi.h in the .h file. Having mpi.h in the .h file would
+   lead to requirements on include-order throughout the NEST
+   code base and is not acceptable.
```
heplesser, line 59:
```suggestion
   of mpi.h in the .h file, which caused problems on BlueGene systems. 
   Having mpi.h in the .h file would lead to requirements on include-order 
   throughout the NEST code base and is not acceptable.
```

heplesser (changes requested):
Thanks for this PR. Please see my comments below. One general question: Is the Fujitsu-Cmake file relevant for Fugaku?

nest/neststartup.cpp
stinebuu: Thanks for spotting this, I was a bit quick when removing things. Fixed now.

stinebuu:
@heplesser Thanks for your review, I have updated the PR now.

Concerning the Fujitsu-Cmake file, I don't think it is needed for Fugako, but I will ask people with more expertise here just in case.

heplesser (approved):

heplesser:
> @heplesser Thanks for your review, I have updated the PR now.
> 
> Concerning the Fujitsu-Cmake file, I don't think it is needed for Fugako, but I will ask people with more expertise here just in case.

I think that for this PR, we should just remove fujitsu-sparc64. We may then if needed add instruction for Fugaku in a separate PR.

gtrensch (approved):
Thank you for this pull request. :+1: Approved.

PR #2307, opened 2022-02-22
Use `updateValue< long >` instead of `update_value_int` in `quantal_stp_synapse`

### LINKED ISSUE #2049: Move `update_value_int` from quantal_stp_synapse to where `updateValue` is defined
The `quantal_stp_synapse` has a function called `update_value_int()`, which is used when setting the integer parameters of the model. This function might be useful for other models as well, and should be placed where `updateValue` is [defined](https://github.com/nest/nest-simulator/blob/master/sli/dictutils.h#L233).

https://github.com/nest/nest-simulator/blob/d6499c6ccf7c3d6926a2cd811e36dfca65124fe0/models/quantal_stp_synapse_impl.h#L40-L49

stinebuu:
This PR fixes #2049

### DIFF
--- a/models/quantal_stp_synapse_impl.h
+++ b/models/quantal_stp_synapse_impl.h
@@ -76,8 +76,8 @@ quantal_stp_synapse< targetidentifierT >::set_status( const DictionaryDatum& d, 
   updateValue< double >( d, names::u, u_ );
   updateValue< double >( d, names::tau_rec, tau_rec_ );
   updateValue< double >( d, names::tau_fac, tau_fac_ );
-  updateValueInt( d, names::n, n_ );
-  updateValueInt( d, names::a, a_ );
+  updateValue< long >( d, names::n, n_ );
+  updateValue< long >( d, names::a, a_ );
 }
 
 } // of namespace nest
--- a/models/quantal_stp_synapse_impl.h
+++ b/models/quantal_stp_synapse_impl.h
@@ -36,38 +36,6 @@
 namespace nest
 {
 
-
-/* Polymorphic version of update_value.
- * This code will take either an int or a double and convert it to an
- * int.
- */
-bool
-update_value_int( const DictionaryDatum& d, Name propname, int& prop )
-{
-  if ( d->known( propname ) )
-  {
-    Datum* dat = ( *d )[ propname ].datum();
-    IntegerDatum* intdat = dynamic_cast< IntegerDatum* >( dat );
-    if ( intdat != 0 )
-    {
-      prop = static_cast< int >( intdat->get() );
-      return true;
-    }
-    DoubleDatum* doubledat = dynamic_cast< DoubleDatum* >( dat );
-    if ( doubledat != 0 )
-    {
-      prop = static_cast< int >( doubledat->get() );
-      return true;
-    }
-    else
-    {
-      throw TypeMismatch();
-    }
-  }
-
-  return false;
-}
-
 template < typename targetidentifierT >
 quantal_stp_synapse< targetidentifierT >::quantal_stp_synapse()
   : ConnectionBase()
@@ -108,8 +76,8 @@ quantal_stp_synapse< targetidentifierT >::set_status( const DictionaryDatum& d, 
   updateValue< double >( d, names::u, u_ );
   updateValue< double >( d, names::tau_rec, tau_rec_ );
   updateValue< double >( d, names::tau_fac, tau_fac_ );
-  update_value_int( d, names::n, n_ );
-  update_value_int( d, names::a, a_ );
+  updateValueInt( d, names::n, n_ );
+  updateValueInt( d, names::a, a_ );
 }
 
 } // of namespace nest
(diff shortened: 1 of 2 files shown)


sli/dictutils.h
```diff
@@ -268,6 +268,38 @@ updateValue2( DictionaryDatum const& d, Name const n, C& obj, void ( C::*setfunc
   }
 }
 
+/** Update an int or a double from a dictionary entry if it exists, and convert it to an int.
+ * Skip call if it doesn't exist.
+ * @ingroup DictUtils
+ * @throws TypeMismatch The entry is neither int nor double.
+ */
```
clinssen, line 275:
```suggestion
 * @returns true if the property was found (and updated) in the dictionary, false otherwise
 */
```

sli/dictutils.h
```diff
@@ -268,6 +268,38 @@ updateValue2( DictionaryDatum const& d, Name const n, C& obj, void ( C::*setfunc
...
+inline bool
+updateValueInt( const DictionaryDatum& d, Name propname, int& prop )
+{
+  if ( d->known( propname ) )
+  {
+    Datum* dat = ( *d )[ propname ].datum();
+    IntegerDatum* intdat = dynamic_cast< IntegerDatum* >( dat );
+    if ( intdat != 0 )
```
clinssen, line 283:
```suggestion
    if ( intdat != nullptr )
```

sli/dictutils.h
```diff
@@ -268,6 +268,38 @@ updateValue2( DictionaryDatum const& d, Name const n, C& obj, void ( C::*setfunc
...
+    {
+      prop = static_cast< int >( intdat->get() );
+      return true;
+    }
+    DoubleDatum* doubledat = dynamic_cast< DoubleDatum* >( dat );
+    if ( doubledat != 0 )
```
clinssen, line 289:
```suggestion
    if ( doubledat != nullptr )
```

models/quantal_stp_synapse_impl.h
```diff
@@ -108,8 +76,8 @@ quantal_stp_synapse< targetidentifierT >::set_status( const DictionaryDatum& d,
   updateValue< double >( d, names::u, u_ );
   updateValue< double >( d, names::tau_rec, tau_rec_ );
   updateValue< double >( d, names::tau_fac, tau_fac_ );
-  update_value_int( d, names::n, n_ );
-  update_value_int( d, names::a, a_ );
+  updateValueInt( d, names::n, n_ );
```
clinssen, line 79: Why couldn't this just be ``updateValue < int >``? (Same for the next line)

clinssen (changes requested):

models/quantal_stp_synapse_impl.h
stinebuu: Good question! I had not thought about this. I tried now, and got the following error message:
```
error: cannot dynamic_cast ‘(& t)->Token::datum()’ (of type ‘class Datum*’) to type ‘int*’ (target is not pointer or reference to class)
   FT* value = dynamic_cast< FT* >( t.datum() );
```
I will update the docstring of `updateValue`.

clinssen: It looks like ``updateValue< long >()`` works fine here, but using an ``int`` would of course be preferred. @jougs: is there any reason why an overloaded version for ``int``s is not defined for ``getValue()`` in ``sli/tokenutils.cc``? If we can just use ``long``, this whole ``updateValueInt()`` function can go away (which is anyway already casting the ``long`` inside ``IntegerDatum`` to an ``int``).

stinebuu: I have updated the comment for `updateValue` now, but agree that not having to use `updateValueInt()` would be good, so I would be happy to change to this.

stinebuu: @jougs have you had a chance to look at @clinssen's comment about `updateValue< long >()`?

stinebuu: @clinssen @jougs I have now updated to use `updateValue< long >`.

jougs:
One question: Does `updateValue< long >`also have the property `this code will take either an int or a double and convert it to an int` that the removed function had according to it's Doxygen comment?

stinebuu:
@jougs Good point. An error is now thrown if we provide a double. It is quite easy to fix by adding a check in the function of `getValue< long >`, should I do that?
https://github.com/nest/nest-simulator/blob/66269a25fbda4bad67c178b82d0537c242892a00/sli/tokenutils.cc#L41-L50

jougs:
I think it would be good if that functionality were there. Otherwise we might again run into trouble when setting data from NEST Desktop, see #2009 and #2024.

clinssen:
I don't understand why this automatic cast is necessary; it could even be dangerous due to loss of precision. The underlying data type in class ``quantal_stp_synapse`` is an ``int``. The cast, if it should happen at all, should be on the Python side, and NEST should error out when passed a ``float``/``double``.

The only actual danger is mixing an ``int`` and a ``long`` here.

jougs:
Argh! You're right, @clinssen. I was thinking the wrong way around all the time. This function updates a C++ variable of type `long` from a dictionary and that should of course throw an error if the datum in the dictionary is of type `double`. Sorry for confusing you, @stinebuu, I think you have to revert your recent commits aaf2fdd and 13055f1 :crying_cat_face:

stinebuu:
I will convert back to 04347251a89ec46d25a2fd3b317341529c40100d :wink:

clinssen (approved):

PR #2311, opened 2022-02-24
Remove abstract structural plasticity methods from Node base class

clinssen:
This is a refactoring that removes abstract methods related to structural plasticity from the ``Node`` base class. This means that a ``Node`` can no longer be assumed to define these methods, and a ``static_cast`` to ``StructuralPlasticityNode`` is necessary. ``static_cast`` is sufficient here (as contrasted with a ``dynamic_cast``) because this code is anyway only reached when structural plasticity is enabled. The runtime overhead is therefore zero; it's a purely-cosmetic change.


clinssen:
@sdiazpier: do you have any idea why removing the virtual functions in the base class (``Node``) breaks things? I'm getting the error "Last update of the calcium concentration does not match the last update of the synaptic element", even for tests that are unrelated to structural plasticity. If this gets too complicated, I will just close this PR without merging!

sdiazpier:
Hi @clinssen, sorry I have not had time to look into this PR. This seems strangely as if the synaptic element update has become detached from the calcium dynamics. 
Is this change very critical for the development in NestML? If it is critical but not urgent, I can promise to look at it by the end of the year.

clinssen:
Thanks for your quick reply! It is not urgent, but it came up during the NEST hackathon. It's a bit of a mystery though why this happens, so it would be interesting to look into it further. I'd be happy to do a pair programming session as well if that's more convenient. Cheers!

nestkernel/simulation_manager.cpp
```diff
@@ -1038,7 +1038,7 @@ nest::SimulationManager::update_()
           i != kernel().node_manager.get_local_nodes( tid ).end();
           ++i )
     {
-      Node* node = i->get_node();
+      StructuralPlasticityNode* node = static_cast< StructuralPlasticityNode* >( i->get_node() );
```
med-ayssar, line 1041: Inside this `for-loop`. The `local_nodes` object might contain `Neurons`, `Devices` and `correlomatrix_detector`. Depending on the Simulation script, you have to be careful if the node object is really of type `StructuralPlasticityNode`. Thus, when you do the conversion, you will get a weird object that has undefined behavior. In my testing case, the `for-loop` in the `update_synaptic_elements` was entered despite that it has the size of `zero`.

med-ayssar:
@clinssen is there any progress regarding the current status.

JanVogelsang (approved):
Making the SP-functions non-virtual is great, but the dynamic casts are not ideal yet. While I think this fine for now, in the long run we should find a cleaner solution (i.e., some way to efficiently iterate over all neurons involved in SP).

JanVogelsang:
@clinssen Could you check Ayssar's comment and could you also provide benchmark data so we can make sure the additional dynamic casts don't slow down the code. If you need support with the benchmarks, feel free to contact me.

heplesser (changes requested):
@clinssen I consider dynamic casts a solution of last resort only. I agree that the interface of the `Node` base class has become far too large. But the solution is a careful re-design of the entire `Node` class concept that should lead to a cleaner interface throughout.

Therefore, I suggest to live with the large base class interface for now and to close this PR.

heplesser:
@clinssen As you did not protest after my previous post a week ago, I will close this PR now.

PR #2316, opened 2022-02-25
Reduce memory consumption in node creation with many threads

hakonsbm:
Creating nodes of different models while using many threads uses an excessive amount of memory. For example, creating 10 populations of 5000 nodes each and different models for each population:

| Threads | Memory [MB] |
|---------:|-----------:|
| 1       |         135 |
| 256     |        2255 |

This is because node creation relies on memory management with `sli::pool`:
https://github.com/nest/nest-simulator/blob/d2e49741374331b23d2b2ed7e218da9a5c993fe1/nestkernel/model.h#L253

By replacing the `sli::pool` with an `std::vector`, the memory consumption is kept at a more manageable level:
| Threads | Memory [MB] |
|---------:|-----------:|
| 1       |         129 |
| 256     |        306 |

### DIFF
--- a/nestkernel/model.h
+++ b/nestkernel/model.h
@@ -89,24 +89,11 @@ public:
    */
   Node* allocate( thread t );
 
-  void free( thread t, Node* );
-
   /**
    * Deletes all nodes which belong to this model.
    */
 
   void clear();
-
-  /**
-   * Reserve memory for at least n additional Nodes.
-   * A number of memory managers work more efficiently if they have
-   * an idea about the number of Nodes to be allocated.
-   * This function prepares the memory manager for the subsequent
-   * allocation of n additional Nodes.
-   * @param t Thread for which the Nodes are reserved.
-   * @param n Number of Nodes to be allocated.
-   */
-  void reserve_additional( thread t, size_t n );
 
   /**
    * Return name of the Model.
@@ -223,14 +210,9 @@ private:
   void set_threads_( thread t );
 
   /**
-   * Initialize the pool allocator with the Node specific values.
-   */
-  virtual void init_memory_( sli::pool& ) = 0;
-
-  /**
    * Allocate a new object at the specified memory position.
    */
-  virtual Node* allocate_( void* ) = 0;
+  virtual Node* allocate_() = 0;
 
   /**
    * Name of the Model.
@@ -250,7 +232,7 @@ private:
   /**
    * Memory for all nodes sorted by threads.
    */
-  std::vector< sli::pool > memory_;
+  std::vector< std::vector< Node* > > memory_;
 };
 
 
@@ -258,14 +240,9 @@ inline Node*
 Model::allocate( thread t )
 {
   assert( ( size_t ) t < memory_.size() );
-  return allocate_( memory_[ t ].alloc() );
-}
-
-inline void
-Model::free( thread t, Node* n )
-{
-  assert( ( size_t ) t < memory_.size() );
-  memory_[ t ].free( n );
+  Node* n = allocate_();
+  memory_[ t ].push_back( n );
+  return n;
 }
 
 inline std::string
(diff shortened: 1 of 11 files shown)


heplesser:
@hakonsbm That is an impressive reduction in memory use. It would be interesting to see construction and simulation times for realistic network sizes as well. And could memory consumption be reduced even further using `BlockVector` without noticeable performance loss?

hakonsbm:
@heplesser For connection and simulation the `memory_` vector is actually not used. The node pointer is added to `local_nodes_`, which is the one used for connection and simulation:
https://github.com/nest/nest-simulator/blob/d2e49741374331b23d2b2ed7e218da9a5c993fe1/nestkernel/node_manager.cpp#L252-L261
Therefore, the connection and simulation times should be unaffected, which a run of hpc_benchmark with a scale of 20 supports.

Using `BlockVector` instead of `std::vector` actually uses more memory (696 MB with `BlockVector` vs 306 MB with `std::vector`). That's because `BlockVector` initializes 1024 elements when created, and that is done 256 times (once for each thread) for every model, leading to a big overhead.

heplesser:
@hakonsbm Good to know that building and simulation time is not affected. But I think the `memory_` is used in the sense that this is where the neuron objects are stored, so if memory layout had an effect on performance, we might have seen changes. But given that `vector` stores objects contiguously, this is presumably the optimal memory layout.

nestkernel/model.h
```diff
@@ -222,15 +209,10 @@ class Model
    */
   void set_threads_( thread t );
 
-  /**
-   * Initialize the pool allocator with the Node specific values.
-   */
-  virtual void init_memory_( sli::pool& ) = 0;
-
   /**
    * Allocate a new object at the specified memory position.
```
heplesser, line 213: Comment is outdated.

nestkernel/model.h
```diff
@@ -250,22 +232,17 @@ class Model
...
    * Memory for all nodes sorted by threads.
    */
-  std::vector< sli::pool > memory_;
+  std::vector< std::vector< Node* > > memory_;
 };
 
 
 inline Node*
 Model::allocate( thread t )
 {
```
heplesser, line 241: Are `allocate()` and `allocate_()` actually good names? Wouldn't `create` or `clone` be more readable?

nestkernel/node_manager.cpp
```diff
@@ -186,20 +186,14 @@ NodeManager::add_node( index model_id, long n )
...
-  const size_t max_new_per_thread =
-    static_cast< size_t >( std::ceil( static_cast< double >( max_node_id - min_node_id + 1 ) / num_vps ) );
 
 #pragma omp parallel
   {
     const index t = kernel().vp_manager.get_thread_id();
 
     try
     {
-      model.reserve_additional( t, max_new_per_thread );
```
heplesser, line 201: Could it make sense to keep `reserve_additional` to avoid unnecessary automatic vector resizings?

heplesser (commented):
Looks generally good, just a few comments/questions.

nestkernel/model.h
```diff
@@ -83,18 +83,23 @@ class Model
...
+  Node* create( thread t );
 
   /**
    * Deletes all nodes which belong to this model.
    */
 
   void clear();
 
+  /**
+   * Reserve space for n Nodes.
```
jougs, line 99:
```suggestion
   * Reserve space for n additional Nodes.
```

jougs (approved):
I'm fine with the changes here. Many thanks!

More fundamentally, I'm wondering if we wouldn't be better of with simple thread-local vectors of nodes of a certain type in the `GenericModel` instead of packing them into custom pool-based allocators that have never been really tested or designed for performance... But that's probably for another time.

heplesser (approved):

PR #2317, opened 2022-02-28
Disallow specifying lists of synapse models

### LINKED ISSUE #2301: Exception when specying synapse models as list in Connect()
An exception happens when connecting with a given list of synapse models. This is (probably?) a rather obscure use case, and we could possibly deprecate this feature, which would amount to allowing only one single synapse type (such as "stdp_synapse") to be instantiated during one single Connect() call.

Minimal reproducer:
```python
import nest
n = nest.Create("iaf_psc_exp", 20)
nest.Connect(n[:10], n[10:], "one_to_one", syn_spec={"synapse_model": 42*["static_synapse"]})
--> NESTError: 'synapse_model' has to be an array of dimension 10, a scalar or a dictionary.
```
OK, so we can specify the synapse models as a list, but we need to give a list with the correct length!
```python
nest.Connect(n[:10], n[10:], "one_to_one", syn_spec={"synapse_model": 10*["static_synapse"]})
```
However, when doing this, we get:
```python
AttributeError                            Traceback (most recent call last)
<ipython-input-5-4f099322e6b2> in <module>
----> 1 nest.Connect(n[:10], n[10:], "one_to_one", syn_spec={"synapse_model": 10*["static_synapse"]})

~/julich/nest-simulator-install/lib/python3.9/site-packages/nest/ll_api.py in stack_checker_func(*args, **kwargs)
    225     def stack_checker_func(*args, **kwargs):
    226         if not get_debug():
--> 227             return f(*args, **kwargs)
    228         else:
    229             sr('count')

~/julich/nest-simulator-install/lib/python3.9/site-packages/nest/lib/hl_api_connections.py in Connect(pre, post, conn_spec, syn_spec, return_synapsecollection)
    281         sps(processed_conn_spec)
    282         if processed_syn_spec is not None:
--> 283             sps(processed_syn_spec)
    284         sr('Connect')
    285 

pynestkernel.pyx in pynestkernel.NESTEngine.push()

pynestkernel.pyx in pynestkernel.python_object_to_datum()

pynestkernel.pyx in pynestkernel.python_object_to_datum()

AttributeError: type object 'NESTError' has no attribute 'PyNESTError'
```

hakonsbm:
Limits `synapse_model` to be a string, and fixes a typo in pynestkernel.pyx which lead to confusing error messages. Fixes #2301.

### DIFF
--- a/pynest/nest/lib/hl_api_connection_helpers.py
+++ b/pynest/nest/lib/hl_api_connection_helpers.py
@@ -75,7 +75,7 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, use_connect_ar
     if (not isinstance(syn_spec, CollocatedSynapses) and
         "synapse_model" in syn_spec and
             not isinstance(syn_spec["synapse_model"], str)):
-        raise kernel.NESTError("'synapse_model' must be a string")
+        raise kernel.NESTError("'synapse_model' must be a string or a CollocatedSynapses object")
 
     elif isinstance(syn_spec, dict):
         for key, value in syn_spec.items():
--- a/pynest/nest/lib/hl_api_connection_helpers.py
+++ b/pynest/nest/lib/hl_api_connection_helpers.py
@@ -71,6 +71,10 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, use_connect_ar
 
     if isinstance(syn_spec, str):
         return {"synapse_model": syn_spec}
+
+    if "synapse_model" in syn_spec and not isinstance(syn_spec["synapse_model"], str):
+        raise kernel.NESTError("'synapse_model' must be a string")
+
     elif isinstance(syn_spec, dict):
         for key, value in syn_spec.items():
             # if value is a list, it is converted to a numpy array
--- a/pynest/nest/lib/hl_api_connection_helpers.py
+++ b/pynest/nest/lib/hl_api_connection_helpers.py
@@ -71,13 +71,9 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, use_connect_ar
 
     if isinstance(syn_spec, str):
         return {"synapse_model": syn_spec}
-
-    if (not isinstance(syn_spec, CollocatedSynapses) and
-        "synapse_model" in syn_spec and
-            not isinstance(syn_spec["synapse_model"], str)):
-        raise kernel.NESTError("'synapse_model' must be a string or a CollocatedSynapses object")
-
     elif isinstance(syn_spec, dict):
+        if "synapse_model" in syn_spec and not isinstance(syn_spec["synapse_model"], str):
+            raise kernel.NESTError("'synapse_model' must be a string")
         for key, value in syn_spec.items():
             # if value is a list, it is converted to a numpy array
             if isinstance(value, (list, tuple)):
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -71,6 +71,12 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, use_connect_ar
 
     if isinstance(syn_spec, str):
         return {"synapse_model": syn_spec}
+
+    if (not isinstance(syn_spec, CollocatedSynapses) and
+        "synapse_model" in syn_spec and
+            not isinstance(syn_spec["synapse_model"], str)):
+        raise kernel.NESTError("'synapse_model' must be a string")
```
jougs, line 78: Should this error message also mention `CollocatedSynapses`?

jougs (approved):
Thanks for addressing this so quickly!

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -71,6 +71,12 @@ def _process_syn_spec(syn_spec, conn_spec, prelength, postlength, use_connect_ar
 
     if isinstance(syn_spec, str):
         return {"synapse_model": syn_spec}
+
+    if (not isinstance(syn_spec, CollocatedSynapses) and
+        "synapse_model" in syn_spec and
+            not isinstance(syn_spec["synapse_model"], str)):
+        raise kernel.NESTError("'synapse_model' must be a string or a CollocatedSynapses object")
+
     elif isinstance(syn_spec, dict):
```
clinssen, line 80: Inserting another ``if`` before this ``elif`` changes the logic of when the ``else`` part will be triggered. I don't think we need the ``else`` part here anyway, as the previous ``if``s either return or raise an exception. Could you change this ``elif`` into an ``if``, and move the two lines of code from below:
```python
elif isinstance(syn_spec, CollocatedSynapses):
  return syn_spec
```
to the top of the function?

pynest/pynestkernel.pyx
```diff
@@ -450,7 +450,7 @@ cdef inline Datum* python_object_to_datum(obj) except NULL:
             elif numpy.issubdtype(obj.dtype, numpy.floating):
                 ret = python_buffer_to_datum[object, double](obj)
             else:
```
clinssen, line 452:
```suggestion
```

pynest/pynestkernel.pyx
```diff
@@ -450,7 +450,7 @@ cdef inline Datum* python_object_to_datum(obj) except NULL:
...
-                raise NESTError.PyNESTError("only vectors of integers or floats are supported")
+                raise NESTErrors.PyNESTError("only vectors of integers or floats are supported")
```
clinssen, line 453:
```suggestion
            raise NESTErrors.PyNESTError("only vectors of integers or floats are supported")
```

pynest/pynestkernel.pyx
clinssen: Return-early pattern: ``else`` can be removed.

clinssen (changes requested):

pynest/pynestkernel.pyx
hakonsbm: We are only storing the value in `ret` here, not returning, so I think we need the `else` statement.

clinssen (approved):
Cheers!

PR #2319, opened 2022-03-01
Make the C++ standard configurable

### LINKED ISSUE #2259: Add CMake flag for C++ standard
The C++ standard `-std=c++11` is currently hardcoded in the CMake files:

https://github.com/nest/nest-simulator/blob/cb61289cc5751f68eb42f5eaf6564acc03823252/cmake/DefaultCompilerFlags.cmake#L25

To make it possible to easily check different standards, this value should be configurable. A CMake option should be added, for example something like `-Dwith-std=<C++ standard>`.

hakonsbm:
The C++ standard can now be specified in CMake configuration with `-Dwith-std=<C++ standard>`. Default is `C++11`. Fixes #2259.


CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ option( static-libraries "Build static executable and libraries [default=OFF]" O
 set( with-optimize ON CACHE STRING "Enable user defined optimizations [default=OFF (uses '-O2')]. When ON, '-O3' is used. Separate multiple flags by ';'." )
 set( with-warning ON CACHE STRING "Enable user defined warnings [default=ON (uses '-Wall')]. Separate  multiple flags by ';'." )
 set( with-debug OFF CACHE STRING "Enable user defined debug flags [default=OFF]. When ON, '-g' is used. Separate  multiple flags by ';'." )
+set( with-std OFF CACHE STRING "User defined C++ standard [default='c++11']." )
```
heplesser, line 76:
```suggestion
set( with-std OFF CACHE STRING "C++ standard to use for compilation [default='c++11']." )
```
We don't really allow users to define their own C++ standards ;).

jougs (commented):
This looks good to me in principle, but could the flag's name be changed to `with-cpp-std`? I think that would be more expressive.

heplesser (changes requested):
Looks fine, just one little suggestion for documentation.

And please also add documentation for the option in the "Generic build configuration" section of https://github.com/nest/nest-simulator/blob/master/doc/userdoc/installation/cmake_options.rst

doc/userdoc/installation/cmake_options.rst
```diff
@@ -120,6 +120,8 @@ Generic build configuration
 | ``-Dwith-intel-compiler-flags=[OFF|<list;of;flags>]``| User defined flags for the Intel compiler                        |
 |                                                      | [default='-fp-model strict']. Separate multiple flags by ';'.    |
 +------------------------------------------------------+------------------------------------------------------------------+
+| ``-Dwith-cpp-std=[OFF|<C++ standard>]``              | C++ standard to use for compilation [default='c++11'].           |
```
heplesser, line 123: Does `-Dwith-cpp-std=OFF` actually work and result in the default value?

jougs (approved):

doc/userdoc/installation/cmake_options.rst
hakonsbm: It does, `OFF` would just mean the default value. But because it's a bit confusing I have corrected the CMake variable now.

heplesser (approved):

Issue #2175, opened 2021-10-04
pycodestyle ``# noqa`` is selectively ignored


clinssen:
pep8 was replaced by pycodestyle in #1132. Unfortunately, pycodestyle does not allow a ``# noqa`` inline comment to suppress certain errors; see https://github.com/PyCQA/pycodestyle/blob/master/docs/intro.rst. This is by design and unlikely to change any time soon.

This means that (some) existing lines containing ``# noqa`` need to be refactored, because the formatting errors on those lines are no longer being ignored. For a concrete example, see the matrix formatting in https://github.com/nest/nest-simulator/blob/c545255f7252308d6e73a851ddeece4f2dffc812/pynest/examples/clopath_synapse_spike_pairing.py#L73

There are several options here:
- Switch from pycodestyle to flake8, which does not have this restriction
- Replace the line-scope noqas with a file-scope noqa
- Store the matrices in a textfile and use ``numpy.loadtxt()``
- (probably more...)

jougs:
@clinssen: Sorry that I've overlooked this one and might have sounded like I discovered this issue in our recent meeting. That was not my intention.

Please also see https://github.com/nest/nest-simulator/pull/2109#issuecomment-921013681 and https://github.com/PyCQA/pycodestyle/issues/924

PR #2328, opened 2022-03-07
Temporary fix for noqa problems.

heplesser:
This PR provides a temporary fix for the PEP8/noqa problem described in #2175, which currently leads to CI failure for many PRs. It brute-force ignores the problematic file.

This should be reversed as soon as a proper fix is in place (flake8 instead of pycodestyle).

### DIFF
--- a/build_support/parse_build_log.py
+++ b/build_support/parse_build_log.py
@@ -1032,7 +1032,7 @@ if __name__ == '__main__':
                                         "MSGBLD0180", "MSGBLD0175")
 
     summary_pep8 = msg_summary_pep8(log_filename, "MSGBLD0190",
-                                    "MSGBLD0200", "MSGBLD0195")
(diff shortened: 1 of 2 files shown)


jougs:
Not sure we should merge this, given that [help is on the way](https://github.com/terhorstd/nest-simulator/pull/14) and will probably arrive later this week.

heplesser:
> Not sure we should merge this, given that [help is on the way](https://github.com/terhorstd/nest-simulator/pull/14) and will probably arrive later this week.

Maybe we can keep it open but not merged, so we can merge it in case CI for other PRs is blocked by this problem.

jougs (approved):
Given that the other PR will probably still take some time and that the file in which this change is will be removed from the repository in that PR, I am now approving to reduce the annoyance by this nasty error. I'm also merging immediately, as the change does not concern the simulation code.

PR #2329, opened 2022-03-07
Remove installation of python2-dev from CI setup.

heplesser:
This PR removes installation of `python2-dev` from `nestbuildmatrix.yml`.


lekshmideepu (approved):
Thank you! That makes sense.

jougs (approved):
Good catch!

PR #2331, opened 2022-03-08
Fix condition in bash script in case of non-existent variable

### LINKED ISSUE #2330: Unbound variable error while running testsuite
When running the testsuite without MPI enabled, I get:

```sh
/home/charl/julich/nest-simulator-install/share/nest/testsuite/do_tests.sh: line 495: MPI_LAUNCHER: unbound variable
```
This refers to the line
```sh
    if test "${HAVE_MPI}" = "true" -a "${MPI_LAUNCHER}" ; then
```
It looks like the second condition is always evaluated, so this should probably be converted into two nested if statements.

clinssen:
Fixes #2330.

### DIFF
--- a/testsuite/do_tests.sh
+++ b/testsuite/do_tests.sh
@@ -492,8 +492,8 @@ if test "${PYTHON}"; then
           --ignore="${PYNEST_TEST_DIR}/mpi" "${PYNEST_TEST_DIR}" 2>&1 | tee -a "${TEST_LOGFILE}"
 
     # Run tests in the mpi* subdirectories, grouped by number of processes
-    if test "${HAVE_MPI}" = "true" ; then
-        if test "${MPI_LAUNCHER}" ; then
+    if test "${HAVE_MPI}" = "true"; then
+        if test "${MPI_LAUNCHER}"; then
             for numproc in $(cd ${PYNEST_TEST_DIR}/mpi/; ls -d */ | tr -d '/'); do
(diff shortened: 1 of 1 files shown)


testsuite/do_tests.sh
```diff
@@ -492,12 +492,14 @@ if test "${PYTHON}"; then
...
 
     # Run tests in the mpi* subdirectories, grouped by number of processes
-    if test "${HAVE_MPI}" = "true" -a "${MPI_LAUNCHER}" ; then
-       for numproc in $(cd ${PYNEST_TEST_DIR}/mpi/; ls -d */ | tr -d '/'); do
-           XUNIT_FILE="${REPORTDIR}/${XUNIT_NAME}_mpi_${numproc}.xml"
-           PYTEST_ARGS="--verbose --timeout $TIME_LIMIT --junit-xml=${XUNIT_FILE} ${PYNEST_TEST_DIR}/mpi/${numproc}"
-           $(sli -c "${numproc} (${PYTHON} -m pytest) (${PYTEST_ARGS}) mpirun =only") 2>&1 | tee -a "${TEST_LOGFILE}"
-       done
+    if test "${HAVE_MPI}" = "true" ; then
+        if test "${MPI_LAUNCHER}" ; then
```
jougs, line 496:
```suggestion
    if test "${HAVE_MPI}" = "true"; then
        if test "${MPI_LAUNCHER}"; then
```

jougs (approved):
Thanks!

PR #2332, opened 2022-03-08
Return empty dictionary if `get()` is called on empty SynapseCollection

### LINKED ISSUE #2318: Inconsistent returned type in `nest.GetConnections().get()`
**Describe the bug**
Once the network is built by calling the connect function in nest, the user can inspect the connection by calling `nest.GetConnections().get()`. The return type of this function is a dictionary, so in case of calling `nest.GetConnections().get()` without previously making any connections, we should expect to get any empty dictionary, but instead an empty `tuple` is returned.

stinebuu:
This fixes #2318.

### DIFF
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -795,9 +795,12 @@ class SynapseCollection:
         if pandas_output and not HAVE_PANDAS:
             raise ImportError('Pandas could not be imported')
 
-        # Return empty tuple if we have no connections or if we have done a nest.ResetKernel()
+        # Return empty dictionary if we have no connections or if we have done a nest.ResetKernel()
         num_conns = GetKernelStatus('num_connections')  # Has to be called first because it involves MPI communication.
-        if self.__len__() == 0 or num_conns == 0:
+        if keys == None and ( self.__len__() == 0 or num_conns == 0 ):
+            return {}
+        elif self.__len__() == 0 or num_conns == 0:
+            # Return empty tuple if get is called with an argument
             return ()
 
         if keys is None:
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -797,11 +797,9 @@ class SynapseCollection:
 
         # Return empty dictionary if we have no connections or if we have done a nest.ResetKernel()
         num_conns = GetKernelStatus('num_connections')  # Has to be called first because it involves MPI communication.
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_types.py
```diff
@@ -795,9 +795,12 @@ def get(self, keys=None, output=''):
...
 
-        # Return empty tuple if we have no connections or if we have done a nest.ResetKernel()
+        # Return empty dictionary if we have no connections or if we have done a nest.ResetKernel()
         num_conns = GetKernelStatus('num_connections')  # Has to be called first because it involves MPI communication.
-        if self.__len__() == 0 or num_conns == 0:
+        if keys == None and ( self.__len__() == 0 or num_conns == 0 ):
+            return {}
+        elif self.__len__() == 0 or num_conns == 0:
+            # Return empty tuple if get is called with an argument
             return ()
```
hakonsbm, line 804: It's cleaner to not check `self.__len__() == 0 or num_conns == 0` twice.
```suggestion
        if self.__len__() == 0 or num_conns == 0:
            # Return empty tuple if get is called with an argument
            return {} if keys is None else ()
```

hakonsbm (changes requested):
Could you also add a test of `conns.get()` on an empty SynapseCollection with specified keys (the case that should return `()`)?

testsuite/pytests/test_synapsecollection.py
```diff
@@ -363,15 +363,16 @@ def test_empty(self):
...
 
         nrns = nest.Create('iaf_psc_alpha', 2)
         nest.Connect(nrns, nrns, 'one_to_one')
         conns = nest.GetConnections()
         self.assertEqual(len(conns), 2)
 
         nest.ResetKernel()
-        self.assertEqual(conns.get(), ())
+        self.assertEqual(conns.get(), {})
         conns.set(weight=10.)
```
med-ayssar, line 376: Is not resetting the `kernel` equivalent to adding or deleting `connections`, and in that case we should get a warning from the `ConnectionManager` stating a similar `info` as the following: `ConnectionManager [Warning]: 
    New connections created, connection descriptors previously obtained using 
    'GetConnections' are now invalid`. In that case, the user should know that setting the `weight` in `conn` object won't be a valid operation, or maybe throw an error that  the `weight` isn't in the connection dictionary anymore.

med-ayssar (approved):
Otherwise, it looks good to me.

PR #2334, opened 2022-03-09
Fix broken links in release notes

jessica-mitchell:
This PR fixes broken internal references and links in the release notes pages, and removes the local toc.


jougs (approved):
LGTM. Merging without second review as this only concerns documentation.

Issue #2116, opened 2021-07-31
Report major (+ minor) NEST version even when cloned from git


clinssen:
#1223 changed the way NEST reports its version when it is cloned from git. Originally, the version consisted of hard-coded major, minor and patch numbers; afterwards, the git hash of the repository at time of cloning was reported.

An issue has emerged where, if NEST was installed from the git repository, NESTML is unable to query the NEST installation for its major version. In this case, only the hash is reported, and it cannot be determined whether the installed version was _based on_ NEST 2, or _based on_ NEST 3. (Due to API changes, NESTML needs to generate different code for them.)

By comparison, to version NESTML itself, we do a release under a major.minor number (for instance, "3.1"), and immediately after releasing, we update the version string on git to "3.1-post-dev". Perhaps a similar major.minor.patch prefix could be used by NEST, followed by the hash as a suffix.

See the template https://github.com/nest/nest-simulator/blob/master/libnestutil/config.h.in — for a relatively recent install, this gets rendered as

```C
#define NEST_VERSION_STRING "master@aa0095a22"
#define NEST_VERSION_BRANCH "master"
#define NEST_VERSION_SUFFIX ""
#define NEST_VERSION "master"
#define NEST_VERSION_GITHASH "aa0095a22"
```

The offending line is probably https://github.com/nest/nest-simulator/blob/b6c1e538610ca5746a0068dd0ac985aeda7787bb/cmake/NestVersionInfo.cmake#L71

Note that this seems to be inconsistent with the comment at https://github.com/nest/nest-simulator/blob/b6c1e538610ca5746a0068dd0ac985aeda7787bb/cmake/NestVersionInfo.cmake#L27
... which is closer to what we're after.

Ideally, these version numbers would also be queriable via ``nest-config``.

jougs:
For me, this boils down to the fact that NESTML (and possibly other external tools) need a way to query NEST version information in a nice way. As different parts of NEST are moving at a different pace, I would even suggest that we introduce different version numbers for the different parts, e.g. the module API, the model API, the PyNEST API, the NEST C++ API, and possibly others.

@sanjayankur31, can you please quickly comment here in how far #2250 can help here or relates to this? Thanks!

sanjayankur31:
The changes we've made in #2250 do allow the different shared objects to be versioned independently of each other and of the nest release value, ideally following the scheme recommended by the libtool documentation:

https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

[...]

PR #2336, opened 2022-03-10
Rename `calibrate` method to `pre_run_hook`

### LINKED ISSUE #2060: Refactor ``Node::calibrate()`` method
The ``calibrate()`` method should be renamed to something more descriptive, such as ``recompute_internal_variables()``.

It should be removed from the ``Node`` base class and no longer be called from ``NodeManager::prepare_node()``. Instead, it should be a private method, that individual ``Node`` classes call internally from three places: their constructor, their copy constructor, and ``SetStatus()``.

pnbabu:
Fixes #2060 

The `calibrate()` method recomputes the internal variables and is called for every call of simulation. It is now renamed to`pre_run_hook()` which is more descriptive.

### DIFF
--- a/doc/userdoc/guides/using_nest_with_music.rst
+++ b/doc/userdoc/guides/using_nest_with_music.rst
@@ -245,7 +245,7 @@ which yields the following output:
 
 ::
 
-	Nov 23 11:18:23 music_message_in_proxy::calibrate() [Info]:
+	Nov 23 11:18:23 music_message_in_proxy::pre_run_hook() [Info]:
 		Mapping MUSIC input port 'msgdata' with width=0 and acceptable latency=0
 		ms.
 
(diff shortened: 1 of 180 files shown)


jougs (commented):
Looks mostly good to me minus the changes in the cm_* models, where some of the functions should probably be left with the name `calibrate()`, as they are actually calibrating some currents based on some other quantity.

@WillemWybo could you please comment on the changes in those models and say which should be kept with the old name?

WillemWybo:
> Looks mostly good to me minus the changes in the cm_* models, where some of the functions should probably be left with the name `calibrate()`, as they are actually calibrating some currents based on some other quantity.
> 
> @WillemWybo could you please comment on the changes in those models and say which should be kept with the old name?

The calibrate functions mostly compute some quantities based on the simulation time step, so the `pre_run_hook` name seems fine to me. The calibrate function in `cm_default` also sets the all the needed pointers between model variables. Also seems reasonable to me that this would fall under the `pre_run_hook` name.

jougs (approved):
Thanks for your assessment, @WillemWybo. I thus concur and approve.

clinssen (approved):
Thank you! Please make concomitant changes in the nest-extension-module repository.

This is tricky to support in NESTML due to #2116, but we should probably deal with that elsewhere.

PR #2339, opened 2022-03-11
Handle single node in PlotLayer

ackurth:
Currently, the following code plotting a single node with spatial metadata with internal plotting routines, fails with a Python level ```TypeError```:
```
import nest

num_neurons = 1
extent = 2

neuron = nest.Create(
    'iaf_psc_exp',
    num_neurons,
    positions=nest.spatial.free(
        pos=nest.random.uniform(min=-extent / 2, max=extent / 2),
        num_dimensions=2,
        edge_wrap=True)
)
nest.PlotLayer(neuron)
```
This PR tries to fix this problem.

### DIFF
--- a/pynest/nest/lib/hl_api_spatial.py
+++ b/pynest/nest/lib/hl_api_spatial.py
@@ -946,7 +946,12 @@ def PlotLayer(layer, fig=None, nodecolor='b', nodesize=20):
         xctr, yctr = layer.spatial['center']
 
         # extract position information, transpose to list of x and y pos
-        xpos, ypos = zip(*GetPosition(layer))
+        try:
+            # try zipping coordinates
+            xpos, ypos = zip(*GetPosition(layer))
+        except TypeError:
+            # handle case of single node
+            xpos, ypos = GetPosition(layer)
 
         if fig is None:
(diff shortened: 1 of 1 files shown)


stinebuu (approved):
Thanks!

hakonsbm (changes requested):
Although the single-node case results in a `TypeError`, I think it's more straight-forward to have something like
```python
if (len(layer) == 1):
    pos = GetPosition(layer)
else:
    pos = zip(*GetPosition(layer))
```
or even as a single line
```python
pos = GetPosition(layer) if len(layer) == 1 else zip(*GetPosition(layer))
```

ackurth:
> Although the single-node case results in a `TypeError`, I think it's more straight-forward to have something like
> 
> ```python
> if (len(layer) == 1):
>     pos = GetPosition(layer)
> else:
>     pos = zip(*GetPosition(layer))
> ```
> 
> or even as a single line
> 
> ```python
> pos = GetPosition(layer) if len(layer) == 1 else zip(*GetPosition(layer))
> ```

I changed to code accordingly.

hakonsbm (approved):
Thanks!

PR #2345, opened 2022-03-17
Correct doc for with-python CMake option and remove Python 2 check.

heplesser:
- For some time, only ON and OFF have been valid values for `-Dwith-python`. This PR fixes the documentation on this point.
- The PR also removes a check for `-Dwith-python=2` which is not longer relevant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ endif()
 ################################################################################
 
 # use Python to build PyNEST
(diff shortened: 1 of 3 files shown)


jessica-mitchell (approved):
looks good from my end!

stinebuu:
Looks good! But could you also update the docstring in `CMakeLists.txt`?

https://github.com/nest/nest-simulator/blob/277114f5ea45c6b34bb33abe2d13e9edd6468249/CMakeLists.txt#L47-L48

stinebuu (approved):

PR #2346, opened 2022-03-17
Enable full macOS build as part of CI

heplesser:
This PR fixes issues with the CI setup for macOS, so that we can run the full testsuite for macOS (except MUSIC, libneurosim, Sionlib). To achieve this, Boost-installation is done via apt/brew now.


gtrensch (approved):
Looks good to me. :+1: Thanks for this contribution.

lekshmideepu (approved):
Thank you! Looks good to me as well 👍

PR #2347, opened 2022-03-17
Fix clang 13 pedantic warnings

heplesser:
This PR fixes warnings emitted by Clang 13 when run with `-pedantic -Wextra -Woverloaded-virtual -Wno-unknown-pragmas -Wall`. Almost all of the warnings point out that defining a copy constructor without an assignment operator is deprecated (see also https://stackoverflow.com/questions/51863588/warning-definition-of-implicit-copy-constructor-is-deprecated).

Note that building pynestkernel will issue lot's of warnings. This is Cython problem and reported there (https://github.com/cython/cython/pull/4687).

@poojanbabu This may also be relevant for NESTML code generation, therefore I assign you as reviewer.


hakonsbm (approved):
LGTM

PR #2349, opened 2022-03-18
Fix CMake check of Random123 generators

hakonsbm:
The CMake if-test does not short-circuit, so both `COMPILE_RESULT` and `RUN_RESULT` were always checked. But if the compilation fails, `RUN_RESULT` is undefined and CMake crashes. 

The if-test is changed to test `COMPILE_RESULT` and `RUN_RESULT` separately. I've verified that it works now in cases with failing compilation, with failing tests, and with working compilation and passing tests.


heplesser (approved):

gtrensch (approved):
Looks good to me.

PR #2351, opened 2022-03-18
Fix version information when not related to a release

terhorstd:
Revert accidental merge of last release (e396edb4f107b8).

### DIFF
--- a/cmake/NestVersionInfo.cmake
+++ b/cmake/NestVersionInfo.cmake
@@ -58,7 +58,7 @@ macro(get_version_info)
     endif()
 
     if (NOT NEST_VERSION_BRANCH)
-        set(NEST_VERSION_BRANCH "3.2")
+        set(NEST_VERSION_BRANCH "UNKNOWN")
     endif()
 
     string(SUBSTRING "${NEST_VERSION_BRANCH}" 0 5 isRelease)


jougs (approved):

clinssen:
Could you comment on how this commit affects https://github.com/nest/nest-simulator/issues/2116?

jougs:
@terhorst: maybe we should rather put `3.4-pre` in that field instead of `UNKNOWN`? @clinssen: would that help with #2116?

clinssen:
We could say, if we are based on a branch (``if (NEST_VERSION_BRANCH)``), then append a "hand-rolled" string to indicate the version it was based on. For instance,
```sh
if (NEST_VERSION_BRANCH)
    set(NEST_VERSION_BRANCH "${NEST_VERSION_BRANCH}@3.4-pre")
endif()
```
results in the version string
```c
#define NEST_VERSION_STRING "master@3.4-pre@71c85a0f8"
```
Undoubtedly you and Dennis will have a better idea for syntax than my repeated at symbol. We are not presently compliant with https://peps.python.org/pep-0386/; how badly would we want this? An alternative, semantic versioning-compliant string would be
```
3.4-pre+71c85a0f8
```
or
```
3.4-pre+master-71c85a0f8
```

Updating the hand-rolled string would in any case need to be part of the steps involved in doing a release.

PR #2353, opened 2022-03-22
Remove pathlib from requirements.txt as this is a Python-builtin now

steffengraber:
This PR replaces the no longer maintained and outdated version of the pathlib module with a new module in requirements.txt.

The problem was noticed in the automatic build of the NEST docker containers and is fixed with the exchange of the packages.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -13,7 +13,7 @@ nbsphinx
 numpy
 numpydoc
 path.py
-pathlib
+pathlib2
 pytest
 recommonmark
 scipy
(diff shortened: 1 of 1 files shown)


heplesser:
`pathlib` is part of standard python now (https://docs.python.org/3/library/pathlib.html, since 3.4) and the PyPi packages `pathlib` and its successors `pathlib2` are backports for earlier Python versions. Is there any reason we need the `pathlib` package in requirements.txt at all?

heplesser (approved):

jougs (approved):
LGTM!

PR #2356, opened 2022-03-22
Create classes for the different propagators

stinebuu:
This PR follows up the discussion in #2294 and #2017. I have created two propagator classes, `PropagateExp` and `PropagateAlpha` for calculating propagators for models with exponential and alpha postsynaptic currents. Both classes have an `evaluate` function where the exp version just calculates `P31`, and the alpha version calculates `P31` and `P32` simultaneously. This way, the exponentials are calculated fewer times than what we do currently. I have not been able to get the simulation time for `iaf_psc_alpha_ps` to precisely match `iaf_psc_alpha_canon`, but the times are closer than with master, see results below.

With this new version, getting P31 and P32 is handled slightly differently than before:
Old version
```
V_.P31_ = propagator_31( P_.tau_syn_, tau_m, P_.c_m_, h );
V_.P32_ = propagator_32( P_.tau_syn_, tau_m, P_.c_m_, h );
```
New version exp model
```
V_.P21_ = PropagatorExp( P_.tau_syn_, tau_m, P_.c_m_ ).evaluate( h );
```
New version alpha model
```
std::tie( V_.P31_, V_.P32_ ) = PropagatorAlpha( P_.tau_syn_, tau_m_, P_.c_m_ ).evaluate( h );
```

Here are some timing results. All measured on my computer.
| model  | Simulation time for `brunel_ps.sli` | Time `ArtificialSyncrony.sli`
| ------------- | ------------- | ------------- |
| `iaf_psc_alpha_ps` with `propagator` class  | 132.82 sec  | 7 min 23 sec |
| master version of `iaf_psc_alpha_ps`  | 222.86 sec  | 9 min 18 sec |
| master version of `iaf_psc_alpha_canon`  | 95.03 sec | 6 min 33 sec |

`iaf_psc_alpha_canon` does not use `propagate_31` and `propagate_32` in it's [`propogate_` function](https://github.com/nest/nest-simulator/blob/master/models/iaf_psc_alpha_canon.cpp#L473). Using canon's version in the ps model decrease the time further, see comment [here](https://github.com/nest/nest-simulator/pull/2294#issuecomment-1065174892).

@heplesser

### DIFF
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -82,12 +82,12 @@ PropagatorExp::evaluate( double h ) const
 }
 
 PropagatorAlpha::PropagatorAlpha()
-  : PropagatorExp( 0., 0.,  0. )
+  : PropagatorExp( 0., 0., 0. )
 {
 }
 
 PropagatorAlpha::PropagatorAlpha( double tau_syn, double tau, double c )
-  : PropagatorExp( tau_syn, tau,  c )
+  : PropagatorExp( tau_syn, tau, c )
 {
 }
 
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -57,7 +57,7 @@ PropagatorExp::evaluate( double h ) const
 
   double P32 = gamma_ * exp_h_tau_syn * expm1_h_tau;
 
-  if ( tau_m_ == tau_syn_ or std::abs( tau_m_ - tau_syn_ ) < 0.1 )
+  if ( std::abs( tau_m_ - tau_syn_ ) < 0.1 )
   {
     const double exp_h_tau = std::exp( -h / tau_m_ );
 
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -86,8 +86,8 @@ PropagatorAlpha::PropagatorAlpha()
 {
 }
 
-PropagatorAlpha::PropagatorAlpha( double tau_syn, double tau, double c )
-  : PropagatorExp( tau_syn, tau, c )
+PropagatorAlpha::PropagatorAlpha( double tau_syn, double tau_m, double c_m )
+  : PropagatorExp( tau_syn, tau_m, c_m )
 {
 }
 
--- a/libnestutil/propagator_stability.cpp
+++ b/libnestutil/propagator_stability.cpp
@@ -28,84 +28,69 @@
 // Includes from libnestutil:
 #include "numerics.h"
 
-propagator::propagator( )
+propagator::propagator()
   : alpha_( 0 )
   , beta_ ( 0 )
   , gamma_ ( 0 )
-  , gamma_sq_ ( 0 )
+  , tau_syn_ ( 0 )
+  , tau_ ( 0 )
+  , c_ ( 0 )
 {
 }
 
-void propagator::calculate_constants( double tau_syn, double tau, double c )
+propagator::propagator( double tau_syn, double tau, double c )
+  : tau_syn_ ( tau_syn )
+  , tau_ ( tau )
+  , c_ ( c )
 {
-  alpha_ = 1 / ( 3. * c * tau * tau ) * ( tau_syn - tau );
+  alpha_ = 1 / ( c * tau * tau ) * ( tau_syn - tau );
   beta_ = tau_syn  * tau / (tau - tau_syn);
   gamma_ = 1 / c * beta_;
-  gamma_sq_ = 1 / c / ( ( 1 / tau_syn - 1 / tau ) * ( 1 / tau_syn - 1 / tau ) );
 }
 
-double
-propagator::propagator_31( double tau_syn, double tau, double C, double h ) const
+void propagator::update_constants( double tau_syn, double tau, double c )
 {
-  const double exp_h_tau_syn = std::exp( -h / tau_syn );
-  const double expm1_h_tau = numerics::expm1( -h / tau + h / tau_syn );
+  alpha_ = 1 / ( c * tau * tau ) * ( tau_syn - tau );
+  beta_ = tau_syn  * tau / (tau - tau_syn);
+  gamma_ = 1 / c * beta_;
 
+  tau_syn_ = tau_syn;
+  tau_ = tau;
+  c_ = c;
+}
 
-  //const double P31_linear = 1 / ( 3. * C * tau * tau ) * h * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
+propogate
+propagator::propagate( double h ) const
+{
+  const double exp_h_tau_syn = std::exp( -h / tau_syn_ );
+  const double expm1_h_tau = numerics::expm1( -h / tau_ + h / tau_syn_ );
 
-  //const double P31 = 1 / C
-  //  * ( std::exp( -h / tau_syn ) * numerics::expm1( -h / tau + h / tau_syn ) / ( tau / tau_syn - 1 ) * tau
-  //    - h * std::exp( -h / tau_syn ) )
-  //  / ( -1 - -tau / tau_syn ) * tau;
+  propogate P;
 
-  const double P31 = gamma_ * ( beta_ * exp_h_tau_syn * expm1_h_tau - h * exp_h_tau_syn);
+  P.P31 = gamma_ * ( beta_ * exp_h_tau_syn * expm1_h_tau - h * exp_h_tau_syn);
+  P.P32 = gamma_ * exp_h_tau_syn * expm1_h_tau;
 
-  //const double P31_singular = h * h / 2 / C * std::exp( -h / tau );
-  //const double dev_P31 = std::abs( P31 - P31_singular );
-
-  if ( tau == tau_syn or std::abs( tau - tau_syn ) < 0.1 )
+  if ( tau_ == tau_syn_ or std::abs( tau_ - tau_syn_ ) < 0.1 )
   {
-    const double P31_linear = alpha_ * h * h * h * std::exp( -h / tau );
-    const double P31_singular = h * h / 2 / C * std::exp( -h / tau );
-    const double dev_P31 = std::abs( P31 - P31_singular );
+    const double exp_h_tau = std::exp( -h / tau_ );
+
+    const double P31_linear = alpha_ * h * h * h * exp_h_tau / 3.;
+    const double P31_singular = h * h / 2 / c_ * exp_h_tau;
+    const double dev_P31 = std::abs( P.P31 - P31_singular );
 
     if ( dev_P31 > 2 * std::abs( P31_linear ) )
     {
-      return P31_singular;
+      P.P31 =  P31_singular;
+    }
+
+    const double P32_linear = alpha_ * h * h * exp_h_tau / 2.;
+    const double P32_singular = h / c_ * exp_h_tau;
+    const double dev_P32 = std::abs( P.P32 - P32_singular );
+
+    if ( dev_P32 > 2 * std::abs( P32_linear ) )
+    {
+      P.P32 = P32_singular;
     }
   }
-
-  return P31;
-}
-
-double
-propagator::propagator_32( double tau_syn, double tau, double C, double h ) const
-{
-  const double P32_linear = 1 / ( 2. * C * tau * tau ) * h * h * ( tau_syn - tau ) * std::exp( -h / tau );
-  const double P32_singular = h / C * std::exp( -h / tau );
-  const double P32 =
-    -tau / ( C * ( 1 - tau / tau_syn ) ) * std::exp( -h / tau_syn ) * numerics::expm1( h * ( 1 / tau_syn - 1 / tau ) );
-
-  const double dev_P32 = std::abs( P32 - P32_singular );
-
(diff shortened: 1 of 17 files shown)


libnestutil/propagator_stability.cpp
```diff
@@ -28,43 +28,69 @@
...
-  else
+propagators
+propagator::propagate( double h ) const
+{
+  const double exp_h_tau_syn = std::exp( -h / tau_syn_ );
+  const double expm1_h_tau = numerics::expm1( -h / tau_ + h / tau_syn_ );
+
+  propagators P;
+
+  P.P31 = gamma_ * ( beta_ * exp_h_tau_syn * expm1_h_tau - h * exp_h_tau_syn);
```
clinssen, line 70:
```suggestion
  P.P31 = gamma_ * ( beta_ * exp_h_tau_syn * expm1_h_tau - h * exp_h_tau_syn );
```

libnestutil/propagator_stability.cpp
```diff
@@ -28,43 +28,69 @@
...
-    return P32;
+    const double exp_h_tau = std::exp( -h / tau_ );
+
+    const double P31_linear = alpha_ * h * h * h * exp_h_tau / 3.;
+    const double P31_singular = h * h / 2 / c_ * exp_h_tau;
+    const double dev_P31 = std::abs( P.P31 - P31_singular );
+
+    if ( dev_P31 > 2 * std::abs( P31_linear ) )
+    {
+      P.P31 =  P31_singular;
```
clinssen, line 83:
```suggestion
      P.P31 = P31_singular;
```

models/iaf_psc_exp_multisynapse.cpp
```diff
@@ -294,7 +294,9 @@ nest::iaf_psc_exp_multisynapse::calibrate()
   {
     V_.P11_syn_[ i ] = std::exp( -h / P_.tau_syn_[ i ] );
     // these are determined according to a numeric stability criterion
-    V_.P21_syn_[ i ] = propagator_32( P_.tau_syn_[ i ], P_.Tau_, P_.C_, h );
+    propagator prop( P_.tau_syn_[ i ], P_.Tau_, P_.C_);
+    propagators propagators = prop.propagate( h );
```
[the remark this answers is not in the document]
stinebuu: I will add a const qualifier. I had problems naming both the struct and the variable, but I will also try renaming, as I agree it is a bit confusing.

models/gif_psc_exp.cpp
```diff
@@ -292,8 +292,10 @@ nest::gif_psc_exp::calibrate()
   const double tau_m = P_.c_m_ / P_.g_L_;
 
   // these are determined according to a numeric stability criterion
-  V_.P21ex_ = propagator_32( P_.tau_ex_, tau_m, P_.c_m_, h );
-  V_.P21in_ = propagator_32( P_.tau_in_, tau_m, P_.c_m_, h );
+  PropagatorExp prop_ex( P_.tau_ex_, tau_m, P_.c_m_ );
+  PropagatorExp prop_in( P_.tau_in_, tau_m, P_.c_m_ );
+  V_.P21ex_ = prop_ex.evaluate( h );
+  V_.P21in_ = prop_in.evaluate( h );
```
heplesser, line 298: For all models which do not need the propagator object elsewhere, I'd write directly
```suggestion
  V_.P21ex_ = PropagatorExp( P_.tau_ex_, tau_m, P_.c_m_ ).evaluate( h );
  V_.P21in_ = PropagatorExp( P_.tau_in_, tau_m, P_.c_m_ ).evaluate( h );
```

libnestutil/propagator_stability.h
```diff
@@ -23,9 +23,79 @@
...
+   * Empty constructor needed for initialization of buffers.
+   */
+  PropagatorAlpha();
+
+  /**
+   * @param tau_syn Time constant of synaptic current in ms
+   * @param tau Membrane time constant in ms
+   * @param Membrane capacitance in pF
+   */
+  PropagatorAlpha( double tau_syn, double tau, double c_m );
```
heplesser, line 89:
```suggestion
  PropagatorAlpha( double tau_syn, double tau_m, double c_m );
```
with corresponding changes elsewhere.

libnestutil/propagator_stability.h
```diff
@@ -23,9 +23,79 @@
...
+   * Empty constructor needed for initialization of buffers.
+   */
+  PropagatorExp();
+
+  /**
+   * @param tau_syn Time constant of synaptic current in ms
+   * @param tau Membrane time constant in ms
+   * @param Membrane capacitance in pF
+   */
+  PropagatorExp( double tau_syn, double tau, double c_m );
```
heplesser, line 55:
```suggestion
  PropagatorExp( double tau_syn, double tau_m, double c_m );
```
with corresponding changes elsewhere.

libnestutil/propagator_stability.h
```diff
@@ -23,9 +23,79 @@
...
+/**
+ * Propagator classes for handling similar tau_m and tau_syn_* time constants.
+ *
+ * Constants are calculated in the constructor, while
+ * propagator_31 and propagator_32 are calculated in `evaluate( h )`.
+ *
+ * Models with exponential postsynaptic currents should use PropagatorExp.
+ * Here `evaluate( h )` returns P32 as a double. Models with postsynaptic
+ * currents modeled as an alpha current on the other hand, should use PropagatorAlpha.
+ * P31 and P32 are then returned as a tuple, where P31 is the first variable.
```
heplesser, line 38: I find the comment confusing here, because the PropExp comment is much about PropAlpha. I'd suggest to comment PropExp only here and PropAlpha there.

libnestutil/propagator_stability.cpp
```diff
@@ -28,43 +28,115 @@
...
+std::tuple< double, double >
+PropagatorAlpha::evaluate( double h ) const
+{
+  const double exp_h_tau_syn = std::exp( -h / tau_syn_ );
+  const double expm1_h_tau = numerics::expm1( -h / tau_ + h / tau_syn_ );
+
+  double P31 = gamma_ * exp_h_tau_syn * ( beta_ * expm1_h_tau - h );
+  double P32 = gamma_ * exp_h_tau_syn * expm1_h_tau;
+
+  if ( tau_ == tau_syn_ or std::abs( tau_ - tau_syn_ ) < 0.1 )
```
heplesser, line 103: Looking at the code here again, I think it does not make much sense to have the first comparison, since it is included in the second part, and I assume that we do not save much by avoiding the subtraction. So
```suggestion
  if ( std::abs( tau_ - tau_syn_ ) < 0.1 )
```

libnestutil/propagator_stability.cpp
```diff
@@ -28,43 +28,115 @@
...
+  return P32;
+}
+
+PropagatorAlpha::PropagatorAlpha()
+  : PropagatorExp( 0., 0., 0. )
+{
+}
+
+PropagatorAlpha::PropagatorAlpha( double tau_syn, double tau, double c )
+  : PropagatorExp( tau_syn, tau, c )
```
heplesser, line 90:
```suggestion
PropagatorAlpha::PropagatorAlpha( double tau_syn, double tau_m, double c_m )
  : PropagatorExp( tau_syn, tau_m, c_m )
```

libnestutil/propagator_stability.cpp
```diff
@@ -28,43 +28,115 @@
...
-  const double P32 =
-    -tau / ( C * ( 1 - tau / tau_syn ) ) * std::exp( -h / tau_syn ) * numerics::expm1( h * ( 1 / tau_syn - 1 / tau ) );
+  const double exp_h_tau_syn = std::exp( -h / tau_syn_ );
+  const double expm1_h_tau = numerics::expm1( -h / tau_ + h / tau_syn_ );
 
-  const double dev_P32 = std::abs( P32 - P32_singular );
+  double P32 = gamma_ * exp_h_tau_syn * expm1_h_tau;
 
-  if ( tau == tau_syn or ( std::abs( tau - tau_syn ) < 0.1 and dev_P32 > 2 * std::abs( P32_linear ) ) )
+  if ( tau_ == tau_syn_ or std::abs( tau_ - tau_syn_ ) < 0.1 )
```
heplesser, line 60: See comment on corresponding line in PropAlpha below.

heplesser (changes requested):
@stinebuu Looks much better now, I think! Is the timing information in the PR descriptionfor this version of the original PR version? See some smaller suggestion in code below.

One aspect I am not quite happy about still is that the code for compute P32 is the same in both classes. How about the following design:
- A base class `Propagator` which holds all data members which we currently have.

- The base class defines `std::tuple <double, double, double> evaluate_P32_( double h )` which returns P32 and the `exp_h_tau_syn` and `expm1_h_tau` values.

- `PropagatorExp::evaluate()` then just calls `evaluate_P32_()` and returns the value for P32; if inlined, this should not cost anything extra.

- `PropagatorAlpha::evaluate()` calls `evaluate_P32_()` and then computes P31, using the returned valued for `exp_...` and `expm1_...`.

libnestutil/propagator_stability.h
```diff
@@ -23,9 +23,154 @@
...
+   */
+  std::tuple< double, double, double, double > evaluate_P32_( double h ) const;
+
+  double tau_syn_; //!< Time constant of synaptic current in ms
+  double tau_m_;   //!< Membrane time constant in ms
+  double c_m_;     //!< Membrane capacitance in pF
+
+  double alpha_; //!< 1/(c*tau*tau) * (tau_syn - tau)
+  double beta_;  //!< tau_syn * tau/(tau - tau_syn)
+  double gamma_; //!< beta_/c
```
heplesser, line 76: Since propagator parameters cannot change once objects are instantiated, maybe making them const can allow for further optimization.
```suggestion
  const double tau_syn_; //!< Time constant of synaptic current in ms
  const double tau_m_;   //!< Membrane time constant in ms
  const double c_m_;     //!< Membrane capacitance in pF

  const double alpha_; //!< 1/(c*tau*tau) * (tau_syn - tau)
  const double beta_;  //!< tau_syn * tau/(tau - tau_syn)
  const double gamma_; //!< beta_/c
```

libnestutil/propagator_stability.h
```diff
@@ -23,9 +23,154 @@
...
+   * Calculate propagators 31 and 32.
+   *
+   * @param h time step
+   *
+   * @returns tuple containing propagators 31 and 32
+   */
+  std::tuple< double, double > evaluate( double h ) const;
+};
+
+inline std::tuple< double, double, double, double >
```
heplesser, line 143: This method is so large that I see no point in inlining, so I suggest to move it to the cpp-file.

models/iaf_psc_alpha_ps.h
```diff
@@ -265,6 +268,10 @@ class iaf_psc_alpha_ps : public ArchivingNode
    */
   void emit_instant_spike_( Time const& origin, const long lag, const double spike_offset );
 
+  /** Propagator object for updating synaptic components */
+  PropagatorAlpha propagator_ex_;
+  PropagatorAlpha propagator_in_;
```
heplesser, line 273: Shouldn't these be part of `V_`?

models/iaf_psc_exp_ps.h
```diff
@@ -260,6 +263,10 @@ class iaf_psc_exp_ps : public ArchivingNode
    */
   void emit_instant_spike_( const Time& origin, const long lag, const double spike_offset );
 
+  /** Propagator object for updating synaptic components */
+  PropagatorExp propagator_ex_;
+  PropagatorExp propagator_in_;
```
heplesser, line 268: Shouldn't these be part of `V_`?

models/iaf_psc_exp_ps_lossless.h
```diff
@@ -271,6 +274,10 @@ class iaf_psc_exp_ps_lossless : public ArchivingNode
    */
   double is_spike_( const double );
 
+  /** Propagator object for updating synaptic components */
+  PropagatorExp propagator_ex_;
+  PropagatorExp propagator_in_;
```
heplesser, line 279: See above.

heplesser (changes requested):
@stinebuu This looks very fine now, just a few small suggestions left.

libnestutil/propagator_stability.h
stinebuu: @heplesser If I make these `const` I get the following error message, one per variable: `error: non-static const member ‘const double Propagator::tau_syn_’, can’t use default assignment operator`. I am not quite sure how to fix this though.

 `‘Propagator& Propagator::operator=(Propagator&&)` is also implicitly deleted, so it is no longer possible to do `propagator_ex_ = PropagatorAlpha( P_.tau_syn_ex_, P_.tau_m_, P_.c_m_ );`

heplesser: Given that we have and `update_constants()` method, making the members `const` seems contradictory, but then so seems the idea of updating *constants*. I took another look at the code and got some ideas, which I will try out and then get back.

heplesser: @stinebuu Please excuse my previous nonsensical comment—I have looked at an outdated version of the code due to some Git confusion. I think we just need to leave things as they are, i.e., non-const.

heplesser:
This PR is replaced by #2385, which builds on the material in this PR.

PR #2365, opened 2022-04-04
Add GIF / GLIF acronyms to model title in docs

jessica-mitchell:
To help alleviate confusion between the two types of models, I've added the acronyms to the titles in the model userdocs section.

Not sure if further differentiation would be helpful (e.g., do we need to add a tag for leaky models?). @heplesser @diesmann perhaps have ideas.

### DIFF
--- a/models/gif_cond_exp.h
+++ b/models/gif_cond_exp.h
@@ -53,7 +53,7 @@ extern "C" int gif_cond_exp_dynamics( double, const double*, double*, void* );
 Short description
 +++++++++++++++++
 
-Conductance-based generalized integrate-and-fire neuron model
+Conductance-based generalized integrate-and-fire neuron (GIF) model
 
 Description
 +++++++++++
--- a/models/gif_cond_exp.h
+++ b/models/gif_cond_exp.h
@@ -53,7 +53,7 @@ extern "C" int gif_cond_exp_dynamics( double, const double*, double*, void* );
 Short description
 +++++++++++++++++
 
-Conductance-based generalized integrate-and-fire neuron (GIF) model - Gerstner Group
+Conductance-based generalized integrate-and-fire neuron (GIF) model (from the Gerstner lab)
 
 Description
 +++++++++++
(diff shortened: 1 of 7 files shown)


clinssen:
Thank you for the PR. This is just an unfortunate naming clash (or almost-clash, GIF vs. GLIF). Both models are leaky (virtually all models in neuroscience are leaky, so I would not make a separate tag for this). If we want to disambiguate further, perhaps we could include the authors' (or first author's) name in the model name; or "Allen Institute" for GLIF.

jessica-mitchell:
OK thanks @clinssen I kind of thought it was common, but was just going based on name :D.
If it's agreed that adding Allen Institute is helpful then we could add it to the title of GLIF models. Especially if that's a search term people might use `ctl+f` for.

models/gif_cond_exp.h
```diff
@@ -53,7 +53,7 @@ extern "C" int gif_cond_exp_dynamics( double, const double*, double*, void* );
 Short description
 +++++++++++++++++
 
-Conductance-based generalized integrate-and-fire neuron (GIF) model
+Conductance-based generalized integrate-and-fire neuron (GIF) model - Gerstner Group
```
heplesser, line 56: How about
```suggestion
Conductance-based generalized integrate-and-fire neuron (GIF) model (from the Gerstner lab)
```

models/glif_cond.h
```diff
@@ -47,7 +47,7 @@
 Short description
 +++++++++++++++++
 
-Conductance-based generalized leaky integrate and fire (GLIF) model
+Conductance-based generalized leaky integrate and fire (GLIF) model - Allen Institute
```
heplesser, line 50:
```suggestion
Conductance-based generalized leaky integrate and fire (GLIF) model (from the Allen Institute)
```

heplesser (commented):
@jessica-mitchell I added suggestions for slightly changed wording.

@diesmann @abigailm I'd much appreciate your comments on what would be the optimal way to make the source and thus nature of the GIF and GLIF models obvious to reduce the risk of confusion.

heplesser (approved):

PR #2372, opened 2022-04-17
first working prototype

med-ayssar:


heplesser:
@med-ayssar Could you change the title of the PR to something more informative? It now sounds as if we, 26 years after the first commit, had finally gotten to the first working NEST prototype ;).

terhorstd:
@med-ayssar how is work going with this one? Could you also edit the description to something more like a summary what these changes actually change? "No description" is not very helpful, as you might agree.

jougs:
@med-ayssar: Should we close this? I think this is still very much work in progress and having this here just causes people to get mails on every push without a lot of benefit for anyone.

med-ayssar:
Really? I thought when it is draft state, nobody will be informed about the subsequent commits.

jougs:
Hmmm... Now that you say this, I indeed can't seem to find any mails. In this case, I guess you can leave this open. Sorry for the noise!

PR #2373, opened 2022-04-20
Ensure that thread-shared variable is initialized by single thread

heplesser:
This PR ensures that the thread-global variable `decrease_buffer_size_spike_data_` is initialized by a single thread before entering the region in which it is used.


jougs:
@heplesser: I am generally positive towards the changes, but can you please provide a description of the problem this is solving or a link to an issue?

heplesser:
> @heplesser: I am generally positive towards the changes, but can you please provide a description of the problem this is solving or a link to an issue?

I have added a short description and merged master so the actual change becomes better visible.

jougs (approved):
Thanks!

suku248 (approved):
Thanks @heplesser !

Just for the record: This is one of the suspicious lines of code we checked during one of our Hackathons?
And we concluded that:
- not having this in a single environment did not result in erroneous behavior, but it has the potential to cause errors in thread-parallel code, and
- runtime measurements showed that the single environment does have an effect on the runtime?

heplesser:
> Just for the record: This is one of the suspicious lines of code we checked during one of our Hackathons? And we concluded that:
> 
> * not having this in a single environment did not result in erroneous behavior, but it has the potential to cause errors in thread-parallel code, and
> * runtime measurements showed that the single environment does have an effect on the runtime?

Yes to all!

PR #2374, opened 2022-04-20
Report actual number of synapse created in PyNEST Brunel examples

heplesser:
This PR ensures that PyNEST examples of Brunel networks report the actual number of excitatory and inhibitory connections created. Previously, the actual total number of synapses was reported, while the expected number was reported for excitatory and inhibitory synapses.


clinssen:
Question: is ``CE * N_neurons`` always precisely equal to ``nest.GetDefaults("excitatory")["num_connections"]`` (both of type int)?

heplesser:
> Question: is `CE * N_neurons` always precisely equal to `nest.GetDefaults("excitatory")["num_connections"]` (both of type int)?

If the script is correct and NEST works correctly, then the following `excitatory` synapses should be created:
- `CE * N_neurons` synapses from excitatory neurons to neurons (E and I)
- `N_neurons` synapses from the poisson generator to neurons (E and I)
- `2 * N_rec` synapses from the recorded E and I neurons to the respective spike recorders

Furthermore `CI * N_neurons` synapses of type `inhibitory` should be created from inhibitory neurons to neurons (E and I).

So the old numbers E-synapse numbers were not quite correct, since they ignored the connections to the recorders, and the sums did not work out correctly.

I think it makes most sense to obtain the total number of synapses and the E/I numbers in the same way to ensure consistency—either expected numbers or measured. I propose to use the measured numbers in this PR.

mlober (approved):

nicolossus (approved):
Thank you!

PR #2296, opened 2022-02-16
Return invalid_thread_ for proxynode

StefanBruens:
Not returning in a non-void function causes undefined-behavior.

Fixes #2295

PR #2376, opened 2022-04-21
Ensure consistent return values in "impossible" situations even if assertions are compiled out

### LINKED ISSUE #2295: proxynode::get_thread should return defined value (or throw)
Currently, when the code is not compiled in DEBUG mode, get_thread will do sometthing random:

https://github.com/nest/nest-simulator/blob/6649606756d1a5fb8ec67550cec35abae1d4b4ae/nestkernel/proxynode.h#L114-L118

It should either return a defined value (e.g. `invalid_thread_`), or throw.

heplesser:
This PR addresses all parts of #2295 and fixes some minor issues (typos, inconsistent naming) as well. It replaces #2296.

Fixes #2295.

### DIFF
--- a/models/bernoulli_synapse.h
+++ b/models/bernoulli_synapse.h
@@ -132,7 +132,7 @@ public:
     port
     handles_test_event( SpikeEvent&, rport )
     {
-      return invalid_port_;
+      return invalid_port;
     }
   };
 
--- a/models/clopath_synapse.h
+++ b/models/clopath_synapse.h
@@ -166,7 +166,7 @@ public:
     port
     handles_test_event( SpikeEvent&, rport )
     {
-      return invalid_port_;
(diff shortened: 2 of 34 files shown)


StefanBruens:
AFAICS this breaks API, as the constant names are part of public headers.

heplesser:
> AFAICS this breaks API, as the constant names are part of public headers.

So far, we have never promised a stable C++ API. The closest we have come to an explicit API is the set of functions defined in `nest.h`. We rather focus on stability of the user interface at the Python level.

Furthermore, I'd argue that the names that are changed in `nest_types.h` ended in `_` in the past, marking them as private implementation detail.

hakonsbm (approved):

jougs (approved):

PR #2377, opened 2022-04-22
Prohibit multimeter-recording from noise_generator in multithreaded mode

### LINKED ISSUE #2282: Multimeter records from noise_generator unreliably and dependent on local_num_threads
**Describe the bug**
The multimeter does not record current changes from a noise_generator reliably. This behavior seems to change in dependence of the number of threads used, as well as the number of neurons  (or perhaps nodes in general) in the simulation. When connection fails, all current values inside the multimeter are set to 0.0.

**To Reproduce**
see screenshots/attached file

**Expected behavior**
multimeter collecting information from the noise_generator consistently

**Screenshots**
image
image

 - OS: Ubuntu 20.0.4 LTS
 - Shell: bash (also attempted through jupyter notebook)
 - Python-Version: Python 3.8.10
 - NEST-Version: nest-3.1, 3.0
 - Installation: apt for 3.1, docker for 3.0

**Additional context**
the noise generator works as expected, and so does the multimeter when connecting the it to a neuron directly. the issue only arises when connecting multimeters to the noise_generator on more than one thread
[noise_generator_minimal_example.zip](https://github.com/nest/nest-simulator/files/7980216/noise_generator_minimal_example.zip)

heplesser:
As described in #2282, recording the average current produced by the `noise_generator` correctly when running on multiple threads would challenging to implement. Since I consider this a corner case anyways, I do not believe the effort would be worth it, as it would considerably complicate our code. Therefore, this PR explicitly prevents connections from `multimeter` to `noise_generator` if NEST runs with multiple threads.

The problem does not affect other generators (ac, dc, step_current, step_rate, sinusoidal_{gamma,poisson}) because they send deterministic values to the recorder which are identical for all threads so that it does not matter from which thread one records.

Fixes #2282.


testsuite/regressiontests/issue-2282.sli
```diff
@@ -0,0 +1,71 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/** @BeginDocumentation
+
+Name: testsuite::issue-2282 Ensure multimeter-recording from noise_generator prohibited in if multithreaded
```
hakonsbm, line 25:
```suggestion
Name: testsuite::issue-2282 Ensure multimeter-recording from noise_generator is prohibited if multithreaded
```

hakonsbm (changes requested):
Just a small suggestion for a clarification in the documentation.

hakonsbm (approved):
Thanks!

jougs (changes requested):
I would actually also want to see a hint about this restriction as a [RST `.. note::` admonition](https://docutils.sourceforge.io/docs/ref/rst/directives.html#toc-entry-1) around [line 131 of `noise_generator.h](https://github.com/nest/nest-simulator/pull/2377/files#diff-4342293d07738fb595c94ee7d3e0fbd1d7782a8c1e279c73d2da4bf3805b2225R131).

I also suggest to convert the test to Python, so you could actually check if the right exception is thrown.

heplesser:
@jougs I have now revised the documentation for the noise generator. Could you take a look? I do not quite see the point of converting to a Python-based test. I would allow me to test that exception `nest.lib.hl_api_exceptions.NESTErrors.Recording from a noise_generator is only possible in single-threaded mode.` is thrown, but extracting the interesting last part from it would be a bit of an effort and any change in the wording of the error message would break the test—or am I overlooking something here?

@jessica-mitchell Could you take a look at the revised documentation?

models/noise_generator.h
```diff
@@ -49,85 +49,72 @@ Generate a Gaussian white noise current
...
 
-   \Sigma \rightarrow std / C_m * \sqrt(dt \cdot \tau / 2).
+.. admonition:: Recording the generated current
 
-To obtain comparable results for different values of dt, you must adapt std.
+   You can use a :doc:`multimeter <multimeter>` to record the average current sent to all targets for each time step
+   if simulating on a single thread; multiple MPI processes with one thread each also work. In this case,
+   the recording interval of the multimeter should be the equal to the simulation resolution to avoid confusing effects
+   due to offset or drift between the recording times of the multimeter and the switching times of the
+   noise generator. In multi-threaded mode, recording is noise current is prohibited for technical reasons.
```
jougs, line 94:
```suggestion
   noise generator. In multi-threaded mode, recording of noise currents is prohibited for technical reasons.
```

jougs (approved):
Just a single small suggestion.

heplesser:
> Just a single small suggestion.

Done :)

heplesser:
@jougs Could you resolve the one open conversation so that we can merge this one?

PR #2379, opened 2022-04-24
Ensure correct model id and model name when models are copied

### LINKED ISSUE #2378: `nest.CopyModel` doesn't assign `new_id` to the new copied model
**Describe the bug**
Once an instance of a copied model is created, `nest` returns again `Unknown` as the model name on the call of the `get` function.
**To Reproduce**
1. Copy the model
2. Create new instance
3. Call the get function
4. Observe the `model` and `model_id`  fields in the `dict`

**Expected behavior**
The copied model `id` must be a non-negative number and `new_model.id > old_mode.id`

**Additional context**
The bug isn't very crucial, as each model is stored in a vector, and the `position` of the model is its `id` and the position is mostly used rather than the stored `id` in the model.

med-ayssar:
Fixes #2378


med-ayssar:
> @med-ayssar Thanks for the quick fix. There also seems to be a problem with the model name, as the following example shows (with current master, not your branch):
> 
> ```
> SLI ] /iaf_psc_alpha /foo CopyModel
> SLI ] /i /iaf_psc_alpha Create def
> SLI ] /f /foo Create def
> SLI ] i GetStatus 0 get [[/model /model_id]] get ==
> [/iaf_psc_alpha 17]
> SLI ] f GetStatus 0 get [[/model /model_id]] get ==
> [/iaf_psc_alpha 17]
> SLI ] /iaf_psc_alpha GetDefaults [[/model /model_id]] get ==
> [/iaf_psc_alpha 17]
> SLI ] /foo GetDefaults [[/model /model_id]] get ==
> [/foo 17]
> ```
> 
> When getting the status information for neuron `f`, it not only shows the `model_id` of the original model, but also the original name. Interestingly, the new name (but the original id) is shown when looking at the model defaults. Could you check this out?
> 
This current branch already fixes both issues. Interestingly, we get either the original model name and its `id` or `Unknownmodel` and `-1` as `id`. The latter happens only when installing external models (i.e. generated model from nestml). The differene lies in the `CopyConstructor`. The built-in models call the CopyConstructor of the `ArchivingNode` whereas the nestml models call `Constructor` of the `ArchivingNode`. That is the reason why we see (`iaf_psc_alpha` and `17`) in the instance of the `foo` model.
> And have you checked if synapse models are also affected?
> 
Coping Synapses take another path and to my understanding it is working pretty fine as expected.
 
> Please also create a regression test for this issue as part of the PR. The PR should check model name and ID and to be on the safe side also check that copying from a copied model works.
> 
Yeah, that would be the next task to include in this PR.
> Finally, the PR contains a number of unrelated changes to CMake files. Can you modify the PR to keep those out? They should go into a separate PR. The changes in models and pynest examples which fix only formatting are ok.

I merged my last PR #2194 to avoid doing the same work fixing the `static_checks` errors in order to reach the build phase.

heplesser:
> This current branch already fixes both issues. Interestingly, we get either the original model name and its `id` or `Unknownmodel` and `-1` as `id`. The latter happens only when installing external models (i.e. generated model from nestml). The differene lies in the `CopyConstructor`. The built-in models call the CopyConstructor of the `ArchivingNode` whereas the nestml models call `Constructor` of the `ArchivingNode`. That is the reason why we see (`iaf_psc_alpha` and `17`) in the instance of the `foo` model.

Excellent.

> Coping Synapses take another path and to my understanding it is working pretty fine as expected.

Could you double check (and add it to a test?)

> > Please also create a regression test for this issue as part of the PR. The PR should check model name and ID and to be on the safe side also check that copying from a copied model works.
> 
> Yeah, that would be the next task to include in this PR.

Looking forward to it!

> > Finally, the PR contains a number of unrelated changes to CMake files. Can you modify the PR to keep those out? They should go into a separate PR. The changes in models and pynest examples which fix only formatting are ok.
> 
> I merged my last PR #2194 to avoid doing the same work fixing the `static_checks` errors in order to reach the build phase.

Keeping the fixes for the formatting problems in is fine, but the changes to the CMake files should not "sneak" in through a back door. So either #2194 should be merged first or you should remove the CMake changes here. 

Generally, I would suggest not to merge one PR-branch into another. Rather remind reviewers ... Otherwise, tracing the history becomes difficult and in the most extreme case the code review process could be sidestepped.

jougs:
@heplesser:
- I'm also not expecting any problems for synapses, as they don't have IDs in the sense nodes have them.
- #2194 has been merged, so the dependency you worried about is gone.

@med-ayssar: can you please create the unit test @heplesser requested above? Given that this is real problem for users, I'd like to get this merged asap.

jougs:
@med-ayssar: can you please also merge master, so the latest changes are reflected in the diff? Thanks!

terhorstd:
@med-ayssar how is the status on this?

heplesser:
@med-ayssar Any updates?

jougs:
@med-ayssar: I've created https://github.com/med-ayssar/nest-simulator/pull/1 with some changes that I'd like to see in this one. Can you please have a look? Thanks!

jougs (approved):
This looks good to me now. Many thanks for the nice work and for accepting my changes.

heplesser (changes requested):
Thanks, looks ready now!

heplesser (approved):
Sorry, clicked the wrong radio button!

jougs:
Yay! :smile:

PR #2380, opened 2022-04-25
Remove remaining uses of `float` datatype from C++ code.

heplesser:
This PR removes the last remaining uses of the `float` data type from NEST C++ code. Some uses may have crept in as a side-effect of too much Python programming ;).


hakonsbm (approved):
Looks good to me!

gtrensch (approved):
Looks good to me!

PR #2394, opened 2022-05-11
Add functions yielding source neurons given target neurons

ackurth:
In my work with spatial networks,I find it oftentimes useful to have an easy interface for accessing, plotting etc. the source neurons of a given target neuron.
Currently, while there are function to get information of target neurons given source neurons ```nest.GetTargetNodes```, ```nest.GetTargetPositions``` as well as a plotting routine  ```nest.PlotTargets```, functionality regarding the other way (i.e. about source neurons given a target neuron) is lacking.
This PR addresses this and introduces ```nest.GetSourceNodes```, ```nest.GetSourcePositions```, ```nest.PlotTargets``` for the NEST Python API.
Or is there a reason I don't see yet for not having these functions as well?


pynest/nest/lib/hl_api_spatial.py
```diff
@@ -728,6 +731,81 @@ def GetTargetNodes(sources, tgt_layer, syn_model=None):
     return tuple(src_tgt_map[snode_id] for snode_id in sources.tolist())
 
 
+def GetSourceNodes(src_layer, targets, syn_model=None):
+    """
+    Obtain sources of `targets` in given `source` population.
```
hakonsbm, line 736: Should it be `src_layer` here?
```suggestion
    Obtain sources of `targets` in given `src_layer` population.
```

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -728,6 +731,81 @@ def GetTargetNodes(sources, tgt_layer, syn_model=None):
...
+    -------
+    tuple of NodeCollection:
+        Tuple of `NodeCollections` of source neurons fulfilling the given criteria, one `NodeCollection` per
+        target node ID in `target`.
+
+    See also
+    --------
+    GetSourcePositions: Obtain positions of sources in a given source layer connected to given target.
+    GetConnections: Return connection identifiers between
+        sources and targets
```
hakonsbm, line 761: Should "Return connection identifiers between" and "sources and targets" be on the same line here?

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -810,6 +888,87 @@ def GetTargetPositions(sources, tgt_layer, syn_model=None):
...
+        ::
+
+            import nest
+
+            # create a spatial population
+            s_nodes = nest.Create('iaf_psc_alpha', positions=nest.spatial.grid(shape=[11, 11], extent=[11., 11.]))
+
+            # connectivity specifications with a mask
+            conndict = {'rule': 'pairwise_bernoulli', 'p': 1.,
+                        'mask': {'rectangular': {'lower_left' : [-2.0, -1.0],
```
hakonsbm, line 935:
```suggestion
                        'mask': {'rectangular': {'lower_left': [-2.0, -1.0],
```

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -1129,6 +1289,150 @@ def PlotTargets(src_nrn, tgt_layer, syn_type=None, fig=None,
...
+            import nest
+            import matplotlib.pyplot as plt
+
+            # create a spatial population
+            s_nodes = nest.Create('iaf_psc_alpha', positions=nest.spatial.grid(shape=[11, 11], extent=[11., 11.]))
+
+            # connectivity specifications with a mask
+            conndict = {'rule': 'pairwise_bernoulli', 'p': 1.,
+                        'use_on_source': True,
+                        'mask': {'rectangular': {'lower_left' : [-2.0, -1.0],
```
hakonsbm, line 1355:
```suggestion
                        'mask': {'rectangular': {'lower_left': [-2.0, -1.0],
```

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -1129,6 +1289,150 @@ def PlotTargets(src_nrn, tgt_layer, syn_type=None, fig=None,
...
+            nest.PlotSources(s_nodes, s_nodes[4])
+            plt.show()
+    """
+
+    # import pyplot here and not at toplevel to avoid preventing users
+    # from changing matplotlib backend after importing nest
+    import matplotlib.pyplot as plt
+
+    if not HAVE_MPL:
+        raise ImportError("Matplotlib could not be imported")
```
hakonsbm, line 1372: Should this check be before we import pyplot? And do we even need it? It's just a slightly nicer error message than the standard one.

testsuite/pytests/test_spatial/test_basics.py
```diff
@@ -345,6 +345,54 @@ def test_GetTargetNodes(self):
...
+        self.assertTrue(all([len(g) == 4 for g in s]))
+
+        s = nest.GetSourceNodes(layer, layer, syn_model='static_synapse')
+        self.assertEqual(len(s), len(layer))
+        self.assertTrue(all([len(g) == 0 for g in s]))  # no static syns
+
+        s = nest.GetSourceNodes(layer, layer, syn_model='stdp_synapse')
+        self.assertEqual(len(s), len(layer))
+        self.assertTrue(
+            all([len(g) == 4 for g in s]))  # 2x2 mask  -> four sources
```
hakonsbm, line 382: No need to break the line here.
```suggestion
        self.assertTrue(all([len(g) == 4 for g in s]))  # 2x2 mask  -> four sources
```

testsuite/pytests/test_spatial/test_basics.py
```diff
@@ -421,6 +469,82 @@ def test_GetTargetPositions(self):
...
+        cdict = {'rule': 'pairwise_bernoulli',
+                 'p': 1.}
+        sdict = {'synapse_model': 'stdp_synapse'}
+
+        # For co-dependent properties, we use `set()` instead of kernel attributes
+        nest.set(use_compressed_spikes=False, sort_connections_by_source=False)
+
+        layer = nest.Create('iaf_psc_alpha',
+                            positions=nest.spatial.grid(shape=[1, 1],
+                                                        extent=(1., 1.),
```
hakonsbm, line 485: Why use a tuple for the extent here?

testsuite/pytests/test_spatial/test_basics.py
```diff
@@ -345,6 +345,54 @@ def test_GetTargetNodes(self):
         self.assertEqual(len(t), 1)
         self.assertEqual(t[0], nest.NodeCollection([1, 3, 7, 9]))
 
+    def test_GetSourceNodes(self):
+        """Interface check for finding source."""
```
hakonsbm, line 349:
```suggestion
        """Interface check for finding source"""
```

hakonsbm (changes requested):
Thanks for these useful functions. I have just a few minor suggestions.

terhorstd:
This seems not critical and might need a bit more time for a reaction, since @ackurth is busy at the moment.
All in good time.

doc/htmldoc/networks/spatially_structured_networks.rst
```diff
@@ -1311,9 +1311,15 @@ The following table presents some query functions provided by NEST.
 | ``nest.GetTargetNodes()``       | Obtain targets of sources in a                      |
 |                                 | given target layer.                                 |
 +---------------------------------+-----------------------------------------------------+
+| ``nest.GetSourceNodes()``       | Obtain sources of targets in a                      |
```
heplesser, line 1314: Here an in corresponding locations below, I suggest to order alphabetically, i.e., GetSourceNodes before GetTargetNodes

doc/htmldoc/networks/spatially_structured_networks.rst
```diff
@@ -1353,6 +1359,9 @@ NEST provides three functions to visualize networks:
 | :py:func:`.PlotTargets`              | Plot all targets of a node in a given    |
 |                                      | NodeCollection.                          |
 +--------------------------------------+------------------------------------------+
+| :py:func:`.PlotSources`              | Plot all source of a node in a given    |
```
heplesser, line 1362:
```suggestion
| :py:func:`.PlotSources`              | Plot all sources of a node in a given    |
```

pynest/examples/spatial/nodes_source_target.py
```diff
@@ -0,0 +1,72 @@
...
+plt.title('Target neurons of center neuron')
+plt.show()
+
+# Plot source neurons of center neuron
+fig = nest.PlotLayer(s_nodes, nodesize=80, nodecolor='coral')
+nest.PlotSources(s_nodes, center_neuron, fig=fig)
+plt.title('Source neurons of center neuron')
+plt.show()
+
+print('Global id of target neurons of center neuron')
```
heplesser, line 62: We've dropped the "global id" terminology a while ago and now use "node id". But that gets a bit redundant in the text, so I'd suggest just
```suggestion
print('Target neurons of center neuron')
```

pynest/examples/spatial/nodes_source_target.py
```diff
@@ -0,0 +1,72 @@
...
+print(nest.GetTargetNodes(center_neuron, s_nodes))
+
+print('Global id of source neurons of center neuron')
```
heplesser, line 65:
```suggestion
print('Source neurons of center neuron')
```

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -728,6 +731,81 @@ def GetTargetNodes(sources, tgt_layer, syn_model=None):
...
+    returned, otherwise only targets of specific type.
+
+    Parameters
+    ----------
+    src_layer : NodeCollection
+        NodeCollection with node IDs of `src_layer`
+    targets : NodeCollection
+        NodeCollection with node IDs of `targets`
+    syn_model : [None | str], optional, default: None
+        Return only target positions for a given synapse model.
```
heplesser, line 749:
```suggestion
        Return only source positions for a given synapse model.
```

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -728,6 +731,81 @@ def GetTargetNodes(sources, tgt_layer, syn_model=None):
...
+
+    conns = GetConnections(src_layer, targets, synapse_model=syn_model)
+
+    # Re-organize conns into one list per target, containing only source node IDs.
+    tgt_src_map = dict((tnode_id, []) for tnode_id in targets.tolist())
+    for src, tgt in zip(conns.sources(), conns.targets()):
+        tgt_src_map[tgt].append(src)
+
+    for tgt in tgt_src_map.keys():
+        tgt_src_map[tgt] = NodeCollection(list(np.unique(tgt_src_map[tgt])))
```
heplesser, line 803: How about inserting the sources into sets instead of lists, since you then get the uniqueness automatically?

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -810,6 +888,87 @@ def GetTargetPositions(sources, tgt_layer, syn_model=None):
     return [src_tgt_pos_map[snode_id] for snode_id in sources.tolist()]
 
 
+def GetSourcePositions(src_layer, targets, syn_model=None):
+    """
+    Obtain positions of sources to a given `NodeCollection` of `targets`.
+
+    For each neuron in `targets`, this function finds all source elements
+    in `src_layer`. If `syn_model` is not given (default), all targets are
+    returned, otherwise only source of specific type.
```
heplesser, line 897:
```suggestion
    returned, otherwise only sources connected via the given synapse model.
```
Please check corresponding methods and make the corresponding change.

pynest/nest/lib/hl_api_spatial.py
```diff
@@ -810,6 +888,87 @@ def GetTargetPositions(sources, tgt_layer, syn_model=None):
...
+                                                 'upper_right': [2.0, 1.0]}}}
+
+            # connect population s_nodes with itself according to the given
+            # specifications
+            nest.Connect(s_nodes, s_nodes, conndict)
+
+            # get the positions of the targets of a source neuron
+            nest.GetSourcePositions(s_nodes, s_nodes[5])
+    """
+    if not isinstance(targets, NodeCollection):
```
heplesser, line 945: Empty line above

heplesser (approved):
Thanks for this PR. It looks largely fine, but I have added some suggestions.

pynest/nest/lib/hl_api_spatial.py
ackurth: Indeed.

pynest/nest/lib/hl_api_spatial.py
ackurth: They should, thanks. Corrected in commit 7617304

pynest/nest/lib/hl_api_spatial.py
ackurth: I think we do not necessarily need it, for the exact reason you mentioned.
I do it here since the other plotting functions do it exactly this way as well.
I can remove it, I then would for consistency need to remove all it's appearances in the script.
So maybe it should be another PR? What do you think?

testsuite/pytests/test_spatial/test_basics.py
ackurth: No deep reason, just consistency with the other tests.

doc/htmldoc/networks/spatially_structured_networks.rst
ackurth: Done in 599eafd.

pynest/examples/spatial/nodes_source_target.py
ackurth: Thanks. I like the suggestion.

pynest/nest/lib/hl_api_spatial.py
ackurth: Done in c034c34

pynest/nest/lib/hl_api_spatial.py
ackurth: In principal I agree, this is a nice idea and the resulting code would look more pythonic.
Hence I tried it, but it does not work.
The point where it seems to break is at `sli_push`, this cannot work with Python sets and throws
`PyNESTError: unknown Python type: <class 'set'>`.

I thus think that this suggestion goes beyond the scope of this PR.
What do you think?

ackurth:
Thanks for the very thorough review and the nice comments.
I hope I addressed all the issues you had.

pynest/nest/lib/hl_api_spatial.py
heplesser: Then just leave it as it is, I hadn't thought about that complication.

heplesser (approved):

pynest/nest/lib/hl_api_spatial.py
hakonsbm: As it stands now, if the user doesn't have Matplotlib, an error will be raised at line 1369, making the check of `HAVE_MPL` useless. While I think the added function should be consistent with the other functions, not adding useless code is more important. Therefore I suggest to remove the check here. Fixing the other functions is not within the scope of this PR, so I will create an issue for that.

hakonsbm (changes requested):
Thanks! See [my reply](https://github.com/nest/nest-simulator/pull/2394#discussion_r957018626) of my previous review.

pynest/nest/lib/hl_api_spatial.py
ackurth: I agree, done in 58b4b98.

ackurth:
For some reason, after bringing the branch up to date with master, the static code check does not pass anymore.
If I understand the output correctly, the errors come from files I did not even tough?
I am confused.

heplesser:
> For some reason, after bringing the branch up to date with master, the static code check does not pass anymore.
> If I understand the output correctly, the errors come from files I did not even tough?
> I am confused.

@ackurth, welcome to the mysteries of your formatting checker ... The test of unchanged files may be induced by the merge. The only fix is to correct the formatting ;).

hakonsbm (approved):

pynest/nest/lib/hl_api_spatial.py
hakonsbm: @heplesser Since you have approved, can this conversation be resolved?

PR #2396, opened 2022-05-16
Use MAKE variable to pass correct info to child.

### LINKED ISSUE #2395: Can't use `-j` on `install-nodoc` target
**Describe the bug**
Using `-j8` on `make install-nodoc` results in this error:

```
make[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```

and results in a terribly slow build, slower than `make install -j8`

Helveg:
Spawning a make subprocess should be done through use of the `$(MAKE)` variable so that the child process receives all the information it needs. closes #2395. I didn't check the whole list of `make` targets, but more could be affected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -362,7 +362,7 @@ install( FILES LICENSE README.md
     )
 
 add_custom_target( install-nodoc
-  COMMAND make NEST_INSTALL_NODOC=true install
+  COMMAND $(MAKE) NEST_INSTALL_NODOC=true install
 )
 
 nest_print_config_summary()


Helveg:
Point of discussion: This syntax doesn't seem to sit well with `ninja`, which complains:

```
ninja: error: build.ninja:126: bad $-escape (literal $ must be written as $$)
```

I have tried any combination of unqouted, quoted, bracket quoted, single $, double $, \$, and VERBATIM you could think of, but can't find any syntax that both `cmake` and `ninja` will run.

terhorstd:
@jougs and I had a long discussion some time ago, that this should probably not be a cmake target at all. The concept of "nodoc" should be a flag that can be set. This would reduce the complexity of the issue a lot. We wrote this down in an issue, but never got around to working on this, also I seem to be unable to find the issue that I have in mind :-/ … does anyone reading this remember the discussion? @gtrensch maybe?

gtrensch:
We discussion this in issue #1905.

terhorstd (approved):
Given the time passing on #1905 and the negligible amount of change this PR does, I would say this can go in and we tackle #1905 separately. This one doesn't solve the problem, but it does no harm and makes #2395 work for the time being.

Helveg:
Well it breaks `make install-nodoc` for `ninja` users 😟

gtrensch (approved):
Approving.

Issue #2381, opened 2022-04-25
Prevent NEST from running under mpirun-control if not built with MPI


heplesser:
If NEST is built without MPI support, invoking NEST with `mpirun -np N` will create N independent NEST instances. This is a behaviour the user most likely does not desire. When writing to files, there is considerable risk that these instances will write to the same output files simultaneously, creating a mess at best. 

Therefore, it would be good if an MPI-free NEST could report an error if it is run from `mpirun` or similar.

Unfortunately, [this discussion on StackOverflow](https://stackoverflow.com/questions/12678337/how-can-my-program-detect-whether-it-was-launch-via-mpirun) suggest this might not be solvable in general.

I still suggest we explore and close this issue if we conclude that there is no possibility to implement this feature.

med-ayssar:
Won't be feasible to check the number of processes (`MPI_num`) and have a flag `HAS_MPI` that either set to true or false during the compilation.

During execution and on kernel startup, we can check if `MPI_num > 1 and not HAS_MPI => exit with error`, else continue execution, or is it a too naive solution?

PR #2400, opened 2022-05-19
Mark MPI as required if explicitly requested during CMake configuration

hakonsbm:
During CMake configuration, if MPI is explicitly requested (with `-Dwith-mpi=ON`), MPI is set as `REQUIRED` which makes CMake error if MPI is not found. Previously CMake would continue, and NEST would be installed without MPI support (this may be loosely related to #2381).

Furthermore, if `with-mpi` is passed a path, it is not used to find MPI as indicated by the documentation. There are no good ways to select an MPI installation based on the path alone, and this is a corner-case anyway, so the related text in the documentation is instead removed.


gtrensch (approved):
Thank you for this contribution! :+1:

jougs (approved):
Thanks for addressing this. I was bitten by this more than once myself.

PR #2401, opened 2022-05-20
Add documentation regarding HPC systems

jessica-mitchell:
This PR adds documentation regarding HPC including
* a general overview of hardware and software components
* an example Slurm script
* explanation of MPI and threading
* reference to beNNch for benchmarking


terhorstd:
@jessica-mitchell, could you comment on the status of this one? The mentioned manuscripts have meanwhile been published
* Kurth AC, Senk J, Terhorst D, Finnerty J and Diesmann M (2022)[ Sub-realtime simulation of a neuronal network of natural density.](https://doi.org/10.1088/2634-4386/ac55fc) Neuromorphic Computing and Engineering(2):021001 (article)

and maybe the workflow for optimizations on HPC would also be interesting to mention:
* Albers J, Pronold J, Kurth AC, Vennemo SB, Mood KH, Patronis A, Terhorst D, Jordan J, Kunkel S, Tetzlaff T, Diesmann M and Senk J (2022)[ A Modular Workflow for Performance Benchmarking of Neuronal Network Simulations.](https://doi.org/10.3389/fninf.2022.837549) Frontiers in Neuroinformatics(16):837549 (article)

jessica-mitchell:
This PR is ready for review.

@ackurth @JoseJVS I have a couple of spots with questions as to how to complete the statement, or if the statement is true. Particularly in the slurm_script.rst and mpi_process.rst.

In addition to the parts on opitimizng HPC for NEST there is also a benchmarking part on beNNch that @jasperalbers contributed to. 

See output here
https://nest-test.readthedocs.io/en/add-hpc-docs/hpc/optimizing_nest.html#optimize-performance
https://nest-test.readthedocs.io/en/add-hpc-docs/hpc/benchmarking.html#benchmark

doc/htmldoc/hpc/benchmarking.rst
```diff
@@ -0,0 +1,47 @@
...
+    Error bars indicate variability across three simulation repeats with different random seeds.
+    The top right graph displays the real-time factor defined as wall-clock time normalized by the model time.
+    Built-in timers resolve four different phases of the state propagation: update, collocation, communication, and delivery.
+    Pink error bars show the same variability of state propagation as the left graph.
+    The lower right graph shows the relative contribution of these phases to the state-propagation time.
+
+
+.. seealso:: 
+
+    For further details, see the accompanying `beNNch GitHub Page <https://inm-6.github.io/beNNch>`_.
```
JoseJVS, line 34: We should also point, ideally integrate, Jasper's github page with the guide to setup beNNch (from 0 to plot):
https://inm-6.github.io/beNNch/walk-through.html

doc/htmldoc/hpc/mpi_processes.rst
```diff
@@ -0,0 +1,84 @@
+.. _mpi_process:
+
+MPI process
+===========
+
+
+An MPI process is a set of data and instructions with its own memory that must be explictly shared between processes (also known as CPUs).
```
JoseJVS, line 7: If the intention with this sentence is to just introduce the whole MPI process section I would suggest to be a little bit more clear like:
"An MPI process is one of the instances of a computer program that is being executed by one or many threads and employs a Message Passing Interface (MPI) to communicate its data with the other instances."

Otherwise, if the intention with the sentence was to be more didactic for users with no knowledge of "MPI processes" I would expand a little bit on what is a process and what MPI protocols are and how these come together.

jessica-mitchell:
@ackurth Can you also take a look at the text?

doc/htmldoc/hpc/mpi_processes.rst
```diff
@@ -0,0 +1,79 @@
...
+===========
+
+
+An MPI process is process of a computer program that is being executed by one or many threads and employs a Message Passing Interface (MPI) to communicate its data with the other processes.
+
+We typically use the standard Message Passing Interface (MPI) to instruct how processes work in parallel (See e.g.,
+`OpenMPI <https://www.open-mpi.org/>`_).
+
+The number of MPI processes on each each node is related to the number of :ref:`threads <threads>` you want for each MPI process.
+Multiplied together, the values should equal the total number of cores in a node. (The number of cores varies depending on what system you are using).
```
JoseJVS, line 13: This is not true, one can have as many as MPI processes and threads one wants, the problem would then be whether the node can run everything in parallel or not.
Along with the previous sentence, these statements confuse the usage of MPI processes along with their threads and the distribution of these to the available computing resources.

JoseJVS:
@jessica-mitchell I think we should reconsider the intention with these guides, showing how NEST can be optimized with scaling the number of threads and using multiple MPI processes is a good thing; I think including a guide on thread pinning/process binding is a delicate topic. Although @ackurth explored the performance impact of different thread pinning and process binding affinity schemes, these types of configurations are not simple to explain and are tightly coupled with the types of machines the user has access to. So trying to explain these topics on short guides would be complicated and worst case scenario would be misleading for the reader. I think for this it would be better to point to the sub-real time paper https://doi.org/10.1088/2634-4386/ac55fc  and maybe some additional literature.

jessica-mitchell:
@JoseJVS @ackurth I think I made all the suggested changes, as per our discussion a few weeks ago. Many thanks for your insights :) Here is the output https://nest-test.readthedocs.io/en/add-hpc-docs/hpc/optimizing_nest.html#optimize-performance. @ackurth I believe you still have one thing you want to add to the document?

ackurth (approved):
Very nice documentation.
I think this will prevent quite some struggle in people who want to use NEST on HPC systems.

JoseJVS (approved):
Thanks for all the hard work Jessica, this will definitely be helpful!

PR #2408, opened 2022-06-03
Expose `prepared` flag via kernel status dictionary

jougs:
In a discussion with developers of the NEST-Desktop and Neurorobotics Platform (NRP), we found that both tools require to know if `prepare()` was already called or not. This is internally tracked, but was not available in the kernel status dictionary. This PR just exposes the bool flag to the API layers.


terhorstd (approved):
LGTM

jougs:
Merging with single review, as this is only adding a feature and apparently does not break anything.

PR #2412, opened 2022-06-08
Remove the SLI-based NEST configuration file

jougs:
This removes the SLI-based  `nestrc` file (which mainly defined the `mpirun` command) in favor of a purely self-contained solution that only relies on the SLI library and variables that are determined during the run of CMake.

Please note that this PR comes with a certain loss of flexibility for defining the pager and browser commands used in the help system from within SLI. Given the small user base of the built-in interpreter, this is likely an acceptable downside, given the increased maintainability of the code and the fact that this PR is a prerequisite for continuing the work on `pip` installable NEST.


jougs:
@terhorstd: I switched you as a reviewer against @Helveg due to the connection to the wheel-building process.

Helveg (changes requested):
Less things to deploy on the target machine! 👍

doc/htmldoc/installation/cmake_options.rst
```diff
@@ -159,19 +161,30 @@ neurons, writing to ASCII files might become prohibitively slow due to
...
+needs to know how to execute the launcher of your MPI implementation.
+CMake is usually able to detect the command line for this, but you can
+customize it using the follwing configuration variables (common
+defaults are shown below)::
 
-See the :ref:`parallel_computing` to learn how to execute
-threaded and distributed simulations with NEST.
+    -DMPIEXEC=/usr/bin/mpiexec
+    -DMPIEXEC_NUMPROCS_FLAG=-np
+    -DMPIEXEC_PREFLAGS=
```
heplesser, line 175: Should we mention `--oversubscribe` as a possible flag, maybe the most useful one for developers?

lib/sli/processes.sli
```diff
@@ -1210,13 +1210,9 @@ Synopsis:  (sli-filename) nest_serial -> (string with shell command)
...
 obtain a non-distributed version. The command nest_serial looks up the appropriate 
-way to call nest and returns a suitable command as a string. The algorithm is as
-follows:
-  1. if nest is not compiled with MPI, return "nest"
-  2. if sli command /mpirun is defined, return this with
-     the -np 1 equivalent
-  3. issue and error with the message that nest was compiled with MPI 
-     but no method to call nest has been specified. 
+way to call nest and returns a suitable command as a string. If NEST was not
+compiled with MPI, it just returns "nest", else it returns the "mpirun" command
```
heplesser, line 1214: Doesn't it return the absolute path?

sli/slistartup.cc
```diff
@@ -173,6 +173,11 @@ SLIStartup::SLIStartup( int argc, char** argv )
   , have_libneurosim_name( "have_libneurosim" )
   , have_sionlib_name( "have_sionlib" )
   , ndebug_name( "ndebug" )
+  , mpiexec_name( "mpiexec" )
```
heplesser, line 176: See comment above on `mpiexec` vs `mpirun`

testsuite/do_tests.sh
```diff
@@ -145,8 +145,14 @@ NEST="nest_serial"
 HAVE_MPI="$(sli -c 'statusdict/have_mpi :: =only')"
 
 if test "${HAVE_MPI}" = "true"; then
-  MPI_LAUNCHER="$(sli -c '1 () () mpirun cst 0 get =only')"
-  MPI_LAUNCHER="$(command -v $MPI_LAUNCHER)"
+    MPI_LAUNCHER="$(sli -c 'statusdict/mpiexec :: =only')"
+    MPI_LAUNCHER_VERSION="$($MPI_LAUNCHER --version | head -n1)"
+    # OpenMPI requires --oversubscribe to allow more processes than available cores
+    if [[ "${MPI_LAUNCHER_VERSION}" =~ "(OpenRTE)" ]]; then
```
heplesser, line 151: Another point here on `mpirun` vs `mpiexec`.  When called with `--version`, the former returns `(Open MPI)`, the latter `(OpenRTE)`. At the same time, `man mpiexec` shows a man page stating that the two are equivalent and that uses `mpirun` in all examples. Can we be robust here?

heplesser (commented):
@jougs Thanks, this looks good. My main question is `mpiexec` vs `mpirun`, see some of my comments for details.

doc/htmldoc/installation/cmake_options.rst
```diff
@@ -159,19 +161,30 @@ neurons, writing to ASCII files might become prohibitively slow due to
...
-documentation on the :ref:`config_options`.
+In order to run the distributed tests upon ``make installcheck``, NEST
+needs to know how to execute the launcher of your MPI implementation.
+CMake is usually able to detect the command line for this, but you can
+customize it using the follwing configuration variables (common
+defaults are shown below)::
 
-See the :ref:`parallel_computing` to learn how to execute
-threaded and distributed simulations with NEST.
+    -DMPIEXEC=/usr/bin/mpiexec
```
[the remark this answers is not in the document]
jougs: This default is not anymore set by us, but is now determined by CMake, which seems to prefer `mpiexec` (likely because that works for both OpenMPI and MPICH to my knowledge)

That said, I have changed the documentation in 3ed30fa8f, because we indeed use the word `mpirun` throughout the whole documentation and that is probably more familiar to users.

doc/htmldoc/installation/cmake_options.rst
jougs: I think we shouldn't. `--oversubscribe` is an implementation-detail of OpenMPI I'd rather keep out of NEST's documentation.  The testsuite now adds it automatically when OpenMPI is used and SLI's `mpirun` command is not used anywhere else.

lib/sli/processes.sli
jougs: It does. Fixed in 0e639f221.

testsuite/do_tests.sh
jougs: I did some checks on my machine and it always returned OpenRTE, but think that's all up to your MPI implementation (and which version CMake happens to find on your machine). I have extended the check to include "Open MPI" in 1eb38cf94.

doc/htmldoc/installation/cmake_options.rst
```diff
@@ -3,25 +3,27 @@
 CMake Options for NEST
 ======================
 
-NEST is installed with ``cmake`` (at least v3.12). In the simplest case, the commands::
+Before compiling and installing NEST, the source code  has to be
+configured with ``cmake``. In the simplest case, the commands::
 
     cmake -DCMAKE_INSTALL_PREFIX:PATH=<nest_install_dir> <nest_source_dir>
```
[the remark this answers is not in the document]
jougs: Good point. CMake will anyway abort with an error if the argument is not given and we're not in a Python environment of some sorts.

sli/slistartup.cc
jougs: Plus this is just the key in the status dict under which the executable name is found. I don't think many users will ever look at this.

doc/htmldoc/installation/cmake_options.rst
heplesser: OK

sli/slistartup.cc
heplesser: OK.

Helveg (approved):
Still, LGTM :)

terhorstd:
:rocket: Thanks!

PR #2417, opened 2022-06-21
Add note about PYNEST_QUIET in docstring

clinssen:
The ``PYNEST_QUIET`` flag is currently only mentioned in the NEST2->3 transition guide, and in the API documentation of ``init()`` in ``ll_api.py``.

This PR places the note where it can more easily be found (for instance, after searching for "verbosity").


pynest/nest/lib/hl_api_info.py
```diff
@@ -200,6 +200,11 @@ def set_verbosity(level):
     - M_ERROR=30, display error messages and above
     - M_FATAL=40, display failure messages and above
 
+    .. note::
+
+       If the environment variable PYNEST_QUIET is set, NEST will not print
```
jessica-mitchell, line 205:
```suggestion
       You can also set ``export PYNEST_QUIET = 1`` before importing NEST.  
   ```

File: pynest/nest/lib/hl_api_info.py [hunk 2]
```diff
@@ -200,6 +200,11 @@ def set_verbosity(level):
...
+       welcome text containing the version and other information.
```
jessica-mitchell, line 206:
```suggestion
       This will prevent the welcome text containing the version and other information from printing.
```

jessica-mitchell (changes requested):
@clinssen  thanks for pointing this out. I think the text should be more explicit as to how it's used.

pynest/nest/lib/hl_api_info.py
clinssen: Not technically the same thing; the former is correct.

jessica-mitchell: ok, how about keep your text but also include a link to see :ref:`refguide_2_3` where the command is written

pynest/nest/lib/hl_api_info.py
jougs, line 206: What about this?

```suggestion
       To suppress the usual output when NEST starts up (e.g. the welcome message and
       version information) you can set the environment variable ``PYNEST_QUIET`` by
       running ``export PYNEST_QUIET = 1`` on the command line before executing your
       simulation script. 
```

jougs:
I can't see how the changes here can make a test fail. @clinssen: can you please merge upstream master?

jougs (approved):

pynest/nest/lib/hl_api_info.py
```diff
@@ -200,6 +200,12 @@ def set_verbosity(level):
     - M_ERROR=30, display error messages and above
     - M_FATAL=40, display failure messages and above
 
+    .. note::
+
+       To suppress the usual output when NEST starts up (e.g. the welcome message and
```
jessica-mitchell, line 205:
```suggestion
       To suppress the usual output when NEST starts up (e.g., the welcome message and
```

pynest/nest/lib/hl_api_info.py
```diff
@@ -200,6 +200,12 @@ def set_verbosity(level):
...
+       version information) you can run ``export PYNEST_QUIET=1`` on the command
```
jessica-mitchell, line 206:
```suggestion
       version information), you can run ``export PYNEST_QUIET=1`` on the command
```

jessica-mitchell (approved):
@clinssen I have a couple of comma additions, but I approve thanks!

Issue #2212, opened 2021-11-17
Enable autapses for structural plasticity in NEST 3


sdiazpier:
**Describe the bug**
When migrating to NEST 3, a set of valid parameters have been specified to be provided to SetKernelStatus which does not include all parameters which were supported before by SetStructuralPlasticityStatus in previous versions of NEST 2. This has as a consequence that the connection builder generated with plastic synapses can not include the autapses parameter.

**To Reproduce**
nest.SetKernelStatus({
    'structural_plasticity_synapses': {
        'syn1': {
            'model': 'copied_synapse_model',
            'post_synaptic_element': 'Den_exc',
            'pre_synaptic_element': 'Axon_exc',
        }
    },
     'autapses': False,
})

In NEST 3.1 produces: Error: "`autapses` is not a valid kernel parameter, valid parameters are: ..."

**Expected behavior**
autapses should still be supported for structural plasticity. Probably the connection parameters need to be sent through the structural_plasticity_synapses parameter.

stinebuu:
@sdiazpier `autapses` has been renamed to `allow_autapses` with NEST 3. It is also not possible to set the parameter with the kernel status, it needs to be sent with the connection dictionary:
```
nest.Connect(nrns, nrns, {'rule': 'fixed_indegree', 'indegree': 3, 'allow_autapses': False})
```

neuroady:
@stinebuu with structural plasticity, connections are not explicitly defined via `nest.Connect(nrns, nrns, conn_dict, syn_dict)`. Rather neurons are given synaptic elements which establish connections based on some rule (`nest.EnableSturcturalPlasticity()`). Therefore, which connections are created cannot (and should not) be explicitly defined, rather it's the Structural Plasticity module which creates them. In NESTv3.1 all connections so made have autapses and multapses by default. And there is no way to switch them off. Well, because there is no way to access these parameters when using structural plasticity. Meanwhile, this behavior could be controlled in NESTv2.20.x via `SetKernelStatus` as @sdiazpier has mentioned above.

sdiazpier:
It is possible to enable or disable autapses for the specific connection builder. The solution is to send this information in the structural_plasticity_synapses parameter and enable this at the moment of creation of the builder. I am working on a prototype, should be available soon.

ErbB4:
Hey, I had the same issue with moving from NEST2.x to NEST3.1 concerning structural plasticity. I am using this line to assign synaptic elements to my synapse with structural plasticity:

[...]

PR #2429, opened 2022-07-08
Add flags for autapses and multapses in `structural_plasticity_synapses`

sdiazpier:
This PR addresses Issue #2212 and allows users to specify if connections to be generated between a pair of synaptic elements will allow for autapses and multapses. 

The user can do this by adding the `allow_autapses` and `allow_multapses` parameters to each type of plastic synapse defined in the `structural_plasticity_synapses` parameter in `SetKernelStatus`, e.g.:
```python
nest.SetKernelStatus({
	    'structural_plasticity_synapses': {
		'synapse_ex': {
		    'synapse_model': 'synapse_ex',
		    'post_synaptic_element': 'Den_ex',
		    'pre_synaptic_element': 'Axon_ex',
		    'allow_autapses': False,
		},
	    }
	})
```


hakonsbm (changes requested):

heplesser:
@sdiazpier Could you check the comments by @hakonsbm ?

terhorstd:
Is there any progress on this?

nestkernel/conn_builder.cpp
```diff
@@ -58,9 +58,15 @@ nest::ConnBuilder::ConnBuilder( NodeCollectionPTR sources,
...
-  updateValue< bool >( conn_spec, names::make_symmetric, make_symmetric_ );
+  if (conn_spec->known(names::allow_autapses)){
+    updateValue< bool >( conn_spec, names::allow_autapses, allow_autapses_ );
+  }
+  if (conn_spec->known(names::allow_multapses)){
+    updateValue< bool >( conn_spec, names::allow_multapses, allow_multapses_ );
+  }
+  if (conn_spec->known(names::make_symmetric)){
+    updateValue< bool >( conn_spec, names::make_symmetric, make_symmetric_ );
+  }
```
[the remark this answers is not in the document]
jougs: Fixed in 25f3a29b8

testsuite/pytests/test_sp/test_sp_autapses.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TestStructuralPlasticityAutapses(unittest.TestCase):
+
+    def setUp(self):
+        nest.ResetKernel()
+        nest.set_verbosity('M_INFO')
+
+    def test_autapses(self):
+        nest.CopyModel('static_synapse', 'synapse_ex')
```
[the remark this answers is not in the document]
jougs: fixed in a779816.

testsuite/pytests/test_sp/test_sp_autapses.py
```diff
@@ -0,0 +1,53 @@
...
+        nest.SetKernelStatus({
+            'structural_plasticity_synapses': {
+                'synapse_ex': {
+                    'synapse_model': 'synapse_ex',
+                    'post_synaptic_element': 'Den_ex',
+                    'pre_synaptic_element': 'Axon_ex',
+                    'allow_autapses': False,
+                },
+            }
+        })
```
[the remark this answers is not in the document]
jougs: fixed in a779816.

testsuite/pytests/test_sp/test_sp_autapses.py
```diff
@@ -0,0 +1,53 @@
...
+
+
```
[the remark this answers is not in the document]
jougs: fixed in a779816.

jougs:
@terhorstd: yes, there is :smile_cat:

@heplesser: I have taken over this PR and have addressed all reviewer comments. I have also used the opportunity to clean up things here and there by simplifying some logic and addressing some (undocumented) issues. Given that I'm an author now, someone else should review.

jougs:
@pnbabu, @heplesser, can you please review? Thanks!

nestkernel/conn_builder.cpp
```diff
@@ -511,39 +500,22 @@ nest::ConnBuilder::set_synapse_params( DictionaryDatum syn_defaults, DictionaryD
 void
 nest::ConnBuilder::set_structural_plasticity_parameters( std::vector< DictionaryDatum > syn_specs )
 {
-  // Check if both pre and postsynaptic element are provided. Currently only possible to have
-  // structural plasticity with single element syn_spec.
-  bool have_both_sp_keys = false;
-  bool have_one_sp_key = false;
-  for ( auto syn_params : syn_specs )
+  DictionaryDatum syn_spec = syn_specs[ 0 ];
```
heplesser, line 503: Very nice cleanup here! If I read the code right, then if neither the pre- nor the postsynaptic element are given, the method will just do nothing. Is this intended behavior?

nestkernel/conn_builder.h
```diff
@@ -468,13 +472,25 @@ class SPBuilder : public ConnBuilder
   std::string
```
heplesser, line 472: Should this return a `const &`?

nestkernel/conn_builder.h
```diff
@@ -468,13 +472,25 @@ class SPBuilder : public ConnBuilder
...
   get_pre_synaptic_element_name() const
   {
-    return pre_synaptic_element_name_.toString();
+    return pre_synaptic_element_name_;
   }
 
   std::string
```
heplesser, line 478: Should this return a `const &`?

nestkernel/conn_builder.h
```diff
@@ -198,17 +209,10 @@ class ConnBuilder
   std::vector< std::shared_ptr< WrappedThreadException > > exceptions_raised_;
 
   // Name of the pre synaptic and postsynaptic elements for this connection builder
-  Name pre_synaptic_element_name_;
-  Name post_synaptic_element_name_;
+  std::string pre_synaptic_element_name_;
+  std::string post_synaptic_element_name_;
```
heplesser, line 213: These seem SP-specific. Do we need them in the base class?

nestkernel/conn_builder.h
```diff
@@ -490,6 +506,8 @@ class SPBuilder : public ConnBuilder
   void sp_connect( const std::vector< size_t >& sources, const std::vector< size_t >& targets );
 
 protected:
+  std::string name_;
```
heplesser, line 509: What is this `name_`?

testsuite/pytests/test_sp/test_sp_autapses_multapses.py
```diff
@@ -0,0 +1,72 @@
...
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import nest
+import unittest
```
heplesser, line 23: Could you remove the unittest-stuff and make it a pytest?

testsuite/pytests/test_sp/test_sp_manager.py
```diff
@@ -59,7 +59,8 @@ def test_register_synapses(self):
                 nest.structural_plasticity_synapses = {"syn1": syn_dict}
                 kernel_status = nest.structural_plasticity_synapses
                 assert "syn1" in kernel_status
-                assert kernel_status["syn1"] == extract_dict_a_from_b(kernel_status["syn1"], syn_dict)
+                for kv in extract_dict_a_from_b(kernel_status["syn1"], syn_dict).items():
+                    assert kv in kernel_status["syn1"].items()
```
heplesser, line 63: I don't understand what is going on here.

nestkernel/conn_builder.cpp
jougs: Yes, it is.

Actually, I only inferred the functionality from the old code and was also confused initially. To avoid such confusion in the future, I have further restructured the code to be more like the original and added a docstring in 4ba134e2b7.

nestkernel/conn_builder.h
jougs: Fixed in ddb4f9dc68.

nestkernel/conn_builder.h
jougs: Fixed in ddb4f9dc68.

nestkernel/conn_builder.h
jougs: I initially also thought so and even tried moving them. However, they are used in `ConnBuilder::change_connected_synaptic_elements()`, which gets called by `sp_disconnect()` in each and every `ConnBuilder` child class.

Maybe this can be further disentangled in a future PR? I would regard this as outside the scope of this on. See also #2311 for a similar effort to split SP related code from general NEST base functionality.

nestkernel/conn_builder.h
jougs: This is the name of the `ConnBuilder`. I admit that this is a bit of a hack, but I didn't know where else to store this variable.

This name is needed to be able to return the same names when retrieving the kernel attribute `structural_plasticity_synapses` as used when setting them. Previously, a user set them using a name, e.g.
```
nest.structural_plasticity_synapses = { "synapse_ex": { ...},  "synapse_in": { ...}  }
```
but got back a numbered, e.g.
```
nest.structural_plasticity_synapses = { "syn1": { ...},  "syn2": { ...}  }
```
which was a) confusing and b) error prone due to the likely loss of ordering.

Now, the name is kept in the `ConnBuilder` object, so it can be used for the retrieval of parameters.

testsuite/pytests/test_sp/test_sp_manager.py
jougs: This just checks that all key/value pairs from the outer dict (the one that was set) are contained in the inner dict (the one that was retrieved after setting).  This broke due to the addition of new values that are present in the retrieved dict with their defaults even though they might not have been present in the set dict.

In general: A lot of the testing for the structural plasticity code is rather simple and not using modern techniques and someone should clean this up real good. I just don't think I can do this right now.

heplesser (commented):
@jougs Thanks for the changes! Now I am just missing a docstring for the `name_` member variable.

nestkernel/conn_builder.h
jougs: Added in 3aa649a5a.

jougs:
I've just added the requested docstring and think this good to go now.

testsuite/pytests/test_sp/test_sp_autapses_multapses.py
jougs: I have renamed the file when I added the test for `allow_multapses`. Probably that lead to the confusion.

heplesser (approved):

jougs:
Do we actually *expect* a review from @hakonsbm? If not, I think we could also merge this without waiting any longer. @hakonsbm any myself did a review on the original code and I merely stepped in to address the reviewer comments in @sdiazpier's place, which you now have reviewed.

pnbabu (approved):
Looks good to me!

Issue #2340, opened 2022-03-11
Described and actual behavior of -Dwith-optimize don't match


jarsi:
**Description**
While doing some benchmarks together with @ackurth, we discovered that there is a discrepancy between the description and behavior of the flag `-Dwith-optimize`. According to the description in the file [CMakeLists.txt](https://github.com/nest/nest-simulator/blob/6565ce172c950cabb0dc7ffee216d1566fbf6c8a/CMakeLists.txt#L73) the default is `OFF` yielding `-O2`. Changing it to `ON` should yield `-O3`. Contrary to the description, the default is `ON`:

`set( with-optimize ON CACHE STRING "Enable user defined optimizations [default=OFF (uses '-O2')]. When ON, '-O3' is used. Separate multiple flags by ';'." )`

Looking at [cmake/ProcessOptions.cmake](https://github.com/nest/nest-simulator/blob/6565ce172c950cabb0dc7ffee216d1566fbf6c8a/cmake/ProcessOptions.cmake#L23) shows that `ON` actually leads to `-O2` and `OFF` yields no optimization, which is afaik `-O0`.

```
function( NEST_PROCESS_WITH_OPTIMIZE )
  if ( with-optimize )
    if ( with-optimize STREQUAL "ON" )
      set( with-optimize "-O2" )
    endif ()
    foreach ( flag ${with-optimize} )
      set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${flag}" PARENT_SCOPE )
      set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${flag}" PARENT_SCOPE )
    endforeach ()
  endif ()
endfunction()
```

**To Reproduce**
Running cmake shows that the described behavior is inconsistent with the actual behavior.

**Expected behavior**
Described and actual behavior should match. In my understanding `-Dwith-optimize=ON` should be `-O3` and `-Dwith-optimize=OFF` should be `-O2`.

heplesser:
I think `with-optimize=OFF` should to that, i.e., turn off all optimization, i.e., `-O0`. The question then is if `ON` should yield `-O2` or `-O3`. I am not sure why we did not use `-O3` so far—maybe check compile times and resulting performance. Users can always give their own choice explicitly if they want. Documentation must obviously be updated.

steffengraber:
Solved with #2432.

PR #2432, opened 2022-07-21
Adjust documentation to the actual behaviour of -Dwith-optimize

steffengraber:
Fix issue #2340


steffengraber:
@jarsi maybe you can have a look

jarsi:
Hi @steffengraber, thanks for starting this!

If I understand the cmake line correctly then the default is still ON although it states that it is OFF. Also the default optimization level is still set to O2 (see [here](https://github.com/nest/nest-simulator/blob/dd5b6134259885fc8a7dbbafd53020a3a9d96cef/cmake/ProcessOptions.cmake#L23-L33)).

I think it's a matter of preference whether with-opimize defaults to O2 or O3. We ran some tests a while ago and did not see huge differences between these two.

If I remember correctly, this  has not been finally resolved, maybe @heplesser and @diesmann could comment on this or we discuss this in one of our upcoming meetings.

CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ set( target-bits-split "standard" CACHE STRING "Split of the 64-bit target neuro
 
 # generic build configuration
 option( static-libraries "Build static executable and libraries [default=OFF]" OFF )
-set( with-optimize ON CACHE STRING "Enable user defined optimizations [default=OFF (uses '-O2')]. When ON, '-O3' is used. Separate multiple flags by ';'." )
+set( with-optimize ON CACHE STRING "Enable user defined optimizations [default=OFF (uses '-O0')]. When ON, '-O3' is used. Separate multiple flags by ';'." )
```
heplesser, line 75:
```suggestion
set( with-optimize ON CACHE STRING "Enable user defined optimizations [default=ON (uses '-O2')]. When OFF, no '-O' flag is passed to the compiler. Explicit compiler flags can be given; separate multiple flags by ';'." )
```

The code in `ProcessOptions.cmake` 

https://github.com/nest/nest-simulator/blob/d2500ca3ee16930c0b372694ec90bfe22cb3461f/cmake/ProcessOptions.cmake#L24

indicates that `-Dwith-optimize=ON` will use `-O2`, while `...=OFF` will add no O-option at all. If `-Dwith-optimize` is not given at all, `-O2` is also added. I confirmed these behaviors with a test. What the compiler does when used without any -O flag is up to the compiler and e.g., the Clang man page does not provide information about this.

doc/htmldoc/installation/cmake_options.rst
```diff
@@ -107,7 +107,7 @@ Generic build configuration
 | ``-Dstatic-libraries=[OFF|ON]``                      | Build static executable and libraries [default=OFF].             |
 +------------------------------------------------------+------------------------------------------------------------------+
 | ``-Dwith-optimize=[OFF|ON|<list;of;flags>]``         | Enable user defined optimizations                                |
-|                                                      | [default=OFF (uses '-O2')]. When ON, '-O3' is used. Separate     |
+|                                                      | [default=OFF (uses '-O0')]. When ON, '-O3' is used. Separate     |
```
heplesser, line 110: Adjust as above. I struggle adding that as a proper suggestion here since the text is split over modified and unmodified lines.

heplesser (changes requested):
@steffengraber Thanks for the PR. Please see my suggestions below to bring doc in step with behavior.

heplesser (approved):

hakonsbm (changes requested):
Looks mostly good, but there is a missing square bracket in the rst-file (I hope my suggestion doesn't mess up the formatting).

hakonsbm (approved):
Thanks!

PR #2439, opened 2022-07-27
Only set relevant runpaths in libraries and executables

StefanBruens:
Libraries and executables install into different directories, and thus the relative directories to the library dir differ.

Only set the RUNPATH which actually applies, avoiding extraneous and invalid entries.

### DIFF
--- a/models/CMakeLists.txt
+++ b/models/CMakeLists.txt
@@ -150,6 +150,18 @@ set_target_properties( models
     )
 target_link_libraries( models nestutil sli_lib nestkernel )
 
+if ( NOT APPLE )
+    set_target_properties( models
+        PROPERTIES
+        INSTALL_RPATH "\$ORIGIN"
+        )
+else ()
+    set_target_properties( models
+        PROPERTIES
+        INSTALL_RPATH "@loader_path"
+        )
+endif ()
+
 target_include_directories( models PRIVATE
     ${PROJECT_SOURCE_DIR}/thirdparty
     ${PROJECT_SOURCE_DIR}/libnestutil
--- a/nestkernel/CMakeLists.txt
+++ b/nestkernel/CMakeLists.txt
@@ -161,6 +161,18 @@ install( TARGETS nestkernel
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
     )
 
+if ( NOT APPLE )
+    set_target_properties( nestkernel
+        PROPERTIES
+        INSTALL_RPATH "\$ORIGIN"
+        )
+else ()
+    set_target_properties( nestkernel
+        PROPERTIES
+        INSTALL_RPATH "@loader_path"
+        )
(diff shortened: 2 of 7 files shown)


terhorstd (changes requested):
Thanks @StefanBruens for the help on MacOS! This looks like a very useful distinction even beyond MacOS itself, and I was wondering if it would be possible to separate `$ORIGIN` into two different variables. With that we maybe could eliminate all `if` blocks and remain with a single block only, like
```
if ( APPLE )
   ORIGIN_LIB = @loader_path
   ORIGIN_EXE = @executable_path
else()
   ORIGIN_LIB = $ORIGIN
   ORIGIN_EXE = $ORIGIN
endif()
```

heplesser (changes requested):
This is just a formal request for changes. Please send us a signed contributor agreement (https://nest-simulator.readthedocs.io/en/latest/_downloads/9b65adbdacba6bfed66e68c62af4e308/NEST_Contributor_Agreement.pdf) to info@nest-initiative.org.

models/CMakeLists.txt
```diff
@@ -150,6 +150,18 @@ set_target_properties( models
     )
 target_link_libraries( models nestutil sli_lib nestkernel )
 
+if ( NOT APPLE )
```
gtrensch, line 153: Avoiding the negation here would result in better readability.

nestkernel/CMakeLists.txt
```diff
@@ -161,6 +161,18 @@ install( TARGETS nestkernel
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
     )
 
+if ( NOT APPLE )
```
gtrensch, line 164: Avoiding the negation here would result in better readability.

pynest/CMakeLists.txt
```diff
@@ -58,6 +58,18 @@ if ( HAVE_PYTHON )
       ${SLI_MODULES} ${EXTERNAL_MODULE_LIBRARIES}
       )
 
+  if ( NOT APPLE )
```
gtrensch, line 61: Avoiding the negation here would result in better readability.

testsuite/cpptests/CMakeLists.txt
```diff
@@ -25,6 +25,18 @@ if ( HAVE_BOOST )
 
   target_link_libraries( run_all_cpptests ${BOOST_LIBRARIES} nestkernel models )
 
+  if ( NOT APPLE )
```
gtrensch, line 28: Avoiding the negation here would result in better readability.

gtrensch (changes requested):

jougs:
@StefanBruens: Do you have any intentions on finishing this? Many thanks!

lekshmideepu:
No response from the user. Closing the PR

Issue #2300, opened 2022-02-16
`make install` unconditionally tries to install python dependencies via pip


StefanBruens:
`make install` should install nest, but should not try to install all its python dependencies.

Build hosts in general do not have an internet connection, but are often isolated hosts for security and reproducibility reasons. 

**To Reproduce**
cd build; cmake --prefix=/usr ..
make
make DESTDIR=/tmp/staging install

**Expected behavior**
All files belonging to `nest` should be installed into `/tmp/staging/usr`. No extra files should be installed.

**Desktop/Environment (please complete the following information):**
 - OS: openSUSE Tumbleweed, openSUSE Leap
 - Python-Version: 3.8
 - NEST-Version: 3.2
 - Installation: rpm

terhorstd:
Hi @StefanBruens , my apologies for the delay.
Would the `-Dwith-python=OFF` flag help in your case? Then `make install` will not build the PyNEST interface.

StefanBruens:
Previously I have been informed the Python interface is the only official, stable API of nest, and API compatibility is not a concern for the C++ interface.

> So far, we have never promised a stable C++ API. The closest we have come to an explicit API is the set of functions defined in `nest.h`. We rather focus on stability of the user interface at the Python level.

_Originally posted by @heplesser in https://github.com/nest/nest-simulator/issues/2376#issuecomment-1105825821_

So disabling the Python API would create a package without any stable, external interface. Obviously this is not viable.

StefanBruens:
This issue is still current (3.3).

terhorstd:
Yes, I completely agree. I don't recall why that should be necessary, @Helveg was this related to the wheel?

terhorstd:
Hey, thanks @Helveg for the fast reply! @StefanBruens would you be interested in joining this effort from the MacOS side?

Helveg:
FWIW: "being a proper binary extension to Python" boils down to having NEST's CMake setup successfully executed by `scikit-build`.

StefanBruens:
I haven't touched a Mac for over 10 years ...

terhorstd:
My appologies @StefanBruens, I wrongly implied this from the `if ( APPLE )` changes in PR 2439 and we're notoriously short in developers on mac, so my hopes went up a bit too much…

heplesser:
@Helveg Will this get fixed with the wheel-based installation approach?

Helveg:
[...]

PR #2440, opened 2022-07-27
Respect DESTDIR when installing python egg-info

StefanBruens:
In case a DESTDIR staging directory is set, everything should be installed
below this prefix.

Fixes part of #2300.

### DIFF
--- a/pynest/CMakeLists.txt
+++ b/pynest/CMakeLists.txt
@@ -82,6 +82,7 @@ if ( HAVE_PYTHON )
   install( CODE "
       execute_process(
           COMMAND ${PYTHON} ${CMAKE_CURRENT_BINARY_DIR}/setup.py install_egg_info
+            --root=\${DESTDIR}/
             --install-dir=${CMAKE_INSTALL_PREFIX}/${PYEXECDIR}
           WORKING_DIRECTORY \"${CMAKE_INSTALL_PREFIX}/${PYEXECDIR}\"
       )


gtrensch (approved):
Looks good to me. :+1:

terhorstd:
Doesn't this break conda, venv, and similar environments?

StefanBruens:
> Doesn't this break conda, venv, and similar environments?

No, using `--root` with an empty argument has the same effect as not specifying the parameter.

pynest/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ if ( HAVE_PYTHON )
   install( CODE "
       execute_process(
           COMMAND ${PYTHON} ${CMAKE_CURRENT_BINARY_DIR}/setup.py install_egg_info
+            --root=\${DESTDIR}/
```
terhorstd, line 85: My question is if the "/" is considered an empty argument. When `DESTDIR` is empty it reads `--root=/`
```suggestion
            --root=\"\${DESTDIR}\"
```
(I did some quote-hunting a while ago, and possibly missing quotes around `CMAKE_CURRENT_BINARY_DIR` jump to my eye. Might be necessary if there is a space somewhere in the install path… :grimacing: )

I don't have much time to actually try this now, but @jougs and @Helveg where building the wheel recently and might be able to comment on best practice here.

Helveg: The suggestion looks good. I'd also add quotes around `${CMAKE_CURRENT_BINARY_DIR}/setup.py' like @terhorstd mentions.

StefanBruens: `--root=/` will install to the root directory. When the option is not specified, it installs to the root directory.

terhorstd: yes, but it should install into the venv or conda env or wherever the current python thinks it's packages should go.

pynest/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ if ( HAVE_PYTHON )
   install( CODE "
       execute_process(
           COMMAND ${PYTHON} ${CMAKE_CURRENT_BINARY_DIR}/setup.py install_egg_info
```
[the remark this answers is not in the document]
StefanBruens: That's an unrelated change and should be a separate MR.

pynest/CMakeLists.txt
StefanBruens: `--root` specifies an additional staging prefix, the actual installation path is relative to this prefix.

StefanBruens: > The suggestion looks good. I'd also add quotes around `${CMAKE_CURRENT_BINARY_DIR}/setup.py` like @terhorstd mentions.

Is this just based on a gut feeling, or have you actually verified this is necessary?

execute_process already adds quoting to commands, so this would just add noise. Also, the `${PYTHON}` interpreter path may contain spaces as well.

Helveg: I can confirm that `python my path/package/setup.py install_egg_info` fails, but I haven't tried it in this context:

```
(bsb3py9) robin@LAPTOP-JF6T3PSU:~/git$ python bsb hdf5/setup.py install_egg_info
/home/robin/.pyenv/versions/bsb3py9/bin/python: can't find '__main__' module in '/mnt/c/Users/pwd06/git/bsb'
```

StefanBruens: > execute_process already adds quoting

When **you**  don't use execute_process, **you** have to add the quoting.

Helveg: Do you mean that:

```
execute_process(
          COMMAND ${PYTHON} ${CMAKE_CURRENT_BINARY_DIR}/setup.py install_egg_info
```

with `${CMAKE_CURRENT_BINARY_DIR}` equal to `my path` would be escaped to:

```
python "my path/setup.py" install_egg_info
```

That seems advanced.

terhorstd:
@Helveg, in how far would this be solved by our recent refactoring of the PyNEST install?

StefanBruens:
Prefix and root are two different things (and pip supports both).

In every installed artifact which contains an absolute path, the absolute path starts with the prefix, while the root never appears inside the artifacts.

pynest/CMakeLists.txt
Helveg: So is this resolved? @StefanBruens I still think we need the quotes for when there's whitespaces in the paths

babsey:
Be honest, I am skeptical of this PR, because the issue #2300 indicates that user do not want to install PyNEST dependencies. Correct me if I am wrong.

gtrensch:
Yes indeed. I will take a closer look again.

terhorstd:
This seems to require more in-depth understanding of the mechanism that are actually used. Additionally there is a lot of parallel work on pynest (see pynest-ng project) that will supersede these changes.

Issue #2437, opened 2022-07-26
`get_K_value()` can still return 0 erroneously in edge case


JanVogelsang:
**Describe the bug**
The bug was already described in #1034 and partially fixed in #1137, but still appears for some parameter combinations.  

Continuing on the description of the issue mentioned above:  
Between two communication episodes, lots of post-synaptic spikes might occur, which can lead to the removal of other post-synaptic spikes from the history-queue, which is needed for the synapse to calculate the post-synaptic trace (K-value).  
Whether a spike can be removed from the history or not depends on two conditions defined in 
[*set_spiketime* of the archiving node class](https://github.com/nest/nest-simulator/blob/dd5b6134259885fc8a7dbbafd53020a3a9d96cef/nestkernel/archiving_node.cpp#L198):
1. Its access counter indicates it has been read out by all connected STDP synapses
2. There is another, later spike, that is strictly more than *max_delay_ + eps* away from the new spike

The second condition is needed for situations when two pre-synaptic spikes arrive at the synapse before there is another postsynaptic spike that can be used as reference for the post-synaptic trace. Max-delay is used to make sure that at least one other spike in the post-synaptic neuron's history is relevant for the weight update of the synapse (i.e. had time to travel to the synapse) before removing the spike.  

Spikes are communicated in fixed intervals, and the weights of the STDP synapses are only updated upon spike delivery. This leads to an issue when the history of a post-synaptic neuron gets updated between two communication interval which might not leave the synapse enough time to access the history in time.  
The issue is quite easy to reproduce in a minimal scenario. We connect two arbitrary neurons using a STDP synapse with some delay, e.g. a purely dendritic delay of 1.0. The delay between these neurons is now the min- and max-delay for our simulation with a resolution h defined by _min-delay = k*h, k>1_. Now imagine these spikes:
1. Post 0.2
2. Pre 1.5
4. Post 2.1
5. Pre 3.1  [get_K_value history: 3.4, 2.1]
6. Post 3.4

[...]

Issue #2420, opened 2022-06-23
Numerical instability in aeif_psc_alpha neuron model


zbarni:
**Describe the bug**
While running some experiments together with @TeEsTeBe, we found the implementation of the `aeif_psc_alpha` neuron model appears to be numerically unstable for certain RNG seeds, with two different behaviors as described below: 

**To Reproduce**
1. Consider the following example script
```python
import nest
import numpy as np

nest.ResetKernel()
nest.set_verbosity('M_ALL')
neuron_pars = {
            'E1': {  # IT cell firing profile
                # "model": "aeif_psc_alpha",
                "C_m": 200.,  # capacity of the membrane (pF)
                "g_L": 12.,  # leak conductance (nS)
                "E_L": -70.,  # leak reversal potential (mV) [in other models: resting potential]
                "V_th": -50.,  # spike initiation threshold (mV)
                "Delta_T": 2.,  # slope factor (mV)
                "a": 2.,  # sub-threshold adaptation (ns)
                "tau_w": 300.,  # adaptation time constant (ms) [large]
                "b": 60.,  # spike-triggered adaptaton (pA)
                "V_reset": -58.,  # reset value for Vm after a spike (mV)
                "I_e": 0.,  # constant external input current (pA)
                'tau_syn_ex': 5.,
                'tau_syn_in': 10.,
                "t_ref": 0.1
            },
            'E2': {  # PT cell firing profile
                # "model": "aeif_psc_alpha",
                "C_m": 200.,  # capacity of the membrane (pF)
                "g_L": 10.,  # leak conductance (nS)
                "E_L": -58.,  # leak reversal potential (mV) [in other models: resting potential]
                "V_th": -50.,  # spike initiation threshold (mV)
                "Delta_T": 2.,  # slope factor (mV)
                "a": 2.,  # sub-threshold adaptation (ns)
                "tau_w": 120.,  # adaptation time constant (ms) [large]
                "b": 100.,  # spike-triggered adaptaton (pA)
                "V_reset": -46.,  # reset value for Vm after a spike (mV)
                "I_e": 0.,  # constant external input current (pA)
                'tau_syn_ex': 5.,
                'tau_syn_in': 10.,
                "t_ref": 0.1
            },
            'I1': {
                # 'model': 'iaf_psc_exp',
                'C_m': 250.,
                'E_L': -70.0,
                'I_e': 0.,
                'V_m': -70.0,
                'V_th': -55.0,
                'V_reset': -70.0,
[...]

PR #2443, opened 2022-08-03
Fix STDP k-value error for edge case

JanVogelsang:
This pull requests aims to fix issue #2437 .  
Regression test "issue-2437.py" provides a minimal example of the error in NEST which fails for the current master.  
The issue is fixed by changing the condition to remove spikes from the post-synaptic neuron's spike history in the *set_spiketime* function.  

The problem to be fixed in this PR happens due to communication on a fixed grid (i.e. in fixed intervals) instead of instantly. As STDP weight updates are only performed right upon sending/communicating a spike, the weight update will not use the state of the history of the post-synaptic neuron when the spike occurs, but when the spike is communicated instead. This can lead to scenarios where the history is updated between two communication episodes and spikes that are still relevant for the weight update might get removed. To prevent early removal of spikes, the time window in the condition of the `set_spiketime` function for which spikes are kept in the history is extended by the `min_delay` (time between two communication cycles).


JanVogelsang:
I am currently testing different solutions, as the proposed solution is not yet optimal and can be further optimized.

JanVogelsang:
Apart from the added regression test, there is only a single line changes in this PR, which adds the (global) `min_delay` in the `ArchivingNode::set_spiketime` to make sure spikes are not removed from the history too early.  

Unfortunetaly, introducing this change makes one test fail: [Pipeline step:9:4671](https://github.com/nest/nest-simulator/actions/runs/3185556568/jobs/5195327124#step:9:4671)
Does someone have an idea what causes the "numerical instability"?

clinssen:
Thanks for finding this! I Why is a new test necessary? Could you add the offending pre/post spike sequence from the linked issue to one of the existing STDP synapse tests?

JanVogelsang:
Alright, I merged the existing stdp_synapse test and the regression test I created for the issue to now cover everything in a single test. I also fixed some bugs in the existing test that I stumbled across when writing the regression test, so the test will now be much safer (e.g. when comparing floating point numbers).

JanVogelsang:
Fails due to numerical instability, without any obvious reason (at least not obvious to me).

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -209,42 +224,52 @@ def Kpost_at_time(t, spikes, inclusive=True):
...
-                idx_next_post_spike = np.where((post_spikes_delayed - t) > 0)[0][0]
+            if idx_next_post_spike >= post_spikes.size:
+                t_next_post_spike = -1
+            else:
                 t_next_post_spike = post_spikes_delayed[idx_next_post_spike]
 
-            if idx_next_pre_spike >= 0 and idx_next_post_spike >= 0 and t_next_post_spike < t_next_pre_spike:
-                handle_post_spike = True
+            if t_next_post_spike == -1:
+                a = 1
```
clinssen, line 255: remove

nestkernel/archiving_node.cpp
```diff
@@ -191,12 +191,13 @@ nest::ArchivingNode::set_spiketime( Time const& t_sp, double offset )
...
     //   STDP synapses, and
     // - there is another, later spike, that is strictly more than
-    //   (max_delay_ + eps) away from the new spike (at t_sp_ms)
+    //   (min_global_delay + max_local_delay + eps) away from the new spike (at t_sp_ms)
     while ( history_.size() > 1 )
     {
       const double next_t_sp = history_[ 1 ].t_;
       if ( history_.front().access_counter_ >= n_incoming_
-        and t_sp_ms - next_t_sp > max_delay_ + kernel().connection_manager.get_stdp_eps() )
+        and t_sp_ms - next_t_sp
```
clinssen, line 199: Could you please wrap the lines at 120 characters rather than 80 characters?

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -75,34 +77,32 @@ def init_params(self):
...
+        # contains the weight at pre *and* post times: check that weights are equal for pre spike times
         assert len(weight_by_nest) > 0
-        for idx_pre_spike_nest, t_pre_spike_nest in enumerate(t_weight_by_nest):
-            idx_pre_spike_reproduced_independently = \
-                np.argmin((t_pre_spike_nest - t_weight_reproduced_independently)**2)
-            np.testing.assert_allclose(t_pre_spike_nest,
-                                       t_weight_reproduced_independently[idx_pre_spike_reproduced_independently])
-            np.testing.assert_allclose(weight_by_nest[idx_pre_spike_nest],
-                                       weight_reproduced_independently[idx_pre_spike_reproduced_independently])
+        np.testing.assert_allclose(t_weight_by_nest,
```
clinssen, line 102: I'm probably being a little slow here, but can you explain how we can now test all points in the vector, including the ones at post spike times where NEST has not handled the spike and updated the weight yet?

clinssen:
> > Fails due to numerical instability, without any obvious reason (at least not obvious to me).
> 
> The `aeif` models are notoriously problematic with respect to numerics. You could try to stabilise this by changing in
> 
> https://github.com/nest/nest-simulator/blob/e8995202a97c49c2d4504199691cc0e2b323f277/testsuite/regressiontests/issue-77.sli#L107-L110
> the value for `E_rev` to something lower, e.g., `/E_rev [ -20.0 ]`.
> 
> See also #2420.

It is not clear why this should pop up in this PR, as the aeif models are not touched by it.

@JanVogelsang: could you merge upstream/master to check this not just a fluke?

heplesser:
> > > Fails due to numerical instability, without any obvious reason (at least not obvious to me).
> > 
> > The `aeif` models are notoriously problematic with respect to numerics. You could try to stabilise this by changing in
> 
> It is not clear why this should pop up in this PR, as the aeif models are not touched by it.
> 
> @JanVogelsang: could you merge upstream/master to check this not just a fluke?

`issue-77.sli` tests that all neuron models correctly support STDP by driving them with a strong Poisson input. Presumably, the changes to the STDP code pushed `aeif_cond_beta_multisynapse` across the edge of instability. As a fix, you can also add `/V_peak -30.0` to the `/params` dictionary shown above. You could do the same for the other `aeif` neurons in the test.

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: See line 318:
```python
return pre_spike_times, [w for t, w in w_log.items() if t in pre_spike_times]
```
The reproduced weights are filtered by pre spikes before comparing them to the weights produced by NEST.

JanVogelsang:
Unfortunately, setting V_peak alone does not fix the instability yet. Setting E_rev to -20.0 does fix it though. Should I therefore just set E_rev (and if so, in which models, as only the aeif_cond_alpha_multisynapse seems to fail)?

heplesser:
> Unfortunately, setting V_peak alone does not fix the instability yet. Setting E_rev to -20.0 does fix it though. Should I therefore just set E_rev (and if so, in which models, as only the aeif_cond_alpha_multisynapse seems to fail)?

Yes, then change `E_rev` in that test as well. I would only change it for the model where the test fails.

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -209,42 +224,49 @@ def Kpost_at_time(t, spikes, inclusive=True):
...
-                idx_next_post_spike = np.where((post_spikes_delayed - t) > 0)[0][0]
+            if idx_next_post_spike >= post_spikes.size:
+                t_next_post_spike = -1
+            else:
                 t_next_post_spike = post_spikes_delayed[idx_next_post_spike]
 
-            if idx_next_pre_spike >= 0 and idx_next_post_spike >= 0 and t_next_post_spike < t_next_pre_spike:
-                handle_post_spike = True
+            if t_next_post_spike >= 0 and (
+                    t_next_post_spike + eps < t_next_pre_spike or t_next_pre_spike < 0):
```
clinssen, line 255: Why do we need an epsilon here now whereas we did not before? Did the test case itself hereby fundamentally change?

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -209,42 +224,49 @@ def Kpost_at_time(t, spikes, inclusive=True):
...
-        t_log = []
-        w_log = []
-        Kpre_log = []
-
-        # logging
-        t_log.append(t)
-        w_log.append(weight)
-        Kpre_log.append(Kpre)
+        t_log = list()
+        w_log = dict()
```
clinssen, line 237: This can just stay a list, right? The timepoint corresponding to ``w[i]`` is ``t[i]`` for all *i*.

clinssen:
This is just a small request: could you undo the changes that reformat to an 80 character line length? Lines of code in NEST Simulator are now wrapped at 120 characters.

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -75,34 +77,32 @@ def init_params(self):
         # append some hardcoded spike sequences:
         # pre: 1       5 6 7   9    11 12 13
```
clinssen, line 78:
```suggestion
```

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -75,34 +77,32 @@ def init_params(self):
...
         # post:  2 3 4       8 9 10    12
```
clinssen, line 79:
```suggestion
```

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: I am not entirely sure anymore if that missing epsilon actually caused any issues, but it is definitely much cleaner, as comparing floats could potentially go wrong (something like 3<3.001 -> False). I know that there was a bug in the test where a missing epsilon caused a problem with some other spike times, not sure if it was this line, though.

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: Yup, you are completely right!

JanVogelsang:
@clinssen Could you check again if this is ready to be merged now?

PR #2444, opened 2022-08-04
Add `+=` operator for compartment model

### LINKED ISSUE #2309: `+=` operator for parameter collections
Collections of node parameters currently have to be set with a single command. We would like the option of appending further node parameters to the original parameter collection, as discussed [here](https://github.com/nest/nest-simulator/pull/1895#discussion_r805829211)

Implementation of the `+=` operator to append to parameter collections.

N/A

**Additional context**
E.g. for compartmental models, we can currently only do this:
```
cm = nest.Create('cm_default')

cm.compartments = [
    {"parent_idx": -1, "params": soma_params},
    {"parent_idx": 0, "params": dend_params}
]
```
But we would also like to be able to do this:
```
cm.compartments = {"parent_idx": -1, "params": soma_params}
cm.compartments += {"parent_idx": 0, "params": dend_params}
```

hakonsbm:
An addition assignment operator is added for compartments, and the `cm_default` model is updated. Fixes #2309.


models/cm_default.cpp
```diff
@@ -112,19 +112,13 @@ nest::cm_default::set_status( const DictionaryDatum& statusdict )
...
-    {
-      throw BadProperty( "\'compartments\' is already defined for this model" );
-    }
-
     Datum* dat = ( *statusdict )[ names::compartments ].datum();
     ArrayDatum* ad = dynamic_cast< ArrayDatum* >( dat );
     DictionaryDatum* dd = dynamic_cast< DictionaryDatum* >( dat );
 
+    // Need to rebuild the compartment tree with new compartments.
+    c_tree_ = {};
```
WillemWybo, line 120: The original behaviour of the set_status functions was to add new compartments to the existing tree (i.e. before this was blocked to only allow setting the compartments a single time). I guess you reset the tree because adding compartments to an existing compartments list is now taken care of entirely by the python API? 

I'm not familiar enough with the API code to understand when the `set_status` function is called exactly. I presume it is called after every `+=` operation? In that case, we do incur a (small) extra initialisation cost since we remove and reset the tree every time, which may grow to be significant if the user adds a compartments one at a time in a loop. It may hence we worth considering whether this resetting of the entire tree could not be left out.

WillemWybo:
Thank you very much for this. I only have one comment pertaining to the `set_status` call. The rest looks good to me.

models/cm_default.cpp
hakonsbm: @WillemWybo That's true, there is an overhead here. I've implemented the add assign operator in a different way now to avoid this.

WillemWybo: Everything appears to work for me. Just added a small extra test (PR to your branch)

A slight caveat with my review is that I don't understand what this line does:
```
const auto add_compartments_list_or_dict = [ this, statusdict ]( const Name name )
{
...
}
```
Would it be possible to clarify?

Fyi, it would be great to implement a `+=` operator for receptors in the same way. As far a `cm_default.cpp` is concerned, the code should be pretty much the same.

hakonsbm: It's a [lambda expression](https://en.cppreference.com/w/cpp/language/lambda), locally defining a function, `add_compartments_list_or_dict()`. Like ordinary functions it takes an argument, `name`, but additional variables are passed within the square brackets. These are captured, meaning that they are outside variables that are accessible within the function.

I can add the equivalent functions for receptors as well.

WillemWybo: Ah ok, I see, now I understand. Very nice!

It would make sense to do the same for the receptors.

hakonsbm:
@WillemWybo I've updated the PR with the equivalent functions for receptors now, please have another look.

WillemWybo (approved):

WillemWybo:
Approved! @heplesser

heplesser:
@pnbabu Ping!

pnbabu (approved):
Looks good to me too!

heplesser:
@hakonsbm I have re-started the FULL_MACOS job, the failure seems spurious, at least nothing in the latest commit that would explain it.

PR #2447, opened 2022-08-15
Fix some dimensions in developer documentation adex neurons

jarsi:
I wanted to try out the adaptive exponential integrate-and-fire neuron models and realized that some of the comments about the dimensions of the variables in the text seem to be incorrect: The slope factor `Delta_T` is mV, not ms, see [the original publication](https://journals.physiology.org/doi/pdf/10.1152/jn.00686.2005?download=true). This PR corrects the dimensions.


heplesser (approved):
Thanks for spotting and fixing this. Fortunately, it has been correct in the user documentation already. Since it is a minimal documentation fix, I merge based on single review.

PR #2451, opened 2022-08-18
Conda packages for Apple's M1 are available now

steffengraber:
osx-arm64 packages are now available for NEST v3.3. The warning in the documentation is obsolete and can be deleted.


heplesser (approved):
Minimal documentation update, thus merging based on single review.

PR #2452, opened 2022-08-19
Add documentation on discretization of uniformly distributed delays

### LINKED ISSUE #833: Uniform distribution of delays isn't uniform
Hi! I just ran into an odd behaviour when trying to create a uniform distribution of delays for my network. I'm new to Nest so I apologise if I'm misinterpreting the results or if this is a known issue.

_Description of problem:_

Uniform distribution of delays is not uniform. Edge values are less probable than middle ones.

_Version-Release or Git commit:_
nest-2.12.0

_How reproducible:_
Always

_Steps to Reproduce:_
This is an example, generating numbers in the **range [1-2]**
```python
import nest
import numpy as np

neurons = nest.Create("iaf_cond_exp", n=100)
nest.Connect(neurons, neurons,
             syn_spec={'delay': {'distribution': 'uniform', 'low': 1., 'high': 2.}})

conns = nest.GetStatus(nest.GetConnections())

delays = [c['delay'] for c in conns]

values, count = np.unique(delays, return_counts=True)
probs = zip(values, count.astype(np.float) / np.sum(count))

for i in [0, 5, -1]:
    print 'Probability of item %.1f: %2f' % probs[i]
```
you can also see the distribution with
```python
from matplotlib import pyplot as plt

_, plots = plt.subplots(nrows=2)
hist, bins = np.histogram(delays, bins=20)
plots[0].bar(bins[:-1], hist, bins[1:] - bins[:-1], align='edge')
plots[1].plot(np.sort(delays), 'o')
plt.show()
```
_Actual results:_
Something like
Probability of item 1.0: 0.05
Probability of item 1.5: 0.01
Probability of item 2.0: 0.05

_Expected results:_
An equal probability for all items
Probability of item 1.0: 0.090909
Probability of item 1.5: 0.090909
Probability of item 2.0: 0.090909

_Additional info:_
This sounds to me like you are generating a floating value for the delays in the whole range [1-2] and afterwards rounding it to one decimal. This is wrong because there're half as many numbers in [1-1.05] (that will get rounded to 1) than there are in [1.05-1.15] (that will get rounded to 1.1). You'd need to either, generate them between [0.95-2.05] or, better, generate integers in the range [10-20] and then divide by 10.

heplesser:
Explain in more detail how delays are discretized when drawn from a uniform distribution.

This fixes #833.

### DIFF
--- a/doc/htmldoc/nest_behavior/random_numbers.rst
+++ b/doc/htmldoc/nest_behavior/random_numbers.rst
@@ -194,7 +194,7 @@ Connection delays in NEST are rounded to the nearest multiple of the simulation 
 even if delays are drawn from a continuous distribution. This will work as expected if the
 distribution has infinite support, for example, the normal distribution. For the uniform distribution,
 though, this rounding will usually lead to lower probabilities for the delays at the edges of
-the distribution. Consider the following case
+the distribution. Consider the following case:
 
 ::
 
--- a/doc/htmldoc/nest_behavior/random_numbers.rst
+++ b/doc/htmldoc/nest_behavior/random_numbers.rst
@@ -200,8 +200,7 @@ the distribution. Consider the following case
 
     nest.resolution = 0.1
     n = nest.Create('iaf_psc_alpha', 100)
-    nest.Connect(n, n, syn_spec={'delay': nest.random.uniform(min=1,
-                                                              max=2)})
+    nest.Connect(n, n, syn_spec={'delay': nest.random.uniform(min=1, max=2)})
     
 This will create 10000 connections in total with delay values 1.0, 1.1, ..., 2.0. But while
 the interior delay values 1.1, ..., 1.9 will occur approximately 1000 times each, the corner
--- a/doc/htmldoc/nest_behavior/random_numbers.rst
+++ b/doc/htmldoc/nest_behavior/random_numbers.rst
@@ -192,7 +192,7 @@ Rounding effects when randomizing delays
 
 Connection delays in NEST are rounded to the nearest multiple of the simulation resolution,
 even if delays are drawn from a continuous distribution. This will work as expected if the
-distribution has infinite support, e.g., the normal distribution. For the uniform distribution,
+distribution has infinite support, for example, the normal distribution. For the uniform distribution,
 though, this rounding will usually lead to lower probabilities for the delays at the edges of
 the distribution. Consider the following case
 
--- a/doc/htmldoc/nest_behavior/random_numbers.rst
+++ b/doc/htmldoc/nest_behavior/random_numbers.rst
@@ -210,7 +210,7 @@ any number from :math:`[1.05, 1.15)` will be rounded to 1.1, but only numbers in
 :math:`[1.0, 1.05)` will be rounded to 1.0.
 
 To achieve equal probabilities of the first and last values, you can either extend the interval
-of the uniform distribution by half the resolution
+of the uniform distribution by half the resolution:
 
 ::
 
--- a/doc/htmldoc/nest_behavior/random_numbers.rst
+++ b/doc/htmldoc/nest_behavior/random_numbers.rst
@@ -204,7 +204,7 @@ the distribution. Consider the following case:
     
 This will create 10000 connections in total with delay values 1.0, 1.1, ..., 2.0. But while
 the interior delay values 1.1, ..., 1.9 will occur approximately 1000 times each, the first and last
-cases 1.0 and 2.0 will occur only approximately 500 times each. This happens because NEST
+cases, 1.0 and 2.0, will occur only approximately 500 times each. This happens because NEST
 first draws the delay uniformly from :math:`[1, 2)` and then rounds to a fixed delay. Thus,
 any number from :math:`[1.05, 1.15)` will be rounded to 1.1, but only numbers in
 :math:`[1.0, 1.05)` will be rounded to 1.0.
--- a/doc/htmldoc/nest_behavior/random_numbers.rst
+++ b/doc/htmldoc/nest_behavior/random_numbers.rst
@@ -136,7 +136,7 @@ The NEST random module
 The ``nest.random`` module provides a range of random distributions that
 can be used to specify parameters for neurons, synapses, and connection
 rules. See :ref:`below for examples <random_examples>` on how to use them in
-practice.
+practice and :ref:`some details on randomizing delays <random_delays>`.
 
 .. automodule:: nest.random.hl_api_random
     :members:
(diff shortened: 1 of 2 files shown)


doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,49 @@ Likewise, synapse parameters can be specified using the random distributions.
...
+
+    nest.Connect(n, n, syn_spec={'delay': nest.random.uniform(min=1-0.5*nest.resolution,
+                                                              max=2+0.5*nest.resolution)})
+
+or use the uniform integer distribution. Since it always draws numbers beginning with zero,
+this is slightly more cumbersome:
+
+::
+
+    nest.Connect(n, n, syn_spec={'delay': 1 + 0.1 * nest.random.uniform_int(11)})
```
hakonsbm, line 226: Generally there should be space around arithmetic operators. When different priorities are used, one may consider adding whitespace around the operators with the lowest priority, but it should be consistent.

hakonsbm (changes requested):
Thanks! I have just a couple minor suggestions.

doc/htmldoc/nest_behavior/random_numbers.rst
heplesser: Looking at PEP8, I think it should be spaces everywhere (except around `**`). I will fix it shortly.

doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,48 @@ Likewise, synapse parameters can be specified using the random distributions.
...
                                 'delay': nest.random.uniform(min=0.5, max=1.5)})
 
+.. _random_delays:
+
+Rounding effects when randomizing delays
+........................................
+
+Connection delays in NEST are rounded to the nearest multiple of the simulation resolution,
+even if delays are drawn from a continuous distribution. This will work as expected if the
+distribution has infinite support, e.g., the normal distribution. For the uniform distribution,
```
jessica-mitchell, line 195:
```suggestion
distribution has infinite support, for example, the normal distribution. For the uniform distribution,
```

doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,48 @@ Likewise, synapse parameters can be specified using the random distributions.
...
+though, this rounding will usually lead to lower probabilities for the delays at the edges of
+the distribution. Consider the following case
```
jessica-mitchell, line 197:
```suggestion
the distribution. Consider the following case:
```

doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,48 @@ Likewise, synapse parameters can be specified using the random distributions.
...
+    
+This will create 10000 connections in total with delay values 1.0, 1.1, ..., 2.0. But while
+the interior delay values 1.1, ..., 1.9 will occur approximately 1000 times each, the corner
+cases 1.0 and 2.0 will occur only approximately 500 times each. This happens because NEST
+first draws the delay uniformly from :math:`[1, 2)` and then rounds to a fixed delay. Thus,
+any number from :math:`[1.05, 1.15)` will be rounded to 1.1, but only numbers in
+:math:`[1.0, 1.05)` will be rounded to 1.0.
+
+To achieve equal probabilities of the corner values, you can either extend the interval
+of the uniform distribution by half the resolution
```
jessica-mitchell, line 213:
```suggestion
of the uniform distribution by half the resolution:
```

doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,48 @@ Likewise, synapse parameters can be specified using the random distributions.
...
+the distribution. Consider the following case
+
+::
+
+    nest.resolution = 0.1
+    n = nest.Create('iaf_psc_alpha', 100)
+    nest.Connect(n, n, syn_spec={'delay': nest.random.uniform(min=1, max=2)})
+    
+This will create 10000 connections in total with delay values 1.0, 1.1, ..., 2.0. But while
+the interior delay values 1.1, ..., 1.9 will occur approximately 1000 times each, the corner
```
jessica-mitchell, line 206:
```suggestion
the interior delay values 1.1, ..., 1.9 will occur approximately 1000 times each, the first and last
```

doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,48 @@ Likewise, synapse parameters can be specified using the random distributions.
...
+cases 1.0 and 2.0 will occur only approximately 500 times each. This happens because NEST
+first draws the delay uniformly from :math:`[1, 2)` and then rounds to a fixed delay. Thus,
+any number from :math:`[1.05, 1.15)` will be rounded to 1.1, but only numbers in
+:math:`[1.0, 1.05)` will be rounded to 1.0.
+
+To achieve equal probabilities of the corner values, you can either extend the interval
```
jessica-mitchell, line 212:
```suggestion
To achieve equal probabilities of the first and last values, you can either extend the interval
```
Using the term first and last as it was used describe these. Although corner sounds ok to me, it might not be good term for non-native English speakers. Unless this is an accepted use of the term. I'm also not sure about using edge, because 'edge case' in English has an idiomatic meaning.

doc/htmldoc/nest_behavior/random_numbers.rst
```diff
@@ -185,6 +185,48 @@ Likewise, synapse parameters can be specified using the random distributions.
...
+
+::
+
+    nest.resolution = 0.1
+    n = nest.Create('iaf_psc_alpha', 100)
+    nest.Connect(n, n, syn_spec={'delay': nest.random.uniform(min=1, max=2)})
+    
+This will create 10000 connections in total with delay values 1.0, 1.1, ..., 2.0. But while
+the interior delay values 1.1, ..., 1.9 will occur approximately 1000 times each, the corner
+cases 1.0 and 2.0 will occur only approximately 500 times each. This happens because NEST
```
jessica-mitchell, line 207:
```suggestion
cases, 1.0 and 2.0, will occur only approximately 500 times each. This happens because NEST
```

jessica-mitchell (changes requested):
Just a few minor suggestions. Thanks @heplesser

heplesser:
@jessica-mitchell @hakonsbm Thank you for your suggestions! All should be fixed now.

jessica-mitchell (approved):
LGTM, thanks

doc/htmldoc/nest_behavior/random_numbers.rst
hakonsbm: @heplesser Should this be committed as well?

heplesser: @hakonsbm Done.

hakonsbm (approved):
Looks good now, thanks!

Issue #2404, opened 2022-05-23
model_details notebooks installed to wrong place or linked incorrectly


heplesser:
The Jupyter notebooks from the `model_details` directory are installed to a different place than they are linked to. The notebooks are installed to 

- https://nest-simulator.readthedocs.io/en/latest/neurons/model_details/aeif_models_implementation.ipynb

while the link from e.g. https://nest-simulator.readthedocs.io/en/latest/models/aeif_psc_alpha.html

goes to

- https://nest-simulator.readthedocs.io/en/latest/model_details/aeif_models_implementation.ipynb

This should be fixed either on the installation location or the link side.

Another suggestion: At present, clicking the link downloads the notebook. Would there be a way to link it such that it is rendered?

heplesser:
@jessica-mitchell Was this fixed by #2421?

jessica-mitchell:
This is fixed with #2453

jessica-mitchell:
Closing, fixed with #2453

PR #2453, opened 2022-08-19
Fix and update model_details notebooks

jessica-mitchell:
This PR moves the model_details notebooks up to top level of hierarchy in htmldoc and fixes issues with rendering of the LaTeX math.

To do this the notebooks were converted to a reStructured text file and output as pngs,
A download link for each notebook was added using a zip file. A direct download of a notebook is not obviously possible with reStructuredText- (if one uses the :download: role with a .ipynb file, the raw notebook code opens).

See for example: https://nest-test.readthedocs.io/en/fix-notebooks-docs/model_details/HillTononiModels.html

*Note:*
The model_details folder is currently not indexed and not discoverable through the table of contents. They are linked in the respective models/ documentation. 

The only notebook not changed is the aeif_models_implementation.ipynb because it does not execute properly. See issue #2454 .


heplesser (commented):
@jessica-mitchell Thanks a lot for this PR! I see three issues with the current state of the PR 

1. The figures render extremely small in both Safari and Firefox on macOS.
2. I would like to keep the notebooks as the primary documents, so that one can easily modify them if needed. RST-files and zip archives should then be automatically generated from the notebooks. Would this be feasible? 
3. Also, adding zip archives and generated figures to Git risks clogging the repo with binary files as these files evolve. Can we avoid it and just "make" those files on our documentation server?

jessica-mitchell:
> @jessica-mitchell Thanks a lot for this PR! I see three issues with the current state of the PR
> 
>     1. The figures render extremely small in both Safari and Firefox on macOS.
Sorry, this is because it is not up-to-date with master!  

>     2. I would like to keep the notebooks as the primary documents, so that one can easily modify them if needed. RST-files and zip archives should then be automatically generated from the notebooks. Would this be feasible?
> 
This is a fair point, but I had trouble getting the HTML rendered notebook to generate proper latex and the restructured text version worked easily. I will take another look to see how I can fix it using the notebook as source.

>     3. Also, adding zip archives and generated figures to Git risks clogging the repo with binary files as these files evolve. Can we avoid it and just "make" those files on our documentation server?
Making these files with sphinxgallery is possible, but it expects a python source. There are other ways I'm sure, which I can look into more.

jessica-mitchell:
@heplesser I looked into this more and discovered that basically the broken equations were missing $$. This seems to resolve the HTML rendering without affecting the output in a notebook running in Jupyter lab. I also updated the mathjax js in the conf, which I dont think did anything but figured we should avoid using outdated versions.

jessica-mitchell:
@heplesser I'm afraid there are some further issues after checking Read the docs rendering which do not seem straightforward

jessica-mitchell:
@heplesser I have now gotten the notebooks to render properly on Read the docs; however local builds continue to not render in some cases.  
For example:
Working on Read the docs:
image

Broken on local build: 
image


I am not sure what needs fixing to get it render in both cases. Something during the build seems to drop `\` (like in `\begin{equation}` ) and ignore the role nbsphinx-math.
HTML Source local build:
```
The repolarizing current is active during, and only during the refractory period: :nbsphinx-math:<a href="[#id1](view-source:file:///home/mitchell/Work/build-repo/gen_build/doc/htmldoc/html/model_details/HillTononiModels.html#id1)"><span class="problematic" id="id2">`</span></a>begin{equation}
g_{text{spike}} = begin{cases}  1  &amp; text{neuron is refractory}\</p>
<blockquote>
<div><p>0 &amp; text{else} end{cases}</p>
</div></blockquote>
<p>end{equation}`</p>
```

But this happens inconsistently and I have yet to pinpoint anything in the source JSON that differs between rendering equations and non-rendering equations. The only thing that worked was adding `$$` but this broke the render on Read the docs.



At the very least the rendering currently works where it needs to for users (both Read the docs and also in Jupyter lab). 
See:
https://nest-test.readthedocs.io/en/fix-notebooks-docs/model_details/HillTononiModels.html
https://nest-test.readthedocs.io/en/fix-notebooks-docs/model_details/noise_generator.html
https://nest-test.readthedocs.io/en/fix-notebooks-docs/model_details/IAF_neurons_singularity.html

heplesser:
This may be due to different versions or configurations of nbconvert. Is there any way to figure out what is installed on the Readthedocs machines, e.g., from logfiles? I just tried after updating my conda packages (including update of nbconvert from 6.5.3 to 7.0.0) and got the same rendering error as you did.

jessica-mitchell:
> This may be due to different versions or configurations of nbconvert. Is there any way to figure out what is installed on the Readthedocs machines, e.g., from logfiles? I just tried after updating my conda packages (including update of nbconvert from 6.5.3 to 7.0.0) and got the same rendering error as you did.

Read the docs uses version 7.0.0 and I have two local conda envs with 6.5.3 and 7.0.0, respectively. Both local envs have the same rendering issue. I also looked at other packages like nbsphinx and its the same across the board.

med-ayssar:
In [The neuron emits a single spike if ](https://github.com/jessica-mitchell/nest-simulator/blame/fix-notebooks-docs/doc/htmldoc/model_details/HillTononiModels.ipynb#L63) we should a line break before the unordered list, otherwise; it won't be rendered as a list.
Same for [Upon spike emission
](https://github.com/jessica-mitchell/nest-simulator/blame/fix-notebooks-docs/doc/htmldoc/model_details/HillTononiModels.ipynb#L67)

doc/htmldoc/model_details/IAF_neurons_singularity.ipynb
```diff
@@ -0,0 +1,606 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Non-singular case ($\\tau_m\\neq \\tau_s$) \n",
+    "The propagator is: "
+   ]
+  },
+  {
+   "cell_type": "raw",
```
med-ayssar, line 49: `cell_type` must be `code` here.

doc/htmldoc/model_details/IAF_neurons_singularity.ipynb
```diff
@@ -0,0 +1,606 @@
...
+    "PA_32_series "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Therefore we have \n",
+    " \n",
+    "$T(PA_{32}(\\tau_s,\\tau_m))=PAs_{32}+PA_{32}^{lin}+O(2)$ where $PA_{32}^{lin}=h^2(-\\tau_m + \\tau_s)*exp(-h/\\tau_m)/(2C\\tau_m^2)$\n",
```
med-ayssar, line 235:
```suggestion
    "$T(PA_{32}(\\tau_s,\\tau_m))=PAs_{32}+PA_{32}^{lin}+O(2)$, where $PA_{32}^{lin}=h^2(-\\tau_m + \\tau_s)*exp(-h/\\tau_m)/(2C\\tau_m^2)$\n",
```

med-ayssar:
Side note, the space and position of the references at the bottom do not seem right.

### Space problem in `aeif_psc_delta_clopath.html`

### Position Problem in `siegert_neuron.html`

Why not simply use the horizontal representation for the references.

jessica-mitchell:
> ### Space problem in `aeif_psc_delta_clopath.html`
Can you be more specific? I see a broken link in the third reference ...
> ### Position Problem in `siegert_neuron.html`
Again, I dont see a problem, can you provide more details or screen shot?
 
> Why not simply use the horizontal representation for the references.

What do you mean by horizontal representation?

med-ayssar:
I do not know if the problem of some settings I have in my local machine, but when building the docs, the references look like that:
Screenshot from 2022-11-15 10-37-30


They are vertical (going from top to bottom), and there is not enough space between columns. I tried building the docs with the current master and also this branch, and I have the same output.

med-ayssar:
I just found another list that is not rendered correctly.

Missing new line before start of list at [Line 250](https://github.com/jessica-mitchell/nest-simulator/blame/fix-notebooks-docs/doc/htmldoc/model_details/HillTononiModels.ipynb#L250)

med-ayssar (commented):
Otherwise, the rest looks good to me.

jessica-mitchell:
> They are vertical (going from top to bottom), and there is not enough space between columns. I tried building the docs with the current master and also this branch, and I have the same output.

Ah this problem, yes this is rendered properly on Read the docs so I didnt notice it, but you're right, in the local build, something strange happens. I have it on my to do list to look into.

jessica-mitchell:
Fixes #2404

doc/htmldoc/model_details/noise_generator.ipynb
```diff
@@ -82,19 +80,19 @@
     "\n",
     "$$\\dot{V} = - \\frac{V}{\\tau} + \\frac{I}{C}$$\n",
     "\n",
-    "where $\\tau$ is the membrane time constant and $C$ the capacitance. We further assume $V(0)=0$ mV. We now focus on the membrane potential at times $t_j=j\\delta$. Let $V_j=V(j\\delta)$ be the membrane potential at time $t_j$. Then, a constant current $I_j$ will be applied to the neuron until $t_{j+1}=t_j+\\delta$, at which time the membrane potential will be\n",
+    "where $\\tau$ is the membrane time constant and $C$ the capacitance. We further assume $V(0)=0$ mV. We now focus on the membrane potential at times $t_j=j\\delta$. Let $V_j=V(j\\delta)$ be the membrane potential at time $t_j$. Then, a constant currant $I_j$ will be applied to the neuron until $t_{j+1}=t_j+\\delta$, at which time the membrane potential will be\n",
```
heplesser, line 83:
```suggestion
    "where $\\tau$ is the membrane time constant and $C$ the capacitance. We further assume $V(0)=0$ mV. We now focus on the membrane potential at times $t_j=j\\delta$. Let $V_j=V(j\\delta)$ be the membrane potential at time $t_j$. Then, a constant current $I_j$ will be applied to the neuron until $t_{j+1}=t_j+\\delta$, at which time the membrane potential will be\n",
```

doc/htmldoc/model_details/noise_generator.ipynb
```diff
@@ -221,19 +219,20 @@
...
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
     "import math\n",
     "import numpy as np\n",
     "import scipy\n",
-    "import matplotlib.pyplot as plt"
+    "import matplotlib.pyplot as plt\n",
+    "%matplotlib inline"
```
heplesser, line 230:
```suggestion
```
The inline magic should not be needed in Jupyter lab.

heplesser (changes requested):
@jessica-mitchell Looks good overall, just a few smaller comments.

heplesser (approved):
Looks good now except for formatting error and merge conflict.

jessica-mitchell:
@med-ayssar Can you do a final review or approve if looks good? Thanks!

med-ayssar (approved):

PR #2455, opened 2022-08-21
Install assimulo from conda-forge

### LINKED ISSUE #2454: aeif_models_implementation fails to execute
**Describe the flaw**

When running this notebook, the following error occurs

error: 'assimulo/thirdparty/hairer/dopri5module.c' missing
additional errors and warnings are created depending on where the notebook is run (tested locally and EBRAINS 22.07 kernel)

**Suggest an improvement**
The dependency Assimulo might need to be upgraded or changed

heplesser:
The `aeif_models_implementation` requires the `assimulo` package to compute reference solutions. The notebook previously suggested installation from PyPi, but assimulo is outdated there (3.0 from 2018) and does not build any longer. A current version is available on conda-forge. This PR therefore adds `assimulo` to `environments.yml` and updates the requirements information in the notebook.

@jougs @terhorstd Since this introduces yet another dependency for a corner case, I'd like your comments.

Fixes #2454.


jessica-mitchell (approved):
The notebook works! Thanks @heplesser

jougs:
Phew... Honestly, I don't like the fact that this adds a dependency on conda to build the full docs. I have now contacted the main developer of Assimulo and asked for an update of the PyPI package and would like to keep this un-merged until I get an answer from him.

heplesser:
@jougs Why do you consider a conda-dependency more problematic than a PyPI dependency? That said, I am wondering if it could make sense to move background documents such as this to a separate repository to avoid too many dependencies in the NEST code base itself.

jougs (approved):
I always feel like conda is the odd one out, because it is not installable/updatable via my system's trusted repository, while `pip` is. I'm usually trying to avoid either if I can, but `assimulo` is also not available via `apt-get`...

Splitting the documentation would not be a good idea in my view, as that has a high potential of things getting outdated and forgotten.

Thinking about it again, I realize that this is only a run-time dependency, and I can probably live well with it being mentioned in the `environment.yaml`, which I'm anyway not using. I'm thus approving.

PR #2459, opened 2022-08-29
Remove padding of extent if not specified when creating free positions

### LINKED ISSUE #2359: Connecting spatial nodes with mask and `edge_wrap=True` yields wrong connections
Creating a population with node positions drawn from a random distribution, without specifying the extent, will calculate the extent and centre based on node-positions.
```python
num_nodes = 1000
r = 0.5
param = nest.random.uniform(-r, r)
free_positions = nest.spatial.free(param, num_dimensions=2, edge_wrap=True)
nodes = nest.Create('iaf_psc_alpha', num_nodes, positions=free_positions)
spatial = nodes.spatial
print(f'centre: {spatial["center"]}')
print(f'extent: {spatial["extent"]}')
```
The above example will print 
```
centre: (2.5454122135615265e-06, 0.00042177573406521063)
extent: (1.199160999965586, 1.198836559276)
```
The extent is not (2r, 2r) = (1.0, 1.0) as one would expect, and the centre is not at (0.0, 0.0).

Furthermore, with edge wrapping turned on, using masks when connecting will not connect to the right nodes.
``` python
mask_radius = r
mask = {'circular': {'radius': mask_radius}}
area = np.pi * mask_radius**2
density = num_nodes / (2 * r)**2
expected_conns_per_node = density * area
expected_total_conns = expected_conns_per_node * num_nodes
print(f'Expecting {expected_total_conns:.0f} connections')
nest.Connect(nodes, nodes, {'rule': 'pairwise_bernoulli', 'p': 1.0, 'mask': mask})
print(f'Num. connections: {nest.GetKernelStatus("num_connections")}')
nest.PlotTargets(nodes[58], nodes, mask=mask)
```
This example will print
```
Expecting 785398 connections
Num. connections: 567914
```
and generate the following plot:
plot

There are fewer connections made than we expect, and from the plot we see that because of the too large extent, there are areas in the mask that do not contain any neurons.

If we instead use `nest.spatial.free(param, extent=[2 * r, 2 * r], edge_wrap=True)`, i.e. specifying the extent, we get (close to) the expected number of connections, and the `PlotTarget()` function generates the following plot:
plot2

The issue comes from how the extent and centre is calculated in free_layer.h. If no extent is specified, the extent is padded. The centre is also calculated from the node positions and cannot be specified.

To fix the issue we may consider always requiring the extent to be specified for free layers, and never padded. Furthermore, the centre should be in the centre of the extent, not calculated from node positions.

hakonsbm:
Fixes #2359 by removing padding of extent if no extent is specified when creating the positions. Extent is instead rounded up and adjusted to a rounded centre. If only a single node is created, extent must be specified.

### DIFF
--- a/nestkernel/free_layer.h
+++ b/nestkernel/free_layer.h
@@ -236,18 +236,14 @@ FreeLayer< D >::set_status( const DictionaryDatum& d )
 
     const auto positional_extent = max_point - this->lower_left_;
     const auto center = ( max_point + this->lower_left_ ) / 2;
-    Position< D > rounded_center;
     for ( int d = 0; d < D; ++d )
     {
-      // Round to the first decimal.
-      rounded_center[ d ] = ( std::round( 10.0 * center[ d ] ) ) / 10;
-
       // Set extent to be extent of the points, rounded up in each dimension.
       this->extent_[ d ] = std::ceil( positional_extent[ d ] );
     }
 
     // Adjust lower_left relative to the rounded center with the rounded up extent.
-    this->lower_left_ = rounded_center - this->extent_ / 2;
+    this->lower_left_ = center - this->extent_ / 2;
   }
 }
 
--- a/nestkernel/free_layer.h
+++ b/nestkernel/free_layer.h
@@ -115,13 +115,11 @@ FreeLayer< D >::set_status( const DictionaryDatum& d )
   Layer< D >::set_status( d );
 
   Position< D > max_point; // for each dimension, the largest value of the positions, aka upper right
-  Position< D > epsilon;   // small value which may be added to layer size ensuring that single-point layers have a size
 
   for ( int d = 0; d < D; ++d )
   {
     this->lower_left_[ d ] = std::numeric_limits< double >::infinity();
     max_point[ d ] = -std::numeric_limits< double >::infinity();
-    epsilon[ d ] = 0.1;
   }
 
   num_local_nodes_ = std::accumulate( this->node_collection_->begin(),
@@ -234,9 +232,25 @@ FreeLayer< D >::set_status( const DictionaryDatum& d )
   }
   else
   {
-    this->extent_ = max_point - this->lower_left_;
-    this->extent_ += epsilon * 2;
-    this->lower_left_ -= epsilon;
+    if ( this->node_collection_->size() <= 1 )
+    {
+      throw KernelException( "If only one node is created, 'extent' must be specified." );
+    }
+
+    const auto positional_extent = max_point - this->lower_left_;
+    const auto center = ( max_point + this->lower_left_ ) / 2;
+    Position< D > rounded_center;
+    for ( int d = 0; d < D; ++d )
+    {
+      // Round to the first decimal.
+      rounded_center[ d ] = ( std::round( 10.0 * center[ d ] ) ) / 10;
+
+      // Set extent to be extent of the points, rounded up in each dimension.
+      this->extent_[ d ] = std::ceil( positional_extent[ d ] );
+    }
+
+    // Adjust lower_left relative to the rounded center with the rounded up extent.
+    this->lower_left_ = rounded_center - this->extent_ / 2;
   }
 }
 
--- a/nestkernel/free_layer.h
+++ b/nestkernel/free_layer.h
@@ -122,8 +122,11 @@ FreeLayer< D >::set_status( const DictionaryDatum& d )
     max_point[ d ] = -std::numeric_limits< double >::infinity();
   }
 
-  num_local_nodes_ =
-    std::accumulate( this->node_collection_->begin(), this->node_collection_->end(), 0, []( size_t a, NodeIDTriple b ) {
+  num_local_nodes_ = std::accumulate( this->node_collection_->begin(),
+    this->node_collection_->end(),
+    0,
+    []( size_t a, NodeIDTriple b )
+    {
       const auto node = kernel().node_manager.get_mpi_local_node_or_device_head( b.node_id );
       return node->is_proxy() ? a : a + 1;
     } );
--- a/nestkernel/free_layer.h
+++ b/nestkernel/free_layer.h
@@ -122,11 +122,8 @@ FreeLayer< D >::set_status( const DictionaryDatum& d )
     max_point[ d ] = -std::numeric_limits< double >::infinity();
   }
 
-  num_local_nodes_ = std::accumulate( this->node_collection_->begin(),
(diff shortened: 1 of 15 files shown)


nestkernel/free_layer.h
```diff
@@ -234,9 +229,25 @@ FreeLayer< D >::set_status( const DictionaryDatum& d )
...
+      throw KernelException( "If only one node is created, 'extent' must be specified." );
+    }
+
+    const auto positional_extent = max_point - this->lower_left_;
+    const auto center = ( max_point + this->lower_left_ ) / 2;
+    Position< D > rounded_center;
+    for ( int d = 0; d < D; ++d )
+    {
+      // Round to the first decimal.
+      rounded_center[ d ] = ( std::round( 10.0 * center[ d ] ) ) / 10;
```
heplesser, line 243: Why this rounding? I would leave just the center that we get from the positions. If the user wants a specific center, they can specify the extent.

testsuite/pytests/test_spatial/test_connect_sliced.py
```diff
@@ -131,7 +131,10 @@ def test_sliced_spatial_inheritance(self):
         extent_pos = nest.spatial.free(nest.random.uniform(-0.5, 0.5), extent=[5., 2.5])
         for pos in [self.free_pos, self.grid_pos, wrapped_pos, extent_pos]:
             nest.ResetKernel()
-            nodes = nest.Create('iaf_psc_alpha', positions=pos)
+            # Don't specify number of nodes if using grid_pos
+            nodes = (nest.Create('iaf_psc_alpha', positions=pos)
+                     if pos == self.grid_pos else
+                     nest.Create('iaf_psc_alpha', 10, positions=pos))
```
heplesser, line 137: To make the code consistent with the comment, I would turn the statement around:
```suggestion
            # Don't specify number of nodes if using grid_pos
            nodes = (nest.Create('iaf_psc_alpha', 10, positions=pos)
                     if pos != self.grid_pos else
                     nest.Create('iaf_psc_alpha', positions=pos))
```

Since assigning to `nodes` is more of a side effect and the main thing is creating, I wonder if it wouldn't be more readable to write an explicit if-else statement:

```suggestion
            if pos == self.grid_pos: 
                # Don't specify number of nodes if using grid_pos
                nodes = (nest.Create('iaf_psc_alpha', positions=pos)
            else:
                nodes = nest.Create('iaf_psc_alpha', 10, positions=pos))
```

testsuite/pytests/test_spatial/test_layerNodeCollection.py
```diff
@@ -53,6 +54,59 @@ def test_addTwoLayers(self):
...
+                       (1, 3)):
+            nest.ResetKernel()
+            param = nest.random.uniform(ll * r, ur * r)
+            free_positions = nest.spatial.free(param, num_dimensions=2, edge_wrap=True)
+
+            nodes = nest.Create('iaf_psc_alpha', num_nodes, positions=free_positions)
+            spatial = nodes.spatial  # Extract spatial information
+            center_coord = r * (ur + ll) / 2.0  # Expected center coordinate
+            self.assertEqual(spatial["center"], (center_coord, center_coord))
+            self.assertEqual(spatial["extent"], (2 * r, 2 * r))
```
heplesser, line 79: Better use assertAlmostEqual?

testsuite/pytests/test_spatial/test_layerNodeCollection.py
```diff
@@ -53,6 +54,59 @@ def test_addTwoLayers(self):
         with self.assertRaises(nest.kernel.NESTError):
             c = layer1 + layer2
 
+    def test_extent_center_mask(self):
+        """Correct extent and center when connecting with mask"""
+        nest.SetKernelStatus({'rng_seed': 1234})
+        # Allowing up to 5 percent relative difference from expected number of connections,
+        # because of randomness.
+        rel_limit = 0.05
```
heplesser, line 62: Do you need to allow for 5% variation? I haven't done the maths ...

testsuite/pytests/test_spatial/test_layerNodeCollection.py
```diff
@@ -53,6 +54,59 @@ def test_addTwoLayers(self):
...
+        r = 0.5
+        # Iterate over values of lower left and upper right.
+        for ll, ur in ((-3, -1),
+                       (-2, 0),
+                       (-1, 1),
+                       (0, 2),
+                       (1, 3)):
+            nest.ResetKernel()
+            param = nest.random.uniform(ll * r, ur * r)
+            free_positions = nest.spatial.free(param, num_dimensions=2, edge_wrap=True)
```
heplesser, line 73: Could you add a corresponding test for 3 dimension?

heplesser (changes requested):
@hakonsbm Thanks for fixing this! I have added some comments, mostly minor points. Since this bug fix will change behavior of existing scripts, please add some explicit information about this in the [release notes](https://github.com/nest/nest-simulator/tree/master/doc/htmldoc/release_notes), although I am right now not certain if for v3.4 or v3.5.

testsuite/pytests/test_spatial/test_layerNodeCollection.py
hakonsbm: Because nodes are randomly distributed in space, the actual number of connections won't be *exactly* the calculated number of expected connections. With 100 nodes, 5% seemed like a sensible value. Do you have a suggestion for a more suitable value?

nicolossus (approved):
LGTM, thanks!

testsuite/pytests/test_spatial/test_layerNodeCollection.py
heplesser: No, I don't really have a better value. I just wondered about the sensitivity of the test. I think one could work out the standard deviation of the expected number of connections based on Eq (4.8) of [Daniel Hjertholm's master thesis](https://nmbu.brage.unit.no/nmbu-xmlui/handle/11250/189117), but for the sake of time it is probably easier to run the test with, let's say, 100 different seeds and measure the variation. And would the test trigger reliably without this fix in place?

heplesser: @hakonsbm Could you put this information in a comment in the test?

hakonsbm: @heplesser Done in 41e8c42ad0f06ce4a8e0d8bd7f6f5e8f32b6f17f.

heplesser (approved):

PR #2468, opened 2022-09-14
Correct library information provided to extension modules

heplesser:
Until now, the library information provided by `nest-config --libs` for extension module building include `-lnest`, i.e., `libnest`. This library only includes `nest_startup.{h,cpp}` and is only needed for the NEST executable or main PyNEST module. It can lead to loading problems when included with extension modules.

This PR ensures that `-lnest` is not returned by `nest-config --libs`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,7 +276,7 @@ foreach ( def ${SIONLIB_DEFINES} )
     set( ALL_CXXFLAGS "${ALL_CXXFLAGS} ${def}" )
 endforeach ()
 
-# libraries required to link extention modules
+# libraries required to link extension modules
 set( MODULE_LINK_LIBS
   "-lnestutil"
   "-lsli"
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -276,10 +276,9 @@ foreach ( def ${SIONLIB_DEFINES} )
     set( ALL_CXXFLAGS "${ALL_CXXFLAGS} ${def}" )
 endforeach ()
 
-# all libraries
-set( ALL_LIBS
+# libraries required to link extention modules
```
clinssen, line 279:
```suggestion
# libraries required to link extension modules
```

clinssen:
NESTML modules seem to build just fine against this branch. Approving!

clinssen (approved):

jougs (approved):
LGTM! Thanks for your investigation and the fix.

PR #2474, opened 2022-09-20
Refactor functions to use `MPI_Allreduce`

### LINKED ISSUE #1925: Implement functions with MPI_Allreduce
Converted from TODOs in #1549. The function `MPIManager::any_true()` and `RandomManager::check_rng_synchrony()` introduced in #1549 should be implemented using `MPI_Allreduce`.

Blocked because #1549 must be merged first.

hakonsbm:
Fixes #1925 by refactoring `MPIManager::any_true()` and `RandomManager::check_rng_synchrony()` to use `MPI_Allreduce`.

### DIFF
--- a/nestkernel/mpi_manager.cpp
+++ b/nestkernel/mpi_manager.cpp
@@ -1073,16 +1073,10 @@ nest::MPIManager::communicate_Allreduce_sum( std::vector< double >& send_buffer,
   recv_buffer.swap( send_buffer );
 }
 
-double
-nest::MPIManager::min_cross_ranks( double value )
+bool
+nest::MPIManager::equal_cross_ranks( const double )
 {
-  return value;
-}
-
-double
-nest::MPIManager::max_cross_ranks( double value )
-{
-  return value;
+  return true;
 }
 
 void
--- a/nestkernel/mpi_manager.cpp
+++ b/nestkernel/mpi_manager.cpp
@@ -799,21 +799,10 @@ nest::MPIManager::any_true( const bool my_bool )
     return my_bool;
   }
 
-  // since there is no MPI_BOOL we first convert to int
-  int my_int = my_bool;
-
-  std::vector< int > all_int( get_num_processes() );
-  MPI_Allgather( &my_int, 1, MPI_INT, &all_int[ 0 ], 1, MPI_INT, comm );
-  // check if any MPI process sent a "true"
-  for ( unsigned int i = 0; i < all_int.size(); ++i )
-  {
-    if ( all_int[ i ] != 0 )
-    {
-      return true;
-    }
-  }
-
-  return false;
+  const int my_int = my_bool;
+  int global_int;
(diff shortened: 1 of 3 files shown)


nestkernel/mpi_manager.cpp
```diff
@@ -719,18 +719,15 @@ nest::MPIManager::communicate_Allreduce_sum( std::vector< double >& send_buffer,
...
+nest::MPIManager::equal_cross_ranks( const double value )
 {
-  MPI_Allreduce( MPI_IN_PLACE, &value, 1, MPI_DOUBLE, MPI_MAX, comm );
-  return value;
+  // Flipping the sign of one argument to check both min and max values.
+  double values[ 2 ];
+  values[ 0 ] = -value;
+  values[ 1 ] = value;
+  MPI_Allreduce( MPI_IN_PLACE, &values, 2, MPI_DOUBLE, MPI_MIN, comm );
+  return values[ 0 ] == -values[ 1 ];
```
heplesser, line 730:
```suggestion
  if ( local_min == local_max )
  {
    values[ 0 ] = std::abs( value );
  }
  else
  {
    values[ 0 ] = -std::numeric_limits< double >::infinity();
  }
  values[ 1 ] = -values[ 0 ];

  MPI_Allreduce( MPI_IN_PLACE, &values, 2, MPI_DOUBLE, MPI_MIN, comm );

  return values[ 0 ] == -values[ 1 ] and values[ 0 ] != -std::numeric_limits< double >::infinity();
```

My idea here is the following: the -1 approach will fail in corner cases (see below). By setting -inf in case of local non-equality, we ensure that all ranks will find out about the local non-equality and will return the same result.

heplesser (changes requested):
Looks good, but I think there is a corner case for which I suggest a solution.

hakonsbm:
@heplesser I didn't think of that corner case, thanks for pointing it out. However, I don't agree with moving the local non-equality check into the `equal_cross_ranks()` function. The `equal_cross_ranks()` function has one job: to check if a value is equal across all ranks. Local non-equality is a special case, which should be handled locally before checking if values are equal across all ranks. Another problem with handling local non-equality in the `equal_cross_ranks()` function by passing `local_min` and `local_max` values is what to do in cases where there is only a single value, like when checking the rank-synchronized RNGs.

In the case where there is a local non-equality, this has to be communicated to all ranks. I've taken your suggestion of using `-inf` instead of `-1` to flag a local non-equality, which will become the result of `values[ 1 ]` after communication and in that way raise an exception in all MPI processes. Please have another look.

heplesser:
@hakonsbm You got a point. My idea was to isolate the "-inf" magic entirely in the `equal_cross_...` function but then one should probably have an `equal_across_vps()` functions. Your solution is fine and I will approve.

heplesser (approved):

hakonsbm:
@mlober Friendly ping

mlober:
Hi, sorry for the super late reply, I completely missed this. I looked through everything and I like your solution. Thanks for the improvement!

PR #2476, opened 2022-09-21
Add spike-transmission test to multithreading test

heplesser:
`test_multithreading.sli` until now only tests that thread numbers are set correctly and that nodes are correctly assigned to VPs. This test also checks that spikes are correctly transmitted between nodes on different threads.


jougs (approved):
LGTM!

I don't think we need a second review here and will merge immediately.

PR #2478, opened 2022-09-22
Use the PR base as reference SHA when getting changed files

### LINKED ISSUE #2471: All files in PRs are not always checked by static code analysis
Currently the static code analysis in the CI doesn't always check all changed files of a PR, only files changed since the *previously pushed* commit. Therefore if a PR is opened with formatting errors in one file, and another file is changed and pushed, the static code analysis will incorrectly pass because the errors are in an unchanged file in the latest commit of that PR.

The problem comes from the way we get changed files of a PR:
https://github.com/nest/nest-simulator/blob/5e8790c31399a4b2e2a0b340b8e91f1809de4c43/.github/workflows/nestbuildmatrix.yml#L77-L80
Here `github.event.before` is the SHA of the last commit pushed, and `github.event.after` the SHA of the current commit. 

Instead of getting files changed since the last push event, we should either get files changed in the current PR, or just check all files.

hakonsbm:
The PR base is the commit in master that the branch is built on. Using that as the reference for changed files ensures that the correct files are detected when getting changed files for the static code analysis. Fixes #2471.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -77,7 +77,7 @@ jobs:
       - name: "Find changed files"
         run: |
           echo "CHANGED_FILES<<EOF" >> $GITHUB_ENV
-          echo "$(git diff --name-only ${{ github.event.before }}..${{ github.event.after }})" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


jougs (approved):
I wholeheartedly approve :-)

heplesser:
@terhorstd approved verbally.

PR #2482, opened 2022-09-27
Update models to accept parameter objects as node parameters

### LINKED ISSUE #2480: Cannot set membrane voltage to random values for several models
**Describe the bug**
A `TypeMismatch` error is raised when trying to set `V_m` with a distribution from `nest.random` on neurons of the following types:

- iaf_psc_alpha_canon
- iaf_psc_alpha_ps
- iaf_psc_delta_ps
- glif_psc
- hh_cond_beta_gap_traub
- glif_cond

Setting `V_m` to a fixed value works.

**To Reproduce**
```python
n = nest.Create('iaf_psc_alpha_ps')
n.V_m = nest.random.uniform(0.0, 1.0)
```

 - NEST-Version: master@66255beb5 

**Additional context**

Here is a reproducer testing all models, trying both setting a fixed value and a random value. It may be useful as starting point for a regression test.

```python
import nest

nest.set_verbosity('M_FATAL')
for m in nest.node_models:
    n = nest.Create(m)
    try:
        n.get('V_m')
    except KeyError:
        continue

    try:
        n.V_m = 0.5
    except nest.NESTError:
        print(f'Constant Failed: {m}')

    try:
        n.V_m = nest.random.uniform(0., 1.)
    except nest.NESTError:
        print(f'Random Failed : {m}')
```

hakonsbm:
Some models were not updated to take parameter objects when setting parameters such as `V_m`. With this PR they are updated, and a test of setting `V_m` for all relevant node models is added. Fixes #2480.

### DIFF
--- a/models/glif_cond.cpp
+++ b/models/glif_cond.cpp
@@ -29,6 +29,7 @@
 #include <limits>
 
 // Includes from libnestutil:
+#include "dict_util.h"
 #include "numerics.h"
 
 // Includes from nestkernel:
@@ -224,7 +225,7 @@ nest::glif_cond::Parameters_::get( DictionaryDatum& d ) const
 }
 
 double
-nest::glif_cond::Parameters_::set( const DictionaryDatum& d )
+nest::glif_cond::Parameters_::set( const DictionaryDatum& d, Node* node )
 {
   // if E_L_ is changed, we need to adjust all variables defined relative to
   // E_L_
@@ -232,7 +233,7 @@ nest::glif_cond::Parameters_::set( const DictionaryDatum& d )
   updateValue< double >( d, names::E_L, E_L_ );
   const double delta_EL = E_L_ - ELold;
 
-  if ( updateValue< double >( d, names::V_reset, V_reset_ ) )
+  if ( updateValueParam< double >( d, names::V_reset, V_reset_, node ) )
   {
     V_reset_ -= E_L_;
   }
@@ -241,7 +242,7 @@ nest::glif_cond::Parameters_::set( const DictionaryDatum& d )
     V_reset_ -= delta_EL;
   }
 
-  if ( updateValue< double >( d, names::V_th, th_inf_ ) )
+  if ( updateValueParam< double >( d, names::V_th, th_inf_, node ) )
   {
     th_inf_ -= E_L_;
   }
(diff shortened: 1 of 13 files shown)


clinssen:
Thanks Håkon! Just to be sure: nodes do not need to implement ``set( const DictionaryDatum& )`` any longer on their ``State_`` and ``Parameters_`` members, but only ``set( const DictionaryDatum&, Node* )``?

hakonsbm:
@clinssen Yes, to be able to use a parameter object to set model parameters, the node pointer has to be passed when using `set()` of `State_` and `Parameters_`. This is because the parameter object may be based on spatial information, which it gets from the node.

heplesser (changes requested):
@hakonsbm Thanks! It looks good, but I wonder if you could extend the test to test setting all parameters that can be set. I believe there is a test somewhere that does something like this, it may be a SLI test. Could you see if you can find that?

I'd also suggest to call the test `test_regression_issue-2480.py` in line with the other Python regression tests we have.

clinssen:
@hakonsbm: could you point me to the changes I should make to make this work for synapses as well? I'm getting the same error when trying ``syn.w = nest.random.uniform(0., 1.)`` but passing ``this`` from within the C++ synapse object to ``updateValueParam()`` does not work; it only takes ``Node`` objects.

hakonsbm:
@heplesser I have moved the test to a new file. I can't find the test you mention which sets all parameters that can be set, though. I also thought about how one could test all settable parameters, but there are a couple problems with that. Getting all parameters that can be set would be complicated. And some node parameters require their values being lower or higher than other parameters, which is impossible to check without consulting the documentation for that model. Ultimately I think a test like that, while useful, would be too complicated and difficult to make reliable.

hakonsbm:
@clinssen Parameter objects that set parameters of non-spatial connections requires the post-synaptic node. So to set the status of a synapse model, you would have to pass the node you get here:

https://github.com/nest/nest-simulator/blob/66255beb5f7d2381f02a87b4a231c6c4220335fc/nestkernel/connection_manager.cpp#L270

It would then have to be passed as an argument down to the model's `set_status()` function.

For spatial connections it's a bit more complicated, because you then need not only the post-synaptic node, but also the layer, source node position, and target node position.

However, using parameter objects to set parameters of synapses is already possible when creating the connections.

heplesser (approved):
@hakonsbm Thanks, looks good to me! Your point that a test checking all parameters would be too difficult, so I think the test we have now is good.

clinssen:
So this is not possible/expected to fail with the same ``TypeMismatch`` exception?
https://github.com/nest/nestml/pull/818/files#diff-fb6a53f8c7efa8bb222864d58cd7a6fd6dd09cc9082324f71cacdff743074150R76

hakonsbm:
@clinssen Yes, without the changes outlined in my previous comment it's impossible to set parameters of a synapse with a parameter object after the connection is created.

clinssen (approved):
The corresponding NESTML PR now contains a working test for setting both neuron and synapse variables with distributions, so I can now pass off on this. Cheers!

PR #2485, opened 2022-09-30
Improved clarity of `Dwith-python` error

Helveg:
The legacy forms of `Dwith-python=2/3` still circulate. If a user now specifies `-Dwith-python=3` they are met with the message `Invalid option: -Dwith-python=3`. This message implies that the option `-Dwith-python` is invalid and unrecognized by CMake. I changed the error message to clarify that the option is valid, but the given value is invalid, and `ON` or `OFF` should be given instead.


hakonsbm (approved):
LGTM!

PR #2491, opened 2022-10-06
Ensure do_tests.sh completes even when tests fail

### LINKED ISSUE #2487: do_tests.sh terminates early if Python tests fail
When Python tests fail (possibly also when MPI, MUSIC or C++ tests fail), `do_tests.sh` will terminate early (immediately after the Python test suite is complete) and will not display the summary table of results and the list of failed tests, as can be seen here

https://github.com/heplesser/nest-simulator/actions/runs/3170175789/jobs/5162607621#step:8:8601

The final summary table when running Github Actions will then report that no tests were run at all

https://github.com/heplesser/nest-simulator/actions/runs/3170175789/jobs/5162607621#step:8:8660

The reason for this appears to be 

https://github.com/nest/nest-simulator/blob/e1c80d01f8318b5b2ce5d2f292cafb1f0e26047d/testsuite/do_tests.sh#L30

This makes sense in principle since it ensures that `do_tests.sh` will return with a non-zero exit code if anything goes wrong. 

For SLI-tests, this is not a problem, because `run_test.sh` writes

https://github.com/nest/nest-simulator/blob/e1c80d01f8318b5b2ce5d2f292cafb1f0e26047d/testsuite/run_test.sh#L91-L92

into every temporary tests script.

Can something similar be done for the tests not run via `run_test()`?

heplesser:
This PR changes do_tests.sh to turn off "exit on error" for the parts that execute Python and C++ tests so that failures can be properly collected and reported by our log analysis. For SLI-based tests, this is handled in `run_test.sh`.

This fixes #2487.

### DIFF
--- a/testsuite/do_tests.sh
+++ b/testsuite/do_tests.sh
@@ -494,16 +494,20 @@ if test "${PYTHON}"; then
 
     # Run all tests except those in the mpi* subdirectories because they cannot be run concurrently
     XUNIT_FILE="${REPORTDIR}/${XUNIT_NAME}.xml"
+    set +e
     "${PYTHON}" -m pytest --verbose --timeout $TIME_LIMIT --junit-xml="${XUNIT_FILE}" --numprocesses=1 \
           --ignore="${PYNEST_TEST_DIR}/mpi" "${PYNEST_TEST_DIR}" 2>&1 | tee -a "${TEST_LOGFILE}"
-
+    set -e
+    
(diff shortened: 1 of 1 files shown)


jougs (approved):
Looks totally good to me. This is probably the behavior that we initially should have had all along.

gtrensch (approved):
Looks good to me.

PR #2496, opened 2022-10-07
Modernizing the codebase - checking for nullptr

JanVogelsang:
**This PR is based on PR #2493.**  
In the C++ standard it is defined that `if ( ptr != nullptr )` gives the same result as `if ( ptr )` as `ptr` is converted to a `bool` in the latter case, which per standard evaluates to `true` only if `ptr != nullptr`.  
We can therefore convert all occurrences of
`if ( ptr != nullptr )` to ` if ( ptr )`
and
`if ( ptr == nullptr )` to `if ( not ptr )`
which makes it quite clear in which cases a pointer is expected to be valid and when not.

### DIFF
--- a/models/aeif_psc_delta.cpp
+++ b/models/aeif_psc_delta.cpp
@@ -352,7 +352,7 @@ nest::aeif_psc_delta::init_buffers_()
   // We must integrate this model with high-precision to obtain decent results
   B_.IntegrationStep_ = std::min( 0.01, B_.step_ );
 
-  if ( B_.s_ == nullptr )
+  if ( not _ )
   {
     B_.s_ = gsl_odeiv_step_alloc( gsl_odeiv_step_rkf45, State_::STATE_VEC_SIZE );
   }
@@ -360,7 +360,7 @@ nest::aeif_psc_delta::init_buffers_()
   {
     gsl_odeiv_step_reset( B_.s_ );
   }
-  if ( B_.c_ == nullptr )
+  if ( not _ )
   {
     B_.c_ = gsl_odeiv_control_yp_new( P_.gsl_error_tol, P_.gsl_error_tol );
   }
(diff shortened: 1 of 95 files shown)


models/aeif_psc_delta.cpp
```diff
@@ -352,15 +352,15 @@ nest::aeif_psc_delta::init_buffers_()
...
+  if ( B_.s_ == nullptr )
   {
     B_.s_ = gsl_odeiv_step_alloc( gsl_odeiv_step_rkf45, State_::STATE_VEC_SIZE );
   }
   else
   {
     gsl_odeiv_step_reset( B_.s_ );
   }
-  if ( B_.c_ == 0 )
+  if ( B_.c_ == nullptr )
```
jougs, line 363: Same here

models/aeif_psc_delta.cpp
```diff
@@ -369,7 +369,7 @@ nest::aeif_psc_delta::init_buffers_()
     gsl_odeiv_control_init( B_.c_, P_.gsl_error_tol, P_.gsl_error_tol, 0.0, 1.0 );
   }
 
-  if ( B_.e_ == 0 )
+  if ( B_.e_ == nullptr )
```
jougs, line 372: And here and probably more below...

JanVogelsang:
Not sure why, but after the latest commit, I now double-checked both `== nullptr` and `!= nullptr` and did not find any more occurrences.

jougs (approved):
What can I say?!

heplesser (approved):
@JanVogelsang Looks good, thank you! In the next PR (replacing `&&`, `||`, and `!` with `and`, `or` and `not`), it would be good to also remove superfluous parentheses in boolean expression, e.g. of to form `(n1) and (n2)`.

heplesser:
And merging ...

JanVogelsang:
> @JanVogelsang Looks good, thank you! In the next PR (replacing `&&`, `||`, and `!` with `and`, `or` and `not`), it would be good to also remove superfluous parentheses in boolean expression, e.g. of to form `(n1) and (n2)`.

I could only find very few occurrences (~25) of redundant parentheses, but I think I got them all. PR is incoming as soon as all tests pass.

PR #2502, opened 2022-10-13
Remove superfluous data structure for secondary events

### LINKED ISSUE #2492: Copying plain and lbl versions of secondary-event synapses can lead to crash
**Describe the bug**
NEST crashes if a secondary-event synapse, eg `gap_junction` is copied, the kernel then reset and after that the corresponding `_lbl` synapse is copied. This applies to

- gap_junction
- rate_connection_instantaneous
- rate_connection_delayed
- diffusion_connection

and also in opposite order.

**To Reproduce**

```python
nest.CopyModel('gap_junction', 'foo')
nest.ResetKernel()
nest.CopyModel('gap_junction_lbl', 'foo_lbl')
```

results in

```
Assertion failed: (not supports_syn_id( synid )), function add_syn_id, file event.h, line 1025.
```

**Desktop/Environment:**
 - NEST-Version: master@8028a79c2

**Additional context**
Discovered during work on #2379.

jougs:
This PR does the following:
* move the definitions/declarations for secondary events to their own file `nestkernel/secondary_event.h`
* remove the data structure for keeping the secondary events from the `ModelManager` in favor of the events that are stored in the connection models
* remove all pertaining functions for creating and deleting secondary events

On the way, I cleaned up the includes in the ModelsModule and some other minor things.

**Update on October 18:**
* clean up the handling of supported synapse IDs in `SecondaryEvent`s
* remove the `pristine_supported_syn_ids` from `SecondaryEvent`s
* simplify checks in `ConnectionManager::connect_()`
* protect a variable in `DynamicLoaderModule` from being used undefined

Fixes #2492.


nestkernel/model_manager.h
```diff
@@ -331,26 +324,11 @@ ModelManager::assert_valid_syn_id( synindex syn_id, thread t ) const
...
-      ( *iit->second ).reset_supported_syn_ids();
-      delete iit->second;
-    }
-  }
-  secondary_events_prototypes_.clear();
-}
-
 inline SecondaryEvent&
-ModelManager::get_secondary_event_prototype( const synindex syn_id, const thread tid ) const
+ModelManager::get_secondary_event_prototype( const synindex syn_id, const thread tid )
```
heplesser, line 328: Why is this no longer `const`? All the "get"ting suggests const-ness.

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,464 @@
...
+ * Bolten, M., Frommer, A. and Diesmann, M.,
+ * A unified framework for spiking and gap-junction interactions
+ * in distributed neuronal network simulations,
+ * Front. Neuroinform. 9:22. (2015),
+ * doi: 10.3389/fninf.2015.00022
+ */
+class SecondaryEvent : public Event
+{
+
+public:
```
heplesser, line 46: Could you add doxygen comments to all methods?

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,464 @@
...
+  virtual std::vector< unsigned int >::iterator& operator<<( std::vector< unsigned int >::iterator& pos ) = 0;
+  virtual std::vector< unsigned int >::iterator& operator>>( std::vector< unsigned int >::iterator& pos ) = 0;
+
+  virtual const std::vector< synindex >& get_supported_syn_ids() const = 0;
+
+  virtual void reset_supported_syn_ids() = 0;
+};
+
+/**
+ * This template function returns the number of uints covered by a variable of
```
heplesser, line 64: Could you rewrite this doxygen comment so that it starts with a concise one-line description and has clear @returns infomation. Similar for others below, with parameter info where applicable.

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,464 @@
...
+ * type T. This function is used to determine the storage demands for a
+ * variable of type T in the NEST communication buffer, which is of type
+ * std::vector<unsigned int>.
+ */
+template < typename T >
+size_t
+number_of_uints_covered()
```
heplesser, line 71: Maybe better 
```suggestion
size_in_uints()
```
And could this entire thing be turned into a constexpr, something like (not 100% sure the math works out)
```
{
  return ( sizeof( T ) + sizeof( unsigned int ) - 1 ) / sizeof( unsigned int );
}
```

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,464 @@
...
+ * duplicates for pyNN) which make it necessary to register several
+ * SecondaryConnectorModels with one SecondaryEvent. Therefore the synindices
+ * of all these models are added to supported_syn_ids_. The
+ * supports_syn_id()-function allows testing if a particular synid is mapped
+ * with the SecondaryEvent in question.
+ */
+template < typename DataType, typename Subclass >
+class DataSecondaryEvent : public SecondaryEvent
+{
+private:
```
heplesser, line 164: Why private section first?

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,464 @@
...
+  // we chose std::vector over std::set because we expect this to be short
+  static std::vector< synindex > pristine_supported_syn_ids_;
+  static std::vector< synindex > supported_syn_ids_;
+  static size_t coeff_length_; // length of coeffarray
```
heplesser, line 168: I suspect that these static vectors may be the cause of the problems in #2492 (@med-ayssar), because they will not be reset by a `ResetKernel()`. Why are they static and how are they reset?

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,464 @@
...
+
+  union CoeffarrayBegin
+  {
+    std::vector< unsigned int >::iterator as_uint;
+    typename std::vector< DataType >::iterator as_d;
```
heplesser, line 173: Maybe
```suggestion
    typename std::vector< DataType >::iterator as_DataType;
```
Same below.

heplesser (changes requested):
@jougs Thanks for tidying this up. I am still a bit in doubt why it is okay to remove the create secondary machinery as you do. Maybe easier to discuss that one-to-one.

jougs:
@heplesser, @med-ayssar: Thanks for your valuable input.

In a video call with @heplesser and @suku248 on Friday, we concluded that these changes probably don't go far enough and we should revisit the handling of secondary events more thoroughly. I will mark this as draft for now with the option of either adding more commits later on or closing it in favor of another PR.

nestkernel/model_manager.h
```diff
@@ -331,26 +324,11 @@ ModelManager::assert_valid_syn_id( synindex syn_id, thread t ) const
...
 }
 
-inline void
-ModelManager::delete_secondary_events_prototypes()
-{
-  for ( auto it = secondary_events_prototypes_.begin(); it != secondary_events_prototypes_.end(); ++it )
-  {
-    for ( std::map< synindex, SecondaryEvent* >::iterator iit = it->begin(); iit != it->end(); ++iit )
-    {
-      ( *iit->second ).reset_supported_syn_ids();
```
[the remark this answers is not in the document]
jougs: I have re-added the call to `reset_supported_syn_ids()` [here](https://github.com/nest/nest-simulator/pull/2502/files#diff-b082494e2589f18477776a5b797e38d3d179522a573213f4e6a534a9d9188993R485-R488). I think this should do the trick.

nestkernel/model_manager.h
jougs: I seem to remember that something down the line was not `const`. Not sure anymore, but there was a good reason.

nestkernel/secondary_event.h
jougs: This PR is not about rewriting the code. Please file another PR if you have suggestions like these.

nestkernel/secondary_event.h
jougs: I don't know. I'm not the author of this code, but have just moved it here.

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,447 @@
...
+
+    CoeffarrayBegin() {}; // need to provide default constructor due to
+                          // non-trivial constructors of iterators
+  } coeffarray_begin_;
+
+  union CoeffarrayEnd
+  {
+    std::vector< unsigned int >::iterator as_uint;
+    typename std::vector< DataType >::iterator as_d;
```
jougs, line 181:
```suggestion
    typename std::vector< DataType >::iterator as_DataType;
```

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,447 @@
...
+  supports_syn_id( const synindex synid ) const override
+  {
+    return ( std::find( supported_syn_ids_.begin(), supported_syn_ids_.end(), synid ) != supported_syn_ids_.end() );
+  }
+
+  void
+  set_coeffarray( std::vector< DataType >& ca )
+  {
+    coeffarray_begin_.as_d = ca.begin();
+    coeffarray_end_.as_d = ca.end();
```
jougs, line 243:
```suggestion
    coeffarray_begin_.as_DataType = ca.begin();
    coeffarray_end_.as_DataType = ca.end();
```

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,447 @@
...
+  /**
+   * The following operator is used to write the information of the
+   * DataSecondaryEvent into the secondary_events_buffer_.
+   * All DataSecondaryEvents are identified by the synid of the
+   * first element in supported_syn_ids_.
+   */
+  std::vector< unsigned int >::iterator&
+  operator>>( std::vector< unsigned int >::iterator& pos ) override
+  {
+    for ( typename std::vector< DataType >::iterator it = coeffarray_begin_.as_d; it != coeffarray_end_.as_d; ++it )
```
jougs, line 276:
```suggestion
    for ( auto it = coeffarray_begin_.as_DataType; it != coeffarray_end_.as_DataType; ++it )
```

nestkernel/secondary_event.h
jougs: The problematic code has been removed.

jougs:
I have added a bunch of new commits to this. It should be much more complete now. The description was updated to explain what was added in detail. This should now also fix #2492.

med-ayssar:
It seems that my changes [here](https://github.com/med-ayssar/nest-simulator/tree/fix_copying_synapse_with_second_event_delivery) also fixes the same issue #2492, and interestingly we both have the same exact failing tests in Full/ OpenMP_only tests. 

I will stop using my branch, and create a PR against this branch, if I will find something useful to fix the new failing tests.

jougs:
@med-ayssar: I don't have any failures locally anymore, let's see what the CI thinks.

nestkernel/dynamicloader.h
```diff
@@ -78,6 +78,21 @@ class DynamicLoaderModule : public SLIModule
...
+private:
+  /**
+   * Provide access to the list of linked modules managed DynamicLoader.
+   * This function controls access to the list of linked modules managed
+   * by DynamicLoaderModule via a Meyers' Singleton (Alexandrescu, ch 6.4).
+   * The list is filled by calls to @c registerLinkedModule().
+   */
+  static vecLinkedModules& getLinkedModules();
+
+  // vector to store handles and pointers to dynamic modules
```
heplesser, line 90:
```suggestion
  //! Vector to store handles and pointers to dynamic modules
```

nestkernel/model_manager.cpp
```diff
@@ -267,26 +269,16 @@ ModelManager::copy_connection_model_( index old_id, Name new_name )
   if ( new_id == invalid_synindex ) // we wrapped around (=63), maximal id of
                                     // connection_model = 62, see nest_types.h
```
heplesser, line 270:
```suggestion
  if ( new_id == invalid_synindex )
```
Values in comment are outdated and depend on configuration, so I suggest to drop.

nestkernel/model_manager.cpp
```diff
@@ -547,48 +543,23 @@ ModelManager::memory_info() const
...
 {
   if ( synapsedict_->known( cf->get_name() ) )
   {
     delete cf;
-    std::string msg = String::compose(
-      "A synapse type called '%1' already exists.\n"
-      "Please choose a different name!",
-      cf->get_name() );
+    std::string msg =
+      String::compose( "A synapse type called '%1' already exists.\nPlease choose a different name!", cf->get_name() );
```
heplesser, line 553: You use here `cf` which you delete one line above. Switch order (Github doesn't let me make a suggestion here, unfortunately).

heplesser (commented):
@jougs Nice work! I added a few more comments. Unfortunately, I haven't spotted yet where things go wrong :(.

nestkernel/connector_model.h
```diff
@@ -302,9 +303,17 @@ class GenericSecondaryConnectorModel : public GenericConnectorModel< ConnectionT
...
 
   ConnectorModel*
-  clone( std::string name ) const
+  clone( std::string name, synindex syn_id ) const
   {
-    return new GenericSecondaryConnectorModel( *this, name ); // calls copy construtor
+    ConnectorModel* new_cm = new GenericSecondaryConnectorModel( *this, name ); // calls copy construtor
+    new_cm->set_syn_id( syn_id );
+
+    if ( not new_cm->is_primary() )
```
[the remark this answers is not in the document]
jougs: The model handling for primary and secondary connection models will be unified in my next PR and I suggest that we keep this as is until then.

nestkernel/model_manager.cpp
```diff
@@ -547,48 +543,23 @@ ModelManager::memory_info() const
...
-    }
-  }
-}
-
 synindex
 ModelManager::register_connection_model_( ConnectorModel* cf )
 {
   if ( synapsedict_->known( cf->get_name() ) )
   {
     delete cf;
```
jougs, line 550:
```suggestion
    std::string msg =
      String::compose( "A synapse type called '%1' already exists.\nPlease choose a different name!", cf->get_name() );
    delete cf;
```

nestkernel/model_manager.cpp
jougs, line 552:
```suggestion
```

jougs: Done.

jougs (commented):
@heplesser: I've addressed your concerns in my latest commits and comments

jougs:
> @jougs Nice work! I added a few more comments. Unfortunately, I haven't spotted yet where things go wrong :(.

@heplesser: What do you mean by "things go wrong?" I don't see anything going wrong anymore.

nestkernel/secondary_event.h
```diff
@@ -0,0 +1,449 @@
...
+   * This function needs to be a virtual function of the base class
+   * DataSecondaryEvent as it is usually called from a pointer or
+   * reference of type SecondaryEvent.
+   *
+   * See also:
+   */
+  void
+  add_syn_id( const synindex synid ) override
+  {
+    if ( not supports_syn_id( synid ) )
```
med-ayssar, line 202: I am a bit sceptical about this change. I understand that it fixes the mentioned issue. But why not using a `Set` instead of vector, if we don't allow multiple insertions of the same element, and the second question, why does it event happen that the same element will be inserted more than once.

heplesser:
> > @jougs Nice work! I added a few more comments. Unfortunately, I haven't spotted yet where things go wrong :(.
> 
> @heplesser: What do you mean by "things go wrong?" I don't see anything going wrong anymore.

When I reviewed, the most recent run of the testsuite on Github still showed many failing tests, but clearly that is fixed now :).

nestkernel/model_manager.cpp
```diff
@@ -476,6 +467,10 @@ ModelManager::clear_connection_models_()
     {
       if ( connection_model )
       {
+        if ( not connection_model->is_primary() )
+        {
+          connection_model->get_event()->reset_supported_syn_ids();
```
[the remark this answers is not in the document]
jougs: The `supported_syn_ids` are not static members of the `ConnectorModel` but of the `Event` type they send. The vector thus only exists once, as both the original and the copied connection model point to the same `Event`. All functions involved are tiny and the compiler probably figures out that they can be `inline`d, so call overhead should be negligible and doing the `if` on the outside would probably not be more efficient than doing it outside.

In addition, the built-in models will completely disappear soon in the context of #1717, so nothing can be called on them.

heplesser:
@jougs Could you remove the `[:66]` in https://github.com/jougs/nest-simulator/blob/7bb2f6440102eea63ddfd0a82eb6d9771b98ce2b/testsuite/pytests/test_copy_model.py#L60-L62 so that that test also includes the secondary connections (and remove the comment)? In that way this test will check that #2492 is solved.

jougs:
> @jougs Could you remove the `[:66]` in https://github.com/jougs/nest-simulator/blob/7bb2f6440102eea63ddfd0a82eb6d9771b98ce2b/testsuite/pytests/test_copy_model.py#L60-L62 so that that test also includes the secondary connections (and remove the comment)? In that way this test will check that #2492 is solved.

Removed in a63a3bb. However, I'm a bit shocked that this was even in (in master). If we have a bug, we ave a bug and I don't think that such a workaround should ever be in master just to allow other PRs to go in. How would I even have known that this needs to be removed if you did not tell me?

heplesser:
> Removed in [a63a3bb](https://github.com/nest/nest-simulator/commit/a63a3bbd0b251b0f1f4009c0ee966cb528fa96b9). However, I'm a bit shocked that this was even in (in master). If we have a bug, we ave a bug and I don't think that such a workaround should ever be in master just to allow other PRs to go in. How would I even have known that this needs to be removed if you did not tell me?

I agree with you that this is borderline and we need to discuss how to handle situations like this in the future. In this case, #2379 fixed a concrete bug (problems copying all kinds of models), but also brought and improved test that detected the error described in #2492. This left the choice of either delaying #2379 until #2492 was fixed or weakening the test. I opted for the latter and put 

https://github.com/nest/nest-simulator/issues/2492#issuecomment-1271518853

into #2492 so that we would not forget to remove the restriction from the test. That info would probably have been more visible in the issue description, maybe even in form of a to-do list.

heplesser (approved):

med-ayssar (approved):

nestkernel/secondary_event.h
jougs: Will be fixed in a follow-up PR by @med-ayssar.

med-ayssar: Please refer to #2507

Issue #2503, opened 2022-10-14
The cortical microcircuit NEST examples do not generate the number of connections with a minimal delay that one might expect


gtrensch:
**Problem Description**
If you read out the connectivity from the cortical microcircuit NEST examples and reconstruct the distribution of synaptic delays, you will see a difference in the number of connections with a minimal synaptic delay compared to what one would expect for a truncated normal distribution. This is illustrated below for inhibitory projections: 

<img src="https://user-images.githubusercontent.com/19706694/195809777-8e71f0d5-84e8-4f99-a6d8-2ef023ee6a13.png" width="50%" height="50%">

Shown in blue is the discretized probability density function reconstructed from the connectivity read from NEST. For comparison, the expected values without cut-off are shown in grey.

The figure below further illustrates the effect. It was created by drawing connections using the corresponding NEST functions: 
`nest.math.redraw(nest.random.normal( mean = MEAN, std = SIGMA), min = MIN, max = np.Inf)`

<img src="https://user-images.githubusercontent.com/19706694/195814328-32edf547-cf08-4103-80ae-00cc1b7757b1.png" width="80%" height="80%">

This behavior is known and arises because delays in NEST are a discrete quantity. This is described in issue #833 and has also been added to the documentation (#2452).

**Solution**
The PyNEST and SLI cortical microcircuit example scripts need to be adapted accordingly.

PR #2504, opened 2022-10-14
Correct rounding of synaptic delays in the cortical microcircuit examples

gtrensch:
This pull request resolves issue #2503.
 
The changes applied by this pull request, slightly change the behavior of the cortical microcircuit (see #2503). It should therefore be discussed whether this is desired or not.


jhnnsnk (commented):
Well spotted! I think that this is the expected behavior and therefore we should update the model as suggested.
On the long run, one could consider of making this default for redrawing - what do you think?
I would suggest two small additions to this PR:
1. Could you please update the new firing rates in the [code](https://github.com/nest/nest-simulator/blob/8c386cec300b3369cfc798c59eee7e33f8cca16e/pynest/examples/Potjans_2014/network_params.py#L76) as they are used for downscaling?
2. To make the documentation clearer (see #2452), I would suggest to mention there that this effect not only appears for the uniform distribution but also for the normal distribution.

gtrensch:
I've adjusted the mean rates and also added a sentence to the documentation. Please review again.

heplesser (approved):
I think this is a good fix.

gtrensch:
I also updated the reference data.

heplesser (approved):

PR #2507, opened 2022-10-19
Replace std::vector with std::set in SecondaryEvent

med-ayssar:
The `SecondaryEvent` class in `secondary_event.h`  uses the `std::vector` to implement the same logic as the `std::set`. Therefore, It is more appropriate to use directly the `set` implementation.

### DIFF
--- a/nestkernel/secondary_event.h
+++ b/nestkernel/secondary_event.h
@@ -24,6 +24,7 @@
 #define SECONDARY_EVENT_H
 
 // c++ includes
+// C++ includes:
 #include <set>
 
 namespace nest
--- a/nestkernel/secondary_event.h
+++ b/nestkernel/secondary_event.h
@@ -23,6 +23,7 @@
 #ifndef SECONDARY_EVENT_H
 #define SECONDARY_EVENT_H
 
+// c++ includes
 #include <set>
 
 namespace nest
@@ -49,8 +50,6 @@ public:
   SecondaryEvent* clone() const override = 0;
 
   virtual void add_syn_id( const synindex synid ) = 0;
-
-  virtual bool has_synid( const synindex synid ) const = 0;
 
   //! size of event in units of unsigned int
   virtual size_t size() = 0;
@@ -156,9 +155,7 @@ read_from_comm_buffer( T& d, std::vector< unsigned int >::iterator& pos )
  * the usage of CopyModel or the creation of the labeled synapse model
  * duplicates for pyNN) which make it necessary to register several
  * SecondaryConnectorModels with one SecondaryEvent. Therefore the synindices
- * of all these models are added to supported_syn_ids_. The
- * has_synid()-function allows testing if a particular synid is mapped
- * with the SecondaryEvent in question.
+ * of all these models are added to supported_syn_ids_.
  */
 template < typename DataType, typename Subclass >
 class DataSecondaryEvent : public SecondaryEvent
@@ -200,7 +197,6 @@ public:
   void
   add_syn_id( const synindex synid ) override
   {
-
     VPManager::assert_single_threaded();
     supported_syn_ids_.insert( synid );
   }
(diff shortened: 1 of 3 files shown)


med-ayssar:
@jougs, could you take a look.

nestkernel/secondary_event.h
```diff
@@ -23,6 +23,8 @@
 #ifndef SECONDARY_EVENT_H
 #define SECONDARY_EVENT_H
 
+#include <set>
```
jougs, line 26: For consistency with other files
```suggestion
// C++ includes:
#include <set>
```

nestkernel/secondary_event.h
```diff
@@ -48,14 +50,14 @@ class SecondaryEvent : public Event
 
   virtual void add_syn_id( const synindex synid ) = 0;
 
-  virtual bool supports_syn_id( const synindex synid ) const = 0;
+  virtual bool has_synid( const synindex synid ) const = 0;
```
jougs, line 53: It seems that this function is not used anywhere after your change from `vector` to `set`. I suggest to remove it.

nestkernel/secondary_event.h
```diff
@@ -149,21 +151,20 @@ read_from_comm_buffer( T& d, std::vector< unsigned int >::iterator& pos )
...
- * SecondaryConnectorModel is stored as first element in the static vector
+ * SecondaryConnectorModel is stored as first element in the static set
  * supported_syn_ids_ on model registration. There are however reasons (e.g.
  * the usage of CopyModel or the creation of the labeled synapse model
  * duplicates for pyNN) which make it necessary to register several
  * SecondaryConnectorModels with one SecondaryEvent. Therefore the synindices
  * of all these models are added to supported_syn_ids_. The
- * supports_syn_id()-function allows testing if a particular synid is mapped
+ * has_synid()-function allows testing if a particular synid is mapped
  * with the SecondaryEvent in question.
```
jougs, line 161: Please also remove the reference to the function here.

nestkernel/secondary_event.h
```diff
@@ -199,14 +200,12 @@ class DataSecondaryEvent : public SecondaryEvent
   void
   add_syn_id( const synindex synid ) override
   {
-    if ( not supports_syn_id( synid ) )
-    {
-      VPManager::assert_single_threaded();
-      supported_syn_ids_.push_back( synid );
-    }
+
```
jougs, line 203:
```suggestion
```

nestkernel/secondary_event.h
```diff
@@ -233,9 +232,11 @@ class DataSecondaryEvent : public SecondaryEvent
...
 
   bool
-  supports_syn_id( const synindex synid ) const override
+  has_synid( const synindex synid ) const override
   {
-    return ( std::find( supported_syn_ids_.begin(), supported_syn_ids_.end(), synid ) != supported_syn_ids_.end() );
+    // Unfortunately, the set::contain function is only available as of C++20
+    auto item = supported_syn_ids_.find( synid );
+    return ( item != supported_syn_ids_.end() );
   }
```
jougs, line 240: Remove.

nestkernel/target_table_devices_impl.h
```diff
@@ -97,8 +97,8 @@ nest::TargetTableDevices::send_to_device( const thread tid,
   const std::vector< ConnectorModel* >& cm )
 {
   const index lid = kernel().vp_manager.node_id_to_lid( source_node_id );
-  const std::vector< synindex >& supported_syn_ids = e.get_supported_syn_ids();
-  for ( std::vector< synindex >::const_iterator cit = supported_syn_ids.begin(); cit != supported_syn_ids.end(); ++cit )
+  const std::set< synindex >& supported_syn_ids = e.get_supported_syn_ids();
+  for ( auto cit = supported_syn_ids.begin(); cit != supported_syn_ids.end(); ++cit )
```
jougs, line 101: Why not go all the way (also renaming the name `cit` [standing for `const_iterator`] to a more meaningful `syn_id`)?
```suggestion
  for ( auto& syn_id : e.get_supported_syn_ids() )
```

jougs (changes requested):
Good in principle, but I would go a bit further. See my comments.

heplesser (approved):
@med-ayssar Thanks! I agree with all suggestions by @jougs. To simplify the workflow, I approve now, assuming that @jougs will only approve once you have followed up on his suggestions.

jougs (approved):

PR #2514, opened 2022-10-25
Add missing tests of MUSIC models

### LINKED ISSUE #1459: Add missing tests of MUSIC models
The following MUSIC models lack tests in the testsuite: 

- `music_rate_in_proxy` (see #1449)
- `music_rate_out_proxy`
- `music_message_in_proxy`

Tests for these models should be added.

Note that because of conflicts with the Arbor backend, running the MUSIC tests causes a deadlock. The MUSIC test stage of `installcheck` is therefore currently disabled. #1404 contains a temporary fix, disabling the Arbor communication and re-enabling the MUSIC tests.

hakonsbm:
Adds tests for `music_rate_in_proxy`, `music_rate_out_proxy`, and `music_message_in_proxy`. 

Should fix #1459, but I'm not sure if we need some additional tests with more processes, as the tests `test_cont_proxy.music` and `issue_1974.sli` already use multiple processes. Any comment on this, @heplesser?


testsuite/do_tests.sh
```diff
@@ -408,6 +408,10 @@ if test "${MUSIC}"; then
         sh_file="${TESTDIR}/$(basename ${music_file} .music).sh"
         if test ! -f "${sh_file}"; then sh_file=""; fi
 
+        # Check if there is an accompanying input data file
+        input_file="${TESTDIR}/$(basename ${music_file} .music)0.dat"
```
heplesser, line 412: I don't get the logic here. Why the 0 at the end? Would it not be better to grep the music file to get the name of the input file from this line
```
# input file: test_message0.dat
```

heplesser (changes requested):
Thanks! Just one little point and I think these tests suffice.

testsuite/do_tests.sh
hakonsbm: That's what I did initially. But if there is no data file specified, grep fails with a nonzero exit code and installcheck crashes. Instead  of jumping through hoops to avid the exit code, we simply assume here that there is a datafile and unset `input_file` if there isn't.

The 0 has to do with how MUSIC somehow handles input files:
https://github.com/nest/nest-simulator/blob/a06b43c9ea67973a17774af21c9c4f5f08a4ecdb/testsuite/musictests/test_message.music#L26-L31

heplesser (approved):

jougs (approved):
LGTM! Thanks for addressing this.

Issue #2515, opened 2022-10-25
MacOS CI tests fail


heplesser:
The last successful NEST build on MacOS was [#2136](https://github.com/nest/nest-simulator/actions/runs/3288963056/jobs/5419998743). The next matrix build and all subsequent failed [#2137(https://github.com/nest/nest-simulator/actions/runs/3296060754/jobs/5464998013)] because [CMake cannot find OpenMP](https://github.com/nest/nest-simulator/actions/runs/3296060754/jobs/5464998013#step:8:286).

Comparing the logs, two changes seem to have taken place between the two runs:
- Github has change their confiugration from 20221010.1 to 20221018.1, but the [diff shows nothing that should have affected us](https://github.com/actions/runner-images/commit/067c72fb350e597919af31f45d6bfd10767e7f84)
- Homebrew has [hiked libomp from 14.0 to 15.0](https://github.com/Homebrew/homebrew-core/commits/e1f828798dd11353d4a8ebec4d59349a2928574c/Formula/libomp.rb).

It seems more likely that the version hike in libomp causes the problem. Unfortunately, there is no (easy) way to install libomp 14.0.6.

As a temporary remedy, I disable OpenMP for the MacOS runner. Once the problem is solved properly, that needs to be re-instated.

- [ ] Solve libomp problem on macOS runner
- [ ] Re-activate OpenMP in [macOS configuration](https://github.com/heplesser/nest-simulator/blob/80a1fdff0b1005ae9a024c02b5bfa46028520d19/build_support/ci_build.sh#L187)

heplesser:
This is due to a known problem in CMAKE: https://gitlab.kitware.com/cmake/cmake/-/issues/24097

PR #2517, opened 2022-10-25
Move connection flags to ``Connection`` static const member

clinssen:
A ``Connection`` knows best about its own required connection flags. These flags are therefore moved from the ``ModelsModule``, where each individual model is registered with the kernel through a ``register_connection_model()`` call, to the ``Connection`` class and its children (such as the ``clopath_synapse``) as a new member ``const static RegisterConnectionModelFlags flags``. Clopath and Urbanczik models initialize them using their own flags, OR-ed with the default flags in the base class.

@jougs: should we keep the non-const ``operator|`` on our custom ``Enum`` type, or ditch it?

If you are happy with how this looks, I will make the same changes for the secondary connection flags.

### DIFF
--- a/models/gap_junction.h
+++ b/models/gap_junction.h
@@ -168,14 +168,12 @@ GapJunction< targetidentifierT >::get_status( DictionaryDatum& d ) const
   def< long >( d, names::size_of, sizeof( *this ) );
 }
 
-
 template < typename targetidentifierT >
 SecondaryEvent*
 GapJunction< targetidentifierT >::get_secondary_event()
 {
   return new GapJunctionEvent();
 }
-
 
 template < typename targetidentifierT >
 void
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -750,7 +750,6 @@ nest::ConnectionManager::connect_( Node& source,
   const double weight )
 {
   ConnectorModel& conn_model = kernel().model_manager.get_connection_model( syn_id, tid );
-  const bool is_primary = conn_model.has_property( ConnectionModelProperties::IS_PRIMARY );
 
   const bool clopath_archiving = conn_model.has_property( ConnectionModelProperties::REQUIRES_CLOPATH_ARCHIVING );
   if ( clopath_archiving and not dynamic_cast< ClopathArchivingNode* >( &target ) )
@@ -764,6 +763,7 @@ nest::ConnectionManager::connect_( Node& source,
     throw NotImplemented( "This synapse model is not supported by the neuron model of at least one  connection." );
   }
 
+  const bool is_primary = conn_model.has_property( ConnectionModelProperties::IS_PRIMARY );
   conn_model.add_connection( source, target, connections_[ tid ], syn_id, params, delay, weight );
   source_table_.add_source( tid, syn_id, s_node_id, is_primary );
 


heplesser:
@clinssen This seems like a good idea. Can you merge master to make macOS pass again (see #2515)?

nestkernel/connection.h
```diff
@@ -120,6 +121,8 @@ class Connection
   // connections not used in primary connectors
   typedef SecondaryEvent EventType;
 
+  const static RegisterConnectionModelFlags flags = default_connection_model_flags;
```
jougs, line 124: Can the definition of the defaults be moved from `nest.h` to this file as well? They should only be needed here, shouldn't they?

nestkernel/model_manager_impl.h
```diff
@@ -55,8 +55,10 @@ ModelManager::register_node_model( const Name& name, std::string deprecation_inf
 
 template < template < typename targetidentifierT > class ConnectionT >
 void
-ModelManager::register_connection_model( const std::string& name, const RegisterConnectionModelFlags flags )
+ModelManager::register_connection_model( const std::string& name )
 {
+  RegisterConnectionModelFlags flags = ConnectionT< TargetIdentifierPtrRport >::flags;
+
```
jougs, line 61: Instead of getting the `flags` out of the `ConnectionT` here, and putting them into the `GenericConnectorModel` in the form of individual `bool`s in lines 64ff. and 75ff., the `GenericConnectorModel` constructor could just get them out of the `ConnectionT` by itself. The `ConnectionT` is a template argument to that after all.

The `if` in line 73 should also be changed to use the flag from the `ConnectionT` directly instead of using the "intermediate" `flags` variable.

jougs (commented):
Very nice :-)

I have added some inline comments to further improve the encapsulation of the flags.

nestkernel/connection.h
heplesser: Following up on this, couldn't this actually become (with some implicit suggestions for name changes)
```c++
constexpr ConnectionModelProperties properties = 
ConnectionModelProperties::SUPPORTS_HPC
  | ConnectionModelProperties::SUPPORTS_LBL
  | ConnectionModelProperties::IS_PRIMARY
  | ConnectionModelProperties::HAS_DELAY;
```
and then in the derived classes on could do
```c++
constexpr ConnectionModelProperties properties = ConnectionT::properties |           
             ConnectionModelProperties::REQUIRES_URBANCZIK_ARCHIVING;
```

And maybe rename `enumFlagSet()` to `has_property()` to focus on meaning and not implementation?

heplesser (commented):
@clinssen This looks very interesting. I agree with @jougs' comments and they inspired me to some more suggestions.

jougs:
@clinssen: I know you are busy, but still: any news on this?!

clinssen:
Thank you for the reviews! The new changes should address all of the comments raised.

nestkernel/connector_model.h
```diff
@@ -115,57 +135,59 @@ class ConnectorModel
...
   {
-    return supports_wfr_;
+    return has_property( properties_, ConnectionModelProperties::SUPPORTS_WFR );
   }
 
+  bool
+  has_flag_set( const ConnectionModelProperties& flag )
+  {
+    return has_property( properties_, flag );
+  }
```
heplesser, line 175: Rename and make const
```suggestion
  bool
  has_property( const ConnectionModelProperties& property ) const
  {
    return has_property( properties_, property );
  }
```

nestkernel/connector_model.h
```diff
@@ -115,57 +135,59 @@ class ConnectorModel
   bool
   is_primary() const
   {
-    return is_primary_;
+    return has_property( properties_, ConnectionModelProperties::IS_PRIMARY );
```
heplesser, line 138: I would change this and all other corresponding code to
```suggestion
    return has_property( ConnectionModelProperties::IS_PRIMARY );
```
where `has_property()` is renamed from `has_flag_set()`, see below.

libnestutil/enum_bitfield.h
```diff
@@ -133,10 +133,10 @@ operator^=( Enum& lhs, Enum rhs )
 
 template < typename Enum >
 bool
-enumFlagSet( const Enum en, const Enum flag )
+has_property( const Enum en, const Enum property )
 {
   using underlying = typename std::underlying_type< Enum >::type;
-  return static_cast< underlying >( en & flag ) != 0;
+  return static_cast< underlying >( en & property ) != 0;
 }
```
heplesser, line 140: I am not entirely happy with the `has_property()` name here (even though I first suggested it). I think it makes more sense to be able to write something like
```
conn_model.has_property( ConnModProp::IS_PRIMARY )
```
and therefore suggest below to rename `has_flag_set()`. Then having `has_property()` once from here and once from `ConnectorModel` gets confusing. Since the code here is more technical, behind the scenes, maybe calling this `flag_is_set()` and renaming parameters for clarity will give clearer code overall, and fits also better with terminology in the documentation above:
```
template < typename Enum >
bool
flag_is_set( const Enum flags, const Enum flag )
{
   using underlying = typename std::underlying_type< Enum >::type;
   return static_cast< underlying >( flags & flag );
}
```
I also dropped the comparison `!=0`, since anything different from 0 is true (see #2499).

By delegating flag testing then to `ConnectorModel::has_property()`, all code that currently looks like (`has_property` from here)
```
    return has_property( properties_, ConnectionModelProperties::IS_PRIMARY );
```
will become
```
    return has_property( ConnectionModelProperties::IS_PRIMARY );
```
with `has_property()` from `ConnectorModel`.

Sorry I didn't think this properly through in my first review.

heplesser (changes requested):
@clinssen This looks so much nicer now that it inspired me to some ideas to make it even tidier :).

jougs:
@clinssen: I have just sent you https://github.com/clinssen/nest-simulator/pull/6 to fix the problem with the failing assertions. Locally, I still see some tests failing and I can look into this hopefully tomorrow.

nestkernel/connector_model.h
```diff
@@ -115,57 +127,59 @@ class ConnectorModel
...
   {
-    return requires_urbanczik_archiving_;
+    return has_property( properties_, ConnectionModelProperties::REQUIRES_URBANCZIK_ARCHIVING );
   }
 
   bool
   supports_wfr() const
   {
-    return supports_wfr_;
+    return has_property( properties_, ConnectionModelProperties::SUPPORTS_WFR );
```
jougs, line 160: Do we still need these forwarders? Can't we just use `ConnModel.has_flag_set( ConnectionModelProperties::BLA )` from the outside? I think I would like that more, as it is more explicit and it would again remove some code here

nestkernel/connector_model.h
jougs: :+1:

jougs (commented):
I second most of @heplesser's comments and have added some additional ones.

nestkernel/connector_model.h
```diff
@@ -46,17 +47,36 @@ class CommonSynapseProperties;
...
+{
+  static const bool enable = true;
+};
+
+const ConnectionModelProperties default_connection_model_properties = ConnectionModelProperties::SUPPORTS_HPC
+  | ConnectionModelProperties::SUPPORTS_LBL | ConnectionModelProperties::IS_PRIMARY
+  | ConnectionModelProperties::HAS_DELAY;
+
+const ConnectionModelProperties default_secondary_connection_model_properties =
+  ConnectionModelProperties::SUPPORTS_WFR | ConnectionModelProperties::HAS_DELAY;
```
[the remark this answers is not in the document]
jougs: Defaults are gone now, so I'm resolving this conversation.

models/rate_connection_instantaneous.h
```diff
@@ -82,6 +82,8 @@ class RateConnectionInstantaneous : public Connection< targetidentifierT >
   typedef Connection< targetidentifierT > ConnectionBase;
   typedef InstantaneousRateConnectionEvent EventType;
 
+  static constexpr ConnectionModelProperties secondaryProperties = ConnectionModelProperties::SUPPORTS_WFR;
```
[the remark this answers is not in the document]
clinssen: Secondary properties have now gone away.

clinssen:
@jougs, @heplesser: thanks very much for your input! The latest commits should address all of your comments exactly as you proposed them, so there are no unresolved conversations remaining. Could you give it another look (or seal of approval)? Cheers!

jougs (changes requested):
Only some minor remarks left.

nestkernel/conn_builder.cpp
```diff
@@ -205,9 +205,12 @@ nest::ConnBuilder::connect()
...
   // We test here, and not in the ConnBuilder constructor, so the derived
   // classes are fully constructed when the test is executed
-  for ( auto syn_model : synapse_model_id_ )
+  for ( auto synapse_model_id : synapse_model_id_ )
   {
-    if ( kernel().model_manager.connector_requires_symmetric( syn_model ) and not( is_symmetric() or make_symmetric_ ) )
+    const ConnectorModel& synapse_model = kernel().model_manager.get_connection_model( synapse_model_id );
+    const bool requires_symmetric = synapse_model.has_property( ConnectionModelProperties::REQUIRES_SYMMETRIC );
+
+    if ( requires_symmetric and not( is_symmetric() or make_symmetric_ ) )
```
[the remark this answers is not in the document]
clinssen: this would make the naming a bit less consistent, no?

jougs (approved):
Thanks!

nestkernel/conn_builder.cpp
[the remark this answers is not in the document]
jougs: You're right.

jougs:
@heplesser, @med-ayssar: could one of you please have a quick look? I think this is good to go. Thanks!

nestkernel/model_manager.h
```diff
@@ -132,12 +133,7 @@ class ModelManager : public ManagerInterface
    * @param name The name under which the ConnectorModel will be registered.
    */
   template < template < typename targetidentifierT > class ConnectionT >
-  void register_connection_model( const std::string& name,
-    const RegisterConnectionModelFlags flags = default_connection_model_flags );
-
-  template < template < typename targetidentifierT > class ConnectionT >
-  void register_secondary_connection_model( const std::string& name,
-    const RegisterConnectionModelFlags flags = default_secondary_connection_model_flags );
+  void register_connection_model( const std::string& name );
```
med-ayssar, line 136: After simplifying the declaration of this function, wouldn't it be even better to discard the `string name` argument, and use a macro to convert the class name to a string inside the definition of `register_connection_model`.

med-ayssar (approved):

nestkernel/model_manager.h
jougs: I remember debugging situations where it came in handy that I was able to register the same model under different names and would thus keep this flexible for now.

PR #2518, opened 2022-10-26
Getter for NodeCollection step on C++ level

nicolossus:
The main purpose of this PR is to make the step of a NodeCollection accessible on C++ level by introducing a getter. The PR also includes a NodeCollection creator that is convenient for C++ test programs. With the new creator, a NodeCollection can be instantiated in C++ by solely providing a vector of node IDs. A kernel must be instantiated before the call to the creator, however, the usual NodeCollection creators require that a node model and several other managers are instantiated.


nestkernel/node_collection.h
```diff
@@ -218,6 +218,19 @@ class NodeCollection
    */
   static NodeCollectionPTR create( const std::vector< index >& node_ids );
 
+  /**
+   * @brief Create a NodeCollection from an array of node IDs without requiring
+   * instantiation of a node model and associated managers. Intended to simplify
+   * C++ test programs where just a NodeCollection is needed. Since the fingerprint
+   * of the NodeCollection superclass require a kernel instance, a kernel must be
+   * instantiated, i.e., KernelManager::create_kernel_manager(), before this
```
hakonsbm, line 226:
```suggestion
   * instantiated, i.e., by KernelManager::create_kernel_manager(), before this
```

nestkernel/node_collection.h
```diff
@@ -413,8 +433,9 @@ class NodeCollectionPrimitive : public NodeCollection
    * @param first The first node ID in the primitive
    * @param last  The last node ID in the primitive
    * @param model_id Model ID of the node IDs
+   * @param for_testing Whether the NodeCollection is intended for C++ test programs
```
hakonsbm, line 436:
```suggestion
   * @param for_testing Whether the NodeCollection is intended for C++ test programs.
   * Defaults to false if not specified.
```

testsuite/cpptests/test_node_collection_step.h
```diff
@@ -0,0 +1,91 @@
...
+#ifndef TEST_NODE_COLLECTION_STEP_H
+#define TEST_NODE_COLLECTION_STEP_H
+
+#define BOOST_TEST_DYN_LINK
+#include <boost/test/unit_test.hpp>
+
+// C++ includes
+#include <vector>
+
+// Includes from nestkerne
```
hakonsbm, line 32:
```suggestion
// Includes from nestkernel
```

testsuite/cpptests/test_node_collection_step.h
```diff
@@ -0,0 +1,91 @@
...
+ * This test checks that a NodeCollection created from a vector of homogeneous and
+ * contiguous node IDs (primitive NodeCollection) has a step size of 1.
+ */
+BOOST_AUTO_TEST_CASE( test_primitive_node_collection )
+{
+  KernelManager::create_kernel_manager();
+  const std::vector< index > node_ids { 1, 2, 3 };
+  auto nc = NodeCollection::create_test_collection( node_ids );
+  size_t expected_nc_step = 1;
+  auto actual_nc_step = nc->step();
```
hakonsbm, line 54:
```suggestion
  const auto nc = NodeCollection::create_test_collection( node_ids );
  const size_t expected_nc_step = 1;
  const auto actual_nc_step = nc->step();
```

testsuite/cpptests/test_node_collection_step.h
```diff
@@ -0,0 +1,91 @@
...
+ * This test checks if an empty NodeCollection has a step size of 1, since an empty
+ * NodeCollection should be primitive.
+ */
+BOOST_AUTO_TEST_CASE( test_empty_node_collection )
+{
+  KernelManager::create_kernel_manager();
+  const std::vector< index > node_ids {};
+  auto nc = NodeCollection::create_test_collection( node_ids );
+  size_t expected_nc_step = 1;
+  auto actual_nc_step = nc->step();
```
hakonsbm, line 69:
```suggestion
  const auto nc = NodeCollection::create_test_collection( node_ids );
  const size_t expected_nc_step = 1;
  const auto actual_nc_step = nc->step();
```

testsuite/cpptests/test_node_collection_step.h
```diff
@@ -0,0 +1,91 @@
...
+ * NodeCollection created from a NodeCollection slice with a step of 2 between node IDs.
+ */
+BOOST_AUTO_TEST_CASE( test_composite_node_collection )
+{
+  KernelManager::create_kernel_manager();
+  std::vector< index > node_ids { 1, 2, 3, 4, 5 };
+  auto nc = NodeCollection::create_test_collection( node_ids );
+  auto nc_slice = nc->slice( 0, 4, 2 );
+  size_t expected_nc_step = 2;
+  auto actual_nc_step = nc_slice->step();
```
hakonsbm, line 84:
```suggestion
  const std::vector< index > node_ids { 1, 2, 3, 4, 5 };
  const auto nc = NodeCollection::create_test_collection( node_ids );
  const auto nc_slice = nc->slice( 0, 4, 2 );
  const size_t expected_nc_step = 2;
  const auto actual_nc_step = nc_slice->step();
```

hakonsbm (changes requested):
Looks pretty good, I have just some minor suggestions.

nestkernel/node_collection.h
```diff
@@ -218,6 +218,19 @@ class NodeCollection
    */
   static NodeCollectionPTR create( const std::vector< index >& node_ids );
 
+  /**
+   * @brief Create a NodeCollection from an array of node IDs without requiring
+   * instantiation of a node model and associated managers. Intended to simplify
+   * C++ test programs where just a NodeCollection is needed. Since the fingerprint
+   * of the NodeCollection superclass require a kernel instance, a kernel must be
+   * instantiated, i.e., KernelManager::create_kernel_manager(), before this
+   * creator can be used.
```
heplesser, line 227: Maybe add a @note that these mock collections can only be use to test certain properties of node collections.

nestkernel/node_collection.h
```diff
@@ -218,6 +218,19 @@ class NodeCollection
...
+   *
+   * @param node_ids Array of node IDs from which to create the NodeCollection
+   * @return a NodeCollection pointer to the created NodeCollection
+   */
+  static NodeCollectionPTR create_test_collection( const std::vector< index >& node_ids );
```
heplesser, line 232:
```suggestion
  static NodeCollectionPTR create_mock_node_collection( const std::vector< index >& node_ids );
```
Mocking is the technical term for incomplete instances like this.

nestkernel/node_collection.h
```diff
@@ -413,8 +433,9 @@ class NodeCollectionPrimitive : public NodeCollection
    * @param first The first node ID in the primitive
    * @param last  The last node ID in the primitive
    * @param model_id Model ID of the node IDs
+   * @param for_testing Whether the NodeCollection is intended for C++ test programs
    */
-  NodeCollectionPrimitive( index first, index last, index model_id );
+  NodeCollectionPrimitive( index first, index last, index model_id, bool for_testing = false );
```
heplesser, line 439:
```suggestion
  NodeCollectionPrimitive( index first, index last, index model_id, mocker = false );
```

nestkernel/node_collection.cpp
```diff
@@ -283,14 +333,21 @@ NodeCollectionPrimitive::NodeCollectionPrimitive( index first,
   assert( first_ <= last_ );
 }
 
-NodeCollectionPrimitive::NodeCollectionPrimitive( index first, index last, index model_id )
+NodeCollectionPrimitive::NodeCollectionPrimitive( index first, index last, index model_id, bool for_testing )
```
heplesser, line 336: I am not really happy with the flag here, and actually the fact that you need the additional create function for a mock version of `NodeCollectionPrimitive`. If I understand the code below correctly, you only need the distinction because the `get_node_model()` call below will fail in C++ test cases not including the full kernel. 

To build the code below, it clearly needs to be linked against the full kernel. I see several possibilities:

1. Add `get_num_models()` to `ModelManager` and execute the `nodes_have_no_proxies_` line only if there are models, otherwise, set to `false`. Could even be a 
```c++
nodes_have_no_proxies_ = kernel().model_manager.get_num_models() > 0 and not kernel().model_manager.get_node_model( model_id_ )->has_proxies();
```
2. Put the call in a try-except and just ignore errors. This is dangerous if something should really be wrong. It could be wrapped using a compiler flag such as `NEST_CXX_TEST` and re-throw if this is off, otherwise pass silently.
3. Create a class `NodeCollectionPrimitiveBase` which contains all of `NodeCollectionPrimitive` but fixes `nodes_have_no_proxies_` to `false`. Then derive `NodeCollectionPrimitive` from it, and set the field correctly. The, make the `CollectionT` a template argument of `create_test_collection()` with `NodeCollectionPrimitive` as default value. In the test program itself, you can then derive `NodeCollectionPrimitiveMocker` from `NodeCollectionPrimitiveBase` and use that as template argument for `create_test_collection()`; you could probably also use `...Base` directly.  With that in place, you can then use the `create_test_collection()` as a general mechanism to build a node collection from arbitrary arrays of GIDs.

I think #3 is the most general and powerful approach which also keeps all test-specific stuff in the test.

nestkernel/node_collection.cpp
```diff
@@ -208,6 +208,56 @@ NodeCollection::create( const std::vector< index >& node_ids_vector )
...
+
+  index current_first = node_ids[ 0 ];
+  index current_last = current_first;
+
+  std::vector< NodeCollectionPrimitive > parts;
+
+  index old_node_id = current_first;
+  for ( auto node_id = ++( node_ids.begin() ); node_id != node_ids.end(); ++node_id )
+  {
+    if ( *node_id == old_node_id )
```
heplesser, line 228: This implies that ids are sorted and will fail if they are not. I would suggest to explicitly require (in documentation) that `node_ids` is sorted and then test this here by `if ( *node_id < old_node_id )`.

heplesser (changes requested):
@nicolossus Thank you for this PR. It made me wonder more generally about how to create mock objects in C++ tests. Since we have only very few C++ tests so far, this is new territory. Please see my comment below, starting with the longest comment. Some of the other comments don't entirely apply if you follow up the longest.

@hakonsbm I thought we explicitly required that the nodes represented by the node_ids in a  NodeCollection exist, but I cannot see that check. Do I remember wrong?

@med-ayssar @JanVogelsang @jougs Since this is new territory in test design and I have very limited experience with C++-level testing, I'd appreciate your comments on how to handle mocking in the best way, assuming you have more experience with this than I do.

hakonsbm:
> @hakonsbm I thought we explicitly required that the nodes represented by the node_ids in a NodeCollection exist, but I cannot see that check. Do I remember wrong?

It's not required by the mocking function to avoid having to initialize the models and actually creating the nodes. If you try creating a NodeCollection normally with a non-existent `node_id`, you get an `UnknownNode` error.

heplesser:
> It's not required by the mocking function to avoid having to initialize the models and actually creating the nodes. If you try creating a NodeCollection normally with a non-existent `node_id`, you get an `UnknownNode` error.

Could you just remind me where that is implemented?

nestkernel/node_collection.cpp
JanVogelsang: No. 3 Would definitely be the way to go if we continue on adding more C++ tests (incl. mocking) in the future. From a design perspective, this would be very clean.
However, if we don't plan on adding any more mocks in the near future, a simple compiler flag wrapping test code would be sufficient as well in my opinion.

heplesser: I expect that we will have more C++ tests in the future. We therefore should start out with a good design as we all have a tendency to build new code on existing examples. Good examples at the beginning are therefore crucial.

One thing I am not entirely happy with in 3. is the need to introduce the additional `...Base` class just to make mocking possible. I experimented a bit to see if one could have the dependency on `kernel()` in a base class and the define it away in a derived class (so that the ...`Mocker` class could derive directly from `NodeCollectionPrimitive`), but since class base constructors never call derived-class methods, I haven't found a way to achieve that.

med-ayssar: I might have a different idea, but I don't know how clean it would sound. 

We can use a **forward-declaration** of a class. This class will be called `NodeCollectionInitializer` (for each target class to test, we create a` [TargetClassName]Initializer`).  In the `TargetClass` (i.e, `NodeCollectionPrimitive`), we add a new constructor that takes a `NodeCollectionPrimitiveInitializer` as an argument.

This `NodeCollectionPrimitiveInitializer` is just a simple class that stores information about the `TargetClass` and provides only a `constructor` and `Getter` functions. 

The new added `constructor` in the `TargetClass` will not be defined in the `TargetClass`'s declaration nor definition inside the nestkernel, but instead the new  `constructor`  will be defined inside the cppunittest folder. So when compiling nest, the  `constructor` won't be used, and also we will not have a linker problem, but when compiling the unitests, we add the file that has the definition of the new  `constructor`.

In the definition of the new  `constructor` in the `TagetClass`, we simply use the ` [TargetClassName]Initializer` to set to  the values that we want, and therefore we can bypass those provided constructors that use the `kernel` or do any other unwanted setups.

med-ayssar: Given that each `TargetClass` will only have one mock class (only one `[TargetClassName]Initializer`), thus we will not pollute the header file of the  `TargetClass`  with many **forward-declarared** mocks. 

The `TargetClass` will just use the  `[TargetClassName]Initializer` to initiate its inner state, and we simply can use the instances of the `TargetClass` normally without any restrictions.

med-ayssar:
@nicolossus, I assume, with the word **creator** in the description you mean the **constructor**.  Could you maybe adjust that?

nestkernel/node_collection.h
```diff
@@ -218,6 +218,19 @@ class NodeCollection
    */
   static NodeCollectionPTR create( const std::vector< index >& node_ids );
 
+  /**
+   * @brief Create a NodeCollection from an array of node IDs without requiring
+   * instantiation of a node model and associated managers. Intended to simplify
+   * C++ test programs where just a NodeCollection is needed. Since the fingerprint
```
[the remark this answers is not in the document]
nicolossus: The wording might be a bit unclear, but by "fingerprint" I mean the unique identity of the kernel that creates the `NodeCollection`. A kernel instance is needed in order to set the fingerprint in the `NodeCollection` superclass [constructor](https://github.com/nest/nest-simulator/blob/master/nestkernel/node_collection.cpp#L142).

nestkernel/node_collection.cpp
med-ayssar: ## Example
in `nestkernel/node_collection.h`
```
namespace nest {
class NodeCollectionPrimitiveInitializer;
class NodeCollectionPrimitive 
  {
       public:
       /....../
          NodeCollectionPrimitive(const NodeCollectionPrimitiveInitializer& mock);
  };
}
```
in `cpptests/node_collection_mock.h`
```
#ifndef TEST_H
#define TEST_H
namespace nest {
     class NodeCollectionPrimitiveInitializer 
     {
           //store the needed information here
     };
}
nest::NodeCollectionPrimitive:: NodeCollectionPrimitive(const NodeCollectionPrimitiveInitializer& mock)
{
       //retreive information from mock object.
       first = mock.get_first();
       last = mock.get_last();

}
#endif
```

hakonsbm:
> Could you just remind me where that is implemented?

I think it's checked when getting the model:
https://github.com/nest/nest-simulator/blob/8c386cec300b3369cfc798c59eee7e33f8cca16e/nestkernel/node_collection.cpp#L235
https://github.com/nest/nest-simulator/blob/8c386cec300b3369cfc798c59eee7e33f8cca16e/nestkernel/modelrange_manager.cpp#L81-L86

nicolossus:
> @nicolossus, I assume, with the word **creator** in the description you mean the **constructor**. Could you maybe adjust that?

"Creator" is a reference to the `create_test_collection` function. In a C++ unit test, the user only needs to call this function in order to create a NodeCollection. Constructors are handled internally within the function, so I think the usage of "creator" might be fine here.

med-ayssar:
> > @nicolossus, I assume, with the word **creator** in the description you mean the **constructor**. Could you maybe adjust that?
> 
> "Creator" is a reference to the `create_test_collection` function. In a C++ unit test, the user only needs to call this function in order to create a NodeCollection. Constructors are handled internally within the function, so I think the usage of "creator" might be fine here.

Maybe then a **factory function** would be better in the description.

nestkernel/node_collection.cpp
hakonsbm: > One thing I am not entirely happy with in 3. is the need to introduce the additional `...Base` class just to make mocking possible. I experimented a bit to see if one could have the dependency on `kernel()` in a base class and the define it away in a derived class (so that the ...`Mocker` class could derive directly from `NodeCollectionPrimitive`), but since class base constructors never call derived-class methods, I haven't found a way to achieve that.

I think this is the cleanest solution. We don't have to call derived-class methods if we bypass the `create` functions and directly create the NodeCollection in the test.

nestkernel/node_collection.cpp
```diff
@@ -208,6 +208,56 @@ NodeCollection::create( const std::vector< index >& node_ids_vector )
...
+    {
+      // store Primitive; node goes in new Primitive
+      parts.emplace_back( current_first, current_last, invalid_index, true );
+      current_first = *node_id;
+      current_last = current_first;
+    }
+  }
+
+  // now push last section we opened
+  parts.emplace_back( current_first, current_last, invalid_index, true );
```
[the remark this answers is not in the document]
heplesser: @med-ayssar This is a neat idea. But I am sceptical about performance for several reasons: We need to perform `std::unique()`, which performs a full loop through the input data and actually modifies the input data if it contains duplicates. That should not happen: if we encounter an error condition, we should leave things unchanged. We also need to make updates to map on every loop iteration and later need another loop through the resulting `map` to create the actual node collections. So I would keep the code as it is.

nicolossus:
After an in-house discussion, we decided that the changes required to enable C++ unit testing of NodeCollection are too excessive for the simple getter functionality introduced by this PR. The C++ tests and changes made to accommodate the NodeCollection class for testing have therefore been removed. The PR now only introduces a getter for the NodeCollection step.

heplesser (approved):

heplesser:
Minimal change and full test failing due to technical problem on Github. Therefore merging in spite of failing test.

Issue #2512, opened 2022-10-24
Repeating creating nodes of same structure causes growing memory consumption and creation time


ChenK19:
Hello, I hope to create the same nest structure and adjust some connection weight in the structure for simulation, i accomplish that in a loop, for every iteration, create the nodes and connect them with newly weight, which  shows **growing memory consumption and creation time for nodes**. 

maybe i need some help to solve the problem.

The problem is similar to  the pynest example in 
[https://nest-simulator.readthedocs.io/en/v3.3/auto_examples/brunel_alpha_evolution_strategies.html](url)

replace the 'N' with 2000 in simulation_parameters and 'max_generation' to 100 in optimization_parameters at the start of main function

`    simulation_parameters = {
        'seed': 123,
        'dt': 0.1,            # (ms) simulation resolution
        'sim_time': 1000.,    # (ms) simulation duration
        'warmup_time': 300.,  # (ms) duration ignored during analysis
        'delay': 1.5,         # (ms) synaptic delay
        'g': None,            # relative ratio of excitation and inhibition
        'eta': None,          # relative strength of external drive
        'epsilon': 0.1,       # average connectivity of network
        'N': 400,             # number of neurons in network
        'gamma': 0.8,         # relative size of excitatory and
                              # inhibitory population
        'N_rec': 40,          # number of neurons to record activity from
    }

    optimization_parameters = {
        'verbosity': 1,             # print progress over generations
        'max_generations': 20,      # maximal number of generations
        'target_rate': 1.89,        # (spikes/s) target rate
        'target_corr': 0.0,         # target correlation
        'target_cv': 1.,            # target coefficient of variation
        'mu': [1., 3.],             # initial mean for search distribution
                                    # (mu(g), mu(eta))
        'sigma': [0.15, 0.05],      # initial sigma for search
                                    # distribution (sigma(g), sigma(eta))

        # hyperparameters of the fitness function; these are used to
        # compensate for the different typical scales of the
        # individual measures, rate ~ O(1), cv ~ (0.1), corr ~ O(0.01)
        'fitness_weight_rate': 1.,    # relative weight of rate deviation
        'fitness_weight_cv': 10.,     # relative weight of cv deviation
        'fitness_weight_corr': 100.,  # relative weight of corr deviation
    }`

my environment is:
 - OS: Ubuntu 22.04
 - Python-Version: 3.10.5
 - NEST-Version: 3.3
 - Installation: conda without MPI

heplesser:
[...]

PR #2520, opened 2022-11-02
Fix memory leak with `ResetKernel`

### LINKED ISSUE #2519: Repeat simulation with memory leakage
Dear Nest Developers team:
I am trying to simulate the same nest model structure with different connection weight(for some local connections) repeatly, and i meet some problems for memory leakage in creation model even with nest.ResetKernel().

Then i test a easy demo like:
```
import nest
import psutil

for i in range(1000):
    nest.ResetKernel()
    epop1_num = 1000
    epop1 = nest.Create('iaf_psc_alpha', epop1_num)
    ipop1_num = 1000
    ipop1 = nest.Create('iaf_psc_alpha', ipop1_num)
    nest.Connect(epop1, ipop1)
    p = psutil.virtual_memory()
    print('memory used:', float(p.used)/1024**3)
```

which causes increasing memory usage for 1.4M in every iteration like:

4.692104339599609
4.693546295166016
4.694988250732422
4.696430206298828
4.697872161865234
4.699314117431641
4.700756072998047
4.701717376708984

And I noticed in your nest pynest API documentation about nest.ResetKernel, you metion that 

>  The only exception is that dynamically loaded modules are not unloaded. This may change in a future version of NEST.

I wonder if there is any way to execute such loop of simulations, I hope to change the connection weight and repeat simulation for many times.  

I am really apreciate for your any reply.
Thanks!

Ken

hakonsbm:
Nodes are created with `new` in `GenericModel`:
https://github.com/nest/nest-simulator/blob/d7ef15baf1ce6b4453a10a0b9708b4e9477abc56/nestkernel/genericmodel.h#L154-L158

Then the node pointers are stored in `memory_` in `Model`. However the nodes are never deleted, which leads to a memory leak when calling `ResetKernel`. With only a few thousand nodes, calling `ResetKernel` multiple times can add up to hundreds of MB of leaked memory.

This PR puts Node pointers in smart pointers within `memory_`, so that the Nodes are automatically deleted when the model is destroyed. The explicit call to the node destructor in `NodeManager` is removed because it interferes with the deletion of the nodes at the end of `Model`s lifetime. This fixes #2519. @ChenK19 Can you check if #2512 is fixed by this as well?

### DIFF
--- a/nestkernel/model.h
+++ b/nestkernel/model.h
@@ -237,7 +237,7 @@ private:
   /**
    * Memory for all nodes sorted by threads.
    */
-  std::vector< std::vector< Node* > > memory_;
+  std::vector< std::vector< std::shared_ptr< Node > > > memory_;
 };
 
 
@@ -246,7 +246,7 @@ Model::create( thread t )
 {
   assert( ( size_t ) t < memory_.size() );
   Node* n = create_();
-  memory_[ t ].push_back( n );
+  memory_[ t ].emplace_back( n );
   return n;
 }
 
--- a/nestkernel/model.h
+++ b/nestkernel/model.h
@@ -237,7 +237,7 @@ private:
   /**
    * Memory for all nodes sorted by threads.
    */
-  std::vector< std::vector< std::shared_ptr< Node > > > memory_;
+  std::vector< std::vector< Node* > > memory_;
 };
 
 
--- a/nestkernel/node_manager.cpp
+++ b/nestkernel/node_manager.cpp
@@ -568,8 +568,12 @@ NodeManager::destruct_nodes_()
 {
 #pragma omp parallel
   {
-    const index t = kernel().vp_manager.get_thread_id();
-    local_nodes_[ t ].clear();
+    const index tid = kernel().vp_manager.get_thread_id();
+    for ( auto node : local_nodes_[ tid ] )
+    {
+      delete node.get_node();
+    }
+    local_nodes_[ tid ].clear();
   } // omp parallel
 }
 
--- a/nestkernel/node_manager.cpp
+++ b/nestkernel/node_manager.cpp
@@ -568,17 +568,7 @@ NodeManager::destruct_nodes_()
 {
 #pragma omp parallel
   {
-    index t = kernel().vp_manager.get_thread_id();
-    SparseNodeArray::const_iterator n;
-    for ( n = local_nodes_[ t ].begin(); n != local_nodes_[ t ].end(); ++n )
-    {
-      // We call the destructor for each node excplicitly. This
-      // destroys the objects without releasing their memory. Since
-      // the Memory is owned by the Model objects, we must not call
-      // delete on the Node objects!
-      n->get_node()->~Node();
-    }
-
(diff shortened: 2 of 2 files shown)


nicolossus (approved):
LGTM!

heplesser:
Is there any performance effect of replacing a normal with a smart pointer?

hakonsbm:
@heplesser `Create` with this branch appears to be initially slightly slower (~4 % for 1e6 nodes), but faster after the first `ResetKernel` (~55 % s for 1e6 nodes).

heplesser:
> @heplesser `Create` with this branch appears to be initially slightly slower (~4 % for 1e6 nodes), but faster after the first `ResetKernel` (~55 % s for 1e6 nodes).

Are there any effects on simulation time? I assume that smart pointers take more memory (reference counting needs to happen somewhere) and thus worry that they are cached less effectively, so that neuron access during spike delivery might become less efficient. And since neurons are created in a single place and deleted in a single place, I don't think we really need smart pointers.

I think in #2316 we just overlooked the need for destruction when we went away from the memory `pool`.  In the longer run, we should keep [this comment by @jougs](https://github.com/nest/nest-simulator/pull/2316#pullrequestreview-901821003) in mind.

nestkernel/model.h
```diff
@@ -237,7 +237,7 @@ class Model
   /**
    * Memory for all nodes sorted by threads.
    */
-  std::vector< std::vector< Node* > > memory_;
```
heplesser, line 240: I would suggest to see if we can stick to a plain `Node*` here.

nestkernel/node_manager.cpp
```diff
@@ -568,17 +568,7 @@ NodeManager::destruct_nodes_()
...
-    index t = kernel().vp_manager.get_thread_id();
-    SparseNodeArray::const_iterator n;
-    for ( n = local_nodes_[ t ].begin(); n != local_nodes_[ t ].end(); ++n )
-    {
-      // We call the destructor for each node excplicitly. This
-      // destroys the objects without releasing their memory. Since
-      // the Memory is owned by the Model objects, we must not call
-      // delete on the Node objects!
-      n->get_node()->~Node();
-    }
```
heplesser, line 580: Here, we should delete nodes explicitly:
```c++
const auto tid = kernel().vp_manager.get_thread_id();
for ( auto node : local_nodes_[ tid ] )
{
  delete node;
}
local_nodes_[ tid ].clear();

heplesser (changes requested):
I think we can solve this without smart pointers and the performance and memory cost they imply.

heplesser:
I just did a little test, comparing a program filling a vector with 2^10 `*int` vs `std::shared_ptr<int>`, using g++12 and clang++14. In both cases, the shared pointer takes very close to 40B extra memory per vector element compared to a plain pointer. 

For more on the implementation in the gcc case, see https://gcc.gnu.org/onlinedocs/libstdc++/manual/memory.html#std.util.memory.shared_ptr.

heplesser (approved):

PR #2523, opened 2022-11-10
Remove superfluous css and js bootstrap files

jessica-mitchell:
This PR removes all but one bootstrap file (bootstrap.min)  from css and js folders.

Output does not seem affected, and cleans up the static folder from excess files.

See output: https://nest-test.readthedocs.io/en/cleanup-css/index.html


steffengraber (approved):
After the latest changes, 404 errors no longer occur.  Thanks for the detective work.

PR #2525, opened 2022-11-14
Fix missing colon and correct comment

JanVogelsang:
There was a missing colon in [models/iaf_cond_alpha_mc.h](https://github.com/nest/nest-simulator/compare/master...JanVogelsang:nest-simulator:comment-fixes?expand=1#diff-71d7ffa760a39bbef5e46bc776b2e5604429c0f79444615fae47e75fd9f521c2) which caused a reference to not be displayed correctly.

A comment in [nestkernel/connection.h](https://github.com/nest/nest-simulator/compare/master...JanVogelsang:nest-simulator:comment-fixes?expand=1#diff-7a75dd629d0cf1fd4cf7910456c209d29267aeec9955b87ac38abc6de527e55b) stated incorrect bit-sizes, probably an artifact of older versions.

### DIFF
--- a/models/iaf_cond_alpha_mc.h
+++ b/models/iaf_cond_alpha_mc.h
@@ -77,7 +77,7 @@ FULLY TESTED. USE AT YOUR OWN PERIL!
 
 ``iaf_cond_alpha_mc`` is an implementation of a multi-compartment spiking
 neuron using IAF dynamics with conductance-based synapses. It serves
-mainly to illustrate the implementation of ref:`multicompartment models
+mainly to illustrate the implementation of :ref:`multicompartment models
 <multicompartment-models>` in NEST.
 
 The model has three compartments: soma, proximal and distal dendrite,
--- a/nestkernel/connection.h
+++ b/nestkernel/connection.h
@@ -306,7 +306,7 @@ protected:
      connection
   */
   targetidentifierT target_;
-  //! syn_id (char) and delay (24 bit) in timesteps of this connection
+  //! syn_id (9 bit), delay (21 bit) in timesteps of this connection and more_targets and disabled flags (each 1 bit)
   SynIdDelay syn_id_delay_;
 };
 


jessica-mitchell (approved):
@JanVogelsang I can only comment on the change to the `iaf_cond_alpha.h`, and that looks good.

Someone who knows the C++  should check the other line @jougs @heplesser ?

heplesser (approved):

heplesser:
@JanVogelsang Thank you for this PR. I will merge it now but would like to ask you to further improve the doxygen comments on data members in `Connection`:

https://github.com/nest/nest-simulator/blob/b692c32a01394f29bf834bf1fa28784f1b115a31/nestkernel/connection.h#L300-L310

The comment on `syn_id_delay_` is partly redundant with the non-doxygen comment before `target_`. Could comments on the two members be combined into a doxygen group? I also think that it is not sensible to give field sizes in the comments, given that depend on compiler flags and could change over time, see

https://github.com/nest/nest-simulator/blob/b692c32a01394f29bf834bf1fa28784f1b115a31/nestkernel/nest_types.h#L84

and 

https://github.com/nest/nest-simulator/blob/b692c32a01394f29bf834bf1fa28784f1b115a31/nestkernel/nest_types.h#L88

JanVogelsang:
Fair point, I will look into this.

Didn't know the bitmasks were different for HPC and non-HPC, is that noted in the user documentation somewhere as well?
A user should definitely know that a model they run on their local system is not guaranteed to run on an HPC system as well, which is a questionable decision anyway. Shouldn't the typical workflow be to test a model on your local system with lower scale and then run the same model on a cluster by just scaling up the network size? And wouldn't you usually use even more different synapse types on an HPC system than you would on a lower scale one? I understand that the maximum number of threads or ranks will be limited on a non-HPC system, therefore reducing the size of the bitmasks won't hurt, but reducing the number of different synapse types might actually hurt.
Although I understand, that there is no other choice basically and I should probably see it the other way around. The number of different synapse types is _not reduced_ on HPC systems, it is instead _increased_ on non-HPC ones, because why not, there are free bits to spend anyway.

PR #2526, opened 2022-11-14
Add note about resetting behavior of local_spike_counter

### LINKED ISSUE #2465: Document local_spike_counter resetting behavior
**Describe the flaw**

`local_spike_counter` will reset in repeated calls to simulate(), which is not documented
The assumed behavior is that it would reset after resetting the kernel.

**Suggest an improvement**
Add text to describe this behavior in PyNEST API.

@JoseJVS let me know if this isn't quite correct

jessica-mitchell:
Resolves #2465 

This PR adds a note to the local_spike_counter and to Run() regarding the resetting behavior.


pynest/nest/__init__.py
```diff
@@ -373,7 +373,7 @@ def __dir__(self):
             "Number of spikes fired by neurons on a given MPI rank during the"
             + " most recent call to :py:func:`.Simulate`. Only spikes from"
             + " \"normal\" neurons are counted, not spikes generated by devices"
-            + " such as ``poisson_generator``"
+            + " such as ``poisson_generator``. Resets on each call to ``Run``."
```
heplesser, line 376:
```suggestion
            + " such as ``poisson_generator``. Resets on each call to ``Simulate`` or ``Run``."
```

heplesser (changes requested):
Looks good, just one small suggestion.

heplesser:
@josejvs Could you be second reviewer on this one?

jessica-mitchell:
> @JoseJVS Could you be second reviewer on this one?

Yes, I wanted to add him but he was not on the list of reviewers that I could pick from.

heplesser (approved):

JoseJVS:
I feel that saying "Resets on each call to ``Simulate`` or ``Run``." is misleading in the sense that Simulate implies Run whereas "Simulate or Run" might give the impression that they may be disjoint.
Although I agree that this makes it clearer for people who don't know that Simulate also executes Run, it might make it confusing for people who do know.
One possible solution would be to have "Resets on each call to  ``Run`` (also called upon by ``Simulate``)." , of course this might not be the best solution but to me this way we avoid the previously stated confusion and those who did not know the relationship between Simulate and Run can learn it by chance.
Of course this is purely from a semantics point of view and in itself the proposed statement is already enough for the objective of this PR.

jessica-mitchell:
> I feel that saying "Resets on each call to `Simulate` or `Run`." is misleading in the sense that Simulate implies Run whereas "Simulate or Run" might give the impression that they may be disjoint. Although I agree that this makes it clearer for people who don't know that Simulate also executes Run, it might make it confusing for people who do know. One possible solution would be to have "Resets on each call to `Run` (also called upon by `Simulate`)." 

I think "called upon" is a bit confusing.  
I assume the intention here is if you write `simulate()` or `run()` in your script both have the affect that local_spike_counter resets. That's how I would interpret this, is that correct? In the end, for a user it doesnt matter if `simulate` triggers `run` so long as they understand the outcome. For developers, if its not clear how these commands operate, then we should consider expanding the documentation on this topic.

JoseJVS:
I see perhaps just "called by" is enough?
In any case, I was just giving my opinion from a semantics point of view but "Simulate or Run" is enough for the information we want to convey in this PR.

jessica-mitchell:
I think I'd rather be more explicit, and in the Run function mention that it is called during simulate but in a separate sentence from the one with local_spike_counter. I think this line 
```
Simulate(t): t' = t/m; Prepare(); for _ in range(m): Run(t'); Cleanup() 
```
is trying to do that. Although I would like to see some text around that.  I could create another PR, just for that.

jessica-mitchell:
@JoseJVS I just added you as a reviewer, I didn't see your handle on the list yesterday, maybe I was blind. If you are happy with this current PR can you please approve thanks

JoseJVS:
As long as the fact that Run is called by Simulate, is properly documented elsewhere I think everything is good.
Also you were not blind yesterday, I was just not a member in the nest initiative group in github but HEP sent me an invite and now I can properly be a reviewer.

JoseJVS (approved):

PR #2527, opened 2022-11-15
Fix output of nest.spatial in Read the Docs

### LINKED ISSUE #2513: Missing spatial user documentation
In both the [3.3 release](https://nest-simulator.readthedocs.io/en/v3.3/ref_material/pynest_apis.html#module-nest.spatial_distributions.hl_api_spatial_distributions) and [latest](https://nest-simulator.readthedocs.io/en/latest/ref_material/pynest_apis.html#module-nest.spatial_distributions.hl_api_spatial_distributions) (as of writing) documentation on `Functions related to spatial distributions`, the documentation from [pynest/nest/spatial/hl_api_spatial.py](https://github.com/nest/nest-simulator/blob/master/pynest/nest/spatial/hl_api_spatial.py) is missing. The module is listed in [pynest_apis.rst](https://github.com/nest/nest-simulator/blob/master/doc/htmldoc/ref_material/pynest_apis.rst#functions-related-to-spatial-distributions), so the origin of the bug is not immediately obvious to me.

Note that the missing documentation also means that the references to `spatial.grid` and `spatial.free` in [`Create`](https://nest-simulator.readthedocs.io/en/latest/ref_material/pynest_apis.html#nest.lib.hl_api_nodes.Create) do not lead anywhere.

med-ayssar:
Fixes #2513

### DIFF
--- a/pynest/nest/spatial/hl_api_spatial.py
+++ b/pynest/nest/spatial/hl_api_spatial.py
@@ -46,8 +46,11 @@ class DistanceParameter(Parameter):
     be used in contexts with two nodes, e.g. when connecting.
     """
 
+    __distanceParam = None
+
     def __init__(self):
         distance_parameter = CreateParameter('distance', {})
+
         super().__init__(distance_parameter._datum)
 
     @property
@@ -82,8 +85,17 @@ class DistanceParameter(Parameter):
         """
         return CreateParameter('distance', {'dimension': dimension})
 
+    @staticmethod
+    def getInstance():
+        if DistanceParameter.__distanceParam:
+            return DistanceParameter.__distanceParam
+        else:
+            DistanceParameter.__distanceParam = DistanceParameter()
+            return DistanceParameter.__distanceParam
 
-distance = DistanceParameter()
+
+def distance():
+    return DistanceParameter.getInstance()
 
 
 class pos:
--- a/pynest/nest/spatial/hl_api_spatial.py
+++ b/pynest/nest/spatial/hl_api_spatial.py
@@ -91,6 +91,7 @@ class DistanceParameter(Parameter):
 def __getattr__(name):
     if name == "distance":
         return DistanceParameter()
+    raise AttributeError(f"module {__name__} has no attribute {name}")
 
 
 class pos:
--- a/pynest/nest/spatial/hl_api_spatial.py
+++ b/pynest/nest/spatial/hl_api_spatial.py
@@ -22,6 +22,7 @@
 import numpy as np
 from ..lib.hl_api_types import CreateParameter, Parameter
 from ..ll_api import sli_func
+import functools
 
 __all__ = [
     'distance',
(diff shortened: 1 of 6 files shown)


jessica-mitchell:
There is nest.spatial and nest.lib.spatial - and these are separate sections in Read the Docs API, for "spatially-distributed nodes" and "spatial distributions", which is a bit confusing. 
I want to know if we can either move lib.spatial into spatial or rename it. And clarify any relevant differences in the docs or simply combine them if that makes sense.

pynest/nest/spatial/hl_api_spatial.py
```diff
@@ -46,6 +46,8 @@ class DistanceParameter(Parameter):
     be used in contexts with two nodes, e.g. when connecting.
     """
 
+    __distanceParam = None
```
hakonsbm, line 49: I think we only need a single underscore here instead of the double one.

pynest/nest/spatial/hl_api_spatial.py
```diff
@@ -82,8 +84,17 @@ def n(dimension):
         """
         return CreateParameter('distance', {'dimension': dimension})
 
+    @staticmethod
+    def getInstance():
```
hakonsbm, line 88: Should be renamed to `get_instance`. Also add a docstring.

pynest/nest/spatial/hl_api_spatial.py
```diff
@@ -82,8 +84,17 @@ def n(dimension):
...
+        if DistanceParameter.__distanceParam:
+            return DistanceParameter.__distanceParam
+        else:
+            DistanceParameter.__distanceParam = DistanceParameter()
+            return DistanceParameter.__distanceParam
+
 
-distance = DistanceParameter()
+def distance():
```
hakonsbm, line 96: Could probably add a docstring to `distance` as well.

hakonsbm (changes requested):
I have just a couple suggestions for the code.

pynest/nest/spatial/hl_api_spatial.py
med-ayssar: I have tested it with double underscores, and it makes the `__distanceParam` not accessible from outside the class.

pynest/nest/spatial/hl_api_spatial.py
```diff
@@ -83,7 +87,11 @@ def n(dimension):
         return CreateParameter('distance', {'dimension': dimension})
 
 
-distance = DistanceParameter()
+@functools.cache
+def __getattr__(name):
+    if name == "distance":
+        return DistanceParameter()
+    raise AttributeError(f"module {__name__} has no attribute {name}")
```
Helveg, line 94: This `__getattr__` call should be moved up one module to `nest.spatial`, otherwise the import in `nest.spatial` triggers this import at import-time again. I propose this:

### In `nest/spatial/__init__.py`

```python
from .hl_api_spatial import _distance_singleton

__all__ = [..., "distance"]

def __getattr__(name):
  if name == "distance":
    return _distance_singleton()
```

### In `hl_api_spatial.py`

```python
@functools.cache
def _distance_singleton():
  return DistanceParameter()
```

pynest/nest/spatial/hl_api_spatial.py
```diff
@@ -33,19 +34,22 @@
...
-    Alternatively the distance in a single dimension may be chosen. Three
-    properties are defined, x, y, and z, which represent the distance in
-    their respective dimensions. Note that the distance parameter can only
-    be used in contexts with two nodes, e.g. when connecting.
-    """
+Alternatively the distance in a single dimension may be chosen. Three
+properties are defined, x, y, and z, which represent the distance in
+their respective dimensions. Note that the distance parameter can only
+be used in contexts with two nodes, e.g. when connecting.
+"""
```
Helveg, line 49: See my comment below, this ought to be moved to `nest/spatial/__init__.py`

Helveg (changes requested):
We should avoid that `distance` is created at import time, which we can do by moving this logic to `nest.spatial`

Helveg:
> There is nest.spatial and nest.lib.spatial - and these are separate sections in Read the Docs API, for "spatially-distributed nodes" and "spatial distributions", which is a bit confusing. I want to know if we can either move lib.spatial into spatial or rename it. And clarify any relevant differences in the docs or simply combine them if that makes sense.

I'm also highly in favor of merging the 2 modules! It's confusing.

pynest/nest/spatial/hl_api_spatial.py
Helveg: docstring has been added

pynest/nest/spatial/hl_api_spatial.py
Helveg: name wrangling with double underscore isn't usually recommended, unless you have a complex hierarchy where you know each class in the hierarchy will need to lookup their own possibly overlapping class attributes.

> and it makes the __distanceParam not accessible from outside the class.

This is not true, and why it's not recommended: access modifiers don't exist in Python, your attribute is still accessible as `f"_{self.__class__.__name__}_distanceParam"` ;p

Helveg (approved):
Looks good. `nest.spatial.distance` works, import time code is avoided and the docs render and reference to distance just fine.

hakonsbm (approved):

PR #2531, opened 2022-11-23
Make CMakeLists install-nodoc cmd ninja compatible

marcelkrueger:
Hello everyone,

https://github.com/nest/nest-simulator/pull/2396 introduced a regression by limiting cmake generation to make-based build systems by using the Makefile specific ` $(MAKE)` variable. 
 ` $(MAKE)` is not substituted by cmake, instead it is only substitute by make-based system when encountering it in the makefile.
Therefore, cmake can only generate correct files for Makefile based build systems as ninja for example does not know how to handle ` $(MAKE)` in ninja.build files.
The breakage w.r.t. ninja was also mentioned by the PR author @Helveg.

This is a quick fix that allows using ninja again as a build system. 
While this is not an optimal and permanent solution it removes the regression introduced by https://github.com/nest/nest-simulator/pull/2396 with minimal changes while https://github.com/nest/nest-simulator/issues/1905 is still discussed.

Best,
Marcel

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -368,8 +368,14 @@ install( FILES LICENSE README.md
     DESTINATION ${CMAKE_INSTALL_DOCDIR}
     )
 
-add_custom_target( install-nodoc
-  COMMAND $(MAKE) NEST_INSTALL_NODOC=true install
-)
+if(${CMAKE_GENERATOR} STREQUAL "Ninja")
+  add_custom_target( install-nodoc
+    COMMAND ninja NEST_INSTALL_NODOC=true install
+    )
+else()
+  add_custom_target( install-nodoc
+    COMMAND $(MAKE) NEST_INSTALL_NODOC=true install
+    )
+endif()
 
 nest_print_config_summary()


Helveg:
I didn't know this! Thanks for setting this right. So then there's no way for something like `make install-nodoc` to work properly?

marcelkrueger:
At least not build system independent as far as me and my colleagues know. 

A probably more idiomatic way to solve it would be to either use cmake flags as proposed in  https://github.com/nest/nest-simulator/issues/1905 or document that user should set the env variable themselves, e.g.: 
`NEST_INSTALL_NODOC=true make install` 
Either way the check can take place in the cmake script and calling the generator scripts can be skipped as from a quick glance the python script doesn't do anything if the ENV var is set besides printing which can be done via cmake as well.

Disclaimer: I am not deeply familiar with the whole build system and just wanted to restore ninja compatibility with this quickfix.

Helveg:
I don't know how many build system NEST supports, or how many there are, but would it make sense to invert this check so that exclusively `make` uses the `$(MAKE)` variable?

jougs:
@marcelkrueger: can you please resolve the conflicts?

Also: #1905 has been closed recently, so this issue can't be solved properly in that context. What's your feeling here? I think I'd actually be fine with the conditional forever.

marcelkrueger:
Looks like https://github.com/nest/nest-simulator/pull/2544/ completely removed the section making trouble w.r.t. to ninja. 
Therefore, the conditional is not necessary anymore.

I just tested the current master (ebdc604eb623714607414b754b9dcdc3f5af14d8) and was able to use ninja again without any modifications to the CMakeLists.txt :+1:

PR #2533, opened 2022-11-25
Abort NEST if run using mpirun but compiled without support for MPI

### LINKED ISSUE #2381: Prevent NEST from running under mpirun-control if not built with MPI

jougs:
This fixes #2381.

### DIFF
--- a/nestkernel/mpi_manager.cpp
+++ b/nestkernel/mpi_manager.cpp
@@ -183,15 +183,15 @@ void
 nest::MPIManager::initialize()
 {
 #ifndef HAVE_MPI
-  char* pmix_rank_set = std::getenv( "PMIX_RANK" );  // set by OpenMPI's launcher
-  char* pmi_rank_set = std::getenv( "PMI_RANK" );    // set by MPICH's launcher
+  char* pmix_rank_set = std::getenv( "PMIX_RANK" ); // set by OpenMPI's launcher
+  char* pmi_rank_set = std::getenv( "PMI_RANK" );   // set by MPICH's launcher
   if ( pmix_rank_set or pmi_rank_set )
   {
     LOG( M_FATAL,
       "MPIManager::initialize()",
       "You seem to be using NEST via an MPI launcher like mpirun, mpiexec or srun "
       "although NEST was not compiled with MPI support. Please see the NEST "
-      "documentation about parallel and distributed computing. Exiting.");
+      "documentation about parallel and distributed computing. Exiting." );
     std::exit( 127 );
   }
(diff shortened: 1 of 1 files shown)


jougs:
Should this be mentioned somewhere in the documentation, or do you consider this just an ordinary safeguard that goes without saying?

nestkernel/mpi_manager.cpp
```diff
@@ -180,6 +182,19 @@ nest::MPIManager::init_mpi( int* argc, char** argv[] )
...
 nest::MPIManager::initialize()
 {
+#ifndef HAVE_MPI
+  char* pmix_rank_set = std::getenv( "PMIX_RANK" );  // set by OpenMPI's launcher
+  char* pmi_rank_set = std::getenv( "PMI_RANK" );    // set by MPICH's launcher
+  if ( pmix_rank_set or pmi_rank_set )
+  {
+    LOG( M_FATAL,
+      "MPIManager::initialize()",
+      "You seem to be using NEST via an MPI launcher like mpirun or mpiexec, "
```
heplesser, line 192:
```suggestion
      "You seem to be using NEST via an MPI launcher like mpirun, mpiexec or srun "
```

heplesser (commented):
Since many will be using NEST via Slurm, I think it makes sense to mention `srun` as well. I don't think we need to add more information to the documentation, since this is a corner case and the error message very informative.

But why did the "Full" test fail? I restarted the test just in case it was a random system problem.

jougs:
> But why did the "Full" test fail? I restarted the test just in case it was a random system problem.

I have no clue. It's the same failure as in https://github.com/nest/nest-simulator/actions/runs/3550596187/jobs/5968895331, so I don't think this is related to the code here.

med-ayssar:
As this new testing feature will not be tested by the CI, I think it is important to add an extra unittest for this case and expect it to fail.

heplesser (approved):

PR #2534, opened 2022-11-25
Fix open TODO in music_cont_in_proxy

### LINKED ISSUE #1390: Open TODO in `music_cont_in_proxy.cpp`
During review of #1245 we encountered an open TODO in `music_cont_in_proxy.cpp`. This should most likely be fixed or turned into a helpful comment.

Unfortunately the history stops at the move from svn to git. Maybe @mdjurfeldt @jougs remember what this was about?

jougs:
Fixes #1390

### DIFF
--- a/models/music_cont_in_proxy.cpp
+++ b/models/music_cont_in_proxy.cpp
@@ -67,9 +67,9 @@ nest::music_cont_in_proxy::Parameters_::get( DictionaryDatum& d ) const
 void
 nest::music_cont_in_proxy::Parameters_::set( const DictionaryDatum& d, State_& s, Node* node )
 {
-  if( d->known(names::port_name) and s.published_ )
+  if ( d->known( names::port_name ) and s.published_ )
   {
-    throw MUSICPortAlreadyPublished(node->get_name(), port_name_);
+    throw MUSICPortAlreadyPublished( node->get_name(), port_name_ );
   }
 
   if ( not s.published_ )
@@ -168,11 +168,11 @@ nest::music_cont_in_proxy::get_status( DictionaryDatum& d ) const
 void
 nest::music_cont_in_proxy::set_status( const DictionaryDatum& d )
 {
-  Parameters_ ptmp = P_;    // temporary copy in case of errors
-  ptmp.set( d, S_, this );  // throws if BadProperty
+  Parameters_ ptmp = P_;   // temporary copy in case of errors
+  ptmp.set( d, S_, this ); // throws if BadProperty
 
   State_ stmp = S_;
-  stmp.set( d, P_ );  // throws if BadProperty
+  stmp.set( d, P_ ); // throws if BadProperty
 
   // if we get here, temporaries contain consistent set of properties
   P_ = ptmp;
--- a/models/music_cont_in_proxy.cpp
+++ b/models/music_cont_in_proxy.cpp
@@ -65,11 +65,12 @@ nest::music_cont_in_proxy::Parameters_::get( DictionaryDatum& d ) const
 }
 
 void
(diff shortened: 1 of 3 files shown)


models/music_cont_in_proxy.cpp
```diff
@@ -167,11 +168,11 @@ nest::music_cont_in_proxy::get_status( DictionaryDatum& d ) const
 void
 nest::music_cont_in_proxy::set_status( const DictionaryDatum& d )
 {
-  Parameters_ ptmp = P_; // temporary copy in case of errors
-  ptmp.set( d, S_ );     // throws if BadProperty
+  Parameters_ ptmp = P_;    // temporary copy in case of errors
```
med-ayssar, line 171: Why do we need a temp attribute here? If something went wrong in the `set` function, then an exception will be thrown, and we end the execution.

jougs: But then we might already have set half of the new parameters on P_, and the user has now way to know which worked and which did not.

This same pattern is used in every model:
1. create a temporary copy
1. check if setting on the copy works
1. if yes, swap the updated copy with the original parameters.

models/music_cont_in_proxy.cpp
```diff
@@ -65,11 +65,12 @@ nest::music_cont_in_proxy::Parameters_::get( DictionaryDatum& d ) const
 }
 
 void
-nest::music_cont_in_proxy::Parameters_::set( const DictionaryDatum& d, State_& s )
+nest::music_cont_in_proxy::Parameters_::set( const DictionaryDatum& d, State_& s, Node* node )
```
heplesser, line 68:
```suggestion
nest::music_cont_in_proxy::Parameters_::set( const DictionaryDatum& d, State_& s, const Node& node )
```

models/music_cont_in_proxy.cpp
```diff
@@ -65,11 +65,12 @@ nest::music_cont_in_proxy::Parameters_::get( DictionaryDatum& d ) const
...
 {
-  // TODO: This is not possible, as P_ does not know about get_name()
-  //  if(d->known(names::port_name) and s.published_)
-  //    throw MUSICPortAlreadyPublished(get_name(), P_.port_name_);
+  if( d->known(names::port_name) and s.published_ )
+  {
+    throw MUSICPortAlreadyPublished(node->get_name(), port_name_);
```
heplesser, line 72:
```suggestion
    throw MUSICPortAlreadyPublished(node.get_name(), port_name_);
```

models/music_cont_in_proxy.cpp
```diff
@@ -167,11 +168,11 @@ nest::music_cont_in_proxy::get_status( DictionaryDatum& d ) const
 void
 nest::music_cont_in_proxy::set_status( const DictionaryDatum& d )
 {
-  Parameters_ ptmp = P_; // temporary copy in case of errors
-  ptmp.set( d, S_ );     // throws if BadProperty
+  Parameters_ ptmp = P_;    // temporary copy in case of errors
+  ptmp.set( d, S_, this );  // throws if BadProperty
```
heplesser, line 172:
```suggestion
  ptmp.set( d, S_, *this );  // throws if BadProperty
```

heplesser (changes requested):
@jougs Thanks! I added suggestions to use `const&` instead of pointer.

heplesser:
OK! And now the FULL run fails again ...

heplesser (approved):

med-ayssar (approved):

models/music_cont_in_proxy.cpp
jougs: @med-ayssar: can you please resolve this conversation if you are happy with my answer? We can only merge once all conversations are resolved.

PR #2535, opened 2022-11-28
Remove leftover reference to nosetests

clinssen:
``nosetests`` is no longer needed or being used at all by NEST; remove this leftover reference to the package.


heplesser:
Minimal clean-up of CMakeFiles, thus merging without second review.

PR #2538, opened 2022-11-28
Revising comments on Event class

### LINKED ISSUE #129: Review role and use of Events in connection creation and signal transmission
We use `Event`s in NEST both during connection creation and during transmission of signals via connections. `Event`s are used differently in both cases, and different subsets of `Event` public functions cater to the different uses. This is not properly documented at present.

As an example, `e.get_sender().get_gid()` and `e.get_sender_gid()` will yield inconsistent results, even though one would expect them to yield the same result. In particular, `e.get_sender_gid()` will trigger an assertion (`sender_gid_>0` fails) when called from `handle_test_event()`.

The reason for this inconsistency is that Event objects are used differently during network construction and network simulation. 

During network construction, we send a pointer to a node of the sender node type. This is not, generally, a pointer to the actual sender: for senders on remote MPI processes, this pointer is not available. Instead, we send a pointer to a proxy node. No GID is set on that proxy node (proxy nodes always have GID 0). 

`send_test_event()` now sets only the `Node* sender_` field on the Event it sends (to `this`), but not the `sender_gid_`, since that is not consistently available (i.e., not for proxy neurons representing remote sources).

When delivering spikes, on the other hand, only the `sender_gid_` field is set in an Event, but not the `sender_`. This because a pointer to sender is not available for remote neurons and passing pointers to proxy neurons would not make sense at this point.

As a further twist, if the target is a recording device, connections will only be created from local sources (the connect mechanism takes care of this), whence the pointer returned by `e.get_sender()` will actually be a pointer to the real sender object (not a proxy), and `e.get_sender().get_gid()` will return the proper GID. `e.get_sender_gid()`, on the other hand, should be used only during spike delivery, not during connection creation.

heplesser:
This PR provides updated developer documentation for the `Event` class. It thus fixes #129. At a later stage, the `Event` class hierarchy should be reviewed fully.

### DIFF
--- a/nestkernel/event.h
+++ b/nestkernel/event.h
@@ -45,41 +45,59 @@ namespace nest
 class Node;
 
 /**
- * Encapsulates information which is sent between Nodes.
+ * Encapsulate information sent between Nodes.
  *
- * For each type of information there has to be a specialized event
- * class.
+ * Event is the base class for transmitting information between Nodes in NEST,
+ * with different subclasses for transmitting different types of information. Event
+ * types come in three categories
+ * -# SpikeEvent can be transmitted between MPI processes
+ * -# SecondaryEvent subclasses can also be transmitted between MPI processes, need to be transmitted via secondary
+ connections. They can transport data.
+ * -# All other Event subclasses can only be transmitted within an MPI process
  *
- * Events are used for two tasks. During connection, they are used as
- * polymorphic connect objects. During simulation they are used to
- * transport basic event information from one node to the other.
+ * Events are used for two tasks:
+ * -# Creating connections
+ * -# Sending signals between nodes during simulation
  *
- * A connection between two elements is physically established in two
- * steps: First, create an event with the two envolved elements.
- * Second, call the connect method of the event.
+ * ## Events during connection
  *
- * An event object contains only administrative information which is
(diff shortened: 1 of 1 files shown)


nestkernel/event.h
```diff
@@ -45,41 +45,59 @@ namespace nest
 class Node;
 
 /**
- * Encapsulates information which is sent between Nodes.
+ * Encapsulate information sent between Nodes.
```
jessica-mitchell, line 48:
```suggestion
 * Encapsulate information sent between nodes.
```

jessica-mitchell: I think nodes should not be capitalized, even if the class is. It seems to me most text we have regarding nodes is lower case.

jessica-mitchell (changes requested):
@heplesser thanks for looking at this ancient issue; I just have some minor fixes from my side.

med-ayssar:
Looks good to me.

med-ayssar (approved):

heplesser:
@jessica-mitchell Thanks for your suggestions, I have accepted them all.

jessica-mitchell (approved):

heplesser:
Since the test failure is due to causes entirely unrelated to this PR, I merge in spite of one failing test case.

PR #2540, opened 2022-11-28
Improved comments on weight positivity for conductance-based multisynapse models

### LINKED ISSUE #2146: Inconsistency between _cond_ and _cond_multisynapse implementations
I think the problem has been discussed before but I could not find it and I run into this issue just now so I thought I would create a new issue to discuss it: *the implementation of conductance weights is inconsistent between ``_cond_`` and ``_cond_multisynapse`` models.*

This is because simple ``_cond_`` models (e.g. ``iaf_cond_alpha`` or ``aeif_cond_alpha``) expect *signed* conductances to tell apart excitatory and inhibitory spikes.
On the other hand, multisynapse models use the receptor ports to tell spikes apart and they expect *positive* conductances (which seems like the most reasonable option).

The problem is that positivity is not enforced when setting weights with multisynapse models, so one can set negative conductances, but the simulation with crash and burn as soon as a spike is delivered with a negative conductance (kernel throws a uncaught BadProperty, then calls terminate recursively and dies).

So my questions are:
- do we want to make all models to behave the same? (I think it would be better)
- should that wait for the models to be moved to NESTML?
- if so, should we at least check the conductance sign in multisynaptic models in the meantime?

heplesser:
The title says it all. This fixed #2146.

### DIFF
--- a/models/gif_cond_exp_multisynapse.cpp
+++ b/models/gif_cond_exp_multisynapse.cpp
@@ -587,7 +587,7 @@ nest::gif_cond_exp_multisynapse::handle( SpikeEvent& e )
   if ( e.get_weight() < 0 )
   {
     throw BadProperty(
-      "Synaptic weights for conductance based models "
+      "Synaptic weights for conductance-based multisynapse models "
       "must be positive." );
   }
   assert( e.get_delay_steps() > 0 );
--- a/models/gif_cond_exp_multisynapse.cpp
+++ b/models/gif_cond_exp_multisynapse.cpp
@@ -588,7 +588,7 @@ nest::gif_cond_exp_multisynapse::handle( SpikeEvent& e )
   {
     throw BadProperty(
       "Synaptic weights for conductance-based multisynapse models "
-      "must be positive." );
+      "must be non-negative." );
   }
   assert( e.get_delay_steps() > 0 );
   assert( ( e.get_rport() > 0 ) and ( ( size_t ) e.get_rport() <= P_.n_receptors() ) );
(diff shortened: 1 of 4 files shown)


models/gif_cond_exp_multisynapse.h
```diff
@@ -113,28 +113,31 @@ positive or negative):
...
-:math:`\eta_i` and :math:`\gamma_i`. It applies the jump of :math:`\eta_i` and
-:math:`\gamma_i` after the refractory period. One can easily convert between
-:math:`q_{\eta/\gamma}` of these two approaches:
+   In the current implementation of the model,
+   the values of :math:`\eta_i` and :math:`\gamma_i` are affected immediately
+   after spike emission. However, `GIF toolbox <http://wiki.epfl.ch/giftoolbox>`_,
+   which fits the model using experimental data, requires a different set of
+   :math:`\eta_i` and :math:`\gamma_i`. It applies the jump of :math:`\eta_i` and
+   :math:`\gamma_i` after the refractory period. One can easily convert between
+   :math:`q_{\eta/\gamma}` of these two approaches:
```
jessica-mitchell, line 124:
```suggestion
   :math:`q_{\eta/\gamma}` with these two approaches:
```

jessica-mitchell (approved):
@heplesser I approve with only one small comment, thanks

models/gif_cond_exp_multisynapse.cpp
```diff
@@ -587,7 +587,7 @@ nest::gif_cond_exp_multisynapse::handle( SpikeEvent& e )
   if ( e.get_weight() < 0 )
   {
     throw BadProperty(
-      "Synaptic weights for conductance based models "
+      "Synaptic weights for conductance-based multisynapse models "
       "must be positive." );
```
clinssen, line 591:
```suggestion
      "must be non-negative." );
```

clinssen (approved):
Just a small suggestion. Cheers!

PR #2542, opened 2022-11-29
Fix rendering of references in local build of documentation

jessica-mitchell:
This PR fixes the rendering issue in Sphinx build where the bibliography section is set to float right.
The offending lines in the CSS are removed. This seemed to be only an issue in the local build and not on Read the Docs.


jougs (approved):
LGTM!

PR #2545, opened 2022-11-30
Remove VERA++ static code checking

heplesser:
This PR removes static code checking with VERA++ since that tool is no longer maintained.


gtrensch (approved):
Looks good to me! :+1:

PR #2549, opened 2022-11-30
Add the ability to patch the documentation on Read the Docs 

steffengraber:
This fix enables the patching of documentation on readthedocs. The patch name must contain the hash of the corresponding release. The location where the patches are stored is stored as an environment variable in the admin console of Readthedocs and can thus be adjusted at any time without having to change the release again.


jougs:
Without having looked at the code changes in detail: could you please add comments about the logic and (security) implications of this? Also, some words about the motivation and possible use cases would be nice.

steffengraber:
Sorry, you are absolutely right. Here are a few comments on the changes.

Motivation 

Every NEST software release also includes documentation, which means that even small changes such as fixing spelling mistakes, broken links or similar are no longer possible or difficult to implement. 
For our online documentation on Readthedocs, we have therefore looked for a way to apply patches without having to create a new software release. 

Method

A patch file is stored on a server that can be accessed online. The name of the patch contains the hash of the release, so that an exact allocation between patch and release is possible. The path to the patch is defined via an environment variable in Readthedocs. In order to apply the patch, only a new build of the documentation for the release in question needs to be triggered.

Implications

The patch is only intended for the documentation online. The source code of NEST is not changed. All changes can be undone at any time; only a new build of the release documentation without patch is required.
To make changes to the documentation, a user must have access to the administration of the NEST readthedocs repository. Only there, the required settings can be done.

jougs:
Many thanks for the nice explanations. Can you please also add those (possibly in a condensed form) to the code, so they can easily be found when looking at it? Thanks!

doc/htmldoc/conf.py
```diff
@@ -25,6 +25,7 @@
 import re
 import pip
 import subprocess
```
jougs, line 27: Is this still needed with the new line 34 in place?

doc/htmldoc/conf.py
```diff
@@ -331,9 +332,41 @@ def copy_acknowledgments_file(src):
...
+    # Download the patch
+    urlretrieve(patch_url, patch_file)
+    print(f"apply {patch_file}")
+    # Patch the documentation
+    result = check_output('patch -p3', stdin=open(patch_file, 'r'), stderr=subprocess.STDOUT, shell=True)
+    print("patch result:")
+    print(result)
+except Exception as exc:
+    print("no patch applied:")
+    print(exc)
```
jougs, line 372:
```suggestion

def patch_documentation():
    """Apply a hot-fix patch to the documentation before building it.

    This function is useful in situations where the online documentation should
    be modified, but the reason for a new documentation build does not justify
    a new release of NEST. Example situations are the discovery of broken links
    spelling errors, or styling issues. Moreover, this mechanism can be used to
    customize to the look and feel of the NEST documentation in individual
    installations.

    In order to make use of this function, the environment variable ``patch_url``
    has to be set to the URL where documentation patch files are located. The
    environment variable must either be set locally or via the admin panel of
    Read the Docs.

    Patch files under the ``patch_url`` are expected to have names in the format
    ``{git_hash}_doc.patch``, where ``{git_hash}`` is the full hash of the version the
    patch applies to.

    The basic algorithm implemented by this function is the following:
    1. obtain the Git hash of the version currently checked out
    2. log the hash by printing it to the console
    3. retrieve the patch

    """
    
    print("Preparing patch...")
    try:
        git_hash = check_output("git rev-parse HEAD", shell=True, encoding='utf8').strip()
        print(f"  current git hash: {git_hash}")
        patch_file = f'{git_hash}_doc.patch'
        patch_url = f'{os.environ["patch_url"]}/{patch_file}'
        print(f"  retrieving {patch_url}")
        urlretrieve(patch_url, patch_file)
        print(f"  applying {patch_file}")
        result = check_output('patch -p3', stdin=open(patch_file, 'r'), stderr=subprocess.STDOUT, shell=True)
        print(f"Patch result: {result}")
    except Exception as exc:
        print(f"Error while applying patch: {exc}")
        
patch_documentation()
```

jougs (commented):
I have added a suggestion to improve the documentation of the patching code by making it into a function.

doc/htmldoc/conf.py
steffengraber: Removed

jougs (approved):
LGTM

PR #2551, opened 2022-11-30
Allow more than one optimization flag to be set when running CMake

### LINKED ISSUE #2548: Cmake ignores the list of optimization options and chooses only the last item in the list
**Describe the bug**
Calling cmake with `-Dwith-optimize="item1;item2;...;itemn"`, only selects the last item in the list, and the other items will be ignored.

med-ayssar:
Fixes #2548


cmake/ProcessOptions.cmake
```diff
@@ -25,10 +25,23 @@ function( NEST_PROCESS_WITH_OPTIMIZE )
     if ( with-optimize STREQUAL "ON" )
```
terhorstd, line 25: There is one unintuitive path along wrongly-cased "on", "On" vs. "ON", which can be caught with `TOUPPER`
```suggestion
    string(TOUPPER "${with-optimize}" WITHOPTIMIZE)
    if ( WITHOPTIMIZE STREQUAL "ON" )
```

terhorstd (changes requested):
Nice addtion :+1: Some minor suggestions for usability

gtrensch (changes requested):
Thank you for this work :+1:. Will be approved when the requested changes are made.

med-ayssar:
> Thank you for this work +1. Will be approved when the requested changes are made.

Actually, I found a more compact version. So maybe if you please take a look at the changes again.

cmake/ProcessOptions.cmake
```diff
@@ -22,13 +22,14 @@
...
       set( with-optimize "-O2" )
     endif ()
-    foreach ( flag ${with-optimize} )
-      set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${flag}" PARENT_SCOPE )
-      set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${flag}" PARENT_SCOPE )
-    endforeach ()
+    set(OPTIMIZATION_FLAGS "")
+    string(JOIN " " OPTIMIZATION_FLAGS  ${with-optimize} )
+    set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OPTIMIZATION_FLAGS}" PARENT_SCOPE )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OPTIMIZATION_FLAGS}" PARENT_SCOPE )
```
terhorstd, line 32: That's a much cleaner solution! Thanks for optimizing the optimization flags :grin:

terhorstd (approved):
All regular cases and most more extravagant cases seem to work. The only unintuitive behavior I found was using multiple `-Dwith-optimize=…` flags, which is however regular CMake behavior. There is little chance to change this from NEST side.

For me this looks good now! :+1:

gtrensch (approved):
Looks good to me! 👍

CMakeLists.txt
```diff
@@ -155,6 +153,11 @@ nest_process_target_bits_split()
 nest_process_version_suffix()
 nest_process_userdoc()
 nest_process_devdoc()
+# IMPORTANT: Please try to keep these two function calls to be the last functions to call in this Process step.
+#            Adding wrong keywords to the "-Dwith-optmize" flag might cause wrong errors, and therefore to avoid that
+#            the step for processing the optmization phase and the debug phase must be the last steps.
```
jougs, line 158:
```suggestion

# These two function calls must come last, as to prevent unwanted interactions of the newly set flags
# with detection/compilation operations carried out in earlier functions. The optimize/debug flags set
# using these functions should only apply to the compilation of NEST.
```

jougs (commented):
I have just some wording changes for the comment on the function call positions.

PR #2560, opened 2022-12-01
Made sphinx build self-sufficient and in-tree-safe

Helveg:
The paths in `conf.py` were complicated and build-env dependent, different between ReadTheDocs, CMake builds and out-of-tree builds. Out-of-tree builds were required because NEST's sphinx configuration littered the source tree. All these problems have been resolved and the documentation can be built without CMake supplying any paths, without env vars supplying `NESTSRCDIR`, and without polluting the NEST source directory:

```
cd doc/htmldoc
sphinx-build . <output-dir> -b html
```

The CMake sphinx process is also simplified, and files no longer need to be copied over, and apart from `resolve_includes.py` also doesn't need to run any more commands. The output is identical.

Most of the littering of the local doc build has been fixed by gitignoring the outputs. Several plugins used by NEST are preprocessors that generate RST files to be used by the build, these mostly don't conflict with any existing files, and are overwritten every time the build is ran. The only downside is that someone might try to change these files and see their changes overwritten.


steffengraber:
Tested `sphinx-build` and it worked without any problems.

doc/htmldoc/examples/.gitignore
```diff
@@ -0,0 +1,5 @@
+# The example's README is copied over from the source directory.
+README.rst
+# The figures are copied over from the source directory as well.
+*.png
+*.svg
```
jougs, line 5: Newline

doc/htmldoc/models/.gitignore
```diff
@@ -0,0 +1,7 @@
+# The UserDocExtractor preprocesses `/models/*.h` and `/nestkernel/*.h` files into
+# .rst and .json files to be picked up during the Sphinx build.
+*.rst
+*.json
+# Only these files should be tracked.
+!models-main.rst
+!models-toc.rst
```
jougs, line 7: Newline

pynest/nest/ll_api_ka.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            return status_root[self._name]
+
+    @stack_checker
+    def __set__(self, instance, value):
+        if self._readonly:
+            msg = f"`{self._name}` is a read only kernel attribute."
+            raise AttributeError(msg)
+        sps({self._name: value})
+        sr('SetKernelStatus')
```
jougs, line 47: Newline

jougs (changes requested):
I think this is a really nice clean-up. I still have some fundamental comments to it.

doc/CMakeLists.txt
```diff
@@ -66,39 +66,22 @@ else ()
...
-  #    installed into the target docdir.
-  # 4. Sphinx generates the models RST documentation from the models/*.h files. But we
-  #    need to copy them into a deeper directory.
-  # 5. To make Python's `help` results more useful, `include` directives are inlined using
-  #    `resolve_includes.py`
-  # 6. This assembled Sphinx folder is marked to be installed, together with the logos.
+  message( STATUS "Configuring Sphinx documentation" )
+  set( _SPHINX_SOURCE_DIR "${PROJECT_SOURCE_DIR}/doc/htmldoc" )
+  set( _SPHINX_BUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}/_build/html" )
   add_custom_target( sphinxdocs
```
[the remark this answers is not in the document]
jessica-mitchell: When it comes generating the documentation from Sphinx, although it mainly is userdocs, it does also contain docs relevant and useful only to developers. There is a whole section on contributing to nest code and docs, and fundamentally that is not necessary user knowledge (this is I think a gray area: at what point does a user become a developer?). It also should, in the future, be home to the actual doxygen docs.  

So I know that Sphinx is merely the tool, but it is what we are using now.  The term htmldoc was chosen to move away from explicit use of user because the developer_space/ is within that directory. And it describes the role of this directory: to provide the files for documentation on the web.

jessica-mitchell:
@helveg Building with sphinx-build works well but I'm having some issues with building the docs from the build directory. I'm still trying to sort out if it's something in my environment that is misbehaving. But I noticed that in the CMakelists.txt in doc/ there is a `PYTHON_EXECUTABLE` which should be changed to `Python_EXECUTABLE` like the others.

doc/CMakeLists.txt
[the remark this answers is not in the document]
Helveg: My 2 cents is that I opted for this separation so that there are 2 layers: userdoc/devdoc as options, toggling and configuring indeed whichever tool we use to build things,  and this doesn't have to be a 1 to 1 relationships, perhaps in the future multiple builds will form the user doc together. Naming it sphinx docs also makes sense as you'll see that anything inside of the block relates strictly to sphinx commands and would have to be replaced in its entirety if we switch to another tool

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
```diff
@@ -90,37 +90,34 @@ Rendering HTML
...
-1. Go to the :ref:`build directory <dev_install>` of NEST:
+1. Clone the NEST repository:
 
 .. code-block:: bash
 
-   cd nest-simulator-x.y.z-build
+  git clone git@github.com:nest/nest-simulatr
 
-2. Add `-Dwith-userdoc=ON` to your regular CMake command:
+2. Navigate to the ``doc/htmldoc`` folder:
```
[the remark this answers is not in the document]
Helveg: With the upcoming trivialization of the installation, and the widespread availability of NEST on other platforms, I'm not sure that that is going to hold up, and I'm not sure that it's good if it does hold up: Ideally users never have to bother with source code, and contributing to the documentation should not require understanding of complicated build systems like CMake. It poses a considerable barrier if one needs to succeed in configuring building through CMake before they can contribute to the docs.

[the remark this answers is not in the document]
Helveg: Occham's razor's instruction leaflet: git clone + sphinx-build is simpler than git clone + cmake + make, especially considering everything that can go wrong in the CMake step, entirely unrelated to the doc they're trying to build

Helveg:
@jessica-mitchell could you list the steps I'd have to take to reproduce your issue?

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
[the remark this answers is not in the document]
Helveg: Oh, and final remark, if people want to build the doc _and_ sources, this is as simple as passing the `with-userdoc` flags. I don't feel that explaining a CMake config option is the main aim of this dedicated "how to build the docs" documentation page.

jessica-mitchell:
@Helveg I run into an issue building the user docs with `make docs`. It does not put the HTML in the build directory

```
The HTML pages are in ../../../../build-repo/gen-build/doc/_build/html.
  File "/home/mitchell/Work/repo/nest-simulator/doc/htmldoc/resolve_includes.py", line 42, in <module>
    path = sys.argv[1]
IndexError: list index out of range
make[3]: *** [doc/CMakeFiles/sphinxdocs.dir/build.make:71: doc/CMakeFiles/sphinxdocs] Error 1
make[2]: *** [CMakeFiles/Makefile2:515: doc/CMakeFiles/sphinxdocs.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:496: doc/CMakeFiles/docs.dir/rule] Error 2
make: *** [Makefile:192: docs] Error 2
```

jessica-mitchell:
@helveg can you resolve the conflicts and fix static checks, thanks!

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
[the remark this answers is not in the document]
Helveg: @jougs could you make an executive decision? Then I can wrap this up

doc/CMakeLists.txt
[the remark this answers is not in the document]
jougs: Alright. Thanks for explaining again. I try to remain silent about this from now on ;-)

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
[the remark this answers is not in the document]
jougs: OK. You win!

jougs (commented):
I've just resolved my remaining conversations. What's left from my side is adding the missing newlines at end of file and resolving the conflicts that have meanwhile crept in.

jessica-mitchell:
@Helveg ping!

Helveg:
I'll have a look over the weekend! I'll be working on a project including next for a couple of weeks so I'll have time :)

Helveg:
@jessica-mitchell There's several doc building requirements. How are they supposed to be installed? There's `sphinx`, clearly, but also `tqdm`, `sphinx_gallery`, `nbsphinx`, `sphinx_design`, `sphinx_material`, and `pillow`. The build errors out with (to me) pretty clear missing module errors, so I think we can just include an install command in the doc building guide to cover them.

cmake/FindSphinx.cmake
```diff
@@ -12,11 +12,20 @@ if( Python_FOUND )
...
 find_program(
     SPHINX_EXECUTABLE
     NAMES sphinx-build sphinx-build.exe
     HINTS ${_PYTHON_PATHS}
+    VALIDATOR check_sphinx
     )
+
 mark_as_advanced(SPHINX_EXECUTABLE)
 
 find_package_handle_standard_args(Sphinx DEFAULT_MSG SPHINX_EXECUTABLE)
```
jougs, line 31: Missing newline! BAM!

doc/htmldoc/_ext/HoverXTooltip.py
```diff
@@ -210,7 +208,7 @@ def get_desc_from_glossary(term):
     """
 
     try:
-        with open(str(doc_build_dir) + '/ref_material/glossary.rst') as f:
+        with open(str('ref_material/glossary.rst')) as f:
```
jougs, line 211:
```suggestion
        with open('ref_material/glossary.rst') as f:
```

doc/htmldoc/_ext/VersionSyncRole.py
```diff
@@ -54,7 +53,7 @@ def role(name, rawtext, text, lineno, inliner, options={}, content=[]):
         else:
             level = "min"
 
-        with open(str(doc_build_dir) + '/_ext/versions.json') as fp:
+        with open(Path(__file__).parent.join('versions.json')) as fp:
```
jougs, line 56: Wouldn't this also work and be more readable?
```suggestion
        with open(Path(__file__).parent / 'versions.json') as fp:
```

doc/htmldoc/conf.py
```diff
@@ -98,22 +59,18 @@
     'VersionSyncRole',
 ]
 
+autodoc_mock_imports = ["nest.pynestkernel", "nest.ll_api"]
```
jougs, line 62: Beautiful! :heart:

doc/htmldoc/conf.py
```diff
@@ -330,21 +286,18 @@ def setup(app):
...
-    copyfile(src, doc_build_dir / "static/img" / src.parts[-1])
-
-
-def copy_acknowledgments_file(src):
-    copyfile(src, doc_build_dir / src.parts[-1])
-
+    copyfile(os.path.join(pynest_dir, src), Path("examples") / Path(src).parts[-1])
 
-# -- Copy Acknowledgments file ----------------------------
-copy_acknowledgments_file(source_dir / "ACKNOWLEDGMENTS.md")
```
jougs, line 341: Don't we need to copy this file anymore?

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
```diff
@@ -90,37 +90,34 @@ Rendering HTML
...
+   browser ../_build/html/index.html
 
-   If you are planning to install NEST, ``make docs`` is included
-   in ``make install`` as long as you toggle on the userdoc or devdoc
+.. tip::
 
-   After installation the files are located in
-   ``<nest_install_dir>/share/doc/nest/html/index.html``
+   You can also build the user documentation during the CMake build using
+   the flag ``-Dwith-userdoc=ON``
```
jougs, line 120: Partially unrelated to this one, but maybe you can still comment on it here:

I noticed that since the merge of #2544, docs are built *before* the code is built when `-Dwith-userdoc=ON` is given. Can this please be changed, so that docs are only built after the code was successfully built?

I find the current behavior rather annoying when I have to do repetitive compilations and it slows me down quite a bit...

pynest/nest/ll_api_ka.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*-
+#
+# ll_api_ka.py
```
jougs, line 3: Can you please rename this file to something more meaningful? Thanks!

jougs (changes requested):

doc/htmldoc/conf.py
Helveg: @jessica-mitchell explained that the achknowledgements are now automatically transcluded in some other way?

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
Helveg: Ok, I'll have a look if that's possible some way.

doc/htmldoc/conf.py
jessica-mitchell: The acknowledgments in master is a link in index.html to the github page

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
jougs: Maybe it is smarter to take this offline first, after all... I'll write you a mail.

doc/htmldoc/conf.py
jougs: Alright. Thanks for the clarification.

jougs:
@jessica-mitchell: if you approve, I'll merge this. Thanks!

jessica-mitchell (commented):
@helveg The build of the docs locally and on Read the Docs work fine. However, the docs that are generated at build (pynest examples( `auto_examples/`) and models) all end up in the source directory. Previously we had them in the build directory which kept the source uncluttered. Can we have these files end up in build? @jougs I otherwise approve this PR, if it's urgent to put in.

Helveg:
They end up in the source directory because they are part of the source that Sphinx picks up. Everything that ends up in the source directory as generated pre-build files should be gitignored though. Are there any files that are not gitignored cluttering your git tree?

jessica-mitchell:
@Helveg yah, adding them to .gitignore works. I didn't think about it when building since this hasnt been an issue for a long while. Should `.gitigonore` be updated in this PR then?   Namely adding, `doc/htmldoc/autoexamples` `doc/htmldoc/models/*` `!doc/htmldoc/models/models-main.rst` and `!doc/htmldoc/models/models-toc.rst`

Helveg:
I was convinced I had done so! 😅 What sreps did you take to produce the not gitignored files?

jougs:
From my perspective, the files created by Sphinx are definitely build artifacts, much like the `.o` files generated by the compiler. As such they should not end up in the source directory. Actually, in my opinion nothing should end up there unless I put it there myself. That's why I created the (complicated) solution that is currently in place in master.

In my workflow, I want to be able to delete the build directory and be sure nothing of the build configuration remains, as I spent way too much time in my life chasing after old tests, docs, and other generated files that should not have been there anymore. Putting files in `.gitignore` is not sufficient for achieving this goal, as this would still lead to outdated documentation be shown if I delete or rename files, but do not delete the old files manually.

jessica-mitchell:
To reproduce: I just built the user documentation - `make docs` in my build directory - the auto generated rst files then appear in source...however, the problem still persists. I thought I had it working but it appears that the files still show up in my source tree. I'm not that familiar with .gitignore so I'm probably doing something wrong.

Helveg:
Thx @jessica-mitchell I'll take a look. 

I believe there's 2 sources of additional source files: the extract userdoc script, and `sphinx-gallery`. Neither tools are actually integrated into the sphinx build process: sphinx-gallery seems to by design just render files into RST for sphinx to pick up, and our in-house extract userdoc does the same when `conf.py` is imported.

The ideal solution is to turn these scripts into extensions that can create [doctrees](https://www.sphinx-doc.org/en/master/extdev/index.html#build-phases) during the build phase for each of these, so that sphinx can render them, without the need of intermediate source files on disk.

For now, @jougs would it suffice to keep the gitignored source trees, but to clear out any possibly stale files that may be there from previous builds?

jougs:
> For now, @jougs would it suffice to keep the gitignored source trees, but as part of the sphinx build to clear out any possibly stale files that these tools may have put there from previous builds?

First of all: sorry that this slipped my attention when reviewing. I was under the impression that the new setup still works in the build directory, but does so without the manual steps and workarounds I have added back in the day.

Back to the topic. I just talked to @terhorstd and we agree that out-of-tree builds should be considered the default with building documentation on Read the Docs being the exception. As a consequence our policy is that no files are put in the source tree when building in a different directory.

For a development workflow with multiple active branches at a time and different configurations on the same source branch, a separation between an unmodified source tree (i.e. what comes from the Git repository) and build directory (i.e. the `cmake`d configuration together with the installation, the docs, etc.) has proven very useful.

The main benefits of such a setup are that
* `.gitignore` files can stay minimal and don't have to be updated regularly
* no build artifacts can be accidentally committed
* build files for different configurations are persistent and cannot be overwritten by other configurations

How hard would it be to change your current setup so it does not leave any artifacts in the source tree?

Helveg:
The in-tree build can't be made any better than this without creating extensions: sphinx has no separate build directory so the source tree needs to be modified. I can add extra safeguards to make sure the stale files are cleaned up first.

I think I can run the same in-tree build from the CMake build directory, rather than the source directory, simply by having CMake copy over all the source files to the build dir.. right?

terhorstd:
> I think I can run the same "in-tree" build but from the CMake build directory, rather than the source directory, simply by having CMake copy over all the source files to the build dir.. right?

It's not a clean solution, but copy could work. I think we could go this way for now, and open a new issue for the proposed Sphinx extension. For the time being, for developers, the CMake options to not build docs could prevent this copying and would keep build dir clean. Additionally, copying with `-l` when possible would keep i-node counts low.

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
jessica-mitchell, line 120: I would explicitly write out here, something like: 
You can also build the user documentation in the build directory with CMake:
`cmake -Dwith-userdoc=ON </path/to/NEST/src>`
`make docs`.

jessica-mitchell (changes requested):
@helveg I just have a couple of minor suggestions I spotted.

terhorstd:
@Helveg, any news on this?

Helveg:
@jougs Can you review again?

jougs (approved):
I have some very minor suggestions, but am immediately approving to not delay this any further. Many thanks for cleaning this up and making it nicer :smile:

doc/htmldoc/clean_source_dirs.py
```diff
@@ -0,0 +1,38 @@
...
+            file_.endswith(f)
+            for f in (".gitignore", "models-main.rst", "models-toc.rst")
+        ):
+            try:
+                try:
+                    os.unlink(file_)
+                except OSError:
+                    shutil.rmtree(file_)
+            except Exception:
+                print(f"Couldn't remove '{file_}'")
```
jougs, line 38: Please add a newline to the end of the file. Some PEP8-check fails, if that is missing...

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
```diff
@@ -90,37 +90,38 @@ Rendering HTML
 You can build and preview the documentation locally by running the following
 commands.
 
-1. Go to the :ref:`build directory <dev_install>` of NEST:
+1. Clone the NEST repository:
 
 .. code-block:: bash
 
-   cd nest-simulator-x.y.z-build
+  git clone git@github.com:nest/nest-simulator
```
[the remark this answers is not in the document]
Helveg: I think people who make forks are likely to know this

doc/htmldoc/conf.py
```diff
@@ -1,466 +1,420 @@
...
-        print(f"  applying {patch_file}")
-        result = subprocess.check_output('patch -p3', stdin=open(patch_file, 'r'), stderr=subprocess.STDOUT, shell=True)
-        print(f"Patch result: {result}")
-    except Exception as exc:
-        print(f"Error while applying patch: {exc}")
-
-
-patch_url = os.getenv("patch_url")
-if patch_url is not None:
-    patch_documentation(patch_url)
```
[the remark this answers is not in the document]
Helveg: Perhaps newlines? I'm not sure. I have `git config autocrlf true`, it should commit as linux newlines

jessica-mitchell (approved):
@Helveg @jougs I also approve - further discussions can be had later but this improves docs a lot, so thanks!

doc/htmldoc/conf.py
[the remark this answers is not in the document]
jougs: Whatever ;-)

PR #2565, opened 2022-12-08
Remove autogenerated file from models directory

jessica-mitchell:
This PR removes the index.rst from models/ folder as it was accidentally added in a previous PR. This file is autogenerated so no file is required in the source.


steffengraber (approved):
Deleting things that are no longer required is always a good idea.

jessica-mitchell:
Merging with 1 approval as its documentation related

PR #2574, opened 2023-01-02
Refactor main update block

med-ayssar:
This PR reduces the size (with respect to the number of lines) of the main update function in the `SimulationManager` into sub-functions.

It is mainly for readability reasons!


terhorstd:
@med-ayssar, could you describe what these changes do, which issue they solve, etc. in the PR description?

med-ayssar:
@med-ayssar ping!

PR #2582, opened 2023-01-10
Change order of pre_run hook in `SimulationManager`

marcelkrueger:
By rearranging the pre_run hook call it allows callees to access information about the simulation time in their pre_run implementation. 
Otherwise callees are not able to reliable get these information before the run starts as the simulation time is only known when calling the run() method and the sim time is currently set after the pre_run hook.

Additionally, the pre_run hook is only called if a run will take place at all.

As far as I see the reordering should be safe as there are no dependencies between the reordered  lines.

### DIFF
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -543,7 +543,6 @@ nest::SimulationManager::run( Time const& t )
   assert_valid_simtime( t );
 
   kernel().random_manager.check_rng_synchrony();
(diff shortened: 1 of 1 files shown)


jougs (approved):
LGTM. Thanks!

med-ayssar:
I have no objection, thanks!

terhorstd:
@med-ayssar , could you give approval in the review, so that we can more easily spot that this is ready to merge?

med-ayssar (approved):

PR #2585, opened 2023-01-11
Allow setting MPI port names of MPI recording and stimulation backends from device labels

marcelkrueger:
This is PR adds the ability to set the port name for the MPI connection additionally via the device label.

Falls back to normal file based address providing if label is not set.

Originally created for the cosim framework by @mfahdaz

### DIFF
--- a/nestkernel/recording_backend_mpi.cpp
+++ b/nestkernel/recording_backend_mpi.cpp
@@ -358,15 +358,24 @@ void
 nest::RecordingBackendMPI::get_port( const RecordingDevice* device, std::string* port_name )
 {
   const std::string& label = device->get_label();
-  std::size_t address_starts_at;
+  std::size_t colon_position;
+
+  // The MPI address can be provided by two different means:
+  // a) the address is given by the label in the format: endpoint_address:address
+  // b) the file is provided via a file: {data_path}/{data_prefix}{label}/{node_id}.txt
+
   // Case a: label is the endpoint address in the following format:
   // endpoint_address:address string
-  if ((address_starts_at = label.find(":",0)) != std::string::npos) {
-    *port_name = label.substr(address_starts_at+1);
-  }
+  colon_position = label.find( ":", 0 );
 
-  // Case b: label is part of the path to a file that contains endpoint address
-  else {
+  // check if string start with "endpoint_address:" and a colon is found
+  if ( label.find( "endpoint_address:" ) == 0 )
+  {
+    *port_name = label.substr( colon_position + 1 );
+  }
+  // Case b: fallback to get_port implementation that reads the address from file
+  else
+  {
     get_port( device->get_node_id(), label, port_name );
   }
 }
@@ -400,7 +409,6 @@ nest::RecordingBackendMPI::get_port( const index index_node, const std::string& 
   {
     getline( file, *port_name );
   }
-  
   file.close();
 }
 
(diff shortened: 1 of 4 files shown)


nestkernel/recording_backend_mpi.cpp
```diff
@@ -357,7 +357,27 @@ nest::RecordingBackendMPI::set_status( const DictionaryDatum& )
 void
 nest::RecordingBackendMPI::get_port( const RecordingDevice* device, std::string* port_name )
 {
-  get_port( device->get_node_id(), device->get_label(), port_name );
+  const std::string& label = device->get_label();
+  std::size_t colon_position;
+
+  // The MPI address can be provided by two different means:
+  // a) the address is given by the label in the format: endpoint_address:address
```
jougs, line 364:
```suggestion
  // a) the address is taken from the device's label, which then has to be in the format: endpoint_address:address
```

nestkernel/recording_backend_mpi.cpp
```diff
@@ -357,7 +357,27 @@ nest::RecordingBackendMPI::set_status( const DictionaryDatum& )
...
+  // b) the file is provided via a file: {data_path}/{data_prefix}{label}/{node_id}.txt
```
jougs, line 365:
```suggestion
  // b) the address is provided via a file: {data_path}/{data_prefix}{device_label}/{node_id}.txt
```

nestkernel/recording_backend_mpi.cpp
```diff
@@ -357,7 +357,27 @@ nest::RecordingBackendMPI::set_status( const DictionaryDatum& )
...
+
+  // Case a: label is the endpoint address in the following format:
```
jougs, line 367:
```suggestion
  // Case a: device label is the endpoint address in the following format:
```

jougs (commented):
Thinking about this again, I have to admit that I rather dislike the way this feature is implemented by hijacking the device label, which was originally meant to provide a descriptive device identifier for the user.

I think the right way of adding this would be a new device-specific backend property (e.g. `mpi_address`). See [this code](https://github.com/nest/nest-simulator/blob/master/nestkernel/recording_backend_ascii.h#L195-L219) in the ASCII recording backend  for what I mean. The semantics would be that the address from this new property would used, or read from file if the property is empty.

jougs:
The same changes would have to be applied also to `stimulation_backend_mpi.cpp`. However, in order to save you from duplicate work, can you please first comment on my more far-reaching proposal?

jougs:
I think it is probably good enough to have an empty `mpi_address` property as the default with that meaning to take te address from a file.

If you're willing to go the full length with checks for mutual exclusivity of mulsiple properties and such, I am also not against it, it just seems to be more work and I'm not sure the small user community of this backend justifies that.

jougs:
@marcelkrueger: what is the status of this? Thanks!

med-ayssar:
@marcelkrueger: a gentle ping!

marcelkrueger:
I'm sorry for not getting back to you sooner.
I already swiftly talked to @jougs in a meeting about it but forgot to update here.

Unfortunately, I do not have the capacity right now to implement this properly.
Additionally, it is not high on my priority list as I am not a direct user but just cleaned up the contribution by the original author @mfahdaz.

I might be able to pick it up at a later stage.
However, if it is needed in the meantime anyone can feel free to pick it up.

marcelkrueger:
I am closing this in favour of the restructured approach in #2858.

PR #2589, opened 2023-01-17
Fix mocker regex to avoid failures when building documentation

### LINKED ISSUE #2576: Read the Docs build fails on master
**Describe the bug**

Read the docs fails to build with the following error:
```
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/envs/2495/lib/python3.8/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/checkouts/2495/doc/htmldoc/conf.py", line 71, in <module>
    mockedmodule += convert(fin)
  File "/home/docs/checkouts/readthedocs.org/user_builds/nest-simulator/checkouts/2495/doc/htmldoc/mock_kernel.py", line 79, in convert
    tree = ast.parse(res_tmp)
  File "/home/docs/.pyenv/versions/3.8.6/lib/python3.8/ast.py", line 47, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 180
    array_long_mv = numpy.ascontiguousarray(array, dtype=)
```

**To Reproduce**
1. Build Read the Docs from the master branch

This is related to a change in the pynestkernel.pyx file in PR #2571:

```
-                array_long_mv = numpy.ascontiguousarray(array, dtype=numpy.int_)
+                array_long_mv = numpy.ascontiguousarray(array, dtype=int)
```
Read the docs build passes using the older syntax.

@heplesser We can revert it back to the original, but not sure if there is a better alternative that doesnt use the numpy syntax?

hakonsbm:
Replaced `\b` with `\s` in the regex to avoid removing wrong `int`s when mocking pynestkernel. Fixes #2576.

### DIFF
--- a/doc/htmldoc/mock_kernel.py
+++ b/doc/htmldoc/mock_kernel.py
@@ -51,7 +51,7 @@ def convert(infile):
 
     cdef_in_classes_re = re.compile(r'    +cdef')
     cdef_class_re = re.compile(r'cdef class (.*)')
-    rmdatatype_re = re.compile(r'\bint\b|\bnew\b|\&(?=\w)|<[^>]+>')
+    rmdatatype_re = re.compile(r'\sint\s|\bnew\b|\&(?=\w)|<[^>]+>')
 
     inclass = False
 


jessica-mitchell (approved):
Tested the local build and RtD, both build successfully! thanks @hakonsbm

jougs (approved):
Many thanks!

PR #2591, opened 2023-01-18
Modify `spike_generator` to have proxies

nicolossus:
In multithreaded simulations with a spike generator, a copy of the spike generator is spawned at each thread. For large spike-times arrays, this cause a significant memory usage which is unnecessary because a spike generator can be shared across threads. 

This PR makes it so that threads share a single spike generator. The sharing is enabled by using the existing framework with `has_proxies`. That is, internally a spike generator now behaves as a neuron and not device. 

In order to preserve NEST's external behavior, changes have been made to prohibit connecting spike recorders to global receivers and exclude spike generator spikes from the `local_spike_counter` count. A couple of tests needed minor modifications in order to accommodate the new spike generator's behavior.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1467,7 +1467,12 @@ nest::ConnectionManager::connection_required( Node*& source, Node*& target, thre
   // currently prohibit connections from devices to global receivers.
   else
   {
-    if ( source->has_proxies() )
+
+    const index spike_gen_id = kernel().model_manager.get_node_model_id( "spike_generator" );
+    const index src_model_id = kernel().model_manager.get_node_model_id( source->get_name() );
+    const auto source_is_spike_gen = spike_gen_id == src_model_id;
+
+    if ( source->has_proxies() and not source_is_spike_gen )
     {
       target = kernel().node_manager.get_node_or_proxy( target->get_node_id(), tid );
       return CONNECT;
--- a/nestkernel/event_delivery_manager_impl.h
+++ b/nestkernel/event_delivery_manager_impl.h
@@ -69,9 +69,18 @@ EventDeliveryManager::send< SpikeEvent >( Node& source, SpikeEvent& e, const lon
   const thread tid = source.get_thread();
   const index source_node_id = source.get_node_id();
   e.set_sender_node_id( source_node_id );
+
   if ( source.has_proxies() )
   {
-    local_spike_counter_[ tid ] += e.get_multiplicity();
+
+    const index spike_gen_id = kernel().model_manager.get_node_model_id( "spike_generator" );
+    const index src_model_id = kernel().model_manager.get_node_model_id( source.get_name() );
+    const auto source_is_spike_gen = spike_gen_id == src_model_id;
+
+    if ( not source_is_spike_gen )
+    {
+      local_spike_counter_[ tid ] += e.get_multiplicity();
+    }
 
     e.set_stamp( kernel().simulation_manager.get_slice_origin() + Time::step( lag + 1 ) );
     e.set_sender( source );
--- a/models/spike_generator.h
+++ b/models/spike_generator.h
@@ -226,6 +226,7 @@ public:
   void get_status( DictionaryDatum& ) const override;
   void set_status( const DictionaryDatum& ) override;
   bool has_proxies() const override;
+  bool is_off_grid() const override;
 
   StimulationDevice::Type get_type() const override;
   void set_data_from_stimulation_backend( std::vector< double >& input_spikes ) override;
@@ -381,6 +382,12 @@ inline bool
 spike_generator::has_proxies() const
 {
   return true;
+}
+
+inline bool
+spike_generator::is_off_grid() const
+{
+  return P_.precise_times_;
 }
 
 } // namespace nest
--- a/models/spike_generator.h
+++ b/models/spike_generator.h
@@ -85,6 +85,7 @@ If false, spike times will be rounded to the nearest step if they are
 less than tic/2 from the step, otherwise NEST reports an error.
 If true, spike times are rounded to the nearest step if within tic/2
 from the step, otherwise they are rounded up to the *end* of the step.
+Irrelevant if precise_times is true.
 
     `shift_now_spikes`   default: false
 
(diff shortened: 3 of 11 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -1467,7 +1467,12 @@ nest::ConnectionManager::connection_required( Node*& source, Node*& target, thre
   // currently prohibit connections from devices to global receivers.
   else
   {
-    if ( source->has_proxies() )
+
+    const index spike_gen_id = kernel().model_manager.get_node_model_id( "spike_generator" );
+    const index src_model_id = kernel().model_manager.get_node_model_id( source->get_name() );
+    const auto source_is_spike_gen = spike_gen_id == src_model_id;
```
hakonsbm, line 1473: Here `"spike_generator"` is converted to a `Name` object before getting the model ID for every connection going into this clause. Isn't it better to store the `spike_gen_id` value in ConnectionManager (and set it in `ConnectionManager::initialize()`) so we only have to get it once?

And you can get the model ID of the source node directly with `source->get_model_id()`.

nestkernel/event_delivery_manager_impl.h
```diff
@@ -69,9 +69,18 @@ EventDeliveryManager::send< SpikeEvent >( Node& source, SpikeEvent& e, const lon
...
   const index source_node_id = source.get_node_id();
   e.set_sender_node_id( source_node_id );
+
   if ( source.has_proxies() )
   {
-    local_spike_counter_[ tid ] += e.get_multiplicity();
+
+    const index spike_gen_id = kernel().model_manager.get_node_model_id( "spike_generator" );
+    const index src_model_id = kernel().model_manager.get_node_model_id( source.get_name() );
+    const auto source_is_spike_gen = spike_gen_id == src_model_id;
```
hakonsbm, line 78: Same here.

nestkernel/connection_manager.cpp
jougs: Additionally, this does not work for models that came into existence by being `CopyModel()'`d from the original `spike_generator` and makes the NEST kernel dependent on the availability of a model called `spike_generator`.

Moreover this adds to the connection time of each device to support this special case...

jougs (commented):
It is true that the replication of `spike_generator`s for each threads leads to a duplication of memory and it would be nice to get rid of that.

However, I'm afraid the proposed solution suffers from misconception about what `has_proxies` means: Nodes _with_ proxies are only existing once in the whole MPI universe and transmit their events via MPI. I don't think we want this for a device, as that might add quite a bit to the volume of data transmitted.

A way out could be using the [`one_node_per_proxies`](https://github.com/nest/nest-simulator/blame/bb43f8ae151fe28ac548b019359e4ffd26deaf8b/nestkernel/node.h#L157) property.  On the other side, I can imagine that it could be quite problematic for the performance, if all threads have to be fed by a single generator that would have to write into another threads's memory.

nestkernel/connection_manager.cpp
nicolossus: I agree with you both, and another procedure is needed to distinguish nodes with the properties we want for `spike_generator`. For the record, it is also not possible to retrieve `spike_generator`s model id in `ConnectionManager::initialize` because the models are registered by `ModelsModule::init` at a later stage.

nicolossus:
@jougs Would you prefer an entirely new way of distinguishing nodes with the properties we want for `spike_generator`? This would of course add another layer of complexity to several routines, but it might also be the cleanest and perhaps more importantly the most efficient solution. 

Just to provide some context, the Allen Institute's V1 model suffers from the current implementation. In this model, retinal input to V1 is mediated by the LGN which is modelled as an external population of spike generators. The spike-times arrays are fairly large. On a single compute node with 128 cores, the memory consumption of the build phase is about 5 GBs when using 1 core and it ramps up to about 35 GBs when using all 128. With hyperthreading (256 threads) the memory consumption is about 50 GBs. The V1 model has roughly 25e4 nodes with 70e6 connections, so it is not particularly large-scale either.

jougs:
Thanks for the explanation. I think it nicely demonstrates the scale of the problem you are facing.

I still think that `has_proxies` is the wrong flag here, as that would result in a _single_ generator feeding _all_ connected neurons in the whole MPI universe and a massive increase of MPI communication, buffer sizes and (most likely) spike delivery times. Without solid benchmarks showing that this is not a problem, I am very unlikely to accept this PR.

Did you try out / think about the `one_node_per_process` flag I pointed out above?

An alternative (also with unknown consequences for performance) would be to let all `spike_generator`s look at the same spike time arrays by giving them a pointer to a shared piece of memory instead of having them all store their own copy of the array. This could even be made configurable, so users could chose if they rather want to save memory or get the best possible performance by having a thread-local replica of the data.

JanVogelsang:
Ideally, what we would want would be to have the memory duplicated only one per NUMA node. If there are 64 threads per NUMA node, these 64 threads would all want a copy of the SpikeGenerator which is connected to only the thread-local neurons, and they would all use the same NUMA-local memory. Getting some information about the NUMA layout would be quite handy in general for some other optimizations (e.g. parallel spike buffer collocation) which suffer from low performance due to no control over the NUMA nodes.  
Using `<numa.h>` it is possible to retrieve the NUMA node which could then be used to duplicate data.

heplesser:
The key question is what role a `spike_generator` has in a model. If a generator represents an individual neuron, its firing rate would be on par with normal neurons, O(10 spikes/s), and in this case having the `spike_generator` implemented like a normal neuron would not lead to unreasonable increases in spike transmission volume. If a `spike_generator` represents an entire population, similar to a `poisson_generator`, the picture is very different, with firing rates of O(10^4) to O(10^5). In that case, transmission bandwidth would be an issue. But is that realistic for `spike_generator`? Should we add a switch so the user can control behaviour?  Have two different models, one for "like a neuron" and one for "like a population"?

JanVogelsang:
Is the `spike_generator` also copied onto nodes where there are no local targets? If that is the case, what you suggested what be as simple as fixing this issue.

jougs:
@all: Maybe we should have a dedicated video call on this? I have a feeling that there's some misconceptions and conflicting goals and ideas that we could probably much easier clear up by talking. Who would be interested in such a discussion?

nicolossus:
@jougs I also think we should have a video meeting about this. I will send you all an email tomorrow so we can arrange one.

jougs:
@nicolossus: Is this PR still relevant now that the `injector_neuron` is on the way?

nicolossus:
@jougs No, the upcoming injector neuron will be a replacement for this PR. Just to have the complete picture in this thread; in internal discussions we decided to instead make a new "injector neuron" model which behaves similar to spike generator but that only resides on a single virtual process (no replication to other vps). Closing this PR.

PR #2597, opened 2023-01-30
Fixing bad coding pattern in loops in connection creation

JanVogelsang:
We cast to `int` in a loop so that we can do `--i` to counteract the loops `++i` if a connection is not created. Instead, `target_number_connections` should just be counted down in a while loop here.


jougs (approved):
This is indeed much nicer. @heplesser: can you please also review?

nestkernel/connection_creator_impl.h
```diff
@@ -438,19 +438,17 @@ ConnectionCreator::fixed_indegree_( Layer< D >& source,
         std::vector< bool > is_selected( positions.size() );
 
         // Draw `target_number_connections` sources
-        for ( int i = 0; i < ( int ) target_number_connections; ++i )
+        while ( target_number_connections > 0)
```
jougs, line 441:
```suggestion
        while ( target_number_connections > 0 )
```

nestkernel/connection_creator_impl.h
```diff
@@ -483,12 +482,11 @@ ConnectionCreator::fixed_indegree_( Layer< D >& source,
...
```
jougs, line 485:
```suggestion
        while ( target_number_connections > 0 )
```

nestkernel/connection_creator_impl.h
```diff
@@ -564,19 +563,17 @@ ConnectionCreator::fixed_indegree_( Layer< D >& source,
         std::vector< bool > is_selected( positions->size() );
 
         // Draw `target_number_connections` sources
-        for ( int i = 0; i < ( int ) target_number_connections; ++i )
+        while ( target_number_connections > 0)
```
jougs, line 566:
```suggestion
        while ( target_number_connections > 0 )
```

nestkernel/connection_creator_impl.h
```diff
@@ -602,19 +600,17 @@ ConnectionCreator::fixed_indegree_( Layer< D >& source,
...
```
jougs, line 603:
```suggestion
        while ( target_number_connections > 0 )
```

jougs (changes requested):
There's another one in [lines 737-750](https://github.com/nest/nest-simulator/pull/2597/files#diff-8c1da9f8deff4ed8c17f28c8669992da99e86bad7ed82361136acea7ee4c09e6L737-R750). Please also change that.

heplesser (commented):
This looks good. Could your add `const` for local variables in the loops that lack it but could be const, e.g., `random_id` or `source_id`.

jougs:
@JanVogelsang: friendly ping!

nestkernel/connection_creator_impl.h
```diff
@@ -714,7 +711,7 @@ ConnectionCreator::fixed_outdegree_( Layer< D >& source,
       probabilities.resize( target_pos_node_id_pairs.size(), 1.0 );
     }
 
-    const auto number_of_connections = std::round( number_of_connections_->value( grng, src ) );
+    long number_of_connections = std::round( number_of_connections_->value( grng, src ) );
```
jougs, line 714: The [new warning](https://github.com/nest/nest-simulator/actions/runs/4231025828/jobs/7349173899#step:9:657) (that is now failing the build) suggests that this should rather be a `size_t`

jougs: The [second new warning](https://github.com/nest/nest-simulator/actions/runs/4231025828/jobs/7349173899#step:9:664) seems to point to the same problem.

JanVogelsang: I think the signed type used there is the main issue here. So using `unsigned long` might be the way to go here. Of course, simply using `size_t` would also work and might also make sense as we are comparing to `size_t` here. Without knowing that we compare against a `size_t` (which one does in theory not know at the time of the declaration), `unsigned long` would be the type I would use here, as this would be the correct type for a simple counter. What do you think?

jougs: I think your reasoning `unsigned long` is sound. Go for it!

JanVogelsang:
Cancelled after 360min, how could that happen?

heplesser:
The problem causing the timeout is here:
https://github.com/nest/nest-simulator/blob/f17346d33b0d7b78d49bb44d60127776dcacf9cd/nestkernel/connection_creator_impl.h#L765-L771

Before the `continue`, we need to insert `--target_number_connections` to account for connections created on other VPs (this is `fixed_outdegree` which does not parallelize: all VPs need to draw all connections and then just skip creating those they are not responsible for).

JanVogelsang:
Thanks Hans Ekkehard, I'm mad at me to not notice this myself earlier. I fixed it, let's see if the testsuite runs through now.

jougs (approved):

nestkernel/connection_creator_impl.h
```diff
@@ -757,6 +757,8 @@ ConnectionCreator::fixed_outdegree_( Layer< D >& source,
         rng_delay_vec.push_back( delay_[ indx ]->value( grng, source_pos_vector, target_pos_vector, target, tgt ) );
       }
 
+      --number_of_connections;
+
       // We bail out for non-local neurons only now after all possible
       // random numbers haven been drawn. Bailing out any earlier may lead
       // to desynchronized global rngs.
```
heplesser, line 764: We should comment a bit better what is going on here. The `if ( not ... )` line should come right after the `--number_...` line.
```suggestion
      // Each VP has now decided to create this connection and drawn any random parameter values
      // required for it. Each VP thus counts the connection as created, but only the VP hosting the
      // target neuron actually creates the connection.
      --number_of_connections;
```

heplesser (changes requested):
@JanVogelsang Thanks! Just a small suggestion for a better comment so we won't wonder a year from now why we decrease the counter half-way through the loop.

heplesser (approved):

PR #2600, opened 2023-02-01
Fix patch application for documentation

jougs:
See the commit messages in this PR for details.

@steffengraber: Can you make sure this still works as intended on RtD?
@terhorstd: Are you fine with the code I've added?


steffengraber:
@jougs works on RTD

terhorstd (approved):
That looks very reasonable!
Thanks for the quick fix!

PR #2601, opened 2023-02-01
Remove NodeCollection pointer from Node class

med-ayssar:
The `Node` class does not need to know about the primitive `NodeCollection` object it belongs to. Therefore, the `NodeManange` must keep track of the mapping between a `Node` instance and the `NodeCollection` container.

**This is a derived requirement for the Vectorization project of the Node class.**


nestkernel/node_collection.cpp
```diff
@@ -268,6 +268,19 @@ NodeCollection::create( const std::vector< index >& node_ids_vector )
...
+NodeCollection::get_first() const
+{
+  return ( *begin() ).node_id;
+}
+
+index
+NodeCollection::get_last() const
+{
+  size_t offset = size() - 1;
+  return ( *( begin() + offset ) ).node_id;
```
med-ayssar, line 281: I could have done the following: `(*(end()  -1).node_id)`, but the `NodeCollection::iterator` does not implement the `operator-()` operator.

It might be my next PR to implement decrement operator.

heplesser (changes requested):
I am seriously concerned about the performance consequences of this change. For spatial connectivity, neurons need to know which node collection they belong to. I expect that this change will make lookup noticeably slower, even though `std::lower_bound()` has logarithmic complexity. We need to perform systematic benchmarks with spatial models before we proceed to avoid performance regressions on building.

I am also not happy with the mixed-camel-casing and would suggest `clear_node_collection_container` and similar.

med-ayssar:
> I am seriously concerned about the performance consequences of this change. For spatial connectivity, neurons need to know which node collection they belong to. I expect that this change will make lookup noticeably slower, even though `std::lower_bound()` has logarithmic complexity. We need to perform systematic benchmarks with spatial models before we proceed to avoid performance regressions on building.
> 
> I am also not happy with the mixed-camel-casing and would suggest `clear_node_collection_container` and similar.

The functions using the `get_nc` in the `spatial.cpp`  have as parameter an array of `IDs` (i.e, an `ArrayDatum` of IDs). In order to avoid the possible runtime complexity that I have introduced, we Can group the items in the `ArrayDatum` by the `NodeCollection` they belong to, without explicitly searching for the `NodeCollection`, and then from each group we extract a representative of the group and call the new implemented `get_nc` function in the `NodeManager` on the representative item.

med-ayssar:
@heplesser, I think last time you said that you might have been overthinking about the performance impact, but those new changes won't have any significant impact on the current performance.

Can we merge this PR, as I needed it for the new devices.

nestkernel/node_collection.cpp
heplesser: OK.

nestkernel/node_manager.cpp
```diff
@@ -308,6 +311,28 @@ NodeManager::add_music_nodes_( Model& model, index min_node_id, index max_node_i
   } // omp parallel
 }
 
+NodeCollectionPTR
+NodeManager::node_id_to_nodeCollection( const index node_id ) const
+{
+  auto it = std::lower_bound( nodeCollection_last_.begin(), nodeCollection_last_.end(), node_id );
```
heplesser, line 317: Please add a comment explaining the logic here. It took me some minutes to get it ...

heplesser (changes requested):
@med-ayssar Please see my comments inline. Also, the compiler seems to miss some things ;). My main concern is still performance. I will see if I can find some good benchmarks.

nestkernel/node_collection.h
```diff
@@ -798,6 +809,7 @@ NodeCollectionPrimitive::has_proxies() const
   return not nodes_have_no_proxies_;
 }
 
+
```
jougs, line 812:
```suggestion
```

nestkernel/node_manager.cpp
```diff
@@ -23,8 +23,10 @@
 #include "node_manager.h"
 
 // C++ includes:
+#include <algorithm>
 #include <set>
 
+
```
jougs, line 29:
```suggestion
```

nestkernel/node_manager.cpp
```diff
@@ -308,6 +311,37 @@ NodeManager::add_music_nodes_( Model& model, index min_node_id, index max_node_i
   } // omp parallel
 }
 
+NodeCollectionPTR
+NodeManager::node_id_to_node_collection( const index node_id ) const
+{
+  // find the largest element `e` in the node_collection_last_ so that `e` < node_id
```
jougs, line 317:
```suggestion
  // find the largest ID in node_collection_last_ that is still smaller than node_id
```

nestkernel/node_manager.cpp
```diff
@@ -308,6 +311,37 @@ NodeManager::add_music_nodes_( Model& model, index min_node_id, index max_node_i
...
+  auto it = std::lower_bound( node_collection_last_.begin(), node_collection_last_.end(), node_id );
+
+  // compute the position of the nodeCollection based on the position of `e`
```
jougs, line 320:
```suggestion
  // compute the position of the nodeCollection based on the position of the ID found above
```

nestkernel/node_manager.h
```diff
@@ -229,6 +229,20 @@ class NodeManager : public ManagerInterface
   bool have_nodes_changed() const;
   void set_have_nodes_changed( const bool changed );
 
+  /**
+   * @brief Map the the node ID to its original primitive NodeCollection object.
```
jougs, line 233:
```suggestion
   * @brief Map the node ID to its original primitive NodeCollection object.
```

nestkernel/node_manager.h
```diff
@@ -229,6 +229,20 @@ class NodeManager : public ManagerInterface
...
+   * @param node_id  node ID
```
jougs, line 234:
```suggestion
   * @param node_id The node ID
```

nestkernel/node_manager.h
```diff
@@ -229,6 +229,20 @@ class NodeManager : public ManagerInterface
...
+   * @return The primitive nodeCollection object containing the node ID that falls in [first, last)
+   */
+  NodeCollectionPTR node_id_to_node_collection( const index node_id ) const;
+
+  /**
+   * @brief Map the the node to its original primitive NodeCollection object.
+   * @param node  Node instance
```
jougs, line 241:
```suggestion
   * @brief Map the node to its original primitive NodeCollection object.
   * @param node The Node instance
```

nestkernel/node_manager.h
```diff
@@ -229,6 +229,20 @@ class NodeManager : public ManagerInterface
...
+   * @return The primitive nodeCollection object containing the node with node ID  falls in [first, last)
```
jougs, line 242:
```suggestion
   * @return The primitive NodeCollection object containing the node with node ID falls in [first, last)
```

nestkernel/node_manager.h
```diff
@@ -229,6 +229,20 @@ class NodeManager : public ManagerInterface
   bool have_nodes_changed() const;
   void set_have_nodes_changed( const bool changed );
 
+  /**
+   * @brief Map the the node ID to its original primitive NodeCollection object.
+   * @param node_id  node ID
+   * @return The primitive nodeCollection object containing the node ID that falls in [first, last)
```
jougs, line 235:
```suggestion
   * @return The primitive NodeCollection object containing the node ID that falls in [first, last)
```

nestkernel/node_manager.h
```diff
@@ -296,6 +318,14 @@ class NodeManager : public ManagerInterface
    */
   std::vector< SparseNodeArray > local_nodes_;
 
+  std::vector< NodeCollectionPTR > node_collection_container_; //!< a vector of the original/primitive NodeCollection
+
+  std::vector< index >
```
jougs, line 323:
```suggestion
  std::vector< size_t >
```

jougs (approved):
Happy if you merge all my suggestions.

jougs:
@heplesser: can you please re-review? Thanks!

med-ayssar:
@heplesser: I think it can be merged now :D

heplesser (approved):
Looks good now. I performed benchmarks small and large and did not find any noticeable slowdown.

nestkernel/node_collection.h
```diff
@@ -667,6 +678,20 @@ NodeCollection::set_metadata( NodeCollectionMetadataPTR )
...
+}
+
+inline size_t
+NodeCollection::get_last() const
+{
+  size_t offset = size() - 1;
+  return ( *( begin() + offset ) ).node_id;
+}
+
+
```
jougs, line 694: I totally realize that it is bad style if
1. one of the original reviewers (me) 
2. comments on a merged pull request
3. by a person not working on NEST anymore

:see_no_evil:

*But*: couldn't this have been done like this?

```suggestion
inline size_t
NodeCollection::get_first() const
{
  return begin()->node_id;
}

inline size_t
NodeCollection::get_last() const
{
  return ( end() - 1 )->node_id;
}
```

Just sayin'...

med-ayssar: That was not that simple or better saying, things might get more ugly.

The `begin` functions returns an `iterator`, and this iterator is a bit complex, as it hides the distinction between the *primitive* and *composite* types. Thus, `begin()->node_id` won't work and also the `operator-()` is not implemented.

PR #2602, opened 2023-02-02
Fix the patch application for the documentation (again)

steffengraber:
During the test of current master, the error `name 'subprocess' is not defined` occurred. This is fixed and at the same time "import mock" is removed, which is there no longer needed.

@terhorstd  Sorry for the late fix. Could you please have a look at this?

### DIFF
--- a/doc/htmldoc/conf.py
+++ b/doc/htmldoc/conf.py
@@ -29,6 +29,7 @@ from pathlib import Path
 from shutil import copyfile
 import json
 
+import subprocess
 from subprocess import check_output
 
 
--- a/doc/htmldoc/conf.py
+++ b/doc/htmldoc/conf.py
@@ -29,8 +29,9 @@ from pathlib import Path
 from shutil import copyfile
 import json
 
-from subprocess import check_output, CalledProcessError
-from mock import Mock as MagicMock
+import subprocess
+from subprocess import check_output
+
 
 
 source_dir = os.environ.get('NESTSRCDIR', False)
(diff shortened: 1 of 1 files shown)


doc/htmldoc/conf.py
```diff
@@ -29,8 +29,8 @@
 from shutil import copyfile
 import json
 
-from subprocess import check_output, CalledProcessError
-from mock import Mock as MagicMock
+import subprocess
+from subprocess import check_output
```
terhorstd, line 33: Why should both be necessary?! I'd expect explicit imports for all, or global import and all uses prefixed.

terhorstd (changes requested):

doc/htmldoc/conf.py
```diff
@@ -378,14 +376,14 @@ def patch_documentation(patch_url):
     print("Preparing patch...")
     try:
         git_dir = str(source_dir / ".git")
-        git_hash = check_output(f"GIT_DIR={git_dir} git rev-parse HEAD", shell=True, encoding='utf8').strip()
+        git_hash = subprocess.check_output(f"GIT_DIR={git_dir} git rev-parse HEAD", shell=True, encoding='utf8').strip()
```
terhorstd, line 379: Minor change to avoid a mess with badly named dirs.
```suggestion
        git_hash = subprocess.check_output(f"GIT_DIR='{git_dir}' git rev-parse HEAD", shell=True, encoding='utf8').strip()
```

terhorstd (approved):
With the minor change this seems good to go.

PR #2604, opened 2023-02-03
Avoid using Dynamic_cast

med-ayssar:
The `Model` class provides a `type_id` field that can be used for applying type checking on the `Node` instance, and avoid using the `dynamic_cast`.

### DIFF
--- a/models/stdp_dopamine_synapse.cpp
+++ b/models/stdp_dopamine_synapse.cpp
@@ -31,6 +31,8 @@
 // Includes from sli:
 #include "dictdatum.h"
 
+#define ClassToString( cls ) #cls
+
 namespace nest
 {
 //
@@ -82,10 +84,16 @@ STDPDopaCommonProperties::set_status( const DictionaryDatum& d, ConnectorModel& 
   long vtnode_id;
   if ( updateValue< long >( d, names::vt, vtnode_id ) )
   {
-    const thread tid = kernel().vp_manager.get_thread_id();
-    Node* vt = kernel().node_manager.get_node_or_proxy( vtnode_id, tid );
-    vt_ = dynamic_cast< volume_transmitter* >( vt );
-    if ( not vt_ )
+    std::string model_name = ClassToString( volume_transmitter );
+    const index model = kernel().modelrange_manager.get_model_id( vtnode_id );
+    const index current_model_type_id = kernel().model_manager.get_node_model( model )->get_type_id();
+    const index expected_model_type_id = kernel().model_manager.get_node_model_id( model_name );
+    if ( current_model_type_id == expected_model_type_id )
+    {
+      const thread tid = kernel().vp_manager.get_thread_id();
+      vt_ = static_cast< volume_transmitter* >( kernel().node_manager.get_node_or_proxy( vtnode_id, tid ) );
+    }
+    else
     {
       throw BadProperty( "Dopamine source must be volume transmitter" );
     }


JanVogelsang (approved):
Apart from the fact that the whole`modelrange_manager` design is quite an antipattern in itself, this change actually makes sense.

models/stdp_dopamine_synapse.cpp
```diff
@@ -82,10 +84,16 @@ STDPDopaCommonProperties::set_status( const DictionaryDatum& d, ConnectorModel&
   long vtnode_id;
   if ( updateValue< long >( d, names::vt, vtnode_id ) )
   {
-    const thread tid = kernel().vp_manager.get_thread_id();
-    Node* vt = kernel().node_manager.get_node_or_proxy( vtnode_id, tid );
-    vt_ = dynamic_cast< volume_transmitter* >( vt );
-    if ( not vt_ )
+    std::string model_name = ClassToString( volume_transmitter );
```
clinssen, line 87: Why is the macro necessary; why can't we just use the const string literal ``"volume_transmitter"``?

JanVogelsang: Now that you say it, using the macro is actually incorrect here, as the name of the class does not have to match the name of the model. This works as a workaround, but what we would actually need would be a way to retrieve the type_id of a model from the type, so a function of the form:
`index ModelManager::get_type_id<volume_transmitter>();`
However, in this case, we could just assume that `volume_transmitter` is also registered as `"volume_transmitter"`, even though this isn't the cleanest way and might break the code if the registration was to be changed.

models/stdp_dopamine_synapse.cpp
```diff
@@ -31,6 +31,8 @@
 // Includes from sli:
 #include "dictdatum.h"
 
+#define ClassToString( cls ) #cls
```
heplesser, line 34: How does this macro work?

heplesser (commented):
I do not understand the purpose of this change. Why do you not want to use the dynamic cast? Isn't that C++'s built-in way of testing whether a pointer points to a given type? Or is there a problem when applying `CopyModel` to a volume transmitter?

models/stdp_dopamine_synapse.cpp
JanVogelsang: I think it's just a matter of performance, but now that Charl mentioned the problem with these changes, I don't think this change makes sense anymore, even if it would change performance at all. Not that this part of the code would be performance-critical/-relevant at all.

JanVogelsang: The macro simply takes the input and interprets it as the actual characters that were given to that macro. So all it does is change `volume_transmitter` to `"volume_transmitter"`.

med-ayssar: If we can apply certain checks if the underlying `Node` object can be converted to `volume_transmitter`, we can avoid using the `dynamic_cast` and therefore avoid the runtime overhead used for the conversion.

JanVogelsang: You would have to perform some profiling, though. All the overhead required for the manual type-checking might actually outweigh the overhead of the `dynamic_cast`, but calling the RTTI might actually not consume that much more time. The branching will be what costs us a lot of cycles as well, and the if will still persist with your solution. So in total there might not even be an advantage of doing manual type-checking here.

PR #2606, opened 2023-02-06
Convert `node_id` to `Node` only when necessary

med-ayssar:


JanVogelsang:
@med-ayssar What is the current state of this PR? If you give me a short summary of the changes and what is left to finish this PR, I am absolutely willing to continue this one for you.

med-ayssar:
> @med-ayssar What is the current state of this PR? If you give me a short summary of the changes and what is left to finish this PR, I am absolutely willing to continue this one for you.

In some cases, the `NodeId` and `Node*` are not used correctly. You would always see the pattern that a `NodeId` was converted to `Node*` and the pointer was simply passed to the function without calling any function on the node, but instead they call the `Node::get_node_id`. That pattern was so annoying, and I was trying to remove some of them, but it was happening almost everywhere, and it was even more annoying to individually decide if the use of that pattern was correct or not.

JanVogelsang:
Assigned @heplesser to check if it should be merged or if more work is needed here to turn this into an actual PR.

terhorstd:
Need to also look at performance effects.

PR #2611, opened 2023-02-09
Make get_thread virtual in Node class

med-ayssar:
The `get_thread` function in the `Node` class is not virtual, which makes the implementation of the function in the `ProxyNode` class useless. Mainly, the `get_thread` function in the `ProxyNode` class will never be called, as we always have a pointer of `Node` type pointing to the proxy instance.

My suggested change is to make the function virtual and remove the implementation from the `ProxyNode`.

### DIFF
--- a/nestkernel/node.h
+++ b/nestkernel/node.h
@@ -723,7 +723,7 @@ public:
   /**
    * Retrieve the number of the thread to which the node is assigned.
    */
-  thread get_thread() const;
+  virtual thread get_thread() const;
 
   /**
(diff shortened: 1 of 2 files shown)


jougs:
As discussed offline, this changes are problematic in that they
1. add a vtable indirection for all calls to `get_thread()`, which might lead to a performance regression
2. shadow the fact that `get_thread()` must not be called on a `proxynode` as it cannot return a sensible value for remote nodes represented by it.

I will thus close it.

PR #2612, opened 2023-02-13
Refactor `using Node::handle…` in Model declarations

med-ayssar:
In the `models` directory, all models are **re-importing** the virtual functions from the `Node` class, but it is not really necessary as we are not [ **hiding**](https://stackoverflow.com/questions/6727087/c-virtual-function-being-hidden) those functions.

In the `kernel_manager::create_kernel_manager()` function, we are using the OpenMP `critical` construct, even though the function is not being called in a **parallel region**.

### DIFF
--- a/models/ac_generator.h
+++ b/models/ac_generator.h
@@ -121,8 +121,6 @@ public:
 
   port send_test_event( Node&, rport, synindex, bool ) override;
 
-  using Node::handle;
-  using Node::handles_test_event;
 
   void handle( DataLoggingRequest& ) override;
 
--- a/models/aeif_cond_alpha.h
+++ b/models/aeif_cond_alpha.h
@@ -195,13 +195,6 @@ public:
   aeif_cond_alpha( const aeif_cond_alpha& );
   ~aeif_cond_alpha() override;
 
-  /**
-   * Import sets of overloaded virtual functions.
-   * @see Technical Issues / Virtual Functions: Overriding, Overloading, and
-   * Hiding
-   */
-  using Node::handle;
-  using Node::handles_test_event;
 
   port send_test_event( Node&, rport, synindex, bool ) override;
 
--- a/models/aeif_cond_alpha_multisynapse.h
+++ b/models/aeif_cond_alpha_multisynapse.h
@@ -186,13 +186,6 @@ public:
 
   friend int aeif_cond_alpha_multisynapse_dynamics( double, const double*, double*, void* );
 
-  /**
-   * Import sets of overloaded virtual functions.
-   * @see Technical Issues / Virtual Functions: Overriding, Overloading, and
-   * Hiding
-   */
-  using Node::handle;
-  using Node::handles_test_event;
 
   port send_test_event( Node&, rport, synindex, bool ) override;
 
--- a/models/aeif_cond_beta_multisynapse.h
+++ b/models/aeif_cond_beta_multisynapse.h
@@ -189,13 +189,6 @@ public:
 
   friend int aeif_cond_beta_multisynapse_dynamics( double, const double*, double*, void* );
 
-  /**
-   * Import sets of overloaded virtual functions.
-   * @see Technical Issues / Virtual Functions: Overriding, Overloading, and
-   * Hiding
-   */
-  using Node::handle;
-  using Node::handles_test_event;
 
   port send_test_event( Node&, rport, synindex, bool ) override;
 
--- a/models/aeif_cond_exp.h
+++ b/models/aeif_cond_exp.h
@@ -196,13 +196,6 @@ public:
   aeif_cond_exp( const aeif_cond_exp& );
   ~aeif_cond_exp() override;
 
-  /**
-   * Import sets of overloaded virtual functions.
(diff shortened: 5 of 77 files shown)


terhorstd:
Could you break the "omp critical" change into a different PR? It seems to need some thinking and might raise more discussion than the removal of "using" directives would require.

med-ayssar:
> Could you break the "omp critical" change into a different PR? It seems to need some thinking and might raise more discussion than the removal of "using" directives would require.

Sure!

heplesser (changes requested):
The `using` directives are in place to ensure correct resolution of overloaded overridden functions. I am digging for documentation on this and will get back with it shortly.

med-ayssar:
> The `using` directives are in place to ensure correct resolution of overloaded overridden functions. I am digging for documentation on this and will get back with it shortly.

I might be missing some background information about **overload resolution**, but to my current understanding the function `handle_test_event` is overloaded `n` times, with `n` being the number of the different supported events in NEST.  The implemented events only have a common base `Event` class, and there is no defined conversion between them.

In the following example, the compiler will generate the following error: ` error: call of overloaded ‘test(int)’ is ambiguous`

```cpp
void test(char x){
    std::cout<< "calling test function with char argument";
}

void test (long x) {
     std::cout<< "calling test function with long argument";
}


int main() {
 test(42); // ambiguous ( compile error )
}
```
Because `int` will be implicitly converted to `char` and `long` and they belong to the same `numeric conversion` ranking order and therefore there is a tie between the set of candidates.

However, in `Event` case in the `Node` class, we have the following:

```cpp

class Event {
    
};

class SpikeEvent: public Event {
    
};

class OtherEvent : public Event {
    
};



void test(SpikeEvent x){
    std::cout<< "calling test function with SpikeEvent argument";
}

void test (OtherEvent x) {
     std::cout<< "calling test function with OtherEvent argument";
}

int main()
{
    test(SpikeEvent());
    return 0;
}

```

There is no direct relation between the `SpikeEvent` class and the `OtherEvent` class, and the compiler will choose the function with `SpikeEvent` argument.

heplesser:
Please find attached my 2008 memo on overriding and overloading virtual functions including the example program mentioned in the memo (zipped for technical reasons).

[VirtualOverrideOverload_HEP_2008.pdf](https://github.com/nest/nest-simulator/files/10768651/VirtualOverrideOverload_HEP_2008.pdf)
[vol.cpp.zip](https://github.com/nest/nest-simulator/files/10768655/vol.cpp.zip)

med-ayssar:
> request

Thank you for the document, but I don't see that we are hiding functions here. The Node class provides all variants of the function with respect to the Event type. So there is exact perfect match between the function in the derived class and the child class. 

In your example, only primitive types are used to illustrate the problem, as there is implicit conversion between these types, but for the event types we are using, there is no such relation.

med-ayssar:
@heplesser, any comments or opinions regarding my last argument?

heplesser:
@med-ayssar Here is a modified example where I have introduced classes `Ea` and `Eb` derived from `E` as arguments to be closer to what we do in NEST. Note the last case which will trigger a compiler error because `using B::g` is missing. The error occurs as soon as one calls `g(const Eb&)` directly through an instance of the derived class `D`. Since we cannot know in principle that this will never happen for methods in model classes, we should write proper code for the model classes with the `using` directives in place.

```c++
#include <iostream>
#include <iomanip>
#include <string>

using namespace std;

class E {
};

class Ea : public E {
};

class Eb : public E {
};

class B {
public:
  virtual ~B() {}

  virtual string f(const Ea&)  { return "B::f(Ea&)"; }
  virtual string f(const Eb&)  { return "B::f(Eb&)"; }
  
  virtual string g(const Ea&)  { return "B::g(Ea&)"; }
  virtual string g(const Eb&)  { return "B::g(Eb&)"; }
};

class D : public B {
public:
  using B::f;
  string f(const Ea&) override { return "D::f(Ea&)"; }

  string g(const Ea&) override { return "D::g(Ea&)"; }
};

int main()
{
  D d;
  B* bd = new D;

  Ea ea;
  Eb eb;

  // The following all work correctly
  cout << "d.f(ea): " << d.f(ea) << endl;
  cout << "d.f(eb): " << d.f(eb) << endl;
  cout << "d.g(ea): " << d.g(ea) << endl;
  cout << endl;
  cout << "bd->f(ea): " << bd->f(ea) << endl;
  cout << "bd->f(eb): " << bd->f(eb) << endl;
  cout << "bd->g(ea): " << bd->g(ea) << endl;
  cout << "bd->g(eb): " << bd->g(eb) << endl;

  // The next leads to a compiler error because "using B::g;" is missing in
  // declaration of D, so overriding g(const Ea&) in D hides B::g(const Eb&).
  // cout << "d.g(eb): " << d.g(eb) << endl;  
  
  return 0;
}
```

med-ayssar:
Yeah, I totally understand the problem, but we are currently only having pointers to `Node` (in the example `B`), so those hidden function won't be really hidden, but then again for completeness, we probably need to make all these functions visible to the SubClass, and therefore I will close this PR.

PR #2615, opened 2023-02-18
Correctly build connection infrastructure in GetConnections also when secondary events in use

### LINKED ISSUE #2614: GetConnections crashes simulation with gap junctions
Adding a `<< >> GetConnections` call between the final `Connect` and the `Simulate` calls in a minimal simulation with gap junctions (), leads in most runs to a segmentation fault or in some runs to NaN values for recorded membrane potential. No problems occur for a corresponding network without gap junctions, so it seems related to configuration of gap junction connection data structures. The same works fine in nest-2.14.1.

heplesser:
This PR fixes #2614 with a straightforward solution: `update_delay_extrema_()` and `check_wfr_use()` are now called before `update_connection_infrastructure()` also when the connection infrastructure is built by `get_connections()`.  This is necessary because if no nodes or connections are added between a `GetConnections` and a `Simulate` call, the infrastructure built by `get_connections()` is used. This requires information about the data size in secondary events, which again depends on `min_delay`.

PEP8 fixes in two unrelated files because they were detected by check_code_style.

This entire process needs to be refactored at a later stage, this is just a bug fix.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -976,6 +976,11 @@ nest::ConnectionManager::get_connections( const DictionaryDatum& params )
   // as this may involve sorting connections by source node IDs.
   if ( connections_have_changed() )
   {
+    // check whether waveform relaxation is used on any MPI process;
+    // needs to be called before update_connection_intrastructure_since
+    // it resizes coefficient arrays for secondary events
+    kernel().node_manager.check_wfr_use();
+
 #pragma omp parallel
     {
       const thread tid = kernel().vp_manager.get_thread_id();
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -981,7 +981,7 @@ nest::ConnectionManager::get_connections( const DictionaryDatum& params )
     update_delay_extrema_();
 
     // Check whether waveform relaxation is used on any MPI process;
-    // needs to be called before update_connection_intrastructure_since
+    // needs to be called before update_connection_infrastructure since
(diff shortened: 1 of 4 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -976,6 +976,15 @@ nest::ConnectionManager::get_connections( const DictionaryDatum& params )
   // as this may involve sorting connections by source node IDs.
   if ( connections_have_changed() )
   {
+    // We need to update min_delay because it is used by check_wfr_use() below
+    // to set secondary event data size.
+    update_delay_extrema_();
+
+    // Check whether waveform relaxation is used on any MPI process;
+    // needs to be called before update_connection_intrastructure_since
```
nicolossus, line 984:
```suggestion
    // needs to be called before update_connection_infrastructure since
```

nicolossus (approved):
Thanks for the bug fix! Besides the minor typo, this looks good to me.

testsuite/unittests/test_hh_psc_alpha_gap.sli
```diff
@@ -85,6 +85,9 @@ neuron1 neuron2
 Connect
 vm neuron2     1.0 h Connect
 
+% The following call to GetConnections makes Simulate crash
+% or generate results containing NaNs, see #2614.
+<< >> GetConnections ;
```
med-ayssar, line 90: I think that it is better to create a specific test for this case rather than inserting the `GetConnection` without relay using the returned connections.

med-ayssar (approved):

PR #2617, opened 2023-02-22
Separate event transmission and event delivery for better performance in high-thread situations

heplesser:
This is a draft PR to integrate the revision of event transmission from the "5G" approach to close to what we had in "4G" (until NEST 2.14). Briefly, the events are now gathered at the end of each update loop and MPI-communicated, and then delivered at the beginning of the next update loop.

At present, all kinds of events are supported again (SpikeEvent on and off grid, SecondaryEvent), but test_mini_brunel_ps.sli fails if NEST is compiled with GCC 12. When compiling with Clang 13, the test passes. This indicates most likely some incorrect memory access. Failure is deterministic with spike times depending on the mpi-thread split. Results are correct for a single MPI process (four threads).

To get more insight into where things go wrong, I attempted to build NEST with

`-D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC -fsanitize=address`

As long as the `*DEBUG*` options were in place, g++ emits error messages concerning overridden deleted functions in the `SecondaryEvent` class hierarchy, which left me entirely confused, so I dropped them. With just `-fsanitize=address`, NEST builds but segfaults on start with a stack trace indicating that the problem occurs even before NEST itself starts. @med-ayssar and @JanVogelsang, could you two take a look at this?

Once this problem is solved, a number of this remain. I will comment on them later in this issue.

I consciously add many reviewers just to make sure you are in the loop.


ikitayama:
Is this series going to be tested on a big system?

JanVogelsang:
We are currently preparing various benchmarks to be run on our local cluster, Jureca, and Juwels. The benchmarks will cover simulations of all scales, including simulations of very large networks which can only be run on big systems. If the PR turns out to improve the performance on the abovementioned systems, benchmarks on additional systems might be performed as well. Do you have any particular system in mind already?

ikitayama:
> We are currently preparing various benchmarks to be run on our local cluster, Jureca, and Juwels. The benchmarks will cover simulations of all scales, including simulations of very large networks which can only be run on big systems. If the PR turns out to improve the performance on the abovementioned systems, benchmarks on additional systems might be performed as well. Do you have any particular system in mind already?

Fugaku - if you happen to have access to it, would you mind running this against it please?

heplesser:
I have made some progess with respect to the bug mentioned above. The different behavior with GCC (bug) and Clang (no bug) was due to the fact that the two compilers use different random deviate generation methods, so that different connection patterns were generated. If one uses the connection pattern from GCC on Clang, the bug occurs for Clang as well. The attached script (rename to ....sli), is a fully deterministic minimal reproducer. For the GCC-generated connection pattern, it produces different spike times with two than with one MPI process (in total four VPs). The master branch produces identical results with one and two MPI processes and passes the test.

[def_bug.txt](https://github.com/nest/nest-simulator/files/10819280/def_bug.txt)

heplesser:
I have narrowed the bug further down to a single spike transmission that seems to go wrong. Further analysis soon.

JanVogelsang:
@ikitayama Sure, we will eventually come back to you on this!

heplesser:
Closing this PR. It is superseeded by #2926.

PR #2616, opened 2023-02-21
Removing well-proven and therefore outdated asserts

JanVogelsang:
The asserts in the `Node::update` method of all models derived from `Node` were checking if the parameters for the requested update interval are indeed correct. These asserts are now generally seen as superfluous, as they have not been triggered in many years now.

PR #2618, opened 2023-02-22
Replace NEST typedefs with plain C++ datatypes

### LINKED ISSUE #2541: Replace typedef'ed types with C++ PODs
This issue is a follow-up to #2215.

After a thorough discussion during the Hackathon we came to the following conclusions:

1. Typedefs (except `weight` and partially `delay`) are only used to index into containers. Therefore, `size_t` is the suitable POD where currently `typedef`s  `index`, `thread`, `port`, `rport`, ... are used.
2. Typedefs do not provide any type safety in C++.
3. Where we use signed types through `typedef`s today (`int`,  `long`), we only use value -1 from the negative range, wasting value range. Use of `size_t` would avoid this.
4. [Named type solutions](https://embeddedartistry.com/blog/2018/04/23/namedtype-the-easy-way-to-use-strong-types-in-c/) seem too complicated, as do implementing our own classes for, e.g., `Thread`, `Delay`, ....
5. We therefore concluded that we will explore replacing all `typedef`ed types with PODs, using unsigned PODs throughout.
6. For marking invalid values, three approaches can be used, as suitable in the situation
    1. Return a `std::pair<>` combining the value and a boolean flag indicating validity.
    2. Use `std::numeric_limits<POD>::max()` as the invalid value.
    3. Raise exceptions instead of returning invalid values. 
7. Certain quantities have implicitly limited value ranges because they need to be transmitted in bitfields of limited size in MPI communication (https://github.com/nest/nest-simulator/blob/e49be2ae64b92b6b6d7c451b2ed8027c0975e480/nestkernel/nest_types.h#L81-L105). Where these values are set (e.g., when the number of threads is changed or new synapse models are created), we need to check if new values are in the permissible range.

pnbabu:
This PR fixes #2541. It replaces the following typedefs 
- `index`, `thread`, `port`, `rport` with `size_t`
- `weight` with `double`
- `delay` with `long`


nestkernel/nest_types.h
```diff
@@ -145,62 +141,23 @@ const synindex invalid_synindex = MAX_SYN_ID;
...
 
 /**
- * Connection port number to distinguish outgoing connections.
- * Connections between Nodes are assigned port numbers.
- * Valid port numbers start at zero (0).
- * The value -1 is used for invalid or unassigned ports.
+ * Value for invalid connection thread id.
  */
-typedef long port;
+const size_t invalid_thread = std::numeric_limits< size_t >::max();
```
JanVogelsang, line 149:
```suggestion
constexpr size_t invalid_thread = std::numeric_limits< size_t >::max();
```

nestkernel/nest_types.h
```diff
@@ -145,62 +141,23 @@ const synindex invalid_synindex = MAX_SYN_ID;
...
-/**
- * Weight of a connection.
- * Connections have a weight which is used to scale the influence
- * of an event.
- * A weight of 0 should have the same influence on the receiving node
- * as a non-existing connection. Otherwise, there is no default range for
- * connection weights.
- */
-typedef double weight;
+const size_t invalid_port = std::numeric_limits< size_t >::max();
```
JanVogelsang, line 154:
```suggestion
constexpr size_t invalid_port = std::numeric_limits< size_t >::max();
```

nestkernel/nest_types.h
```diff
@@ -145,62 +141,23 @@ const synindex invalid_synindex = MAX_SYN_ID;
...
- * Delay of a connection.
- * The delay defines the number of simulation steps which elapse
- * before an Event arrives at the receiving Node.
- * Delays must be equal or larger than one.
+ * Values for min and max delay.
  */
-typedef long delay;
-const long delay_min = LONG_MIN;
-const long delay_max = LONG_MAX;
+const long delay_min = std::numeric_limits< long >::min();
```
JanVogelsang, line 159:
```suggestion
constexpr long delay_min = std::numeric_limits< long >::min();
```

nestkernel/nest_types.h
```diff
@@ -145,62 +141,23 @@ const synindex invalid_synindex = MAX_SYN_ID;
...
+const long delay_max = std::numeric_limits< long >::max();
```
JanVogelsang, line 160:
```suggestion
constexpr long delay_max = std::numeric_limits< long >::max();
```

nestkernel/nest_types.h
```diff
@@ -125,11 +125,7 @@ const tic_t tic_t_min = LONG_MIN;
 /**
  *  Unsigned long type for enumerations.
  */
-typedef size_t index;
-#ifndef SIZE_MAX
-#define SIZE_MAX ( static_cast< std::size_t >( -1 ) )
-#endif
-__attribute__( ( __unused__ ) ) const index invalid_index = SIZE_MAX;
+__attribute__( ( __unused__ ) ) const size_t invalid_index = std::numeric_limits< size_t >::max();
```
JanVogelsang, line 128:
```suggestion
__attribute__( ( __unused__ ) ) constexpr size_t invalid_index = std::numeric_limits< size_t >::max();
```

nestkernel/nest_types.h
```diff
@@ -145,62 +141,23 @@ const synindex invalid_synindex = MAX_SYN_ID;
 //! target index into thread local node vector
 typedef unsigned short targetindex;
```
JanVogelsang, line 142: Shouldn't we get rid of all typedefs here? There are 3 left which we haven't discussed yet:  
`synindex`: Currently `unsigned int`, but only gets 9 Bits when sent over MPI. Primarily used as index, should therefore become `size_t` as all places where memory matters bitfields are used anyway.  
`targetindex`: Currently `unsigned short` and not sent over MPI. Here, size actually matters and should therefore either keep `unsigned short` or could be changed to `size_t` together with a bitfield (16b) in the targetidentifier. The first would require an implicit conversion from `unsigned short` to `size_t` while the latter would require some bitmasking and zero-padding to get the actual `size_t` value. Optimized compiled code will probably do the exact same for both, so I think using `size_t` with a bitfield ist actually cleaner here to indicate that this is indeed an index.  
`tic_t`: Unsure about this one, this might actually be a valid typedef.

JanVogelsang (commented):
These constant values here are fixed at compile time, thus using `constexpr` is a good thing to do here.  
Apart from that, everything looks right to me.

JanVogelsang:
Edit: I didn't check yet if you missed any instances. Jochen, can you check that while you are reviewing?

nestkernel/nest_types.h
jougs: My take on this was that as these `typedef`s are actually descriptive names for something and thus they could stay as is. Also, I thought, if we want to change their definition, it would be easier if they were defined at a central point. If there is no danger to increase memory requirements by changing them, I might reconsider.

@suku248:  What is your opinion on this?

JanVogelsang: The `targetindex` is only ever used in a single file (*target_identifier.h*), so I think instead of a typedef at a central point, one might just change the type directly in that file (or use a local typedef there). Of course, having it at a central point raised the awareness to that typedef and somewhat defines it as a parameter which could be changed if desired. However, changing the type without thinking about it carefully might increase the memory usage much more than expected (due to alignment).  
For the `synindex` we already have the `NUM_BITS_SYN_ID` which controls the size at all places where size actually matters.  
So my conclusion would be that it is safe to remove these typedefs. But I am quite interested in Susanne's opinion as well.

heplesser: `TargetIdentifierIndex::target_` must be 16 bit and not longer to keep HPC-Synapse small. Changing the type to `size_t` would increase memory requirement for these synapses by at least 48 bit (see https://github.com/nest/nest-simulator/blob/4192435db972ce333e6a33e5769407fd9d8c3b06/nestkernel/connection.h#L308). So the underlying type must remain `unsigned short` and I would argue to use that explicitly instead of `typedef'ing`.

models/ac_generator.cpp
```diff
@@ -201,7 +201,7 @@ nest::ac_generator::pre_run_hook()
 void
 nest::ac_generator::update( Time const& origin, const long from, const long to )
 {
-  assert( to >= 0 and ( delay ) from < kernel().connection_manager.get_min_delay() );
+  assert( to >= 0 and ( long ) from < kernel().connection_manager.get_min_delay() );
```
heplesser, line 204: Since `get_min_delay()` now returns `long`, we don't need the cast here, which anyways does not make sense since `from` is already `long` (and if we needed it, it should be a proper `static_cast< long >(...)`. But here we can reduce to
```suggestion
  assert( to >= 0 and from < kernel().connection_manager.get_min_delay() );
```
The same applies for all similar lines, which are actually removed by #2616. I suggest to wait for that PR before doing anything more here.

heplesser:
@pnbabu I have just merged #2616, which removes all the from-to asserts at the beginning of `update()` in models. This has unfortunately caused a lot of conflicts, but they should be straightforward to resolve by accepting the removal of the asserts.

nestkernel/nest_types.h
JanVogelsang: Actually you are right, thought about it again and we musn't chang ethe type. Using 32b would be fine, as we have 16b offset at the moment anyway (as the SynIdDelay forces a 32b alignment), but 64b (size_t) would increase the memory footprint.  

```cpp
typedef unsigned short targetindex;
targetindex target;
```
will not be the same as
```cpp
size_t target : 16;
```
which is what I first thought. But actually, the bitfields will always allocate the size of the largest member type (i.e., size_t), even if only using 16b of that type. The TargetIdentifierIndex will thus have a size of 8B instead of the 2B it currently has. Thanks Hans Ekkehard!  

Still, then we should remove the typedef and use the `unsigned short` explicitly as you suggested.

models/ac_generator.cpp
JanVogelsang: Actually, we shouldn't touch the casts in this PR, but in #2506 which is on hold until this PR is merged, right?

JanVogelsang: Sure, but there (probably) are many more such casts which were not removed in #2616.

med-ayssar:
@pnbabu any updates?

heplesser:
@pnbabu This looks good now overall, but having just read Stroustrup's "Tour", I am wondering about replacing so much with `size_t` (I know I argued for it before). The Core Guidelines don't say anything explicitly about `size_t`, but generally favour signed types (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Res-nonnegative) and point to suggest `gsl::index` for many cases where we currently use `size_t`, see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Res-subscripts. This leaves me wondering a bit. @jougs what to you think?

JanVogelsang:
Looks good.  
@heplesser Interesting, a few weeks ago I was asking myself the same question, as in university we got taught to use `int` for indexing, but personally I preferred `size_t` as well, because it seemed more descriptive. The core guidelines' arguments are valid, but I would have liked more depth, as I can't quite find anything regarding optimization/performance of unsigned vs signed types. I could only find an argument _for_ unsigned types, as those can be divided by 2 faster than signed ones, but that's about it. And I am not sure why signed types enable better error detection either..  
And apart from that, I don't really see an advantage of `ptrdiff_t` over `size_t`. In our code, we use `size_t` strictly for indexing or memory sizes, both of which can only be positive and will never be used in any operation that could return a negative value (or can you think of any?). And as we heavily rely on the STL, in my opinion, it makes more sense to stick with their "native" type.

models/aeif_cond_alpha_multisynapse.cpp
```diff
@@ -567,10 +567,10 @@ aeif_cond_alpha_multisynapse::update( Time const& origin, const long from, const
   } // for-loop
 }
 
-port
-aeif_cond_alpha_multisynapse::handles_test_event( SpikeEvent&, rport receptor_type )
+size_t
+aeif_cond_alpha_multisynapse::handles_test_event( SpikeEvent&, size_t receptor_type )
 {
-  if ( receptor_type <= 0 or receptor_type > static_cast< port >( P_.n_receptors() ) )
+  if ( receptor_type <= 0 or receptor_type > static_cast< size_t >( P_.n_receptors() ) )
```
jougs, line 573: See below.

```suggestion
  if ( receptor_type <= 0 or receptor_type > P_.n_receptors() )
```

models/iaf_psc_exp_multisynapse.cpp
```diff
@@ -344,10 +344,10 @@ iaf_psc_exp_multisynapse::update( const Time& origin, const long from, const lon
   }
 }
 
-port
-iaf_psc_exp_multisynapse::handles_test_event( SpikeEvent&, rport receptor_type )
+size_t
+iaf_psc_exp_multisynapse::handles_test_event( SpikeEvent&, size_t receptor_type )
 {
-  if ( receptor_type <= 0 or receptor_type > static_cast< port >( P_.n_receptors_() ) )
+  if ( receptor_type <= 0 or receptor_type > static_cast< size_t >( P_.n_receptors_() ) )
```
jougs, line 350: See above, and also change everywhere else, where `P_.n_receptors_()` is used in a similar way.
```suggestion
  if ( receptor_type <= 0 or receptor_type > P_.n_receptors_() )
```

jougs (commented):
Mostly fine by me. Just one repetitive pattern where I think the `static_cast` is not needed.

jougs (approved):

PR #2621, opened 2023-02-24
Remove `get_thread()` function from proxynode class

med-ayssar:
When creating a `Node` instance, the `thread` property of the node must be initialized to `invalid_thread`, and then the `NodeManager` should be responsible for assigning the correct thread `ID` to the node instance.

Therefore and by construction, the `Proxy` node will always be initialized with an `invalid_thread` ID.


heplesser (approved):
Looks fine.

PR #2622, opened 2023-02-24
Add `spike_train_injector` model

nicolossus:
This PR adds a new model named ~`static_injector_neuron`~ `spike_train_injector` and is a replacement for the closed https://github.com/nest/nest-simulator/pull/2591. 

The spike train injector simply emits spikes at prescribed spike times which are given as an array. Incoming spikes will be ignored. The spike train injector behaves similarly to a spike generator, but is treated internally as a neuron and not a device. Unlike a spike generator which is replicated at each virtual process, the spike train injector (neuron) resides on a single virtual process.

The spike train injector is meant to be used in multithreaded simulations where spike emitting neurons in a somewhat  large external population are modeled on an individual-basis. In this case, modeling the external neurons as `spike_generator`s will bloat memory consumption with replication at each virtual process. This will not be the case if the external neurons are modeled as `spike_train_injector`s.

The name of the new model was open to discussion.

## Poll for model name -- CLOSED

React to this comment with the emoticon that matches your favorite name or suggest another name to be added as an option in a comment.

* `injector_neuron`
  * Signals that the model injects spikes. However, this mechanism is inherent to neuron models in general. As pointed out be @clinssen below, the plain use of `injector` here opens for interpreting this as a current injecting mechanism. 
  * Vote with :thumbsup:
* `static_injector_neuron`
  * Signals that the model injects spikes and cannot be changed, i.e., a neuron just emitting prescribed spikes. This name is currently used. The comment about `injector` above also applies here.
  * Vote with :smile:
* `spike_injector`
  * Signals explicitly that this is a spike injecting mechanism to avoid confusion with current injection. Does not signal that the injected spikes must be pre-defined (deterministic).
  * Vote with :tada:
* `static_spike_injector` 
  * Same as above but signals that the mechanism cannot be altered once defined.  
  * Vote with :hearts: 
* `spike_train_injector`
  * Clearly signals that  (prescribed) spike trains are injected. 
  * Vote with :rocket:
* `spike_generator_neuron`
  * The model behaves similar to a `spike_generator` but is treated as a neuron internally, which is reflected by this name. However, `<name>_generator` is a naming scheme associated with devices, and using `spike_generator_neuron` might obfuscate that this model is not a device.
  * Vote with :eyes:
* `global_spike_generator`
  * Since there are no local copies of the model, it can be said to be global. However, this naming scheme might be a bit technical.
  * Vote with :confused:


clinssen:
"Injector" or "injection" might suggest that it's a current-based device (injecting currents). 

Something close to "spike_generator" would seem to make the most sense to follow existing terminology. ``spike_generator_local`` or ``_neuron`` could both work here.

testsuite/pytests/test_spike_train_injector.py
```diff
@@ -0,0 +1,101 @@
...
+
+    nest.Simulate(simtime)
+
+    spike_data = srec.events
+    post_time = spike_data['times']
+
+    expected_post_time = spike_time + delay
+    if parrot_model == 'parrot_neuron':
+        # round-up
+        expected_post_time = np.true_divide(np.ceil(expected_post_time * 10), 10)
```
heplesser, line 99: Since the time is a scalar there is no need for NumPy here. And instead of using `10`, it is more expressive and robust to use the actual resolution (needs `import math` above)
```suggestion
        expected_post_time = math.ceil(expected_post_time / nest.resolution) * nest.resolution
```

testsuite/pytests/test_spike_train_injector.py
```diff
@@ -0,0 +1,101 @@
...
+
+    assert post_time == expected_post_time
```
heplesser, line 101: I'd use approx to avoid precision problems (needs `from pytest import approx`)
```suggestion
    assert post_time == approx(expected_post_time)
```

testsuite/pytests/test_spike_train_injector.py
```diff
@@ -0,0 +1,101 @@
...
+    nest.set(resolution=0.1, tics_per_ms=1000.0)
+
+    inj_nrn = nest.Create("spike_train_injector",
+                          params={'spike_times': in_spike_times,
+                                  'precise_times': precise_times,
+                                  'allow_offgrid_times': allow_offgrid_times}
+                          )
+
+    out_spike_times = nest.GetStatus(inj_nrn, "spike_times")[0]
+    assert np.array_equal(out_spike_times, np.array(expected_spike_times))
```
heplesser, line 69: The neatest way I have found especially to get nice output if things go wrong is
```suggestion
    assert out_spike_times == approx(expected_spike_times)
```
This handles lists and arrays transparently. Needs `from pytest import approx`.

testsuite/pytests/test_spike_train_injector.py
```diff
@@ -0,0 +1,101 @@
...
+)
+def test_set_spike_times(
+    reset_kernel,
+    in_spike_times,
+    expected_spike_times,
+    precise_times,
+    allow_offgrid_times
+):
+    """
+    Verify correct handling of spike times not coinciding with a step for
```
heplesser, line 56: First line of docstring should be single-line description followed by empty line.

testsuite/pytests/test_spike_train_injector.py
```diff
@@ -0,0 +1,101 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import nest
+import pytest
+import numpy as np
```
heplesser, line 24:
```suggestion
from pytest import approx
import math
```
See below for details.

models/spike_train_injector.h
```diff
@@ -0,0 +1,505 @@
...
+
+Short description
++++++++++++++++++
+
+Neuron that emits prescribed spike trains.
+
+Description
++++++++++++
+
+The spike train injector neuron simply emits spikes at prescribed spike times
```
heplesser, line 49:
```suggestion
The spike train injector neuron emits spikes at prescribed spike times
```

models/spike_train_injector.h
```diff
@@ -0,0 +1,505 @@
...
+which are given as an array. Incoming spikes will be ignored.
+
+The spike train injector neuron behaves similarly to a spike generator, but
```
heplesser, line 52: This and the next section should be much shorter and less technical, just a recommendation: use this one if you spike trains have similar rate to normal neurons, use spike_generator if you have very high rates.

heplesser (changes requested):
@nicolossus Thank you for preparing this. I have a few issues:

- Since this is a neuron, we cannot support `DSSpikeEvents` and thus no `spike_weights`. All code and doc on this should be removed.
- We need an mpi-test that also test if multiplicities are handled properly. That can be a normal Py-MPI test as in the pytest/mpi/4 directory.
- There should be a note in the spike_generator doc pointing to this new neuron.
- Check the docstrings for proper cross-references.

models/spike_train_injector.cpp
```diff
@@ -0,0 +1,463 @@
...
+    if ( tnext_stamp > tstop )
+    {
+      break;
+    }
+
+    long step = tnext_stamp.get_steps();
+
+    if ( get_t_min_() < step and step <= get_t_max_() )
+    {
+      SpikeEvent* se;
```
heplesser, line 425: No need for a dynamically allocated object here, since we only have plain `SpikeEvent`s here. Corresponding changes further down and no `delete` at the end.
```suggestion
      SpikeEvent se;
```

models/spike_train_injector.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+      if ( not P_.spike_multiplicities_.empty() )
+      {
+        se->set_multiplicity( P_.spike_multiplicities_[ S_.position_ ] );
+      }
+
+      // we need to subtract one from stamp which is added again in send()
+      long lag = Time( tnext_stamp - sliceT0 ).get_steps() - 1;
+
+      // all spikes are sent locally, so offset information is always preserved
```
heplesser, line 441: Comment is outdated and incorrect, remove.
```suggestion
```

models/spike_train_injector.cpp
```diff
@@ -0,0 +1,463 @@
...
+      kernel().event_delivery_manager.send( *this, *se, lag );
+      delete se;
+    }
+
+    ++S_.position_;
+  }
+}
+
+void
```
heplesser, line 450: I don't think this is necessary any more. Since the injector behaves like a normal neuron on its output side, I cannot see why we should need this restriction. @jougs, can you remember why we introduced it?

heplesser (changes requested):
Thanks for the update. I have found some more issues and have sent you a slightly revised MPI test script.

doc/htmldoc/devices/stimulate_the_network.rst
```diff
@@ -17,6 +17,10 @@ The most commonly used generators for spike trains are:
 - :doc:`../models/poisson_generator`
 - :doc:`../models/spike_generator`
 
+For injecting prescribed spike trains with rate similar to normal neurons:
```
jougs, line 20:
```suggestion
For injecting prescribed spike trains (e.g. to mimic the output of regular neurons):
```

models/spike_train_injector.cpp
jougs: I don't see a reason for this restriction.

jougs (changes requested):

testsuite/pytests/mpi/4/test_spike_train_injector_mpi.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_spike_train_injector_multiplicities():
+    """
+    Test spike train multiplicity with spike_train_injector model.
+
+    This test verifies the behavior of the spike train injector neuron in
+    MPI parallelsimulations by using parrot neuron's spike repetition
+    properties.
+    """
+    inj = nest.Create("spike_train_injector",
```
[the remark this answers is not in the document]
nicolossus: Resolved by [b7f86af](https://github.com/nest/nest-simulator/pull/2622/commits/b7f86af06f2cef262faad42124a28e6e9aff63ce).

models/spike_train_injector.h
```diff
@@ -0,0 +1,474 @@
...
+
+/**
+ * @brief Spike train injector node.
+ *
+ * See UserDocs for details.
+ *
+ * @note Spikes emitted by a spike train injector neuron will be counted by
+ * the local spike count.
+ */
+class spike_train_injector : public Node
```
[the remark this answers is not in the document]
med-ayssar: I think the `Device` class won't be touched for now, only the `DeviceNode` will be deleted (at least with this first iteration).

nicolossus:
@heplesser @jougs Thanks for all the feedback. I have now incorporated all of your suggestions, so please have another look in case I missed something.

heplesser (approved):

models/spike_train_injector.h
```diff
@@ -0,0 +1,374 @@
...
+will actually use, i.e., for grid-based simulation the spike times rounded
+to the appropriate point on the time grid. This means that ``GetStatus`` may
+return different `spike_times` values at different resolutions.
+
+Example:
+
+  ::
+
+     nest.Create("spike_train_injector",
+                 params={"spike_times": [1.0, 2.0, 3.0]})
```
jougs, line 108: @jessica-mitchell: can you please give this model description (especially the code blocks) a quick run-over to see if everything corresponds to our guidelines? Thanks!

jougs (changes requested):
Many thanks for your changes! In my heart, I approve the changes and only leave the request for changes in as a merge block until @jessica-mitchell had a look at the model documentation.

models/spike_generator.h
```diff
@@ -70,6 +72,11 @@ are delivered with the respective weight multiplied with the
 weight of the connection. To disable this functionality, the
 spike_weights array can be set to an empty array.
 
+The spike generator supports spike times that do not coincide with a time
+step, i.e., are not falling on the grid defined by the simulation resolution.
+There are three options that control how spike times that do not coincide
+with a step are handled (see also examples below):
+
     `precise_times`  default: false
```
jessica-mitchell, line 80: Please change these lines as suggested in spike_train_injector.h

models/spike_train_injector.h
```diff
@@ -0,0 +1,374 @@
...
+in the future. Trying to set a spike time in the past or at the current time
+step will cause a NEST error. Setting a spike time of 0.0 will also result
+in an error.
+
+Multiple occurrences of the same time indicate that more than one event is to
+be generated at this particular time.
+
+The spike generator supports spike times that do not coincide with a time
+step, i.e., are not falling on the grid defined by the simulation resolution.
+There are three options that control how spike times that do not coincide
```
jessica-mitchell, line 71: Reword like 71/72

Spike times that do not coincide with a step are handled with one of three options (see examples below).

jessica-mitchell (changes requested):
@nicolossus I have a few minor suggestions for changes

models/spike_train_injector.cpp
```diff
@@ -0,0 +1,381 @@
...
+  for ( size_t n = 0; n < n_spikes; ++n )
+  {
+    times_ms->push_back( spike_stamps_[ n ].get_ms() );
+    if ( precise_times_ )
+    {
+      ( *times_ms )[ n ] -= spike_offsets_[ n ];
+    }
+  }
+
+  ( *d )[ names::spike_times ] = DoubleVectorDatum( times_ms );
```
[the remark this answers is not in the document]
nicolossus: Yes! It will be destructed by the implementation of lockPTRDatum.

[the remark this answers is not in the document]
med-ayssar: Ok, Good!

nicolossus:
@jessica-mitchell Thanks for the feedback. I think all of your suggestions are incorporated now.

models/spike_train_injector.cpp
```diff
@@ -0,0 +1,381 @@
...
+  ( *d )[ names::spike_multiplicities ] = IntVectorDatum( new std::vector< long >( spike_multiplicities_ ) );
+  ( *d )[ names::precise_times ] = BoolDatum( precise_times_ );
+  ( *d )[ names::allow_offgrid_times ] = BoolDatum( allow_offgrid_times_ );
+  ( *d )[ names::shift_now_spikes ] = BoolDatum( shift_now_spikes_ );
+}
+
+void
+spike_train_injector::Parameters_::assert_valid_spike_time_and_insert_( double t, const Time& origin, const Time& now )
```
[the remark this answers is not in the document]
nicolossus: I think the explicit name is fine here, as this is used by set_status and it clearly signals that validation of the spike trains will be performed by calling the function. @heplesser what is your opinion?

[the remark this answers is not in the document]
heplesser: @nicolossus I agree, let's leave it as it is.

jessica-mitchell (approved):
Thanks @nicolossus looks good!

nicolossus:
@jessica-mitchell Great! @jougs Could you please give the final approval?

PR #2627, opened 2023-03-06
Make calcium concentration available for setting

clinssen:
Check that we can set and get calcium concentration for all nodes that inherit from the ``StructuralPlasticityNode``.


testsuite/pytests/test_calcium.py
```diff
@@ -0,0 +1,51 @@
...
+            if "Ca" not in n.get():
+                continue
+
+            n_models_tested += 1
+            n = nest.Create(model, params={"Ca": 123.})
+            np.testing.assert_almost_equal(n.Ca, 123.)
+            n.Ca = 42.
+            np.testing.assert_almost_equal(n.Ca, 42.)
+
+        assert n_models_tested > 63
```
med-ayssar, line 51: Having this hard-coded is not a good idea, as the number of models might increase in the future, and I don't see that this checking has really any importance for the given issue, as long you can set the `Ca` value and read it again later.

med-ayssar (changes requested):
Otherwise, the rest looks fine.

testsuite/pytests/test_calcium.py
clinssen: Would you be OK making it ``> 1``, just to have this in there as a sanity check?

med-ayssar: I don't think @heplesser would agree to that, as he wants more modern structure, and if you check the examples he is preparing: [pytest](https://github.com/heplesser/nest-simulator/blob/improve-parameter-test-and-implementation/testsuite/pytests/test_parameter_operators.py), the desired pattern would be the following:

```python
import nest
import pytest

def get_neurons(attribute):
   all_models = nest.node_models()
   neurons = [model if attribute in nest.GetDefaults(model) for model in all_models]
   return neurons

neurons = get_neurons("Ca")


@pytest.hookimpl(tryfirst=True)
def test_at_least_one():
      assert len(neurons) > 0


@pytest.mark.parametrize('model', neurons)
def test_calcium(model):
     n = nest.Create(model, params={"Ca": 123.})
     assert n.Ca  == pytest.approx(123.)
     n.Ca = 42.
    assert n.Ca  == pytest.approx(42.)
     
```

heplesser: @med-ayssar Very nice :). `get_neurons()` should probably go to a separate module in which we collect helper functions for tests and certain pre-configured fixtures. And maybe be called `select_neurons_by_attribute()` or so.


The test itself could be even more robust as follows:
```python
ca_default = nest.GetDefaults(model, 'Ca')
n = nest.Create(model, params={'Ca': ca_default + 5})
assert n.Ca  == pytest.approx(ca_default + 5)
n.Ca = ca_default + 12
assert n.Ca  == pytest.approx(ca_default + 12)
```
Then, the test will even work for neurons with default Ca values of 123 and 42 :).

med-ayssar: We can even make that offset a `random constant`

heplesser: "Random constant" ... We alreaduy use [pytest-randomly](https://pypi.org/project/pytest-randomly/) in our test setup, but for now to keep the random numbers fixed. But fuzzing tests wrt random numbers would be neat, provided we can turn fuzzing off for debugging.

heplesser: > Would we still prefer to put all the calcium-getting-setting tests into a class (like `TestCalcium`)? Otherwise variables like `neurons` here become globals which might accidentally leak state.

Variables defined in a module are in the namespace of the module, so I do not see any risk of leaking. 

The main cross-test risk is, I think, that the kernel may not be in default state, so I think all tests should call `ResetKernel()`.

clinssen: I tried to refactor the test to use ``@pytest.mark.parameterize`` while avoiding global variables, but class methods cannot be parameterized by instance members (such as ``self.models_list``) as ``self`` is not available within the parameterize marker. The current solution therefore still uses a for loop to loop over models. I don't want to call ``get_node_models_by_attribute("Ca")`` twice, because we should just grab this list once, check that its length is greater than zero, and iterate over it.

This solution is not ideal, but I believe it's the cleanest option (under the constraint of not having global variables).

By the way, be careful with ``@pytest.hookimpl``, these functions need to be in separate files to be actually run. It is probably better to just use fixtures for this purpose instead.

nicolossus (changes requested):
@clinssen Thanks! The addition to the kernel looks good. However, I would like to see some changes to the corresponding test code, see https://github.com/clinssen/nest-simulator/pull/7.

Moreover, could you pull master, so that the new directory structure and CI pipeline are merged?

nicolossus (approved):
@clinssen LGTM, thanks!

testsuite/pytests/test_calcium.py
```diff
@@ -0,0 +1,59 @@
...
+that inherit from the strutural plasticity node class in the kernel.
+"""
+
+import nest
+import numpy as np
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def reset_kernel(self):
```
nicolossus, line 34:
```suggestion
def reset_kernel():
```
Missed this error from a merge conflict in the review.

med-ayssar:
Looks good!

med-ayssar (approved):

nicolossus:
@heplesser Ping!

heplesser (approved):

PR #2628, opened 2023-03-07
Improve parameter test and implementation

heplesser:
As a step towards developing good test styles, this PR provides a re-write of `test_parameter.py` using PyTest features. It extends the scope of the tests and includes minor fixes:

- float / Parameter is now supported
- when attempting ** with Parameter as exponent, a TypeError is raised to conform with usual Python practice
-

### DIFF
--- a/testsuite/pytests/test_parameter.py
+++ b/testsuite/pytests/test_parameter.py
@@ -20,107 +20,92 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Simple Parameter tests
+Test that Parameters support arithmetic operations correctly.
+
+This set of tests confirms that arithmetic operations on NEST Parameter objects yield
+correct results by comparing with operations on the undelying numeric values.
+It also confirms that operations on Parameter objects and plain numbers work.
+
+@note These tests only confirm that operators on parameters work in principle. Therefore,
+      we can use constant parameters for simplicity. We need to test elsewhere that
+      other types of Parameters produce correct values.
 """
 
 import nest
 import pytest
-
-va, vb = 5, 7
-a = nest.CreateParameter('constant', {'value': va})
-b = nest.CreateParameter('constant', {'value': vb})
+import operator
 
 
-@pytest.fixture(autouse=True)
-def reset_kernel():
-    nest.ResetKernel()
+@pytest.mark.parametrize('op', [operator.pos,
+                                operator.neg])
+def test_binary_operators(op):
+    """
+    Perform tests for unary operators.
 
+    Parametrization is over operators
+    """
 
-@pytest.mark.parametrize('op_a, op_b', [[a, b], [a, vb], [vb, a]])
-def test_add(op_a, op_b):
-
-    try:
-        val_a = op_a.GetValue()
-    except AttributeError:
-        val_a = op_a
-    try:
-        val_b = op_b.GetValue()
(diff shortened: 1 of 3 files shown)


testsuite/pytests/test_parameter.py
```diff
@@ -20,108 +20,127 @@
...
-
-
-def run():
-    runner = unittest.TextTestRunner(verbosity=2)
-    runner.run(suite())
+This set of tests confirms that arithmetic operations on NEST Parameter objects yield
+correct results by comparing with operations on the undelying numeric values.
+It also confirms that operations on Parameter objects and plain numbers work.
 
+@note These tests only confirm that operators on parameters work in principle. Therefore,
```
nicolossus, line 29: Isn't `@note` more of a Doxygen command? I don't mind the usage, just wondering if there is a more appropriate way to highlight notes. For instance, my editor highlights `NOTE` (and also `TODO`) in Python docstrings but not `@note`.

nicolossus (approved):
LGTM, thanks!

testsuite/pytests/test_parameter.py
jougs: I agree. I think you can use proper Restructured Text markup in docstrings, as those also get rendered nicely when we include the tests in the documentation (which we probably should at one point).

heplesser: Fixed.

testsuite/pytests/test_parameter.py
```diff
@@ -20,108 +20,127 @@
...
+      we can use constant parameters for simplicity. We need to test elsewhere that
```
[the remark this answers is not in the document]
heplesser: Those tests do not exist yet. I removed the "We need ..." part of the comment and created #2630 instead. We should use issues for things to be done, not hide todos in comments ;).

testsuite/pytests/test_parameter_operators.py
```diff
@@ -0,0 +1,163 @@
...
+                           31,
+                           nest.CreateParameter('constant', {'value':  5})],
+                          [operator.mod,
+                           nest.CreateParameter('constant', {'value': 31}),
+                           nest.CreateParameter('constant', {'value':  5})],
+                          [operator.mod,
+                           nest.CreateParameter('constant', {'value':  31}), 5],
+                          [operator.pow,
+                           31,
+                           nest.CreateParameter('constant', {'value':  5})]])
```
jougs, line 54:
```suggestion
import operator as op


def const_param(val):
    return nest.CreateParameter('constant', {'value': val})


@pytest.mark.parametrize('op, a, b', [
    [op.pow, const_param(31), const_param(5)],
    [op.pow,             31 , const_param(5)],
    [op.mod, const_param(31), const_param(5)],
    [op.mod, const_param(31),             5 ],
    [op.pow,             31 , const_param(5)]
])
```

jougs (commented):
For me, the `parametrize` lines are really hard to read. I have added a suggestion to make them easier.

heplesser:
@jougs I have now applied your suggestion to the remaining tests as well. I changed `import operator as op` to `... as ops` to avoid having the same name for the module with all operands and for the local variable representing the individual operation inside each test.

jougs (approved):
Thanks! I think this is very nice now.

heplesser:
@jougs Could you resolve the conversations so we can merge?

PR #2631, opened 2023-03-12
Further improvement of parameter operator test implementation

heplesser:
This PR makes the parameter operator tests even more readable and efficient. It is a follow-up to #2628.

Concerning efficiency, tests using Py 3.11 indicate that using `hasattr()` instead of `try-except` improves performance by almost half.

One could argue that the test code here would become even simpler of the nest Parameter objects supported direct casting to `int` or `float`. But these parameters are only to be evaluated inside the C++ kernel, so adding ``__int__()`` or ``__float()__`` methods at the Python level would only simplify test-writing, with no bearing on actual use. But writing code for the sake of testing only seems not sensible.


testsuite/pytests/test_parameter_operators.py
```diff
@@ -94,37 +96,32 @@ def test_binary_operators(op, a, b):
...
-    try:
-        val_a = a.GetValue()
-    except AttributeError:
-        val_a = a
-    try:
-        val_b = b.GetValue()
-    except AttributeError:
-        val_b = b
+    val_a = to_numeric(a)
+    val_b = to_numeric(b)
```
jougs, line 100: Much nicer indeed :smile:

testsuite/pytests/test_parameter_operators.py
```diff
@@ -94,37 +96,32 @@ def test_binary_operators(op, a, b):
...
+    [ops.pow, _const_param(31), 5, True],
+    [ops.pow, _const_param(31), _const_param(5), False],
+    [ops.pow, 31, _const_param(5), False],
+    ])
+def test_incomplete_binary_operators(op, a, b, is_supported):
     """
-    Perform tests for binary operators that do not support Parameter as any operand.
+    Perform tests for binary operators that only support Parameters as one operand.
+
+    Pass `False` for `is_supported` to indicate unsupported operand combinations.
```
jougs, line 114: Can you please elaborate on "indicate" here? From the current description it does not really become clear what the function would do if `True` or `False` is passed. Something like:

> If `True` is passed, the function asserts that the operator works as expected, if `False` is passed, it checks if an Exception is raised.

testsuite/pytests/test_parameter_operators.py
```diff
@@ -94,37 +96,32 @@ def test_binary_operators(op, a, b):
...
+    val_a = to_numeric(a)
+    val_b = to_numeric(b)
 
     assert op(a, b).GetValue() == op(val_a, val_b)
 
 
-@pytest.mark.parametrize('op, a, b', [
-    [ops.pow, const_param(31), 5]])
-def test_incomplete_binary_operators(op, a, b):
+@pytest.mark.parametrize('op, a, b, is_supported', [
```
nicolossus, line 105: Another option to `is_supported` is to use the `pytest.mark.xfail` decorator. By default, we then expect the test to fail and when it fails the test is reported as `XFAIL`. However, `pytest.mark.xfail` has a `strict` parameter that defines two distinct behaviors for the mark. With `strict=False`, the default, the test will be counted separately as an `XPASS` if it passes. The decorator would simplify the test code to just:

```Python
@pytest.mark.xfail
@pytest.mark.parameterize('op, a, b', [...])
def test_incomplete_binary_operators(op, a, b):
    val_a = to_numeric(a)
    val_b = to_numeric(b)
    assert op(a, b).GetValue() == op(val_a, val_b)
``` 

with output summary (run `pytest -rxX`): 

```
================================================== short test summary info ==================================================
XFAIL ../nest-simulator/testsuite/pytests/test_parameter_operators.py::test_incomplete_binary_operators[pow-a1-b1]
XFAIL ../nest-simulator/testsuite/pytests/test_parameter_operators.py::test_incomplete_binary_operators[pow-31-b2]
XPASS ../nest-simulator/testsuite/pytests/test_parameter_operators.py::test_incomplete_binary_operators[pow-a0-5] 
========================================= 47 passed, 2 xfailed, 1 xpassed in 0.96s ==========================================
```

Note that in this context, both `xfailed` and `xpassed` are regarded as successful tests. 

We might want to consider `pytest.mark.xfail` to document features not yet implemented, as these then can be extracted from the output summary.

heplesser:
I dug deeper into the xfail documentation and found that one can mark individual cases explicitly as xfail, while cases that should pass can be provided as usual. I hope you like the new implementation.

nicolossus (approved):
This looks very nice! I'm guessing the newly added `.bak` file is unintentional and should therefore be removed before merge.

testsuite/pytests/test_parameter_operators.py.bak
jougs, line 1: Please remove. This file should not be here.

heplesser:
> This looks very nice! I'm guessing the newly added `.bak` file is unintentional and should therefore be removed before merge.

Thanks for noticing, I have removed the file again.

testsuite/pytests/test_parameter_operators.py
```diff
@@ -94,35 +96,37 @@ def test_binary_operators(op, a, b):
...
-        val_b = b
+    val_a = to_numeric(a)
+    val_b = to_numeric(b)
 
     assert op(a, b).GetValue() == op(val_a, val_b)
 
 
+def _unsupported_binary_op(op, a, b):
+    """
+    Represent test cases where at op is not supported for at least one of a or b.
```
jougs, line 107:
```suggestion
    Represent test cases where an op is not supported for at least one of a or b.
```

jougs (approved):
Just a small typo. Otherwise nice work!

PR #2652, opened 2023-03-23
Convert multimeter test from SLI to Python implementation

heplesser:
This PR provides a polished version of the Python implementation of the multimeter test ported from SLI.


heplesser:
I am not sure why the Python tests fail here. It is again the numpy datatype problem, it seems, but that was fixed with #2571 and this PR does not touch that code. Also, all tests pass on my branch (https://github.com/heplesser/nest-simulator/actions/runs/4505551444) and locally on my machine.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,78 @@
...
+import pytest
+import nest
+
+
+@pytest.fixture
+def _set_resolution():
+    """
+    Set resolution to power of two to avoid rounding issues.
+    """
+
```
nicolossus, line 31:
```suggestion

```
Remove whitespace

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,78 @@
...
+    Set resolution to power of two to avoid rounding issues.
+    """
+    
+    nest.ResetKernel()
+    nest.resolution = 2**-3
+
+
+# Obtain all models with non-empty recordables list
+models = [model for model in nest.node_models
+          if ('recordables' in (dflts := nest.GetDefaults(model))
```
nicolossus, line 38: Here, PEP8 complains about `E203 whitespace before ':'`. It is a bit strange that the walrus operator is not recognized.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,78 @@
...
+              and dflts['recordables'])]
+
```
nicolossus, line 40:
```suggestion


```
Additional blank line before function.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,78 @@
...
+    recording from all recordables and test that data is provided.
+    I also checks that the recording interval can be set.
+
+    @note This test does not check if the data is meaningful.
+    """
+
+    recording_interval = 2
+    simtime = 10
+    num_data_expected = simtime / recording_interval - 1
+
```
nicolossus, line 56:
```suggestion

```
Remove whitespace.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,78 @@
...
+    nest.Connect(mm, nrn)
+    nest.Simulate(simtime)
+
+    result = mm.events
+
+    for r in recordables + ('times', 'senders'):
+        assert r in result
+        assert len(result[r]) == num_data_expected
+
+
```
nicolossus, line 70:
```suggestion

```

Remove whitespace.

testsuite/pytests/test_multimeter.py
nicolossus, line 38:
```suggestion
def models():
    """Generate all models with non-empty recordables list"""
    yield from filter(lambda model: nest.GetDefaults(model).get('recordables'), nest.node_models)
```
If we want a more generic way of obtaining models with certain properties, perhaps a function like this could be a viable solution? What to get from the dictionary, here `'recordables'`, can also then be provided as a function argument.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,78 @@
...
+    nest.ResetKernel()
+    nest.resolution = 2**-3
+
+
+# Obtain all models with non-empty recordables list
+models = [model for model in nest.node_models
+          if ('recordables' in (dflts := nest.GetDefaults(model))
+              and dflts['recordables'])]
+
+@pytest.mark.parametrize('model', models)
```
nicolossus, line 41:
```suggestion
@pytest.mark.parametrize('model', models())
```
Needs to be changed if `models` is a function.

nicolossus (commented):
@heplesser To me it looks like it is the static code check that fails? See comments. I also made a suggestion for how we can obtain models with certain properties.

testsuite/pytests/test_multimeter.py
heplesser: We used an old system pycodestyle. Should be fixed with next push.

heplesser: I am not quite sure that I find the function more readable, but it is an idea to keep in mind for a library of helper functions. For now, I will just convert the list to a generator expression.

heplesser:
@nicolossus The static checks now pass, we now install an up-to-date pycodestyle via pip; previously, we took an old version from a system location, thus the trouble with the :=.

I don't think the models() function an improvement in this case so I have not adopted it yet.

nicolossus:
@heplesser I see, thanks for adding an up-to-date installation. 

I agree that the `models()` function (with a better function name) would fit better in a helper library. However, I still think that the if-statement

```python
if ('recordables' in (dflts := nest.GetDefaults(model)) and dflts['recordables'])
```

should be changed to (in my opinion) the more concise and readable

```python
if nest.GetDefaults(model).get('recordables')
```

heplesser:
> @heplesser I see, thanks for adding an up-to-date installation.
> 
> I agree that the `models()` function (with a better function name) would fit better in a helper library. However, I still think that the if-statement ...

You are right, I didn't realize `get()` could be used like that. Changed it now, much more readable indeed.

nicolossus (approved):
Thanks! LGTM.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def test_multimeter_freeze():
+    """
+    Ensure that frozen parameter can be set to False but not True on multimeter.
+    """
+
+    nest.Create('multimeter', params={'frozen': False})
+    with pytest.raises(Exception):
+        nest.Create('multimeter', params={'frozen': True})
```
med-ayssar, line 77:
```suggestion
@pytest.mark.parametrize(("value"), [False, pytest.param(True, marks=pytest.mark.xfail(raises=Exception))])
def test_multimeter_freeze(value):
    """
    Ensure that frozen parameter can be set to False but not True on multimeter.
    """
    nest.Create('multimeter', params={'frozen': value})
```
I think the suggestion is slightly better.

heplesser: Thank you for the suggestion, but I do not agree. To me, the original code seems clearer because it is quite straightforward Python code where the first line must execute without exception and the last raising an exception. Parsing that the `parametrize()` will have the same effect will take most readers of the code more time, I believe.

@jougs @nicolossus What do you think?

med-ayssar: With detailed template of how to write Pytests, the readers must (should) be accustomed to such syntax.

nicolossus: @heplesser @med-ayssar I am strongly in favor of keeping the original code. I think the original test conveys what we expect to pass and fail in a clear and concise manner. My main problem with the suggested `parametrize`, however, is that the test with `frozen=True` will be marked as `xfailed`. I think we should use `xfail` to document features not yet implemented, or bugs not yet fixed. The exception raised by `frozen=True` is intended behavior and should therefore not be marked as `xfail` in my opinion.

testsuite/pytests/test_multimeter.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.mark.parametrize('model', models)
+def test_recordables_are_recorded(_set_resolution, model):
+    """
+    Test that recordables are recorded.
+
+    For each model with recordables, set up minimal simulation
+    recording from all recordables and test that data is provided.
+    I also checks that the recording interval can be set.
+
+    @note This test does not check if the data is meaningful.
```
nicolossus, line 49:
```suggestion
    .. note::
       This test does not check if the data is meaningful.
```
Missed this in my first review. Notes should use the rST directive.

med-ayssar (approved):

testsuite/pytests/test_multimeter.py
heplesser: Since @med-ayssar gave a thumbs up on @nicolossus comment, I will mark this as resolved.

PR #2658, opened 2023-04-10
Make it an error to set multiple threads if NEST was built without thread support

heplesser:
Until now, setting more than one thread if NEST was built without thread support resulted only in a warning. This led to a considerable number of tests being run in a meaningless way, and potentially causing problems in MPI-based test.

This PR makes it an error to try to set more that one local thread if NEST was built without thread support. The PR also ensures that all tests requiring multiple threads are skipped.

### DIFF
--- a/nestkernel/vp_manager.cpp
+++ b/nestkernel/vp_manager.cpp
@@ -95,9 +95,7 @@ nest::VPManager::set_status( const DictionaryDatum& d )
 
   if ( force_singlethreading_ and n_threads > 1 )
   {
-    std::string msg = "Multithreading requested, but unavailable. Using a single thread.";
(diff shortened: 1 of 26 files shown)


JanVogelsang:
Apart from the formatting issues, the changes look good to me!

JanVogelsang (approved):

med-ayssar (approved):
I was just randomly checking some interesting PRs.

@jougs, if you don't mind, I can review it instead of you :D

@heplesser: It looks good!

heplesser:
@med-ayssar Thanks!

@jougs Since it (minimally) changes behaviour (some scripts will fail now), I'd appreciate your input as well.

jougs (approved):
I think this is a very good change and am shocked that this was only a warning. Thanks!

PR #2660, opened 2023-04-14
Fix OpenMP issues on CI under macOS and re-activate testing with threads for macOS

### LINKED ISSUE #2515: MacOS CI tests fail

heplesser:
This PR fixes the CI build of NEST under macOS by providing the location of the OpenMP lib through a `brew --prefix` query. It also re-activates testing with threads under macOS. A meaningless `brew info` line is removed—it only printed canned package information.

This fixes #2515.

PR developed in collaboration with @nicolossus.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -226,7 +226,6 @@ jobs:
       - name: "Install MacOS system dependencies"
         run: |
           brew install coreutils gsl open-mpi libomp automake autoconf libtool boost
-          brew info python
 
       - name: "Restore pip cache"
         env:
--- a/build_support/ci_build.sh
+++ b/build_support/ci_build.sh
@@ -178,7 +178,7 @@ if [ "$xNEST_BUILD_TYPE" = "FULL_MACOS" ]; then
     xLTDL=1
     xMPI=1
     xMUSIC=0
-    xOPENMP=0
+    xOPENMP=1
     xPYTHON=1
     xREADLINE=1
     xSIONLIB=0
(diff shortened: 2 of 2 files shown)


nicolossus (approved):
Very elegant!

lekshmideepu:
> This PR fixes the CI build of NEST under macOS by providing the location of the OpenMP lib through a `brew --prefix` query. It also re-activates testing with threads under macOS. A meaningless `brew info` line is removed—it only printed canned package information.
> 
> This fixes #2515.
> 
> PR developed in collaboration with @nicolossus.

Looks good to me. Yes, I think I have added brew info line  in the initial phase to get the package information. This is not at all necessary. Use of `brew prefix` is a great fix.

PR #2661, opened 2023-04-18
Fix STDP k-value error for edge case

JanVogelsang:
PR #2443 contained a bug in the fix, this PR hotfixes the bugfix.  
Adding `max_delay_ + kernel().connection_manager.get_min_delay()` looked correct, but was actually adding `double + long`, i.e., `ms + delay`.  
It had to be `max_delay_ + Time::delay_steps_to_ms( kernel().connection_manager.get_min_delay() )` instead.


clinssen:
Can we amend the unit test so it fails on the old code, but passes on the new code?

JanVogelsang:
True, I'll prepare one.

JanVogelsang:
I improved the test to correctly detect the original error correctly, tested some more parameter combinations and fixed some bugs in the test which occurred for that parameters. Additionally, I added a test for the `stdp_pl_synapse_hom` which is quite similar to the regular STDP test, mainly with different STDP weight dynamics.

JanVogelsang:
Strange, NEST aborts, but why? How can I debug this issue when it doesn't occur locally?

heplesser:
> Strange, NEST aborts, but why? How can I debug this issue when it doesn't occur locally?

Do you have the same setup locally (Python version etc) as on the CI runner? And have you tried to run the test locally as the test runner would do it, i.e., with `python -m pytest --verbose --timeout 30 --numprocesses=1 test_stdp_synapse.py`?

clinssen:
@JanVogelsang : could you pull master and fix the merge conflicts? There is also one "TODO" marked in the code, could you re-enable plotting? Cheers!

JanVogelsang:
Sure thing! Re-enabled plotting, removed some warnings while I was at it, and don't get any errors or warning locally when I run pytest the same way the CI runner does (although with a higher Python version, I use 3.11).

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -87,15 +89,51 @@ def do_nest_simulation_and_compare_to_reproduced_weight(self, fname_snip):
...
+        ) = self.reproduce_weight_drift(pre_spikes, post_spikes, self.init_weight, fname_snip=fname_snip)
 
         # ``weight_by_nest`` contains only weight values at pre spike times, ``weight_reproduced_independently``
-        # contains the weight at pre *and* post times: check that weights are equal for pre spike times
+        # contains the weight at pre *and* post times: check that weights are equal only for pre spike times
         assert len(weight_by_nest) > 0
-        np.testing.assert_allclose(t_weight_by_nest, t_weight_reproduced_independently)
-        np.testing.assert_allclose(weight_by_nest, weight_reproduced_independently)
+
+        difference_matrix = t_weight_by_nest[t_weight_by_nest < self.simulation_duration].reshape(
```
clinssen, line 103: Why this more complicated formulation of the assert on times?

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -104,7 +142,13 @@ def do_the_nest_simulation(self):
         """
         nest.set_verbosity("M_WARNING")
         nest.ResetKernel()
-        nest.SetKernelStatus({"resolution": self.resolution})
+        nest.SetKernelStatus(
+            {
+                "resolution": self.resolution,
+                "min_delay": min(self.min_delay, self.dendritic_delay),
```
clinssen, line 148: Why do you need to set min and max_delay explicitly?

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -273,51 +326,62 @@ def Kpost_at_time(t, spikes, inclusive=True):
...
+        Kpre_log=None,
+        Kpost_log=None,
+        pre_indices=slice(-1),
+        fname_snip="",
+        title_snip="",
     ):
         fig, ax = plt.subplots(nrows=3)
 
         n_spikes = len(pre_spikes)
+        print(n_spikes)
```
clinssen, line 384: remove

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -348,11 +412,20 @@ def plot_weight_evolution(
...
+                    self.max_delay = max(self.min_delay, self.max_delay)
+                    for self.nest_neuron_model in ("iaf_psc_exp", "iaf_cond_exp"):
+                        for self.neuron_parameters["t_ref"] in (self.resolution, 0.5, 1.0, 1.1, 2.5):
+                            fname_snip = "_[nest_neuron_mdl=" + self.nest_neuron_model + "]"
+                            fname_snip += "_[dend_delay=" + str(self.dendritic_delay) + "]"
+                            fname_snip += "_[t_ref=" + str(self.neuron_parameters["t_ref"]) + "]"
+                            self.do_nest_simulation_and_compare_to_reproduced_weight(fname_snip=fname_snip)
+
+
+if __name__ == "__main__":
```
clinssen, line 430: remove

testsuite/pytests/test_stdp_synapse.py
```diff
@@ -63,20 +65,20 @@ def init_params(self):
...
+            "tau_minus": self.tau_post,
+        }
+
+        # While the random sequences, fairly long, would supposedly reveal small differences in the weight change
+        # between NEST and ours, some low-probability events (say, coinciding spikes) can well not have occurred.
+        # To generate and test every possible combination of pre/post order, we append some hardcoded spike sequences.
         self.hardcoded_pre_times = np.array([1, 5, 6, 7, 9, 11, 12, 13, 14.5, 16.1], dtype=float)
-        self.hardcoded_post_times = np.array([2, 3, 4, 8, 9, 10, 12, 13.2, 15.1, 16.4], dtype=float)
-        self.hardcoded_trains_length = 2.0 + max(np.amax(self.hardcoded_pre_times), np.amax(self.hardcoded_post_times))
+        self.hardcoded_post_times = np.array([2, 3, 4, 8, 9, 10, 12, 12.2, 14.1, 15.4], dtype=float)
```
clinssen, line 76: Does the test still pass on the old spike times? Can we run the test on both sequences?

clinssen (commented):
Could it be that some unrelated changes were pulled in in ``connector_base.h`` and ``event_delivery_manager.cpp`` (and some other kernel files)?

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: Because STDP (in NEST) strongly depends on when pre-synaptic spikes are communicated. And in theory things might become funky when spikes are communicated in different intervals than the delay of the specific connection. And things get especially funky when axonal delays are involved, so this is mainly a preparation for the future.

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: Why? Is it bad to make the test runnable by itself? Or because you should always run it using pytest?

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: Sure, I can include the previous spike trains as well

JanVogelsang:
Unrelated, but not worth their own PR. Just some lines that Pooja forgot in a recent PR (#2618).

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: Weight recorders in master only record pre-synaptic spikes and this will change in the future. As a preparation for the future, the weights at all times are calculated and returned in the test. The complicated asserts then extract the weights changes due to pre-synaptic spikes from all these weights. Not sure if there was any other reason, the test is already a year old, I can't quite remember all details.

testsuite/pytests/test_stdp_synapse.py
clinssen: OK, cool! I just thought there could be an option like "None" in there, which would allow NEST to compute the min and max-delay values internally.

testsuite/pytests/test_stdp_synapse.py
clinssen: Yeah, the idea is that the tests should always be run through pytest.

testsuite/pytests/test_stdp_synapse.py
JanVogelsang: Not sure if this is necessary, as there will be other tests ensuring that the min_delay is always set correctly. For this test it is fine to always set the min and max delay explicitly.

JanVogelsang:
@abigailm Can you review this PR?

JanVogelsang:
@abigailm Now that you are back from vacation, can you review this?

abigailm:
All looks fine to me

abigailm (approved):

PR #2663, opened 2023-04-19
Add CMake option `-Dminimal-model-set` to reduce compile cycle times

jougs:
This is mainly useful in kernel development scenarios where you have many compile-test cycles and don't want to re-compile all the models all the time.

@jessica-mitchell: Where would be a good point in the developer space to add a sentence like this?
```
If you are working on the simulation kernel and have many compile-test cycles but don't want to
re-compile all the models all the time, you may consider adding `-Dminimal-model-set` to your
CMake command line. 
```


jessica-mitchell:
@jougs Of the [several pages](https://nest-simulator.readthedocs.io/en/latest/developer_space/index.html#developer-space) we have for developers,  none seem to quite fit this topic in my eyes. Should we have a page for things like recommendations / best practises for ... testing/debugging the kernel?

heplesser (approved):

heplesser:
Closing this PR on suggestion by @jougs as an even better solution is on its way.

PR #2664, opened 2023-04-20
Refactor rm -rf from testsuite script

clinssen:
Remove footgun from test script. It removed my build directory while I was working on getting ``make installcheck`` to run on JUSUF.

The approach in this PR uses mktemp to create a new directory with a unique name. The directory is created in ``(the installation directory)/share/nest/testsuite`` and its name can no longer be manually specified by the user.


jougs (approved):
LGTM!

heplesser (approved):

PR #2665, opened 2023-04-20
Set sender node id info with correct target lcid

JanVogelsang:
In the `event_delivery_manager` the `sender_node_id_info` is set during the event delivery using the same `lcid` for all contiguous connections with the same source instead of the actual lcid of the connection. This is not wrong per se, but prevents us from grabbing the correct lcid lateron from the `set_sender_node_id_info`. One could argue if that shouldn't be done at all, but why not. In NEST there is currently no use case in the kernel or any existing model, but it might potentially be used in NESTML or when users write their own models.  
Setting the correct lcid does not add any overhead and even reduces some redundancy in the code.

### DIFF
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -617,13 +617,7 @@ EventDeliveryManager::deliver_events_( const size_t tid, const std::vector< Spik
       {
         if ( spike_data.get_tid() == tid )
         {
-          const size_t syn_id = spike_data.get_syn_id();
-          const size_t lcid = spike_data.get_lcid();
-
-          // non-local sender -> receiver retrieves ID of sender Node from SourceTable based on tid, syn_id, lcid
-          // only if needed, as this is computationally costly
-          se.set_sender_node_id_info( tid, syn_id, lcid );
-          kernel().connection_manager.send( tid, syn_id, lcid, cm, se );
+          kernel().connection_manager.send( tid, spike_data.get_syn_id(), spike_data.get_lcid(), cm, se );
         }
       }
       else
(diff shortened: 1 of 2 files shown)


jougs:
@JanVogelsang: can you please comment on the status of this (and maybe merge master)? Thanks!

JanVogelsang:
Well, status is that I am waiting for a review. It's a rather small change and even a non-breaking one, but one that might help reduce errors in the future (as I had a bug in one of my branches because of this).

JanVogelsang:
Interesting, the RTD CI fails with error 429: Too many requests

nestkernel/event_delivery_manager.cpp
```diff
@@ -617,13 +617,7 @@ EventDeliveryManager::deliver_events_( const size_t tid, const std::vector< Spik
       {
         if ( spike_data.get_tid() == tid )
         {
-          const size_t syn_id = spike_data.get_syn_id();
-          const size_t lcid = spike_data.get_lcid();
-
-          // non-local sender -> receiver retrieves ID of sender Node from SourceTable based on tid, syn_id, lcid
-          // only if needed, as this is computationally costly
-          se.set_sender_node_id_info( tid, syn_id, lcid );
```
jougs, line 625: I am afraid I don't understand why the event does not have to be properly parametrized anymore. Can you please comment on this?

JanVogelsang: Well, this is the point of this PR, _before_ it wasn't parametrized correctly, because the lcid was actually not correct (it always pointed to the very first connection, instead of the n-th one after the first). This line just got moved from the event delivery manager to the connector.

jougs: Ah! Thanks! I failed to grasp that the line was *moved*. Thanks.

jougs (approved):
Many thanks for this fix.

nestkernel/event_delivery_manager.cpp
mlober: I stumbled across this too. Which piece of code is now equivalent to parameterizing the event with the correct lcid? The changes in the connector seem to only regard the variable names, don't they?

mlober: Yes, I see now. Thanks.

mlober (approved):
Thanks for this improvement. Looks good to me.

PR #2676, opened 2023-04-24
Fix typo in existing `CopyModel` Pytest and remove obsolete SLI test

suku248:


jougs (approved):
Trivial, so merging without second review

PR #2677, opened 2023-04-24
Catch all exceptions in SimulationManager::update_()

clinssen:
In ``SimulationManager::update_()``, exceptions in a thread-parallel context are caught and buffered, to be re-thrown later outside of the parallel context. However, the try..catch block involved only wrapped the ``Node::update()`` call, not any of the other code in ``SimulationManager::update_()``.

This came up because the following line should be "fail_butnocrash_or_die"; the "hard failure" (exception not handled; SLI interpreter process terminates) was not being caught by the test: https://github.com/nest/nest-simulator/blob/0888491769edaa279e4c43fa5383112a9f33733b/testsuite/unittests/test_binary.sli#L170

This PR moves the try..catch block one level higher so it wraps all code in ``SimulationManager::update_()``.

### DIFF
--- a/nestkernel/simulation_manager.cpp
+++ b/nestkernel/simulation_manager.cpp
@@ -777,49 +777,53 @@ nest::SimulationManager::update_()
   {
     const thread tid = kernel().vp_manager.get_thread_id();
 
-    do
+    // We update in a parallel region. Therefore, we need to catch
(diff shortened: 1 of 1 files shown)


heplesser (approved):

jougs (approved):
Nice catch. Thanks for fixing.

PR #2679, opened 2023-04-24
Ported `test_iaf_psc_alpha_1to2`

Helveg:


jougs:
There are genuine formatting errors that fail the CI. They need to be addressed before this can be considered for merging.

PR #2680, opened 2023-04-24
Converted `test_iaf.sli`

Helveg:


jougs:
There are genuine formatting errors that fail the CI. They need to be addressed before this can be considered for merging.

PR #2713, opened 2023-04-26
Improve documentation on delays

mlober:
This PR improves the documentation on the effects of resolution change on the min/max delays and other time-based model parameters.

I suggest @jessica-mitchell as a reviewer.


jessica-mitchell (approved):
@mlober thanks for taking some time to make clarify this text. I have some very minor comments to address, but I will still approve.

PR #2715, opened 2023-04-27
Transition version handling to proper semantic versioning when building from Git and release tarballs

### LINKED ISSUE #2003: PyNEST's `setup.py` version does not comply with PEP440
**Describe the bug**
The `setup.py` produced by CMake contains invalid version specifiers such as `python-cmake@6b17f32b1` which do not comply with [PEP440](https://www.python.org/dev/peps/pep-0440/) and lead to warnings in #1959 :

```
/home/robin/.pyenv/versions/3.9.0/envs/nest/lib/python3.9/site-packages/setuptools/dist.py:472: UserWarning: The version specified ('python-cmake@6b17f32b1') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI.
```

### LINKED ISSUE #2116: Report major (+ minor) NEST version even when cloned from git

jougs:
Fixes #2003 and by providing the basic NEST version in a new file `VERSION` in the top-level source directory.

Fixes #2116 by specifying a proper [semantic version](https://semver.org/) as `major.minor.patch-post0.dev0` in the repository. The `-post0.dev0` suffix will be removed from releases.

This PR also removes the possibility to set a custom version suffix for builds with the `-Dwith-version-suffix` switch for CMake as that is now easily possible with the new mechanism based on the `VERSION` file. Less redundancy = more better.


build_support/version_info.sh
```diff
@@ -0,0 +1,43 @@
...
+  REMOTE="unknown"
+fi
+
+# Might fail if we are not on a branch (i.e. in 'detached HEAD' mode)
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [ ! $? -eq 0 ] || [ ${BRANCH} = "HEAD" ]; then
+  BRANCH="unknown"
+fi
+
+# Print ingwith semicolons as separators, as this will be interpreted as
```
Helveg, line 41:
```suggestion
# Printing with semicolons as separators, as this will be interpreted as
```

Helveg (changes requested):
Typo. I find `remote` a confusing name for the remote _branch_. What about `remote_branch` instead?

build_support/version_info.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# This script extracts version control information to be used in
+# cmake/NestVersionInfo.cmake
+
+HASH=$(git rev-parse HEAD)
+
+# Might fail if not on a branch, or no remote tracking branch is set
+REMOTE_BRANCH=$(git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>&1)
+if [ $? -eq 0 ]; then
+  REMOTE=$(echo ${REMOTE_BRANCH} | cut -d\/ -f1)
```
jougs, line 30:
```suggestion
BRANCH_REMOTE=$(git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>&1)
if [ $? -eq 0 ]; then
  REMOTE=$(echo ${BRANCH_REMOTE} | cut -d\/ -f1)
```

jougs:
@Helveg: `remote` is really the name of the remote (as set when `git remote add`ing it). Not sure there's any better name for it...

cmake/NestVersionInfo.cmake
```diff
@@ -17,59 +17,31 @@
...
-#  NEST_VERSION_BRANCH, the current git branch (nest-3.0)
-#  NEST_VERSION_SUFFIX, set using -Dwith-version-suffix=<suffix>. ("-pre")
-#  NEST_VERSION, the numeric version number plus the suffix  ("3.0-pre")
-#  NEST_VERSION_GITHASH, the current git revision hash (empty for tarballs) ("dd47c39ce")
-#  NEST_VERSION_STRING, the full NEST version string ("nest-3.0-pre@dd47c39ce")
+#   NEST_VERSION, the numeric version number including a suffix
+#   NEST_VERSION_GIT, a boolean indicating if source is managed by git
+#   NEST_VERSION_GIT_HASH, the current git revision hash (unset for tarballs)
+#   NEST_VERSION_GIT_BRANCH, the current git branch (unset for tarballs)
+#   NEST_VERSION_GIT_REMOTE, the tracked remote branch (unset for tarballs)
```
jougs, line 27:
```suggestion
#   NEST_VERSION_GIT_REMOTE, the upstream remote (if any) of the tracked branch (unset for tarballs)
```

Helveg (approved):

PR #2718, opened 2023-04-27
Add `-lnest` to the list of libraries to link extension modules

pnbabu:
The build for `nest-extension-module` fails for MacOS with a link error because the `libnest` is missing in the list of libraries to link for extension modules. Adding the `-lnest` option resolves the error.


heplesser (approved):

jougs (commented):
Thanks!

jougs (approved):

PR #2743, opened 2023-04-28
Port test_row_col_pos

med-ayssar:


testsuite/pytests/sli2py_other/test_rows_cols_pos.py
```diff
@@ -0,0 +1,48 @@
...
+def test_correct_position():
+    nest.Create("iaf_psc_alpha", positions=nest.spatial.free(pos=[[0, 0]], extent=[1.0, 1.0]))
+
+
+def test_invalid_shape():
+    with pytest.raises(Exception):
+        nest.Create("iaf_psc_alpha", positions=nest.spatial.grid([1]))
+
+
+def test_shape_with_position():
```
heplesser, line 45: @med-ayssar We need to look at this one again.

heplesser (changes requested):
See inline, will get back to it soon.

testsuite/pytests/sli2py_other/test_rows_cols_pos.py
med-ayssar: Yeah! Those tests do not add much value!

heplesser: One weakness with this test is that it will pass if either of the two `Create()` calls raises and exception, but we want to be sure both do. Each call needs its own `pytest.raises()`. Furthermore, as written, both will fail because `positon` is not a legal argument name. To make some more sense, one could test that `grid` does not allow `pos` and `free` does not allow `shape`. Or one could just drop this test.

heplesser:
I think we don't need this test at all in PyNEST, since one can only pass a `spatial.grid` or a `spatial.free` object as `positions` argument to `Create()`.  Therefore, I close this PR and create a new one moving the SLI test to the retired tests folder.

Issue #2741, opened 2023-04-28
(Python API) MPI deadlock when accessing kernel data


JoseJVS:
**Describe the bug**
When running in a multi-MPI process environment, if a subset of the MPI processes access kernel information (i.e. through direct attribute access or GetKernelStatus function), said subset of processes will enter a deadlock state and never return.
Strangely enough this does not happen if all MPI processes access the kernel information.
I have added a minimal reproducer script but the issue can also be seen in the example microcircuit model when running with more than 1 MPI process.
In the microcircuit case, in the __setup_nest function of the Network class [network.py#L263](https://github.com/nest/nest-simulator/blob/7bb77336e9e8566fe30daa91f050ec477d3d480d/pynest/examples/Potjans_2014/network.py#L263), MPI rank 0 is supposed to print the random number generation seed and the total number of virtual processes. With more than 1 MPI processes, rank 0 is left in a deadlock while the rest of the ranks proceed with the network construction and attempt to simulate at which point the MPI library crashes. In this case, there is an error message printed out but points to a failure of an MPI_Allgather call which is misleading as the original problem is caused by rank 0 entering deadlock state.

**To Reproduce**
1. (Minimal) reproducing example
The issue can be reproduced with the following script.
```python
import nest

rank = nest.Rank()

rng_attr = nest.rng_seed
rng_get = nest.GetKernelStatus('rng_seed')

procs_func = nest.NumProcesses()
procs_attr = nest.num_processes
procs_get = nest.GetKernelStatus('num_processes')

threads_attr = nest.local_num_threads
threads_get = nest.GetKernelStatus('local_num_threads')

vps_attr = nest.total_num_virtual_procs
vps_get = nest.GetKernelStatus('total_num_virtual_procs')

print(f"rank (var): {rank}")
print(f"rank: {nest.Rank()}")

print(f"rng_attr (var): {rng_attr}")
print(f"rng_attr: {nest.rng_seed}")
print(f"rng_get (var): {rng_get}")
print(f"rng_get: {nest.GetKernelStatus('rng_seed')}")

print(f"procs_func (var): {procs_func}")
print(f"procs_func: {nest.NumProcesses()}")
print(f"procs_attr (var): {procs_attr}")
print(f"procs_attr: {nest.num_processes}")
print(f"procs_get (var): {procs_get}")
print(f"procs_get: {nest.GetKernelStatus('num_processes')}")

print(f"threads_attr (var): {threads_attr}")
[...]

PR #2749, opened 2023-04-28
Microcircuit model quick fix for issue #2741

JoseJVS:
This pull request proposes a quick fix for the MPI deadlock at the Python API level when accessing kernel information with a subset of the MPI processes available (cf issue #2741) in the microcircuit model.
In the model, this bug is caused by the __setup_next function in [network.py#L263](https://github.com/nest/nest-simulator/blob/7bb77336e9e8566fe30daa91f050ec477d3d480d/pynest/examples/Potjans_2014/network.py#L263) when printing out the rng seed and total number of virtual processes only by rank 0.
Although this pull request does not fix the issue causing this bug, the proposed quickfix does not alter the behavior of the model in any way.


heplesser:
This quickfix is just the right fix: One should *never* call any `nest.xyz` function inside a block that will only be executed on a subset of MPI ranks.

Since it is a small fix to informational output in an example, I will merge without second review. The failing sphinx_rtd test is due to a configuration error (see #2753), so it should not block merging this PR.

PR #2590, opened 2023-01-17
Separated checks

terhorstd:
These changes decompose the `build_ci.sh` into separate steps to be run by the Github Runner. This allows for a better overview and easier management of job dependencies, matrix jobs, etc.

PR #2750, opened 2023-04-28
Reformat the codebase and enable black check in CI

Helveg:


Helveg:
@heplesser @jougs If you could review the none-cake containing commits and merge this before merging any other Python-containing PRs then we have black in.

Prominent changes:

* Added `black` pre commit hook and enabled CI check; dev docs on the topic was already included in the previous PR #2590 
* Reformatted the Python codebase and added the commit sha to the ignore revs file so that blame isn't cluttered
* Added E203 to the flake8 ignore list because it is not PEP8 compliant and Black and flake do not agree about it.

heplesser (changes requested):
@Helveg Thank you for this PR. My main concern is that it reduces the line length to 80. While this is in line with PEP8, we have concluded that for C++ code we allow 120, because it makes code more readable. In the diff I noticed numerous lines that are now broken although just slightly over 80, not improving legibility. So I would strongly argue for using 120 also for Python code.

And could you elaborate on the SHAs? I don't quite understand what that is about.

Helveg:
Oh, I thought I had set the line length config to 120, I will double check! The SHA removes the commit from any `git blame` checking, otherwise my reformatting commit would show up everywhere as the person last changing that code, and more meaningful history and blame is occluded

Helveg:
@heplesser I force pushed a new attempt; the line length is now set in `pyproject.toml`. 

Small question: why are so many tool settings set in `tox.ini` and not in `pyproject.toml`? Maybe this predates the `pyproject.toml` support?

heplesser:
> @heplesser I force pushed a new attempt; the line length is now set in `pyproject.toml`.
> 
> Small question: why are so many tool settings set in `tox.ini` and not in `pyproject.toml`? Maybe this predates the `pyproject.toml` support?

@Helveg History is probably the most plausible explanation. I don't think it was an explicit choice of `tox.ini` in favour of `pyproject.toml`. I think when I first started with tox (in another context), I came across instructions that gave a minimal "do not change" pyproject file and lots of details of what to put into `tox.ini`. Feel free to create a new PR with a better structure (and ideally links to documentation of the file formats/purposes).

pynest/examples/balancedneuron.py
```diff
@@ -65,16 +65,16 @@
...
+n_ex = 16000  # size of the excitatory population
+n_in = 4000  # size of the inhibitory population
+r_ex = 5.0  # mean rate of the excitatory population
+r_in = 20.5  # initial rate of the inhibitory population
+epsc = 45.0  # peak amplitude of excitatory synaptic currents
+ipsc = -45.0  # peak amplitude of inhibitory synaptic currents
+d = 1.0  # synaptic delay
+lower = 15.0  # lower bound of the search interval
+upper = 25.0  # upper bound of the search interval
+prec = 0.01  # how close need the excitatory rates be
```
heplesser, line 77: In places like this, I think readbility is better if the comments are lined up. Is that feasible?

pynest/examples/clopath_synapse_spike_pairing.py
```diff
@@ -46,56 +46,59 @@
...
-    [ 10.0,   60.0,  110.0,  160.0,  210.0],          # noqa
-    [ 10.0,   43.3,   76.7,  110.0,  143.3],          # noqa
-    [ 10.0,   35.0,   60.0,   85.0,  110.0],          # noqa
-    [ 10.0,   30.0,   50.0,   70.0,   90.0],          # noqa
-    [130.0,  230.0,  330.0,  430.0,  530.0,  630.0],  # noqa
-    [ 80.0,  130.0,  180.0,  230.0,  280.0,  330.0],  # noqa
-    [ 63.3,   96.6,  130.0,  163.3,  196.6,  230.0],  # noqa
-    [ 55.0,   80.0,  105.0,  130.0,  155.0,  180.0],  # noqa
-    [ 50.0,   70.0,   90.0,  110.0,  130.0,  150.0]]  # noqa
+    [10.0, 110.0, 210.0, 310.0, 410.0],  # noqa
```
heplesser, line 91: The `# noqa` seems to have no effect here and the table formatting is destroyed, similarly above. Can this be avoided?

pyproject.toml
```diff
@@ -5,3 +5,6 @@ markers = [
     "skipif_missing_threads: skip test if NEST was built without multithreading support",
     "simulation: the simulation class to use. Always pass a 2nd dummy argument"
 ]
+
+[tool.black]
+line-length = 120
```
heplesser, line 10: Newline missing at end, it seems.

heplesser (commented):
I have a few questions concerning places where code had been hand-formatted for readability. I have not systematically searched for all such cases yet, would like your feedback first.

pynest/examples/balancedneuron.py
Helveg: Yes, we can use `# fmt: off` and `# fmt: on` blocks

pynest/examples/clopath_synapse_spike_pairing.py
Helveg: same as above

pynest/examples/balancedneuron.py
Helveg: @heplesser How can we go about identifying these locations?

heplesser: I think we would need to build a little parser that can identify subsequent lines in which `#` are
- aligned
- at least one `#` is more than two spaces from preceding code
- not sure if we have any cases with just a single space in front of the `#`, if so we would need to handle that, too.
Not sure if one can express that as a regular expression.
Then run it on the pre-black code.

Helveg: I think in this case a Python script might be much easier to make than a regular expression, especially given the one off nature of the task :) I'll get on that.

nicolossus: Just a suggestion. We could also denote comments that should be aligned by `#:`, that is,

```
 n_ex = 16000  #: size of the excitatory population
 n_in = 4000   #: size of the inhibitory population
 r_ex = 5.0    #: mean rate of the excitatory population
```

This will most likely make parser script easier and less prone for making unwanted alignments elsewhere.

nicolossus: This would be a convention introduced by us, and as far as I know not something that can be utilized by black. I think we have to manually add `#:` to the blocks where we want comments to be aligned for readability. These will predominately be located in example scripts, as the one @heplesser referred to above. 

Ideally, I think we want a parser script that runs after black and aligns the comment blocks with `#:`. The example scripts are prone to changes and we want the alignments to be continuously updated in case we, for instance, change the length of a variable name. 

If we want to go with this solution, it could also probably be a separate PR. @heplesser Thoughts?

Helveg: Usually scripts that modify the source code from `push`-CI is discouraged, because it adds more commits and CI-runs, and because we'd have to make and maintain a GitHub API integrations (the runners have no git authentication with write-access to the repo, iirc, and by design).

If we want such a script it might be an idea to add it to the same `pre-commit` tool that runs the black pre-commit hook locally on a dev's machine before it is even pushed to the remote and any CI check runs. Then, in CI, we can run a simple check and reject the CI if it contains any misaligned `#:`. This would be completely analogous to how `black` is used in the repo.

heplesser: Isn't it possible to put a `#noqa` tag around such blocks?

heplesser (approved):

terhorstd (approved):
LGTM! :+1:

nicolossus (approved):
LGTM!

Helveg:
@heplesser I am surprised? So we are fine with disrupting all the currently aligned comments in the codebase?

PR #2754, opened 2023-04-29
Port `test_multiple_multimeter` from SLI-2-Py

nicolossus:
The diff is a bit messy, partly due to our new auto-formatter. I have ported the tests from `test_multiple_multimeter.sli` into `test_multimeter.py`. One test needed a subset of the models with recordables, which necessitated defining `skip_models` and `extra_params`. We should be able to remove most of this clutter when we add more complex model extraction directly from PyNEST in the future.

Also, due to the theme of `test_multimeter_freeze()`, I relocated it from `test_multimeter.py` to `test_multimeter_freeze_thaw.py`.


testsuite/pytests/sli2py_recording/test_multimeter_freeze_thaw.py
```diff
@@ -52,16 +57,30 @@ def simulate_freeze_thaw(num_neurons):
 
     nrns[0].frozen = True
     nest.Simulate(5.0)
+    nest.Simulate(5.0)
```
heplesser, line 60: Here and in a number of other places down to line 120 it seems as if some lines are doubled up in the new version.

heplesser (changes requested):
Overall fine, but see inline comment.

testsuite/pytests/sli2py_recording/test_multimeter_freeze_thaw.py
nicolossus: Duplicates from merge error removed in [4b020be](https://github.com/nest/nest-simulator/pull/2754/commits/4b020be7abd443702c60e241f987d8963979f8e5).

testsuite/pytests/test_multimeter.py
```diff
@@ -68,44 +152,35 @@ def test_recordables_are_recorded(set_resolution, model):
         assert len(result[r]) == num_data_expected
 
 
-@pytest.mark.parametrize("model", models)
-def test_correct_amount_data_collected(model):
+@pytest.mark.parametrize("model", subset_models_with_rec)
+def test_identical_recording_from_multiple_multimeters(model):
```
heplesser, line 156: Comparing to the SLI test, I don't think it it necessary to run this test for all neuron models. Doing it for a single model should suffice and save quite a bit of energy over time. We then also do not need the skip list.

heplesser (approved):

PR #2758, opened 2023-05-02
Add documentation about deadlock in mpi processes 

jessica-mitchell:
This PR add a note in the documentation in the distributed computing section of the parallel computing page, regarding the issue resolved in  #2749.

See output [here](https://nest-test.readthedocs.io/en/docs-mpi-update/hpc/parallel_computing.html#using-distributed-computing)

### DIFF
--- a/doc/htmldoc/hpc/parallel_computing.rst
+++ b/doc/htmldoc/hpc/parallel_computing.rst
@@ -241,6 +241,31 @@ command should look like this
 Please refer to the documentation of your MPI implementation to learn
 more about the usage of ``mpirun``.
 
+
+.. important::
+
+    One should never call any ``nest.*`` function inside a block that will only be executed on a subset of MPI ranks.
+
+    Trying to access kernel information with a subset of MPI processes causes a deadlock.
+
+    For example:
+
+    **Don't do this**
+
+    .. code-block::
+
+      if nest.Rank() == 0:
+            print('RNG seed: {}'.format(
+                nest.rng_seed))
+
+
+    **Do this**
+
+    .. code-block::
+
+     if nest.Rank() == 0:
+        print('RNG seed: {}'.format(rng_seed))
+
 MPI related commands
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/htmldoc/hpc/parallel_computing.rst
+++ b/doc/htmldoc/hpc/parallel_computing.rst
@@ -228,7 +228,7 @@ Run distributed simulations
 
 Distributed simulations **cannot be run interactively**, which means that
 the simulation has to be provided as a script. However, the script can be the same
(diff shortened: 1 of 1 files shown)


doc/htmldoc/hpc/parallel_computing.rst
```diff
@@ -241,6 +241,31 @@ command should look like this
...
+
+    For example:
+
+    **Don't do this**
+
+    .. code-block::
+
+      if nest.Rank() == 0:
+            print('RNG seed: {}'.format(
+                nest.rng_seed))
```
nicolossus, line 259:
```suggestion
      if nest.Rank() == 0:
            rng_seed = nest.rng_seed
            print(f"RNG seed: {rng_seed}")
```

I think for the purpose of illustration, the `nest.*` call should be on a a line of its own. This also makes the comparison with the next code block more clear. Moreover, we should use f-strings.

doc/htmldoc/hpc/parallel_computing.rst
```diff
@@ -241,6 +241,31 @@ command should look like this
...
+
+
+    **Do this**
+
+    .. code-block::
+
+     if nest.Rank() == 0:
+        print('RNG seed: {}'.format(rng_seed))
```
nicolossus, line 267:
```suggestion
     rng_seed = nest.rng_seed
     if nest.Rank() == 0:
        print(f"RNG seed: {rng_seed}")
```
The `nest.*` call should also be included here.

nicolossus (changes requested):
@jessica-mitchell Thanks! I have some minor suggestions for the example code, but overall this looks good and is a very useful addition to the documentation.

nicolossus (approved):

JoseJVS (commented):
Great! This is exactly what was needed. Perhaps could you add a sentence like:
```
The PyNEST API was designed so that all parallel operations are handled through the nestkernel, hence the users don't need to handle MPI communication from the python script.
Ideally the script should be written in a serial manner, however if specific operations unrelated to the simulation execution are needed, the NEST API offers the Rank() function to retrieve the MPI rank of the process and handle process specific tasks.
Note that one should never call any ....
```

So that it is clear that parallelization is fully handled from the kernel and not the python script.

heplesser (changes requested):
@jessica-mitchell Thanks for this PR! I wonder if it would not work even better if the "Important" box came at the end of the "MPI related commands" section that is now just below the "Important" box. Also, in the "Run distributed simulations" section, an "r" is missing in "distibuted simulation scripts". Could you fix that as part of this PR?

Concerning the suggestion by @JoseJVS, I think this is pretty much covered if one looks at the "Using distributed computing" section as a whole, so adding another sentence would make it a bit repetitive.

heplesser (approved):

PR #2762, opened 2023-05-02
Build NEST with Python, OpenMP, MPI, Boost and GSL for macOS on CI

### LINKED ISSUE #2756: New test setup runs only minimal tests on macOS
The new test setup created in #2590 only performs minimal checks under macOS, as it is built without support for MPI, OpenMP, GSL, Python (see https://github.com/nest/nest-simulator/actions/runs/4858443145/jobs/8659888539#step:8:11, https://github.com/nest/nest-simulator/actions/runs/4858443145/jobs/8659888539#step:12:295). It should run the full set of tests except those requiring sionlib, music, libneurosim. 

See also #2660 for fixes to the old CI setup that allowed running test with OpenMP again under macOS. The crucial point was to inform Cmake about the location of the OpenMP library installed by Homebrew:

```
CONFIGURE_OPENMP="${CONFIGURE_OPENMP} -DOpenMP_ROOT=`brew --prefix libomp`"
```

I'd create a PR myself but feel a bit uncertain about the new setup, so I create this issue.

heplesser:
This PR ensures NEST is build with all components on macOS on the CI (except sionlib, libneurosim and music). It also removes some no longer needed parts of the `build_macos` stage. It incorporates changes from #2660. It fixes #2756.


.github/workflows/nestbuildmatrix.yml
```diff
@@ -666,6 +659,7 @@ jobs:
               -Dwith-warning=${{ contains(matrix.use, 'warning') && 'ON' || 'OFF' }} \
               -Dwith-boost=${{ contains(matrix.use, 'boost') && 'ON' || 'OFF' }} \
               -Dwith-openmp=${{ contains(matrix.use, 'openmp') && 'ON' || 'OFF' }} \
+              ${{ contains(matrix.use, 'openmp') && '-DOpenMP_ROOT=`brew --prefix libomp`' }} \
```
terhorstd, line 662: The use of back ticks is discouraged due to unclear nesting in many situations.
```suggestion
              ${{ contains(matrix.use, 'openmp') && '-DOpenMP_ROOT=$(brew --prefix libomp)' }} \
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -666,6 +659,7 @@ jobs:
...
               -Dwith-mpi=${{ contains(matrix.use, 'mpi') && 'ON' || 'OFF' }} \
```
terhorstd, line 663: Check the logs for the MPI config in the `cmake` step. There seems to be some mixup between `brew` installed and system installed MPI versions:
```
 Use MPI             : Yes (MPI: /usr/local/bin/mpicxx)
    Includes        : /usr/local/Cellar/open-mpi/4.1.5/include
    Libraries       : /usr/local/Cellar/open-mpi/4.1.5/lib/libmpi.dylib
    Link Flags      : -L/usr/local/Cellar/libevent/2.1.12/lib
    Launcher        : /usr/local/bin/mpiexec -n <np> <prog> <args>
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -687,7 +681,6 @@ jobs:
           make install
         env:
           xNEST_BUILD_TYPE: ${{ matrix.NEST_BUILD_TYPE }}
```
terhorstd, line 683: If this is removed from the matrix, the empty variable can probably go away here too.
```suggestion
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -652,12 +651,6 @@ jobs:
         run: |
           mkdir -p "$NEST_VPATH/reports" "$NEST_RESULT"
 
-          if [ "$xNEST_BUILD_COMPILER" = "CLANG" ]; then
-              export CC=clang-11
-              export CXX=clang++-11
-          fi
-          export CXX_FLAGS="-pedantic -Wextra -Wno-unknown-pragmas -D_GLIBCXX_ASSERTIONS"
```
terhorstd, line 659: Removing this line could make the build a lot less strict than the Linux build. Was this intended?
```bash
export CXX_FLAGS="-pedantic -Wextra -Wno-unknown-pragmas -D_GLIBCXX_ASSERTIONS"
```

terhorstd (changes requested):
Thanks for starting to clean this up! There are a couple of smaller things that might need another look before merging.

.github/workflows/nestbuildmatrix.yml
heplesser: Welcome to the wondrous world of brew. The binaries in `/usr/local/bin` are symlinks to the actual files in `/usr/local/Cellar/open-mpi/...`, so this is consistent.

.github/workflows/nestbuildmatrix.yml
heplesser: Here I am out of my depth. What do you mean by "empty variable"? And could we then remove the entire `env:` section? And where is `NEST_BUILD_TYPE` actually set?

.github/workflows/nestbuildmatrix.yml
heplesser: Here we need more cleanup, I think, since the linux setup also seems inconsistent. For Linux, we first set

https://github.com/nest/nest-simulator/blob/7d9150c0bffd71e3fdce0318d94406303554045e/.github/workflows/nestbuildmatrix.yml#L401-L404

and then later change this to

https://github.com/nest/nest-simulator/blob/7d9150c0bffd71e3fdce0318d94406303554045e/.github/workflows/nestbuildmatrix.yml#L540

Shouldn't we set it in one place only?

Also, why do we have 

https://github.com/nest/nest-simulator/blob/7d9150c0bffd71e3fdce0318d94406303554045e/.github/workflows/nestbuildmatrix.yml#L536-L538

where is `NEST_BUILD_COMPILER` set and why are we binding ourselves to `clang-11`, while the most recent version is 16.0?

Once we have a definite version for Linux, we can use the same for macOS, although we may need to remove `-pedantic` in light of this comment from the old setup:

https://github.com/nest/nest-simulator/blob/03469faebf14f79fb091e170fcc0531ea680acce/build_support/ci_build.sh#L185-L188

Finally and unrelated, what is the purpose of this commented-out section (and the corresponding one for Linux):

https://github.com/nest/nest-simulator/blob/7d9150c0bffd71e3fdce0318d94406303554045e/.github/workflows/nestbuildmatrix.yml#L250

.github/workflows/nestbuildmatrix.yml
terhorstd: empty meaning it is not set and therefore results in an empty string being set as `xNEST_BUILD_TYPE` variable for the script.
The expression `matrix.NEST_BUILD_TYPE` refers to the `strategy:matrix:` defining the jobs, not a shell variable. The `env:` section would transfer this into a `xNEST_BUILD_TYPE`-named shell environment variable.
If it's empty, the `env:` can be dropped completely.

.github/workflows/nestbuildmatrix.yml
terhorstd: You are right, the `CXX_FLAGS` should be set only in one place, preferably the `env:` but probably with the value of the `export`.  In a separate issue we can take a deeper look into `installcheck`, in how far this still is a problem.

The explicit setting of `CC` and `CXX` was necessary at some point because for the clang builds on the runners a wrong compiler was picked up somewhere. This might also be completely obsolete, you are right.

Given that this PR handles the MacOS build, please leave the `build_linux` as is and open a separate issue.

heplesser: OK, tidied up and will create new issue for the Linux part.

heplesser: #2771 is the follow-up for Linux.

heplesser:
@terhorstd @lekshmideepu Ping!

lekshmideepu:
> @terhorstd @lekshmideepu Ping!

Am working on this.

.github/workflows/nestbuildmatrix.yml
lekshmideepu: > You are right, the `CXX_FLAGS` should be set only in one place, preferably the `env:` but probably with the value of the `export`. In a separate issue we can take a deeper look into `installcheck` to see in how far the `-pedantic` still is a problem.
> 
> The explicit setting of `CC` and `CXX` was necessary at some point because for the clang builds on the runners a wrong compiler was picked up somewhere. This might also be completely obsolete, you are right.
> 
> Given that this PR handles the MacOS build, please leave the `build_linux` as is and open a separate issue.

yes, you are right. Explicit setting of CC and CXX was kind of necessary due to the small technical issue like it was picking up a wrong compiler..

lekshmideepu (commented):
Thanks @heplesser for this nice work for the macOS builds. Few of my comments/ questions were asked by Dennis and got it answered.  It looks good to me 👍

lekshmideepu (approved):

nicolossus (changes requested):
I am not sure why there is no conflict, but it looks to me that this PR will remove the build with SONATA support on macOS if it is merged now. Perhaps a merge with master will retrieve all the SONATA related changes to the build matrix.

heplesser:
> I am not sure why there is no conflict, but it looks to me that this PR will remove the build with SONATA support on macOS if it is merged now. Perhaps a merge with master will retrieve all the SONATA related changes to the build matrix.

I have merged master into this branch.

.github/workflows/nestbuildmatrix.yml
```diff
@@ -621,19 +621,18 @@ jobs:
       matrix:
         os: [macos-latest]
         cpp_compiler: ["clang"]
-        NEST_BUILD_TYPE: ["FULL_MACOS"]
+
+        # available use flags (all default to "OFF"):
+        # openmp, mpi, python, gsl, ltdl, boost, sionlib, libneurosim, optimize, warning, userdoc, music
+        use:
+          - "openmp, mpi, python, gsl, ltdl, boost, optimize, warning"
```
nicolossus, line 628:
```suggestion
          - "openmp, mpi, python, gsl, ltdl, boost, hdf5, optimize, warning"
```

heplesser: @nicolossus That should be fixed now.

nicolossus: @heplesser LGTM. You may resolve the conversation.

heplesser:
@terhorstd Are you happy as well?

terhorstd (approved):
LGTM! :+1:

PR #2763, opened 2023-05-02
Get Sphinx build working in CI

jessica-mitchell:
This  PR updates the nestbuildmatrix for sphinx-rtd so that it builds Sphinx.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -335,7 +335,7 @@ jobs:
       - name: "Set up Python 3.x"
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.8
 
       - name: "Install dependencies"
         run: |
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -335,6 +335,7 @@ jobs:
       - name: "Set up Python 3.x"
         uses: actions/setup-python@v4
         with:
+          # Using 3.8 because Read the docs does not work with higher versions.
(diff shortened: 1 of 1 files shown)


.github/workflows/nestbuildmatrix.yml
```diff
@@ -335,22 +335,24 @@ jobs:
       - name: "Set up Python 3.x"
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.8
```
terhorstd, line 338: What is the critical part depending on the old version?

terhorstd (changes requested):
Hey! Thanks for the fixes! – Maybe add some comments with a bit of context, so that the next person looking at this understands the motivation.

jessica-mitchell:
@terhorstd I added some comments with links to better explain the install for pandoc and specifying the Python version

.github/workflows/nestbuildmatrix.yml
terhorstd: Thanks for adding the comment.

.github/workflows/nestbuildmatrix.yml
```diff
@@ -335,22 +335,24 @@ jobs:
...
 
       - name: "Install dependencies"
         run: |
           pip install -r doc/requirements.txt
+          sudo apt-get install pandoc
```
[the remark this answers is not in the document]
jessica-mitchell: I added a comment (See my change) , pip does not lead to broken version, from what I understand is the 'executable' part is not pip installable

terhorstd (approved):
LGTM! :)

PR #2764, opened 2023-05-03
Port `test_GetConnections` from SLI-2-Py

nicolossus:
Port `test_GetConnections.sli` to Pytest.


heplesser (changes requested):
As discussed in person.

heplesser (approved):

PR #2766, opened 2023-05-05
Update readthedocs.yml to resolve build error from incompatible urllib3 version

jessica-mitchell:
This PR resolves an error brought on by a recent external update to the urllib3 version that was incompatible with the version of Read the Docs we are using. The solution is to update the yml file to include the os build so a more recent version is used to build the docs. 

See the issue on the Read the Docs issue tracker: https://github.com/readthedocs/readthedocs.org/issues/10290

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,13 +1,14 @@
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.8"
+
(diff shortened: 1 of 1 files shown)


steffengraber (approved):
:+1:  Looks good!

jessica-mitchell:
Merging this with 1 reviewer, as its documentation

Issue #545, opened 2016-11-11
SetDefaults on connections with wrong DataType crashes Nest


flinz:
In the [master branch](https://github.com/nest/nest-simulator/commit/1fd378f8d95d64ed0af32fadd373a9500a68e719) calling ``SetDefaults`` on a connection with a wrong datatype causes Nest to ``terminate called without an active exception``.  In Nest 2.10 this correctly raised a NestError. Importantly, for neuron models this still works correctly, see below. 

Missing we be missing something in our unittests?
I am unsure whether this is connected to #249, since this actually causes a crash not a silent fail.

Test Bash Script
```bash
echo "Try setting neuron C_m with float"
nest -c "/iaf_neuron << /C_m 10. >> SetDefaults"

echo "------"
echo "Try setting synapse tau_plus with float"
nest -c "/stdp_synapse << /tau_plus 10. >> SetDefaults"

echo "------"
echo "Try setting neuron C_m with int (should fail)"
nest -c "/iaf_neuron << /C_m 10 >> SetDefaults"

echo "------"
echo "Try setting synapse tau_plus with int (should fail)"
nest -c "/stdp_synapse << /tau_plus 10 >> SetDefaults"
```

In [Nest 2.10](https://github.com/nest/nest-simulator/releases) this behaves as expected
```bash
Try setting neuron C_m with float
------
Try setting synapse tau_plus with float
------
Try setting neuron C_m with int (should fail)

Nov 11 16:13:39 SetDefaults_l_D [Error]: TypeMismatch
    Expected datatype: doubletype
    Provided datatype: integertype
------
Try setting synapse tau_plus with int (should fail)

Nov 11 16:13:39 SetDefaults_l_D [Error]: TypeMismatch
    Expected datatype: doubletype
    Provided datatype: integertype
```

In the [master branch](https://github.com/nest/nest-simulator/commit/1fd378f8d95d64ed0af32fadd373a9500a68e719) the last command crashes Nest
```bash
Try setting neuron C_m with float
------
Try setting synapse tau_plus with float
------
Try setting neuron C_m with int (should fail)

Nov 11 16:23:52 SetDefaults_l_D [Error]: TypeMismatch
    Expected datatype: doubletype
    Provided datatype: integertype
------
Try setting synapse tau_plus with int (should fail)

terminate called without an active exception
test.sli: line 14: 41387 Abort trap: 6           nest -c "/stdp_synapse << /tau_plus 10 >> SetDefaults"
```

[...]

PR #704, opened 2017-04-06
Fixed SetDefaults for synapse models

janhahne:
This PR fixes #545.

PR #2769, opened 2023-05-05
Port `issue-545.sli` to py

otcathatsya:
It looks like the original SLI test may not have covered the intended error case to begin with? The [original report](https://github.com/nest/nest-simulator/issues/545) specifies that passing integers leads to a crash/exception, though the SLI test seems to use floats only. In either case this is not an issue with python, as even integers are correctly cast.

Edit: it appears that a 2021 commit removed the integer part of the original SLI test since it became obsolete, purpose of the current test is a bit questionable


heplesser:
The key point here was that prior to #704 which fixed #545, NEST would actually crash when `SetDefaults()` contained unacceptable values. This would happen because we did not catch exceptions inside threads. A proper test for this regression would be to use parameters forcing an exception (e.g., setting a time constant to a negative value) and test for this with `pytest.raises()`. If NEST then crashed, the test would fail. We don't need to include passing cases here, and we only need to do this for a case with > 1 thread.

nicolossus:
@otcathatsya Can you please pull master into this branch so that you get the CI pipeline with the black checker enabled?

heplesser (approved):

Issue #2315, opened 2022-02-25
Exceptions for unused dictionary items for multiple compartments/receptors


boettcherj:
**Is your feature request related to a problem? Please describe.**
There is not error for unused dictionary items when using SetStatus for multiple compartments and or receptors at the same time. However there is an error when adding a single compartment or a single receptor.

**Describe the solution you'd like**
Implementation of the same DictError that occurs when single compartments are added with false parameters.

**Describe alternatives you've considered**
N/A

**Additional context**
For example if we want a compartmental neuron with two compartments, we currently need to add both at the same time. However the example below does not cause any errors, which makes it hard to debug the simulations since there are natually a lot of parameters in all the dictionaries.
```
cm = nest.Create('cm_default')

soma_params = {'unused key': 0}

dend_params = {'unused key': 0}

cm.compartments = [
    {"parent_idx": -1, "params": soma_params},
    {"parent_idx":  0, "params": dend_params}
]
 
receptors = [
    {"comp_idx": 0, "receptor_type": "GABA", "unused_key": 0, "params": {"unused_key": 0}},
    {"comp_idx": 0, "receptor_type": "AMPA", "unused_key": 0, "params": {"unused_key": 0}}
]

cm.receptors = receptors
```

heplesser:
@WillemWybo Ping!

heplesser:
Should be fixed by #2773.

PR #2773, opened 2023-05-08
User configurable voltage initialization for the compartmental model

WillemWybo:
The initial voltage is now made configurable through a parameter "V_init" in the status dictionary.

I have also added functionality to check for unused dictionary items in compartment and receptor parameter dictionaries. However, someone should probably review it to check whether the way I implemented it is the best way. Solves #2315


clinssen:
@WillemWybo : could you fix the merge conflict? Cheers!

testsuite/pytests/test_compartmental_model.py
```diff
@@ -372,33 +372,34 @@ def create_2tdend_4comp(dt=0.1):
...
+        DP[4]['g_L'] * (v_init / 2. - DP[4]['e_L'])
+
+    bb[6] = DP[5]['C_m'] / dt * v_init - \
+        DP[5]['g_L'] * (v_init / 2. - DP[5]['e_L'])
+
+    bb[7] = DP[6]['C_m'] / dt * v_init - \
+        DP[6]['g_L'] * (v_init / 2. - DP[6]['e_L'])
+
+    bb[8] = DP[7]['C_m'] / dt * v_init - \
+        DP[7]['g_L'] * (v_init / 2. - DP[7]['e_L'])
```
jougs, line 402: I don't see why these lines need breaking. They don't seem to be longer than 120 characters. The same applies above.

models/cm_compartmentcurrents.h
```diff
@@ -466,6 +468,8 @@ class CompartmentCurrents
     {
       assert( false );
     }
+
+    ALL_ENTRIES_ACCESSED( *receptor_params, "receptor_params", "Unread dictionary entries: " );
```
jougs, line 472: The dict access checking looks good to me.

models/cm_compartmentcurrents.cpp
```diff
@@ -50,6 +50,40 @@ nest::Na::Na( const DictionaryDatum& channel_params )
   }
 }
 
+void
+nest::Na::pre_run_hook( const double v_init )
+{
+  // compute limiting values for m and h given v_init as the initial state
```
jougs, line 56: This looks a bit like it should rather be a Doxygen-style comment for the function in the header file. Also the one below with the link to the reference.

models/cm_default.cpp
```diff
@@ -70,6 +72,7 @@ void
 cm_default::get_status( DictionaryDatum& statusdict ) const
 {
   def< double >( statusdict, names::V_th, V_th_ );
+  def< double >( statusdict, "V_init", V_init_ );
```
jougs, line 75: For consistency reasons, I think it would be better to also define `V_init` as a pre-defined name in `nestkernel/nest_names.{h,cpp}` instead of using the string directly.

testsuite/pytests/test_compartmental_model.py
```diff
@@ -962,6 +966,145 @@ def test_add_assign(self):
...
         self.assertEqual(len(list(cm2.receptors)), 3)
 
+    def test_custom_v_init(self):
+        sp0 = {'C_m': 1.00, 'g_C': 0.00, 'g_L': 0.100, 'e_L': -60.0}
+        dp0 = {'C_m': 0.10, 'g_C': 0.10, 'g_L': 0.010, 'e_L': -60.0}
+        sp1 = {'C_m': 1.00, 'g_C': 0.00, 'g_L': 0.100, 'e_L': -75.0}
+        dp1 = {'C_m': 0.10, 'g_C': 0.10, 'g_L': 0.010, 'e_L': -80.0}
+        sp2 = {'C_m': 1.00, 'g_C': 0.00, 'g_L': 0.100, 'e_L': -70.0}
+        dp2 = {'C_m': 0.10, 'g_C': 0.10, 'g_L': 0.010, 'e_L': -70.0}
+        v_init = -60.
```
med-ayssar, line 976: Maybe also an opportunity to change your test to the new modern `pytest` and would be better to put those lines inside a **pytest fixture**.

testsuite/pytests/test_compartmental_model.py
```diff
@@ -962,6 +966,145 @@ def test_add_assign(self):
...
+        h_Na_0 = 1. / (1. + np.exp((v + 65.) / 6.2))
+
+        # activation functions
+        alpha_n = 0.02 * (v - 25.) / (1. - np.exp(-(v - 25.) / 9.))
+        beta_n = -0.002 * (v - 25.) / (1. - np.exp((v - 25.) / 9.))
+        n_K_0 = alpha_n / (alpha_n + beta_n)
+
+        self.assertTrue(np.allclose(mm.events["m_Na_0"][0], m_Na_0))
+        self.assertTrue(np.allclose(mm.events["h_Na_0"][0], h_Na_0))
+        self.assertTrue(np.allclose(mm.events["n_K_0"][0], n_K_0))
```
med-ayssar, line 1058: Those are too many tests for different cases. Better to have this pattern for each test: *create nodes -> create network -> simulate -> test result*. Again with **pytest fixtures** it will be cleaner.

med-ayssar (commented):
Mainly my review was focusing on the *unitest*. Maybe take a look at our newly transpiled *sli tests* to *pytest* to have an idea about the syntax being used.

For the *pre_run_hook* it is better to stay consistent with the other interfaces provided by the NEST kernel, and thus I am in favor of removing the newly introduced parameter to this method.

models/cm_compartmentcurrents.h
WillemWybo: Ok, thx!

models/cm_compartmentcurrents.cpp
WillemWybo: I have implemented this comment style (while also following @med-ayssar 's comment to refactor the state variable activations into their own functions)

models/cm_default.cpp
WillemWybo: I have added this to `nest_names.{h,cpp}`. Note that as per @med-ayssar 's suggestion, I have removed the `v_init` argument from all `pre_run_hook()` functions. This means that the initial voltage is now set for each compartment as part of the `compartment_params` dictionary.

models/cm_compartmentcurrents.cpp
```diff
@@ -169,6 +203,40 @@ nest::K::K( const DictionaryDatum& channel_params )
...
+    alpha_n = 0.02 * v_init_minus_25 * exp_vm25_div_9 * frac_evm25d9;
+    double beta_n = 0.002 * v_init_minus_25 * frac_evm25d9;
+    frac_alpha_plus_beta_n = 1. / ( alpha_n + beta_n );
+  }
+  else
+  {
+    alpha_n = 0.18;
+    double beta_n = 0.018;
+    frac_alpha_plus_beta_n = 1. / ( alpha_n + beta_n );
+  }
```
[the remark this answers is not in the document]
WillemWybo: I have refactored these code segments into separate function (i.e. `compute_statevar_m()` and `compute_statevar_h()` for the Na-channel, and `compute_statevar_n()` for the K-channel)

testsuite/pytests/test_compartmental_model.py
WillemWybo: I would propose to implement a refactorization of the compartmental model tests in a separate PR, as it is not really related to the topic of this PR specficially, and create an associated issue outlining all problems that need to be addressed.

testsuite/pytests/test_compartmental_model.py
WillemWybo: I would propose to implement a refactorization of the compartmental model tests in a separate PR, as it is not really related to the topic of this PR specficially, and create an associated issue outlining all problems that need to be addressed.

models/cm_default.h
```diff
@@ -124,19 +124,28 @@ variables, use the receptor index ``{state_variable_name}{receptor_index}``:
...
 
     dc = nest.Create('dc_generator', {...})
-    nest.Connect(dc, cm, syn_spec={..., 'receptor_type': 0}
+    nest.Connect(dc, cm, syn_spec={..., 'receptor_type': 0})
 
 Parameters
 ++++++++++
 
+Note that the compartmental model does not explicitly ensure that units are consistent.
+Therefor, it is on the user to ensure that units are consistent throughout the model.
```
clinssen, line 141:
```suggestion
Therefore, it is on the user to ensure that units are consistent throughout the model.
```

models/cm_default.h
```diff
@@ -146,10 +155,11 @@ The following parameters can be set in the status dictionary.
 The following parameters can be used when adding compartments using ``SetStatus()``
 
 =========== ======= ===============================================================
- C_m        uF      Capacitance of compartment (default: 1 uF)
+ C_m        nF      Capacitance of compartment (default: 1 nF)
  g_C        uS      Coupling conductance with parent compartment (default: 0.01 uS)
  g_L        uS      Leak conductance of the compartment (default: 0.1 uS)
  e_L        mV      Leak reversal of the compartment (default: -70. mV)
+ V_init     mV      Initialization voltage of the compartment (default: -75. mV)
```
clinssen, line 162: Why is the default different from the leak potential? Doesn't this cause a transient at the start of the simulation?

testsuite/pytests/test_compartmental_model.py
```diff
@@ -959,8 +959,12 @@ def test_continuerun(self, dt=0.1):
         m_neat = nest.Create("multimeter", 1, {"record_from": recordables, "interval": 1.0})
         nest.Connect(m_neat, n_neat)
 
-        nest.Simulate(12.0)
-        nest.Simulate(88.0)
+        # test the case where we continue a run without resetting the neural state
```
clinssen, line 962: Should repeated calls to Simulate() still be tested? Or should this be forbidden?

testsuite/pytests/test_compartmental_model.py
```diff
@@ -1018,6 +1022,132 @@ def test_add_assign(self):
...
+            {"parent_idx": -1, "params": sp0},
+            {"parent_idx": 0, "params": dp0},
+        ]
+
+        mm = nest.Create("multimeter", 1, {"record_from": ["v_comp0", "v_comp1"], "interval": 1.0})
+        nest.Connect(mm, cm)
+
+        nest.Simulate(10.0)
+
+        self.assertTrue(np.allclose(mm.events["v_comp0"], -60.0))
```
clinssen, line 1047:
```suggestion
        self.assertTrue(np.allclose(mm.events["v_comp0"], sp0["V_init"]))
```

models/cm_default.h
WillemWybo: I've set the default to -70 mV now. This would've indeed caused a transient which now won't happen if there are no other conductances. Nevertheless, when there are non-zero ion channels the equilibrium will most likely not be equal to the initialization, nor the leak reversal, so there will be a transient anyway.

testsuite/pytests/test_compartmental_model.py
WillemWybo: I don't know how this code with the two `Simulate()` calls got implemented in the test, my version (and I've reverted the implementation to this now) is:
"""
        nest.Prepare()
        nest.Run(12.0)
        nest.Run(88.0)
        nest.Cleanup()
 """
It tests whether continuing the simulation without resetting the internal state of the model works as expected. Or is this not a proper use of these functions? As far as I'm aware, two `Simulate()` calls don't have the same behaviour.

clinssen (approved):
cheers!

clinssen:
@heplesser

heplesser (changes requested):
@WillemWybo Please excuse my much delayed review. I have some comments on code style details (not top priority, can be fixed later, except for the double initialization of data members in the class declaration and the constructor; should happen in constructor only).

My main question is if it is really necessary to introduce `v_init_`. Can the user not set `v_comp_` to its initial value directly, i.e., provide `{"C_m": 1.00, "g_C": 0.00, "g_L": 0.100, "e_L": -60.0, "V_m": -60.0}`? That would avoid a new parameter name and be more consistent with the behavior of NEST models otherwise.

I also wonder if it is necessary to store `v_init_` as a member in each channel instance. As far as I can see, it is used only in `pre_run_hook()`, so if you pass the initial membrane potential from the compartment's `pre_run_hook()`, you don't need it as a member variable. That will save memory and ensure consistency. Note that since the channel classes are not derived from `Node`, you can add arguments to their `pre_run_hook()` method as you like.

Finally, I wonder why you defined `compute_statevar_m()` with two reference arguments instead of returning the two results as a pair as in `f_numstep()`.  In my opinion this makes for much more readable code. So unless it is proven that this gives a noticeable performance benefit, I would prefer returning a tuple.

models/cm_compartmentcurrents.cpp
```diff
@@ -144,19 +176,21 @@ nest::Na::f_numstep( const double v_comp )
 
 
 nest::K::K()
-  // initialization state variables
+  // state variables
```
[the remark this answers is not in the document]
WillemWybo: implemented with [d3a4f2e](https://github.com/nest/nest-simulator/pull/2773/commits/d3a4f2ed3240e1e6f4ce379c8b08ee27f78c80a0)

models/cm_compartmentcurrents.h
```diff
@@ -47,26 +47,36 @@ namespace nest
 class Na
 {
 private:
-  // state variables sodium channel
+  //!< state variables sodium channel
   double m_Na_ = 0.0;
```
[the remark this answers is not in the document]
WillemWybo: implemented with [d3a4f2e](https://github.com/nest/nest-simulator/pull/2773/commits/d3a4f2ed3240e1e6f4ce379c8b08ee27f78c80a0)

models/cm_compartmentcurrents.h
```diff
@@ -84,53 +94,61 @@ class Na
...
 class AMPA
 {
 private:
-  // global synapse index
+  //!<  global synapse index
   long syn_idx = 0;
 
-  // state variables
+  //!<  state variables
   double g_r_AMPA_ = 0., g_d_AMPA_ = 0.;
```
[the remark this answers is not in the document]
WillemWybo: implemented with [d3a4f2e](https://github.com/nest/nest-simulator/pull/2773/commits/d3a4f2ed3240e1e6f4ce379c8b08ee27f78c80a0)

models/cm_compartmentcurrents.h
```diff
@@ -47,26 +47,36 @@ namespace nest
 class Na
 {
 private:
-  // state variables sodium channel
+  //!< state variables sodium channel
```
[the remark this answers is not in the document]
WillemWybo: implemented with [d3a4f2e](https://github.com/nest/nest-simulator/pull/2773/commits/d3a4f2ed3240e1e6f4ce379c8b08ee27f78c80a0)

WillemWybo:
@heplesser I modified the `compute_statevar` functions to returns `std::pair`s.

models/cm_compartmentcurrents.cpp
```diff
@@ -50,18 +52,25 @@ nest::Na::Na( const DictionaryDatum& channel_params )
   {
     e_Na_ = getValue< double >( channel_params, "e_Na" );
   }
-  if ( channel_params->known( "V_init" ) )
+
+  // set the initialization of the channel
+  double v_init = -75.;
```
heplesser, line 57: I have an idea how to avoid this.

heplesser (changes requested):
@WillemWybo Thanks for the fixes. I have a few more small comments, a bit more on code formatting, and then a solution to avoid "magic" -75 mV in the Na/K constructors.

models/cm_compartmentcurrents.cpp
WillemWybo: fixed with commit [ca5580f](https://github.com/nest/nest-simulator/pull/2773/commits/ca5580f5f1bd7fc5ca0c8ab4c460c7f4c36d366a)

models/cm_compartmentcurrents.cpp
```diff
@@ -220,8 +242,9 @@ nest::K::append_recordables( std::map< Name, double* >* recordables, const long
   ( *recordables )[ Name( "n_K_" + std::to_string( compartment_idx ) ) ] = &n_K_;
 }
 
-void
-nest::K::compute_statevar_n( const double v_comp, double& tau_n_K, double& n_inf_K )
+std::pair< double, double >
+// nest::K::compute_statevar_n( const double v_comp, double& tau_n_K, double& n_inf_K )
```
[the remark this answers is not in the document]
WillemWybo: fixed with [6387307](https://github.com/nest/nest-simulator/pull/2773/commits/6387307fac84c4bd7a7b42a9c00945184199da3a)

models/cm_compartmentcurrents.h
```diff
@@ -171,41 +168,44 @@ class AMPA
...
 
-  //!< assigned variables
+  //! assigned variables
   double g_norm_ = 1.0;
 
-  //!< propagators
-  double prop_r_ = 0., prop_d_ = 0.;
+  //! propagators
+  double prop_r_ = 0.;
+  double prop_d_ = 0.;
```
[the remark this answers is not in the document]
WillemWybo: fixed with [6387307](https://github.com/nest/nest-simulator/pull/2773/commits/6387307fac84c4bd7a7b42a9c00945184199da3a)

models/cm_compartmentcurrents.h
```diff
@@ -319,28 +322,34 @@ class NMDA
...
   double g_norm_NMDA_ = 1.0;
 
-  //!< propagators
-  double prop_r_AMPA_ = 0., prop_d_AMPA_ = 0.;
-  double prop_r_NMDA_ = 0., prop_d_NMDA_ = 0.;
+  //! propagators
+  double prop_r_AMPA_ = 0.;
+  double prop_d_AMPA_ = 0.;
+  double prop_r_NMDA_ = 0.;
+  double prop_d_NMDA_ = 0.;
```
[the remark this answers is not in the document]
WillemWybo: fixed with [6387307](https://github.com/nest/nest-simulator/pull/2773/commits/6387307fac84c4bd7a7b42a9c00945184199da3a)

heplesser (approved):
@WillemWybo Thanks!

PR #2780, opened 2023-05-11
Fix tutorials in documentation

mlober:
We (intern Dorothee Germer and I) found a list of things to improve in the tutorials in the nest documentation:

Part 1:
- in Creating Nodes: I_e must be set to 376. instead of 375. for the neuron to elicit spikes.
- in Extracting and Plotting Data from Devices: changed the naming of variables for better readability. 

Part 2: 
- in Generating Populations of Neurons with Deterministic Connections: multimeters must be connected with the one_to_one rule, otherwise they all record the same 10 neurons, which is redundant.

Part 3:
- in Parameterizing Synapse Models: command nest.Models() is deprecated. Use command synapse_models instead.

Part 4:
- in Defining Spatially Distributed Nodes - On Grid: missing comma in code.

We suggest @jessica-mitchell as a reviewer.


jessica-mitchell (approved):
@mlober thanks for taking some time to fix up some docs. LGTM :)

PR #2786, opened 2023-05-13
Fix indentation in example in docs

clinssen:
Also fix ``np`` -> ``numpy`` and ``numpy.int64`` -> ``int``.

The examples in the docs mix code and code output, for example, in that same file:

```py
::

    print(nest.GetDefaults('static_synapse'))

        {'delay': 1.0,
         'has_delay': True,
         'num_connections': 0,
         'receptor_type': 0,
         'requires_symmetric': False,
         'sizeof': 32,
         'synapse_model': 'static_synapse',
         'weight': 1.0,
         'weight_recorder': ()}

    nest.SetDefaults('static_synapse', {'weight': 2.5})
```

What's the best way to split these up? Different blocks demarcated by ``::``?


jessica-mitchell:
@clinssen For code that has input / output , the `>>>` syntax should be used 
I would suggest given that example something like 
```
>>>    print(nest.GetDefaults('static_synapse')) 
          {'delay': 1.0,
           'has_delay': True,
           'num_connections': 0,
           'receptor_type': 0,
           'requires_symmetric': False,
           'sizeof': 32,
           'synapse_model': 'static_synapse',
           'weight': 1.0,
           'weight_recorder': ()}
::

    nest.SetDefaults('static_synapse', {'weight': 2.5})

This will separate the blocks of course, but I think it makes the output consistent with other blocks that use that syntax.

clinssen:
Guideline for the use of ">>>" in Python code examples: restrict to single-line (or very-few-line) snippets of code. Code that prepares for the line containing ">>>", for instance by instantiating neurons and connections, will go without ">>>".

jessica-mitchell (approved):
@clinssen this looks good to me, I'm gonna follow up with a PR so that the copy button des not copy the prompt symbol

jessica-mitchell:
merging this with one reviewer as its doc related

PR #2788, opened 2023-05-15
Return valid version information also when git is unavailable

jougs:
This fixes the extraction of version information also if `git` is not a runnable command. The problem was introduced in #2715.


nicolossus (approved):
Thanks! I can confirm that this fixes the build on a JSC HPC system (JUSUF) without git loaded. A local build with git available also works as expected.

heplesser (approved):

PR #2796, opened 2023-05-22
Update the package index files on all CI runners

nicolossus:
According to the [GitHub Actions documentation](https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners), one should always run `sudo apt-get update` before installing a package. This PR adds the command to all runners. This seems to fix the installation of Vale, which has been failing recently.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -80,6 +80,7 @@ jobs:
 
       - name: "Install dependencies"
         run: |
+          sudo apt-get update
           sudo apt install python3-docutils
           wget --progress=dot:mega 'https://github.com/errata-ai/vale/releases/download/v2.15.2/vale_2.15.2_Linux_64-bit.tar.gz'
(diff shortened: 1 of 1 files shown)


.github/workflows/nestbuildmatrix.yml
```diff
@@ -80,6 +80,7 @@ jobs:
 
       - name: "Install dependencies"
         run: |
+          sudo apt-get update
```
jessica-mitchell, line 83:
```suggestion
          sudo apt update
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -341,6 +342,7 @@ jobs:
...
```
jessica-mitchell, line 345:
```suggestion
          sudo apt update
```

jessica-mitchell (changes requested):
@nicolossus thanks for this change, I think we are moving towards the apt only syntax, so  I suggest to remove the 'get'

jessica-mitchell (approved):
thanks!

PR #2798, opened 2023-05-23
Update vale version and fix CI package index

jessica-mitchell:
This PR fixes the vale CI job by adding an apt update and fix to a style syntax error.
It also updates the version of vale to the most recent version.


jessica-mitchell:
I just saw #2796 solves the CI problem, I'd still like to do the fixes to vale. @terhorstd what do you think

PR #2819, opened 2023-06-02
Add `sli2py_ignore` subdirectories

nicolossus:
This PR adds a `sli2py_ignore` subdirectory to both `unittests` and `regressiontests`. The `sli2py_ignore` subdirectories are meant to hold SLI tests that will not be ported to Pytest.


jougs (approved):
LGTM. Merging without second review as this was discussed in a larger group on-site.

PR #2823, opened 2023-06-06
Update microcircuit docs regarding virtual process constraint and fix images

jessica-mitchell:
This PR adds a note to clarify that a single virtual process cannot be used to run the microcircuit model.
At the same time the images used in the microcircuit's README are fixed so they work now.


pynest/examples/Potjans_2014/README.rst
```diff
@@ -54,6 +58,15 @@ The output will be saved in the ``data`` directory.
 
 The code can be `parallelized <https://nest-simulator.readthedocs.io/en/latest/guides/parallel_computing.html>`_ using OpenMP and MPI, if NEST has been built with these features.
 The number of threads (per MPI process) can be chosen by adjusting ``local_num_threads`` in ``sim_params.py``.
+
+
+.. important::
+
+   The microcircuit model does not run correctly on a single virtual process. This means
+   the number of threads must be greater than 1. You can either increase
+   the number of MPI processes or set ``local_num_threads`` to a value > 1.
```
JanVogelsang, line 67:
```suggestion
   the number of MPI processes or set either ``local_num_threads`` to a value > 1 or replace ``local_num_threads`` by ``total_num_virtual_procs`` and set it to a value > 1.
```
Not quite sure if this addition really is required, but it would be more complete.

JanVogelsang (changes requested):

pynest/examples/Potjans_2014/README.rst
JanVogelsang: @ackurth Before we merge this, what is your opinion on this?

JanVogelsang (approved):

pynest/examples/Potjans_2014/README.rst
```diff
@@ -54,6 +58,15 @@ The output will be saved in the ``data`` directory.
 
 The code can be `parallelized <https://nest-simulator.readthedocs.io/en/latest/guides/parallel_computing.html>`_ using OpenMP and MPI, if NEST has been built with these features.
 The number of threads (per MPI process) can be chosen by adjusting ``local_num_threads`` in ``sim_params.py``.
+
+
+.. important::
+
+   The microcircuit model does not run correctly on a single virtual process. This means
+   the number of threads must be greater than 1. You can either increase
+   the number of MPI processes or set either ``local_num_threads`` to a value > 1 or replace ``local_num_threads`` by ``total_num_virtual_procs`` and set it to a value > 1.
```
ackurth, line 67: I just build with the current master and it appears that two is not enough.
Could you please check @JanVogelsang?
This is also consistent with the error
```
Too many connections: at most 134217727 connections supported per virtual process and synapse model.
```
since the microcricuit has approximately has 300 million synapses.
Four are definitely enough.

JanVogelsang: Hm, when I just execute the `run_microcircuit.py` as it is in the repo, it works just fine (which uses a single process and a single thread). However, that version uses a scale of 0.1 by default. When increasing the scaling of both N and K to 1.0, I can reproduce your observations. Three virtual processes are sufficient as well.

JanVogelsang (approved):

pynest/examples/Potjans_2014/README.rst
jessica-mitchell: Resolving this based on discussion in ODV, where we agreed to use 4 as the base

PR #2829, opened 2023-06-13
Port `ticket-80-175-179.sli` to py

otcathatsya:


nicolossus (changes requested):
@otcathatsya Thanks for decoding and porting the test. I have created [a PR](https://github.com/otcathatsya/nest-simulator/pull/1) with some suggested improvements.

nicolossus (approved):
Thanks!

testsuite/pytests/sli2py_regressions/test_ticket_80_175_179.py
```diff
@@ -0,0 +1,138 @@
...
+
+import nest
+import numpy.testing as nptest
+import pytest
+
+vm_stop = 100.0
+total_sim_time = vm_stop + 2.0
+
+
+@pytest.fixture(scope="module")
```
heplesser, line 42: If I understand the documentation here (https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) correctly, then the `setup()` method would be run only once for the entire module and we just every time look at the same data. 
```suggestion
```

Have you tried to break the test by making some changes to you are sure it does a fresh simulation for each block length?

heplesser (commented):
Thanks, looks a lot tidier now, but I am uncertain about one point, see below.

testsuite/pytests/sli2py_regressions/test_ticket_80_175_179.py
otcathatsya: Oh I was under the impression that the network setup was static for all runs and only the sim block varied, the module scope definitely only runs it once for all. Can change it later once back at my PC.

heplesser (approved):

PR #2833, opened 2023-06-19
Fix incorrect name in model docs

jessica-mitchell:
Replace`dc_generator` with   `step_current_generator`. Not sure if the rest of sentence is correct for this model or if it needs further changes.


jougs (approved):
Obviously correct. Merging without additional reviews

PR #2835, opened 2023-06-21
Replace hand-written modelsmodule by a code generation framework

jougs:
This replaces the hand-written `modelsmodule.cpp` by a mechanism to generate it dynamically from CMake arguments.

I think the documentation I wrote best explains the rationale for this change:

> By default, NEST will compile and register *all* neuron and synapse models that are shipped in the source distribution. This is very convenient for an explorative development of simulation scripts, but leads to quite long compilation times and is often not necessary.
> Now, there are two ways to restrict the set of built-in models to tailor NEST to your needs:
> * `-Dwith-modelset=<modelset>` allows you to pecify the modelset to include. Sample configurations are in the [modelsets](https://github.com/jougs/nest-simulator/tree/wip/modelsets) directory in the top-level of the source tree. This flag is mutually exclusive with -Dwith-models. [default=full].
> * `-Dwith.models=<model_list>` allows you to specify the models to include as a semicolon-separated list. This option is mutually exclusive with -Dwith-modelset [default=OFF].

This is the better solution I have promised in #2663.


jougs:
@JanVogelsang: Voilà! Only nine weeks after I said "this is really easy!" :innocent: :sweat_smile:

models/CMakeLists.txt
```diff
@@ -18,127 +18,14 @@
...
-    pp_cond_exp_mc_urbanczik.h pp_cond_exp_mc_urbanczik.cpp
-    ppd_sup_generator.h ppd_sup_generator.cpp
-    pulsepacket_generator.h pulsepacket_generator.cpp
-    quantal_stp_synapse.h quantal_stp_synapse_impl.h
-    rate_connection_delayed.h
-    rate_connection_instantaneous.h
-    rate_neuron_opn.h rate_neuron_opn_impl.h
+    cm_tree.h cm_tree.cpp
     rate_neuron_ipn.h rate_neuron_ipn_impl.h
+    rate_neuron_opn.h rate_neuron_opn_impl.h
```
clinssen, line 26: Why are these still hard-coded?

modelsets/binary
```diff
@@ -0,0 +1,9 @@
+corellospinmatrix_detector
+erfc_neuron
+ginzburg_neuron
+mcculloch_pitts_neuron
+spin_detector
+spike_generator
+spike_recorder
+static_synapse
+multimeter
```
clinssen, line 9: newline at EOF (also in other files in ``modelsets``)

models/CMakeLists.txt
jougs: These are the auxiliary helper/base classes we talked about the other day. They need compilation/inclusion (see [lines 48-50](https://github.com/nest/nest-simulator/pull/2835/files/ae884a60c9546a64b8ac8f79d1e5473bfaafe4b0#diff-d0e03f64b9b96929ef3763efbeca17202c027ecf72059b6a12a676aefee1a372R48-R50) but don't define any models themselves. I thought the easiest way was to include them here irregardless of whether they are needed or not and assume that a modern compiler is smart enough to leave their symbols/definitions out of the resulting executable if they are not needed.

modelsets/binary
jougs: Fixed in e5c3b89.

models/CMakeLists.txt
clinssen: I understand, but why can't we anyway just put them in the "rate" (and "full") modelsets?

jougs: Because then 
1. users who want to build NEST with certain custom modelset would have to know about these (purely technical dependencies), i.e. `-Dwith-models=gauss_rate,step_rate_generator,multimeter` would also have to include `rate_ipn`, `rate_opn` and `rate_transformer_node` (and in the future possible others).
2. the generation pipeline would potentially have to treat the auxiliary files differently, as they don't actually define any models

build_support/generate_modelsmodule.py
```diff
@@ -0,0 +1,340 @@
...
+
+    fname = "modelsmodule.cpp"
+    modeldir = Path(blddir) / "models"
+    modeldir.mkdir(parents=True, exist_ok=True)
+    with open(modeldir / fname, "w") as file:
+        file.write(copyright_header.replace("{{file_name}}", fname))
+        file.write(
+            dedent(
+                """
+            #include "modelsmodule.h"
```
pnbabu, line 252: Is it better to generate this file via a jinja template?

pnbabu (changes requested):

build_support/generate_modelsmodule.py
jougs: I also considered this, but since this is part of the run of `cmake` and should work everywhere, I thought it would be better to avoid external dependencies.

doc/htmldoc/installation/cmake_options.rst
```diff
@@ -38,6 +38,26 @@ Options for configuring NEST
...
+By default, NEST will compile and register *all* neuron and synapse
+models that are shipped in the source distribution. This is very
+convenient for an explorative development of simulation scripts, but
+leads to quite long compilation times and is often not necessary.
+
+There are two ways to restrict the set of built-in models to tailor
+NEST to your needs:
+
++-------------------------------------+------------------------------------------------------------------------------+
+| ``-Dwith-modelset=<modelset>``      | Specify the modelset to include. Sample configurations are in the            |
```
pnbabu, line 53: Could you add more documentation about the different modelsets?

clinssen (approved):

doc/htmldoc/installation/cmake_options.rst
jougs: In b2ba3fd82 I have added a bit more information about what a modelset actually is and in 03282f155 I have added informative headers to the sample modelsets.

I don't want to go into more detail about the different modelsets themselves in the documentation, as they might change over time and keeping the documentation consistent with the actual modelsets would be hard.

jougs:
> @jougs, hm they seem to work for me both in this PR and master: I checked the link from models-main page and the link in the examples index page and both correctly displayed the Microcircuit model docs.

Are you using a build directory separate from the source directory or are you running `cmake` right in the Git clone?

jessica-mitchell:
> Are you using a build directory separate from the source directory or are you running `cmake` right in the Git clone?

I was checking the build on Read the docs, but the local build (current master) with a separate build directory also works for me.

pnbabu (approved):
Looks good to me!

PR #2836, opened 2023-06-22
Set copybutton to only copy input lines in docs

jessica-mitchell:
This PR modify the copy button behaviour so it won't copy the '>>>' in input lines, it's also set to not copy the output. But this can be changed if the desired state is that the input and output should be copied with the button.

so if code looks like this:
```
>>> len(syn_spec)
2
```
The copybutton will paste 
` len(syn_spec)`


jougs (approved):
Awesome!

PR #2837, opened 2023-06-22
Fix `test_calcium.py` formatting to comply with Black

nicolossus:
Somehow https://github.com/nest/nest-simulator/pull/2627 was able to merge without having Black compliant formatting and the CI in master now fails. This PR fixes the formatting and thus the CI. This should be merged ASAP as the CI now also fails for branches with PRs toward master.

### DIFF
--- a/testsuite/pytests/test_calcium.py
+++ b/testsuite/pytests/test_calcium.py
@@ -21,6 +21,7 @@
 
 """
 Test models with calcium concentration.
+
 This set of tests verify the behavior of the calcium concentration in models
 that inherit from the strutural plasticity node class in the kernel.
 """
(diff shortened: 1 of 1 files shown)


terhorstd (approved):
These are all no-op changes. LGTM. The passing build for the #2627 needs to be investigated, though.

jougs (approved):
LGTM. Thanks for catching and fixing

nicolossus:
> These are all no-op changes. LGTM. The passing build for the #2627 needs to be investigated, though.

@terhorstd @jougs I took a closer look at the build for #2627 and the reason it passed was that it still used the black_IGNORED checker, which passed even though the formatting not was Black compliant. The Black checker was changed to fail for non-compliant code with #2750. 

Since #2750 was recently merged, we should ensure other current PRs pull master with the correct CI pipeline when reviewing.

PR #2839, opened 2023-06-26
Fix unit typo

aoot:
The rate parameter should be in Hz instead of pA.

The rate parameter is indicated as Hz a few lines down with the function to convert the units of rate to fit the simulation resolution (ms).


clinssen (approved):
Thanks for the fix!

PR #2841, opened 2023-06-27
Bump Ubuntu and Python versions for GitHub Actions workflows

nicolossus:
This PR changes the `ubuntu-20.04` runner label in workflows to `ubuntu-22.04`. As of the end of 2022, `ubuntu-latest` is `ubuntu-22.04`, see [github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/).

The PR also bumps the Python version from 3.9 to 3.10. This change is based on that Python 3.10 is the default in Ubuntu 22.04 ([github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#language-and-runtime)) and the Python development cycle:  [devguide.python.org/versions/](https://devguide.python.org/versions/).


.github/workflows/nestbuildmatrix.yml
```diff
@@ -8,7 +8,7 @@ on: [push, pull_request]
 
 jobs:
   clang-format:
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-latest"
```
jougs, line 11: I think I'd like the explicit version number better:
```suggestion
    runs-on: "ubuntu-22.04"
```
Same below...

jougs (changes requested):
Mostly fine by me.

@jessica-mitchell: didn't we have a file that states which minimal version of Python we require? Maybe that should also be updated here, when we don't test against 3.9 anymore?

jessica-mitchell:
@jougs we do have a `versions.json` in `doc/htmldoc/_ext/`. It states Python min = "3.8". I assume that NEST  still works with 3.8 though. We can change it, if that makes sense to have it match the version we are testing.
Also in regards to the comment (in the `nestbuildmatrix`) that we need to use Python 3.8 for Read the docs, I just tested 3.10 and everything seems to work fine, as far as I have seen. I think the only thing that's a little buggy is the HillTononiModels notebook, but that should be resolved with PR #2842. 
This is the [build output](https://nest-test.readthedocs.io/en/test-py3.10/) using Python 3.10

nicolossus:
@jessica-mitchell How is actually the `versions.json` used? Shouldn't the version number in `troubleshooting.rst` use the version listed in `versions.json`?

jessica-mitchell:
@nicolossus Yes you're right. The idea was we could use the VersionSyncRole extension when we needed to include a version in the docs. So we only need to keep the json file up-to-date. We haven't actually used it though. I think the syntax should be something like 
```
:version:`python`

```

jougs (approved):
Nice!

terhorstd (approved):
Thanks! LGTM :+1:

PR #2842, opened 2023-06-27
Fix various issues in docs that cause Sphinx warnings

jessica-mitchell:
This PR fixes several typos and broken links and should resolve many Sphinx warnings that happen at build time. 
It should resolve all warnings except the ones about the mc images (resolved in PR #2823), index_stimulation backend (discussed in another PR) and nest.spatial (modified in PR #2588).

To resolve the warnings about the HillTononiModels.ipynb, the markdown cells cannot have lines that begin with a space, so these were removed. Did not seem to affect the equations or output in anyway.

In the connection_management file there were duplicate headings which Sphinx didnt like; the markup for the headers was also not correct and needed fixing. The headings under 'Synapse specification' were changed to 'rule: All to all' to distinguish them from the previous section that also had 'All to all' as heading. 

The other files had minor typos and broken links.


jougs (approved):
LGTM!

PR #2843, opened 2023-06-27
NEST-SONATA: Enclose `pandas` import in `try` block

nicolossus:
Enclosing the `pandas` import in a `try` block allows NEST to be built even if `pandas` is not installed. `pandas` is still a requirement for NEST-SONATA and an error will be raised if the `SonataNetwork` class is instantiated without `pandas` installed.

### DIFF
--- a/pynest/nest/lib/hl_api_sonata.py
+++ b/pynest/nest/lib/hl_api_sonata.py
@@ -25,7 +25,6 @@ import os
 from pathlib import Path, PurePath
 
 import numpy as np
-import pandas as pd
 
 from .. import pynestkernel as kernel
 from ..ll_api import sli_func, sps, sr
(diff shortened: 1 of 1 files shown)


pynest/nest/lib/hl_api_sonata.py
```diff
@@ -34,6 +33,13 @@
 from .hl_api_simulation import SetKernelStatus, Simulate
 from .hl_api_types import NodeCollection
 
+try:
+    import pandas as pd
+
```
jougs, line 38:
```suggestion
```

pynest/nest/lib/hl_api_sonata.py
```diff
@@ -34,6 +33,13 @@
...
+    have_pandas = True
+except ImportError:
+    have_pandas = False
+
 try:
     import h5py
```
jougs, line 45:
```suggestion
```

jougs (approved):
Nice.

pynest/nest/lib/hl_api_sonata.py
nicolossus: Empty line after import is added by formatter.

heplesser (approved):

terhorstd:
Thanks for the quick solution, @nicolossus !

jougs:
@terhorstd: Does that also mean that you are happy with the solution? ;-)

PR #2168, opened 2021-09-19
Add STDP dopamine synapse unit testing

YounesBouhadjar:
This adds a unit test for the STDP dopamine synapse. It compares the weight evolution due to pre-, post, and dopamine spikes between NEST and a Python reference implementation.

Todo
- [ ] finish the reference implementation
- [ ] improve documentation

Test stimuli:
- [x] stimulating the pre, post neurons as well as the dopa release site using a Poisson generator  
- [x] hard-coded edge cases the pre and post spikes
- [ ] hard-coded edge cases for the dopa spikes 

Neuron models:
- [x] iaf_psc_exp
- [x] iaf_cond_exp

Most of the code here is inspired by the STDP unit testing.
fixes #820

PR #2853, opened 2023-07-06
Allow setting volume_transmitter via NodeCollection, instead of naked ID

### LINKED ISSUE #2152: Type mismatch between the volume transmitter and synaptic property
In NEST3, the `volume_transmitter` is of type NodeCollection, but in the synaptic specification it's required to provide only a node ID of the `vt`, For example when running: 

```
import nest
vt = nest.Create("volume_transmitter")
nest.CopyModel("stdp_dopamine_synapse", "stdp_dopa", {"vt": vt})
```
You get a type mismatch error. A workaround for this is to convert the NodeColletion to a list of node IDs :

```
import nest
vt = nest.Create("volume_transmitter")
nest.CopyModel("stdp_dopamine_synapse", "stdp_dopa", {"vt": vt.tolist()[0]})
```

I think it would be cleaner if the synaptic property "vt" can accept a NodeCollection.

jougs:
This PR mainly fixes #2152 by allowing to set the `volume_transmitter` on the `stdp_dopamine_synapse` using a NodeCollection instead of a naked ID.

* As this breaks all existing code involving these two models, I took the liberty to also rename the property *vt* of `stdp_dopamine_synapse` to *volume_transmitter* as suggested in [a comment](https://github.com/nest/nest-simulator/issues/2152#issuecomment-913373342) in the original issue.

* As the `weight_recorder` is meant to be handled in pretty much the same way as the `volume_transmitter`, I also streamlined the code doing that, so it is more similar. I have also added a type check to that code, so it is less prone to errors.

* On the way, I added `NodeCollection::create(Node*)`, which allows to conveniently create a NodeCollection from an existing node from a pointer.

* I also removed the apparently unused function `CommonSynapseProperties::get_node()`


JanVogelsang:
We should probably add a 4.0-tag here, unless that's what "Behavior changes" stands for. But for me that sounds more like a "no API change, just behavioral changes" and that's not the case here.

doc/htmldoc/whats_new/v3.6/index.rst
```diff
@@ -0,0 +1,40 @@
...
+the property `volume_transmitter` of the synapse's common properties:
+
++--------------------------------------------------+--------------------------------------------------+
+| Up to NEST 3.5                                   | from NEST 3.6 on                                 |
++==================================================+==================================================+
+|  ::                                              |  ::                                              |
+|                                                  |                                                  |
+|     vt = nest.Create("volume_tranmitter")        |     vt = nest.Create("volume_tranmitter")        |
+|     nest.SetDefaults(                            |     nest.SetDefaults(                            |
+|         "stdp_dopamine_synapse",                 |         "stdp_dopamine_synapse", {"vt": vt}      |
```
JanVogelsang, line 36: Should probably be "volume_transmitter" on the right.

nestkernel/CMakeLists.txt
```diff
@@ -152,6 +152,7 @@ target_include_directories( nestkernel PRIVATE
     ${PROJECT_SOURCE_DIR}/thirdparty
     ${PROJECT_SOURCE_DIR}/libnestutil
     ${PROJECT_BINARY_DIR}/libnestutil
+    ${PROJECT_SOURCE_DIR}/models
```
JanVogelsang, line 155: How is this related to the changes of this PR?

models/stdp_dopamine_synapse.cpp
```diff
@@ -79,16 +74,23 @@ STDPDopaCommonProperties::set_status( const DictionaryDatum& d, ConnectorModel&
...
     const size_t tid = kernel().vp_manager.get_thread_id();
-    Node* vt = kernel().node_manager.get_node_or_proxy( vtnode_id, tid );
-    vt_ = dynamic_cast< volume_transmitter* >( vt );
-    if ( not vt_ )
+    Node* vt_node = kernel().node_manager.get_node_or_proxy( ( *vt_datum )[ 0 ], tid );
+    volume_transmitter* vt = dynamic_cast< volume_transmitter* >( vt_node );
+    if ( not vt )
     {
-      throw BadProperty( "Dopamine source must be volume transmitter" );
+      throw BadProperty( "Property vt must be set to a node of type volume_transmitter" );
```
JanVogelsang, line 90: The property is now called "volume_transmitter" instead of "vt".

jougs:
> We should probably add a 4.0-tag here, unless that's what "Behavior changes" stands for. But for me that sounds more like a "no API change, just behavioral changes" and that's not the case here. Or do we not count renaming a parameter ("vt"->"volume_transmitter") as API change?

As we don't have too many releases and the major releases are often many years apart, we don't adhere to a *no-API-changes-in-minor-versions* policy too strictly. We usually try to deprecate things at least one release in advance, but have not been totally consistent with that if we deemed that only a small fraction of users will be affected by the change. For example, in [NEST 3.4](https://nest-simulator.readthedocs.io/en/stable/whats_new/v3.4) we have removed some deprecated exotic models and changed the default port for NEST Server. In case of problems, we can always assist people on the mailing list or here in the issue tracker.

nestkernel/CMakeLists.txt
jougs: This is required because of the [added](https://github.com/nest/nest-simulator/blob/f91ab2b41cc864bbc151e413bb26c3d0dcaca868/nestkernel/common_synapse_properties.cpp#L31-L32) `#include "weight_recorder.h"` in `common_synapse_properties.cpp`.

models/stdp_dopamine_synapse.cpp
jougs: Fixed in 0818dc83e.

doc/htmldoc/whats_new/v3.6/index.rst
jougs: Fixed in 0818dc83e.

YounesBouhadjar:
The PR looks good, we just need to correct #2168

YounesBouhadjar (approved):

JanVogelsang:
Interestingly, clang can't compile the same code that gcc compiles.

jougs:
> Interestingly, clang can't compile the same code that gcc compiles.

Yes, this is odd. But then again, the code is also not really nice and I even struggled a bit to get it to compile on gcc...

In the new code we now have `weight_recorder* CommonSynapseProperties::get_weight_recorder()`. Then we have `Event::set_receiver(Node&)`, which we want to use in [`Connector<ConnectionT>::send_weight_event()`](https://github.com/nest/nest-simulator/pull/2853/files#diff-de507fa01ea3f48057f6aba07852d1426dbbe8c4b5b7751d09f96ab9d33c5cc3R55-R56).

My naive thinking was that I should be able to upcast the `weight_recorder*` obtained from the getter  to `Node*` using a `static_cast` and dereference the result of that, all in a single line right in the call to `set_receiver()` like this:
```c++
wr_e.set_receiver( *static_cast< Node* >( cp.get_weight_recorder() ) );
```
This, however did not even work on gcc and I resorted to the `dynamic_cast`, which clang now seems unhappy about.

Interestingly, the equivalent C-cast seems to work (at least on gcc):
```c++
wr_e.set_receiver( *( Node* )( cp.get_weight_recorder() ) );
```
As does a two line construct that skips explicit casts altogether:
```c++
Node* wr_tmp = cp.get_weight_recorder();
wr_e.set_receiver( *wr_tmp );
```

I am leaning towards the last solution, as that is most explicit. What do you think, @JanVogelsang?

JanVogelsang:
I don't understand why a simple `static_cast` wouldn't work here. Do you know the error it produced? The last solution isn't too bad, it's longer than it has to be and I don't feel like we have to be explicit in this place as no one really cares if the weight recorder is casted to a node-pointer before or not. But it's definitely a solution that one could use here which still is quite clean, so in the interest of saving time I think it might be the best solution right now.

jougs:
Ha!

It definitely helps to read the *full* error message, including the notes :sweat_smile:

```text
/home/jochen/work/src/nest-simulator/models/nestkernel/connector_base_impl.h:58:25: error: invalid ‘static_cast’ from type ‘nest::weight_recorder*’ to type ‘nest::Node*’
   58 |     wr_e.set_receiver( *static_cast< Node* >( cp.get_weight_recorder() ) );
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/jochen/work/src/nest-simulator/models/nestkernel/connection.h:27:
/home/jochen/work/src/nest-simulator/models/nestkernel/common_synapse_properties.h:40:7: note: class type ‘nest::weight_recorder’ is incomplete
   40 | class weight_recorder;
      |       ^~~~~~~~~~~~~~~
```

The problem was that the file in question did not include `weight_recorder.h` and thus the cast could not be performed.
Fixed in db99d60f9.

JanVogelsang (approved):

PR #2854, opened 2023-07-07
Add sphinx_copybutton dependency for doc building

heplesser:
With this PR, all necessary packages for building documentation are included in `environment.yml`. With it in place,

```
cmake ... -Dwith-userdoc=ON
make docs
```

works on macOS.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -106,9 +106,9 @@ dependencies:
     - example
     - Image
     - sphinx_autobuild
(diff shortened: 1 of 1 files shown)


jessica-mitchell (commented):
@heplesser Why is the notfound-page extension removed?   The sphinx-copybutton extension was correct with the hyphen. And this works in the build on Linux.

PR #2856, opened 2023-07-07
Completely revised iaf_psc_alpha user documentation

heplesser:
This PR provides completely new version of the user documentation of `iaf_psc_alpha` with a focus on the equations underlying the model. I provide this as a suggestion for how to re-write model documentation. It can probably do with a bit of polishing before being used as template for other models.


models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
+
+.. math::
 
-   The present implementation uses individual variables for the
-   components of the state vector and the non-zero matrix elements of
-   the propagator. Because the propagator is a lower triangular matrix,
-   no full matrix multiplication needs to be carried out and the
-   computation can be done "in place", i.e. no temporary state vector
-   object is required.
+   I_{\text{syn, ex}}(t) = \sum_{(t_m, w_m)\\w_m\geq 0} w_m i_{\text{syn, ex}}(t-t_m)
```
clinssen, line 88:
```suggestion
   I_{\text{syn, ex}}(t) = \sum_{(t_m, w_m)\\w_m > 0} w_m i_{\text{syn, ex}}(t-t_m)
```

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
+
+   i_{\text{syn, X}}(t) = \frac{e}{\tau_{\text{syn, X}}} t e^{-\frac{t}{\tau_{\text{syn, X}}}} \Theta(t)
+
+where :math:`\Theta(x)` is the Heaviside step function. The PSCs are normalized to unit maximum, i.e.,
+
+.. math::
+
+   i_{\text{syn, X}}(t) = 1 \Leftrightarrow t = \tau_{\text{syn, X}} \;.
+
+As a consequence, the total charge transferred by a single PSC depends
```
clinssen, line 103:
```suggestion
As a consequence, the total charge :math:`q` transferred by a single PSC depends
```

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
+on the synaptic time constant according to
+
+.. math::
+
+   \int_0^{infty}  i_{\text{syn, X}}(t) dt = e \tau_{\text{syn, X}} \;.
```
clinssen, line 108:
```suggestion
   q = \int_0^\infty  i_{\text{syn, X}}(t) dt = e \tau_{\text{syn, X}} \;.
```

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
+
+.. math::
+
+   i_{\text{syn, X}}(t) = \frac{e}{\tau_{\text{syn, X}}} t e^{-\frac{t}{\tau_{\text{syn, X}}}} \Theta(t)
+
+where :math:`\Theta(x)` is the Heaviside step function. The PSCs are normalized to unit maximum, i.e.,
+
+.. math::
+
+   i_{\text{syn, X}}(t) = 1 \Leftrightarrow t = \tau_{\text{syn, X}} \;.
```
clinssen, line 101: I don't understand the notation here

models/iaf_psc_alpha.h
```diff
@@ -115,8 +142,8 @@ The following parameters can be set in the status dictionary.
  tau_syn_ex ms      Rise time of the excitatory synaptic alpha function
  tau_syn_in ms      Rise time of the inhibitory synaptic alpha function
  I_e        pA      Constant input current
- V_min      mV      Absolute lower value for the membrane potenial
-=========== ======  ==========================================================
+ V_min      mV      Absolute lower value for the membrane potenial (default `math:-\infty`)
```
clinssen, line 145:
```suggestion
 V_min      mV      Absolute lower value for the membrane potenial (default :math:`-\infty`)
```

clinssen (changes requested):

environment.yml
```diff
@@ -106,6 +106,7 @@ dependencies:
     - example
     - Image
     - sphinx_autobuild
+    - sphinx_copybutton
```
jessica-mitchell, line 109: The installation name for this extension is `sphinx-copybutton` and already exists in the list.
See [here](https://sphinx-copybutton.readthedocs.io/en/latest/index.html)

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
 
-See also [3]_.
 
 Parameters
 ++++++++++
 
 The following parameters can be set in the status dictionary.
 
-=========== ======  ==========================================================
+=========== ======  =======================================================================
```
jessica-mitchell, line 134: Should we add a title to this table? Are these membrane parameters?

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
+
+   V_m(t_k) < V_{th} \quad\text{and}\quad V_m(t_{k+1})\geq V_{th} \;.
+
+Subsequently,
+
+.. math::
+
+   V_m(t) = V_{\text{reset}} \quad\text{for}\quad t^* \leq t < t^* + t_{\text{ref}} \;,
+
+i.e., the membrane potential is clamped to :math:`V_{\text{reset}}` during the refractory period.
```
jessica-mitchell, line 76:
```suggestion
that is., the membrane potential is clamped to :math:`V_{\text{reset}}` during the refractory period.
```

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
-   address the problem of efficient usage of appropriate vector and
-   matrix objects.
+is the sum over all incoming excitatory spikes, with a corresponding expression for inhibitory spikes (:math:`w_m < 0`).
+Finally, individual post-synaptic currents are given by
+
+.. math::
+
+   i_{\text{syn, X}}(t) = \frac{e}{\tau_{\text{syn, X}}} t e^{-\frac{t}{\tau_{\text{syn, X}}}} \Theta(t)
+
+where :math:`\Theta(x)` is the Heaviside step function. The PSCs are normalized to unit maximum, i.e.,
```
jessica-mitchell, line 97:
```suggestion
where :math:`\Theta(x)` is the Heaviside step function. The PSCs are normalized to unit maximum, that is.,
```

jessica-mitchell (changes requested):

models/iaf_psc_alpha.h
jougs: Isn't there a '.' too much now? Also below...

jessica-mitchell: fixed

models/iaf_psc_alpha.h
janskaar: Since the function has a single global maximum of 1 when t=tau_syn, left-hand side implies right-hand side and vice versa.

models/iaf_psc_alpha.h
janskaar: There are membrane parameters, various other neuron parameters and also synaptic parameters. I think just "Parameters" is the most fitting title.

models/iaf_psc_alpha.h
janskaar, line 88: I don't like m as a summation index here, since we're already using it as a subscript for membrane parameters. I also think the summation notation is a bit confusing.

```suggestion
.. math::

   I_{\text{syn, ex}}(t) = \sum_{j} w_j \sum_k i_{\text{syn, ex}}(t-t_j^k-d_j) \;,

where :math:`j` indexes excitatory presynaptic neurons, :math:`k` indexes the spike times of neuron :math:`j`, and :math:`d_j` is the delay from neuron :math:`j`.

```

```suggestion
.. math::

   I_{\text{syn, ex}}(t) = \sum_{j \in \Gamma} w_j \sum_{t' \in \Pi_j} i_{\text{syn, ex}}(t-t'-d_j) \;,

where :math:`\Gamma` is the set of indices of excitatory presynaptic neurons, :math:`\Pi_j` is the set of spike times from neuron :math:`j`, and :math:`d_j` is the delay from neuron :math;`j`.

```

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
-with alpha-function shaped synaptic currents. Thus, synaptic currents
-and the resulting postsynaptic potentials have a finite rise time.
+``iaf_psc_alpha`` is a leaky integrate-and-fire neuron model with
 
-The threshold crossing is followed by an absolute refractory period
-during which the membrane potential is clamped to the resting potential.
+* a hard threshold,
+* a fixed refractory period,
+* no adaptation mechanisms,
+* :math:`\alpha`-shaped synaptic input currents.
```
jougs, line 53: Please consider to make the descriptions in the list more expressive. Right now they are very short and might be intimidating to new users.

models/iaf_psc_alpha.h
```diff
@@ -40,71 +40,98 @@ namespace nest
...
+The membrane potential evolves according to
 
-An additional state variable and the corresponding differential
-equation represents a piecewise constant external current.
+.. math::
 
-The general framework for the consistent formulation of systems with
-neuron like dynamics interacting by point events is described in
-[1]_.  A flow chart can be found in [2]_.
+   \frac{dV_m}{dt} = -\frac{V_m - E_L}{\tau_m} + \frac{I_{\text{syn}} + I_e}{C_m}
```
jougs, line 59:
```suggestion
   \frac{dV_\text{m}}{dt} = -\frac{V_\text{m} - E_\text{L}}{\tau_m} + \frac{I_{\text{syn}} + I_\text{e}}{C_\text{m}}
```

jougs: Please check throughout the whole document.

environment.yml
heplesser: Sorry I overlooked that. I will remove the duplicate and order the sphinx dependencies alphabetically.

models/iaf_psc_alpha.h
heplesser: @jougs I was aiming for a concise overview here, assuming that our users are computational neuroscientists. Especially for users looking for if a model fits their needs, such a brief bullet list with quite standardized terms seems useful to me.

models/iaf_psc_alpha.h
heplesser: I changed this line, I hope you find the new version more readable.

models/iaf_psc_alpha.h
heplesser: I included you first suggestion.

heplesser:
@jessica-mitchell @jougs @clinssen @janskaar Thank you for your comments, I have followed up almost all of them. Please have a look.

I am struggling with one problem now: The parameter table in the comment is 129 characters wide, so clang-format complains. Is there any way to either suppress clang-format for this comment or break lines in a RST table spec?

https://github.com/heplesser/nest-simulator/blob/1e203c9e5c9ac914ea59f640b1bc185217c35c5d/models/iaf_psc_alpha.h#L137-L151

janskaar (approved):

models/iaf_psc_alpha.h
jougs: In my view, our users are either computational neuroscientists already *or* people who want to become one of those.

I like the brevity for the former and agree that standardized terms are a *good thing*<sup>TM</sup>, but I'm also afraid that for the latter, terms like *hard threshold* (compared to what?), *α-shaped synaptic input currents* (how does this shape look like? what other shapes could there be?) and *no adaptation mechanisms* (what is not adapted? and what could be in other models?) might raise a lot of questions and could be too intimidating.

I suggest to add at least some links to [the glossary](https://nest-simulator.readthedocs.io/en/v3.5/ref_material/glossary.html) or suitable Scholarpedia/Wikipedia articles here.

nicolossus:
> I am struggling with one problem now: The parameter table in the comment is 129 characters wide, so clang-format complains. Is there any way to either suppress clang-format for this comment or break lines in a RST table spec?

@heplesser I think you can disable and re-enable clang-format like this:

```
// clang-format off
/*
<docs>
*/
// clang-format on
```

This will, however, suppress clang-format for the all the documentation. I don't think it is possible to suppress the table only.

models/iaf_psc_alpha.h
jessica-mitchell: I also agree with @jougs that we need to provide some more context for users less familiar with these terms.
Can we expand on this bullet list in a subsequent paragraph?  
We could add a link to the glossary, but do users need a standard definition of these terms or do they need to know how these terms are used within this model? In the latter case, I think the glossary is insufficient.

heplesser:
> > I am struggling with one problem now: The parameter table in the comment is 129 characters wide, so clang-format complains. Is there any way to either suppress clang-format for this comment or break lines in a RST table spec?
> 
> @heplesser I think you can disable and re-enable clang-format like this:
> 
> ```
> // clang-format off
> /*
> <docs>
> */
> // clang-format on
> ```
> 
> This will, however, suppress clang-format for the all the documentation. I don't think it is possible to suppress the table only.

Thanks, this worked!

models/iaf_psc_alpha.h
```diff
@@ -150,6 +183,7 @@ See also
 iaf_psc_delta, iaf_psc_exp, iaf_cond_exp
 
 EndUserDocs */
+// clang-format off
```
nicolossus, line 186:
```suggestion
EndUserDocs */
// clang-format on
```
Clang-format should probably be re-enabled again ;)

heplesser: Indeed, done :).

models/iaf_psc_alpha.h
```diff
@@ -34,89 +34,122 @@
...
+ ``C_m``         pF       :math:`C_{\text{m}}`           Capacity of the membrane
+ ``tau_m``       ms       :math:`\tau_{\text{m}}`        Membrane time constant
+ ``t_ref``       ms       :math:`t_{\text{ref}}`         Duration of refractory period
+ ``V_th``        mV       :math:`V_{\text{th}}`          Spike threshold
+ ``V_reset``     mV       :math:`V_{\text{reset}}`       Reset potential of the membrane
+ ``tau_syn_ex``  ms       :math:`\tau_{\text{syn, ex}}`  Rise time of the excitatory synaptic alpha function
+ ``tau_syn_in``  ms       :math:`\tau_{\text{syn, in}}`  Rise time of the inhibitory synaptic alpha function
+ ``I_e``         pA       :math:`I_\text{e}`             Constant input current
+ ``V_min``       mV       :math:`V_{\text{min}}`         Absolute lower value for the membrane potenial (default :math:`-\infty`)
+=============== ======== =============================== ========================================================================
```
nicolossus, line 152: While I think the table is very neat, a tabular format might actually become quite restrictive for describing parameters due to the limited horizontal space. I suggest to instead use a list format akin to the parameter section in Python documentation. We already use that here https://nest-simulator.readthedocs.io/en/latest/models/poisson_generator.html#parameters. I also think that we should provide the default value of the parameters in the documentation.

jougs:
While I generally agree with many things said here to improve the model documentation, I also feel we should not overthink (and overdo) this by putting too much effort into adding new things.

In my opinion, this PR should rather serve as a playground and result in a solid template that can be ported to the NESTML documentation generation pipeline. This is where the magic happens now and improvements there would immediately benefit *all* models (c.f. model equations, units and defaults for parameters, model characterization figures in this [example](https://nestml.readthedocs.io/en/latest/models_library/iaf_psc_alpha.html)).

In the last three years we only gained 3 new models in NEST (jonke_synapse, cm_default, spike_injector) and I would rather put some increased effort into a better integration of NESTML into the build process of NEST and let all models be generated instead of fixing equations and parameters in manually written model documentation.

@clinssen, @poojanbabu, @jessica-mitchell.

Just my 2 cents.

models/iaf_psc_alpha.h
jougs: @jessica-mitchell: do you think a kind of "reverse glossary" would be possible as a Sphinx extension? Something along the lines of what you did in #2859, but for listing all the documentation pages where a term is used next to the entry in the glossary (in some kind of hidden sliding block thingy). That could possibly solve the problem you outline in an elegant and semi-automatic way.

jessica-mitchell: @jougs yes, that should be possible- maybe for now we can leave the documentation as @heplesser suggests and I can look into developing something like this

models/iaf_psc_alpha.h
heplesser: Also in view of @jougs comment below, I would not want to change this now.

heplesser:
@jougs I fully agree with you. I now have one question concerning your earlier comments vs the playground perspective. I entirely agree that it would be useful to link terms in the documentation to glossary entries. But I fear it would be at least another two weeks until I will get around to creating the glossary entries. Would you be ok with merging this now and creating a follow-up issue for glossaryfication?

models/iaf_psc_alpha.h
jougs: @jessica-mitchell: can you please create an issue so we don't forget about this?

jougs (approved):

models/iaf_psc_alpha.h
```diff
@@ -34,89 +34,122 @@
...
 
 .. note::
 
-   If ``tau_m`` is very close to ``tau_syn_ex`` or ``tau_syn_in``, the model
-   will numerically behave as if ``tau_m`` is equal to ``tau_syn_ex`` or
-   ``tau_syn_in``, respectively, to avoid numerical instabilities.
+   NEST uses exact integration [1]_, [2]_ to integrate subthreshold membrane
+   dynamics with maximum precision; see also [3]_.
+
+   If :math:`\tau_\text{m}\approx \tau_{\text{syn, ex}}` or
```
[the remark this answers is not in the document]
heplesser: While it is a mathematical detail for corner cases, for the sake of completeness of the model description it is required here in my opinion.

heplesser:
@clinssen @jessica-mitchell Could you re-review?

jessica-mitchell (approved):
thanks lgtm!

clinssen:
You define I_syn,ex but not I_syn,in. Perhaps you could use the notation I_syn,x or just mention that I_syn,in is defined similarly to I_syn,ex. I am otherwise happy with the changes. Much obliged!

heplesser:
> You define I_syn,ex but not I_syn,in. Perhaps you could use the notation I_syn,x or just mention that I_syn,in is defined similarly to I_syn,ex. I am otherwise happy with the changes. Much obliged!

Thanks for catching this, @clinssen! It should now be fixed.

clinssen (approved):

PR #2857, opened 2023-07-12
Improve error checking in MPI IO backends

marcelkrueger:
This PR improves error handling in two cases:

1. If the file that is supposed to be read via the path given in the device label does not exist, it will throw an appropriate exception.
2. Check the MPI connect call's return code to ensure the connection was successful. 
Previously, it failed silently if, e.g., a wrong MPI address was supplied.

### DIFF
--- a/nestkernel/exceptions.cpp
+++ b/nestkernel/exceptions.cpp
@@ -410,7 +410,8 @@ std::string
 nest::MPIPortsFileMissing::message() const
 {
   std::ostringstream msg;
-  msg << "The node with ID " << node_id_ << " has no file that contains the MPI address.";
+  msg << "The node with ID " << node_id_ << " expects a file with the MPI address at location " << path_
+      << ". The file does not seem to exist.";
   return msg.str();
 }
 
--- a/nestkernel/exceptions.cpp
+++ b/nestkernel/exceptions.cpp
@@ -31,6 +31,12 @@
 // Includes from sli:
 #include "interpret.h"
 
+// Include MPI for MPI error string
+#ifdef HAVE_MPI
+#include <mpi.h>
+#endif
+
+
 std::string
 nest::UnknownModelName::message() const
 {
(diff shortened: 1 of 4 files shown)


nestkernel/exceptions.cpp
```diff
@@ -31,6 +31,12 @@
 // Includes from sli:
 #include "interpret.h"
 
+// Include MPI for MPI error string
+#ifdef HAVE_MPI
+#include <mpi.h>
+#endif
+
+
```
jougs, line 39:
```suggestion
```

nestkernel/recording_backend_mpi.cpp
```diff
@@ -383,8 +385,12 @@ nest::RecordingBackendMPI::get_port( const size_t index_node, const std::string&
   }
 
   basename << "/" << index_node << ".txt";
-  std::cout << basename.rdbuf() << std::endl;
+  std::cout << basename.str() << std::endl;
   std::ifstream file( basename.str() );
+  if ( !file.good() )
+  {
+    throw MPIPortsFileMissing( index_node );
```
jougs, line 392: Wouldn't it be nicer to pass both the expected file name and the node_id here and then print a message like this?
```
Node <node_id> expects a file with the MPI address at location <file_name>. This does not seem to exist.
```

nestkernel/recording_backend_mpi.cpp
```diff
@@ -383,8 +385,12 @@ nest::RecordingBackendMPI::get_port( const size_t index_node, const std::string&
   }
 
   basename << "/" << index_node << ".txt";
-  std::cout << basename.rdbuf() << std::endl;
+  std::cout << basename.str() << std::endl;
```
jougs, line 388: Looks like forgotten debug output to me.
```suggestion
```

nestkernel/stimulation_backend_mpi.cpp
```diff
@@ -311,8 +313,12 @@ nest::StimulationBackendMPI::get_port( const size_t index_node, const std::strin
   }
   // add the id of the device to the path
   basename << "/" << index_node << ".txt";
-  std::cout << basename.rdbuf() << std::endl;
+  std::cout << basename.str() << std::endl;
```
jougs, line 316: See above.
```suggestion
```

jougs (changes requested):

jougs (approved):
Very nice. Many thanks for your contribution!

PR #2859, opened 2023-07-12
Add custom Sphinx extension that lists examples based on model name

jessica-mitchell:
This PR adds a custom Sphinx extension that displays a list of examples based on model name.

**How to use it**
Add directive to reStructured Text with required argument - the model_name:

`.. listexamples:: model_name`

This was used in `models/iaf_psc_alpha.h` and `aeif_cond_alpha`
See output [here](https://nest-simulator--2859.org.readthedocs.build/en/2859/models/iaf_psc_alpha.html#examples-using-this-model)

**Updated**

- Script now includes function that finds matches of model names in examples
- The titles of examples are displayed rather than filename
- Using the ref role syntax, which should allow the use of this directive anywhere in the documentation without issue (hopefully) 

**TODO**

- [x] Rename some example titles that are unclear
- [x] Fix sorting so it's used on the titles rather than filenames


doc/htmldoc/_ext/list_examples.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def run(self):
+        my_arg = self.arguments[0]
+
+        # bullet list equivalent to HTML <ul>
+        bullet_list = nodes.bullet_list()
+        examples_list = ModelMatchExamples()
+
+        sorted_examples = {key: sorted(values) for key, values in examples_list.items()}
+        for key, values in sorted_examples.items():
```
pnbabu, line 77: Do we need to loop through the `sorted_examples`? Since it's a dictionary, we can directly get the value as `values = sorted_examples[key]`.

models/iaf_psc_alpha.h
```diff
@@ -149,6 +149,12 @@ See also
 
 iaf_psc_delta, iaf_psc_exp, iaf_cond_exp
 
+
+Examples using this model
++++++++++++++++++++++++++
+
+.. listexamples:: iaf_psc_alpha
```
pnbabu, line 156: Should we also add this for all the other models?

pnbabu (changes requested):
I think sorting the list of files by title text rather than the filenames is more useful. I see that it's listed as a TODO in this PR. Is that planned to be addressed here?

doc/htmldoc/_ext/list_examples.py
jessica-mitchell: You're right, Thanks for pointing that out!

models/iaf_psc_alpha.h
jessica-mitchell: In principle yes, the main goal of this PR is to create the extension. I have only added it to 2 models, to see if the logic works and the correct output is displayed, but I can add it to all.

doc/htmldoc/_ext/list_examples.py
```diff
@@ -0,0 +1,136 @@
...
+                # equivalent to HTML <li>
+                list_item = nodes.list_item()
+                newnode = nodes.reference("", "")
+                newnode["internal"] = True
+                newnode["refdocname"] = filename
+                newnode["refuri"] = app.builder.get_relative_uri(docname, filename)
+                newnode.append(nodes.Text(title))
+                para = nodes.paragraph()
+                para += newnode
+                # para.append(link_node)
```
jessica-mitchell, line 69:
```suggestion
```

pnbabu (approved):
Looks good to me! This list is super helpful for someone browsing through the models. I think we could add the directive to all model files once you have approval from other reviewers as well.

models/iaf_psc_alpha.h
terhorstd: Thanks, Jessica!

pynest/examples/spatial/conncon_targets.py
```diff
@@ -20,8 +20,8 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Convergent projection and rectangular mask, from source perspective
---------------------------------------------------------------------
+Spatial networks: Convergent projection and rectangular mask, from source perspective
+--------------------------------------------------------------------------------------
```
terhorstd, line 24:
```suggestion
Spatial networks: Convergent projection and rectangular mask, from source perspective
-------------------------------------------------------------------------------------
```

pynest/examples/spatial/connex_ew.py
```diff
@@ -20,8 +20,8 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Circular mask and flat probability, with edge wrap
----------------------------------------------------
+Spatial networks: Circular mask and flat probability, with edge wrap
+---------------------------------------------------------------------
```
terhorstd, line 24:
```suggestion
Spatial networks: Circular mask and flat probability, with edge wrap
--------------------------------------------------------------------
```

terhorstd (approved):
Two bytes to cut, then I think it's ready to go.
Thanks for the cool upgrade! :+1:

PR #2862, opened 2023-07-24
Changed total_num_virtual_procs to local_num_threads in hpc benchmark

JanVogelsang:
The HPC benchmark was setting the number of virtual processes directly using the `total_num_virtual_procs` kernel parameter instead of only setting the number of threads per process using `local_num_threads`.  
The number of virtual processes should never be set by the user and must always be implicitly given by the number of spawned MPI processes and the number of threads, which should be set explicitly per process (even though this setting must also be fixed across all processes). As a follow-up PR, setting the number of virtual processes could be disabled entirely, as this should never be done in practice anyway.


diesmann:
The change is fine and consistent with the user-level documentation of NEST.

diesmann (commented):
fine

JanVogelsang:
@diesmann Did you mean to close this PR?

JanVogelsang:
@ackurth Can you review this?

ackurth (approved):
LGTM

PR #2866, opened 2023-08-07
Change tutorials and examples organisation

clinssen:
Based on mailing list feedback about not being able to find a tutorial about weight normalisation.

It is not clear what the difference between a tutorial is and the items on the "how NEST works page". Therefore, I propose to make the split as follows:
- **Tutorials** contains the video tutorial, "first steps" PyNEST tutorial and all the items from "how NEST works"
- **Examples** contains all the use case examples from "PyNEST example scripts"

This also removes one step of indirection ("Get started" page removed--one less click). 

Further improvements:
- MUSIC tutorial was merged into one page
- "PyNEST API" renamed to "API reference" in the menu

Please let me know if you think this is sensible!


jessica-mitchell (commented):
Hi @clinssen thanks for the PR. Overall I like the suggested changes. 
I would still like to see a link to the PyNEST API on the index page since people who are familiar with NEST might want do directly dive into that. 
I also am wondering if it is obvious to all users that all our various texts would be under tutorials. We have 'guides'  on topics like HPC and built-in timers which don't fit the definition of a tutorial . At the same time I like the idea of keeping all the topics and tutorials together so it's easier to have an overview.

clinssen:
Suggested new organisation:

- Installation
- Getting started
	- PyNEST (4 pages)
	- Video tutorial
- Guides
	- All about neurons..
	- All about...
- Examples
	- Pong 
	- ...

jessica-mitchell:
@clinssen I would opt to close this in lieu of #2969.

clinssen:
Agreed!

Issue #2795, opened 2023-05-22
`nest.Connect([0], [0])` aborts the process


Helveg:
**Describe the bug**
Crash 💥 

**To Reproduce**

```pycon
>>> import nest
>>> nest.Connect([0],[0])
python: /home/robin/git/nest-simulator/nestkernel/modelrange_manager.cpp:90: nest::index nest::ModelRangeManager::get_model_id(nest::index) const: Assertion `right >= 1' failed.
[LAPTOP-JF6T3PSU:02513] *** Process received signal ***
[LAPTOP-JF6T3PSU:02513] Signal: Aborted (6)
[LAPTOP-JF6T3PSU:02513] Signal code:  (-6)
[LAPTOP-JF6T3PSU:02513] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x143c0)[0x7fae625513c0]
[LAPTOP-JF6T3PSU:02513] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7fae6223403b]
[LAPTOP-JF6T3PSU:02513] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7fae62213859]
[LAPTOP-JF6T3PSU:02513] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x22729)[0x7fae62213729]
[LAPTOP-JF6T3PSU:02513] [ 4] /lib/x86_64-linux-gnu/libc.so.6(+0x34006)[0x7fae62225006]
[LAPTOP-JF6T3PSU:02513] [ 5] /home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/nest/../../../nest/libnest.so.3(+0x3fea56)[0x7fae381d6a56]
[LAPTOP-JF6T3PSU:02513] [ 6] /home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/nest/../../../nest/libnest.so.3(_ZN4nest14NodeCollection7create_ERKSt6vectorImSaImEE+0x64)[0x7fae381a3ac4]
[LAPTOP-JF6T3PSU:02513] [ 7] /home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/nest/../../../nest/libnest.so.3(_ZN4nest14NodeCollection6createERK10TokenArray+0xe8)[0x7fae381a3ff8]
[...]

Issue #2636, opened 2023-03-14
`nest.NodeCollection([0])` aborts the process


Helveg:
**Describe the bug**

```pycon
>>> import nest
>>> nest.NodeCollection([0])
python: /home/robin/git/nest-simulator/nestkernel/modelrange_manager.cpp:90: nest::index nest::ModelRangeManager::get_model_id(nest::index) const: Assertion `right >= 1' failed.
[TNG2019:01879] *** Process received signal ***
[TNG2019:01879] Signal: Aborted (6)
[TNG2019:01879] Signal code:  (-6)
[TNG2019:01879] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x46210)[0x7fe0664a6210]
[TNG2019:01879] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7fe0664a618b]
[TNG2019:01879] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7fe066485859]
[TNG2019:01879] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x25729)[0x7fe066485729]
[TNG2019:01879] [ 4] /lib/x86_64-linux-gnu/libc.so.6(+0x36f36)[0x7fe066496f36]
[TNG2019:01879] [ 5] /home/robin/.pyenv/versions/bsb/lib/python3.9/site-packages/nest/../../../nest/libnestkernel.so.3(+0x14e67e)[0x7fe03cab867e]
[TNG2019:01879] [ 6] /home/robin/.pyenv/versions/bsb/lib/python3.9/site-packages/nest/../../../nest/libnestkernel.so.3(_ZN4nest14NodeCollection7create_ERKSt6vectorImSaImEE+0x5d)[0x7fe03ca80ddd]
[TNG2019:01879] [ 7] /home/robin/.pyenv/versions/bsb/lib/python3.9/site-packages/nest/../../../nest/libnestkernel.so.3(_ZN4nest14NodeCollection6createERK10TokenArray+0x110)[0x7fe03ca813d0]
[TNG2019:01879] [ 8] /home/robin/.pyenv/versions/bsb/lib/python3.9/site-packages/nest/../../../nest/libnestkernel.so.3(_ZNK4nest10NestModule27Cvnodecollection_iaFunction7executeEP14SLIInterpreter+0x55)[0x7fe03ca95fd5]
[...]

PR #2872, opened 2023-08-08
Fail gracefully when `node_id` 0 is used to access non-existent node

nicolossus:
Fixes [#2636](https://github.com/nest/nest-simulator/issues/2636), fixes https://github.com/nest/nest-simulator/issues/2795. 

This issue occurred when no nodes had been created and ``node_id=0`` was provided as an array-like object to ``Connect`` or to access a ``NodeCollection``. That is, the following cases would result in a crash instead of an error message:

```python
import nest
nest.Connect([0], [0])    # Crash case 1
nest.NodeCollection([0])  # Crash case 2
```

Providing a `node_id=0` is not really valid and will fail if a `NodeCollection` has been created. However, if no nodes have been created before trying to connect or access, `node_id=0` passes through the consistency check of `ModelRangeManager::is_in_range` because `first_node_id_` and `last_node_id_` both have 0 as default values. Instead of introducing two additional checks ( `first_node_id_ > 0` and `last_node_id_ > 0`) to account for this edge case, we can just ensure that `node_id > 0` with the same effect.

An alternative solution to this issue could be to ensure that `0` is not contained in array-like objects passed to `Connect` or `NodeCollection` on the PyNEST level.

### DIFF
--- a/nestkernel/modelrange_manager.h
+++ b/nestkernel/modelrange_manager.h
@@ -99,7 +99,7 @@ nest::ModelRangeManager::get_status( DictionaryDatum& )
 inline bool
 nest::ModelRangeManager::is_in_range( size_t node_id ) const
 {
-  return ( node_id <= last_node_id_ and node_id >= first_node_id_ );
+  return ( node_id > 0 and node_id <= last_node_id_ and node_id >= first_node_id_ );
 }
 
 inline std::vector< modelrange >::const_iterator
--- a/testsuite/pytests/sli2py_regressions/test_issue_2636_2795.py
+++ b/testsuite/pytests/sli2py_regressions/test_issue_2636_2795.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# test_issue_2795.py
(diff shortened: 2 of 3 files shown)


Helveg (approved):
Thanks! Looks good. I think this solution in `is_in_range` is probably better fit than at `PyNEST`. Future C++ API users thank you already :)

nicolossus:
@heplesser The `ModelRangeManager` class creates an array of `modelrange` class instances. The `is_in_range` check is there to safeguard the `get_model_id` method. In the current implementation, both `ModelRangeManager` and `modelrange` needs the `is_in_range` check and in theory they should be identical. However, since `ModelRangeManager::is_in_range` is run first with the check for `node_id>0`, a `node_id=0` cannot be passed to `modelrange::is_in_range`. This means that we strictly don't have to add another condition to check in `modelrange::is_in_range`. I would perhaps sacrifice the small performance impact of an additional condition for consistency. What do you prefer?

heplesser (approved):

PR #2873, opened 2023-08-08
Test iaf ps psp accuracy sli2py

janskaar:


testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -0,0 +1,127 @@
...
+    response = 0.
+    for sp in spiketrain:
+        t = T - delay - sp
+        response += alpha_fn(t)
+    return response
+
+
+@pytest.fixture(scope="module")
+def reference_potential():
+    return spiketrain_response(spike_emission)
```
heplesser, line 103: Wouldn't it make more sense to have the code of `spiketrain_response` directly here instead of having that function which is only called in a single place?

heplesser (changes requested):
@janskaar Good job in deciphering pretty confusing SLI code. I added a few suggestions for even sleeker Python code.

testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -0,0 +1,127 @@
...
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+test_iaf_ps_psp_accuracy.sli checks the voltage response of the precise versions of the iaf_psc_alpha
```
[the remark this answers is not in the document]
janskaar: Fixed

testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -0,0 +1,127 @@
...
+                 "tau_m": tau_m,
+                 "tau_syn_ex": tau_syn,
+                 "tau_syn_in": tau_syn,
+                 "C_m": C_m}
+
+spike_emission = [2.]
+
+
+def alpha_fn(t):
+    prefactor = analytical_u = weight * math.e / (tau_syn * C_m)
```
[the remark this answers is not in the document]
janskaar: Removed

testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -0,0 +1,127 @@
...
+        response += alpha_fn(t)
+    return response
+
+
+@pytest.fixture(scope="module")
+def reference_potential():
+    return spiketrain_response(spike_emission)
+
+
+@pytest.mark.parametrize("h", [(i) for i in range(-12, 1, 2)])
```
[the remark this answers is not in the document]
janskaar: Yes, that's better

testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -0,0 +1,127 @@
...
+    print(h)
+    nest.set(tics_per_ms=2**14, resolution=2**h)
+
+    sg = nest.Create("spike_generator")
+    sg.set(precise_times=False,
+           origin=0.,
+           spike_times=spike_emission,
+           start=0.,
+           stop=5.
+           )
```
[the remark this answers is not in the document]
janskaar: That's a good point, it was like that in the original test and I did not question the wisdom of the NEST elders. I updated the test slightly now, such that an actual off-grid spike is used. I additionally simulate the non-precise version and confirm that they give different results. I'd like your feedback on this.

testsuite/pytests/test_iaf_ps_psp_accuracy.py
janskaar: Yes, fixed.

testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -61,67 +66,59 @@
...
+
     u = neuron.get("V_m")
-    nptest.assert_allclose(reference_potential, u)
+    u_ps = neuron_ps.get("V_m")
+
+    t = T - delay - spike_emission[0]
+    reference_potential = alpha_fn(t)
+
+    assert reference_potential == pytest.approx(u_ps)
+    assert reference_potential != pytest.approx(u)
```
heplesser, line 124: Github doesn't let me put in explicit suggestions here, so they come piecemeal:
- For explicitness, I would call `neuron` rather `neuron_grid`, and the other `neuron_precise`
- I would drop the `u` and `u_ps` variables and directly write `... pytest.approx(neuron_precise.V_m)`

testsuite/pytests/test_iaf_ps_psp_accuracy.py
```diff
@@ -20,14 +20,19 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-test_iaf_ps_psp_accuracy.sli checks the voltage response of the precise versions of the iaf_psc_alpha
-model neurons to a single incoming spike. The voltage excursion is
-called postsynaptic potential (PSP). In the model neurons
-the postsynaptic current is described by an alpha-function
+Compares the voltage response of iaf_psc_alpha_ps from a single incoming
+excitatory spike at an off-grid time to the analytical solution. 
+Also checks that the equivalent non-ps iaf_psc_alpha does not give the same result.
```
heplesser, line 25: Start with a single line which could be included in a table of tests, e.g.,
```suggestion
Test that neuron with precise timing has correct post-synaptic potential for off-grid input.

Compares the voltage response of iaf_psc_alpha_ps from a single incoming
excitatory spike at an off-grid time to the analytical solution. 
Also checks that the equivalent non-ps iaf_psc_alpha does not give the same result.
```

heplesser (changes requested):
Much better :)! I added two more suggestions. Adding the extra check with grid-based neuron I think rather confuses the test—it is about precise timing. I suggest to take that out again. If you want to tighten the test, I would make more sense to check V_m at several points in time, e.g., the last time step when the effect of the spike is not felt, the first where it should be felt, around the maximum and somewhere in the tail.

testsuite/pytests/test_iaf_ps_psp_accuracy.py
janskaar: Done

testsuite/pytests/test_iaf_ps_psp_accuracy.py
janskaar: Removed the non-ps stuff, added a second time point to check intead

heplesser (approved):

PR #2883, opened 2023-08-09
Sort and format Python imports and enable `isort` check in CI 

nicolossus:
This PR sorts and formats imports in the Python codebase. The PR also introduces an `isort` runner to the CI and pre-commit. 

The following files, in particular, should be reviewed closely:
* `pyproject.toml`
* `.pre-commit-config.yml`
* `./github/workflows/nestbuildmatrix.yml`

In order to have the `nest` import sorted properly in user facing examples, `nest` is marked as a known third-party in the `isort` configuration (by default `isort` recognizes that `nest` actually is first-party and will sort `import nest` in a separate block at the end of the import section).


.github/workflows/nestbuildmatrix.yml
```diff
@@ -298,8 +311,9 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+
       - name: "Run black..."
-        uses: psf/black@bf7a16254ec96b084a6caf3d435ec18f0f245cc7   # 23.3.0
+        uses: psf/black@193ee766ca496871f93621d6b58d57a6564ff81b  # 23.7.0
```
nicolossus, line 316: During our developer meeting we discussed that we should be careful when bumping versions due to security concerns.

I have looked at the changes in v23.7.0 compared to v23.3.0 and have not seen any changes that raise a security concern. This particular package is maintained by the Python Software Foundation, which I think we generally can trust have a process in place for preventing nefarious code entering a release.

doc/htmldoc/networks/scripts/connections.py
```diff
@@ -21,8 +21,8 @@
 
 # create connectivity figures for spatial manual
 
-import nest
 import matplotlib.pyplot as plt
+import nest
```
heplesser, line 25: I am not really sure if I like it that NEST is hidden here between matplotlib and numpy.

Is there any way to configure isort so that nest always comes last?

heplesser (commented):
@nicolossus This looks neat for many cases and is a huge improvement, but as pointed out inline, but can we get more control about where our own stuff goes? I would like a three-block setup:

1. Python Library modules
1. Third-party modules
1. `import nest` with all its variants

Is that feasible?

doc/htmldoc/networks/scripts/connections.py
nicolossus: @heplesser Yes, the default sorting scheme of `isort` is actually as you describe with 1. Python library, 2. third-party and 3. first-party (in this case `nest`).

The reason `nest` is sorted as third-party is because I declared it to be counted as that. The rationale for that decision was that `nest` will then be sorted correctly for users copying our examples for their own use. 

I have no strong feelings about `nest` being counted as third- or first-party, so if want `nest` to be sorted as a first-party module I can fix that fairly quick.

heplesser: That is a good point. Maybe my view of the world is just too NEST-centric, seeing all other imports than `nest` to be just helpers in scripts importing `nest` as the main workhorse.  @terhorstd What do you think?

heplesser:
@nicolossus Could you update the PR to account for changes in master?

heplesser (approved):

doc/htmldoc/networks/scripts/connections.py
heplesser: Ok, I agree treating NEST as third party in example scripts makes sense.

nicolossus:
@jessica-mitchell Could you take a look at the changed documentation?

jessica-mitchell (approved):
thanks lgtm!

PR #2884, opened 2023-08-10
Hike minimum required CMake version to 3.19

heplesser:
#2835 took into use CMake function COMMAND_ERROR_IS_FATAL which is only available from CMake 3.19 on (https://github.com/nest/nest-simulator/blob/1a5838eb15321909ffed31de84cfa0b6e048b321/cmake/ProcessOptions.cmake#L664C10-L664C10).

This PR hikes the CMake requirement correspondingly and simplifies the requirement to just a minimum version.

This can cause problems for reproducibility runs on some supercomputers; e.g., JUSUF Stages/2020 supports only CMake 3.18, but is required for use of Python 3.8.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,9 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>
 
-# Version range from minimum required (3.12 for macOS OpenMP)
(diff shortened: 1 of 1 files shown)


heplesser:
> If you find the implications for JUSUF too severe, we could alternatively also try to find a replacement for `COMMAND_ERROR_IS_FATAL`.

It turned out we can use 3.21 on JUSUF, so no problem :).

gtrensch (approved):
Approved!

JanVogelsang (approved):

PR #2888, opened 2023-08-15
Move deprecated tests to `sli2py_ignore` directory

nicolossus:
This PR moves the SLI tests that will not be ported to pytest to designated directories.


terhorstd (approved):
Jointly reviewed. Approved! :+1:

PR #2895, opened 2023-08-16
Port `test_step_current_generator` from SLI-2-Py

nicolossus:


testsuite/pytests/sli2py_stimulating/test_step_current_generator.py
```diff
@@ -0,0 +1,367 @@
...
+"""
+
+import nest
+import numpy.testing as nptest
+import pandas as pd
+import pandas.testing as pdtest
+import pytest
+
+
+@pytest.fixture(autouse=True)
```
heplesser, line 36: Don't we have this globally somewhere?

testsuite/pytests/sli2py_stimulating/test_step_current_generator.py
```diff
@@ -0,0 +1,367 @@
...
+
+    nest.Simulate(25.0)
+
+    actual_recording = vm.events["V_m"]
+
+    nptest.assert_almost_equal(actual_recording, reference_recording, decimal=5)
+
+
+@pytest.fixture(scope="module")
+def reference_recording():
```
heplesser, line 124: Unfortunately, the original test does not say where the reference comes from, but the solution may have been explicitly computed since we know the closed-form solution. So I'd call it
```suggestion
def expected_V_m():
```

heplesser (changes requested):

testsuite/pytests/sli2py_stimulating/test_step_current_generator.py
nicolossus: The global `safety_reset` in `conftest.py` only resets the kernel for each module. An local reset is needed in order to reset the kernel between tests within a module.

heplesser (approved):

PR #2899, opened 2023-08-22
Port `test_ntree_split` from SLI-2-Py

nicolossus:


testsuite/pytests/test_spatial/test_ntree_split.py
```diff
@@ -0,0 +1,67 @@
...
+    r = 0.45
+
+    low_xy = -r
+    high_xy = r
+    dx = (high_xy - low_xy) / (pre_n_x - 1)
+    xs = np.arange(low_xy, high_xy + dx, dx)
+    y = 0.0
+    z = 0.0
+
+    positions = [[x, y, z] for x in xs]
```
heplesser, line 51: I don't like `arange()` because there could be rare roundoff problems. I'd suggest to simplify this to
```suggestion
    # Generate positions. The values used here have cause rounding errors previously.
    # previously. We need to have more than 100 position to force the ntree to split.
    positions = [[x, 0.0, 0.0] for x in np.linspace(-0.45, 0.45, 110)]
```

heplesser (changes requested):
Thanks, just a single suggestion for simplification.

heplesser (approved):

PR #2904, opened 2023-08-23
Port `test_oversize_mask` from SLI-2-Py

nicolossus:


heplesser (approved):
@nicolossus Thanks for the port! I guess one could have made the test even more compact by parameterizing over all variants for grid and for free respectively on a single test, but we have many tests to port ...

PR #2905, opened 2023-08-24
Port `test_rate_connections` from SLI-2-Py

nicolossus:


heplesser (approved):
Very nice, parameterization rocks!

PR #2906, opened 2023-08-25
Remove assimulo from requirements

heplesser:
As the Assimulo package (with a number of dependencies of its own) is only needed to run the aeif_models_implementation notebook. To reduce external dependencies, this PR removes it from our `environment.yml`.  

The notebook itself is updated to provide up-to-date installation guidance for Assimulo.


doc/htmldoc/model_details/aeif_models_implementation.ipynb
```diff
@@ -75,19 +76,9 @@
...
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "# Install assimulo package in the current Jupyter kernel\n",
-    "import sys\n",
+    "* [matplotlib](http://matplotlib.org/)\n",
     "\n",
-    "!{sys.executable} -m pip install assimulo"
+    "Note that the last version of Assimulo available from PyPI is version 3.0, which is not compatible with current versions of Python distribution tools. If you use conda/mamba, you can install a current version of Assimulo from `conda-forge`."
```
jessica-mitchell, line 81:
```suggestion
    "Note that the latest version of Assimulo available from PyPI is version 3.0, which is not compatible with current versions of Python distribution tools. If you use conda/mamba, you can install a current version of Assimulo from `conda-forge`."
```

jessica-mitchell: Also can we mention which minimal version of Assimulo works?

jessica-mitchell (changes requested):
Just minor question about Assimilo versions, see inline

doc/htmldoc/model_details/aeif_models_implementation.ipynb
heplesser: @jessica-mitchell I revised these two sentences to achieve greater clarity.

jessica-mitchell (approved):

clinssen (approved):
Makes sense to remove this from environments.yml, and the instructions in the notebook are clear. Thanks!

PR #2917, opened 2023-08-30
Port ticket-459.sli to Pytest and fix name for ticket_754.py

heplesser:
This PR port ticket-459.sli to Pytest with significantly simpler code. It also adds the needed `test_` to the ticket_754.py test so that those tests will actually be run.


nicolossus (approved):

heplesser:
@clinssen Ping!

clinssen (approved):

PR #2918, opened 2023-08-31
Test conda package build with NEST master branch integrated in CI

steffengraber:
The building of the package is kept as close as possible to the conda-forge package building. The tooling also corresponds to that of conda-forge.
The build runs under Ubuntu 22.04 with python3.10. But this can be changed or extended. Tests with MacOSX are also generally possible.

The build package is saved as an artifact under `nest_conda_Linux_x86_64` and can be tested afterward. 
To do this, download the zip file, unzip it and thus install NEST with the appropriate environment:

    mamba create -n MY-ENV -c "file://${PWD}/nest_conda_Linux_x86_64" nest-simulator

This also works with `conda` instead of `mamba`.


steffengraber:
As discussed in the NEST meeting, we want to use the conda-forge build workflows via github API calls.

steffengraber:
API calls to start and test the CI workflow for NEST are now integrated. 
@terhorstd and @gtrensch Some secrets need to be set, and we should discuss the integration in the NEST CI workflow.

heplesser:
@steffengraber This looks fine in principle, but the packaging test fails. Could you check that?

steffengraber:
The whole project is being carefully reconsidered. There should be better ways to integrate CIs. Therefore this PR will be closed.

Issue #2629, opened 2023-03-07
`DumpLayerConnections` fails when connecting source layer to more than one target layer


nicolossus:
This is a follow-up to an [issue reported on the mailing list](https://www.nest-initiative.org/mailinglist/hyperkitty/list/users@nest-simulator.org/thread/KI4HY3HLCECRLD2KEMMAEFF5476O77PI/). When using NEST 3.4 (or current master), the following script fails:

```Python 
import nest

pos = nest.spatial.grid(shape=[3, 3])
input = nest.Create('iaf_psc_alpha', positions=pos)

layer_0 = nest.Create('iaf_psc_alpha', positions=pos)
layer_1 = nest.Create('iaf_psc_alpha', positions=pos)

conn_neur = {'rule': 'pairwise_bernoulli',
             'use_on_source': True, 'mask': {'grid': {'shape': [3, 3]}}}

nest.Connect(input, layer_0, conn_neur)
nest.Connect(input, layer_1, conn_neur)
nest.DumpLayerConnections(input, layer_0, 'static_synapse', 'conn.txt')
```

The script works fine with NEST 3.3 (and Python 3.9). Hence, it seems like a bug has been introduced at some point, which makes `DumpLayerConnections` fail when a source layer is connected to more than one target layer. 

The error message reads: 

```
Assertion failed: (lid < dims_[ 0 ]), function lid_to_position, file grid_layer.h, line 223.
[machine-id] *** Process received signal ***
[machine-id] Signal: Abort trap: 6 (6)
[machine-id] Signal code:  (0)
[machine-id] [ 0] 0   libsystem_platform.dylib            0x00007ff80b9c2dfd _sigtramp + 29
[machine-id] [ 1] 0   ???                                 0x0000000000000000 0x0 + 0
[machine-id] [ 2] 0   libsystem_c.dylib                   0x00007ff80b8f8d24 abort + 123
[machine-id] [ 3] 0   libsystem_c.dylib                   0x00007ff80b8f80cb err + 0
[machine-id] [ 4] 0   libnestkernel.master.dylib          0x000000010868f873 _ZNK4nest9GridLayerILi2EE12get_positionEm.cold.1 + 35
[machine-id] [ 5] 0   libnestkernel.master.dylib          0x000000010867095c _ZNK4nest9GridLayerILi2EE12get_positionEm + 204
[...]

PR #2920, opened 2023-09-04
Filter connections also by target when dumping layer connections

nicolossus:
Fixes https://github.com/nest/nest-simulator/issues/2629.

The main purpose of this PR is to fix the bug reported in https://github.com/nest/nest-simulator/issues/2629. The reason for the bug was that the internal kernel call to `get_connections` did not filter by the provided target `NodeCollection`. 

The PR also introduces some minor changes regarding more informative variable names and declaring the appropriate data type. More majorly, the `find` functions of `node_collection.h` are renamed to `get_lid` to better reflect that the functions return an index and not an iterator (the `find` function for a container typically returns an iterator).

The final addition of this PR is that path-like objects are now allowed to be passed as filename to `DumpLayerConnections`.


testsuite/pytests/sli2py_regressions/test_issue_2629.py
```diff
@@ -0,0 +1,83 @@
...
+
+import nest
+
+
+@pytest.fixture(autouse=True)
+def reset():
+    nest.ResetKernel()
+
+
+def test_dump_layer_connections_multiple_targets(tmp_path):
```
heplesser, line 40: I'd suggest to split this into a fixture building the network and then two separate tests.

I assume the tmp_path fixture automatically deletes the files after the test?

heplesser (commented):
Thanks @nicolossus. I just have one question and suggestion concerning the test.

testsuite/pytests/sli2py_regressions/test_issue_2629.py
nicolossus: The test has been updated as you suggested in https://github.com/nest/nest-simulator/pull/2920/commits/398ec15a3d76e54c0467905512a40b7b66b5c275.

No, `tmp_path` directories are not deleted immediately after the test. According to the [Pytest documentation](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-default-base-temporary-directory), the last 3 temporary directories are kept. I guess this is for having artifacts available in case tests fail.

heplesser (approved):
@nicolossus Thank you! I will approve this now, but am slightly irritated that we needed the `stringify_path()` to support SLI shortly before SLI will be a matter of the past. Could you create a follow up issue to remove it again? An maybe create another follow-up issue to remind us to see if we want to convert all path handling to [pathlib](https://docs.python.org/3/library/pathlib.html).

jougs (approved):
This looks good to me. I guess we can handle the removal of the `stringify_path` function in the merge to `new-pynest`.

Issue #2902, opened 2023-08-23
Add thread id to the Connection class


pnbabu:
Currently, the `send()` function in the `Connection` class is the only place where the `thread_id` of the connection is available. This issue is to propose adding a new variable, `thread_id` to the `Connection` class as a private member and a corresponding private function `get_thread()` that returns the thread id. This makes it easier to query the `thread_id` from all the functions in the synapse models, along with the models written in NESTML.

The idea is to introduce a new variable `unsigned short int thread_id` in the `Connection` class. The connection object has a target identifier `target_` is 16 bits + 16 bits of padding + 32 bits of `SynIdDelay` data structure. The 16 bits of padding could be used for the connection's thread ID.

pnbabu:
As mentioned in #2921, the `thread_id` can be accessed via `kernel().vp_manager.get_thread_id()` from the synapse. A new issue is opened for this in NESTML: https://github.com/nest/nestml/issues/1158. Closing this issue for now.

PR #2921, opened 2023-09-06
Make thread id available inside connection's constructor and set_status

JanVogelsang:
This PR solves #2902. Some model parameters could potentially depend on the RNG (see NESTML), while each generator is bound to a specific thread. Currently, connections (i.e., synapses) are not aware of the thread that they correspond to. One solution would be to add the thread id to the connection class. However, while this wouldn't increase the total object size in the case of index-based node targeting (HPC case, `TargetIdentifierIndex`) due to a 16-bit padding that is currently unused, it would possibly increase the size when using the pointer-based node targeting (non-HPC case, `TargetIdentifierPtrRport`). In theory, it could be possible to restructure the connection class members in order to include the thread id in both cases without increasing the total object size, however this would not be guaranteed for all systems/architectures.  

Another solution would be to simply pass the thread id as parameter to all relevant functions of the connection class. When calling a function on a connection, it will always happen inside a threaded environment, therefore the thread id will always be available at that time. The only exception are default connections, which are not bound to a specific thread. In these cases however, the RNG of the main thread (thread id 0) can be used, if required at all.


JanVogelsang:
@heplesser Could you review this or alternatively choose another reviewer?

heplesser:
Adding @suku248 as a reviewer as this touches upon connection representation (even though it does not add data to connections in this PR).

@pnbabu Could you provide a use case that needs the `thread_id` for the synapse?

heplesser (commented):
As discussed in the NEST Kernel Team meeting on Thursday, passing the thread around should not be necessary since in almost all cases the synaptic parameters will be drawn by the thread responsible for the target neuron, so just getting the current thread will give the necessary information.

The next step will be to check this. If this holds, code could be much simplified.

heplesser:
@JanVogelsang @pnbabu As I pointed out earlier, I think we can just use `kernel().vp_manager.get_thread_id()` whenever a random number would be required during connection creation, because the connection always belongs to the target thread and is created by its thread. Then, we do not need to pass or store the thread id.

Do you think this is a viable idea? @pnbabu Could you test this with code generated from NESTML?

JanVogelsang:
Yes, that sounds like a good idea.

heplesser:
I will close this PR now in favour of a different solution as discussed [above](https://github.com/nest/nest-simulator/pull/2921#issuecomment-1816141880).

PR #2922, opened 2023-09-06
Port `issue-521.sli` to Pytest

nicolossus:


testsuite/pytests/sli2py_regressions/test_issue_521.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Ensure that *_hpc connections also return correct global target node ID on ``GetConnections``.
+    """
+
+    nest.ResetKernel()
+    nest.local_num_threads = num_threads
+
+    nc = nest.Create("iaf_psc_alpha", 20)
+
+    # connect two neurons (prime number node ids to increase variations of thread placement)
```
heplesser, line 43: The comment is confusing, since 4 and 12 are not prime numbers (for once, zero indexing really gets confusing here). I think it would actually be better to just choose 4 and 12 here as is, and then use prime numbers of threads, e.g., 1, 2, 5, 11 (I don't think we need more).

nicolossus: Fixed in [2d7a609](https://github.com/nest/nest-simulator/pull/2922/commits/2d7a609c8e88f4d08211d38a864271236d878857).

heplesser (approved):

PR #2933, opened 2023-09-11
Add details on Hodgkin Huxley models, based on asynchronicity discussion

### LINKED ISSUE #2910: Add additional documentaton regarding model Hodgkin Huxley
Add documentation based on discussion from mailing list regarding HH
Need to determine where this would fit best in the docs.

"Just a comment on the spike emission for Hodgkin-Huxley models. This is a non-trivial topic that requires the modeller to make a choice of what they want. Remember that the original Hodgkin-Huxley model is not a model of a neuron but a model for spike propagation down the squid axon. In this model, there is no such thing as a singular spike  in the sense we use it commonly when discussing integrate-and-fire point neuron models.

 In the HH model, under certain conditions, we have a rapidly rising Na current depolarizing the membrane up to around +30 mV, followed by a rapidly activating K current repolarizing the memrane again towards -70 mV. The resulting memrane potential excursion looks like a spike, but the dynamics are entirely continuous and described by the HH ODEs. These ODEs also incorporate the refractory mechanism. If one creates a chain of HH-compartments, activation will thus travel one way, again as an entirely continuous process. Only when the activation reaches an axon terminal will something new happen: exocytosis of transmitter substances into the synaptic cleft.

In typical point-neuron models, on the other hand, we have on the one hand the sub-threshold membrane potential dynamics, described by continuous ODEs, and some threshold condition. When the threshold condition is met, we say the neuron emits a spike (in the sense of a singular event causing exocytosis at axon terminals) and typically some reset mechanism applies, followed by some refractory mechanism.

 When we now force the HH model into this point-neuron framework subthreshold dynamics + threshold mechanism + reset + refractoriness, the modeller needs to make choices. For the subthreshold dynamics, one will usually use the HH equations, since this is a HH model, after all.

 For the threshold mechanism, various choices would be possible, but somehow one needs to detect the fast rise followed by fast fall of the membrane potential. In the built-in hh_psc_alpha model, this is done by

    else if ( S_.y_[ State_::V_M ] >= 0 and U_old > S_.y_[ State_::V_M ] ) // ( threshold and maximum )

The first criterion makes sure that we are well into the rising flank of the Na-driven depolarization, the second that we have just passed the maximum of that excursion.

For the reset, hh_psc_alpha does nothing, which makes sense, since the subthreshold HH dynamics have the reset built in—the K current.

 For refractoriness, hh_psc_alpha has a fixed refractory time (2 ms by default). In this model, the only effect of refractoriness is that it prohibits spiking. The subthreshold dynamics evolve freely during refractoriness. This again makes sense, since the HH dynamics have refractoriness built in. This is different in iaf-models, where the membrane potential is typically clamped to a reset potential during refractoriness.

 Now the latter point raises the question of why hh_psc_alpha has an explicit refractory mechanism even though the HH-dynamics include refractory effects already. The reason is technical.  Given that the Na-current will push the membrane potential to around +30 mV, it will take several time steps before the K-current will pull the potential to below 0 mV again. Then, if

 S_.y_[ State_::V_M ] >= 0 and U_old > S_.y_[ State_::V_M ]

 is the criterion for spike emission, a spike would be emitted for every time step during the downward flank of the membrane potential excursion until V_m < 0 again. This might be the effect you see. The refractory period in NEST's hh_psc_alpha simply suppresses these spikes during the downward flank.
A better criterion could be to not just compare V_m at two points in time but at three time steps to look for an actual maximum.

jessica-mitchell:
Resolves #2910 

This PR adds the text from the mailing list discussing asynchronicity in Hodgkin Huxley models.
The text is added to model_details folder, which might not fit there because the other files are Jupyter notebooks
but it is technical info on HH models. The text is linked in each HH model.

See output here: https://nest-simulator--2933.org.readthedocs.build/en/2933/model_details/hh_details.html#hh-details

### DIFF
--- a/doc/htmldoc/model_details/hh_details.rst
+++ b/doc/htmldoc/model_details/hh_details.rst
@@ -1,0 +1,67 @@
+.. _hh_details:
+
+A note on Hodgkin Huxley models
+===============================
+
+Asynchronicity between spikes and firing pattern
+------------------------------------------------
+
+Original question:
+~~~~~~~~~~~~~~~~~~
+
+I made a Hodgkin Huxley neuron model and I can get an action potential from it, 
+but it seems there is a problem with spikes.
+Action potential shape shows one spike, but the spike recorder shows several spikes 
+and there is not any synchronicity between them.
+
+
+Answer:
+~~~~~~~
+
+This is a non-trivial topic that requires the modeler to make a choice of what they want.
+The original Hodgkin-Huxley model is not a model of a neuron, but a model for spike propagation down the squid axon.
+In this model, there is no such thing as a singular spike in the sense we commonly use it when discussing
+integrate-and-fire point neuron models.
+
+In the HH model, under certain conditions, we have a rapidly rising `Na` current depolarizing the membrane up to around +30 mV,
+followed by a rapidly activating `K` current repolarizing the memrane again towards -70 mV.
+The resulting memrane potential excursion looks like a spike, but the dynamics are entirely continuous and described by the
+HH ODEs (ordinary differential equations).
+These ODEs also incorporate the refractory mechanism.
+If one creates a chain of HH-compartments, activation will thus travel one way, again as an entirely continuous process.
+Only when the activation reaches an axon terminal will something new happen: exocytosis of transmitter substances into the synaptic cleft.
+
+In contrast, typical point-neuron models have, on one hand, the sub-threshold membrane potential dynamics, described by continuous ODEs,
+and some threshold condition, on the other hand. When the threshold condition is met,
+we say the neuron emits a spike (in the sense of a singular event causing exocytosis at axon terminals),`
+and typically some reset mechanism applies, followed by some refractory mechanism.
+
+When we now force the HH model into this point-neuron framework (subthreshold dynamics + threshold mechanism + reset + refractoriness), the modeler needs to make choices.
+For the subthreshold dynamics, one will usually use the HH equations, since this is a HH model, after all.
+
+For the threshold mechanism, various choices would be possible, but somehow one needs to detect the fast rise followed by fast fall of the membrane potential.
+In the built-in ``hh_psc_alpha model``, this is done by
+
+.. code-block:: cpp
+
+    else if ( S_.y_[ State_::V_M ] >= 0 and U_old > S_.y_[ State_::V_M ] ) // ( threshold and maximum )
+
+The first criterion makes sure that we are well into the rising flank of the Na-driven depolarization; the second criterion ensurs that
+we have just passed the maximum of that excursion.
+For the reset, ``hh_psc_alpha`` does nothing, which makes sense, since the subthreshold HH dynamics have the reset built in—the `K` current.
+For refractoriness, ``hh_psc_alpha`` has a fixed refractory time (2 ms by default). In this model, the only effect of refractoriness is that it prohibits spiking.
+The subthreshold dynamics evolve freely during refractoriness. This again makes sense, since the HH dynamics have refractoriness built in.
+This is different in iaf-models, where the membrane potential is typically clamped to a reset potential during refractoriness.
+
+Now the latter point raises the question of why ``hh_psc_alpha`` has an explicit refractory mechanism even though the HH-dynamics include refractory effects already.
+The reason is technical. Given that the Na-current will push the membrane potential to around +30 mV,
+it will take several time steps before the K-current will pull the potential to below 0 mV again. Then, if
+
+.. code-block:: cpp
+
+    S_.y_[ State_::V_M ] >= 0 and U_old > S_.y_[ State_::V_M ]
+
+is the criterion for spike emission, a spike would be emitted for every time step during the downward flank of the membrane potential excursion until `V_m < 0` again.
+This might take the effect of the action potential shape showing one spike, but spike recorder showing several spikes with no synchronicity between them.
+The refractory period in NEST's ``hh_psc_alpha`` simply suppresses these spikes during the downward flank.
+A better criterion could be to not just compare V_m at two points in time but at three time steps to look for an actual maximum.
--- a/doc/htmldoc/model_details/hh_details.rst
+++ b/doc/htmldoc/model_details/hh_details.rst
@@ -1,39 +1,28 @@
 .. _hh_details:
 
-A note on Hodgkin Huxley models
-===============================
-
-Asynchronicity between spikes and firing pattern
-------------------------------------------------
-
-Original question:
-~~~~~~~~~~~~~~~~~~
-
-I made a Hodgkin Huxley neuron model and I can get an action potential from it, 
-but it seems there is a problem with spikes.
-Action potential shape shows one spike, but the spike recorder shows several spikes 
-and there is not any synchronicity between them.
+How NEST generates spikes in Hodgkin-Huxley style models
+========================================================
 
 
-Answer:
-~~~~~~~
 
-This is a non-trivial topic that requires the modeler to make a choice of what they want.
+
+The way in which Hogkin-Huxley style neuron models in NEST generate outgoing spikes is not trivial. In particular when implementing new HH-style models, for example via NESTML, this can lead to confusing observations, where spikes recorded by a ``spike_recorder`` seem to disagree with membrane potential traces recorded with a ``voltmeter``. This document addresses the underlying challenges.
+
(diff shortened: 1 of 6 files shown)


doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Answer:
+~~~~~~~
+
+This is a non-trivial topic that requires the modeler to make a choice of what they want.
+The original Hodgkin-Huxley model is not a model of a neuron, but a model for spike propagation down the squid axon.
+In this model, there is no such thing as a singular spike in the sense we commonly use it when discussing
+integrate-and-fire point neuron models.
+
+In the HH model, under certain conditions, we have a rapidly rising `Na` current depolarizing the membrane up to around +30 mV,
```
clinssen, line 26: "sodium current" or ``:math:`\text{Na}^+` ``; same for K+

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+followed by a rapidly activating `K` current repolarizing the memrane again towards -70 mV.
+The resulting memrane potential excursion looks like a spike, but the dynamics are entirely continuous and described by the
```
clinssen, line 28: typo

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+HH ODEs (ordinary differential equations).
+These ODEs also incorporate the refractory mechanism.
+If one creates a chain of HH-compartments, activation will thus travel one way, again as an entirely continuous process.
+Only when the activation reaches an axon terminal will something new happen: exocytosis of transmitter substances into the synaptic cleft.
+
+In contrast, typical point-neuron models have, on one hand, the sub-threshold membrane potential dynamics, described by continuous ODEs,
+and some threshold condition, on the other hand. When the threshold condition is met,
+we say the neuron emits a spike (in the sense of a singular event causing exocytosis at axon terminals),`
```
clinssen, line 36:
```suggestion
we say the neuron emits a spike (in the sense of a singular event causing exocytosis at axon terminals),
```

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+For the subthreshold dynamics, one will usually use the HH equations, since this is a HH model, after all.
+
+For the threshold mechanism, various choices would be possible, but somehow one needs to detect the fast rise followed by fast fall of the membrane potential.
+In the built-in ``hh_psc_alpha model``, this is done by
+
+.. code-block:: cpp
+
+    else if ( S_.y_[ State_::V_M ] >= 0 and U_old > S_.y_[ State_::V_M ] ) // ( threshold and maximum )
+
+The first criterion makes sure that we are well into the rising flank of the Na-driven depolarization; the second criterion ensurs that
```
clinssen, line 49:
```suggestion
The first criterion makes sure that we are well into the rising flank of the Na-driven depolarization; the second criterion ensures that
```

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Now the latter point raises the question of why ``hh_psc_alpha`` has an explicit refractory mechanism even though the HH-dynamics include refractory effects already.
+The reason is technical. Given that the Na-current will push the membrane potential to around +30 mV,
+it will take several time steps before the K-current will pull the potential to below 0 mV again. Then, if
+
+.. code-block:: cpp
+
+    S_.y_[ State_::V_M ] >= 0 and U_old > S_.y_[ State_::V_M ]
+
+is the criterion for spike emission, a spike would be emitted for every time step during the downward flank of the membrane potential excursion until `V_m < 0` again.
```
clinssen, line 64:
```suggestion
is the criterion for spike emission, a spike would be emitted for every time step during the downward flank of the membrane potential excursion until :math:`V_m < 0` again.
```

clinssen (changes requested):

doc/htmldoc/model_details/hh_details.rst
heplesser: I suggest to write out "sodium" and "potassium", respectively.

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+
+This is a non-trivial topic that requires the modeler to make a choice of what they want.
+The original Hodgkin-Huxley model is not a model of a neuron, but a model for spike propagation down the squid axon.
+In this model, there is no such thing as a singular spike in the sense we commonly use it when discussing
+integrate-and-fire point neuron models.
+
+In the HH model, under certain conditions, we have a rapidly rising `Na` current depolarizing the membrane up to around +30 mV,
+followed by a rapidly activating `K` current repolarizing the memrane again towards -70 mV.
+The resulting memrane potential excursion looks like a spike, but the dynamics are entirely continuous and described by the
+HH ODEs (ordinary differential equations).
```
heplesser, line 29:
```suggestion
Hogkin-Huxley ordinary differential equations (ODEs).
```

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+Asynchronicity between spikes and firing pattern
+------------------------------------------------
+
+Original question:
+~~~~~~~~~~~~~~~~~~
+
+I made a Hodgkin Huxley neuron model and I can get an action potential from it, 
+but it seems there is a problem with spikes.
+Action potential shape shows one spike, but the spike recorder shows several spikes 
+and there is not any synchronicity between them.
```
heplesser, line 15:
```suggestion
How NEST generates spikes in Hodgkin-Huxley style models
========================================================

```

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+
+
+Answer:
+~~~~~~~
+
+This is a non-trivial topic that requires the modeler to make a choice of what they want.
+The original Hodgkin-Huxley model is not a model of a neuron, but a model for spike propagation down the squid axon.
+In this model, there is no such thing as a singular spike in the sense we commonly use it when discussing
+integrate-and-fire point neuron models.
```
heplesser, line 24:
```suggestion

The way in which Hogkin-Huxley style neuron models in NEST generate outgoing spikes is not trivial. In particular when implementing new HH-style models, for example via NESTML, this can lead to confusing observations, where spikes recorded by a ``spike_recorder`` seem to disagree with membrane potential traces recorded with a ``voltmeter``. This document addresses the underlying challenges.

The original Hodgkin-Huxley model is not a model of a neuron, but a model for spike propagation down the squid axon.
In this model, there is no such thing as a singular spike in the sense we commonly use it when discussing
integrate-and-fire point neuron models.
```

heplesser (changes requested):
@jessica-mitchell Thanks! I agree with the comments by @clinssen. I made some further suggestions for improvement. In particular, I would modify the opening. I think my form makes it more comprehensible for a reader.

doc/htmldoc/model_details/hh_details.rst
jessica-mitchell, line 26:
```suggestion
In the HH model, under certain conditions, we have a rapidly rising sodium current depolarizing the membrane up to around +30 mV,
```

doc/htmldoc/model_details/hh_details.rst
```diff
@@ -0,0 +1,67 @@
...
+
+In the HH model, under certain conditions, we have a rapidly rising `Na` current depolarizing the membrane up to around +30 mV,
+followed by a rapidly activating `K` current repolarizing the memrane again towards -70 mV.
```
jessica-mitchell, line 27:
```suggestion
followed by a rapidly activating potassium current repolarizing the membrane again towards -70 mV.
```

doc/htmldoc/model_details/hh_details.rst
jessica-mitchell, line 28:
```suggestion
The resulting membrane potential excursion looks like a spike, but the dynamics are entirely continuous and described by the
```

doc/htmldoc/model_details/hh_details.rst
jessica-mitchell: done

heplesser (approved):

heplesser (approved):

clinssen (approved):
Cheers!

Issue #2779, opened 2023-05-10
Should the CI fail again when the NEST build raises warnings?


nicolossus:
Before the merge of https://github.com/nest/nest-simulator/pull/2590 with the new CI setup, the CI would fail if any warnings were raised in the building of NEST. Known warnings could be excluded in the now removed `parse_build_log.py`. 

Currently, the CI passes even though warnings are raised. Do we want this new behavior or should the old behavior be reinstated?

nicolossus:
@heplesser @jougs @terhorstd Any opinions?

jougs:
Generally, I think we should strive for zero warnings during the build.

Can you have a look at how to get the old behavior back? I assume this is not so easy now that we don't write the logs to our own files anymore, or is there a way to programmatically look at the output that GitHub Actions caches?

PR #2935, opened 2023-09-12
Add GCC compiler problem matcher to CI

nicolossus:
Resolves https://github.com/nest/nest-simulator/issues/2779.

This PR introduces a GitHub Actions [problem matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) which scans the GCC compiler output for a specified regex pattern and surface the information in the UI as [annotations]( https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#annotations-object-1).

The problem matcher is identical to the one used in VS Code's `cpptools`: 
https://github.com/microsoft/vscode-cpptools/blob/2cc55ea6ebe96b91c118c84784e980bdfaf20304/Extension/package.json#L303-L320.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -692,11 +692,19 @@ jobs:
               -Dwith-libneurosim=${{ contains(matrix.use, 'libneurosim') && '$HOME/.cache/libneurosim.install' || 'OFF' }} \
               ..
 
+      - name: "Add GCC problem matcher"
+        run: |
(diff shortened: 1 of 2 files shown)


heplesser:
@jougs @terhorstd I think you had already looked at this and expressed support. Could you give formal feedback?

terhorstd (approved):
I don't have a strong opinion on including this in the CI. The GitHub parsing of stdout magic strings is very irritating, but the changes don't constitute a strong binding to this proprietary service.

PR #2940, opened 2023-09-13
Fix up HTML output to clean up browser warnings

jessica-mitchell:
This PR fixes some warnings that are output in the browser.
Removing extra stylesheet links in layout.html that aren't needed and add `version_dropdown` option to false to HTML config.


steffengraber (approved):
:+1:

PR #2944, opened 2023-09-14
Add calibration function for models, called when changing resolution, to NEST2

clinssen:
This would be nice to have for NESTML. It was added as part of the big NEST 3 pull request by @hakonsbm. This PR is only a cherry-pick of that commit (2dac11d).

### DIFF
--- a/nestkernel/genericmodel.h
+++ b/nestkernel/genericmodel.h
@@ -60,6 +60,7 @@ public:
   bool potential_global_receiver();
   bool one_node_per_process();
   bool is_off_grid();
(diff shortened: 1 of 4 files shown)


pnbabu:
@clinssen should this change be on the `2.20.2` branch instead since NESTML supports only that version for NEST2?

clinssen:
Oh, yes, it should be on the latest NEST 2 development branch. Which is that? (cc @terhorstd)

terhorstd:
This seems to be not relevant anymore, since no more 2.x releases are planned.

PR #2945, opened 2023-09-14
Correct limit for test on target fields and remove duplicate assertions

heplesser:
The merge of #2926 introduced a subtle problem with a C++ test for setting values in the `Target` bitfield. The problem occurred only sporadically since the test was randomized based on walltime. This PR fixes the test and the random seed. It also removes duplicate assertions.

This i critical because it breaks the testsuite.


nicolossus (approved):

jougs (approved):
LGTM!

PR #2948, opened 2023-09-18
Fix bug in reconstruction of connection data structures after disconnect and fix small problems with examples

heplesser:
This PR contains two parts:
- A critical bug fix including new test to ensure connection data structures are properly rebuilt after disconnect.
- Fixes to some examples and the example runner to ensure they work with current master.

This PR changes the order of structural plasticity and spike delivery: When using compressed spikes, we can only deliver spikes through the same connection setup in which they were gathered.

@terhorstd @jougs I label this one as critical because it must make it into 3.6.

### DIFF
--- a/examples/run_examples.sh
+++ b/examples/run_examples.sh
@@ -93,11 +93,13 @@ for i in $EXAMPLES; do
     echo "  output_dir: '$output_dir'" >>"$metafile"
     echo "  log: '$logfile'" >>"$metafile"
 
-    export NEST_DATA_PATH=""  # $output_dir"
+    export NEST_DATA_PATH="$output_dir"
     touch .start_example
     sleep 1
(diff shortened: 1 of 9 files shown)


otcathatsya (approved):

jhnnsnk (approved):
Thanks!
I have run a structural plasticity example and it gives the same result as with NEST 3.5. The test also works.
I am just wondering if in the `simulation_manager.cpp` the `#pragma`'s are needed because if I try to run with multiple threads I get the error:
`nest.lib.hl_api_exceptions.Structural plasticity can not be used with multiple threads: Structural plasticity can not be used with multiple threads in SLI function EnableStructuralPlasticity: `. But this goes beyond the specific changes of this PR.

PR #2954, opened 2023-09-22
Fix image generation in Sudoku and Pong examples for compatibility with Matplotlib 3.8

### LINKED ISSUE #2951: Images from Sudoku and Pong examples do not render properly
I just tried the Sudoku and Pong examples in NEST (see also #2950, change from `fps` to `duration` in `imageio`). At least on my computer, the images do not render properly: later images seem to be put on top of older ones, and in the parts of the Sudoku with grey background one cannot see the discovered numbers, see the examples below.

The problems are apparent in the PNGs, so it is likely a Matplotlib problem. I used Matplotlib 3.8.0 with the MacOSX backend.

@krachbumm3nte Could you check what might be going wrong?

img_000496

krachbumm3nte:
Fixes #2951. I tested the changes with Matplotlib 3.8.0, but do not have access to MacOS. Since the Issue was caused by [a documented change](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.6.0.html#auto-removal-of-overlapping-axes-by-plt-subplot-and-plt-subplot2grid) in Matplotlib, the fix should work cross-system.


heplesser (approved):
Thanks for the quick fix! I have created a PR towards you repo with minor formatting fixes, but will approve already.

steffengraber (approved):
Checked it on my Ubuntu machine and everything worked fine. Assuming that you still fix the minor format problems, and the CI runs through, I give it a :+1:. Thanks for your work!.

PR #2956, opened 2023-09-23
Code re-organization for shorter build times (or taming the 96% troll)

heplesser:
This PR re-organizes the way in which node and connection models are registered in modelsmodule. Up to now, compiling `modelsmodule.cpp` triggers a large number of template instantiations and compiling this file alone takes about 112 seconds on my MacBook Pro with Apple Clang 15. This causes a noticeable wait during compilation around 96% complete. And this happens every time you touch the h-file for a single model. Experiments indicate that of the 112s, about 90s were spent on synapse models and 11s on node models. Even with all `register*` calls commented out compiling the file took still about 9s—not really surprising given that `modelsmodule.ii` is some 166.000 lines long after all `#include`s have been handle (yes onehundredandsixtysixthousand).

To avoid this single stumbling block, this PR ensures that each model forces its own template instantiation. It does to by declaring (in the h-file) and defining (in the cpp-file) a function
```
void
register_multimeter()
{
  kernel().model_manager.register_node_model< multimeter >( "multimeter" );
}
```
This function is then called from `multimeter.cpp`. The function is not declared as a (static) method because some neuron models are templated classes themselves, so we need a method outside the class. 

Since there is now a specific function to register each model, I decided to call the function `register_<name>` for all kind of models, with no distinction between nodes and synapses.

I had to add cpp-files for all synapses which did not yet have one to have a compilation unit in which the respective templates can be instantiated.

I auto-generated the changes with an ipython notebook (you can find it accidentally commited a few commits back), so the placement of declarations and definitions might not be perfect but was dictated by what a reasonably simple parser could achieve.

So what is the outcome?

1. On my MacBook Pro with four cores, overall build times do not differ much (close to 5 minutes).
2. On a dual processor EPYC Rome system with 128 cores in total, build time with `make -j128` is down from about 5 minutes to about 3.5 minutes.
3. The real gain is when you change a single model file, e.g. `touch iaf_psc_alpha.h`: With current master, you will wait for about 2 minutes, with the change here you will be done after less than 20 seconds.

Many files are changed here, but practically all files are changed in the same way.

I give this one high priority because it will make developers so much happier and more productive :).


models/ac_generator.cpp
```diff
@@ -41,6 +43,12 @@
 
 namespace nest
 {
+void
+register_ac_generator()
+{
+  kernel().model_manager.register_node_model< ac_generator >( "ac_generator" );
+}
```
jougs, line 50: I think this would be a good time to create a wrapper for `register_node_model()` in `nest.h`/nest_impl.h` as we already have it for `register_connection_model()`. I guess that we could then also replace most of the includes above by just `nest_impl.h`

jougs (commented):
What a heroic effort! :clap: :tada:

I have added two inline comments that apply to all of the models changed here.

And one more thing: Could it be smart to change the function signature of `register_*()` to accept the name under which the model will be registered as an argument?

That way, the call to the function could decide what the name should be and the generator for the modelsmodule could be extended later on to support a modelset syntax like "iaf_psc_alpha -> standard_neuron"

models/bernoulli_synapse.cpp
```diff
@@ -0,0 +1,34 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "bernoulli_synapse.h"
+
+// Includes from nestkernel:
+#include "connector_model_impl.h"
+#include "model_manager_impl.h"
```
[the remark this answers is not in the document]
heplesser: I may have solved the right problem in the wrong place. The two includes should be in `nest_impl.h`, not in the individual synaspe cpp files. We need both to resolve the entire call chain of templated functions.

models/ac_generator.cpp
heplesser: Indeed.

heplesser:
> What a heroic effort! 👏 🎉

Thanks 😊!

> And one more thing: Could it be smart to change the function signature of `register_*()` to accept the name under which the model will be registered as an argument?
> 
> That way, the call to the function could decide what the name should be and the generator for the modelsmodule could be extended later on to support a modelset syntax like "iaf_psc_alpha -> standard_neuron"

You are right, should not be hard to fix.

jougs:
@heplesser: can you maybe prioritize this? I'd love to have this in the module refactoring PR and time is running out...

heplesser:
> @heplesser: can you maybe prioritize this? I'd love to have this in the module refactoring PR and time is running out...

On it!

heplesser:
@jougs @pnbabu I have revised the code now. Could you review again?

jougs (approved):
Many thanks for addressing my concerns. I wholeheartedly approve ;-)

PR #2960, opened 2023-09-25
New connection rule for Poisson-distributed number of synapses between neuron pairs

ackurth:
In my work, I found it practical to adapt the ```pairwise_bernoulli``` connection method to a ```pairwise_poisson``` method, where the number of synapses between pre- and post-synaptic neurons is drawn from a Poisson distribution.

The choice of the Poisson distribution is not arbitrary.
Imagine you have the average number of synapses ```p``` between two neurons of specific populations.
Oftentimes, these specific populations have distinct targeting patters: same inhibitory neurons target preferentially basal, other distal dendrites etc.
Assume that the process on which a synapses is established has overall length L and that this L is similar for all neurons in the target populations
Since the overall number of synapses between two neurons of pre- and postsynaptic populations is known (```p```), on a small piece of the target process with length ```x````, the probability of the synapses being established on that piece is p*x.
Here of course we assume that the connection probability is identical everywhere on the target process.
To first order, this approximation seems fine.
You can probably already see where I am going: The synapses are thus a Poisson process on the line, the number of synapses is Poisson distributed.

This has two additional advantages:
First, the heterogeneity in the network is bigger in comparison the the standard Erdos-Renyi graphs generated by the ```pairwise_bernoulli``` connection method.
Indeed, in this case variance relative to the mean (i.e. the Fano Factor) is ```N_1*N_2*p*(1-p) / (N_1 * N_2 * p) = (1-p) <= 1```with equality iff ```p = 1```.
Here, ```N_1, N_2``` are the number of neurons in the pre- and post-synaptic populations.
For the Poisson based rule the Fano Factor equals ```1```.

Additionally, especially when working with spatial networks it frequently occurs that the average number of synapses between to populations (then interpreted as connection probability)  on short distances exceeds ```1```.
This leads to errors when using ```pairwise_bernoulli```, but is not a problem for the ```pairwise_poissson```.

Anyway, in the current PR such a connection scheme is implemented.
If you deem this sufficiently interesting to potentially merge it into master I will add the documentation.


heplesser (commented):
Hello @ackurth! Thank you for this contribution, which seems useful. I have only had a first look at the code, so here a few initial observations:
- Please make sure the formatting is in order.
- I don't like the `lam` parameter name, and yes, the one true design flaw in Python is `lambda` as a keyword, but even `lambda` would not be good. Can you find a name that is neurobiologically meaningful?
- As far as I could see, while you give a very nice justification for the new rule in the PR here, there is no documentation or example for the new rule. Could you add that?

jessica-mitchell:
@ackurth I second the need for documentation; there should be a section added in the `doc/htmldoc/synapses/connection_management.rst` at the very least

ackurth:
@heplesser 
I changed the variable name from ```lam``` to ```pairwise_avg_num_conns```.
It's a bit wordy but I think the most precise. Unless one wants to exchange ```average``` with ```expected```.

I also checked the and corrected to formatting now. I have to say the new (to me) workflow for this is quite nice.

@heplesser @jessica-mitchell 
Sure, documentation must be added.
This is what I was aiming at with the last sentence of the message of the PR.
I now added some documentation in the ```doc/htmldoc/synapses/connection_management.rst``` as well as in the ```doc/htmldoc/networks/spatially_structured_networks.rst``` (latter I think is in order since the new connection rule is also nice and works for spatial networks).
Anyway, I did find that the description of why the connection rule is nice did not really fit in the ```doc/htmldoc/synapses/connection_management.rst```.
Is there another place where I should write it down? Do we want this to be part of the documentation anyway?

jessica-mitchell:
@ackurth I see what you mean, but I can't think of another document that would suit  such a description. In this case, I would suggest adding a note or subsection in connection_management. @heplesser do you have any other suggestions?

nestkernel/random_generators.h
```diff
@@ -115,6 +115,13 @@ class BaseRandomGenerator
    * @param N Maximum value that can be drawn.
    */
   virtual unsigned long ulrand( unsigned long N ) = 0;
+
+  /**
+   * @brief Uses the wrapped RNG engine to draw an unsigned long from a Poisson distribution with mean lam.
```
nicolossus, line 120:
```suggestion
   * @brief Uses the wrapped RNG engine to draw an unsigned long from a Poisson distribution with mean mu.
```

In my opinion, `mu` is a bit better than `lam` as variable name here. Applies to the occurrences below as well.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -213,6 +216,10 @@ def _process_spatial_projections(conn_spec, syn_spec):
             projections["connection_type"] = "pairwise_bernoulli_on_target"
             if "use_on_source" in projections:
                 projections.pop("use_on_source")
+    elif conn_spec["rule"] == "pairwise_poisson":
+        if "use_on_source" in conn_spec:
+            raise ValueError("'use_on_source' can only be set when using pairwise_bernoulli")
```
nicolossus, line 221:
```suggestion
            raise ValueError("'use_on_source' can only be set when using 'pairwise_bernoulli'.")
```

testsuite/pytests/test_connect_pairwise_poisson.py
```diff
@@ -0,0 +1,111 @@
...
+# NEST is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import unittest
```
nicolossus, line 23: @ackurth Could you convert this test to the new style using Pytest? 

For inspiration, see, e.g., [test_multimeter.py](https://github.com/nest/nest-simulator/blob/master/testsuite/pytests/test_multimeter.py), [test_rate_connections.py](https://github.com/nest/nest-simulator/blob/master/testsuite/pytests/sli2py_connect/test_rate_connections.py), [test_spike_generator.py](https://github.com/nest/nest-simulator/blob/master/testsuite/pytests/sli2py_stimulating/test_spike_generator.py) and [test_threaded_spike_recorder.py](https://github.com/nest/nest-simulator/blob/master/testsuite/pytests/sli2py_recording/test_threaded_spike_recorder.py).

nicolossus (changes requested):
@ackurth Thanks! Overall this looks good to me. I only have some minor nitpicks, see inline comments.

@heplesser @jessica-mitchell This PR got me thinking that it would be nice to have a minimalistic example where the differences (including the different parametrizations) between all connection rules are shown. I think this can be handled in a separate PR. If you agree, I can create an issue to remind us.

testsuite/pytests/test_connect_pairwise_poisson.py
ackurth: Hey @nicolossus.
Thanks for the review and the nice comments.
To my understanding of pytest (which is limited, so please correct me if I am wrong) converting the this test would mean to rework the general testing suite for connections.
Currently, the tests for connections rules inherit from ```connect_test_base``` which defines (among other things) universal tests for connection routines.
These, I would argue, one wants to parameterize using the available functionality from ```pytest``` and then have specific scripts for the specific tests of the different connection rules.
The only quick way to achieve the proposed would be to mix ```pytest``` and ```unittest``` which might not be the best idea.
A proper solution would in my understanding require a rewriting of the "test hierarchy" for the connection rules and is, I think, beyond the scope of this PR.
If we aim for this, you can just create an issue and assign me to it.
As soon as I find the time I'd take care of this then.

nicolossus: @ackurth You are right, I didn't see the huge effort a conversion would require. A conversion should happen at some point, but I agree it is beyond the scope of this PR.

nestkernel/random_generators.h
```diff
@@ -115,6 +115,13 @@ class BaseRandomGenerator
    * @param N Maximum value that can be drawn.
    */
   virtual unsigned long ulrand( unsigned long N ) = 0;
+
+  /**
+   * @brief Uses the wrapped RNG engine to draw an unsigned long from a Poisson distribution with mean mu.
+   *
+   * @param lam Mean value of Poisson distribution.
```
nicolossus, line 122:
```suggestion
   * @param mu Mean value of Poisson distribution.
```

nestkernel/random_generators.h
```diff
@@ -115,6 +115,13 @@ class BaseRandomGenerator
...
+   */
+  virtual unsigned long prand( double lam ) = 0;
```
nicolossus, line 124:
```suggestion
  virtual unsigned long prand( double mu ) = 0;
```

nestkernel/random_generators.h
```diff
@@ -266,10 +273,18 @@ class RandomGenerator final : public BaseRandomGenerator
     return uniform_ulong_dist_( rng_, param );
   }
 
+  inline unsigned long
+  prand( double lam ) override
+  {
+    std::poisson_distribution< unsigned long >::param_type param( lam );
```
nicolossus, line 279:
```suggestion
  prand( double mu ) override
  {
    std::poisson_distribution< unsigned long >::param_type param( mu );
```

nicolossus (approved):
@ackurth LGTM! Just a few variable names that needs to be updated, but I'll approve now.

jessica-mitchell:
> @heplesser @jessica-mitchell This PR got me thinking that it would be nice to have a minimalistic example where the differences (including the different parametrizations) between all connection rules are shown. I think this can be handled in a separate PR. If you agree, I can create an issue to remind us.

@nicolossus Could you give me a little example of what you mean just so I can get clearer picture?

ackurth:
> @ackurth LGTM! Just a few variable names that needs to be updated, but I'll approve now.

Thanks @nicolossus! I committed your suggestions. Sorry that they slipped.

nestkernel/random_generators.h
```diff
@@ -115,6 +115,13 @@ class BaseRandomGenerator
    * @param N Maximum value that can be drawn.
    */
   virtual unsigned long ulrand( unsigned long N ) = 0;
+
+  /**
+   * @brief Uses the wrapped RNG engine to draw an unsigned long from a Poisson distribution with mean mu.
+   *
+   * @param mu Mean value of Poisson distribution.
+   */
+  virtual unsigned long prand( double mu ) = 0;
```
heplesser, line 124: You should not introduce new random number functions here, the interface is limited to `drand()` (for double-valued densities) and `ulrand()` (for integer-valued distributions). You need to create a `poisson_distribution` object in the connection builder and obtain the random number from it, similar to the `FixedTotalNumber` builder:

https://github.com/nest/nest-simulator/blob/9575a63b2c65cab5ea1f2bc092a86ba1acccb95a/nestkernel/conn_builder.cpp#L1385

nestkernel/random_generators.h
```diff
@@ -266,10 +273,18 @@ class RandomGenerator final : public BaseRandomGenerator
     return uniform_ulong_dist_( rng_, param );
   }
 
+  inline unsigned long
+  prand( double mu ) override
+  {
+    std::poisson_distribution< unsigned long >::param_type param( mu );
+    return poisson_ulong_dist_( rng_, param );
+  }
+
```
heplesser, line 282: See above

nestkernel/random_generators.h
```diff
@@ -266,10 +273,18 @@ class RandomGenerator final : public BaseRandomGenerator
...
 private:
   RandomEngineT rng_; //!< Wrapped RNG engine.
   std::uniform_int_distribution< unsigned long > uniform_ulong_dist_;
   std::uniform_real_distribution<> uniform_double_dist_0_1_;
+  std::poisson_distribution< unsigned long > poisson_ulong_dist_;
```
heplesser, line 287: See above

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -233,6 +240,13 @@ def _connect_layers_needed(conn_spec, syn_spec):
         rule_is_bernoulli = "pairwise_bernoulli" in str(conn_spec["rule"])
         if "mask" in conn_spec or ("p" in conn_spec and not rule_is_bernoulli) or "use_on_source" in conn_spec:
             return True
+        rule_is_poisson = "pairwise_poisson" in str(conn_spec["rule"])
+        if (
+            "mask" in conn_spec
+            or ("pairwise_avg_num_conns" in conn_spec and not rule_is_poisson)
+            or "use_on_source" in conn_spec
+        ):
+            return True
```
heplesser, line 249:
```suggestion
        rule_is_poisson = "pairwise_poisson" in str(conn_spec["rule"])
        if (
            "mask" in conn_spec
            or ("pairwise_avg_num_conns" in conn_spec and not rule_is_poisson)
        ):
            return True
```

nestkernel/connection_creator_impl.h
```diff
@@ -110,6 +115,56 @@ ConnectionCreator::connect_to_target_( Iterator from,
...
+  RngPtr rng = get_vp_specific_rng( tgt_thread );
+
+  // We create a source pos vector here that can be updated with the
+  // source position. This is done to avoid creating and destroying
+  // unnecessarily many vectors.
+  std::vector< double > source_pos( D );
+  const std::vector< double > target_pos = tgt_pos.get_vector();
+
+  // Declare number of connections variable
+  unsigned long num_conns;
```
heplesser, line 136: Declare locally in loop.
```suggestion
```

nestkernel/connection_creator_impl.h
```diff
@@ -110,6 +115,56 @@ ConnectionCreator::connect_to_target_( Iterator from,
...
+  for ( Iterator iter = from; iter != to; ++iter )
+  {
+    if ( not allow_autapses_ and ( iter->second == tgt_ptr->get_node_id() ) )
+    {
+      continue;
+    }
+    iter->first.get_vector( source_pos );
+
+    // Sample number of connections that are to be established
+    num_conns = rng->prand( kernel_->value( rng, source_pos, target_pos, source, tgt_ptr ) );
```
heplesser, line 148:
```suggestion
    const unsigned long num_conns = rng->prand( kernel_->value( rng, source_pos, target_pos, source, tgt_ptr ) );
```

nestkernel/connection_creator_impl.h
```diff
@@ -110,6 +115,56 @@ ConnectionCreator::connect_to_target_( Iterator from,
...
+    if ( without_kernel or num_conns )
```
heplesser, line 149: `without_kernel` will have the same value for all iterations of the loop, won't it? Then you could return at the top of the method if it is `false`. For clarity, I would write `num_conns > 0`—although you don't need it since the subsequent loop is not run if `num_conns == 0`.  In the loop, I would stick to C++ conventions and let it run from zero.

nestkernel/connection_creator_impl.h
```diff
@@ -337,6 +392,82 @@ ConnectionCreator::pairwise_bernoulli_on_target_( Layer< D >& source,
...
+  else
+  {
+    pool.define( source.get_global_positions_vector( source_nc ) );
+  }
+
+  std::vector< std::shared_ptr< WrappedThreadException > > exceptions_raised_( kernel().vp_manager.get_num_threads() );
+
+// sharing specs on next line commented out because gcc 4.2 cannot handle them
+#pragma omp parallel // default(none) shared(source, target, masked_layer,
+                     // target_begin, target_end)
```
heplesser, line 424: We don't use sharing info on any other parallel sections, so just remove these comments (please also do so in all other places they occur in this file):
```suggestion
#pragma omp parallel
```

nestkernel/conn_builder.cpp
```diff
@@ -1585,6 +1585,130 @@ nest::BernoulliBuilder::inner_connect_( const int tid, RngPtr rng, Node* target,
...
+  NodeCollectionPTR targets,
+  const DictionaryDatum& conn_spec,
+  const std::vector< DictionaryDatum >& syn_specs )
+  : ConnBuilder( sources, targets, conn_spec, syn_specs )
+{
+  ParameterDatum* pd = dynamic_cast< ParameterDatum* >( ( *conn_spec )[ names::pairwise_avg_num_conns ].datum() );
+  if ( pd )
+  {
+    pairwise_avg_num_conns_ = *pd;
+    // TODO: Checks of parameter range
```
heplesser, line 1598: Our `Parameter` class does not provide information on the possible range of values, so this is not feasible and I'd remove the comment. Have you tried what happens if you supply a negative value?
```suggestion
```

nestkernel/conn_builder.cpp
```diff
@@ -1585,6 +1585,130 @@ nest::BernoulliBuilder::inner_connect_( const int tid, RngPtr rng, Node* target,
...
+      if ( loop_over_targets_() )
+      {
+        NodeCollection::const_iterator target_it = targets_->begin();
+        for ( ; target_it < targets_->end(); ++target_it )
+        {
+          const size_t tnode_id = ( *target_it ).node_id;
+          Node* const target = kernel().node_manager.get_node_or_proxy( tnode_id, tid );
+          if ( target->is_proxy() )
+          {
+            // skip array parameters handled in other virtual processes
```
heplesser, line 1637: Array parameters are only sensible for rules with a predetermined number of connections (see https://nest-simulator.readthedocs.io/en/stable/synapses/connection_management.html#array-parameters), so I think this makes no sense/is not needed. But make sure there is a check somewhere that stops users from providing array parameters.

nestkernel/conn_builder.cpp
```diff
@@ -1585,6 +1585,130 @@ nest::BernoulliBuilder::inner_connect_( const int tid, RngPtr rng, Node* target,
...
+    num_conns = rng->prand( pairwise_avg_num_conns_->value( rng, target ) );
+
+    if ( num_conns == 0 )
+    {
+      continue;
+    }
+    for ( unsigned long conn_counter = 1; conn_counter <= num_conns; ++conn_counter )
+    {
+      single_connect_( snode_id, *target, target_thread, rng );
+    }
```
heplesser, line 1708: This can be done simpler:
```suggestion
    for ( size_t n= 0 ; n < num_conns ; ++n )
    {
      single_connect_( snode_id, *target, target_thread, rng );
    }
```

heplesser (changes requested):
@ackurth Sorry for the long wait, but here are some comments for minor fixes.

doc/htmldoc/networks/spatially_structured_networks.rst
```diff
@@ -471,6 +471,8 @@ Pool
    +================================+==============+==============+
    | ``rule='pairwise_bernoulli'``  | source layer | target layer |
    +--------------------------------+--------------+--------------+
+   | ``rule='pairwise_poisson'``    | source layer | target layer |
+   +--------------------------------+--------------+--------------+
```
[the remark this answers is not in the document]
ackurth: Adressed in 546a01f5caed8443befa238db85d9a3fc0ea4fe6.

doc/htmldoc/networks/spatially_structured_networks.rst
```diff
@@ -832,10 +841,14 @@ Probabilistic connection rules
...
-``pairwise_bernoulli`` connection rule. The probability can then be a constant,
-depend on the position of the source or the target neuron, or on the
-distance between a driver and a pool node to a connection probability. To
-create dependencies on neuron positions, NEST parameters objects are used.
+``pairwise_bernoulli`` and the ``pairwise_poisson`` connection rule.
+For the ``pairwise_bernoulli`` rule, the probability can then be a
+constant, depend on the position of the source or the target neuron,
+or on the distance between a driver and a pool node to a connection probability.
+For the ``pairwise_poisson`` rule, the pairwise average number of connections
+again can depend on the position of the source or the target neuron,
```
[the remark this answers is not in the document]
ackurth: Addressed in 85d88b716ae483053fbaeac0e12af22ffe5eb55f.

nestkernel/random_generators.h
ackurth: Addressed in 29f6d751013b363b90ed241b363c7205d37cfb60.

nestkernel/random_generators.h
ackurth: Addressed in 29f6d751013b363b90ed241b363c7205d37cfb60.

nestkernel/random_generators.h
ackurth: Addressed in 29f6d751013b363b90ed241b363c7205d37cfb60.

nestkernel/connection_creator_impl.h
ackurth: outdated by answer to other comment

nestkernel/connection_creator_impl.h
ackurth: Addressed in 9af0d30ec25d3d223a5283a4c638e54b25fbaebc.

Regarding ```without_kernel```: I am not sure you can move it outside, at least not without breaking the current logic, c.f [here](https://github.com/nest/nest-simulator/blob/12cff3d370f503130142e575b3e1ea202a92ce49/nestkernel/connection_creator_impl.h#L97): if ```rng->drand() < kernel_->value( rng, source_pos, target_pos, source, tgt_ptr )``` evaluates to ```false``` but ```without_kernel``` to true a connection is established.
This realization also lead to a bug fix of the current code to make it consistent with the cod  [here](https://github.com/nest/nest-simulator/blob/12cff3d370f503130142e575b3e1ea202a92ce49).
But what this case should catch is unclear to me - maybe I don't understand it and thus don't understand your comment.

nestkernel/connection_creator_impl.h
ackurth: Addressed in 1920cdc163f953f6bb4e44f980bb1dcfd62e507d.

nestkernel/connection_creator_impl.h
ackurth: After reconsidering:
The ```without_kernel``` is not needed at this point, nor does it need to be checked at this level.
An error is thrown immediately if the required parameter ```pairwise_avg_num_conns``` is not provided.

nestkernel/connection_creator_impl.h
```diff
@@ -94,7 +98,55 @@ ConnectionCreator::connect_to_target_( Iterator from,
...
+  const Layer< D >& source )
+{
+  RngPtr rng = get_vp_specific_rng( tgt_thread );
+
+  // We create a source pos vector here that can be updated with the
+  // source position. This is done to avoid creating and destroying
+  // unnecessarily many vectors.
+  std::vector< double > source_pos( D );
+  const std::vector< double > target_pos = tgt_pos.get_vector();
+
```
heplesser, line 133:
```suggestion
```

nestkernel/connection_creator_impl.h
```diff
@@ -94,7 +98,55 @@ ConnectionCreator::connect_to_target_( Iterator from,
...
+
+  poisson_distribution poi_dist;
+
+
```
heplesser, line 137:
```suggestion
```

testsuite/pytests/test_spatial/test_connect_layers.py
```diff
@@ -237,6 +278,16 @@ def test_connect_layers_bernoulli_mask(self):
...
+    def test_connect_layers_poisson_mask(self):
+        """Connecting layers with pairwise_poisson and mask"""
+        conn_spec = {
+            "rule": "pairwise_poisson",
+            "pairwise_avg_num_conns": 0.5,
+            "mask": {"rectangular": {"lower_left": [-5.0, -5.0], "upper_right": [0.1, 0.1]}},
+        }
+        conns = self._check_connections(conn_spec, None, return_conns=True)
+        np.testing.assert_allclose(54, len(conns), atol=5)
+
```
heplesser, line 290: This test generates exactly the same connectivity pattern as `test_connect_layers_poisson_kernel_mask()` below, just that the test below applies a more stringent test. I suggest to remove this test therefore.
```suggestion
```

heplesser (changes requested):
Hi @ackurth I have a few more suggestions, please see inline.

doc/htmldoc/synapses/connection_management.rst
```diff
@@ -256,6 +256,26 @@ must be ``True``.
                       'allow_autapses': False, 'make_symmetric': True}
     nest.Connect(A, B, conn_spec_dict)
 
+pairwise poisson
+~~~~~~~~~~~~
```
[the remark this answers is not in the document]
terhorstd: @jessica-mitchell, shouldn't this be caught by some pydocstyle or so?

nestkernel/conn_builder.h
```diff
@@ -515,6 +527,8 @@ class TripartiteBernoulliWithPoolBuilder : public ConnBuilder
   void connect_() override;
 
 private:
+  void inner_connect_( const int, RngPtr, Node*, size_t );
+  ParameterDatum pairwise_avg_num_conns_; //!< Mean number of connections
```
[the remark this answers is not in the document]
ackurth: Oh sorry, merge error.

ackurth:
I addressed [this](https://github.com/nest/nest-simulator/pull/2960#discussion_r1425124668) in 995f22d .

heplesser (approved):
@ackurth Thanks for the changes, this looks good to me now!

PR #2968, opened 2023-09-29
Allow `isort` pre-commit to make in-place changes

nicolossus:
This PR removes the `--diff` flag from the `isort` pre-commit hook, thus allowing `isort` to make in-place changes instead of just showing the changes `isort` would make.


heplesser (approved):

heplesser:
Minimal flag to isort pre-commit hook, thus merging with single review.

PR #2970, opened 2023-10-09
Fix Pylint errors caused by new, stricter Pylint version

nicolossus:
Currently the Pylint checker fails due to a new, stricter version. This PR fix the errors Pylint points out.


heplesser (changes requested):
Hi @nicolossus! We cannot unconditionally import matplotlib. Tests should run on systems without it. Therefore, I created #2972 as a replacement. Can you close this one?

PR #2974, opened 2023-10-12
Add EBRAINS badges to README

terhorstd:
This adds badges from the EBRAINS badge service to the top level README.


jessica-mitchell:
@terhorstd  would linking to https://www.ebrains.eu/tools/nest this page be optimal?

heplesser:
@terhorstd I wonder about two things here:
- https://badge.apps-dev.hbp.eu/badge/nest/current shows 23.06 with NEST 3.5, why not 23.09 with NEST 3.6?
- On https://www.ebrains.eu/tools/nest the "Launch Online" button takes one to https://www.nest-simulator.org/download/, which isn't really online launching in my view. Could one link to some welcoming example in a collab?

jessica-mitchell:
Deployment has to wait for migration of EBRAINS services to new hosting compute centres

terhorstd:
Badge service is currently under discussion to become an actual production service, so we can keep this as "Blocked" for now. Probably the URL will change soon-ish.

PR #2975, opened 2023-10-13
Add plantuml to docs

jessica-mitchell:
This PR adds plantuml diagram possibilities to the documentation.

plantuml requires Java so this is installed during build on Read the Docs- This assumes we don't want to add Java as a dependency in our sphinx-build
For local builds, instructions are added to the User documentation workflow on how you can get plantuml diagrams to work.
In addition the requirements for the docs are no longer pinned to a specific version (rather changed to `>=`), along with version 3.8 of Python being updated to 3.10 in the readthedocs.yml. This didnt seem to break anything and I believe is the more desired way of handling requirements.
 There is a [temp file showing the output.](https://nest-simulator--2975.org.readthedocs.build/en/2975/testuml.html) An actual useful example would be better, so setting as draft for now.


terhorstd (approved):
Looks good for now. :+1:

jessica-mitchell:
Merging with 1 approval, as its docs

PR #2980, opened 2023-10-31
Minor cleanup of code and comments

jougs:
This is stuff that was lying in my cleanup branch and I don't want to be lost. See the list of commits for what and why.

### DIFF
--- a/nestkernel/logging_manager.h
+++ b/nestkernel/logging_manager.h
@@ -118,8 +118,7 @@ private:
 private:
   std::vector< deliver_logging_event_ptr > client_callbacks_;
   nest::severity_t logging_level_;
-  bool dict_miss_is_error_; //!< whether to throw exception on missed dictionary
-                            //!< entries
+  bool dict_miss_is_error_; //!< whether to throw exception on missed dictionary entries
 };
 
 } // namespace nest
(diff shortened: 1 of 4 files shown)


heplesser (changes requested):
Hi @jougs! Thanks for these fixes. I agree with most of them, and also with moving `register_*` calls out of `nestmodule.cpp`. But I think the `*_manager`s are not the right place for them: The managers take care of NEST parts, but they do not define which parts NEST has. With the new `modelsmodule`, this definition of what is part of NEST wrt neurons and synapses has become nicely separated, and ideally conn-builders and growth rules should be handled correspondingly. I see three options in no particular order and enumerated only for ease of reference:

1. Leave the `register_` calls where they are right now until we find a good place for them.
2. Call `register_conn_` from `conn_builder.cpp` and `register_growth_` from `growth_curve.cpp` (provide registers components).
3. Add the `register_{conn,growth}` calls to the auto-generated `modelsmodule.cpp` and maybe call that `register_nest_components.cpp` then.

heplesser:
@jougs If you could reset the `connection_manager.cpp`, `nestmodule.cpp` and `sp_manager.cpp` in this PR, I'd be happy to approve it :).

jougs:
Done.

heplesser (approved):
@jougs Thanks!

heplesser:
Minimal code tidying, thus merging based on single review.

PR #2981, opened 2023-11-04
Update clang-format version to 17.0.4 and activate `InsertBraces`

heplesser:
This PR hikes the clang-format version to 17.0.4, the current version at this point, which is available via PyPi for all relevant OS (https://pypi.org/project/clang-format/#files). It also activates automatic insertion of braces around blocks (https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces); this feature became available with clang-format 15.

Code format changes due to  the transition from 13 to 17 are minimal, see below.

Also fixes a forgotten requirements hike for cmake.

@terhorstd  @jessica-mitchell I'd appreciate it if you could review this quickly as #2867 would benefit from it.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -76,6 +76,7 @@ IndentCaseLabels: false
 IndentPPDirectives: None
 IndentWidth:     2
 IndentWrappedFunctionNames: false
+InsertBraces: true
 JavaScriptQuotes: Leave
 JavaScriptWrapImports: true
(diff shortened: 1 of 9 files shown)


jessica-mitchell (approved):
From my side it looks good

terhorstd (approved):
Together with some discussion this looks good to me. :+1:

PR #2983, opened 2023-11-07
Fix guards

heplesser:
@akorgor @JesusEV This PR brings back checks agains empty histories that are required (reverses part of 6ccefa076421d7). 

With this in place, all tests should pass and even for these "internal" PRs, we should only merge if all is green.


heplesser:
Accidentally created, should not go here.

PR #2984, opened 2023-11-10
Improvements to examples

heplesser:
@akorgor With the example from the pynest/examples directory, everything works :). This PR makes the examples more robust and informative. First, the `IPython` import and the `Image` call are protected with a `try`, which is essential for automated running of the examples. Second, I changed the verification test to also use `np.allclose()`, to print out information on failure, and to call `exit(1)` in that case. For automated checking of examples, it is important to return a non-zero exit code if the example failed.

In `np.allclose()`, I used `rtol`. I suggest you do that in the pytests as well. Given that floating-point numerics always work on 53 bits multiplied with some exponent, relative error is the most sensible quantity to look for, except when the expected value is zero, when on needs to use `atol`.


heplesser:
PR against wrong repo.

PR #2985, opened 2023-11-12
port test_iaf_psc_exp to py

janskaar:


heplesser (approved):

heplesser (changes requested):
The test is fine, but it should go to `pytest/sli2py_models`

heplesser (approved):

PR #2986, opened 2023-11-14
bug fix iaf_psc_exp_multisynapse / sli2py test_iaf_psc_exp_multisynapse

janskaar:
Before, S_.current_ was computed _before_ the updating individual synaptic ports, causing them to be 1 time step behind. Additionally, it was only computed when the neuron was not refractory, while the individual synaptic ports were updated every time step. Moving the S_.current_ computation down a few lines fixes this.

Port test_iaf_psc_exp_multisynapse.sli to py


testsuite/pytests/test_iaf_psc_exp_multisynapse.py
```diff
@@ -0,0 +1,164 @@
...
+
+    for i, syn_id in enumerate(range(1, 5)):
+        syn_spec = {
+            "synapse_model": "static_synapse",
+            "delay": delays[i],
+            "weight": weights[i],
+            "receptor_type": syn_id,
+        }
+
+        nest.Connect(sg, nrn, conn_spec="one_to_one", syn_spec=syn_spec)
```
heplesser, line 105: A more elegant solution:

```suggestion
    for syn_idx, (delay, weight) in enumerate(zip(delays, weights)):
        syn_spec = {
            "synapse_model": "static_synapse",
            "delay": delay,
            "weight": weight,
            "receptor_type": syn_idx + 1,
        }

        nest.Connect(sg, nrn, conn_spec="one_to_one", syn_spec=syn_spec)
```

testsuite/pytests/test_iaf_psc_exp_multisynapse.py
```diff
@@ -0,0 +1,164 @@
...
+
+    mm = nest.Create(
+        "multimeter",
+        params={"record_from": ["I_syn_1", "I_syn_2", "I_syn_3", "I_syn_4", "V_m", "I_syn"], "interval": dt},
+    )
+
+    nest.Connect(mm, nrn)
+    nest.Simulate(simtime)
+    times = mm.get("events", "times")
+    I_syn = np.sum([mm.get("events", f"I_syn_{i}") for i in range(1, 5)], axis=0)
```
heplesser, line 115: This weakens the test and explains why the wrong `current_` was not detected by the tests. We should test the `I_syn` recordable here (not sure if it needs and extra `[0]` at the end):
```suggestion
    I_syn = mm.get("events", "I_syn")
```

testsuite/pytests/test_iaf_psc_exp_multisynapse.py
```diff
@@ -0,0 +1,164 @@
...
+    V_m_analytical = np.zeros_like(times, dtype=np.float64)
+    for i in range(4):
+        I_syns_analytical.append(exp_psc_fn(times - delays[i] - spike_time, tau_syn[i]) * weights[i])
+        V_m_analytical += exp_psc_voltage_response(times - delays[i] - spike_time, tau_syn[i], tau_m, C_m, weights[i])
+
+    nptest.assert_array_almost_equal(mm.get("events", "I_syn"), np.sum(I_syns_analytical, axis=0))
+    nptest.assert_array_almost_equal(mm.get("events", "I_syn_1"), I_syns_analytical[0])
+    nptest.assert_array_almost_equal(mm.get("events", "I_syn_2"), I_syns_analytical[1])
+    nptest.assert_array_almost_equal(mm.get("events", "I_syn_3"), I_syns_analytical[2])
+    nptest.assert_array_almost_equal(mm.get("events", "I_syn_4"), I_syns_analytical[3])
```
heplesser, line 127:
```suggestion
    for idx, I_syn_analytical in enumerate(I_syns_analytical):
        nptest.assert_array_almost_equal(mm.get("events", f"I_syn_{idx+1}"), I_syn_analytical)
```

heplesser (changes requested):
@janskaar Good catch! I have an alternative solution for summing the currents and some suggestions for the tests. The same problem applies to `iaf_psc_alpha_multisynapse`.  I think it would be a good idea to fix both neurons within this PR.

heplesser (changes requested):
The new test should got to `pytest/sli2py_models`.

testsuite/pytests/test_iaf_psc_exp_multisynapse.py
```diff
@@ -0,0 +1,164 @@
...
+@pytest.fixture(autouse=True)
+def reset():
+    nest.ResetKernel()
+
+
+def exp_psc_fn(t, tau_syn):
+    vals = np.zeros_like(t)
+    zero_inds = t <= 0.0
+    nonzero_inds = ~zero_inds
+    vals[nonzero_inds] = np.exp(-t[nonzero_inds] / tau_syn)
```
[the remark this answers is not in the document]
janskaar: Changed this to a slighly different solution

heplesser (changes requested):
@janskaar This looks fine now, but there seem to be some problems with formatting. Could you fix those? And how about a corresponding test for alpha multisynapse? Is that in a separate PR?

janskaar:
> @janskaar This looks fine now, but there seem to be some problems with formatting. Could you fix those? And how about a corresponding test for alpha multisynapse? Is that in a separate PR?

Formatting should be fixed now. The alpha multisynapse test was merged a couple of weeks ago, but I have updated it to be more in line with this one now.

janskaar:
@heplesser Sorry, just noticed that test_mini_brunel_ps.sli fails, but only on MacOS. Have you encountered something like this before? It's for a completely different neuron model, so it's difficult to see why it should fail here.

heplesser:
@otcathatsya Ping!

otcathatsya (approved):
Looks good! Do we have a preference for which approximate comparison method to use? We have quite a few `pytest.approx` and I just saw that numpy's `assert_almost_equal` works with absolute decimal places instead, though in this case they'd work interchangeably too.

heplesser:
> Looks good! Do we have a preference for which approximate comparison method to use? We have quite a few `pytest.approx` and I just saw that numpy's `assert_almost_equal` works with absolute decimal places instead, though in this case they'd work interchangeably too.

I don't think we have made a definite decision. I am not certain how pytest.approx works on NumPy arrays, because `==` between arrays returns an array of bools, not a single truth value.

PR #2988, opened 2023-11-14
Fix misplaced `assert` statement causing a lot of warnings

muffgaga:
I just saw this warning quite often (in the EBRAINS spack build it gets repeated for 162 times ;) ):
```
/tmp/spack-stage/spack-stage-nest-3.6-uy4os6lc7odhv7bryyp2atuprbdl2xwn/spack-src/nestkernel/connection.h: In member function 'nest::SecondaryEvent* nest::Connection<targetidentifierT>::get_secondary_event()':
/tmp/spack-stage/spack-stage-nest-3.6-uy4os6lc7odhv7bryyp2atuprbdl2xwn/spack-src/nestkernel/connection.h:404:1: warning: no return statement in function returning non-void [-Wreturn-type]
  404 | }
      | ^
```

Also: I guess even if the `assert` is ensured to never be removed (contrary to standard `NDEBUG` behavior) it would be reasonable to clearly state the intention here (an exception might be more user-friendly, but I guess the code was intentionally avoiding an exception) by calling abort?

### DIFF
--- a/nestkernel/connection.h
+++ b/nestkernel/connection.h
@@ -402,7 +402,8 @@ template < typename targetidentifierT >
 SecondaryEvent*
 Connection< targetidentifierT >::get_secondary_event()
 {
-  assert( false );
+  assert( false && "non-primary connections have to provide get_secondary_event()");
+  return nullptr;
 }
 
 } // namespace nest
--- a/nestkernel/connection.h
+++ b/nestkernel/connection.h
@@ -402,7 +402,7 @@ template < typename targetidentifierT >
 SecondaryEvent*
 Connection< targetidentifierT >::get_secondary_event()
 {
-  assert( false && "non-primary connections have to provide get_secondary_event()");
+  assert( false and "Non-primary connections have to provide get_secondary_event()" );
   return nullptr;
 }
 


nestkernel/connection.h
```diff
@@ -401,6 +403,7 @@ SecondaryEvent*
 Connection< targetidentifierT >::get_secondary_event()
 {
   assert( false );
+  std::abort();
```
terhorstd, line 406: The assertion there is indeed plain wrong. So we can also stick to only have `std::abort()` there.
```suggestion
  std::abort();
```

terhorstd (changes requested):
minor change

gtrensch (approved):
Approved.

nestkernel/connection.h
heplesser: I think that the assertion makes more sense here than abort: this method should simply never been called in the base class. And the `assert()` can be compiled out with `-DNDEBUG`.

nestkernel/connection.h
```diff
@@ -134,7 +136,7 @@ class Connection
    * To prevent erronous calls of this function on primary connections, the base class implementation
    * below just contains `assert(false)`.
    */
-  SecondaryEvent* get_secondary_event();
+  [[noreturn]] SecondaryEvent* get_secondary_event();
```
heplesser, line 139: The method is overridden in derived classes implementing `SecondaryEvent`s, and there it returns a pointer. I assume there is no problem with having the `[[noreturn]]` attribute in the base class but not in the derived classes?

heplesser (changes requested):
@muffgaga Thanks for this fix! Which compiler and which compiler options do you use? I have not seen this warning before, so our warning settings are apparently not strict enough. Also, even though the limited "Schaffungshöhe" of this PR does not strictly make it necessary, would you send us a contributor agreement (https://nest-simulator.readthedocs.io/en/latest/_downloads/9b65adbdacba6bfed66e68c62af4e308/NEST_Contributor_Agreement.pdf)?

nestkernel/connection.h
```diff
@@ -400,7 +402,7 @@ template < typename targetidentifierT >
 SecondaryEvent*
 Connection< targetidentifierT >::get_secondary_event()
 {
-  assert( false );
+  std::abort();
```
heplesser, line 405: I still don't quite understand why `abort()` is better than a failing assertion.

muffgaga: Yes, `assert(false)` would call `std::abort()`, but as it is a macro which is conditionally removed by the define `NDEBUG` (according to the standard, see https://en.cppreference.com/w/cpp/error/assert) it's not safe (nest might have some non-standard behavior w.r.t. NDEBUG/assert, so I'm not sure if this is the case here).

nestkernel/connection.h
muffgaga: I don't think this attribute is inherited by derived classes (I did try to look at the standard documents that I have access to). Looking at https://en.cppreference.com/w/cpp/language/attributes/noreturn I would expect that this only affects the "one" function where it was declared.

muffgaga:
> @muffgaga Thanks for this fix! Which compiler and which compiler options do you use? I have not seen this warning before, so our warning settings are apparently not strict enough.

It was the EBRAINS collab one — i.e. gcc@10.3.0, and "Release" build was mentioned in the build log (see #2993 for a build issue using the same compiler/software environment).

> Also, even though the limited "Schaffungshöhe" of this PR does not strictly make it necessary, would you send us a contributor agreement (https://nest-simulator.readthedocs.io/en/latest/_downloads/9b65adbdacba6bfed66e68c62af4e308/NEST_Contributor_Agreement.pdf)?

Yes, will do.

muffgaga:
> > NEST_Contributor_Agreement
>
> Yes, will do.

→ sent to @terhorstd (he asked me directly via chat).

nestkernel/connection.h
terhorstd: if `abort()` is what should be done here, then we need to explicitly call it here. Especially the production builds (apt, spack, …) all have the build-type "rel" (release) which removes debugging stuff. So the assert resolves to `;` and does not cause nest to `abort()` as is intended.

muffgaga: Maybe I'm missing something, but why provide this function at all? There's no dynamic casting going on, as far as I could see. I.e. just remove the function and there would be a build-time error if someone wrongly calls this (as a base class).
Maybe it is about the intend to communicate the idea of an "API" → you could still leave out the implementation, and only a link-time error would occur. Or, if modern C++ is okay,  you could just redefine `Connection` to be a concept that the "derived" Connection classes would have to fulfill.

```
template<typename C>
concept ConnectionLike = requires(C const src, C dst)
{
    C{};
    C{src};
    dst = src;
    
    // a non-const get_secondary_event member function
    dst.get_secondary_event();
};
```

heplesser: @muffgaga Yes, we could use constexpr-evaluation here. But the entire "secondary event" implementation needs a thorough overhaul, so I would not like to spend any significant effort on code improvements in this area now. But I like your suggestion to return a nullptr after the assert (and remove the ``[[noreturn]]``). But then I would really like to wrap up this PR—it is a tiny fix to get to a well-defined state which avoids many compiler warnings and that should not be delayed by possible larger improvements. One can always create a follow-up issue ...

muffgaga: Ok, got rid of the other things, and only added a `return nullptr;` (+ a small helpful msg to the assert).

heplesser:
@muffgaga Thanks a lot, will merge now!

heplesser (approved):

heplesser:
@terhorstd Since you had previously requested changes to this PR, you need to approve it now :).

terhorstd (approved):
LGTM :+1:
Thanks!

PR #2990, opened 2023-11-15
Models no longer wrapped in SLI Module, model registration and thread-setting in managers revised

heplesser:
The main purpose of this PR (developed by @jougs) is to register node and connection models directly in the `ModelManager` without wrapping them in a `SLIModule`.  This change will allow implementation of extension module unloading at a later stage.

As a side effect of this change, this PR also:

1. Creates one connection model instance per thread instead of storing the same pointer to a single model for each thread; thus common properties of connections can now be stored locally per thread.
2. Changes the way managers are adjusted to a changed number of threads. In the past, the new number of threads was first set and then all managers finalized and initialized anew. Thus finalization of data structures with the old thread number was done with the new thread number already set, which can lead to problems. Now, all managers are first finalized (in opposite order of initialization), then the new thread number is set and then all managers are initialized. Both `initialize()` and `finalize()` now have a `reset_kernel` argument (default `true`) which controls if everything should be reset or only data structures adjusted to the new thread numbers. In the past, we actually reset more than we intended when just setting a new number of threads.


heplesser:
@terhorstd @mlober Ping!

nestkernel/model_manager.h
mlober, line 1: Typo in line 183: This function deleteS

nestkernel/node_manager.cpp
mlober, line 1: Typo in line 181: ... must be called ...

mlober (approved):
The code compiles and runs nicely on my laptop. From what I can gather, the code and added documentation both look good to me (except for small typos - see comments).

nestkernel/nest.h
```diff
@@ -53,9 +53,6 @@ void enable_dryrun_mode( const size_t n_procs );
 
 void register_logger_client( const deliver_logging_event_ptr client_callback );
 
-/**
- * Register connection model (i.e. an instance of a class inheriting from `Connection`).
- */
```
terhorstd, line 58: Why did this vanish?

testsuite/regressiontests/ticket-478.sli
```diff
@@ -84,7 +85,7 @@ ResetKernel
   { /st Set
     % create new nodes for each connection to avoid any issues with prohibited multiple connections
     ds_models
-    { Create /iaf_psc_alpha Create st Connect } forall
+    { Create /iaf_psc_alpha Create /one_to_one st Connect } forall
```
terhorstd, line 88: Why is this change necessary? Does this mean a change in `Connect` behavior?

nestkernel/nest.h
heplesser: Re-instated.

heplesser:
@terhorstd I have now committed the changes discussed in person today and addressed the two comments above. I have also merged in master.

terhorstd (approved):
This looks OK from my perspective. There is obvious need for refactoring in some places, but that should not clutter this PR.
:+1: 
Big thanks to everyone involved!

PR #2996, opened 2023-11-20
Add pre-commit to NEST environment.yml and limit pylint to pre-3.0

heplesser:
This PR 

1. adds `pre-commit` to a new "Contributing to NEST" section of `environment.yml`
2. limits the `pylint` version installed to <3. This is need to run `pytest`, which invokes `pylint`, due to some recent changes in `pylint` (see https://github.com/pylint-dev/pylint/issues/8862).

Probably #2834 needs adjusting as a consequence.

Tagging as high priority as it may brake testing.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -74,7 +74,7 @@ dependencies:
   - jupyter
   - jupyterlab
 
-  # Contributin to NEST -----
+  # Contributing to NEST -----
   - pre-commit
 
   # Building NEST documentation
--- a/environment.yml
+++ b/environment.yml
@@ -47,6 +47,8 @@ dependencies:
 
   # Testing NEST ------------
   - attrs
+  # Pytest currently incompatible with pylint 3, see https://github.com/pylint-dev/pylint/issues/8862
+  - pylint < 3
   - pytest
   - pytest-timeout
   - pytest-xdist
--- a/environment.yml
+++ b/environment.yml
@@ -74,6 +74,9 @@ dependencies:
   - jupyter
   - jupyterlab
 
+  # Contributin to NEST -----
+  - pre-commit
+
   # Building NEST documentation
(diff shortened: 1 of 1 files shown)


nicolossus:
@heplesser With `pylint 3.02`, I am able to locally run both the `pylint` check and the `pytest` suite from the CI without problems. Do you have an example of where `pylint > 3` causes problems? 

From looking at the issue you linked to, it doesn't seem like requiring `pylint < 3` should be necessary as long as we don't depend on a package that still use the now deprecated `pylint.config.PYLINTRC`.

heplesser:
@nicolossus This is strange. If I create an environment with just
```
mamba create -n t312pt "python>=3.12" pylint pytest 
```
then `pytest --version` works. When I create an entire environment using the `environment.yml`, then `pytest --version` triggers an error.

nicolossus:
@heplesser Thanks for the info, I figured out where the problem lies. With the `environment.yml` from current master, the following versions are installed:

```
pylint 3.0.2
pytest 7.4.3
pytest-pylint 0.18.0
```

`pytest-pylint` added support for the new PyLint RC discovery in version `0.19.0` (see [release notes](https://pypi.org/project/pytest-pylint/)).

The reason it worked for me with my `nest` environment is that mine is built from https://github.com/nest/nest-simulator/pull/2834, which installs the following versions:

```
pylint 3.0.2
pytest 7.4.3
pytest-pylint 0.21.0
```

https://github.com/nest/nest-simulator/pull/2834 now has two approvals and `pre-commit` is already included. I suggest that we close this PR and merge https://github.com/nest/nest-simulator/pull/2834 instead.

heplesser:
@nicolossus Thanks for you detective work. I will close this in favour of #2834.

PR #2997, opened 2023-11-20
Replace assertDictContainsSubset() which has been removed in Python 3.12

heplesser:
See https://stackoverflow.com/questions/20050913/python-unittests-assertdictcontainssubset-recommended-alternative for the solution I used.

High priority because it breaks testsuite in Python 3.12.

### DIFF
--- a/testsuite/pytests/test_sp/test_synaptic_elements.py
+++ b/testsuite/pytests/test_sp/test_synaptic_elements.py
@@ -22,6 +22,7 @@
 import unittest
 
 import nest
+from testutil import dict_is_subset_of
 
 __author__ = "naveau"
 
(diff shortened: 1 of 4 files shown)


otcathatsya (approved):

testsuite/pytests/test_sp/test_synaptic_elements.py
```diff
@@ -37,7 +37,7 @@ def test_set_status(self):
         nest.SetStatus(neuron, {"synaptic_elements": synaptic_element_dict})
         neuron_synaptic_elements = nest.GetStatus(neuron, "synaptic_elements")[0]
         self.assertIn("SE", neuron_synaptic_elements)
-        self.assertDictContainsSubset(synaptic_element_dict["SE"], neuron_synaptic_elements["SE"])
+        self.assertEqual(neuron_synaptic_elements["SE"], synaptic_element_dict["SE"] | neuron_synaptic_elements["SE"])
```
janskaar, line 40: I believe the positions of the two dictionaries in the last argument should be switched. 

For two dictionaries a, b which share a subset of keys, it seems `c = a | b` for any keys present in both a and b, will be given the value from b, not a. 

In this case, if e.g.
`neuron_synaptic_elements["SE"]["z"]` != `synaptic_element_dict["SE"]["z"]`, 
the test will not fail, since
`neuron_synaptic_elements["SE"]["z"]` == `(synaptic_element_dict["SE"] | neuron_synaptic_elements["SE"])["z"]`.

Same would then be the case for the other tests.

janskaar (changes requested):

testsuite/pytests/test_sp/test_synaptic_elements.py
heplesser: Thanks for noticing this! I also noted now that the dict merge operator was only introduced with Python 3.9, while we still want to support 3.8. I will look for a better solution.

heplesser:
@janskaar I have now fixed the PR, putting the dictionaries in proper order and making it a function with its own test. Since #2834 hiked Python requirements to 3.10, using the dict merge operator is no problem. I removed one test for getverbosity—no need to explicitly check it returns and int value. The remaining test will fail if it doesn't.

janskaar (approved):
Looks good

PR #2936, opened 2023-09-12
Add `clang-format` pre-commit hook

nicolossus:
This PR introduces a pre-commit hook for `clang-format`. By default, the hook uses the `.clang-format` file from the root. 

The PR also adds the `end-of-file-fixer` and `trailing-whitespace` hooks. They are out-of-the-box hooks from `pre-commit`, which already is a dependency. These particular hooks might save a lot of back-and-forth with CI.

PR #3002, opened 2023-11-21
Add `shellcheck` and `pre-commit` to CI jobs

terhorstd:
This is an addendum to #2880 adding a `pre-commit` run also to the CI pre-build checks.
The `shellcheck` is also added separately to address language specific linters on the same level.

Related PRs are:
* #2936 
* #2968
* #2996


.github/workflows/nestbuildmatrix.yml
```diff
@@ -33,6 +33,44 @@ jobs:
...
+      - name: "Set up Python 3.x"
+        uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4.7.0
+        with:
+          python-version: 3.9
+
+      - name: "Install dependencies"
+        run: |
+          pip install pre-commit
+
+      - name: "Run shellcheck..."
```
heplesser, line 70:
```suggestion
      - name: "Run pre-commit ..."
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -33,6 +33,44 @@ jobs:
...
+      - name: "Install Linux system dependencies"
+        run: |
+          sudo apt-get update
+          sudo apt-get install shellcheck
+
+      - name: "Run shellcheck..."
+        run: |
+          find . -iname "*.sh" | xargs shellcheck
+
+  pre-commit-hooks:
```
heplesser, line 53: Check if this has already been integrated.

heplesser (changes requested):
See comments and suggestions.

.github/workflows/nestbuildmatrix.yml
terhorstd: This is now fixed with the correct `find . -iname "*.sh" -o -iname "*.sh.in" | …`. Have a look also into `.pre-commit-config.yaml` where currently a hand-written list needs to be maintained.

heplesser: @terhorstd Given that we also run the pre-commit tests just below, does this not lead to running shellcheck twice in the test process?

.github/workflows/nestbuildmatrix.yml
```diff
@@ -33,6 +33,62 @@ jobs:
...
+        run: |
+          pre-commit run --all-files
+
+  no-crlf:
+    # While pre-commit-hook mixed-line-endings could check for this, it would
+    # be incompatible with "git smudge", as users may decide to use any
+    # formatting in their editor consistently.
+    # https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#mixed-line-ending
+    #
+    # This check makes sure the repository is always consistent NL line endings.
```
heplesser, line 81: What does `NL` mean? There is also some grammatical problem in the sentence.

doc/htmldoc/static/img/EBRAINS.svg
heplesser, line 1: Why does git show that all SVG files here have been replaced by identical-looking files?

heplesser (commented):
@terhorstd This looks good, but see the few inline comments. Did you run the shell scripts that have been modified significantly and which are not run by the CI, e.g., run_micrcocircuit and run_examples?

heplesser:
I am exploring the test failure on macOS.

.github/workflows/nestbuildmatrix.yml
terhorstd: yes, and with potentially different settings. The idea behind was that pre-commit tests if the developer has run the checks, but we still have each test separately in the CI overview.

`clang-format` is another candidate and that basically makes sure that our clang versions don't differ between developer and CI machines.

Currently whole process is 49 seconds, could be much less if we add caching of the pre-commit environments.

doc/htmldoc/static/img/EBRAINS.svg
terhorstd: A single newline at the end was changed. Should I add an ignore for those or are we fine with this?

terhorstd:
The reason for the MacOS failure is described in https://stackoverflow.com/a/752893. And I think I found a workaround.

.github/workflows/nestbuildmatrix.yml
heplesser: OK, thanks for the explanation.

doc/htmldoc/static/img/EBRAINS.svg
heplesser: Fine now, since the files now are fixed leave them in.

.github/workflows/nestbuildmatrix.yml
```diff
@@ -78,7 +78,8 @@ jobs:
     # formatting in their editor consistently.
     # https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#mixed-line-ending
     #
-    # This check makes sure the repository is always consistent NL line endings.
+    # This check makes sure the files in the repository are always consistent
+    # in their line endings, i.e. new line, and not carrige return+line feed.
```
heplesser, line 82:
```suggestion
    # in their line endings, i.e. new line, and not carriage return+line feed.
```

.github/workflows/nestbuildmatrix.yml
heplesser: Thanks for the improved implementation!

terhorstd:
Hey, could you have another look? I think it's ready to go

build_support/install_sionlib.sh
```diff
@@ -2,15 +2,16 @@
 set -euo pipefail
 
 SIONLIBVERSION="1.7.4"
+SIONLIB_INSTALL_PATH="${1:-${HOME}/.cache/csa.install}"
```
heplesser: Are you sure the `csa.install` here is correct? Why not `sionlib.install`?

examples/run_examples.sh
```diff
@@ -31,23 +31,28 @@ if ! python3 -c "import nest" >/dev/null 2>&1; then
...
+
 declare -a EXAMPLES
 if [ "${#}" -eq 0 ]; then
     # Find all examples that have a line containing "autorun=true"
     # The examples can be found in subdirectory nest and in the
     # examples installation path.
     EXAMPLELIST="$(mktemp examplelist.XXXXXXXXXX)"
     if [ -d "nest/" ] ; then
-        grep -rl --include='*.sli' 'autorun=true' nest/ >>"$EXAMPLELIST"
+        grep -rl --include='*.sli' 'autorun=true' "${sourcedir}/nest/" >>"$EXAMPLELIST" || true
```
heplesser, line 51: What is the purpose of `|| true`?

examples/run_examples.sh
```diff
@@ -64,10 +69,34 @@ done
...
 export MPLCONFIGDIR
 
-time_format="  time: {real: %E, user: %U, system: %S}\n\
-  memory: {total: %K, max_rss: %M}"
+INFO_OS="$(uname -s)"
 
-basedir=$PWD
+# find "time" command
+if test "${INFO_OS:-}" = "Darwin"; then
+    if test -x "/usr/local/bin/gtime"; then
```
heplesser, line 76: For some reason, I cannot make a suggestion here, but the path should be `/opt/homebrew/bin/gtime` (also below). Homebrew installs everything under `/opt/homebrew` on Apple Silicon.

heplesser (changes requested):
Hi @terhorstd! I spotted one potential copy-paste-error (but I may have looked at an outdated file version), have one suggestion and one question.

examples/run_examples.sh
terhorstd: If grep does not find the string (i.e. there are not exmples in that directory) then the script exits due to the error code. `true` always returns a successful return-code.

examples/run_examples.sh
terhorstd: A nicer way to define this would be a `$(which gtime)` but I don't know if that is correctly found. Could you check?

build_support/install_sionlib.sh
terhorstd: Not shure how you see this, but it has been fixed in https://github.com/nest/nest-simulator/pull/3002/commits/6f937ebcf974557718fd253514e8d38ae25347cc, hence the "(outdated)" above.

examples/run_examples.sh
heplesser: Seems to work:
```
(nn) ~> echo $(which gtime)
/opt/homebrew/bin/gtime
```

terhorstd: Thanks, I added the `which`

heplesser (approved):

PR #3011, opened 2023-11-23
Sort results of NEST releases on Zenodo by most recent

jessica-mitchell:
Zenodo changed how sorting is done, so this PR updates the link so the first results are the most recent NEST releases


jessica-mitchell:
Merging without review as its a single line change in documentation - link tested and works

PR #3015, opened 2023-11-28
Add missing keyword to glif neuron model documentation

jessica-mitchell:
The glif neuron models are missing the keyword neuron, so they don't appear in the directory under that category.


jessica-mitchell:
Merging with no reviewer as its a minor documentation change

PR #3021, opened 2023-11-28
Ensure Connect() does not allow passing parameters that need to be set at the synapse-model level

heplesser:
Until now, users could pass parameters shared by all synapses of a given type to the `syn_spec` in `Connect()`. These parameters were silently ignored as illustrated by the example below, where `alpha` is not set to 10 either in the connection (which does not have that property) nor in the defaults.

The PR also ports the pertaining tests to Python, in part simplifying and in part extending them.

```
In [2]: n = nest.Create('iaf_psc_alpha')
In [4]: nest.Connect(n, n, syn_spec={'synapse_model': 'stdp_synapse_hom', 'alpha': 10})
In [5]: c = nest.GetConnections()
In [6]: c.get()
Out[6]: 
{'delay': 1.0,
 'Kplus': 0.0,
 'port': 0,
 'receptor': 0,
 'sizeof': 48,
 'source': 1,
 'synapse_id': 54,
 'synapse_model': 'stdp_synapse_hom',
 'target': 1,
 'target_thread': 0,
 'weight': 1.0}

In [7]: nest.GetDefaults('stdp_synapse_hom')
Out[7]: 
{'alpha': 1.0,
 'delay': 1.0,
 'element_type': 'synapse',
 'has_delay': True,
 'Kplus': 0.0,
 'lambda': 0.01,
 'mu_minus': 1.0,
 'mu_plus': 1.0,
 'num_connections': 1,
 'receptor_type': 0,
 'requires_symmetric': False,
 'sizeof': 48,
 'synapse_model': 'stdp_synapse_hom',
 'synapse_modelid': 54,
 'tau_plus': 20.0,
 'Wmax': 100.0,
 'weight': 1.0,
 'weight_recorder': NodeCollection(<empty>)}
```


akorgor (approved):
This PR prevents the scenario that users pass common synapse properties directly to the `syn_spec` in `Connect()` and don't notice that these are ignored by throwing an error suggesting to set the respective parameter via `SetDefaults` on the synpase model, which is an important fix.


The mechanism and the test work like a charm and the code is clear and well-documented.

nestkernel/connector_model_impl.h
```diff
@@ -137,6 +137,28 @@ GenericConnectorModel< ConnectionT >::set_status( const DictionaryDatum& d )
...
+  // This is called just once per Connect() call, so we need not worry much about performance.
+  // We get a dictionary with synapse default values and check if any of its keys are in syn_spec.
+  DictionaryDatum dummy( new Dictionary );
+  cp_.get_status( dummy );
+
+  for ( [[maybe_unused]] const auto& [ key, val ] : *syn_spec )
+  {
+    if ( dummy->known( key ) )
+    {
+      throw NotImplemented(
```
nicolossus, line 153: Same question as above regarding the use of `NotImplemented`.

testsuite/pytests/sli2py_connect/test_common_properties_setting.py
```diff
@@ -0,0 +1,135 @@
...
+def test_no_setting_on_connection(syn_model, specs):
+    """Test that common property cannot be set on individual connection."""
+
+    if specs["setup"]:
+        specs["setup"]()
+
+    n = nest.Create(specs["neuron"])
+    nest.Connect(n, n, syn_spec={"synapse_model": syn_model})
+    conn = nest.GetConnections()
+    with pytest.raises(nest.kernel.NESTError):
```
nicolossus, line 122: If the specific error is surfaced to the Python level, I think it is better to test whether the specific error is raised.

testsuite/pytests/sli2py_connect/test_common_properties_setting.py
```diff
@@ -0,0 +1,135 @@
...
+
+@pytest.mark.parametrize("syn_model, specs", available_cp_models.items())
+def test_no_setting_on_connect(syn_model, specs):
+    """Test that common property cannot be set in a Connect call"""
+
+    if specs["setup"]:
+        specs["setup"]()
+
+    n = nest.Create(specs["neuron"])
+    with pytest.raises(nest.kernel.NESTError):
```
nicolossus, line 134: Specific error?

testsuite/pytests/sli2py_connect/test_common_properties_setting.py
```diff
@@ -0,0 +1,135 @@
...
+    assert new_val == specs["value"]
+
+    # Set parameter back on copied model, original must not be changed
+    nest.SetDefaults(new_model, {specs["parameter"]: old_val})
+    assert nest.GetDefaults(syn_model)[specs["parameter"]] == specs["value"]
+
+
+@pytest.mark.parametrize("syn_model, specs", available_cp_models.items())
+def test_no_setting_on_connection(syn_model, specs):
+    """Test that common property cannot be set on individual connection."""
```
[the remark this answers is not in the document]
nicolossus: I would suggest to fix this the other way around, that is, add punctuation marks to the other docstrings.

nicolossus (changes requested):
@heplesser Thanks! This looks mostly good to me, I only have a few remarks/questions. See inline comments.

models/tsodyks_synapse_hom.h
```diff
@@ -251,7 +251,7 @@ class tsodyks_synapse_hom : public Connection< targetidentifierT >
   void
   set_weight( double )
   {
-    throw BadProperty(
+    throw NotImplemented(
```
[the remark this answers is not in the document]
heplesser: I decided to use `NotImplemented` here for consistency with the (now deleted) checks in the Jonke and STDP dopamine synapse (see above). I also think `NotImplemented` makes sense: The parameter is not wrong in the sense of having an out-of-range value—we just do not support changing common properties through individual synapses.

testsuite/pytests/sli2py_connect/test_common_properties_setting.py
heplesser: Fixed. The specific error here is DictError because the common parameter is not read in this case.

testsuite/pytests/sli2py_connect/test_common_properties_setting.py
heplesser: Fixed. Here we get the NotImplemented from the checking function.

nicolossus (approved):
LGTM!

PR #3027, opened 2023-12-01
Add compile-time checks w.r.t. `is_primary` or existence of `get_secondary_event()`

muffgaga:
That is not a polished commit, but rather just a draft showing that some more compile-time checking of the interfaces is possible without any restructuring. (Cf. discussion in PR #2988).

* By compile-time checking for the `IS_PRIMARY` property, we now can drop `Connection::get_secondary_event()`, which could only be reached by illegal code.
* It introduces a compile-time check on `not IS_PRIMARY` and the existence of the derived Connection's `get_secondary_event` — and it provides a compile-time error message in cases where this isn't the case.


nestkernel/connector_model.h
```diff
@@ -199,7 +219,19 @@ class GenericConnectorModel : public ConnectorModel
   SecondaryEvent*
   get_secondary_event() override
   {
-    return default_connection_.get_secondary_event();
+    constexpr bool is_primary = flag_is_set( ConnectionT::properties, nest::ConnectionModelProperties::IS_PRIMARY );
+    constexpr bool has_get_secondary_event = has_get_secondary_event_t< ConnectionT >::value;
+    static_assert(
+      is_primary xor has_get_secondary_event, "Non-primary connections have to provide get_secondary_event()" );
+    if constexpr ( ( not is_primary ) and has_get_secondary_event )
```
otcathatsya, line 226: should be sufficient to only check one of them after the assert

otcathatsya (approved):

nestkernel/connector_model.h
JanVogelsang: Should be sufficient, but also doesn't hurt to check for both at compile time, right?

nestkernel/connector_model.h
```diff
@@ -199,7 +219,19 @@ class GenericConnectorModel : public ConnectorModel
...
+    {
+      return default_connection_.get_secondary_event();
+    }
+    else
+    {
+      // unreachable code
+      return nullptr;
```
JanVogelsang, line 233:
```suggestion
    UnsupportedEvent( "Using secondary events on primary connections is not possible. Secondary connections must implement get_secondary_event!" );
```

JanVogelsang (changes requested):

heplesser:
@muffgaga Could you respond to the remarks in the code review? Also, could you merge master into your branch and fix the conflicts that have arisen since you created this PR?

nestkernel/connector_model.h
```diff
@@ -59,6 +59,26 @@ enum class ConnectionModelProperties : unsigned
   REQUIRES_CLOPATH_ARCHIVING = 1 << 6,
   REQUIRES_URBANCZIK_ARCHIVING = 1 << 7
 };
+}
+
+namespace
+{ // FIXME: utils namespace?
```
muffgaga: will move to util lib?

muffgaga: It's `ConnectorModel`-specific helper code, so I guess that's ok?

nestkernel/connector_model.h
muffgaga: (I am sorry for ignoring this for very long… Dennis pinged me today.)

I'd say "no change of behavior" compared to before — but now with the suggested change by @JanVogelsang from the other comment, there's this `UnsupportedEvent` instead of a `return nullptr`.

muffgaga:
I rebased it; it was almost conflict free (except for a conflict in the comment of get_secondary_event in the base class, which we remove in this PR anyways…)

nestkernel/connector_model.h
```diff
@@ -200,7 +220,19 @@ class GenericConnectorModel : public ConnectorModel
   std::unique_ptr< SecondaryEvent >
   get_secondary_event() override
   {
-    return default_connection_.get_secondary_event();
+    constexpr bool is_primary = flag_is_set( ConnectionT::properties, nest::ConnectionModelProperties::IS_PRIMARY );
+    constexpr bool has_get_secondary_event = has_get_secondary_event_t< ConnectionT >::value;
+    static_assert(
+      is_primary xor has_get_secondary_event, "Non-primary connections have to provide get_secondary_event()" );
+    if constexpr ( ( not is_primary ) and has_get_secondary_event )
```
muffgaga: If C++20 would be okay, we could simplify this and remove the helpers above, and just have:
```suggestion
    if constexpr ( ( not is_primary ) and requires { default_connection_.get_secondary_event(); } )
```

heplesser: Yes, we now require C++20 for NEST, so we should use `requires` here, and correspondingly below in `connector_model_impl.h`. Then we would not need to introduce `has_get_secondary_event_t` above at all, right?

And could we then not use `static_assert()` also here instead of `if constexpr`?

heplesser (commented):
@muffgaga Thanks for the updates! We now require C++20 for NEST, so we should use the `requires` approach, see also my inline comment.

Your code fails some of your formatting checks. Could you run the pre-commit hooks to fix that, see https://nest-simulator.readthedocs.io/en/stable/developer_space/guidelines/coding_guidelines_check.html#pre-commit ?

nestkernel/connector_model.h
muffgaga: Great, I changed it to a `requires` expression (and yes, I removed the helpers above as I mentioned before).

Regarding:

> And could we then not use `static_assert()` also here instead of `if constexpr`?

I believe there are still callers where types aren't known at compile type (maybe they could be known, but the code doesn't explicitly use constexpr) — so I think we still need to guard the code path by this conditional (as this would be illegal code for "primary" connection types). (…but I did not fully (re-)investigate — that's basically what I remember from the time I created this PR…)

heplesser:
@muffgaga Thanks! I sent you a PR towards your branch with a few more improvements.

muffgaga:
> @muffgaga Thanks! I sent you a PR towards your branch with a few more improvements.

Thanks → I just rebased onto `main` and cherry-picked your change.
(And thanks for the bug fix in the exception throwing; I just took the suggested change and never looked it up myself.)

heplesser (approved):
@muffgaga Thanks! Currently, the `isort` check fails due to changes to `isort`, see #3939. Nothing to worry about, just another merge from main once that issue will have been fixed.

heplesser:
@muffgaga Could you merge the newest main to fix the problem with the isort check?

PR #3030, opened 2023-12-04
Update documentation of video tutorial to clarify version

jessica-mitchell:
The video tutorial is old and uses 2.12, and also does not work with the newer syntax.
Until a  new video is created, I added the version of NEST that is in the video and a note to 3.x users that there is incompatibility.


jessica-mitchell:
Merging with self-review as its minor documentation change

PR #3031, opened 2023-12-05
Remove unused doxygen groups in developer documentation

jessica-mitchell:
This PR removes various groups that either do not have a definition (defgroup) or they do not have any members (ingroup).
In a future PR, some groups may be re-introduced but with proper structuring.


terhorstd (approved):
LGTM. Probably we will add some back in later, once the new defgroups are available.
:+1:

PR #3044, opened 2023-12-11
Register only spikes actually sent with weight_recorder

heplesser:
Until now, spikes would be registered with a `weight_recorder` even if the spike was not transmitted to the target, e.g., by a `bernoulli_synapse`. This PR changes all `Connection<T>::send()` methods to return `true` if they actually transmit a spike to the target node (call `e();`) and `false` otherwise.

The PR also simplifies the code in the `bernoulli_synapse`. Since this synapse model can only be used between normal nodes (not to/from devices), it will only receive events with multiplicity 1, allowing for significant simplification.

@JesusEV @terhorstd I'd appreciate if you could fast-track this PR since e-prop work depends on it.

### DIFF
--- a/nestkernel/connector_base.h
+++ b/nestkernel/connector_base.h
@@ -399,7 +399,11 @@ public:
       if ( not conn.is_disabled() )
       {
         // Some synapses, e.g., bernoulli_synapse, may not send an event after all
-        conn.send( e, tid, cp ) and send_weight_event( tid, lcid + lcid_offset, e, cp );
+        const bool event_sent = conn.send( e, tid, cp );
+        if ( event_sent )
+        {
+          send_weight_event( tid, lcid + lcid_offset, e, cp );
+        }
       }
       if ( not conn.source_has_more_targets() )
       {
--- a/nestkernel/connector_base.h
+++ b/nestkernel/connector_base.h
@@ -399,11 +399,7 @@ public:
       if ( not conn.is_disabled() )
       {
         // Some synapses, e.g., bernoulli_synapse, may not send an event after all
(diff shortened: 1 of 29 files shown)


nestkernel/connector_base.h
```diff
@@ -394,16 +394,18 @@ class Connector : public ConnectorBase
...
+      if ( not conn.is_disabled() )
       {
-        conn.send( e, tid, cp );
-        send_weight_event( tid, lcid + lcid_offset, e, cp );
+        // Some synapses, e.g., bernoulli_synapse, may not send an event after all
+        const bool event_sent = conn.send( e, tid, cp );
+        if ( event_sent )
+        {
+          send_weight_event( tid, lcid + lcid_offset, e, cp );
+        }
```
terhorstd, line 406:
```suggestion
        conn.send( e, tid, cp ) and send_weight_event( tid, lcid + lcid_offset, e, cp );
```

JesusEV (approved):
I applied these changes to a copy of the latest `eprop_feature`  and applied the corresponding changes to `eprop_synapse.h` (returning false if the spike is to be ignored) and `learning_signal_connection.h`.

All verifications were successful.

By themselves, and all in all, the changes seem very appropriate

terhorstd (approved):
LGTM :+1:

nestkernel/connector_base.h
terhorstd: OK, this doesn't work for `send_weight_event()` not being cast-able to `bool`.

terhorstd (approved):

JesusEV (approved):

PR #3051, opened 2023-12-15
Properly handle OpenMP linker flags

heplesser:
Until now, we placed `OpenMP_CXX_FLAGS` in the `MODULE_LINK_LIBS` for nest. This is not quite correct and meant that `libomp` was missing from the list of libraries returned by `nest-config --libs`, which in turn caused trouble for building external models with OpenMP, at least on macOS. This PR fixes this.


gtrensch (changes requested):
Thank you for this contribution. It looks good to me. I would just like to make one suggestion for improvement. In order to report on the OpenMP libraries used, we could add the following to `ConfigureSummary.cmake`, after line 102:
`message( "    Libraries       : ${OpenMP_CXX_LIBRARIES}" )`

heplesser:
> Thank you for this contribution. It looks good to me. I would just like to make one suggestion for improvement. In order to report on the OpenMP libraries used, we could add the following to `ConfigureSummary.cmake`, after line 102: `message( " Libraries : ${OpenMP_CXX_LIBRARIES}" )`

@gtrensch Thanks for the suggestion, fixed!

heplesser:
@gtrensch @clinssen Could you (re-)review asap, this is blocking progress elsewhere.

pnbabu (approved):

gtrensch (approved):
Looks good to me!

PR #3060, opened 2024-01-02
Add  to values to be set for synaptic elements

ackurth:
The ```z_connected``` variables in synaptic elements in used to keep track of the connected synaptic elements, in contrast to the ```z``` variable that is used to determine how many connections ought to be established or deleted.
There are use cases when one wants to set ```z_connected``` by hand (for example when creating connections with a higher level connections routine that subsequently should be subject to potential removal by structural plasticity).
This PR add ```z_connected``` to the parameters that can be set by the user.

### DIFF
--- a/nestkernel/synaptic_element.cpp
+++ b/nestkernel/synaptic_element.cpp
@@ -115,6 +115,7 @@ nest::SynapticElement::set( const DictionaryDatum& d )
   updateValue< double >( d, names::tau_vacant, new_tau_vacant );
   updateValue< bool >( d, names::continuous, continuous_ );
   updateValue< double >( d, names::z, z_ );
+  updateValue< double >( d, names::z_connected, z_connected_ );
 
   if ( d->known( names::growth_curve ) )
   {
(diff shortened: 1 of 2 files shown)


heplesser (commented):
I don't think I can follow the logic here. So far, `z_connected_` is counted up or down as actual synapses are created. Setting this to some arbitrary value would break the consistency of the `SynapticElement` object, I fear.

clinssen:
@ackurth: why could this not be done through calling ``SynapticElement::connect()``?

heplesser:
@ackurth Could you respond to the comments above?

heplesser:
@ackurth Is this PR still relevant? If so, kindly respond to my questions, otherwise I will close the PR.

nestkernel/synaptic_element.cpp
```diff
@@ -115,6 +115,7 @@ nest::SynapticElement::set( const DictionaryDatum& d )
   updateValue< double >( d, names::tau_vacant, new_tau_vacant );
   updateValue< bool >( d, names::continuous, continuous_ );
   updateValue< double >( d, names::z, z_ );
+  updateValue< double >( d, names::z_connected, z_connected_ );
```
heplesser, line 118: `z_connected` is an integer-value parameter, so shouldn't it be

```suggestion
  updateValue< int >( d, names::z_connected, z_connected_ );
```

PR #3061, opened 2024-01-02
Rename  to ```tau_vacant``` to ```decay_vacant```

ackurth:
In the structural plasticity routine vacant (i.e. at a given time unused) synaptic elements decay with a "rate" (word used in the source code) ```tau_vacant```, also referred to as a "time constant" in one case.
I would argue the decay constant is neither a rate (units 1 / time unit) nor a time constant (units of time unit) , but a quantity without any unit.
Since ```tau_vacant``` insinuates a time constant I suggest renaming the decay constant in just ```decay_vacant``` (in line with the naming for this in [this](https://www.frontiersin.org/articles/10.3389/fnana.2016.00057/full) paper referring to it as a decay constant).
Also, in one case it is required that the decay constant is greater than 0.
I find this impractical and unnecessary and thus suggest to alter this to greater or equal than 0.

### DIFF
--- a/nestkernel/nest_names.cpp
+++ b/nestkernel/nest_names.cpp
@@ -534,7 +534,7 @@ const Name tau_u_bar_minus( "tau_u_bar_minus" );
 const Name tau_u_bar_plus( "tau_u_bar_plus" );
 const Name tau_V_th( "tau_V_th" );
 const Name tau_v( "tau_v" );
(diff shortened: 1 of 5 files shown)


heplesser:
@ackurth Could you follow up the comments and merge master?

ackurth:
I close the PR, issue can be addressed when improving on structural plasticity in general.

PR #3065, opened 2024-01-17
Modify doc extensions to check for correct files

jessica-mitchell:
This PR modifies two extension scripts so they only use the autogenerated files created from sphinx-gallery rather than all rst files in the pynest/examples directory. This means any rst files that are added manually will be skipped by the extensions.


jessica-mitchell:
Merging with no reviewers as its minor doc change

PR #3090, opened 2024-02-02
Ensure all MPI communication to is handled by master thread

terhorstd:
According to the MPI specifications, calls have to be done from the *same* thread (`master`) instead of from a `single` thread in MPI_FUNNELED mode . This means a number of `omp single` pragmas needed to be changed to `omp master`. This should not impact performance overly much, but needs checking!

:warning: Since an exit-barrier is not implicit for `omp master`, I added explicit ones where not given anyway.


heplesser (approved):
@terhorstd Thank you for this fix. While strictly speaking the resize calls could be done in `single`,  since they do not perform MPI communication, it makes sense to place them in master so we are certain that the buffer data structures always are allocated on the thread performing MPI communication.

PR #3092, opened 2024-02-02
Ensure assert_valid_syn_id() is always called with correct thread id

heplesser:
Automated thread analysis revealed a data race because `ModelManager::get_connection_model()` calls `assert_valid_syn_id()` without thread argument

https://github.com/nest/nest-simulator/blob/d5aa3fe65e3d4e6eb368de561d407a572dc4e467/nestkernel/model_manager.h#L303

thus using the default thread number 0. This can lead to data races during parallel construction of new models during `CopyModel`.  This PR removes the default value for the thread so avoid data races.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -207,7 +207,7 @@ nest::ConnectionManager::get_synapse_status( const size_t source_node_id,
   const synindex syn_id,
   const size_t lcid ) const
 {
-  kernel().model_manager.assert_valid_syn_id( syn_id );
+  kernel().model_manager.assert_valid_syn_id( syn_id, kernel().vp_manager.get_thread_id() );
 
   DictionaryDatum dict( new Dictionary );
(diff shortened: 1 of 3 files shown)


otcathatsya (approved):
Looks good; would it be an option to set the current thread id as the parameter default? From what I can see the method is only used inside connection and model manager who always call it with the same parameterization.

heplesser:
> Looks good; would it be an option to set the current thread id as the parameter default? From what I can see the method is only used inside connection and model manager who always call it with the same parameterization.

The parameter default must be a constant, so we can't make the current thread the default.

otcathatsya:
> > Looks good; would it be an option to set the current thread id as the parameter default? From what I can see the method is only used inside connection and model manager who always call it with the same parameterization.
> 
> The parameter default must be a constant, so we can't make the current thread the default.
My bad; could do an overload to produce the same behaviour?

suku248 (approved):
Looks good, thanks @heplesser

PR #3093, opened 2024-02-06
Rename the variable "Ca" in astrocyte_lr_1994 to "Ca_astro"

HanjiaJiang:
This PR is proposed to solve a conflict due to the same variable name "Ca" in astrocyte_lr_1994 and StructuralPlasticityNode.

We found that nest.GetStatus() and nest.GetDefaults() give wrong values of the calcium concentration (under the name "Ca") in astrocyte_lr_1994. This is because the parent class StructuralPlasticityNode also has a variable for calcium concentration under the same name. This conflict does not affect the simulations, as the underlying dynamics is not affected. Nevertheless, we would like change the variable name in astrocyte_lr_1994 to "Ca_astro" to avoid this conflict.

Corresponding changes have been made for all relevant kernel, documentation, and test files.


heplesser (approved):
Thanks, looks good to me.

jessica-mitchell (approved):
LGTM! thanks

Issue #3100, opened 2024-02-11
Properly control validity of NodeCollections and SynapseCollections with hashes


heplesser:
`NodeCollection`s and `SynapseCollection`s remain existent in Python even after a `ResetKernel()`. We currently protect against failing C++ assertions on `set()`/`get()` on `SynapseCollection` by checking against an empty network, but that is approximate at best.

A robust solution would be to generate a time-based hash every time `ResetKernel()` is called. Each `NodeCollection` is assigned the kernel hash valid when it is created, and any operation using the `NodeCollection` to interact with the kernel checks that the NC-hash is equal to the current kernel hash.

For `SynapseCollection`s, one can do things correspondingly, but here a new hash should be set every time `connections_have_changed` is set to true. The "no longer valid" warning can then disappear, as the user will get an error.

clinssen:
This would solve an issue with slowdowns happening due to repeatedly setting a parameter of a SynapseCollection.

@heplesser tracked this down to the following.

> Specifically, in pynest/nest/lib/hl_api_types.py, SynapseCollections.get() and .set() both contain 
>
>        if self.__len__() == 0 or GetKernelStatus("network_size") == 0:
>
> The underlying problem is that GetKernelStatus() first collects the full network status and then picks out the network size. Again, I have no idea why this took about 75% of the total runtime and why it got slower over time (there is not more data to return).

PR #3101, opened 2024-02-11
Avoid MPI deadlocks on SynapseCollection.get()/set()

### LINKED ISSUE #3099: NEST hangs when setting connection weight in parallel simulations
The tests below will pass if run serially, but hang when run with `mpirun -np 2 pytest ...`. This also holds if only one of the three tests is executed. Usually, one MPI process will report success while the other never completes. Note that for two processes, only one will actually have the connection. Presumably, this leads to incorrect MPI communication, where one rank hangs.

This PR is based on a bug report by Xavier Otzau on the NEST User Mailing List.

```python
import nest
import pytest


@pytest.fixture
def conns():
    nest.ResetKernel()
    nrn = nest.Create('parrot_neuron')
    nest.Connect(nrn, nrn)
    return nest.GetConnections()

def test_conn_weight(conns):
    if conns:
        conns.weight = 2.5

def test_set_weight(conns):
    if conns:    
        conns.set({'weight': 2.5})

def test_set_status_weight(conns):
    if conns:
        nest.SetStatus(conns, 'weight', 2.5)
```

heplesser:
This fixes #3099. MPI deadlocks could occur if only some MPI ranks performed `get()` or `set()` operations on `SynapseCollection`s. These operations are local operations so while not elegant, they should work even if executed only on some ranks and not on other, e.g., when a rank has no connections.

The hang occurred because the `get()`/`set()` methods called `GetKernelStatus()` to check for the number of connections to decide whether an SC was still valid. This was misguided to begin with, since the number of connections is only known locally. This is now changed to use the network size, whose global value is locally known. For a proposed proper solution, see #3100.

A further problem then was that `GetKernelStatus()` called `ConnectionManager::update_delay_extrema_()`, which unconditionally performed MPI exchange of `min/max_delay`. This PR changes this so that delay extrema are MPI-exchanged only if their values could have changed remotely. This also eliminates the MPI communication hidden in every single `nest.<some property>` use and might thus lead to faster execution of simulation scripts.

The test simply checks that it does not time out.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -464,6 +464,13 @@ nest::ConnectionManager::connect( TokenArray sources, TokenArray targets, const 
 void
 nest::ConnectionManager::update_delay_extrema_()
 {
+  if ( kernel().simulation_manager.has_been_simulated() )
+  {
+    // Once simulation has started, min/max_delay can no longer change,
+    // so there is nothing to update.
+    return;
+  }
+
   min_delay_ = get_min_delay_time_().get_steps();
   max_delay_ = get_max_delay_time_().get_steps();
 
@@ -475,7 +482,13 @@ nest::ConnectionManager::update_delay_extrema_()
     max_delay_ = std::max( max_delay_, kernel().sp_manager.builder_max_delay() );
   }
 
-  if ( kernel().mpi_manager.get_num_processes() > 1 )
+  // If the user explicitly set min/max_delay, this happend on all MPI ranks,
+  // so all ranks are up to date already. Also, once the user has set min/max_delay
+  // explicitly, Connect() cannot induce new extrema. Thuse, we only need to communicate
+  // with other ranks if the user has not set the extrema and connections may have
+  // been created.
+  if ( not kernel().connection_manager.get_user_set_delay_extrema()
+    and kernel().connection_manager.connections_have_changed() and kernel().mpi_manager.get_num_processes() > 1 )
   {
     std::vector< long > min_delays( kernel().mpi_manager.get_num_processes() );
     min_delays[ kernel().mpi_manager.get_rank() ] = min_delay_;
--- a/testsuite/pytests/mpi/2/test_issue_3099.py
+++ b/testsuite/pytests/mpi/2/test_issue_3099.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+#
+# test_issue_3099.py
+#
+# This file is part of NEST.
+#
+# Copyright (C) 2004 The NEST Initiative
+#
+# NEST is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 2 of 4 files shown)


nestkernel/connection_manager.cpp
```diff
@@ -464,6 +464,13 @@ nest::ConnectionManager::connect( TokenArray sources, TokenArray targets, const
 void
 nest::ConnectionManager::update_delay_extrema_()
 {
+  if ( kernel().simulation_manager.has_been_simulated() )
+  {
+    // Once simulation has started, min/max_delay can no longer change,
+    // so there is nothing to update.
+    return;
+  }
```
nicolossus, line 472: @heplesser In `ConnectionManager::get_connections()` there is a call to `update_delay_extrema_()` if `connections_have_changed`. It seems like it is used for setting the secondary event data size. I'm not sure if your change affect this, so I just wanted to check if you are aware and have taken that into consideration with the change?

testsuite/pytests/mpi/2/test_issue_3099.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.fixture
+def conns():
+    nest.ResetKernel()
+    nrn = nest.Create("parrot_neuron")
+    nest.Connect(nrn, nrn)
+    return nest.GetConnections()
+
+
+def test_conn_weight(conns):
+    """Test that operation does cause MPI deadlock."""
```
nicolossus, line 36:
```suggestion
    """Test that operation does not cause MPI deadlock."""
```
Since the test should fail if a deadlock happens, a negation is more correct I think. Also applies to the docstrings below.

nicolossus (approved):
@heplesser LGTM. I have some minor comments/clarifications, but I'll go ahead and approve.

otcathatsya (approved):

nestkernel/connection_manager.cpp
heplesser: Thanks for pointing this out, but this is safe:  `update_delay_extrema()`  is called at least once while `has_been_simulated()` is still false, so we will get the correct min/max values.

testsuite/pytests/mpi/2/test_issue_3099.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_conn_weight(conns):
+    """Test that operation does not cause MPI deadlock."""
+
+    if conns:
+        conns.weight = 2.5
+
+
+def test_set_weight(conns):
+    """Test that operation does cause MPI deadlock."""
```
heplesser, line 43:
```suggestion
    """Test that operation does not cause MPI deadlock."""
```

testsuite/pytests/mpi/2/test_issue_3099.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if conns:
+        conns.set({"weight": 2.5})
+
+
+def test_set_status_weight(conns):
+    """Test that operation does cause MPI deadlock."""
```
heplesser, line 50:
```suggestion
    """Test that operation does not cause MPI deadlock."""
```

PR #3113, opened 2024-02-24
Attempt cache thrashing fix for low synapse count vector access

otcathatsya:
- This is an attempt to fix an observed niche issue where, on a C++ level, connection creation will be slowed down by the `num_connections_` vector accessing inner indices on the same cache line and therefore invalidating memory across threads, usually when there is only a few synapse models (e.g. when running with `-Dwith-modelset=iaf_minimal`)
- Very much experimental, this fixes the `time_construction_connect` difference in SLI tests but makes no difference on a Python level (neither did the problem occur when timing from Python to begin with. Too insignificant?) 
- This does increase memory consumption up to a certain point (the L1 cache line usually is 64 bytes), but stays below full modelset levels.
- Maybe the destructive interference constexprs can be useful for something else further down the line
- If this does get considered it should probably be amended to include `-Wno-interference-size` to suppress the compile warnings about this value varying between platforms, or be an optional CMake flag altogether

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -131,7 +131,7 @@ nest::ConnectionManager::initialize( const bool reset_kernel )
   std::vector< DelayChecker > tmp( kernel().vp_manager.get_num_threads() );
   delay_checkers_.swap( tmp );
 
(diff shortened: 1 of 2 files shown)


terhorstd:
@suku248 and @heplesser want to be in the loop for this.

PR #3117, opened 2024-02-27
Fix  tests and examples to pass Pylint test R1730

heplesser:
The R1730 test in Pylint seems to have become active with the recent update from Pylint 3.0.3 to Pylint 3.1.0 in the Github Actions.

This PR makes our tests and examples compatible.

Labeling as critical as it blocks our CI.


suku248 (approved):
:+1:

PR #3120, opened 2024-02-28
Fix data race in nestkernel/per_thread_bool_indicator.cpp

jprotze:


nestkernel/per_thread_bool_indicator.h
```diff
@@ -106,6 +108,34 @@ class PerThreadBoolIndicator
...
+  void
+  set_true( const size_t tid )
+  {
+    if ( per_thread_status_[ tid ].is_false() )
+    {
+      are_true_++;
+      per_thread_status_[ tid ].set_true();
+    }
+  }
+  void
```
heplesser, line 120:
```suggestion

  void
```

nestkernel/per_thread_bool_indicator.h
```diff
@@ -106,6 +108,34 @@ class PerThreadBoolIndicator
...
+  set_false( const size_t tid )
+  {
+    if ( per_thread_status_[ tid ].is_true() )
+    {
+      are_true_--;
+      per_thread_status_[ tid ].set_false();
+    }
+  }
+  void
```
heplesser, line 129:
```suggestion

  void
```

nestkernel/per_thread_bool_indicator.h
```diff
@@ -106,6 +108,34 @@ class PerThreadBoolIndicator
...
+  logical_and( const size_t tid, const bool status )
+  {
+    if ( per_thread_status_[ tid ].is_true() && !status )
```
heplesser, line 132: It is NEST convention to always write out boolean operators:
```suggestion
    if ( per_thread_status_[ tid ].is_true() and not status )
```

nestkernel/per_thread_bool_indicator.h
```diff
@@ -133,6 +163,7 @@ class PerThreadBoolIndicator
 
 private:
   std::vector< BoolIndicatorUInt64 > per_thread_status_;
+  std::atomic< int > size_ { 0 }, are_true_ { 0 };
```
heplesser, line 166: We generally avoid multiple definitions in a single statement. 

```suggestion
  std::atomic< int > size_ { 0 };
  
  /** Number of per-thread indicators currently true
   *
   * are_true_ == 0 -> all are false
   * are_true_ == size_  -> all are true
   * 0 < are_true_ < size_  -> some true, some false
   */
  std::atomic< int > are_true_ { 0 };
```

Another question: Does `size_` need to be atomic, or is there any advantage to it being atomic, given that it is only set by `initialize()` and then just read later?

nestkernel/per_thread_bool_indicator.cpp
```diff
@@ -50,62 +50,47 @@ PerThreadBoolIndicator::initialize( const size_t num_threads, const bool status
   kernel().vp_manager.assert_single_threaded();
   per_thread_status_.clear();
   per_thread_status_.resize( num_threads, BoolIndicatorUInt64( status ) );
+  size_ = num_threads;
+  if ( status )
+    are_true_ = num_threads;
+  else
+    are_true_ = 0;
```
heplesser, line 57: We require braces also for single-statement blocks.
```suggestion
  if ( status )
  {
    are_true_ = num_threads;
  }
  else
  {
    are_true_ = 0;
  }
```

nestkernel/per_thread_bool_indicator.cpp
```diff
@@ -50,62 +50,47 @@ PerThreadBoolIndicator::initialize( const size_t num_threads, const bool status
...
 }
 
 bool
 PerThreadBoolIndicator::all_false() const
 {
 #pragma omp barrier
```
heplesser, line 63:
```suggestion
// We need two barriers here to ensure that no thread can continue and change the result
// before all threads have determined the result.
#pragma omp barrier
```

heplesser (changes requested):
@jprotze Thank you for the PR. I just have a few suggestions mainly related to NEST coding conventions.

heplesser:
@ackurth Could you benchmark this branch on a single node with microcircuit, up to 128 threads? Just to check that the change does not introduce a performance regression.

ackurth:
> @ackurth Could you benchmark this branch on a single node with microcircuit, up to 128 threads? Just to check that the change does not introduce a performance regression.

Will do.

nestkernel/per_thread_bool_indicator.h
jprotze: I think, size_ doesn't need to be atomic

heplesser (commented):
@jprotze This looks good to me now. I still wait for results from systematic benchmarks on larger models (first checks look good) and will wait for final approval until then.

ackurth:
> @ackurth Could you benchmark this branch on a single node with microcircuit, up to 128 threads? Just to check that the change does not introduce a performance regression.

I compared the suggested changes to the current NEST Master, there is no performance regression in our usual benchmarks of the microcircuit model.

heplesser:
> > @ackurth Could you benchmark this branch on a single node with microcircuit, up to 128 threads? Just to check that the change does not introduce a performance regression.
> 
> I compared the suggested changes to the current NEST Master, there is no performance regression in our usual benchmarks of the microcircuit model.

Excellent, then we can proceed. Thanks!

heplesser (approved):
With positive benchmark results, I approve.

suku248 (approved):
Thanks @jprotze for this PR and @ackurth for benchmarking!

PR #3129, opened 2024-03-01
Automatically adjust column widths in testsuite report

heplesser:
This PR makes sure that columns format nicely even if phase name become longer. Also switches to format strings.


steffengraber (approved):
Looks good

terhorstd (approved):
LGTM

PR #3133, opened 2024-03-05
Fix minor integer type mismatches

heplesser:
This PR fixes some warnings by GCC 12.3 about integer type mismatches.

Note that we have a bit of a mess between `int` (e.g., `MPIManager::num_processes_`) and `size_t` (e.g., return type of `MPIManager::get_num_processes()`), but cleaning that up is for another day.

### DIFF
--- a/nestkernel/mpi_manager.cpp
+++ b/nestkernel/mpi_manager.cpp
@@ -400,7 +400,7 @@ nest::MPIManager::communicate_Allgather( std::vector< unsigned int >& send_buffe
   int disp = 0;
   unsigned int max_recv_count = send_buffer_size_;
   bool overflow = false;
-  for ( int pid = 0; pid < get_num_processes(); ++pid )
+  for ( size_t pid = 0; pid < get_num_processes(); ++pid )
   {
     unsigned int block_disp = pid * send_buffer_size_;
     displacements[ pid ] = disp;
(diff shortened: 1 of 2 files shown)


otcathatsya (approved):
LGTM! As a related question, would we be interested in applying type/tidiness suggestions (e.g. "can be made const", "can be made static") to all files? Some IDEs conveniently provide the feature to analyze all files in a scope and they should be safe.

JanVogelsang:
> LGTM! As a related question, would we be interested in applying type/tidiness suggestions (e.g. "can be made const", "can be made static") to all files? Some IDEs conveniently provide the feature to analyze all files in a scope and they should be safe.

While this is a good idea, it is preferable to find some tools which can be invoked from the CI pipeline so that the code-refactoring would be a continuous process and also applies to new code rather than just a one-time thing for the current codebase.  
Most IDEs just use clang-format/clang-tidy or similar tools under the hood, which we could use in our pipelines as well.

JanVogelsang (approved):

JanVogelsang:
> but cleaning that up is for another day

Could you open an issue in which you mention the mess? It would be great if we would collect a list of potential refactorings that one could collectively do at some point, or ideally use these as a test case for automated refactoring tools (e.g. clang-tidy).

otcathatsya:
> > LGTM! As a related question, would we be interested in applying type/tidiness suggestions (e.g. "can be made const", "can be made static") to all files? Some IDEs conveniently provide the feature to analyze all files in a scope and they should be safe.
> 
> While this is a good idea, it is preferable to find some tools which can be invoked from the CI pipeline so that the code-refactoring would be a continuous process and also applies to new code rather than just a one-time thing for the current codebase. Most IDEs just use clang-format/clang-tidy or similar tools under the hood, which we could use in our pipelines as well.

I definitely agree that we could use that for CI, though might not have to be mutually exclusive. A lot of the IDE & static analysis tools go beyond clang-tidy, though it'd probably be best to have a session and go over some of that output together at some point since not every refactor is safe but may very well cut down build or even runtime (missing constexprs, unused imports, redundant checks etc)

PR #3136, opened 2024-03-05
Fix raster plot physical units in label

clinssen:
With props to @abigailm

### DIFF
--- a/pynest/nest/raster_plot.py
+++ b/pynest/nest/raster_plot.py
@@ -268,7 +268,7 @@ def _make_plot(ts, ts1, node_ids, neurons, hist=True, hist_binwidth=5.0, graysca
 
         plt.bar(t_bins, heights, width=hist_binwidth, color=color_bar, edgecolor=color_edge)
         plt.yticks([int(x) for x in numpy.linspace(0.0, int(max(heights) * 1.1) + 5, 4)])
-        plt.ylabel("Rate (Hz)")
+        plt.ylabel("Rate (spks/s)")
         plt.xlabel(xlabel)
         plt.xlim(xlim)
         plt.axes(ax1)


doc/htmldoc/ref_material/glossary.rst
```diff
@@ -41,22 +41,19 @@ Units of measure
...
  spike rates
-   Spikes/second.
-
- modulation frequencies
-   Herz (Hz).
+   The number of spikes that occurred in a certain time interval, usually expressed in terms of spikes per second (spks/s or s^-1).
 
  frequency
-   Frequncy in Hertz (Hz).
+   Frequency in Hertz (Hz). Note that spike rates are often better expressed in terms of spikes per second (spks/s or s^-1).
```
jessica-mitchell, line 56: Would it make more sense since  it seems all instances of Hertz are changed, that we just say Frequency in spikes per second (spks/s)?

clinssen: Frequency ↔ Hz and rate ↔ spks/s might be good to keep consistent. I'm open to suggestions though.

heplesser: That is exactly how it should be :)

pynest/examples/balancedneuron.py
```diff
@@ -123,13 +123,13 @@
 
 
 def output_rate(guess):
-    print("Inhibitory rate estimate: %5.2f Hz" % guess)
+    print("Inhibitory rate estimate: %5.2f spks/s" % guess)
```
heplesser, line 126: Please update to use Python format strings here and in other places touched by this PR.

pynest/nest/raster_plot.py
```diff
@@ -268,7 +268,7 @@ def _make_plot(ts, ts1, node_ids, neurons, hist=True, hist_binwidth=5.0, graysca
 
         plt.bar(t_bins, heights, width=hist_binwidth, color=color_bar, edgecolor=color_edge)
         plt.yticks([int(x) for x in numpy.linspace(0.0, int(max(heights) * 1.1) + 5, 4)])
-        plt.ylabel("Rate (Hz)")
+        plt.ylabel("Rate (spks/s)")
```
heplesser, line 271: There seems to be some inconsistency in using () and [] around units. @jessica-mitchell Do we have a policy on this?

heplesser (changes requested):
Looks good overall, just some small suggestions. And obviously I do not see here what you may have overlooked.

pynest/nest/raster_plot.py
jessica-mitchell: No we don't; it seems from a quick look into the notation that both parentheses and square brackets are simply style choices - so my opinion is that whatever the majority of cases show that we do that. What do you think @heplesser

pynest/examples/balancedneuron.py
clinssen: Done!

pynest/nest/raster_plot.py
clinssen: Spun off to https://github.com/nest/nest-simulator/issues/3375.

pynest/examples/one_neuron_with_noise.py
```diff
@@ -56,9 +56,9 @@
 ###############################################################################
 # Third, the Poisson generator is configured using ``SetStatus``, which expects
 # a list of node handles and a list of parameter dictionaries. We set the
-# Poisson generators to 80,000 Hz and 15,000 Hz, respectively. Note that we do
-# not need to set parameters for the neuron and the voltmeter, since they have
-# satisfactory defaults.
+# Poisson generators to 8,000 spks/s and 15,000 spks/s, respectively. Note that
```
heplesser, line 59: In the code below, it looks like eighty-thousand, so it should be
```suggestion
# Poisson generators to 80,000 spks/s and 15,000 spks/s, respectively. Note that
```

heplesser (changes requested):
@clinssen Looks good, I just have one small correction, see inline.

PR #3138, opened 2024-03-06
astrocyte_lr_1994: Store result of expression evaluation in proper variable

### LINKED ISSUE #3137: Compiler warns about initializing reference with temporary in astrocyte_lr_1994
The newest AppleClang (15.0.0.15000309) warns about assigning the result of a temporary expression evaluation result to a reference:

https://github.com/nest/nest-simulator/blob/56395725b3bb89bda0c0307c6038b0bed0281e0b/models/astrocyte_lr_1994.cpp#L84-L89

 `conc` should be a proper variable. The two references above and below are fine, since they point to vector elements.

heplesser:
This PR fixes #3137. I re-ordered lines a bit so that all reference initialization come first, then initialization of genuine variables based on expression evaluation.

### DIFF
--- a/models/astrocyte_lr_1994.cpp
+++ b/models/astrocyte_lr_1994.cpp
@@ -83,11 +83,11 @@ astrocyte_lr_1994_dynamics( double time, const double y[], double f[], void* pno
 
   // shorthand for state variables
   const double& ip3 = y[ S::IP3 ];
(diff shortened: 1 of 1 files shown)


HanjiaJiang (approved):
Thanks for fixing this!

PR #3139, opened 2024-03-06
Another approach to fix the data race in PerThreadBoolIndicator

jprotze:
This approach avoids some of the barriers introduced by the other PR. 
But, more significantly it does not introduce the atomic counter for the instances of PerThreadPoolIndicator that don't use the collective logical operations.

### DIFF
--- a/nestkernel/event_delivery_manager.cpp
+++ b/nestkernel/event_delivery_manager.cpp
@@ -67,7 +67,6 @@ EventDeliveryManager::EventDeliveryManager()
   , send_recv_buffer_shrink_spare_( 0.1 )
   , send_recv_buffer_grow_extra_( 0.5 )
   , send_recv_buffer_resize_log_()
-  , gather_completed_checker_()
 {
 }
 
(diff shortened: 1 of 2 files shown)


heplesser:
@jprotze In the OpenMP only (no MPI) build of NEST this variant fails several of our tests. With MPI and OpenMP, tests run into deadlocks under Linux and macOS, so I am afraid something is not quite right yet.

jprotze:
Would you consider this alternative approach at all? 
Then I would look into the issue and what is going wrong.
Otherwise, I would just drop this PR. The other one is ready to merge, I think.

heplesser:
> Would you consider this alternative approach at all? Then I would look into the issue and what is going wrong. Otherwise, I would just drop this PR. The other one is ready to merge, I think.

@jprotze This approach looks quite a bit simpler than the one based on the `PerThreadBoolIndicator`, so if you can make it work and it did not affect performance, I would actually prefer it. As we would like to fix this bug within a few days and #3120 does so an passes, I would consider merging that first and then replace it with this approach later on when it works.

heplesser:
@jprotze Since #3120 is merged now, I will close this PR for now to avoid confusion. Feel free to re-open it later.

PR #3145, opened 2024-03-12
Stop IPython from time consuming deep inspection of NodeCollection objects

### LINKED ISSUE #3144: Displaying NodeCollection is very slow in IPython
Running the following in `IPython`

```python
import nest
n = nest.Create('parrot_neuron', positions=nest.spatial.free([(x, 0) for x in range(10000)]))
n
```

the last line takes extraordinarily long, over 10 seconds, I think. But if one instead does
 
```python
print(n)
```

it executes instantaneously.
 
Also, executing the same code in plain python, the Node collection is displayed immediately.

heplesser:
This solution fixes #3144 and was suggested by Harold Gutch via NEST User (12 March 2024) with the following explanation

> What happens is that IPython's pretty printing calls (a bit down the line) the `IPythonDisplayFormatter()` class.  The first thing that does is test if for the given object there is an `_ipython_display_()` method — if that exists, it calls that, otherwise it falls back to a more generic method.  In your case the latter happens, it falls back to the more generic method and that is what then produces the output.
> 
> When looking up if there is an `_iython_display_()` method, the first thing IPython does is check for a method called `_ipython_canary_method_should_not_exist_()`— and it is precisely this lookup that takes so long.  You can verify this by
> 
> ```python
> getattr(n, "_ipython_canary_method_should_not_exist_")
> ```
> 
> which will also take rather long.  As the output here also contains some SLI things, one might need to dig deeper to actually speed this up.
> 
> Instead, adding a check specifically for exactly the above mentioned lookup of `_ipython_canary_method_should_not_exist_()` also does the job ... .  With that workaround, your example
> returns instantaneously — as one would expect it to.
>

### DIFF
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -548,9 +548,11 @@ class NodeCollection:
             raise AttributeError("Cannot get attribute of empty NodeCollection")
 
         # IPython looks up this method when doing pretty printing
-        # Without special casing this, this would go through SLI which is *slow*
+        # As long as we do not provide special methods to support IPython prettyprinting,
+        # HTML-rendering, etc, we stop IPython from time-consuming checks by raising an
+        # exception here. The exception must *not* be AttributeError.
         if attr == "_ipython_canary_method_should_not_exist_":
-            raise KeyError(attr)
+            raise NotImplementedError("_ipython_canary_method_should_not_exist_")
 
         if attr == "spatial":
             metadata = sli_func("GetMetadata", self._datum)
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -547,6 +547,11 @@ class NodeCollection:
         if not self:
             raise AttributeError("Cannot get attribute of empty NodeCollection")
 
+        # IPython looks up this method when doing pretty printing
+        # Without special casing this, this would go through SLI which is *slow*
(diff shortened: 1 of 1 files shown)


pynest/nest/lib/hl_api_types.py
```diff
@@ -547,6 +547,11 @@ def __getattr__(self, attr):
         if not self:
             raise AttributeError("Cannot get attribute of empty NodeCollection")
 
+        # IPython looks up this method when doing pretty printing
+        # Without special casing this, this would go through SLI which is *slow*
+        if attr == "_ipython_canary_method_should_not_exist_":
+            raise KeyError(attr)
```
Helveg, line 553: How about 

```suggestion
            super().__getattribute__(attr)
```

as this will raise exactly the same AttributeError a normal object would. One could do the same for the numpy compatibility check right below.

Helveg (approved):
Just a nitpick, fine either way :)

JanVogelsang (approved):

pynest/nest/lib/hl_api_types.py
heplesser: @Helveg Actually, this does not work. Some more experimentation showed that to stop IPython here, one needs to raise any kind of exceptions *except* an `AttributeError`. But `super.__getattribute__(attr)` wil raise exactly that and we are as far as before. I will change the exception to `NotImplementedError`,  which seems to make more sense here and add another comment, but then merge without further review if the tests pass.

PR #3112, opened 2024-02-23
Fix processing of CMake option `with-mpi`

gtrensch:
Meant to address #2993, but does not.

PR #3148, opened 2024-03-12
Make -Dwith-mpi respect provided path and ensure "default off" options are turned on only by explicit ON

gtrensch:
Unify the handling of CMake options and explicitly check the contents of variables.

To be more precise:

- Some options with default `OFF` could previously be turned on by any non-empty string. Now `ON` is explicitly required.
- Previously, passing `-Dwith-mpi=/opt/local` would not make CMake look for MPI in `/opt/local`, it would search in default locations. Now it respects the path given.

This is a replacement for #3112.


heplesser (approved):
Fixed CMake syntax, thus approving.

terhorstd (approved):
LGTM :+1:

PR #3149, opened 2024-03-12
Move registration of ConnBuilders, Backends, growth curves from nestmodule to respective managers, allowing unloading

heplesser:
This PR is a follow-up to #3103 and should be merged only after it (the set of changes in this one then becomes much smaller). It moves the registration of ConnBuilders, Recording/StimulationBackends and SP growth curves from `nestmodule::init()` to the `initialize()` methods of Connection, IO and SP Manger respectively, with corresponding clean-up in the `finalize()` methods. The actual changes are small, since `nestmodule::init()` in the past called `register_*` methods in the respective managers.

Thus, these elements will be unloaded on `ResetKernel()` and such elements provided by extension modules can be reloaded.  Therefore, if at all possible, this should still make it into 3.7 together with #3103.

Parameters and Masks should be re-organizedi in the same way, but they are fully managed by nestmodule and seemingly with a rather complicated approach (parameters are also available as SLI objects), so their transition to the new scheme is more work and needs to wait.


heplesser:
@pnbabu @terhorstd #3149 is now ready for review as #3103 has been merged. Two comments:

-  In `io_manager.cpp` I have moved `initialize()` and `finalize()` forward in the file so that they appear right after constructor and destructor, as in the code for other managers.
- In `KernelManager`, `io_manager` must be initialized before `model_manager` as registration of stimulation and recording device models requires that the backends are registered.

pnbabu (approved):
Looks good to me.

terhorstd (approved):
LGTM :+1:

PR #3152, opened 2024-03-14
Fix padding around links in docs

jessica-mitchell:
This PR fixes a minor issue with padding that was meant to only be applied to the buttons on the index page  but instead was applied to all links.


jessica-mitchell:
Merging with  self review as its minor documentation fix

PR #3175, opened 2024-04-09
Fix MUSIC support in NEST

### LINKED ISSUE #3174: NEST cannot be built with MUSIC and this is not detected by the CI
I forgot to update the MUSIC models in NEST in #2956, so NEST cannot be built against MUSIC any more. Our CI setup does not build NEST with MUSIC at present, thus this problem was not detected previously.

Reported by @sanjayankur31 via email.

heplesser:
This PR fixes #3174. In particular, it does the following:

1. It activates MUSIC for the CI all-in Linux build.
2. In `do_tests.sh` it turns `REPORTDIR` and names generated from it into absolute paths so they can be used from MUSIC test's `runner.sh` running in any directory. `REPORTDIR` is a uniquely named directory in the directory from which `do_tests.sh` is run, i.e., where `make installcheck` is run. The PR also removes `TMPDIR` gymnastics which actually did not have any effect on `mktemp`.
3. All `music_*_proxy` models now use the new registration mechanism. Please check extra careful that the namings for each registration function are consistent. Note that the `registrer_...()` functions are called from `$BUILD_DIR/models/models.cpp`, which is auto-generated by CMake.
4. Previously, `music_{event,rate}_in_proxy` models were set up so that they registered an `event_in`/`rate_in` port as the default input port. Since we no longer have "pristine" models (which survived a `ResetKernel()`), this registration is now done in the default constructor for the model. In addition, the registration is done again in the copy constructor every time an instance of the model is created, since `MusicManager` keeps track of ports via reference counting.

The registration mentioned in 4. is inconsistently applied, since it is only done for `event` and `rate`, but not for `cont` and `message` proxies. I have created #3178 as a follow up issue to check this out.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -527,7 +527,7 @@ jobs:
           - "boost, optimize, warning"
           - "openmp, python, gsl, ltdl, boost, optimize, warning"
           - "mpi, python, gsl, ltdl, boost, optimize, warning"
-          - "openmp, mpi, python, gsl, ltdl, boost, hdf5, sionlib, libneurosim, optimize, warning"
+          - "openmp, mpi, python, gsl, ltdl, boost, hdf5, sionlib, libneurosim, optimize, warning, music"
 
     steps:
       - name: Harden Runner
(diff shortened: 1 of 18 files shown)


heplesser:
@sanjayankur31 @terhorstd This PR fixing the NEST-MUSIC interface for NEST 3.7 is now ready for review.

heplesser:
BTW, I have run the examples in `examples/nest/music` and `pynest/examples/music_cont_...`. Of the examples in the first directory, `{clock,conf}test.music` examples cannot run because the needed `clocksource` and `constsource` binaries are not provided by Music 1.2.1.  I will create a separate issue for that.

terhorstd (approved):
LGTM :+1: Thanks for the fix and cleanup!

sanjayankur31:
LGTM too: https://koji.fedoraproject.org/koji/taskinfo?taskID=116213022

heplesser:
@terhorstd With the second approval by @sanjayankur31 in place (in form of a comment), can you merge this? I leave it to you to merge so you can do it in the way that most easily works wrt integration into both master and 3.7.

PR #3176, opened 2024-04-09
Add "How NEST Works" diagram and text to NEST docs index page

jessica-mitchell:
This PR adds a conceptual diagram and associated text to the index page.

Originally we thought to include a mermaid diagram but found a nice-looking alternative with SVG output that was generated from a mermaid -> excalidraw workflow. 

The only modification needed to the SVG was to add a target so the links in the diagram can open correctly in new tab.

See Read the Docs output here: https://nest-simulator--3176.org.readthedocs.build/en/3176/


heplesser:
@jessica-mitchell This looks very nice! In the graphics, I think I'd prefer slightly higher contrast for the "Experimental protocols" and "Data analysis" parts. I will go through the texts in more details shortly.

ddahmen:
@jessica-mitchell Great work, Jessica! I agree with HEP to have a bit more contrast for the two panels. Otherwise, the diagram looks very good to me. I also like the text next to the diagram. For the text at the top of the page, I would replace "artificial learning" by "learning", because "artificial" might have connotations to learning mechanisms (such as backprop) in artificial neural networks, while NEST rather provides more biologically plausible learning rules.

ddahmen:
Thanks a lot, Jessica. The sketch looks good to me now. I still would remove the word "artificial" in the text (see comment above). Otherwise, I am fine with the PR.

doc/htmldoc/index.rst
```diff
@@ -75,6 +83,36 @@ Welcome to the NEST Simulator documentation
...
+
+        .. raw:: html
+
+           <object width="89%" height="89%" data="_static/img/excalidraw_nestconcept.svg" type="image/svg+xml"></object>
+
+
+    .. grid-item::
+        :columns: 5
+
+        **How NEST works - Conceptual approach**
```
heplesser, line 101:
```suggestion
        **How NEST works — The big picture**
```

doc/htmldoc/index.rst
```diff
@@ -75,6 +83,36 @@ Welcome to the NEST Simulator documentation
...
+
+        This diagram provides an overview of the components essential for building and simulating a network model with NEST:
```
heplesser, line 103: I would drop this sentence entirely. I think it is all clear from the header and the following text.
```suggestion
```

doc/htmldoc/index.rst
```diff
@@ -75,6 +83,36 @@ Welcome to the NEST Simulator documentation
...
+
+        A NEST simulation is created with input from :doc:`stimulation devices </models/index_generator>`,
+        :doc:`neuron models </models/index_neuron>`, and :doc:`synapse models </models/index_synapse>`,
+        along with  :ref:`connection rules <connectivity_concepts>`.
+        You can choose what data to record with :doc:`recording devices </models/index_recorder>`.
+        After simulation, the output is ready for analysis with external tools.
```
heplesser, line 109: We should also point to our built-in analysis support and advertise our friends a little. I can never remember how to link in RST, I hope you can fix my sketch below. 
```suggestion
        After simulation, the output is ready for analysis with NEST's built in <raster_plot> and <voltage_trace> modules or external tools such as <Elephant>.
```

doc/htmldoc/index.rst
```diff
@@ -75,6 +83,36 @@ Welcome to the NEST Simulator documentation
...
+.. grid::
+
+    .. grid-item::
+        :columns: 6
+        :child-direction: row
+        :child-align: end
+
+        .. raw:: html
+
+           <object width="89%" height="89%" data="_static/img/excalidraw_nestconcept.svg" type="image/svg+xml"></object>
```
heplesser, line 95: I wonder if it would make sense to flip the figure by 90 degrees, with "Protocols" -> ... -> "Analysis" on top left-to-right and the other three boxes below. At least on a computer screen that seems to be more effective use of screen real estate.

heplesser (commented):
@jessica-mitchell This looks pretty good! I have added a few suggestions. I am also wondering if the top headline "Welcome to the ..." should not end in an exclamation mark?

jessica-mitchell:
@heplesser I applied the changes you requested, and I added a horizontal version of the diagram along with the vertical so you can compare

heplesser:
Thanks! I prefer the horizontal version, with the headline above the figure and the text below.

ddahmen (approved):
Thanks a lot, @jessica-mitchell for the great work on this index page!

jessica-mitchell:
The horiz diagram is now in

heplesser (changes requested):
Thanks for the changes. It looks very nice now with Firefox and Edge, but renders horribly with Safari 17.5, see screen shot. Any ideas?

And shouldn't it be en emdash (—) in the heading instead of hyphen?

image

jessica-mitchell:
@heplesser I modified the svg slightly - can you check Safari to see if it fixed the problem? I do not have a mac to test on at the moment.

heplesser (commented):
The image rendered now very slowly in Firefox and Chromium. Once rendered, it looked correct in Firefox, Edge and Chromium. In Safari, it is now only about 1/3 of the text width and thus so small that the text is not legible.

heplesser (approved):
@jessica-mitchell Thanks, it looks good now on all browsers I tested on my laptop and phone.

PR #3187, opened 2024-04-22
Make test log parsing robust

### LINKED ISSUE #3155: Failing MPI tests not properly detected
It appears that if tests in the `testsuite/mpi/{2,3,4}` directories fail (at least if failing due to a failed assertion), then the corresponding rows in the test report will miss instead of reporting failures, and the testsuite reports overall a passing result. 

I mark this critical because it invalidates the results of our CI.

heplesser:
This PR fixes #3155. It catches errors in parsing XML files generated by tests, displays them in the results table exits with non-zero exit code as if tests had actually failed.

### DIFF
--- a/testsuite/summarize_tests.py
+++ b/testsuite/summarize_tests.py
@@ -71,10 +71,15 @@ if __name__ == "__main__":
 
     for pfile in sorted(glob.glob(os.path.join(test_outdir, "*.xml"))):
         ph_name = os.path.splitext(os.path.split(pfile)[1])[0].replace("_", " ")
-        ph_res = parse_result_file(pfile)
-        results[ph_name] = ph_res
-        for k, v in ph_res.items():
-            totals[k] += v
+        try:
+            ph_res = parse_result_file(pfile)
+            results[ph_name] = ph_res
+            for k, v in ph_res.items():
+                totals[k] += v
+        except Exception as err:
+            msg = f"ERROR: {pfile} not parsable with error {err}"
+            results[ph_name] = {"Tests": 0, "Skipped": 0, "Failures": 0, "Errors": 0, "Time": 0, "Failed tests": [msg]}
+            totals["Failed tests"].append(msg)
 
     cols = ["Tests", "Skipped", "Failures", "Errors", "Time"]
 
@@ -97,10 +102,13 @@ if __name__ == "__main__":
     print(tline)
     for pn, pr in results.items():
         print(f"{pn:<{first_col_w}s}", end="")
-        for c in cols:
-            fmt = ".1f" if c == "Time" else "d"
-            print(f"{pr[c]:{col_w}{fmt}}", end="")
-        print()
+        if pr["Tests"] == 0 and pr["Failed tests"]:
+            print(f"{'--- XML PARSING FAILURE ---':^{len(cols) * col_w}}")
+        else:
+            for c in cols:
+                fmt = ".1f" if c == "Time" else "d"
+                print(f"{pr[c]:{col_w}{fmt}}", end="")
+            print()
 
     print(tline)
     print(f"{'Total':<{first_col_w}s}", end="")
@@ -111,7 +119,8 @@ if __name__ == "__main__":
     print(tline)
     print()
 
-    if totals["Failures"] + totals["Errors"] > 0:
+    # Second condition handles xml parsing failures
+    if totals["Failures"] + totals["Errors"] > 0 or totals["Failed tests"]:
         print("THE NEST TESTSUITE DISCOVERED PROBLEMS")
         print("    The following tests failed")
         for t in totals["Failed tests"]:


heplesser:
@terhorstd @lekshmideepu Ping!

testsuite/summarize_tests.py
```diff
@@ -71,10 +71,15 @@ def parse_result_file(fname):
...
-            totals[k] += v
+        try:
+            ph_res = parse_result_file(pfile)
+            results[ph_name] = ph_res
+            for k, v in ph_res.items():
+                totals[k] += v
+        except Exception as err:
+            msg = f"ERROR: {pfile} not parsable with error {err}"
+            results[ph_name] = {"Tests": 0, "Skipped": 0, "Failures": 0, "Errors": 0, "Time": 0, "Failed tests": [msg]}
+            totals["Failed tests"].append(msg)
```
terhorstd, line 82: Why is this exception not handled inside `parse_result_file()`? If the exception can be handled as is done here, it should be handled where it is first encountered; otherwise every caller of `parse_result_file` needs to implement it. Seems the results can be handled quite symmetrically to the "normal" return values, which would be even more reason to move the `try … except` inside.

testsuite/summarize_tests.py
```diff
@@ -71,10 +71,15 @@ def parse_result_file(fname):
...
-        for k, v in ph_res.items():
-            totals[k] += v
+        try:
+            ph_res = parse_result_file(pfile)
+            results[ph_name] = ph_res
+            for k, v in ph_res.items():
+                totals[k] += v
+        except Exception as err:
+            msg = f"ERROR: {pfile} not parsable with error {err}"
+            results[ph_name] = {"Tests": 0, "Skipped": 0, "Failures": 0, "Errors": 0, "Time": 0, "Failed tests": [msg]}
```
terhorstd, line 81:
```suggestion
            results[ph_name] = {"Tests": 0, "Skipped": 0, "Failures": 0, "Errors": 1, "Time": 0, "Failed tests": [msg]}
```
If we mark the "Failed tests" entry also in the "Errors" column, we could keep `len(results["Failed tests"]) == results["Errors"] + results["Failures"]`, which seems like an easy to sustain symmetry and follows the principle of least surprise.

testsuite/summarize_tests.py
```diff
@@ -97,10 +102,13 @@ def parse_result_file(fname):
     print(tline)
     for pn, pr in results.items():
         print(f"{pn:<{first_col_w}s}", end="")
-        for c in cols:
-            fmt = ".1f" if c == "Time" else "d"
-            print(f"{pr[c]:{col_w}{fmt}}", end="")
-        print()
+        if pr["Tests"] == 0 and pr["Failed tests"]:
+            print(f"{'--- XML PARSING FAILURE ---':^{len(cols) * col_w}}")
+        else:
```
terhorstd, line 107: Not necessary if marked as in previous comment. The extra `XML PARSING FAILURE` string can be absorbed into `msg` and I don't see why centering necessary in addition to all-caps.

testsuite/summarize_tests.py
```diff
@@ -111,7 +119,8 @@ def parse_result_file(fname):
     print(tline)
     print()
 
-    if totals["Failures"] + totals["Errors"] > 0:
+    # Second condition handles xml parsing failures
+    if totals["Failures"] + totals["Errors"] > 0 or totals["Failed tests"]:
```
terhorstd, line 123:
```suggestion
    if totals["Failures"] + totals["Errors"] > 0:
```
See above.

terhorstd (changes requested):
I'd suggest some small refactoring to reduce cross-code dependencies and code complexity due to introduction of "another case".

lekshmideepu (commented):
My preference would be to have the failed and errored tests separately.

testsuite/summarize_tests.py
lekshmideepu: My preference would be to have the failed and errored tests separately.

lekshmideepu (commented):
There is a typo the link for posting the issues. LN: 130
**nest-simulator** instead of nest_simulator in the below link.
print(" https://github.com/nest/nest_simulator/issues")

heplesser:
> There is a typo the link for posting the issues. LN: 130 **nest-simulator** instead of **nest_simulator** in the below link. print(" https://github.com/nest/nest_simulator/issues")

Thanks, fixed with next push.

testsuite/summarize_tests.py
heplesser: Done with next push.

testsuite/summarize_tests.py
heplesser: When parsing of the XML file fails, we lose all information about an entire phase of the testing process. We do not even know how many tests were in this phase and could therefore be considered "Failed". Indeed, in all cases I have seen so far, the tests actually all passed, we just could not parse this because the XML file was corrupted.

Therefore, we need to handle parse failures differently and cannot include them in the other statistics.

testsuite/summarize_tests.py
heplesser: I have removed the all caps. As explained above, we need to treat this case special. And I find the centred message much more visually pleasing :).

testsuite/summarize_tests.py
heplesser: See above.

testsuite/summarize_tests.py
terhorstd: The exact number of failures is not relevant, since the type of failures determines criticality and required actions. So I think it's ok to have parsing failure recorded as "one failure". The same happens in other cases where we treat one wrong result different from any number of other errors. If there is at least one failure in a stage we need to look into it and understand what this means anyway. It's not possible to dismiss a number of failures just because the number is small.

terhorstd: @lekshmideepu, yes, that's why there are `Errors` and `Failures`. My point is that I think we do not need a third type that requires extra condition checks when interpreting the results. For me it would be intuitively clean to assert correct function of the testsuite by an
```python
assert results["Failures"] + results["Errors"] == len(results["Failed tests"])
```
Where the list of failed tests contains messages for each of the errors and failures. The level on which something went wrong is not known for each failure anyway. If a test fails, this can be a problem anywhere between external libraries, nest code or broken test or test-suite. This is not special for a parsing error.

lekshmideepu: @terhorstd Thanks for the clarification. I thought it differently. I agree with your point.

lekshmideepu (approved):
Thanks @heplesser Looks good to me 👍

testsuite/summarize_tests.py
heplesser: @terhorstd In Pytest, "error" and "failure" have specific meanings (see https://stackoverflow.com/questions/21513718/pytest-error-vs-fail): If an assertion is triggered while a `test_...()` is executed, this counts as a failure, while a failed assertion outside a `test_`, especially in fixtures, is counted as an error. The sum of passes, failures, errors and skips should always equal the number of tests in total. This should apply for each row in our table and the totals at the bottom.

If we cannot parse an XML results file, we have no information and just counting that as an error would break the logic of the PyTest result counts. Therefore, I still think that it is most sensible to treat XML parsing errors separately from PyTest result tabulation.

heplesser: @terhorstd After our further discussion I have updated the PR. Could you re-review?

terhorstd (approved):
Thanks for the thorough discussion! I think this is now a very effective and concise solution!
LGTM :+1:

testsuite/summarize_tests.py
terhorstd: Thanks for the good discussion!

PR #3188, opened 2024-04-22
Fix uploading of CI artifacts

heplesser:


.github/workflows/nestbuildmatrix.yml
```diff
@@ -848,6 +847,5 @@ jobs:
         with:
           name: "${{ matrix.NEST_BUILD_TYPE }}-build-logs-${{ matrix.os }}-${{ matrix.cpp_compiler }}"
           path: |
-            install_manifest.txt
-            **.log
-            build/reports/**
+            build/install_manifest.txt
```
lekshmideepu, line 850: Perfect. This then fixes the issue with the warning that those files were not found in the  provided path:

lekshmideepu (commented):
👍

.github/workflows/nestbuildmatrix.yml
heplesser: @lekshmideepu Github now sees your positive comment as an unresolved, merge-blocking comment. Could you resolve it and approve, unless there are other issues?

lekshmideepu (approved):
👍

terhorstd (approved):
Thanks for the fix!
LGTM :+1:

PR #3214, opened 2024-06-03
Fix corner case for node collections

### LINKED ISSUE #3213: GetConnections() triggers assertion for some node collections given as sources
The following code triggers an [assertion in `NodeCollection::get_node_index()`](https://github.com/nest/nest-simulator/blob/4ea6520834a294f0a3430d84c3c57c17f9b681d2/nestkernel/node_collection.cpp#L1340):

```python
import nest
n = nest.Create('parrot_neuron', 12)
nest.Connect(n, n)
pick = [3, 7, 9, 11]
print(n[pick])
print(nest.GetConnections(source=n[pick]))
```

heplesser:
This PR fixes #3213.

### DIFF
--- a/nestkernel/node_collection.cpp
+++ b/nestkernel/node_collection.cpp
@@ -1337,9 +1337,9 @@ NodeCollectionComposite::get_nc_index( const size_t node_id ) const
     }
   }
 
-  assert( lower == upper );
+  assert( lower >= upper );
 
-  if ( node_id < parts_[ lower ][ 0 ] or parts_[ lower ][ parts_[ lower ].size() - 1 ] < node_id )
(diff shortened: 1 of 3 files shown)


clinssen:
I confirm that this fixes the issues we were seeing in NESTML. Thank you for the quick fix!

clinssen (approved):

otcathatsya:
Looks good to me, although the test also passes without the fix (on master)

otcathatsya (approved):

clinssen:
> Looks good to me, although the test also passes without the fix (on master)

That's strange! I double-checked that the test (at least the code snippet from ``num_n = ...`` to the assert) fails for me on master and passes on this branch. Do the lines fail for you if you paste them into an ipython session, but the test passes? Or they are both passing?

PR #3225, opened 2024-06-17
Add note on parameter corrections for the AMAT2 model (Heiberg et al 2018)

heplesser:
The title says it all.

Just one reviewer since it is a small doc improvement.


jessica-mitchell (approved):
LGTM thanks

PR #3227, opened 2024-06-17
Replace aliases that have been removed with NumPy 2.0

otcathatsya:
Checked via NPY201 in ruff, no other syntax issues found in the NEST repo; will PR this separately for the INM-6 models.


heplesser (approved):

terhorstd (approved):
That looks good. We should really add the python testing matrix again, though. (Different PR).
:+1:

Issue #3228, opened 2024-06-18
NEST behaves incorrectly and potentially segfaults if number of threads changed after parameters are created


heplesser:
The test below will fail if run as is. If the number of threads is changed *before* `p = nest.random.normal(mean=0, std=1)` all works well. Otherwise, there is also a risk of a seg.fault. This applies at present to `normal` and `lognormal` parameters only and is due to the fact that these two parameter types maintain per-thread objects internally (https://github.com/nest/nest-simulator/blob/c6aacad4bb5a0685bdf8414e0bf366ed16a88a9b/nestkernel/parameter.cpp#L95, https://github.com/nest/nest-simulator/blob/c6aacad4bb5a0685bdf8414e0bf366ed16a88a9b/nestkernel/parameter.cpp#L121).

We currently have no way of detecting the existence of parameter objects. Since there objects live outside the kernel (at the Python level), they also are not subject to `ResetKernel`. Fixing this will thus require some careful thought.

Until we have a proper fix, we should add a clear warning on https://nest-simulator.readthedocs.io/en/stable/neurons/parametrization.html.

```python
import nest

p = nest.random.normal(mean=0, std=1)
nest.local_num_threads = 2

n = nest.Create('iaf_psc_exp')

num_rounds = 10
v = []
for _ in range(num_rounds):
    n.set(V_m=p)
    v.append(n.V_m)

assert len(set(v)) == num_rounds  # all v must differ
```

heplesser:
#3229 adds a warning to documentation, but does *not* provide a fix.

med-ayssar:
A version mechanism might be implemented during the simulation to continuously monitor and ensure that the kernel version remains consistent throughout its execution. Each time the kernel is updated, its version number should be incremented starting from zero. Every internal function should compare the stored version number with the actual kernel version, and in the event of a mismatch, the simulation should be stopped, with a warning issued to the user about the conflict.

This concept might applied to re-using old deleted NodeCollections too.

jessica-mitchell:
@med-ayssar @heplesser what is the status of this issue? is it still relevant?

heplesser:
Yes, it is still relevant.

med-ayssar:
@heplesser  Any concrete plans or design ideas that were discussed internally?

PR #3229, opened 2024-06-18
Add warning concerning order of parameter creation and change in number of threads

heplesser:
This PR provides a temporary patch for #3228.

Since this only adds documentation, a single reviewer suffices.


jessica-mitchell (approved):
thanks, looks good

PR #3230, opened 2024-06-19
Update VM Badge to v3.7

steffengraber:
Updated the VM a longer time ago and only now have seen that the badge also needs to be replaced.


jessica-mitchell (approved):
@steffengraber looks good
can you merge master since we have some failing tests that have been fixed?

PR #3232, opened 2024-06-19
Fix minor Sphinx warnings related to broken links 

jessica-mitchell:
This PR fixes warnings that Sphinx emits due to links or references not being correct in some way.
Biggest change is renaming EI_clustered network Python scripts so they do not collide with the names of the scripts in microcircuit (Potjans_2014).


jessica-mitchell:
Merging with no reviewers as these are minor changes to docs

PR #3233, opened 2024-06-19
Fixed conductance based neuron model header files

aoot:
[Similar conductance based neuron model](https://nest-simulator.readthedocs.io/en/stable/models/iaf_cond_exp.html) with conductance modulated by an exponential function states "conductance" instead of "current." By fixing the documentation, the conductance based neuron models are consistent so is the unit specified (i.e., nS) in respective documentations.


jessica-mitchell (approved):
thanks @aoot Looks good

PR #3241, opened 2024-06-27
Update models with adaptation mechanisms and change tag 'adaptive threshold' to 'adaptation'

jessica-mitchell:
This PR updates the model tag for models that use an adaptation mechanism to 'adaptation' after discussions with @ddahmen 
The reason is that not all the models have the same adaptation mechanism, and the ones labeled with adaptive threshold (i.e. 'aeif' models) technically do not have an adaptive threshold. 
We were also missing tags for several models that have an adaptation mechanism but were not 'aeif' models.

See Read the Docs output here: https://nest-simulator--3241.org.readthedocs.build/en/3241/models/index.html#adaptation

@shimoura agreed to review, thanks!

### DIFF
--- a/models/aeif_cond_alpha.h
+++ b/models/aeif_cond_alpha.h
@@ -68,7 +68,7 @@ extern "C" int aeif_cond_alpha_dynamics( double, const double*, double*, void* )
  */
 extern "C" int aeif_cond_alpha_dynamics_DT0( double, const double*, double*, void* );
 
-/* BeginUserDocs: neuron, integrate-and-fire, adaptive threshold, conductance-based
+/* BeginUserDocs: neuron, integrate-and-fire, adaptation, conductance-based
 
 Short description
 +++++++++++++++++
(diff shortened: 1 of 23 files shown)


shimoura (changes requested):
The ones included seem right. I noted one that is missing. The "Izhikevich" model can present spike frequency adaptation.

jessica-mitchell:
> The ones included seem right. I noted one that is missing. The "Izhikevich" model can present spike frequency adaptation.

I don't see, however, an obvious parameter that reflects an adaptive mechanism in the model code

shimoura:
> > The ones included seem right. I noted one that is missing. The "Izhikevich" model can present spike frequency adaptation.
> 
> @shimoura I don't see, however, an obvious parameter that reflects an adaptive mechanism in the model code

The model is similar to aeif category. The main difference is the dependence of membrane voltage with quadratic terms instead of exponential. And the parameters "a", "b", and "d" are in some way similar to the parameters "a" and "b" of aeif models. 
In summary, different values of these parameters will give similar behavior as aeif models. The spike frequency adaptation is one of them. Fig. 2 of the main Izhikevich reference shows it. 

Obs: I noticed in Izhikevich documentation that du/dt should be dU_m/dt to match the variable names.

shimoura (approved):

jessica-mitchell:
Merging with 1 reviewer, since it's documentation related, also discussed in person at NEST meeting, to general approval of concept

PR #3247, opened 2024-07-09
Exponential lookup tables for homogenous STDP synapses

otcathatsya:
Updated implementation of lookup tables for exponential computations in homogenous STDP synapses (original commits by @suku248 @jarsi), for now just applied to `stdp_pl_synapse_hom`. There are a few potential concerns:

- Ideally we'd be able to keep `ArchivingNode` as a common base class and support both (potentially) precise spike times as well as `size_t`-implemented grid steps, but due to the way history is stored this is impossible without costly conversion on multiple occasions. A dynamic polymorphism approach + conversions was implemented, but did considerably worse in benchmarks (albeit still better than the baseline, see figures below)

- Since we now require a separate archiving node (could be a static template impl but that also has minor overhead), we also need to duplicate the nodes that support homogenous STDP synapses; we introduced `iaf_psc_alpha_hom`. A better solution would probably require a restructuring for the way archivers are implemented, e.g. as pointers.

In the figures below, "plain" is an implementation that just replaced existing structures for testing purposes while "separate" is the current version with duplicated classes, they are both listed as a sanity check and performance should be the same. "Converted" uses a templated histentry and conversion approach in archiving node. Std and mean values taken over 3 rng seeds.
exp_tables
exp_tables_2

### DIFF
--- a/models/iaf_psc_alpha_hom.cpp
+++ b/models/iaf_psc_alpha_hom.cpp
@@ -1,0 +1,408 @@
+/*
+*  iaf_psc_alpha_hom.cpp
+*
+*  This file is part of NEST.
+*
+*  Copyright (C) 2004 The NEST Initiative
+*
+*  NEST is free software: you can redistribute it and/or modify
+*  it under the terms of the GNU General Public License as published by
+*  the Free Software Foundation, either version 2 of the License, or
+*  (at your option) any later version.
+*
+*  NEST is distributed in the hope that it will be useful,
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*  GNU General Public License for more details.
+*
+*  You should have received a copy of the GNU General Public License
+*  along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+#include "iaf_psc_alpha_hom.h"
+
+// C++ includes:
+#include <limits>
+
+// Includes from libnestutil:
+#include "dict_util.h"
+#include "exceptions.h"
+#include "iaf_propagator.h"
+#include "kernel_manager.h"
+#include "nest_impl.h"
+#include "numerics.h"
+#include "ring_buffer_impl.h"
+#include "universal_data_logger_impl.h"
+
+// Includes from sli:
+#include "dictutils.h"
+
+nest::RecordablesMap< nest::iaf_psc_alpha_hom > nest::iaf_psc_alpha_hom::recordablesMap_;
+
+namespace nest
+{
+void
+register_iaf_psc_alpha_hom( const std::string& name )
+{
+ register_node_model< iaf_psc_alpha_hom >( name );
+}
+
+
+/*
+* Override the create() method with one call to RecordablesMap::insert_()
+* for each quantity to be recorded.
+*/
+template <>
+void
+RecordablesMap< iaf_psc_alpha_hom >::create()
+{
+ // use standard names wherever you can for consistency!
+ insert_( names::V_m, &iaf_psc_alpha_hom::get_V_m_ );
+ insert_( names::I_syn_ex, &iaf_psc_alpha_hom::get_I_syn_ex_ );
+ insert_( names::I_syn_in, &iaf_psc_alpha_hom::get_I_syn_in_ );
+}
+
+/* ----------------------------------------------------------------
+* Default constructors defining default parameters and state
+* ---------------------------------------------------------------- */
+
+iaf_psc_alpha_hom::Parameters_::Parameters_()
+ : Tau_( 10.0 )             // ms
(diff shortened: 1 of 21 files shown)


models/iaf_psc_alpha_hom.cpp
```diff
@@ -0,0 +1,408 @@
+/*
```
clinssen, line 1: Perhaps this could be named differently from "hom[ogeneous]", like "lookup" or "fast"? Please indicate in the model docstring how this model is different from iaf_psc_alpha.

nestkernel/archiving_node_hom.h
```diff
@@ -0,0 +1,156 @@
...
+#include "dictdatum.h"
+
+#define DEBUG_ARCHIVER 1
+
+namespace nest
+{
+
+constexpr double TAU_MINUS_PLACEHOLDER = 1.0; // indicate that tau minus has not been set yet
+
+/**
```
clinssen, line 47: Please add documentation here about the fixed time step based lookup table.

nestkernel/archiving_node_hom.h
```diff
@@ -0,0 +1,156 @@
...
+#include "histentry.h"
+#include "nest_time.h"
+#include "nest_types.h"
+#include "node.h"
+#include "structural_plasticity_node.h"
+
+// Includes from sli:
+#include "dictdatum.h"
+
+#define DEBUG_ARCHIVER 1
```
clinssen, line 40: Should this be 1?

nestkernel/histentry.h
```diff
@@ -43,6 +43,18 @@ class histentry
   size_t access_counter_; //!< access counter to enable removal of the entry, once all neurons read it
 };
 
+class histentry_step
```
clinssen, line 46: please add a docstring describing how this is different from ``histentry``

nestkernel/node.h
```diff
@@ -479,6 +479,8 @@ class Node
    * @throws IllegalConnection
    *
    */
+  virtual void register_stdp_connection( size_t, size_t, const double tau_minus );
```
clinssen, line 482: please add a detailed docstring

nestkernel/node.h
```diff
@@ -761,7 +763,9 @@ class Node
    * return the Kminus value at t (in ms).
    * @throws UnexpectedEvent
    */
-  virtual double get_K_value( double t );
+  virtual double get_K_value( long t, size_t& dt_steps );
```
clinssen, line 766: please add a detailed docstring

nestkernel/node.h
```diff
@@ -777,6 +781,11 @@ class Node
    * return the spike history for (t1,t2].
    * @throws UnexpectedEvent
    */
+  virtual void get_history( long t1,
```
clinssen, line 784: please add a detailed docstring

pynest/examples/hpc_benchmark.py
```diff
@@ -224,10 +224,10 @@ def build_network(logger):
     nest.overwrite_files = True
 
     nest.message(M_INFO, "build_network", "Creating excitatory population.")
-    E_neurons = nest.Create("iaf_psc_alpha", NE, params=model_params)
+    E_neurons = nest.Create("iaf_psc_alpha_hom", NE, params=model_params)
```
clinssen, line 227: should it be a user-configurable option which version of the model to select? Should the default remain the same (as the changed proposed in this PR is a numerical approximation for the purpose of runtime performance benefit)?

testsuite/pytests/test_stdp_pl_synapse_hom.py
```diff
@@ -18,12 +18,12 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
-
+import unittest
```
clinssen, line 21: Could you please use pytest instead? ``unittest`` is deprecated.

testsuite/regressiontests/ticket-689.sli
```diff
@@ -42,8 +42,8 @@ M_ERROR setverbosity
   /total_num_virtual_procs 1
 >> SetKernelStatus
 
-/iaf_psc_alpha Create /n1 Set
-/iaf_psc_alpha Create /n2 Set
+/iaf_psc_alpha_hom Create /n1 Set
```
clinssen, line 45: Should this remain the original? Should the test for the "hom" version be added rather than replacing the original?

clinssen (changes requested):

terhorstd:
Hi @otcathatsya, what's your status on this?

heplesser:
@JanVogelsang Ping for review!
@otcathatsya Relabeld to "Normal" because it is an enhancement, not a bug or blocker.

terhorstd:
@otcathatsya, kind _ping_ for updates on this matter…

PR #3249, opened 2024-07-12
Remove pointless initialisation of integration step size

clinssen:
<s>Add note about maximum enforced timestep in all aeif models.</s>

Remove pointless initialisation of integration step size to max 0.01 ms (see discussion below).

### DIFF
--- a/models/aeif_cond_alpha.cpp
+++ b/models/aeif_cond_alpha.cpp
@@ -384,9 +384,9 @@ nest::aeif_cond_alpha::init_buffers_()
   B_.logger_.reset();
 
   B_.step_ = Time::get_resolution().get_ms();
-
-  // We must integrate this model with high-precision to obtain decent results
-  B_.IntegrationStep_ = std::min( 0.01, B_.step_ );
+  B_.IntegrationStep_ =
+    B_.step_; // reasonable initial value for numerical integrator step size; this will anyway be overwritten by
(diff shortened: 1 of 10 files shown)


terhorstd:
@clinssen, could you please provide a description for this PR?

models/aeif_cond_alpha.h
```diff
@@ -103,6 +103,11 @@ and
 For the reference implementation of this model, see
 `aeif_models_implementation <../model_details/aeif_models_implementation.ipynb>`_ notebook.
 
+.. note::
+
+   A maximum integration step size of 10 μs is enforced, regardless of the
+   resolution set in the NEST kernel.
```
heplesser, line 109: This is not correct. `init_buffers_()` initialises `B_.IntegrationStep_` to `min(0.01, Time::get_resolution().get_ms())`, but after that the adaptive stepsize mechanism controls `IntegrationStep_`. The initialization line has been unchanged since 2.6.1 at least, and I am not sure why we have it at all, given that we are using an adaptive stepsize solver. It might be a leftover from very early versions with a fixed-stepsize solver.

heplesser (changes requested):
@clinssen I think in this case the code is actually wrong/outdated. We us an adaptive-stepsize solver, so we should not need to control the integration time step manually. See also inline comment. I'd suggest to change the code instead (but check for consequences).

heplesser (approved):
Thanks for tidying up!

otcathatsya (approved):

PR #3257, opened 2024-07-22
Add more minimal CI test

terhorstd:
This change adds one more build to the CI/CD chain to test for the Boost dependency.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -524,6 +524,7 @@ jobs:
         # available use flags (all default to "OFF"):
         # openmp, mpi, python, gsl, ltdl, boost, sionlib, libneurosim, optimize, warning, userdoc, music
         use:
+          - "optimize, warning"
           - "boost, optimize, warning"
           - "openmp, python, gsl, ltdl, boost, optimize, warning"
           - "mpi, python, gsl, ltdl, boost, optimize, warning"


terhorstd:
It seems difficult to uninstall the boost headers due to dependencies of other packages. Important would be a proof that the CI fails when #3258 is reverted. Do you have an idea how to best go about that? :thinking:

terhorstd:
This seems good to go once the break is reverted. With the commit https://github.com/nest/nest-simulator/pull/3257/commits/243449d35f5c154a90fcf54e5441380d94d34a94 correctly being flagged broken by the CI:
```
Error: /home/runner/work/nest-simulator/nest-simulator/libnestutil/iterator_pair.h:26:10: fatal error: boost/iterator/iterator_adaptor.hpp: No such file or directory
   26 | #include <boost/iterator/iterator_adaptor.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
:warning: This is however only partial protection as we have seen in https://github.com/nest/nest-simulator/pull/3257/commits/ff5dbc3161c8c892b5e5553d5b9f500e5ca6c516 not being flagged. This additional build can only protect against `boost`-includes in parts of the code that is not include-guarded by a combination of other dependencies (such as GSL in this case). To protect against this we would need to duplicate the whole test matrix. :zap: 

@gtrensch, @lekshmideepu, could you have another look?

gtrensch:
@terhorstd, all builds and checks have passed! Depending on CMake option '-Dwith-mpi=on/off' and the availability of 'mpi4py', MPI tests are executed or skipped correctly.

gtrensch:
@terhorstd, many thanks for your kind comments. I agree completely. The test suite should handle this at a higher level.

gtrensch (approved):
LGTM 👍

terhorstd:
Thanks for the very enlightening process and inspiring reviews :+1:

Issue #3217, opened 2024-06-07
nest as a SPEC CPU benchmark


heshpdx:
Hello friends,

I’m a CPU architect at Ampere Computing where I do performance analysis and workload characterization. I also serve on the [SPEC CPU](https://spec.org/cpu2017/) committee, working on benchmarks for the next version of SPEC CPU, [CPUv8](https://spec.org/cpuv8/) . We try to find computationally intensive workloads in diverse fields, to help measure performance across a wide variety of behaviors and application domains. Based on the longevity of nest, its large active community in biology and its use in education, I have proposed the nest neural network model be included in the next set of marquee benchmarks in SPEC CPU. 

As part of the effort, we have ported and integrated the nest mainline code into the SPEC CPU harness so that it can be tested on a wide variety of systems in a controlled environment to produce reproducible results. We have even built it on native Windows using MSVC and the Intel compiler for Windows – we are happy to share the changes if someone is interested in testing and integrating it back into the upstream mainline for the benefit of the community.

The piece we need help with is an understanding of the multithreaded workloads. Right now, we have *single-threaded* nest command lines which run and produce verifiable output across many compilers (llvm, gcc, icc, aocc, nvhpc, cray), ISAs (aarch64, x86, power) and operating systems (linux, windows, android). We verify the run via checking the .dat files which come out of the simulation runs to make sure that there are no differences in the resulting output. A problem arises when we run with multiple threads, since there are a different number of files produced, and I am unfamiliar with how to coalesce them to verify.

First some fundamental questions: Does a nest invocation with 8 threads perform the same amount of work as a run with 16 threads? Or is it that the problem being solved is larger? If it is the same, how can we verify that?  Does this answer change based on the .sli script used?

In the example below, if I run `examples/nest/brunel-2000_newconnect_dc.sli` (with a small edit to make it run longer)... I tried with 8 threads and 16. It looks like I am simulating the same number of Neurons and Synapses. The 8-thread version outputs 16 files and the 16-threaded version outputs 24 files. The total lines in the files are close. Do they fundamentally contain the same information, just at different sample points?

```
$ ./nest_s_base.O3-64 --userargs=threads=8 brunel-2000_newconnect_dc_LONG.sli
NEST 3.5.0-post0.dev0 (C) 2004 The NEST Initiative
Configuring neuron parameters.
Creating the network.
Configuring neuron parameters.
Creating excitatory spike recorder.
Creating inhibitory spike recorder.
Connecting excitatory neurons.
[...]

PR #3272, opened 2024-08-05
Fix memory safety issue in slistartup.cc

heshpdx:
Also fixes the memory leak in parser.h

As referred to in https://github.com/nest/nest-simulator/issues/3217#issuecomment-2269949314

### DIFF
--- a/sli/parser.h
+++ b/sli/parser.h
@@ -61,7 +61,10 @@ class Parser
 public:
   Parser();
   Parser( std::istream& );
-  ~Parser() { delete s; }
+  ~Parser()
+  {
+    delete s;
+  }
 
   bool operator()( Token& );
   bool
--- a/sli/parser.h
+++ b/sli/parser.h
@@ -61,7 +61,7 @@ class Parser
 public:
   Parser();
   Parser( std::istream& );
-  ~Parser() { if (s) delete s; }
+  ~Parser() { delete s; }
 
   bool operator()( Token& );
   bool
--- a/sli/parser.h
+++ b/sli/parser.h
@@ -38,7 +38,7 @@
 
 class Parser
 {
-  Scanner* s;
+  Scanner* s = nullptr;
 
   Token arraytoken;
   Token proctoken;
@@ -61,6 +61,7 @@ class Parser
 public:
   Parser();
   Parser( std::istream& );
+  ~Parser() { if (s) delete s; }
 
   bool operator()( Token& );
   bool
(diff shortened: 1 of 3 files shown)


sli/parser.h
```diff
@@ -61,6 +61,7 @@ class Parser
 public:
   Parser();
   Parser( std::istream& );
+  ~Parser() { if (s) delete s; }
```
heplesser, line 64: According to *The C++ Programming Language*, 4th ed, §11.2, "[a]pplying `delete` to the `nullptr` has no effect." So the `if` is not needed here:
```suggestion
  ~Parser() { delete s; }
```

In general, we require braces around single-line blocks (as a consequence of the OpenSSH goto-goto bug some years back).

sli/slistartup.cc
```diff
@@ -214,53 +214,55 @@ SLIStartup::SLIStartup( int argc, char** argv )
     StringDatum* sd = new StringDatum( argv[ i ] );
     args_array.push_back( Token( sd ) );
 
-    if ( *sd == "-d" or *sd == "--debug" )
+    std::string myarg = argv[i];
```
heplesser, line 217: Our general policy is to declare `const` for anything that does not change for clarity. We also require whitespace inside parentheses and brackets:
```suggestion
    const std::string myarg = argv[ i ];
```

heplesser (changes requested):
Hi @heshpdx! Thank you for your pull request. I have added a few suggestions to get in line with NEST coding standards and solve the clang-format failure.

sli/parser.h
```diff
@@ -61,6 +61,7 @@ class Parser
 public:
   Parser();
   Parser( std::istream& );
+  ~Parser() { delete s; }
```
heplesser, line 64: There is another little formatting problem which I had overlooked yesterday. This should fix it:
```suggestion
  ~Parser()
  {
    delete s;
  }
```  
BTW, we have set up pre-commit checks for NEST to make it easier for contributors to submit code in accordance to our formatting guidelines, see https://nest-simulator.readthedocs.io/en/stable/developer_space/guidelines/coding_guidelines_check.html#tooling.

heshpdx: Can you share why line 60 formatting is ok? I matched the destructor formatting to the constructor formatting.

heplesser: The destructor is defined inline, while the constructor is only declared in line 63, so there is no body, and the formatting problem related to the method body.

heplesser (approved):
All good now!

heplesser:
Small fix in SLI parser plus small syntax cleanup, this merging with single review.

PR #3284, opened 2024-08-08
Configure Sphinx gallery to copy files from `pynest/examples`

### LINKED ISSUE #3246: Provide extra files needed to run examples on Read the Docs
Some PyNEST examples have extra files that are needed to run; they are available in the repo but it's not obvious from Read the Docs how to access them.

Suggestion:
Provide a download link to these files from Read the Docs

jessica-mitchell:
This PR adjusts the configuration for Sphinx gallery so it now copies additional files from `pynest/examples`.
This makes it easier to keep all files related to the example together and link to them appropriately.
This means the function for copying hpc_benchmark_connectivity.svg can be removed from `conf.py`

This also fixes #3246  

**Note** however, files (txt, images etc) required for Python examples, need to be somehow linked to in the appropriate script. For eprop handwriting text file - it is added as a `:download:` link.
In the case of Sonata, there is an additional folder with many types of files. These are not copied and a direct link to the repo needs to be used. Sphinx gallery does not support >1 subfolders. (see [mention here](https://sphinx-gallery.github.io/stable/configuration.html#manage-multiple-galleries))
Thus, when creating Python scripts and additional data or image files need to be used, care must be taken that links are included.


steffengraber (approved):
:+1:

PR #3285, opened 2024-08-08
Update boost detection and issue warning message

### LINKED ISSUE #3261: Add warning for missing Boost similar to missing GSL
When configuring a source build without GSL a visible paragraph is printed after the "Configuration Summary" that highlights consequences for the availability of models.
```
ATTENTION!
You are about to compile NEST without the GNU Scientific
Library or your GSL is too old (before v1.11). This means
that conductance-based neuron models and some random number
generators will not be available.
```
Currently it is possible to compile `-Dwith-boost=OFF`, but that also has *significant* consequences for performance and availability of models.

The consequences of compiling without Boost should be clearly noted in the configuration summary (and explained in the Build docs).

**Additional context**
For some deployments it may be desirable to have a minimal install. Removing library dependencies may however cause the NEST kernel to switch to *much slower* local implementations internally (without any warnings at run-time) or unexpectedly make many models or features unavailable.

### LINKED ISSUE #3283: CMake Boost detection changed
Per CMake 3.30, the way Boost is detected is changed. Our CMake file needs to be updated according to the suggestion below.

See also #3261.

```
CMP0167
-------

.. versionadded:: 3.30

The ``FindBoost`` module is removed.

CMake 3.29 and below provide a ``FindBoost`` module, but it needs constant
updates to keep up with upstream Boost releases.  Upstream Boost 1.70
and above provide a ``BoostConfig.cmake`` package configuration file.
``find_package(Boost CONFIG)`` finds the upstream package directly,
without the find module.

CMake 3.30 and above prefer to not provide the ``FindBoost`` module
so that ``find_package(Boost)`` calls, without the ``CONFIG`` or
``NO_MODULE`` options, find the upstream ``BoostConfig.cmake`` directly.
This policy provides compatibility for projects that have not been ported
to use the upstream Boost package.

The ``OLD`` behavior of this policy is for ``find_package(Boost)`` to
load CMake's ``FindBoost`` module.  The ``NEW`` behavior is for
``find_package(Boost)`` to search for the upstream ``BoostConfig.cmake``.

This policy was introduced in CMake version 3.30.
It may be set by ``cmake_policy()`` or ``cmake_minimum_required()``.
If it is not set, CMake warns, and uses ``OLD`` behavior.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

heplesser:
This PR fixes #3283 and fixes #3261. 

- Boost detection now uses the "CONFIG" scheme in CMake, see https://cmake.org/cmake/help/latest/policy/CMP0167.html and https://cmake.org/cmake/help/latest/command/find_package.html
- This also hikes Boost requirements from 1.69 to 1.70 because the CONFIG scheme for Boost cam with 1.70. This seems unproblematic, as 1.70 was released in April 2019 and current Boost is 1.85.
- The new approach only reports an error while CMake executes, positive Boost results are only presented in the summary. This seems to be a property of the CONFIG scheme.
- A warning is printed if Boost is not found.
- I also simplified version checking for GSL.


gtrensch (approved):
Thank you for this contribution! Looks good to me. 👍

terhorstd (approved):
LGTM :+1:

PR #3296, opened 2024-08-21
Add note to `aeif` models about refractory period and peculiarities

### LINKED ISSUE #3088: aeif-model documentation does not point out peculiarities
The current documentation for the aeif models does not point out two peculiarities of these models:

1. The default refractory period to these models is *zero*. This is consistent with the model definition in Brette & Gerstner, but most likely unexpected, since NEST models generally have a non-zero refractory period.
2. Because NEST uses an adaptive stepsize solver for the aeif ODEs with spike detection *inside* the update loop, aeif neurons can fire multiple spikes in a single time step if the refractory period is zero (the default). This can lead to exploding spike numbers and extreme slow-down of simulations.

Documentation should point this out clearly in note boxes and recommend users to set `t_ref > 0`.

jessica-mitchell:
Fixes #3088

This PR adds a note to all `aeif` models, including the reference if needed based on the comment in #3088. 
The technical reason was left out, since the focus should be on usage imo. 

See output from Read the Docs here: https://nest-simulator--3296.org.readthedocs.build/en/3296/models/aeif_psc_exp.html


clinssen:
The formulation "The default refractory period [...] is zero, [...]. This means if the refractory period is zero, [...]" is perhaps a little bit of a non sequitur. Suggest to remove "This means" from the second sentence.

models/aeif_cond_alpha.h
```diff
@@ -103,7 +103,15 @@ and
...
 `aeif_models_implementation <../model_details/aeif_models_implementation.ipynb>`_ notebook.
 
-See also [1]_.
+
+.. note::
+
+    The default refractory period for ``aeif`` models is zero, consistent with the model definition in
+    Brette & Gerstner [1]_.
+    This means if the refractory period is zero, the ``aeif`` neurons can fire multiple spikes in a single time step,
+    which can lead to exploding spike numbers and extreme slow-down of simulations.
```
heplesser, line 112: Based on Charl's comment, I suggest the following (applies to all other models as well):
```suggestion
    The default refractory period for ``aeif`` models is zero, consistent with the model definition in
    Brette & Gerstner [1]_.  Thus, an ``aeif`` neuron with default parameters can fire multiple spikes in a single time step,
    which can lead to exploding spike numbers and extreme slow-down of simulations.
```

models/aeif_cond_alpha.h
```diff
@@ -103,7 +103,15 @@ and
...
+
+    We recommend users set ``t_ref > 0``.
```
heplesser, line 114:
```suggestion
    To avoid such unphysiological behavior, you should set a refractory time ``t_ref > 0``.
```

heplesser (changes requested):

models/aeif_cond_alpha.h
jessica-mitchell: changes applied to all models

models/aeif_cond_alpha.h
jessica-mitchell: changes applied to all models

heplesser (approved):

clinssen (approved):

PR #3299, opened 2024-08-26
Allow passing OpenMP root directory to -Dwith-openmp

heplesser:
This PR brings the handling of `-Dwith-openmp` in line with MPI and other libraries according to modern CMake practice, so that the root directory of the `libomp` installation can be passed as an argument. In existing code, and argument different from `ON` or `OFF` is actually handled incorrectly. Also update related variable names to include lowercase letters, as is modern CMake practice, see [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html).

Furthermore, in the cmake summary, results for threading and MPI are now reported next to each other early in the report.


gtrensch (approved):
Looks good to me! Thanks for this contribution.

heplesser:
@JoshuaBoettcher  I have now added you as a reviewer and am looking forward to your comments.

cmake/ProcessOptions.cmake
```diff
@@ -384,20 +384,17 @@ endfunction()
...
-      printInfo( "Set OpenMP argument: ${with-openmp}")
-      # set variables in this scope
-      set( OPENMP_FOUND ON )
-      set( OpenMP_C_FLAGS "${with-openmp}" )
-      set( OpenMP_CXX_FLAGS "${with-openmp}" )
-      set( OpenMP_CXX_LIBRARIES "${with-openmp}" )
-    else ()
-      find_package( OpenMP )
+  if ( NOT "${with-openmp}" STREQUAL "OFF" )
+    if ( NOT ${with-openmp} STREQUAL "ON" )
```
pnbabu, line 388: Should `${with-openmp}` be with quotes `"${with-openmp}"`?

heplesser: Thanks for catching this, fixed.

pnbabu (approved):
Looks good to me!

PR #3304, opened 2024-09-03
Optimize computation in GrowthCurveGaussian

heshpdx:
Improves one of the hottest functions be reducing unnecessary computation. Removes one sqrt and three divs outside the loop, and two divs inside the loop. I measured about 10% speedup on my machine, depending on the workload. 

- Replace FP divides with FP multiplies. Most systems have divides slower than multiplies.
- Pre-compute constant `1/(2*sqrt(ln(2)))` instead of calculating it each time the function is called.

Can you please check my math? For some reason the spike recorder on structural_plasticity_benchmark.sli seems to be giving different output. My other workloads still produce the same results. Thanks!

### DIFF
--- a/nestkernel/growth_curve.cpp
+++ b/nestkernel/growth_curve.cpp
@@ -32,6 +32,9 @@
 // Includes from sli:
 #include "dictutils.h"
 
+// 1.0 / (2.0 * sqrt( log( 2.0 ) ))
+#define INV_TWO_SQRT_LOG_TWO 0.6005612043932248974
+
 /* ----------------------------------------------------------------
  * GrowthCurveLinear
  * ---------------------------------------------------------------- */
@@ -106,16 +109,17 @@ nest::GrowthCurveGaussian::update( double t,
   // Numerical integration from t_minus to t
   // use standard forward Euler numerics
   const double h = Time::get_resolution().get_ms();
-  const double zeta = ( eta_ - eps_ ) / ( 2.0 * sqrt( log( 2.0 ) ) );
-  const double xi = ( eta_ + eps_ ) / 2.0;
+  const double inv_zeta = ( eta_ - eps_ ) * INV_TWO_SQRT_LOG_TWO;
+  const double xi = ( eta_ + eps_ ) * 0.5;
+  const double inv_tau_Ca = 1.0 / tau_Ca;
 
   double z_value = z_minus;
   double Ca = Ca_minus;
 
-  for ( double lag = t_minus; lag < ( t - h / 2.0 ); lag += h )
+  for ( double lag = t_minus; lag < ( t - h * 0.5 ); lag += h )
   {
-    Ca = Ca - ( ( Ca / tau_Ca ) * h );
-    const double dz = h * growth_rate * ( 2.0 * exp( -pow( ( Ca - xi ) / zeta, 2 ) ) - 1.0 );
+    Ca = Ca - ( ( Ca * inv_tau_Ca ) * h );
+    const double dz = h * growth_rate * ( 2.0 * exp( -pow( ( Ca - xi ) * inv_zeta, 2 ) ) - 1.0 );
     z_value = z_value + dz;
   }
 


nestkernel/growth_curve.cpp
```diff
@@ -106,16 +109,17 @@ nest::GrowthCurveGaussian::update( double t,
   // Numerical integration from t_minus to t
   // use standard forward Euler numerics
   const double h = Time::get_resolution().get_ms();
-  const double zeta = ( eta_ - eps_ ) / ( 2.0 * sqrt( log( 2.0 ) ) );
-  const double xi = ( eta_ + eps_ ) / 2.0;
+  const double inv_zeta = ( eta_ - eps_ ) * INV_TWO_SQRT_LOG_TWO;
+  const double xi = ( eta_ + eps_ ) * 0.5;
+  const double inv_tau_Ca = 1.0 / tau_Ca;
```
heplesser, line 114: Since `eta_` and `eps_` can only be changed via the `set()` method, I'd suggest to make these three variables data members of the class, create a private method to compute their values and call that method from the constructor and the `set()` method.

And `inv_zeta` needs to be `2 * SQRT_LOG_TWO / ( eta_ - eps_ )` to make the math right (see failing test at https://github.com/nest/nest-simulator/actions/runs/10678447769/job/29595542950?pr=3304#step:23:7813). I'd suggest to keep the factor 2 out of the constant since it is cheap to compute and makes the constant more generally applicable.

nestkernel/growth_curve.cpp
```diff
@@ -106,16 +109,17 @@ nest::GrowthCurveGaussian::update( double t,
...
   double z_value = z_minus;
   double Ca = Ca_minus;
 
-  for ( double lag = t_minus; lag < ( t - h / 2.0 ); lag += h )
+  for ( double lag = t_minus; lag < ( t - h * 0.5 ); lag += h )
   {
-    Ca = Ca - ( ( Ca / tau_Ca ) * h );
-    const double dz = h * growth_rate * ( 2.0 * exp( -pow( ( Ca - xi ) / zeta, 2 ) ) - 1.0 );
+    Ca = Ca - ( ( Ca * inv_tau_Ca ) * h );
+    const double dz = h * growth_rate * ( 2.0 * exp( -pow( ( Ca - xi ) * inv_zeta, 2 ) ) - 1.0 );
```
heplesser, line 122: Since you are touching this line, let's bring it up to NEST standard which ask for `std::` for everything from the standard library, even if not strictly needed due to C-heritage:
```suggestion
    const double dz = h * growth_rate * ( 2.0 * std::exp( -std::pow( ( Ca - xi ) * inv_zeta, 2 ) ) - 1.0 );
```

nestkernel/growth_curve.cpp
```diff
@@ -106,16 +109,17 @@ nest::GrowthCurveGaussian::update( double t,
...
     z_value = z_value + dz;
```
heplesser, line 123: And this might be improved to
```suggestion
    z_value += dz;
```
to be even more expressive.

heplesser (changes requested):
@heshpdx Thanks for spotting this. I added some comments pertaining to maths and style.

nestkernel/growth_curve.cpp
```diff
@@ -32,6 +32,9 @@
 // Includes from sli:
 #include "dictutils.h"
 
+// 1.0 / (2.0 * sqrt( log( 2.0 ) ))
+#define INV_TWO_SQRT_LOG_TWO 0.6005612043932248974
```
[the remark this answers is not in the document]
heshpdx: Thank you for this, I will work on it. I think computing once per run is the right solution.

nestkernel/growth_curve.cpp
heshpdx: Completed, thank you. Let's see if the CI tests pass. My benchmarks do verify again now.

nestkernel/growth_curve.cpp
```diff
@@ -78,6 +78,7 @@ nest::GrowthCurveGaussian::GrowthCurveGaussian()
   , eta_( 0.1 )
   , eps_( 0.7 )
 {
+  compute_local();
```
heplesser, line 81:
```suggestion
  compute_local_();
```

nestkernel/growth_curve.cpp
```diff
@@ -93,6 +94,7 @@ nest::GrowthCurveGaussian::set( const DictionaryDatum& d )
 {
   updateValue< double >( d, names::eps, eps_ );
   updateValue< double >( d, names::eta, eta_ );
+  compute_local();
```
heplesser, line 97:
```suggestion
  compute_local_();
```

nestkernel/growth_curve.cpp
```diff
@@ -106,22 +108,30 @@ nest::GrowthCurveGaussian::update( double t,
...
+    Ca = Ca - ( ( Ca * inv_tau_Ca ) * h );
+    const double dz = h * growth_rate * ( 2.0 * std::exp( -std::pow( ( Ca - xi ) * inv_zeta, 2 ) ) - 1.0 );
+    z_value += dz;
   }
 
   return std::max( z_value, 0.0 );
 }
 
+void
+nest::GrowthCurveGaussian::compute_local()
```
heplesser, line 127:
```suggestion
nest::GrowthCurveGaussian::compute_local_()
```

nestkernel/growth_curve.h
```diff
@@ -227,8 +227,12 @@ class GrowthCurveGaussian : public GrowthCurve
   update( double t, double t_minus, double Ca_minus, double z, double tau_Ca, double growth_rate ) const override;
 
 private:
+  void compute_local();
```
heplesser, line 230: Private attributes should have names ending with an underscore:
```suggestion
  void compute_local_();
```

heplesser (commented):
@heshpdx Thanks for the update! It looks good now, I just have a tiny suggestion relating to naming, see below.

heplesser (approved):
Thanks!

otcathatsya (approved):

PR #3305, opened 2024-09-03
Optimize computation in iaf_psc_alpha_ps model

heshpdx:
Replace FP divides with FP multiplies and pre-compute loop invariants. Gives about +10% performance. Similar to https://github.com/nest/nest-simulator/pull/3304.

### DIFF
--- a/models/iaf_psc_alpha_ps.cpp
+++ b/models/iaf_psc_alpha_ps.cpp
@@ -274,8 +274,8 @@ nest::iaf_psc_alpha_ps::pre_run_hook()
   V_.inv_tau_syn_ex_ = 1.0 / P_.tau_syn_ex_;
   V_.inv_tau_syn_in_ = 1.0 / P_.tau_syn_in_;
 
-  V_.psc_norm_ex_ = 1.0 * numerics::e * V_.inv_tau_syn_ex_;
-  V_.psc_norm_in_ = 1.0 * numerics::e * V_.inv_tau_syn_in_;
+  V_.psc_norm_ex_ = numerics::e * V_.inv_tau_syn_ex_;
+  V_.psc_norm_in_ = numerics::e * V_.inv_tau_syn_in_;
 
   // pre-compute matrix for full time step
   V_.expm1_tau_m_ = numerics::expm1( -V_.h_ms_ * V_.inv_tau_m_ );
--- a/models/iaf_psc_alpha_ps.cpp
+++ b/models/iaf_psc_alpha_ps.cpp
@@ -296,7 +296,6 @@ nest::iaf_psc_alpha_ps::pre_run_hook()
   V_.refractory_steps_ = Time( Time::ms( P_.t_ref_ ) ).get_steps();
   // since t_ref_ >= sim step size, this can only fail in error
   assert( V_.refractory_steps_ >= 1 );
-
 }
 
 /* ----------------------------------------------------------------
--- a/models/iaf_psc_alpha_ps.cpp
+++ b/models/iaf_psc_alpha_ps.cpp
@@ -85,6 +85,7 @@ nest::iaf_psc_alpha_ps::Parameters_::Parameters_()
   , U_min_( -std::numeric_limits< double >::infinity() ) // mV
   , U_reset_( -70.0 - E_L_ )                             // mV, rel to E_L_
 {
+  compute_local_();
 }
 
 nest::iaf_psc_alpha_ps::State_::State_()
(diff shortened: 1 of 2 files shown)


clinssen:
I'm confused about why this would lead to a ~10% improvement in performance; all of the constants here are only computed prior to simulation, not in the main simulation loop (as opposed to #3304).

models/iaf_psc_alpha_ps.cpp
```diff
@@ -290,6 +296,7 @@ nest::iaf_psc_alpha_ps::pre_run_hook()
   V_.refractory_steps_ = Time( Time::ms( P_.t_ref_ ) ).get_steps();
   // since t_ref_ >= sim step size, this can only fail in error
   assert( V_.refractory_steps_ >= 1 );
+
```
heshpdx, line 299:
```suggestion
```

heplesser (approved):
@heshpdx Thanks, all good now! One CI runner failed because a dependency could not be downloaded (SIONLIB). I have restarted the runner now.

models/iaf_psc_alpha_ps.cpp
```diff
@@ -268,15 +268,21 @@ nest::iaf_psc_alpha_ps::pre_run_hook()
...
 
-  V_.psc_norm_ex_ = 1.0 * numerics::e / P_.tau_syn_ex_;
-  V_.psc_norm_in_ = 1.0 * numerics::e / P_.tau_syn_in_;
+  // pre-compute inverse member variables
+  V_.inv_c_m_ = 1.0 / P_.c_m_;
+  V_.inv_tau_m_ = 1.0 / P_.tau_m_;
+  V_.inv_tau_syn_ex_ = 1.0 / P_.tau_syn_ex_;
+  V_.inv_tau_syn_in_ = 1.0 / P_.tau_syn_in_;
+
+  V_.psc_norm_ex_ = 1.0 * numerics::e * V_.inv_tau_syn_ex_;
```
clinssen, line 277: Why multiply by 1?

models/iaf_psc_alpha_ps.cpp
```diff
@@ -268,15 +268,21 @@ nest::iaf_psc_alpha_ps::pre_run_hook()
...
+  V_.psc_norm_in_ = 1.0 * numerics::e * V_.inv_tau_syn_in_;
 
   // pre-compute matrix for full time step
-  V_.expm1_tau_m_ = numerics::expm1( -V_.h_ms_ / P_.tau_m_ );
-  V_.exp_tau_syn_ex_ = std::exp( -V_.h_ms_ / P_.tau_syn_ex_ );
-  V_.exp_tau_syn_in_ = std::exp( -V_.h_ms_ / P_.tau_syn_in_ );
+  V_.expm1_tau_m_ = numerics::expm1( -V_.h_ms_ * V_.inv_tau_m_ );
```
clinssen, line 281: In floating point math, foo/bar is of course not always precisely equivalent to foo*(1/bar), although it's probably close enough to not matter in practice. Most/all of the performance gain in this PR is anyway coming from the changes in propagate() and threshold_distance() below.

clinssen (changes requested):

models/iaf_psc_alpha_ps.cpp
heshpdx: I was going with the existing style. I didn't change the 1.0 thing that was already there.

models/iaf_psc_alpha_ps.cpp
```diff
@@ -268,15 +268,21 @@ nest::iaf_psc_alpha_ps::pre_run_hook()
...
-  V_.psc_norm_ex_ = 1.0 * numerics::e / P_.tau_syn_ex_;
-  V_.psc_norm_in_ = 1.0 * numerics::e / P_.tau_syn_in_;
+  // pre-compute inverse member variables
+  V_.inv_c_m_ = 1.0 / P_.c_m_;
+  V_.inv_tau_m_ = 1.0 / P_.tau_m_;
+  V_.inv_tau_syn_ex_ = 1.0 / P_.tau_syn_ex_;
+  V_.inv_tau_syn_in_ = 1.0 / P_.tau_syn_in_;
+
+  V_.psc_norm_ex_ = 1.0 * numerics::e * V_.inv_tau_syn_ex_;
+  V_.psc_norm_in_ = 1.0 * numerics::e * V_.inv_tau_syn_in_;
```
heshpdx, line 278:
```suggestion
  V_.psc_norm_ex_ = numerics::e * V_.inv_tau_syn_ex_;
  V_.psc_norm_in_ = numerics::e * V_.inv_tau_syn_in_;
```

models/iaf_psc_alpha_ps.cpp
heshpdx: I reverted this code. Now the only changes are the ones which are significant for performance.

heshpdx:
@clinssen anything left for me to do here?

clinssen:
Hi, thanks for the updates, this looks great and I'm giving it my thumbs up!

clinssen (approved):

PR #3308, opened 2024-09-09
Remove code for `dry_run` mode

jessica-mitchell:
The dry_run mode has not been used for several releases, thus removing as it will not be updated.


suku248 (approved):
Thanks @jessica-mitchell ! I think this PR removes all dry-run related code.

heplesser (approved):

PR #3339, opened 2024-10-18
Loosen pylint threshold

terhorstd:
This PR loosens the seemingly changed settings for `too-many-positional-arguments`. Instead of ignoring or refactoring this at this point, we loosen the threshold.
:construction:  A follow-up needs to be created to fix those bad spots in the code by appropriate refactoring.


gtrensch (approved):
Looks good to me!

Scutua (approved):

PR #3369, opened 2024-11-18
Distance dependent structural plasticity

FinnBurkhardt:
## Description

This pull request introduces optional distance-dependent connection probabilities for structural plasticity in the `SPManager` class.

### `gather_global_positions_and_ids()`

Collects global neuron positions and IDs across MPI ranks.

This method validates that neurons with valid spatial positions are available when distance-dependent structural plasticity is enabled. If no valid positions are found, an error is raised.

### `gaussian_kernel()`

Computes the Gaussian distance-dependent connection weight between two neuron positions.

The kernel is used as an unnormalized probability weight during spatial target selection. Larger values of `gaussian_kernel_sigma` reduce the sensitivity to distance, making distant neurons more likely to be selected.

### `squared_distance()`

Computes the squared Euclidean distance between two position vectors.

### `within_max_distance()`

Checks whether two neurons are within the configured maximum distance.

If `max_distance` is finite, candidate pairs beyond this distance are excluded from spatial target selection. If `max_distance` is infinite, no hard distance cutoff is applied.

### `global_shuffle_spatial()`

Implements spatially weighted pairing of vacant pre- and post-synaptic elements.

For each pre-synaptic element, candidate post-synaptic targets are weighted by the Gaussian distance kernel and filtered by the optional maximum-distance mask. A post-synaptic target is then selected using weighted random selection.

### `roulette_wheel_selection()`

Performs weighted random selection from a vector of probability weights.

This is used by `global_shuffle_spatial()` to select post-synaptic targets according to their distance-dependent probabilities.

### `enable_structural_plasticity()`

Structural plasticity can now be enabled with optional distance-dependent parameters:

```python
nest.EnableStructuralPlasticity(
    use_gaussian_kernel=False,
    gaussian_kernel_sigma=1.0,
    max_distance=float("inf"),
)
```

The default behavior remains the existing non-spatial structural plasticity behavior.

When `use_gaussian_kernel=True`, `gaussian_kernel_sigma` must be positive. The optional `max_distance` parameter can be used as a hard distance mask.

### `create_synapses()`

Synapse creation now dispatches between the existing non-spatial pairing logic and the new spatial Gaussian pairing logic.

If `use_gaussian_kernel=False`, the previous structural-plasticity behavior is preserved.

If `use_gaussian_kernel=True`, vacant pre- and post-synaptic elements are paired using distance-dependent weighted selection.

### `set_status()`

The handling of `allow_autapses` and `allow_multapses` for structural-plasticity synapse specifications was cleaned up.

A fresh connection specification is now created for each structural-plasticity synapse entry, so `allow_autapses` / `allow_multapses` settings cannot leak between multiple synapse specifications.

### Enable Distance Dependence

Distance dependence is enabled through:

```python
nest.EnableStructuralPlasticity(
    use_gaussian_kernel=True,
    gaussian_kernel_sigma=<positive value>,
)
```

The parameter `gaussian_kernel_sigma` controls the width of the Gaussian distance kernel. Higher values make the selection less sensitive to distance.

### Maximum-Distance Mask

An optional hard distance cutoff can be provided:

```python
nest.EnableStructuralPlasticity(
    use_gaussian_kernel=True,
    gaussian_kernel_sigma=<positive value>,
    max_distance=<positive value>,
)
```

Candidate pairs with distances larger than `max_distance` are excluded from spatial target selection.

The default value is `max_distance=float("inf")`, which disables the hard cutoff.

### Probability Caching

Previously implemented probability caching was removed from this PR.

It needs separate design and benchmarking, so this PR only implements direct distance-dependent target selection without a precomputed probability cache.

## Autapses and Multapses

Autapses and multapses are handled as follows:

- `allow_autapses=False` is respected for newly created structural-plasticity synapses. In the spatial path, self-connections are excluded during candidate selection. The final `SPBuilder` connection step also checks again and skips autapses when autapses are disabled.
- `allow_multapses=True` remains the supported structural-plasticity behavior. Structural plasticity pairs vacant synaptic elements, so multiple synapses between the same source and target can occur when both neurons have multiple vacant elements.
- `allow_multapses=False` is now rejected explicitly with `NotImplemented`. Suppression of multapses is not implemented in the structural-plasticity creation path, so the flag is no longer silently accepted.

## Performance Comparison

I added a benchmark comparing ordinary structural plasticity with the new spatial / Gaussian distance-dependent variant.

The benchmark is designed to measure structural-plasticity overhead during synapse creation. Newly created structural-plasticity synapses use zero recurrent weights in this benchmark, so the measurement focuses on the cost of structural-plasticity updates and distance-dependent partner selection rather than on changes in network dynamics caused by newly formed recurrent connections.

For each network scale, the same spatial positions are reused across benchmark conditions. This keeps the comparison between non-spatial and spatial structural plasticity controlled.

### Benchmark setup

The benchmark uses two neuron populations:

- excitatory population size: `800 * scale`
- inhibitory population size: `200 * scale`
- random 2D spatial positions in a `100 x 100` domain
- one Poisson background input population
- structural-plasticity update interval: `1000 ms`
- one MPI process and one local thread

The benchmark was run for:

- scale factors: `1`, `2`, `8`, `12`
- simulation lengths: `80000`, `120000`, `160000` ms
- repeats: `3` per condition

### Compared conditions

| Case | Meaning |
|---|---|
| `no_sp` | Structural plasticity disabled |
| `sp_no_gaussian` | Ordinary structural plasticity without spatial distance dependence |
| `sp_uniform_spatial` | Spatial path enabled with an effectively uniform kernel, `sigma = 1e9` |
| `sp_gaussian_sigma_3` | Spatial path enabled with Gaussian distance weighting, `sigma = 3` |
| `sp_uniform_spatial_maxdist_20` | Uniform spatial path with a hard maximum-distance mask |
| `sp_uniform_spatial_maxdist_inf` | Uniform spatial path with no finite maximum-distance mask |

The `sp_uniform_spatial` condition is useful because it separates the overhead of the spatial target-selection path from the overhead of a genuinely distance-sensitive Gaussian kernel.

### Measured quantities

| Quantity | Meaning |
|---|---|
| `T_enable` | Wall-clock time spent in `EnableStructuralPlasticity(...)` |
| `T_sim` | Total wall-clock time spent in chunked `nest.Simulate(...)` calls |
| `total_secs` | `T_enable + T_sim` |
| `median_sec_per_pair` | Median over SP update intervals of wall-time divided by newly created pairs |
| `total_pairs_created` | Total number of newly created structural-plasticity pairs |

The main comparison is based on total runtime overhead. `median_sec_per_pair` is reported only as a diagnostic quantity because it can be noisy when few pairs are created during an update interval.

### Summary over all scale factors and simulation lengths

| Case | Mean overhead vs. no SP | Mean overhead vs. non-Gaussian SP | Mean `EnableStructuralPlasticity` time |
|---|---:|---:|---:|
| SP, no Gaussian | +0.63% | baseline | 0.018 ms |
| Spatial uniform control | +1.33% | +0.70% | 0.981 ms |
| Gaussian, sigma = 3 | +1.51% | +0.87% | 0.979 ms |
| Spatial uniform + max distance 20 | +1.39% | +0.75% | 0.984 ms |
| Spatial uniform + max distance inf | +1.45% | +0.82% | 0.982 ms |

Overall, the Gaussian distance-dependent implementation adds a small runtime overhead compared with ordinary structural plasticity, around `0.9%` on average in this benchmark. Compared with no structural plasticity, the overhead is around `1.5%` on average.

The setup cost of `EnableStructuralPlasticity(...)` remains negligible compared with total simulation time.

### Benchmark plots
01_overall_runtime_overhead_vs_no_sp

02_spatial_overhead_vs_nonspatial_sp_by_scale

## Testing

The following tests have been added or updated to validate the new functionality.

#### `test_gaussian_kernel`

Tests Gaussian kernel computations for different distances and sigma values.

#### `test_distance_dependent_without_positions()`

Ensures an error is raised when distance-dependent structural plasticity is enabled but neurons do not have valid spatial positions.

#### `test_structural_plasticity_with_positions()`

Ensures structural plasticity works correctly when neuron positions are defined.

#### `test_synapse_creation_distance_dependent()`

Validates distance-dependent synapse creation using spatially positioned neurons.

#### Autapse / multapse tests

The structural-plasticity autapse and multapse tests were updated to reflect the current behavior:

- `allow_autapses=False` is supported.
- `allow_multapses=True` is supported.
- `allow_multapses=False` is rejected explicitly, because multapse suppression is not implemented in the structural-plasticity creation path.

## Addressed Review Comments

This update addresses the previous review comments as follows:

- The distance-dependent interface is now configured through `nest.EnableStructuralPlasticity(...)`.
- The misleading negative sigma default was removed.
- `gaussian_kernel_sigma` is only meaningful when `use_gaussian_kernel=True`, and it must be positive.
- Probability caching was removed from this PR.
- A benchmark comparison with and without Gaussian distance dependence was added.
- Spatial structural-plasticity tests were moved into a separate test file.
- The duplicate `test_distance_dependent_without_positions()` was removed.
- Autapse and multapse behavior was clarified and tested.
- A maximum-distance mask is now supported through `max_distance`.


alpinangu:
Looks good. We might reduce some overhead by passing a pointer to the relevant element in the global_position vector along with its dimension directly to the gaussianKernel function instead of creating and passing subvectors on each iteration, but for a small number of neurons it would not be noticeable

jhnnsnk (changes requested):
Thank you very much for this PR! At a first glance, I have the following comments:
- I suggest to improve the interface a bit: Can `nest.EnableStructuralPlasticity()` get arguments instead of having `nest.structural_plasticity_gaussian_kernel_sigma` and `nest.structural_plasticity_cache_probabilities` separately?
- Using `structural_plasticity_gaussian_kernel_sigma = -1` as default is misleading. How about an argument such as `use_Gaussian_kernel=true` instead and then allowing only values for sigma that actually make sense (and allowing to set this value only if the Gaussian kernel is enabled)?
- How does the performance compare if you connect and simulate without the Gaussian kernel? And with and without cached probabilities?
- Could you please document what caching probabilities actually does?
- I suggest to move the pytests for the spatial version into a separate file in `test_sp`.
- The test `test_distance_dependent_without_positions` appears twice.
- How are autapses and multapses handled during the creation and later during the simulation?
- Can masks be supported (as a maximum distance)?

FinnBurkhardt:
@alpinangu ,@jhnnsnk Thank you so much for your feedback! I'm working on the suggested changes and the performance comparisons.

FinnBurkhardt:
Hi @jhnnsnk @alpinangu,

I have now finished the implementation and addressed the review comments. I also included a benchmark, which shows only a small overhead. I am happy to go into more detail if useful.

Thanks again for the feedback. I'd be grateful for another review.

nestkernel/sp_manager.cpp
```diff
@@ -166,6 +181,134 @@ SPManager::set_status( const Dictionary& d )
...
+  std::vector< int > displacements;
+
+  // Collect local positions and IDs
+  for ( size_t tid = 0; tid < kernel().vp_manager.get_num_threads(); ++tid )
+  {
+    const SparseNodeArray& local_nodes = kernel().node_manager.get_local_nodes( tid );
+
+    for ( auto node_it = local_nodes.begin(); node_it < local_nodes.end(); ++node_it )
+    {
+      int node_id = node_it->get_node_id();
```
alpinangu:
```suggestion
      const size_t node_id = node_it->get_node_id();
```

nestkernel/sp_manager.cpp
```diff
@@ -166,6 +181,134 @@ SPManager::set_status( const Dictionary& d )
...
+    id_pos_pairs.emplace_back( node_id, pos );
+  }
+
+  // Sort id_pos_pairs based on node_id to ensure ordering from 1 to num_neurons
+  std::sort( id_pos_pairs.begin(),
+    id_pos_pairs.end(),
+    []( const std::pair< int, std::vector< double > >& a, const std::pair< int, std::vector< double > >& b ) -> bool
+    { return a.first < b.first; } );
+
+  // Verify that IDs are sequential
```
alpinangu: Do these global ids of neurons actually have to go from 1 to num_neurons? If they are the GIDs then if we create a device or a non spetial population (if it is allowed) first then the GIDs would not start from 1. Also if we create a population then a device and then a population again the GIDs would not necceserly be sequential.

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +835,114 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+    if ( pre_ids.empty() || post_ids.empty() )
+    {
+      break;  // Stop if either vector is empty
+    }
+
+    size_t pre_id = pre_ids.back();
+    pre_ids.pop_back();
+
+    std::vector< double > pre_pos(
+      global_positions.begin() + ( pre_id - 1 ) * pos_dim, global_positions.begin() + pre_id * pos_dim );
```
alpinangu: This also seems to assume array index = GID-1, see the previous comment about the GIDs

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +835,114 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+    for ( size_t post_id : post_ids )
+    {
+      if ( post_id == pre_id && !allow_autapses )
+      {
+        continue;  // Skip self-connections
+      }
+
+      // fetch post position
+      std::vector< double > post_pos(
+        global_positions.begin() + ( post_id - 1 ) * pos_dim, global_positions.begin() + post_id * pos_dim );
```
alpinangu: Likewise

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +835,114 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+void
+SPManager::global_shuffle_spatial( std::vector< size_t >& pre_ids,
+  std::vector< size_t >& post_ids,
+  std::vector< size_t >& pre_ids_results,
+  std::vector< size_t >& post_ids_results,
+  bool allow_autapses )
+{
+  size_t maxIterations = std::min( pre_ids.size(), post_ids.size() );
+
+  for ( size_t iteration = 0; iteration < maxIterations; ++iteration )
```
alpinangu:
```suggestion
  while (!pre_ids.empty() && !post_ids.empty())
```

If we have lets say 100 pre-synaptic neuron and 10 post-synaptic neurons, we would itterate over 10 times, but if the first 10 checked pre-synaptic neurons are not eligible, shouldn't we still check the other neurons? If so this while loop would be better

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +835,114 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+      if ( post_id == pre_id && !allow_autapses )
+      {
+        continue;  // Skip self-connections
+      }
+
+      // fetch post position
+      std::vector< double > post_pos(
+        global_positions.begin() + ( post_id - 1 ) * pos_dim, global_positions.begin() + post_id * pos_dim );
+
+      if ( !within_max_distance( pre_pos, post_pos ) )
```
alpinangu: [Docs](https://nest-simulator.readthedocs.io/en/v3.9/ref_material/pynest_api/nest.spatial.hl_api_spatial.html?utm_source=chatgpt.com) says spatial positions can use periodic boundary wrapping, you might want to use the function they use for distance if you use the same coordiantes

nestkernel/sp_manager.cpp
FinnBurkhardt: Good Point!

alpinangu (approved):
👍
it just might be better to use size_t instead of int for GIDs for large networks and for consistency

heplesser (changes requested):
@FinnBurkhardt Thank you for your PR and my apologies for coming late to the party :). You seem to re-implement a number of things that are already available in the code handling spatial connectivity—and there for 2D and 3D, e.g., exchange of positions, distance calculations and gaussian kernels. Building on that, you could directly use the full power of NEST's `Parameter` support. I am also concerned about hard-coding a gaussian profile into `EnableStructuralPlasticity()`—what if someone wants to test an exponential dependency next? Using the tools we have for spatial connectivity, we should be able to provide general support.

nestkernel/nest.h
```diff
@@ -56,7 +56,7 @@ void install_module( const std::string& module_name );
 
 void reset_kernel();
 
-void enable_structural_plasticity();
+void enable_structural_plasticity( ParameterPTR spatial_kernel, MaskPTR spatial_mask );
```
heplesser, line 59:
```suggestion
void enable_structural_plasticity( ParameterPTR spatial_kernel=nullptr, MaskPTR spatial_mask=nullptr );
```

Please add Doxygen documentation for the function and its arguments.

nestkernel/simulation_manager.cpp
```diff
@@ -853,6 +853,8 @@ nest::SimulationManager::update_()
 
   std::vector< std::exception_ptr > exceptions_raised( kernel().vp_manager.get_num_threads() );
 
+  std::exception_ptr sp_exception_raised;
```
heplesser, line 856: Remove, we already have a catching mechanism in place.

nestkernel/simulation_manager.cpp
```diff
@@ -1039,7 +1041,18 @@ nest::SimulationManager::update_()
...
+              kernel().sp_manager.update_structural_plasticity();
+            }
+            catch ( ... )
+            {
+              sp_exception_raised = std::current_exception();
+            }
+          }
+          if ( sp_exception_raised )
+          {
+            std::rethrow_exception( sp_exception_raised );
```
heplesser, line 1055:
```suggestion
```

nestkernel/sp_manager.h
```diff
@@ -121,9 +123,12 @@ class SPManager : public ManagerInterface
...
 
   /**
-   * Enable structural plasticity
+   * Enable structural plasticity.
+   *
+   * @param spatial_kernel  Weights candidates by their positions; zero excludes. Null: uniform matching.
+   * @param spatial_mask    Excludes candidates whose source-target displacement lies outside it.
    */
-  void enable_structural_plasticity();
+  void enable_structural_plasticity( ParameterPTR spatial_kernel = ParameterPTR(), MaskPTR spatial_mask = MaskPTR() );
```
heplesser, line 131: Provide default arguments in nest.h at user-interface level.

```suggestion
  void enable_structural_plasticity( ParameterPTR spatial_kernel, MaskPTR spatial_mask );
```

nestkernel/sp_manager.h
```diff
@@ -185,6 +193,49 @@ class SPManager : public ManagerInterface
...
+
+  /**
+   * Gather global neuron positions and IDs from all nodes.
+   */
+  void gather_global_positions_and_ids();
+
+  /**
+   * Select an index at random, with probability proportional to the given weights.
+   *
+   * @param weights Non-negative, need not be normalised; at least one must be positive.
```
heplesser, line 222:
```suggestion
   * @param weights Non-negative, need not be normalised; at least one must be strictly positive.
```

nestkernel/sp_manager.h
```diff
@@ -185,6 +193,49 @@ class SPManager : public ManagerInterface
...
+   * @param rnd Random number drawn from [0, 1).
+   * @return Selected index.
+   */
+  size_t roulette_wheel_selection( const std::vector< double >& weights, double rnd );
+
+  /**
+   * Global list of neuron IDs used for structural plasticity computations.
+   */
+  std::vector< int > global_ids;
```
heplesser, line 231: This and global_positions must be private. Provide const getters if required. Consider vector<pair<id, pos>> instead of two vectors.

nestkernel/sp_manager.h
```diff
@@ -185,6 +193,49 @@ class SPManager : public ManagerInterface
...
+
+  /**
+   * Global list of neuron positions used for spatial computations in
+   * structural plasticity.
+   */
+  std::vector< double > global_positions;
```
heplesser, line 237: Use `Position` class.

nestkernel/sp_manager.h
```diff
@@ -197,6 +248,18 @@ class SPManager : public ManagerInterface
    * Off (False).
    */
   bool structural_plasticity_enabled_;
+
+  //! Optional spatial kernel weighting candidate partners; null means uniform matching.
+  ParameterPTR structural_plasticity_kernel_;
```
heplesser, line 253: Consider name, also below.

nestkernel/sp_manager.cpp
```diff
@@ -41,6 +45,9 @@ SPManager::SPManager()
   : ManagerInterface()
   , structural_plasticity_update_interval_( 10000. )
   , structural_plasticity_enabled_( false )
+  , structural_plasticity_kernel_()
+  , structural_plasticity_mask_()
+  , pos_dim( 0 )
```
heplesser, line 50: Can probably go if `Position` class is used.

nestkernel/sp_manager.cpp
```diff
@@ -126,8 +135,6 @@ SPManager::set_status( const Dictionary& d )
   }
 
   Dictionary syn_specs;
```
heplesser, line 137: Consider to rename to `sp_specs`.

nestkernel/sp_manager.cpp
```diff
@@ -166,6 +179,124 @@ SPManager::set_status( const Dictionary& d )
   }
 }
 
+void
+SPManager::gather_global_positions_and_ids()
```
heplesser, line 183: Use instead `Layer::get_global_positions_vector()` if possible.

nestkernel/sp_manager.cpp
```diff
@@ -166,6 +179,124 @@ SPManager::set_status( const Dictionary& d )
...
+  if ( not std::isfinite( sum ) or sum <= 0.0 )
+  {
+    throw std::runtime_error( "Sum of selection weights must be finite and greater than zero." );
+  }
+
+  // Generate a random number in the range [0, sum)
+  const double random_value = rnd * sum;
+
+  const auto it = std::upper_bound( cumulative.begin(), cumulative.end(), random_value );
+  if ( it == cumulative.end() )
```
heplesser, line 288: This handling will work for zero weights at the end, but not in the middle of the data. How to handle this properly? We must not return an index corresponding to a zero weight.

Maybe just loop downwards as long as the element to the left is not smaller.

nestkernel/sp_manager.cpp
```diff
@@ -166,6 +179,124 @@ SPManager::set_status( const Dictionary& d )
...
+  }
+
+  // Update global_positions with sorted positions
+  global_positions = std::move( temp_positions );
+}
+
+
+// Method to perform roulette wheel selection
+size_t
+SPManager::roulette_wheel_selection( const std::vector< double >& weights, double rnd )
```
heplesser, line 268: Pass in RNG, not rnd.

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +822,109 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+SPManager::match_vacant_elements_spatially( std::vector< size_t >& pre_ids,
+  std::vector< size_t >& post_ids,
+  std::vector< size_t >& pre_ids_results,
+  std::vector< size_t >& post_ids_results,
+  bool allow_autapses )
+{
 
+  const auto get_global_position = [ this ]( const size_t node_id )
+  {
+    const auto id_it = std::lower_bound( global_ids.begin(), global_ids.end(), node_id );
```
heplesser, line 835: This strongly suggests that `global_ids` and `global_positions` should be an `std::map`.

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +822,109 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+    for ( size_t post_id : post_ids )
+    {
+      if ( post_id == pre_id and not allow_autapses )
+      {
+        continue;  // Skip self-connections
+      }
+
+      const std::vector< double > post_pos = get_global_position( post_id );
+
+      // Via the layer, so that periodic boundary conditions are respected.
```
heplesser, line 867: Comment is bit confusing here.

nestkernel/sp_manager.cpp
```diff
@@ -671,10 +822,109 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
+      }
+    }
+
+    if ( weights.empty() )
+    {
+      continue;  // No admissible partner; the element stays vacant.
+    }
+
+    // Select a post-synaptic neuron using roulette wheel selection
+    const size_t selected_post_idx = roulette_wheel_selection( weights, rng->drand() );
```
heplesser, line 911: Pass ids also in and then return selected neuron.

testsuite/pytests/structural_plasticity/test_sp_autapses_multapses.py
```diff
@@ -45,22 +44,22 @@ def test_autapses(self):
         assert not nest.structural_plasticity_synapses["synapse_ex"]["allow_autapses"]
 
     def test_multapses(self):
-        """Test if allow_multapses can be set on structural plasticity synapses."""
+        """Test if allow_multapses=False is rejected for structural plasticity synapses."""
```
heplesser, line 47:
```suggestion
        """Test if allow_multapses==False is rejected for structural plasticity synapses."""
```

testsuite/pytests/structural_plasticity/test_sp_manager_spatial.py
```diff
@@ -0,0 +1,394 @@
...
+    if not isinstance(sources, (list, tuple)):
+        sources, targets = [sources], [targets]
+    return list(sources), list(targets)
+
+
+def gaussian_kernel(std, mean=0.0):
+    return nest.spatial_distributions.gaussian(nest.spatial.distance, mean=mean, std=std)
+
+
+def test_gaussian_follows_nest_convention():
```
heplesser, line 90: NeedeD?

testsuite/pytests/structural_plasticity/test_sp_manager_spatial.py
```diff
@@ -0,0 +1,394 @@
...
+        nest.EnableStructuralPlasticity(spatial_kernel=gaussian_kernel(std=std))
+        nest.Simulate(5.0)
+
+        _, targets = sp_connections()
+        assert len(targets) == 1
+        num_near += targets[0] == near_id
+
+    # Four standard errors of the binomial estimate, so the tolerance does not depend on the seeds.
+    tolerance = 4 * np.sqrt(expected * (1 - expected) / num_runs)
+    assert num_near / num_runs == pytest.approx(expected, abs=tolerance)
```
heplesser, line 135: Mayb use existing stats tests from scipy.

heplesser (changes requested):
Nice work, see remaining comments inline.

PR #3378, opened 2024-12-09
Add optional support for per-thread timers and timing MPI synchronization

JanVogelsang:
Added option to measure timers per thread and added timers to measure thread synchronization time. With this PR, we are hoping to be able to identify idle time of individual threads and identify black time present with the current timers.


CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ set( with-models OFF CACHE STRING "The models to include as a semicolon-separate
 set( tics_per_ms "1000.0" CACHE STRING "Specify elementary unit of time [default=1000 tics per ms]." )
 set( tics_per_step "100" CACHE STRING "Specify resolution [default=100 tics per step]." )
 set( with-detailed-timers OFF CACHE STRING "Build with detailed internal time measurements [default=OFF]. Detailed timers can affect the performance." )
+set( with-threaded-timers ON CACHE STRING "Build with one internal timer per thread [default=ON]. Multi-threaded timers can affect the performance." )
```
heplesser, line 71: Do we really want to make this the default? Does it only apply to detailed timers?

nestkernel/stopwatch.h
```diff
@@ -273,18 +304,53 @@ nest::Stopwatch::print( const char* msg, timeunit_t timeunit, std::ostream& os )
...
+
+class StopwatchArray
+{
+public:
+  void reset();
+
+  std::vector< double > elapsed();
+
+  SingleStopwatch&
+  operator[]( const size_t tid )
```
heplesser, line 328: Do we really want to make the individual stopwatches directly available? Wouldn't it be much tidier to define `start()` and `stop()` for the array and then call `kernel().vp_manager.get_thread_id()` inside the start/stop, or use 0 if we do not use threaded timers? That would also avoid the macros, I think. The threaded-timers ifdef would then occur only inside the  `StopwatchArray` class.

heplesser (changes requested):
@JanVogelsang Thanks for the PR! Please look at my comments starting at the very end—I am wondering another approach that would avoid macros.

nestkernel/stopwatch.h
JanVogelsang: I like the idea. In the NEST code, we typically add tid as a parameter in threaded functions instead of calling `get_thread_id` every time, which would most likely be cleaner though. I can change this for this PR, but we might create a follow-up issue to do the same for all the other occurrences in the code.

nestkernel/stopwatch.h
```diff
@@ -273,18 +304,53 @@ nest::Stopwatch::print( const char* msg, timeunit_t timeunit, std::ostream& os )
...
+
+inline std::vector< double >
+StopwatchArray::elapsed()
+{
+  std::vector< double > times;
+  times.reserve( timers_.size() );
+  for ( auto& timer : timers_ )
+  {
+    times.push_back( timer.elapsed() );
+  }
```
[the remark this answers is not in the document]
JanVogelsang: Done!

nestkernel/stopwatch.h
```diff
@@ -273,18 +304,53 @@ nest::Stopwatch::print( const char* msg, timeunit_t timeunit, std::ostream& os )
...
+  return ( nest::SingleStopwatch::timestamp_t ) now.tv_usec
+    + ( nest::SingleStopwatch::timestamp_t ) now.tv_sec * nest::SingleStopwatch::SECONDS;
+}
+
+class StopwatchArray
+{
+public:
+  void reset();
+
+  std::vector< double > elapsed();
```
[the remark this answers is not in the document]
JanVogelsang: One can certainly do this! However, after searching for a solution, I stumbled across this StackOverflow post, which claims that it is best practice to not specify the move semantics explicitly:
https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-with-return-statement

nestkernel/stopwatch.h
```diff
@@ -121,7 +152,7 @@ class Stopwatch
    * runtime is added. If you want only the last measurement, you
    * have to reset the timer, before stating the measurement.
    * Does not change the running state.
-   * In contrast to Stopwatch::elapsed(), only the timestamp is returned,
+   * In contrast to SingleStopwatch::elapsed(), only the timestamp is returned,
    * that is the number if microseconds as an integer.
```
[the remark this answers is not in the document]
JanVogelsang: I don't have a strong opinion on this topic. We can either open a new issue or just quickly discuss this in person some time during the hackathon.

nestkernel/kernel_manager.cpp
```diff
@@ -162,6 +162,12 @@ nest::KernelManager::change_number_of_threads( size_t new_num_threads )
   // is in place again, we can tell modules to re-register the components
   // they provide.
   module_manager.reinitialize_dynamic_modules();
+
+  // Prepare timers
+  kernel().simulation_manager.reset_timers_for_preparation();
+  kernel().simulation_manager.reset_timers_for_dynamics();
+  kernel().event_delivery_manager.reset_timers_for_preparation();
+  kernel().event_delivery_manager.reset_timers_for_dynamics();
```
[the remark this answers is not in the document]
JanVogelsang: Done!

nestkernel/event_delivery_manager.h
```diff
@@ -469,9 +469,9 @@ class EventDeliveryManager : public ManagerInterface
 #ifdef TIMER_DETAILED
   // private stop watches for benchmarking purposes
   // (intended for internal core developers, not for use in the public API)
-  Stopwatch sw_collocate_spike_data_;
-  Stopwatch sw_communicate_spike_data_;
-  Stopwatch sw_communicate_target_data_;
+  SingleStopwatch sw_collocate_spike_data_;
```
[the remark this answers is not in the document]
JanVogelsang: See below.

nestkernel/mpi_manager.cpp
```diff
@@ -870,7 +870,7 @@ nest::MPIManager::time_communicatev( int num_bytes, int samples )
   }
 
   // start time measurement here
-  Stopwatch foo;
+  SingleStopwatch foo;
```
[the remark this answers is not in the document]
JanVogelsang: For comment see above.

nestkernel/node_manager.h
```diff
@@ -356,7 +356,7 @@ class NodeManager : public ManagerInterface
   std::vector< std::shared_ptr< WrappedThreadException > > exceptions_raised_;
 
   // private stop watch for benchmarking purposes
-  Stopwatch sw_construction_create_;
+  SingleStopwatch sw_construction_create_;
```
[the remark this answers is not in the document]
JanVogelsang: See above.

nestkernel/simulation_manager.cpp
```diff
@@ -523,6 +525,7 @@ nest::SimulationManager::prepare()
   // this call sets the member variables
   kernel().connection_manager.update_delay_extrema_();
   kernel().event_delivery_manager.init_moduli();
+  kernel().node_manager.ensure_valid_thread_local_ids();
```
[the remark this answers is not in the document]
JanVogelsang: Thanks for noticing. This is indeed coming from different changes, unrelated to this PR.

nestkernel/simulation_manager.cpp
```diff
@@ -738,17 +741,21 @@ nest::SimulationManager::call_update_()
...
-    sw_communicate_prepare_.start();
-  }
+  DETAILED_TIMER_STOP( sw_idle_, tid );
+
+  DETAILED_TIMER_START( sw_communicate_prepare_, tid );
 
-  kernel().connection_manager.restructure_connection_tables( tid );
   kernel().connection_manager.sort_connections( tid );
+  DETAILED_TIMER_START( sw_gather_target_data_, tid );
+  kernel().connection_manager.restructure_connection_tables( tid );
```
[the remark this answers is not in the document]
JanVogelsang: Restructuring the source and target table should be part of gathering target data in my opinion, as this is only needed when we actually want to communicate connectivity (i.e., "gather target data"). I think it is fine to alter detailed timers, as those are mainly used to guide optimizations anyway. I would be more cautious with non-detailed timers, which will always be output, even by non-developers. We can add this to the documentation somewhere, but I am unsure about a good location for such an information. And we can also discuss this in a larger group, as one might also argue that backwards-compatability of detailed timers might make sense.

nestkernel/simulation_manager.cpp
```diff
@@ -972,8 +950,10 @@ nest::SimulationManager::update_()
 // add done value of thread p to done vector
 #pragma omp critical
             done.push_back( done_p );
```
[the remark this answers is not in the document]
JanVogelsang: Yes, done!

nestkernel/simulation_manager.cpp
```diff
@@ -1074,46 +1049,41 @@ nest::SimulationManager::update_()
...
-        }
-#endif
+        // parallel section ends, wait until all threads are done -> synchronize
+        DETAILED_TIMER_START( sw_idle_, tid );
+#pragma omp barrier
+        DETAILED_TIMER_STOP( sw_idle_, tid );
 
         // the following block is executed by the master thread only
-// the other threads are enforced to wait at the end of the block
-#pragma omp master
```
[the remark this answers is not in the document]
JanVogelsang: Hm, true. I don't see how this could be done differently though.

nestkernel/simulation_manager.h
```diff
@@ -189,6 +189,14 @@ class SimulationManager : public ManagerInterface
   Time get_eprop_learning_window() const;
   bool get_eprop_reset_neurons_on_update() const;
 
+#ifdef TIMER_DETAILED
+  Stopwatch&
+  get_idle_stopwatch()
```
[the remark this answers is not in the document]
JanVogelsang: Yes, exactly. Any suggestions for a better location? Maybe in the kernel manager?

nestkernel/simulation_manager.h
```diff
@@ -228,7 +236,7 @@ class SimulationManager : public ManagerInterface
   double max_update_time_;         //!< longest update time seen so far (seconds)
 
   // private stop watches for benchmarking purposes
-  Stopwatch sw_simulate_;
+  SingleStopwatch sw_simulate_;
   Stopwatch sw_communicate_prepare_;
```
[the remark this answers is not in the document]
JanVogelsang: Should we really? I think we should rather focus on getting as much information as possible from our timers as it doesn't really hurt to use different stopwatch types based on where timers are being used.

nestkernel/stopwatch.h
```diff
@@ -29,12 +29,43 @@
...
+#define DETAILED_TIMER_START( timer, tid ) TIMER( timer, tid ).start();
+#else
+#define DETAILED_TIMER_START( timer, tid )
+#endif
+
+#ifdef TIMER_DETAILED
+#define DETAILED_TIMER_STOP( timer, tid ) TIMER( timer, tid ).stop();
+#else
+#define DETAILED_TIMER_STOP( timer, tid )
+#endif
```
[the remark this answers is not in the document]
JanVogelsang: Done!

nestkernel/stopwatch.h
```diff
@@ -273,18 +304,53 @@ nest::Stopwatch::print( const char* msg, timeunit_t timeunit, std::ostream& os )
...
   // * MacOS 10.9
   struct timeval now;
   gettimeofday( &now, static_cast< struct timezone* >( nullptr ) );
-  return ( nest::Stopwatch::timestamp_t ) now.tv_usec
-    + ( nest::Stopwatch::timestamp_t ) now.tv_sec * nest::Stopwatch::SECONDS;
+  return ( nest::SingleStopwatch::timestamp_t ) now.tv_usec
+    + ( nest::SingleStopwatch::timestamp_t ) now.tv_sec * nest::SingleStopwatch::SECONDS;
+}
+
+class StopwatchArray
```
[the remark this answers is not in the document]
JanVogelsang: Done!

nestkernel/stopwatch.h
```diff
@@ -29,12 +29,43 @@
...
+#define DETAILED_TIMER_START( timer, tid )
+#endif
+
+#ifdef TIMER_DETAILED
+#define DETAILED_TIMER_STOP( timer, tid ) TIMER( timer, tid ).stop();
+#else
+#define DETAILED_TIMER_STOP( timer, tid )
+#endif
+
+// TODO JV: Update docs
```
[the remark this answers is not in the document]
JanVogelsang: Will do that once the rest of the PR is reviewed/accepted, so that I only have to write the docs once based on the final version.

doc/htmldoc/nest_behavior/built-in_timers.rst
```diff
@@ -86,7 +86,7 @@ available as kernel attributes:
 +--------------------------------+----------------------------------+----------------------------------+
 |``time_communicate_target_data``|Cumulative time for core MPI      |``time_gather_target_data``       |
 |                                |communication when gathering      |                                  |
-|				 |target data                       |                                  |
+|			                	 |target data                       |                                  |
```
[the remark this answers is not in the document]
JanVogelsang: Yes!

CMakeLists.txt
JanVogelsang: I would argue that this is a good idea to make people aware of the limitations of the current timers which lead to many problems (e.g. dark time).

nestkernel/simulation_manager.cpp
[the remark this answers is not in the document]
heplesser: I agree, changing the timer definition here is fine, but the change should be documented well, both in the documentation of the timer and the documentation of changes for NEST 3.9.

nestkernel/simulation_manager.cpp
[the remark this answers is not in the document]
heplesser: Isn't it just one master section in the current code?

nestkernel/simulation_manager.h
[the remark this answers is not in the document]
heplesser: Yes, KernelManager might be a good place. I was starting to think about a TimerManager ...

CMakeLists.txt
heplesser: Ok, please also document that this only applies to detailed timers.

mlober: I think it is more intuitive to have the threaded timers OFF by default and if the user wants to go a step further with the performance measurements they are turned ON. Especially if the threaded timers affect performance. @JanVogelsang do you see an effect in simulation speed?

JanVogelsang: Didn't measure performance yet, will do that once we added CI-beNNch support for these timers.
But I don't think that hiding these timers behind a flag is a good idea. These timers are not about going a step further, they are more of a fix for existing timers. We found that the current timer implementation doesn't provide us with meaningful data and we are often just guessing where dark time comes from, for example. I personally see these new timers more as a replacement of the previous timers, rather than an addition.

nestkernel/event_delivery_manager.cpp
```diff
@@ -826,18 +814,21 @@ EventDeliveryManager::gather_target_data( const size_t tid )
...
 #pragma omp barrier
+    kernel().get_omp_synchronization_stopwatch().stop();
     kernel().connection_manager.clean_source_table( tid );
 
 #pragma omp master
     {
-#ifdef TIMER_DETAILED
-      sw_communicate_target_data_.start();
+#if defined( HAVE_MPI ) && defined( TIMER_DETAILED )
+      kernel().get_mpi_synchronization_stopwatch().start();
```
mlober, line 825: Since we are only really interested in the synchronization during spike communication we could remove these lines. Documentation of the timer might have to be adjusted to state that it only measures synchronization during gather spike data.

mlober (changes requested):
#

doc/htmldoc/nest_behavior/built-in_timers.rst
```diff
@@ -82,11 +82,11 @@ available as kernel attributes:
...
-|				 |postsynaptic to presynaptic side  |                                  |
+|				                 |postsynaptic to presynaptic side  |                                  |
 +--------------------------------+----------------------------------+----------------------------------+
 |``time_communicate_target_data``|Cumulative time for core MPI      |``time_gather_target_data``       |
 |                                |communication when gathering      |                                  |
-|				 |target data                       |                                  |
+|			                	 |target data                       |                                  |
 +--------------------------------+----------------------------------+----------------------------------+
 |``time_update``                 |Time for neuron update            |``time_simulate``                 |
 +--------------------------------+----------------------------------+----------------------------------+
```
mlober, line 92: Documentation on "time_mpi_synchronization" needed. E.g. "Time for synchronization of MPI processes during spike communication"

mlober (changes requested):

nestkernel/event_delivery_manager.cpp
```diff
@@ -426,12 +415,13 @@ EventDeliveryManager::gather_spike_data_( std::vector< SpikeDataT >& send_buffer
...
-      sw_collocate_spike_data_.stop();
-      sw_communicate_spike_data_.start();
-    }
+    sw_collocate_spike_data_.stop();
+#if defined( HAVE_MPI ) && defined( TIMER_DETAILED )
+    kernel().get_mpi_synchronization_stopwatch().start();
+    kernel().mpi_manager.synchronize();
+    kernel().get_mpi_synchronization_stopwatch().stop();
 #endif
+    sw_communicate_spike_data_.start();
```
mlober, line 424: So far, we always considered the synchronization as a part of communication. Communication = synchronization + data exchange. I think we should keep it this way and have the communicate_spike_data Stopwatch start before synchronization.

CMakeLists.txt
mlober: Okay, I agree. As we discussed, we then have to keep in mind that this will be a breaking change to existing benchmarking workflows like e.g. ci-beNNch.

cmake/ConfigureSummary.cmake
```diff
@@ -150,6 +150,13 @@ function( NEST_PRINT_CONFIG_SUMMARY )
     message( "Detailed timers     : No" )
   endif ()
 
+  message( "" )
+  if ( THREADED_TIMERS )
+    message( "Threaded timers     : Yes" )
+  else ()
+    message( "Threaded timers     : No" )
+  endif ()
+
```
mlober, line 159: Here, the message for MPI sync timers need to be added.

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+}
+
+template < clockid_t clock_type >
+inline void
+StopwatchBase< clock_type >::print( const char* msg, timeunit_t timeunit, std::ostream& os ) const
+{
+#ifndef DISABLE_TIMING
+  double e = elapsed( timeunit );
+  os << msg << e;
+  switch ( timeunit )
```
heplesser, line 238: Add nanoseconds as well to have all bases covered :).

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+    break;
+  case MINUTES:
+    os << " min.";
+    break;
+  case HOURS:
+    os << " h.";
+    break;
+  case DAYS:
+    os << " days.";
+    break;
```
heplesser, line 257: And just for safety's sake a `default:` clause raising an error?

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+  }
+#ifdef DEBUG
+  os << " (running: " << ( _running ? "true" : "false" ) << ", begin: " << _beg << ", end: " << _end
```
heplesser, line 260:
```suggestion
  os << std::boolalpha << " (running: " <<  _running << ", begin: " << _beg << ", end: " << _end
```

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+inline std::ostream&
+operator<<( std::ostream& os, const StopwatchBase< clock_type >& stopwatch )
+{
+  stopwatch.print( "", timeunit_t::SECONDS, os );
+  return os;
+}
+
+}
+
+enum StopwatchVerbosity
```
heplesser, line 287: I think `StopwatchGranularity` would be a better name, since it is about what we measure, not how much we talk about the measurements we take.

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+{
+  Normal,  //<! Always measure stopwatch
+  Detailed //<! Only measure if detailed stopwatches are activated
+};
+
+enum StopwatchType
+{
+  MasterOnly, //<! Only the master thread owns a stopwatch
+  Threaded    //<! Every thread measures an individual stopwatch
+};
```
heplesser, line 297: Also here I think we could use more expressive names:
```suggestion
enum StopwatchParallelism
{
  MasterOnly, //<! Only the master thread owns a stopwatch
  PerThread   //<! Every thread owns an individual stopwatch
};
```

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+
+
+/** This is the base template for all Stopwatch specializations.
+ */
+/** Base timer class, which only measures a single timer, owned by the master thread. Might only actually measure time
+ * if detailed timers are enabled.
```
heplesser, line 303: I am a bit confused here that `StopwatchBase` is not the base class for stopwatches. Maybe rethink the name `StopwatchBase` and provide some explanation of how that class relates to the classes being defined in the following. Also add a short explanation for the `typename = void` template argument.

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+  {
+#pragma omp master
+    {
+      walltime_timer_.reset();
+      cputime_timer_.reset();
+    }
+  }
+
+  void
+  print( const char* msg = "",
```
heplesser, line 362: As above, why C-string pointer and not C++ string?

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+  NANOSEC = 1,
+  MICROSEC = NANOSEC * 1000,
+  MILLISEC = MICROSEC * 1000,
+  SECONDS = MILLISEC * 1000,
+  MINUTES = SECONDS * 60,
+  HOURS = MINUTES * 60,
+  DAYS = HOURS * 24
+};
+
+template < clockid_t clock_type >
```
heplesser, line 93: Add  a doxygen comment for this class. Since it later is used to define variables called `..._timer_`, maybe call it `StopwatchTimer`?  That would also avoid confusion about the fact that this class is called `StopwatchBase` but then does *not* serve as base class for stopwatches.

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+  print( const char* msg = "",
+    timers::timeunit_t timeunit = timers::timeunit_t::SECONDS,
+    std::ostream& os = std::cout ) const
+  {
+#pragma omp master
+    walltime_timer_.print( msg, timeunit, os );
+  }
+
+  void
+  output_timer( DictionaryDatum& d, const Name& walltime_name, const Name& cputime_name )
```
heplesser, line 371:
```suggestion
  get_status( DictionaryDatum& d, const Name& walltime_name, const Name& cputime_name )
```
to make naming consistent with all other NEST code extracting information from parts and pieces of NEST. Also, why is the method not `const`?

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+    def< double >( d, cputime_name, cputime_timer_.elapsed() );
+  }
+
+private:
+  timers::StopwatchBase< CLOCK_MONOTONIC > walltime_timer_;
+  timers::StopwatchBase< CLOCK_THREAD_CPUTIME_ID > cputime_timer_;
+};
+
+/** If the user deactivated detailed timers, Stopwatch instance with the detailed flag will become an empty Stopwatch,
+ * which will be safely ignored by the compiler, as if the instance was not declared (e.g. as a member).
```
heplesser, line 383: Doxygen comments should start with a brief one-liner, then an empty line, then the full info.

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+  void
+  print( const char* = "", timers::timeunit_t = timers::timeunit_t::SECONDS, std::ostream& = std::cout ) const
+  {
+  }
+  void
+  output_timer( DictionaryDatum&, const Name&, const Name& )
+  {
+  }
+};
+template < StopwatchVerbosity detailed_timer >
```
heplesser, line 462: Insert an empty line above.

I get confused by how the selection of the different templates here works. Could you add an overview in the comment for the top-level `Stopwatch` template declaration about the approach to the selection logic?

nestkernel/stopwatch.h
```diff
@@ -0,0 +1,505 @@
...
+  void
+  output_timer( DictionaryDatum& d, const Name& walltime_name, const Name& cputime_name )
+  {
+    std::vector< double > wall_times( walltime_timers_.size() );
+    std::transform( walltime_timers_.begin(),
+      walltime_timers_.end(),
+      wall_times.begin(),
+      []( const timers::StopwatchBase< CLOCK_MONOTONIC >& timer ) { return timer.elapsed(); } );
+    def< ArrayDatum >( d, walltime_name, ArrayDatum( wall_times ) );
+    std::vector< double > cpu_times( cputime_timers_.size() );
```
heplesser, line 491: Insert empty line.

heplesser (changes requested):
@JanVogelsang Thanks a lot, this is a very nice implementation! I have some comments, mostly related to naming, const-ness and documentation.

heplesser:
@JanVogelsang This looks very good now. I just sent you a PR that solves the Clang compilation problem (gcc is too lenient ...).

JanVogelsang:
@heplesser Thanks for the PR! Should we also get rid of the timers namespace?

heplesser:
> @heplesser Thanks for the PR! Should we also get rid of the timers namespace?

We can keep it for now, maybe remove it in a later PR.

heplesser (approved):
@JanVogelsang Good to go now!

PR #1672, opened 2020-07-02
Fix for first spike passing through "tsodyks2" synapse

clinssen:
I believe this fixes #957. @MatthiasKohler: could you confirm? See in particular the diff for ``models/tsodyks2_connection.h``.

It could be that the follow-up issue, #959, is too broad. I would suggest to close #959 if this PR is accepted.

PR #3384, opened 2024-12-16
Fix issue with first spike passing through adaptive synapse

clinssen:
The fix for first spike passing through an adaptive synapse in https://github.com/nest/nest-simulator/pull/1672/ was erroneous and affects more than just the first spike. This PR fixes the fix.


models/tsodyks2_synapse.h
```diff
@@ -227,9 +231,19 @@ tsodyks2_synapse< targetidentifierT >::send( Event& e, size_t t, const CommonSyn
...
-  double x_decay = std::exp( -h / tau_rec_ );
-  double u_decay = ( tau_fac_ < 1.0e-10 ) ? 0.0 : std::exp( -h / tau_fac_ );
+
+  if ( t_lastspike_ >= 0.0 )
+  {
+    // only update x and u if this is not the first spike to pass through the synapse
+
+    const double h = t_spike - t_lastspike_;
+    double x_decay = std::exp( -h / tau_rec_ );
+    double u_decay = ( tau_fac_ < 1.0e-10 ) ? 0.0 : std::exp( -h / tau_fac_ );
```
heplesser, line 241:
```suggestion
    const double x_decay = std::exp( -h / tau_rec_ );
    const double u_decay = ( tau_fac_ == 0 ) ? 0.0 : std::exp( -h / tau_fac_ );  // tau_fac == 0 disables facilitation
```
I find this check here a bit confusing. Since `tau_fac_ == 0` is allowed to turn off facilitation (actually the default), I think it is most sensible to actually test for zero.

heplesser (changes requested):
@clinssen Thanks for this PR. It looks mostly good to me, but see my small suggestions inline. Could you add a little more text to the PR description to explain why/how things go wrong with the current implementation? Just for future reference.

models/tsodyks2_synapse.h
clinssen: There is the risk of a floating point underflow, but I don't think that will be a problem in practice (by default it shouldn't throw an exception).

PR #3402, opened 2025-02-10
Fix cmake issues finding the compiler

steffengraber:


steffengraber:
This PR was opened by mistake. Therefore, it will be closed immediately.

PR #3403, opened 2025-02-10
Remove deprecated environment variable from SLI init script

clinssen:
Remove undocumented and (seemingly) unused environment variable from SLI init script.


heplesser (approved):
Thanks for cleaning up :)!

PR #3405, opened 2025-02-14
Corrected unit of q_stc to pA in gif_psc_exp

Laxmi-gupta:
This is my first contribution, and I noticed that the adaptive current q_stc was mentioned in nA, but it should be in pA.

I made two changes:

Updated the comment to correctly state pA instead of nA.
Converted the value by multiplying it by 1000 to properly reflect pA in the calculation.

Changes made in:
models/gif_psc_exp.cpp
Line: 357

I haven’t tested this yet since it’s my first issue, but I followed the logic and made the necessary correction.

### DIFF
--- a/models/gif_psc_exp.cpp
+++ b/models/gif_psc_exp.cpp
@@ -354,7 +354,7 @@ nest::gif_psc_exp::update( Time const& origin, const long from, const long to )
     if ( S_.r_ref_ == 0 ) // neuron is not in refractory period
     {
 
-      S_.V_ = V_.P30_ * ( S_.I_stim_ + P_.I_e_ - S_.stc_ ) + V_.P33_ * S_.V_ + V_.P31_ * P_.E_L_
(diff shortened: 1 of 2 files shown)


Laxmi-gupta:
Thank you for your feedback and for pointing out the equivalence between defining q_stc in nA and pA. You're absolutely correct that mathematically, defining q_stc in nA and scaling it by 1e3 is equivalent to defining it directly in pA. However, this change is primarily rooted in consistency and clarity within the model.
Thank you also for the tip about running the testsuite locally with make installcheck—I’ll definitely make use of that.

clinssen:
Could you say in more detail what the consistency is between? There are many variants of the "gif" model that also define ``q_stc`` in nA, so to keep it consistent, they would all need to be changed. Could you give a reference to an equation or parameters table in the literature (like to the original publication of the model) where the value is given in pA? Much obliged!

clinssen:
Closing for now due to inactivity. If this is picked up again, we need to check the original publications to check which unit was used there, and decide whether we prefer to stick to the convention of the original publication, or the NEST convention of having currents in pA. 

Furthermore, changes need to be made in a way that covers all "gif" family models to ensure consistency.

PR #3420, opened 2025-02-28
Move critical section so it can be avoided by limiting verbosity

heplesser:
This PR moves the critical section in the SLI Interpreter's `message()` functions close to the actual output code, which needs protection. In that way, the critical section will only be encountered if the verbosity level is of the kernel is at least as "chatty" as the level of the message. By setting a level such as `M_FATAL`, the critical section can thus be avoided. 

For background, see https://github.com/nest/nest-simulator/issues/3217#issuecomment-2688688177.

### DIFF
--- a/sli/interpret.cc
+++ b/sli/interpret.cc
@@ -773,54 +773,45 @@ SLIInterpreter::terminate( int returnvalue )
 void
 SLIInterpreter::message( int level, const char from[], const char text[], const char errorname[] ) const
 {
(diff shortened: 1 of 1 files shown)


heplesser:
@heshpdx This PR should fix the nested critical region problem reported for the Cray compiler in #3217. Could you give it a test and review this PR? Setting verbosity to `M_FATAL` should avoid the critical region now.

JanVogelsang (approved):

PR #3423, opened 2025-03-04
Correctly detect that substring was not found

heplesser:
This PR updates ancient code to use modern C++ techniques to detect that a substring was not found (see https://en.cppreference.com/w/cpp/string/basic_string/npos).

This is related to #3217.

@bowers Could you review this change and see if it fixes the problems under Win32?


JanVogelsang (approved):

bowers:
I tested and this does fix the problems on Windows.  For posterity, I tested with MSVC 19 and with mingw/gcc13.1, both on Windows Server 2019.

PR #3428, opened 2025-03-20
Improve discoverablility of network models in NEST docs

jessica-mitchell:
This PR adds some additional links for network models in the models directory as well as in the examples

Output on Read the Docs: https://nest-simulator--3428.org.readthedocs.build/en/3428/index.html


doc/htmldoc/models/index.rst
```diff
@@ -5,8 +5,32 @@ Model directory
...
+      **Network Models**
+
+
+
+   .. grid-item-card::
+      :columns: 4
+      :link: model_selector
+      :link-type: ref
+
+      **Neurons, synapses, and devices**
```
jhnnsnk: Instead of these cards, I suggest to start the page with the section "What we mean by models" that is currently further down (https://nest-simulator--3428.org.readthedocs.build/en/3428/models/index.html#what-we-mean-by-models). This section should also mention devices.

jhnnsnk (changes requested):
Thanks  a lot for this PR!
The card on large-scale models in Examples looks good to me.
I've left a comment with the suggestion to reorder the content of the Models page.
On the start page, would it be possible to add links to the examples or sections shown automatically one after the other?

doc/htmldoc/models/index.rst
jessica-mitchell: This was discussed before, and the decision was made that we want to see the model selector near the top, so this was pushed further down. but maybe we can add a link at the top of the page with the cards like 'what's the difference?'

jessica-mitchell:
@jhnnsnk reminder

> Thanks a lot for this PR! The card on large-scale models in Examples looks good to me. I've left a comment with the suggestion to reorder the content of the Models page. On the start page, would it be possible to add links to the examples or sections shown automatically one after the other?

If I understand what you mean, you want to be able to click on the image/text in the carousel and go straight to that example? I don't exactly remember, but there was an issue with doing that. I can look into it again

doc/htmldoc/models/index.rst
jhnnsnk: @jessica-mitchell I think for a new user it might be unclear that the model selector does not refer to network models after showing the cards of both. Yes, the "What's the difference?" already helps. In addition, I suggest to extend "Select a tag to display corresponding models." to "Select a tag to display corresponding neuron, synapse or device models."

doc/htmldoc/models/index.rst
```diff
@@ -5,8 +5,34 @@ Model directory
...
 
+
+.. grid::
+
+   .. grid-item-card::
+      :columns: 4
+      :link: network_models
+      :link-type: ref
+
+      **Network Models**
```
jhnnsnk: Lower-case "models"?

doc/htmldoc/models/index.rst
jhnnsnk: Maybe "Neuron, synapse, and device models"?

jhnnsnk:
> @jhnnsnk I added links to the images in the carousel, and modified the models/index page text a bit

Looks and works great, thank you!

jhnnsnk (changes requested):
I left some tiny suggestions for text modifications, otherwise it's already great!

doc/htmldoc/models/index.rst
jessica-mitchell: Technically the devices are not 'models', at least how I understand we define them. I did add some text below in the 'what we  mean by models' section to explain a bit more. Hopefully without confusing it more.

doc/htmldoc/models/index.rst
jessica-mitchell:
```suggestion
      **Network models**
```

jhnnsnk (approved):
:+1: Looks good to me now, thank you!

jessica-mitchell:
merging with 1 reviewer because doc only change

PR #3439, opened 2025-03-24
Ported issue 1974 slitest to python

JanVogelsang:
There still is an issue with the Python-version, as it triggers the "assert_thread_parallel" assertion (line 207 in vp_manager.cpp). I am not sure why, @heplesser could you take a look at that?


testsuite/pytests/sli2py_mpi/test_issue_1974.py
```diff
@@ -0,0 +1,50 @@
...
+    specifically to the proxy on the rank on which the parrot neuron exists. Therefore, the pooled connections across
+    ranks must be identical independent of number of ranks.
+    """
+
+    import nest
+
+    if not nest.ll_api.sli_func("statusdict/have_music ::"):
+        return
+
+    nest.total_num_virtual_procs = 4
```
heplesser, line 39: @JanVogelsang  This line is not in the SLI version. Can you check if the test passes when you remove it? Then we could port this test. At the same time, could you then create a follow-up issue about MUSIC and threading not working fully?

JanVogelsang: Oh you are right, strange, not sure why I added this line.

The script (i.e., runner.py) indeed doesn't fail without that line, which also makes sense of course, but the test itself still fails. How can I debug this?

JanVogelsang: The issue is not related to MUSIC as it seems, I created a minimal reproducer (#3442 ).

heplesser: I don't have music installed on my computer. Could you run `runner.py` for 1, 2, 4 MPI processes and post the resulting OTHER files?

JanVogelsang: Sure, I can't find the files though. Should they be written in the same temp directory where I find the runner.py file?

JanVogelsang: The OTHER files look sane (identical if one would combine the files from all processes and sort them), but somehow the test still fails, even if not setting the number of VPs to 4.
If one explicitly sets the number of VPs to 4, the original bug is now fixed with PR #3445, but now there is another issue, which I described in issue #3446.

heplesser:
@JanVogelsang This one fails currently on macOS, I believe, because we skip the test when Music is not present, which confuses the result collection process. I will need to think about the skipping logic for these MPI tests once again.

It also seems that this PR contains many more files than it should. It would be good to wait with this one until we have #3450 merged.

JanVogelsang:
Oh true, this PR now also contains changes that #3445 contains, because that fix was required for this test to work. So we have to wait until that one is merged. #3450 is not a blocker for this one. While they are related, fixing it is not required for this test to work.

heplesser:
@JanVogelsang Could you review #3467? Once that is in place it should be straightforward to implement music-skipping by decorator here.

JanVogelsang:
@heplesser I implemented the music-skipping decorator and added it to the isuee_1974 python test.

JanVogelsang:
Blocked by #3470

heplesser:
> @JanVogelsang This looks in principle good to me now, but I would prefer to merge #3450 first so that the change set for this PR becomes smaller, reduced to just the port of the test and the missing-music decorator.

Sorry, I meant to merge #3445 first, not #3450.

JanVogelsang:
@heplesser Can you rerun the test-suite now that the CI issue is merged?

JanVogelsang:
> > @JanVogelsang This looks in principle good to me now, but I would prefer to merge #3450 first so that the change set for this PR becomes smaller, reduced to just the port of the test and the missing-music decorator.
> 
> Sorry, I meant to merge #3445 first, not #3450.

Yes, I agree. We just need Dennis to take a very quick look at #3445

heplesser:
@JanVogelsang If think you need to merge master into your branch for this PR so the Ubuntu-update in master takes effect here. That should automatically trigger the CI.

PR #3445, opened 2025-03-26
Fix for issue 3442

JanVogelsang:
This PR fixes issue #3442.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1219,14 +1219,12 @@ nest::ConnectionManager::split_to_neuron_device_vectors_( const size_t tid,
 
 void
 nest::ConnectionManager::get_connections_( const size_t tid,
-  std::deque< ConnectionID >& connectome,
-  NodeCollectionPTR source,
-  NodeCollectionPTR target,
+  std::deque< ConnectionID >& conns_in_thread,
+  NodeCollectionPTR,
+  NodeCollectionPTR,
   synindex syn_id,
   long synapse_label ) const
 {
-  std::deque< ConnectionID > conns_in_thread;
-
   ConnectorBase* connections = connections_[ tid ][ syn_id ];
   if ( connections )
   {
@@ -1240,26 +1238,16 @@ nest::ConnectionManager::get_connections_( const size_t tid,
   }
 
   target_table_devices_.get_connections( 0, 0, tid, syn_id, synapse_label, conns_in_thread );
-
-  if ( conns_in_thread.size() > 0 )
-  {
-#pragma omp critical( get_connections )
-    {
-      extend_connectome( connectome, conns_in_thread );
-    }
-  }
 }
 
 void
 nest::ConnectionManager::get_connections_to_targets_( const size_t tid,
-  std::deque< ConnectionID >& connectome,
-  NodeCollectionPTR source,
+  std::deque< ConnectionID >& conns_in_thread,
+  NodeCollectionPTR,
   NodeCollectionPTR target,
   synindex syn_id,
   long synapse_label ) const
 {
-  std::deque< ConnectionID > conns_in_thread;
-
   // Split targets into neuron- and device-vectors.
   std::vector< size_t > target_neuron_node_ids;
   std::vector< size_t > target_device_node_ids;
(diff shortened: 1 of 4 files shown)


testsuite/pytests/test_issue_3442.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import nest
+import pytest
+
+"""
+Tests that get_connections works if called when also using multithreading.
```
heplesser, line 27:
```suggestion
Tests that get_connections works if called when also using multithreading.

This test would only fail before fixing #3442 if NEST was built with `-Dwith-detailed-timers=ON`.
```

testsuite/pytests/test_issue_3442.py
```diff
@@ -0,0 +1,44 @@
...
+
+    nest.total_num_virtual_procs = n_threads
+
+    neuron = nest.Create("parrot_neuron")
+    nest.Connect(neuron, neuron)
+
+    nest.GetConnections()
+
+
+test_threaded_get_connections(1)
```
heplesser, line 44: I'd remove these so we have a clean test. If one wants to run it from the command line, `pytest -vs ...` will do.
```suggestion
```

nestkernel/connection_manager.cpp
```diff
@@ -1218,188 +1218,202 @@ nest::ConnectionManager::split_to_neuron_device_vectors_( const size_t tid,
...
+      const size_t source_node_id = source_table_.get_node_id( tid, syn_id, lcid );
+      connections->get_connection( source_node_id, 0, tid, lcid, synapse_label, conns_in_thread );
+    }
   }
 
-  const size_t num_connections = get_num_connections( syn_id );
+  target_table_devices_.get_connections( 0, 0, tid, syn_id, synapse_label, conns_in_thread );
 
-  if ( num_connections == 0 )
+  if ( conns_in_thread.size() > 0 )
```
heplesser, line 1244: This part is in all three `get_connections*_()` methods. I'd suggest to move it to the outer method and then define `conns_in_threads` in that method and pass it into the "submethods" instead of `connectome`. That avoids code duplication.

nestkernel/connection_manager.cpp
```diff
@@ -1218,188 +1218,202 @@ nest::ConnectionManager::split_to_neuron_device_vectors_( const size_t tid,
 }
 
 void
-nest::ConnectionManager::get_connections( std::deque< ConnectionID >& connectome,
+nest::ConnectionManager::get_connections_( const size_t tid,
+  std::deque< ConnectionID >& connectome,
   NodeCollectionPTR source,
```
heplesser, line 1223: Leaving out the unused name should avoid the warnings.
```suggestion
  NodeCollectionPTR,
```

nestkernel/connection_manager.cpp
```diff
@@ -1218,188 +1218,202 @@ nest::ConnectionManager::split_to_neuron_device_vectors_( const size_t tid,
...
   NodeCollectionPTR target,
```
heplesser, line 1224:
```suggestion
  NodeCollectionPTR,
```

nestkernel/connection_manager.cpp
```diff
@@ -1218,188 +1218,202 @@ nest::ConnectionManager::split_to_neuron_device_vectors_( const size_t tid,
...
+    {
+      extend_connectome( connectome, conns_in_thread );
+    }
   }
+}
+
+void
+nest::ConnectionManager::get_connections_to_targets_( const size_t tid,
+  std::deque< ConnectionID >& connectome,
+  NodeCollectionPTR source,
```
heplesser, line 1256:
```suggestion
  NodeCollectionPTR,
```

heplesser (changes requested):
Looks good already. I have a few suggestions for further improvement. I also wonder if we should go "all in" in the CI and in our "everything" test also activate detailed and mpi-sync timer. That will slow the testsuite a little, but at least we test all. That would require adding something like "detailed_timers" and "mpi_sync_timer" to https://github.com/nest/nest-simulator/blob/09b349c55298d346e8aceea4f0ae643dea8ee7a3/.github/workflows/nestbuildmatrix.yml#L530 and append the corresponding switches here: https://github.com/nest/nest-simulator/blob/09b349c55298d346e8aceea4f0ae643dea8ee7a3/.github/workflows/nestbuildmatrix.yml#L681

JanVogelsang:
I agree, adding detailed timers and the mpi-sync-timer to the "full" CI testcase makes sense. It shouldn't slow it down too much, so we should be fine.

testsuite/pytests/test_issue_3442.py
JanVogelsang: Done

heplesser (approved):

heplesser:
@terhorstd Could you take a look at this one especially with a view to the changes in the build matrix?

heplesser:
@terhorstd Ping!

JanVogelsang:
@terhorstd You only need to review the changes in `.github/workflows/nestbuildmatrix.yml`, which should roughly take 12 seconds.

JanVogelsang:
Ping

otcathatsya (approved):

PR #3449, opened 2025-03-27
Fix incorrect conversion of string inf to double seen on MinGW

bowers:
This PR changes the string-to-double conversion when reading SLI from std::atoi() to std::strtod(). This resolves an error observed in MinGW in which the value "inf" is converted to 0 instead of infinity. 

This is related to https://github.com/nest/nest-simulator/issues/3217.

### DIFF
--- a/sli/slidata.cc
+++ b/sli/slidata.cc
@@ -1383,7 +1383,7 @@ Cvd_sFunction::execute( SLIInterpreter* i ) const
 
   StringDatum* obj = dynamic_cast< StringDatum* >( i->OStack.top().datum() );
   assert( obj );
-  Token t( new DoubleDatum( std::atof( obj->c_str() ) ) );
+  Token t( new DoubleDatum( std::strtod( obj->c_str(), nullptr ) ) );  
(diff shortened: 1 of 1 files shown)


sli/slidata.cc
```diff
@@ -1383,7 +1383,7 @@ Cvd_sFunction::execute( SLIInterpreter* i ) const
 
   StringDatum* obj = dynamic_cast< StringDatum* >( i->OStack.top().datum() );
   assert( obj );
-  Token t( new DoubleDatum( std::atof( obj->c_str() ) ) );
+  Token t( new DoubleDatum( std::strtod( obj->c_str(), nullptr ) ) );
```
heplesser, line 1386: Just a tiny fix to make our code format checker happy:
```suggestion
  Token t( new DoubleDatum( std::strtod( obj->c_str(), nullptr ) ) );
```

heplesser (changes requested):
Thanks for the PR, looks good except for two stray spaces.

heplesser (approved):

Issue #3446, opened 2025-03-26
Segfault when using MUSIC with more than 3 threads


JanVogelsang:
The following code snipped produces a segmentation fault, as it tries to access a non-existent Connector in line 218 of `target_table_devices.h` (`target_from_devices_[ tid ][ ldid ][ syn_id ]->get_synapse_status( tid, lcid, dict );`):

```python
nest.local_num_threads = 3

nrns = nest.Create("parrot_neuron", 2)

music_in = nest.Create("music_event_in_proxy", 1, {"port_name": "in_spikes"})
music_out = nest.Create("music_event_out_proxy", 1, {"port_name": "out_spikes"})

nest.Connect(music_in, nrns)
nest.Connect(nrns, music_out)

nest.GetConnections().get()
```

med-ayssar:
@heplesser (CC @JanVogelsang ), I tried to debug it, and I had some interesting issues. Somehow I think I forgot a bit about the logic behind devices and proxies in the context of MUSIC (I know that neurons on thread `i` as none-device models are represented as proxies on the other threads `j =! i`), and devices are replicated over all threads), however; MUSIC devices (according to `add_music` func) are only added to thread `t=0`, no proxies!

While debugging I found those issues:

1.  if we **first** create `music_event_in_proxy` MUSIC device-alike models, then we get a different `segfault`.
2.  if we set `local_num_threads = 2`, the result of `getConnections` is wrong, compared to `local_num_threads = 1` ( 1X edge is missing and 1X is duplicated, independent of [1])

Question: is a MUSIC node, considered a device, and should it be added to the logic of `target_to_devices_` and `target_from_devices_`?

```cpp
void
nest::ConnectionManager::connect( const size_t snode_id,
  Node* target,
  size_t target_thread,
  const synindex syn_id,
  const DictionaryDatum& params,
  const double delay,
  const double weight )
{
  kernel().model_manager.assert_valid_syn_id( syn_id, kernel().vp_manager.get_thread_id() );

  Node* source = kernel().node_manager.get_node_or_proxy( snode_id, target_thread );

  ConnectionType connection_type = connection_required( source, target, target_thread );

  switch ( connection_type )
  {
  case CONNECT:
    connect_( *source, *target, snode_id, target_thread, syn_id, params, delay, weight );
    break;
  case CONNECT_FROM_DEVICE:
    connect_from_device_( *source, *target, target_thread, syn_id, params, delay, weight );
    break;
[...]

PR #3450, opened 2025-03-27
Add regression test for issue #3446

JanVogelsang:
This PR adds a regression test for issue #3446. Please fork this branch to implement a fix for the issue.

### DIFF
--- a/testsuite/pytests/test_issue_3446.py
+++ b/testsuite/pytests/test_issue_3446.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+#
+# test_issue_3446.py
+#
+# This file is part of NEST.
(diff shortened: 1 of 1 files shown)


terhorstd:
Is this still active? If so:
Could you comment on why a fix should not be based on latest master? What other changes are injected when starting from your branch, as you suggest?

JanVogelsang:
This branch, which is based on master, makes the CI fail by adding a regression test.

terhorstd:
This PR will stay available and is linked directly from the issue. There is no use having it as draft and continuously occupy our attention on the PR board. It's still an issue, hence we find it in the issues.
I'll close this for now, please re-open when works are picked up again.

PR #3456, opened 2025-04-02
Port regression test 659 to Pytests

jessica-mitchell:
This conversion was done with Mistral AI.


testsuite/pytests/sli2py_regressions/test_issue_659.py
```diff
@@ -0,0 +1,68 @@
...
+import nest
+import pytest
+
+
+def test_run_without_prepare():
+    """
+    Test that calling Run without Prepare results in an error.
+    """
+    nest.ResetKernel()
+    with pytest.raises(Exception):
```
heplesser, line 36: We can and should be more specific here:
```suggestion
    with pytest.raises(nest.kernel.NESTError):
```
This applies also to the remaining cases.

heplesser (changes requested):
This looks fine, but we can make the test a bit more specific, see inline comment.

heplesser (approved):

PR #3460, opened 2025-04-03
Fix corner cases uncovered by Clang sanitizer

heplesser:
This PR fixes some errors that popped up when building NEST with a range of Clang sanitizers (see https://github.com/nest/nest-simulator/issues/3217#issuecomment-2774216910). Specifically, NEST was configured with

```
 -Dwith-optimize="-g -O3 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize=address,undefined -fsanitize=float-divide-by-zero,unsigned-integer-overflow,implicit-conversion,local-bounds  -fno-sanitize-recover=undefined -fsanitize-recover=float-divide-by-zero -fsanitize-address-use-after-return=always -fsanitize-address-use-after-scope"
```

## Notes
- Running the SLI-based parts of the testsuite works out of the box under macOS.
- Running the Python-based parts is more complicated. Under macOS, the system will complain that the address sanitizer is loaded too late and gives suggestions how to fix this. Based on this, the following works (adjust the path to pytests):
```
DYLD_INSERT_LIBRARIES=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/17/lib/darwin/libclang_rt.asan_osx_dynamic.dylib python -m pytest <PATH>/pytests
```
- The sanitizers will always report integer overflows from randutils.hpp. This can be ignored.
- If the sanitizer detects a problem with, pytesting will terminate immediately with a relatively uninformative stack trace.
- The only way to get a good stack trace is to use lldb:
```
DYLD_INSERT_LIBRARIES=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/17/lib/darwin/libclang_rt.asan_osx_dynamic.dylib lldb -- python -m pytest pytests/sli2py_connect/test_getconnections_multiple_syn_models.py
```
- When running with lldb, execution will stop at each of the integer overflows from randutils.hpp. One then needs to `c(ontinue)` until one gets to the actual problem.
- I have gotten to the point where all SLI based tests (including mpi tests) and all plain (non-mpi) Python test pass.

### DIFF
--- a/libnestutil/config.h.in
+++ b/libnestutil/config.h.in
@@ -96,10 +96,10 @@
 /* Define to 1 if you have usable long long type. */
 #cmakedefine HAVE_LONG_LONG 1
 
-/* Define to 1 if you have u_int46_t type. */
+/* Define to 1 if you have u_int64_t type. */
 #cmakedefine HAVE_U_INT64_T 1
 
-/* Define to 1 if you have uint46_t type. */
+/* Define to 1 if you have uint64_t type. */
 #cmakedefine HAVE_UINT64_T 1
 
 /* Define to 1 if you have the <mach/mach.h> header file. */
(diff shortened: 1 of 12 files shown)


JanVogelsang (approved):

otcathatsya (approved):
Getting some signedness comparison warnings (int <-> long) but assuming those are in a context where that'd never be an issue it looks good

PR #3464, opened 2025-04-09
Fix example for EI clustered network

jessica-mitchell:
The filenames are renamed so the example can run properly


jessica-mitchell:
Merge with no extra review as its doc related

Issue #3466, opened 2025-04-10
Use OpenMPI 5.0.7 or later for CI testing under Linux


heplesser:
OpenMPI seems to contain a bug at least up to 4.1.6 which makes our new MPI Test setup complicated, because with the bug present, one cannot subprocess-mpirun anything after a module (such as NEST) which call MPI_Init() has been imported, e.g. in Pytest's `conftest.py` file. The error is definitely fixed in OpenMPI 5.0.7. See also https://github.com/python/cpython/issues/132377.

Under macOS, Github actions already use OpenMPI 5.0.7, so tests pass there (see, e.g., https://github.com/heplesser/nest-simulator/actions/runs/14229583317/job/39877167366), but under Linux we still use 4.1.2 (see https://github.com/heplesser/nest-simulator/actions/runs/14229583317/job/39877167373#step:7:414), leading to test failure.

There are cumbersome ways to work around this problem (such as adding `nest.finalize_mpi()` just for testing), but I would much rather use OpenMPI 5.0.7 for testing that to hack in such a function.

Can we organise this without having to recompile OpenMPI manually every time (which would waste a lot of electricity)? Alternatively, we could skip the mpitests for OpenMPI < 5.0.7. Since macOS uses 5.0.7 or later, the mpitests would definitely be run, so the risk of this is not too bad.

heplesser:
I took this down from "Critical" to "Normal" again after I realized that I can easily implement a pytest skip decorator based on the reproducer I created for https://github.com/python/cpython/issues/132377 by putting into conftest.py something like

```
import nest   # is in place
import subprocess

res = subprocess.run(["mpirun", "-np", "1", "echo"])
need_to_skip = res.retcode != 0
```

PR #3467, opened 2025-04-12
Improved wrappers and OpenMPI handling in Python MPI tests

heplesser:
This PR fixes the problem we had with `subprocess` and `mpirun`. Thus, mpitests can now be run with `conftest.py` and the usual "missing" decorators can be applied to MPI tests.

See also https://github.com/nest/nest-simulator/issues/3466.

### DIFF
--- a/testsuite/pytests/sli2py_mpi/test_self_get_conns_with_empty_ranks.py
+++ b/testsuite/pytests/sli2py_mpi/test_self_get_conns_with_empty_ranks.py
@@ -26,9 +26,11 @@ from mpi_test_wrapper import MPITestAssertEqual
 # Parametrization over the number of nodes here only to show hat it works
 @pytest.mark.skipif_incompatible_mpi
 @MPITestAssertEqual([1, 2, 4], debug=False)
-def test_self_get_conns_with_empty_ranks():
+def test_get_conns_with_empty_ranks():
     """
-    Selftest: Confirm that connections can be gathered correctly even if some ranks have no neurons.
+    Confirm that connections can be gathered correctly even if some ranks have no neurons.
+
+    The test compares connection data written to OTHER_LABEL.
     """
 
     import nest
--- a/testsuite/pytests/sli2py_mpi/test_self_get_conns_with_empty_ranks.py
+++ b/testsuite/pytests/sli2py_mpi/test_self_get_conns_with_empty_ranks.py
@@ -19,13 +19,12 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-import numpy as np
-import pandas
 import pytest
 from mpi_test_wrapper import MPITestAssertEqual
 
 
 # Parametrization over the number of nodes here only to show hat it works
+@pytest.mark.skipif_incompatible_mpi
 @MPITestAssertEqual([1, 2, 4], debug=False)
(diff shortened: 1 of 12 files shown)


testsuite/pytests/sli2py_mpi/test_self_get_conns_with_empty_ranks.py
```diff
@@ -19,13 +19,12 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-import numpy as np
-import pandas
```
JanVogelsang, line 23: Was this intentional?

JanVogelsang (approved):

testsuite/pytests/sli2py_mpi/test_self_get_conns_with_empty_ranks.py
heplesser: Yes, the imports are not needed here. I just pushed another commit that removes superfluous imports in other tests as well, removes old-style GSL-checking and adds comments on data generated to all test docstrings.

nicolossus (approved):
Nice, LGTM!

One suggestion you may consider (maybe in a future PR): As I understand the documentation, you can use the [pytest_ignore_collect](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_ignore_collect) hook in `conftest.py` to skip tests in a specified directory based on a condition. This should work well for the MPI tests since they are in a dedicated directory, and would save you from having to decorate every MPI test function with `skipif_incompatible_mpi`.

heplesser:
> Nice, LGTM!
> 
> One suggestion you may consider (maybe in a future PR): As I understand the documentation, you can use the [pytest_ignore_collect](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_ignore_collect) hook in `conftest.py` to skip tests in a specified directory based on a condition. This should work well for the MPI tests since they are in a dedicated directory, and would save you from having to decorate every MPI test function with `skipif_incompatible_mpi`.

Thanks for the hint, I had been looking for some way to skip entire directories. But I think it is not exactly what I would like, because if I understand this right, it would completely ignore the directory, so the test placed in it would be invisible in the test report. I would like them to be reported as skipped, so one is aware of the fact that tests were not run.

nicolossus:
> > Nice, LGTM!
> > One suggestion you may consider (maybe in a future PR): As I understand the documentation, you can use the [pytest_ignore_collect](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_ignore_collect) hook in `conftest.py` to skip tests in a specified directory based on a condition. This should work well for the MPI tests since they are in a dedicated directory, and would save you from having to decorate every MPI test function with `skipif_incompatible_mpi`.
> 
> Thanks for the hint, I had been looking for some way to skip entire directories. But I think it is not exactly what I would like, because if I understand this right, it would completely ignore the directory, so the test placed in it would be invisible in the test report. I would like them to be reported as skipped, so one is aware of the fact that tests were not run.

Good point. This would skip collecting the tests, so they will probably not show up in the summary. Another alternative could be to skip the tests at module level with `pytest.skip(reason, allow_module_level=True)`, see https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-test-functions

JanVogelsang (approved):

PR #3488, opened 2025-04-25
Update the data serialization in hl_api_server.py

vvorobjov:
int64 cannot be serialized by default jsonify,

this change adds the serialization of all int* types to `serialize_data` function and uses it before sending the server response

### DIFF
--- a/pynest/nest/lib/hl_api_types.py
+++ b/pynest/nest/lib/hl_api_types.py
@@ -1290,8 +1290,6 @@ def serialize_data(data):
                             numpy.int16, numpy.int32, numpy.int64, numpy.uint8,
                             numpy.uint16, numpy.uint32, numpy.uint64)):
         return int(data)
-    if isinstance(data, (numpy.ndarray, NodeCollection)):
-        return data.tolist()
     elif isinstance(data, SynapseCollection):
         # Get full information from SynapseCollection
         return serialize_data(data.get())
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_types.py
```diff
@@ -1286,6 +1286,10 @@ def serialize_data(data):
 
     if isinstance(data, (numpy.ndarray, NodeCollection)):
         return data.tolist()
+    if isinstance(data, (numpy.int_, numpy.intc, numpy.intp, numpy.int8,
+                            numpy.int16, numpy.int32, numpy.int64, numpy.uint8,
+                            numpy.uint16, numpy.uint32, numpy.uint64)):
+        return int(data)
```
med-ayssar, line 1292: You could maybe make use of `np.integer`, instead of explicitly listing most of the integer types.
```suggestion
    if np.issubdtype(d, np.integer):
        return int(data)
```

vvorobjov: Yes, indeed, it works as well, thanks!

heplesser:
@babsey Ping for review!

babsey (changes requested):
LGTM!

I would suggest a code at line 264 in `pynest/nest/server/hl_api_server.py`:
`    response = [nest.serialize_data(master_response)]`

vvorobjov:
@babsey the proposed changes are added

babsey (approved):
LGTM!

PR #3490, opened 2025-04-30
Update built-in timer documentation

jessica-mitchell:
This PR adds a diagram describing a simplified simulation run focusing on timers.
The diagram is generated via a script that converts YAML into a dot diagram. Thanks to @JoseJVS for providing the original version.
In addition, some links are added to try to connect various parts of the documentation together with respect to perforance, benchmarking, and timers.

https://nest-simulator--3490.org.readthedocs.build/en/3490/nest_behavior/built-in_timers.html#built-in-timers


mlober (changes requested):
Thank you for this, it already looks really good. I had a look at the code again and have a few small improvements:
* diagram: change "Repeat once if buffer too small" to "Repeat once if not all spikes transmitted". 
* diagram: block "if buffer can shrink" must be outside of "Repeat once if not all spikes transmitted".
* diagram: include another block after time_communicate_spike_data-block that is the opposite of "if buffer can shrink": "if buffer too small -> grow MPI buffer size".
* OMP parallel and OMP master with same capitalization
* question to @JanVogelsang : shouldn't the OMP parallel end before OMP master?

JanVogelsang (approved):

JanVogelsang:
I'm not sure if "Repeat once if not all spikes transmitted" is actually better than "Repeat once if buffer too small". That makes it sound as if in the first round, the communicated spikes are already used and only the remaining spikes are communicated in the second round.

mlober (approved):

Issue #3495, opened 2025-05-15
Latest Cython release breaks NEST build


clinssen:
Cython 3.1.0 was released on 2025-05-08. It deprecates the "long" data type in favour of "int" (see https://cython.readthedocs.io/en/latest/src/changes.html).

This results in the following error during the NEST build:
```
pynest/pynestkernel.pyx
Error compiling Cython file:
------------------------------------------------------------
...

    cdef string obj_str

    if isinstance(obj, bool):
        ret = <Datum*> new BoolDatum(obj)
    elif isinstance(obj, (int, long)):
                               ^
------------------------------------------------------------

/home/runner/work/nestml/nestml/nest-simulator/pynest/pynestkernel.pyx:364:31: undeclared name not builtin: long
```

I believe that fixing this should involve only replacing "long" with "int" in ``pynestkernel.pyx``.

clinssen:
N.B.: just to note that the fix needs to be backported to older NEST releases as well. For instance, at present, NESTML tests run in GitHub Actions CI for NEST versions 2.20.2, 3.0, 3.8 and master.

clinssen:
Yes, that would probably avoid the issue, but if that is the fix that we end up going for, then I assume we would prefer to put that constraint upstream, i.e. into NEST's own requirements.txt (and even backport this update to requirements.txt).

heplesser:
@clinssen I cannot find the "long" deprecation in the Cython release notes. Could you point me to where in the release notes this is mentioned? I don't doubt the reality of it, but if we mention the release notes in the issue, I'd want to find the matter :).

PR #3497, opened 2025-05-23
Fix issue with latest Cython release

steffengraber:
There is currently an error in the CI:
```

Error compiling Cython file:
------------------------------------------------------------
...

    cdef string obj_str

    if isinstance(obj, bool):
        ret = <Datum*> new BoolDatum(obj)
    elif isinstance(obj, (int, long)):
                               ^
------------------------------------------------------------

/home/runner/work/nest-simulator/nest-simulator/pynest/pynestkernel.pyx:364:31: undeclared name not builtin: long

```
As a quick fix, i have removed the 'long' type check from python_object_to_datum function.
This allows the CI to function again.

Fixes issue #3495.

### DIFF
--- a/pynest/pynestkernel.pyx
+++ b/pynest/pynestkernel.pyx
@@ -361,7 +361,7 @@ cdef inline Datum* python_object_to_datum(obj) except NULL:
 
     if isinstance(obj, bool):
         ret = <Datum*> new BoolDatum(obj)
-    elif isinstance(obj, (int, long)):
+    elif isinstance(obj, int):  # removed 'long'
         ret = <Datum*> new IntegerDatum(obj)
     elif isinstance(obj, float):
         ret = <Datum*> new DoubleDatum(obj)


pynest/pynestkernel.pyx
```diff
@@ -361,7 +361,7 @@ cdef inline Datum* python_object_to_datum(obj) except NULL:
 
     if isinstance(obj, bool):
         ret = <Datum*> new BoolDatum(obj)
-    elif isinstance(obj, (int, long)):
+    elif isinstance(obj, int):  # removed 'long'
```
heplesser, line 364: I think the comment is not really needed here, we always have the git history:
```suggestion
    elif isinstance(obj, int):
```

heplesser (changes requested):
@steffengraber Thanks! I have just one little suggestion.

heplesser:
Adding @clinssen as issue-creator and @terhorstd for question about backports, see #3495.

steffengraber:
> What about all the other mentions of `long` in `pynestkernel.pyx`? If they do not need to be replaced, why not—why do they not trigger the issue?

@clinssen I asked myself the same questions, but couldn't find an answer. Maybe @heplesser or @gtrensch  can help. All I can say is that this small change fixes the CI.

PR #3511, opened 2025-06-25
Port regression test 451 to Pytests

jessica-mitchell:
Using AI


testsuite/pytests/sli2py_regressions/test_ticket_451.py
```diff
@@ -0,0 +1,88 @@
...
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+import nest
+import pytest
+
+
+def test_ticket_451():
+    """
+    Guard against infinite loops in Random*Connect.
+
```
jessica-mitchell, line 72:
```suggestion

    Test ported from SLI regression test
```

PR #3513, opened 2025-06-25
Port regression test 464 to Pytests

jessica-mitchell:
Uses AI

### DIFF
--- a/testsuite/pytests/sli2py_regressions/test_ticket_464.py
+++ b/testsuite/pytests/sli2py_regressions/test_ticket_464.py
@@ -37,7 +37,7 @@ def skip_if_have_mpi(func):
 def test_ticket_464():
     """
     Ensure that NEST triggers an error with a frozen multimeter connected to a node.
-    
+
     Test ported from SLI regression test
     This test has been modified to reflect that NEST now protects multimeter against being frozen.
     The first test triggers an error instead of an assertion, and crash_or_die has been replaced by fail_or_die.


testsuite/pytests/sli2py_regressions/test_ticket_464.py
```diff
@@ -0,0 +1,70 @@
...
+        return pytest.mark.skip(reason="Test skipped if NEST has MPI support")(func)
+    except ImportError:
+        return func
+
+
+@skip_if_have_mpi
+def test_ticket_464():
+    """
+    Ensure that NEST triggers an error with a frozen multimeter connected to a node.
+
```
jessica-mitchell, line 40:
```suggestion
    
    Test ported from SLI regression test
```

testsuite/pytests/sli2py_regressions/test_ticket_464.py
```diff
@@ -0,0 +1,71 @@
...
+    Test ported from SLI regression test
+    This test has been modified to reflect that NEST now protects multimeter against being frozen.
+    The first test triggers an error instead of an assertion, and crash_or_die has been replaced by fail_or_die.
+
+    Remarks:
+    This test has been modified (2011-02-11) to reflect the fact that NEST now protects
+    multimeter against being frozen. Thus, the first test triggers an error instead of
+    and assertion, and crash_or_die has been replaced by fail_or_die.
```
noah-ostendorf, line 48: Remarks of the original test have been duplicated. Hint that test will not be executed with MPI support is missing. 

```suggestion
    Test ported from SLI regression test
    This test has been modified to reflect that NEST now protects multimeter against being frozen.
    The first test triggers an error instead of an assertion, and crash_or_die has been replaced by fail_or_die.

    Remarks:
    This test has been modified (2011-02-11) to reflect the fact that NEST now protects
    multimeter against being frozen. Thus, the first test triggers an error instead of
    and assertion, and crash_or_die has been replaced by fail_or_die.
```

    Remarks:
    This test has been modified (2011-02-11) to reflect the fact that NEST now protects
    multimeter against being frozen. Thus, the first test triggers an error instead of
    and assertion, and crash_or_die has been replaced by fail_or_die.
    
    This test will only be executed if NEST has been compiled without MPI support.
```

PR #3516, opened 2025-06-25
Port regression test 478 to Pytests

jessica-mitchell:
Uses AI cursor
NOTE:  Initial conversion failed the pytest, but then the AI suggested adding try /except blocks - pytest then passed but I think these parts of the code should be carefully checked.


testsuite/pytests/sli2py_regressions/test_ticket_478.py
```diff
@@ -0,0 +1,160 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Regression test for Ticket #478.
+
+Ensure that devices can only be connected using static synapses.
+
```
jessica-mitchell, line 26:
```suggestion

Test ported from SLI regression test
```

clinssen (approved):

testsuite/pytests/sli2py_regressions/test_ticket_478.py
```diff
@@ -0,0 +1,156 @@
...
+Test ported from SLI regression test.
+
+Ensure that NEST throws an exception if one tries to connect poisson_generator
+(sending DSSpikeEvents), noise_generator (sending DSCurrentEvents) or multimeter
+(sending DataLoggingRequest) to a neuron using a plastic synapse.
+"""
```
jessica-mitchell, line 35:
```suggestion
spike_generator slips through this test, since it usually sends SpikeEvent. But it
is no danger if it sends DSSpikeEvents for weighted spikes, since it sends precisely
one DSSpikeEvent per spike.
"""
```

PR #3524, opened 2025-07-10
PyNEST-NG: A new Python interface for NEST directly connecting Python to C++

heplesser:
A detailed description will follow. I create the PR now to allow references to it.


nestkernel/kernel_manager.cpp
```diff
@@ -42,7 +157,7 @@ nest::KernelManager::create_kernel_manager()
 void
 nest::KernelManager::destroy_kernel_manager()
 {
-  kernel_manager_instance_->logging_manager.set_logging_level( M_QUIET );
+  assert( false ); // Just to check if this ever gets called
```
akorgor: Is `assert( false )` a remainder of debugging? Also, is deleting `kernel_manager_instance_->logging_manager.set_logging_level( M_QUIET );` intentional?

nestkernel/layer.h
```diff
@@ -234,10 +232,10 @@ class Layer : public AbstractLayer
    *
    * @param d Dictionary with named parameter settings.
    */
-  void set_status( const DictionaryDatum& ) override;
+  void set_status( const Dictionary& ) override;
 
   //! Retrieve status, slice according to node collection if given
-  void get_status( DictionaryDatum&, NodeCollection const* ) const override;
+  void get_status( Dictionary&, NodeCollection const* const ) const override;
```
akorgor, line 238: Why the extra `const` here?

nestkernel/recordables_map.h
```diff
@@ -107,7 +103,7 @@ class RecordablesMap : public std::map< Name, double ( HostNode::* )() const >
 private:
   //! Insertion functions to be used in create(), adds entry to map and list
   void
-  insert_( const Name& n, const DataAccessFct f )
+  insert_( const std::string& n, const DataAccessFct f )
```
akorgor, line 106: In l. 112 there is a `LiteralDatum` and in l. 122 an `ArrayDatum` remaining.

nestkernel/recordables_map.h
```diff
@@ -90,13 +86,13 @@ class RecordablesMap : public std::map< Name, double ( HostNode::* )() const >
    *       build the list every time, even though that beats the
    *       goal of being more efficient ...
    */
-  ArrayDatum
+  std::vector< std::string >
```
akorgor, line 89: In l. 82 there is SLI mentioned in the comment.

nestkernel/recordables_map.h
akorgor, line 106: In l. 116 there is SLI mentioned in the comment.

nestkernel/recordables_map.h
```diff
@@ -199,32 +195,31 @@ class DynamicRecordablesMap : public std::map< Name, const DataAccessFunctor< Ho
...
```
akorgor, line 198: In l. 191 there is SLI mentioned in the comment.

nestkernel/sp_manager.h
```diff
@@ -103,8 +100,8 @@ class SPManager : public ManagerInterface
    */
   void disconnect( NodeCollectionPTR sources,
     NodeCollectionPTR targets,
-    DictionaryDatum& conn_spec,
-    DictionaryDatum& syn_spec );
+    const Dictionary& conn_spec,
+    const std::vector< Dictionary >& syn_spec );
```
akorgor: Maybe this should be named `syn_specs` to reflect that it holds multiple items and to match the name used in the .cpp file.

nestkernel/source_table.cpp
```diff
@@ -314,10 +314,10 @@ nest::SourceTable::source_should_be_processed_( const size_t rank_start,
 {
   const size_t source_rank = kernel().mpi_manager.get_process_id_of_node_id( source.get_node_id() );
 
-  return not( source.is_processed()
-    or source.is_disabled()
-    // is this thread responsible for this part of the MPI buffer?
-    or source_rank < rank_start or rank_end <= source_rank );
+  // is this thread responsible for this part of the MPI buffer?
```
akorgor: This comment reads as if it was a debug comment. Maybe better:

`// Check if thread responsible for this part of the MPI buffer.`

nestkernel/recording_backend_sionlib.cpp
```diff
@@ -206,7 +201,7 @@ nest::RecordingBackendSIONlib::open_files_()
           "Please change data_path, or data_prefix, or set /overwrite_files "
           "to true in the root node.",
           filename );
-        LOG( M_ERROR, "RecordingBackendSIONlib::open_files_()", msg );
+        LOG( nest::VerbosityLevel::ERROR, "RecordingBackendSIONlib::open_files_()", msg );
```
akorgor: `nest::` here necessary?

nestkernel/recording_backend_sionlib.cpp
```diff
@@ -593,7 +584,8 @@ nest::RecordingBackendSIONlib::SIONBuffer::write( const char* v, size_t n )
   else
   {
     std::string msg = String::compose( "SIONBuffer: buffer overflow: ptr=%1, n=%2, max_size=%3.", ptr_, n, max_size_ );
-    LOG( M_ERROR, "RecordingBackendSIONlib::write()", msg );
+    LOG( nest::VerbosityLevel::ERROR, "RecordingBackendSIONlib::write()", msg );
```
akorgor: Same comment as above: `nest::` here necessary?

nestkernel/recordables_map.h
```diff
@@ -56,14 +51,15 @@ namespace nest
  *       creation is deferred to the plain constructor of the host
  *       Node class, which is called only once to create the
  *       model prototype instance.
+ *       PYNEST-NG-FUTURE: is this still the case?
```
akorgor, line 54: Should this be answered with this PR?

nestkernel/spatial.cpp
```diff
@@ -360,92 +354,223 @@ distance( const ArrayDatum conns )
...
-create_mask( const DictionaryDatum& mask_dict )
+MaskPTR
+create_mask( const Dictionary& mask_dict )
+{
+  mask_dict.init_access_flags();
+
+  // The dictionary should contain one key which is the name of the
+  // mask type, and optionally the key 'anchor'. To find the unknown
+  // mask type key, we must loop through all keys. The value for the
+  // anchor key will be stored in the anchor_token variable.
```
akorgor: I do not see the `anchor_token` variable - is this a remainder from some old code?

pynest/CMakeLists.txt
```diff
@@ -19,68 +19,58 @@
 
 if ( HAVE_PYTHON )
 
+  # We use python3 here directly, as some of the CI jobs don't seem to have PYTHON
+  # or Python_EXECUTABLE set properly.
+  execute_process(
+    COMMAND "python3" "${PROJECT_SOURCE_DIR}/pynest/generate_exception_header.py"
```
otcathatsya: Can we rely on the alias being python3? I think there's a FindPython CMake command, would that be an option?

nestkernel/sonata_connector.cpp
```diff
@@ -91,24 +94,26 @@ SonataConnector::connect()
...
 
-  auto edges_container = getValue< ArrayDatum >( graph_specs_->lookup( "edges" ) );
+  auto edges_container = boost::any_cast< std::vector< Dictionary > >( graph_specs_.at( "edges" ) );
 
   // synapse-specific parameters that should be skipped when we set default synapse parameters
   skip_syn_params_ = {
     names::weight, names::delay, names::min_delay, names::max_delay, names::num_connections, names::synapse_model
   };
 
+
```
akorgor: Unintentional blank line?

nestkernel/sonata_connector.cpp
```diff
@@ -91,24 +94,26 @@ SonataConnector::connect()
...
 
-    const auto edge_dict = getValue< DictionaryDatum >( edge_dict_datum );
-    cur_fname_ = getValue< std::string >( edge_dict->lookup( "edges_file" ) );
     const auto file = open_file_( cur_fname_ );
     const auto edges_top_level_grp = open_group_( file, "edges" );
 
     // Create map of edge type ids to NEST synapse_model ids
-    cur_edge_params_ = getValue< DictionaryDatum >( edge_dict->lookup( "syn_specs" ) );
+    // cur_edge_params_ = edge_dict[ "syn_specs" ];
+    cur_edge_params_ = boost::any_cast< Dictionary >( edge_dict.at( "syn_specs" ) );
```
akorgor: Maybe `// cur_edge_params_ = edge_dict[ "syn_specs" ];` can be removed unless it is a purposeful reminder.

nestkernel/sonata_connector.cpp
```diff
@@ -563,13 +545,14 @@ SonataConnector::read_subset_( const H5::DataSet& dataset,
 }
 
 void
-SonataConnector::create_edge_type_id_2_syn_spec_( DictionaryDatum edge_params )
+SonataConnector::create_edge_type_id_2_syn_spec_( Dictionary edge_params )
 {
-  for ( auto it = edge_params->begin(); it != edge_params->end(); ++it )
+  for ( auto& syn_kv_pair : edge_params )
```
akorgor: Maybe here `for ( auto& [k_syn, v_syn] : edge_params )` and the corresponding changes in the loop?

nestkernel/sonata_connector.cpp
```diff
@@ -580,53 +563,49 @@ SonataConnector::create_edge_type_id_2_syn_spec_( DictionaryDatum edge_params )
...
 void
-SonataConnector::set_synapse_params_( DictionaryDatum syn_dict, size_t synapse_model_id, int type_id )
+SonataConnector::set_synapse_params_( Dictionary syn_dict, size_t synapse_model_id, int type_id )
 {
-  DictionaryDatum syn_defaults = kernel().model_manager.get_connector_defaults( synapse_model_id );
+  Dictionary syn_defaults = kernel().model_manager.get_connector_defaults( synapse_model_id );
   ConnParameterMap synapse_params;
 
-  for ( Dictionary::const_iterator default_it = syn_defaults->begin(); default_it != syn_defaults->end(); ++default_it )
+  for ( auto& syn_kv_pair : syn_defaults )
```
akorgor: As above, maybe `for ( auto& [k_syn, v_syn] : edge_params )` and the corresponding changes in the loop?

nestkernel/sonata_connector.cpp
```diff
@@ -638,40 +617,36 @@ SonataConnector::get_synapse_params_( size_t snode_id,
 {
   for ( auto const& syn_param : edge_type_id_2_syn_spec_.at( edge_type_id ) )
   {
-    const Name param_name = syn_param.first;
+    const auto param_name = syn_param.first;
     const auto param = syn_param.second;
```
akorgor, line 620: Maybe this would also work: `const auto [param_name, param] = syn_param;`?

nestkernel/conn_builder.cpp
```diff
@@ -499,13 +489,11 @@ nest::BipartiteConnBuilder::loop_over_targets_() const
...
+nest::BipartiteConnBuilder::set_synapse_model_( const Dictionary& syn_params, size_t synapse_indx )
 {
-  if ( not syn_params->known( names::synapse_model ) )
-  {
-    throw BadProperty( "Synapse spec must contain synapse model." );
-  }
-  const std::string syn_name = ( *syn_params )[ names::synapse_model ];
+  const std::string syn_name = syn_params.known( names::synapse_model )
+    ? syn_params.get< std::string >( names::synapse_model )
+    : std::string( "static_synapse" );
```
suku248, line 496: Why is it okay to set this to "static_synapse" instead of throwing an exception in case the synapse model is not known?

pynest/examples/run_examples.sh
```diff
@@ -19,12 +19,10 @@
...
 
-if ! command -v nest; then
-    echo "ERROR: command 'nest' not found. Please make sure PATH is set correctly"
-    echo "       by sourcing the script nest_vars.sh from your NEST installation."
-    exit 1
-fi
+# NOTE: This script requires bash version 4 or later.
+#       On macOS, you need to brew install bash for this script to work.
 
+# TODO-PYNEST-NG: What do we do about nest_vars.sh?
```
otcathatsya, line 25: Delete nest_vars.sh?

nestkernel/conn_builder.cpp
```diff
@@ -577,52 +564,58 @@ nest::BipartiteConnBuilder::set_synapse_params( DictionaryDatum syn_defaults,
...
+  // We now know that we have SP-parameters and can perform SP-specific checks and operations
   if ( syn_specs.size() > 1 )
   {
     throw KernelException( "Structural plasticity can only be used with a single syn_spec." );
   }
 
-  const DictionaryDatum syn_spec = syn_specs[ 0 ];
-  if ( syn_spec->known( names::pre_synaptic_element ) xor syn_spec->known( names::post_synaptic_element ) )
+  // We know now that we only have a single syn spec, so we extract that.
+  // We must take a reference here, otherwise access registration will not work, because the
```
suku248: This comment is incomplete.

nestkernel/nest.h
```diff
@@ -123,16 +162,19 @@ void connect_arrays( long* sources,
   long* targets,
   double* weights,
   double* delays,
-  std::vector< std::string >& p_keys,
+  const std::vector< std::string >& p_keys,
   double* p_values,
   size_t n,
-  std::string syn_model );
+  const std::string& syn_model );
+
```
akorgor, line 169: Maybe unintentional blank line?

nestkernel/nest.cpp
```diff
@@ -63,89 +110,207 @@ reset_kernel()
...
 
+Dictionary
+get_nc_status( NodeCollectionPTR nc )
+{
+  Dictionary result;
+  size_t node_index = 0;
+  for ( NodeCollection::const_iterator it = nc->begin(); it < nc->end(); ++it, ++node_index )
+  {
+    const auto node_status = get_node_status( ( *it ).node_id );
+    for ( auto& kv_pair : node_status )
```
akorgor: Maybe `for ( auto& [k, v] : node_status )`?

akorgor (commented):
Although I tried to be thorough, this review is likely superficial because I’m not familiar with the background of these changes and this part of the code base.

pynest/nestkernel_api.pyx
```diff
@@ -0,0 +1,944 @@
...
+cdef object any_vector_to_list(vector[any] cvec):
+    cdef tmp = []
+    cdef vector[any].iterator it = cvec.begin()
+    while it != cvec.end():
+        tmp.append(any_to_pyobj(deref(it)))
+        inc(it)
+    return tmp
+
+
+cdef object dict_vector_to_list(vector[Dictionary] cvec):
```
otcathatsya: duplicate with vec_of_dict_to_list (can delete one of them)

pynest/nestkernel_api.pyx
```diff
@@ -0,0 +1,944 @@
...
+    cdef std_map[string, vector[Dictionary]] syn_param_map
+    for k, colloc_syns in synapse_params.items():
+        syn_param_map[pystr_to_string(k)] = pylist_to_dictvec(colloc_syns.syn_specs)
+
+    connect_tripartite(pre.thisptr, post.thisptr, third.thisptr,
+    	               pydict_to_Dictionary(conn_params), pydict_to_Dictionary(third_factor_conn_params),
+                       syn_param_map)
+
+
+def llapi_disconnect(NodeCollectionObject pre, NodeCollectionObject post, object conn_params, object synapse_params):
```
otcathatsya, line 446: shares implementation with llapi_connect minus the very last call, potentially refactor?

pynest/nestkernel_api.pyx
```diff
@@ -0,0 +1,944 @@
...
+    elif isinstance(params, (list, tuple)):
+        if len(params) != len(conns):
+            raise ValueError('params list length must be equal to number of connections')
+        set_connection_status(conn_deque, list_of_dict_to_vec(params))
+    else:
+        raise TypeError('params must be a dict or a list of dicts')
+
+
+def llapi_connect_arrays(sources, targets, weights, delays, synapse_model, syn_param_keys, syn_param_values):
+    """Calls connect_arrays function, bypassing SLI to expose pointers to the NumPy arrays"""
```
otcathatsya: it says SLI :(

nestkernel/conn_builder.cpp
heplesser: Because `static_synapse` is our default synapse model which is used when the user does not explicitly specify one. Earlier, we put the default model into the params dict somewhere in the SLI interface.

nestkernel/conn_builder.cpp
heplesser: I rewrote the comment, hope the new one is clearer.

nestkernel/kernel_manager.cpp
heplesser: The method is never called, so I will remove it.

nestkernel/layer.h
heplesser: To provide as much const-ness as possible. Should be read from right to left: the parameter is a constant pointer to a constant node collection.

nestkernel/recordables_map.h
heplesser: Thanks for catching! Replaced now with standard data types in the comments there; a reminder to improve code there in a future step.

nestkernel/recordables_map.h
heplesser: Fixed.

nestkernel/recordables_map.h
heplesser: Fixed.

nestkernel/recordables_map.h
heplesser: Fixed.

nestkernel/sp_manager.h
heplesser: Fixed.

nestkernel/source_table.cpp
heplesser: Thanks!

nestkernel/recordables_map.h
heplesser: No, anything marked -FUTURE are things we should get back to once this PR is behind us.

nestkernel/spatial.cpp
heplesser: Fixed.

nestkernel/sonata_connector.cpp
heplesser: Removed.

nestkernel/sonata_connector.cpp
heplesser: Removed.

nestkernel/recording_backend_sionlib.cpp
heplesser: Removed.

nestkernel/recording_backend_sionlib.cpp
heplesser: Removed.

nestkernel/sonata_connector.cpp
heplesser: Thanks, if one first changes this, one should do it properly.  This actually reminded me of a point that needs more work, namely removing various `boost::any_cast<>()` from around the code and insulating them in the `Dictionary` class.

nestkernel/sonata_connector.cpp
heplesser: See above.

nestkernel/nest.h
heplesser: Removed

nestkernel/layer.h
akorgor: Aha, thank you for the explanation.

nestkernel/sonata_connector.cpp
akorgor: I see!

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+construct your networks slightly faster.
+
+If you have implemented your models using PyNEST, the only change you will need to make is to replace
+
+.code::
+  nest.set_verbosity("M_WARNING")
+
+with
+
+.code::
```
jessica-mitchell:
```suggestion
.. code-block::
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+
+What does this mean for you as a NEST user?
+...........................................
+
+Fortunately, almost nothing, except that you can now `pip install nest-simulator` and that scripts might
+construct your networks slightly faster.
+
+If you have implemented your models using PyNEST, the only change you will need to make is to replace
+
+.code::
```
jessica-mitchell:
```suggestion
.. code-block::
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+  nest.set_verbosity("M_WARNING")
+
+with
+
+.code::
+  nest.verbosity = nest.VerbosityLevel.WARNING
+
+We have also deprecated `nest.GetStatus()` and `nest.SetStatus()`, so over time you may want to replace
+
+.code::
```
jessica-mitchell:
```suggestion
.. code-block::
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+  nest.GetStatus(node_coll)
+
+with
+
+.code::
```
jessica-mitchell:
```suggestion
.. code-block::
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+Also, `Dictionary`, `Datum`, and `Token` are a matter of the past. Instead, we now
+have class `dictionary` based directly on `std::map` using `boost::any` to store entries of
+arbitrary type. Instead of our own `lockPTR`, we now use `std::unique_ptr` to manage objects with
+reference counting. Where changes are necessary in code for neuron or synapse models, they will
+likely be limited to slightly different notation in the `set()/get()` methods to support the new
+`dictionary` class.
+
+To test whether certain errors are raised when writing tests, instead of
+
+.code::
```
jessica-mitchell:
```suggestion
.. code-block::
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+  with pytest.raises(nest.kernel.NESTErrors.IllegalConnection):
+
+you now have
+
+.code::
```
jessica-mitchell:
```suggestion
.. code-block::
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+If you still have your models implemented in SLI, it is time now to migrate to PyNEST.
+
+This has been a huge change for us and we would be very interested in hearing about your experiences!
+
+
+What does this mean for you as a developer?
+...........................................
+
+The key change from a developer perspective are that the entire SLI interpreter code has been
+removed, noticeably reducing compile times. We therefore no longer have the `SLIModule`` concept.
```
jessica-mitchell:
```suggestion
removed, noticeably reducing compile times. We therefore no longer have the ``SLIModule`` concept.
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+Also, `Dictionary`, `Datum`, and `Token` are a matter of the past. Instead, we now
+have class `dictionary` based directly on `std::map` using `boost::any` to store entries of
```
jessica-mitchell:
```suggestion
have class ``Dictionary`` based directly on ``std::map`` using ``boost::any`` to store entries of
```

nestkernel/conn_builder.cpp
suku248: Thanks for clarifying this.

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+arbitrary type. Instead of our own `lockPTR`, we now use `std::unique_ptr` to manage objects with
```
jessica-mitchell:
```suggestion
arbitrary type. Instead of our own ``lockPTR``, we now use ``std::unique_ptr`` to manage objects with
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+reference counting. Where changes are necessary in code for neuron or synapse models, they will
+likely be limited to slightly different notation in the `set()/get()` methods to support the new
```
jessica-mitchell:
```suggestion
likely be limited to slightly different notation in the ``set()/get()`` methods to support the new
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+`dictionary` class.
```
jessica-mitchell:
```suggestion
``Dictionary`` class.
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+  nest.GetStatus(node_coll)
+
+with
+
+.code::
+  node_coll.get()
+
+and correspondingly for connection collections and use `nest.get()/set()` for kernel status parameters.
+
+If you have used direct interaction with SLI from the Python level through `sli_func()`, `sli_run()`
```
jessica-mitchell:
```suggestion
If you have used direct interaction with SLI from the Python level through ``sli_func()``, ``sli_run()``
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+or `sr()` you will need to now use the direct interface. Please get in touch via the NEST User
```
jessica-mitchell:
```suggestion
or ``sr()`` you will need to now use the direct interface. Please get in touch via the NEST User
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+
+.code::
+  nest.set_verbosity("M_WARNING")
+
+with
+
+.code::
+  nest.verbosity = nest.VerbosityLevel.WARNING
+
+We have also deprecated `nest.GetStatus()` and `nest.SetStatus()`, so over time you may want to replace
```
jessica-mitchell:
```suggestion
We have also deprecated ``nest.GetStatus()`` and ``nest.SetStatus()``, so over time you may want to replace
```

doc/htmldoc/whats_new/v3.10/index.rst
```diff
@@ -0,0 +1,100 @@
...
+
+.code::
+  with pytest.raises(nest.kernel.NESTErrors.IllegalConnection):
+
+you now have
+
+.code::
+  with pytest.raises(nest.NESTErrors.IllegalConnection):
+
+When
```
jessica-mitchell: Is there a forgotten statement here, or can this be removed?

pynest/nest/lib/hl_api_helper.py
```diff
@@ -362,7 +301,7 @@ def get_help_fname(obj):
         File name of the help text for obj
     """
 
-    docdir = sli_func("statusdict/prgdocdir ::")
+    docdir = nestkernel.ll_api_get_kernel_status()["docdir"]
```
babsey:
```suggestion
    docdir = nestkernel.llapi_get_kernel_status()["docdir"]
```

pynest/nest/lib/hl_api_info.py
```diff
@@ -80,7 +57,7 @@ def helpdesk():
 
     """
 
-    docdir = sli_func("statusdict/prgdocdir ::")
+    docdir = nestkernel.ll_api_get_kernel_status()["docdir"]
```
babsey:
```suggestion
    docdir = nestkernel.llapi_get_kernel_status()["docdir"]
```

models/amat2_psc_exp.h
```diff
@@ -261,12 +261,12 @@ class amat2_psc_exp : public ArchivingNode
 
     Parameters_(); //!< Sets default parameter values
 
-    void get( DictionaryDatum& ) const; //!< Store current values in dictionary
+    void get( Dictionary& ) const; //!< Store current values in dictionary
```
pnbabu, line 264: Should we use `Dictionary` instead of `dictionary` in the comments here? I saw this in other models, but not here.

models/astrocyte_lr_1994.h
```diff
@@ -331,8 +331,8 @@ class astrocyte_lr_1994 : public Node
 
     Parameters_(); //!< Sets default parameter values
 
-    void get( DictionaryDatum& ) const;             //!< Store current values in dictionary
-    void set( const DictionaryDatum&, Node* node ); //!< Set values from dicitonary
+    void get( Dictionary& ) const;             //!< Store current values in Dictionary
+    void set( const Dictionary&, Node* node ); //!< Set values from dicitonary
```
pnbabu, line 335:
```suggestion
    void set( const Dictionary&, Node* node ); //!< Set values from Dictionary
```

pynest/nest/lib/hl_api_types.py
```diff
@@ -574,9 +569,10 @@ def __bool__(self):
         """Converts the NodeCollection to a bool. False if it is empty, True otherwise."""
         return len(self) > 0
 
-    def __array__(self, dtype=None):
+    def __array__(self, dtype=None, copy=None):
```
babsey: The `copy` argument of `numpy.array` is a boolean.

```suggestion
    def __array__(self, dtype=None, copy=False):
```

models/bernoulli_synapse.h
```diff
@@ -195,21 +195,21 @@ constexpr ConnectionModelProperties bernoulli_synapse< targetidentifierT >::prop
...
 
 template < typename targetidentifierT >
 void
-bernoulli_synapse< targetidentifierT >::set_status( const DictionaryDatum& d, ConnectorModel& cm )
+bernoulli_synapse< targetidentifierT >::set_status( const Dictionary& d, ConnectorModel& cm )
 {
   ConnectionBase::set_status( d, cm );
-  updateValue< double >( d, names::weight, weight_ );
-  updateValue< double >( d, names::p_transmit, p_transmit_ );
+  d.update_value( names::weight, weight_ );
```
pnbabu: Why don't we use `update_value_param` here?

models/cm_compartmentcurrents.cpp
```diff
@@ -34,8 +34,7 @@ nest::Na::Na( double v_comp )
   // some default initialization
   init_statevars( v_comp );
 }
-
-nest::Na::Na( double v_comp, const DictionaryDatum& channel_params )
+nest::Na::Na( double v_comp, const Dictionary& channel_params )
```
pnbabu, line 37:
```suggestion

nest::Na::Na( double v_comp, const Dictionary& channel_params )
```

pynest/nest/lib/hl_api_types.py
```diff
@@ -942,21 +931,26 @@ def set(self, params=None, **kwargs):
             return
 
         if isinstance(params, (list, tuple)) and self.__len__() != len(params):
-            raise TypeError("status dict must be a dict, or a list of dicts of length {}".format(self.__len__()))
+            raise TypeError(f"status dict must be a dict, or a list of dicts of length {self.__len__()}")
```
babsey:
```suggestion
            raise TypeError(f"Status dict must be a dict, or a list of dicts of length {self.__len__()}")
```

pynest/nest/lib/hl_api_types.py
```diff
@@ -1080,30 +1066,28 @@ class Parameter:
     # The constructor should not be called by the user
     def __init__(self, datum):
         """Parameters must be created using the CreateParameter command."""
-        if not isinstance(datum, kernel.SLIDatum) or datum.dtype != "parametertype":
-            raise TypeError("expected parameter datum")
+        if not isinstance(datum, nestkernel.ParameterObject):
+            raise TypeError(
+                "expected low-level parameter object; use the 'CreateParameter()' function to create a 'Parameter'."
```
babsey:
```suggestion
                "Expected low-level parameter object; use the 'CreateParameter()' function to create a 'Parameter'."
```

doc/htmldoc/whats_new/v3.10/index.rst
heplesser: I will remove this with the next push.

pynest/nest/ll_api.py
```diff
@@ -50,236 +42,78 @@
...
-    """
-
-    if "mpi4py" not in sys.modules:
-        raise kernel.NESTError("set_communicator: mpi4py not loaded.")
-
-    engine.set_communicator(comm)
+    def __set__(self, instance, value):
+        if self._readonly:
+            msg = f"`{self._name}` is a read only kernel attribute."
+            raise AttributeError(msg)
```
babsey:
```suggestion
            raise AttributeError(f"`{self._name}` is a read only kernel attribute.")
```

pynest/nest/lib/hl_api_info.py
```diff
@@ -80,7 +57,7 @@ def helpdesk():
...
     help_fname = os.path.join(docdir, "html", "index.html")
 
     if not os.path.isfile(help_fname):
```
babsey, line 64: For better consistence, please include msg content into `FileNotFoundError`.

nestkernel/conn_parameter.h
```diff
@@ -27,14 +27,12 @@
 #include <limits>
 #include <vector>
 
+#include <boost/any.hpp>
```
suku248, line 30: With this NEST depends on boost.

models/correlation_detector.h
```diff
@@ -326,15 +352,15 @@ correlation_detector::handles_test_event( SpikeEvent&, size_t receptor_type )
 }
 
 inline void
-correlation_detector::get_status( DictionaryDatum& d ) const
+nest::correlation_detector::get_status( Dictionary& d ) const
```
pnbabu: Why does this need the `nest` namespace here when it's already in the namespace?

models/correlation_detector.h
```diff
@@ -326,15 +352,15 @@ correlation_detector::handles_test_event( SpikeEvent&, size_t receptor_type )
...
 {
   device_.get_status( d );
   P_.get( d );
   S_.get( d );
 }
 
 inline void
-correlation_detector::set_status( const DictionaryDatum& d )
+nest::correlation_detector::set_status( const Dictionary& d )
```
pnbabu: Same as above.

babsey (changes requested):
LGTM! 

Please see several code suggestions inline.

models/correlomatrix_detector.h
```diff
@@ -315,15 +315,15 @@ correlomatrix_detector::handles_test_event( SpikeEvent&, size_t receptor_type )
 }
 
 inline void
-correlomatrix_detector::get_status( DictionaryDatum& d ) const
+nest::correlomatrix_detector::get_status( Dictionary& d ) const
```
pnbabu: Same comment here. Why is the namespace qualifier `nest` required?

models/correlospinmatrix_detector.h
```diff
@@ -311,15 +311,15 @@ correlospinmatrix_detector::handles_test_event( SpikeEvent&, size_t receptor_typ
 }
 
 inline void
-correlospinmatrix_detector::get_status( DictionaryDatum& d ) const
+nest::correlospinmatrix_detector::get_status( Dictionary& d ) const
```
pnbabu, line 314: Same as above

models/eprop_iaf.h
```diff
@@ -475,7 +475,7 @@ class eprop_iaf : public EpropArchivingNodeRecurrent< false >
     double learning_signal_;
 
     //! Number of remaining refractory steps.
-    int r_;
+    long r_;
```
pnbabu, line 478: Why change it to `long`? Same comment for all eprop models.

models/eprop_iaf.h
```diff
@@ -531,7 +531,7 @@ class eprop_iaf : public EpropArchivingNodeRecurrent< false >
     double P_i_in_;
 
     //! Total refractory steps.
-    int RefractoryCounts_;
+    long RefractoryCounts_;
```
pnbabu: Why `long` here?

pynest/nest/lib/hl_api_info.py
babsey: You can find it at line 64. Sorry for this comment, it was out of reach.

models/eprop_synapse_bsshslm_2020.h
```diff
@@ -576,36 +576,38 @@ eprop_synapse_bsshslm_2020< targetidentifierT >::send( Event& e,
...
-eprop_synapse_bsshslm_2020< targetidentifierT >::set_status( const DictionaryDatum& d, ConnectorModel& cm )
+eprop_synapse_bsshslm_2020< targetidentifierT >::set_status( const Dictionary& d, ConnectorModel& cm )
 {
   ConnectionBase::set_status( d, cm );
-  if ( d->known( names::optimizer ) and optimizer_ )
+  if ( d.known( names::optimizer ) and optimizer_ )
   {
-    optimizer_->set_status( getValue< DictionaryDatum >( d->lookup( names::optimizer ) ) );
+    // We must pass here if called by SetDefaults. In that case, the user will get and error
+    // message because the parameters for the synapse-specific optimizer have not been accessed.
```
pnbabu:
```suggestion
    // We must pass here if called by SetDefaults. In that case, the user will get an error
    // message because the parameters for the synapse-specific optimizer have not been accessed.
```

pynest/nest/lib/hl_api_helper.py
```diff
@@ -517,17 +457,10 @@ def get_parameters(nc, param):
...
-        cmd = "/{} get".format(param)
-        sps(nc._datum)
-        try:
-            sr(cmd)
-            result = spp()
-        except kernel.NESTError:
-            result = nc.get()[param]  # If the NodeCollection is a composite.
-    # param is array of strings
+    if isinstance(param, str):
+        result = nestkernel.llapi_get_nc_status(nc._datum, param)
```
otcathatsya, line 461: In general would it be possible to just pass the wrapper object and do all the unrwapping (._datum, ptr) inside to hide it on the high level calls?

models/iaf_cond_alpha_mc.cpp
```diff
@@ -39,33 +39,10 @@
...
 
-// Includes from sli:
-#include "dict.h"
-#include "dictutils.h"
-
-/* ----------------------------------------------------------------
- * Compartment name list
- * ---------------------------------------------------------------- */
-
-/* Harold Gutch reported some static destruction problems on OSX 10.4.
```
pnbabu, line 50: Should we keep this comment?

models/multimeter.cpp
```diff
@@ -105,23 +105,24 @@ nest::multimeter::Buffers_::Buffers_()
...
+  d[ names::interval ] = interval_.get_ms();
+  d[ names::offset ] = offset_.get_ms();
+
+  // JME: Why can't this be done using std::copy?
+  std::vector< std::string > ad;
   for ( size_t j = 0; j < record_from_.size(); ++j )
   {
-    ad.push_back( LiteralDatum( record_from_[ j ] ) );
+    ad.push_back( record_from_[ j ] );
   }
```
pnbabu: Why do we copy this to a new vector?

models/music_cont_out_proxy.cpp
```diff
@@ -102,42 +94,42 @@ nest::music_cont_out_proxy::Buffers_::Buffers_( const Buffers_& b )
...
+  d[ names::port_name ] = port_name_;
+  d[ names::interval ] = interval_.get_ms();
+
+  ///  std::vector< std::string > record_from;
+  ///
+  ///  for ( size_t j = 0; j < record_from_.size(); ++j )
+  ///  {
+  ///    record_from.push_back( record_from_[ j ] );
+  ///  }
+
```
pnbabu: Remove comment.

pynest/nest/lib/hl_api_connection_helpers.py
```diff
@@ -255,21 +245,17 @@ def _connect_layers_needed(conn_spec, syn_spec):
...
                 d[k] = v._datum
         return d
 
-    projections = fixdict(projections)
-    sps(projections)
-    sr("ConnectLayers")
+    nestkernel.llapi_connect_layers(pre, post, fixdict(projections))
 
 
 def _process_input_nodes(pre, post, conn_spec):
```
otcathatsya, line 253: Not declared in \_\_all__ but used in `hl_api_connections` with explicit import. My Pytorch didn't like that but I guess it is just a conceptual issue? Do we actually use the wildcard import anywhere to justify declaring \_\_all__ headers?

pynest/nest/lib/hl_api_connections.py
```diff
@@ -279,20 +275,14 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, return_synapsecollection=F
...
-        _connect_spatial(pre, post, spatial_projections)
+        _connect_spatial(pre._datum, post._datum, spatial_projections)
     else:
-        sps(processed_conn_spec)
-        if processed_syn_spec is not None:
-            sps(processed_syn_spec)
-        sr("Connect")
+        nestkernel.llapi_connect(pre._datum, post._datum, processed_conn_spec, processed_syn_spec)
 
     if return_synapsecollection:
```
otcathatsya, line 282: should we add a default return path `return None`?

pynest/nest/lib/hl_api_connections.py
```diff
@@ -279,20 +275,14 @@ def Connect(pre, post, conn_spec=None, syn_spec=None, return_synapsecollection=F
...
         return GetConnections(pre, post)
 
 
-@check_stack
 def TripartiteConnect(pre, post, third, conn_spec, third_factor_conn_spec, syn_specs=None):
```
otcathatsya, line 288: docstring says syn_spec but param name is syn_specs

models/weight_recorder.cpp
```diff
@@ -59,81 +52,42 @@ nest::weight_recorder::weight_recorder( const weight_recorder& n )
...
+  d[ names::senders ] = senders_;
+  d[ names::targets ] = targets_;
 }
 
 void
-nest::weight_recorder::Parameters_::set( const DictionaryDatum& d )
+nest::weight_recorder::Parameters_::set( const Dictionary& d )
 {
-  if ( d->known( names::senders ) )
+  auto get_or_create_nc = [ &d ]( NodeCollectionPTR& nc, const std::string& key )
```
pnbabu: Shouldn't this be called just `get_nc` instead of `get_or_create_nc`, as an empty NC sis already created in the constructor?

nestkernel/grid_layer.h
```diff
@@ -294,9 +297,15 @@ template < int D >
 void
 GridLayer< D >::insert_global_positions_ntree_( Ntree< D, size_t >& tree, NodeCollectionPTR node_collection )
 {
-  insert_global_positions_( std::inserter( tree, tree.end() ), node_collection );
+  //  constexpr std::insert_iterator< Ntree< D, size_t > > ins1( std::insert_iterator( tree, tree.end() ) );
```
suku248: Can this commented out code be removed?

libnestutil/CMakeLists.txt
```diff
@@ -17,38 +17,45 @@
...
+)
 
-add_library( nestutil STATIC ${nestutil_sources} )
+add_library(nestutil STATIC ${nestutil_sources})
 
-set_target_properties( nestutil
-    PROPERTIES
-    POSITION_INDEPENDENT_CODE ON
-    )
+if(APPLE)
```
JanVogelsang, line 41: Are we sure that these properties are "Apple"-specific or does it apply to clang in general?

libnestutil/dict_util.h
```diff
@@ -25,43 +25,36 @@
...
+#include "dictionary.h"
 
 namespace nest
 {
-
-
-/** Update a variable from a dictionary entry if it exists, skip call if it
- * doesn't. If the dictionary entry is a parameter, return value generated from
- * the parameter parameter.
+/**
```
JanVogelsang, line 34: The changes in this file are fine, but what was the reason to put this function into a separate file in the first place? Why is this not in `dictionary.h`?

nestkernel/spatial.cpp
```diff
@@ -360,92 +354,222 @@ distance( const ArrayDatum conns )
   return result;
 }
 
-MaskDatum
-create_mask( const DictionaryDatum& mask_dict )
+MaskPTR
```
JanVogelsang, line 359: I don't quite understand where these changes to the `create_mask` function come from or how they are related to PyNEST-NG.

nestkernel/grid_layer.h
JanVogelsang: Looks like temporary/debug code which should be removed.

nestkernel/conn_builder.cpp
heplesser: @suku248 Could you check if the comment is now fine and resolve the conversation if it is?

nestkernel/spatial.cpp
heplesser: @akorgor If you are happy with the revised comment, please resolve the conversation, otherwise, let me know that I need to work more on it.

nestkernel/nest.cpp
heplesser: Fixed.

nestkernel/sonata_connector.cpp
heplesser: I think here I would for now leave it as it is, just to limit the number of changes in this PR.

nestkernel/sonata_connector.cpp
heplesser: @akorgor I thoroughly updated this part of the code to avoid explicit `boost::any_cast<>()` calls in the code here (they should be encapsuled in the `Dictionary` class and its helpers). Could you check that you are content with the new code and if so resolve the conversation?

nestkernel/conn_parameter.h
heplesser: Yes, it does for the moment. Once we switch to C++17 variants, we avoid the dependency again.

@suku248 Do you want to keep this comment open or resolve it?

pynest/nest/lib/hl_api_helper.py
heplesser: @babsey Good catch, fixed! And it needed a bit more work, too.

pynest/nest/lib/hl_api_info.py
heplesser: See above.

pynest/nest/lib/hl_api_types.py
heplesser: `None` is a meaningful default value for `copy` according to the [NumPy documentation](https://numpy.org/doc/stable/reference/generated/numpy.array.html):
> if True (default), then the array data is copied. If None, a copy will only be made if __array__ returns a copy, if obj is a nested sequence, or if a copy is needed to satisfy any of the other requirements (dtype, order, etc.). ...

models/amat2_psc_exp.h
heplesser: I think in the comment we can consider "dictionary" as an ordinary English word, not as a class name, so I would not want to systematically fix this now.

models/correlation_detector.h
heplesser: It doesn't need it, removed.

models/correlation_detector.h
heplesser: Removed.

models/correlomatrix_detector.h
heplesser: Removed

models/correlospinmatrix_detector.h
heplesser: Removed

nestkernel/nest.cpp
akorgor: Ok!

nestkernel/spatial.cpp
akorgor: Ok!

nestkernel/sonata_connector.cpp
akorgor: Ok!

nestkernel/sonata_connector.cpp
akorgor: Looks good to me!

Just as info `boost::any_cast` is still called in `nestkernel/free_layer.h`,  `nestkernel/conn_parameter.cpp` , `nestkernel/nest.cpp` , `nestkernel/conn_builder.cpp`, unless these are the helpers you are referring to.

heplesser: Thanks for the info! The helpers I meant were from `dictionary.{h,cpp}` and `dict_utils.h`.

pynest/nest/lib/hl_api_helper.py
heplesser: @babsey If you are happy with my answers/changes, can you either hit the "Resolve" button here or, if you don't have that button, leave a little "OK" so I know I can push "Resolve"?

pynest/nest/lib/hl_api_types.py
heplesser: Fixed.

pynest/nest/lib/hl_api_types.py
heplesser: Accepted.

pynest/nest/ll_api.py
heplesser: Accepted.

models/astrocyte_lr_1994.h
heplesser: Accepted.

models/cm_compartmentcurrents.cpp
heplesser: Accepted.

models/eprop_iaf.h
heplesser: `r_` gets its value from `V_.RefractoryCounts_`, which in turn is set to `Time( Time::ms( P_.t_ref_ ) ).get_steps()`, and `get_steps()` returns a `long`. Using `int` thus risks (in theory) losing bits, although in practice it does not make a difference as values are small enough.

models/eprop_synapse_bsshslm_2020.h
heplesser: Accepted.

models/iaf_cond_alpha_mc.cpp
heplesser: Yes, I would leave it in so it is clar why we define with fixed size here.

models/weight_recorder.cpp
heplesser: Actually, `update_nc` seems even better. Fixed.

models/music_cont_out_proxy.cpp
heplesser: Removed.

models/multimeter.cpp
heplesser: We don't need to do this any more, inherited from SLI times. Since we assign to a `boost::any`, a copy will be created behind the scenes anyways (also for std::variant). I removed the explicit copy here.

libnestutil/CMakeLists.txt
heplesser: @JanVogelsang I am not sure. Could you test with Clang under Linux?

libnestutil/dict_util.h
heplesser: The function needs access to the VP-specific RNG and that requires interaction with the `vp_manager`, which in turn lead to circular inclusion problems. Placing this function in a separate file avoids that.

nestkernel/spatial.cpp
heplesser: Spatial masks can be created from the Python level using `nest.CreateMask()`, which eventuelly ends up here. The resulting mask the needs to be returned in a form that can be presented at the Python level. This allows the user to define new masks, e.g., as the intersection or union of elementary masks. Previously, the representation at the SLI and Python level was handled via a `MaskDatum`. With `Datum`s gone, we now use a `MaskPTR`.

nestkernel/grid_layer.h
heplesser: Removed.

JanVogelsang: Thanks!

libnestutil/dict_util.h
JanVogelsang: Alright, I see, I think there's no way around that at the moment (until the kernel manager refactoring)

nestkernel/spatial.cpp
JanVogelsang: But why all the additional logic here (~130 lines)? Was this logic just moved from `libsli` somewhere?

heplesser: The logic here was previously in `NestModule`, see here: https://github.com/nest/nest-simulator/blob/5d15d571de3693fa071117a8b5446361d9ac42ee/nestkernel/nestmodule.cpp#L176

heplesser:
> LGTM!

Thanks @babsey ! The reviewer panel indicates that you had some comments. Could you formally approve the PR?

akorgor (approved):

nestkernel/conn_builder.cpp
heplesser: @suku248 Ping!

libnestutil/CMakeLists.txt
```diff
@@ -17,38 +17,49 @@
...
+    set_target_properties(nestutil PROPERTIES LINK_FLAGS "-Wl,-undefined -Wl,dynamic_lookup")
+endif()
 
-target_link_libraries( nestutil ${GSL_LIBRARIES} ${SIONLIB_LIBS} )
+set_target_properties(nestutil PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
-target_include_directories( nestutil PRIVATE
+target_link_libraries(nestutil ${GSL_LIBRARIES} ${SIONLIB_LIBS} OpenMP::OpenMP_CXX)
```
terhorstd, line 47: Did we drop support for `-Dwith-openmp=OFF`, or do we need to?

libnestutil/CMakeLists.txt
```diff
@@ -17,38 +17,49 @@
...
     regula_falsi.h
     sort.h
-    string_utils.h
+    string_util.h
     vector_util.h
-    )
+)
 
-add_library( nestutil STATIC ${nestutil_sources} )
+add_library(nestutil STATIC ${nestutil_sources})
```
terhorstd, line 39: Maybe add a comment about the interaction of `STATIC` here and `POSITION_INDEPENDENT_CODE` (`-fPIC`) below. Isn't PIC only needed for dynamic libs?

(If this turns out to need correction rather than explanation, I'd go with the unchanged setup now and flag an issue to revisit this after the merge.)

libnestutil/dict_util.h
```diff
@@ -25,43 +25,36 @@
...
+template < typename T >
 bool
-updateValueParam( DictionaryDatum const& d, Name const n, VT& value, nest::Node* node )
+update_value_param( Dictionary const& d, const std::string& key, T& value, nest::Node* node )
 {
-  const Token& t = d->lookup( n );
-
-  ParameterDatum* pd = dynamic_cast< ParameterDatum* >( t.datum() );
-  if ( pd )
+  assert( node != nullptr ); // PYNEST-NG-FUTURE: Receive node as const Node&, but that needs many changes throughout
```
terhorstd: Create an issue instead of comments. …maybe with a bit more explanation.

libnestutil/dictionary.cpp
```diff
@@ -0,0 +1,589 @@
...
+    {
+      return false;
+    }
+  }
+  else if ( is_type< long >( first ) )
+  {
+    if ( not is_type< long >( second ) )
+    {
+      return false;
+    }
```
terhorstd: Wouldn't it be cleaner to turn the logic? There seem to be only very specific cases where this can be `true`, but for most possible arguments it is `false`.
```suggestion
  if ( is_type< int >( first ) and is_type< int >( second ) )  // is short-cut by compiler
  {
    const auto this_value = boost::any_cast< int >( first );
    const auto other_value = boost::any_cast< int >( second );
    retrun this_value == other_value;
  }
  else if ( is_type< long >( first ) and is_type< long >( second ) )
  {
    // or even shorter
    retrun boost::any_cast< long >( first ) == boost::any_cast< long >( second );
  }
  // ...
  return false;
```

terhorstd (changes requested):
I left a few comments for the `libnestutil/` part. The OpenMP comment has to be clarified.
The others may be worth a look, and may be opened as issues after the merge. Just comment that you are aware and if you think it makes sense to create issues.

libnestutil/CMakeLists.txt
heplesser: The `OpenMP` search process makes sure that a dummy `OpenMP::OpenMP_CXX` is available even if we do not use OpenMP, see https://github.com/heplesser/nest-simulator/blob/9c6a1524ec680f36a626a555d371f4de49f8723e/cmake/ProcessOptions.cmake#L380.

libnestutil/CMakeLists.txt
heplesser: I just changed whitespace here, the original code is from @jougs ;). The same pattern applies also in `nestkernel` and `models`. From the [pertaining CMake documentation](https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html) it appears to me that combining `STATIC` and `POSITION_INDEPENDENT_CODE` is entirely fine and can, e.g., improve security.

libnestutil/dictionary.cpp
heplesser: Good point! And then one should be able to create a local templated lambda function for that ... Will look at it soon.

libnestutil/dict_util.h
heplesser: Good point, will look at it soon.

libnestutil/dictionary.cpp
terhorstd: Lambda was also my first thought, but asking for the type of the variant only gives run-time usable typeinfo, as far as I see. Haven't found a way to get any compile time available info useful for templates yet. :/

libnestutil/dict_util.h
heplesser: I have created #3734 for follow-up and removed the comment.

libnestutil/dictionary.cpp
heplesser: I found a solution I consider very neat and hope you like it, too!

terhorstd: Thanks for working out the details! It's really much better now!
:+1:

pynest/CMakeLists.txt
heplesser: Good point. I have changed it and hope things will work now on the CI.

pynest/examples/run_examples.sh
heplesser: Good point, but I think beyond the scope of this PR. I created #3740 as follow up and therefore will mark this comment as resolved.

pynest/nestkernel_api.pyx
heplesser: OK, I opted for the `vec_to_dict_...` variant because we used that form for several other conversion functions.

pynest/nestkernel_api.pyx
heplesser: Done (with next push).

pynest/nestkernel_api.pyx
heplesser: Removed.

pynest/nest/lib/hl_api_connection_helpers.py
heplesser: Good point. Actually, these helpers should be (and are) imported only via relative imports explicitly, so I removed the `__all__` here.

pynest/nest/lib/hl_api_connections.py
heplesser: Done, explicit is better than implicit :).

pynest/nest/lib/hl_api_connections.py
heplesser: Fixed, now plural also in docstring.

pynest/nest/lib/hl_api_helper.py
heplesser: I would see this as a larger refactoring and have created #3741 for follow-up, therefore resolving this here.

suku248 (approved):

nestkernel/grid_layer.h
heplesser: Done.

pynest/nest/__init__.py
```diff
@@ -109,7 +110,6 @@ def __init__(self, name):
         type(self).visualization = _lazy_module_property("visualization")  # noqa: F821
         type(self).voltage_trace = _lazy_module_property("voltage_trace")  # noqa: F821
 
-        self.__version__ = ll_api.sli_func("statusdict /version get")  # noqa: F821
```
pnbabu, line 112: Since this variable has been removed, it should also be removed from the docstring on L26. What is the alternative way of fetching the version information of NEST on the Python level?

heplesser: `nest.__version__` will be available after I merge the next fixes into PyNEST-NG.

heplesser:
After a long process initiated by @jougs and @hakonsbm, we are today finally ready to merge the PyNEST-NG branch in the main NEST development branch. This removes in particular the SLI interpreter. All future development shall build on this change.

PR #3525, opened 2025-07-11
Port regression test ticket 772 to Pytests

jessica-mitchell:
Using cursor

Note the second test fails:  ('weight' has the wrong type. One-dimensional parameter arrays can only be used in conjunction with rule 'one_to_one' or 'fixed_total_number'.)
, but I was not sure if the original sli test should pass or if the constraints are different


testsuite/pytests/sli2py_regressions/test_ticket_772.py
```diff
@@ -0,0 +1,58 @@
...
+
+import nest
+import pytest
+
+
+def test_integer_weights_and_delays_do_not_crash():
+    """
+    Regression test for Ticket #772.
+
+    Ensure that integer values for weight and delay do not crash NEST in Connect.
```
jessica-mitchell, line 35:
```suggestion
    Test ported from SLI regression test
    Ensure that integer values for weight and delay do not crash NEST in Connect.
```

PR #3530, opened 2025-07-11
Port regression test ticket 800 to Pytests

jessica-mitchell:
Uses cursor

Note that this test fails, but I am unsure what the problem is, it seems to be structured nicely so maybe the issue is minor.


testsuite/pytests/sli2py_regressions/test_ticket_800.py
```diff
@@ -0,0 +1,181 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import nest
+import pytest
+
```
jessica-mitchell, line 24:
```suggestion

"""
ticket 800: Ensure that spatial parameters check their parameters

Test ported from SLI regression test

This ticket ensures that spatial parameters check their parameters for
validity, e.g., strictly positive sigma for a Gaussian.

Author: Hans Ekkehard Plesser, 2014-12-13
"""
```

testsuite/regressiontests/ticket-800.sli
```diff
@@ -1,373 +0,0 @@
...
-  /lspec Set
-  ResetKernel
-  (Test 6: ) =only
-  lspec keys ==
-  { lspec CreateLayer } fail_or_die
-  clear
-} forall
-
-%% seventh: test ConnectLayers with good kernels
-good_distance_parameters [ 0.5 1.0 ] join
```
shimoura, line 254: I don't see these parameters listed in the Python version. Was it intentional?

testsuite/regressiontests/ticket-800.sli
```diff
@@ -1,373 +0,0 @@
...
-  (Test 8: ) =only
-  kspec DictQ { keys == } { == } ifelse
-  { << /connection_type /pairwise_bernoulli_on_source /kernel kspec >>
-    conn_layers_test_2d
-  } fail_or_die
-  clear
-} forall
-
-%% ninth: test ConnectLayers with good weights
-good_parameters [ 0.5 1.0 ] join
```
shimoura, line 278: These parameters are also not listed in the Python version. Was it intentional?

testsuite/pytests/sli2py_regressions/test_ticket_800.py
```diff
@@ -0,0 +1,252 @@
...
+bad_positions = [
+    nest.spatial.grid(shape=[1]),
+    nest.spatial.grid(shape=[2, 2], extent=[2.0, 0.5], center=[0.2, 0.3], edge_wrap=3),  # bad datatype
+    nest.spatial.free(pos=[[0.1, 0.1], [0.2, 0.2, 0.2]], extent=[2.0, 0.5], edge_wrap=True),  # dimension mix-up
+    nest.spatial.free(pos=[[0.1, 0.1, 0.1], [0.2, 0.2, 0.2]], extent=[2.0, 0.5], edge_wrap=True),  # dimension mix-up
+]
+
+# --- Good and bad conn specs ---
+good_connspecs = [
```
shimoura, line 119: From the sli version, there are two sets of parameters I don't see here:
- `/connection_type /pairwise_bernoulli_on_source /synapse_model /static_synapse`
- `/connection_type /pairwise_bernoulli_on_source /delay 2.0`

testsuite/regressiontests/ticket-800.sli
```diff
@@ -1,373 +0,0 @@
...
-     /extent [2. .5] /edge_wrap true >>
-  << /positions [[0.1 0.1 0.1] [0.2 0.2 0.2]] /elements /iaf_psc_alpha
-     /extent [2. .5 3.] /edge_wrap true >>
-]
-def
-
-/bad_layers
-[
-  << /shape [ 1 ] /elements /iaf_psc_alpha >>    % too few elements in shape
-  << /shape [ 2 2 ] /elementsss /iaf_psc_alpha >>  % misspelling
```
shimoura, line 136: This misspelled key test case isn't present in the Python version. Not sure where it would fit, or if it was removed on purpose.

shimoura (commented):
All tests passed. 
Besides that, some parameters present in the Sli version are not in the Python version, but I'm not sure if this was intentional or not.

testsuite/regressiontests/ticket-800.sli
heplesser: I dropped this intentionally because there no longer is an `/elements` parameter: The type of nodes to create is now given as the first argument to `Create()`. The other parameters in these dicts are now passed as named arguments to `grid()` or `free()`, so Python does the name-checking for us.

testsuite/pytests/sli2py_regressions/test_ticket_800.py
heplesser: Those two cases are dropped because `/synapse_model` and `/delay` now are part of the `syn_spec`, not the `conn_spec` as was the case for layers earlier.

testsuite/regressiontests/ticket-800.sli
heplesser: I overlooked these, added in new PR towards @jessica-mitchell (https://github.com/jessica-mitchell/nest-simulator/pull/36)

Issue #3532, opened 2025-07-13
disconnect() does not work reliably without compressed spikes


heplesser:
This code creates four connections and deletes the second of them:
```python
import nest
nest.use_compressed_spikes = False
n = nest.Create("parrot_neuron", 2)
nest.Connect(n, n)
c = nest.GetConnections()
c[1].disconnect()
```
If run as is, it will raise
```
NESTErrors.The connection does not exist: The connection does not exist in SLI function Disconnect_a: Deletion of connection is not possible.
```
even though the connection obviously is there.

When run with `nest.use_compressed_spikes = True` (default), the test completes successfully.

This raises the overall question whether there are other tests that will fail if they were run without compressed spikes?

med-ayssar:
Hey @heplesser,

**TL;DR**

1. The cause of the problem is  an **undefined behavior** when calling `std::lower_bound` without sorting the array first, and this happens when `compressedSpike=False`
2.  Plz check my changes and give me your feedback.

### Observation

Given the above script with two nodes: `p1` and `p2` -> connection list is as `SourceTable:[(source, target)]` -> `[(1, 1), (2 ,1), (1, 2), (2, 2)]`, and `TableConnection:[lcid->target]`  - > `[0->1, 1->1, 2->2, 3->2]` 

- [Info]: when calling getConnection -> we call `SimulationManager::update_connection_infrastructure`, which will **sort** the `SourceTable` and apply the permutation to the `TableConnection`.

- Using CompressedSpike=True
    * `AfterSort:getConnections` -> `[(1,1), (1, 2), (2, 1), (2, 2)]`
    * `AfterSort:TableConnection -> [0->1, 1->2, 2->1, 3->2]`
- Using CompressedSpike=False
    * Same Original Input (`SourceTable` & `TableConnection`) without **SORT**!!

Now let's check the following Code, and we want to delete the `Connection (edge) with id=1`

```cpp
size_t
nest::ConnectionManager::find_connection( const size_t tid,
  const synindex syn_id,
  const size_t snode_id,
  const size_t tnode_id )
{
  // lcid will hold the position of the /first/ connection from node
  // snode_id to any local node, or be invalid
  size_t lcid = source_table_.find_first_source( tid, syn_id, snode_id );
  if ( lcid == invalid_index )
  {
    return invalid_index;
  }

  // lcid will hold the position of the /first/ connection from node
  // snode_id to node tnode_id, or be invalid
[...]

PR #3536, opened 2025-07-20
Fix issue: #3532: Disconnect given compressedSpikes

med-ayssar:
Fix for #3532

### DIFF
--- a/nestkernel/source_table.h
+++ b/nestkernel/source_table.h
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <map>
 #include <set>
+#include <utility>
 #include <vector>
 
 // Includes from nestkernel:
@@ -300,7 +301,10 @@ public:
    * Finds the first entry in sources_ at the given thread id and
    * synapse type that is equal to snode_id.
    */
-  size_t find_first_source( const size_t tid, const synindex syn_id, const size_t snode_id ) const;
+  size_t find_first_source( const size_t tid,
+    const synindex syn_id,
+    const size_t snode_id,
+    bool isCompressedEnabled = false ) const;
 
   /**
    * Marks entry in sources_ at given position as disabled.
@@ -471,27 +475,75 @@ SourceTable::no_targets_to_process( const size_t tid )
 }
 
 inline size_t
-SourceTable::find_first_source( const size_t tid, const synindex syn_id, const size_t snode_id ) const
+SourceTable::find_first_source( const size_t tid,
+  const synindex syn_id,
+  const size_t snode_id,
+  bool isCompressedEnabled /* default = false */ ) const
 {
-  // binary search in sorted sources
-  const BlockVector< Source >::const_iterator begin = sources_[ tid ][ syn_id ].begin();
-  const BlockVector< Source >::const_iterator end = sources_[ tid ][ syn_id ].end();
-  BlockVector< Source >::const_iterator it = std::lower_bound( begin, end, Source( snode_id, true ) );
+  using SourceIter = BlockVector< Source >::const_iterator;
 
-  // source found by binary search could be disabled, iterate through
-  // sources until a valid one is found
-  while ( it != end )
+  const SourceIter begin = sources_[ tid ][ syn_id ].begin();
+  const SourceIter end = sources_[ tid ][ syn_id ].end();
+
+  if ( isCompressedEnabled )
   {
-    if ( it->get_node_id() == snode_id and not it->is_disabled() )
-    {
-      const size_t lcid = it - begin;
-      return lcid;
-    }
-    ++it;
-  }
+    // binary search in sorted sources
+    SourceIter it = std::lower_bound( begin, end, Source( snode_id, true ) );
 
-  // no enabled entry with this snode ID found
-  return invalid_index;
+    // source found by binary search could be disabled, iterate through
+    // sources until a valid one is found
+    while ( it != end )
+    {
+      if ( it->get_node_id() == snode_id and not it->is_disabled() )
+      {
+        const size_t lcid = it - begin;
+        return lcid;
+      }
+      ++it;
+    }
+
+    // no enabled entry with this snode ID found
+    return invalid_index;
+  }
+  else
+  {
+
+    auto nth_equal =
+      []( SourceIter first, SourceIter last, const Source& value, size_t n ) -> std::pair< bool, SourceIter >
+    {
+      if ( n == 0 )
+      {
+        auto iter = std::find( first, last, value );
+        return { iter != last, iter };
+      }
+      auto iter = std::find( first, last, value );
+      while ( n > 0 && iter != last )
+      {
+        --n;
+        iter = std::find( std::next( iter ), last, value );
+      }
+      return { iter != last, iter };
+    };
+    size_t pos = 0;
+    auto res = nth_equal( begin, end, Source( snode_id, true ), pos );
+    if ( !res.first )
+    {
+      return invalid_index;
+    }
+
+    while ( res.first )
+    {
+      if ( res.second->get_node_id() == snode_id && not res.second->is_disabled() )
+      {
+        // found a valid source
+        size_t lcid = res.second - begin;
+        return lcid;
+      }
+      ++pos;
+      res = nth_equal( std::next( res.second ), end, Source( snode_id, true ), pos );
+    }
+    return invalid_index;
+  }
 }
 
 inline void
@@ -521,8 +573,8 @@ SourceTable::num_unique_sources( const size_t tid, const synindex syn_id ) const
   size_t n = 0;
   size_t last_source = 0;
   for ( BlockVector< Source >::const_iterator cit = sources_[ tid ][ syn_id ].begin();
-        cit != sources_[ tid ][ syn_id ].end();
-        ++cit )
+    cit != sources_[ tid ][ syn_id ].end();
+    ++cit )
   {
     if ( last_source != ( *cit ).get_node_id() )
     {
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -996,7 +996,7 @@ nest::ConnectionManager::find_connection( const size_t tid,
 {
   // lcid will hold the position of the /first/ connection from node
   // snode_id to any local node, or be invalid
-  size_t lcid = source_table_.find_first_source( tid, syn_id, snode_id );
+  size_t lcid = source_table_.find_first_source( tid, syn_id, snode_id, use_compressed_spikes() );
   if ( lcid == invalid_index )
   {
     return invalid_index;
@@ -1446,7 +1446,7 @@ nest::ConnectionManager::get_targets( const std::vector< size_t >& sources,
   {
     for ( size_t i = 0; i < sources.size(); ++i )
     {
-      const size_t start_lcid = source_table_.find_first_source( tid, syn_id, sources[ i ] );
+      const size_t start_lcid = source_table_.find_first_source( tid, syn_id, sources[ i ], use_compressed_spikes() );
       if ( start_lcid != invalid_index )
       {
         connections_[ tid ][ syn_id ]->get_target_node_ids( tid, start_lcid, post_synaptic_element, targets[ i ] );


nestkernel/source_table.h
```diff
@@ -471,27 +475,77 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+    while ( res.first )
+    {
+      if ( res.second->get_node_id() == snode_id && not res.second->is_disabled() )
+      {
+        // found a valid source
+        size_t lcid = res.second - begin;
+        return lcid;
+      }
+      ++pos;
+      res = nth_equal( std::next( res.second ), end, value, pos );
```
heplesser: I don't get your logic here with counting `pos` up and then doing `pos` rounds in `nth_equal()`.  Since you call `nth_equal()` with `first == std::next( res.second )`, aren't you already looking for the next occurence of `value`?

nestkernel/source_table.h
```diff
@@ -471,27 +475,77 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+      while ( n > 0 && iter != last )
+      {
+        --n;
+        iter = std::find( std::next( iter ), last, value );
+      }
+      return { iter != last, iter };
+    };
+    size_t pos = 0;
+    auto res = nth_equal( begin, end, value, pos );
+    if ( !res.first )
```
heplesser: According to NEST coding standards, please write out `not`, `and`, `or`. A bit surprised that our CI setup didn't detect his, @terhorstd.

heplesser (changes requested):
@med-ayssar Thanks for the PR! I am not entirely sure I follow the logic around `nth_equal()`, see comments below.

I also wonder if, for the uncompressed case, we should not take a different approach. In that case, I believe, we simply need to run through all connections of a given synapse type linearly and disconnect the first one for which source and target match, and which has not yet been disconnected. I guess that means parallel iteration through source and target tables. The split into `find_first_source()` and `find_first_target()` does not make sense in that case. 

We also need some tests, including tests where there are multiple connections between the same source-target pair and we delete one at a time.

nestkernel/source_table.h
med-ayssar: You are right, I was thinking to put the `nth_equal` function in some different `namespace` or `file`. The idea was just to make it too general, trying to find the`nth` element equal to `val`.

med-ayssar: I will try applying your suggestion and let you know ASAP!

med-ayssar: I am using now `std:find_if` and a custom predicate for `std::lower_bound`.

nestkernel/source_table.h
```diff
@@ -300,7 +301,10 @@ class SourceTable
    * Finds the first entry in sources_ at the given thread id and
    * synapse type that is equal to snode_id.
    */
-  size_t find_first_source( const size_t tid, const synindex syn_id, const size_t snode_id ) const;
+  size_t find_first_source( const size_t tid,
+    const synindex syn_id,
+    const size_t snode_id,
+    bool isCompressedEnabled = false ) const;
```
heplesser: Please use camel casing only for types. I'd also keep the name closer to the name in connection manager:
```suggestion
    bool using_compressed_spikes = false ) const;
```

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
 {
-  // binary search in sorted sources
-  const BlockVector< Source >::const_iterator begin = sources_[ tid ][ syn_id ].begin();
-  const BlockVector< Source >::const_iterator end = sources_[ tid ][ syn_id ].end();
-  BlockVector< Source >::const_iterator it = std::lower_bound( begin, end, Source( snode_id, true ) );
-
-  // source found by binary search could be disabled, iterate through
-  // sources until a valid one is found
-  while ( it != end )
+  using SourceIter = BlockVector< Source >::const_iterator;
```
heplesser: Do you need to introduce this alias or could you use `auto` below instead? But maybe we need the explicit type to ensure we get a const iterator? Then see suggestions below.

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+
+  const SourceIter begin = sources_[ tid ][ syn_id ].begin();
+  const SourceIter end = sources_[ tid ][ syn_id ].end();
+
+  Source value { snode_id, true };
```
heplesser:
```suggestion
  const auto sources_begin = sources_[ tid ][ syn_id ].cbegin();
  const auto sources_end = sources_[ tid ][ syn_id ].cend();

  const Source wanted_source { snode_id, /* is_disabled */ true };    // is_disabled entry is ignored by Source::operator<()
```

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+    //     return true;
+    //   }
+    //   return false;
+    // };
+    // binary search in sorted sources
+    SourceIter it = std::lower_bound( begin, end, value );
+
+    // source found by binary search could be disabled, iterate through
+    // sources until a valid one is found
+    if ( it != end )
```
heplesser: I think we don't need the extra if here. If `it == end`, then `find_if()` will do nothing and return `end`, so we get the same result with simpler code.

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
     {
-      const size_t lcid = it - begin;
+      auto sourceIter = std::find_if( it,
+        end,
+        [ &value ]( const Source& src ) { return src.get_node_id() == value.get_node_id() && not src.is_disabled(); } );
```
heplesser:
```suggestion
        [ &value ]( const Source& src ) { return src.get_node_id() == value.get_node_id() and not src.is_disabled(); } );
```

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+      }
+    }
+    // no enabled entry with this snode ID found
+    return invalid_index;
+  }
+  else
+  {
+    auto sourceIter = std::find_if( begin,
+      end,
+      [ &value ]( const Source& src ) { return src.get_node_id() == value.get_node_id() && not src.is_disabled(); } );
```
heplesser:
```suggestion
      [ &value ]( const Source& src ) { return src.get_node_id() == value.get_node_id() and not src.is_disabled(); } );
```

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+    // };
+    // binary search in sorted sources
+    SourceIter it = std::lower_bound( begin, end, value );
+
+    // source found by binary search could be disabled, iterate through
+    // sources until a valid one is found
+    if ( it != end )
     {
-      const size_t lcid = it - begin;
+      auto sourceIter = std::find_if( it,
```
heplesser: Again, no camel casing:
```suggestion
      const auto source_iter = std::find_if( it,
```

nestkernel/source_table.h
```diff
@@ -471,27 +475,59 @@ SourceTable::no_targets_to_process( const size_t tid )
...
+        end,
+        [ &value ]( const Source& src ) { return src.get_node_id() == value.get_node_id() && not src.is_disabled(); } );
+      if ( sourceIter != end )
```
heplesser: This code is duplicated below for the non-compressed case. Wouldn't it be neater to move this part out of the if-else?

med-ayssar:
I found another issue that occurs when compressed spikes is turned off. It should be fixed now, and once the CI is green, I’ll update the description with an explanation of the problem.

med-ayssar:
@heplesser please take a look again.

med-ayssar:
- With a `custom-predicate, ` it is then 100% UB (check the `std::lower_bound` description). 

- Disabling a node in the source table causes the table to become unsorted when using compressed-spikes -> then UB with `std::lower_bound(predicate`)

med-ayssar:
### To Summarize:

1. With `compressed-spikes` off,  `std::lower_bound` was being applied on `unsorted` vector -> **now it is fixed.**

2. With `compressed-spikes` off, and regardless of (1), disconnecting a connection was not implemented correctly, because the query on the `snode_id` and `tnode_id` delivers a wrong `lcid` -> **now is fixed.**
3. While testing, I discovered that disconnecting a connection, it changes the entry in the `sourceTable`, and when working with a sorted `sourceTable` (`compressed-spikes=true`), we might get an `undefined behaviour` -> **not fixed yet.**

med-ayssar:
[Off-topic]: In NEST, disconnecting means **marking** certain objects (i.e, `sourceTable` and `connectionTable` entries, in such case, calling the `disable` function. Calling `nest.GetConnections` should then remove those marked entries. Unfortunately, this is not the case. However; the result of`nest.GetConnections` is still correct, but the underlying infrastructure might be in an inconsistent state.

It happens here: check the while-loop, and you will see why nothing is removed in some cases, despite they are marked for removal (disabled).
````cpp
size_t
nest::SourceTable::remove_disabled_sources( const size_t tid, const synindex syn_id )
{
  if ( sources_[ tid ].size() <= syn_id )
  {
    return invalid_index; // no source table entry for this synapse model
  }

  BlockVector< Source >& mysources = sources_[ tid ][ syn_id ];
  const size_t max_size = mysources.size();
  if ( max_size == 0 )
  {
    return invalid_index; // no connections for this synapse model
  }

  // lcid needs to be signed, to allow lcid >= 0 check in while loop
  // to fail; afterwards we can be certain that it is non-negative and
  // we can static_cast it to index
  long lcid = max_size - 1;
  while ( lcid >= 0 and mysources[ lcid ].is_disabled() )
  {
    --lcid;
  }
  const size_t first_invalid_lcid = static_cast< size_t >( lcid + 1 ); // loop stopped on first valid entry or -1
  if ( first_invalid_lcid == max_size )
  {
    return invalid_index; // all lcids are valid, nothing to remove
  }

  mysources.erase( mysources.begin() + first_invalid_lcid, mysources.end() );
  return first_invalid_lcid;
}

````

**Comment by heplesser** [comment]
@med-ayssar In isolation, you are right that `remove_disabled_sources()` will not work correctly. It strictly depends on a source table that has been sorted after any connections have been disabled. But then it will work correctly.

**Comment by med-ayssar** [comment]
Fair enough. Otherwise the PR is ready for reviewing.

File: testsuite/pytests/test_connectivity_with_compressed_spikes.py [hunk 11]
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
```
heplesser: Since this is a regression test, please rename it to `test_regression_issue-3532.py`.

models/eprop_synapse.cpp
cjotade, line 1: I don't understand how this code changes are related to this PR, as this wouldn't be an e-prop fix?

models/eprop_synapse_bsshslm_2020.cpp
cjotade, line 1: Same question that in eprop_synapse.cpp, how are this e-prop changes related with the current PR? sorry if I misunderstood something.

testsuite/pytests/test_issue_3532.py
cjotade, line 1: All test passed for me. @heplesser recommended adding a disconnect_all_at_once() method, @med-ayssar I sent you a PR with my suggested changes. After that, everything seems good from my side.

testsuite/pytests/test_issue_3532.py
```diff
@@ -0,0 +1,143 @@
...
+        pre_conns = post_conns
+
+
+def test_disconnect_one_to_one(network):
+    n_nodes = len(network)
+    nest.Disconnect(network, network)
+
+    assert len(nest.GetConnections()) == n_nodes**2 - n_nodes
+
+def test_disconnect_all_at_once(network):
```
heplesser, line 88:
```suggestion

def test_disconnect_all_at_once(network):
```

testsuite/pytests/test_issue_3532.py
```diff
@@ -0,0 +1,143 @@
...
+    n_nodes = len(network)
+    nest.Disconnect(network, network, conn_spec="all_to_all")
+
+    assert nest.num_connections == 0
+
+def test_disconnect_by_edge(network):
```
heplesser, line 95:
```suggestion

def test_disconnect_by_edge(network):
```

heplesser (commented):
Adding empty lines to fulfill coding standards.

PR #3541, opened 2025-08-01
Add Ornstein–Uhlenbeck noise generator

FinnBurkhardt:
# Add Ornstein–Uhlenbeck Noise Generator

This pull request adds a new `ou_noise_generator` device and corresponding pytest-based tests. The device generates temporally correlated current following an Ornstein–Uhlenbeck (OU). The tests verify both parameter handling and statistical properties.

## Model Implementation

- **Files added**  
  - `models/ou_noise_generator.h`
  - `models/ou_noise_generator.cpp`
  - `pytests/sli2py_stimulating/test_ou_noise_generator.py`

- **Process equation** 
  The state variable \(x\) evolves as

$$
dx = \frac{1}{\tau}(\mu - x)\ dt + \sigma\ dW
$$

  where  
  - \($\mu$\) (`mean` parameter) is the long-term mean of the process,  
  - \($\sigma$\) (`std` parameter) its stationary standard deviation,  
  - \($\tau$\) (`tau` parameter) is the time constant,  
  - `dt` is the internal integration step.

- **Integration and output**  
  - Internally integrates the OU equation at user-specified `dt`.  
  - On each simulation step it emits the current \(x\) to connected targets.

## Tests (`test_ou_noise_generator.py`)

- **Parameter setting and defaults**  
  - Verifies that `set()` and `SetDefaults()` both correctly configure all parameters (`mean`, `std`, `tau`, `dt`).

- **Error conditions**  
  - Asserts that attempting to create the generator with a `dt` not divisible by `nest.resolution` raises `StepMultipleRequired`.

- **Mean and variance**  
  - Runs 100 trials of 1000 ms each, records the last current value, and checks that the sample mean and variance match the target $\mu$ and $\sigma^2$ within reasonable statistical bounds.

- **Autocorrelation**  
  - Collects a long trace, discards the initial $10\tau$ burn-in, and computes the empirical lag-1 autocorrelation.  
  - Confirms it matches $\exp(-\text{dt}/\tau)$ within a small tolerance.

- **Cross-correlation**  
  - Connects a single OU generator to two independent neurons and records their membrane potentials.  
  - Computes the cross-correlation of the two voltage traces and verifies it remains near zero.


models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+The generator integrates this process at a user-defined interval `dt` and delivers the resulting current to its targets.
+A larger time constant :math:`\tau` results in a more slowly varying noise signal.
+
+All targets of a noise generator receive different, independent noise currents, but the currents for all targets are
+updated at the same points in time. The interval `dt` between updates must be a multiple of the simulation time step.
+
+.. admonition:: Recording the generated current
+
+   You can use a :doc:`multimeter <multimeter>` to record the average current sent to all targets for each time step if
+simulating on a single thread; multiple MPI processes with one thread each also work. In this case, the recording
```
jessica-mitchell:
```suggestion
   simulating on a single thread; multiple MPI processes with one thread each also work. In this case, the recording
```

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+interval of the multimeter should be equal to the `dt` of the generator to avoid aliasing effects. In multi-threaded
```
jessica-mitchell:
```suggestion
   interval of the multimeter should be equal to the `dt` of the generator to avoid aliasing effects. In multi-threaded
```

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+mode, recording of noise currents is prohibited for technical reasons.
```
jessica-mitchell:
```suggestion
   mode, recording of noise currents is prohibited for technical reasons.
```

jessica-mitchell: The entire paragraph needs to be indented to be included in the admonition

jessica-mitchell (changes requested):
Thanks for the contribution!
I have a minor change for the docs, and a question if there is a good use case we can use an [example](https://nest-simulator.readthedocs.io/en/stable/examples/index.html) for this generator?

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+    The mean value :math:`\mu` to which the process reverts (pA).
+
+std
+    The stationary standard deviation :math:`\sigma_{stat}` of the process (pA).
+
+tau
+    The correlation time constant :math:`\tau` of the process (ms).
+
+dt
+    The interval :math:`\delta` between updates of the noise current (ms).
```
clinssen:
```suggestion
    The time interval between updates of the noise current (ms).
```

jakobkramp (approved):

models/ou_noise_generator.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if ( not StimulationDevice::is_active( Time::step( now ) ) )
+    {
+      B_.logger_.record_data( origin.get_steps() + offs );
+      continue;
+    }
+
+    // >= in case we woke from inactivity
+    if ( now >= B_.next_step_ )
+    {
+      // std::cout << "I am in update" << std::endl;
```
heplesser: Please delete the commented-out line.

models/ou_noise_generator.cpp
```diff
@@ -0,0 +1,374 @@
...
+nest::ou_noise_generator::State_::get( DictionaryDatum& d ) const
+{
+}
+
+void
+nest::ou_noise_generator::Parameters_::set( const DictionaryDatum& d, const ou_noise_generator& n, Node* node )
+{
+  updateValueParam< double >( d, names::mean, mean_, node );
+  updateValueParam< double >( d, names::std, std_, node );
+  updateValueParam< double >( d, names::tau, tau_, node );
```
heplesser: Add a check for positive tau:
```suggestion
  updateValueParam< double >( d, names::tau, tau_, node );
  if ( tau_ <= 0 )
  {
    throw BadProperty( "tau > 0 required." );
  }
```

heplesser:
@FinnBurkhardt Would you be able to address the open issues within a few days? I also added two minor suggestions. Please merge master into your branch for this PR to ensure that the testsuite works properly.

pynest/examples/ou_noise_single_neuron.py
```diff
@@ -0,0 +1,78 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+One neuron with Ornstein-Uhlenbeck noise
+----------------------------------------
+
+This script simulates a neuron with input from the ou_noise_generator and
```
jessica-mitchell:
```suggestion
This script simulates a neuron with input from the ``ou_noise_generator`` and
```

pynest/examples/ou_noise_single_neuron.py
```diff
@@ -0,0 +1,78 @@
...
+###############################################################################
+# Second, we create the neuron, noise generator and voltmeter.
+
+neuron = nest.Create("iaf_psc_alpha")
+ou_noise = nest.Create("ou_noise_generator")
+voltmeter = nest.Create("voltmeter")
+
+###############################################################################
+# Third, we configure the OU-noise generator. The parameters are:
+# mean and std in pA, tau (correlation time) and dt in ms.
```
jessica-mitchell:
```suggestion
# ``mean`` and ``std`` in `pA`, ``tau`` (correlation time) and ``dt`` in `ms`.
```

pynest/examples/ou_noise_single_neuron.py
```diff
@@ -0,0 +1,78 @@
...
+# The update interval dt must be a multiple of the simulation resolution.
```
jessica-mitchell:
```suggestion
# The update interval ``dt`` must be a multiple of the simulation resolution.
```

pynest/examples/ou_vs_white_noise.py
```diff
@@ -0,0 +1,121 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Comparing white and OU noise
+----------------------------
+
+This example compares two neurons: one driven by the noise_generator (white noise)
```
jessica-mitchell:
```suggestion
This example compares two neurons: one driven by the ``noise_generator`` (white noise)
```

pynest/examples/ou_vs_white_noise.py
```diff
@@ -0,0 +1,121 @@
...
+and one driven by the ou_noise_generator (temporally correlated noise).
```
jessica-mitchell:
```suggestion
and one driven by the ``ou_noise_generator`` (temporally correlated noise).
```

pynest/examples/ou_vs_white_noise.py
```diff
@@ -0,0 +1,121 @@
...
+n_ou = nest.Create("iaf_psc_alpha", params=lif_params)
+
+mm_white = nest.Create("multimeter", params={"record_from": ["V_m"], "interval": dt})
+mm_ou = nest.Create("multimeter", params={"record_from": ["V_m"], "interval": dt})
+sr_white = nest.Create("spike_recorder")
+sr_ou = nest.Create("spike_recorder")
+
+###############################################################################
+# Third, we create and configure the generators. We use the same mean and
+# stationary standard deviation for both inputs and set tau only for the
```
jessica-mitchell:
```suggestion
# stationary standard deviation for both inputs and set ``tau`` only for the
```

pynest/examples/ou_vs_white_noise.py
```diff
@@ -0,0 +1,121 @@
...
+nest.Connect(n_ou, sr_ou)
+nest.Connect(mm_ou, n_ou)
+
+###############################################################################
+# Now we simulate the network.
+
+nest.Simulate(simtime)
+
+###############################################################################
+# Finally, we compute the membrane-potential mean/std (after a 1 s burn-in),
```
jessica-mitchell:
```suggestion
# Finally, we compute the membrane-potential `mean/std` (after a 1 s burn-in),
```

jessica-mitchell (changes requested):
Thanks @FinnBurkhardt looks good, just a few minor formatting suggestions

jessica-mitchell (approved):
thanks!

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+  interval of the multimeter should be equal to the `dt` of the generator to avoid aliasing effects. In multi-threaded
+  mode, recording of noise currents is prohibited for technical reasons.
+
+
+.. include:: ../models/stimulation_device.rst
+
+mean
+    The mean value :math:`\mu` to which the process reverts (pA).
+
+std
```
clinssen, line 85: On line 66 of models/ou_noise_generator.cpp, this unit is given as pA/sqrt(s)

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+
+  typedef std::vector< double > AmpVec_;
+
+  /**
+   * Store independent parameters of the model.
+   */
+  struct Parameters_
+  {
+    double mean_; //!< mean current, in pA
+    double std_;  //!< standard deviation of current, in pA
```
clinssen: same here

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+    double tau_;  //!< Standard frequency in Hz
```
clinssen: unit here seems wrong

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+    Time dt_;     //!< time interval between updates
+
+    /**
+     * Number of targets.
+     * This is a hidden parameter; must be placed in parameters,
```
clinssen, line 177: Just curious why not put it into Variables?

clinssen (commented):
Thanks for the PR! I just have some small comments/questions.

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+
+void register_ou_noise_generator( const std::string& name );
+
+class ou_noise_generator : public StimulationDevice
+{
+
+public:
+  ou_noise_generator();
+  ou_noise_generator( const ou_noise_generator& );
+
```
heplesser, line 118:
```suggestion
```

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+  // ------------------------------------------------------------
+
+  struct Variables_
+  {
+    normal_distribution normal_dist_; //!< normal distribution
+
+    long dt_steps_;    //!< update interval in steps
+    double prop_;      //!< frequency [radian/s]
+    double noise_amp_; //!<
+    double tau_inv_;   //!<
```
heplesser: Comment texts are missing here. I find `tau_inv_` misleading (would expect value 1/tau from the name), while it is 1-exp(-h/tau). Can you find a better name? Further, if I understand the code right, it is only used in in the combination `P_.mean_ * V_.tau_inv_`, so why not store that value in a variable `mean_decayed_` (or whatever would be a good name)?

models/ou_noise_generator.h
heplesser: I am wondering the same thing, but it is the same in `noise_generator` and several other generators. This comment may be very old, from the time when we still had a (non-working) `ResetNetwork` command. But changing that should be done for all pertaining generators. I have created #3617 for follow-up and suggest we keep it as is here.

models/ou_noise_generator.cpp
```diff
@@ -0,0 +1,377 @@
...
+    LOG( M_INFO, "ou_noise_generator::pre_run_hook()", "The number of targets has changed, drawing new amplitudes." );
+    init_buffers_();
+  }
+
+  V_.dt_steps_ = P_.dt_.get_steps();
+
+  const double h = Time::get_resolution().get_ms();
+  const double t = kernel().simulation_manager.get_time().get_ms();
+
+  // scale Hz to ms
```
heplesser: This comment seems to make no sense here.

models/ou_noise_generator.cpp
```diff
@@ -0,0 +1,377 @@
...
+  const double noise_amp = P_.std_ * std::sqrt( -1 * std::expm1( -2 * h / P_.tau_ ) );
+  const double prop = std::exp( -1 * h / P_.tau_ );
+  const double tau_inv = -std::expm1( -h / P_.tau_ );
+
+  V_.noise_amp_ = noise_amp;
+  V_.prop_ = prop;
+  V_.tau_inv_ = tau_inv;
```
heplesser: I would simplify this a bit

```suggestion
  V_.noise_amp_ = P_.std_ * std::sqrt( -1 * std::expm1( -2 * h / P_.tau_ ) );
  V_.prop_ = std::exp( -1 * h / P_.tau_ );
  V_.tau_inv_ = -std::expm1( -h / P_.tau_ );
```

models/ou_noise_generator.cpp
```diff
@@ -0,0 +1,377 @@
...
+    const long now = start + offs;
+
+    if ( not StimulationDevice::is_active( Time::step( now ) ) )
+    {
+      B_.logger_.record_data( origin.get_steps() + offs );
+      continue;
+    }
+
+    // >= in case we woke from inactivity
+    if ( now >= B_.next_step_ )
```
heplesser, line 289: I wonder if the logic for handling inactivity is correct here. Consider a situation like the following:
```python
ou = nest.Create('ou_noise_generator', params={'start': 200, 'stop': 500'})
for k in range(10):
    ou.offset = k * 1000
    nest.Simulate(1000)
```
This runs a simulation in sections of 1000 ms and in each section the generator is on from 200 to 500 ms and off in between. 

With the code as it is now, we freeze the OU process while the generator is inactive. But would it not make more sense to assume that the process runs permanently and we only send its values out between start and stop? Then, `amp` should always be updated and only `I_amp` should be kept to 0 and no events sent out while the generator is inactive. Or does this not matter? Or could on be more efficient and correct in probability by updating according to the time that has passed since the generator was last active?

models/ou_noise_generator.h
```diff
@@ -0,0 +1,320 @@
...
+
+All targets of a noise generator receive different, independent noise currents, but the currents for all targets are
+updated at the same points in time. The interval `dt` between updates must be a multiple of the simulation time step.
+
+.. admonition:: Recording the generated current
+
+  You can use a :doc:`multimeter <multimeter>` to record the average current sent to all targets for each time step if
+  simulating on a single thread; multiple MPI processes with one thread each also work. In this case, the recording
+  interval of the multimeter should be equal to the `dt` of the generator to avoid aliasing effects. In multi-threaded
+  mode, recording of noise currents is prohibited for technical reasons.
```
heplesser, line 77: I did not see that prohibition in the code.

testsuite/pytests/stimulating/test_ou_noise_generator.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
```
heplesser, line 1: This file should be directly in `pytests`. The `sli2py_*` directories are only for tests ported from SLI.

heplesser (changes requested):
Thanks for the revisions. I noticed a few more details and added one principal questions concerning the return from periods of inactivity.

models/ou_noise_generator.cpp
FinnBurkhardt: Good point! I hadn’t considered that case before. I’ve now implemented a version where the OU process keeps evolving even while the generator is inactive.
I tested it by repeatedly turning the generator on and off and checking the correlation between the last value before the pause and the first after resuming.
Would you consider this a useful test to add to the unit tests?

I also believe it’s possible to compute the update in a single large step, based on the time elapsed since the last update, with some bookkeeping overhead. I'll try to implement that version to assess how much more efficient it is.

models/ou_noise_generator.h
FinnBurkhardt: I think the prohibition is enforced in handles_test_event() in line 255-258:
```cpp
if (kernel().vp_manager.get_num_threads() > 1) {
    throw KernelException("Recording from a ou_noise_generator is only possible in single-threaded mode.");
}
```
That part is copied from the white noise generator, so maybe i got it wrong.

models/ou_noise_generator.cpp
heplesser: Yes, "jumping" over that gap mathematically should help efficiency. And if you can include the correlation check across the gap as a test, that would be very nice. Then, to test the test, try to break it by going temporarily back to the old implementation in NEST and see if the test fails.

PR #3564, opened 2025-09-03
Fix data race in mechanism preventing excessive update times

heplesser:
This PR fixes a data race in our mechanism monitoring the wall-clock duration of update cycles, discovered by @alpinangu.  The mechanism ensures that simulations can be terminated when they "get stuck". The data race did not affect correctness of results. Now, all measurements are strictly taken by the master thread.

The PR also re-orders some lines for better logic consistency and removes superfluous conditions from the do-while loop:
- There is no need to test the `update_time_limit_exceeded` variable because we throw an exception inside the loop in that case.
- There is no point in checking `exceptions_raised.at(tid)`, since we only set that after an exception has happened in the loop.

### DIFF
--- a/testsuite/do_tests.sh
+++ b/testsuite/do_tests.sh
@@ -177,14 +177,14 @@ list_files() {
     #
     # instead use list_files like
     #
-    #   list_files <somedir> <some_pattern> | while read -r test_name; do ... done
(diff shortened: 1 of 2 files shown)


mlober (approved):
Thank you for fixing the bug and also improving the readability of this code section. It looks good to me.

PR #3569, opened 2025-09-04
Ensure testsuite runs all tests again and detects missing reports

### LINKED ISSUE #3565: Testsuite does not run SLI-based unit, regression, mpi or music tests any more
#3002 brought major modifications to `do_tests.sh`. Since then, it seems that no unit, regression, mpi or music tests are run any more, as is visible here:

- Linux: https://github.com/nest/nest-simulator/actions/runs/16630029794/job/47802492554?pr=3002#step:23:9094
- macOS: https://github.com/nest/nest-simulator/actions/runs/16630029794/job/47802492572?pr=3002#step:15:8550

Closer inspection of the logs shows that a single test is run for each time `list_files()` is invoked, see here: https://github.com/nest/nest-simulator/actions/runs/16630029794/job/47802492554?pr=3002#step:23:118. 

Strangely enough, the mpitests all run, even though they use exactly the same approach to iterating over test files. But they are still not reported in the table at the end.

A first analysis on my local computer (macOS 15.6.2) indicates that `list_files()` (https://github.com/nest/nest-simulator/blob/416efe4dfe7844656631ef70a5cda12e1d91e65e/testsuite/do_tests.sh#L172) works correctly, so that the while loop at https://github.com/nest/nest-simulator/blob/416efe4dfe7844656631ef70a5cda12e1d91e65e/testsuite/do_tests.sh#L356 and in corresponding places must do something wrong, i.e., not iterate over the files provided by `list_files()`. It seems that the loop only runs once, for the first file returned by `list_files()`.

The macOS action log also shows an error just before displaying the final table (https://github.com/nest/nest-simulator/actions/runs/16630029794/job/47802492572?pr=3002#step:15:8535).

heplesser:
This PR fixes #3565. It reverts to use of for loops instead of `while read ...`. This is done because `run_test` seems to interfere with `while`. The PR also adds a check for phases reporting 0 tests. One can see here (tested on a different branch) that this works: https://github.com/heplesser/nest-simulator/actions/runs/17476583295/job/49638650526#step:32:8920


heplesser:
The missing results reported [here](https://github.com/nest/nest-simulator/actions/runs/17477690548/job/49642023473?pr=3569#step:32:9180) is a false positive: When testing with MPI and OpenMP under Linux on the Github runners, we do not run the tests for issue 3108 due to excessive runtimes. They are disabled using PyTest's deselection mechanism here: https://github.com/nest/nest-simulator/blob/416efe4dfe7844656631ef70a5cda12e1d91e65e/testsuite/pytests/test_issue_3108.py#L39 .

Unfortunately, deselected tests are not counted and reported as skipped, they are not counted at all (deselection happens during test collection, skipping during execution as far as I understand). And so the XML file reports zero tests for the pytest/mpi/3 case, where we only have the 3108 tests. I am looking for a solution.

heplesser:
I have now implemented a detection of missing tests based on current test counts. Numbers will need to be adjusted as we move tests from SLI to Py, but it turned out to be too difficult to export counts of deselected tests to the junit-xml output.

testsuite/summarize_tests.py
```diff
@@ -35,6 +36,26 @@
...
+# fewer tests under pytest. To keep complexity of the testing logic in bounds, minima are used below.
+expected_num_tests = {
+    "01 basetests": 6,
+    "02 selftests": 8,
+    "03 unittests": 27,
+    "04 regressiontests": 45,  # does not include the two python-dependent tests
+    "05 mpitests": 77,
+    "06 musictests": 1,
+    "07 pynesttests": 3719,  # without thread-dependent cases
+    "07 pynesttests mpi 2": (230, 172),  # first case without thread-depenent cases
```
pnbabu:
```suggestion
    "07 pynesttests mpi 2": (230, 172),  # first case without thread-dependent cases
```

testsuite/do_tests.sh
```diff
@@ -547,8 +548,18 @@ if test "${PYTHON}"; then
...
+
+	    if test "${INFO_OS:-}" = "Darwin"; then
+		# ref https://stackoverflow.com/a/752893
+		# Note that on GNU systems an additional '-r' would be needed for
+		# xargs, which is not available here.
+		proc_nums=$(cd "${PYNEST_TEST_DIR}/mpi/"; find ./* -maxdepth 0 -type d -print0 | xargs -0 -n1 basename)
+	    else
+		proc_nums=$(cd "${PYNEST_TEST_DIR}/mpi/"; find ./* -maxdepth 0 -type d -printf "%f\n")
+	    fi
+
```
pnbabu: Could you fix the indentation here?

pnbabu (commented):
It looks good to me; I have some minor comments. I would be fine merging this once the music tests pass.

testsuite/do_tests.sh
heplesser: Some tabs had sneaked in ...

heplesser:
Since this is critical and has been discussed in the NEST Open VC on 8 Sep in addition to the one review, I merge this now without a second review. It also does not touch on NEST code directly.

PR #3575, opened 2025-09-12
Fix data race in NodeManager

### LINKED ISSUE #3574: Data race condition in NodeManager
Analysis with LLVM Clang/OpenMP/Archer shows a data race in `NodeManager::ensure_valig_thread_local_ids()` between reading `wfr_network_size_` here https://github.com/nest/nest-simulator/blob/83713a2854fa0553c7247dea3c472e6ce7485e3d/nestkernel/node_manager.cpp#L532 and writing it here https://github.com/nest/nest-simulator/blob/83713a2854fa0553c7247dea3c472e6ce7485e3d/nestkernel/node_manager.cpp#L577.

The solution is to no longer call this method, which is entirely related to nodes, from parallel context during connection creation, but to make sure that local ids are updated before entering parallel connection creation.

In a connection context, this method is until now only called from here https://github.com/nest/nest-simulator/blob/83713a2854fa0553c7247dea3c472e6ce7485e3d/nestkernel/target_identifier.h#L171 when creating `_hpc` synapses, so any testing must be based on such synapses. `hpc_benchmark.sli` lead to the discovery of this bug.

heplesser:
This PR fixes #3574 by moving the check for updated thread-local data structures from `TargetIdentifierIndex::set_target()`,  which is called very deeply in the creation of every single connection in parallel context, out to top-level connection code (just below the level of the functions in `NestModule`). This makes sense, because this update is only related to nodes, not to connections. `set_target()` only requires that each node knows its up-to-date local node id.

In the process I have also renamed some variables and methods, because the code affected here does more than just update WFR-related data structures.

I realize that the code in https://github.com/heplesser/nest-simulator/blob/18e75eecf33f291bf2602eac843772bfb0918ef0/nestkernel/node_manager.cpp#L526 is not ideal. We should be able to eliminate at least the last loop over the local nodes and we could probably let the loop over the threads run in parallel. But that needs to come at a later stage due to time constraints.

I also realize that we should try to integrate this thread-local-is-up-to-date with the `have_nodes_changed` checks, but also that will take extra time.

### DIFF
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -489,32 +489,6 @@ nest::ConnectionManager::connect( NodeCollectionPTR sources,
 
 
 void
-nest::ConnectionManager::connect( TokenArray sources, TokenArray targets, const DictionaryDatum& syn_spec )
-{
-  // Get synapse id
-  size_t syn_id = 0;
-  auto synmodel = syn_spec->lookup( names::model );
-  if ( not synmodel.empty() )
-  {
-    const std::string synmodel_name = getValue< std::string >( synmodel );
-    // The following throws UnknownSynapseType for invalid synmodel_name
-    syn_id = kernel().model_manager.get_synapse_model_id( synmodel_name );
-  }
-  // Connect all sources to all targets
-  for ( auto&& source : sources )
-  {
-    auto source_node = kernel().node_manager.get_node_or_proxy( source );
-    for ( auto&& target : targets )
-    {
-      auto target_node = kernel().node_manager.get_node_or_proxy( target );
-      auto target_thread = target_node->get_thread();
-      connect_( *source_node, *target_node, source, target_thread, syn_id, syn_spec );
-    }
-  }
-}
-
-
-void
 nest::ConnectionManager::update_delay_extrema_()
 {
   if ( kernel().simulation_manager.has_been_simulated() )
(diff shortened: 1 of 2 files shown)


nestkernel/nest.cpp
```diff
@@ -221,6 +221,9 @@ get_connections( const DictionaryDatum& dict )
 void
 disconnect( const ArrayDatum& conns )
 {
+  // probably not strictly necessarye here, but does nothing if all is up to date
```
suku248: typo "necessary"

heplesser: Fixed.

nestkernel/node_manager.h
```diff
@@ -401,6 +411,15 @@ NodeManager::set_have_nodes_changed( const bool changed )
   have_nodes_changed_ = changed;
 }
 
+inline bool
+NodeManager::thread_local_data_is_up_to_date() const
+{
+  // Our logic assumes that we never delete nodes from a network
+  assert( size() >= size_last_local_data_update_ );
```
JoseJVS, line 418: Why not directly assert size() == size_last... Since anyways the result will be used in an assertion when returned?

heplesser: We cannot assert equality here: New nodes may have been added since the last call, and then `size()` is larger than `size_last_local_data_update_`.

PR #3026, opened 2023-12-01
 Fix warnings regarding int-type variables

muffgaga:
Sorry for the close/reopen, github reacted weird when I pushed with full reference name `refs/heads/fix_warnings2`…

PR #3576, opened 2025-09-15
Fix some warnings about mismatching integer types (replaces #3026)

heplesser:
Since the code in #3026 does not compile (any more; see https://github.com/nest/nest-simulator/pull/3026#discussion_r2348724494), I provide a simpler solution here based on `auto` and, where necessary, `size_t`. Partly inspired by Herb Sutter recommending [almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/), which I discovered from [here](https://stackoverflow.com/a/19236169).


jessica-mitchell (approved):
lgtm

terhorstd (approved):
LGTM :+1:

PR #3581, opened 2025-09-18
Bump C++ standard to C++20

JanVogelsang:
This PR consists of two changes:
1. Bump the C++ standard from C++17 to C++20 which is required to use C++ concepts (see `requires` keyword) that this PR uses to ~~fix~~ improve the timer implementation.
2. ~~Fix~~ Improve the detailed and threaded timer implementation by now correctly turning off detailed or threaded timers when desired.

The PR is based on #3544, which should be merged first. After #3544 is fixed, the number of changes in this PR will reduce drastically.

Edit: #3582 already fixed the timer issue, but the changes to the stopwatches here might still be more desirable compared to the solution implemented in #3582.

Now that the timer issue is already fixed, we can take more time to improve other aspects, such as the CMake-to-C++ interface, which is currently mainly C-macro-based. This could be changed to a C++-based solution which uses compile-time constants (`constexpr`) instead of C-macros. Further, C++20 modules could be introduced to further cut compile times and clean up the include tree (in combination with #3544).

### DIFF
--- a/nestkernel/stopwatch.h
+++ b/nestkernel/stopwatch.h
@@ -115,7 +115,7 @@ enum class timeunit_t : size_t
 template < clockid_t clock_type >
 class StopwatchTimer
 {
-  template < StopwatchGranularity, StopwatchParallelism, typename >
+  template < StopwatchGranularity, StopwatchParallelism >
   friend class nest::Stopwatch;
 
 public:


terhorstd:
Discussion in today's meeting concluded:
* This is too entagled with #3544 
* this should be split into one "bump version" and different PRs for specific changes
* postponed till after the PyNEST-ng merge

PR #3582, opened 2025-09-19
Provide a non-invasive fix for the stopwatch issue

JanVogelsang:
This PR is a simplified version of #3581 which doesn't require C++20.
It simplifies the stopwatch implementation by combining all template specializations into a single one which uses `if constexpr` constructs and some clever `std::conditional_t` usage to use the correct timer containers and functionality based on compile flags.

### DIFF
--- a/nestkernel/stopwatch_impl.h
+++ b/nestkernel/stopwatch_impl.h
@@ -46,6 +46,10 @@ Stopwatch< detailed_timer, threaded_timer >::start()
     }
     else
     {
+// This code applies for MasterOnly timers started/stopped from serial context, but also for
+// Threaded timers called from parallel contexts if `-Dwith-threaded-timers=OFF` for backward
+// compatibility with pre-3.9 timers. Therefore, we need to restrict to the master thread here and
+// cannot assert a single-threaded context.
 #pragma omp master
       {
         walltime_timer_.start();


nestkernel/stopwatch_impl.h
```diff
@@ -20,93 +20,203 @@
...
+  {
+    if constexpr ( use_timer_array )
+    {
+      kernel().vp_manager.assert_thread_parallel();
+      walltime_timer_[ kernel().vp_manager.get_thread_id() ].start();
+      cputime_timer_[ kernel().vp_manager.get_thread_id() ].start();
+    }
+    else
+    {
+#pragma omp master
```
heplesser, line 53:
```suggestion
// This code applies for MasterOnly timers started/stopped from serial context, but also for
// Threaded timers called from parallel contexts if `-Dwith-threaded-timers=OFF` for backward
// compatibility with pre-3.9 timers. Therefore, we need to restrict to the master thread here and
// cannot assert a single-threaded context.
#pragma omp master
```

heplesser (approved):
@JanVogelsang Very nice work! I just added one suggestion for a comment.

heplesser:
BTW, I have checked all four detailed/threaded combinations and they work as intended. Very nice that now only the actually active timers show in the kernel dictionary.

JanVogelsang:
I see, so essentially we have two different non-threaded timer types:
1. True "master-only" timers which are called in a parallel environment, but only by the master thread.
2. Timers in a serial environment.
In theory we could distinguish these two types as those are different cases, but as the `#pragma omp master` works even in serial environments, it's fine to just use the same code for both cases.

mlober (approved):
Thanks for fixing this so quickly. All my tests with the various configuration options were successful.

PR #3586, opened 2025-09-26
Fix a number of problems with GetNodes

heplesser:
This PR is a backport of a corresponding fix in PyNEST-NG. It fixes code leading to undefined MPI behaviour (passing `nullptr` for `sendbuf` to `MPI_Allgatherv()` and corrects the selection logic if only neurons with certain parameters are requested, as well as fixing a doubling of device entries from threads. It furthermore extends the tests to cover more cases, especially parallel and including devices.

The code for `GetNodes()` could be further improved. I have done this in the PyNEST-NG branch, see https://github.com/heplesser/nest-simulator/blob/da99c325a540b1a769c59530b5e1436bae65f1ea/nestkernel/node_manager.cpp#L338, but will not backport this here any more.

### DIFF
--- a/testsuite/pytests/test_getnodes.py
+++ b/testsuite/pytests/test_getnodes.py
@@ -23,50 +23,58 @@
 Test GetNodes
 """
 
-import unittest
-
 import nest
+import pytest
 
 
-@nest.ll_api.check_stack
-class GetNodesTestCase(unittest.TestCase):
-    """Test GetNodes function"""
+@pytest.fixture(autouse=True, params=[1, 2])
+def create_neurons(request):
+    nest.ResetKernel()
+    nest.local_num_threads = request.param
 
-    def setUp(self):
-        nest.ResetKernel()
+    nest.Create("iaf_psc_alpha", 3)
+    nest.Create("iaf_psc_delta", 2, {"V_m": -77.0})
+    nest.Create("iaf_psc_alpha", 4, {"V_m": [-77.0, -66.0, -77.0, -66.0], "tau_m": [10.0, 11.0, 12.0, 13.0]})
+    nest.Create("iaf_psc_exp", 4)
+    nest.Create("spike_generator", 3)
 
-        a = nest.Create("iaf_psc_alpha", 3)  # noqa: F841
-        b = nest.Create("iaf_psc_delta", 2, {"V_m": -77.0})  # noqa: F841
-        c = nest.Create(
-            "iaf_psc_alpha", 4, {"V_m": [-77.0, -66.0, -77.0, -66.0], "tau_m": [10.0, 11.0, 12.0, 13.0]}  # noqa: F841
-        )
-        d = nest.Create("iaf_psc_exp", 4)  # noqa: F841
 
-    def test_GetNodes(self):
-        """test GetNodes"""
-        all_nodes_ref = nest.NodeCollection(list(range(1, nest.network_size + 1)))
-        all_nodes = nest.GetNodes()
+@pytest.mark.parametrize("local_only", [True, False])
+class TestGetNodes:
+    """Tests for GetNodes() function."""
 
-        self.assertEqual(all_nodes_ref, all_nodes)
+    def test_GetNodes(self, local_only):
+        """Test that nodes are correctly retrieved if on parameters are specified."""
 
-    def test_GetNodes_with_params(self):
-        """test GetNodes with params"""
-        nodes_Vm = nest.GetNodes({"V_m": -77.0})
-        nodes_Vm_ref = nest.NodeCollection([4, 5, 6, 8])
+        nodes_ref = nest.NodeCollection(list(range(1, nest.network_size + 1)))
+        if local_only:
+            nodes_ref = sum(n for n in nodes_ref if n.local)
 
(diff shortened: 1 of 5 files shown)


testsuite/pytests/test_getnodes.py
```diff
@@ -23,69 +23,74 @@
...
+    nest.Create("iaf_psc_alpha", 3)
+    nest.Create("iaf_psc_delta", 2, {"V_m": -77.0})
+    # Work around because Create() with arrays does not work with more than one rank
+    # nest.Create("iaf_psc_alpha", 4, {"V_m": [-77.0, -66.0, -77.0, -66.0], "tau_m": [10.0, 11.0, 12.0, 13.0]})
+    c = nest.Create("iaf_psc_alpha", 4, {"V_m": -77.0, "tau_m": 10.0})
+    c[1].V_m = -66.0
+    c[3].V_m = -66.0
+    c[1].tau_m = 11.0
+    c[2].tau_m = 12.0
+    c[3].tau_m = 13.0
```
med-ayssar, line 46: I am curious how did this work, a local node (proxy-node) doesn't allow calling `set_status`, when working with`mpi`.

testsuite/pytests/test_getnodes.py
```diff
@@ -23,69 +23,74 @@
...
+        assert nodes == nodes_ref
+
+    def test_GetNodes_no_match(self, local_only):
         """
         Ensure we get an empty result if nothing matches.
 
         This would lead to crashes in MPI-parallel code before #3460.
         """
 
         nodes = nest.GetNodes({"V_m": 100.0})
```
med-ayssar: Just to make sure that it works in all variants.
```suggestion
        nodes = nest.GetNodes({"V_m": 100.0}, local_only=local_only)
```

heplesser: Thanks for pointing out my oversight!

testsuite/pytests/test_getnodes.py
heplesser: Not entirely sure, but it does:
```
(brew_venv_venv) fix-getnodes> mpirun -np 2 nest -c "/iaf_psc_alpha 10 Create dup 3 Take << /V_m -99.0 >> SetStatus { GetStatus dup /V_m known { /V_m get }{ pop (none) } ifelse } Map =="
[-9.900000e+01 (none) -9.900000e+01 (none) -7.000000e+01 (none) -7.000000e+01 (none) -7.000000e+01 (none)]
[(none) -9.900000e+01 (none) -7.000000e+01 (none) -7.000000e+01 (none) -7.000000e+01 (none) -7.000000e+01]
```
Note that `3 Take` takes the first three neurons created.

med-ayssar: I would try this example in python, because python has a different implementation, so maybe `sli` drops the update for `non-local` nodes

heplesser: I would suspect the other way around, since the PyNEST interface does quite a bit of work on arguments.  The same works nicely with PyNEST as well:
```
(brew_venv_venv) fix-getnodes> PYNEST_QUIET=1 mpirun -np 2 python -c "import nest; n=nest.Create('iaf_psc_alpha', 10); n[:3].V_m = -99.; print(n.get('V_m'))"
(-99.0, None, -99.0, None, -70.0, None, -70.0, None, -70.0, None)
(None, -99.0, None, -70.0, None, -70.0, None, -70.0, None, -70.0)
```

And the tests pass as well when run on two processes (each process outputs pytest info):
```
(brew_venv_venv) fix-getnodes> mpirun -np 2 pytest ../../src/fix-getnodes/testsuite/pytests/test_getnodes.py 
=============================================================== test session starts ================================================================
platform darwin -- Python 3.13.7, pytest-8.4.2, pluggy-1.6.0
=============================================================== test session starts ================================================================
platform darwin -- Python 3.13.7, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/plesser/NEST/code/src/fix-getnodes
configfile: pyproject.toml
plugins: mypy-1.0.1, pylint-0.21.0, xdist-3.8.0, timeout-2.4.0, anyio-4.10.0, cov-7.0.0
rootdir: /Users/plesser/NEST/code/src/fix-getnodes
configfile: pyproject.toml
plugins: mypy-1.0.1, pylint-0.21.0, xdist-3.8.0, timeout-2.4.0, anyio-4.10.0, cov-7.0.0
collected 24 items                                                                                                                                 
collected 24 items                                                                                                                                 

../../src/fix-getnodes/testsuite/pytests/test_getnodes.py 
../../src/fix-getnodes/testsuite/pytests/test_getnodes.py ................................................                                                           [100%]                                                           [100%]


================================================================ 24 passed in 0.27s ================================================================

================================================================ 24 passed in 0.27s ================================================================
```

med-ayssar: Yeah, It works, I just tested it too, we simply get a `nullptr` node when querying on a proxy-node, and we do nothing, on the other hand, PyNEST does not seem to check that when working with list of nodes (a.k.a `nodeCollection` with more than one item [issue: #3588]

med-ayssar (approved):
Looks good to me!

PR #3591, opened 2025-09-29
Create() does not work MPI-parallel with arrays of parameter values

### LINKED ISSUE #3587: Create() does not work MPI-parallel with arrays of parameter values
The following should work

```bash
mpirun -np 2 python -c 'import nest; nest.Create("iaf_psc_alpha", 4, {"V_m": [-77.0, -66.0, -77.0, -66.0], "tau_m": [10.0, 11.0, 12.0, 13.0]})'
```

but causes strange type errors both with master and pynest-ng. It works when run with just a single process. I suspect the problem lies in the Python interface that somehow manipulates the dictionary content.

med-ayssar:
Fixes #3587

### DIFF
--- a/pynest/nest/lib/hl_api_info.py
+++ b/pynest/nest/lib/hl_api_info.py
@@ -22,7 +22,7 @@
 """
 Functions to get information on NEST.
 """
-
+import itertools
 import os
 import textwrap
 import webbrowser
@@ -271,12 +271,15 @@ def SetStatus(nodes, params, val=None):
 
     params_is_dict = isinstance(params, dict)
     set_status_nodes = isinstance(nodes, nest.NodeCollection)
+    only_local_nodes = []
     if set_status_nodes:
         local_nodes = [nodes.local] if len(nodes) == 1 else nodes.local
-        set_status_nodes = set_status_nodes and all(local_nodes)
+        only_local_nodes = list(itertools.compress(nodes, local_nodes))
+        if len(only_local_nodes) == 0:
+            return
 
-    if params_is_dict and set_status_nodes:
-        node_params = nodes[0].get()
+    if params_is_dict:
+        node_params = only_local_nodes[0].get()
         contains_list = [
             is_iterable(vals) and key in node_params and not is_iterable(node_params[key])
             for key, vals in params.items()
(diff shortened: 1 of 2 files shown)


pynest/nest/lib/hl_api_info.py
```diff
@@ -271,12 +271,15 @@ def SetStatus(nodes, params, val=None):
 
     params_is_dict = isinstance(params, dict)
     set_status_nodes = isinstance(nodes, nest.NodeCollection)
+    only_local_nodes = []
     if set_status_nodes:
         local_nodes = [nodes.local] if len(nodes) == 1 else nodes.local
-        set_status_nodes = set_status_nodes and all(local_nodes)
+        only_local_nodes = list(itertools.compress(nodes, local_nodes))
+        if len(only_local_nodes) == 0:
+            return
```
heplesser, line 279: I am afraid of corner cases here. If `SetStatus` should trigger MPI communication, but some rank never calls `SetStatus` in the kernel, we will hang. So I would not want to return here.

pynest/nest/lib/hl_api_info.py
```diff
@@ -271,12 +271,15 @@ def SetStatus(nodes, params, val=None):
...
 
-    if params_is_dict and set_status_nodes:
-        node_params = nodes[0].get()
+    if params_is_dict:
+        node_params = only_local_nodes[0].get() if set_status_nodes else nodes[0].get()
```
heplesser, line 282: Here, I think the `and set_status_nodes` needs to stay where it is. The function here mixes in a not very good way setting node and synapse status and everything under this if, I believe, applies only to nodes.

heplesser (changes requested):
@med-ayssar Thanks for the PR, but I see some issues, see below. I wonder if we shouldn't leave the fix for PyNEST-NG—no user has complained about this ever ;).

heplesser:
@med-ayssar This issue should be addressed on a deeper level with a thorough revision of our approach to parameter setting, preferably moving operations to the C++ level, so we do not need to pull much information inefficiently up to the Python level to facilitate inspection of expected types. As #3587 will lead to NEST terminating with an error, there is no risk of users performing simulations with incorrectly set parameters. Since there is a risk of hangs in corner cases, I close this PR.

PR #3592, opened 2025-09-29
nest-server: BadRequest: 400 Bad Request: name '<function name>' is not defined

### LINKED ISSUE #2666: nest-server: BadRequest: 400 Bad Request: name '<function name>' is not defined
**Describe the bug**
The nest-server has issues with nest simulation scripts that rely on (nested?) function definitions, like the `LambertWm1' function in the PyNest `brunel_alpha_nest.py` example.

**To Reproduce**
```
$ export NEST_SERVER_RESTRICTION_OFF=false  # doesn't matter
$ nest-server start -d
$ ipython
>>> from NESTServerClient import NESTServerClient
>>> nsc = NESTServerClient()
>>> response = nsc.from_file('brunel_alpha_nest.py')
...
nest.raster_plot.from_device(espikes, hist=True)
plt.show()

--------------------
BadRequest                                Traceback (most recent call last)
Cell In[5], line 4
      1 from NESTServerClient import NESTServerClient
      2 nsc = NESTServerClient()
----> 4 response = nsc.from_file('brunel_alpha_nest.py')

File ~/miniforge3/envs/hybridlfpy/lib/python3.10/site-packages/NESTServerClient/NESTServerClient.py:68, in NESTServerClient.from_file(self, filename, return_vars)
     66 print(script)
     67 print(20 * '-')
---> 68 return self.exec_script(script, return_vars)

File ~/miniforge3/envs/hybridlfpy/lib/python3.10/site-packages/NESTServerClient/NESTServerClient.py:57, in NESTServerClient.exec_script(self, source, return_vars)
     52 params = {
     53     'source': source,
     54     'return': return_vars,
     55 }
     56 response = requests.post(self.url + 'exec', json=params, headers=self.headers)
---> 57 return encode(response)

File ~/miniforge3/envs/hybridlfpy/lib/python3.10/site-packages/NESTServerClient/NESTServerClient.py:35, in encode(response)
     33     return response.json()
     34 elif response.status_code == 400:
---> 35     raise BadRequest(response.text)

BadRequest: 400 Bad Request: name 'LambertWm1' is not defined
```

**Expected behavior**
Should work I suppose

 - OS: MacOS 13.3.1 
 - Shell: zsh
 - Python-Version: 3.10.10
 - NEST-Version: 3.4
 - Installation: from source

med-ayssar:
Fixes #2666

 #####  `exec(script: string|bytes, globals: dict, locals: dict)`:
```python
# script.py
def foo():  # stored in locals
    pass

def bar(): # stored in locals
    foo()  # is it in `bar.__locals__`?, if yes then ok, found it, 
           # if not, seach in bar._globals__ (same as the provided global), if not found, then search in __builtin__  
           # --> result `foo` not found
```
here, we must set `locals` to `globals`, to be able to call the defined functions within other defined functions.

### DIFF
--- a/pynest/nest/server/hl_api_server.py
+++ b/pynest/nest/server/hl_api_server.py
@@ -198,14 +198,14 @@ def do_exec(args, kwargs):
         with Capturing() as stdout:
             globals_ = globals().copy()
             globals_.update(get_modules_from_env())
-            get_or_error(exec)(source_cleaned, globals_, locals_)
+            get_or_error(exec)(source_cleaned, globals_)
         if len(stdout) > 0:
             response["stdout"] = "\n".join(stdout)
     else:
         code = RestrictedPython.compile_restricted(source_cleaned, "<inline>", "exec")  # noqa
         globals_ = get_restricted_globals()
         globals_.update(get_modules_from_env())
-        get_or_error(exec)(code, globals_, locals_)
+        get_or_error(exec)(code, globals_)
         if "_print" in locals_:
             response["stdout"] = "".join(locals_["_print"].txt)
 
@@ -303,7 +303,7 @@ def route_api():
 @app.route("/api/<call>", methods=["GET", "POST"])
 def route_api_call(call):
     """Route to call function in NEST."""
-    print(f"\n{'='*40}\n", flush=True)
+    print(f"\n{'=' * 40}\n", flush=True)
     args, kwargs = get_arguments(request)
     log("route_api_call", f"call={call}, args={args}, kwargs={kwargs}")
     response = api_client(call, args, kwargs)
@@ -588,7 +588,14 @@ def combine(call_name, response):
         return None
 
     # return the master response if all responses are known to be the same
-    if call_name in ("exec", "Create", "GetDefaults", "GetKernelStatus", "SetKernelStatus", "SetStatus"):
+    if call_name in (
+        "exec",
+        "Create",
+        "GetDefaults",
+        "GetKernelStatus",
+        "SetKernelStatus",
+        "SetStatus",
+    ):
         return response[0]
 
     # return a single response if there is only one which is not None
--- a/pynest/nest/server/hl_api_server.py
+++ b/pynest/nest/server/hl_api_server.py
@@ -192,20 +192,20 @@ def do_exec(args, kwargs):
     source_code = kwargs.get("source", "")
     source_cleaned = clean_code(source_code)
 
-    locals_ = dict()
+    locals_ = globals()
     response = dict()
     if RESTRICTION_DISABLED:
         with Capturing() as stdout:
             globals_ = globals().copy()
             globals_.update(get_modules_from_env())
-            get_or_error(exec)(source_cleaned, globals_)
+            get_or_error(exec)(source_cleaned, globals_, locals_)
         if len(stdout) > 0:
             response["stdout"] = "\n".join(stdout)
     else:
         code = RestrictedPython.compile_restricted(source_cleaned, "<inline>", "exec")  # noqa
         globals_ = get_restricted_globals()
         globals_.update(get_modules_from_env())
-        get_or_error(exec)(code, globals_)
+        get_or_error(exec)(code, globals_, locals_)
         if "_print" in locals_:
             response["stdout"] = "".join(locals_["_print"].txt)
 
--- a/pynest/nest/server/hl_api_server.py
+++ b/pynest/nest/server/hl_api_server.py
@@ -192,28 +192,27 @@ def do_exec(args, kwargs):
     source_code = kwargs.get("source", "")
     source_cleaned = clean_code(source_code)
 
-    locals_ = globals()
     response = dict()
     if RESTRICTION_DISABLED:
         with Capturing() as stdout:
             globals_ = globals().copy()
             globals_.update(get_modules_from_env())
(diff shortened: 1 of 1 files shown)


Helveg:
globals/locals are a known pain-point with undefined behavior, rectified only in [PEP667](https://peps.python.org/pep-0667/) since Python 3.13.

Example:

```
source = """
def y():
 return 5

def z():
 return y()

print(z())
"""

exec(source)
> 5

>>> exec(x, {})
> 5

exec(source, {}, {})
  File "<stdin>", line 1, in <module>
  File "<string>", line 8, in <module>
  File "<string>", line 6, in z
NameError: name 'y' is not defined
```

Which is explained by the note in the docs on `exec` [here](https://docs.python.org/3.13/library/functions.html#exec):

> Note: When exec gets two separate objects as globals and locals, the code will be executed as if it were embedded in a class definition. This means functions and classes defined in the executed code will not be able to access variables assigned at the top level (as the “top level” variables are treated as class variables in a class definition).

The solution proposed by @med-ayssar is summarized as follows:

```python
globals_ = get_restricted_globals()  # Create a safe dictionary
locals_ = globals_
exec(source, globals_, locals_)
```

which is actually equivalent to `exec(source, globals_, globals_)` which avoids the noted 2-object condition, but which is actually also equivalent to simply `exec(source, globals_)`.

So why are we even passing in `locals_` at all? This is a historically relatively widespread trick to "fish out" variables from an `exec` statement without polluting the current namespace, and in our source code looks like this:

```python
globals_ = ...
locals_ = {}
exec(source, globals_, locals_)
response["stdout"] = "".join(locals_["_print"].txt)
```

The purpose of this seems to be to capture stdout while inside of the `RestrictedPython` environment. The pattern if I had to guess would be phylogenetically related to the relative widespread (ab)use of:

```python
exec("x = 5") # which is equivalent to exec(code, globals(), locals())
print(locals()["x"]) # prints 5, so exec code occurs within current namespace
```

which breaks in Python 3.13 after the change of the semantics (see https://github.com/python/cpython/issues/118888).

Given all this information, I think the safest and clearest way to achieve the desired behavior here is to simply not create any `locals_` dict, and fish it out from the `globals` we pass in:

```python
def do_exec(args, kwargs):
    source_code = kwargs.get("source", "")
    source_cleaned = clean_code(source_code)

    response = dict()
    if RESTRICTION_DISABLED:
        with Capturing() as stdout:
            globals_ = globals().copy()
            globals_.update(get_modules_from_env())
            get_or_error(exec)(source_cleaned, globals_)
        if len(stdout) > 0:
            response["stdout"] = "\n".join(stdout)
    else:
        code = RestrictedPython.compile_restricted(source_cleaned, "<inline>", "exec")  # noqa
        globals_ = get_restricted_globals()
        globals_.update(get_modules_from_env())
        get_or_error(exec)(code, globals_)
        if "_print" in globals_:
            response["stdout"] = "".join(globals_ ["_print"].txt)
```

The only possible change that might have (which is also a caveat of @med-ayssar current proposal) is when we would define `_print` in the globals for some reason?

Helveg (changes requested):
See my [discourse](https://github.com/nest/nest-simulator/pull/3592#issuecomment-3351023622) in the conversation. The last code block is my suggested change, but I can't propose it here because it involves changing things outside of the code range I can propose suggestions to in the Review tab.

pynest/nest/server/hl_api_server.py
```diff
@@ -198,13 +198,15 @@ def do_exec(args, kwargs):
         with Capturing() as stdout:
             globals_ = globals().copy()
             globals_.update(get_modules_from_env())
+            locals_ = globals_
             get_or_error(exec)(source_cleaned, globals_, locals_)
```
babsey:
```suggestion
            get_or_error(exec)(source_cleaned, globals_)
```

pynest/nest/server/hl_api_server.py
```diff
@@ -198,13 +198,15 @@ def do_exec(args, kwargs):
...
         if len(stdout) > 0:
             response["stdout"] = "\n".join(stdout)
     else:
         code = RestrictedPython.compile_restricted(source_cleaned, "<inline>", "exec")  # noqa
         globals_ = get_restricted_globals()
         globals_.update(get_modules_from_env())
+        locals_ = globals_
         get_or_error(exec)(code, globals_, locals_)
```
babsey:
```suggestion
        get_or_error(exec)(code, globals_)
```

babsey (changes requested):
Remove `locals_` variable at line 195 and rename `locals_` to `globals_` at lines 215 - 218

med-ayssar:
Sorry, here you go!

Helveg (changes requested):
I think this still does not do what reviewers agreed on:

* `locals_` on line 195 should be removed
* The `get_or_error(exec)(code, globals_, locals_)` calls should be changed to `get_or_error(exec)(code, globals_)`

babsey (approved):
LGTM! 

Thank you for the fix. 
It works with function in nest-server exec call.

med-ayssar:
> main needs to be merged for the macos build to succeed. (i think, @heplesser sent an email on this)

Done

Helveg (approved):

Helveg:
@heplesser @terhorstd  this one has been pending for quite a bit and seems fully ready to merge

heplesser:
@Helveg Thanks for the reminder and sorry about the delay.

PR #3599, opened 2025-10-08
Add CI workflow to check authors name against release tools 

jessica-mitchell:
This PR adds a workflow in nestbuildmatrix/buld_support that, for every PR,  checks the author(s) who contributed on that PR. It  then compares these authors to those in our nest-release repo to see if they are included in our 'validated authors' list. This includes authors who have contributed but whose author names or email addresses have changed.

This CI job will flag any author who is not found so we can add them to the release workflow, ensure we have their CA and other info. 
The reasoning for this is to minimize the amount of work needed to be done at release time;
If we have all the author information at time of PR we can save time during the release process.

The CI job is currently set up so that even if access to the private nest-releases repo is not authorized; the names of the authors of the current pr will be logged.


jessica-mitchell:
I am going to close this PR and look at this issue in the future; there are still too many open discussion points.

PR #3616, opened 2025-11-03
Add note in docs about how to use volume transmitter during network setup

clinssen:
With thanks to Alexandra Chatzikalymniou for pointing this out!

### DIFF
--- a/models/stdp_dopamine_synapse.h
+++ b/models/stdp_dopamine_synapse.h
@@ -66,7 +66,7 @@ Parameters
 
 =================== =========================== ======================================================
 **Common properties**
------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


models/stdp_dopamine_synapse.h
```diff
@@ -64,40 +64,41 @@ dopaminergic dynamics is calculated in the synapse itself.
 Parameters
 ++++++++++
 
-=================== ============== ======================================================
+=================== =========================== ======================================================
 **Common properties**
 -----------------------------------------------------------------------------------------
```
jessica-mitchell:
```suggestion
------------------------------------------------------------------------------------------------------
```
Make sure that this line is the same length as the top and bottom or else the table won't render

PR #3622, opened 2025-11-10
Port test_delay_exchange from SLI to Py

heplesser:
This test does not need the MPIWrapper construct since it is run only for one number of MPI ranks and results can be checked on each rank.


suku248 (approved):
This test conversion looks good. Thanks @heplesser !

PR #3626, opened 2025-11-11
Port test_gamma_sup_generator from SLI to Py

heplesser:


heplesser:
Incorporated into #3633 by parametrization, thus closing.

PR #3630, opened 2025-11-12
Port mpitest/test_global_rng from SLI to Python

heplesser:
This PR also adds a new MPITestWrapper subclass.

### DIFF
--- a/testsuite/pytests/mpi/4/test_consistent_local_vps.py
+++ b/testsuite/pytests/mpi/4/test_consistent_local_vps.py
@@ -19,14 +19,11 @@
 # You should have received a copy of the GNU General Public License
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
-import unittest
-
 import nest
+import pytest
 
-HAVE_OPENMP = nest.ll_api.sli_func("is_threaded")
 
-
-@unittest.skipIf(not HAVE_OPENMP, "NEST was compiled without multi-threading")
+@pytest.mark.skipif_missing_threads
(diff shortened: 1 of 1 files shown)


testsuite/pytests/sli2py_mpi/test_global_rng.py
```diff
@@ -0,0 +1,50 @@
...
+    p = nest.CreateParameter("uniform", {"min": 0, "max": 1})
+
+    # Uncomment one of the two for loops to provoke failure
+    # for _ in range(nest.num_processes):
+    #     p.GetValue()
+    # for _ in range(nest.Rank()):
+    #     p.GetValue()
+
+    vals = pd.DataFrame([p.GetValue() for _ in range(5)])
+    vals.to_csv(OTHER_LABEL.format(nest.num_processes, nest.Rank()), sep="\t")  # noqa: F821
```
otcathatsya, line 50: Is communicating with the wrapper via file system the easiest way? Feels a bit unintuitive but probably fine?

otcathatsya (approved):
Should we have separate tests with a passing + failing case instead of the commented-out part?

testsuite/pytests/sli2py_mpi/test_global_rng.py
heplesser: It is the only robust way when we have data coming from multiple ranks sending data. The old SLI-based approach used pipes, but that limited severely the amount of data that can be sent and it still had some risk of failure because messages from different ranks got garbled.

PR #3633, opened 2025-11-13
Port most mpitests/test_*_generator to Python

heplesser:
This PR ports almost all mpitests/*generator* tests using parameterization. For the sinusoidal generator tests, the variants with multiple generators in the network are dropped—if one works, multiple will also work. `test_sinusoidal_poisson_generator_{5,6}` have a different goal and will be ported separately.


JanVogelsang (approved):
Nice!

PR #3635, opened 2025-11-18
Port mpi-based tests of spatial connectivity from SLI to Py

heplesser:
This ports mpi-based tests of spatial connectivity, including connecting generators. All cases now use a mask and space-dependent weights and delays. Note the significant change in syntax for spatial connectivity between SLI and Python. Connecting from generators can be handled in the same way as for neurons, while for recorders only pairwise Bernoulli iterating over the source population is supported. Therefore, the latter case is handled in a separate test.

The original SLI tests only sent limited data on some connections for comparison between the different cases. This was due to the risk of "data garbling" if too much data was sent. In the new MPI testing scheme, this problem no longer exists, so source, target, weight and delay information for all connections are written to file and compared. Therefore, we also now randomize weights and delays in all cases.


testsuite/pytests/sli2py_mpi/test_spatial_connections.py
```diff
@@ -0,0 +1,95 @@
...
+)
+@MPITestAssertEqual([1, 2, 4], debug=False)
+def test_spatial_connections(source_model, geometry, conn_spec):
+    """
+    Confirm that spatial connections are created consistently for fixed VP.
+
+    The test is performed on connection data written to OTHER_LABEL.
+    """
+
+    import nest
```
otcathatsya, line 67: can we move this to the top? Also, do we need pandas for csv write or could we use numpy's txt format / the more lightweight csv lib? Probably not a big difference though, tried it locally and the timing diff was negligible.

otcathatsya (approved):

testsuite/pytests/sli2py_mpi/test_spatial_connections.py
heplesser: Maybe personal preference, but I am quite fond of Pandas  ...

Issue #3634, opened 2025-11-18
tsodyks(2)_synapse strange behavior in paired-pulse simulations


HDictus:
**Describe the bug**
The paired-pulse behavior of both tsodyks_synapse and tsodyks2_synapse is highly unpredictable. While the same parameters, simulation settings, and interval lead to the same strength (it is  deterministic), the difference between intervals with a given parameterization is fully chaotic. When I do trains of multiple spikes at 10Hz the behavior of the synapse is reasonable, but paired-pulse simulations specifically seem to have a problem.

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/03a66739-0ef4-4121-b75e-c88a2257aea3" />

**To Reproduce**
recreate_bug.py : 
```
import pandas as pd
import matplotlib.pyplot as plt
from pathlib import Path
import numpy as np
import nest


plots_dir = Path('intervalplots')
plots_dir.mkdir(exist_ok=True)

times_run = [0]
def sim(interval, nmodel, modparams, receptor_type):
    nest.ResetKernel()
    times_run[0] += 1
    nest.rng_seed = times_run[0]
    src = nest.Create('spike_train_injector')
    src.spike_times = [1., 1.+interval]
    tgt = nest.Create(nmodel, modparams)
    synapse_model = {
        'synapse_model': 'tsodyks2_synapse',
        'u': 0.5,
        'U': 0.5,
        'receptor_type': receptor_type,
        'tau_fac': 0.0,
        'tau_rec': 1600
    }
    syn = nest.Connect(src, tgt, 'one_to_one', synapse_model, return_synapsecollection=True)
    mm = nest.Create('multimeter', params={'interval': 0.25, 'record_from': ['g_1', 'V_m']})
    nest.Connect(mm, tgt, 'one_to_one')
    nest.Simulate(interval + 50)
    df = pd.DataFrame(mm.get()['events'])
    plt.clf()
    plt.plot(df['times'], df['g_1'])
    plt.savefig(plots_dir / f'{nmodel}-{interval}.png')
    return df

stds = {}
peaks = {}
for interval in (5, 10, 11, 12, 13, 14, 15, 20, 30, 40, 50, 75, 100, 150, 200, 300, 400, 600, 800):
    pks = []
    for _ in range(10):
        df = sim(interval, 'glif_cond', {'tau_syn': [0.1, 0.1]}, 1)
        pks.append(df['g_1'][df['times'] > interval].max())

[...]

PR #3636, opened 2025-11-18
multimeter/voltmeter now raise an exception if one tries to set a recording interval that is not a multiple of the resolution

heplesser:
It was always intended that multimeter/voltmeter should only accept recording intervals that were multiples of the resolution. Due to an implementation error, this did not happen and multimeter silently rounded to the nearest step. This PR ensures expected behavior and makes this explicit in the documentation. See also #3634.

### DIFF
--- a/testsuite/pytests/test_multimeter.py
+++ b/testsuite/pytests/test_multimeter.py
@@ -134,5 +134,7 @@ def test_bad_intervals_detected(interval):
     Test that NEST raises BadProperty if interval cannot be represented as multiple of resolution.
     """
 
+    nest.resolution = 0.1
+
     with pytest.raises(nest.kernel.NESTError, match="BadProperty"):
         nest.Create("multimeter", params={"interval": interval})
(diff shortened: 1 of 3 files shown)


testsuite/pytests/test_multimeter.py
```diff
@@ -126,3 +126,13 @@ def test_identical_recording_from_multiple_multimeters(model):
...
     for recordable in recordables:
         nptest.assert_array_equal(mm1.events[recordable], mm2.events[recordable])
+
+
+@pytest.mark.parametrize("interval", [0, 0.05, 0.15])
+def test_bad_intervals_detected(interval):
+    """
+    Test that NEST raises BadProperty if interval cannot be represented as multiple of resolution.
+    """
+
```
clinssen, line 136: Although the fixture of this test will run ResetKernel() for each of the three invocations of this test function (once for each of the listed ``interval`` parameter values), do you think it may be best to at least set an assertion on the actual value of ``nest.resolution`` here, just as a sanity check/being explicit about the preconditions?

heplesser: Good point! I now set the resolution explicitly.

clinssen (approved):

clinssen:
Thanks for the explanation, all signals green then from my side!

jessica-mitchell (approved):
lgtm!

PR #3638, opened 2025-11-20
Port test_multiple_synapses from SLI to Py

heplesser:
I renamed the tests to more specific names and dropped explicit checks for expected values, because we already have those checks in place in the non-MPI case in https://github.com/nest/nest-simulator/blob/master/testsuite/pytests/test_multiple_synapses.py, so here we only need to check that we get consistent results when going MPI parallel.


testsuite/pytests/sli2py_mpi/test_collocated_synapses_mpi.py
```diff
@@ -0,0 +1,62 @@
...
+import pandas as pd
+import pytest
+from mpi_test_wrapper import MPITestAssertEqual
+
+# Functions to be passed via decorators must be in module namespace without qualifiers
+from numpy import meshgrid
+
+
+@pytest.mark.skipif_incompatible_mpi
+@MPITestAssertEqual([1, 2, 4], debug=True)
```
JanVogelsang:
```suggestion
@MPITestAssertEqual([1, 2, 4], debug=False)
```

JanVogelsang:
```suggestion
@MPITestAssertEqual([1, 2, 4], debug=False)
```

JanVogelsang (approved):

PR #3640, opened 2025-11-20
Port test_neuron_vp* from SLI to Py

heplesser:
This port combines the plain and *_global.sli variants of the test. Since results can be checked on a per-rank basis, I move this to plain python tests and mpi/{2,3}. I chose 2 and 3 ranks instead of 1, 2, 4, because the odd number might turn up different problems than all evens.


suku248 (approved):
Looks good. Thanks @heplesser !

PR #3642, opened 2025-11-20
Port test_parallel_conn_and_rand from SLI to Py

heplesser:


testsuite/pytests/sli2py_mpi/test_parallel_conn_and_rand.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.mark.skipif_missing_threads
+@MPITestAssertEqual([1, 2, 4], debug=False)
+def test_parallel_cond_and_rand():
+    """
+    Parallel Connect and Randomize Test Script
+
+    This script creates a neuron population to itself, randomizing
+    weight, delay, receptor type and one synaptic parameter. It runs
+    with a fixed number of virtual processes, and checks that connections
+    are invariant on executing with a varying number of MPI processes.
```
suku248:
```suggestion
    This test connects a neuron population to itself, randomizing
    weight, delay, receptor type and one synaptic parameter. It runs
    with a fixed number of virtual processes, and checks that connections
    are invariant on executing with a varying number of MPI processes.
```

suku248 (approved):
Thanks @heplesser !

PR #3645, opened 2025-11-21
Port test_spatial_distributed_positions from SLI to Py

heplesser:


testsuite/pytests/sli2py_mpi/test_spatial_distributed_positions.py
```diff
@@ -0,0 +1,50 @@
...
+@MPITestAssertEqual([1, 2, 4], debug=False)
+def test_spatial_distributed_positions():
+    """
+    Confirm that layers have metadata and that positions are consistent in parallel simulations.
+
+    The test is performed on connection data written to OTHER_LABEL.
+    """
+
+    import nest
+    import pandas as pd
```
otcathatsya, line 37: should we move the imports to the top?

otcathatsya (approved):

testsuite/pytests/sli2py_mpi/test_spatial_distributed_positions.py
heplesser: In these MPI tests, all imports that are needed in the actual test code need to be inside the test function, so they get written to the runner.py file created and executed.

PR #3646, opened 2025-11-21
Port test_spatial_pairwise_bernoulli_on_source_param from SLI to Py

heplesser:
The only new thing about this test compared to those already ported in #3635 to test_spatial.py is that the positions here are specified through random parameters. Therefore, I modified this test to checking the positions of the neurons for consistency, not connections drawn based on those positions.


suku248:
@otcathatsya The problem is likely that MPI processes get in each other's way when writing to the XML file. The current solution is to rerun the tests, and I have triggered such a rerun :crossed_fingers:

otcathatsya (approved):

PR #3649, opened 2025-11-21
Port ticket-516 test from SLI to Py

heplesser:
The Python implementation is considerably stronger since it simulates the network with neurons that actually spike and compares the resulting spike trains.


suku248 (approved):
Looks okay and the test ran at least on MacOS.

PR #3663, opened 2025-12-01
Port regression test 686 positive params to Pytest

jessica-mitchell:
Converted using AI

### DIFF
--- a/testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
+++ b/testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
@@ -20,14 +20,9 @@
 # along with NEST.  If not, see <http://www.gnu.org/licenses/>.
 
 import math
-from collections.abc import Sequence
-
-try:
-    import numpy as np
-except ImportError:  # pragma: no cover - numpy is available in the test environment
-    np = None
 
 import nest
+import numpy as np
 
 """
 Regression test for Ticket #686.
@@ -47,28 +42,9 @@ SKIPPED_MODELS = {
 }
 
 POSITIVE_KEYS_CACHE: dict[str, tuple[str, ...]] = {}
-# Mapping of parameters to their companion parameters that must be updated together.
-#
-# Some neuron model parameters have companion parameters that must maintain matching
-# dimensions. For example, tau_sfa (spike-frequency adaptation time constant) and
-# q_sfa (adaptation amplitude) must have the same vector length. Similarly for
-# tau_stc and q_stc (spike-triggered current parameters).
-#
-# When updating these parameters via SetStatus, the model implementations validate
-# that companion parameters have matching dimensions. If only one parameter is
-# updated without its companion, a dimension mismatch error can occur, which would
-# mask the actual positive-value validation we're testing.
-#
-# The original SLI test did not handle companion keys because it tested models
-# before these dimension constraints were strictly enforced, or because the SLI
-# implementation handled partial updates differently. The Python port includes
-# companion keys to ensure the test focuses on positive-value validation rather
-# than dimension mismatches.
-#
-# When testing a parameter that has companions, we include the companion parameters
-# in the update dictionary with their current values to maintain dimension
-# consistency and avoid spurious errors.
-COMPANION_KEYS = {
+
+# Parameters that must be updated together to maintain matching dimensions
+DIMENSION_PAIRS = {
     "tau_sfa": ("q_sfa",),
     "tau_stc": ("q_stc",),
 }
@@ -87,52 +63,42 @@ def _positive_keys(model: str) -> tuple[str, ...]:
 
 
 def _as_numeric_tuple(value) -> tuple[float, ...]:
-    if np is not None and isinstance(value, np.ndarray):
+    """Convert value to tuple of floats, handling scalars, lists, and arrays."""
+    if isinstance(value, np.ndarray):
         return tuple(float(v) for v in value.tolist())
-    if isinstance(value, Sequence) and not isinstance(value, (str, bytes)):
+    if isinstance(value, (list, tuple)):
         return tuple(float(v) for v in value)
     return (float(value),)
 
 
-def _cast_like(reference, numeric_values: Sequence[float]):
-    if np is not None and isinstance(reference, np.ndarray):
+def _cast_like(reference, numeric_values: tuple[float, ...]):
+    """Cast numeric values to match the type of reference value."""
+    if isinstance(reference, np.ndarray):
         return np.array(numeric_values, dtype=reference.dtype)
     if isinstance(reference, tuple):
         return tuple(numeric_values)
     if isinstance(reference, list):
         return list(numeric_values)
-    return numeric_values[0]
+    return numeric_values[0] if len(numeric_values) == 1 else numeric_values
 
 
-def _allclose(values_a: Sequence[float], values_b: Sequence[float], *, abs_tol: float = 1e-12) -> bool:
+def _allclose(values_a: tuple[float, ...], values_b: tuple[float, ...], *, abs_tol: float = 1e-12) -> bool:
     return len(values_a) == len(values_b) and all(
         math.isclose(a, b, rel_tol=0.0, abs_tol=abs_tol) for a, b in zip(values_a, values_b)
     )
 
 
-def _get_companion_updates(neuron, key):
-    """Get companion parameter values for a given key to include in updates."""
-    return {companion: neuron.get(companion) for companion in COMPANION_KEYS.get(key, ())}
+def _get_dimension_pairs(neuron, key):
+    """Get dimension-paired parameter values to include in updates."""
+    return {pair: neuron.get(pair) for pair in DIMENSION_PAIRS.get(key, ())}
 
 
 def _test_parameter_update(neuron, key, raw_value, new_values):
-    """
-    Test updating a parameter with new values, including companion parameters.
-
-    Args:
-        neuron: The neuron object
-        key: Parameter key to update
-        raw_value: Original raw value (used to preserve type)
-        new_values: New numeric values as tuple[float, ...]
-
-    Returns: (success: bool, error_message: str | None)
-    """
-    # Include companion parameters to avoid dimension mismatch errors
-    companion_updates = _get_companion_updates(neuron, key)
-
+    """Test updating a parameter with new values, including dimension-paired parameters."""
+    dimension_pairs = _get_dimension_pairs(neuron, key)
     try:
         update = {key: _cast_like(raw_value, new_values)}
-        update.update(companion_updates)
+        update.update(dimension_pairs)
         nest.SetStatus(neuron, update)
         return True, None
     except nest.kernel.NESTError as err:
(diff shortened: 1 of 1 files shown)


testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,171 @@
...
+}
+
+POSITIVE_KEYS_CACHE: dict[str, tuple[str, ...]] = {}
+COMPANION_KEYS = {
+    "tau_sfa": ("q_sfa",),
+    "tau_stc": ("q_stc",),
+}
+
+
+def _models_to_check():
```
abigailm, line 53: should check that list is not empty, otherwise the test will succeed trivially

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,171 @@
...
+    return [model for model in nest.node_models if model not in SKIPPED_MODELS]
+
+
+def _positive_keys(model: str) -> tuple[str, ...]:
+    if model not in POSITIVE_KEYS_CACHE:
+        defaults = nest.GetDefaults(model)
+        keys = tuple(key for key in defaults if key == "C_m" or (len(key) >= 4 and key.startswith("tau_")))
```
abigailm: Remove the len(key) >= 4 condition as it is redundant

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,171 @@
...
+            continue
+
+        nest.ResetKernel()
+        nest.SetKernelStatus({"dict_miss_is_error": False})
+
+        for key in positive_keys:
+            for candidate in (0.0, -1.0):
+                neuron = nest.Create(model)
+                raw_value = neuron.get(key)
+                original_values = _as_numeric_tuple(raw_value)
```
abigailm, line 131: why the conversion to numeric tuple?

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,171 @@
...
+                if not original_values:
+                    continue
+                invalid_values = tuple(float(candidate) for _ in original_values)
```
abigailm, line 134: why the conversion to float, it already is one

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,171 @@
...
+                companion_updates = {companion: neuron.get(companion) for companion in COMPANION_KEYS.get(key, ())}
```
abigailm: what is this? There is no mention of companion_updates in the sli version

abigailm: In general, this function seems convoluted and is not documented. It should be simplified - the thing it is testing is quite simple, so I don't see why the code is so complicated. If the complex code is justified, it must be better documented.

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,171 @@
...
+            continue
+
+        positive_keys = _positive_keys(model)
+        if not positive_keys:
+            continue
+
+        nest.ResetKernel()
+        nest.SetKernelStatus({"dict_miss_is_error": False})
+
+        for key in positive_keys:
```
abigailm, line 171: See also comment above - reduce complexity if possible; add explanation of complexity if not.

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: see line 149

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: comments are added

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: used the ai to try to add comments and reduce complexity

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: removed

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,225 @@
...
+                    continue
+                # Create invalid values matching the dimension (scalar -> 1 element, vector -> N elements)
+                invalid_values = tuple(float(candidate) for _ in original_values)
+
+                success, _ = _test_parameter_update(neuron, key, raw_value, invalid_values)
+                if success:
+                    # Update succeeded but shouldn't have - check if value actually changed
+                    current_values = _as_numeric_tuple(neuron.get(key))
+                    if not _allclose(current_values, original_values):
+                        failing_cases.append((model, key, f"accepted_non_positive_{candidate}"))
```
abigailm, line 140: So if I understand correctly, the code finds it acceptable if `_test_parameter_update` returnds with success, as long as the value doesn't change. But that isn't ok - it suggests that the user could think they have set the parameter but they haven't. Perhaps it would be better to add this to `failing_cases` but with a different string, indicating that the the neuron lies about accepting the non positive value.

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,225 @@
...
+        positive_keys = _positive_keys(model)
+        if not positive_keys:
+            continue
+
+        nest.ResetKernel()
+        nest.SetKernelStatus({"dict_miss_is_error": False})
+
+        for key in positive_keys:
+            for candidate in (0.0, -1.0):
+                neuron = nest.Create(model)
```
abigailm, line 129: if you add a line saying skip if model is iaf_tum_2000 and candidate == 0.0 (similar to line 195) then we can remove iaf_tum_2000 from the SKIPPED_MODELS list

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: Ok added a test case to handle 'ignored' case

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: done

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,231 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+import math
+from collections.abc import Sequence
+
+try:
+    import numpy as np
```
otcathatsya: do we need this safeguard? afaik it is not used in other py tests

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,231 @@
...
+#
+# The original SLI test did not handle companion keys because it tested models
+# before these dimension constraints were strictly enforced, or because the SLI
+# implementation handled partial updates differently. The Python port includes
+# companion keys to ensure the test focuses on positive-value validation rather
+# than dimension mismatches.
+#
+# When testing a parameter that has companions, we include the companion parameters
+# in the update dictionary with their current values to maintain dimension
+# consistency and avoid spurious errors.
```
otcathatsya: bit confused by this terminology, is companion a term we actually use in NEST?

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
otcathatsya: I think in general the test is excessively verbose; e.g. super long comments and a lot of boilerplate to handle the types correctly, which only applies to a single model (glif_cond is the only that has 2+ values for a tau_ parameter by default but there's 3 methods dedicated to handling this)

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,231 @@
...
+def _positive_keys(model: str) -> tuple[str, ...]:
+    if model not in POSITIVE_KEYS_CACHE:
+        defaults = nest.GetDefaults(model)
+        keys = tuple(key for key in defaults if key == "C_m" or key.startswith("tau_"))
+        POSITIVE_KEYS_CACHE[model] = keys
+    return POSITIVE_KEYS_CACHE[model]
+
+
+def _as_numeric_tuple(value) -> tuple[float, ...]:
+    if np is not None and isinstance(value, np.ndarray):
```
otcathatsya: again not sure if we need np checks etc

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: now it uses dimension pairs, which I think sounds better, but I don't really know what's best

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: updated

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
jessica-mitchell: Made some modifications, less verbose comments removed the try block etc

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
```diff
@@ -0,0 +1,187 @@
...
+    if isinstance(reference, tuple):
+        return tuple(numeric_values)
+    if isinstance(reference, list):
+        return list(numeric_values)
+    return numeric_values[0] if len(numeric_values) == 1 else numeric_values
+
+
+def _allclose(values_a: tuple[float, ...], values_b: tuple[float, ...], *, abs_tol: float = 1e-12) -> bool:
+    return len(values_a) == len(values_b) and all(
+        math.isclose(a, b, rel_tol=0.0, abs_tol=abs_tol) for a, b in zip(values_a, values_b)
```
otcathatsya, line 87: I think we usually use `np.testing.assert__allclose` but since we need the dimensionality check here I think the wrapper is fine

otcathatsya (approved):

testsuite/pytests/sli2py_regressions/test_ticket_686_positive_parameters.py
heplesser: Resolving after direct discussion.

PR #3664, opened 2025-12-01
Move regression test 681 to sli2py_ignore

jessica-mitchell:
Converted with AI


jessica-mitchell:
@heplesser There is a note in the SLI test:
% preparatory work for proper test code in case NEST is complied with MPI support
% **For now we just ignore this test, this will later be replaced**
% by a restart of NEST with a serial binary.

is this test then necessary to port?

PR #3669, opened 2025-12-01
Port regression test 740 to Pytests

jessica-mitchell:
Convert using AI

### DIFF
--- a/testsuite/pytests/sli2py_regressions/test_issue_740.py
+++ b/testsuite/pytests/sli2py_regressions/test_issue_740.py
@@ -89,12 +89,7 @@ def test_issue_740_step_current_generator_time_checks(resolution, allow_offgrid,
     """
 
     if should_raise:
-        with warnings.catch_warnings():
-            warnings.filterwarnings(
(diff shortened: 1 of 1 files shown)


testsuite/pytests/sli2py_regressions/test_issue_740.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    if should_raise:
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                message="SetStatus\\(\\) call failed, but nodes have already been created!",
+            )
+            with pytest.raises(nest.kernel.NESTError, match="(increasing|representable)"):
+                _simulate(resolution, allow_offgrid)
```
terhorstd: Not sure why the warning needs to be cought, would have thought the `nest.ResetKernel()` also resets created nodes counter…
```suggestion
        with pytest.raises(nest.kernel.NESTError, match="(increasing|representable)"):
            _simulate(resolution, allow_offgrid)
```

PR #3674, opened 2025-12-01
Port regression test 573 to Pytests

jessica-mitchell:
Converted with AI

### DIFF
--- a/nestkernel/model_manager.cpp
+++ b/nestkernel/model_manager.cpp
@@ -128,7 +128,9 @@ ModelManager::get_status( DictionaryDatum& dict )
   def< ArrayDatum >( dict, names::synapse_models, synapse_models );
 
   // syn_ids start at 0, so the maximal number of syn models is MAX_SYN_ID + 1
-  def< int >( dict, names::max_num_syn_models, MAX_SYN_ID + 1 );
+  // the last ID is however used as "invalid_synindex", so the final array
+  // position will always be empty in the `connections` and `source_table`.
+  def< int >( dict, names::max_num_syn_models, MAX_SYN_ID );
 }
 
 void


testsuite/pytests/sli2py_regressions/test_ticket_573.py
```diff
@@ -0,0 +1,48 @@
...
+
+    nest.ResetKernel()
+
+    max_synapse_models = nest.max_num_syn_models
+    existing_synapse_models = len(nest.synapse_models)
+    max_additional_models = max_synapse_models - existing_synapse_models
+
+    with pytest.raises(nest.kernel.NESTError, match="Synapse model count exceeded"):
+        for idx in range(max_additional_models + 1):
+            nest.CopyModel("static_synapse", f"ticket_573_syn_{idx}")
```
terhorstd: Splitting the loop into `n` that should pass and `+1` that should fail.
```suggestion
    # Fill up all possible slots for synapse models
    for idx in range(max_additional_models):
        nest.CopyModel("static_synapse", f"ticket_573_syn_{idx}")

    # make sure adding an extra does raise
    with pytest.raises(nest.kernel.NESTError, match="Synapse model count exceeded"):
        nest.CopyModel("static_synapse", f"ticket_573_syn_{max_additional_models}")
```

testsuite/pytests/sli2py_regressions/test_ticket_573.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    nest.ResetKernel()
+
+    max_synapse_models = nest.max_num_syn_models
+    existing_synapse_models = len(nest.synapse_models)
+    max_additional_models = max_synapse_models - existing_synapse_models
+
+    # Fill up all possible slots for synapse models
+    for idx in range(max_additional_models):
```
babsey, line 47: It produces error in filling up copied synapse models. I would change this:

```suggestion
    for idx in range(max_additional_models - 1):
```

terhorstd: I've fixed NEST instead of adjusting the test to tolerate wrong `max_num_syn_models`.

PR #3676, opened 2025-12-01
Port regression test 881 to Pytests

jessica-mitchell:
converted using AI

### DIFF
--- a/testsuite/pytests/sli2py_regressions/test_ticket_881.py
+++ b/testsuite/pytests/sli2py_regressions/test_ticket_881.py
@@ -39,10 +39,6 @@ NUM_NODES = len(NODE_POSITIONS)
 NUM_EXPECTED = NUM_NODES * NUM_NODES
 
 
-def _create_layer():
-    return nest.Create("iaf_psc_alpha", n=NUM_NODES, positions=spatial.free(NODE_POSITIONS))
-
-
 @pytest.mark.skipif_missing_threads
 @pytest.mark.parametrize(
     "conn_spec",
@@ -71,7 +67,7 @@ def test_ticket_881_multithreaded_spatial_connectivity(conn_spec):
     nest.ResetKernel()
     nest.local_num_threads = NUM_THREADS
 
-    layer = _create_layer()
+    layer = nest.Create("iaf_psc_alpha", n=NUM_NODES, positions=spatial.free(NODE_POSITIONS))
 
     nest.Connect(layer, layer, conn_spec=conn_spec)
 


testsuite/pytests/sli2py_regressions/test_ticket_881.py
```diff
@@ -0,0 +1,70 @@
...
+def _create_layer():
+    return nest.Create("iaf_psc_alpha", n=NUM_NODES, positions=spatial.free(NODE_POSITIONS))
+
+
+@pytest.mark.skipif_missing_threads
+@pytest.mark.parametrize(
+    "conn_spec",
+    (
+        {"rule": "pairwise_bernoulli", "p": 1.0, "allow_autapses": True},
+        {"rule": "pairwise_bernoulli", "p": 1.0, "allow_autapses": True},
```
pnbabu: These two rules are the same. We should remove one of them.

testsuite/pytests/sli2py_regressions/test_ticket_881.py
```diff
@@ -0,0 +1,70 @@
...
+        {"rule": "fixed_outdegree", "outdegree": NUM_NODES, "allow_autapses": True},
+        {"rule": "fixed_indegree", "indegree": NUM_NODES, "allow_autapses": True},
```
pnbabu, line 53: `allow_autapses` is by default `True`. Should it be explicitly set?

jessica-mitchell: I think for a test, it's better to be explicit?

testsuite/pytests/sli2py_regressions/test_ticket_881.py
jessica-mitchell: yes, but there are 4 scenarios that should be tested, so I'm wondering if something should be changed in on of these

pnbabu: I just checked the C++ code, and I think for the first two `conn_spec`s, we need an additional attribute `use_on_source=True/False`. It also in the docs here: https://nest-simulator.readthedocs.io/en/latest/networks/spatially_structured_networks.html#connections

testsuite/pytests/sli2py_regressions/test_ticket_881.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+
+
+NUM_THREADS = 4
+NODE_POSITIONS = [[0.0, 0.0]] * 4
+NUM_NODES = len(NODE_POSITIONS)
+NUM_EXPECTED = NUM_NODES * NUM_NODES
+
+
+def _create_layer():
```
pnbabu: Since this function only calls `Create()`, maybe we do not need a function here.

testsuite/pytests/sli2py_regressions/test_ticket_881.py
abigailm: I agree. What if, in the future, that default were to be changed? Then the test would silently become invalid.

PR #3678, opened 2025-12-01
Port MUSIC tests from SLI to Python

JanVogelsang:
This PR ports the cont_proxy MUSIC test from SLI to Python.

@heplesser and I also discovered a bug in MUSIC #3679 related to the `cont_proxy` test case. However, this PR is already valid even without the bug being fixed, as this specific test setup already works.

### DIFF
--- a/testsuite/sli2py_music/test_event_proxies_issue-696_receiver.py
+++ b/testsuite/sli2py_music/test_event_proxies_issue-696_receiver.py
@@ -24,6 +24,9 @@ import nest
 n_neurons = 11
 
 neurons = nest.Create("iaf_psc_alpha", n_neurons)
+
+# We need a range of [0, N-1] here, not a list of neuron ids. We therefore use list(range(n_neurons)) instead of
+# neurons.neuron_id.
 inputs = nest.Create(
     "music_event_in_proxy", n_neurons, params={"port_name": ["in"] * n_neurons, "music_channel": list(range(n_neurons))}
 )
--- a/testsuite/sli2py_music/test_event_proxies_issue-696_sender.py
+++ b/testsuite/sli2py_music/test_event_proxies_issue-696_sender.py
@@ -33,7 +33,8 @@ nest.Connect(generators, neurons, "one_to_one", {"delay": 0.1})
 
 meop = nest.Create("music_event_out_proxy", params={"port_name": "out"})
 
-
+# We need a range of [0, N-1] here, not a list of neuron ids. We therefore use list(range(n_neurons)) instead of
+# neurons.neuron_id.
 nest.Connect(neurons, meop, "all_to_all", {"music_channel": [list(range(n_neurons))]})
 
 nest.Simulate(1)


testsuite/sli2py_music/test_cont_proxy.music
```diff
@@ -1,5 +1,5 @@
 /*
- *  test_cont_proxies.music
+ *  test_rate_proxy.music
```
heplesser:
```suggestion
 *  test_cont_proxy.music
```

testsuite/sli2py_music/test_event_proxies.music
```diff
@@ -0,0 +1,12 @@
+
```
heplesser, line 19: Lacks copyright header.

testsuite/do_tests.sh
```diff
@@ -144,16 +144,18 @@ HAVE_OPENMP="$(sli -c 'is_threaded =only')"
 if test "${HAVE_MPI}" = "true"; then
     MPI_LAUNCHER="$(sli -c 'statusdict/mpiexec :: =only')"
     MPI_LAUNCHER_VERSION="$($MPI_LAUNCHER --version | head -n1)"
+    SLI_MPIEXEC_PREFLAGS="--prefix $(python -c 'import sys; print(sys.prefix)')"
```
heplesser:
```suggestion
    MPIEXEC_PREFLAGS="--prefix $(python -c 'import sys; print(sys.prefix)')"
```
with corresponding changes below.

heplesser (changes requested):
@JanVogelsang Great that this works now :)! The slight modernization of the tests you have done is good. I also like that you removed the "skip if not music" checks—they make no sense here because these files are never seen directly by pytest.

I have a few comments in the code and then a few I could not attach to files:
- `empty.sli` should probably become `empty.py` or even better `test_music_failure_1_empty.py` so it is clear where it belongs.
- The two `test_music_failure_*` cases still need to be adapted.

testsuite/sli2py_music/test_event_proxies.music
terhorstd: I think `.music` files are not checked in [check-copyright-headers](https://github.com/nest/nest-simulator/blob/master/build_support/check_copyright_headers.py#L89). Would it make sense to include them?

testsuite/sli2py_music/test_music_failure1.music
terhorstd, line 1: This will also break if there is no more `nest` binary after the merge.

testsuite/sli2py_music/test_event_proxies_issue-696_receiver.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+import nest
+
+n_neurons = 11
+
+neurons = nest.Create("iaf_psc_alpha", n_neurons)
+inputs = nest.Create(
+    "music_event_in_proxy", n_neurons, params={"port_name": ["in"] * n_neurons, "music_channel": list(range(n_neurons))}
```
terhorstd, line 31: Why `list(range(n_neurons))` and not `list(neurons)`?

testsuite/sli2py_music/test_event_proxies_issue-696_sender.py
```diff
@@ -0,0 +1,39 @@
...
+n_neurons = 11
+generators = nest.Create("spike_generator", n_neurons, params={"spike_times": [0.1, 0.2, 0.3]})
+neurons = nest.Create("parrot_neuron", n_neurons)
+
+nest.Connect(generators, neurons, "one_to_one", {"delay": 0.1})
+
+meop = nest.Create("music_event_out_proxy", params={"port_name": "out"})
+
+
+nest.Connect(neurons, meop, "all_to_all", {"music_channel": [list(range(n_neurons))]})
```
terhorstd, line 38: Same question here: `[list(range(n_neurons))]` → `[list(neurons)]` ?

testsuite/do_tests.sh
```diff
@@ -422,16 +424,21 @@ if test "${MUSIC}"; then
...
 
-        # Collect the list of SLI files from the '.music' file.
-        sli_files="$(grep '\.sli' "${music_file}" | sed -e "s#args=#${TESTDIR}#g")"
-        sli_files="$(for f in ${sli_files}; do if test -f "${f}"; then echo "${f}"; fi; done)"
-        sli_files="${sli_files//$'\n'/ }"
+        # Collect the list of Python files from the '.music' file.
+        if py_file_matches="$(grep '\.py' "${music_file}")"; then
+            py_files="${py_file_matches//binary=/${TESTDIR}}"
+            py_files="$(for f in ${py_files}; do if test -f "${f}"; then echo "${f}"; fi; done)"
+            py_files="${py_files//$'\n'/ }"
```
terhorstd: When handling the `for x in $manywords; do`, the `$manywords` will expand to separate items in case any `$IFS` chars appear in the names, as it is unquoted.
What could help is "… `${name[@]}` expands each element of name to a separate word" (see man bash), which we can use in a similar way also in `${array[@]//…/…}` and other expressions:
```suggestion
        readarray py_file_matches < <(grep '\.py' "${music_file}")
        if [ ${#py_file_matches} -gt 0 ]; then
            readarray py_files < <(for f in "${py_file_matches[@]//binary=/${TESTDIR}}"; do if test -f "${f}"; then echo "${f}"; fi; done)
```

testsuite/sli2py_music/test_event_proxies_issue-696_receiver.py
JanVogelsang: `music_channel` expects a list of integers, not a NodeCollection (as returned by nest.Create)

testsuite/do_tests.sh
JanVogelsang: I'm not much of a bash expert, so I'll just accept these changes. If there are any other places that need to be adapted, please let me know.

testsuite/sli2py_music/test_music_failure1.music
JanVogelsang: Fixed

testsuite/do_tests.sh
JanVogelsang: Fixed

testsuite/sli2py_music/test_event_proxies.music
JanVogelsang: Added music files to check-copyright-headers

testsuite/sli2py_music/test_cont_proxy.music
JanVogelsang: Fixed

testsuite/sli2py_music/test_event_proxies_issue-696_receiver.py
terhorstd: isn't `list(neurons)` of type `list[int]`?

JanVogelsang: If it should be that would be a bug in PyNEST, as I just tested it and NodeCollections don't seem to be converted to a list of ints automatically:
```py
>>> n = nest.Create("iaf_psc_alpha", 2)
>>> n
NodeCollection(metadata=None, model=iaf_psc_alpha, size=2, first=1, last=2)
>>> list(n)
[NodeCollection(metadata=None, model=iaf_psc_alpha, size=1, first=1), NodeCollection(metadata=None, model=iaf_psc_alpha, size=1, first=2)]
>>> 
```

testsuite/sli2py_music/test_event_proxies_issue-696_sender.py
JanVogelsang: Added comment

testsuite/sli2py_music/test_event_proxies_issue-696_receiver.py
JanVogelsang: Added comment

terhorstd: OK, I wasn't aware that there is no connection. Now it's much better to unerstand. :+1:

PR #3681, opened 2025-12-01
Make parsing of XML files robust against extra > at end of file

heplesser:
This occurs regularly for mpi/4 tests and I have never observed any other problem there. The cause is multiple MPI processes writing simultaneously.

### DIFF
--- a/requirements_testing.txt
+++ b/requirements_testing.txt
@@ -28,4 +28,3 @@ mypy >= 0.8
 junitparser >= 2
 clang-format == 17.0.4
 attrs
-lxml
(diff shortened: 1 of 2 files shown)


requirements_testing.txt
```diff
@@ -28,3 +28,4 @@ mypy >= 0.8
 junitparser >= 2
 clang-format == 17.0.4
 attrs
+lxml
```
terhorstd:
```suggestion
```

terhorstd (approved):
LGTM :+1:

PR #3682, opened 2025-12-02
Convert unittest test_symmetric_connections to Pytest

jessica-mitchell:
uses ai for conversion


testsuite/pytests/sli2py_connect/test_symmetric_connections.py
```diff
@@ -0,0 +1,352 @@
...
+        set1,
+        set2,
+        conn_spec={"rule": "one_to_one", "make_symmetric": True},
+        syn_spec={"synapse_model": "static_synapse"},
+    )
+
+    # Get forward and backward connections
+    fwd = nest.GetConnections(source=set1)
+    bck = nest.GetConnections(source=set2)
+
```
shimoura, line 119: I wonder in case the `make_symmetric` fails somehow and creates an extra backward connection. For instance, if `len(bck)` is 6 and `len(fwd)` is 5. The for loop below can be completed without error since it can find all matches. But the extra `bck` connection would be ignored. 

A simple way to address it could be just ensuring:
 `assert len(fwd) == len(bck)`

testsuite/pytests/sli2py_connect/test_symmetric_connections.py
```diff
@@ -0,0 +1,352 @@
...
+            "weight": [1.0, 2.0, 3.0, 4.0, 5.0],
+            "delay": [1.5, 2.5, 3.5, 4.5, 5.5],
+            "alpha": [1.2, 2.2, 3.2, 4.2, 5.2],
+        },
+    )
+
+    # Get forward and backward connections
+    fwd = nest.GetConnections(source=set1)
+    bck = nest.GetConnections(source=set2)
+
```
shimoura, line 196: Add `assert len(fwd) == len(bck)` if needed. Same reason as the other comment.

shimoura (commented):
All tests passed, but worth checking if my comment makes sense or if this is already being safe by other means.

testsuite/pytests/sli2py_connect/test_symmetric_connections.py
jessica-mitchell: ok, assert is added

testsuite/pytests/sli2py_connect/test_symmetric_connections.py
jessica-mitchell: ok done

shimoura (approved):

PR #3690, opened 2025-12-02
Port unittest test_stdp_synapse to Pytest

jessica-mitchell:
Uses AI for conversion
Note that there is already a test_stdp_synapse.py in `pytests` but seems to be a different test, name for this one added `_basic` to separate the two


testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
```diff
@@ -0,0 +1,220 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with NEST.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Name: testsuite::test_stdp_synapse - basic test of stdp_synapse
+
+Synopsis: (test_stdp_synapse) run
```
abigailm: remove SLI code, replace with an accurate synopsis if one is required for a unit test, otherwise remove line completely

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
```diff
@@ -0,0 +1,220 @@
...
+    sr_post = nest.Create("spike_recorder")
+
+    # Set defaults for stdp_synapse
+    # Note: "lambda" is a Python keyword, but can be used as a dict key
+    # Note: weight is not set in defaults, it's set in the Connect call
+    nest.SetDefaults(
+        "stdp_synapse",
+        {
+            "alpha": alpha,
+            "lambda": lambda_val,  # noqa: A001
```
abigailm: remove comment or replace by one that can be understood

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
```diff
@@ -0,0 +1,220 @@
...
+        if pre_spike == post_spike:
+            # pre- and post-syn. spike at the same time
+            if last_post != post_spike:
+                facilitate()
+            if last_pre != pre_spike:
+                depress()
+            if j + 1 < len(pre_spikes):
+                update_K_plus()
+                next_pre_spike()
+                if i + 1 < len(post_spikes):
```
abigailm, line 189: is it correct that this `if` is only carried out if the previous `if` is true? Because the overall `while` loop continues as long as either of the lists have still got spikes.

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
```diff
@@ -0,0 +1,220 @@
...
+            if j + 1 < len(pre_spikes):
+                next_pre_spike()
+            else:
+                # we don't consider the post-syn. spikes after the last pre-syn. spike
+                break
+        else:
+            # next spike is a post-syn. spike
+            facilitate()
+            update_K_minus()
+            if i + 1 < len(post_spikes):
```
abigailm: can't we just call `next_post_spike` without this condition? It anyway checks the length of `post_spikes` and sets `post_spike` to something big if we are at the end of the vector

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
jessica-mitchell: I think for now we just leave the name and remove SLI syntax

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
jessica-mitchell: so I assume the #noqa is required, but added a bit more context

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
jessica-mitchell: some comments added, according to the sli it seems to be the case

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
jessica-mitchell: simplified

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
```diff
@@ -181,13 +179,18 @@ def depress():
...
+            # Exit if no more pre-spikes (even if post-spikes remain, we can't form new STDP pairs)
             if j + 1 < len(pre_spikes):
                 update_K_plus()
                 next_pre_spike()
+                # Advance post-spike if we also advanced pre-spike
+                # next_post_spike() handles end-of-vector case internally, but only update_K_minus if more spikes exist
                 if i + 1 < len(post_spikes):
                     update_K_minus()
-                    next_post_spike()
+                next_post_spike()
```
akorgor, line 191: Was de-denting `next_post_spike()` on purpose?

akorgor (approved):
I added one comment to confirm whether a change was intentional. If it was, please feel free to merge.

testsuite/pytests/sli2py_synapses/test_stdp_synapse_basic.py
jessica-mitchell: yes it was

PR #3691, opened 2025-12-03
Port unittest test_syn_hom_w to Pytest

jessica-mitchell:
Converted with AI, uses new testutil function


testsuite/pytests/sli2py_synapses/test_syn_hom_w.py
```diff
@@ -0,0 +1,126 @@
...
+    # Prepare actual data array
+    actual_data = np.column_stack((times, voltages))
+
+    # Reference data needs to have times in ms, while data above is in 0.1 ms steps
+    REFERENCE_DATA[:, 0] *= 0.1
+
+    actual, expected = testutil.get_comparable_timesamples(dt, actual_data, REFERENCE_DATA)
+
+    # Check that the function did not return empty arrays
+    assert len(actual) > 0, (
```
otcathatsya: I think this is unnecessary since get_comparable_timesamples already does
```python

    assert len(actual[simulated_points]) > 0, "The recorded data did not contain any relevant timesamples"```
before returning it

**Review by otcathatsya — APPROVED** [review]

File: testsuite/pytests/sli2py_synapses/test_syn_hom_w.py (hunk 1 above)
Comment (heplesser) [reply testsuite/pytests/sli2py_synapses/test_syn_hom_w.py]: Good point, will fix this here.

File: testsuite/pytests/sli2py_synapses/test_syn_hom_w.py [hunk 2]
```diff
@@ -0,0 +1,126 @@
...
+        f"get_comparable_timesamples() returned empty arrays - " f"no matching time points found for resolution dt={dt}"
+    )
```
heplesser:
```suggestion
    # We know that get_comparable_timesamples() will return only if it has at least one data point to return
```

PR #3702, opened 2025-12-05
Extend test suite to run more tests also in pre-commit

terhorstd:
The test framework uses some outdated versions that start to become difficult with newer python versions. The main change in this PR is the update of some `pyproject.toml` sections and the `.pre-commit-config.yaml`. New checks are added to improve developer-local checks as well as the CI side.

In more detail:
* The `mypy.ini` was translated to `pyproject.toml` to have config in one place
* More pylint errors are silenced for now to have them being fixed in a separate PR.
* `nest-server` code was slightly refactored to meet absolute minimum quality requirements.
* some files with implicit relative import require explicit `from . import connect_test_base` to be checked. (which is subsequently resorted by isort)
* Some test implementation type-checking suffers from untyped pynest API, thus requiring explicit annotations like `n_events: int = spike_recorder.get("n_events")`
* mpi-tests can only be checked if directory-names are valid python module names, therefore the rename of `mpi/2/*.py` → `mpi/nproc2/*.py`
* Some updates caused a re-formatting of cpp file indention levels. (change "files changed" settings to hide white-space changes to de-clutter the review)
* minor changes to shell-scripts as suggested by `shellcheck`


gtrensch (approved):
LGTM! :+1:  Approved.

.pre-commit-config.yaml
```diff
@@ -10,43 +12,95 @@ repos:
       - id: fix-byte-order-marker
 
   - repo: https://github.com/gitleaks/gitleaks
-    rev: v8.16.3
+    rev: v8.10.0
```
heplesser, line 15: Why is this downgraded to an earlier version?

bin/nest-server-mpi
```diff
@@ -41,41 +37,47 @@ def log(call_name, msg):
...
-
-        log(call_name, f"sending reponse gather, data={response}")
-        comm.gather(nest.serialize_data(response), root=0)
+def main():
+    opt = docopt(__doc__)
+    if rank == 0:
+        logger.info("==> Starting NEST Server Master on rank 0")
+        nest.server.set_mpi_comm(comm)
+        nest.server.run_mpi_app(host=opt.get("--host", HOST), port=opt.get("--port", PORT))
+
```
heplesser: Remove empty line:
```suggestion
```

bin/nest-server-mpi
```diff
@@ -41,41 +37,47 @@ def log(call_name, msg):
...
-                response = call(*args, **kwargs)
-            except Exception:
-                logger.error("Failed to execute call")
-                continue
-
-        log(call_name, f"sending reponse gather, data={response}")
-        comm.gather(nest.serialize_data(response), root=0)
+def main():
+    opt = docopt(__doc__)
+    if rank == 0:
```
heplesser, line 42: Here, `rank` is a global variable, is it not? Can we avoid that, maybe by restructuring the code?

bin/nest-server-mpi
```diff
@@ -41,41 +37,47 @@ def log(call_name, msg):
...
+        logger.info("==> Starting NEST Server Master on rank 0")
+        nest.server.set_mpi_comm(comm)
+        nest.server.run_mpi_app(host=opt.get("--host", HOST), port=opt.get("--port", PORT))
+
+    else:
+        logger.info(f"==> Starting NEST Server Worker on rank {rank}")
```
heplesser, line 47: The `else` block is very large. Could one restructure the code to avoid such a long `else` block? Maybe return at the end of the `if` block?

nestkernel/position.h
```diff
@@ -805,7 +805,8 @@ Position< D, T >::length() const
 }
 
 template < int D, class T >
-Position< D, T >::operator std::string() const
+Position< D, T >::
+operator std::string() const
```
heplesser, line 809: This reformatting looks very strange to me, significantly reducing code readability. Can we keep the old form and get the formatter to understand that this should be one line? Breaking on `::` seems really weird to me.

nestkernel/simulation_manager.cpp
```diff
@@ -940,8 +940,8 @@ nest::SimulationManager::update_()
             // this loop may be empty for those threads
             // that do not have any nodes requiring wfr_update
             for ( std::vector< Node* >::const_iterator i = thread_local_wfr_nodes.begin();
-                  i != thread_local_wfr_nodes.end();
-                  ++i )
+              i != thread_local_wfr_nodes.end();
+              ++i )
```
heplesser, line 944: I do not like this change. I think the loop is more readable if all parts are aligned under each other as in the old version here. Can we not keep the old rules?

heplesser (changes requested):
@terhorstd Thanks for this PR. Unfortunately, it comes with several changes that I believe hurt code readability, that I do not understand and that break the testsuite. 

Concerning code formatting for C++, I think the changes to for-loop header indentation reduces readability. Even worse is the the line break after `::` in many method declarations. Here, we need to stick to the old style.

I also do not understand why all tests suddenly are converted to importable Python packages. As far as I understand the Pytest documentation, when keeping tests outside the installable package as we do, tests should not be packages (https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code).

All tests currently fail because `connect_test_base.py` is not found when tests try to import it.

.pre-commit-config.yaml
terhorstd: Newer version showed problems that we need to look into in another issue.

heplesser:
@terhorstd How is this PR related to #3849?

PR #3707, opened 2025-12-10
Fix issue with sorting when simulating more than 2 billion neurons

JanVogelsang:
When using Boost's `integer_sort` method, we were currently sorting the Sources by converting them to `int`s, which on most platforms are represented using 32 bits. For global node ids requiring more than 32 bits (>2,147,483,647), the sorting fails.


libnestutil/iterator_pair.h
```diff
@@ -170,7 +170,7 @@ struct rightshift_iterator_pair
   }
 
   template < typename T >
-  inline int
+  inline uint64_t
```
heplesser: Why did you choose `uint64_t`?  Given that `get_node_id()` returns `size_t`, wouldn't that be a better choice?

JanVogelsang: `get_node_id()` actually returns a `uint64_t` that's why I chose this type: https://github.com/nest/nest-simulator/blob/master/nestkernel/source.h#L56

JanVogelsang: But we might just change the type used in `Source` used to store the node-id to `size_t`. It's identical to uint64_t on most platforms anyway and there is also no apparent reason why we would need exactly 64 bits there. We just need the same number of bits we use anywhere else for node-ids, which as you already stated is `size_t`.

heplesser: I noticed that I had looked at the wrong `get_node_id()`, namely the one in `SourceTable`:

https://github.com/nest/nest-simulator/blob/cd8c487ee959eb0d87ca73d9996b914d8795727e/nestkernel/source_table.cpp#L155-L163

It actually takes the value from `Source` and returns it. I think the best way to sort this out is indeed to change to `size_t` in `Source`.

JanVogelsang:
> @heplesser (cc @JanVogelsang) I get your point, but maybe the return type of `Source::get_node_id` might change in the future, or maybe we use the sort on different types; therefore, I suggest the following:
> 
> ```c++
> template < typename T >
>   inline uint64_t
>   operator()( boost::tuples::tuple< nest::Source&, T& > s, unsigned offset )
>   {
>     using RetType = decltype( boost::get< 0 >( s ).get_node_id() );
>     constexpr auto lower_bound = std::numeric_limits<uint64_t >::lowest() <= std::numeric_limits< RetType >::lowest();
>     constexpr auto upper_bound = std::numeric_limits<uint64_t>::max() >= std::numeric_limits< RetType >::max();
> 
>     static_assert( lower_bound && upper_bound, "Lossy conversion" );
> 
>     return boost::get< 0 >( s ).get_node_id() >> offset;
>   }
> ```

I personally don't think that's necessary, simply because a couple of months ago we decided to replace all custom typedefs with built-in types and at that point also decided to use `size_t` for all ID-types (node-id, synapse-id, etc.). We might of course at one point decide to use different types again, but in that case we would probably want to come up with a more centralized logic to define the type to use for any ID-values and based on that logic adapt the abovementioned function again.

libnestutil/iterator_pair.h
JanVogelsang: Done!

heplesser (approved):
Thanks @JanVogelsang! I approve now, merging will need to wait until #3717 is merged so tests also pass on macOS.

heplesser:
> I personally don't think that's necessary, simply because a couple of months ago we decided to replace all custom typedefs with built-in types and at that point also decided to use `size_t` for all ID-types (node-id, synapse-id, etc.). We might of course at one point decide to use different types again, but in that case we would probably want to come up with a more centralized logic to define the type to use for any ID-values and based on that logic adapt the abovementioned function again.

I agree.

PR #3709, opened 2025-12-16
Redirect performance benchmarks docs to new repository

jessica-mitchell:
This PR adds an intersphinx link to the new nest/performance-benchmarks repo. It removes the images and rst files and updates the links.
The benchmark results will live in this  repo rather than here on nest-simulator. This allows us to modify and update the results as needed and not have to patch and mess with released versions of NEST.
The performance benchmark pages are found at https://performance-benchmarks.readthedocs.io/en/latest/


steffengraber (approved):
:+1: LGTM

PR #3712, opened 2025-12-17
Fix assertion to ensure range index is properly cast to size_t

akorgor:
Fix the following warning thrown by the compiler:

```bash
modelrange_manager.cpp:107:23: warning: comparison of integer expressions of different signedness: 'long int' and 'std:
:vector<nest::modelrange>::size_type' {aka 'long unsigned int'} [-Wsign-compare]                 
  107 |     assert( range_idx < modelranges_.size() );                        
```

### DIFF
--- a/nestkernel/modelrange_manager.cpp
+++ b/nestkernel/modelrange_manager.cpp
@@ -104,6 +104,7 @@ ModelRangeManager::get_model_id( size_t node_id ) const
       range_idx -= ( range_idx - left ) / 2;
     }
     assert( left + 1 < right );
+    assert( range_idx >= 0 );
     assert( static_cast< size_t >( range_idx ) < modelranges_.size() );
   }
   return modelranges_[ range_idx ].get_model_id();


nestkernel/modelrange_manager.cpp
```diff
@@ -104,7 +104,7 @@ ModelRangeManager::get_model_id( size_t node_id ) const
       range_idx -= ( range_idx - left ) / 2;
     }
     assert( left + 1 < right );
-    assert( range_idx < modelranges_.size() );
+    assert( static_cast< size_t >( range_idx ) < modelranges_.size() );
```
heplesser, line 108: Just to be entirely on the safe side that we do not cast away negative values (which we did not control for in the old version):
```suggestion
    assert( range_idx >= 0 );
    assert( static_cast< size_t >( range_idx ) < modelranges_.size() );
```

heplesser (commented):
@akorgor After approving, I had another small improvement idea. Can you take a look?

akorgor:
> @akorgor After approving, I had another small improvement idea. Can you take a look?

Looks good to me!

PR #3713, opened 2025-12-17
Fix destructor warning by updating PoolWrapper destructor signature

akorgor:
Fix the following warning thrown by the compiler:

```bash
connection_creator_impl.h:170:39: warning: template-id not allowed for destructor in C++20 [-Wtemplate-id-cdtor]
  170 | ConnectionCreator::PoolWrapper_< D >::~PoolWrapper_< D >()            
```


heplesser:
Merging based on single review as it only addresses a compiler warning.

PR #3720, opened 2025-12-26
Attempt to Fix merge of Sliced NodeCollections

med-ayssar:


nestkernel/nest.cpp
```diff
@@ -386,33 +395,64 @@ node_collection_array_index( const Datum* datum, const long* array, unsigned lon
 {
   const NodeCollectionDatum node_collection = *dynamic_cast< const NodeCollectionDatum* >( datum );
   assert( node_collection->size() >= n );
-  std::vector< size_t > node_ids;
-  node_ids.reserve( n );
 
-  for ( auto node_ptr = array; node_ptr != array + n; ++node_ptr )
+  if ( n == 0 )
```
heplesser, line 399: Am I overlooking something here or does this change essentially turn 8 lines of code into 20 without changing the logic? I suspect that the new code will be less efficient, because it requires n-1 temporary single-element NodeCollections to be created and merged, while the original code just creates a vector of node IDs and then creates the NodeCollection in one go.

libnestutil/vector_util.h
```diff
@@ -47,6 +49,65 @@ grow( std::vector< T >& v )
...
+
+template < typename Container >
+inline std::vector< std::pair< size_t, size_t > >
+split_into_contiguous_slices(
+  const Container& container,
+  const bool& useIndex = false,
+  Predicate< typename Container::value_type > pred = []( const auto& current, const auto& next )
+  { return next == current + 1; },
+  Filter< typename Container::value_type > filter = []( const auto& ) { return true; } )
+{
```
heplesser, line 68: I find this code challenging to follow without comments explaining the logic.

nestkernel/nest.cpp
```diff
@@ -386,33 +395,64 @@ node_collection_array_index( const Datum* datum, const long* array, unsigned lon
...
   assert( node_collection->size() == n );
-  std::vector< size_t > node_ids;
-  node_ids.reserve( n );
 
-  auto nc_it = node_collection->begin();
-  for ( auto node_ptr = array; node_ptr != array + n; ++node_ptr, ++nc_it )
+  std::span< const bool > span { array, n };
+
+  auto slices = vector_util::split_into_contiguous_slices(
+    span, true, []( auto current, auto next ) { return current == next; }, []( const auto& v ) { return v == true; } );
```
heplesser, line 433: This code seems even more complex and likely performance-problematic that the variant working on node IDs directly.

nestkernel/node_collection.cpp
```diff
@@ -958,34 +958,57 @@ NodeCollectionComposite::operator+( NodeCollectionPTR rhs ) const
...
+  auto check_overlap = []( const NodeCollectionComposite& c1, const NodeCollection& c2 )
   {
-    assert( stride_ > 1 or last_part_ != 0 or last_elem_ != 0 );
-    throw BadProperty( "Cannot add NodeCollection to a sliced composite." );
-  }
+    // Check if smaller contains any element in larger is usually faster
+    // But here we iterate c1 and check if c2 contains.
+    for ( const auto& elem : c1 )
+    {
+      if ( c2.contains( elem.node_id ) )
```
heplesser, line 968: For composite collections `c2`, `contains()` is a complex operation, so I am afraid this is too costly. See below for other ideas.

nestkernel/node_collection.cpp
```diff
@@ -958,34 +958,57 @@ NodeCollectionComposite::operator+( NodeCollectionPTR rhs ) const
...
-        throw BadProperty( "Cannot join overlapping NodeCollections." );
+        if ( part.overlapping( *rhs_ptr ) )
+        {
+          throw BadProperty( "Cannot join overlapping NodeCollections." );
+        }
       }
     }
-    return NodeCollectionPTR( *this + *rhs_ptr );
+
+    std::vector< NodeCollectionPrimitive > new_parts = to_primitives_();
```
heplesser, line 995: I think the following could be integrated with the code below if that is revised as suggested below.

nestkernel/node_collection.cpp
```diff
@@ -958,34 +958,57 @@ NodeCollectionComposite::operator+( NodeCollectionPTR rhs ) const
...
+    merge_parts_( new_parts );
+    if ( new_parts.size() == 1 )
+    {
+      return std::make_shared< NodeCollectionPrimitive >( new_parts[ 0 ] );
+    }
+    else
+    {
+      return std::make_shared< NodeCollectionComposite >( new_parts );
+    }
   }
```
heplesser, line 1007: If I read the code above right, then execution of the `if`-branch will always lead to either `throw` or `return` and execution will in that case never pass this point. That needs to be clear from the code, right now it is hidden.

nestkernel/node_collection.cpp
```diff
@@ -1002,9 +1025,14 @@ NodeCollectionComposite::operator+( NodeCollectionPTR rhs ) const
...
+    std::vector< NodeCollectionPrimitive > parts1 = to_primitives_();
+    std::vector< NodeCollectionPrimitive > parts2 = rhs_ptr->to_primitives_();
+
+    std::vector< NodeCollectionPrimitive > new_parts;
+    new_parts.reserve( parts1.size() + parts2.size() );
+    new_parts.insert( new_parts.end(), parts1.begin(), parts1.end() );
+    new_parts.insert( new_parts.end(), parts2.begin(), parts2.end() );
+
     std::sort( new_parts.begin(), new_parts.end(), primitive_sort_op );
     merge_parts_( new_parts );
```
heplesser, line 1037: Wouldn't it be more efficient to apply merge-sort logic here, iterating over both sides in parallel? In that process, one would also detect overlaps, so that no extra overlap detection is needed, improving efficiency. Since `parts1` and `parts2` are sorted per precondition, no sorting here is needed. One can then cover the case where `rhs_ptr` is a primitive connection by just constructing a single-element vector containing it.

nestkernel/node_collection.cpp
```diff
@@ -1027,16 +1055,29 @@ NodeCollectionComposite::operator+( const NodeCollectionPrimitive& rhs ) const
     throw BadProperty( "can only join NodeCollections with the same metadata" );
   }
 
-  // check primitives in the composites for overlap
-  for ( auto part_it = parts_.begin(); part_it < parts_.end(); ++part_it )
+  // check overlap
+  if ( is_sliced_ )
   {
-    if ( part_it->overlapping( rhs ) )
+    for ( const auto& elem : *this )
```
heplesser, line 1061: Iterating over all elements of a collection is potentially very expensive, since collections can be large. We need to find a way to avoid it.

nestkernel/node_collection.cpp
```diff
@@ -1275,6 +1316,74 @@ NodeCollectionComposite::slice( size_t start, size_t end, size_t stride ) const
...
+  if ( empty() )
+  {
+    return primitives;
+  }
+
+  // Iterate over the composite to identify contiguous ranges
+  auto it = begin();
+  const auto end_it = end();
+
+  // Initialize the first range
```
heplesser, line 1340: Might want to point out here that the algorithm is very similar to `create_()`.

med-ayssar:
Sorry for wasting your time, This branch was only for testing purposes, thus closing.

PR #3721, opened 2026-01-06
Implement activation events to reduce memory usage in the e-prop mechanism

akorgor:
In the NEST implementation of the e-prop plasticity rule, all variables required to calculate weight updates are archived at every time step. When a spike activates an e-prop synapse, the archived history is retrieved from the postsynaptic neuron, and the used section is marked as no longer needed by that synapse in `update_history`. Once no synapse requires a specific part of the postsynaptic history, that part is deleted and the memory is freed. In biologically plausible scenarios, neurons often fire infrequently or stop firing during the learning process, causing their histories to accumulate indefinitely because they are never released. This PR resolves the issue by introducing activation events that, similar to spike events, trigger synapse activation so that obsolete history segments can be deleted, while not being sent to the postsynaptic neuron unlike spike events. Users can configure the interval for sending these activation events since the last spike.

Memory and runtime comparison for 100 training and 10 test iterations of the N-MNIST task with the `eprop_iaf` neuron model:

memory-comparison

- [x] Enable this mechanism also for static synapses, which would make a dedicated `eprop_input_neuron` / `eprop_parrot_neuron` obsolete.
- [x] Resolve the issue that the activation interval of the presynaptic neuron has to be larger or equal the `eprop_isi_trace_cutoff` of the postsynaptic neuron.
- [x] Preserve flush event flag in `set_sender_node_id_info` during spike delivery.
- [x] Document `erase_used_history`.
- [x] Document activation event mechanism.
- [x] Benchmark the runtime with HPC benchmark.

### DIFF
--- a/nestkernel/flush_event_mechanism.cpp
+++ b/nestkernel/flush_event_mechanism.cpp
@@ -34,15 +34,15 @@ namespace nest
 
 FlushEventMechanism::FlushEventMechanism()
   : flush_event_send_interval_( std::numeric_limits< double >::infinity() )
+  , flush_event_send_interval_steps_( Time( Time::ms( flush_event_send_interval_ ) ).get_steps() )
   , last_event_time_( 0 )
-  , flush_event_send_interval_steps_( std::numeric_limits< long >::max() )
 {
 }
 
 FlushEventMechanism::FlushEventMechanism( const FlushEventMechanism& n )
   : flush_event_send_interval_( n.flush_event_send_interval_ )
-  , last_event_time_( n.last_event_time_ )
   , flush_event_send_interval_steps_( n.flush_event_send_interval_steps_ )
+  , last_event_time_( n.last_event_time_ )
 {
 }
 
--- a/nestkernel/flush_event_mechanism.h
+++ b/nestkernel/flush_event_mechanism.h
@@ -104,11 +104,11 @@ protected:
   //! Interval since previous event after which a flush event is sent (ms).
   double flush_event_send_interval_;
 
-  //! Time of last spike or flush event (steps).
-  long last_event_time_;
-
   //! Interval since previous event after which a flush event is sent (steps).
   long flush_event_send_interval_steps_;
+
+  //! Time of last spike or flush event (steps).
+  long last_event_time_;
 };
 
 }  // namespace nest


models/eprop_iaf.cpp
```diff
@@ -209,11 +212,16 @@ eprop_iaf::Parameters_::set( const DictionaryDatum& d, Node* node )
     throw BadProperty( "Firing rate low-pass filter for regularization kappa_reg from range [0, 1] required." );
   }
 
-  if ( eprop_isi_trace_cutoff_ < 0.0 )
+  if ( activation_interval_ < 0 )
   {
-    throw BadProperty( "Cutoff of integration of eprop trace between spikes eprop_isi_trace_cutoff ≥ 0 required." );
+    throw BadProperty( "Interval between activations activation_interval ≥ 0 required." );
```
heplesser: I'd keep it shorter:
```suggestion
    throw BadProperty( "activation_interval ≥ 0 required." );
```

models/eprop_iaf.cpp
```diff
@@ -209,11 +212,16 @@ eprop_iaf::Parameters_::set( const DictionaryDatum& d, Node* node )
...
   }
 
+  if ( eprop_isi_trace_cutoff_ < 0.0 or eprop_isi_trace_cutoff_ > activation_interval_ )
+  {
+    throw BadProperty(
+      "Computation cutoff of eprop trace 0 ≤ eprop trace eprop_isi_trace_cutoff ≤ activation_interval required." );
```
heplesser:
```suggestion
      "0 ≤ eprop_isi_trace_cutoff ≤ activation_interval required." );
```

models/eprop_iaf.h
```diff
@@ -230,6 +230,8 @@ Parameter                   Unit    Math equivalent         Default          Des
 ----------------------------------------------------------------------------------------------------------------
 Parameter                       Unit    Math equivalent             Default            Description
 =============================== ======= =========================== ================== =========================
+``activation_interval``         ms                                              3000.0 Interval between two
```
heplesser: Check the indentation on the "3000.0". It is right-aligned with the "0.0" in the row below?

nestkernel/archiving_node.h
```diff
@@ -98,6 +98,12 @@ class ArchivingNode : public StructuralPlasticityNode
   void set_status( const DictionaryDatum& d ) override;
 
 protected:
+  //! Interval between two activations in steps.
+  long activation_interval_;
```
heplesser: Would it not make more sense to put declare these eprop-specific fields in `EpropArchivingNode`?

nestkernel/archiving_node.cpp
```diff
@@ -34,7 +34,9 @@ namespace nest
 // member functions for ArchivingNode
 
 nest::ArchivingNode::ArchivingNode()
-  : n_incoming_( 0 )
+  : activation_interval_( 3 * kernel().simulation_manager.get_eprop_update_interval().get_steps() )
```
heplesser: Why 3 and not, e.g., 2, 5 or 10?

models/eprop_input_neuron.h
```diff
@@ -0,0 +1,170 @@
...
+Examples using this model
++++++++++++++++++++++++++
+
+.. listexamples:: eprop_input_neuron
+
+EndUserDocs */
+
+void register_eprop_input_neuron( const std::string& name );
+
+class eprop_input_neuron : public ArchivingNode
```
heplesser: I would rather call this `eprop_parrot_neuron` to make clear that it is a parrot neuron with the only difference being the activation spikes. This should also be explained explicitly in the user documentation.

nestkernel/eprop_archiving_node.h
```diff
@@ -104,7 +107,7 @@ class EpropArchivingNode : public Node
    *
    * @param eprop_isi_trace_cutoff The cutoff value for the inter-spike integration of the eprop trace.
    */
-  void erase_used_eprop_history( const long eprop_isi_trace_cutoff );
+  void erase_used_eprop_history( const long t_spike, const long t_spike_previous );
```
heplesser: From the warnings below, it seems an `override` may be missing. Why did that not miss before?

nestkernel/eprop_archiving_node_impl.h
```diff
@@ -72,6 +79,8 @@ template < typename HistEntryT >
 void
 EpropArchivingNode< HistEntryT >::write_update_to_history( const long t_previous_update,
   const long t_current_update,
+  const bool activation,
+  const bool previous_event_was_activation,
   const long eprop_isi_trace_cutoff )
```
heplesser: The following should avoid the warnings below
```suggestion
  const long )
```

nestkernel/eprop_archiving_node_recurrent.h
```diff
@@ -239,12 +239,25 @@ class EpropArchivingNodeRecurrent : public EpropArchivingNode< HistEntryEpropRec
...
+  long get_last_event_time();
+
   //! Firing rate regularization.
   double firing_rate_reg_;
 
   //! Average firing rate.
   double f_av_;
 
+  //! Last time the neuron spiked.
+  long last_event_time_;
```
heplesser: Could this be moved up to `EpropArchivingNode`?

nestkernel/spike_data.h
```diff
@@ -96,16 +102,17 @@ class SpikeData
...
   static constexpr int MAX_LAG = generate_max_value( NUM_BITS_LAG );
 
-  size_t lcid_ : NUM_BITS_LCID;                      //!< local connection index
-  unsigned int marker_ : NUM_BITS_MARKER_SPIKE_DATA; //!< status flag
-  unsigned int lag_ : NUM_BITS_LAG;                  //!< lag in this min-delay interval
-  unsigned int tid_ : NUM_BITS_TID;                  //!< thread index
-  synindex syn_id_ : NUM_BITS_SYN_ID;                //!< synapse-type index
+  size_t lcid_ : NUM_BITS_LCID;                                 //!< local connection index
+  unsigned int marker_ : NUM_BITS_MARKER_SPIKE_DATA;            //!< status flag
+  unsigned int activation_marker_ : NUM_BITS_MARKER_ACTIVATION; //!< activation flag
```
heplesser: I wonder if it would be better to move this after `syn_id_`, so that flags that markers used in spike routing stay close. Just a hunch that that might have some effect on performance, but not sure.

nestkernel/spike_data.h
```diff
@@ -183,10 +190,25 @@ class SpikeData
    */
   bool is_invalid_marker() const;
 
+  /**
+   * Returns whether the marker is the default marker.
```
heplesser: I don't understand the concept of "default" here. The marker is binary, so either it is on/off or set/unset. From the perspective of looking at a spike, shouldn't the method rather be called `is_activation_spike()`?

nestkernel/spike_data.h
```diff
@@ -204,24 +227,31 @@ inline SpikeData::SpikeData()
...
   , tid_( rhs.tid_ )
   , syn_id_( rhs.syn_id_ )
 {
 }
 
-inline SpikeData::SpikeData( const Target& target, const size_t lag )
+inline SpikeData::SpikeData( const Target& target, const size_t lag, const bool activation )
   : lcid_( target.get_lcid() )
   , marker_( SPIKE_DATA_ID_DEFAULT )
+  , activation_marker_( SPIKE_DATA_ID_ACTIVATION_DEFAULT )
```
heplesser:
```suggestion
  , activation_marker_( NOT_ACTIVATION_SPIKE )
```

nestkernel/spike_data.h
```diff
@@ -204,24 +227,31 @@ inline SpikeData::SpikeData()
...
   , lag_( lag )
   , tid_( target.get_tid() )
   , syn_id_( target.get_syn_id() )
 {
+  if ( activation )
+  {
+    activation_marker_ = SPIKE_DATA_ID_ACTIVATION;
```
heplesser:
```suggestion
    activation_marker_ = IS_ACTIVATION_SPIKE;
```

nestkernel/spike_data.h
```diff
@@ -85,6 +85,12 @@ enum enum_status_spike_data_id
   SPIKE_DATA_ID_INVALID
 };
 
+enum enum_activation_id
+{
+  SPIKE_DATA_ID_ACTIVATION_DEFAULT,
+  SPIKE_DATA_ID_ACTIVATION
+};
```
heplesser: Since we have binary values here, I'd choose a different naming scheme:
```suggestion
enum enum_activation_marker
{
  NOT_ACTIVATION_SPIKE,
  IS_ACTIVATION_SPIKE
};
```

libnestutil/nest_types.h
```diff
@@ -90,6 +90,7 @@ constexpr uint8_t NUM_BITS_SYN_ID = 6U;
 constexpr uint8_t NUM_BITS_LCID = 27U;
 constexpr uint8_t NUM_BITS_PROCESSED_FLAG = 1U;
 constexpr uint8_t NUM_BITS_MARKER_SPIKE_DATA = 2U;
+constexpr uint8_t NUM_BITS_MARKER_ACTIVATION = 1U;
```
heplesser: This adds one bit to `SpikeData`. Have you confirmed that `SpikeData` then still fits into 64 bits?

models/eprop_iaf.cpp
```diff
@@ -85,6 +85,7 @@ eprop_iaf::Parameters_::Parameters_()
   , kappa_( 0.97 )
   , kappa_reg_( 0.97 )
   , eprop_isi_trace_cutoff_( 1000.0 )
+  , activation_interval_( 3000.0 )
```
heplesser: Why is this set explicitly here (in ms), while somewhere further down it is set as a multiple of the cutoff and in steps? It should be consistent across all models and ideally be handled in some parent class.

models/eprop_iaf.cpp
akorgor: We describe the variable in words before its name in all BadProperty throws, so this message is consistent with the existing pattern and gives clearer context.

models/eprop_iaf.cpp
akorgor: We describe the variable in words before its name in all BadProperty throws, so this message is consistent with the existing pattern and gives clearer context.

nestkernel/archiving_node.cpp
akorgor: I did not tune this value extensively or base it on specific measurements. The optimal choice depends on available system memory and network size. In some scenarios a shorter `eprop_history` speeds up computation because vector search operations are cheaper, while activation events also incur a cost. A default of 3 is a reasonable trade-off for the standard experiments.

libnestutil/nest_types.h
akorgor: Yes, @suku248 noticed during our pair programming session that there was one unused bit.

models/eprop_iaf.cpp
akorgor: The value is specified in ms rather than steps because all e-prop neuron model parameters are defined in ms. To avoid pitfalls for the user, the input is given in ms and converted to the required steps internally.

If with "further down" you are referring to the `bsshslm` models: there the `activation_interval` must be a multiple of the `update_interval`. In the eprop+ models there is no `update_interval` anymore, so the `activation_interval` is set directly in physical time.

For simplicity, I also enforce `activation_interval` > `eprop_isi_trace_cutoff`. Relaxing this would be possible, but would require more careful handling in the algorithm to ensure correctness.

I agree that `activation_interval` should ideally be handled in a parent class. To ensure consistency across all models, I need to check whether the requirement of being a multiple of the `update_interval` can be lifted.

nestkernel/spike_data.h
akorgor: In my eyes, your proposed variable names are much clearer! @suku248, do you have any objections against the renaming of these variables?

- `SPIKE_DATA_ID_ACTIVATION_DEFAULT` → `NOT_ACTIVATION_EVENT`
- `SPIKE_DATA_ID_ACTIVATION` → `IS_ACTIVATION_EVENT`
- `enum_activation_id` → `enum_activation_event`
- `activation_marker` → `activation_event`
- `MARKER_ACTIVATION` → `ACTIVATION_EVENT`

nestkernel/spike_data.h
akorgor: Yes, I agree, I would rename the variables like this and adjust the doc strings accordingly:

```cpp
  /**
   * Returns whether this spike is an activation event.
   */
  bool is_activation_event() const;
```
```cpp
  /**
   * Marks this spike as an activation event.
   */
  void set_activation_event();

  /**
   * Marks this spike as a regular (non-activation) event.
   */
  void unset_activation_event();
```
@suku248, do you have any objections?

nestkernel/spike_data.h
akorgor: See comment below.

nestkernel/spike_data.h
akorgor: See comment below.

nestkernel/spike_data.h
akorgor: The current ordering is naive and was not chosen with performance in mind. I cannot give a qualified answer on whether this would have a measurable impact; perhaps @suku248 can comment on that? Is the grouping of the other fields already optimized in this respect? To clarify this properly, I think benchmarks would be necessary.

nestkernel/eprop_archiving_node_impl.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/5e4b6667769b60132940d7562ae9098b8456b2f9

models/eprop_input_neuron.h
akorgor: I renamed it because the term input layer is more common in a machine learning context, but I agree that the origin of the code should remain clear for NEST developers. I will rename it and add explanatory documentation. If support for static synapses were enabled, the standard parrot neuron could provide this functionality, making a dedicated `eprop_parrot_neuron` model obsolete. I will add a note to the open to-dos.

models/eprop_iaf.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/f1c2f2745206579afb31977a3ce9b9cc0bdcab7e.

nestkernel/archiving_node.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/60b9a6e1e62122bc08b54f4bc0a895667a08e3e5.

models/eprop_input_neuron.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/0f1aa8f4e63b0e6f1985276027b18e5b3aec61d8 and https://github.com/nest/nest-simulator/pull/3721/commits/df9a0af9f49f45df2879d7d0666eb64d2af7f6e5.

nestkernel/eprop_archiving_node.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/cc581de088cd57c19033534df2d34f1bc4953a9d.

nestkernel/eprop_archiving_node_recurrent.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/60b9a6e1e62122bc08b54f4bc0a895667a08e3e5.

nestkernel/spike_data.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/4ef6fbc88224a44897fe8c2c66c92086c3d8fef0.

nestkernel/spike_data.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/4ef6fbc88224a44897fe8c2c66c92086c3d8fef0.

nestkernel/spike_data.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/4ef6fbc88224a44897fe8c2c66c92086c3d8fef0.

models/eprop_iaf.cpp
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/60b9a6e1e62122bc08b54f4bc0a895667a08e3e5.

nestkernel/spike_data.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/4ef6fbc88224a44897fe8c2c66c92086c3d8fef0.

JesusEV:
#### Spike recorder test

I suggest adding an explicit test case to verify that the `spike_recorder` does not "catch" or log activation events. This could serve as a safeguard against future problems if the event delivery logic or the spike recorder implementation is refactored. This could prevent activation events from leaking into the recorder, which could corrupt firing rate calculations and analysis for any user unaware of the underlying housekeeping mechanism.

akorgor:
> #### Spike recorder test
> 
> I suggest adding an explicit test to verify that the `spike_recorder` does not "catch" or log activation events. This could serve as a safeguard against future problems if the event delivery logic or the spike recorder implementation is refactored. This could prevent activation events from leaking into the recorder, which could corrupt firing rate calculations and analysis for any user unaware of the underlying housekeeping mechanism.

Adden in c221d9e8d89b54786093a9dfb7d1f4405d21242f.

nestkernel/spike_data.h
akorgor: I tested the following bit-field configurations:

```cpp
  struct bit_field_1 {
    // original order
    size_t NUM_BITS_LCID : 27;
    unsigned int NUM_BITS_MARKER_SPIKE_DATA : 2;
    unsigned int NUM_BITS_ACTIVATION_EVENT : 1;
    unsigned int NUM_BITS_LAG : 14;
    unsigned int NUM_BITS_TID : 9;
    unsigned int NUM_BITS_SYN_ID : 9;
  };

  struct bit_field_2 {
      // NUM_BITS_ACTIVATION_EVENT moved to the end
      // all unsigned int except NUM_BITS_SYN_ID converted to size_t
      size_t NUM_BITS_LCID : 27;
      size_t NUM_BITS_MARKER_SPIKE_DATA : 2;
      size_t NUM_BITS_LAG : 14;
      size_t NUM_BITS_TID : 9;
      unsigned int NUM_BITS_SYN_ID : 9;
      size_t NUM_BITS_ACTIVATION_EVENT : 1;
  }; 

  struct bit_field_3 {
      // NUM_BITS_ACTIVATION_EVENT moved to the end
      // all unsigned int converted to size_t
      size_t NUM_BITS_LCID : 27;
      size_t NUM_BITS_MARKER_SPIKE_DATA : 2;
      size_t NUM_BITS_LAG : 14;
      size_t NUM_BITS_TID : 9;
      size_t NUM_BITS_SYN_ID : 9;
      size_t NUM_BITS_ACTIVATION_EVENT : 1;
  };

  struct bit_field_4 {
      // NUM_BITS_ACTIVATION_EVENT moved to the end
      size_t NUM_BITS_LCID : 27;
      unsigned int NUM_BITS_MARKER_SPIKE_DATA : 2;
      unsigned int NUM_BITS_LAG : 14;
      unsigned int NUM_BITS_TID : 9;
      unsigned int NUM_BITS_SYN_ID : 9;
      unsigned int NUM_BITS_ACTIVATION_EVENT : 1;
  }; 

  struct bit_field_5 {
      // NUM_BITS_ACTIVATION_EVENT moved to the end
      // size_t converted to unsigned int
      unsigned int NUM_BITS_LCID : 27;
      unsigned int NUM_BITS_MARKER_SPIKE_DATA : 2;
      unsigned int NUM_BITS_LAG : 14;
      unsigned int NUM_BITS_TID : 9;
      unsigned int NUM_BITS_SYN_ID : 9;
      unsigned int NUM_BITS_ACTIVATION_EVENT : 1;
  };
  ```
as implemented in [bit-field.cpp](https://github.com/user-attachments/files/25154136/bit-field.cpp), yielding the sizes in bytes:
  ```bash
  1: 8
  2: 8
  3: 8
  4: 16
  5: 12
  ```
Moving `NUM_BITS_ACTIVATION_EVENT` preserves the 64-bit layout only if other fields' types are adjusted. Only the first three variants remain 8 bytes. In NEST benchmark simulations (128k sparsely connected eprop ignore and fire neurons simulated on 1 node), the first two variants show nearly identical real-time factors, while the third a slightly higher one: 
nest_version_comparison
Since the original ordering satisfies the size constraint without changing types and the quick performance benchmark shows no benefit from moving `NUM_BITS_ACTIVATION_EVENT` to the end, in my eyes, the simplest and safest choice is to keep the original ordering.

File: models/eprop_iaf.cpp (hunk 16 above)
Comment (heplesser) [reply models/eprop_iaf.cpp]: I would treat the activation interval similar to how we handle the refractory time in neurons: The user sees and passes values in ms, but the internal representation is in steps. The conversion is done in `set/get_status()`.

File: models/eprop_readout_bsshslm_2020.cpp
Comment (heplesser) [inline models/eprop_readout_bsshslm_2020.cpp]: If `lag` is not used in the method, then

```suggestion
eprop_readout_bsshslm_2020::compute_error_signal_mean_squared_error( const long )
```

will avoid the warnings.

models/eprop_synapse.h
```diff
@@ -335,6 +341,9 @@ class eprop_synapse : public Connection< targetidentifierT >
   //! Value of spiking variable one time step before t_previous_spike_.
   double z_previous_buffer_ = 0.0;
 
+  //! Sum of gradients.
```
heplesser, line 346: Summed over what?

nestkernel/eprop_archiving_node_impl.h
```diff
@@ -189,6 +261,57 @@ EpropArchivingNode< HistEntryT >::get_eprop_history_duration() const
   return Time::get_resolution().get_ms() * eprop_history_.size();
 }
 
+template < typename HistEntryT >
+inline void
+EpropArchivingNode< HistEntryT >::get_status( DictionaryDatum& d ) const
+{
+  if ( get_name().find( "bsshslm_2020" ) == std::string::npos )
```
heplesser: This should be done in the derived class, same for the activation interval and set_status

nestkernel/event_delivery_manager.cpp
```diff
@@ -685,6 +693,14 @@ EventDeliveryManager::deliver_events_( const size_t tid, const std::vector< Spik
         syn_id_batch[ j ] = spike_data.get_syn_id();
         lcid_batch[ j ] = spike_data.get_lcid();
         se_batch[ j ].set_sender_node_id_info( tid_batch[ j ], syn_id_batch[ j ], lcid_batch[ j ] );
+        if ( spike_data.is_activation_event() )
```
heplesser: This could would get simpler if we could pass the result of `is_activation_event()` to `set_activation()`.

nestkernel/node.h
```diff
@@ -487,6 +487,16 @@ class Node
...
 
+  /**
+   * @brief Registers an eprop synapse and initializes the update history.
+   *
+   * The time for the first entry of the update history is set to the neuron specific shift for `bsshslm_2020`
+   * models and to the negative transmission delay from the recurrent to the output layer otherwise.
+   *
+   * @throws IllegalConnection
+   */
+  virtual void register_synapse();
```
heplesser: If this is for eprop, why not mention it in the name? And maybe for consistency with the stdp-related method above, call it `register_eprop_connection()?

nestkernel/activation_event_node.h
```diff
@@ -0,0 +1,116 @@
...
+namespace nest
+{
+/**
+ * @brief Base class providing activation event tracking functionality.
+ *
+ * This class provides the basic infrastructure for tracking activation events,
+ * including the interval between activation events and the time of the last event.
+ * It can be used by neurons that need to send periodic activation events.
+ */
+class ActivationEventNode
```
heplesser: I like the idea of implementing this as a mix-in class, but that class is not a node. So maybe better `ActivationEventMechanism`?

models/parrot_neuron.h
```diff
@@ -82,7 +94,7 @@ EndUserDocs */
 
 void register_parrot_neuron( const std::string& name );
 
-class parrot_neuron : public ArchivingNode
+class parrot_neuron : public ArchivingNode, public ActivationEventNode
```
heplesser: I do not really like the idea of parrot neurons sending out activation events every 3 seconds by default. One solution would be to set the default to `inf` (which only works for `double`, so one would need to "trick" a bit for the internal steps representation). Or one would have a separate `eprop_parrot_neuron` class derived from `parrot_neuron` and the ActivationEvent mechanism.

heplesser (changes requested):
Thanks for all the fixes. I have a few comments left, see inline.

models/eprop_synapse.h
JesusEV: @akorgor  @heplesser  In the current `compute_gradient()` implementation, there is:
```cpp
sum_grad += grad;
```
I suggest renaming these variables to better reflect their mathematical roles:

`grad` <- `sum_grad`: The actual gradient vector passed to the optimizer.

`grad_contrib` <- `grad`: The individual instantaneous contribution from a specific timestep.

```
grad += grad_contrib;
```

The "gradient" used by the optimizer is the addition of its parts. Encapsulating the "accumulation" notion totally inside `compute_gradient()`  eliminates possible sources of confusion regarding the nature of the "sum" in remote code locations.

models/eprop_readout_bsshslm_2020.cpp
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/e146bb9b86a19109879439fbaf13d15a24c5c38e.

nestkernel/event_delivery_manager.cpp
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/ae2c12b8e6210a8a123b951f56221d21c299a7e8.

nestkernel/activation_event_node.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/435968705704e475e46465725801a2f18bbc0b4e. It makes sense to name it `Mechanism` instead of `Node`. For my understanding, could you please explain why the base `ArchivingNode` and its derived classes as well as the `StructuralPlasticityNode` class are considered nodes, in constrast?

nestkernel/node.h
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3721/commits/1304c34148db703c85eec1bf0003d69401689b5e. In this commit, I reduced duplication and improved consistency between registering connections in the `bsshslm_2020` and the non-`bbshslm_2020` models.

models/eprop_synapse.h
akorgor: Thank you for this valuable comment, @JesusEV. I refactored the naming of the gradient related variables in https://github.com/nest/nest-simulator/pull/3721/commits/cec3e8cca33242aaf4cc1296de090dd9e1a5767c. Note that I deviated slightly from your suggestion (`gradient += gradient_increment`) and I changed `gradients_sum` to `cumulative_gradient` in the `WeightOptimizer`, both in order to further increase clarity.

nestkernel/eprop_archiving_node_impl.h
akorgor: The `set_status` and `get_status` functions are inspired by similar functions in  `ArchivingNode`. Moving their checks into the derived neuron classes would duplicate the same boilerplate across every `::set`, `::get`, `::set_status`, and `::get_status` implementation. It would also force parameters such as `activation_interval_` and `eprop_isi_trace_cutoff_` to be reintroduced separately in each derived model, increasing maintenance overhead and the risk of inconsistencies.

This shared implementation makes the required checks explicit in one place. While it requires distinguishing between neuron models at this level, it also provides a clear overview of which eprop models depend on which constraints, which is harder to see once the logic is scattered across classes.

If you still think these checks should live in the derived classes, could you explain your reasoning so I can better understand it?

models/eprop_iaf.cpp
akorgor: The conversion is done in the `pre_run_hook` and it is assigned to a new variable `V_.RefractoryCounts_` https://github.com/nest/nest-simulator/blob/9dbf52bdd54a3f8bc4c4f9f9246a5ef4c03488b5/models/iaf_psc_delta.cpp#L237-L268

In the meantime, I have introduced the function `get_activation_interval_steps` which only has to be used in the function below `is_activation_event_due`

https://github.com/akorgor/nest-simulator/blob/cec3e8cca33242aaf4cc1296de090dd9e1a5767c/nestkernel/activation_event_mechanism.h#L79-L92

Would it be fine to keep it like this?

models/parrot_neuron.h
akorgor: Given that these events would also have use cases outside eprop, this argues against introducing a separate `eprop_parrot_neuron`. Since as far as I understand the [`get_steps`](https://github.com/akorgor/nest-simulator/blob/b2387c072c36dc2a3ca34952a4f56eea317fc7b3/nestkernel/nest_time.h#L513C1-L528C4) function, it maps values exceeding the limits to meaningful values for the long type, I've set now the limit to `inf` in https://github.com/nest/nest-simulator/pull/3721/commits/1ab2599302c87638fb9fc21c6f5123a5b7672563.

testsuite/pytests/other/test_flush_event_filtering.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
```
akorgor, line 1: @heplesser `test_flush_event_filtering.py` under `testsuite/pytests` seems somewhat out of place given the new `testsuite` organization. Where would be the most appropriate location for this test?

I also noticed that after the reorganization the two e-prop related tests are located in different directories: `synapses/test_eprop_plasticity.py` and `neurons/test_eprop_bsshslm_2020_plasticity.py`. Deciding where they should live is a bit tricky since they involve both neuron and synapse models, but perhaps placing them under `synapses` would make more sense, since the synapse models are the main component being tested.

doc/htmldoc/model_details/flush_event_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- Spike - spike: The standard mechanism is applied.
+
+For e-prop with additional biological features, the mechanism is implemented primarily in the gradient computation function and the cutoff in the e-prop gradient computation requires tracking the remaining steps until cutoff:
+
+- Spike - spike / flush: The remaining steps until cutoff are initialized to the cutoff value. The gradient is reset to 0 and the spike state buffer is set to 1.
+- Spike / flush - spike / flush: The gradient is evaluated (and, if per-step optimization is used, the update weight is calculated) for each time step since the previous event, and the remaining steps until cutoff are decremented accordingly. The computation stops once either all steps up to the current event are processed or the cutoff is reached. If per-step optimization is used, the updated weight is calculated.
+- Spike / flush - spike: If there are remaining steps until the current event, the signal is decayed over those steps. If per-step optimization is not used, the weight is computed from the gradient and applied in the send function of the eprop synapse.
```
jessica-mitchell:
```suggestion
- Spike / flush - spike: If there are remaining steps until the current event, the signal is decayed over those steps. If per-step optimization is not used, the weight is computed from the gradient and applied in the send function of the e-prop synapse.
```

doc/htmldoc/model_details/flush_event_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Therefore, we introduce a dynamic flushing mechanism which alleviates this memory build-up (see the curve labeled "With flush events" in the figure). Each neuron tracks the time of its last spike. If the time since the last spike exceeds a user-defined threshold, the neuron emits a flush event. In each outgoing synapse, a flush event triggers gradient computation and retrieval of the dynamic variable histories from the corresponding postsynaptic neurons required for the weight update. The new weight is computed but not applied. The retrieved portion of the history can then be deleted. The flush event is not sent to the postsynaptic neuron and does not affect the postsynaptic membrane voltage, unlike a spike event. It does not alter the mathematical formulation of the gradient computation, but only evaluates an intermediate gradient value, and therefore does not interfere with the plasticity algorithm.
```
jessica-mitchell:
```suggestion
Therefore, we introduce a dynamic flushing mechanism that alleviates this memory buildup (see the curve labeled "With flush events" in the figure). Each neuron tracks the time of its last spike. If the time since the last spike exceeds a user-defined threshold, the neuron emits a flush event. In each outgoing synapse, a flush event triggers gradient computation and retrieval of the dynamic variable histories from the corresponding postsynaptic neurons required for the weight update. The new weight is computed but not applied. The retrieved portion of the history can then be deleted. The flush event is not sent to the postsynaptic neuron and does not affect the postsynaptic membrane voltage, unlike a spike event. It does not alter the mathematical formulation of the gradient computation, but only evaluates an intermediate gradient value, and therefore does not interfere with the plasticity algorithm.
```

Build-up is not wrong, but changed just to be consistent with other uses

doc/htmldoc/model_details/flush_event_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+
+The details of the flush event mechanism differ between the two e-prop models, and the computations depend on whether the previous and current events are spikes or flush events. For the original e-prop model, this mechanism is implemented entirely in the e-prop synapse's send function. Different event transitions requires different actions:
```
jessica-mitchell:
```suggestion
The details of the flush event mechanism differ between the two e-prop models, and the computations depend on whether the previous and current events are spikes or flush events. For the original e-prop model, this mechanism is implemented entirely in the e-prop synapse's send function. Different event transitions require different actions:
```

doc/htmldoc/model_details/flush_event_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+
+- Flush - flush: No action is required.
+- Spike - flush: Histories are retrieved from the postsynaptic neuron and the gradient is computed.
+- Flush - spike: The weight update is computed and applied, the spike is sent to the postsynaptic neuron, and it is recorded in the update history.
+- Spike - spike: The standard mechanism is applied.
+
+For e-prop with additional biological features, the mechanism is implemented primarily in the gradient computation function and the cutoff in the e-prop gradient computation requires tracking the remaining steps until cutoff:
+
+- Spike - spike / flush: The remaining steps until cutoff are initialized to the cutoff value. The gradient is reset to 0 and the spike state buffer is set to 1.
+- Spike / flush - spike / flush: The gradient is evaluated (and, if per-step optimization is used, the update weight is calculated) for each time step since the previous event, and the remaining steps until cutoff are decremented accordingly. The computation stops once either all steps up to the current event are processed or the cutoff is reached. If per-step optimization is used, the updated weight is calculated.
```
jessica-mitchell:
```suggestion
- Spike / flush - spike / flush: The gradient is evaluated for each time step since the previous event, and the remaining steps until cutoff are decremented accordingly. The computation stops once either all steps up to the current event are processed or the cutoff is reached. If per-step optimization is used, the updated weight is calculated.
```
Redundant text removed

doc/htmldoc/model_details/flush_event_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+
+The details of the flush event mechanism differ between the two e-prop models, and the computations depend on whether the previous and current events are spikes or flush events. For the original e-prop model, this mechanism is implemented entirely in the e-prop synapse's send function. Different event transitions requires different actions:
+
+- Flush - flush: No action is required.
+- Spike - flush: Histories are retrieved from the postsynaptic neuron and the gradient is computed.
+- Flush - spike: The weight update is computed and applied, the spike is sent to the postsynaptic neuron, and it is recorded in the update history.
```
jessica-mitchell:
```suggestion
- Flush - spike: The weight update is computed and applied, the spike is sent to the postsynaptic neuron, and the spike is recorded in the update history.
```
reduce ambiguity

doc/htmldoc/model_details/flush_event_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
... realistic sparse spiking scenarios. As a result, all outgoing synapses of a silent neuron require the corresponding postsynaptic neurons to store the relevant history until it emits a spike. In the original e-prop with fixed-length update intervals, if no presynaptic spike occurs during an interval, the weight update is zero. Thus, the synapse does not require that portion of the postsynaptic history and it can be discarded. However, if even a single presynaptic spike occurs within an interval, the history of that interval must be retained until the next spike arrives. In the e-prop model with additional biological features, weights are updated at every spike and no update intervals exist, so intermediate parts of the history cannot be discarded. Consequently, in both e-prop models, and especially in the latter, a single neuron can cause history to accumulate across many others, increasing memory usage. The buildup may also slow down operations on the history vector, as the algorithm frequently searches for specific entries. Experiments show continuous buildup that exceeds available memory even on supercomputing systems (see the curve labeled "Without flush events" in the figure).
```
jessica-mitchell: "Especially in the latter" is unclear to which model you are referring to 
and "across many others" also is vague and should refer to a specific term

models/eprop_synapse.h
```diff
@@ -67,9 +67,11 @@ For more information on e-prop plasticity, see the documentation on the other e-
  * :doc:`eprop_readout<../models/eprop_readout/>`
  * :doc:`eprop_learning_signal_connection<../models/eprop_learning_signal_connection/>`
 
-For more information on the optimizers, see the documentation of the weight optimizer:
+For more information on the optimizers, see the documentation of the
```
jessica-mitchell: To reduce some of this redundancy "for more information" for every link I would propose
creating a note to collect all these additional thoughts
```
.. note:: 

  For more information, you can find additional documentation on the following topics:

  * other e-prop plasticity models 
        * :doc:`eprop_iaf<../models/eprop_iaf/>`
        * :doc:`eprop_iaf_adapt<../models/eprop_iaf_adapt/>`
        * :doc:`eprop_readout<../models/eprop_readout/>`
        * :doc:`eprop_learning_signal_connection<../models/eprop_learning_signal_connection/>`
  
  * :doc:`weight optimizer <../models/weight_optimizer/>`.
  * triggering synaptic plasticity computations to reduce memory usage with the :ref:`flush event mechanism <flush_event_mechanism>`
```
also note that the syntax is different: you need to specify the filename or label inside the `< >` but the custom text outside that can be whatever you want, so spaces between words is fine.
I recommend using labels for rst files. the reference label is the first line you added to the rst file `.. _flush_event_mechanism:`

models/eprop_synapse_bsshslm_2020.h
```diff
@@ -70,9 +70,11 @@ For more information on e-prop plasticity, see the documentation on the other e-
  * :doc:`eprop_readout_bsshslm_2020<../models/eprop_readout_bsshslm_2020/>`
  * :doc:`eprop_learning_signal_connection_bsshslm_2020<../models/eprop_learning_signal_connection_bsshslm_2020/>`
 
-For more information on the optimizers, see the documentation of the weight optimizer:
+For more information on the optimizers, see the documentation of the
```
jessica-mitchell: see comment on eprop_synapse.h

jessica-mitchell (changes requested):
thanks @akorgor looks good, just a few minor comments :)

doc/htmldoc/model_details/flush_event_mechanism.rst
akorgor: Fixed.

doc/htmldoc/model_details/flush_event_mechanism.rst
akorgor: Thank you for pointing this out. Apparently the brackets made it unclear what I meant here. Hopefully, it is more understandable with the fix.

models/eprop_synapse.h
akorgor: Fixed.

models/eprop_synapse_bsshslm_2020.h
akorgor: Fixed.

nestkernel/eprop_archiving_node_impl.h
akorgor: https://github.com/nest/nest-simulator/pull/3721/commits/641ae11c9e7921995a00bc019049fde1d407382f  refactors `get_status` and `set_status` as discussed by moving these functions into the derived `EpropArchivingNode` classes. https://github.com/nest/nest-simulator/pull/3721/commits/f736d2b7c2beee07ceefffccfd4779912870795f removes the function `history_shift_required` since it is not used anywhere. Note that it is cleaner to leave the function `model_dependent_history_shift` in the derived classes because it uses `hist_shift_required` and moving it into the base class would require to add a second template parameter.

nestkernel/flush_event_mechanism.h
```diff
@@ -110,6 +106,9 @@ class FlushEventMechanism
 
   //! Time of last spike or flush event (steps).
   long last_event_time_;
+
+  //! Interval since previous event after which a flush event is sent (steps).
+  long flush_event_send_interval_steps_;
```
heplesser, line 108: I would place this one directly after `flush_event_send_interval_` for readability.

nestkernel/flush_event_mechanism.cpp
```diff
@@ -35,15 +35,23 @@ namespace nest
 FlushEventMechanism::FlushEventMechanism()
   : flush_event_send_interval_( std::numeric_limits< double >::infinity() )
   , last_event_time_( 0 )
+  , flush_event_send_interval_steps_( std::numeric_limits< long >::max() )
```
heplesser: For consistency, I'd initialize this as
```
flush_event_send_interval_steps_( Time( Time::ms( flush_event_send_interval_ ) ).get_steps() )
```
for consistency and reorder (see comment below).

heplesser:
Benchmark results obtained by @JanVogelsang on JURECA with the hpc_benchmark indicate no loss of performance (one simulation for each of three different seeds for different numbers of compute nodes, 2 MPI processes @ 64 threads per node, weak scaling: problem size scales 32x, runtime only approx 3.5x). Results indicate that there is no performance loss due to introduction of the flush-event mechanism.

image

PR #3722, opened 2026-01-08
E-prop optimizations

akorgor:
This PR primarily introduces optimizations to improve the runtime of the e-prop algorithms. Additionally, it resolves git check errors, updates the reference, and introduces changes to ensure deterministic computations.

Note that all commits prior to  `Replace insert with emplace_back for efficiency` are part of PR #3721, on which this PR builds. Thus, this PR is blocked until #3721 is merged.


models/eprop_iaf.cpp
```diff
@@ -380,61 +397,70 @@ eprop_iaf::compute_gradient( const long t_spike,
...
 
-  if ( not optimize_each_step )
+  const long trace_decay_interval = t_spike - ( previous_event_was_activation ? t_spike_previous : t_compute_until );
+
+  if ( trace_decay_interval > 0 )
   {
-    weight = optimizer->optimized_weight( *ecp.optimizer_cp_, t_compute_until, grad, weight );
+    z_bar *= std::exp( std::log( V_.P_v_m_ ) * trace_decay_interval );
+    e_bar *= std::exp( std::log( P_.kappa_ ) * trace_decay_interval );
+    e_bar_reg *= std::exp( std::log( P_.kappa_reg_ ) * trace_decay_interval );
```
heplesser: You replace here `std::pow(x, n)` with `std::exp( std::log(x) * n)`. I wonder about four things:
- How much faster is this?
- How does this affect accuracy?
- All `x` here are constant during simulation. Wouldn't it make sense to pre-compute the logs?
- `trace_delay_interval` is integer-valued. If the range of expected values is not too large, should one consider to tabulate values?

akorgor: > You replace here `std::pow(x, n)` with `std::exp( std::log(x) * n)`. I wonder about four things:
> 
>  * How much faster is this?
> 
I will provide measurements.
>  * How does this affect accuracy?
> 
It did not yield a measurable deviation in learning performance. All tests also pass.
>  * All `x` here are constant during simulation. Wouldn't it make sense to pre-compute the logs?
> 
Yes, they could be pre-computed.
>  * `trace_delay_interval` is integer-valued. If the range of expected values is not too large, should one consider to tabulate values?

The range of expected values is `[0, activation_interval - eprop_isi_trace_cutoff]`, so a lookup table would be possible. I will test this optimization.

akorgor: The measurements surprisingly did not yield any speed-up anymore. Therefore, I removed this optimization to be on the safe side and for clarity.

heplesser:
@jessica-mitchell This PR almost exclusively changes references to papers and some figures. I have reviewed the small code changes. Could you quickly check that the documentation updates are fine?

akorgor:
> @akorgor I noticed that the see also sections are empty in the model docs; i dont think this is a problem worth blocking a pr, but you may want to remove it (I can take a look later as I thought we had something autogenerating those)

@jessica-mitchell thank you for pointing this out. Actually, I could move the sections listing the other e-prop models there.

akorgor:
> > @akorgor I noticed that the see also sections are empty in the model docs; i dont think this is a problem worth blocking a pr, but you may want to remove it (I can take a look later as I thought we had something autogenerating those)
> 
> @jessica-mitchell thank you for pointing this out. Actually, I could move the sections listing the other e-prop models there.

We just discussed that this issue will be addressed later and not within this PR.

PR #3276, opened 2024-08-06
Add PyNEST example using `ignore_and_fire` model for scaling experiments

jessica-mitchell:
This PR adds an example for the `ignore_and_fire` model 
Thanks to @tomtetzlaff for developing the content!

The documentation to generate the sketch from odg file is removed, as I don't think it's suitable for the purposes of the docs here. 
See Read the Docs output here: https://nest-simulator--3276.org.readthedocs.build/en/3276/auto_examples/ignore_and_fire/index.html

PR #3723, opened 2026-01-08
Introduce generic IgnoreAndSpikeMechanism and add ignore-and-spike support to eprop neurons and iaf_psc_delta 

akorgor:
This PR introduces a flag to enable ignore-and-fire mode for all e-prop neurons.

Note that all commits prior to  `Implement ignore-and-fire neuron models` are part of PR #3721 and #3722, on which this PR builds. Thus, this PR is blocked until #3721 and #3722 are merged. 

- [x] Discuss if `Node` class should inherit this mechanism
- [x] Add documentation page
- [x] Discuss name `paced_spiking_mechanism` vs. `ignore_and_fire_mechanism`
- [x] Abstract ignore-and-fire methods currently implemented in the neuron models into `IgnoreAndFireNode`
- [x] Implement test to check correctness of implementation
- [x] Resolve function name clash with `emit_spike_`
- [x] Add issue to discuss removing / keeping the `ignore_and_fire` model and maybe adapting the example from #3276


tomtetzlaff (changes requested):
Very nice. I think it would be great to add this functionality also to other neuron models. 

Two comments:

1.  I prefer the term "ignore_and_fire" over "paced_spiking", because the former is self-explaining, the latter is not (*):

    a) The name "paced_spiking" doesn't make clear that the input integration dynamics remains fully intact, but is "ignored".

    b) All neuron models are "paced" in the sense that they produce spikes in response to some input. So "paced" doesn't give any hint what this mechanism is about.

    (*)  Some time ago, I tried to explain to a colleague what the "ignore-and-fire neuron" does, but she interrupted me and said: "You don't have to explain it, the name makes it very clear."

2. The `offset` is not set correctly for all cases. The `test_paced_spiking_mechanism.py` fails for all `offset` which are not integer multiples of `interval`, for example, for

   ```
   interval = 10
   offset = 11
   ```

   In particular, the test fails for `offset = 0`.

akorgor:
Thank you for the prompt and helpful review, @tomtetzlaff!

> Very nice. I think it would be great to add this functionality also to other neuron models.

I have added a ToDo point to consider passing on this mechanism to the `Node` class.

>  1. I prefer the term "ignore_and_fire" over "paced_spiking", because the former is self-explaining, the latter is not (*):
>        a) The name "paced_spiking" doesn't make clear that the input integration dynamics remains fully intact, but is "ignored".
>        b) All neuron models are "paced" in the sense that they produce spikes in response to some input. So "paced" doesn't give any hint what this mechanism is about.
>        (*)  Some time ago, I tried to explain to a colleague what the "ignore-and-fire neuron" does, but she interrupted me and said: "You don't have to explain it, the name makes it very clear."

The new name is intended to improve clarity in two ways:

* The NEST codebase generally uses "spike" rather than "fire", except in "integrate-and-fire". Using "spike" keeps the variable names for this mechanism consistent with the surrounding code.
* "Paced spikes" indicates spike emission driven externally at a fixed pace, i.e., at regular intervals.

However, I am open to keeping the old name if you still think it is preferable.

> 2. The `offset` is not set correctly for all cases. The `test_paced_spiking_mechanism.py` fails for all `offset` which are not integer multiples of `interval`, for example, for
>        ```
>        interval = 10
>        offset = 11
>        ```        
>        In particular, the test fails for `offset = 0`.

Fixed in https://github.com/nest/nest-simulator/pull/3723/commits/0e5dca071890e77ac0876db178ab6d20d82c6f9d.

heplesser:
@tomtetzlaff Agnes and I have discussed a bit more about naming and the interval logic.

1. What would you think about "ignore-and-spike" instead of "ignore-and-fire"? That would ensure  consistency between the systematic used of "spiking" instead of "firing" in NEST code and documentation.
2. Concerning the offset, we wonder if it would not make sense to rename this to `time_of_first_spike`, to be given as absolute time point in the simulation, which must be later than "now" at the time of setting it. Then, it would always have to be > 0 (consistent with spike_generator) and, when set during the simulation, in the future. As default, one could use now + interval.

What do you think?

models/eprop_iaf.cpp
```diff
@@ -294,7 +295,7 @@ eprop_iaf::update( Time const& origin, const long from, const long to )
 
     S_.surrogate_gradient_ = ( this->*compute_surrogate_gradient_ )( S_.r_, S_.v_m_, P_.V_th_, P_.beta_, P_.gamma_ );
 
-    if ( S_.v_m_ >= P_.V_th_ and S_.r_ == 0 )
+    if ( update_and_check_spike_emission( S_.v_m_ >= P_.V_th_ and S_.r_ == 0 ) )
```
heplesser:
```suggestion
    if ( dynamic_spike_valid_or_paced_spike_due( S_.v_m_ >= P_.V_th_ and S_.r_ == 0 ) )
```

tomtetzlaff:
> @tomtetzlaff Agnes and I have discussed a bit more about naming and the interval logic.
> 
>     1. What would you think about "ignore-and-spike" instead of "ignore-and-fire"? That would ensure  consistency between the systematic used of "spiking" instead of "firing" in NEST code and documentation.
> 
>     2. Concerning the offset, we wonder if it would not make sense to rename this to `time_of_first_spike`, to be given as absolute time point in the simulation, which must be later than "now" at the time of setting it. Then, it would always have to be > 0 (consistent with spike_generator) and, when set during the simulation, in the future. As default, one could use now + interval.
> 
> 
> What do you think?

1. "ias" -- nice ;). Yes, "ignore-and-spike" is totally fine with me. 
   
2. I agree. "time_of_first_spike" is somewhat lengthy, but it makes the meaning clearer.

tomtetzlaff:
> > 2. The `offset` is not set correctly for all cases. The `test_paced_spiking_mechanism.py` fails for all `offset` which are not integer multiples of `interval`, for example, for
> >    `interval = 10 offset = 11`
> >    In particular, the test fails for `offset = 0`.
> 
> Fixed in [0e5dca0](https://github.com/nest/nest-simulator/commit/0e5dca071890e77ac0876db178ab6d20d82c6f9d).

Great. Thanks @akorgor. I have tested this again and can confirm that it behaves as expected now.

tomtetzlaff (approved):

akorgor:
A documentation page now informs about the purpose and functioning of the mechanism (https://github.com/nest/nest-simulator/pull/3723/commits/044c7d45db4134c8b1d7b6377173b5d90126d43b). Additionally, `ArchivingNode` now inherits from `IgnoreAndSpikeMechanism`, so in principle any neuron model can use the mechanism by adjusting the spike emission condition; as an example, this has been applied to the `iaf_psc_delta` model (https://github.com/nest/nest-simulator/pull/3723/commits/6dffd800465744a88ba3a8afaebeccd6e8ce6769). The only remaining decisions, which can be addressed in a separate issue, is whether to adjust the spike emission conditions also in other neuron models and whether to remove the `ignore_and_fire` neuron model.

models/eprop_iaf.cpp
akorgor: Fixed in https://github.com/nest/nest-simulator/pull/3723/commits/6c006e35bc5ba119db0c39c4e3df703741afe7f6. Note that I chose `spike_event_is_due` for brevity and consistency with `flush_event_is_due`.

akorgor:
@heplesser This benchmark shows that the ignore-and-spike mechanism does not negatively affect the runtime for `iaf_psc_delta` measured over 10 random seeds used for the input spike generation:

runtime_comparison_ignore-and-fire_main

The benchmark was run on JURECA on the `dc-cpu` partition:
```python
config = {
  "indegree_in": 100,
  "indegree_out": 1000,
  "indegree_rec": 10000,
  "job_cpus_per_task": 32,
  "job_exclusive": true,
  "job_nodes": 1,
  "job_ntasks_per_node": 4,
  "n_in_base": 1000,
  "n_out_base": 10,
  "n_rec_base": 128000,
  "outdegree_fb": 100,
  "plasticity": "static",
  "rate_in": 5.0,
  "rate_rec": 5.0,
  "steps": 6000,
}
 ```

**Comment by akorgor** [comment]
@jessica-mitchell Could you please review `doc/htmldoc/model_details/ignore_and_spike_mechanism.rst`?

File: doc/htmldoc/model_details/ignore_and_spike_mechanism.rst [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+
+On creation, the first forced spike is scheduled relative to the current simulation time using the offset. Subsequent spikes follow at the given interval. If the parameters are changed later, the mechanism restarts from the current simulation time using the new offset and interval. If the parameters remain unchanged across consecutive ``Simulate()`` calls, the spike train continues from the previous forced spike using the interval, i.e., the offset is not applied again. Thus, the offset defines the first spike after creation or parameter change, while the interval defines all following spikes until the parameters change.
+
+Let :math:`t_0` denote the current simulation time when the mechanism is created or its parameters changed, :math:`o` the offset, and :math:`\Delta t` the interval. The forced spike times are then
```
heplesser:
```suggestion
Let :math:`t_0` denote the simulation time at which a neuron using the mechanism is created or its parameters changed, :math:`o` the offset, and :math:`\Delta t` the interval. The forced spike times are then
```

models/eprop_iaf.cpp
heplesser:
```suggestion
  double& /* epsilon */,
```

nestkernel/archiving_node.cpp
```diff
@@ -32,7 +32,8 @@ namespace nest
 // member functions for ArchivingNode
 
 nest::ArchivingNode::ArchivingNode()
-  : n_incoming_( 0 )
+  : IgnoreAndSpikeMechanism()
```
heplesser: For consistency:

```suggestion
  : StructuralPlasticityNode()
  , IgnoreAndSpikeMechanism()
```

nestkernel/flush_event_mechanism.cpp
```diff
@@ -59,15 +59,27 @@ FlushEventMechanism::get_status( Dictionary& d ) const
...
-      "Interval since previous event after which a flush event is sent flush_event_send_interval > 0 required." );
+    throw BadProperty( "flush_event_send_interval > 0 required." );
   }
+
+  if ( check_eprop_constraint )
+  {
+    const double eprop_update_interval = kernel().simulation_manager.get_eprop_update_interval().get_ms();
+    if ( flush_event_send_interval_tmp < eprop_update_interval )
+    {
+      throw BadProperty( "flush_event_send_interval >= eprop_update_interval required." );
```
heplesser:
```suggestion
      throw BadProperty( "flush_event_send_interval ≥ eprop_update_interval required." );
```

nestkernel/ignore_and_spike_mechanism.h
```diff
@@ -0,0 +1,125 @@
...
+// nestkernel
+#include "nest_time.h"
+
+// libnestutil
+#include "dict_util.h"
+
+namespace nest
+{
+/**
+ * @brief Class implementing an ignore-and-fire mechanism for neuron models.
```
heplesser, line 35:
```suggestion
 * @brief Class implementing an ignore-and-fire mechanism for neuron models.
 * 
 * @note See model_details/ignore_and_spike_mechanism.rst for details.
```

heplesser (changes requested):
Looks good, just a few tiny details, see suggestions.

models/eprop_iaf.cpp
akorgor: This is fixed in PR #3820. To not create a merge-conflict, I would drop this change here.

heplesser (approved):
Thanks, good to go now from my side!

jessica-mitchell:
@akorgor can you add the link to the example in  `doc/htmldoc/examples/index.rst` I am not sure where it best fits

akorgor:
> @akorgor can you add the link to the example in `doc/htmldoc/examples/index.rst` I am not sure where it best fits

Fixed in https://github.com/nest/nest-simulator/pull/3723/commits/849ea46d4c7619a20d647f1bcd34ad7797e35397.

doc/htmldoc/examples/index.rst
```diff
@@ -266,6 +266,7 @@ PyNEST examples
            :img-top: ../auto_examples/hpc_benchmark_connectivity.svg
 
            * :doc:`../auto_examples/hpc_benchmark`
+           * :doc:`../auto_examples/ignore_and_spike`
```
jessica-mitchell:
```suggestion
           * :doc:`../auto_examples/ignore_and_spike_mechanism`
```

doc/htmldoc/examples/index.rst
jessica-mitchell, line 290:
```suggestion
   ../auto_examples/pong/index
   ../auto_examples/ignore_and_spike_mechanism
```

jessica-mitchell (changes requested):

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Thus, the first forced spike occurs at :math:`t_0 + o`, and each following forced spike satisfies
+
+.. math::
+
+   t_{k+1} - t_k = \Delta t.
+
+Currently, the following models are endowed with this mechanism:
+
+- `eprop_iaf_adapt_bsshslm_2020.cpp`
```
jessica-mitchell:
```suggestion
- ``eprop_iaf_adapt_bsshslm_2020.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- `eprop_iaf_adapt.cpp`
```
jessica-mitchell:
```suggestion
- ``eprop_iaf_adapt.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- `eprop_iaf_bsshslm_2020.cpp`
```
jessica-mitchell:
```suggestion
- ``eprop_iaf_bsshslm_2020.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- `eprop_iaf_psc_delta_adapt.cpp`
```
jessica-mitchell:
```suggestion
- ``eprop_iaf_psc_delta_adapt.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- `eprop_iaf_psc_delta.cpp`
```
jessica-mitchell:
```suggestion
- ``eprop_iaf_psc_delta.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- `eprop_iaf.cpp`
```
jessica-mitchell:
```suggestion
- ``eprop_iaf.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+- `iaf_psc_delta.cpp`
```
jessica-mitchell:
```suggestion
- ``iaf_psc_delta.cpp``
```

doc/htmldoc/model_details/ignore_and_spike_mechanism.rst
```diff
@@ -0,0 +1,32 @@
...
+
+For developers: All `Node`s that inherit from `ArchivingNode` automatically inherit the `IgnoreAndSpike` class. To enable this functionality for a model that is not listed above, wrap the spike-emission condition in `spike_event_is_due()`.
```
jessica-mitchell:
```suggestion
For developers: All Node`s that inherit from ``ArchivingNode`` automatically inherit the``IgnoreAndSpike`` class. To enable this functionality for a model that is not listed above, wrap the spike-emission condition in ``spike_event_is_due()``.
```

jessica-mitchell (approved):
I will approve, but there are some minor formatting that should be changed, otherwise looks good!

tomtetzlaff (approved):
This looks very good to me. Thanks a lot. 

A small comment on the wording in the documentation:

"The ignore-and-spike mechanism supports this by ignoring spikes from the internal dynamics and generating spikes at defined intervals."

How about changing this to:

"The ignore-and-spike mechanism supports this by ignoring the internal dynamics and generating spikes at defined intervals. The input integration and internal dynamics hence remain intact, but have no effect on the output spikes."

The wording "by ignoring spikes from the internal dynamics" may lead to misunderstandings. The user may think that input spikes are not processed at all, which is not the case, right? For the output spikes, this doesn't make a difference. It is important, however, for performance benchmarking.

And how about adding a crosslink in the documentation to 

https://nest-simulator.readthedocs.io/en/stable/models/ignore_and_fire.html 

and/or 

https://nestml.readthedocs.io/en/latest/models_library/ignore_and_fire_neuron.html, 

just to make clear that this is basically the same thing?

akorgor:
> A small comment on the wording in the documentation:
> 
> "The ignore-and-spike mechanism supports this by ignoring spikes from the internal dynamics and generating spikes at defined intervals."
> 
> How about changing this to:
> 
> "The ignore-and-spike mechanism supports this by ignoring the internal dynamics and generating spikes at defined intervals. The input integration and internal dynamics hence remain intact, but have no effect on the output spikes."
> 
> The wording "by ignoring spikes from the internal dynamics" may lead to misunderstandings. The user may think that input spikes are not processed at all, which is not the case, right? For the output spikes, this doesn't make a difference. It is important, however, for performance benchmarking.
> 

Incorporated in https://github.com/nest/nest-simulator/pull/3723/commits/ec14275bc1edac19284f922322906a65010f28a6.

> And how about adding a crosslink in the documentation to
> 
> https://nest-simulator.readthedocs.io/en/stable/models/ignore_and_fire.html
> 
> and/or
> 
> https://nestml.readthedocs.io/en/latest/models_library/ignore_and_fire_neuron.html,
> 
> just to make clear that this is basically the same thing?

I have now created issue #3867 as a reminder to discuss whether to keep or remove the ignore-and-fire neuron model and, in the former case, to add the crosslinks.

PR #3727, opened 2026-01-12
New feature: cycle timers

alpinangu:
Logs information about the duration and spike counts within each update cycle to the KernelStatus to support temporal performance profiling. 

**CMake option:** `-Dwith-cycle-timers=ON/OFF`

requires detailed timers to be enabled

**Accsess pattern:**

_d = nest.GetKernelStatus()_ 

`d['cycle_time_log']['time_update'] `: Total duration of each individual update cycles
 
`d['cycle_time_log']['time_communicate_spike_data']` : Time spent in MPI spike-data communication within each individual update cycles

`d['cycle_time_log']['local_spike_counter']` : Number of generated spike events (both off- and on-grid) within each individual update cycles

### DIFF
--- a/cmake/ProcessOptions.cmake
+++ b/cmake/ProcessOptions.cmake
@@ -434,6 +434,9 @@ function(NEST_PROCESS_WITH_CYCLE_TIMERS)
   if ("${with-detailed-timers}" STREQUAL "ON" AND "${with-cycle-timers}"   STREQUAL "ON")
     set(CYCLE_TIMERS ON PARENT_SCOPE)
   endif()
+  if ("${with-cycle-timers}" STREQUAL "ON" AND NOT "${with-detailed-timers}" STREQUAL "ON")
+    message(FATAL_ERROR "To enable cycle timers, you must also enable detailed timers")
+  endif()
 endfunction()
 
 


nestkernel/cycle_time_log.h
```diff
@@ -0,0 +1,54 @@
...
+public:
+  CycleTimeLog();
+  void clear();
+  void add_entry( double cycle_update_time, double communicate_time, long local_spike_counter );
+  void to_dict( DictionaryDatum& ) const;
+
+private:
+  std::vector< double > cycle_update_time_;  //!< Time of one update cycle
+  std::vector< double > communicate_time_;   // Time of communicate in current cycle
+  std::vector< long > local_spike_counter_;  // Local spike count in current cycle
```
JanVogelsang: Just a minor change, but rather make this a `size_t` or `unsigned long`, as the counter can never be negative.
```suggestion
  std::vector< size_t > local_spike_counter_;  // Local spike count in current cycle
```

nestkernel/event_delivery_manager.cpp
```diff
@@ -1084,4 +1084,18 @@ nest::EventDeliveryManager::distribute_target_data_buffers_( const size_t tid )
   return are_others_completed;
 }
 
+#ifdef CYCLE_TIMERS
+double
+EventDeliveryManager::get_sw_communicate_spike_data() const
+{
+  return sw_communicate_spike_data_.elapsed();
+}
```
JanVogelsang: The name doesn't match the return type. It should either return the stopwatch itself (which I would prefer here) or be called something like `get_communicate_spike_data_elapsed`.


```suggestion
Stopwatch<StopwatchGranularity::Detailed, StopwatchParallelism::MasterOnly>
EventDeliveryManager::get_sw_communicate_spike_data() const
{
  return sw_communicate_spike_data_;
}
```

nestkernel/event_delivery_manager.cpp
```diff
@@ -1084,4 +1084,18 @@ nest::EventDeliveryManager::distribute_target_data_buffers_( const size_t tid )
...
+
+unsigned long
+EventDeliveryManager::get_local_spike_counter() const
+{
+  return std::accumulate( local_spike_counter_.begin(), local_spike_counter_.end(), 0UL );
```
JanVogelsang: `0UL` is an unsigned 0, while the type is currently declared as `long` (see comment above). Just change the type above. And you can actually drop the `UL` here, just a simple 0 suffices and is more readable.


```suggestion
  return std::accumulate( local_spike_counter_.begin(), local_spike_counter_.end(), 0 );
```

nestkernel/simulation_manager.cpp
```diff
@@ -491,6 +494,12 @@ nest::SimulationManager::get_status( Dictionary& d )
   d[ names::eprop_update_interval ] = eprop_update_interval_;
   d[ names::eprop_learning_window ] = eprop_learning_window_;
   d[ names::eprop_reset_neurons_on_update ] = eprop_reset_neurons_on_update_;
+
+#ifdef CYCLE_TIMERS
+  DictionaryDatum log_events = DictionaryDatum( new Dictionary );
+  ( *d )[ names::cycle_time_log ] = log_events;
```
JanVogelsang:
```suggestion
  Dictionary log_events;
  d[ names::cycle_time_log ] = log_events;
```

nestkernel/simulation_manager.cpp
```diff
@@ -843,6 +852,11 @@ nest::SimulationManager::update_()
 
   std::vector< std::exception_ptr > exceptions_raised( kernel().vp_manager.get_num_threads() );
 
+#ifdef CYCLE_TIMERS
+  double start_current_communicate = kernel().event_delivery_manager.get_sw_communicate_spike_data();
+  long start_local_spike_counter = kernel().event_delivery_manager.get_local_spike_counter();
```
JanVogelsang: See comment above.
```suggestion
  size_t start_local_spike_counter = kernel().event_delivery_manager.get_local_spike_counter();
```

nestkernel/simulation_manager.cpp
```diff
@@ -1104,6 +1118,20 @@ nest::SimulationManager::update_()
...
           update_time_limit_exceeded = update_time > update_time_limit_;
+
+#ifdef CYCLE_TIMERS
+
+          const double end_current_communicate = kernel().event_delivery_manager.get_sw_communicate_spike_data();
+          const double communicate_time = end_current_communicate - start_current_communicate;
+          start_current_communicate = end_current_communicate;
+
+          long end_local_spike_counter = kernel().event_delivery_manager.get_local_spike_counter();
+          const long local_spike_counter = end_local_spike_counter - start_local_spike_counter;
```
JanVogelsang: See comment above.
```suggestion
          const size_t local_spike_counter = end_local_spike_counter - start_local_spike_counter;
```

nestkernel/simulation_manager.cpp
```diff
@@ -1104,6 +1118,20 @@ nest::SimulationManager::update_()
...
           // after the master section.
           update_time_limit_exceeded = update_time > update_time_limit_;
+
+#ifdef CYCLE_TIMERS
+
+          const double end_current_communicate = kernel().event_delivery_manager.get_sw_communicate_spike_data();
+          const double communicate_time = end_current_communicate - start_current_communicate;
+          start_current_communicate = end_current_communicate;
+
+          long end_local_spike_counter = kernel().event_delivery_manager.get_local_spike_counter();
```
JanVogelsang: See comment above.
```suggestion
          const size_t end_local_spike_counter = kernel().event_delivery_manager.get_local_spike_counter();
```

nestkernel/simulation_manager.h
```diff
@@ -236,6 +237,8 @@ class SimulationManager : public ManagerInterface
   Stopwatch< StopwatchGranularity::Detailed, StopwatchParallelism::Threaded > sw_deliver_spike_data_;
   Stopwatch< StopwatchGranularity::Detailed, StopwatchParallelism::Threaded > sw_deliver_secondary_data_;
 
+  CycleTimeLog cycle_time_log_;
```
JanVogelsang, line 241: It should be possible to just remove this field when not needed right?
```suggestion
#ifdef CYCLE_TIMERS
  CycleTimeLog cycle_time_log_;
#endif
```

nestkernel/simulation_manager.cpp
```diff
@@ -106,6 +107,8 @@ nest::SimulationManager::initialize( const bool adjust_number_of_threads_or_rng_
 
   reset_timers_for_preparation();
   reset_timers_for_dynamics();
+
+  cycle_time_log_.clear();
```
JanVogelsang, line 111:
```suggestion
#ifdef CYCLE_TIMERS
  cycle_time_log_.clear();
#endif
```

nestkernel/simulation_manager.cpp
```diff
@@ -65,6 +65,7 @@ nest::SimulationManager::SimulationManager()
   , eprop_update_interval_( 1000. )
   , eprop_learning_window_( 1000. )
   , eprop_reset_neurons_on_update_( true )
+  , cycle_time_log_()
```
JanVogelsang: I think you can just remove this line. It definitely fails when compiling without cycle timers, but I think it is not strictly required anyway.

PR #3730, opened 2026-01-19
Pin libNeurosim/CSA version in CI install

terhorstd:
Dependencies [libneurosim](https://github.com/INCF/libneurosim) and [CSA](https://github.com/INCF/csa) are installed from development branches. In libCSA development a header file was changed which breaks NEST CI. This PR fixes the CI build by correcting the install source to last released version.


doc/htmldoc/_ext/versions.json
```diff
@@ -23,6 +23,12 @@
 	"libtool": {
 		"min": "2.4.6"
 	},
+	"libneurosim": {
+		"min": "1.2.0"
+	},
+	"csa": {
+		"max": "0.1.13"
```
heplesser, line 31: Versions before 0.1.13 will not work with Python 3.14, so we also need a minimum
```suggestion
		"min": "0.1.13",
		"max": "0.1.13"
```

heplesser (changes requested):
@terhorstd Thanks for the quick fix! I have just one small suggestion, see below.

And then I wonder how we can remind ourselves to from time to time check versions pinnings and update them where that makes sense.

PR #3735, opened 2026-02-09
Add pandoc to the environment dependencies

akorgor:
Building the user documentation with the provided conda environment recipes fails due to a missing pandoc dependency on a clean setup. This PR adds pandoc to the conda environment file.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -41,6 +41,7 @@ dependencies:
   - cppcheck
   - doxygen
   - graphviz
+  - pandoc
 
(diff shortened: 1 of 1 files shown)


jessica-mitchell:
Hi @akorgor pandoc is a bit of a tricky situation because you can't use just any version -- you need the system install with apt package - the instructions for [building the docs](https://nest-simulator.readthedocs.io/en/latest/developer_space/workflows/documentation_workflow/user_documentation_workflow.html#set-up-your-environment) mention this too

PR #3738, opened 2026-02-11
Align behavior of random parameters with numpy.random.

### LINKED ISSUE #3736: Permit std=0 in nest.random.normal()
Calling `nest.random.normal(mean, std)` with `std=0` raises an error:

```
NESTErrors.BadProperty: BadProperty in SLI function CreateParameter_D: nest::NormalParameter: std > 0 required.
```

This is slightly uncomfortable, for example, when comparing homogeneous and heterogeneous neuron or connection parameters (e.g., delays).

For comparison: `numpy.random.normal(mean, 0)` deterministically returns `mean` without complaints.

heplesser:
This PR aligns the non-spatial normal, lognormal and exponential parameters in NEST by allowing `std == 0` or `beta == 0`, respectively. For the exponential parameter, a test against negative beta is introduced.

This PR only modifies the plain random parameters. For spatial random parameters, we still require that parameters are strictly positive.

This fixes #3736.


tomtetzlaff (approved):
Thanks a lot, @heplesser, for this fast fix. All three functions `nest.random.normal()`, `nest.random.lognormal()` and `nest.random.exponential()` behave as expected now.

testsuite/pytests/regressions/test_ticket_800.py
```diff
@@ -52,11 +52,9 @@
 bad_random_parameters = [
     ("uniform", {"min": 0.0, "max": 0.0}),
     ("normal", {"mean": 0.0, "sigma": 1.0}),  # bad param name
-    ("normal", {"mean": 0.0, "std": 0.0}),
     ("normal", {"mean": 0.0, "std": -1.0}),
-    ("lognormal", {"mu": 0.0, "std": 0.0}),
     ("lognormal", {"mu": 0.0, "std": -1.0}),
-    ("lognormal", {"mu": 0.0, "std": 1.0}),
+    ("exponential", {"beta": -1.0}),
 ]
```
terhorstd, line 60: Are acceptable parameter values also tested? I'd have expected also a `good_random_parameters = […, ("normal", {"mean": 1.0, "std": 0.0}, 1.0)]`

terhorstd (commented):
Looks like a good change. For consistency we could briefly check testing strategy for the changed behavior covers also new cases.

testsuite/pytests/regressions/test_ticket_800.py
terhorstd: After considering to decouple this into a separate issue, it seems too directly related, so I add to this comment:

In this change of aligning random parameters to numpy behavior formerly `throw`ing tests are removed. However, no corresponding tests of the – now possible – good behavior are introduced.

Currently the new expected behavior is not tested. Arguments that other statistical tests cover this implicitly are invalid, as
* the other tests did not run into `std==0` cases before (otherwise they would have `thrown`, and would now not throw anymore).
* any `catch` behavior outside the changed functionality that defines values may already include branches for `std==0` case, causing the new functionality not to be used.

Additionally the `std=0` parameter would explicit test that the exact mean value is returned
```
good_random_parameters = [
    …,
    ("normal", {"mean": 0.0, "std": 0.0}, 0.0),
    ("normal", {"mean": 1.0, "std": 0.0}, 1.0),
]
```
Testing this for different `mean` values makes sure
* the new functionality *does actually work*, and ensures that the code does not bypass calculations returning default values.
* returned values are *exact* and internal calculations don't cause float bit-differences, i.e. `mean = 0.1` → `result == 0.1` without rounding errors.

terhorstd (changes requested):

testsuite/pytests/regressions/test_ticket_800.py
heplesser: I have now added tests to check correct behavior for std==0.

terhorstd: That looks amazingly better :+1:

terhorstd (approved):
Thanks for digging so deeply into the standard library definitions. This is a nicely instructive PR!
:+1:

PR #3742, opened 2026-02-24
Update N-MNIST download URL and robustify download method in e-prop tutorial

akorgor:
This PR fixes the failing N-MNIST dataset download by updating the URL and improves robustness by streaming and chunking the download.


jessica-mitchell:
@akorgor thanks for the PR, I am just testing this on EBRAINS and it looks like the link works as expected now, but the kernel gets killed because memory requirements for this example exceed 2048 Mb (the max on EBRAINS Jupyter hub)
Is it possible to scale this example down in to reduce memory usage so it can be used on EBRAINS?

PR #3748, opened 2026-03-03
Move from boost::any to std::variant 

JanVogelsang:
`boost::any` allocates memory on the heap for every object while `std::variant` only allocates space on the stack. Further, `std::variant` performs all type-checking at compile-time rather than during runtime.

### DIFF
--- a/libnestutil/dictionary.h
+++ b/libnestutil/dictionary.h
@@ -76,8 +76,8 @@ struct EmptyList;
  * information for a @ref NodeCollection is returned.
  *
  * @note Conversion from @ref Dictionary elements to and from Python datatypes is done by
- * @ref nestkernel_api.pyx:any_to_pyobj and @ref nestkernel_api.pyx:pydict_to_Dictionary, respectively. If you add a type
- * to the variant, you most likely need to add corresponding conversions to those functions.
+ * @c nestkernel_api.pyx:any_to_pyobj and @c nestkernel_api.pyx:pydict_to_Dictionary, respectively. If you add a
+ * type to the variant, you most likely need to add corresponding conversions to those functions.
  *
  * @ingroup nestdict
  */


libnestutil/numerics.cpp
```diff
@@ -164,8 +164,8 @@ mod_inverse( long a, long m )
   assert( 0 < a );
   assert( 2 <= m );
 
-  const long a_orig = a;
   const long m_orig = m;
+  const long a_orig = a;
```
heplesser: Why not keep the old, alphabetical order

```suggestion
  const long a_orig = a;
  const long m_orig = m;
```

nestkernel/node_collection.h
```diff
@@ -520,9 +517,11 @@ class nc_const_iterator
    */
   size_t get_step_size() const;
 
-  void print_me( std::ostream& ) const;
+  friend std::ostream& operator<<( std::ostream& os, const nc_const_iterator& e );
 };
 
+std::ostream& operator<<( std::ostream& os, const nc_const_iterator& e );
+
```
akorgor: For consistency:
```suggestion
  friend std::ostream& operator<<( std::ostream& out, const nc_const_iterator& nc );
};

std::ostream& operator<<( std::ostream& out, const nc_const_iterator& nc );

```

nestkernel/parameter.h
```diff
@@ -74,7 +72,13 @@ class Parameter
    * @param node pointer to the node, used when the node position is relevant
    * @returns the value of the parameter.
    */
-  virtual double value( RngPtr rng, Node* node ) = 0;
+  // TODO-PYNEST-NG: We need to either make this non-abstract or add every single non-abstract parameter subclass to the
+  //  std::variant acceptec-types collection.
+  virtual double
+  value( RngPtr rng, Node* node )
```
akorgor: To resolve the github warnings:
```suggestion
  value( RngPtr, Node* )
```

nestkernel/sonata_connector.cpp
```diff
@@ -548,9 +547,7 @@ SonataConnector::create_edge_type_id_2_syn_spec_( Dictionary edge_params )
   for ( const auto& [ syn_k, syn_v ] : edge_params )
```
akorgor, line 547:
```suggestion
  for ( const auto& syn_kv_pair : edge_params )
```

pynest/nestkernel_api.pyx
```diff
@@ -396,7 +390,7 @@ def llapi_distance(object conn):  # PYNEST-NG-FUTURE: should there be a SynapseC
 
 def llapi_make_nodecollection(object node_ids):
     cdef NodeCollectionPTR gids
-    # node_ids list is automatically converted to an std::vector
+    # node_ids list is automatically converted to an vector
```
akorgor:
```suggestion
    # node_ids list is automatically converted to a vector
```

CMakeLists.txt
```diff
@@ -24,9 +24,9 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 
 project( nest CXX C )
 
-set( CMAKE_CXX_STANDARD 17 )
+set( CMAKE_CXX_STANDARD 20 )
 set( CMAKE_CXX_STANDARD_REQUIRED True )
-set( CMAKE_CXX_EXTENSIONS OFF )          # -std=c++17 instead of gnu++17
+set( CMAKE_CXX_EXTENSIONS OFF )          # -std=c++20 instead of gnu++17
```
akorgor: Note that this is based on https://github.com/nest/nest-simulator/pull/3748/changes/9301168782ccf9f1020ee4e1261781dbcc692fe2 but there exists a very similar commit https://github.com/nest/nest-simulator/pull/3749/commits/230f36eefc90aeb05b945311da2ef9795a8583fd. The latter also corrects gnu++17 to gnu++20.

libnestutil/dictionary.h
```diff
@@ -23,73 +23,183 @@
...
+  }
+  bool operator==( const EmptyList& ) const = default;
+};
+
+template < typename... Scalars >
+struct DictionarySchemaBuilder
+{
+  template < typename T >
+  static constexpr bool is_defined_scalar = ( std::is_same_v< T, Scalars > or ... );
+  ;
```
akorgor:
```suggestion
```

libnestutil/dictionary.h
```diff
@@ -216,26 +290,22 @@ class Dictionary : public std::map< std::string, DictEntry_ >
   /**
    * Return reference to vector of type T stored under key.
    *
-   * If key does not exist in dict, create empty vector<T> and return it.
+   * * If key does not exist in dict, create empty vector<T> and return it.
```
akorgor:
```suggestion
   * If key does not exist in dict, create empty vector<T> and return it.
```

libnestutil/dictionary.h
```diff
@@ -216,26 +290,22 @@ class Dictionary : public std::map< std::string, DictEntry_ >
...
-      // to create a new element would result in an empty boost::any, not an
-      // empty vector<T>.
-      ( *this )[ key ] = std::vector< T >();
-    }
-    return boost::any_cast< std::vector< T >& >( ( *this )[ key ] );
+    // try_emplace will only insert if key doesn't exist.
+    auto [ iter, success ] = maptype_::try_emplace( key, std::vector< T >() );
+
+    return std::get< std::vector< T > >( iter->second.item );
   };
```
akorgor:
```suggestion
  }
```

nestkernel/sonata_connector.cpp
```diff
@@ -548,9 +547,7 @@ SonataConnector::create_edge_type_id_2_syn_spec_( Dictionary edge_params )
...
   {
     const int type_id = std::stoi( syn_k );
-
-    // PYNEST-NG-FUTURE: Could avoid explicit any_cast here by adding extraction method to DictEntry_
-    const auto& d = boost::any_cast< Dictionary >( syn_v.item );
+    auto d = std::get< Dictionary >( syn_kv_pair.second.item );
 
     const auto& syn_name = d.get< std::string >( "synapse_model" );
```
akorgor, line 552: Here CI reports the following check failure:
```
expected primary-expression before ‘>’ token
```

akorgor (commented):
Looks good to me overall. I unfortunately could only do a shallow review due to my lack of expertise on these advanced type concepts. For me the following test fails:
```
.share.nest.testsuite.pytests.sli2py_stimulating.test_iaf_ps_psp_poisson_generator_accuracy
```

heplesser:
@JanVogelsang Working on resolving the conflicts now.

libnestutil/dictionary.cpp
```diff
@@ -171,133 +232,76 @@ operator<<( std::ostream& os, const Dictionary& dict )
...
 {
   // Compiler will short-circuit to false if not same type
-  return is_type< T >( first ) and is_type< T >( second )
-    and boost::any_cast< T >( first ) == boost::any_cast< T >( second );
+  return is_type< T >( first ) and is_type< T >( second ) and std::get< T >( first ) == std::get< T >( second );
 }
 
 bool
-value_equal( const boost::any& first, const boost::any& second )
+value_equal( const any_type& first, const any_type& second )
```
heplesser: Will not work correctly if first and second have different types.

libnestutil/dictionary.cpp
```diff
@@ -171,133 +232,76 @@ operator<<( std::ostream& os, const Dictionary& dict )
...
     or equal_as_< std::string >( first, second )
-    or equal_as_< std::vector< int > >( first, second )
+    or equal_as_< std::vector< long > >( first, second )
     or equal_as_< std::vector< double > >( first, second )
-    or equal_as_< std::vector< size_t > >( first, second )
     or equal_as_< std::vector< std::string > >( first, second )
     or equal_as_< std::vector< std::vector< double > > >( first, second )
     or equal_as_< Dictionary >( first, second )
     or equal_as_< std::shared_ptr< nest::Parameter > >( first, second )
+    or fail( first, second )
```
heplesser: Why does this never trigger the exception? Don't we get here whenever to dicts are unequal?

libnestutil/dictionary.cpp
heplesser: Actually, this entire mechanism (which I moved here from the boost::any-based code) is not needed, see below.

libnestutil/dictionary.cpp
heplesser: As C++17 natively provides `operator==()` for variants with just the logic I tried to implement here, we do not need this incorrect code here. I will remove it with the next iteration.

nestkernel/sonata_connector.cpp
heplesser: This is fixed.

nestkernel/node_collection.h
heplesser: You are right that  `e` doesn't fit here, replacing with `nci`.

nestkernel/sonata_connector.cpp
heplesser: I don't understand your suggestion. The following code is based on the `[syn_k, syn_v]` unpacking and I think it makes for cleaner code, because one does not need to write `syn_kv_pair.first` and `syn_kv_pair.second` in the loop.

akorgor: Currently `syn_kv_pair` in line 550 is undeclared and I proposed one of the two possible fix options. I agree that `[syn_k, syn_v]` would be cleaner., so, I would suggest to change it in l. 550 to `syn_v` instead.

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+/**
+ * Dictionary class for interface to Python and C++ API.
+ *
+ * Only this class should be used in code outside @c dictionary.{h,cpp}.
+ *
+ * Methods that behave as in @c std::map are not documented explicitly.
+ *
+ * @ingroup nestdict
+ */
+class Dictionary : public std::shared_ptr< dictionary_ >
```
akorgor, line 210: Maybe this suggestion belongs rather in a separate issue than in this PR.

The current design inherits from `std::shared_ptr<dictionary_>`, which makes `Dictionary` behave like a smart pointer rather than a dictionary. I found the code `( *this )->` in the wrappers confusing, as the origin of the `->` operator is not immediately clear - it is inherited from the smart pointer. Therefore, I would suggest using composition instead, i.e., storing a `std::shared_ptr` as a member.

Given that `dictionary_` is already exposed through the implicit conversion on line 214, it could instead be exposed explicitly (e.g. via `impl()`), which would allow many of the forwarding wrappers to be removed. Calls would then become explicit, e.g.:

```cpp
dict.impl().size();
dict.impl().empty();
dict.impl().find( "key" );
```

Additionally, the code would be easier to understand if `dictionary_` were renamed to `DictionaryImpl`. Taken together:

```cpp
class DictionaryImpl;

class Dictionary
{
public:
  Dictionary()
    : impl_( std::make_shared< DictionaryImpl >() )
  {}

  DictionaryImpl& impl() { return *impl_; }
  const DictionaryImpl& impl() const { return *impl_; }

private:
  std::shared_ptr< DictionaryImpl > impl_;
};
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+   */
+  bool operator==( const Dictionary& other ) const;
+
+
+  auto begin() const;
+  auto end() const;
+
+  auto size() const;
+  auto empty() const;
+  void clear() const;
```
akorgor: Some functions are marked `const` even though they can modify the underlying dictionary, for example `clear()` and `operator[]`. One would expect `const` objects to be read-only. Therefore I would suggest removing `const` from these functions:
```cpp
any_type& operator[]( const std::string& key );
any_type& operator[]( std::string&& key );
void clear();
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
 {
-  return operand.type() == typeid( T );
-}
+public:
+  Dictionary()
+    : std::shared_ptr< dictionary_ >( std::make_shared< dictionary_ >() )
+  {
+  }
+
+  operator dictionary_&() const
```
akorgor: This implicit conversion exposes the internal `dictionary_` type and allows bypassing the `Dictionary` interface. On the call site, this conversion happens hidden. Maybe consider removing the implicit conversion and requiring explicit access via `impl()` as suggested in the comment above.

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+  /**
+   * Initialize access flags to prepare for later error checking with @ref all_entries_accessed.
+   *
+   * @param thread_local_dict Pass @c true if calling this method on a thread-specific dictionary. Used for internal
+   * error checking (assert).
+   */
+  void init_access_flags( const bool thread_local_dict = false ) const;
+
+  /**
+   * Confirm that all entries in dictioary have been accessed.
```
akorgor: Typo: dictioary -> dictionary

libnestutil/dictionary.cpp
```diff
@@ -56,88 +53,115 @@ operator<<( std::ostream& os, const std::vector< T >& vec )
...
     }
-    throw boost::bad_any_cast();  // deflect to error handling below
+    throw std::bad_variant_access();  // deflect to error handling below
   }
-  catch ( const boost::bad_any_cast& )
+  catch ( const std::bad_variant_access& )
   {
     const std::string msg =
-      std::string( "Failed to cast '" ) + key + "' from " + debug_type( value ) + " to type std::vector<double>.";
+      String::compose( "Failed to cast '%1' from %2 to type std::vector<double>", key, debug_type( value ) );
```
akorgor:
```suggestion
      String::compose( "Failed to cast '%1' from %2 to type std::vector<string>", key, debug_type( value ) );
```

libnestutil/dictionary.cpp
```diff
@@ -335,44 +309,44 @@ Dictionary::register_access_( const DictEntry_& entry ) const
...
   register_access_( entry );
   return entry.item;
 }
 
-boost::any&
-Dictionary::operator[]( std::string&& key )
+any_type&
+dictionary_::operator[]( std::string&& key )
 {
   auto& entry = maptype_::operator[]( key );
```
akorgor:
```suggestion
  auto& entry = maptype_::operator[]( std::move( key ) );
```

libnestutil/dictionary.cpp
```diff
@@ -171,160 +232,73 @@ operator<<( std::ostream& os, const Dictionary& dict )
```
akorgor: `value_stream` is not used anymore and should be deleted. `type` is only used as `type.length()` in l. 237 which is always 0. Both should be deleted.

libnestutil/dictionary.cpp
```diff
@@ -171,160 +232,73 @@ operator<<( std::ostream& os, const Dictionary& dict )
...
-    or equal_as_< std::vector< std::string > >( first, second )
-    or equal_as_< std::vector< std::vector< double > > >( first, second )
-    or equal_as_< Dictionary >( first, second )
-    or equal_as_< std::shared_ptr< nest::Parameter > >( first, second )
-  );
-  // clang-format on
+  os << "[";
+  for ( auto v : av )
+  {
+    std::visit( []( const auto& arg ) { std::cout << arg << " "; }, v );
```
akorgor:
```suggestion
    std::visit( []( const auto& arg ) { os << arg << " "; }, v );
```

libnestutil/dictionary.cpp
```diff
@@ -171,160 +232,73 @@ operator<<( std::ostream& os, const Dictionary& dict )
...
-    or equal_as_< std::vector< double > >( first, second )
-    or equal_as_< std::vector< size_t > >( first, second )
-    or equal_as_< std::vector< std::string > >( first, second )
-    or equal_as_< std::vector< std::vector< double > > >( first, second )
-    or equal_as_< Dictionary >( first, second )
-    or equal_as_< std::shared_ptr< nest::Parameter > >( first, second )
-  );
-  // clang-format on
+  os << "[";
+  for ( auto v : av )
```
akorgor:
```suggestion
  for ( const auto& v : av )
```

libnestutil/dictionary.cpp
```diff
@@ -171,160 +232,73 @@ operator<<( std::ostream& os, const Dictionary& dict )
...
   // All elements are equal
   return true;
 }
 
 void
-Dictionary::register_access_( const DictEntry_& entry ) const
+dictionary_::register_access_( const DictEntry_& entry ) const
 {
   if ( not entry.accessed )
   {
```
akorgor, line 312: Typo: behind-the-scences -> behind-the-scenes

nestkernel/grid_layer.h
```diff
@@ -213,7 +213,10 @@ GridLayer< D >::get_status( Dictionary& d, NodeCollection const* const nc ) cons
 {
   Layer< D >::get_status( d, nc );
 
-  d[ names::shape ] = std::vector< size_t >( dims_.get_vector() );
+  std::vector< long > dtmp;
+  const auto& dv = dims_.get_vector();
+  std::copy( dv.begin(), dv.end(), std::back_inserter( dtmp ) );
+  d[ names::shape ] = dtmp;
```
akorgor:
```suggestion
  const auto dv = dims_.get_vector();
  d[ names::shape ] = std::vector< long >( dv.begin(), dv.end() );
```
This suggestion removes `&` to avoid confusion since `get_vector` returns a copy. Additionally, it simplifies the code.

Besides that, is the silent conversion from `size_t` to `long` intended? If yes, maybe it should be documented.

nestkernel/mpi_manager.cpp
```diff
@@ -240,7 +240,17 @@ nest::MPIManager::set_status( const Dictionary& dict )
...
+
+  long mbstd = 0;
+  if ( dict.update_value( names::max_buffer_size_target_data, mbstd ) )
+  {
+    if ( mbstd < 0 )
+    {
+      throw BadProperty( "max_buffer_size_target_data ≥ 0 required." );
+    }
+    max_buffer_size_target_data_ = mbstd;
+  };
```
akorgor:
```suggestion
  }
```

nestkernel/nest.cpp
```diff
@@ -591,30 +598,30 @@ get_model_defaults( const std::string& component )
...
+  return std::visit(
+    [ &value ]( auto&& val ) -> ParameterPTR
+    {
+      using T = std::decay_t< decltype( val ) >;
+      if constexpr ( std::is_same_v< T, ParameterPTR > )
+      {
+        return val;
+      }
+      else if constexpr ( std::is_same_v< T, double > or std::is_same_v< T, long > or std::is_same_v< T, Dictionary& >
+        or std::is_same_v< T, int > )
```
akorgor:
```suggestion
      else if constexpr ( std::is_same_v< T, double > or std::is_same_v< T, long > or std::is_same_v< T, Dictionary > )
```
`std::decay_t` removes references, therefore there should be no `&` after `Dictionary`, also `any_type` cannot be `int`, so this does not have to be checked for.

nestkernel/sonata_connector.cpp
akorgor: Sorry, this is resolved by now. Github was showing me here an outdated view:

Image

nestkernel/nest.cpp
```diff
@@ -170,12 +164,15 @@ get_kernel_status()
...
+// TODO: Add the possibility to filter for specific keys
 Dictionary
 get_nc_status( NodeCollectionPTR nc )
 {
   Dictionary result;
+  const size_t num_nodes = nc->size();
+
   size_t node_index = 0;
-  for ( NodeCollection::const_iterator it = nc->begin(); it < nc->end(); ++it, ++node_index )
+  for ( auto it = nc->begin(); it < nc->end(); ++it, ++node_index )
```
akorgor:
```suggestion
  for ( auto it = nc->begin(); it != nc->end(); ++it, ++node_index )
```
More idiomatic - just a suggestion.

nestkernel/nest.cpp
```diff
@@ -640,10 +647,10 @@ create_parameter( const Dictionary& param_dict )
   {
     throw BadProperty( "Parameter definition dictionary must contain one single key only." );
   }
-  const auto n = param_dict.begin()->first;
-  const auto pdict = param_dict.get< Dictionary >( n );
+  const std::string n = param_dict.begin()->first;
+  const Dictionary& pdict = param_dict.get< Dictionary >( n );
```
akorgor, line 648:
```suggestion
  const Dictionary pdict = param_dict.get< Dictionary >( n );
```
Since `get` returns a `Dictionary` by value not by reference, removing `&` would make it clear that `pdict` is a local copy.

nestkernel/node_collection.cpp
```diff
@@ -265,11 +265,13 @@ nc_const_iterator::advance_local_iter_to_new_part_( size_t n )
   }
 }
 
-void
-nc_const_iterator::print_me( std::ostream& out ) const
+std::ostream&
+operator<<( std::ostream& out, const nc_const_iterator& nc )
```
akorgor: Maybe rename `nc` to `nc_it` otherwise it reads as a node collection.

nestkernel/node_collection.cpp
```diff
@@ -265,11 +265,13 @@ nc_const_iterator::advance_local_iter_to_new_part_( size_t n )
...
 {
-  out << "[[" << this << " pc: " << primitive_collection_ << ", cc: " << composite_collection_ << ", px: " << part_idx_
-      << ", ex: " << element_idx_ << "]]";
+  out << "[[" << &nc << " pc: " << nc.primitive_collection_ << ", cc: " << nc.composite_collection_
```
akorgor: Is it intended here to print addresses?

pynest/nestkernel_api.pyx
```diff
@@ -197,6 +135,68 @@ cdef object Dictionary_to_pydict(Dictionary cdict):
         inc(it)
     return tmp
 
+cdef anyvec_to_objtuple(any_type operand):
+    cdef AnyVector a_vec = get[AnyVector](operand)
+    return tuple(any_to_pyobj(obj) for obj in a_vec)
+
+cdef object any_to_pyobj(any_type operand):
```
akorgor, line 143: Couldn't `operand` also hold an object of type `std::shared_ptr< nest::Parameter >`?

pynest/nestkernel_api.pyx
```diff
@@ -197,6 +135,68 @@ cdef object Dictionary_to_pydict(Dictionary cdict):
...
+    if holds_alternative[vector[string]](operand):
+        return [s.decode("utf-8") for s in get[vector[string]](operand)]
+
+    if holds_alternative[Dictionary](operand):
+        return dictionary_to_pydict(get[Dictionary](operand))
+    if holds_alternative[vector[Dictionary]](operand):
+        return vec_of_dict_to_list(get[vector[Dictionary]](operand))
+    if holds_alternative[EmptyList](operand):
+        assert False, "Should never get an EmptyList from C++"
+        return None
```
akorgor: Is it intended that `return None` is only reached if assertions are turned on?

pynest/nestkernel_api.pyx
```diff
@@ -197,6 +135,68 @@ cdef object Dictionary_to_pydict(Dictionary cdict):
...
+    if holds_alternative[long](operand):
+        return get[long](operand)
+    if holds_alternative[double](operand):
+        return get[double](operand)
+    if holds_alternative[cbool](operand):
+        return get[cbool](operand)
+    if holds_alternative[string](operand):
+        return string_to_pystr(get[string](operand))
+
+    if holds_alternative[vector[long]](operand):
```
akorgor, line 157: Is it intended that some vectors map to `np.array`, some to `list`, and `AnyVector` to `tuple`?

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
 /**
- * @brief Get the typename of the operand.
+ * Add a set of standard datatypes to @ref any_type variant.
+ *
+ * The main purpose of this type is to automatically generate variants representing the basic scalar types,
+ * strings, and dictionaries and one-dimensional @c std::vector<> of these. More specific types are added
+ * later in the actual declaration of @ref any_type.
+ *
+ * The first entry in the @c std::variant<> must be @c std::monostate , so that default-initialized @ref any_value
+ * instances will have this type. This is needed fill the entries corresponding to non-local nodes when status
```
akorgor:
```suggestion
 * instances will have this type. This is needed to fill the entries corresponding to non-local nodes when status
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+}
+
+/**
+ * @defgroup nestdict Data types related to the @ref Dictionary class in NEST
+ *
+ * The @ref Dictionary class provides the data exchange interface between the NEST kernel
+ * and the Python level as well as with C++ applications. This group collects data types
+ * related to the implementation of the @Dictionary class.
+ *
+ * The @ref Dictionary class is essentially an @c std::map<> with access checking to allow
```
akorgor: This would be more precise:
```suggestion
 * The @ref Dictionary class provides a map-like interface with access checking to allow
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+ * detection of misspelled or unknown entries passed in to NEST. Since dictionaries need to
+ * be able to contain dictionaries, the outward-facing @ref Dictionary class holds a pointer
+ * to the @ref dictionary_ instance actually implementing the dictionary as a @c std::map<>.
```
akorgor, line 56:
```suggestion
 * detection of misspelled or unknown entries passed in to NEST. Since dictionaries need to
 * be able to contain dictionaries, the outward-facing @ref Dictionary class acts as a handle
 * to the @ref dictionary_ instance actually implementing the dictionary as a @c std::map<>.
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+  /**
+   * @brief Check whether the dictionary is equal to another dictionary.
+   *
+   * Two dictionaries are equal if and only if they contain the exact same entries with the same values.
+   *
+   * @param other dictionary to check against.
+   * @return true if the dictionary is equal to the other dictionary, false if not.
+   */
+  bool operator==( const Dictionary& other ) const;
+
```
akorgor, line 232:
```suggestion
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+  template < typename T >
+  std::vector< T >& get_or_create_vector( const std::string& key );
+
+  /**
+   * @brief Update the specified non-vector value if there exists a value at key.
+   *
+   * @param key key where the value may be located in the dictionary.
+   * @param value object to update if there exists a value at key.
+   * @throws TypeMismatch if the value at key is not the same type as the value argument.
+   * @return @true if @c key was found in dictionary and @c value was updated, else @false
```
akorgor: Are @true and @false valid Doxygen commands?
```suggestion
   * @return @c true if @c key was found in dictionary and @c value was updated, else @c false
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+  template < typename T >
+  bool update_value( const std::string& key, T& value ) const;
+
+  /**
+   * @brief Update the specified value if there exists an integer value at key.
+   *
+   * @param key key where the value may be located in the dictionary.
+   * @param value object to update if there exists a value at key.
+   * @throws TypeMismatch if the value at key is not an integer.
+   * @return @true if @c key was found in dictionary and @c value was updated, else @false
```
akorgor:
```suggestion
   * @return @c true if @c key was found in dictionary and @c value was updated, else @c false
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+   *
+   * @param dict_out the dictionary to be updated.
+   * @return @c true if any values were updated in or added to @c dict_out .
+   */
+  bool update_dictionary( dictionary_& dict_out ) const;
+};
 
 /**
- * @brief Check whether two boost::any values are equal.
+ * @brief Return typename of @c operand
```
akorgor:
```suggestion
 * @brief Return typename of @c operand.
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
  *
- * Values are stored as boost::any objects, with std::string keys.
+ * @ingroup nestdict
+ */
+std::string debug_dict_types( const Dictionary& dict );
+
+/**
+ * Represent value in dictionary entry with access information.
+ *
+ * @c accessed flag is mutable so access (error) checking does not interfer with logcial const-ness
```
akorgor:
```suggestion
 * @c accessed flag is mutable so access (error) checking does not interfere with logcial const-ness.
```

libnestutil/dictionary.h
```diff
@@ -23,75 +23,377 @@
...
+    return item == other.item;
+  }
+
+  any_type item;          //!< actual item stored
+  mutable bool accessed;  //!< initially false, set to true once entry is accessed
 };
 
-class Dictionary : public std::map< std::string, DictEntry_ >
+/**
+ * @brief A Python-like dictionary_, based on std::map.
```
akorgor:
```suggestion
 * @brief A Python-like dictionary_, based on @c std::map.
```

libnestutil/dictionary.cpp
heplesser: Thanks, I will revise the code.

libnestutil/dictionary.cpp
heplesser: Good catch!

libnestutil/dictionary.cpp
heplesser: Fixed.

nestkernel/grid_layer.h
heplesser: Thanks for the suggestion, that code is much cleaner. Since all integer data is converted to `long` upon insertion to a dictionary, I think it is not necessary to comment that in every location.

nestkernel/nest.cpp
heplesser: A good point, but here I want to think a bit more about `get()`, since I really only want a reference here. We only want to read from the dictionary so there is no point in creating a local copy, particularly since dictionaries might be large.

nestkernel/nest.cpp
heplesser: Well spotted! One can then also simplify the code in the if block, I will do this separately.

nestkernel/node_collection.cpp
heplesser: This operator was actually only added for debugging and prints out highly technical information relevant for debugging at some point (see below). Since it is not used and future debugging efforts will need other information, I remove it entirely.

nestkernel/node_collection.cpp
heplesser: Yes, for debugging, but removing it now, see above.

libnestutil/dictionary.h
heplesser: Thank you for the analysis. In my view, this discussion is outside the scope of this PR, which is about a different representation of dictionary entries. Therefore, it should go into a separate issue.

But I'd also point out that a dictionary *is* a map, and is-a relationships are usually expressed via inheritance, while has-a relationships suggest composition. So in my view, the current implementation is suitable.

libnestutil/dictionary.h
heplesser: An interesting point. It turns out that this conversion was only needed because we had `Dictionary::update_dict(dictionary_& dict_out)`. But this method signature should always have been `Dictionary::update_dict(Dictionary& dict_out)` and with that fixed, this conversion can be removed.

libnestutil/dictionary.h
heplesser: Good catch! Actually, a bit surprising that this compiled at all, but that is a consequence of the shared_ptr construction—the pointer does not change, and that is all const here guaranteed.

libnestutil/dictionary.h
heplesser: Fixed.

nestkernel/nest.cpp
heplesser: Not entirely happy with this (if RVO does not avoid copying, this could be expensive), but since we need to construct temporaries in `get()` in certain cases, we need to return by value.

libnestutil/dictionary.h
```diff
@@ -257,67 +499,45 @@ class Dictionary : public std::map< std::string, DictEntry_ >
...
+   * See @ref Dictionary::update_dictionary
    */
-  template < typename T >
   bool
-  update_integer_value( const std::string& key, T& value ) const
+  update_dictionary( Dictionary& dict_out ) const
   {
-    auto it = find( key );
-    if ( it != end() )
+    for ( auto [ key, value ] : *this )
```
otcathatsya: Rather `for ( const auto& [ key, value ] : *this )` to avoid copies?

libnestutil/dictionary.h
```diff
@@ -23,75 +23,371 @@
...
+  mutable bool accessed;  //!< initially false, set to true once entry is accessed
 };
 
-class Dictionary : public std::map< std::string, DictEntry_ >
+/**
+ * @brief A Python-like dictionary_, based on @c std::map.
+ *
+ * Values are stored as @ref DictEntry_ items providing access checking.
+ */
+class dictionary_ : public std::map< std::string, DictEntry_ >
```
otcathatsya, line 393: Do we need order for anything here? Otherwise `unordered_map` has better lookup time complexity (amortized O(1))

libnestutil/dictionary.h
otcathatsya: Does `Dictionary` get fully exposed to the Python user end via Cython? If so then I'd be slightly concerned that one could call smart pointer methods from there and somehow mismanage memory or create dangling pointers, but if it's C++ kernel side only then that probably doesn't need safeguarding.

heplesser: The C++ `Dictionary` and its content are converted to Python here: https://github.com/JanVogelsang/nest-simulator/blob/043ae83d35f6dc1abc4df00e5203bc5e226b66eb/pynest/nestkernel_api.pyx#L142. So this should be safe, the C++ `Dictionary` never "gets" beyond `nestkernel_api.pyx`.

pynest/nestkernel_api.pyx
heplesser: That could only happen if a neuron or synapse model internally held a parameter object that needed to be returned as part of a `GetStatus` call. But at least the case should be covered, and if only by throwing an exception.

pynest/nestkernel_api.pyx
heplesser: I think in Python, assertions are always on, so we never get to `return None`. It might be necessary because this is a value-returning function. But the `assert` should be turned into a proper exception.

pynest/nestkernel_api.pyx
heplesser: Overall, the idea is to convert to `numpy.array` where possible. When we get an `AnyVector`, that is in response to `GetStatus()` and there NEST 3.9 returns tuples. So for compatibility we return a tuple as well.
Dictionaries cannot (easily) be turned into `numpy.array` (because they may contain a lot of different things ...), so a list of dicts seems sensible. But right now not sure where this would occur and if NEST 3.9 then returns tuple or list. We should do what NEST 3.9 does.

libnestutil/dictionary.h
heplesser: Good point, I will fix it.

libnestutil/dictionary.h
heplesser: I would like to revisit this at a later point. In the end, we want to show dictionaries sorted alphabetically by key at the Python level to make life easy for the user. We get this right now because `map` sorts alphabetically and is then copied sequentially to Python `dict`,  which is ordered in order of insertion, if I remember correctly.

pynest/nestkernel_api.pyx
heplesser: But then, since Python supports empty lists, why not just return an empty list? I will implement that.

pynest/nestkernel_api.pyx
heplesser: After some thought, since here currently are no functions in NEST that insert `Parameter` objects into dictionaries, there would be no way to test if a converter worked correctly. Therefore, I think it is best not to implement a converter now. If anyone ever creates code that puts a parameter object into a dictionary, this will be caught by https://github.com/JanVogelsang/nest-simulator/blob/043ae83d35f6dc1abc4df00e5203bc5e226b66eb/pynest/nestkernel_api.pyx#L198 
and one can then add the necessary converter.

libnestutil/dictionary.h
```diff
@@ -23,75 +23,375 @@
...
+ *
+ * The first entry in the @c std::variant<> must be @c std::monostate , so that default-initialized @ref any_value
+ * instances will have this type. This is needed to fill the entries corresponding to non-local nodes when status
+ * information for a @ref NodeCollection is returned.
  *
- * @param operand to get the typename of.
- * @return std::string of the typename.
+ * @note Conversion from Dictionary elements to and from Python datatypes is done by @c
+ * nestkernel_api.pyx:any_to_pyobj() and @c nestkernel_api.pyx:pydict_to_Dictionary(), respectively. If you add a type
+ * to the variant, you most likely need to add corresponding conversions to those functions.
```
akorgor: Maybe:
```suggestion
 * @note Conversion from @ref Dictionary elements to and from Python datatypes is done by
 * @ref nestkernel_api.pyx:any_to_pyobj and @ref nestkernel_api.pyx:pydict_to_Dictionary, respectively. If you add a type
 * to the variant, you most likely need to add corresponding conversions to those functions.
```
But please check if Doxygen is able to correctly resolve the latter two and the syntax.

pynest/nestkernel_api.pyx
```diff
@@ -197,6 +136,69 @@ cdef object Dictionary_to_pydict(Dictionary cdict):
         inc(it)
     return tmp
 
+cdef anyvec_to_objtuple(any_type operand):
+    cdef AnyVector a_vec = get[AnyVector](operand)
+    return tuple(any_to_pyobj(obj) for obj in a_vec)
+
+cdef object any_to_pyobj(any_type operand):
+    """Convert an object of any of the types in the any_type variant to a Python object."""
```
akorgor:
```suggestion
    """Convert an object of any of the types in the ``any_type`` variant to a Python object."""
```

pynest/nestkernel_api.pyx
akorgor: Ok, thank you for the explanation!

pynest/nestkernel_api.pyx
```diff
@@ -211,14 +213,17 @@ cdef is_list_tuple_ndarray_of_int(v):
 
 
 cdef Dictionary pydict_to_Dictionary(object py_dict) except *:  # Adding "except *" makes cython propagate the error if it is raised.
+    """Convert a Python dictionary to a C++ Dictionary with elements represented by the any_type variant."""
```
akorgor:
```suggestion
    """Convert a Python dictionary to a C++ ``Dictionary`` with elements represented by the any_type variant."""
```
or "C++ dictionary"

testsuite/pytests/other/test_long_overflow.py
```diff
@@ -0,0 +1,43 @@
...
+
+@pytest.mark.parametrize("python_int", [-(2**63), 2**63 - 1])
+def test_within_bounds(neuron, python_int):
+    neuron.V_m = python_int
+
+
+@pytest.mark.parametrize("python_int", [-(2**63) - 1, 2**63])
+def test_outside_bounds(neuron, python_int):
+    with pytest.raises(OverflowError):
+        neuron.V_m = python_int
```
akorgor, line 49: For clarity, consider introducing variables with descriptive names:

```suggestion
import pytest

LONG_MIN = -(2**63)
LONG_MAX = 2**63 - 1

LONG_BELOW_MIN = LONG_MIN - 1
LONG_ABOVE_MAX = LONG_MAX + 1


@pytest.fixture
def neuron():
    return nest.Create("iaf_psc_alpha")


@pytest.mark.parametrize("python_int", [LONG_MIN, LONG_MAX])
def test_within_bounds(neuron, python_int):
    neuron.V_m = python_int


@pytest.mark.parametrize("python_int", [LONG_BELOW_MIN, LONG_ABOVE_MAX])
def test_outside_bounds(neuron, python_int):
    with pytest.raises(OverflowError):
        neuron.V_m = python_int
```

PR #3749, opened 2026-03-03
Modernize the stopwatch implementation

JanVogelsang:
This PR slightly modernizes the Stopwatch implementation. C++20 concepts are not used however, as it doesn't offer any advantages over the already elegant solution.

Depends on #3751.

### DIFF
--- a/nestkernel/stopwatch_impl.h
+++ b/nestkernel/stopwatch_impl.h
@@ -29,28 +29,31 @@
 namespace nest
 {
 
-// In each of the following methods compile-time-evaluated ifs decide which code is actually included.
+/* In each of the following methods compile-time-evaluated ifs decide which code is actually included. */
 
 template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_timer >
 void
 Stopwatch< detailed_timer, threaded_timer >::start()
 {
-  if constexpr ( enable_timer and use_timer_array )
+  if constexpr ( enable_timer )
   {
-    kernel().vp_manager.assert_thread_parallel();
-    walltime_timer_[ kernel().vp_manager.get_thread_id() ].start();
-    cputime_timer_[ kernel().vp_manager.get_thread_id() ].start();
-  }
-  else
-  {
+    if constexpr ( use_timer_array )
+    {
+      kernel().vp_manager.assert_thread_parallel();
+      walltime_timer_[ kernel().vp_manager.get_thread_id() ].start();
+      cputime_timer_[ kernel().vp_manager.get_thread_id() ].start();
+    }
+    else
+    {
 // This code applies for MasterOnly timers started/stopped from serial context, but also for
 // Threaded timers called from parallel contexts if `-Dwith-threaded-timers=OFF` for backward
 // compatibility with pre-3.9 timers. Therefore, we need to restrict to the master thread here and
 // cannot assert a single-threaded context.
 #pragma omp master
-    {
-      walltime_timer_.start();
-      cputime_timer_.start();
+      {
+        walltime_timer_.start();
+        cputime_timer_.start();
+      }
     }
   }
 }
@@ -59,18 +62,21 @@ template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_ti
 void
 Stopwatch< detailed_timer, threaded_timer >::stop()
 {
-  if constexpr ( enable_timer and use_timer_array )
+  if constexpr ( enable_timer )
   {
-    kernel().vp_manager.assert_thread_parallel();
-    walltime_timer_[ kernel().vp_manager.get_thread_id() ].stop();
-    cputime_timer_[ kernel().vp_manager.get_thread_id() ].stop();
-  }
-  else
-  {
-#pragma omp master
+    if constexpr ( use_timer_array )
     {
-      walltime_timer_.stop();
-      cputime_timer_.stop();
+      kernel().vp_manager.assert_thread_parallel();
+      walltime_timer_[ kernel().vp_manager.get_thread_id() ].stop();
+      cputime_timer_[ kernel().vp_manager.get_thread_id() ].stop();
+    }
+    else
+    {
+#pragma omp master
+      {
+        walltime_timer_.stop();
+        cputime_timer_.stop();
+      }
     }
   }
 }
@@ -79,40 +85,48 @@ template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_ti
 bool
 Stopwatch< detailed_timer, threaded_timer >::is_running_() const
 {
-  if constexpr ( enable_timer and use_timer_array )
+  if constexpr ( enable_timer )
   {
-    kernel().vp_manager.assert_thread_parallel();
-    return walltime_timer_[ kernel().vp_manager.get_thread_id() ].is_running_();
-  }
-  else
-  {
-    bool is_running_ = false;
-#pragma omp master
+    if constexpr ( use_timer_array )
     {
-      is_running_ = walltime_timer_.is_running_();
+      kernel().vp_manager.assert_thread_parallel();
+      return walltime_timer_[ kernel().vp_manager.get_thread_id() ].is_running_();
     }
-    return is_running_;
+    else
+    {
+      bool is_running_ = false;
+#pragma omp master
+      {
+        is_running_ = walltime_timer_.is_running_();
+      }
+      return is_running_;
+    }
   }
+  return false;
 }
 
 template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_timer >
 double
 Stopwatch< detailed_timer, threaded_timer >::elapsed( timers::timeunit_t timeunit ) const
 {
-  if constexpr ( enable_timer and use_timer_array )
+  if constexpr ( enable_timer )
   {
-    kernel().vp_manager.assert_thread_parallel();
-    return walltime_timer_[ kernel().vp_manager.get_thread_id() ].elapsed( timeunit );
-  }
-  else
-  {
-    double elapsed = 0.;
-#pragma omp master
+    if constexpr ( use_timer_array )
     {
-      elapsed = walltime_timer_.elapsed( timeunit );
-    };
-    return elapsed;
+      kernel().vp_manager.assert_thread_parallel();
+      return walltime_timer_[ kernel().vp_manager.get_thread_id() ].elapsed( timeunit );
+    }
+    else
+    {
+      double elapsed = 0.;
+#pragma omp master
+      {
+        elapsed = walltime_timer_.elapsed( timeunit );
+      };
+      return elapsed;
+    }
   }
+  return 0.f;
 }
 
 template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_timer >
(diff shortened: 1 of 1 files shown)


nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
 Stopwatch< detailed_timer, threaded_timer >::start()
 {
-  if constexpr ( enable_timer )
+  if constexpr ( use_timer_array and enable_timer )
+  {
+    kernel().vp_manager.assert_thread_parallel();
+    walltime_timer_[ kernel().vp_manager.get_thread_id() ].start();
+    cputime_timer_[ kernel().vp_manager.get_thread_id() ].start();
+  }
+  else
```
heplesser: Just to double-check the logic here: We end up in the `else` block if `enable_timer` is false. But in that case the function calls inside the `else` block will not have any effect (and be presumably optimized away in compilation) because `StopwatchTimerType` is then `NoOpStopwatch`, isn't it?

nestkernel/stopwatch_impl.h
```diff
@@ -142,21 +121,18 @@ Stopwatch< detailed_timer, threaded_timer >::print( const std::string& msg,
...
     }
-    else
-    {
-#pragma omp master
-      {
-        walltime_timer_.print( msg, timeunit, os );
-      }
-    } // use_timer_array
-  }   // enable_timer
+  } // use_timer_array
```
akorgor:
```suggestion
  } // use_timer_array and enable_timer
```
or given the comments above
```suggestion
  } // enable_timer and use_timer_array
```
However, I find it readable also without the comment.

nestkernel/stopwatch_impl.h
```diff
@@ -196,29 +172,26 @@ template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_ti
...
-        walltime_timer_.reset();
-        cputime_timer_.reset();
-      }
-    } // use_timer_array
-  }   // enable_timer
+    {
+      walltime_timer_.reset();
+      cputime_timer_.reset();
+    }
+  } //
```
akorgor:
```suggestion
  } // use_timer_array and enable_timer
```
or given the comments above
```suggestion
  } // enable_timer and use_timer_array
```
However, I find it readable also without the comment.

nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
-// In each of the following methods compile-time-evaluated ifs decide which code
-// is actually included.
+// In each of the following methods compile-time-evaluated ifs decide which code is actually included.
 
 template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_timer >
 void
 Stopwatch< detailed_timer, threaded_timer >::start()
 {
-  if constexpr ( enable_timer )
+  if constexpr ( use_timer_array and enable_timer )
```
akorgor:
```suggestion
  if constexpr ( enable_timer and use_timer_array )
```
Before, `enable_timer` was evaluated first. Wouldn't it make sense to keep this order for efficiency and logic since I understand `enable_timer` to guard `use_timer_array`? But maybe my understanding is wrong. This appears in further 8 instances further below.

nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
-        walltime_timer_.start();
-        cputime_timer_.start();
-      }
-    } // use_timer_array
-  }   // enable_timer
+    {
+      walltime_timer_.start();
+      cputime_timer_.start();
+    }
+  } // use_timer_array
```
akorgor:
```suggestion
  } // use_timer_array and enable_timer
```
or given the comments above
```suggestion
  } // enable_timer and use_timer_array
```
However, I find it readable also without the comment. This appears in further 8 instances below.

nestkernel/mpi_manager_impl.h
```diff
@@ -45,7 +45,7 @@ nest::MPIManager::get_process_id_of_vp( const size_t vp ) const
 
 // Variable to hold the MPI communicator to use.
 #ifdef HAVE_MUSIC
-extern MPI::Intracomm comm;
+extern MPI_Comm comm;
 #else  /* #ifdef HAVE_MUSIC */
 extern MPI_Comm comm;
 #endif /* #ifdef HAVE_MUSIC */
```
akorgor: The two conditions are now the same and this whole block can be simplified to:
```cpp
extern MPI_Comm comm;
```

.github/workflows/nestbuildmatrix.yml
```diff
@@ -111,7 +111,7 @@ jobs:
 
       - name: "Run cppcheck..."
         run: |
-          cppcheck --enable=all --language=c++ --std=c++17 -i sli/ --suppressions-list=.cppcheck_suppressions ./
+          cppcheck --enable=all --language=c++ --std=c++20 -i sli/ --suppressions-list=.cppcheck_suppressions ./
```
akorgor, line 114: Wouldn't it be safer to switch to C++20 in a separate commit such that it is explicit before adding the stopwatch modernization?

nestkernel/stopwatch.h
```diff
@@ -274,23 +254,58 @@ operator<<( std::ostream& os, const StopwatchTimer< clock_type >& stopwatch )
...
+  }
+  void
+  print( const std::string&, timers::timeunit_t, std::ostream& ) const
+  {
+  }
+  double
+  elapsed( ... ) const
+  {
+    return std::numeric_limits< double >().quiet_NaN();
+  }
```
akorgor:
```suggestion
  double
  elapsed() const
  {
    return std::numeric_limits< double >().quiet_NaN();
  }
  double
  elapsed( timers::timeunit_t ) const
  {
    return std::numeric_limits< double >().quiet_NaN();
  }
```
Wouldn't an overload here be cleaner and ensure type-safety compared to the variadic?

nestkernel/stopwatch.h
```diff
@@ -274,23 +254,58 @@ operator<<( std::ostream& os, const StopwatchTimer< clock_type >& stopwatch )
...
+};
```
akorgor, line 290: Maybe consider marking these simple no-op functions as `constexpr` and `noexcept` (`print` not because it takes a stream):
```suggestion
struct NoOpStopwatch
{
 constexpr void
  start() noexcept
  {
  }
 constexpr void
  stop() noexcept
  {
  }
  constexpr void
  reset() noexcept
  {
  }
  constexpr bool
  is_running_() const noexcept
  {
    return false;
  }
  void
  print( const std::string&, timers::timeunit_t, std::ostream& ) const
  {
  }
  constexpr double
  elapsed( ... ) const noexcept
  {
    return std::numeric_limits< double >().quiet_NaN();
  }
};
```

.github/workflows/nestbuildmatrix.yml
akorgor: Ah, sorry, I just saw that the respective commits indeed are part of another PR.

nestkernel/stopwatch.h
JanVogelsang: I would believe that these no-ops are always compiled away anyway, but it of course doesn't hurt to add `constexpr` and `noexcept`.

nestkernel/stopwatch_impl.h
heplesser: Since it is a `constexpr`, I think the order does not matter for performance. But logically, I think it indeed is better to first check that we have a timer at all an then whether it is an array of timers. But it is just for looks, not for consequences.

nestkernel/stopwatch.h
JanVogelsang: Makes sense!

nestkernel/stopwatch_impl.h
JanVogelsang: Done

nestkernel/stopwatch_impl.h
JanVogelsang: Yes, exactly!

nestkernel/stopwatch_impl.h
akorgor: Just for my understanding: Why here not also `enable_timer` and `use_timer_array`?

nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
-        walltime_timer_.start();
-        cputime_timer_.start();
-      }
-    }  // use_timer_array
-  }  // enable_timer
+    {
+      walltime_timer_.start();
+      cputime_timer_.start();
+    }
+  }  // use_timer_array and enable_timer
```
akorgor: For clarity and consistency as this unnecessary comment was also removed in `reset()`:
```suggestion
  }
```

nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
-    else
-    {
-#pragma omp master
-      {
-        walltime_timer_.stop();
-        cputime_timer_.stop();
-      }
-    }  // use_timer_array
-  }  // enable_timer
+  }  // use_timer_array
```
akorgor: For clarity and consistency as this unnecessary comment was also removed in `reset()`:
```suggestion
  }
```

nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
   {
-    return false;
-  }  // enable_timer
+    bool is_running_ = false;
+#pragma omp master
+    {
+      is_running_ = walltime_timer_.is_running_();
+    }
+    return is_running_;
+  }  // use_timer_array
```
akorgor: For clarity and consistency as this unnecessary comment was also removed in `reset()`:
```suggestion
  }
```

nestkernel/stopwatch_impl.h
```diff
@@ -29,111 +29,90 @@
...
   {
-    return std::numeric_limits< double >().quiet_NaN();
-  }  // enable_timer
+    double elapsed = 0.;
+#pragma omp master
+    {
+      elapsed = walltime_timer_.elapsed( timeunit );
+    };
+    return elapsed;
+  }  // use_timer_array
```
akorgor: For clarity and consistency as this unnecessary comment was also removed in `reset()`:
```suggestion
  }
```

nestkernel/stopwatch_impl.h
```diff
@@ -142,21 +121,18 @@ Stopwatch< detailed_timer, threaded_timer >::print( const std::string& msg,
...
     }
-    else
-    {
-#pragma omp master
-      {
-        walltime_timer_.print( msg, timeunit, os );
-      }
-    }  // use_timer_array
-  }  // enable_timer
+  }  // use_timer_array
```
akorgor: For clarity and consistency as this unnecessary comment was also removed in `reset()`:
```suggestion
  }
```

nestkernel/stopwatch_impl.h
akorgor, line 164: Just for my understanding: Why here not also `enable_timer and use_timer_array`?

akorgor: For clarity and consistency as this unnecessary comment was also removed in `reset()`: remove `// use_timer_array` and `//enable_timer`. I could not add a code suggestion here since Github does not render it correctly outside of changed code.

nestkernel/stopwatch_impl.h
```diff
@@ -196,29 +172,26 @@ template < StopwatchGranularity detailed_timer, StopwatchParallelism threaded_ti
...
-      }
+      walltime_timer_[ i ].reset();
+      cputime_timer_[ i ].reset();
     }
-    else
-    {
+  }
+  else
+  {
 #pragma omp master
```
akorgor: Why here `#pragma omp master` despite the single threaded execution? If it's just a defensive guard, for consistency, I would add it for l. 165-166, too.

nestkernel/stopwatch_impl.h
JanVogelsang: Because the following code wouldn't compile to a no-op, while it would in the other cases:
```cpp
d[ walltime_name ] = walltime_timer_.elapsed();
d[ cputime_name ] = cputime_timer_.elapsed();
```
Therefore, we have to actually always disable this if timers are not enabled.

But I don't really see why we wouldn't just do it this way in all other places as well. Would probably not change the compiled machine code, but would make the code more readable.

nestkernel/stopwatch_impl.h
JanVogelsang: True, this is not needed

nestkernel/stopwatch_impl.h
JanVogelsang: Thanks! Done

PR #3753, opened 2026-03-03
Replace String::compose by std::format

JanVogelsang:
This PR replaces our custom string compose functionality by the C++ standard format function introduced in C++20.

Quote @heplesser:
> It seems we need to stick with std::format for some time, since GCC only supports std::format from v 13 onwards (https://en.cppreference.com/w/cpp/compiler_support/20) and our standard Linux runners have gcc 11.


heplesser:
I am afraid that his will have to wait until we no longer support Ubuntu 22.04 with gcc-11 as default compiler.

Issue #1933, opened 2021-02-22
Revise sp_manager::global_shuffle


heplesser:
Currently, the behavior of [`sp_manager::global_shuffle()`](https://github.com/nest/nest-simulator/blob/5af666418730e3a0a7f7a1c0f717d19dd23a6434/nestkernel/sp_manager.cpp#L703) does not agree with the comment describing the method:

- The comment states that the method "shuffles the first n elements of the vector", implying that the elements from n+1 on remain in the vector in original order
- The method randomly picks n elements from the entire vector without replacement and then replaces the content of the vector with these first n elements, dropping the rest

The way the method is used (eg here https://github.com/nest/nest-simulator/blob/5af666418730e3a0a7f7a1c0f717d19dd23a6434/nestkernel/sp_manager.cpp#L439-L440) indicates that the actual behavior is what is intended (but the `resize()` after calling the method is then not necessary).

Comment and code need to be brought into agreement. One might also consider a more informative method name, including the "chopping" part. The implementation should be reviewed as well. It should be able to work in-place without need of the "erase" function (costly). It might make sense to use suitable C++11 functions, e.g. https://www.cplusplus.com/reference/algorithm/random_shuffle/. Finally, variable names in the method are very uninformative and should be revised.

heplesser:
@sdiazpier @pnbabu Ping!

pradkrish:
Hello, Is this issue still open? I am happy to give it a go.

heplesser:
Yes, the issues appears still to be open.

PR #3779, opened 2026-03-11
Improve sp_manager::global_shuffle

alpinangu:
Shuffles without creating a copy of the passed vector in O(n). Original version uses erase on each itteration which makes it O(n^2). Also addresses [#1933](https://github.com/nest/nest-simulator/issues/1933)


nestkernel/sp_manager.cpp
```diff
@@ -655,22 +655,15 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
-    rndi = v.begin();
-    v.erase( rndi + rnd );
-  }
-  v = v2;
+  // We select an element from [ith index, vector end-index(remaining range)]
+  for ( size_t i = 0; i < n; ++i )
+  {
+    size_t rnd = i + get_rank_synced_rng()->ulrand( v.size() - i );  // Random index in range [i, v.size()-1]
+    std::swap( v[ i ], v[ rnd ] );                                   // Swap selected element to the front
+  }
```
heplesser: Since `v` does not change in size, this may be even clearer and faster:

```suggestion
  for ( size_t next = 0, remaining = v.size() ; next < n ; ++next, --remaining )
  {
    const size_t rnd = next + get_rank_synced_rng()->ulrand( remaining );  // Random index in range [next, remaining-1]
    std::swap( v[ next ], v[ rnd ] );                                   // Swap selected element to the front
  }
```

For testing, I'd compile NEST with array bounds checking activated for [] (some compiler flag I don't remember right now). Also, I noticed that the comment here 
https://github.com/nest/nest-simulator/blob/83eec3756eba42ce4adb4a33cf4074856ee132a9/nestkernel/random_generators.h#L120
is wrong: The maximum value that can drawn is is N-1. This is correct in the line above, where the interval is given as [0, N), i.e., right-open, excluding N. Could you add a fix for the comment to this PR?

heplesser (changes requested):
Nice! Just a small suggestion for further improvement. For testing on larger networks, you could try using MPI parallelization.

nestkernel/sp_manager.cpp
alpinangu: Thank you, this loop is much more readable.
Could it be that the comment for ulrand was meant to say that there are N values that can be drawn as in N maximum values, starting from 0? I’ve made it clearer now.

heplesser (approved):

nestkernel/sp_manager.cpp
```diff
@@ -655,22 +655,15 @@ nest::SPManager::global_shuffle( std::vector< size_t >& v, size_t n )
...
-    rndi = v.begin();
-    v.erase( rndi + rnd );
-  }
-  v = v2;
+  // We select an element from [ith index, vector end-index(remaining range)]
+  for ( size_t next = 0, remaining = v.size(); next < n; ++next, --remaining )
+  {
+    const size_t rnd = next + get_rank_synced_rng()->ulrand( remaining );  // Random index in range [next, remaining-1]
+    std::swap( v[ next ], v[ rnd ] );                                      // Swap selected element to the front
+  }
```
heplesser, line 663: This code would benefit from better documentation:

```suggestion
  // Randomly select elements from [next, v.size).
  // Invariant: next + remaining == v.size()
  for ( size_t next = 0, remaining = v.size(); next < n; ++next, --remaining )
  {
    const size_t rnd = next + get_rank_synced_rng()->ulrand( remaining );  // Random index in range [next, v.size()-1]
    std::swap( v[ next ], v[ rnd ] );                                      // Swap selected element to the front
  }
```

heplesser (changes requested):
Please see suggestion for improved comment.

heplesser:
See #3896 for a possible alternative.

alpinangu:
Closing this and converging on [#3896](https://github.com/nest/nest-simulator/pull/3896) since this branch is behind on main

PR #3784, opened 2026-03-18
Fix connection generator code to work with PyNEST-NG

jessica-mitchell:
This is a solution found by Claude code.
csa_example.py and csa_spatial_example.py both worked in Docker container with these modifications
**The following text and code changes were done with AI:**

| File | Change |
|------|--------|
| `pynest/pynestkernel_aux.h` | Remove `conngenmodule.h` include; replace `Datum*`-returning `CYTHON_unpackConnectionGeneratorDatum` with `CYTHON_insertConnectionGenerator` that stores `shared_ptr<ConnectionGenerator>` into a `boost::any` Dictionary slot |
| `pynest/nestkernel_api.pxd` | Add `cdef extern` declarations for `CYTHON_isConnectionGenerator` and `CYTHON_insertConnectionGenerator` |
| `pynest/nestkernel_api.pyx` | Add `elif CYTHON_isConnectionGenerator` branch in `pydict_to_Dictionary` before the `AttributeError` fallthrough |
| `pynest/CMakeLists.txt` | Add `${LIBNEUROSIM_LIBRARIES}` to `nestkernel_api` link libraries |
| `nestkernel/conn_builder_conngen.cpp` | Read `params_map` indices as `long` not `size_t`; `static_cast` to `size_t` |

Fix: ConnectionGenerator (libneurosim/CSA) support in pynest-ng

## Background

When pynest-ng (Cython) was merged into `main`, the ConnectionGenerator
integration from the old SWIG-based pynest was not fully ported. There are
three distinct problems:

### Problem 1 — `pynestkernel_aux.h` exists but is broken and unused

The file `pynest/pynestkernel_aux.h` was carried over from the SWIG era but
was never wired into pynest-ng. It has two bugs that make it unusable as-is:

**Bug A — includes a module that no longer exists:**
```c
// Includes from conngen:
#include "conngenmodule.h"    // <-- ConngenModule was removed from NEST main
```
This line causes a compile error if the header is ever included while
`HAVE_LIBNEUROSIM` is defined.

**Bug B — returns `Datum*`, which is incompatible with pynest-ng:**
```c
Datum*
CYTHON_unpackConnectionGeneratorDatum( PyObject* obj )
{
  ...
  ret = static_cast< Datum* >( new nest::ConnectionGeneratorDatum( cg ) );
  return ret;
}
```
pynest-ng does not use the SLI `Datum` type. All values are stored in
`boost::any` fields inside a `Dictionary`. A function returning `Datum*` cannot
be used to insert a ConnectionGenerator into a pynest-ng Dictionary.

### Problem 2 — `pydict_to_Dictionary` has no ConnectionGenerator branch

`nestkernel_api.pyx` falls through to `raise AttributeError` for any type it
does not recognise. `pynestkernel_aux.h` is not included and its macros are
never called:

```
AttributeError: value of key (cg) is not a known type,
got <class 'csa.connset.ConnectionSet'>
```

### Problem 3 — `conn_builder_conngen.cpp` reads `params_map` as `size_t`

`boost::any` requires an exact type match. `pydict_to_Dictionary` stores Python
`int` values as `long`, but the conngen builder reads them as `size_t`:

```
NESTErrors.TypeMismatch: Failed to cast 'delay' from long to type unsigned long
```

---

## Reproducer

With NEST built with `libneurosim` support:

```python
import nest
import csa

pre  = nest.Create("iaf_psc_alpha", 4)
post = nest.Create("iaf_psc_alpha", 4)

cg = csa.cset(csa.oneToOne, 10000.0, 2.0)
params_map = {"weight": 0, "delay": 1}
connspec = {"rule": "conngen", "cg": cg, "params_map": params_map}

nest.Connect(pre, post, connspec)  # raises AttributeError  (problem 2)
                                   # after fixing problem 2: raises TypeMismatch (problem 3)
```

The existing test suite in
`testsuite/pytests/connect/test_connect_conngen.py` covers all four conngen
scenarios and serves as the regression suite for this fix.

### DIFF
--- a/nestkernel/conn_builder_conngen.cpp
+++ b/nestkernel/conn_builder_conngen.cpp
@@ -100,8 +100,8 @@ ConnectionGeneratorBuilder::connect_()
       throw BadProperty( "The parameter map has to contain the indices of weight and delay." );
     }
 
-    const size_t d_idx = params_map_.get< size_t >( names::delay );
-    const size_t w_idx = params_map_.get< size_t >( names::weight );
+    const size_t d_idx = static_cast< size_t >( params_map_.get< long >( names::delay ) );
+    const size_t w_idx = static_cast< size_t >( params_map_.get< long >( names::weight ) );
 
     const bool d_idx_is_0_or_1 = d_idx == 0 or ( d_idx == 1 );
     const bool w_idx_is_0_or_1 = w_idx == 0 or ( w_idx == 1 );
--- a/pynest/CMakeLists.txt
+++ b/pynest/CMakeLists.txt
@@ -36,7 +36,7 @@ if ( HAVE_PYTHON )
   endif ()
 
   # PYNEST-NG-FUTURE: somehow it works without models being in here, but why?
-  target_link_libraries( nestkernel_api nestutil nestkernel )
+  target_link_libraries( nestkernel_api nestutil nestkernel ${LIBNEUROSIM_LIBRARIES} )
 
   if ( APPLE )
     set_target_properties( nestkernel_api PROPERTIES LINK_FLAGS "-undefined dynamic_lookup" )
--- a/pynest/nestkernel_api.pxd
+++ b/pynest/nestkernel_api.pxd
@@ -130,6 +130,11 @@ cdef extern from "mask.h" namespace "nest":
         MaskPTR()
 
 
+cdef extern from "pynestkernel_aux.h":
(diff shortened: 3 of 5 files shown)


heplesser:
Further analysis as mentioned in #3798 has revealed that Connection Generator/Libneurosim/CSA support has not been working in NEST from the Python level since at least NEST 3.6. Therefore, I am closing this PR.

PR #3795, opened 2026-03-25
Use std::expm1() if available instead of hand-written implementation 

heplesser:
In many of our models, we use `expm1()` to compute propagator elements. Since `expm1()` historically was not guaranteed to be available in `math.h` or `cmath`, we let CMake check if `expm1()` was available in `math.h` and if not, we provided our own implementation.

It now turns out that this CMake mechanism was incorrectly implemented: The detection worked correctly, but `HAVE_EXPM1` was never set in `config.h` due to an extra argument to the pertaining CMake function that got things messed up. This PR fixes this, so that `HAVE_EXPM1` is now properly set.

One could wonder whether we need the check at all, given that according to [CPPReference](https://cppreference.com/w/cpp/numeric/math/expm1.html) is part of the standard; but a Perplexity query gave a more ambiguous result.

Now the problem here is that this fix will change NEST results, because the `expm1()` function from C++ may return different results from our implementation, which has been used in all NEST experiments over at least the past decade due to the bug described and fixed here.

To test the consequences, I performed simulations with the `brunel_alpha_nest.py` example and a modified version using the precise `iaf_psc_alpha_ps` neuron. For the normal `iaf_psc_alpha` neuron, I did not find a single spike differed when recording from 500 spikes for 10s. But for the precise neurons, already the time of the first spikes differed by about 10 eps with "macroscopic" differences in spike patterns later in the simulation. The precise neurons are a much harder test since they evaluate `expm1()` essentially for every incoming spike and thus also probe a much wider range of argument values, especially values close to 0. The implementation from `math.h`, btw, gives about 25% faster network simulation times for the ps-case.

AI Assistance: [moderate]
Model: Claude Opus 4.8 (for `testutil.py:assert_approx_equal()`)


diesmann (approved):
This change is fine. Appendix A.3 of https://doi.org/10.1162/neco.2007.19.1.47 describes some of the background.

heplesser:
@suku248 I have now fixed the merge conflicts.

heplesser:
@suku248 It turned out that the expm1() change here changed the last bit in some V_m values, so converting to approx() comparisons.

suku248 (approved):
👍

PR #3796, opened 2026-03-26
Add Apple' stub-based libraries (.tbd) to list of library suffixes

heplesser:
This fixes a CMake problem that currently breaks all macOS builds including HDF5, because the FindHDF5 process chokes on a ZLIB dependency on up-to-date Macs (Apple introduced the .tbd ending in 2015, we just never noticed so far ...).

Marking this as critical because it breaks the CI.


JanVogelsang (approved):

gtrensch (approved):
LGTM 👍

PR #3797, opened 2026-03-27
Set up examples.yml as metadata file and add workflow to check for discrepancies in pynest examples

jessica-mitchell:
This PR create a single source of truth for metadata about the `pynest/examples` in `examples.yml` and includes workflow that checks it's correctness at pre-release.

This file is planned to be used for several downstream workflows and Sphinx scripts.

The idea is that this yml file should be  manually updated when there are changes to pynest/examples  and a contributor can run the check_examples_registry.py. (instructions added to docs), to check. 

But we also run a workflow in nest-simulator that checks for discrepancies between the example files and the yml file at time of pre-release (including more thorough checks). If any change has happened to one but not the other this will cause the workflow to fail.
Changes need to be manually fixed and workflow re-run.
Once the workflow passes, this triggers a downstream repo (nest-simulator-examples) to run. 

Currently, the downstream repo workflows still need to be finalized, and the Sphinx extension need to be updated (obviously the file needs to exist first).

this was developed with AI (Claude code)


terhorstd (changes requested):
The structure looks okay to me, however the code quality of the script files is quite low. Maybe avoid single-letter variable names and add type annotations. The `ignore_patterns` can be scanned with `fnmatch.filter` more concisely and without deeply nested `for` and `if` statements. Many of the functions could be reduced to very few lines when using generators. Using `logging` module would avoid the `print(…, file=sys.stderr)` and keep the output functions separate also for future refactoring.
I don't get what the "PREVIOUS_TAG" entry is supposed to be used for. If it's change tracking, the first script could add a checksum or commit hash. If it's for annotating new examples "new since", then maybe the upcoming release would be a better value than a hard-coded magic string. In any case, a different field name should be considered.

jessica-mitchell:
@terhorstd 
thanks for the feedback, the scripts have been updated with logger, list comprehensions and better naming conventions for variables, and some other checks
I also changed the 'type' in examples.yml to two separate entities: 'runner' which now can be python or snakemake and 'run_in_ci' [bool]; this is to separate out the 'skip' which was included in the 'type' with python and snakemake before, but didnt make much sense as a third option

pynest/examples/examples.yml
```diff
@@ -0,0 +1,1218 @@
...
+#   that cannot be executed in the standard CI environment (missing optional
+#   library, MPI setup, special hardware, etc.).  Entries with run_in_ci: false
+#   must have a non-empty 'notes' field explaining why; they appear in the
+#   dashboard skipped-examples dropdown.
+#
+# convert_to_notebook: true | false
+#   If true, convert-notebooks-on-prerelease.yml converts the example to a
+#   Jupyter notebook for EBRAINS.  Independent of run_in_ci, but the
+#   convention is to set this false when run_in_ci is false — an untested
+#   notebook can rot silently and break for end users.
```
heplesser: I don't understand your intention here. We want to avoid rot—does this contribute to avoiding it?

pynest/examples/examples.yml
```diff
@@ -0,0 +1,1218 @@
...
+#
+# category: str
+#   Thematic group (e.g. neurons, networks, synapses, spatial).
+#   Used by the Sphinx listexamples directive to group examples on model pages.
+#
+# last_change: "no change" | "changes since version vMAJOR.MINOR" | "new since version vMAJOR.MINOR"
+#   Must be updated for every .py file that changes between releases.
```
heplesser, line 46: Will this even apply for adding a single missing comma? That seems tedious, and why not auto-detect such changes?

pynest/examples/examples.yml
```diff
@@ -0,0 +1,1218 @@
...
+#
+# models: list[str]
+#   NEST model names the example demonstrates.
+#   Planned use: Sphinx listexamples directive (see EXAMPLES_PIPELINE.md).
+#   Note: do not list example-local CopyModel aliases — only the source model names.
+#
+# ── Optional fields ──────────────────────────────────────────────────────────────
+#
+# notes: str
+#   Free-text explanation.  Required for entries with run_in_ci: false (shown in
```
heplesser: What do you mean with "explanation" here? Here it sounds mostly like technical aspects.

pynest/examples/examples.yml
```diff
@@ -0,0 +1,1218 @@
...
+  - iaf_psc_alpha
+  - poisson_generator
+  - spike_recorder
+  - voltmeter
+- name: clopath_synapse_small_network
+  path: clopath_synapse_small_network.py
+  runner: python
+  run_in_ci: true
+  notes: SIGABRT in poisson_generator::pre_run_hook with OpenMP+MPI build; possible
+    NEST bug
```
heplesser: Is there already an issue for this SIGABRT?

pynest/examples/examples.yml
```diff
@@ -0,0 +1,1218 @@
...
+  last_change: "no change"
+  notes: Requires MUSIC library and special MPI setup
+- name: music_cont_out_proxy_receiver
+  path: music_cont_out_proxy_example/receiver_script.py
+  runner: python
+  run_in_ci: false
+  convert_to_notebook: false
+  category: special
+  last_change: "no change"
+  notes: Helper script for music_cont_out_proxy; not a standalone example
```
heplesser, line 1213: Would it not make more sense to list the helpers as helpers under the actual example?

pynest/examples/examples.yml
```diff
@@ -0,0 +1,1218 @@
...
+#   prerelease_check.py verifies all listed paths exist on disk (CI Check 4).
+#
+# data_files: list[{path: str, formats: list[str], description: str}]
+#   Data files bundled with the example (relative to pynest/examples/).
+#   prerelease_check.py verifies that each 'path' exists on disk (CI Check 4).
+#
+# post_script: str
+#   A script to run after the main example, relative to pynest/examples/.
+#   Used by run-examples.yml (e.g. to generate a GIF from simulation output).
+#
```
heplesser:
```suggestion

```

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
```diff
@@ -171,14 +171,36 @@ If you want to edit Model docs, PyNEST API files, or PyNEST examples, you will n
...
+
+   To verify your changes are consistent with the files on disk, run from the repo root:
+
+   .. code-block:: bash
+
+      python .github/scripts/check_examples_registry.py --examples-dir pynest/examples
+
+   The script exits with a summary of any missing or orphaned entries and prints
+   copy-paste–ready YAML stubs for anything that needs to be added.
+   It requires only ``pyyaml`` (``pip install pyyaml``).
```
heplesser, line 202: Is pyyaml in our requirements?

heplesser (changes requested):
@jessica-mitchell This looks interesting, see also my inline comments. I have one major question: Would it not make more sense to include the metadata as comments in the individual example files and collect them from there? I believe that would make maintenance easier. "Main" files would then list their "helpers", and that way the system can check that all files are either mains or known through mains.

pynest/examples/examples.yml
jessica-mitchell: Yes, so there are a few examples where-- at least with my initial test of running them downstream in CI-- I couldnt get them to work  (e.g. too large and timed out), and I was thinking we might want to skip some examples, but we also then probably dont want them as notebooks with 'try it on EBRAINS button' either. So the convention should be to have both of these settings false. 

Of course the better solution is to figure out how to to test all the examples, but that still needs to be looked at more carefully.

pynest/examples/examples.yml
jessica-mitchell: so the way the script works, it will find any change and flag it and expect the value in the  yml to be updated, which is not ideal in the case of minor typos, but I have not come up with a good solution for this case. I figured, this is likely not to happen very often.

pynest/examples/examples.yml
jessica-mitchell: yah this could be reworded but the intent is that it's not expecting a specific string or value

pynest/examples/examples.yml
jessica-mitchell: No I dont think so, and I think we can probably remove it; during my tests, I think I managed to get this example to work (I had tested the examples using an artifact built from nestbuildmatrix where this failed, but when I tried with a docker image it worked).

pynest/examples/examples.yml
jessica-mitchell: I dont know, I thought it would be more efficient for setting up downstream scripts to have the info they needed (like to know which scripts need to be converted) in one place rather than have to scrape this info from many files.

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
jessica-mitchell: good catch I will add it

pynest/examples/examples.yml
heplesser: I would remove the part about rot, it is more project philosophy than documentation of the meaning of flags:

```suggestion
# convert_to_notebook: true | false
#   If true, convert-notebooks-on-prerelease.yml converts the example to a
#   Jupyter notebook for EBRAINS.  Independent of run_in_ci, but the
#   convention is to set this to false when run_in_ci is false.
```

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
jessica-mitchell: ah it's actually there in doc/requirements.txt but it's ordered by uppercase  then lower case, so i missed it

heplesser (approved):

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+    the source of truth for entry points.
+
+    Returns the guessed entry-point Path, or None when it is ambiguous (no
+    conventional filename and more than one .py file), in which case each file
+    is suggested under its own stem.
+    """
+    for pattern in ENTRY_POINTS:
+        for path in py_files:
+            if path.name == pattern:
+                return path
```
terhorstd, line 118: This does not check for `pattern` as it gives the impression. Either this has to be `fnmatch()`ed (as in `should_ignore`), or (as in the above case) there are no wildcards and we can just match directly:
```suggestion
    for path in py_files:
        if path.name in ENTRY_POINTS:
            return path
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+    if len(py_files) == 1:
+        return py_files[0]
```
terhorstd, line 120: So if there is just a single file in the directory, that does not match any entry-point, we take that as entry point?! – It cannot be assumed that's even a script in any sense. What's the reasoning behind assuming this is an entry-point vs. just raising an error directly?

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+                discovered.setdefault(key, []).append(rel_path)
+
+    return discovered
+
+
+def load_yaml_file(yaml_file: Path) -> dict:
+    """Load and parse examples.yml."""
+    if not yaml_file.exists():
+        logger.error("%s not found", yaml_file)
+        sys.exit(1)
```
terhorstd, line 161: This doesn't make sense, as file-not-found would raise anyways and potentially cause an exit with exactly this error message anyway. Having something like this in the code however prevents calling context to catch this and possibly handle the situation different from dying.
```suggestion
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+    try:
+        return yaml.safe_load(yaml_file.read_text(encoding="utf-8"))
+    except yaml.YAMLError as exc:
+        logger.error("Invalid YAML file: %s", exc)
+        sys.exit(1)
+
```
terhorstd, line 167: Same argument here.
```suggestion
        raise
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+
+def validate_yaml(data: dict) -> None:
+    """Validate YAML structure."""
+    if not isinstance(data, dict):
+        logger.error("YAML root must be a dictionary")
+        sys.exit(1)
```
terhorstd, line 173: Same argument here
```suggestion
        raise ValueError("YAML root must be a dictionary")
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+    if "examples" not in data:
+        logger.error("YAML must contain 'examples' key")
+        sys.exit(1)
```
terhorstd, line 176:
```suggestion
        raise ValueError("YAML must contain 'examples' key")
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+    if not isinstance(data["examples"], list):
+        logger.error("'examples' must be a list")
+        sys.exit(1)
```
terhorstd, line 179:
```suggestion
        raise ValueError("'examples' must be a list")
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+    seen_paths: dict[str, str] = {}
+    for idx, example in enumerate(data["examples"]):
+        if not isinstance(example, dict):
+            logger.error("Example %d must be a dictionary", idx)
+            sys.exit(1)
```
terhorstd, line 184:
```suggestion
            raise ValueError("Example %d must be a dictionary", idx)
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+        if "name" not in example:
+            logger.error("Example %d missing 'name' field", idx)
+            sys.exit(1)
```
terhorstd, line 187:
```suggestion
            raise ValueError("Example %d missing 'name' field", idx)
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+        if "path" not in example:
+            logger.error("Example '%s' missing 'path' field", example.get("name"))
+            sys.exit(1)
```
terhorstd, line 190:
```suggestion
            raise ValueError("Example '%s' missing 'path' field", example.get("name"))
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+        path = example["path"]
+        if path in seen_paths:
+            logger.error(
+                "Duplicate path %r in examples.yml (entries %r and %r)",
+                path,
+                seen_paths[path],
+                example["name"],
+            )
+            sys.exit(1)
```
terhorstd, line 199:
```suggestion
            raise ValueError(f"Duplicate path {path} in examples.yml (entries {seen_paths[path] and {example['name']})")
```

.github/scripts/check_examples_registry.py
```diff
@@ -0,0 +1,355 @@
...
+            "run_in_ci": True,
+            "convert_to_notebook": True,
+            "category": item["category"],
+            "last_change": last_change_value,
+        }
+        # safe_dump handles quoting for any special characters in name/path/category.
+        rendered = yaml.safe_dump([entry], default_flow_style=False, sort_keys=False).rstrip()
+        chunks.append(textwrap.indent(rendered, "  "))
+
+    return "\n".join(chunks) + "\n"
```
terhorstd, line 279: It's a bit of pain to see this piece-wise rendering + sort + indent + append + join ... – All just because pyyaml doesn't do comments. There are tons of reasons why to switch to `ruamel.yaml`…

However, for the sake of this PR leave this here. I'll cry silently 😭

doc/htmldoc/developer_space/workflows/documentation_workflow/user_documentation_workflow.rst
terhorstd: What was this added for originally? If this was not used, I'd strongly opt for `ruamel.yaml` instead.

.github/workflows/trigger-examples-on-prerelease.yml
```diff
@@ -0,0 +1,142 @@
...
+          echo "tag=$PREV" >> "$GITHUB_OUTPUT"
+          echo "Previous stable release : $PREV"
+          echo "Current ref             : $REF_NAME"
+
+      - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v5.6.0
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: pip install pyyaml
```
terhorstd, line 93: Why not use an already curated spot here? We wouldn't want to remember updating yet another line of pip install somewhere.
```suggestion
        run: pip install -r doc/requirements.txt
```

.github/scripts/prerelease_check.py
```diff
@@ -0,0 +1,240 @@
...
+    result = subprocess.run(
+        ["git", "rev-parse", "--show-toplevel"],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return Path(result.stdout.strip())
+
+
+def run_git(git_args: list[str], repo_root: Path) -> subprocess.CompletedProcess[str]:
```
terhorstd, line 74: `run_git` is called only once in the whole script. Having a separate function just adds complexity and the actual call in L87 isn't that much cleaner either.

.github/scripts/prerelease_check.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def load_yaml_entries(examples_dir: Path) -> list[dict]:
+    yaml_file = examples_dir / "examples.yml"
+    with yaml_file.open(encoding="utf-8") as yaml_fh:
+        data = yaml.safe_load(yaml_fh)
+    return data["examples"]
+
+
+def main() -> None:
```
terhorstd, line 126: Hey, the 80's are back. 🍝 

This would profit a lot from OOP. For example having the `repo_root` calculated above, then called here and handed around to each function just adds more lines to the not very readable script everywhere.
Also the constant jumps between absolute and relative paths seem very verbose, as we should be able to handle most things relative anyway. The absolute path cannot be relevant on CI or across user laptops, …
Also, doesn't this do exactly the same as `check_examples_registry.py`, just with `discovered` replaced by `changed_files`?! Why go through the same pain again? Also the "Checing required fields" is basically `validate_yaml`, "Checking … files on disk" is `compare_discovered_with_yaml` and so on. It lines up almost perfectly.

Please refactor `get_changed_files` into the other script and use the same code-
base. Don't use different code to do the same thing.

terhorstd (changes requested):
A couple of things to look into. Especially the maintainability needs to be improved.

PR #3801, opened 2026-04-13
Fix use-after-free of ltdl error string

clinssen:
This PR fixes a small use-after-free bug in handling the ltdl error string in case of a specific failure.

The issue is as follows: we grab a pointer to the error string (a char*) returned by ltdl. We then call ``lt_dlclose()``, which frees the memory where the string is at, and then we construct a std::string using the pointer. This can result in incorrect detection of the error condition and garbled output, like:

``
DynamicModuleManagementError: Module '[snip]/foomodule.so' could not be loaded.
The dynamic loader returned the following error: ''00'.
``

``
DynamicModuleManagementError: Module '[snip]/foomodule.so' could not be loaded.
The dynamic loader returned the following error: 'j/j'.
``

This PR fixes the issue by converting to a std::string straight away.

### DIFF
--- a/nestkernel/module_manager.cpp
+++ b/nestkernel/module_manager.cpp
@@ -104,7 +104,7 @@ ModuleManager::get_status( Dictionary& d )
 }
 
 void
-ModuleManager::set_status( const Dictionary& d )
+ModuleManager::set_status( const Dictionary& )
 {
 }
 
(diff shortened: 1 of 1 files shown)


nestkernel/module_manager.cpp
heplesser: I realize this is unrelated, but since it is a tiny fix for a warning, I include it here anyways:

```suggestion
ModuleManager::set_status( const Dictionary& )
```

heplesser (approved):
@clinssen Thanks for diagnosing and fixing this. I added a small suggestion to fix a warning.

med-ayssar (approved):
Looks good!

PR #3805, opened 2026-04-16
Add CMake Presets

med-ayssar:
NEST has no `CMakePresets.json`, forcing every developer to manually remember and type the same long `-D flag` chains every time they configure a build. This adds friction for new contributors and makes it easy to misconfigure.

This PR adds `CMakePresets.json` with 7 presets covering the most common build scenarios (`default`, `debug`, `release`, `mpi`, `mpi-debug`, `minimal, docs`), so developers can replace verbose `cmake -Dwith-mpi=ON -Dwith-optimize=ON ...` invocations with a single `cmake --preset <name>.`

## CMake Presets
| Preset | Description | Key flags |
|--------|-------------|-----------|
| `default` | Standard dev build | `-O2`, OpenMP, Python, GSL, Boost |
| `debug` | Debug build | `-g`, no optimization |
| `release` | Optimized release | `-O3` |
| `mpi` | MPI + OpenMP | `-O2`, MPI enabled |
| `mpi-debug` | MPI debug | `-g`, MPI enabled |
| `minimal` | No optional deps | No Python/MPI/Boost/GSL |
| `docs` | With user docs | Requires Sphinx + Pandoc |

Each preset puts its binary dir under `build/<preset-name>/` to avoid conflicts.

### List all presets

```bash
cmake --list-presets
```

### Configure
```bash
cmake --preset default
cmake --preset debug
cmake --preset release
cmake --preset mpi
cmake --preset mpi-debug
cmake --preset minimal
cmake --preset docs
```

### Build
```bash
cmake --build --preset default
cmake --build --preset debug
```
### Install
```bash
cmake --build --preset default --target install
```

## Example
```bash
cmake --preset debug
cd build/debug
make install -j7
```

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,142 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": { "major": 3, "minor": 19, "patch": 0 },
+  "configurePresets": [
+    {
+      "name": "base",
+      "hidden": true,
+      "description": "Common settings shared by all presets",
+      "binaryDir": "${sourceDir}/build/${presetName}",
+      "cacheVariables": {
+        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+        "with-python": "ON",
+        "cythonize-pynest": "ON",
+        "with-openmp": "ON",
+        "with-boost": "ON",
+        "with-ltdl": "ON",
+        "with-gsl": "ON",
+        "with-warning": "ON",
+        "with-modelset": "full",
+        "with-threaded-timers": "ON",
+        "target-bits-split": "standard"
+      }
+    },
+    {
+      "name": "default",
+      "displayName": "Default",
+      "description": "Standard build with optimizations (mirrors existing build/ configuration)",
+      "inherits": "base",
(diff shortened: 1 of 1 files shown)


med-ayssar:
@heplesser, @terhorstd and @gtrensch: could you maybe give it a try and approve it if you see it fit.

heplesser:
Questions raised in in-person discussion
- Which presets? Should have a "production" preset as well, review definition of others
- Can presets be combined?
- How will directories work, especially install directories?
- How does this relate to cmake build type variable?

heplesser:
@med-ayssar I will get back to this in a while, the main question here is which presets we want to have precisely how to define them.

As part of my big CMake refactoring, Claude remarked the following, which we might want to take into account here (line numbers are certainly off as they relate to code per #3912):

> The CMAKE_BUILD_TYPE elseif chain (CMakeLists.txt:240-257) hard-errors on any build type other than the five standard CMake ones — blocks legitimate custom types (e.g. Coverage, ASan) some projects rely on. Might be intentional; worth a deliberate decision rather than a reflexive fix.

PR #3808, opened 2026-04-21
Fix failing weight plot in e-prop examples

akorgor:
This PR fixes the failing weight plotting routine by introducing a guard against the behavior of the weight recorder to record weights from non-existent connections.


jessica-mitchell (approved):
thanks @akorgor for the quick response. I have tested a couple of the examples and they work with the nest 3.10_rc1. One note to consider for the future - the examples take a long time to run locally and by default at least, no logging is output to the terminal so the user cant tell what's going on. It might be nice to add some status messages.

akorgor:
> thanks @akorgor for the quick response. I have tested a couple of the examples and they work with the nest 3.10_rc1. One note to consider for the future - the examples take a long time to run locally and by default at least, no logging is output to the terminal so the user cant tell what's going on. It might be nice to add some status messages.

Thank you @jessica-mitchell, yes, I agree, this point is on my to-do list.

PR #3816, opened 2026-04-24
Fix Compiler Error when using Clang22

### LINKED ISSUE #3813: Compilation fails with Clang 22
When building NEST (main@97cf79d24) with Clang 22.1.4 (from homebrew), compilation fails with the following error message:
```
[ 34%] Building CXX object models/CMakeFiles/models.dir/bernoulli_synapse.cpp.o
In file included from /Users/plesser/NEST/code/src/main/models/bernoulli_synapse.cpp:23:
In file included from /Users/plesser/NEST/code/src/main/models/bernoulli_synapse.h:28:
In file included from /Users/plesser/NEST/code/src/main/nestkernel/connection.h:27:
In file included from /Users/plesser/NEST/code/src/main/nestkernel/common_synapse_properties.h:27:
In file included from /Users/plesser/NEST/code/src/main/nestkernel/connector_model.h:28:
In file included from /opt/homebrew/Cellar/llvm/22.1.4/bin/../include/c++/v1/string:654:
In file included from /opt/homebrew/Cellar/llvm/22.1.4/bin/../include/c++/v1/string_view:975:
In file included from /opt/homebrew/Cellar/llvm/22.1.4/bin/../include/c++/v1/algorithm:1875:
In file included from /opt/homebrew/Cellar/llvm/22.1.4/bin/../include/c++/v1/__algorithm/make_heap.h:16:
/opt/homebrew/Cellar/llvm/22.1.4/bin/../include/c++/v1/__algorithm/sift_down.h:49:39: error: no matching function for call to object of type 'std::less<boost::tuples::tuple<nest::Source,
      nest::bernoulli_synapse<nest::TargetIdentifierPtrRport>>>'
   49 |   } else if ((__child + 1) < __len && __comp(__first[__child], __first[__child + 1])) {
      |                                       ^~~~~~
...
```
and similar or other synapse models. Building works fine with Clang 21.1.8 from Homebrew and Apple Clang 21.0.0.

The issue seems related to our connection sorting algorithm at https://github.com/nest/nest-simulator/blob/97cf79d249af6b89eeca19bd62c36c96ef64c4ab/libnestutil/sort.h#L180, which uses iterator pairs (https://github.com/nest/nest-simulator/blob/97cf79d249af6b89eeca19bd62c36c96ef64c4ab/libnestutil/iterator_pair.h#L92) to sort one array based on the order in another array.

A first look at the change log for Clang 22 does not reveal any immediate explanation (https://releases.llvm.org/22.1.0/tools/clang/docs/ReleaseNotes.htm). Building NEST 3.9 (which still uses `-std=C++17`) fails in the same way. I also tried the "compatibility flags" mentioned in https://releases.llvm.org/22.1.0/projects/libcxx/docs/ReleaseNotes.html#abi-affecting-changes, which did not resolve the problem.

med-ayssar:
Fixes #3813

### DIFF
--- a/libnestutil/iterator_pair.h
+++ b/libnestutil/iterator_pair.h
@@ -96,8 +96,17 @@ class IteratorPair : public boost::iterator_facade< IteratorPair< sort_iter_type
                        typename iterator_pair_types< sort_iter_type_, perm_iter_type_ >::difference_type >
 {
 public:
+  using ReferenceType = iterator_pair_types< sort_iter_type_, perm_iter_type_ >::ref_type;
+  using DifferenceType = iterator_pair_types< sort_iter_type_, perm_iter_type_ >::difference_type;
+
   IteratorPair() = default;
   IteratorPair( sort_iter_type_, perm_iter_type_ );
+
+  ReferenceType
+  operator[]( DifferenceType n ) const
+  {
+    return *( *this + n );
+  }
 
 private:
   friend class boost::iterator_core_access;


heplesser (approved):
@med-ayssar Thank you for the fix, looks good and I confirmed that it works with Clang 22.1.4 under macOS 26.4.1! Also works for #3544 with LTO, @JanVogelsang.

JanVogelsang (approved):

terhorstd:
Discussed that this only used as kernel internal type. The defined types in `using` are not a promise to the user facing API, not even used by pynest. So this is fine. :+1:

PR #3817, opened 2026-04-24
Fix warnings raised by Clang22

heplesser:
This PR addresses the following warnings raised by Clang22:

```
[  4%] Building CXX object nestkernel/CMakeFiles/nestkernel.dir/nest_time.cpp.o
.../nestkernel/nest_time.cpp:64:8: warning: implicit conversion from 'const long' to 'double' changes value from 9223372036854775807 to 9223372036854775808
      [-Wimplicit-const-int-float-conversion]
   64 |   if ( lmax < tmax * Range::TICS_PER_STEP_INV )  // step size is limiting factor
      |        ^~~~ ~
.../nestkernel/nest_time.cpp:64:15: warning: implicit conversion from 'const tic_t' (aka 'const long long') to 'double' changes value from 9223372036854775807 to
      9223372036854775808 [-Wimplicit-const-int-float-conversion]
   64 |   if ( lmax < tmax * Range::TICS_PER_STEP_INV )  // step size is limiting factor

.../models/iaf_cond_alpha_mc.h:353:58: warning: arithmetic between different enumeration types ('nest::iaf_cond_alpha_mc::State_::StateVecElems_' and
      'nest::iaf_cond_alpha_mc::Compartments_') is deprecated [-Wdeprecated-enum-enum-conversion]
  353 |     static const size_t STATE_VEC_SIZE = STATE_VEC_COMPS * NCOMP;
```

Defining `to_underlying()` in `nest_types.h` is maybe not perfect, but I didn't want to create another header file for such a simple function. And it will disappear again when we move to C++23, because then we get `std::to_underlying()`.


JanVogelsang (approved):

terhorstd (approved):
LGTM :+1:

PR #3821, opened 2026-04-29
Ensure NEST can be built without Boost

heplesser:
This PR ensures that NEST can be built if Boost is not available or deselected `-Dwith-boost=OFF`. It also improves some function names.
As before, connection sorting falls back on a hand-written quicksort instead of spreadsort when Boost is missing.
The change leads to slightly less pretty type names displayed in error messages:

- With Boost: `TypeMismatch: Expected datatype: Failed to cast 'V_m' from std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> to type double.`
- Without Boost: `TypeMismatch: Expected datatype: Failed to cast 'V_m' from NSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEE to type double.`

### DIFF
--- a/libnestutil/dictionary.cpp
+++ b/libnestutil/dictionary.cpp
@@ -79,7 +79,7 @@ double
 dictionary_::cast_value_< double >( const any_type& value, const std::string& key ) const
 {
   return std::visit(
-    [ &key, debug_name = debug_type( value ) ]( auto&& arg ) -> double
+    [ &key, debug_name = get_typename( value ) ]( auto&& arg ) -> double
     {
       using T = std::decay_t< decltype( arg ) >;
       if constexpr ( std::is_arithmetic_v< T > )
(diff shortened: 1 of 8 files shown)


akorgor (approved):
The changes look good to me!

I am just wondering: Is it intended that all cpp tests are skipped during `make installcheck` when NEST was compiled with `-Dwith-boost=OFF`?

```
Phase 8: Running C++ tests (experimental)
-----------------------------------------
  Not running C++ tests because NEST was compiled without Boost.


----------------------------------------------------------------
NEST Testsuite Results
----------------------------------------------------------------
Phase              Tests   Skipped  Failures    Errors      Time
----------------------------------------------------------------
07 pynesttests      4608        23         0         0     122.1
----------------------------------------------------------------
Total               4608        23         0         0     122.1
----------------------------------------------------------------

THE NEST TESTSUITE DISCOVERED PROBLEMS
    The following test phases did not report any results:
    | 08 cpptests
```

heplesser:
> I am just wondering: Is it intended that all cpp tests are skipped during `make installcheck` when NEST was compiled with `-Dwith-boost=OFF`?

Yes—we use Boost as our test framwork for CPP tests.

terhorstd:
@heplesser could you elaborate, why we should not release 3.10 without this? Which test should have failed in our test-suite that is fixed by this?

terhorstd (approved):
Was discussed in the ONDVC, all questions clarified.
LGTM :+1:

PR #3822, opened 2026-04-30
Reinstate get/set_verbosity for backward compatiblity

heplesser:
Might not be the tidies implementation, but it works.

Single reviewer should suffice, since it is just for backward compatibility.


terhorstd (approved):
Not performance critical, deprecation anounced, implementation could be more robust, yes, but should suffice to keep existing scripts running. :+1:
Thanks for the quick fix.

PR #3823, opened 2026-05-04
Test on CI without Boost and GSL and adjust tests where necessary

heplesser:
This PR re-instates the two most minimal test cases we had in the testsuite. They were in the past always built without Python and therefore "fell off" when PyNEST-NG was merged—no more need to test without Python. But we overlooked then that this would remove the no-boost test case.

Once I added the cases without Boost or GSL, a number of test failed because they did not skip tests properly in the absence of GSL (or avoided models missing if GSL is missing). This is fixed here. Furthermore, when built without Boost, the testsuite summary no longer looks for Cpptests, which require Boost's unit testing framework.

### DIFF
--- a/testsuite/pytests/other/test_status.py
+++ b/testsuite/pytests/other/test_status.py
@@ -30,6 +30,8 @@ class StatusTestCase(unittest.TestCase):
     def test_kernel_attributes(self):
         """Test nest attribute access of kernel attributes"""
 
+        self.maxDiff = None
+
         nest.ResetKernel()
 
         # Remove entry containing numpy arrays from status dicts since they do not compare well
--- a/testsuite/pytests/other/test_status.py
+++ b/testsuite/pytests/other/test_status.py
@@ -30,15 +30,17 @@ class StatusTestCase(unittest.TestCase):
     def test_kernel_attributes(self):
         """Test nest attribute access of kernel attributes"""
 
-        self.maxDiff = None
-
         nest.ResetKernel()
 
-        # Remove entry containing numpy arrays from status dicts since they do not compare well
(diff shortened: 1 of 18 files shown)


testsuite/pytests/other/test_status.py
```diff
@@ -30,6 +30,8 @@ class StatusTestCase(unittest.TestCase):
     def test_kernel_attributes(self):
         """Test nest attribute access of kernel attributes"""
 
+        self.maxDiff = None
+
```
terhorstd: This seems random and unused. If this is important for something non-obvious, please add a comment.
```suggestion
```

terhorstd (changes requested):
Thanks @heplesser!
Found some seemingly spurious code, but then it looks fine to me. Bit more extensive changes than I'd expect, but no need to disentangle into separate PRs, imho.

gtrensch (approved):
LGTM 👍

testsuite/pytests/other/test_status.py
heplesser: This setting ensured that if gks_result and ks_result differed, the full dictionaries would be printed so one could find where they differed. This occurred sometimes when memory size changed slightly between the two calls fetching the kernel status. Such a difference is expected behavior, so I now added code removing `memory_size` from the dictionaries before comparison. I also commented this better.

terhorstd (approved):
Perfect, thanks for the fast response! 🚀

testsuite/pytests/other/test_status.py
terhorstd: Perfect, thanks for the fast response! 🚀

PR #3827, opened 2026-05-05
Restructure scikit-build version metadata configuration

steffengraber:
- Moved `tool.scikit-build.metadata.version` settings to inline `tool.scikit-build` section
- Updated version regex pattern
- Version will still be read from `VERSION` file
- No functional changes to version detection logic

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,13 +120,16 @@ Documentation = "https://nest-simulator.readthedocs.io/en/stable/"
 "Bug Tracker" = "https://github.com/nest/nest-simulator/issues"
 Changelog = "https://nest-simulator.readthedocs.io/en/stable/whats_new/index.html"
 
+
+[tool.scikit-build.metadata.version]
+provider = "scikit_build_core.metadata.regex"
+input = "VERSION"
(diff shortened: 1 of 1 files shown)


terhorstd (changes requested):
Thanks for taking care of this section.
* Could you provide a reason why this change is necessary?
* If the section is split off, please move it _under_ the `[tool.scikit-build]` section, not above.

steffengraber:
We don't have to split it into two sections. Splitting creates visual separation and clearer meaning for the purpose of the sections.  Both TOML forms are functionally equivalent.

The regex improvement, '^(?P<version>.*)'\s*$, is not strictly necessary for basic functionality, but it is more robust.

- `^` anchors the match to the start of the line and $ anchors it to the end, preventing partial matches.
- `\s*$` strips any trailing whitespace (spaces and tabs) that might exist in the VERSION file.

However, moving the section under [tool.scikit-build] would make sense in terms of consistency.

terhorstd (approved):
Thanks for the fix!
LGTM :+1:

heplesser:
Package-building related, therefore merging based on single review. Failure on macOS due to SciPy regression, see #3881.

PR #3828, opened 2026-05-07
Clarify installation instructions for mamba

clinssen:
With thanks to @abigailm for the suggestions!


terhorstd (approved):
Good catch with the typo! It also reads much cleaner now. Thanks! 👍

terhorstd:
For doc changes, one reviewer is conventionally sufficient. Thanks for putting this in.

PR #3829, opened 2026-05-11
Restore GetStatus backward compatibility with NEST 3.9 for plain output

heplesser:
This PR ensures that `nest.GetStatus(n)` returns results in the same format as in NEST 3.9 for backward compatibility, both for normal output (Python data structures) and for JSON-serialized output. Pandas-output has never been supported for `GetStatus()`, so this needs no attention. It also provides a test for this, which additionally checks for equivalent results also when using `n.get()`.  Various combinations of keys are checked.

AI Assistance: light
Model: Claude Sonnet 4.6
Details: Helper functions `convert_to_lists()` and `remove_keys()` generated by AI.


terhorstd (approved):
For the purpose of pushing the release this is okay. We should schedule some time discussing a proper data format and typing, though.

testsuite/pytests/other/test_getstatus_39_compatibility.py
```diff
@@ -0,0 +1,497 @@
...
+        rf.close()
+
+
+def convert_to_lists(obj):
+    """
+    Convert to pure Python types to allow json serialization.
+
+    Skip values that may change for good reason, currently only model_id.
+
+    Created based on suggestions by Claude Sonnet 4.6.
```
akorgor:
```suggestion
```
I do not think that this information is useful for anyone reading the docstring. If at all I would put into the commit message or add an according tag to the PR.

testsuite/pytests/other/test_getstatus_39_compatibility.py
```diff
@@ -0,0 +1,497 @@
...
+    else:
+        return obj
+
+
+def remove_keys(obj):
+    """
+    Skip values that may change for good reason, currently only model_id.
+
+    Created based on suggestions by Claude Sonnet 4.6.
+    """
```
akorgor, line 98:
```suggestion
```
See above.

testsuite/pytests/other/test_getstatus_39_compatibility.py
```diff
@@ -0,0 +1,497 @@
...
+        rf = open("test_getstatus_39_reference.py", "w")
+        rf.write("""def reference():\n    return {\n""")
+    else:
+        rf = None
+    yield rf
+
+    # teardown
+    if rf:
+        rf.write("""}\n""")
+        rf.close()
```
akorgor, line 75: Are you expecting the reference data to change frequently? Otherwise, I would suggest keeping reference data generation separate from this test. Generating a Python file while running the test suite seems risky, both because it creates executable code and because it makes the test run stateful. This workflow may also unexpectedly leave the generated file in the working tree, where it can appear in `git status` and be accidentally committed unless it is ignored by `.gitignore`. It seems safer to me for tests to be read-only and static, with a separate script or the test running in a different mode generating reference data. If regeneration is needed, writing deterministic JSON to a checked-in `.json` file, or printing it with `json.dumps` to standard output, would be safer.

akorgor (approved):
I approve the PR given the time-sensitivity of this patch. However, the issues in the comments should be addressed at a later point.

testsuite/pytests/other/test_getstatus_39_compatibility.py
heplesser: I think this suggestion was meant to remove line 101, not 102. I will make that change in a separate commit.

testsuite/pytests/other/test_getstatus_39_compatibility.py
heplesser: Thanks for raising a good point. We should discuss this at a later stage for handling corresponding cases. In this case, I do not expect any changes to the reference data, since it represents the status per NEST 3.9 (although one might want to add additional cases (requested keys) later). I wanted to use the same code for generating the reference data with 3.9 as is used for testing, to make sure we really compare apples with apples.

heplesser:
> For the purpose of pushing the release this is okay. We should schedule some time discussing a proper data format and typing, though.

Not entirely sure which data format and typing you mean precisely, but I agree that a revision is needed, also for the hierarchical indexing supported by `get()`. But this would break user code, so I think that is a matter for NEST 4.

heplesser:
@terhorstd I have applied the suggestions by @akorgor and will merge once all tests are through.

PR #3830, opened 2026-05-15
Introduce more intuitive parameterization of surrogate gradients

akorgor:
This PR improves consistency across the surrogate gradient implementations and increases clarity for the user. While before the surrogate gradient was parametrized with `gamma` and `beta`, now it is parameterized with `surrogate_gradient_height` and `surrogate_gradient_width`, where the latter corresponds to `1/beta`. Additionally,  the `arctan` surrogate function was renamed to `arctan_derivative` and it incorporates $\pi$ into the height and width parameter.


heplesser (approved):
@akorgor Good from my side, thanks!

JesusEV (approved):
Very nice! These changes definitely make surrogate gradients usage much more transparent

heplesser (changes requested):
@akorgor Just one tiny request before merging: Can you modify the title of the PR to be more informative about what changes, rather than just say "fix"? And create a little follow up to remind us to add information about the variable name changes to "What's new" for 3.11?

akorgor:
> @akorgor Just one tiny request before merging: Can you modify the title of the PR to be more informative about what changes, rather than just say "fix"?

Done.

> And create a little follow up to remind us to add information about the variable name changes to "What's new" for 3.11?

I have already contacted @jessica-mitchell offline and also it has the tag `I: User Interface` so this is probably already on the radar for the "What's new". Let me know if you still want me to do anything further.

heplesser (approved):

PR #3831, opened 2026-05-19
Change dependabot versioning-strategy for Python packages

terhorstd:
This changes the behavior of dependabot to stop suggesting to increase lower version bounds (such as #3825 and similar). To keep NEST compatible with various ecosystems, the lower version bounds are only to be increased when new functionality is used that requires the stricter requirement.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,6 @@ updates:
     directory: /doc
     schedule:
       interval: daily
+    # package version bound updates
+    # see https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--
+    versioning-strategy: "widen"


terhorstd:
Also should be considered in other NEST projects, such as in context of https://github.com/nest/nestml/pull/1273

terhorstd:
@heplesser, @akorgor, could you have a quick look in the near future? gtrensch seems unavailable at the moment. I'm very annoyed by the default behavior of dependabot and am already tempted to just merge this without review… :-/
As soon as one of you agrees, you can also hit 'merge' already, so we don't collect more closed PRs to bother us next release…

akorgor (approved):
If I understand this PR correctly, it makes Dependabot widen the Python dependency version ranges in `/doc`, allowing newer versions without raising the minimum supported version. My understanding is that this does not affect libraries, which already default to `widen`, but rather apps, such as Sphinx, which default to `increase`. Unlike `widen`, `increase` always raises the minimum version to match the new version.

I would personally drop the link and the comment, since the keyword is easy to find in the GitHub docs via search.

akorgor:
> As soon as one of you agrees, you can also hit 'merge' already

Unfortunately, hitting the merge button exceeds my permissions. However, a second pair of eyes would be anyway beneficial because of my superficial understanding of this matter.

heplesser (approved):

terhorstd:
Thanks for the quick resolve

PR #3846, opened 2026-05-26
Update Sphinx extension and JSON for version numbers  in docs

jessica-mitchell:
this pr improves the extension so it handles failures better and  reads the json in at build time

- added a line to the docs on how to use the version role

updates to the versions.json file 
- based mainly on the doc/requirements.txt - since this is where a minimum is pinned for dependencies;
- removed sli based dependencies
- 
under python, the doc_build refers to the .readthedocs python version requirement, and the wheel_min/wheel_max is for pip installation

to discuss

should all dependencies be included in this file? should we pin versions in other places?


terhorstd:
Taking the version bounds from Python files is probably safest, i.e. `requirements.txt`, or better `pyproject.toml`. Ideally I would like to have the `versions.json` only used for building the strings in the documentation, and eventually superseded by directly reading from an authoritative source. For now we don't have that, but we're slowly getting there. However, at some point we still need to discuss the different requirement types (sources, python, wheel, docs, testing, …) and how to implement the requirements definition. It may not be possible to express all complexities the way we need in too simple formats such as the `requirements.txt` files.

terhorstd (commented):
The comments of the inner functions got dropped. They can be quite helpful for people unfamiliar with the workings of sphinx. On top of that, it should really get picked up by our code quality checks...

jessica-mitchell:
> The comments of the inner functions got dropped. They can be quite helpful for people unfamiliar with the workings of sphinx. On top of that, it should really get picked up by our code quality checks...

@terhorstd  yes good point, I can rea-add comments 
but I am surprised that no other checks got triggered- did something change?

Issue #3731, opened 2026-01-20
Tidy up dependency management


terhorstd:
Currently dependencies of NEST are tracked in various locations:
* At the base of the repository [`/requirements*.txt`](https://github.com/nest/nest-simulator/) for source installs (dev, use, doc)
* For documentation we have the [`:version:` role definitions](https://github.com/nest/nest-simulator/blob/master/doc/htmldoc/_ext/versions.json) (dev, doc)
*  the [EBRAINS Software Distribution (ESD) `nest/package.py`](https://gitlab.ebrains.eu/ri/tech-hub/platform/esd/ebrains-spack-builds/-/blob/master/packages/nest/package.py?ref_type=heads) / [repo](https://gitlab.ebrains.eu/ri/tech-hub/platform/esd/ebrains-spack-builds/-/tree/master/packages/nest?ref_type=heads) (use, hpc, cloud)
* Conda → [conda-forge](https://packages.fedoraproject.org/pkgs/nest/nest/) / [repo](https://github.com/conda-forge/nest-simulator-feedstock/) (use, laptop)
* Fedora →[nest package](https://packages.fedoraproject.org/pkgs/nest/nest/) / [repo](https://src.fedoraproject.org/rpms/nest/tree/rawhide) (use, laptop)
* APT → [launchpad](https://launchpad.net/~nest-simulator/+archive/ubuntu/nest) / [recipes](https://code.launchpad.net/~nest-simulator/+recipes) / [+git/nestml repo](https://git.launchpad.net/~nest-simulator/nest-simulator/+git/nestml/tree/) (use, laptop, cluster(?))
  * Snap → [launchpad](https://code.launchpad.net/~nest-simulator/+snaps)
* pip → *blocked* – waiting for [pynest](https://github.com/orgs/nest/projects/12) / [repo](https://github.com/heplesser/nest-simulator/tree/pynest-ng), generated [setup.py](https://github.com/heplesser/nest-simulator/tree/pynest-ng) (use, laptop)

The most complete and flexible definition is currently in ESD.
* dependency trees differ for each environment, installation goals and required feature sets (develop, doc, test, hpc, container, laptop, …)
* namespaces differ between packaging systems making automatic translation difficult.
* not all features are supported by the different systems (version ranges, feature flags, etc.)
* Currently we maintain only a subset of the available package definitions directly within the NEST teams.

[...]

PR #3859, opened 2026-05-26
Hike Cython version requirement to >=3.0.0

heplesser:
This PR hikes the Cython version requirement to 3.0.0 (released July 2023) to support the Python assignment operator (`: =`) and scoped C++ `enum class`.

Tested on Ubuntu 24.04 with gcc 13.3 and Cython 3.0.0.


terhorstd (approved):
lgtm 👍

terhorstd:
The multitude of files that need changing for this single dependency is worrying. It's a symptom of #3731.

steffengraber (approved):
Looks good to me! I tested it for pip-packaging.

Issue #3737, opened 2026-02-11
Tidy up "pip install → cmake" vs. "cmake → pip install"


terhorstd:
Currently the [pynest-ng](https://github.com/heplesser/nest-simulator/tree/pynest-ng) version of the Python packaging includes a configure (`cmake`) and compile (`make && make install`) step. This requires all possible configuration to be available on the python packaging side and internally another translation to the cmake flags.

We should investigate the possibility to configure the Python package via cmake, such that it automatically has the right configuration (that cmake knows and compiles nest with). Once configured, the `pynest/` directory would be a fully configured pip installable Python package and wheel building could directly use the built binaries (custom or generic, depending on the build environment).

jessica-mitchell:
Decision after long discussion, `cmake` should create the Python package

PR #3860, opened 2026-05-27
Accept plain venv as Python context

terhorstd:
Formerly only Conda was hard-coded to be checked for virtual environment paths. This adds the `VIRTUAL_ENV` variable to be checked, too. Neither is the perfect approach and the Python install mechanism needs to be cleaned up (#3737), for now this allows to install with plain `venv`/`virtualenv` setup.


heplesser (approved):

steffengraber (approved):
LGTM :+1:
VIRTUAL_ENV is not only found in a venv environment, but also in poetry, virtualenv and uv.

PR #3868, opened 2026-06-01
Enable distributed parameter values for `EpropArchivingNode`, `FlushEventMechanism`, `IgnoreAndSpikeMechanism`

akorgor:
This PR enables setting distributed parameter values by using  `update_value_param()` instead of `update_value()`.

### DIFF
--- a/models/parrot_neuron.cpp
+++ b/models/parrot_neuron.cpp
@@ -97,7 +97,7 @@ void
 parrot_neuron::set_status( const Dictionary& d )
 {
   ArchivingNode::set_status( d );
-  FlushEventMechanism::set_status( d );
+  FlushEventMechanism::set_status( d, this );
 }
 
 void
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -265,7 +265,7 @@ nest::ArchivingNode::set_status( const Dictionary& d )
     clear_history();
   }
 
-  IgnoreAndSpikeMechanism::set_status( d );
+  IgnoreAndSpikeMechanism::set_status( d, this );
 }
 
 void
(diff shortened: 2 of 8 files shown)


heplesser (changes requested):
Thanks for this PR. For offsets and intervals, supporting distributed values makes sense. But turning history shifting, flushing or ignore-and-spike on or off really changes the neuron model and I find it very difficult to see any sensible use case for randomly turning some neurons into ignore-and-spike neurons. I would therefore allow setting those parameters with fixed values only. Actually, I'd almost want to allow setting them only on model defaults to encourage users to use them only via a `CopyModel()` operation, but that we can leave for later.

akorgor:
> Thanks for this PR. For offsets and intervals, supporting distributed values makes sense. But turning history shifting, flushing or ignore-and-spike on or off really changes the neuron model and I find it very difficult to see any sensible use case for randomly turning some neurons into ignore-and-spike neurons. I would therefore allow setting those parameters with fixed values only.

For the ignore-and-spike bool this is now fixed in https://github.com/nest/nest-simulator/pull/3868/commits/48cc00886843169b5540bb213f9c851c84293e53. Note that none of the parameters turns the history shifting on or off, the double `eprop_isi_trace_cutoff` just determines after how many steps the computation of the weight update is cut off. Also, there is no bool to turn the flush event mechanism on or off, instead it is controlled solely by the double `flush_event_send_interval`. But I can disallow those, too - currently I only have a use case for the ignore-and-spike parameters.

heplesser:
@akorgor Thank you for the clarification. I guess one could in principle foresee a use case for distributed flush intervals, essentially a kind of random forgetfulness of neurons, so allowing the flush interval to be drawn from a distribution makes sense or is at least justifiable.

heplesser (approved):

tomtetzlaff (approved):
Looks good to me. At least I can confirm that this code 

```
import nest

nrn = nest.Create("iaf_psc_delta", n=10, params = dict(ignore_and_spike=True))

nrn.set(ignore_and_spike_offset = nest.random.uniform(min = 1.0, max = 10.), ignore_and_spike_interval=nest.random.uniform(min = 10.0, max = 100.))  
print(nrn.get('ignore_and_spike_offset'))
print(nrn.get('ignore_and_spike_interval'))

nest.Simulate(1000)
```

runs without error and gives reasonable numbers.

PR #3881, opened 2026-06-22
Exclude SciPy version 1.18.0 due to regression

heplesser:
SciPy 1.18.0 has a regression that breaks some of our tests. The problem is known on the SciPy side (https://github.com/scipy/scipy/issues/25448) and they have a PR for it (https://github.com/scipy/scipy/pull/25468), so I assume it will be fixed with SciPy 1.18.1 and exclude only this specific version.

Labeling as critical as it breaks the CI.


terhorstd (approved):
lgtm :+1:

steffengraber (approved):
👍

PR #3882, opened 2026-06-22
Fix mailing list URLs in documentation return HTTP 400

Gasterbuzzer:
**Describe the flaw**

The mailing list links in the documentation point to outdated URLs under nest-initiative.org, which return an HTTP 400 error.

**Affected Files**

`doc/htmldoc/developer_space/guidelines/mailing_list_guidelines.rst`

**Changes**

Replaced outdated mailing list URLs:

From: 
 - https://www.nest-initiative.org/mailinglist/
 - https://www.nest-initiative.org/mailinglist/hyperkitty/list/users@nest-simulator.org/

To: 
 - https://www.nest-simulator.org/mailinglist/
 - https://www.nest-simulator.org/mailinglist/hyperkitty/list/users@nest-simulator.org/

**Supporting materials**
grafik

PR #3883, opened 2026-06-22
Add link checker to CI

clinssen:
See #3882.


jessica-mitchell (changes requested):
thanks for getting on this @clinssen 

can we pin lychee to a full commit SHA like we do in other checks? also include the harden runner

we also should include link_checker in the "needs" list for the linux and macos builds, similar to the other static checks

clinssen:
> can we pin lychee to a full commit SHA like we do in other checks? also include the harden runner

done! The "harden runner" is already fixed to a SHA hash.

> we also should include link_checker in the "needs" list for the linux and macos builds, similar to the other static checks

We decided against this for NESTML because there are often fluke 404s on zenodo.org and doi.org, which go away when you re-run the stage, and broken links often arise spontaneously as the result of an outside change, and the fixes are thus typically spun off to different PRs. However, I'm happy to include it here as a hard dependency.

jessica-mitchell:
> > can we pin lychee to a full commit SHA like we do in other checks? also include the harden runner
> 
> done! The "harden runner" is already fixed to a SHA hash.

 
> > we also should include link_checker in the "needs" list for the linux and macos builds, similar to the other static checks
> 
> We decided against this for NESTML because there are often fluke 404s on zenodo.org and doi.org, which go away when you re-run the stage, and broken links often arise spontaneously as the result of an outside change, and the fixes are thus typically spun off to different PRs. However, I'm happy to include it here as a hard dependency.


I think for all jobs we include the stepsecurity; I am not sure how strict we are on including all static checks for the linux and macos builds but @terhorstd should weigh in on these topics

.github/workflows/nestbuildmatrix.yml
```diff
@@ -254,18 +254,13 @@ jobs:
...
-          tar -xzf 'lychee-v0.8.2-x86_64-unknown-linux-gnu.tar.gz'
-
-      - name: "Run lychee..."
-        continue-on-error: true
-        run: "pwd; ls -l; ./lychee build_support/ doc/{*.txt,*.md,htmldoc} examples/ lib* models/ nest* pynest/ testsuite/ *.md"
-        # TODO: doc/auto_examples and doc/models need to be added here, but exists only after a `make html`
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # v2.8.0
+        with:
+          args: build_support/ doc/{*.txt,*.md,htmldoc} examples/ lib* models/ nest* pynest/ testsuite/ *.md
```
terhorstd: How do the `args` work? Are they recursive for directories? If so, why is the `doc/{*.txt,*.md,htmldoc}` needed explicitly instead of just `doc/`? Or even more generally, we might consider `.` and adding exceptions to the `exclude_path` config. I'm a bit worried having the long hard-coded list here that would default to *not* checking anything that moved or new dirs.

terhorstd: Additionally the `htmldoc` may need a `**.rst` kind of syntax...

lychee.toml
```diff
@@ -73,4 +73,4 @@ exclude_link_local = true
 exclude_loopback = true
 
 # Exclude all mail addresses from checking
-exclude_mail = true
+include_mail = false
```
terhorstd, line 82: How does checking mail addresses work? What does it do to find out?

terhorstd (changes requested):
Some things are not clear to me...

terhorstd:
> I think for all jobs we include the stepsecurity; I am not sure how strict we are on including all static checks for the linux and macos builds but @terhorstd should weigh in on these topics

Stepsecurity is already in, that should be fine. For making the builds dependent on the link checks, I am more thinking about the purpose of this CI: We're interested to see if the code and docs are as they should be, so external deps being down shouldn't stop our development. However, the links being broken should show up somewhere. Maybe a combination of `needs` and [`continue-on-error`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error) could show us an exclamation mark without stopping the more interesting builds. I don't have experience on github with this, but GitLab can do this also specific to a dependency, i.e. on the [`needs`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds) level not above.

.github/workflows/nestbuildmatrix.yml
jessica-mitchell: I ran the Sphinx linkchecker against NEST which explictly checks the docs and found many "errors"  :  most were 403 that actually are working links, and a few were unusual catches in local references, that also work. In the end there were less than 10 (some already resolved in open PRs) broken links.  

So I'd like to see how lychee does in comparison, because it doesnt seem to be checking rst files

lychee.toml
clinssen: There is some MX record/mail server magic (the equivalent of a ping: lychee connects to the mail server, then starts sending a message and checks the response, but then doesn't actually send any message).

clinssen:
I added it under "needs" of the sphinx-rtd build stage.

``continue-on-error`` causes the stage to show up with a green checkmark even on failure, so I guess that's not what we want.

.github/workflows/nestbuildmatrix.yml
clinssen: I changed this so it now checks all files.

requirements_nest_server.txt
```diff
@@ -4,7 +4,7 @@
 # It is meant for those who would like to compile NEST or build NEST
 # documentation themselves. If you just want to execute NEST, you should
 # install NEST directly as described in
-# https://www.nest-simulator.org/installation.
+# https://nest-simulator.readthedocs.io/en/stable/installation/index.html.
```
jessica-mitchell, line 7:
```suggestion
#  https://www.nest-simulator.org/documentation/installation.
```

requirements_pynest.txt
```diff
@@ -4,7 +4,7 @@
 # with PyNEST. It is meant for those who would like to compile NEST
 # or build NEST documentation themselves. If you just want to execute
 # NEST, you should install NEST directly as described in
-# https://www.nest-simulator.org/installation.
+# https://nest-simulator.readthedocs.io/en/stable/installation/index.html.
```
jessica-mitchell, line 7:
```suggestion
#  https://www.nest-simulator.org/documentation/installation.
```

requirements_testing.txt
```diff
@@ -4,7 +4,7 @@
 # testsuite. It is meant for those who would like to compile NEST
 # or build NEST documentation themselves. If you just want to execute
 # NEST, you should install NEST directly as described in
-# https://www.nest-simulator.org/installation.
+# https://nest-simulator.readthedocs.io/en/stable/installation/index.html.
```
jessica-mitchell, line 7:
```suggestion
# https://www.nest-simulator.org/documentation/installation.
```

jessica-mitchell (changes requested):

jessica-mitchell:
@clinssen I'm testing lychee locally - and I think we need to specify  `doc/htmldoc/**/*.rst` and accept 403 (along with 200 300x) because many of the doi links get 403s but work

update: my options that seem to output something reasonable (just testing docs) is --accept '200,301,302,307,308,403' --exclude-loopback --include-verbatim  '**/*.rst'
Screenshot from 2026-06-25 10-49-56

.github/workflows/nestbuildmatrix.yml
jessica-mitchell: > I changed this so it now checks all files.

We need to explictly include rst files; it's the only way i've gotten lychee to check them

clinssen:
Hi, I cleaned things up a bit (especially the file discovery, added caching, excluded gnu.org as it seems to error out by an unreasonable amount); this uncovered many more 404s that I'm afraid will also need to be fixed one-by-one by hand.

PR #3889, opened 2026-06-23
Make summarization robust against lack entirely missing results

heplesser:
`max([])` does not work, while `max([], default=0)` will return zero, which makes sense in this case.

See https://github.com/terhorstd/nest-simulator/actions/runs/28035709820/job/82990320660#step:32:253 for an example where his was needed.

A single reviewer should suffice here.


terhorstd (approved):
LGTM :+1:

PR #3890, opened 2026-06-24
Add note in iaf_cond* models about setting inhibitory and excitatory synapses

### LINKED ISSUE #3615: DOC How to set excitatory and inhibitory synapses in conductance models
After much research, I cannot figure out how to set a connection to a conductance-based neuron to be excitatory or inhibitory. I am opening this issue to ask how this is done, and to suggest that a clear example is added to the documentation/examples, since this is a very basic question.

My understanding of conductance-based neurons like `iaf_cond_exp` is that they allow to use two different types of synapses, excitatory and inhibitory, each with their own conductance and reversal potential. I understand that these two different synapses are already implemented in the base model,  from this page: https://nest-simulator.readthedocs.io/en/v3.9/models/iaf_cond_exp.html

Thus, excitatory neurons should be connected using excitatory synapses, and inhibitory neurons using inhibitory synapses. 

However, how can I set the type of synapse used in a connection? I thought that this could be done using the `"receptor_type"` field in `syn_spec` dictionary (as described here https://nest-simulator.readthedocs.io/en/v3.7/tutorials/pynest_tutorial/part_1_neurons_and_simple_neural_networks.html#synapse). However, setting this input is not working for me:

```python

import nest

nest.ResetKernel()
n = nest.Create("iaf_cond_exp")

pg_e = nest.Create("poisson_generator", params={"rate": 80.0})
pg_i = nest.Create("poisson_generator", params={"rate": 80.0})

# excitatory to EXC port 0
nest.Connect(pg_e, n, syn_spec={
    "synapse_model": "static_synapse",
    "receptor_type": 0,   # EXC
    "weight": 1,
    "delay": 1.0,
})

# inhibitory to INH port 1
nest.Connect(pg_i, n, syn_spec={
    "synapse_model": "static_synapse",
    "receptor_type": 1,   # INH
    "weight": 1,
    "delay": 1.0,
})
```
```
NESTErrors.UnknownReceptorType: UnknownReceptorType in SLI function Connect_g_g_D_D: Receptor type 1 is not available in iaf_cond_exp.
```

I looked through the tutorials and the examples in the documentation, but I couldn't locate any instance where excitatory and inhibitory neurons are used in a conductance-based network. I might have missed an example since I didn't look exhaustively, but I feel like this shouldn't be that difficult to find. 

In sum, how would I implement the intended model above? Can the documentation be improved to make this easier to find?

jessica-mitchell:
This PR takes the explanation from the issue about how iaf_cond models set excitatory and inhibitory synapsaes and puts it in the model and synapse_spec docs

closes #3615

### DIFF
--- a/doc/htmldoc/synapses/synapse_specification.rst
+++ b/doc/htmldoc/synapses/synapse_specification.rst
@@ -440,6 +440,33 @@ shows the setup and connection of such a model in more detail:
 
     nest.Connect(A, B, syn_spec={'receptor_type': 2})
 
+.. note::
+
+   For single-compartment conductance-based neuron models (``iaf_cond_exp``,
+   ``iaf_cond_alpha``, ``iaf_cond_beta``, ``iaf_cond_exp_sfa_rr``), ``receptor_type``
+   is **not** used to select the excitatory or inhibitory synapse. Instead, the **sign of
+   the connection weight** determines the target synapse: a positive weight connects to
+   the excitatory synapse (``E_ex``), while a negative weight connects to the inhibitory
+   synapse (``E_in``). Only the absolute value of the weight determines the conductance
+   amplitude. Attempting to set ``receptor_type`` to a value other than 0 for these models
+   raises an ``UnknownReceptorType`` error at runtime.
+
+   .. code-block:: python
+
+       n = nest.Create('iaf_cond_exp')
+       pg_e = nest.Create('poisson_generator', params={'rate': 80.0})
+       pg_i = nest.Create('poisson_generator', params={'rate': 80.0})
+
+       # Positive weight → excitatory synapse (E_ex, tau_syn_ex)
+       nest.Connect(pg_e, n, syn_spec={'weight': 1.0, 'delay': 1.0})
+
+       # Negative weight → inhibitory synapse (E_in, tau_syn_in)
+       nest.Connect(pg_i, n, syn_spec={'weight': -1.0, 'delay': 1.0})
+
+   This convention is shared by all ``iaf_cond_*`` models and mirrors the sign convention
+   used in the ``iaf_psc_*`` current-based models, making it straightforward to switch
+   between the two families.
+
 
 .. _synapse-types:
 
--- a/doc/htmldoc/synapses/synapse_specification.rst
+++ b/doc/htmldoc/synapses/synapse_specification.rst
@@ -442,14 +442,14 @@ shows the setup and connection of such a model in more detail:
 
 .. note::
 
-   For single-compartment conductance-based neuron models (``iaf_cond_exp``,
-   ``iaf_cond_alpha``, ``iaf_cond_beta``, ``iaf_cond_exp_sfa_rr``), ``receptor_type``
-   is **not** used to select the excitatory or inhibitory synapse. Instead, the **sign of
+   Many single-compartment conductance-based neuron models (``iaf_cond_exp``,
+   ``iaf_cond_alpha``, ``iaf_cond_beta``, ``iaf_cond_exp_sfa_rr``) do **not** use
+   ``receptor_type`` to select the excitatory or inhibitory synapse. Instead, the **sign of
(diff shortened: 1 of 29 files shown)


doc/htmldoc/synapses/synapse_specification.rst
```diff
@@ -440,6 +440,33 @@ shows the setup and connection of such a model in more detail:
...
+.. note::
+
+   For single-compartment conductance-based neuron models (``iaf_cond_exp``,
+   ``iaf_cond_alpha``, ``iaf_cond_beta``, ``iaf_cond_exp_sfa_rr``), ``receptor_type``
+   is **not** used to select the excitatory or inhibitory synapse. Instead, the **sign of
+   the connection weight** determines the target synapse: a positive weight connects to
+   the excitatory synapse (``E_ex``), while a negative weight connects to the inhibitory
+   synapse (``E_in``). Only the absolute value of the weight determines the conductance
+   amplitude. Attempting to set ``receptor_type`` to a value other than 0 for these models
+   raises an ``UnknownReceptorType`` error at runtime.
```
heplesser:
```suggestion
   Many single-compartment conductance-based neuron models (``iaf_cond_exp``,
   ``iaf_cond_alpha``, ``iaf_cond_beta``, ``iaf_cond_exp_sfa_rr``) do **not** use
   ``receptor_type`` to select the excitatory or inhibitory synapse. Instead, the **sign of
   the connection weight** determines the target synapse: a positive weight connects to
   the excitatory synapse (``E_ex``, ``tau_syn_ex``), while a negative weight connects
   to the inhibitory synapse (``E_in``, ``tau_syn_in``). The absolute value of the weight
   determines the conductance amplitude. Attempting to set ``receptor_type`` to a value
   other than 0 for these models raises an ``UnknownReceptorType`` error at runtime.
```

models/iaf_cond_alpha.h
```diff
@@ -74,6 +74,18 @@ at :math:`t = \tau_{syn}`.
 
 See also [1]_, [2]_, [3]_.
 
+.. note::
+
+   The **sign of the synaptic weight** determines which synapse receives the input: a positive
```
heplesser:
```suggestion
   The **sign of the synaptic weight** determines which synapse receives an incoming spike: a positive
```

models/iaf_cond_alpha.h
```diff
@@ -74,6 +74,18 @@ at :math:`t = \tau_{syn}`.
...
+   weight routes the spike to the excitatory synapse (governed by ``E_ex`` and ``tau_syn_ex``),
+   while a negative weight routes it to the inhibitory synapse (governed by ``E_in`` and
+   ``tau_syn_in``). Only the absolute value of the weight determines the conductance amplitude.
```
heplesser:
```suggestion
   ``tau_syn_in``). The absolute value of the weight determines the change in conductance amplitude.
```

The two suggestions here apply also the remaining models in this PR.

heplesser (changes requested):
@jessica-mitchell Looks pretty good, I just have small suggestion for textual changes.

heplesser (commented):
Thanks! I have a few more thoughts: 

1. This also applies to most `aeif_cond_*` neurons. One way to find candidates is to search for `if ( e.get_weight() > 0.0 )`.
2. Do we have a description of the corresponding `_psc_` behavior anywhere? If yes, should we connect the documentation for cond to that and only have a short reference here. If not, we should write about our sign logic somewhere in a more prominent location.
3. I'd like to keep the text in the individual neuron models shorter and include a link to the main "article" on the topic.

jessica-mitchell:
@heplesser I have asked claude to look into which models use the sign of the connection weight, and it came up with quite a number of models. Right now I have added a [section](https://nest-simulator--3890.org.readthedocs.build/en/3890/neurons/neuron_types.html#selecting-the-target-synapse) in `neuron_types` about how the different models use connection weight (which is quite verbose and may need rewording), and then those model files also get a short comment with a link to that page. I can provide more details about what claude found if you're interested

PR #3896, opened 2026-06-24
Implement efficient partial shuffling

heplesser:
This PR replaces "homemade" code for random sampling with the C++ standard library `std::sample()`.

Initial timing experiments using the structural_plasticity.py example indicate small performance difference both towards main and #3779 and some variation in results. We should perform some systematic tests to (a) ensure that all three versions generate statistically consistent results and (b) benchmark performance, especially for cases requiring random selection of larger numbers of synaptic elements.

@cjotade and @alpinangu: Could you work together on this?

This is likely a replacement for #3779.


nestkernel/random_generators.h
```diff
@@ -130,6 +130,11 @@ class BaseRandomGenerator
     NodeCollection::const_iterator last,
     std::back_insert_iterator< std::vector< NodeIDTriple > > dest,
     size_t n ) = 0;
+
+  /**
+   * @biref Return vector containing n elements randomly selected from data.
```
Magnushst:
```suggestion
   * @brief Return vector containing n elements randomly selected from data.
```

nestkernel/random_generators.h
```diff
@@ -130,6 +130,11 @@ class BaseRandomGenerator
...
+   */
```
Magnushst, line 147:
```suggestion
   *
   * @return n elements drawn from data without replacement and in random order;
   *         fewer than n if data.size() < n.
   */
```
std::sample clamps to min(n, data.size()), and the order matters to the sp_manager.cpp callers

nestkernel/random_generators.h
```diff
@@ -290,6 +295,15 @@ class RandomGenerator final : public BaseRandomGenerator
     std::sample( first, last, dest, n, rng_ );
   }
 
+  inline std::vector< size_t >
+  sample( const std::vector< size_t >& data, const size_t n ) override
+  {
+    std::vector< size_t > tmp;
+    tmp.reserve( n );
+    std::sample( data.begin(), data.end(), std::back_inserter( tmp ), n, rng_ );
+    return tmp;
```
Magnushst:
```suggestion
    const size_t m = std::min( n, data.size() );
    std::vector< size_t > tmp( data );

    // Partial Fisher-Yates: at each step select an element from the remaining range
    // [ next, tmp.size() ) and swap it to the front. Draws m random numbers and
    // leaves the selection in random order, as global_shuffle() did.
    for ( size_t next = 0, remaining = tmp.size(); next < m; ++next, --remaining )
    {
      const size_t rnd = next + ulrand( remaining );
      std::swap( tmp[ next ], tmp[ rnd ] );
    }
    tmp.resize( m );  // keep only the m selected elements.
    return tmp;
```
std::sample keeps the input order, so create_synapses loses the random pairing global_shuffle gave it; with the two element lists equally long it hands back the list unchanged and every synapse becomes an autapse. 

Same as #3779's 'partial Fisher-Yates' algorithm implementation moved into this PR's API, which fixes the order and the draw count together: same distribution as main, n draws instead of about N/2, output identical to #3779 over 50 seeds.

Magnushst:
`SPBuilder::connect_` wires the i-th free axon slot to the i-th free dendrite slot, and both lists come in node-id order, so `global_shuffle` was the only source of randomness in the pairing. `std::sample` hands its picks back in input order instead. The delete call sites don't mind; `create_synapses` does. When `Axon_ex` and `Den_ex` grow on the same neurons under the same curve the lists are the same length, `std::sample` returns them unchanged, and every neuron is wired to itself.

Four builds of v3.10 from one source tree, same model: 300 neurons, eight growth intervals then six of pruning. Synapse counts match in every build; the partners don't. Across the 14 updates: the share of synapses a neuron makes onto itself, and the average node-id gap between partners. The bottom row is what random wiring gives.

| build | self-connections | mean gap, source to target |
|---|---|---|
| main | 0.08% to 0.56% | 98.9 to 101.6 |
| **#3896** | **100%, every update** | **0.00** |
| #3779, and #3896 + Fisher-Yates | 0.28% to 1.00% | 98.5 to 100.8 |
| random wiring (analytic) | 0.33% | 100.0 |

Different lengths collapse less far, but still collapse: 3000 axon against 2000 dendrite slots on 1000 neurons give 9.3% and 4.5, where main gives 0.08% and 334.2.

`std::sample` also draws in proportion to N, not `n`, and the delete sites pass `n` = 1. At N = 10⁶, n = 10³, ms / RNG calls: main 270.2 / 1 000, #3779 **0.017 / 1 000**, #3896 5.313 / 499 207, Fisher-Yates behind this API **1.032 / 1 000**.

#3779 and #3896 fix different halves; putting #3779's partial Fisher-Yates inside `sample()` does both: same distribution as main, exactly `n` draws, output bit-identical to #3779 over 50 seeds.

nestkernel/random_generators.h
```diff
@@ -124,12 +124,27 @@ class BaseRandomGenerator
...
     size_t n ) = 0;
+
+  /**
+   * @brief Select randomly shuffled elements from vector.
+   *
+   * Places `n` elements chosen from `data` without replacement at the beginning of `data` and then
+   * reduce size of `data` to `n`.
+   *
+   * @param data Vector to be shuffled and possibly shortened
+   * @oaram n Number of elements to shuffle. `n <= data.size()` required.
```
Magnushst:
```suggestion
   * Places `n` elements chosen from `data` without replacement, in random order, at the beginning of
   * `data` and then reduces the size of `data` to `n`. If `n` exceeds `data.size()`, all elements are
   * kept and only shuffled.
   *
   * @param data Vector to be shuffled and possibly shortened.
   * @param n Number of elements to select.
```
@oaram typo. 

The doc says n <= data.size() is required, but the first line of the implementation is std::min( n, data.size() ), which is there precisely to handle the case where it is not. Saying it is clamped keeps the two in agreement; as it stands, someone could reasonably delete the min() as dead code.

nestkernel/random_generators.h
```diff
@@ -282,14 +297,30 @@ class RandomGenerator final : public BaseRandomGenerator
...
+
+    // Partial Fisher-Yates-Durstenfeld-Knuth: at each step select an element from the remaining range
+    // [ next, data.size() ) and swap it to the front. Draws num_elems random numbers and
+    // leaves the selection in random order.
+    for ( size_t next = 0, remaining = data.size(); next < num_elems; ++next, --remaining )
+    {
+      const size_t rnd = next + ulrand( remaining );
+      std::swap( data[ next ], data[ rnd ] );
+    }
+    data.resize( num_elems );  // keep only the m selected elements.
```
Magnushst:
```suggestion
    data.resize( num_elems );  // keep only the selected elements.
```

'm' was renamed to 'num_elems'.

Magnushst (approved):

heplesser:
@Magnushst Could you run a benchmark again, just to make sure performance is now the same as for #3779?

Magnushst:
Same output as #3779. Speed wise: equal at n=1, and 1.3 to 1.6 times faster at larger n on my laptop.

 Each run takes the median over 25 to 201 repetxitions; the table is the median of four such runs. Values are miliseconds.

| N | n | main | #3779 | #3896 | this PR |
|---|---|---|---|---|---|
| 10⁶ | 1 | 0.204 | 0.0003 | 1.41 | **0.0003** |
| 10⁶ | 10³ | 207 | 0.016 | 3.45 | **0.012** |
| 10⁵ | 10⁴ | 117 | 0.070 | 0.588 | **0.044** |
| 10⁴ | 10⁴ | 3.85 | 0.068 | 0.033 | **0.042** |

Magnushst (approved):

alpinangu (approved):
:+1:

PR #3897, opened 2026-06-24
Protect against setting parameters before threads

### LINKED ISSUE #3228: NEST behaves incorrectly and potentially segfaults if number of threads changed after parameters are created

heplesser:
This fixes #3228 at least in the sense that NEST will now throw an assertion if the user creates parameters before setting the number of threads. Unfortunately, there doesn't seem to be a way to do this more elegantly without requiring every parameter creation to update a "parameters exist" flag, which may lead to thread contention.

The indirection in the test was suggested by Opus 4.6.


alpinangu (approved):
👍

PR #3899, opened 2026-06-25
Weight recorder and recording backend changes

johnpapad24:
Added Weight recorder support for exporting connection labels.

Added Recording backend support for setting the delimiter of the exported file, skipping exportation of header comment and also the abillity to write to a single file from multiple threads.


models/weight_recorder.cpp
```diff
@@ -102,8 +104,16 @@ nest::weight_recorder::Parameters_::set( const Dictionary& d )
...
+  if ( !P_.include_label_ )
+  {
+    RecordingDevice::pre_run_hook(
+      { nest::names::weights }, { nest::names::targets, nest::names::receptors, nest::names::ports } );
+  }
+  else
+  {
+    RecordingDevice::pre_run_hook( { nest::names::weights },
+      { nest::names::label, nest::names::targets, nest::names::receptors, nest::names::ports } );
+  }
```
heplesser, line 117: We do not use symbolic boolean operators in NEST (see styleguide; @terhorstd why are out pre-commit checks still not detecting this?). Also, avoid negations if possible, as they make code more challenging to read.

Avoid code duplication if possible, e.g.,

```suggestion
auto long_headers = { nest::names::targets, nest::names::receptors, nest::names::ports };
if ( P_.include_label_ )
{
  long_headers.push_back( nest::names::label );
}
RecordingDevice::pre_run_hook({ nest::names::weights }, long_headers);
```
and corresponding changes in `write()`;

models/weight_recorder.cpp
```diff
@@ -167,11 +177,22 @@ nest::weight_recorder::handle( WeightRecorderEvent& e )
     {
       return;
     }
-
-    write( e,
-      { e.get_weight() },
-      { static_cast< long >( e.get_receiver_node_id() ),
-        static_cast< long >( e.get_rport() ),
-        static_cast< long >( e.get_port() ) } );
+    if ( !P_.include_label_ )
```
heplesser, line 181: See comments on `pre_run_hook()`.

models/weight_recorder.h
```diff
@@ -61,6 +61,10 @@ weight recorder accepts NodeCollections in the parameters ``senders`` and
 ``targets``. If set, they restrict the recording of data to only
 synapses that fulfill the given criteria.
 
+For labeled connection identification an optional boolean argument called
```
heplesser, line 64: @jessica-mitchell I think weight recorder documentation needs a proper parameter section.

models/weight_recorder.h
```diff
@@ -141,6 +145,7 @@ class weight_recorder : public RecordingDevice
   {
     NodeCollectionPTR senders_;
     NodeCollectionPTR targets_;
+    bool include_label_ = false;
```
heplesser: For consistency with all other models, initialize in attribute list in constructor definition, not here at attribute declaration.

nestkernel/connector_base.h
```diff
@@ -414,6 +414,7 @@ class Connector : public ConnectorBase
       ConnectionT& conn = C_[ lcid + lcid_offset ];
 
       e.set_port( lcid + lcid_offset );
+      e.set_conn_label( conn.get_label() );
```
heplesser: See comment in `event.h`.

nestkernel/event.h
```diff
@@ -365,6 +377,14 @@ class Event
    */
   size_t rp_;
 
+
+  /**
+   * Connection Label (conn_label)
+   *
+   * The label of the connection (if available).
+   */
+  long label_;
```
heplesser: Do we really need to add this (and the pertaining methods) to the `Event` base class? We only need it in `WeightRecorderEvent`. I'd prefer not to bloat the base class.

nestkernel/nest_names.h
```diff
@@ -120,6 +120,7 @@ const std::string capacity( "capacity" );
 const std::string center( "center" );
 const std::string circular( "circular" );
 const std::string clear( "clear" );
+const std::string combine_thread_write( "combine_thread_write" );
```
heplesser: I fear this name is not very informative for users. How about `one_file_per_process`?

nestkernel/nest_names.h
```diff
@@ -495,6 +498,7 @@ const std::string sion_chunksize( "sion_chunksize" );
 const std::string sion_collective( "sion_collective" );
 const std::string sion_n_files( "sion_n_files" );
 const std::string size_of( "sizeof" );
+const std::string skip_header_comment( "skip_header_comment" );
```
heplesser: Metadata in files is very useful and clients who do not want to bother with it can always skip it. So I am sceptical about introducing this option as I am sceptical about option inflation. If we add this option, I think "skip" fits more to reading and I would suggest `no_metadata` or similar.

nestkernel/recording_backend_ascii.cpp
```diff
@@ -244,92 +248,198 @@ nest::RecordingBackendASCII::DeviceData::set_value_names( const std::vector< std
...
+    }
+    file_ << std::endl;
   }
-  for ( auto& val : long_value_names_ )
+  else
   {
-    file_ << "\t" << val;
+#pragma omp single
+    {
+      std::ifstream test( filename.c_str() );
```
heplesser: Please avoid code duplication.

nestkernel/recording_backend_ascii.cpp
```diff
@@ -244,92 +248,198 @@ nest::RecordingBackendASCII::DeviceData::set_value_names( const std::vector< std
...
+
+    file_ << "\n";
   }
-  for ( auto& val : long_values )
+  else
   {
-    file_ << "\t" << val;
+#pragma omp critical
+    {
+      file_ << event.get_sender_node_id() << delimiter_;
```
heplesser: Please avoid code duplication.

heplesser (changes requested):
@johnpapad24 Thank you for your pull request. I am afraid there are a number of issues with it. For minor ones, please see the inline comments. Is see some major issues:
1. I am afraid of option inflation and am in particular sceptical about the idea of not writing file metadata. Readers can always ignore it.
2. There is quite some code duplication.
3. There are no test at all yet for the new functionality.
4. I find it difficult to believe that the "single file" mechanism works. When adding a recording device, `NodeManager` will add one instance of the recorder for each thread, so you have an individual `file_` handle for each thread. Having multiple handles for the same file, with threads taking turns at writing to the same file through their individual handles seems like a bad idea.

Overall, I wonder if it might not make more sense to create a recording backend and recorder more suitable to your particular needs for coupling with your software, e.g., as an extension module.

johnpapad24:
About code duplication: I tried to avoid it using wrappers but didn't work correctly on omp code,

I have also found a new way to parallel write to files (through appending) but the new way (although faster) does not preserve the file's order. I can add it through a new parameter though.

regarding tests, I have tested it with my own code for now but we must found a new mutually agreed way to test it by implementing weight recorder and test the produced output against the one produced by only one thread and process like the example.

johnpapad24:
I will make changes to the write function by writing two separate functions for synchronous and asynchronous parallel file write. 

Please test the asynchronous one against the current write function and replace it if better.

however complete code deduplication (as suggested in some examples) cannot be avoided because even because it won't even compile (as in pushing elements in initializer lists) or throws compiler warnings.

johnpapad24:
New version published:

This new version deals with parameter naming schemes, issues with code in base classes adds the ability for asynchronous and Synchronous single file writing, as well as tests. 

It also partially deals with code deduplication.

Please do a performance check on this code because the write function has been changed for avoiding code deduplication.
If the performance is poorer than the previous solution a revert of changes will be done.

P.S. the check should be performed with the default options of course.

nestkernel/event.h
```diff
@@ -231,6 +231,7 @@ class Event
    */
   size_t get_rport() const;
 
+
```
JoseJVS: We usually do not use double empty lines to separate within class declarations.
```suggestion
```

nestkernel/nest_names.h
```diff
@@ -716,6 +719,8 @@ const std::string wfr_interpolation_order( "wfr_interpolation_order" );
 const std::string wfr_max_iterations( "wfr_max_iterations" );
 const std::string wfr_tol( "wfr_tol" );
 const std::string with_reset( "with_reset" );
+const std::string write_to_single_file( "write_to_single_file" );
+
```
JoseJVS, line 723: Within the names namespace we only use a single empty line to separate name blocks.
```suggestion
```

nestkernel/recording_backend_ascii.cpp
JoseJVS, line 1: Like on event.h and nest_names.h there are empty lines inconsistently added in between method definitions and within definitions too.

nestkernel/recording_backend_ascii.cpp
```diff
@@ -229,7 +230,11 @@ nest::RecordingBackendASCII::DeviceData::DeviceData( std::string modelname, std:
   , modelname_( modelname )
   , vp_node_id_string_( vp_node_id_string )
   , file_extension_( "dat" )
+  , delimiter_( "\t" )
   , label_( "" )
+  , no_metadata_( false )
+  , write_to_single_file_( "Off" )
+//, write_mutex_(std::make_unique<std::mutex>())
```
JoseJVS: Please do not leave commented code for PRs.
```suggestion
```

nestkernel/recording_backend_ascii.cpp
```diff
@@ -244,84 +249,196 @@ nest::RecordingBackendASCII::DeviceData::set_value_names( const std::vector< std
...
+
+  if ( not file_.good() )
   {
-    file_ << "\t" << val;
+    std::string msg = String::compose( "I/O error while opening file '%1'.", filename );
+    LOG( VerbosityLevel::ERROR, "RecordingBackendASCII::prepare()", msg );
+    throw IOError();
   }
-  file_ << std::endl;
+  // file_ << std::fixed << std::setprecision( precision_ );
```
JoseJVS:
```suggestion
```

nestkernel/recording_backend_ascii.cpp
JoseJVS: Please also check for and remove leftover commented code.

JoseJVS, line 1: There is a lot of code duplication between open_file and open_file_omp, this could be certainly refactored, especially as open_file_omp is called within open_file.

JoseJVS, line 1: I find hard to follow the logic between of the write_single_file options, could you document how you converged to this synchronization scheme?

nestkernel/recording_backend_ascii.h
```diff
@@ -193,13 +212,14 @@ class RecordingBackendASCII : public RecordingBackend
 
 private:
   const std::string compute_vp_node_id_string_( const RecordingDevice& device ) const;
-
   struct DeviceData
   {
+    // std::unique_ptr<std::mutex> write_mutex_;
```
JoseJVS:
```suggestion
```

nestkernel/recording_backend_ascii.h
```diff
@@ -25,6 +25,7 @@
 
 // C++ includes:
 #include <fstream>
+#include <mutex>
```
JoseJVS: I believe this is no longer in use now
```suggestion
```

nestkernel/recording_backend_ascii.h
```diff
@@ -133,6 +134,24 @@ time_in_steps
...
+
+write_single_file
+  A string value that gives the user the option to write in a single file
+  from mutiple threads and processes either by keeing the time order of the file
+  (*"Syncronous" *option) or not (*"Asyncronous"* option). Please note the using
+  this option in large distributed systems may result in considerable simulation
+  slowdowns especially when using the *"Syncronous"* option.
+  (default *"Off"*).
+
+
```
JoseJVS:
```suggestion
```

pynest/nest/raster_plot.py
```diff
@@ -123,20 +127,24 @@ def from_file(fname, **kwargs):
...
     """Use pandas."""
 
     data = None
+    head = 2
+    if no_metadata == True:
+        head = 0
     for f in fname:
         # pylint: disable=possibly-used-before-assignment
-        dataFrame = pandas.read_table(f, header=2, skipinitialspace=True)
+
```
JoseJVS:
```suggestion
```

johnpapad24:
Dear all,

All these will be fixed very soon.

John Papadopoulos

JoseJVS:
I find it hard to justify the usefulness of the delimiter and no_metadata options, as @heplesser mentioned we should avoid bloating the kernel with parameters and our general tendency is to allow users to get as much information as possible from their simulations.
The file headers are not costly neither in disk space nor in performance, so I see no reason to remove them.

I also find the recording test lacking, in particular I am not convinced that this test demonstrates proper synchronization just by looking at a single output value.

Lastly, could you clarify what is your target use case for the synchronization? As you previously mentioned (and rightly so) synchronizing file writes across threads would have a huge impact on distributed simulations, potentially even outweighing performance gains from parallelization during simulation.
Is this necessary for you as a single thread cannot contain your full network?

johnpapad24:
My application involves the combination of biodynamo and nest in order to
produce a model that combines both the neural outgrowth and electrical
activity.

The no_metadata option allows the use of custom, simplier csv readers
wheras the delimitter allows the user to choose the delimitter in order to
have the data exported unified file formats.

I can also set a second (or even final) line test in the test file to
ensure the correct order. (This test was copied whoever from the write file
tests which I think it are lacking overall).

John Papadopoulos

JoseJVS:
Thanks for the quick answer!

> My application involves the combination of biodynamo and nest in order to
produce a model that combines both the neural outgrowth and electrical
activity.

But have you tried running NEST with your network on a single thread? If so is the performance still slower than multi-threaded with synchronized file writes?

> The no_metadata option allows the use of custom, simplier csv readers
wheras the delimitter allows the user to choose the delimitter in order to
have the data exported unified file formats.

I see that you have a use case for this, however I would still argue that it would still be simpler to load the file with numpy (which is our target data loader application) and then export it to whatever format you see fit within Python.
This generalizes with ease to many other formats instead of trying to support this from the kernel side.

> I can also set a second (or even final) line test in the test file to
ensure the correct order. (This test was copied whoever from the write file
tests which I think it are lacking overall).

Thank you! The original tests were meant to check simple outputs, however for a more complex algorithm such as thread-synchronized file writing this is indeed insufficient.

johnpapad24:
Generally nest is called multiple times throughout the whole simulation and
using numpy and other tricks to bring the files to the right format in the
middle of the simulation (as i am doing right now) is very very slow.

Also I am using 16 openmp threads for my simulation which delivers better
performance but having 16 different files for each diffrerent case I
examine (ex. Excitatory neurons, inhibitrory neurons. Etc) is very very
messy.

johnpapad24:
Also the syncronization code is tested against both mpi and openmp
simutanously.

JoseJVS:
> Generally nest is called multiple times throughout the whole simulation and
using numpy and other tricks to bring the files to the right format in the
middle of the simulation (as i am doing right now) is very very slow.

Could you give me an estimate of how large are your files?

> Also I am using 16 openmp threads for my simulation which delivers better
performance but having 16 different files for each diffrerent case I
examine (ex. Excitatory neurons, inhibitrory neurons. Etc) is very very
messy.

Have you tested the performance impact of the synchronized writing with your network? If so could you give me an estimate of the slowdown?

johnpapad24:
Basically my files are 4+Gb,

Generally my project does not need synchronous file writing but the performance impact of this synchronization depends on the environment.

johnpapad24:
yesterday's comments are also solved

JoseJVS:
> Basically my files are 4+Gb

I see, I understand better now the logic for the delimiter and metadata header, as having to transform several Gb of data in between NEST calls would indeed be cumbersome.
However, I have another question out of curiosity, how does having a single 4+Gb instead of multiple files help you? Wouldn't it be easier to parse multiple files in parallel?

> Generally my project does not need synchronous file writing but the performance impact of this synchronization depends on the environment.

You are indeed right that it would depend on the system, would it be fine for you to run a simple performance analysis of with a file for each thread vs a single file for all threads in your system? If you are fine with that just run your NEST model, while recording, for both versions and post some information about your network architecture, parameters, ideally for multiple seeds and post some information about the hardware and software stack you are using.

 > yesterday's comments are also solved

I'm sorry but I still see some unresolved comments in some files (e.g. https://github.com/nest/nest-simulator/pull/3899/changes#r3922684027 ) in particular I did not mark every single occurrence of misplaced added or deleted empty lines as well as commented code in the PR and I would ask you to please double check that.

johnpapad24:
in fact, I don't have just one 4+gb file as I have different files each time nest is called.

NEST produces me at least 4 large files corresponding to each case (spikes inhibitory/excitatory and weights inhibitory/excitatory) at a time. If each time I call nest I had 10s of files for each case it would be a huge problem for me.

also I will double check the code and do the performance analysis

johnpapad24:
I have created a python file benchmark based on the reproducibility test

Simulation Time: 100000ms on 16 omp threads and 4 mpi procs

Asynchronous: Simulation time : 11.87 s

Synchronous:  Simulation time : 19.45 s - 33.9 s (depends on synchronization)

Write to different files: Simulation time   : 14.85 s
[benchmark.py](https://github.com/user-attachments/files/31913005/benchmark.py)

JoseJVS:
> I have created a python file benchmark based on the reproducibility test

Thanks for the benchmarking work, I checked your script and found an issue, you set both ```local_num_threads``` and ```total_num_virtual_procs```. Given that you wrote:

> Simulation Time: 100000ms on 16 omp threads and 4 mpi procs

I am assuming you set ```total_num_virtual_procs = 4``` to indicate that you "run" with 4 MPI processes, however this is incorrect. The number of MPI processes you use is set as an argument to your MPI executor e.g. ```mpirun -np 4``` NEST then knows automatically how many MPI processes are active. In practice, both ```local_num_threads``` and ```total_num_virtual_procs``` set the number of computing threads but following a different formula which is explained here: https://nest-simulator.readthedocs.io/en/stable/hpc/parallel_computing.html

All in all, as the instruction ```total_num_virtual_procs = 4``` is executed after ```local_num_threads = 16 ``` (overwriting it), the actual number of computing threads used is 4. If you correctly launched the script with 4 MPI processes this means that only one thread per process is used, if you did not launch it via MPI then this means 4 threads per process were used (i.e. running with a single process).

johnpapad24:
I think that one argument is for the openmp procs and the other for the total processes.

JoseJVS:
> I think that one argument is for the openmp procs and the other for the total processes.

I assumed as much, however that is wrong, please read this guide: https://nest-simulator.readthedocs.io/en/stable/hpc/parallel_computing.html

johnpapad24:
the correct solution for reproducibility is setting the nest.total_num_virtual_procs = 64

however this has made the simulation very sluggish.

JoseJVS:
> the correct solution for reproducibility is setting the nest.total_num_virtual_procs = 64

From our previous messages, I understand that you want to you use 64 threads in total, with 16 threads per process and 4 processes, so this should be indeed correct.

On a side note, I personally find that setting ```local_num_threads = 16``` in this case is more clear, "local_num_threads" directly translates to the number of threads per process, but that's just my personal opinion.
Naturally, the equivalence with ```total_num_virtual_procs = 64``` is only true if you run the simulation with 4 processes.
Also, I do not know if you changed this but as previously mentioned, do not set both ```total_num_virtual_procs``` and ```local_num_threads``` in the same script, as one will overwrite the other.

> however this has made the simulation very sluggish.

A likely cause to this slowdown on your simulation could be that you are oversubscribing the number of threads compared to the number of physical cores (previously you were only running your simulations with 4 threads in total). Furthermore, Hyper-threading (Intel nomenclature) or Simultaneous Multi-Threading (more generally speaking; SMT for short) is generally not beneficial to the performance of NEST simulations due to our thread-local memory architecture.

Could you tell me how many physical CPU cores does your system have? (if you run something like ```lscpu``` you will get information about the CPU of your system, there you can see information on the number of threads per core, cores per socket, and sockets; having multiple threads per core means that your system is capable of SMT; the product of the number of cores per socket with the number of sockets will result in the total number of physical cores)

Ideally, you want to use as many virtual processes as you have cores available and, unless you are running your simulations across multiple sockets, there is no reason to use multiple MPI processes on a local workstation.

johnpapad24:
In that case I am oversubscribing because my machine contains just 4 physical cores (8 with hyperthreading) but using 16 opemnp threads in total is fine with me but not in the case of using mpi.

JoseJVS:
> In that case I am oversubscribing because my machine contains just 4 physical cores (8 with hyperthreading) but using 16 opemnp threads in total is fine with me but not in the case of using mpi.

I see, you can of course run your simulations however you want, however just for this benchmarking exercise, I would like to see performance results with only **4 threads** i.e. ```local_num_threads = 4```, **without MPI**, and prefacing the launch argument with the **time** command i.e. ```time python3 benchmark.py```.

The time command will give some more information about cumulative thread time as well as time spent in kernel space. Once the benchmark is finished, the time command prints this information to stderr.

With that setup, could you please run the benchmark again using vanilla NEST, your fork with the single file asynchronous configuration, as well as single file synchronous configuration, and paste the output of the time command for each configuration in a comment.

PR #3900, opened 2026-06-26
Significantly accelerate GetConnection with source or target selection

heplesser:
The existing implementation of `GetConnection()` is very inefficient and does not scale properly under parallelization. This PR fixes this.

The benchmark shown below, picking 5 million out of 125 million connections, achieved the following times for `GetConnection()` for different code and parallelization choices (MacBook Pro M3, -O2):

 | MPI Procs | Threads | Main | This branch |
|-----:|---:|---:|---:|
| 1 | 4  | 10.5 | 0.39 |
| 2 | 2 | 10.4 | 0.27 |
| 4 | 1 | 10.3  | 0.21 |

Running on a single MPI process is slower than on four, because all 5 million connections need to be moved to the Python level.

The key change in is that connection collection is now done thread-parallel. Connections are collected into a `vector< deque< ConnectionID > >` structure, with one deque per thread. This structure is passed as-is up to the Python level and is flattened into a Python list only there.

Connection selection needs to handle connections to/from devices differently from connections between neurons proper. Distinguishing between these types of nodes currently depends on the `has_proxy()` and `local_receiver()` properties. This anything but an optimal solution, but needs to be addressed in the context of re-designing our devices.

```python
"""
Mini-benchmark for GetConnections with source and target selection

- 5 populations à 10.000 => 50.000 neurons
- In-degree 2.500 => 125 million synapses
- Select connections between two populations => 5 million synapses collected
- One VP by default
- Runs by default 10 times over to get decent times.

Usage

python demo.py [n_vp [n_reps]]
"""

import nest
from time import time
import sys

n_vp = int(sys.argv[1]) if len(sys.argv) > 1 else 1
n_reps = int(sys.argv[2]) if len(sys.argv) > 2 else 10

nest.total_num_virtual_procs = n_vp


pops = [nest.Create("parrot_neuron", n=10000) for _ in range(5)]
nrns = sum(pops)

tic = time()
nest.Connect(nrns, nrns, {"rule": "fixed_indegree", "indegree": 2500})
t_conn = time() - tic

# Force creation of pre-synaptic connection infrastructure
nest.GetConnections(source=pops[0][0])

tic = time()
for _ in range(n_reps):
    cg = nest.GetConnections(source=pops[0], target=pops[2])
t_get = time() - tic


print(f"{t_conn          = }")
print(f"t_get (per rep) = {t_get/n_reps}")
print(f"{len(cg)         = }")
```

### DIFF
--- a/models/music_event_in_proxy.h
+++ b/models/music_event_in_proxy.h
@@ -101,12 +101,17 @@ public:
   music_event_in_proxy( const music_event_in_proxy& );
 
   bool
-  has_proxies() const
+  has_proxies() const override
   {
     return false;
   }
   bool
-  one_node_per_process() const
+  one_node_per_process() const override
+  {
+    return true;
+  }
+  bool
+  local_receiver() const override
   {
     return true;
   }
--- a/nestkernel/connection_manager.cpp
+++ b/nestkernel/connection_manager.cpp
@@ -1097,6 +1097,10 @@ nest::ConnectionManager::get_connections( const Dictionary& params )
     {
       throw KernelException( "GetConnection requires valid source NodeCollection." );
     }
+    if ( source_a->all_connect_as_neurons() == source_a->all_connect_as_devices() )
+    {
+      throw KernelException( "source must either contain only neurons or only devices." );
+    }
   }
   if ( params.known( names::target ) )
   {
@@ -1104,6 +1108,10 @@ nest::ConnectionManager::get_connections( const Dictionary& params )
     if ( not target_a->valid() )
     {
       throw KernelException( "GetConnection requires valid target NodeCollection." );
+    }
+    if ( target_a->all_connect_as_neurons() == target_a->all_connect_as_devices() )
+    {
+      throw KernelException( "target must either contain only neurons or only devices." );
     }
   }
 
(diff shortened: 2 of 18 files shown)


heplesser:
@leonwoelki @alpinangu I have now merged main into this PR to reduce the change set and added a bit more on the changes made above. I'd appreciated it if you could review it in the near future.

nestkernel/node_collection.cpp
```diff
@@ -1337,6 +1342,20 @@ NodeCollectionComposite::has_proxies() const
     parts_.begin(), parts_.end(), []( const NodeCollectionPrimitive& prim ) { return prim.has_proxies(); } );
 }
 
+bool
+NodeCollectionComposite::all_connect_as_neurons() const
+{
+  return std::all_of(
+    parts_.begin(), parts_.end(), []( const NodeCollectionPrimitive& prim ) { return prim.all_connect_as_neurons(); } );
```
alpinangu, line 1349: Sorry for the late review. 
This fails if, before calling this function, we slice a NodeCollectionComposite consisting of multiple neuron types and a device in a way that the resulting NodeCollection contains only neurons. Because the result is still a composite, but in its constructor, we copy the parts_ from the original composite, which still includes the device part.

nestkernel/node_collection.cpp
```diff
@@ -1337,6 +1342,20 @@ NodeCollectionComposite::has_proxies() const
...
+}
+
+bool
+NodeCollectionComposite::all_connect_as_devices() const
+{
+  return std::all_of(
+    parts_.begin(), parts_.end(), []( const NodeCollectionPrimitive& prim ) { return prim.all_connect_as_devices(); } );
```
alpinangu, line 1356: Same issue here but the other way around

nestkernel/node_collection.cpp
heplesser: @alpinangu Thanks for noticing! I consider this such an exotic corner case that I would not want to support it. So I now mention in the documentation that this case is not supported. It will trigger an error message, so it is safe in the sense that it will not silently do unexpected things.

I did do a brief test and things work as expected if you have a combination of one neuron + one device node collection. Things only go wrong if you do something like

```python
m, r = nest.NodeCollection(), nest.NodeCollection()

for _ in range(3):
     m += nest.Create("parrot_neuron")
     r += nest.Create("spike_recorder")
mr = m + r

nest.Connect(m, r, "one_to_one")

nest.GetConnections(target=mr[1::2])
```

But that would be so poor style that it needs not be supported.

models/music_event_in_proxy.h
leonwoelki, line 1: I do not know if they are still maintained, but I found three more headers that contain a similar structure (music_cont_in_proxy.h, music_rate_in_proxy.h, music_message_in_proxy.h), which are not touched in this PR.

nestkernel/connection_manager.cpp
```diff
@@ -1145,41 +1153,48 @@ ConnectionManager::get_connections( const Dictionary& params )
...
-    else
+    else if ( source.get() )
     {
-      device_node_ids.push_back( node_id );
+      get_connections_from_sources_( tid, conns_in_thread, source, target, syn_id, synapse_label );
+    }
+
+    if ( conns_in_thread.size() > 0 )
+    {
+#pragma omp critical
```
leonwoelki, line 1194: Couldn't this unnamed marker introduce a delay, since there are multiple unnamed cases, e.g., in simulation_manager.cpp?
Shouldn't be an issue currently, but it could become one, and it would not cost anything to name it again?

PR #3901, opened 2026-06-27
CMake: Improve handling of Boost and GSL

heplesser:
This PR is the first in a series aiming to modernize our CMake setup. This PR addresses the detection of Boost and GSL.

Note that a plain `if ( with-gsl )` will be true for a considerable range of "positive" values passed, case insensitive, and is therefore recommended over explicit `STREQUAL "OFF"` and similar. Standardization of this for all is on my backlog.

PR #3902, opened 2026-06-27
CMake: Modernize MPI and OpenMP detection including autodetection of libomp on macOS

heplesser:
This PR is the second CMake modernization PR and should be considered after #3901 to minimize the changeset.

The main user benefit of this PR is that brew-installed libomp will be automatically detected on macOS.


JanVogelsang (approved):

terhorstd (approved):
Looks good :+1:

PR #3903, opened 2026-06-27
CMake: Remove historic checks, C-dependencies and rely on functions that are provided by the C++ standard library

heplesser:
The third of the CMake-improvement commits, should be handled after #3901 and #3902. It mainly removes stuff that is very outdated and drops any C dependencies. CMake runs a lot faster now with the superfluous checks out of the way.

Tests for uint64_t and other specific types are removed because they are provided in `cstdlib` by all known STL implementations.


JanVogelsang (approved):

terhorstd (approved):
Joint rewview with @JoseJVS. look good. 👍

PR #3904, opened 2026-06-28
CMake: Properly mark linked libraries as `PUBLIC` or `PRIVATE`

heplesser:
The fourth CMake modernization PR. Should be handled after #3903 is in.

This PR brings us closer to best practice by marking whether libraries that one links against are "passed on" via headers or only needed in cpp files.

I developed this PR with the help of Claude.ai. It summarized the changes as follows (I reviewed the summary in view of the code changes):

**PUBLIC/PRIVATE keyword sweep**. `target_link_libraries` in `libnestutil`, `nestkernel`, `models`, and `testsuite/cpptests` now use explicit keywords. `Boost::headers` stays `PUBLIC` on `nestutil` (its public headers `dictionary.h`/`sort.h` expose Boost-dependent templates used pervasively elsewhere); GSL/OpenMP/MPI/SIONlib/MUSIC/LibNeurosim/HDF5/LTDL are `PRIVATE` wherever linked, since their compile-time include paths are already applied globally in `cmake/ProcessOptions.cmake` and CMake still correctly propagates `PRIVATE` link *libraries* of static targets to the final link step. `models` gained an explicit `GSL::gsl PRIVATE` (29 headers use `gsl/` directly) and `OpenMP::OpenMP_CXX PRIVATE`; `nestkernel` gained an explicit `OpenMP::OpenMP_CXX PRIVATE` — both previously only worked by accident via the un-keyworded (effectively `PUBLIC`) chain through `nestutil`.


JanVogelsang (approved):

terhorstd (approved):
Agreed on open questions regarding cross-library dependencies in online discussion.
* different gsl/mpi/… packages in the environment are a problem of packaging, not to be solved by NEST's cmake.
* noticed the include order in Cmake modules is opposite to expectations: kernel includes everything else (see also #3905)

With these points understood, this one is good to go. 👍

Issue #3905, opened 2026-06-28
Circular dependency between nestkernel and models


heplesser:
Logically, our `models` library will depend on `nestkernel`, but `nestkernel` should not depend on `models`. But because `ModelManager` registers models, we have a circular dependency. That should be removed. See below for a full analysis provided by Claude. This came up in connection with #3906.

@JanVogelsang I assume this is relevant to and possibly already addressed by your LTO work in #3544, so I assign this to you.

----

nestkernel/model_manager.cpp:44 does #include "models.h" and calls register_models() at line 86. That function is declared in models/models.h but implemented in the generated models/models.cpp (which contains the actual calls to register every concrete neuron/synapse class). So nestkernel's own object code has a real undefined-symbol reference into models — it's not a CMake mistake, there's a genuine function call across that boundary.
The underlying design smell: nestkernel owns the model registry (ModelManager, register_models()'s declaration), but the call that populates it is made from inside nestkernel, reaching out to models — i.e. the lower layer is reaching up into the layer built on top of it. That's what creates the real circular coupling (models needs nestkernel's Node/Connection base classes to define models; nestkernel needs models to list them).
The architecturally clean fix is to invert who calls whom: instead of model_manager.cpp calling register_models(), the registration call should happen from wherever models is already linked in (e.g. the final executable/nestkernel_api module's startup code, or a self-registration mechanism where each model statically registers itself into the kernel's registry without nestkernel ever calling into models). That would let nestkernel drop its link to models entirely, breaking the cycle for good.

PR #3906, opened 2026-06-28
CMake: Reduce duplicate library inclusion

heplesser:
This is the 5th PR in the CMake modernization effort. It reduces duplicate inclusion of libraries on linking. It cannot fully resolve this problem due to #3905, but still provides and improvement. See there for details on the remaining issues.

This PR should be handled after #3901 to #3904.


JanVogelsang (approved):

JoseJVS (approved):
Looks good to me!

PR #3908, opened 2026-06-28
CMake: Do not build files that are empty due to missing MUSIC

heplesser:
Part 6 of the CMake modernization. Some compilation units will be completely empty when MUSIC is missing. This PR avoids warnings about empty .o files.

Should be applied after #3906 is merged.


JanVogelsang (approved):

terhorstd (approved):
very nice. 👍

PR #3909, opened 2026-06-28
CMake: Remove redundant 32bit-architecture flag

heplesser:
This is part 7 of the CMake cleanup.  It removes a redundant flag related to 32-bit architectures.

Should be reviewed once #3908 is done.


JanVogelsang (approved):

cmake/CheckIncludesSymbols.cmake
```diff
@@ -19,101 +19,17 @@
...
-check_include_files( "stdint.h" HAVE_STDINT_H )
-check_include_files( "sys/types.h" HAVE_SYS_TYPES_H )
-
-include( CheckIncludeFileCXX )
-check_include_file_cxx( "istream" HAVE_ISTREAM )
-check_include_file_cxx( "ostream" HAVE_OSTREAM )
-check_include_file_cxx( "sstream" HAVE_SSTREAM )
-
 # Check types exist
 include( CheckTypeSize )
```
JoseJVS:
```suggestion
```
Is not used anymore after test removal

cmake/ProcessOptions.cmake
```diff
@@ -386,28 +388,16 @@ function( NEST_PROCESS_WITH_MPI )
   # Find MPI
   set( HAVE_MPI OFF PARENT_SCOPE )
   if ( NOT "${with-mpi}" STREQUAL "OFF" )
-    if ( NOT ${with-mpi} STREQUAL "ON" )
+    if ( NOT "${with-mpi}" STREQUAL "ON" )
       # if set, use this prefix
       set( MPI_ROOT "${with-mpi}" )
     endif ()
-    find_package( MPI REQUIRED )
+    find_package( MPI REQUIRED QUIET )
```
JoseJVS, line 395:
```suggestion
    find_package( MPI REQUIRED )
```
Conflicting option as of https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature
i.e. quiet is used for silent failures while required implies an error should be thrown if package not found

libnestutil/CMakeLists.txt
```diff
@@ -42,14 +42,23 @@ endif()
...
+# Temporary: nestutil includes nestkernel headers that pull in mpi.h, so nestutil needs
+# the MPI include path. Remove once the nestutil/nestkernel circular include is resolved.
+target_link_libraries(nestutil
+  PUBLIC
+    Boost::headers
+  PRIVATE
+    GSL::gsl
+    ${SIONLIB_LIBS}
+    OpenMP::OpenMP_CXX
+    MPI::MPI_CXX
```
JoseJVS, line 54: Shouldn't MPI and OpenMP also propagate through PUBLIC linkage?

libnestutil/nest_types.h
```diff
@@ -108,20 +100,9 @@ constexpr uint64_t MAX_NODE_ID = DISABLED_NODE_ID - 1;
...
-const tic_t tic_t_max = LONG_LONG_MAX;
-const tic_t tic_t_min = LONG_LONG_MIN;
-#endif
-#else
-typedef long tic_t;
-const tic_t tic_t_max = LONG_MAX;
-const tic_t tic_t_min = LONG_MIN;
-#endif
+const tic_t tic_t_max = std::numeric_limits< tic_t >::max();
+const tic_t tic_t_min = std::numeric_limits< tic_t >::min();
```
JoseJVS:
```suggestion
constexpr const tic_t tic_t_max = std::numeric_limits< tic_t >::max();
constexpr const tic_t tic_t_min = std::numeric_limits< tic_t >::min();
```

numeric limits library provides constant expressions for min/max types

libnestutil/numerics.h
```diff
@@ -32,11 +32,9 @@
 #include <cstddef>
 #include <limits>
 
-#if defined( HAVE_STD_ISNAN )
-#include <cmath>
-#elif defined( HAVE_ISNAN )
-#include <math.h>
-#endif
+// Generated includes:
```
JoseJVS:
```suggestion
```

libnestutil/numerics.h
```diff
@@ -46,55 +44,17 @@ extern const double pi;
...
-    }
-
-    return sum;
-  }
-#endif
+  return std::expm1( x );
 }
 
 template < typename T >
 bool
```
JoseJVS:
```suggestion
inline bool
```

.gitignore
```diff
@@ -74,3 +74,7 @@ cscope.out
 
 # Merge conflict artifacts
 *.orig
+
+# User-specific LLM files
+CLAUDE.md
+.claude/settings.local.json
```
JoseJVS:
```suggestion
```

This should be user specific settings that can be configured e.g. global git settings
```git config --global core.excludesfile $HOME/.gitignore```

File: cmake/CheckIncludesSymbols.cmake (hunk 1 above)
Comment (heplesser) [reply cmake/CheckIncludesSymbols.cmake]: Fixing, also removing the comment line above it and and empty line.

File: cmake/ProcessOptions.cmake (hunk 2 above)
Comment (heplesser) [reply cmake/ProcessOptions.cmake:395]: I understand the logic differently here and tests confirm this, so I want to keep the `QUIET`. With MPI present on the system, without `QUIET` we get three lines of output
```
-- Found MPI_CXX: /opt/homebrew/Cellar/open-mpi/5.0.9_1/lib/libmpi.dylib (found version "3.1")
-- Found MPI: TRUE (found version "3.1")
-- Found MPI: /opt/homebrew/bin/mpicxx (found version )
```
while with `QUIET` we only get the last line, and I think that is all that is needed.

If MPI is not present on the system, we get
```
-- Checking for module 'mpi-cxx'
--   Package 'mpi-cxx' not found
CMake Error at /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:290 (message):
  Could NOT find MPI (missing: MPI_CXX_FOUND)
  /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:654 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/share/cmake/Modules/FindMPI.cmake:2006 (find_package_handle_standard_args)
  cmake/ProcessOptions.cmake:395 (find_package)
  CMakeLists.txt:152 (nest_process_with_mpi)
```
no matter whether `QUIET` is present or not.

But I noticed that in the desired output line 
```
-- Found MPI: /opt/homebrew/bin/mpicxx (found version )
```
the MPI version number is missing. This could be fixed by replacing `${MPI_VERSION}` by `${MPI_CXX_VERSION}`, but for my installation, e.g., this would output "3.1" as the certified fully supported MPI standard—it does not provide the version of the MPI library found (5.0.9 on my machine). Therefore, I decided to drop the version number here entirely.

libnestutil/CMakeLists.txt
heplesser: All use of GSL, SION, OpenMP and MPI in libnestutil is local to libnestutil code, therefore PRIVATE linking suffices here. Ideally, we would want the same for Boost, but (see comment on lines 45–46 in CMakeLists.txt) we need PUBLIC linking for now due to certain dependencies.

Claude provided the following rule of thumb: if a dependency only shows up in .cpp files → PRIVATE. If it leaks into a .h that other targets #include → PUBLIC. (There's also INTERFACE, for header-only usage that doesn't affect the target's own build — not needed here since nestutil is a compiled library.)

libnestutil/nest_types.h
heplesser: Fixing in next commit (and a few more similar ones).

libnestutil/numerics.h
heplesser: Fixing in next commit.

libnestutil/numerics.h
heplesser: Fixing in next commit.

.gitignore
heplesser: OK, I will take it out. We might in a separate step want to review our `.gitignore` file as it contains some more lines that are not NEST-specific.

heplesser:
@JoseJVS I have now addressed all your comments, I hope you are content with the answers and changes.

cmake/ProcessOptions.cmake
JoseJVS: I see, strange combination of flags but as long as that an error is thrown if MPI is not found (and it is being asked for) then I am fine with it :+1:

libnestutil/CMakeLists.txt
JoseJVS: That's a good rule of thumb thanks for checking :+1:

.gitignore
JoseJVS: Fair enough :+1:

JoseJVS (approved):

terhorstd (approved):
Clarification of PR-title and description, make this a good thing.

Thanks for the cleanup! 👍

PR #3910, opened 2026-06-28
CMake: Minor refactoring

heplesser:
This is part 8 of the cmake project. It is not directly related (but came up in the context), so I still put it here. It mainly provides extra documentation, removes a detour to doubles and uses C++11 loop syntax instead of indexed loops.

Should be considered after #3909.


JanVogelsang (approved):

terhorstd (approved):
LGTM :+1:

PR #3911, opened 2026-06-28
Use pi and e from C++20 std::numbers

heplesser:
This PR was created as part 9 of cmake modernization, even though it is not directly related. It removes a lot of rigg around pi and e detection since they are available from `std::numbers` in C++20.

Consider once #3910 is in.


JanVogelsang (approved):

terhorstd (approved):
Good improvements!
Thanks 👍

PR #3912, opened 2026-06-28
CMake: Revision of CMake option handling for greater robustness (with some irrelevant options removed and a few renamed)

heplesser:
Part 10 of the CMake makeover. I have thoroughly revised and extended this PR after #3911 was merged. The work is based on use of claude-code.

This PR ensures a systematic and robust approach to all NEST-provided CMake options. Key changes are described below and user-facing changes are summarized in [What's new in v3.11](https://github.com/heplesser/nest-simulator/blob/913bef862cd7db96c09b2555d4055dbe6b155bae/doc/htmldoc/whats_new/v3.11/index.rst#cmake-configuration-modernization).

## User-facing changes

- Require Python 3.10 instead of 3.8 (had that implicitly for a while)
- All NEST-specific options now take the form `-Dwith-...`
- All options now translate any true/false-ish value to `ON` and `OFF`, respectively. By CMake default, anything that is not explicitly false-ish is considered true, so `-Dwith-mpi=foo` will turn MPI on. Sticking to general CMake behavior seems sensible.
- All options now fall into one of two categories
    a. **Pure boolean:** Only `ON` and `OFF` allowed (i.e., all true/false-ish values)
    b. **ON/OFF/String:**
	    - `ON` uses a default value, e.g., `-O2`
	    - `OFF` sets to the empty string
	    - `String` uses the given string as is
- Any options that are not used by CMake will be reported as errors to detect typos. If an unknown name is a removed or renamed option, an informative message is given.
- To specify non-default library locations, use `-Dxyz_ROOT=<path>` (or set `xyz_ROOT` in the environment or include it in `CMAKE_PREFIX_PATH`. If given as `-Dxyz_ROOT=...`, only that path will be checked for lib `xyz`.

Notes:
- This still allows passing arbitrary compiler flags using the `-Dwith-optimize`, `-Dwith-debug` and `-Dwith-defines` options.
- The ON/OFF/String variant should be reviewed in connection with #3805: Once we have presets, we can drop the `-Dwith-optimize|debug|warnings` options (the presets will handle what you get with ON/OFF now) and then just have a standard way to pass additional compiler or linker options.

### Options removed

- The `-Dwith-python` option is gone, since Python is mandatory now. The Python executable can still be specified using `-DPython_EXECUTABLE`.
- The `-Dtics_per_[ms|step]` options are gone, since the parameters can be set at runtime and there is no advantage to having them cmake-configurable in addition.

### Options renamed

- `-Dstatic-libraries` is now `-Dwith-static-linking` and its only effect is that whereever possible libraries will be linked into `pynestkernel.so` statically. This may be useful on HPC systems to avoid having to load many `*.so` files dynamically.
- `-Dtarget-bit-split` is now `-Dwith-target-bit-split` and the default value is now called `default` instead of `standard`. The only alternative is `hpc`. I decided not to convert this to a boolean option since it is possible that we will introduce other splits at some point.
- `-Dwith-intel-compiler-flags=<flags>` is now `-Dwith-intel-compiler-strict-math=ON|OFF` with default `ON`. Default behavior is as before that when using the Intel compiler, we add the `-fp-mode strict` compiler flag to ensure IEEE754 conformant arithmetic. Any other compiler flags should still be set via the usual options related to compiler flags.
- `-Dcythonize-nest=[ON|OFF]` is replaced with`-DNESTKERNEL_API_CXX=<path/to/cxx_file>`. If not given, Cython is run to create `nestkernel_api.cxx` in `<nest_build_directory>/pynest`. If given, there must be a cxx file at the given location at the latest when running `make`. This approach is in keeping with the `xyz_ROOT` variables above, while an explicit `ON/OFF` options seems overkill for this very peculiar case.

## Behind the scenes

- Some clean-up and better code documentation.
- Remove `python` from `matrix.use` in `nestbuildmatrix.yml`, because we now always use python.
- Add `no-python-in-kernel` check to `nestbuildmatrix.yml` to verify that no files linked into `libnestkernel` include `Python.h` to protect against Python-dependencies in NEST C++ code.
- Remove `libneurosim` from our test matrix. The test (and `ConnectionGenerator`) haven't worked for a very long time and now it breaks the new config. Since we have decided to remove it entirely, I see no point in fixing the test setup around it.


JanVogelsang (approved):

heplesser:
Comments from meeting with @terhorstd and others:

- [x] Add a flag that allows NEST to be built without Python, especially also in the absence of `Python.h`, to ensure that we avoid accidentally introducing Python-dependencies into the NEST core.
- [x] Revisit the cythonize flag: it should be possible to run cython manually.
- [x] Should one, instead of `-Dwith-xyz=ON|OFF|Path` use `-Dwith-xyz=ON|OFF` plus `-Dwith-xyz-path=Path`?
- [x] Ensure Python requirement in `pyproject.toml` is consistent with CMake

heplesser:
@terhorstd @JanVogelsang @JoseJVS I have now worked on the open issues from the last discussion:

- Protection against "Python creep" into the kernel or models is now done by a new `no-python-in-kernel` check in the CI setup, which verifies that no C++ files include `Python.h`. This seems to me a cleaner and cheaper solution that one that would require building NEST in the absence of Python just to see it fail (and in almost all cases pass without doing anything).
- I re-instated the possibility to use a pre-built cxx file in case Cython cannot be run on the build system. The flag is now called `-Dwith-prebuilt-pynest-cxx` and follows the usual rules for ON/OFF/<path> options, default OFF.
- pyproject.toml now requires Python 3.10, in sync with what we require in CMake

Concerning the question of whether to have three-way `ON/OFF/<path>` options or a combination of `ON/OFF` option with a separate `<path>` option, I asked Claude for recommendations and using binary `ON/OFF` is strongly recommended and consistent with large-scale projects such as LLVM and CMake itself. For specific paths, one can then use CMakes own variables:

| Go with the two-variable pattern, using CMake's own naming convention so it's discoverable:
```
cmake
option(WITH_LIB_A "Enable support for libA" ON)
# No custom path variable needed if you just tell users to use LibA_ROOT / CMAKE_PREFIX_PATH

if(WITH_LIB_A)
    find_package(LibA REQUIRED)
endif()
``` 
| Then document that a specific install location is chosen via `-DLibA_ROOT=/path/to/liba` (or `CMAKE_PREFIX_PATH`), not via a custom variable of your own. This gives you the click-together boolean semantics, reuses infrastructure users already know, and avoids writing/maintaining a parser for a hybrid variable type.

I put this PR back into draft mode until that change is in place.

heplesser:
Ok, all should be ready now, see updated description at the top.

cmake/ConfigureSummary.cmake
```diff
@@ -243,11 +228,9 @@ function( NEST_PRINT_CONFIG_SUMMARY )
...
     message( "" )
   endif()
-  if ( HAVE_PYTHON )
-    message( "PyNEST will be installed to:" )
-    message( "    ${CMAKE_INSTALL_PREFIX}/${PYEXECDIR}" )
-    message( "" )
-  endif ()
+  message( "PyNEST will be installed to:" )
+  message( "    ${CMAKE_INSTALL_PREFIX}/${PYEXECDIR}" )
+  message( "" )
```
JoseJVS: If we do not compile the Python binding this message should not appear

cmake/ConfigureSummary.cmake
```diff
@@ -42,22 +42,13 @@ function( NEST_PRINT_CONFIG_SUMMARY )
...
-      message( "    Cython          : No. Make sure to cythonize `pynestkernel.pyx` manually." )
-    endif ()
-    if ( HAVE_MPI4PY )
-      message( "    MPI4Py          : Yes (${PY_MPI4PY}/include)" )
-    else()
-      message( "    MPI4Py          : No " )
-    endif ()
+  message( "Python bindings     : Yes (Python ${Python_VERSION}: ${PYTHON})" )
+  message( "    Includes        : ${Python_INCLUDE_DIRS}" )
+  message( "    Libraries       : ${Python_LIBRARIES}" )
```
JoseJVS: If no binding is compiled this message should not be printed

cmake/ConfigureSummary.cmake
```diff
@@ -275,6 +258,19 @@ function( NEST_PRINT_CONFIG_SUMMARY )
     message( "" )
   endif ()
 
+  if ( NOT NEST_PREBUILT_PYNEST_CXX STREQUAL ""
```
JoseJVS: Shouldn't this be an error?

JanVogelsang:
(Removing myself as a reviewer as I currently do not have the capacity to do an in-depth review and don't know CMake well enough to judge whether some of the changes are implemented optimally. From a functional perspective, I agree to the changes made in the PR. If required, I can possibly review this PR in an AI-assisted fashion during the September Hackathon.

cmake/ConfigureSummary.cmake
heplesser: No, this is a corner case, but a legal an intended one. The logic is as follows:
- In almost all cases, the default is used and `cython` will be run to generate `<nest_build_dir>/pynest/nestkernel_api.cxx`.
- In special cases, the user may want to generate that file by manually running `cython` and then telling CMake where to find it. In this situation, we have again two subcases:
    - The generated `cxx` file is already in place when CMake is executed. Then, CMake finds that the file exists and all is well.
    - The `cxx` file has not yet been put in place when CMake runs. This should be possible so that the user can place the `cxx` file into its default location `<nest_build_dir>/pynest/`. Here we have a chicken-and-egg problem, since that directory is only created by CMake. Therefore the warning instead of the error.

But looking at this, I noticed that there still is a variable `NEST_PREBUILT_PYNEST_CXX` around from an earlier iteration, while now all should be based on `NESTKERNEL_API_CXX`. I revised that now.

heplesser (commented):
@JoseJVS Thanks, see my inline comment.

.github/workflows/nestbuildmatrix.yml
```diff
@@ -513,13 +531,13 @@ jobs:
...
-          - "python, optimize, warning"
-          - "python, boost, optimize, warning"
-          - "openmp, python, gsl, ltdl, boost, optimize, warning"
-          - "mpi, python, gsl, ltdl, boost, optimize, warning"
-          - "openmp, mpi, python, gsl, ltdl, boost, hdf5, sionlib, libneurosim, optimize, warning, music, detailed-timers, threaded-timers, mpi-sync-timer"
+          - "optimize, warning"
+          - "boost, optimize, warning"
+          - "openmp, gsl, ltdl, boost, optimize, warning"
+          - "mpi, gsl, ltdl, boost, optimize, warning"
+          - "openmp, mpi, gsl, ltdl, boost, hdf5, sionlib, optimize, warning, music, detailed-timers, threaded-timers, mpi-sync-timer"
```
terhorstd: What is the reason for removing `libneurosim`? Would it be possible to make this decision in another PR to decouple from the refactoring?

.github/workflows/nestbuildmatrix.yml
```diff
@@ -837,15 +848,15 @@ jobs:
...
               -Dwith-gsl=${{ contains(matrix.use, 'gsl') && 'ON' || 'OFF' }} \
               -Dwith-ltdl=${{ contains(matrix.use, 'ltdl') && 'ON' || 'OFF' }} \
               -Dwith-hdf5=${{ contains(matrix.use, 'hdf5') && 'ON' || 'OFF' }} \
-              -Dwith-sionlib=${{ contains(matrix.use, 'sionlib') && '$HOME/.cache/sionlib.install' || 'OFF' }} \
-              -Dwith-libneurosim=${{ contains(matrix.use, 'libneurosim') && '$HOME/.cache/libneurosim.install' || 'OFF' }} \
-              -Dwith-music=${{ contains(matrix.use, 'music') && '$HOME/.cache/music.install' || 'OFF' }} \
+              -Dwith-sionlib=${{ contains(matrix.use, 'sionlib') && 'ON' || 'OFF' }} \
+              -DSIONlib_ROOT=$HOME/.cache/sionlib.install \
+              -Dwith-music=${{ contains(matrix.use, 'music') && 'ON' || 'OFF' }} \
+              -DMusic_ROOT=$HOME/.cache/music.install \
```
terhorstd: Also here libneurosim is silently dropped...

cmake/ConfigureSummary.cmake
```diff
@@ -42,22 +42,13 @@ function( NEST_PRINT_CONFIG_SUMMARY )
...
-      message( "    MPI4Py          : No " )
-    endif ()
+  message( "Python bindings     : Yes (Python ${Python_VERSION}: ${PYTHON})" )
+  message( "    Includes        : ${Python_INCLUDE_DIRS}" )
+  message( "    Libraries       : ${Python_LIBRARIES}" )
+  if ( NOT NESTKERNEL_API_CXX )
+    message( "    Cython          : Yes (Cython ${CYTHON_VERSION}: ${CYTHON_EXECUTABLE})" )
   else ()
-    message( "Python bindings     : No" )
+    message( "    Cython          : Not run; using pre-generated ${NESTKERNEL_API_CXX}" )
```
terhorstd: Wouldn't it make sense to keep the informative `mpi4py` choice, given our history of users problems with MPI setup? There are also a bunch of occurrences of `mpi4py` in various other places. If this is to be removed, another PR that removes mpi4py first would disentangle this decision.

cmake/ProcessOptions.cmake
```diff
@@ -410,209 +446,176 @@ function( NEST_PROCESS_WITH_MPI )
...
-
-      # export found variables to parent scope
-      set( LIBNEUROSIM_FOUND "${LIBNEUROSIM_FOUND}" PARENT_SCOPE )
-      set( LIBNEUROSIM_LIBRARIES "${LIBNEUROSIM_LIBRARIES}" PARENT_SCOPE )
-      set( LIBNEUROSIM_INCLUDE_DIRS "${LIBNEUROSIM_INCLUDE_DIRS}" PARENT_SCOPE )
-      set( LIBNEUROSIM_VERSION "${LIBNEUROSIM_VERSION}" PARENT_SCOPE )
-    endif ()
-  endif ()
+  nest_restrict_find_if_pinned( LibNeurosim )
+  if ( NOT with-libneurosim )
```
terhorstd: Here, libneurosim is kept…

cmake/ProcessOptions.cmake
```diff
@@ -410,209 +446,176 @@ function( NEST_PROCESS_WITH_MPI )
...
     set( BUILTIN_MODELS ${with-models} )
-  else()
+  else ()
     if ( NOT EXISTS "${PROJECT_SOURCE_DIR}/modelsets/${with-modelset}" )
       printError( "Cannot find modelset configuration 'modelsets/${with-modelset}'" )
     endif ()
-    file(STRINGS "${PROJECT_SOURCE_DIR}/modelsets/${with-modelset}" BUILTIN_MODELS)
-  endif()
+    file( STRINGS "${PROJECT_SOURCE_DIR}/modelsets/${with-modelset}" BUILTIN_MODELS )
+  endif ()
```
terhorstd: Minor remark: The introduction of additional space around parentheses breaks convention. However, we'd need a format checker on cmake files to sanction this appropriately.

nestkernel/nest_time.cpp
```diff
@@ -25,16 +25,13 @@
...
-
 // Includes from libnestutil:
 #include "numerics.h"
 
 using namespace nest;
 
-const double Time::Range::TICS_PER_MS_DEFAULT = CONFIG_TICS_PER_MS;
-const tic_t Time::Range::TICS_PER_STEP_DEFAULT = CONFIG_TICS_PER_STEP;
+const double Time::Range::TICS_PER_MS_DEFAULT = 1000.0;
+const tic_t Time::Range::TICS_PER_STEP_DEFAULT = 100;
```
JoseJVS: Shouldn't these two be constant expressions?
```suggestion
constexpr double Time::Range::TICS_PER_MS_DEFAULT = 1000.0;
constexpr tic_t Time::Range::TICS_PER_STEP_DEFAULT = 100;
```

doc/htmldoc/whats_new/v3.11/index.rst
JoseJVS, line 22: Is this still correct, I no longer see the HAVE_EXPM1 definition in libnestutil/config.h.in?

heplesser: Yes, it is still correct. `numerics::expm1()` calls `std::expm1()` and since this is guaranteed to be available, we have no longer a `HAVE_EXPM1` guard. Previously, we had that guard but actually always used our hand-coded `expm1()` implementation by accident. That is gone now per 23a1b0d1c011a12adff7b814e783740e0f64147c.

nestkernel/nest_time.cpp
heplesser: Good catch! I noticed a few more possible `constexpr` a few lines down, will push that shortly.

doc/htmldoc/whats_new/v3.11/index.rst
JoseJVS: I see, thanks for the clarification, then I would remove the last sentence of this paragraph as there is no HAVE_EXPM1 guard to undefine:
> The old code is still available; undefine ``HAVE_EXPM1`` in ``<build_dir>/libnestutil/config.h`` to use it on systems where ``std::expm1()`` is available.

heplesser:
@JoseJVS I have now implemented the changes we discussed in person. Could you re-review?

JoseJVS (approved):
Thanks a lot for the great work!

clinssen:
Hi, could it be that this PR broke the extant extension module build infrastructure? It seems that NESTML CI runs are failing since this merge. Could the PR at https://github.com/nest/nest-extension-module/pull/42 be brought up-to-date so we can make the same changes on the NESTML side? Many thanks!

PR #3913, opened 2026-06-28
CMake: Provide support for FindPackage(NESTSimulator) and modernize nest-config generation

heplesser:
Part 11 of the CMake series (consider after #3912 is done) modernizes the creation of nest-config by extracting information from CMake high-level objects such as `Boost::headers` only at the last moment. Furthermore, it creates the files need so that `FindPackage` can find nest. Pertaining files are generated by the same `nest_export_module_build_config()` function as nest-config to ensure coherence. Files are installed to `<install>/lib/cmake/NESTSimulator` and can then be used with

```
find_package(NESTSimulator REQUIRED)
add_library(mymodule MODULE mymodule.cpp)
target_link_libraries(mymodule PRIVATE NEST::nest)
```

See also https://github.com/nest/nest-extension-module/pull/42 for an early use of the FindPackage() support.

As this touches on the extension module mechanism, I add @clinssen and @pnbabu as reviewers.


heplesser:
See also https://github.com/nest/nest-extension-module/pull/42 for an early use of the `FindPackage()` support.

We should discuss if we really need `nest-config` in the long run. It is only available after we have run CMake (and make install, actually), and then the CMake CONFIG file is also in place. So the only advantage is that `nest-config` is usable also outside the CMake universe.

CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ set( with-intel-compiler-flags OFF CACHE STRING "User defined flags for the Inte
 set( with-libraries OFF CACHE STRING "Link additional libraries [default=OFF]. Give full path. Separate multiple libraries by ';'." )
 set( with-includes OFF CACHE STRING "Add additional include paths [default=OFF]. Give full path without '-I'. Separate multiple include paths by ';'." )
 set( with-defines OFF CACHE STRING "Additional defines, e.g. '-DXYZ=1' [default=OFF]. Separate multiple defines by ';'." )
+set( CMAKE_INSTALL_MESSAGE LAZY CACHE STRING "Verbosity of cmake --install output [default=LAZY]." )
```
JanVogelsang: This new options needs documentation.

heplesser: Done!

JanVogelsang (approved):

PR #3916, opened 2026-06-28
CMake: Directly use math functions and constants from std:: where possible

heplesser:
This is a follow-up to #3911, but builds on #3913. It is the 12th (and for now last) PR in the big CMake makeover. This PR replaces forwarding via our `numerics.h` header with direct access to `std::` functions or constants where possible.

I am actually not quite certain if it wouldn't have been smarter to leave at least `e` and `pi` in `numerics`, in case we want to add `pi_half` or `e_squared` in a consistent way at some point. And `std::nan("")` isn't very pretty either.

This also has side-effects for NESTML, thus including @pnbabu and @clinssen .


JanVogelsang (approved):

JanVogelsang (approved):
Great series of changes!
I reviewed the whole series and didn't identify any flaws. I just a minor comment in #3913, that's it. After my own review, I also asked Github Copilot to review all changes and it also didn't find any flaws that you were not already aware of (e.g. NESTML compatibility due to the "numerics.h" changes).

heplesser:
@terhorstd @JanVogelsang  I converted this to draft status to ensure that we discuss in a broader forum if we really want to move e, pi, nan out of our `numerics` namespace, see my remark in the PR description.

heplesser:
I have thought a bit more about this and have come to the conclusion that our code will remain more readable if all numerical values are defined in a single namespace. With the current setup, several values come from `std::numbers` and others still from our own `numerics` namespace. This seems messier than needed and I will therefore close this PR.

PR #3922, opened 2026-07-03
Improve docs search for AI and search engines

jessica-mitchell:
This PR adds an llms.txt file which is supposed to help agents and coding assistants avoid scraping HTML 
The llms.txt provides a lightweight,  clean Markdown index map of your site's structure and core links

There is a sphinx-extenstion which would have the benefit of auto updating links, but it provides a long list of all file links, and I think it would serve us better to curate this file manually so we can provide extra context.

The meta directives are supposed to help with search engines as it adds a meta description tag into the HTML. This is what will be shown on results page.  Only added to some main pages now.


doc/htmldoc/llms.txt
```diff
@@ -0,0 +1,143 @@
...
+
+## Troubleshooting
+- [Installation Issues](installation/index.html#troubleshooting)
+- [Import Errors](troubleshooting.html#import-error)
+- [MPI Configuration](hpc/parallel_computing.html#mpi-setup)
+- [Performance Issues](hpc/optimize_performance.html)
+
+## Community
+- [GitHub Issues](https://github.com/nest/nest-simulator/issues)
+- [Mailing List](https://www.nest-initiative.org/mailinglist)
```
jessica-mitchell:
```suggestion
- [Mailing List](https://www.nest-simulator.org/mailinglist)
```

jessica-mitchell:
This PR only touches docs so I will merge with no reviewers

PR #3923, opened 2026-07-03
Add warning banners to non-stable and main versions of docs on Read the Docs

### LINKED ISSUE #3377: Version warning not showing up on Read the Docs
In the Read the Docs config, version warnings should appear for older versions of nest. But they do not
image

jessica-mitchell:
Fixes #3377 

Read the Docs has a custom script add-on that can be used to inject js into doc pages. It is loaded asynchronously after the initial page load.

Our current theme does not allow us to use the default banners provided by Read the Docs.
A URL is addd tothe add-on settings page in Read the Docs, pointing to main where the js file will live, and then should be good to go. (See screenshot on test fork for how it would look).

**One caveat** with current setup this warning banner applies to all numbered versions including the most recent release (nest-simulator.readthedocs.io/en/3.10/ gets a warning). To accommodate this, the warning says you may be on and "outdated or non-stable" version with link to stable. We could modify the script to exclude the 3.10, but we would have to update this at every release.  

The colours are not 'nest colours' but I think that makes it stand out more, but they can be adjusted.

Screenshot from 2026-09-09 13-38-23

image

### DIFF
--- a/doc/htmldoc/static/js/rtd-version-banner.js
+++ b/doc/htmldoc/static/js/rtd-version-banner.js
@@ -1,0 +1,161 @@
+/**
+ * Non-stable version warning banner for the NEST Simulator documentation.
+ *
+ * Read the Docs' built-in "addons" version-warning notification does not
+ * render correctly with the sphinx_material theme used here, so this script
+ * is a workaround: it is loaded as a Read the Docs "Custom Script"
+ * (Admin -> Settings -> Addons -> Custom Script), which injects it as a
+ * <script> tag on every page of every version, including old versions that
+ * are never rebuilt.
+ *
+ * Deploy: point the Custom Script setting at the built copy of this file on
+ * a version that is rebuilt on every change (e.g.
+ * https://nest-simulator.readthedocs.io/en/main/_static/js/rtd-version-banner.js),
+ * so all versions always load the latest banner logic.
+ *
+ * Everything (styles included) lives in this one file because the Custom
+ * Script setting only accepts a single script URL -- there is nowhere to
+ * also register a separate stylesheet.
+ *
+ * See: https://docs.readthedocs.com/platform/latest/custom-script.html
+ */
+(function () {
+  "use strict";
+
+  if (document.querySelector(".rtd-version-banner")) {
+    return;
+  }
+
+  var STABLE_DOCS_URL = "https://nest-simulator.readthedocs.io/en/stable/";
+  var MAIN_DOCS_URL = "https://nest-simulator.readthedocs.io/en/main/";
+
+  // Read the Docs URLs always follow /en/<version>/..., so the version slug
+  // is read straight from the path. This avoids depending on any RTD
+  // addons/meta-tag API, which may not be available if addons are broken
+  // for this theme (the reason this workaround exists in the first place).
+  function getVersionSlug() {
+    var parts = window.location.pathname.split("/").filter(Boolean);
+    var enIndex = parts.indexOf("en");
+    var slug = enIndex !== -1 ? parts[enIndex + 1] : parts[0];
+    if (!slug || !/^[a-zA-Z0-9_.-]+$/.test(slug)) {
+      return null;
+    }
+    return slug;
+  }
+
+  var slug = getVersionSlug();
+  if (!slug || slug === "stable") {
+    return;
+  }
+
+  var isDevelopment = slug === "main" || slug === "latest";
+  var bannerType = isDevelopment ? "development" : "outdated";
+  var dismissKey = "rtd-banner-dismissed-" + bannerType;
+
+  try {
+    if (window.localStorage && window.localStorage.getItem(dismissKey) === "true") {
+      return;
+    }
+  } catch (e) {
+    // localStorage unavailable (e.g. privacy mode) -- fall through and show the banner.
+  }
+
+  function injectStyles() {
+    var style = document.createElement("style");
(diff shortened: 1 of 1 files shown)


doc/htmldoc/static/js/rtd-version-banner.js
```diff
@@ -0,0 +1,161 @@
...
+    var style = document.createElement("style");
+    style.textContent =
+      ".rtd-version-banner{position:fixed;left:0;right:0;bottom:0;z-index:10000;" +
+      "display:flex;align-items:center;justify-content:center;gap:1rem;" +
+      "padding:0.75rem 1.25rem;" +
+      "padding-bottom:calc(0.75rem + env(safe-area-inset-bottom, 0px));" +
+      "font-family:'Roboto','Helvetica Neue',Helvetica,Arial,sans-serif;" +
+      "font-size:0.875rem;line-height:1.4;" +
+      "box-shadow:0 -2px 6px rgba(0,0,0,0.15);}" +
+      ".rtd-version-banner--development{background-color:#2196f3;color:#fff;}" +
```
jessica-mitchell:
```suggestion
      ".rtd-version-banner--development{background-color:#16c8ff;color:#fff;}" +
```

doc/htmldoc/static/js/rtd-version-banner.js
```diff
@@ -0,0 +1,161 @@
...
+      ".rtd-version-banner--outdated{background-color:#ff9800;color:#212121;}" +
```
jessica-mitchell:
```suggestion
      ".rtd-version-banner--outdated{background-color:#ffec51;color:#212121;}" +
```

JoseJVS:
> One caveat with current setup this warning banner applies to all numbered versions including the most recent release (nest-simulator.readthedocs.io/en/3.10/ gets a warning). To accommodate this, the warning says you may be on and "outdated or non-stable" version with link to stable. We could modify the script to exclude the 3.10, but we would have to update this at every release.

If its not too much work you could actually try to fetch the latest tag using the REST API of GH (see https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#get-the-latest-release ).
To not have to use octokit you could just use a fetch request with a header similar to the one used for curl e.g.:

```javascript
async function getLatestNESTRelease() {
  const headers = {
    Accept: "application/vnd.github+json",
    "X-GitHub-Api-Version": "2026-03-10",
  };

  const response = await fetch(
    `https://api.github.com/repos/nest/nest-simulator/releases/latest`,
    { headers },
  );

  if (!response.ok) {
      const body = await response.text();
      throw new Error(`GitHub API error ${response.status}: ${body}`);
  }

  return response.json();
}

try {
  const release = await getLatestNESTRelease();

  console.log("Version:", release.tag_name);
} catch (error) {
  console.error(error.message);
}
```
(Full example generated with AI)

You could compare the numbered version with the release tag name property and only show the warning banner if they do not match, and we would only need to update the ```getLatestNESTRelease``` function if the current API version of GH is no longer supported (or we somehow rename our repo).

**The caveat here is that if the GH API is down then the warning label would be printed regardless if it is the latest version or not.**

What do you think?

doc/htmldoc/static/js/rtd-version-banner.js
```diff
@@ -0,0 +1,308 @@
...
+      ".rtd-version-banner{position:fixed;left:0;right:0;bottom:0;z-index:10000;" +
+      "display:flex;align-items:center;justify-content:center;gap:1rem;" +
+      "padding:0.75rem 1.25rem;" +
+      "padding-bottom:calc(0.75rem + env(safe-area-inset-bottom, 0px));" +
+      "font-family:'Roboto','Helvetica Neue',Helvetica,Arial,sans-serif;" +
+      "font-size:0.875rem;line-height:1.4;" +
+      "box-shadow:0 -2px 6px rgba(0,0,0,0.15);}" +
+      ".rtd-version-banner--development{background-color:#16c8ff;color:#fff;}" +
+      ".rtd-version-banner--outdated,.rtd-version-banner--prerelease" +
+      "{background-color:#ff9800;color:#212121;}" +
```
jessica-mitchell:
```suggestion
      "{background-color:#ffdd27;color:#212121;}" +
```

doc/htmldoc/static/js/rtd-version-banner.js
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  var STABLE_DOCS_URL = "https://nest-simulator.readthedocs.io/en/stable/";
+
+  // Fallback for when the releases API is unreachable, rate-limited or slow.
+  // Bump this when tagging a release: it keeps the banner correct offline and
+  // covers the cases the API cannot (a tag with no published GitHub release,
+  // or a patch for an older line published after a newer minor, either of
+  // which would make releases/latest point at the wrong version).
+  var LATEST_KNOWN_RELEASE = "v3.10";
```
JoseJVS: As we discussed please remove this, and decide whether as a fallback no banner should be displayed or the outdated banner, but let's avoid hard-coding versions inside of here.
```suggestion
```

jessica-mitchell: done. I decided to keep the banner off if if it fails that way it won't ever show the banner when it shouldnt; not sure if it's the best way to go but we can always change it later

JoseJVS (approved):
Thanks for the work!

PR #3924, opened 2026-07-08
Upgrade Read the Docs and Sphinx configs for docs

jessica-mitchell:
This PR makes a few updates to requirements/config for docs including the Ubuntu, Python, plantuml in readthedocs.yml and Sphinx in requirements.txt; this required removing the carousel extension, which is incompatible with newer Sphinx extensions and no longer maintained. It was replaced with bootstrap carousel. Other changes made to extensions also made


JoseJVS (approved):
Thanks for the update!

Issue #3802, opened 2026-04-13
ResetKernel() does not clear state between module loading


clinssen:
I'm running into a strange issue pertaining to module loading and unloading. It seems that the internal state of NEST is not completely reset when calling ``ResetKernel()``.

I am generating code for two variants of a NESTML model, let's say ``a_synapse.nestml`` and ``b_synapse.nestml``. I'm generating code for one, compiling and building it into ``nestmlmodule.so``, and then running ``nest.Install("/absolute/path/to/nestmlmodule.so")`` and running a simulation. Next, I call ``nest.ResetKernel()`` and do the same for the second model. The issue is that the second time the module gets loaded (with the same filename, but different contents), everything is as if the first module is the one that had been loaded. This is dependent on there being a ``nest.CopyModel("a_synapse", "foo_synapse")`` just before the call to ResetKernel.

Actual result:

```
>>> import nest;nest.ResetKernel();nest.Install("/absolute/path/to/nestmlmodule.so")
MODULE NAME "nestmlmodule"
REGISTERING SYNAPSE "a_synapse"
>>> nest.CopyModel("a_synapse", "foo_synapse")
```

So far so good. Now I generate the code for b_synapse and compile it into a fresh .so by the same name (``/absolute/path/to/nestmlmodule.so``). Then:

```
>>> import nest;nest.ResetKernel();nest.Install("/absolute/path/to/nestmlmodule.so")
MODULE NAME "nestmlmodule"
REGISTERING SYNAPSE "a_synapse"
```

This is wrong; it should say "b_synapse". If I launch a fresh NEST process and run the second code block in isolation, it correctly reports having registered "b_synapse".

I have already tried to fix this issue by passing a "keep symbols local" parameter to lt_dlopen as follows:
```C++
lt_dladvise advise;
lt_dladvise_init(&advise);
lt_dladvise_local(&advise); // <-------

const lt_dlhandle hModule = lt_dlopenadvise(name.c_str(), advise);
lt_dladvise_destroy(&advise);
```
but it did not change the behaviour.

JanVogelsang:
Unfortunately, I don't have any idea why this happens. I just want to add that I had some issues with `ResetKernel` in the std::variant PR as well, as some strange things were happening with the manager `finalize` calls. In the end I believed the issue was the order in which `finalize` was called which changed in the PR. But I think I never actually found out what exactly was causing the issue.

med-ayssar:
[...]

PR #3929, opened 2026-07-27
Add documentation about module unloading behaviour

clinssen:
Partial (but not complete) fix for #3802.

### DIFF
--- a/pynest/nest/lib/hl_api_simulation.py
+++ b/pynest/nest/lib/hl_api_simulation.py
@@ -182,8 +182,14 @@ def ResetKernel():
     * time
     * random generators
 
-    are reset. All dynamically loaded modules (via :py:func:`.Install()`) are unloaded.
+    are reset.
 
+    Additionally, all dynamically loaded modules (via :py:func:`.Install()`) are
+    unloaded. However, note that some systems do not unload dynamic library symbols
+    (see https://github.com/nest/nest-simulator/issues/3802); as a result of this,
+    it is recommended to always use unique names for models inside modules, or,
+    alternatively, use the ``-fno-gnu-unique`` flag when building NEST extension
+    modules.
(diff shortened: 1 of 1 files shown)


pynest/nest/lib/hl_api_simulation.py
```diff
@@ -182,8 +182,14 @@ def ResetKernel():
...
     * random generators
 
-    are reset. All dynamically loaded modules (via :py:func:`.Install()`) are unloaded.
-
+    are reset.
+
+    Additionally, all dynamically loaded modules (via :py:func:`.Install()`) are
+    unloaded. However, note that some systems do not unload dynamic library symbols
+    (see https://github.com/nest/nest-simulator/issues/3802); as a result of this,
+    it is recommended to always use unique names for models inside modules, or,
```
heplesser, line 190: I am confused about the "use unique names" recommendation here. In your example in #3802 you once had `a_synapse` and once `b_synapse` and this did not work as expected with `ResetKernel`—the `a_synapse` persisted. So in what sense and for what purpose do you recommend unique names? 

And isn't the right solution to always build with `-fno-gnu-unique` (or corresponding flags for other compilers) when building an extension module to properly avoid the problem?

clinssen: The idea of unique names would be that even though ``a_synapse`` may persist, that would be okay, as we would not be re-using that name, but use ``b_synapse`` instead for the next run (or imagine a hash instead of the single letters). Admittedly it's not a very elegant workaround, but it's better than not having a workaround at all, so I feel that it merits a mention.

I would be happy to always enable ``-fno-gnu-unique``, but where should we document the rationale for using this flag? Perhaps on the NESTML and NEST Extension Module readthedocs pages?

PR #3932, opened 2026-08-13
Add explanation for obtaining values from parameter objects

Pedro-RP:
Added an admonition explaining how to obtain specific values from a parameter object in NEST.


heplesser (changes requested):
@Pedro-RP Thank you for your PR! It explains an important detail that we haven't document well so far. Still, I'd like to make a few suggestions for further improvement:
- I think this could be a subheading under "Random parameters" instead of an admonition.
- Focus on what you can do, not what NumPy is doing.
- Add a small example.

While looking at the doc page, I notice that in the table in the "Mathematical functions" part of the page, the table shows `nest.random.exp(x)` etc, while it should be `nest.math.exp(x)`. Could you fix that as part of this PR as well?

jessica-mitchell:
@heplesser should this really be under random parameters? isn't is applicable to all parameter objects?

heplesser:
> @heplesser should this really be under random parameters? isn't is applicable to all parameter objects?

Spatial parameters cannot be samples as such, since they always need distances between neurons. So the `GetValue()` approach only applies to random parameters and combinations of those that do not include any `nest.spatial` parts. Strictly speaking, it also applies to the constant parameter, but that seems to me rather irrelevant.

jessica-mitchell (approved):
thanks!

doc/htmldoc/neurons/parametrization.rst
```diff
@@ -86,6 +86,17 @@ distribution.
...
+.. _returning_random_param_values:
+
+Obtaining specific values from a parameter object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Parameter objects in NEST are only converted to concrete values
+internally, after nodes are created and/or connections are defined.
+
+To check if a parameter is drawn from the expected distribution, or to
+generate example samples, use :py:meth:`GetValue()<nest.lib.hl_api_types.Parameter.GetValue>`.
```
heplesser: Could you add a little example here?

doc/htmldoc/neurons/parametrization.rst
```diff
@@ -86,6 +86,17 @@ distribution.
 
 .. image:: ../static/img/NEST3_13_0.png
 
+.. _returning_random_param_values:
+
+Obtaining specific values from a parameter object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Parameter objects in NEST are only converted to concrete values
+internally, after nodes are created and/or connections are defined.
```
heplesser:
```suggestion
internally after nodes are created and/or connections are defined.
```

doc/htmldoc/neurons/parametrization.rst
```diff
@@ -281,7 +292,7 @@ Mathematical functions
 +============================+=============================================+
 | ::                         |                                             |
 |                            |                                             |
-|     nest.random.exp(x)     | | Calculates the exponential of a parameter |
+|     nest.math.exp(x)     | | Calculates the exponential of a parameter   |
```
heplesser: cos() and sin() below need the same fix.

heplesser (changes requested):
Thanks for the revision, just a few small things left, see below.

doc/htmldoc/neurons/parametrization.rst
Pedro-RP: The linked page that mentions the GetValue() function already includes an example of how to use it. I considered repeating that example or adding a new one, but decided it would be redundant. What do you think?

heplesser: OK!

heplesser (approved):

jessica-mitchell:
@Pedro-RP there is a failing test with isort - can you fix the order of imports so it will pass 
```
  Error: ERROR: /home/runner/work/nest-simulator/nest-simulator/pynest/nest/spatial/__init__.py Imports are incorrectly sorted and/or formatted.
  --- /home/runner/work/nest-simulator/nest-simulator/pynest/nest/spatial/__init__.py:before	2026-09-07 11:19:53.358506
  +++ /home/runner/work/nest-simulator/nest-simulator/pynest/nest/spatial/__init__.py:after	2026-09-07 11:19:57.824946
  @@ -21,8 +21,8 @@
   
   import functools as _functools
   
  +from .hl_api_spatial import *  # noqa: F401,F403
   from .hl_api_spatial import DistanceParameter as _DistanceParameter
  -from .hl_api_spatial import *  # noqa: F401,F403
   
```

PR #3849, opened 2026-05-26
Add pre-commit hooks

terhorstd:
Preparatory work for #3848:
* add more linters to `.pre-commit-config.yaml` → tool coverage
* configure tools to actually scan all files → file coverage
* keeps failing tests minimal, but fixes smaller problems → base issue coverage

After this PR we can start to remove the ignores one by one and proceed with #3848.

PR #3935, opened 2026-08-19
Post-release updates from 3.10 

terhorstd:
These are the version bump and DOI changes imposed by the NEST 3.10 release.


heplesser (approved):

heplesser:
@terhorstd The test for the merged code on macOS currently fails because the macOS runner uses recent Python packages, in particular SciPy 1.18.0, which as a regression. This was fixed in main by https://github.com/nest/nest-simulator/pull/3881, but changes to requirements handling in https://github.com/nest/nest-simulator/pull/3849 broke this (see https://github.com/nest/nest-simulator/pull/3849/changes#diff-89de577a8ded6cb539f7c66e2989d107870f3dc0d96a3f12d40fd8ccd1c92687).

Right now, Python dependencies are specified via `-r requirements_testing.py`, which installs SciPy only as a dependency, thus ignoring the `!=1.18.0` prohibition in `requirements_pynest.py`. If think the correct approach here would be to change

https://github.com/nest/nest-simulator/blob/d6074ff3cd7233644aaf5c84b094e117d4609d0b/.github/workflows/nestbuildmatrix.yml#L824

to

```
python -m pip install --force-reinstall --upgrade -r requirements_pynest.txt -r requirements_testing.txt
```

I anyways find it strange that we never seem to install the PyNEST requirements before building NEST.

heplesser:
The cancelled test is due to a regression in SciPy 1.18.0. With #3937 merged, this is fixed in main, so this PR can be merged without rerunning all tests.

Issue #3936, opened 2026-08-21
Clean up dependency management


terhorstd:
**Describe the bug**
The CI installs packages in `nestbuildmatrix.yml`, which is not in sync with and only partially installs dependencies also mentioned in `requirements_*txt` files and `pyproject.toml`.

**Expected behavior**
Unifided declaration of dependencies based on required features.

**Additional context**
Docs also mention dependencies in `versions.json` and external package definitions (conda-forge, Spack, …) also need to be considered.

PR #3937, opened 2026-08-21
Fix pynest dependencies in macOS build

terhorstd:
This fixes the issue seen on #3935. The changes are only a symptomatic fix, the `nestbuildmatrix.yml` dependency management should be cleaned up in a separate PR. See #3936.


gtrensch (approved):
LGTM 👍

heplesser (approved):
Thanks!

PR #3938, opened 2026-08-26
Limit scipy to <1.18 due to regression in scipy

heplesser:
This is a follow-up to #3881 and #3937. In #3881, I had assumed that the problem would be fixed upstream by SciPy 1.18.1 since a PR already existed ((scipy/scipy#25468[https://github.com/scipy/scipy/pull/25468]), but that was clearly optimistic: The regression persists in SciPy 1.18.1.

Therefore, this PR changes the requirement from "!= 1.18.0" to "< 1.18".

This PR needs only on reviewer. Marked as Critical because it breaks CI on macOS.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -315,7 +315,7 @@ jobs:
 
       - name: "Install dependencies"
         run: |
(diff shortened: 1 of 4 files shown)


terhorstd (approved):
👍

PR #3942, opened 2026-09-02
Pin isort version to 9.0.1 and adjust import ordering

### LINKED ISSUE #3939: isort 9.0 fails on NEST code
isort 9.0 was released on 26 August and changes the order of `from foo import *` and `from foo import abc`, see https://isort.readthedocs.io/en/latest/changelog.html and in particular https://github.com/PyCQA/isort/pull/2619 (lines changed in the test show the change clearly).

This leads to isort failing on `pynest/nest/spatial/__init__py` here: https://github.com/nest/nest-simulator/actions/runs/33166026182/job/98831515152?pr=3027#step:3:56

and thus breaks our CI. I see three options:

- Adapt the code to isort 9.0 expectations (likely most sensible, but everyone needs to update their pre-commit setup to isort 9)
- Pin our isort-checker to isort < 9.0 (but we will then run into problems when developers have newer isort in their pre-commit setup)
- Mark the particular lines to be ignored by isort (not neat).

I mark this as critical because it breaks our CI.

heplesser:
This PR fixes #3939 by pinning isort to 9.0.1 and adapting import order in one file correspondingly.

This PR is critical because the problem blocks our CI. As single reviewer should suffice here.

### DIFF
--- a/.github/workflows/nestbuildmatrix.yml
+++ b/.github/workflows/nestbuildmatrix.yml
@@ -359,7 +359,7 @@ jobs:
       - name: "Run isort..."
         uses: isort/isort-action@24d8a7a51d33ca7f36c3f23598dafa33f7071326  # 1.1.1
         with:
-          isort-version: 8.0.1
+          isort-version: 9.0.1
           configuration: --profile=black --thirdparty="nest" --check-only --diff
 
   black:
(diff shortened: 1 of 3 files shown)


JoseJVS (approved):
Thanks for the quick fix, this looks good to me 👍

PR #3945, opened 2026-09-07
Root module cleanup and import-time creep

Helveg:
I am the original author that introduced lazy importing and kernel attributes to the root NEST module. It had 2 goals:

* Defer any actions other than starting the NEST kernel to AFTER `import nest` had completed. This required lazy importing of submodules, and lazy attribute accessors for submodules (for example `nest.Create` as a shortcut for `nest.hl_api_nodes.Create`). The rationale was to shave off import time which could take up to 2 seconds on slow machines and was at that time dropped to 40ms on the same slow machines.
* Create kernel attribute accessors to get and set kernel attributes simply as `nest.resolution` instead of `nest.GetStatus("resolution")`

Both required a dynamic module, which up until today is still only supported by what is considered by most as a "dirty hack": manipulating/replacing the module object. This is still the case as recent PEPs to try to allow for more module customizations were rejected with this pattern once more confirmed (https://peps.python.org/pep-0726/) with the rejection message reading:

> There are already workarounds that people can use to achieve similar results. That some people consider those workarounds such as replacing a module with another object to be gross is not considered a problem as it is an uncommon need.

The initial implementation achieved its goal but drastically reduced clarity by requiring at least 4 distinct patches on top of the already dirty hack in order to obtain a well behaved `nest` module object during bootstrap, at runtime, and during doc building (Sphinx is finnicky).

After more tinkering it turns out that all of these patches collapse if instead of **duplicating and reconstructing** a quasi-identical novel model object that we replace in `sys.modules`, we instead simply retype the existing module object with a `obj.__class__ = NestModule` reassignment.

This PR as a result can present a much cleaner `__init__.py` and adds full static typing on top (so IDEs and other tools can correctly autocomplete and inspect the public API of the `nest` module).  I've also extended the comments in the file to make sure any first time reader gets a thorough introduction to each of the mechanisms and the structure of the file.

I've also once more moved all optional imports of heavy modules that aren't on the main code path for most users (such as `matplotlib`, `pandas`, and `h5py`) into local imports to optimize import times from ~400ms to ~120ms on my machine. The last 120ms is mostly due to importing `numpy` which we can maybe avoid importing but it was deeper in the Cython code so I didn't touch it for now.

### DIFF
--- a/doc/htmldoc/_ext/extract_api_functions.py
+++ b/doc/htmldoc/_ext/extract_api_functions.py
@@ -21,7 +21,6 @@
 import ast
 import glob
 import os
-import re
 
 from sphinx.application import Sphinx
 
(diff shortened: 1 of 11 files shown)


Helveg:
It seems to be failing on black changes outside of my diff. Since it's just 2 hunks I'll add a commit addressing them, let me know if you want me to keep that out of this PR and address the failing checks another way.

Helveg:
PS: To interested readers who wonder why we are in the category of "uncommon needs": NEST has always had global state in the Python process and has always been a singleton kernel by design for many reasons. Using your root module as a representation of your singleton-kernel-self is logical; suddenly separating that (which is a Herculean task) just for a Python implementation detail of its module system to end up with a less ergonomic outcome doesn't add up.

PR #3946, opened 2026-09-08
Add ArchivingNode::pre_run_hook_() and call it from every model

JanVogelsang:
Gives ArchivingNode a protected pre_run_hook_() extension point (in addition to the existing `Node::pre_run_hook()`) and wires it into every model's own pre_run_hook() override.
The main reason for this PR is to reduce the changeset of the #2989 ax-delay PR.

### DIFF
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -66,8 +66,6 @@ nest::ArchivingNode::ArchivingNode( const ArchivingNode& n )
 void
 ArchivingNode::pre_run_hook_()
 {
-  // Intentionally empty for now; gives subclasses a place to hook into
-  // per-run setup without overriding Node::pre_run_hook() themselves.
 }
 
 void
--- a/nestkernel/archiving_node.cpp
+++ b/nestkernel/archiving_node.cpp
@@ -64,6 +64,13 @@ nest::ArchivingNode::ArchivingNode( const ArchivingNode& n )
 }
 
 void
+ArchivingNode::pre_run_hook_()
+{
+  // Intentionally empty for now; gives subclasses a place to hook into
+  // per-run setup without overriding Node::pre_run_hook() themselves.
+}
+
+void
 ArchivingNode::register_stdp_connection( double t_first_read, double delay )
 {
   // Mark all entries in the deque, which we will not read in future as read by
--- a/models/aeif_cond_alpha.cpp
+++ b/models/aeif_cond_alpha.cpp
@@ -424,6 +424,8 @@ nest::aeif_cond_alpha::init_buffers_()
 void
 nest::aeif_cond_alpha::pre_run_hook()
 {
+  ArchivingNode::pre_run_hook_();
+
   // ensures initialization in case mm connected after Simulate
   B_.logger_.init();
 
--- a/models/aeif_cond_alpha_astro.cpp
+++ b/models/aeif_cond_alpha_astro.cpp
@@ -426,6 +426,8 @@ nest::aeif_cond_alpha_astro::init_buffers_()
 void
 nest::aeif_cond_alpha_astro::pre_run_hook()
 {
+  ArchivingNode::pre_run_hook_();
+
   // ensures initialization in case mm connected after Simulate
   B_.logger_.init();
 
--- a/models/aeif_cond_alpha_multisynapse.cpp
+++ b/models/aeif_cond_alpha_multisynapse.cpp
@@ -435,6 +435,8 @@ aeif_cond_alpha_multisynapse::init_buffers_()
 void
 aeif_cond_alpha_multisynapse::pre_run_hook()
 {
+  ArchivingNode::pre_run_hook_();
+
   // ensures initialization in case mm connected after Simulate
   B_.logger_.init();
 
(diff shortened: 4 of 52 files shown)


nestkernel/archiving_node.cpp
```diff
@@ -63,6 +63,13 @@ nest::ArchivingNode::ArchivingNode( const ArchivingNode& n )
 {
 }
 
+void
+ArchivingNode::pre_run_hook_()
+{
+  // Intentionally empty for now; gives subclasses a place to hook into
+  // per-run setup without overriding Node::pre_run_hook() themselves.
```
JanVogelsang:
```suggestion
```

heplesser (approved):

clinssen:
NESTML models do not derive from the ArchivingNode (but from the StructuralPlasticityNode); thus, adding axonal delays machinery to the ArchivingNode could cause some difficulties with NESTML generated code. Could you say why this functionality is made part of the ArchivingNode, specifically? Should there instead be another class inbetween the StructuralPlasticityNode and the ArchivingNode in the hierarchy, containing the necessary machinery for axonal delays?

JanVogelsang:
For the axonal delay information, we introduced another "archive", this time of incoming presynaptic spikes. The `pre_run_hook` is then used to initialize this archive, which can only be done at the start of the simulation when it is known whether the neuron has any incoming connections with predominant axonal delays or not.

clinssen:
Would it be possible to cleanly separate this newly added functionality from the existing ArchivingNode by introducing another level in the class hierarchy as I suggested? Like a PreSpikesArchivingNode or an AxonalDelayArchivingNode. I think that could help a lot from the perspective of NESTML code generation.

Magnushst (changes requested):
Six potentially missing, they predate the branch point (last touched 2026-03 to 2026-05; branch base 2026-09-07):

| file | line | override |
|---|---|---|
| `models/iaf_bw_2001.cpp` | 380 | `nest::iaf_bw_2001::pre_run_hook()` |
| `models/iaf_bw_2001_exact.cpp` | 375 | `nest::iaf_bw_2001_exact::pre_run_hook()` |
| `models/iaf_tum_2000.cpp` | 298 | `nest::iaf_tum_2000::pre_run_hook()` |
| `models/ignore_and_fire.cpp` | 163 | `ignore_and_fire::pre_run_hook()` |
| `models/parrot_neuron.h` | 129 | inline; calls only `FlushEventMechanism::pre_run_hook()` |
| `models/parrot_neuron_ps.h` | 118 | inline; empty body, `// no variables` |

JanVogelsang:
> Would it be possible to cleanly separate this newly added functionality from the existing ArchivingNode by introducing another level in the class hierarchy as I suggested? Like a PreSpikesArchivingNode or an AxonalDelayArchivingNode. I think that could help a lot from the perspective of NESTML code generation.

Sounds as horrible as the current `ArchivingNode`/`StructuralPlasticityNode` hierarchy design, so it's probably a good idea to just do what you suggested and clean it up later in the Node refactoring that is planned anyway.

clinssen:
Haha, thanks for the thumbs up Jan and sorry for the extra work. I am responsible for disentangling the ArchivingNode into an ArchivingNode and StructuralPlasticityNode, which was done so that in NESTML, we can inherit from the latter without having to inherit all the unnecessary stuff contained in the former. After the axonal delays PR gets merged, it would be great if we could then inherit from this "AxonalDelayArchivingNode", which would sit between the ArchivingNode and the StructuralPlasticityNode in the class hierarchy. Many thanks in advance!

JanVogelsang:
@clinssen I definitely think it made sense to disentangle the two, but it would be even better to not have these two base classes at all and instead use composition instead of inheritance (e.g. via true composition or just Mixins).

clinssen:
Fully agreed.

JanVogelsang:
@clinssen Wouldn't it make more sense that NESTML would then just generate the axonal delay archive just as it generates the archive for other STDP models? Then we would just keep this axonal delay functionality inside the `ArchivingNode` to make it available to NEST built-in models and NESTML would just do its thing for the model it generates.

clinssen:
A fair suggestion, but I'd rather not copy large blocks of code verbatim, as this represents a maintainability hurdle. What if the code changes upstream—we'd have to manually copy everything over. It would be better if we could re-use as much existing NEST code. Even aside from the NESTML perspective, I would maintain that this needs a separate class because the functionality is completely unrelated to the postsynaptic spike archiving that the ArchivingNode is doing.

JanVogelsang:
Fair. But I wouldn't call it completely unrelated. It's both code that's exclusively used for STDP synapses.

clinssen:
Ah, but it's true that the functionality can be cleanly separated, with ArchivingNode staying mostly the way it is now?

The reason for not wanting the ArchivingNode functionality in NESTML, is that this code needs to be dynamically generated, as it depends on the details of the model (e.g. how many trace variables there are). For the axonal delays functionality, I assume that NESTML does not need to dynamically change anything about this code depending on a how a model is built—all NESTML models will support axonal delays because we will simply change the templates. But ideally, we would not pull in all the existing ArchivingNode functionality when we do that.

JanVogelsang:
I applied Charl's suggestion in the original ax-delay PR and now close this PR as it does not make sense to add `pre_run_hook` to all models anymore.

PR #3947, opened 2026-09-08
Unify the nest:: namespacing in the entire codebase

JanVogelsang:
Replaces all occurrences of `nest::` by a `namespace nest` section covering the whole file.
The main reason for this PR is to reduce the changeset of the #2989 ax-delay PR.

### DIFF
--- a/models/ac_generator.cpp
+++ b/models/ac_generator.cpp
@@ -35,6 +35,7 @@
 #include "nest_impl.h"
 #include "universal_data_logger_impl.h"
 
+
 namespace nest
 {
 void
@@ -51,13 +52,12 @@ RecordablesMap< ac_generator >::create()
 {
   insert_( names::I, &ac_generator::get_I_ );
 }
-}
 
 /* ----------------------------------------------------------------
  * Default constructors defining default parameters and state
  * ---------------------------------------------------------------- */
 
-nest::ac_generator::Parameters_::Parameters_()
+ac_generator::Parameters_::Parameters_()
   : amp_( 0.0 )      // pA
   , offset_( 0.0 )   // pA
   , freq_( 0.0 )     // Hz
@@ -65,7 +65,7 @@ nest::ac_generator::Parameters_::Parameters_()
 {
 }
 
-nest::ac_generator::Parameters_::Parameters_( const Parameters_& p )
+ac_generator::Parameters_::Parameters_( const Parameters_& p )
   : amp_( p.amp_ )
   , offset_( p.offset_ )
   , freq_( p.freq_ )
@@ -73,8 +73,8 @@ nest::ac_generator::Parameters_::Parameters_( const Parameters_& p )
 {
 }
 
-nest::ac_generator::Parameters_&
-nest::ac_generator::Parameters_::operator=( const Parameters_& p )
+ac_generator::Parameters_&
+ac_generator::Parameters_::operator=( const Parameters_& p )
 {
   if ( this == &p )
   {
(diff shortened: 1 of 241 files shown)


heplesser:
I think Claude has some good arguments against this change, especially 2 & 3:

Good context: [models/iaf_bw_2001.cpp](vscode-webview://1dg09ublobsut8i0l5e0srog5qiramu5ohkbjl8ahpe42j2tv47c/models/iaf_bw_2001.cpp) actually shows both styles already — a namespace nest { } block wrapping only a free function and an explicit template specialization (lines 48–72), while every actual member-function definition afterward uses the fully-qualified form at file scope. That's a useful concrete data point for the tradeoffs below.

No semantic difference for ordinary members. For an out-of-line member function defined with a qualified-id (void N::C::f() {}), unqualified name lookup inside the body already searches the class scope and then N's enclosing namespace scopes — exactly as if you'd textually reopened namespace N { }. So this is a style choice, not a correctness one, for plain member functions. (The one place it can matter is explicit template specializations, like the RecordablesMap<iaf_bw_2001>::create() in that file — some forms of those are easier or required to write inside a reopened namespace block, which is presumably why that block exists there at all.)

Arguments against making the switch:

Diff size and blame churn. This would touch essentially every .cpp file in models/ and nestkernel/ — hundreds of files, thousands of lines — for a change with no behavioral effect. That's a lot of git blame noise on lines that didn't conceptually change, and a lot of conflict surface against any other branch (including your own CMake work) touching the same files concurrently.

Definitions stop being self-describing. With the qualified form, any single line, viewed in isolation — grep output, a compiler error, a stack trace, a PR diff with limited context, a random scroll position in a 1000-line model file — tells you exactly what class/namespace it belongs to. Inside a namespace block spanning a whole file, you lose that unless you scroll back to the opening brace. This matters especially in NEST's model files, which tend to be long.

Grep/paste friction. Compiler errors and stack traces print fully-qualified names (nest::iaf_bw_2001::update). With the qualified form you can paste that straight into grep and land on the definition. With the wrapped form, only iaf_bw_2001::update appears in the file — the nest:: prefix isn't there to match against.

Indentation dilemma. Indent the block's contents and you touch every single line (defeats even the goal of a "small" diff); don't indent it (common in LLVM/Google-style code) and you need clang-format's NamespaceIndentation: None configured and reviewers used to that convention.

Files rarely map 1:1 onto "one clean namespace block". As iaf_bw_2001.cpp shows, a single .cpp can mix ordinary members, free functions, and explicit specializations that have their own namespace-scoping needs. You often don't get one tidy wrapper for the whole file — you get several non-contiguous blocks interleaved with qualified-form code anyway, undermining the main visual win.

Existing convention. The codebase is already overwhelmingly qualified-form, with the wrapped form reserved for cases that specifically need it (free functions, specializations). A wholesale switch either happens in one very large PR, or you live with two competing idioms side by side for a long transition — itself a maintenance cost.

In favor, briefly: less visual repetition per line (bigger win the deeper the nesting, e.g. N::M::C::f), and it's the natural place to add new namespace-scope free functions/operators next to the class.

Given NEST's current convention already does this selectively (wrap only when needed, qualify otherwise), I'd lean toward not doing a blanket rewrite — the qualified form's self-describing/grep-friendly property is a genuine, ongoing readability win in a codebase this size, not just inertia.

JanVogelsang:
I personally think the arguments Claude provided are all very weak arguments against this change. So I would argue that those arguments should not be what drives our final decision whether we want this PR or not.

I checked C++ coding conventions and found that there are different opinions.
Google's and Mozilla's coding conventions say that one should prefer the `namespace nest` syntax:
>Wrap an entire source file in a namespace, after its includes, gflags definitions/declarations and forward declarations of classes from other namespaces.

LLVM's and Microsoft's coding styleguides say that one should prefer local namespace qualifiers:
> When providing an out of line implementation of a function in a source file, do not open namespace blocks in the source file. Instead, use namespace qualifiers to help ensure that your definition matches an existing declaration.

LLVM's argument for that is rather weak, though, because one can simply add compiler checks (`-Wmissing-declarations`) to prevent declaration and definition namespace mismatches.

heplesser:
After careful reflection, Claude's arguments carry little weight for us given our generally clean code and file structure. Therefore, the advantages of cleaning up outweigh potential problems.

heplesser (approved):
Thanks for tidying this up!

diesmann (approved):
I agree to the outcome of our discussion at the hackathon

PR #3948, opened 2026-09-18
Update workflow to run with new cmake config requirements

jessica-mitchell:
Need to update workflow for C++ doc build on Githubpages to work with new cmake config


jessica-mitchell:
Merging with no reviewr because it's doc related, tested on my fork and it works; changes minor

